PR #3054, opened 2016-03-02
Renaming clashing services and controllers

### LINKED ISSUE #2921: Generated entity service name could clash with JHipster service name
Moving the dissussion from end of old closed issue #1966 to here.

The problem occurs e.g. when generating an entity called `Language` where a service containing the REST API is created with the name `Language`. This will clash with the internal JHipster service also called `Language` which is used to switch the current language. The problem was introduced after issue #2645 was solved since the reserved keywords was removed.

Solutions:
1. Prefix all the generated internal services with `Jhi` e.g. the `JhiLanguage`
2. Postfix all the generated entity service with `Resource` e.g. `LanguageResource` (as it is named in backend)
3. Postfix all the generated entity service with `Api` e.g. `LanguageApi` 
4. Postfix all the generated entity service with `Rest` e.g. `LanguageRest` 
5. Reintroduce the reserved entity names

@jhipster/developers 
Please upvote and downvote the ones you like or dislike with a comment and add more solution suggestions if you have any.

wmarques:
Fix #2921.
I renamed services that could clash with user-generated services with Jhi prefix. Also renamed Health and Metrics Controllers into HealthCheck and MetricsMonitoring as suggested by @deepu105.

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -105,6 +105,7 @@ module.exports = JhipsterClientGenerator.extend({
         this.searchEngine = this.options['search-engine'];
         this.hibernateCache = this.options['hb-cache'];
         this.jhiPrefix = this.options['jhi-prefix'];
+        this.jhiPrefixCapitalized = _.capitalize(this.jhiPrefix);
         this.testFrameworks = [];
         this.options['protractor'] &&  this.testFrameworks.push('protractor');
         var lastQuestion = configOptions.lastQuestion;
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -87,6 +87,13 @@ module.exports = JhipsterClientGenerator.extend({
             type: String
         });
 
+        // This adds support for a `--jhi-prefix` flag
+        this.option('jhi-prefix', {
+            desc: 'Add prefix before services, controllers and states name',
+            type: String,
+            defaults: 'jhi'
+        });
+
         this.skipServer = configOptions.skipServer || this.config.get('skipServer');
         this.skipUserManagement = configOptions.skipUserManagement ||  this.config.get('skipUserManagement');
         this.authenticationType = this.options['auth'];
@@ -97,6 +104,7 @@ module.exports = JhipsterClientGenerator.extend({
         this.enableSocialSignIn = this.options['social'];
         this.searchEngine = this.options['search-engine'];
         this.hibernateCache = this.options['hb-cache'];
+        this.jhiPrefix = this.options['jhi-prefix'];
         this.testFrameworks = [];
         this.options['protractor'] &&  this.testFrameworks.push('protractor');
         var lastQuestion = configOptions.lastQuestion;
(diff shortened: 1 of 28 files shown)


deepu105:
On second thought if @jdubois is ok with it then we should just rename the controllers and state name to have jhi prefix as well for uniformity

deepu105:
also make the jhi prefix into a variable and use it in templates so that we can add an option `--jhi-prefix` which can be used to override the prefix if someone doesnt want to see jhi everywhere so that they can pass `--jhi-prefix foo` and generated code will have prefix foo instead of jhi. we will set jhi as the default value for the option

deepu105:
another Option is to prepend the angular app name to these instead of jhi as the default

wmarques:
Personally I prefer the jhi-prefix option, the angular app name may be too long and reduce code readability.

generators/client/templates/src/main/webapp/app/account/settings/_settings.controller.js
```diff
@@ -5,9 +5,9 @@
         .module('<%=angularAppName%>')
         .controller('SettingsController', SettingsController);
 
-    SettingsController.$inject = ['Principal', 'Auth'<% if (enableTranslation){ %>, 'Language', '$translate'<% } %>];
+    SettingsController.$inject = ['Principal', 'Auth'<% if (enableTranslation){ %>, '<%=jhiPrefix%>LanguageService', '$translate'<% } %>];
```
deepu105: all this should be `jhiPrefixCapitalized`

generators/client/index.js
```diff
@@ -97,6 +104,7 @@ module.exports = JhipsterClientGenerator.extend({
         this.enableSocialSignIn = this.options['social'];
         this.searchEngine = this.options['search-engine'];
         this.hibernateCache = this.options['hb-cache'];
+        this.jhiPrefix = this.options['jhi-prefix'];
```
deepu105: we need to assign this to to variables
```
this.jhiPrefix = this.options['jhi-prefix'];
this.jhiPrefixCapitalized = _.capitalize(this.jhiPrefix);
```

generators/client/templates/src/main/webapp/app/admin/metrics/_metrics.state.js
```diff
@@ -8,7 +8,7 @@
     stateConfig.$inject = ['$stateProvider'];
 
     function stateConfig($stateProvider) {
-        $stateProvider.state('metrics', {
+        $stateProvider.state('<%=jhiPrefix%>metrics', {
```
deepu105: these are correct

deepu105: on second thought i guess these states should be `<%=jhiPrefix%>-metrics`

wmarques:
Oh yes I forgot that controllers and services names should be capitalized ! It's done

kkorada:
how to pass this flag is it like this ?
yo jhipster --jhi-prefix xxx

wmarques:
@kkorada `yo jhipster:client --jhi-prefix foo`

deepu105:
@wmarques i think we need to add this to the main app generator so that it can be passed to client sub gen and document it as well

kkorada:
yes it would be good to add it in main generator as almost many use main generator.
And could you also please add it to generator:modules so that it will be available for creating new modules
Currently it is not present it in jhipsterVar could you please add it ?

PR #3059, opened 2016-03-02
Fix activate Account never fails

### LINKED ISSUE #3041: Activate Account never fails

Activate account API never fails: Because UserService#activateRegistration returns an Optional, the Wrapper Optional.ofNullable in AccountResource.activateAccount produces false negatives.

already used or unknown activation codes does not raise an error

##### **JHipster Version(s)** 

2.27.0

##### **Reproduce the error** 

Calling /api/activate with random activation key always returns 200 OK

##### **Suggest a Fix** 

Remove Optional.ofNullable wrapper in AccountResource.activateAccount.
Remove always return Optional.empty() in UserService.activateRegistration(..), return result of userRepository.findOneByActivationKey(key) instead.

notflorian:
Fix #3041

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -50,7 +50,7 @@ public class UserService {
 
     public Optional<User> activateRegistration(String key) {
         log.debug("Activating user for activation key {}", key);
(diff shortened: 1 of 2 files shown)


deepu105:
@jdubois are you ok with this? seems ok to me

PR #2970, opened 2016-02-23
Ask for native language and additional languages to install when generating a new jhipster app

andidev:
see #2917

PR #3062, opened 2016-03-02
use native language short name instead of en language

andidev:
this should have been a part of #2970


generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -128,7 +128,7 @@ public User createUser(ManagedUserDTO managedUserDTO) {
         user.setLastName(managedUserDTO.getLastName());
         user.setEmail(managedUserDTO.getEmail());
         if (managedUserDTO.getLangKey() == null) {
-            user.setLangKey("en"); // default language is English
+            user.setLangKey("<%= nativeLanguageShortName %>"); // default language is English
```
vivekmore: Should probably modify the comment to say something like 
// defaults to native language

andidev:
Yes true missed that. I'll fix this in another pr

PR #3065, opened 2016-03-03
Format code - aws folder

vivekmore:
format code as per project's .editorconfig

### DIFF
--- a/generators/aws/lib/s3.js
+++ b/generators/aws/lib/s3.js
@@ -66,7 +66,7 @@ S3.prototype.uploadWar = function uploadWar(params, callback) {
     var buildTool = params.buildTool;
     var buildFolder;
 
-    if(buildTool === 'gradle') {
+    if (buildTool === 'gradle') {
         buildFolder = 'build/libs/'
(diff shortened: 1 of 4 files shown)


generators/aws/lib/s3.js
```diff
@@ -107,8 +107,8 @@ var findWarFilename = function findWarFilename(buildFolder, callback) {
             error(err, callback);
         }
         files.filter(function (file) {
-            return file.substr(-FILE_EXTENSION.length) === FILE_EXTENSION;
-        })
+                return file.substr(-FILE_EXTENSION.length) === FILE_EXTENSION;
```
deepu105: is this correct?

deepu105: the original indent seems correct to me

deepu105: im correcting this

vivekmore: Good catch !
Yes the original looks correct.
I don't know why Idea would format it that way.
I'll see if that is a bug in Idea.

PR #3070, opened 2016-03-03
Generate front-end code on Gateway from microservice entity

### LINKED ISSUE #2987: [Entities] Generate front-end code on the gateway, based on a microservice
This is a JHipster 3.0 only feature, using microservices.

Here is the use case:

- The user generates a "Foo" entity on microservice "app1". This only generates the back-end code.
- Now the user wants to generate the corresponding front-end code on the gateway, that connects to the back-end code

This is in fact quite easy:

- We must re-use the `Foo.json` file from the "app1" microservice, and run it again on the gateway with the --skip-server flag

Now there are a few issues, running `yo jhipster:entity Foo --skip-server` on the gateway:

- How do we move around that `Foo.json` file, from the "app1" microservice? My proposal is to add a new question, asking where the file is (as the user may have cloned the "app1" repository anywhere on his filesystem). This question would be triggered by a new flag, for example (as you might want to create a "normal" entity on the gateway)
- How do we map the AngularJS service to the REST back-end? By default, it will be mapped on `app1/api/Foos` but we should ask for that in a new question (and in the gateway we don't know the `app1` prefix anyway)
- Currently running the above command fails -> it looks like the entity sub-generator isn't finished yet.

wmarques:
Fix #2987.
I worked with @PierreBesson on this. We updated the entity generator by following this use case : 

1. Create an entity in a microservice app with the `--skip-client` flag. We added a "microserviceName" field in the entity JSON. This is only added to the entities generated in a microservice. We do this because we need this information to prefix the service name to the URL in angular services.
So the gateway's client entity knows to request the microservice at `/app1/api/foos` and not at `/api/foos`.

2. Create an entity in the gateway with` --skip-server` flag.

3. You will be asked if you to generate the entity from a microservice. You will have to put the path of the microservice folder (eg :` /home/ippon/microservice`) and it will copy and paste Foo.JSON from the microservice to the `.jhipster` folder of the gateway.

4. Front end is generated in gateway. Everything works, even elasticsearch. Amazing !

NB : Gateway and microservices must have the same secret in application.yml config file.

As we added a field in JSON, the JSON files from jhipster-UML won't work if you try to import it. 
Maybe we can do a fallback and ask the user the microservice name if this field is not present in the JSON ? Or we could ask the jhipster uml guys to support this feature.

Also, should we put the `skip-client` and `skip-server` flags activated by default respectively in microservice and gateway ?

Finally, elasticsearch has to be enabled in the microservice to be available in the gateway. Should we add a "searchable" field in the JSON when generating the entity and then display or not the search bar in the gateway ?

Here is an example JSON entity file. Only a microserviceName field has been added.
```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "name",
            "fieldType": "String"
        }
    ],
    "changelogDate": "20160303141317",
    "dto": "no",
    "service": "no",
    "entityTableName": "foo",
    "pagination": "no",
    "microserviceName": "app1"
}
```

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -118,7 +118,6 @@ module.exports = EntityGenerator.extend({
             this.languages = this.config.get('languages');
             this.buildTool = this.config.get('buildTool');
             this.testFrameworks = this.config.get('testFrameworks');
-            this.pathConfigs = this._globalConfig.get('pathConfigs');
             // backward compatibility on testing frameworks
             if (this.testFrameworks == null) {
                 this.testFrameworks = ['gatling'];
@@ -135,18 +134,6 @@ module.exports = EntityGenerator.extend({
                 this.log(chalk.green('\nFound the ' + this.filename + ' configuration file, entity can be automatically generated!\n'));
                 this.useConfigurationFile = true;
                 this.fromPath = this.filename;
-            }
-            else {
-                if (this.pathConfigs) {
-                    var pathConfig = this.pathConfigs[this.entityNameCapitalized];
-                    if (pathConfig) {
-                        this.log(chalk.green('\nFound the ' + this.filename + ' configuration file in remote microservice ' +  pathConfig.appName));
-                        this.log(chalk.green('\nUsing configuration file ' + pathConfig.entityPath));
-                        this.useConfigurationFile = true;
-                        this.fromPath = pathConfig.entityPath;
-                        this.microserviceName = pathConfig.appName;
-                    }
-                }
             }
         },
 
@@ -1637,15 +1624,8 @@ module.exports = EntityGenerator.extend({
             if (_.isUndefined(this.microservicePath)) {
                 return;
             }
-
-            if (_.isUndefined(this.pathConfigs)) this.pathConfigs = {};
-
-            this.pathConfigs[this.entityNameCapitalized] = {
-                appName: this.microserviceName,
-                entityPath: this.microservicePath + '/' + this.jhipsterConfigDirectory + '/' + this.entityNameCapitalized + '.json'
-            };
-
-            this._globalConfig.set('pathConfigs', this.pathConfigs);
+            
+            this.copy(this.microservicePath + '/' + this.jhipsterConfigDirectory + '/' + this.entityNameCapitalized + '.json', this.destinationPath(this.jhipsterConfigDirectory + '/' + this.entityNameCapitalized + '.json'));
         },
 
         writeEnumFiles: function() {


deepu105:
Nice work guys.
Ok few points on this.

1. If the app type is microservice we should `--skip-client` by default and user shouldnt have to pass the flag as microservice wouldn't have any client config setup anyway

2. For the new field JHUML should ideally support it. Either you can PR there or ask someone on that team for it.

3. For gateway `--skip-server` shouldn't be default. What if I want to generate a normal entity on the gateway itself. Ideal flow would be when someone generates an entity from gateway we first scan local `.jhipster` folder for json if not we promt them to ask if its a microservice located in a relative/absolute path. Once they answer we can fetch json from there and generate just the client. if they answer **no** we proceed to generate it as as normal.

4. We might have to maintain the path for each entity generated as such on gateway for regeneration without asking path again. May be we can have a file in the `.jhipster` folder of gateway where we store a key : object pair like below for each such entity. The file could be named **jhipster-remote-entities.json** or something like that. When there is an entry here we skip asking path again.
```
entityName: {
    appName : fooApp,
    entityPath : 'home/some/path'
}
```

5. The `searchable` or `searchEnabled` key on emtity sounds ok to me. Again this also needs to be supported by JHUML 

Ill further test and review the PR latter today.

generators/entity/index.js
```diff
@@ -83,20 +83,39 @@ module.exports = EntityGenerator.extend({
             defaults: ''
         });
 
+        // This adds support for a `--skip-client` flag
```
deepu105: I would prefer this to be read from yo-rc file rather than passing this for each entity as this should be aligned to the app structure itself

generators/entity/index.js
```diff
@@ -108,7 +127,6 @@ module.exports = EntityGenerator.extend({
             this.languages = this.config.get('languages');
             this.buildTool = this.config.get('buildTool');
             this.testFrameworks = this.config.get('testFrameworks');
-            this.skipClient = this.config.get('skipClient');
```
deepu105: This shouldn't be removed

generators/entity/index.js
```diff
@@ -1000,12 +1020,53 @@ module.exports = EntityGenerator.extend({
...
+                            return this.filename + ' not found in ' + input + '/';
+                        }
+                    }.bind(this)
+                }
+            ];
+
+            this.prompt(prompts, function(props) {
+                this.log(chalk.green('\nFound the ' + this.filename + ' configuration file, entity can be automatically generated!\n'));
+                this.microservicePath = props.microservicePath;
+                this.importMicroserviceJson = true;
```
deepu105: Instead of having a new boolen why not set useConfigurationFile true here?

generators/entity/index.js
```diff
@@ -1207,6 +1268,34 @@ module.exports = EntityGenerator.extend({
 
     configuring : {
 
+        loadEntityJson: function() {
+            if(!this.importMicroserviceJson) {
+                return;
+            }
+
+            this.fromPath = this.microservicePath + '/' + this.filename;
```
deepu105: Instead of repeating all this again we can ask the question before reading the json file in init stage and just pass the path to be read which would be far simpler

generators/entity/index.js
wmarques: You are right the boolean is useless. But in some case I have to differentiate if I'm importing a file of if I have to write a new JSON, for example before executing                 
`this.template(this.fromPath, this.rootDir + '/' + this.filename, this,{});`

But I can check if I'm importing a file by checking if `microservicePath` is defined via `!_.isUndefined(this.microservice)`

wmarques: But if we decide to use a `remote-entities` JSON file should I still copy the json in `.jhipster` folder ?

jdubois:
I'm OK with all comments from @deepu105 except the 4)

-> this path is specific for each user, as people might clone each repo anywhere on their filesystem. So I don't think we should store the path, at least not in the project's folder. We might store it in the user's `.yo-rc-global.json`, like for the default package name: what do you think of this?

wmarques:
I've done it as @deepu105 said for the moment, I tested it and works fine even if you update a JSON in microservice, you can reload it in gateway ! I'm going to add the `searchEngine` field in entity JSON now. Then I'll look for adding entities' paths in global config.

deepu105:
@jdubois the path will be added in the gateway for that user alone right? may be we can gitignore it as well so that it stays local it will be simpler to manage then storing in global config

jdubois:
@deepu105 yes it is specific to the user, they can clone their app anywhere on their filesystem.
We already use the `.yo-rc-global.json` file, and it's done for this kind of use-case, I think it's a better solution than having our own config with a gitignore

deepu105:
But the global file is manged by yeoman i dont know if there is an yeoman
api for us to access that we might have to do it manually

jdubois:
I haven't used it yet, but yes the idea is that there is a Yeoman API for this, and that's why it's less work for us

generators/entity/index.js
```diff
@@ -1582,7 +1689,7 @@ module.exports = EntityGenerator.extend({
             this.template(SERVER_MAIN_SRC_DIR + 'package/repository/_EntityRepository.java',
             SERVER_MAIN_SRC_DIR + this.packageFolder + '/repository/' + this.entityClass + 'Repository.java', this, {});
 
-            if (this.searchEngine == 'elasticsearch') {
+            if (this.searchEngines[this.entityName] == 'elasticsearch') {
```
deepu105: i dont understand why this needs to be done this way?

generators/entity/index.js
```diff
@@ -83,13 +82,23 @@ module.exports = EntityGenerator.extend({
...
+            type: Boolean,
+            defaults: false
+        });
+
         this.regenerate = this.options['regenerate'];
         this.entityTableName = this.options['table-name'] || this.name;
+        this.entityName = _s.capitalize(this.name);
         this.entityTableName = _s.underscored(this.entityTableName).toLowerCase();
         this.entityAngularJSSuffix = this.options['angular-suffix'];
+        this.skipServer = this.options['skip-server'];
```
deepu105: I think `skipServer` should be decided based on below and not just based on the passed option
1. If the apps `yo-rc` has `skipServer` property that should be used
2. If its a gateway from where entity is being generated and if the `useMicroserviceJson` option is true we should set `skipServer` to true
3.if its a gateway and if `useMicroserviceJson` is not true we should use what is passed in the option `--skip-server`
4. if its a normal app and if apps yo-rc doesnt have `skipserver` set we should use whats passed in `--skip-server` option

generators/entity/index.js
```diff
@@ -97,22 +106,32 @@ module.exports = EntityGenerator.extend({
...
             this.packageName = this.config.get('packageName');
+            this.applicationType = this.config.get('applicationType');
             this.packageFolder = this.config.get('packageFolder');
             this.authenticationType = this.config.get('authenticationType');
             this.hibernateCache = this.config.get('hibernateCache');
             this.databaseType = this.config.get('databaseType');
             this.prodDatabaseType = this.config.get('prodDatabaseType');
             this.searchEngine = this.config.get('searchEngine');
+            this.searchEngines = [];
+            this.searchEngines[this.entityName] = this.searchEngine;
```
deepu105: why is this required? I thought this will be read from the entity json file

generators/entity/index.js
```diff
@@ -97,22 +106,32 @@ module.exports = EntityGenerator.extend({
...
             this.languages = this.config.get('languages');
             this.buildTool = this.config.get('buildTool');
             this.testFrameworks = this.config.get('testFrameworks');
-            this.skipClient = this.config.get('skipClient');
             // backward compatibility on testing frameworks
             if (this.testFrameworks == null) {
                 this.testFrameworks = ['gatling'];
             }
+
+            if(this.applicationType == 'microservice') {
```
deepu105: this can be simplified to 
```
this.skipClient = this.applicationType === 'microservice' || this.config.get('skipClient');
```

generators/entity/index.js
```diff
@@ -120,13 +139,21 @@ module.exports = EntityGenerator.extend({
...
-                    this.log(chalk.red('\nThe configuration file could not be read!\n'));
-                    return;
-                }
                 this.useConfigurationFile = true;
+                this.fromPath = this.filename;
+            }
+            else {
+                var filePath = this.jhipsterConfigDirectory + '/jhipster-remote-entities.json';
+                var allData = this.fs.readJSON(filePath);
+                for(var key in allData) {
```
deepu105: I think no need to iterate since its an object you can do
```
var pathConfig = allData[this.entityName];
if(pathConfig ) {

this.fromPath = pathConfig.entityPath;
do your stuff

}
```

generators/entity/index.js
```diff
@@ -120,13 +139,21 @@ module.exports = EntityGenerator.extend({
...
-                } catch (err) {
-                    this.log(chalk.red('\nThe configuration file could not be read!\n'));
-                    return;
-                }
                 this.useConfigurationFile = true;
+                this.fromPath = this.filename;
+            }
+            else {
+                var filePath = this.jhipsterConfigDirectory + '/jhipster-remote-entities.json';
+                var allData = this.fs.readJSON(filePath);
```
deepu105: may be we can have it called `pathConfigs`and the enity key value as `pathConfig` or something similar

generators/entity/index.js
```diff
@@ -180,29 +207,43 @@ module.exports = EntityGenerator.extend({
...
+        this.relationships && this.relationships.forEach(function (rel) {
+            relNameChoices.push({name: rel.relationshipName + ':' + rel.relationshipType, value: rel.relationshipName + ':' + rel.relationshipType});
+        }, this);
+        if (this.fileData.angularJSSuffix !== undefined){
+            this.entityAngularJSSuffix = this.fileData.angularJSSuffix;
+        }
+        if (this.applicationType == 'gateway'){
+            this.microserviceName = this.fileData.microserviceName;
+            this.searchEngines = [];
+            this.searchEngines[this.entityName] = this.fileData.searchEngine;
```
deepu105: again I dont understand whats happening here am i missing something?

generators/entity/index.js
```diff
@@ -1000,6 +1040,50 @@ module.exports = EntityGenerator.extend({
...
+                    message: 'Do you want to generate this entity from an existing microservice?',
+                    default: true
+                },
+                {
+                    when: function(response) {
+                        return response.useMicroserviceJson == true;
+                    },
+                    type: 'input',
+                    name: 'microservicePath',
+                    message: 'Enter the absolute path to the microservice root directory:',
```
deepu105: may be we can support relative path as well with respect to the gateways root folder

generators/entity/index.js
```diff
@@ -1000,6 +1040,50 @@ module.exports = EntityGenerator.extend({
...
+                        }
+                    }.bind(this)
+                }
+            ];
+
+            this.prompt(prompts, function(props) {
+                if(props.useMicroserviceJson) {
+                    this.log(chalk.green('\nFound the ' + this.filename + ' configuration file, entity can be automatically generated!\n'));
+                    this.microservicePath = props.microservicePath;
+                    this.fromPath = this.microservicePath + '/' + this.jhipsterConfigDirectory + '/' + _s.capitalize(this.name) + '.json';
```
deepu105: you can use `this.entityName` instead of `_s.capitalize(this.name)` rite?

deepu105:
@jdubois whatever @wmarques has done already works somewhat like you expect if im not wrong and it would work regardless of where the user clones the repos I guess. can you try out tell us if you are expecting something different?

jdubois:
Thanks @deepu105 for reviewing the code, as I didn't have the time to!!
This looks like what I had in mind, now I will need to test this -> this is major for our 3.0 release so your help is warmly welcome!!

deepu105:
I have already tested this, once my comments are corrected I will test
again and merge if ok

Thanks & Regards,
Deepu

generators/entity/index.js
wmarques: As I load the JSON in two cases (local json or external json from a microservice), I've made this helper function to prevent code repetition.

generators/entity/index.js
wmarques: The problem I got by using relative paths is that as we are in the jhipster generator folder, the relative path is from the jhipster generator folder and not from the gateway folder... That's why I precisely ask to the user the absolute path.
If you know a way to handle relative path from where the generator is launched let me know.

wmarques:
Thanks alot for reviewing the code @deepu105 ! I tried to take your advices in count, I tested it and works fine.

generators/entity/index.js
```diff
@@ -83,13 +82,27 @@ module.exports = EntityGenerator.extend({
...
+            defaults: false
+        });
+
         this.regenerate = this.options['regenerate'];
         this.entityTableName = this.options['table-name'] || this.name;
+        this.entityName = _s.capitalize(this.name);
         this.entityTableName = _s.underscored(this.entityTableName).toLowerCase();
         this.entityAngularJSSuffix = this.options['angular-suffix'];
-        if (this.entityAngularJSSuffix && !this.entityAngularJSSuffix.startsWith('-')) {
+        if(!_.isUndefined(this.config.get('skipServer'))) {
```
deepu105: sorry to be a pain in the ass, but we could simplify this to below which will do same as your code
```
this.skipServer = this.config.get('skipServer') || this.options['skip-server'];
```

generators/entity/index.js
deepu105: you can get path of [current directory ](http://yeoman.io/authoring/file-system.html)from where the generator is called this way `this.destinationRoot()` which is what you have assigned to `this.rootDir` so if I invoke `yo jhipster:entity Foo` from `/usr/someDir/myGatewayDir` `this.rootDir` should be the same and we can use a relative path to this folder as well like `../myMicroserviceOne` which is in `/usr/someDir/myMicroserviceOne`

generators/entity/index.js
wmarques: You're not I need to learn this shorten syntax, I was used to use `if/else` at school.

generators/entity/index.js
```diff
@@ -1499,7 +1576,7 @@ module.exports = EntityGenerator.extend({
             var entityNameSpinalCased = _s.dasherize(_s.decapitalize(this.name));
             var entityNamePluralizedAndSpinalCased = _s.dasherize(_s.decapitalize(pluralize(this.name)));
 
-            this.entityClass = _s.capitalize(this.name);
+            this.entityClass = this.entityName;
```
deepu105: the naming `this.entityName` is little confusing may be it should be `this.entityNameCapitalized` what do you think?

generators/entity/index.js
```diff
@@ -1664,7 +1760,8 @@ module.exports = EntityGenerator.extend({
         },
 
         writeClientTestFiles: function () {
-            if (this.skipClient) return;
+            if(this.skipClient) return;
```
deepu105: this `if (this.skipClient) return;` is correct formatting with space between things, I guess you have removed the space in many places can you plz check your diff

generators/entity/index.js
deepu105: @wmarques I was talking about these lines, ofcourse i understood the rest, but couldnt figure out what the below were doing
```
this.searchEngines = [];
this.searchEngines[this.entityName] = this.fileData.searchEngine;
```

deepu105:
@wmarques sorry for giving more comments, but hope it helps you

wmarques:
It's fine I'm also here to learn ! Agree with you on the entityNameCapitalized, it's more consistent with the other variables.

deepu105:
@jdubois Im quite ok with this, plz check and merge if it works as per your requirement

generators/entity/index.js
```diff
@@ -180,38 +201,48 @@ module.exports = EntityGenerator.extend({
...
+        }, this);
+        this.relationships && this.relationships.forEach(function (rel) {
+            relNameChoices.push({name: rel.relationshipName + ':' + rel.relationshipType, value: rel.relationshipName + ':' + rel.relationshipType});
+        }, this);
+        if (this.fileData.angularJSSuffix !== undefined){
+            this.entityAngularJSSuffix = this.fileData.angularJSSuffix;
+        }
+        if (this.applicationType == 'gateway'){
+            this.microserviceName = this.fileData.microserviceName;
+            this.searchEngine = this.fileData.searchEngine;
```
deepu105: this needs to be `this.searchEngine = this.fileData.searchEngine || this.searchEngine` to handle all cases

deepu105:
Ok I actually have just one more comment

wmarques:
It's done thanks for your feebacks !

jdubois:
I'm currently testing this.
Why do we have a `.jhipster/jhipster-remote-entities.json` in the gateway?

- I wanted to store the path in the user's `.yo-rc-global.json` file as the path to the entity is specific per user
- As long as we have the path stored in `.yo-rc-global.json` I don't think this `.jhipster/jhipster-remote-entities.json` is useful anymore. At least it should not have the user's path hardcoded inside it.

wmarques:
I switched into the `.yo-rc json`. I couldn't find out how to write in the `.yo-rc.global.json` but I think it's no longer used by Yeoman because in my `.yo-rc.global.json` I don't have any entry for `generator-jhipster`. I think yeoman is reading the` .yo-rc.json` now, check [this link](http://yeoman.io/authoring/storage.html).

wmarques:
I added `store: true` when we ask the user the path of the microservice so next time it will be this path by default.

deepu105:
the `.yo-rc.global.json` is used by yeoman when you use the `store   : true` property in a prompt there is no API for us to access it with yeoman. So if you want the config to be saved there we need to read it using normal `fs` methods and that means we need to handle for all OS and user directory structure as well.

To me if we `.gitignore` the `.jhipster/jhipster-remote-entities.json` it is as good as saving the values in the `.yo-rc.global.json` file and this way we dont put anything outside of the app directory as well. SO each user can have whatever path they want as this file will not be checked in

deepu105:
@wmarques using store means the path will be overwritten for each entity and its not ideal when you will have multiple microservice apps, means if you try to regenerate it will still promt you and in most case you have to type the path again

wmarques:
Oh yes I just saw that now I have an entry in global file ! Personally I prefer using `gitignore` with `jhipster-remote-entities.json`, it avoids writing in the global yeoman file using `fs`.

For the microserve path, I just stored it to have a default value, if the user wants to import many entities from the same microservice it could be useful. But if you don't like it I can remove this !

By the way, I can handle relative path so I'm going to implement this using `path` node library.

deepu105:
Im ok with store, lets wait for what Julien thinks

Thanks & Regards,
Deepu

wmarques:
I'm now always saving relative paths, it should resolve the problem we got by saving absolute paths.

deepu105:
@wmarques I dont think the pathConfig need to be saved in yo-rc file as it would be checked in to git. May be lets discuss and finalize before you make any further changes

wmarques:
Yes but if it's relative path it will probably be common between users, that's why I put it in yo-rc.

deepu105:
need not be common, depends on where I downloaded the folder

Thanks & Regards,
Deepu

deepu105:
Any way dont keep changing lets discuss and finalize first so that its less
work for you

Thanks & Regards,
Deepu

deepu105:
did some digging the `yo-rc.global.json` is created [here ](https://github.com/yeoman/generator/blob/master/lib/base.js#L158)

so try doing the below and see if it works, its bit hacky way of doing this, another way would be to use this lib https://github.com/yeoman/configstore

### to set 
```
this._globalConfig.set('pathConfigs', this.pathConfigs);
```

### to fetch
```
this.pathConfigs = this._globalConfig.get('pathConfigs');
```

wmarques:
So I save absolute path there right ?

deepu105:
yes coz its for the particular user only so absolute should be fine

Thanks & Regards,
Deepu

deepu105:
@jdubois are you ok with the above solution?

jdubois:
I haven't looked at the API in details, but yes that's the idea: we store the user-specific configuration (like the absolute path) in the global configuration file.
That's what we do for the package name, have a look at the `store: true` parameter in the server/index.js file, in the prompts.

deepu105:
Store is bit different as its for per question type and we cant store
multiple entity config with that anyway if my solution above works then it
will be similar to what store does. Lets see

wmarques:
So I try using globalconfig functions ?

deepu105:
Ya try

wmarques:
It's done !

deepu105:
Is it working?

wmarques:
Yes paths are correctly saved and it works

jdubois:
I've tested it and indeed nearly everything works fine, excepted 2 things:

- I don't have anymore the json file copied in `.jhipster` in the gateway -> we need it to be able to re-create the entity, and I thought you also needed it to store the Elasticsearch option
- This isn't working out-of-the-box as it doesn't have the path to the microservice from the gateway. This is going to [change soon](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jhipster-dev/krdM25gyj6k/BB8lSDGUAgAJ) so be careful

wmarques:
After a discussion with @jdubois, we decided to only copy the JSON from the microservice into the gateway `.jhipster` folder. This is to handle most of the use cases. For example, when we were using the pathConfigs, the `--with-entities` option wasn't working anymore.

deepu105:
But a problem with this would be that if someone edits an entity in the
microservice they need to copy it manually to gateway. Anyway lets go ahead
with this first we can do improvements later

deepu105:
older version of this seems to be merged already into master https://github.com/jhipster/generator-jhipster/commit/20b533ffeccf2169ebc3ad3a4603d99c921d6039

was it by accident? in either case @wmarques you need to rebase this to current master

jdubois:
yes @deepu105 this is probably my fault. I'm merging this one, and testing everything.

PR #3071, opened 2016-03-03
Access control filter for the gateway

### LINKED ISSUE #2954: The Gateway should only forward the "/api" and Swagger URLs
Currently the gateway forwards applications automatically, which is the default with Zuul.

However, for security reasons, we shouldn't forward everything, like all the "/info" URLs.

So the gateway should be reconfigured to only forward URLs to:

- "/api": where all the application's APIs are
- "/v2/api-docs": where Swagger is -> this should only work in "dev" mode, not "prod" mode (still for security reasons)

PierreBesson:
Fix #2954 

It can be configured by changing properties under jhipster.gateway.authorized-microservices-endpoints.
By default, all microservices endpoints are available (spring cloud default configuration) but a sample recommended config is provided for dev and prod.

### DIFF
--- a/generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
+++ b/generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
@@ -1,0 +1,92 @@
+package <%=packageName%>.gateway.accesscontrol;
+
+import <%=packageName%>.config.JHipsterProperties;
+import com.netflix.zuul.ZuulFilter;
+import com.netflix.zuul.context.RequestContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cloud.netflix.zuul.filters.ProxyRouteLocator;
+import org.springframework.http.HttpStatus;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Zuul filter for restricting access to backend micro-services endpoints
+ */
+public class AccessControlFilter extends ZuulFilter{
+
+    private final Logger log = LoggerFactory.getLogger(AccessControlFilter.class);
+
+    @Inject
+    private ProxyRouteLocator routeLocator;
+
+    @Inject
+    private JHipsterProperties jHipsterProperties;
+
+    @Override
+    public String filterType() {
+        return "pre";
+    }
+
+    @Override
+    public int filterOrder() {
+        return 0;
+    }
+
+    @Override
+    // Filter requests on endpoints that are not in the list of authorized microservices endpoints
+    public boolean shouldFilter() {
+        Map<String, String> routes = routeLocator.getRoutes();
+        String requestUri = RequestContext.getCurrentContext().getRequest().getRequestURI();
+
+        // If the request Uri does not start with the path of the authorized endpoints, we block the request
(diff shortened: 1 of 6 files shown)


generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.http.HttpStatus;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Zuul filter for restricting access to backend micro-services endpoints
+ */
+public class AccessControlFilter extends ZuulFilter{
```
jdubois: Formatting: you miss a space at `ZuulFilter {`

generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return true;
+    }
+
+    private boolean isAuthorizedRequest(String serviceUrl, String serviceName, String requestUri) {
+        Map<String, List<String>> authorizedMicroservicesEndpoints = jHipsterProperties.getGateway().getAuthorizedMicroservicesEndpoints();
+
+        // If the authorized endpoints list was left empty for this route, all access are allowed
+        if(authorizedMicroservicesEndpoints.get(serviceName) == null){
+            log.debug("Access Control: allowing access for: " + requestUri + ", as no access control policy has been set up for service: " + serviceName);
```
jdubois: For the log.debug you need to use the `log.debug("my text {}", value);` pattern as it's more efficient

generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
```diff
@@ -0,0 +1,92 @@
...
+            return true;
+        }
+        else {
+            List<String> authorizedEndpoints = authorizedMicroservicesEndpoints.get(serviceName);
+
+            // Go over the authorized endpoints to control that the request URI matches it
+            for (String endpoint : authorizedEndpoints) {
+                String gatewayEndpoint = serviceUrl.substring(0, serviceUrl.length() - 3) + endpoint;
+                if (requestUri.startsWith(gatewayEndpoint)) {
+                    log.debug("Access Control: allowing access for: " + requestUri + ", as it matches the following authorized microservice endpoint: " + gatewayEndpoint);
```
jdubois: same here for logging

generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.cloud.netflix.zuul.filters.ProxyRouteLocator;
+import org.springframework.http.HttpStatus;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Zuul filter for restricting access to backend micro-services endpoints
```
jdubois: you need a "." at the end to have correct Javadoc

deepu105:
LGTM +1

jdubois:
OK let's do this :-)

PR #3073, opened 2016-03-03
Reset language cookie if language is not installed

andidev:
Language cookie should be cleared if the language is not installed in the running jhipster application.

This is to prevent a cryptic error when running two different jhipster applications with different language support localhost.

e.g. see atomfreds comment https://github.com/jhipster/generator-jhipster/pull/2970#issuecomment-191408930

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -328,6 +328,7 @@ module.exports = JhipsterClientGenerator.extend({
             if (this.enableTranslation) {
                 this.template(ANGULAR_DIR + 'blocks/handlers/_translation.handler.js', ANGULAR_DIR + 'blocks/handlers/translation.handler.js', this, {});
                 this.template(ANGULAR_DIR + 'blocks/config/_translation.config.js', ANGULAR_DIR + 'blocks/config/translation.config.js', this, {});
+                this.template(ANGULAR_DIR + 'blocks/config/_translation.storage.js', ANGULAR_DIR + 'blocks/config/translation.storage.js', this, {});
             }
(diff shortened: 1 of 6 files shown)


generators/client/index.js
```diff
@@ -328,6 +328,7 @@ module.exports = JhipsterClientGenerator.extend({
             if (this.enableTranslation) {
                 this.template(ANGULAR_DIR + 'blocks/handlers/_translation.handler.js', ANGULAR_DIR + 'blocks/handlers/translation.handler.js', this, {});
                 this.template(ANGULAR_DIR + 'blocks/config/_translation.config.js', ANGULAR_DIR + 'blocks/config/translation.config.js', this, {});
+                this.template(ANGULAR_DIR + 'blocks/config/_translation.storage.js', ANGULAR_DIR + 'blocks/config/translation.storage.js', this, {});
```
deepu105: I think the file name should be `_translation-storage.provider.js`

andidev:
Aha okey. I'll fix that right away. 
I was not sure about the location of where I put it. Any remarks on that?

generators/client/templates/src/main/webapp/app/blocks/config/_translation.storage.js
```diff
@@ -0,0 +1,28 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('<%=angularAppName%>')
+        .factory('translationStorage', translationStorage);
```
andidev: should this be called translationStorageProvider as well?

deepu105: Yes I guess, coz this acts as a service provider  for translate config rite

andidev:
fixed

PR #3029, opened 2016-02-29
Containerize microservice and gateway applications

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/2988
All changes:
- all docker-compose files use v2 version, so you need docker 1.9+ and compose 1.6+
- gateway and microservice can now be easily containerized:
    - `mvn package docker:build -Pprod -DskipTests=true`
    - `./gradlew -Pprod bootRepackage -x test buildDocker`
- gateway and microservice apps generate 4 new files. As always, these files are here for samples, but can be used directly:
    - src/main/docker/app.dev.yml : start microservice/gateway app with registry in dev profile
    - src/main/docker/app.noregistry.prod.yml : start microservice/gateway app with registry in prod profile
    - src/main/docker/app.dev.yml : start microservice/gateway app without registry in dev profile
    - src/main/docker/app.noregistry.prod.yml : start microservice/gateway app without registry in prod profile

Issue #2988, opened 2016-02-25
[Feature] Microservice workflow with docker


PierreBesson:
For microservices, we would like to be able to have a workflow that uses docker to run the registry, the gateway, microservices, databases, monitoring tools.
This could take the form of one or several docker-compose file.

What needs to be done:
- Every app should have a Dockerfile that is used to create a docker image of this app.
- The app must be composed with it's dependencies: for example a database
- All apps are composed with the registry and  additional tools (ELK, Graphite/Grafana).

What is already decided:
- Our images will be based on Alpine as having small images is important.

Interrogation:
- How to build the docker image ? Should we build the jar with Maven/Gradle (using a docker plugin) and then copy it to a container ? Should we use a separate container to build the apps from sources ? The latter way would allow to (re)build the image from docker only if the sources have been updated.

I started experimenting for some time and have managed to start everything at once using the following  setup:

**docker-compose.yml**
```
jhipster-registry:
  container_name: jhipster-registry
  build: ./jhipster-registry/
  environment:
    - "SPRING_PROFILES=dev,native"
  ports:
    - "8761:8761"
  volumes:
    - ./central-config:/central-config
gateway:
  container_name: gateway
  build: ./gateway/
  ports:
    - "8080:8080"
  links:
    - jhipster-registry:registry
app1:
  container_name: app1
  build: ./app1/
  ports:
    - "8081:8081"
  links:
    - jhipster-registry:registry
app2:
  container_name: app2
  build: ./app2/
  ports:
    - "8082:8081"
  links:
    - jhipster-registry:registry
```

**Microservice and gateway Dockerfile**
```
FROM anapsix/alpine-java
ADD target/*.war app.war
RUN bash -c 'touch /app.war'
EXPOSE 8080
VOLUME /tmp
CMD java -jar -Deureka.client.serviceUrl.defaultZone=http://registry:8761/eureka/ -Dspring.cloud.config.uri=http://registry:8761/config app.war
```

If you are already familiar with docker and are interested in this, please give your feedback and contribute ideas so that we can release a first POC. It would be nice to have this ready for the JH3 release in two weeks.

PierreBesson:
Giving a link to @pascalgrimaud's amazing [JHipster Docker Module](https://github.com/pascalgrimaud/generator-jhipster-docker) as we can build upon what he has done.

pascalgrimaud:
[...]

PR #3085, opened 2016-03-05
Migrate docker base image to alpine for gateway and microservices

pascalgrimaud:
- related to https://github.com/jhipster/generator-jhipster/issues/2988
- migrate base image from `java:openjdk-8u72-jre` to `alpine:3.3`
- remove noregistry docker-compose files, after discussion here https://github.com/jhipster/generator-jhipster/pull/3029
- start microservice with registry:
    - `docker-compose -f src/main/docker/registy.yml -f src/main.docker/app.prod.yml up`
- start microservice (without registry because it's already started):
    - `docker-compose -f src/main/docker/app.prod.yml up`

Everything works fine, with scaling microservices.
jhipster-microservices


generators/server/templates/src/main/docker/_app.dev.yml
```diff
@@ -72,5 +65,5 @@ services:
         <%_ } _%>
         <%_ if (applicationType == 'microservice') { _%>
         expose:
-            - "8080"
+            - "8081"
```
jd-carroll: Please forgive me, but I am confused by this... 
Why are you moving to port 8081, did I miss part of a thread somewhere?  And if so, why aren't you moving the [prod port](https://github.com/jhipster/generator-jhipster/pull/3085/files#diff-5e46f70be7afa0bd3a118c08d94e06b5L75) as well?

pascalgrimaud: Thanks for reviewing my docker PR, @jd-carroll, it's really appreciated!

About 8081 in dev profile : [application-dev.yml](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/_application-dev.yml#L139)
We should probably do the same thing for prod profile, which is currently 8080 : [application-prod.yml](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/_application-prod.yml#L136-L137)

deepu105: Can you plz add a todo item in our documentation ticket for these as well

pascalgrimaud: It is already the point 5 and 6, about docker updates and microservices

PR #3091, opened 2016-03-07
Adding i18n Greek language

PeteShag:
Included an  i18n translation of JHipster to Greek.


generators/generator-base.js
```diff
@@ -292,6 +292,7 @@ Generator.prototype.getAllSupportedLanguageOptions = function () {
         {name: 'French', value: 'fr'},
         {name: 'Galician', value: 'gl'},
         {name: 'German', value: 'de'},
+		{name: 'Greek', value: 'el'},
```
deepu105: formatting is wrong

test/test-languages.js
```diff
@@ -19,6 +19,7 @@ describe('JHipster generator languages', function () {
         {name: 'Dutch', value: 'nl'},
         {name: 'Galician', value: 'gl'},
         {name: 'German', value: 'de'},
+		{name: 'Greek', value: 'el'},
```
deepu105: formatting indent is wrong

jdubois:
Thanks but don't forget to update also the website, if it's not documented people will not know it's there

andidev:
Could you also translate the entity? You can find entity translations over here :
https://github.com/jhipster/generator-jhipster/tree/master/generators/entity/templates/src/main/webapp/i18n

PR #3096, opened 2016-03-07
Greek i18n translation for entity

PeteShag:
Greek i18n translation for entity in reference to comment jhipster/generator-jhipster#3091


deepu105:
Thank you

Issue #3082, opened 2016-03-04
[FEATURE] Add centralized monitoring for microservices


PierreBesson:
Following this [thread](https://groups.google.com/forum/#!topic/jhipster-dev/v3VphFzI9jA) on the JHipster mailing list, I successfully managed to integrate metrics into the application logs and have them shipped to ELK. So I would like to propose a PR for this if you are OK with it.

I used dropwizard's Slf4jReporter on the advice of @cbornet and then used the logstash socket forwarder I recently added to microservice and gateway apps to ship logs to ELK. Finally I parse those logs with logstash to extract metrics data ([logstash config file](https://github.com/PierreBesson/jhipster-microservices-docker/blob/f0404f2e0502a2031b85c0095966b90a03844f37/log-monitoring/log-config/logstash.conf)).
Here is what I get in Kibana:

![screenshot from 2016-03-04 16 35 34](https://cloud.githubusercontent.com/assets/513471/13531574/8afda0cc-e227-11e5-9955-f5ebbf3aaa7a.png)

![screenshot from 2016-03-04 16 37 38](https://cloud.githubusercontent.com/assets/513471/13531576/8dff23e0-e227-11e5-83b6-7b3c7b359afd.png)

As you can see we have all the fields we need in Kibana to start building dashboards. So I will need some help here. It would be amazing if someone experienced with Kibana could come forward.

Then I was thinking that we could have those dashboards exported as json files and imported on startup of the ELK docker-compose with [this script](https://github.com/elastic/beats-dashboards/blob/master/load.sh).

The next thing to do would be alerting. I really want to try to use [Yelp/elastalert](https://github.com/Yelp/elastalert) here but I'm open to other solutions.

gmarziou:
It looks really promising

PierreBesson:
It should be discussed if my approach of using logstash-logback-encoder is better than [elastic/elasticsearch-metrics-reporter-java](https://github.com/elastic/elasticsearch-metrics-reporter-java). Both collect the same dropwizard metrics. My solution has the advantage that it is using the same socket to send logs and metrics. Also it is already enriching the logs to add the appname, instanceId, host and port.

jdubois:
- I'd rather have only one ELK for everything, and not an ELK stack and a Graphite/Grafana. This looks easier to set up and use.
- As you have a solution for alerting, then ELK doesn't have any big drawbacks
[...]

PR #3098, opened 2016-03-07
Metrics forwarding to logs

PierreBesson:
Enable metrics forwarding to ELK. Linked to #3082

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-prod.yml
+++ b/generators/server/templates/src/main/resources/config/_application-prod.yml
@@ -181,6 +181,9 @@ jhipster:
             host: localhost
             port: 2003
             prefix: <%= baseName %>
+        logs: # report metrics in the logs
+            enabled: false
+            reportFrequency: 60 # in seconds
     <%_ if (applicationType == 'microservice' || applicationType == 'gateway') { _%>
     logging:
         logstash: # Forward logs to logstash over a socket, used by LoggingConfiguration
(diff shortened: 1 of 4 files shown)


generators/server/templates/src/main/resources/config/_application-prod.yml
```diff
@@ -181,6 +181,9 @@ jhipster:
             host: localhost
             port: 2003
             prefix: <%= baseName %>
+        logs: # report metrics in the logs
+            enabled: false
+            reportFrequency: 60 # in seconds
```
deepu105: does doing this every minute have any performance impact? especially in prod when enabled?

PierreBesson: Actually I don't really know what default frequency should be used. The spark reporter also report logs every minute and does not even provide a way to configure the frequency in the app properties.
Anyway, It should not have too much performance impact as it is only sending a few dozens of logs every minute minute.

deepu105: Ok was just curious :)

Thanks & Regards,
Deepu

pascalgrimaud:
@PierreBesson : what do you think about adding your `elk.yml` too ?
I thought it was already done to generator-jhipster project but I don't see this file 

It will be easy to test, right ?
- change config false to true
- start elk with docker-compose
- start microservice

PierreBesson:
Currently, the elk.yml is in a [PR on the registry](https://github.com/jhipster/jhipster-registry/pull/15).
I didn't want to clutter every apps with this so I didn't put this in the main generator.
Actually, the new idea is to write a Docker Compose generator to start all the micro-services at once (see #2988). I will put the elk.yml there.

pascalgrimaud:
Ok I see ! thanks Pierre

cbornet:
The elk.yml would be needed for monolith apps I guess

Issue #2785, opened 2016-02-04
update documentation for next


deepu105:
we need to update docs for next major. plz add on if i missed anything @jhipster/developers 

- [x] 1. new angular folder structure & john papa style guideline
- [x] 2. ~~removal of fast~~ - PR submitted
- [x] 3. skip-client option - PR submitted
- [x] 4. grunt to gulp
- [x] 5. docker updates
- [x] 6. microservices - PR submitted, needs more contents
- [x] 7. naming convention updates for angular side and api endpoints, pluralization
- [x] 8. ~~entity update option~~ -PR submitted
- [x] 9. new advanced server & client sub gen 
- [x] 10. update contribution.md to advice new structure changes if required
- [x] 11. we also need to update the presentation 
- [x] 12. ~~update the videos~~ - Done
- [x] 13. We need to update the setting-CI page with gulp and remove the "Configuring Javascript Tests" which is now included by default.
- [x] 14. new --table-name flag in entity sub gen
- [x] 15. --with-entities to regen all entities
- [x] 16. --no-notification flag for gulp
- [x] 17 --angular-suffix option to add entity suffix
- [x] 18 switch to eslint instead of jshint
- [x] 19 --skip-user-management flag
- [x] 20 Internationalization - support for setting native language and additional languages
- [x] 21 have links to the script-base's jhipsterFunc from the creating-a-module page to have a quick view of the protoype and the function doc.
- [x] 22 changes in the generated war structure, no more single war for dev and prod
- [x] 23 Inject of js files with gulp serve and build
- [x] 24 microservice use cases, diagrams and advice
- [x] 25 update customizing_bootstrap.md with the new Gulp configuration
- [x] 26 update the technology stack to add the Netflix stack
- [x] 27 import-jdl sub generator
- [x] 28 docker compose sub generator
- [x] 29 gradle support in cloudfoundry sub gen
- [x] 30 new profile switches for swagger, liquibase
- [x] 31 jhipster registry (spring cloud config part) and jhipster console.

anything else?

cbornet:
--with-entities to regen all entities

jdubois:
Great job Deepu!
I will also create a shared Google doc so we can prepare a clean release
notes document.
Btw the one from the 2.0 release got vandalized, so next time it won't be
opened to comments.
Le 4 févr. 2016 6:45 AM, "Christophe Bornet" <notifications@github.com> a
écrit :

deepu105:
awesome.

[...]

PR #3101, opened 2016-03-07
Add an architecture diagram to README.md of generated gateway

gmarziou:
Add a micro services architecture diagram to README.md of generated gateway

I'll add some description later on.

Ref #2785


deepu105:
+1000₩

PR #3104, opened 2016-03-08
Auto inject js path inside index.html

### LINKED ISSUE #3103: Add Gulp Inject to automatically inject js in index.html
I have implemented gulp-inject with gulp-angular-file to inject automatically (in a good order) new js files inside index.html and it's work well.

I will make a PR, but before, i want to know if it's a good idea to remove the functionality that add js inside index.html and let the gulp task to do the job?

moifort:
* Add gulp inject + angular-file sort
* Add inject task associated with install and build task
* Remove js file path of the generator

Close #3103

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -282,7 +282,7 @@ module.exports = JhipsterClientGenerator.extend({
 
         composeLanguages: function () {
             if (configOptions.skipI18nQuestion) return;
-            
+
             this.composeLanguagesSub(this, configOptions, 'client');
         }
     },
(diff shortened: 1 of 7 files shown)


generators/client/index.js
```diff
@@ -541,168 +541,9 @@ module.exports = JhipsterClientGenerator.extend({
...
-            if (this.applicationType == 'gateway') {
-                appScripts = appScripts.concat([
-                    'app/admin/gateway/gateway.state.js',
-                    'app/admin/gateway/gateway.controller.js',
-                    'app/admin/gateway/gateway.routes.service.js'
-                ]);
-            }
-
-            indexFile = html.appendScripts(indexFile, 'app/app.js', appScripts, {});
+            indexFile = html.append(indexFile, 'body', '\n<!-- build:js app/app.js  -->\n<!-- inject:js -->\n<!-- endinject -->\n<!-- endbuild -->\n');
```
deepu105: we could add this placeholder in the index.html template directly and avoid touching the file here

generators/client/templates/gulpfile.js
```diff
@@ -30,6 +30,8 @@ var gulp = require('gulp'),
     serve = require('./gulp/serve'),
     util = require('./gulp/utils'),
     gulpIf = require('gulp-if'),
+    inject = require('gulp-inject'),
+    angularFilesort = require('gulp-angular-filesort'),
```
deepu105: I dont think `gulp-angular-filesort` is required for our use case coz we use the same module name for all files and since all the files needs to be loaded for app to work. angular does all the injections at run time so the file order actually doesn't matter

generators/client/templates/gulpfile.js
```diff
@@ -281,15 +283,21 @@ gulp.task('watch', function () {
...
 
+gulp.task('inject', function() {
+    return gulp.src(config.app + 'index.html')
+        .pipe(inject(gulp.src(config.app + 'app/**/*.js').pipe(angularFilesort()), {relative: true}))
+        .pipe(gulp.dest(config.app));
+});
+
 gulp.task('build', function (cb) {
-    runSequence('clean', 'copy', 'wiredep:app', 'ngconstant:prod', 'eslint', 'usemin', cb);
+    runSequence('clean', 'copy', 'wiredep:app', 'ngconstant:prod', 'inject', 'eslint', 'usemin', cb);
```
deepu105: we can add it to watch also so that if you add a file with serve its automatically added to index

deepu105:
You need to remove the manual injections in entity sub gen as well. I think those are causing double injection in files which is failing protractor tests

generators/client/index.js
moifort: :+1: smart move

generators/client/templates/gulpfile.js
moifort: Sadly no, you need to add it, because gulp put `<script src="app/app.constants.js"></script>` before `<script src="app/app.module.js"></script>` and is not working.

generators/client/templates/gulpfile.js
moifort: :+1:

moifort:
:+1:  I think it before to go sleep!

generators/client/templates/gulpfile.js
deepu105: ok

moifort:
Working on why I break the test

deepu105:
Im merging this manually I fixed the tests

moifort:
Thanks you @deepu105!

andidev:
The order of the injected files seems change each time running `gulp inject`. Is this suppose to be this way?

moifort:
I didn't notice, for the first inject yes but, after not in my case. I will test and see If I have a solution.

andidev:
Would also be nice to inject sass files into main.scss. I'll do a PR for that

moifort:
@andidev I just check I dont have the problem, how you reproduce it?

andidev:
Every time I run gulp inject the inject order seems to be a little bit different. Only a couple of files seems to be moved around.

PR #3111, opened 2016-03-08
added injection of jhipsterproperties to CustomPersistentRememberMeServices

alex-charos:
If the jhipsterproperties class gets refactored (i.e. name change), then the CustomPersistentRememberMeServices would not work as it was looking for the properties in a kind of hard-coded way. 

By injecting the JHIpster properties to the constructor method, we have a cleaner way of creating the CustomPersistentRememberMeServices instance without hard-coded properties.


jdubois:
You are correct!
And it's not the only `env.getProperty` still being around in the code, I need to check why they haven't been removed (and this is probably an error).

jdubois:
The build is failing for your PR, as you have a missing import

alex-charos:
Oops! My bad. Will fix this and also have a look to replace any other "env.getProperty"s I find!

jdubois:
Great, thanks!
Btw, there are some of those in the _Application.java file: don't change those ones, it's in a static method so that's easier to use that way.

PR #3112, opened 2016-03-08
Gulp improvements

### LINKED ISSUE #3093: Fonts and Images are not revisioned during gulp build
With our grunt build we used to revision the images and fonts as well but with our current gulp build it doesnt seem to be the case.

Does the images and fonts need to be versioned? They would change very rarely in most of the cases

deepu105:
1. switch to useref and remove usemin. this gives us more control is more gulp like
2. add html template cache for production build, means no html partials in prod and better performance
3. revision images like in our old grunt build
4. source maps without mangling
5. moved configs to its own file
6. moved build task to its own file

~~I havent revisioned bootstarp fonts, Im not sure if they should be revisioned~~
Edit: fonts are revisioned as well thanks to commits by @erikkemperman 
fix #3093


deepu105:
@wmarques @erikkemperman can you check and let me know if this is good?

wmarques:
That seems good for me :+1:

erikkemperman:
I will take a closer look when I get home, but just curious: why not revision the fonts?

I can't think of a case where it might hurt, and without it there will remain unrevisioned material in `content`.

Which means we can't crank up the Expires header safely, which we could otherwise basically set to "never".

Am I missing something?

erikkemperman:
So I just tried this, works like a charm!

I also tried with passing the fonts through gulp-rev, and this also works. That means it is safe to crank up the Expires header.

I made a "meta-PR" on your branch, that way if you like these changes you can easily make them a part of your PR to the main repo.

deepu105:
Since no one has any objections Ill merge this once travis passes

erikkemperman:
@deepu105 Were my commits causing problems after all? I saw you merged them into your `useref` branch but don't see them here, or in your branch... So it looks like you rebased them out again and force-pushed the branch?

deepu105:
@erikkemperman Oops yes Ill merge them again

erikkemperman:
Thanks, and great job -- this will all mean many fewer requests!

One thing I thought later, but might be quite complicated to achieve, would be to put admin-only stuff in separate files, so mere mortals don't have to download all of those... Thoughts?

deepu105:
it would be good but will complicate the build a lot so may be at a later stage

deepu105:
More important thing I would like to do is improve the gulp build time. earlier usemin used to take nearly 99% of the build time but there was no way to speed that up. Since we are using useref now may be we can try to do more stuff in parallel and reduce some seconds. I guess the uglify task takes the most time

erikkemperman:
Uglify vendor and main js in parallel would make a decent improvement, sure!

deepu105:
when you find time could you take a look at gulp/build.js and see if there is any improvement that can be done, I couldn't find a way to run those in parallel.

If possible we should run these in parallel
1. vendor js concat, uglify & rev 
2. main js ngAnnotate, concat, uglify and rev
3. vendor css prefix, concat, minify and rev
4. main css prefix, concat, minify and rev
5. html minify

erikkemperman:
@deepu105 All right, but still quite busy with that other project :-(  In general though, gulp runs everything in parallel which isn't explicitly serial (dependency or runSequence).

deepu105:
But pipes inside a task are run in series not parellel

Thanks & Regards,
Deepu

erikkemperman:
Haha, yeah of course. I meant that splitting tasks would be a natural way to achieve parallelism. Now that we have reusable (lazy) pipes, this should be pretty easy I expect.

PR #3113, opened 2016-03-08
Gateway filter to rewrite microservices Swagger URL path

### LINKED ISSUE #3046: Swagger PathMapping in micro service should not be hard coded
JH 3.0 master: when generating a micro service, the path mapping in SwaggerConfiguration is hard coded to the name of the service, it should be configurable.

Here the name of the service is "tester".

~~~java
        Docket docket = new Docket(DocumentationType.SWAGGER_2)
            .apiInfo(apiInfo)
            .pathMapping("/tester")
~~~

I propose to add it to JHipsterProperties.Swagger so that it can be configured in application properties and managed in config server.

~~~yaml
jhipster:
    swagger:
        pathMapping: /${spring.application.name}
~~~

Or it could be set to `/${eureka.instance.appname}`

So the code to use it would be:

~~~java
        Docket docket = new Docket(DocumentationType.SWAGGER_2)
            .apiInfo(apiInfo)
            .pathMapping(jHipsterProperties.getSwagger().getPathMapping())
~~~

PierreBesson:
Fix #3046 
Don't rely anymore on a hard-coded path-mapping to access the micro-services API throught Swagger UI. This is achieved by rewriting the "Base URL" (see the bottom part of Swagger UI) when the JSON swagger-docs is passing through the gateway.

screenshot from 2016-03-08 18 22 44

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -1063,6 +1063,11 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_TEST_RES_DIR + 'config/_application.yml', SERVER_TEST_RES_DIR + 'config/application.yml', this, {});
             this.template(SERVER_TEST_RES_DIR + '_logback-test.xml', SERVER_TEST_RES_DIR + 'logback-test.xml', this, {});
 
+            // Create Gateway tests files
+            if (this.applicationType == "gateway"){
+                this.template(SERVER_TEST_SRC_DIR + 'package/gateway/responserewriting/_SwaggerBasePathRewritingFilterTest.java', testDir + 'gateway/responserewriting/SwaggerBasePathRewritingFilterTest.java', this, {});
+            }
+
             if (this.hibernateCache == "ehcache") {
                 this.template(SERVER_TEST_RES_DIR + '_ehcache.xml', SERVER_TEST_RES_DIR + 'ehcache.xml', this, {});
             }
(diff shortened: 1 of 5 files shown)


PierreBesson:
@gmarziou Thanks for your comments, it is a good way to learn for me. I will amend my PR to include your ideas. About the status code that should be returned, I don't know what could be put except 200 OK.

PierreBesson:
I tried to serialize the response to Springfox's ApiListing but it doesn't work. I think that it would need some tweaking or adding some JSON annotations to the class which is not possible because it belongs to the springfox package.

gmarziou:
OK forget ApiListing then, it's not worth the pain.

PierreBesson:
OK, after running your tests I understand that the status code you mean is in case of an error.

gmarziou:
Exactly, in unit test, I was not sure how to handle this becauseI don't know exactly what the Zuul proxy expects, so this is more an integration test in some ways.

PierreBesson:
@gmarziou I've implemented your advice, however run_on_error_404() is still not passing. I tried many things but I didn't succeed in writing the appropriate code to deal with an empty response.

gmarziou:
Well the unit test is not that important as long as you have tested the real thing: the filter against a service down.

gmarziou:
So to clarify further, I was thinking to this scenario:
- registry and gateway are started
- service starts and registers
- tests  on gateway that swagger-ui works for proxified service
- stop service
- tests that error is correctly handled on gateway swagger

PierreBesson:
I tested it. It returns the following error on swagger-ui when I select the swagger doc of a micro-service that has just been stopped.

`500 : {"timestamp":"2016-03-09T13:12:00.410+0000","status":500,"error":"Internal Server Error","exception":"com.netflix.zuul.exception.ZuulException","message":"Forwarding error"} http://localhost:8080/micro/v2/api-docs`

In Chrome dev tools I have `Status Code:500 Internal Server Error`

Then after waiting some time, when the Zuul route is removed from the gateway, the error becomes:
`Can't read swagger JSON from http://localhost:8080/micro/v2/api-docs` and `Status Code:404 Not Found`.

I guess this is the behavior I want to have, so I will remove the run_on_error_404 unit test.
However I think writing tests for our filters is a good idea. I will try to write some more tests in the future.

gmarziou:
Fine for me.

PierreBesson:
Build number 4 failing doesn't seem to be my fault.

moifort:
@PierreBesson update your branch with the last master. I think @deepu105 fix it.

PierreBesson:
Thanks, just done that.

Issue #2848, opened 2016-02-10
Social Login Support for stateless Authentication


BhawaniSingh:
Enabling social login support for oAuth and x-auth should be done

deepu105:
@moifort

moifort:
@deepu105  @jdubois I was thinking if it will not better to migrate all the social authentification to a module and remove it from the jhipster app?

cicoub13:
:+1: for the request and for the module

jdubois:
I'm happy to have this in the main generator, I guess a lot of people want this, and it doesn't look annoying to maintain (never had any issue for that)

deepu105:
I'm fine either way :)

Thanks & Regards,
Deepu

BhawaniSingh:
So anyone planning to work on this?

moifort:
Normally me, but I do not have much time right now.

amitjindal:
@moifort Could you please guide us if we were to start a JWT project and add some classes to support Social auth, how would we go about that?
Or any other alternate way to do this in a JWT project.

Thanks!

amitjindal:
Here's what I did:

I created a JWT project. Then manually edited the .yo-rc.json and set:
`"enableSocialSignIn": true,`

After that when I regenerated, I am able to get the register with Google+ Working. However the login is not working (expected). The social login is not getting registered in Spring security. I think I need to make some changes in SecurityConfiguration.java to get this to work (possibly).

jdubois:
Great news @amitjindal -> can you share work with @moifort on this?

amitjindal:
@jdubois Actually it did not work. I tried a lot of ways. I can share what I did but the problem seems to be that I am not finding where to set the JWT token when social authentication succeeds. Because from what I could understand the authentication is some sort of form based and its returning RedirectViews. I cannot figure out where is the response object that I can set the JWT to get this working.

I am on my way home. I will type up what I have tried so far. Most likely its very easy but I am just not familiar enough with this.

mraible:
This could be similar to the problem I had with Ionic 2. To fix it, I had
to add the token to the response body.

amitjindal:
@mraible Actually I am trying to figure out how to get a response object :-(
I am probably missing something very basic that I don't yet understand.

mraible:
In the current implementation of JWT on master, there is no response body.

amitjindal:
So here's what I have done so far:

[...]

PR #3116, opened 2016-03-09
Add social login support for JWT authenfication

moifort:
* Add new question to enable social login

How I implement, step by step:
1. Authenticate to social login external server (i.e. Google) 
2. Authenticate to the JHipster server
3. JHipster server send request to angular with a cookie containing the jwt token
4. Angular controller store the token and valid the session

Linked to #2848


deepu105:
Does twitter still works?

moifort:
I didn't check twitter yet.

moifort:
:boom: all the test are failed. Go to sleep...

mraible:
Just curious - why a cookie instead of a header or body? The `UserJWTController` uses both header and body (for Angular 2).

https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_UserJWTController.java#L43

moifort:
I try to make like the normal way, but I think social sign in cannot use the normal way to connect the app and need to have a special login authentication. There are my problems:

1. I try to use `api/authenticate` to authenticate the user, but the problem you need a non encrypted password to authorize the user. In my case, my password is already encoded, because user was already created by social sign in.

2. The social sign in mechanism is redirects urls (I didn't success to manage it with $http), and it's finish (if the social connection was successful) by sending a sign in response to the app, customizable through `CustomSignInAdapter.java`. I can custom my response request with `HttpServletResponse` object, this one can set header but not body. I try to add it in the header, but the problem, is how to get the header with a angular controller through ui-router, I didn't find a way. I have an other solution, catch it with an interceptor, but for me, it's a little to heavy.

My solution was to send a token in a expirable cookie and allow angular controller to use a login with token to authenticate the app. 

I don't think is the best solution, but I didn't find other. Maybe I miss some point or mechanism. So if you have other solution or implementation, tell me!

generators/server/templates/src/main/java/package/security/social/_CustomSignInAdapter.java
```diff
@@ -10,17 +14,57 @@
 import org.springframework.social.connect.Connection;
 import org.springframework.social.connect.web.SignInAdapter;
 import org.springframework.web.context.request.NativeWebRequest;
+<%_ if (authenticationType == 'jwt') { _%>
+import com.mycompany.myapp.security.jwt.TokenProvider;
```
amitjindal: This import is incorrect.

moifort: :+1: well seen!

moifort:
@deepu105 I will need some help, I try to understand the error on Travis. I download the `.yo-rc` who test the social and I didn't reproduce the error, I think I miss something but I have no clue...

deepu105:
Ok ill take a look at the error

generators/server/index.js
```diff
@@ -250,6 +246,25 @@ module.exports = JhipsterServerGenerator.extend({
                 },
                 {
                     when: function (response) {
+                        return applicationType == 'monolith' && (response.authenticationType == 'session' || response.authenticationType == 'jwt');
+                    },
+                    type: 'list',
+                    name: 'enableSocialSignIn',
```
pascalgrimaud: you have to change the current tests 
https://github.com/jhipster/generator-jhipster/blob/master/test/test-creation.js#L676-L692

moifort: :+1: @pascalgrimaud, I will add test for JWT social too!

moifort:
It's seems working (#2848), maybe someone want to make some remarks or improvement? otherwise we can merge it.

deepu105:
Was the two issues reported by amit solved or do we need to track it in a
new issue?

deepu105:
Also a small doubt does it work with non email usernames?

deepu105:
@jdubois can we merge this?

moifort:
Need to check some little things it will be done tonight. 

Thibaut.

moifort:
@deepu105 I just analyze the problem with the user delete. Since we removed the `login` rule on the entity `User`, in user management screen, the delete action on user use the login and spring rest had some problem to parse a query param with an email address format. I dont know about that, but maybe it's better to use id instead of login to update/delete/get an user?

amitjindal:
@moifort I also checked and found that when the DELETE url is:
`http://localhost:8080/api/users/amitjindal@domain.com?cacheBuster=1458063515416`

In the `UserResource::deleteUser(@PathVariable String login)`
here the login is being set to: `amitjindal@domain`

So the delete is not working.


Also in this function we need to delete from `SocialUserConnection`

moifort:
@amitjindal I think I will make a PR to fix that, but before I need to know if we can use `id` instead of `login`.

deepu105:
Since the user entity id would be unique anyway I dont see any problem in
using that for delete and update

moifort:
Thanks @deepu105 I create an issue to discuss about it. We can merge this PR, I will make the fix in another PR.

generators/server/index.js
```diff
@@ -283,6 +279,25 @@ module.exports = JhipsterServerGenerator.extend({
                 },
                 {
                     when: function (response) {
+                        return applicationType == 'monolith' && (response.authenticationType == 'session' || response.authenticationType == 'jwt');
```
deepu105: @moifort why did you add this as a new question, the way it was done earlier seemed better to me as its a type of auth and should be part of the same question

deepu105: Oh ignore me you are enbling it for JWT i forgot

fmatuszewski:
Since microservice gateway has also jwt authentication type is it possible to merge it there?

fmatuszewski:
I checked the source code. It is the same codebase so updating enableSocialSignIn whould do the trick I suppose. So the only thing that is needed to add option in wizard(form).

chegola:
When generate Microservices gateway, there is no option to choose Spring Social but I have manually added enableSocialSignIn: true in yo-rc.json then re-generate. jhipster generate social related files and it works great. Thanks you guys.

gnoopy:
@chegola  May I double check with you?  Recently I've been curious about the social login feature with oauth2 configuration. Since there's no option on the wizard I think social login works only with JWT configuration.

Based on your comment,  you probably use oauth2 to enable microservices. Under this configuration you could manage to use social signin feature and no issues to use it. And this is just possible by modifying yo-rc.json and re-generate even though the wizard doesn't show the social login option with oauth2.  Is this correct on your experience ?

fmatuszewski:
That is correct.
As far as I remember you also need to make some changes in the code
yourserlf. But I tried it on Angular js version.
I didn't try this with Angular 4 version.


Filip Matuszewski
604 59 99 36


2017-08-28 10:05 GMT+02:00 Francis <notifications@github.com>:

gnoopy:
@fmatuszewski  thanks for your kind information. I will try this.

fmatuszewski:
Let me know if you succeeded.

Filip Matuszewski
604 59 99 36


2017-08-29 2:01 GMT+02:00 Francis <notifications@github.com>:

PR #3120, opened 2016-03-09
inject sass into main.scss

andidev:
as we now inject app js files into index.html (PR #3104) we should also inject sass files into main.scss

### DIFF
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -41,7 +41,7 @@ gulp.task('clean', function () {
 });
 
 gulp.task('copy', function () {
-    return es.merge( <% if(enableTranslation) { /* copy i18n folders only if translation is enabled */ %>
+    return es.merge(<% if(enableTranslation) { /* copy i18n folders only if translation is enabled */ %>
         gulp.src(config.app + 'i18n/**')
         .pipe(plumber({errorHandler: handleErrors}))
         .pipe(changed(config.dist + 'i18n/'))
@@ -125,6 +125,21 @@ gulp.task('inject', function () {
         .pipe(inject(gulp.src(config.app + 'app/**/*.js').pipe(angularFilesort()), {relative: true}))
         .pipe(gulp.dest(config.app));
 });
+
+<%_ if(useSass) { _%>
+gulp.task('inject:sass', function () {
+    return gulp.src(config.scss + 'main.scss')
+        .pipe(inject(gulp.src(config.app + 'app/**/*.{scss,sass}').pipe(angularFilesort()), {
+            relative: true,
+            starttag: '// inject:sass',
+            endtag: '// endinject',
+            transform: function (filepath) {
+                return "@import '"+ filepath + "';";
+            }
+        }))
+        .pipe(gulp.dest(config.scss));
+});
+<%_ } _%>
 
 gulp.task('wiredep', ['wiredep:test', 'wiredep:app']);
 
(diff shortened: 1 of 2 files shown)


gmarziou:
Isn't it a bit dangerous as you don't control the order of imports?

andidev:
I think the order will be hierarchical to the folder structure and that should be good enough. But I need to check that. 

Though maybe the sass should be injected at the bottom of the main.scss file. Did not think of that.

generators/client/templates/gulpfile.js
```diff
@@ -126,6 +126,21 @@ gulp.task('inject', function () {
         .pipe(gulp.dest(config.app));
 });
 
+<%_ if(useSass) { _%>
+gulp.task('inject:sass', function () {
+    return gulp.src(config.scss + 'main.scss')
+        .pipe(inject(gulp.src(config.app + 'app/**/*.{scss,sass}').pipe(angularFilesort()), {
```
deepu105: You can use config.sassSrc here

generators/client/templates/gulpfile.js
```diff
@@ -279,11 +294,14 @@ gulp.task('watch', function () {
     gulp.watch(<% if(useSass) { %>config.sassSrc<% } else { %>config.app + 'content/css/**/*.css'<% } %>, ['styles']);
     gulp.watch(config.app + 'content/images/**', ['images']);
     gulp.watch(config.app + 'app/**/*.js', ['inject']);
+    <%_ if(useSass) { _%>
+    gulp.watch(config.app + 'app/**/*.{scss,sass}', ['inject:sass']);
```
deepu105: Here as well

generators/client/templates/gulpfile.js
```diff
@@ -279,11 +294,14 @@ gulp.task('watch', function () {
...
+    <%_ } _%>
     gulp.watch([config.app + '*.html', config.app + 'app/**', config.app + 'i18n/**']).on('change', browserSync.reload);
 });
 
 gulp.task('install', function () {
-    runSequence(['wiredep', 'ngconstant:dev'<% if(useSass) { %>, 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>], 'inject');
+    runSequence(['wiredep', 'ngconstant:dev'<% if(useSass) { %>, 'inject:sass', 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>], 'inject');
```
deepu105: Inject:sass should be before sass task, i mean should be before the square brace start as things inside square brace are run in parellel

generators/client/templates/gulpfile.js
```diff
@@ -291,7 +309,7 @@ gulp.task('serve', function () {
 });
 
 gulp.task('build', ['clean'], function (cb) {
-    runSequence(['copy', 'wiredep:app', 'ngconstant:prod'<% if(enableTranslation) { %>, 'languages'<% } %>], 'inject', 'assets:prod', cb);
+    runSequence(['copy', 'wiredep:app', 'ngconstant:prod'<% if(useSass) { %>, 'inject:sass', 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>], 'inject', 'assets:prod', cb);
```
deepu105: Same here

generators/client/templates/gulpfile.js
andidev: Yes of course. I'll fix this

generators/client/templates/gulpfile.js
andidev: Well maybe both config sass src and this path since if you would like to add component specific css you would most likely keep it along the other component files as John Papa  style guide suggests.

andidev:
I'm closing this one since it is not that obvious how this should work. gmarziou was right that the order is important when injecting sass files from the sass folder depending on how you use sass.

To make this work there is a need for a plugin like gulp-angular-filesort but for sass.

erikkemperman:
@andidev How about [`sass-graph`](https://github.com/xzyfer/sass-graph)? It gives you the dependency graph, and works pretty well. I've used it in the past to help prevent unneeded re-compilation, and conversely to detect needed re-compilation when only a sub-dependency had changed.

andidev:
Do you know how to configure this in gulp? I have no clue since I'm new to gulp.

erikkemperman:
I'd have to look at how gulp-angular-filesort does it, but I guess it would require just a gulp-plugin which buffers all the incoming vinyl objects until the end, and then pushes them through in the right order.

This kind of plugin is usually done with `through2` and might very roughly look something like this, say in `gulp-sass-filesort.js`:

```js
var graph = require('sass-graph');
var through2 = require('through2');

module.exports = function( /* options? */ ) {
  var files = [];

  function add(file, enc, cb) {
    files.push(file);
    cb();
  }

  function flush() {
    // Here the interesting bit: traverse the array of files, and determine an order
    // which meets dependency requirements, which we can get from sass-graph.

    sortedFiles.forEach(function(file) { this.push(file); });
    cb();
  }

  return through2.obj(add, flush);
}
```

Such a plugin would be used exactly as the angular one in the main gulpfile.

Disclaimer: I haven't the time to try this out properly, but I'm pretty sure something like this ought to work :-)

andidev:
Thanks for the example. Maybe I'll look into this later though it requires some time that I don't have right now.

erikkemperman:
@andidev Just wanted to check if you were planning to do anything with this? If not I might have a go at it this weekend.

andidev:
Please do if you want to. I have limited time and knowledge for this plus I'm sick right now.

erikkemperman:
All right, hope you feel better soon!

PR #3123, opened 2016-03-09
add missing sass call in gulp build task

andidev:


generators/client/templates/gulpfile.js
```diff
@@ -291,7 +291,7 @@ gulp.task('serve', function () {
 });
 
 gulp.task('build', ['clean'], function (cb) {
-    runSequence(['copy', 'wiredep:app', 'ngconstant:prod'<% if(enableTranslation) { %>, 'languages'<% } %>], 'inject', 'assets:prod', cb);
+    runSequence(['copy', 'wiredep:app', 'ngconstant:prod'<% if(useSass) { %>, 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>], 'inject', 'assets:prod', cb);
```
deepu105: It wasnt missing its already called by assets:prod task as dependency and having here would make it run twice thats why i didn't put it here

deepu105:
Im reverting this

generators/client/templates/gulpfile.js
deepu105: Btw assets has styles as dependency which has sass as dependency so that its run in order

andidev:
Sorry that's right. My gulp skills lack

Issue #3122, opened 2016-03-09
setDocumentRoot should not be called when running .war with java -jar


wmarques:
##### **Overview of the issue**
When i do `docker-compose -f app.dev.yml up` it throws me an EmbeddedServletContainerException.
StackTrace:
```
java.util.concurrent.ExecutionException: org.apache.catalina.LifecycleException: Failed to start component [StandardEngine[Tomcat].StandardHost[localhost].StandardContext[]]
gateway-dev-app_1 | 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
gateway-dev-app_1 | 	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
gateway-dev-app_1 | 	at org.apache.catalina.core.ContainerBase.startInternal(ContainerBase.java:916)
gateway-dev-app_1 | 	at org.apache.catalina.core.StandardHost.startInternal(StandardHost.java:871)
gateway-dev-app_1 | 	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:147)
gateway-dev-app_1 | 	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1408)
gateway-dev-app_1 | 	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1398)
gateway-dev-app_1 | 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
gateway-dev-app_1 | 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
gateway-dev-app_1 | 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
gateway-dev-app_1 | 	at java.lang.Thread.run(Thread.java:745)
gateway-dev-app_1 | Caused by: org.apache.catalina.LifecycleException: Failed to start component [StandardEngine[Tomcat].StandardHost[localhost].StandardContext[]]
gateway-dev-app_1 | 	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:153)
gateway-dev-app_1 | 	... 6 common frames omitted
gateway-dev-app_1 | Caused by: org.apache.catalina.LifecycleException: Failed to start component [org.apache.catalina.webresources.StandardRoot@67862846]
gateway-dev-app_1 | 	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:153)
[...]

PR #3125, opened 2016-03-10
Only do setDocumentRoot if not running from jar/war

erikkemperman:
This fixes issue [#3122](https://github.com/jhipster/generator-jhipster/issues/3122#issuecomment-194733219).

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
+++ b/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
@@ -116,11 +116,14 @@ public class WebConfigurer implements ServletContextInitializer, EmbeddedServlet
         // CloudFoundry issue, see https://github.com/cloudfoundry/gorouter/issues/64
         mappings.add("json", "text/html;charset=utf-8");
         container.setMimeMappings(mappings);<% if (!skipClient) { %>
-        // Set document root
-        if (env.acceptsProfiles(Constants.SPRING_PROFILE_PRODUCTION)) {
-            container.setDocumentRoot(new File("<%= CLIENT_DIST_DIR %>"));
-        } else if (env.acceptsProfiles(Constants.SPRING_PROFILE_DEVELOPMENT)) {
-            container.setDocumentRoot(new File("<%= CLIENT_MAIN_SRC_DIR %>"));
+
+        // Set document root if we're not running from a jar/war
+        if (getClass().getProtectionDomain().getCodeSource().getLocation().getProtocol().equals("file")) {
+            if (env.acceptsProfiles(Constants.SPRING_PROFILE_PRODUCTION)) {
+                container.setDocumentRoot(new File("<%= CLIENT_DIST_DIR %>"));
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/config/_WebConfigurer.java
```diff
@@ -116,11 +116,14 @@ public void customize(ConfigurableEmbeddedServletContainer container) {
...
         mappings.add("json", "text/html;charset=utf-8");
         container.setMimeMappings(mappings);<% if (!skipClient) { %>
-        // Set document root
-        if (env.acceptsProfiles(Constants.SPRING_PROFILE_PRODUCTION)) {
-            container.setDocumentRoot(new File("<%= CLIENT_DIST_DIR %>"));
-        } else if (env.acceptsProfiles(Constants.SPRING_PROFILE_DEVELOPMENT)) {
-            container.setDocumentRoot(new File("<%= CLIENT_MAIN_SRC_DIR %>"));
+
+        // Set document root if we're not running from a jar/war
+        if (getClass().getProtectionDomain().getCodeSource().getLocation().getProtocol().equals("file")) {
```
notflorian: I think it would be better that the String `"file"` is on the left of the .equals to be complient with the squid rule [S1132 : Strings literals should be placed on the left side when checking for equality](https://dev.eclipse.org/sonar/rules/show/squid:S1132)

erikkemperman: Well sure, if you like that better... The reason for doing that doesn't apply here, though, the `getProtocol` is never going to return `null`.

deepu105:
@jdubois what do you think of this

jdubois:
@deepu105 I have no idea :-)

erikkemperman:
Sorry, maybe I should have provided more context here than just linking to the issue...

There was a problem, due to my refactoring the directory layout, which caused the war file to be non portable. This is because of the `setDocumentRoot` call in WebConfigurer.java, which is needed when running in the IDE, but breaks things when running from the war -- except in the special case where the cwd is the project root when issuing the `java -jar` command, which was what I tested when I did that layout change.

So this PR detects whether we're running from a jar or not, and it calls `setDocumentRoot` only in the latter case.

deepu105:
Ok lets go with this for now untill we have a better solution

erikkemperman:
Thanks!

deepu105:
But we should try to find a better solution for this

Thanks & Regards,
Deepu

deepu105:
would setting the document root to a relative path work?

Thanks & Regards,
Deepu

erikkemperman:
I agree this is not pretty, at all... But I don't really know a better way, just now. A relative path doesn't work -- in fact that is what it is now, and this is precisely what breaks the portability of the generated war file.

I suppose that running the production profile within the IDE (which is the only case the `setDocumentRoot` is actually needed) might be rare. Maybe we don't really have to support that?

deepu105:
@jdubois do you think there is a valid reason to support this, I mean running prod profile from IDE would be only for debug purpose I guess

PR #3131, opened 2016-03-10
Bump version to 2.27.2

MathieuAA:
It may have been forgotten

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "generator-jhipster",
-  "version": "2.27.0",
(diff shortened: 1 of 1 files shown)


jdubois:
No it's good in the v2 branch.

I don't bump version number in the master branch until I do a release: with npm you can't overwrite an old version, so that means I can't do a new release by mistake.
I only did this once, and of course that's the day I made a release by error :-)

But thanks for spotting this !

MathieuAA:
Ahah ok! Quite a good idea

PR #3145, opened 2016-03-13
docker-compose : replace links by external_links

pascalgrimaud:
related to https://github.com/jhipster/generator-jhipster/issues/2988


generators/server/templates/src/main/docker/_app.dev.yml
```diff
@@ -61,9 +31,9 @@ services:
             <%_ } _%>
         <%_ if (applicationType == 'gateway') { _%>
         ports:
-            - "8080:8080"
```
jdubois: Ah, I just did exactly the same thing 2 minutes ago :-) I'm merging this

Issue #3127, opened 2016-03-10
REST Api Call returns 406 instead of 400 for invalid passwords


gbrehmer:
##### **Overview of the issue**

Calling api/account/change_password returns HTTP Status Code 406 in case of invalid password requests (instead of Bad Request 400).

##### **JHipster Version(s)** 

2.27.0

##### **Reproduce the error** 

Calling REST api with an invalid (e.g. too short) password

##### **Suggest a Fix** 

Change Produces-Annotation ContentType to TEXT_PLAIN

cbornet:
Well, if we look at the code we see that we try to send a code HttpStatus.BAD_REQUEST with a text/plain content ("Incorrect password").

gbrehmer:
Not acceptable status code is often bound to Accept/Content-Type Header use cases, e.g. if you set Accept-Header as client to text/plain and backend can only produce application/json Content-Type.

gbrehmer:
btw it's the same for api/register. In case of validation errors, the ExceptionTranslator try to return a ErrorDTO, but register method only defines text/plain as possible content-type. I would say it's more a Spring MVC missing feature/bug, but currently you have to define both content-types (because complex type like ErrorDto can not be converted to text/plain automatically) and to add explicitly Content-Type header for text/plain responses in such a case.

deepu105:
closed via #3158

PR #3158, opened 2016-03-14
correct response types in account resource

atomfrede:
I've implemented the mentioned workaround and checked some other endpoints for missing plain text response type.

update #3127


notflorian:
:+1:

atomfrede:
Oh seems I missed a ``(`` will fix it today.

deepu105:
@atomfrede can we close #3127 or does it need more work?

atomfrede:
From my point of view it can be closed.

Issue #2244, opened 2015-10-29
Add integration tests for elasticsearch


cbornet:
The regression #2241 made me think that we should implement some IT for elasticsearch REST endpoints

geraldhumphries:
Here is a quick implementation of an integration test based on the getAll test:

```
    @Test
    @Transactional
    public void searchAddresses() throws Exception {
        // Clear preexisting index
        elasticsearchTemplate.deleteIndex(Address.class);

        // Initialize the database
        addressRepository.saveAndFlush(address);
        
        // Initialize the index with the test address
        addressSearchRepository.save(address);

        // Search for an address
        restAddressMockMvc.perform(get("/api/_search/addresss/id:" + address.getId() + "?sort=id,desc"))
                .andExpect(status().isOk())
                .andExpect(content().contentType(MediaType.APPLICATION_JSON))
                .andExpect(jsonPath("$.[*].id").value(hasItem(address.getId().intValue())))
                .andExpect(jsonPath("$.[*].streetLine1").value(hasItem(DEFAULT_STREET_LINE1.toString())))
                .andExpect(jsonPath("$.[*].streetLine2").value(hasItem(DEFAULT_STREET_LINE2.toString())))
                .andExpect(jsonPath("$.[*].city").value(hasItem(DEFAULT_CITY.toString())))
                .andExpect(jsonPath("$.[*].provinceState").value(hasItem(DEFAULT_PROVINCE_STATE.toString())))
                .andExpect(jsonPath("$.[*].postalCode").value(hasItem(DEFAULT_POSTAL_CODE.toString())));
    }
```

ElasticsearchTemplate also needs to be injected but nothing else needs to be changed.

Elasticsearch will start embedded in the same folder (by default) as when you're running on develop. The line `elasticsearchTemplate.deleteIndex(Address.class);` will delete all data from the entity's Elasticsearch index. This line is absolutely necessary to make sure the test will output properly and that it won't break on a type mismatch.

If this integration test is run the dev will lose all ES data. This can be hugely annoying. If it somehow runs in prod (super unlikely as the dev would need to configure the test profile's ES instance and their prod profile's ES instance to run in the same folder), that could be disastrous. Both situations can be mitigated with the [Elasticsearch Reindexer module](https://github.com/geraldhumphries/generator-jhipster-elasticsearch-reindexer) if they do happen.

So I think we should also configure ES to run in a different path from dev:

```
src/test/config/application.yml

    data:
        elasticsearch:
[...]

PR #3168, opened 2016-03-15
Add integration tests for elasticsearch #2244

lordlothar99:


generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -4,8 +4,8 @@
 import <%=packageName%>.<%= mainClass %>;
 import <%=packageName%>.domain.<%= entityClass %>;
 import <%=packageName%>.repository.<%= entityClass %>Repository;<% if (service != 'no') { %>
-import <%=packageName%>.service.<%= entityClass %>Service;<% } else { if (searchEngine == 'elasticsearch') { %>
```
deepu105: why did you remove the else statement?  when there is service layer tests should be triggered via that instead of repository

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -195,7 +195,8 @@ public void setup() {
 
     @Before
     public void initTest() {<% if (databaseType == 'mongodb' || databaseType == 'cassandra') { %>
-        <%= entityInstance %>Repository.deleteAll();<% } %>
+        <%= entityInstance %>Repository.deleteAll();<% } if (searchEngine == 'elasticsearch') { %>
+        <%= entityInstance %>SearchRepository.deleteAll();<% } %>
```
deepu105: shouldnt this be handled by the ` <%= entityInstance %>Repository.deleteAll();`

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -231,7 +232,12 @@ public void initTest() {<% if (databaseType == 'mongodb' || databaseType == 'cas
         assertThat(test<%= entityClass %>.is<%=fields[idx].fieldInJavaBeanMethod%>()).isEqualTo(<%='DEFAULT_' + fields[idx].fieldNameUnderscored.toUpperCase()%>);
         <%_ } else { _%>
         assertThat(test<%= entityClass %>.get<%=fields[idx].fieldInJavaBeanMethod%>()).isEqualTo(<%='DEFAULT_' + fields[idx].fieldNameUnderscored.toUpperCase()%>);
-        <%_ }} _%>
+        <%_ }} if (searchEngine == 'elasticsearch') { _%>
+
+        // Validate the <%= entityClass %> in ElasticSearch
+        <%= entityClass %> <%= entityInstance %>Es = <%= entityInstance %>SearchRepository.findOne(test<%= entityClass %>.getId());
```
deepu105: this should be `<%= entityClass %>Service.findOne` when entity has service and else `<%= entityInstance %>SearchRepository.findOne`

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -308,7 +314,8 @@ public void initTest() {<% if (databaseType == 'mongodb' || databaseType == 'cas
     @Transactional<% } %>
     public void update<%= entityClass %>() throws Exception {
         // Initialize the database
-        <%= entityInstance %>Repository.save<% if (databaseType == 'sql') { %>AndFlush<% } %>(<%= entityInstance %>);
+        <%= entityInstance %>Repository.save<% if (databaseType == 'sql') { %>AndFlush<% } %>(<%= entityInstance %>);<% if (searchEngine == 'elasticsearch') { %>
+        <%= entityInstance %>SearchRepository.save(<%= entityInstance %>);<%_ } _%>
```
deepu105: this is required only if entity doesnt have a service else its handled in the service already

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
deepu105: Ok this is needed for non service entity, if entity has service this is handled in its delete method

deepu105:
can you take care of my comments plz. once done ill merge it

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: Direct access to SearchRepository is needed whether service is present or not, cause some new assertions are added in create / update / delete test methods

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: Actually , `EntityService.findOne()` invokes `Entity repository.findOne()`.. While the purpose of this assertion is to read from Elasticsearch. So I need direct usage of `EntitySearchRepository.findOne()`

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: You are right. I just wanted to keep the code similar to the other tests.. But I can change that if you prefer service invocation?

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: yes, we need both `deleteAll()` invocations

lordlothar99:
Hi @deepu105, just let me know what u think about my explanations plz

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
geraldhumphries: What about sending a request to `_search/entities` with the query `"id:" + test<%= entityClass %>.getId()`?

geraldhumphries: Never mind, I see that it is following the same pattern as above.

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
geraldhumphries: If the mapping of an entity changes, trying to insert an object into an old index can cause exceptions (ex. if a field type changes). Deleting the data is not enough, you need to delete the entity's index then recreate it with the new mapping:
```
        elasticsearchTemplate.deleteIndex(<%= entityClass %>.class);
        elasticsearchTemplate.createIndex(<%= entityClass %>.class);
        elasticsearchTemplate.putMapping(<%= entityClass %>.class);
```

This will also delete the data.

lordlothar99: yeah I read that, actually I was wondering we could just use `deleteAll()`, as it's fast, and during the tests mappings won't change .. no?

geraldhumphries: They won't change during the tests, but the ES indexes persist (they're stored on the hard drive) so the mappings may change between test sessions. That's why I was trying to find a solution to delete all indexes before any test is run.

deepu105: Ill look at them today

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
deepu105: ok since you are using `<%= entityInstance %>SearchRepository.deleteAll();` anyway

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
deepu105: Ill let you guys decide this

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
deepu105: ok

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
deepu105: yes since service already does that its a good way to test if service does that so i would prefer to use service when its enabled

deepu105:
Im ok with all except the last save `<%= entityInstance %>SearchRepository.save(<%= entityInstance %>);` stuff I would prefer that via service when its enabled

lordlothar99:
I'm on it

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: @geraldhumphries ok, I understand now that I've read once again your comment. I will try to find a better way to handle that...

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
lordlothar99: ok I will

lordlothar99:
@deepu105 as we said, I replaced `EntityRepository.save()` and `EntitySearchRepository.save()` by `EntityService.save()` when possible
@geraldhumphries I introduced new `EntityIndexInitializer` beans, which reset ElasticSearch indexes at startup only (`@PostConstruct` method), it takes more or less 2s on my (slow) computer. Still `EntityResourceIntTest` `@Before` method invokes `EntitySearchRepository.deleteAll()` which is very fast (<10ms).
Can you please review these changes, and let me know ?
thx !

lordlothar99:
seems like Travis is failing sometimes, but I think that's nothing to do with this PR, no ?

jdubois:
No we are having other build issues at the moment, don't worry!
Le 18 mars 2016 9:11 AM, "François Lecomte" <notifications@github.com> a
écrit :

generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -361,6 +382,35 @@ public void initTest() {<% if (databaseType == 'mongodb' || databaseType == 'cas
 
         // Validate the database is empty
         List<<%= entityClass %>> <%= entityInstancePlural %> = <%= entityInstance %>Repository.findAll();
-        assertThat(<%= entityInstancePlural %>).hasSize(databaseSizeBeforeDelete - 1);
-    }
+        assertThat(<%= entityInstancePlural %>).hasSize(databaseSizeBeforeDelete - 1);<% if (searchEngine == 'elasticsearch') { %>
+
+        // Validate tElasticSearch is empty
+        boolean <%= entityInstance %>ExistsInEs = <%= entityInstance %>SearchRepository.exists(<%= entityInstance %>.getId());
```
cicoub13: When entityInstance is same that entityInstancePlural, there is a problem here (with entity people for example). 

`<%= entityInstance %>.getId()` does not exist because entityInstance was redefined in a `List<<%= entityClass %>>`

deepu105: @lordlothar99 can you check this plz?

cicoub13: Inversing elasticSearch validation and database validation works.
Replcaing `List<<%= entityClass %>> <%= entityInstancePlural %>` by `List<<%= entityClass %>> <%= entityInstancePlural %>Database` works too.

lordlothar99: I'm on it ; do you prefer a new PR, or update this one ?

deepu105: I think you cant update this as its merged you need to do a new PR

lordlothar99: yeah u r right, I pushed but nothing happened ...
I created fixed #3227

Issue #3172, opened 2016-03-15
Jpa EntityListener used to synchronize ElasticSearch with database


lordlothar99:
##### **Overview of the issue** 

[Enhancement] For SQL + ElasticSearch applications, usage of EntityListener

##### **Motivation for or Use Case** 

For applications using ElasticSearch, `EntityResource` / `EntityServiceImpl` call both `EntityRepository.save()` and `EntitySearchRepository.save()` ( resp `update()` / `delete()` ), which is not very elegant ; but the real problem is when an attached entity is fetch from `EntityRepository`, and updated, ElasticSearch is not automatically updated, so user need to be careful, and explicitly call `EntitySearchRepository.save()` method. 

##### **JHipster Version(s)** 

2.27.0

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

```
{
  "generator-jhipster": {
    "applicationType": "monolith",
    "baseName": "samplePsqlEsNoi18n",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "postgresql",
    "prodDatabaseType": "postgresql",
    "searchEngine": "elasticsearch",
    "useSass": true,
    "buildTool": "maven",
    "enableTranslation": false,
    "enableSocialSignIn": false,
    "rememberMeKey": "efaf561e5400af2877060b33324f34288d671c97",
    "testFrameworks": [
      "gatling",
      "protractor"
    ]
  }
}
```

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory** 

Example :
```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "label",
            "fieldType": "String",
            "fieldValidateRules": [
                "required",
                "minlength"
            ],
            "fieldValidateRulesMinlength": "3"
        }
    ],
    "changelogDate": "20150805124936",
    "dto": "no",
    "pagination": "pagination",
    "service": "serviceClass"
}
```

##### **Related issues** 

Desynchronization problems are silent until #2244 is merged 

##### **Suggest a Fix** 

I have the code ready, and will submit a PR soon

deepu105:
closed as PR is closed

PR #3174, opened 2016-03-15
#3172 Jpa EntityListener used to synchronize ElasticSearch with database

lordlothar99:
- ElasticSearchConfiguration moved to dedicated config.elasticsearch
package
- ElasticSearchUpdater entityListener installed on JPA entities
- removed useless explicit EntitySearchRepository.save/update/delete
calls

### DIFF
--- a/generators/entity/templates/src/main/java/package/common/delete_template.ejs
+++ b/generators/entity/templates/src/main/java/package/common/delete_template.ejs
@@ -1,7 +1,7 @@
 <% if (!viaService) {
     if (databaseType == 'sql' || databaseType == 'mongodb') { %>
         <%= entityInstance %>Repository.delete(id);<% } %><% if (databaseType == 'cassandra') { %>
-        <%= entityInstance %>Repository.delete(UUID.fromString(id));<% } %><% if (searchEngine == 'elasticsearch') { %><% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
+        <%= entityInstance %>Repository.delete(UUID.fromString(id));<% } %><% if (searchEngine == 'elasticsearch') { %><% if (databaseType == 'mongodb') { %>
         <%= entityInstance %>SearchRepository.delete(id);<% } %><% if (databaseType == 'cassandra') { %>
         <%= entityInstance %>SearchRepository.delete(UUID.fromString(id));<% } %><% }
 } else { %>
--- a/generators/entity/templates/src/main/java/package/common/save_template.ejs
+++ b/generators/entity/templates/src/main/java/package/common/save_template.ejs
@@ -11,6 +11,6 @@ if (!viaService) {
         <%= entityInstance %> = <%= entityInstance %>Repository.save(<%= entityInstance %>);
         <%= instanceType %> result = <%= entityToDto %>(<%= entityInstance %>);<% } else {
     resultEntity = 'result'; %>
-        <%= entityClass %> result = <%= entityInstance %>Repository.save(<%= entityInstance %>);<% } if (searchEngine == 'elasticsearch') { %>
+        <%= entityClass %> result = <%= entityInstance %>Repository.save(<%= entityInstance %>);<% } if (searchEngine == 'elasticsearch' && databaseType != 'sql') { %>
         <%= entityInstance %>SearchRepository.save(<%= resultEntity %>);<% }} else { %>
         <%= instanceType %> result = <%= entityInstance %>Service.save(<%= instanceName %>);<% } %>
--- a/generators/entity/templates/src/main/java/package/domain/_Entity.java
+++ b/generators/entity/templates/src/main/java/package/domain/_Entity.java
@@ -15,6 +15,10 @@ import org.springframework.data.mongodb.core.mapping.Document;
 import org.springframework.data.mongodb.core.mapping.Field;<% } %><% if (searchEngine == 'elasticsearch') { %>
 import org.springframework.data.elasticsearch.annotations.Document;<% } %>
 <% if (databaseType == 'sql') { %>
+<%_ if (searchEngine == 'elasticsearch') { _%>
+import <%=packageName%>.config.elasticsearch.ElasticSearchUpdater;
+
+<%_ } _%>
 import javax.persistence.*;<% } %><% if (validation) { %>
 import javax.validation.constraints.*;<% } %>
 import java.io.Serializable;<% if (fieldsContainBigDecimal == true) { %>
@@ -43,7 +47,8 @@ import <%=packageName%>.domain.enumeration.<%= element %>;<% }); %>
 @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType == 'mongodb') { %>
 @Document(collection = "<%= entityTableName %>")<% } %><% if (databaseType == 'cassandra') { %>
 @Table(name = "<%= entityInstance %>")<% } %><% if (searchEngine == 'elasticsearch') { %>
-@Document(indexName = "<%= entityInstance.toLowerCase() %>")<% } %>
+@Document(indexName = "<%= entityInstance.toLowerCase() %>")<% if (databaseType == 'sql') { %>
+@EntityListeners(ElasticSearchUpdater.class)<% }} %>
 public class <%= entityClass %> implements Serializable {
 
     private static final long serialVersionUID = 1L;
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -1069,7 +1069,11 @@ module.exports = JhipsterServerGenerator.extend({
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/liquibase/_package-info.java', javaDir + 'config/liquibase/package-info.java', this, {});
             }
             if (this.searchEngine == 'elasticsearch') {
-                this.template(SERVER_MAIN_SRC_DIR + 'package/config/_ElasticSearchConfiguration.java', javaDir + 'config/ElasticSearchConfiguration.java', this, {});
+                this.template(SERVER_MAIN_SRC_DIR + 'package/config/elasticsearch/_ElasticSearchConfiguration.java', javaDir + 'config/elasticsearch/ElasticSearchConfiguration.java', this, {});
+                this.template(SERVER_MAIN_SRC_DIR + 'package/config/elasticsearch/_package-info.java', javaDir + 'config/elasticsearch/package-info.java', this, {});
+                if (this.databaseType == "sql") {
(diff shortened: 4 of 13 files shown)


lordlothar99:
@jdubois @deepu105, what do you think about this ? :)

deepu105:
LGTM this might have some effevt on my entity audit module but thats fine, i will find a way to make that work

jdubois:
- Very good idea indeed. However, I'm not sure of what people want: maybe they like more control on what is searched? As far as I'm concerned this is better than the current solution, but I'm not sure everybody would agree.
- I hate the BeanFactoryHolder that you introduce. I understand you need this for your listener, but there should be a better way of doing this -> we need to study this before merging (and that's why I add the "needs discussion" tag
- Maybe I haven't looked closely enough, but don't you also need to delete all the current save/update/delete code in the entities?

lordlothar99:
- @deepu105 I haven't checked consequences on audit module, please give me your feedback on that
- @jdubois I'm not very proud of `BeanFactoryHolder`, the fact is I wasn't able to find an easier way to get the `EntitySearchRepository`... At least, I will rewrite it using a dedicated `@Configuration` bean, and move `BeanFactoryHolder` to its own file (but anyway it will still contain `static` accessor).
- I indeed removed `EntitySearchRepository.delete()` and `EntitySearchRepository.save()` invocations from `delete_template.ejs` and `save_template` (only when both database is sql, and elasticsearch is present) ; I guess that's what you were talking about ?

deepu105:
I think its impossible to get a spring bean in entity listener as its not
managed by spring, I did something similar in the entity audit module and
not proud of that as well. If someone finds a spring way I would adapt that
as well. Btw your beanholder looks better than my implementation

lordlothar99:
I pretty sure few years ago Spring had some kind of `ApplicationContextHolder` or smthg... but anyway, I tried to apply KISS, and I got a new implementation (`BeanFactoryHolder` has been removed), I will update the PR in few minutes

lordlothar99:
Travis build failed because of phantomjs download error ... 
anyway, the PR is ok

deepu105:
ok this looks cleaner than what I was doing :+1:

lordlothar99:
@deepu105 plz tell me if you rebase/merge these commits

deepu105:
@jdubois LGTM plz merge if you are ok

lordlothar99:
I just changed `ElasticSearchUpdater` log level to `INFO`

amitjindal:
@lordlothar99 We did something similar in a project last year. You mentioned `ApplicationContextHolder`. Here's what we did. See if this is of any use for you.

`InjectionHelper.java`
```java
@Component
public class InjectionHelper implements ApplicationContextAware {

    private static ApplicationContext applicationContext;

    @Override
    public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
        InjectionHelper.applicationContext = applicationContext;
    }
    public static <T> T getBean(Class<T> clazz) { return applicationContext.getBean(clazz); }
    public static <T> T getBeanByName(String clazz) { return (T) applicationContext.getBean(clazz); }
}
```

Usage in EntityListener: `ElasticsearchEntityListener.java`
```java
@Component
public class ElasticsearchEntityListener {

    private final Logger log = LoggerFactory.getLogger(ElasticsearchEntityListener.class);

    private static final String CUSTOM_SEARCH_REPOSITORY = "SearchRepository";

    @PostPersist
    public void save(Object entity) {
        Optional<ElasticsearchCrudRepository> repo = getSearchRepository(entity);
        repo.ifPresent(esCrudRepository -> esCrudRepository.save(entity));
        if(!repo.isPresent()) log.warn("Could not save entity '"+entity.getClass().getSimpleName()+"' in elastic search");
    }

    @PostUpdate
    public void update(Object entity){
        Optional<ElasticsearchCrudRepository> repo = getSearchRepository(entity);
        repo.ifPresent(esCrudRepository -> esCrudRepository.save(entity));
        if(!repo.isPresent()) log.warn("Could not update entity '"+entity.getClass().getSimpleName()+"' in elastic search");
    }

    @PostRemove
    public void delete(Object entity) {
        Optional<ElasticsearchCrudRepository> repo = getSearchRepository(entity);
        repo.ifPresent(esCrudRepository -> esCrudRepository.delete(entity));
        if(!repo.isPresent()) log.warn("Could not delete entity '"+entity.getClass().getSimpleName()+"' from elastic search");
    }

    private Optional<ElasticsearchCrudRepository> getSearchRepository(Object entity){
        String searchRepository = entity.getClass().getSimpleName();
        searchRepository = Character.toLowerCase(searchRepository.charAt(0))+searchRepository.substring(1);
        searchRepository = searchRepository + CUSTOM_SEARCH_REPOSITORY;
        ElasticsearchCrudRepository elasticsearchCrudRepository = null;
        try {
            elasticsearchCrudRepository = InjectionHelper.getBeanByName(searchRepository);
        } catch (BeansException e) {
            log.error("Exception getting Elastic Repo: "+ searchRepository, e);
        }
        return Optional.ofNullable(elasticsearchCrudRepository);
    }
}
```

gzsombor:
We had similar EntityListener, with that ugly static global variable to link Spring with hibernate, but we abandoned this approach totally, as we encountered bugs, which caused by that hibernate called the listeners in a state when some of the ids are not initialized yet, so we ended up creating a home-grown 'entity change event listener' called manually.

lordlothar99:
@amitjindal thks, your version is somewhat close to the first I wrote, but as we discussed earlier with @deepu105 , it was decided to avoid the singleton ..
@gzsombor I'm interested in your feedback about bugs, would you plz give us some more details about your implementation and the bugs ?

jdubois:
I guess we could have some synchronization issues between the database and Elasticsearch, so maybe some people would want more control on their code, and thus would prefer the previous approach to this one.
We are too close to JHipster 3.0 to do such a big change, with nearly 20% of our users working with Elasticsearch -> I'd like to postpone this to another version, as this needs discussion and testing

lordlothar99:
Actually I'm using this implementation in several applications and never
had issues so far... But I will add some more stress tests just to ensure
everything's fine, so it will be ready for next version

Le dim. 20 mars 2016 09:13, Julien Dubois <notifications@github.com> a
écrit :

deepu105:
ok lets put this on hold then

gzsombor:
@lordlothar99 : Currently I can't find the exact issue in our issue tracker, but if I remember correctly, the problem was that we had a parent-child-grand child entity structure, which we wanted to index into one ParentEntityDocument. The problem was that, in some cases, the entity listener is called before all the entities are persisted and received a valid id - and this was caused a whole bunch of other problems later. It is entirely possible, that our implementation was flawed, and this solution would work correctly, I just want to share our experiences with the EntityListeners.

jdubois:
What I meant by synchronization issues:

- We need to be sure that commit/rollbacks are working the same way for both the database and ES. And as we have no transactions between them, I'm sure there are some cases where the data isn't correct in one of them
- Anyway ES isn't a true data store, so you can have lost or corrupted data with it

So for both cases, maybe some users would like to have something less magical than the listener, so they can control and correct their data manually in case of error. I do this all the time with Cassandra.

lordlothar99:
@gzsombor @jdubois you are right, this kind of problems may be hard to diagnose ..  I will add some tests soon

lordlothar99:
actually, it seems like both implementations (with explicit `EntitySearchRepository.save`, or with `EntityListener`) have same bug : if transaction is committed, no pb. But if transaction is rollbacked, then ES contains data it shouldn't, as ES call is not wrapped in JPA transaction... :/
Actually, this bug is already in current JHipster version... 
I think we could fix that easily with a transaction listener plugged in `ElasticsearchUpdater`, I'm working on it

jdubois:
yes, indeed! But my point is that it's not possible to achieve this correctly, and that anyway ES gives no guarantee to save an entity, so you will always have consistency issues.
That's why some people might want to have more control, so they can correct this manually.

lordlothar99:
I wasn't aware of theses ES limitations...
Anyway I implemented transaction handling in `ElasticSearchUpdater` (only activated in production mode, cause during integration tests, we have assertions on ElasticSearch containing data, and transaction is rollbacked).
And I also added a new `ElasticSearchUpdaterTest`, which might not be added to jhipster as is, but I found interesting to compare both solutions (with/without `ElasticSearchUpdater`). You can notice I deactivated `should_ES_contain_data_after_commit_without_ElasticSearchUpdater`, as this test demonstrates current JHipster issue on ES desynchronization when rollback occur. If I'm not mistaken, with `ElasticSearchUpdater` activated, the issue is fixed (see `should_ES_not_contain_data_after_rollback_with_ElasticSearchUpdater`).
Just let me know if this improvement might be included in next JHipster version or not, so I would cleanup the explicit `EntitySearchRepository.save()` invocations, in order to make this PR mergeable

deepu105:
@lordlothar99 this is on hold and that means wont be part of 3.0 release we can have this on a minor version later on as enhancement

deepu105:
and Thank you for the patience

lordlothar99:
Yes of course , I just wanted to go deeper into this issue cause I'm using this implementation in my project

deepu105:
@jdubois may be we should revisit this. I kind of think this is ok, there might be some edge cases as @gzsombor highlighted but that should only affect very few people and this should be good enough for common cases.

deepu105:
@jdubois @gzsombor so what are we gonna do here, we should either merge or close this.

jdubois:
- I think that's a very smart approach to have ES in sync with the database "magically"
- But I also think this is going to add 2 major issues:
  - There will be sync issues. There is no way to remove them, that's just the way ES works. So in those cases people will not be happy with the "magical" solution as they need to solve those errors manually.
  - Some people will be more happy with the current solution, as they can tune everything. In fact, this is also why we don't do everything "magically" in the REST controllers using Spring Data REST: people want those methods as placeholders, where they can add their business logic.

So I'm sorry but I'm not convinced by this approach in the end. Still, it's very smart, and should be at least documented, or maybe provided as a JHipster Module?

deepu105:
@lordlothar99 we are very sorry that we couldn't merge this. We highly value your contribution and we will atleast want to see this as a module as @jdubois said. Hope you would understand.

If you are interested to make this into a JHipster module so that others can benefit from this, please check this https://jhipster.github.io/modules/creating-a-module/

jdubois:
yes very sorry @lordlothar99 as I really find this is a very smart idea. I'd love to find another solution.

lordlothar99:
hi ! yeah no problem with that !
I had a lookup at the module API, but I first thought it would be quite hard to create a module..
I have to read once again, but I understood it was easy in a module to add some new logic, but not that much if you need to remove already generated stuff. And in that case, I would have to remove already generated `<%= entityInstance %>SearchRepository` invocations :/

deepu105:
you can always do some regex replace magic :) thats what I did in my entity
audit module

Thanks & Regards,
Deepu

On Tue, May 10, 2016 at 11:57 PM, François Lecomte <notifications@github.com

lordlothar99:
ok! I will have a look at that

PR #3178, opened 2016-03-16
For microservices and gateways, create application properties for Registry

gmarziou:
For microservices and gateways, create application properties named according to app baseName so that they can be moved to Registry.

In main application, check that t(hese files are not present in resources otherwise exit with instructions as below.

~~~
Exception in thread "main" java.lang.IllegalStateException: You must move src/main/resources/config/srv1*.yml files to JHipster Registry's central-config folder
	at com.mycompany.myapp.Srv1App.checkCentralConfigFiles(Srv1App.java:71)
~~~

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -860,9 +860,13 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_RES_DIR + '_logback-spring.xml', SERVER_MAIN_RES_DIR + 'logback-spring.xml', this, {'interpolate': INTERPOLATE_REGEX});
 
             this.template(SERVER_MAIN_RES_DIR + 'config/_application.yml', SERVER_MAIN_RES_DIR + 'config/application.yml', this, {});
-            this.template(SERVER_MAIN_RES_DIR + 'config/_application-dev.yml', SERVER_MAIN_RES_DIR + 'config/application-dev.yml', this, {});
-            this.template(SERVER_MAIN_RES_DIR + 'config/_application-prod.yml', SERVER_MAIN_RES_DIR + 'config/application-prod.yml', this, {});
-
+            if (this.applicationType == 'microservice' || this.applicationType == 'gateway') {
+                this.template(SERVER_MAIN_RES_DIR + 'config/_application-dev.yml', SERVER_MAIN_RES_DIR + 'config/' + this.baseName + '-dev.yml', this, {});
+                this.template(SERVER_MAIN_RES_DIR + 'config/_application-prod.yml', SERVER_MAIN_RES_DIR + 'config/' + this.baseName + '-prod.yml', this, {});
+            } else {
+                this.template(SERVER_MAIN_RES_DIR + 'config/_application-dev.yml', SERVER_MAIN_RES_DIR + 'config/application-dev.yml', this, {});
+                this.template(SERVER_MAIN_RES_DIR + 'config/_application-prod.yml', SERVER_MAIN_RES_DIR + 'config/application-prod.yml', this, {});
+            }
             if (this.databaseType == "sql") {
                 this.template(SERVER_MAIN_RES_DIR + '/config/liquibase/changelog/_initial_schema.xml', SERVER_MAIN_RES_DIR + 'config/liquibase/changelog/00000000000000_initial_schema.xml', this, {'interpolate': INTERPOLATE_REGEX});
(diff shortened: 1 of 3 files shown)


generators/server/index.js
```diff
@@ -860,9 +860,13 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_RES_DIR + '_logback-spring.xml', SERVER_MAIN_RES_DIR + 'logback-spring.xml', this, {'interpolate': INTERPOLATE_REGEX});
 
             this.template(SERVER_MAIN_RES_DIR + 'config/_application.yml', SERVER_MAIN_RES_DIR + 'config/application.yml', this, {});
-            this.template(SERVER_MAIN_RES_DIR + 'config/_application-dev.yml', SERVER_MAIN_RES_DIR + 'config/application-dev.yml', this, {});
-            this.template(SERVER_MAIN_RES_DIR + 'config/_application-prod.yml', SERVER_MAIN_RES_DIR + 'config/application-prod.yml', this, {});
-
+            if (this.applicationType == 'microservice' || this.applicationType == 'gateway') {
+                this.template(SERVER_MAIN_RES_DIR + 'config/_application-dev.yml', SERVER_MAIN_RES_DIR + 'config/' + this.baseName + '-dev.yml', this, {});
```
deepu105: Shouldn't we use slugified basename?

gmarziou: The important thing is that it should match eureka client name / spring application name property.

gmarziou:
There's an issue with tests, the user is expected to move the files to config server before running the app in dev or prod but maybe I have something to do in test also.

PierreBesson:
I'm having some reservations about this. Although I really like the idea of having configuration in one place. How would it work in "dev" when the developer doesn't have access to prod configuration (maybe this is managed by another team).

I  also liked that the config server needed only the conf that changed from the default and that it could overwrite the local config for the specific properties we need. For example I made use of that to enable log and metrics forwarding. But what you are proposing is to remove all local conf.

gmarziou:
> How would it work in "dev" when the developer doesn't have access to prod configuration (maybe this is managed by another team).

We should not mix 2 things : profiles and environments.

In my case, I plan to have 2 config repos:
- one for local and integration environments, so *-dev.yml would map to local environment and *-prod.yml would map to integration environment
- one for production environment, not accessible from developers' PC with some ciphered properties and with only *-prod.yml

Regarding your second point, yes that's right but how do you want to manage it?
Working in dev with a config server is mandatory I think as you may encounter issues that are only visible in this case like configuration refreshes.

gmarziou:
Another possibility would be to change the name of the config file even in standalone mode or to add another possible name different from "application".

This could be done using [this approach](http://stackoverflow.com/questions/29897802/spring-boot-override-convention-used-to-find-application-properties-config-file/29926689#29926689) by calling `ctx.setSearchNames("application, myService")`

This way we could have same filename `baseName*.yml` for config in both contexts: locally in resources  or moved to Registry folder.

deepu105:
@gmarziou @jdubois did you guys come to any conclusion on this?

jdubois:
What I don't like about this is that it doesn't work "out of the box". I'd like to have something that works straight away after generation, and that can be improved afterwards if necessary.

gmarziou:
Well with my proposal to use `ctx.setSearchNames("application, myService")` it would work out of the box and then the files can be moved to the regisitry without having to rename them

gmarziou:
@jdubois either we agree on my last proposal and I'll modify the PR accordingly or we close it

jdubois:
I don't have the time right now to review it, but I agree the current system is not good enough, we should at least have an easy way to share the JWT token after a first generation, or it's really not user-friendly.
I'll have more time in a couple of weeks.

gmarziou:
OK, it's not urgent, I can work on something else in mean time.

jdubois:
@gmarziou so you would have a default name that works out-of-the-box, but still generate a specific .yml file that could be moved to the registry, if the user wants it?
That would work in all cases, and having a "squeleton" .yml file that would be moved to the registry is a normal JHipster approach for generating config files for the user.
Still that looks a bit complex, and you have quite a lot of things to code and document - as this has been opened for a long time, what do you think about it? I'm not sure many people wants this, as nobody commented or liked the issue.

gmarziou:
So I close it.

PR #3179, opened 2016-03-16
differentiate between win32 and linux/mac environment when running

atomfrede:
use ``./gradlew`` on linux/mac and just ``gradlew`` on windows. 

@jdubois  @deepu105 Why is the build command generated again (and without ``-Pprod``) when setting up the push command? It is not used afterwards as far as I can tell.

### DIFF
--- a/generators/cloudfoundry/index.js
+++ b/generators/cloudfoundry/index.js
@@ -1,6 +1,7 @@
 'use strict';
 var util = require('util'),
     path = require('path'),
+    os = require('os'),
     generators = require('yeoman-generator'),
     childProcess = require('child_process'),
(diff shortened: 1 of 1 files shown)


deepu105:
@atomfrede its used in a log statement [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/cloudfoundry/index.js#L203) `-Pprod` is appended there based on profile selected. I guess we can avoid duplicate logic and assign it to a `this.buildCmd` and reuse it for the logging

deepu105:
since we are shipping `mvnw` wrapper should we also change to that from `mvn`

PR #3188, opened 2016-03-17
Add import-jdl sub generator

### LINKED ISSUE #3118: Add sub generator to import JDL files
This is to give an out of the box experience for our users. From GA stats seems like lot of people are using JDL studio already and providing a sub generator in the main jhipster generator will benefit more people to use JDL effectively and easily. 

Ill look at adding jhipster uml as a dependency and to invoke it with provided JDL file. May be we can add more features like pasting JDL directly on CLI to generate etc as well

@MathieuAA @CarlKlagba any comments

deepu105:
I would like to keep this as the first iteration of the sub generator (its based on jhuml 1.6.1), this works with JDL files from JDL studio well and I have tested it with few combinations

TODO:
1. Need to move to jhipster-domain-language module instead of JHUML 1.6.1

cc @MathieuAA lets add the feature first so it 3.0 has it and we will move to the new lib when its ready as an enhancement.

fix #3118

### DIFF
--- a/generators/app/USAGE
+++ b/generators/app/USAGE
@@ -1,0 +1,7 @@
+Description:
+    Creates a new JHipster application based on the selected options
+
+Example:
+    yo jhipster
+
+    This will compose jhipster:client, jhipster:server and jhipster:languages to scaffold a full application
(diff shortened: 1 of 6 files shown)


deepu105:
Im merging this as the first iteration

PR #3194, opened 2016-03-17
Remove entityUrlType variable in entity subgen

### LINKED ISSUE #3192: Error in JSON parse on get request entity in MongoDB app
When my app tries to get all Foos, the URL in href is wrong, it only put the first 2 numbers of the ID. So when I click on it, angular throws me an error `unexpected end of input`.
This occurred when running the mongoDB app with gateway in prod profiles, using docker-compose generator.
It works in a monolithic application.

##### **JHipster Version(s)** 

last master

```
{
  "generator-jhipster": {
    "jhipsterVersion": "2.27.0",
    "baseName": "mongoapp",
    "serverPort": "8083",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "databaseType": "mongodb",
    "devDatabaseType": "mongodb",
    "searchEngine": "no",
    "buildTool": "maven",
    "jwtSecretKey": "2485a077bbc237484be1341853dd81b6f29641d8",
    "enableTranslation": false,
    "testFrameworks": [],
    "skipClient": true,
    "skipUserManagement": true
  }
}
```

```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "name",
            "fieldType": "String"
        }
    ],
    "dto": "no",
    "service": "no",
    "entityTableName": "mongo_foo",
    "pagination": "no",
    "microserviceName": "mongoapp",
    "searchEngine": "no"
}
```

Last Chrome, Ubuntu 14.04

##### **Reproduce the error** 
Create an app with mongoDB, an entity and check the URL and click on the ID.

wmarques:
Fix #3192.
This variable caused bugs when using different DB types in microservice architecture. For example, if the gateway has an SQL DB and a microservice a Mongo one, the type in gateway front-end generated code will be wrong.
Moreover I didn't understand the point of validating the id using a regex (or a type) in client part. It was made by @andidev in b071e5bbb68ca821beb00d8d0c91d111c099ef93 so if you have any suggestion please tell me

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -1584,13 +1584,6 @@ module.exports = EntityGenerator.extend({
             this.entityAngularJSName = this.entityClass + _.camelCase(this.entityAngularJSSuffix);
             this.entityStateName = entityNameSpinalCased + this.entityAngularJSSuffix;
             this.entityUrl = entityNameSpinalCased + this.entityAngularJSSuffix;
-            if (databaseType == 'sql') {
-                this.entityUrlType = 'int';
-            } else if (databaseType == 'mongodb') {
-                this.entityUrlType = '[0-9a-fA-F]{24}';
-            } else if (databaseType == 'cassandra') {
-                this.entityUrlType = '[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}';
-            }
             this.entityTranslationKey = this.entityInstance;
             this.entityTranslationKeyMenu = _.camelCase(this.entityStateName);
 
@@ -1804,7 +1797,6 @@ module.exports = EntityGenerator.extend({
                         entityServiceFileName: this.entityServiceFileName,
                         entityStateName: this.entityStateName,
                         entityUrl: this.entityUrl,
-                        entityUrlType: this.entityUrlType,
                         entityTranslationKey: this.entityTranslationKey
                     };
                     // run through all post entity creation module hooks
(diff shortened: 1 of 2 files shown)


andidev:
I can't seem to see the reason just by looking at the code but there where a reason. The url where clashing as I remember it. Have you tried creating and viewing entities with this removed?

wmarques:
Yes I tried and it seems to work nicely

deepu105:
@wmarques I guess that code was added as a fix for issues caused by having uuid in the url if you remove that the issue might re appear especially for cassandra. I would suggest that a better fix would be to check the microservice app and use its database to get this parameter while generating client on gateway which wpuld be possible as you already read the entity path you just need to go a level above and read the yo-rc file

jdubois:
I don't know what can issue this could do or fix -> for me the id is just a String, we should not have to validate it, especially as it comes from the database
So what @wmarques looks good to me

deepu105:
I think angular route tries to validate the id thats were it was creating
issues may be just test once for cassandra and mongo db as well with
angular-prefix option enabled as well

wmarques:
I have no problem with mongodb with and without the angular-suffix option. I couldn't manage to test cassandra (don't know what to do to launch the database in dev mode...)

andidev:
Okey so now I remember. The typing of the id was needed since otherwise the url for creating new entities would clash with the url for showing an entity. 


myentity/new
and 
myentity/1

where both routed to the entity detail state and when trying to load an entity with the string new as the id we got an error. 

So creating new entities was not possible. 

Maybe this is fixed in ui router that routes with no routeparams are prioritiezed over routes with params.

jdubois:
@andidev oh, that's logical, but also quite unfortunate. We shouldn't have the same URL in the first place

PR #3195, opened 2016-03-17
add docker task for gradle build

PierreBesson:
This is doing the same thing that we currently have on maven.
It adds a docker gradle task to create a docker image using the war in `build/libs`.
Run with `./gradlew build docker`.
Tell me if the task name `docker` is OK. Or if you would like something more explicit like `buildDocker`, `dockerImage`, etc.

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -24,6 +24,7 @@ targetCompatibility=1.<%= javaVersion %>
 apply plugin: 'maven'
 apply plugin: 'spring-boot'
 apply plugin: 'war'
+apply plugin: 'docker'
 apply plugin: 'propdeps'
 defaultTasks 'bootRun'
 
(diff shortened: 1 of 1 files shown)


deepu105:
I thought this was working already

jdubois:
It's just me who doesn't know how to use it! Can you tell me how it works?
Le 17 mars 2016 12:45 PM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

deepu105:
Ok let me reconfirm, I saw the tasks listed when using gradle but didn't
actually try running it. SO im not sure if it was working, could have been
the default task from the plugin as we had the plugin included already.

In either case in the steps you just need to replace the maven command with
`gradlew build docker`

Thanks & Regards,
Deepu

PierreBesson:
Oh I just realized there was already a dockerBuild task. It is in `gradle/docker.gradle` that's why I had not noticed it before.

PierreBesson:
As it already exit I'm closing this PR.

PR #3197, opened 2016-03-17
Add docker-compose sugenerator

### LINKED ISSUE #3193: Migrate generator-jhipster-docker-compose to a sub-generator
The [JHipster Docker Compose generator(https://github.com/jhipster/generator-jhipster-docker-compose) is useful to everyone who wants to do microservices.

Migrating it to a sub-generator means that:

- It's installed for everyone from the start
- It's documented and easy to use in the main documentation

My proposal is that we call it "docker-compose", so that people only need to run "yo jhipster:docker-compose"

wmarques:
Fix #3193 .
Just a migration of the [Jhipster docker compose generator](https://github.com/jhipster/generator-jhipster-docker-compose). 
For now it works with SQL and Mongo databases, I got some trouble with Cassandra so if someone can help on this I'd be pleased.
As we discussed, shall I remove the question about profile as docker will be probably used to deploy application in production profile ?

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -39,7 +39,6 @@ module.exports = yeoman.generators.Base.extend({
         loadConfig: function() {
 
             this.defaultAppsFolders = this.config.get('appsFolders');
-            this.appConfigs = this.config.get('appConfigs');
             this.useElk = this.config.get('useElk');
             this.profile = this.config.get('profile');
             this.jwtSecretKey = this.config.get('jwtSecretKey');
@@ -226,7 +225,6 @@ module.exports = yeoman.generators.Base.extend({
         saveConfig: function() {
             if(this.abort) return;
             this.config.set('appsFolders', this.appsFolders);
-            this.config.set('appConfigs', this.appConfigs);
             this.config.set('useElk', this.useElk);
             this.config.set('profile', this.profile);
             this.config.set('jwtSecretKey', this.jwtSecretKey);
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -69,7 +69,7 @@ module.exports = yeoman.Base.extend({
                             if(file.isDirectory()) {
                                 if(shelljs.test('-f', file.name + '/.yo-rc.json')) {
                                     var fileData = this.fs.readJSON(file.name + '/.yo-rc.json');
-                                    if(fileData['generator-jhipster'] !== undefined) {
+                                    if(fileData['generator-jhipster'].baseName !== undefined) {
                                         this.appsFolders.push(file.name.match(/([^\/]*)\/*$/)[1]);
                                     }
                                 }
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -189,7 +189,6 @@ module.exports = yeoman.Base.extend({
             this.log('\nChecking Docker images in applications directories...');
 
             for (var i = 0; i < this.appsFolders.length; i++) {
-                this.log(this.appConfigs[i]);
                 if(this.appConfigs[i].buildTool === 'maven') {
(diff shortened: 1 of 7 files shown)


deepu105:
I dont understand why cant it be used for dev?

generators/docker-compose/index.js
```diff
@@ -0,0 +1,264 @@
+'use strict';
+var yeoman = require('yeoman-generator');
+var chalk = require('chalk');
+var shelljs = require('shelljs');
+var crypto = require('crypto');
+var _ = require('lodash');
+
+module.exports = yeoman.generators.Base.extend({
+    initializing: {
+        printJHipsterLogo: function() {
```
deepu105: remove the logo, we dont use the logo on sub generators just put some thing like `generating docker Files for your apps`

generators/docker-compose/index.js
```diff
@@ -0,0 +1,264 @@
...
+                    this.log(chalk.yellow.bold('WARNING!') + ' docker-compose is not found on your computer.\n' +
+                        '         Read https://docs.docker.com/compose/install/\n');
+                }
+                done();
+            }.bind(this));
+        },
+
+        loadConfig: function() {
+
+            this.defaultAppsFolders = this.config.get('appsFolders');
```
deepu105: this would make the yo-rc file of existing app look bad, may be we should save the config under a new file using http://yeoman.io/generator/Storage.html `new Storage(name, fs, configPath)` method

generators/docker-compose/index.js
```diff
@@ -0,0 +1,264 @@
...
+            for (var i = 0; i < this.appsFolders.length; i++) {
+                var path = this.destinationPath(this.directoryPath + this.appsFolders[i]);
+                this.appsFolderPaths.push(path);
+            }
+        },
+
+        saveConfig: function() {
+            if(this.abort) return;
+            this.config.set('appsFolders', this.appsFolders);
+            this.config.set('appConfigs', this.appConfigs);
```
deepu105: do we have to save each apps config as it would be present on the path anyway and we can read from there always since we know `appsFolders`

wmarques:
Actually it can be used for dev but I don't find an interest in that, because when you develop you use gulp and mvn commands with local databases. Docker is used to deploy your war files, so it's more for production usage. But maybe I'm wrong !

deepu105:
But If I want to make sure everything works together well I have run
registry and console from docker anyway right? Im not sure I kind of feel
that there might be a need since microservice use cases can be so different
across teams and we shouldnt restrict

Thanks & Regards,
Deepu

jdubois:
Yes it's better if we can test it in dev, then there isn't much difference now between both profiles, only the app is different, no the databases, etc...

generators/docker-compose/index.js
wmarques: Yes or we detect if the user run this subgen inside an existing app (by checking if the yo-rc has a `basename` property for example) and then abort the generation as it is made to generate files in an empty directory.

wmarques:
Ok I'll keep the dev profile then! I put @xetys work to detect if the project has maven or gradle

jdubois:
I'm merging this now as it's bad to have code duplicated between this repo and the original https://github.com/jhipster/generator-jhipster-docker-compose

jdubois:
@deepu105 there is no need to have it in dev, now that all the services are the same in dev and prod.

There are 2 use cases:

- In prod: you just run everything with Docker Compose
- In dev: you run the services you want with Docker Compose (that's why we now have a mysql.yml, elasticsearch.yml, etc), and then run your app outside Docker with Maven or your IDE -> that's much easier to develop

deepu105:
@jdubois Ok then lets merge this. I want to do some improvements here but ill do that on master once its merged.

Issue #3151, opened 2016-03-14
add profile switches to disable swagger, cache, liquibase


deepu105:
Based on discussion https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/jhipster-dev/Jeh8bYV_sdM

jdubois:
Closing this as the PR is merged

deepu105:
i havent added the no-cache yet but its ok

Thanks & Regards,
Deepu

PR #3198, opened 2016-03-17
Profile switches for no-swager and no-liquibase

deepu105:
profile switches to disable liquibase/swagger

For jar|war it can be added as `--spring.profiles.active=prod|dev,no-liquibase|no-swagger`

For Gradle local bootRun `./gradlew -Pno-liquibase -Pno-swagger` or `./gradlew -Pprod -Pno-liquibase -Pno-swagger`

Can someone check for maven local run 

part of #3151 but no-cache seems to be tougher than I thought will do it as a seperate PR

### DIFF
--- a/generators/server/templates/gradle/_liquibase.gradle
+++ b/generators/server/templates/gradle/_liquibase.gradle
@@ -1,30 +1,30 @@
 configurations {
-  liquibase
+    liquibase
 }
 
 dependencies {
-  liquibase group: 'org.liquibase.ext', name: 'liquibase-hibernate4', version: liquibase_hibernate4_version
+    liquibase group: 'org.liquibase.ext', name: 'liquibase-hibernate4', version: liquibase_hibernate4_version
 }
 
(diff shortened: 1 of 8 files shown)


PierreBesson:
I tested it. It works for the war. Some configuration should be done for maven in the pom.xml so that it would work the same as with gradle. Does anyone have an idea on how to do that ?

deepu105:
my maven skills suck so i need help on that

Thanks & Regards,
Deepu

deepu105:
@jhipster/developers can anybody do this for maven, should this be done using maven profiles like [here ](http://stackoverflow.com/questions/7513319/maven-command-line-arguments)or is there a better way?

jdubois:
Oh we should do this before our 3.0 release!
Le 23 mars 2016 2:57 AM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

jdubois:
@deepu105 you can't "merge" several Maven profiles like you do with Gradle. Yes, Gradle is more powerful :-)
So basically you can only do this at runtime. I'm updating the documentation accordingly.

Issue #3206, opened 2016-03-18
Set default spring.jpa.show_sql to false


lordlothar99:
##### **Overview of the issue**

Hibernate queries executions slow down build execution, and fill the console ; developers would rather activate them when necessary, but parameter default value should be `false`

##### **JHipster Version(s)** 

2.27.0

jdubois:
I thought this was only true in the "dev" profile? Are you sure?

lordlothar99:
it was in both `src/main/resources/config/_application-dev.yml` and `src/test/resources/config/_application.yml` ... in my opinion it's better to set it to `false`, so builds are faster ; anyway, developers can always put it to `true` in their own dev environment

lordlothar99:
PR is #3207

jdubois:
No that's on purpose: we output SQL statements in dev so that people see what they are doing. It's not slowing anything down, it's just outputed to the console. Or you have a much, much bigger issue as you are doing a ton of SQL requests (and that's what is slow).

PR #3207, opened 2016-03-18
#3206 Set default spring.jpa.show_sql to false

lordlothar99:


deepu105:
I actually wouldn't mind this being true in dev so that you get clearer messages if something goes wrong, its false in prod anyways

deepu105:
for test profile its ok to be false

jdubois:
Oh yes maybe for tests profile, then it has already helped me several times, when a test is wrong it's useful to see the SQL logs

Issue #3220, opened 2016-03-19
setDocumentRoot breaks apps in cloudfoundry


deepu105:
This is a tracking ticket for issue discussed in commit https://github.com/jhipster/generator-jhipster/commit/c9f5949654d4f6d6fd01e296a85ae47f72ed9ef7#commitcomment-16742432

This is actually a blocking issue for cloudfoundry deployment and we need to do something for this

cc @jdubois @atomfrede @erikkemperman 

My suggestion is to do something like below as a workaround for now, this would only affect the case where someone runs the app from the IDE with prod profile, in that case they would have to set an additional profile param `spring.profiles.active=prod,debug`

```
if (env.acceptsProfiles(Constants.SPRING_PROFILE_PRODUCTION) && env.acceptsProfiles(Constants.PROD_DEBUG)) {
        container.setDocumentRoot(new File("build/www/"));
}
```

erikkemperman:
I still think it should be possible to accurately detect whether we're running from a war or not, without a separate profile, but unfortunately I have very little time these days. So this proposal of `prod,debug` looks awkward to me but is obviously better than current broken situation.

jdubois:
I'm sorry I didn't follow this closely enough -> why do we have setDocumentRoot in the first place? I looked in the code and couldn't find an explanation (oh, and this should be documented, also)
We didn't have this before, and it seems that it's causing quite a lot of issues

erikkemperman:
It's needed when running prod profile outside of the war (eg IDE or spring-boot:run) because the optimized web assets are no longer in the default location at `src/main/webapp`. Before they were in `src/main/webapp/dist` and there was a server filter to append the last bit of that path.

The problem is that `setDocumentRoot` shouldn't be called when running inside of the war: it is unnecessary because the assets are packaged along, and actually breaks things if we do call it. But we haven't figured out a nice way yet to detect accurately if this is the case.

The latest attempt, checking the protocol of the source location, is obviously not working out very well; it fixed the case for `java -jar target/foo.war` but not for deploying the war (eg to cloud foundry).

I don't have a lot of time to try this out, but now I'm wondering if we couldn't simply check, in between creating the File object and calling `setDocumentRoot`, if the directory exists? Not sure why I didn't think of that earlier, but it just might solve the whole thing.

By the way, @deepu105  mentioned me in a locked conversation... I've tried to reply by mail but not sure if that works for locked issues: where should I look for updating documentation about the changed directory layout?

[...]

PR #3222, opened 2016-03-19
Detect properly if setDocumentRoot will work

erikkemperman:
I believe this (finally) addresses [#3220](https://github.com/jhipster/generator-jhipster/issues/3220) properly.

Tested cases:
- running in IDE
- running with `mvn spring-boot:run`
- running with `java -jar target/foo.war` from an abitrary cwd
- deploying to external servlet container (tomcat 8)


atomfrede:
Works well with cloudfoundry.

deepu105:
This is better than the current solution anyday :)

jdubois:
Good idea, indeed!!

Issue #3180, opened 2016-03-16
Remove the "dev" Docker and Docker Compose configurations


jdubois:
This ticket comes from [this conversation](https://github.com/jhipster/generator-jhipster/issues/3154#issuecomment-197182011).

The idea is from @erikkemperman and I'd like to have feedback from at least @pascalgrimaud as he's the expert of this part.

- Having a "dev" packaged WAR file doesn't have any usage, and in fact it could be an issue as people could deploy it in "prod" by mistake -> so the original idea is to have only one "prod" packaged WAR file, and that should be enough
- The same thing occurs for Docker and Docker Compose. The whole idea is to have the same configuration everywhere -> we should only have "prod" images that we use to test and run in prod

So I would like to delete the app.dev.yml and db.dev.yml files, as well as all references to "dev" profiles in the Docker and Docker Compose configuration.

pascalgrimaud:
I will read all conversations and answer it tonight

erikkemperman:
Perhaps to avoid misunderstandings... I was thinking to set the maven/gradle profile used when building the war as the default spring profile when running/deploying that war.

So if I do `mvn -Pprod package` then I would like to get the `prod` profile when I run `java -jar target/xxx.war` (or deploying it) without having to specify `--spring.profiles.active=prod` (or `-Dspring.profiles.active=prod`).

Of course, this was already working as expected for `dev`, because it is default active profile both in maven/gradle and spring.

So I think what I am proposing is actually different from "remove `dev`" as stated in this issue...?

I believe this is the best way forward for war files -- make a sane default, but don't make it impossible to make a `dev` war.

I'm not sure if this makes sense for Docker though.

jdubois:
Oh, maybe I'm going to far here....

OK, the current system is working well, I've never heard complaints about it: let's not break it now that the 3.0 release is approaching.

I'm closing this, but this is something we should think of in the future.

erikkemperman:
So you're just closing this issue of "removing `dev`" -- but do you still agree that the default profile for running the war should be the one that was used to build it? If not please let me know, then I won't spend any time on that :-)

jdubois:
Yes the default profile is the one used to generate the war -> maybe we should write this in the war name, otherwise you can't know if a war is for prod or dev, and that can lead to errors

deepu105:
Thats exactly what i did in my last app I appended the profile to war name
so that i know what war goes were, I had dev, sit, uat, dr and prod
[...]

PR #3224, opened 2016-03-19
Use Maven/Gradle profile(s) as default Spring profile(s)

erikkemperman:
Sorry for the many PRs, I will stop now :-)

As discussed, among other places, in issue [#3180](https://github.com/jhipster/generator-jhipster/issues/3180), this promotes the profiles used with Maven/Gradle to be the default profile(s) in Spring -- unless an explicit `spring.profiles.active` is given, of course.

This means that `mvn -Pprod package && java -jar target/foo.war` does what you'd expect without the need for arguments. Likewise when deploying the war.

It might be too late for the 3.0 release...? Although I kind of feel that it belongs together with the change that made the `dev` and `prod` assets mutually exclusive in the war file.

I did not change the default profile to be different for `run` or `package` goals, as was suggested: I am not sure this is even possible with Maven.

I briefly looked into changing the war filename depending on the profiles used, as was suggested in the linked issue, but the generators for AWS, Cloudfoundry, and Heroku seem to only expect a single war file to be present:  Cloudfoundry and Heroku use `*.war`, while AWS (arbitrarily?) picks the last `*.war.original` file returned by `fs.readdirSync`.

By the way, shouldn't the uploaded war for those generators always be the `.original` one, also for Heroku and Cloudfoundry?


deepu105:
I know for sure that .original us required for AWS but not sure of heroku
and CF. I can check CF but someone needs to check heroku

deepu105:
And yes this should be part if 3.0

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -21,20 +21,19 @@ if (project.hasProperty('no-cache')) {
...
 }<% } %>
 
-task setProdProperties(dependsOn: bootRun) << {
-    doFirst {
-        System.setProperty('spring.profiles.active', profiles)
-    }
-}
-
 processResources {
+    filesMatching('**/.build-properties.xml') {
```
deepu105: I hate to have an xml for this its kind of ugly :( cant we do the same to the profiles .yml file?  Or using the application.yml file?

deepu105:
I had a look and im not sure. I dont like having additional xml and a class for this, isnt there a simpler way? 

Im not sure if this additional complexity which is a bit non standard for normal spring apps IMO is worth the value provided which is letting the user run without passing a profile.

Any way we need to see if there is any simpler alternative to this

erikkemperman:
Absolutely agree that this is not very pretty! This is why I made it a hidden file -- slightly embarrassing :-)

I haven tried to find a nicer way to do this, but couldn't find one. If anyone has ideas... Please let me know!

For what it's worth though, putting this in the application.yml definitely won't work; it has to be done before spring parses those yaml files, because it will be used to determine which of those to load.

erikkemperman:
Of course the same mechanism, with almost no extra effort, could now be used to deliver more properties from build to runtime. For my own application I'm thinking to include build tool (maven/gradle) date, git revision (hash of latest commit), ...

deepu105:
May be lets take some time and see what can be done, this is not a blocker
for 3.0 anyway so we can take our time

Thanks & Regards,
Deepu

deepu105:
lets do this after 3.0 so im gonna put this on hold

erikkemperman:
Well, despite maybe being subject to a better implementation, I think that it functionally accomplishes a useful detail -- and it shouldn't get in the way of anything else, I don't think, contact surface with rest of the code is unobtrusive and limited to the main and webxml classes.

But fair enough, please let me know if / when to pick this up, or if anyone sees better ways to do this!

erikkemperman:
Maybe I should mention one other implementation I considered, but decided against: the `@variable@` subs could of course be done on (a) Java file(s) directly, without an extra xml file or properties class. But we were interpolating/filtering the xml files anyway (for `@loglevel@`) and having the class seems preferable as it's a natural place to handle fallback for unsubstited values (e.g running in IDE).

deepu105:
lets wait and see if someone has a better idea

Thanks & Regards,
Deepu

erikkemperman:
Minor FYI: removed a now unneeded import from ApplicationWeb.xml, and rebased onto latest master (had to adapt to a commit to make gradle/protractor).

I would still love to hear if anyone has ideas on how to improve on this implementation!

erikkemperman:
Travis is flaky; restarted the failing test and now it [passed](https://travis-ci.org/erikkemperman/generator-jhipster/builds/118368654).

deepu105:
@jhipster/developers guys we need to progress on this, any comments from you guys?

atomfrede:
Looks good to me. I have no feeling against xml, in case it stays as small as possible

PierreBesson:
I'm in favor of including this. It will be great for on-boarding because users would expect the app to start in prod be prod when packaging a jar for prod. Also if spring.profiles.active always take precedence it should not cause issues.
It will also be very helpful as we started to make more use of profiles recently like no-swagger and no-liquibase.

erikkemperman:
Rebased again and fixed an inaccurate comment in BuildProperties.java.

deepu105:
Im not a spring boot expert but I was wondering. Spring boot does support loading stuff from an external `application.properties` or `application.yml` file as stated here https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html

may be we could figure something out using that and the springs various annotations to load specific stuff based on things available in classpath or environment. Thats how spring does most of the autoconfiguration magic isnt it?

erikkemperman:
Well, I'm no spring expert either, so maybe I'm missing something... 

But my thinking was we can't use spring's autoconfig voodoo for this particular case, because we need it done *before* spring looks for external `.properties` or `.yaml` files. This is because the profile we are loading here will be used to determine which external config files spring must load, e.g. `application-dev.yml` or `application-prod.yml`.

I could change it from `.build-properties.xml` to a regular properties file if you prefer, I chose xml just because the maven pom was already set up to do filtering (which is its odd name for what is actually substitution) on xml files.

But I think we'd still need to load it manually, and provide a fallback for unsubstuted values, which makes me think having a separate class (`BuildProperties.java`) is actually a pretty natural fit for the purpose.

deepu105:
Lets see what @jdubois has to say

Thanks & Regards,
Deepu

jdubois:
Yes Spring first reads all those profiles, so you can't modify them afterwards (or you need to refresh your app context, but that's another story).
I'm not a big fan of having the same name for the WAR file both in dev and prod, as then it's very easy to pick up the wrong version.
I'd like to have a "-dev.war" and "-prod.war" file generated, depending on the Maven profile used to do the packaging. Yes we do use "*.war" in some sub-generators, as the War file might as been changed by the user (it depends on the app name), but we can still do a "*-dev.war" and "*-prod.war" instead.

erikkemperman:
Maybe the war name ahould be separate from this PR?

Like I said I looked into it but several cloudy generators would need work as well, they don't expect more than one war file to exist.

jdubois:
Yes but we should change those sub generators.

erikkemperman:
I'd propose dealing with war names separately, for the moment I think the question for this PR was if there is a better way to pass maven/gradle profile to spring?

erikkemperman:
Rebased and fixed a conflict that arose due to recent commit(s) on master.

I'm curious how to proceed here... It seems that functionally the consensus is that this would be a useful addition, but I haven't seen any suggestions as to how to improve the implementation?

deepu105:
@erikkemperman sorry again to let you hang in like this. Gimme some time to test this out

erikkemperman:
No worries, it's just that I have to keep rebasing because I have a project based off this branch which I want to keep up to date with JH master, at least until the next minor release (which will hopefully have this, or something like this, in it).

Thanks for taking the time, whenever you find some!

deepu105:
Ok so i tried few other ways and I guess whatever is said [here](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html) doesnt completely work in our case and hence Im ok to proceed with this PR half hearted. So @jhipster/developers if no body has an objection we can merge this.

@erikkemperman thanks for hanging on with us :)

erikkemperman:
Thanks for looking into this!

I'm a little uncomfy with your "half-hearted" remark, wouldn't want to inflict anything on you guys you're not happy with... So can I ask if that's just cosmetic, aesthetic considerations? Or are you actually worried if it might cause functional headaches?

deepu105:
For now its just aesthetics :)

deepu105:
@jdubois we can merge this if you are ok

erikkemperman:
I don't mean to be too forward... But any chance of getting this in soon? I am working up to a release and would much prefer not to base it off my own branch. Thanks in advance!

deepu105:
Im waiting for @jdubois to take a look. He is in vacation rite now

jdubois:
No I'm back :-)
But lots of "real" work, and there's Devoxx Fr next week... I don't have much time for coding on JHipster at the moment.

jdubois:
I've had a look:

- I still find having an XML file quite complex. Does this work also when you right-click and run the Main application class?
- I'd rather have a "-dev.war" and "-prod.war" files generated, so you know which file you have

Another option would be to only generate a production build - anyway I don't think there's a valid use case for doing a "dev" WAR file. We used to have this because we couldn't have source maps in JavaScript, but that's solved now.

erikkemperman:
@jdubois Running in IDE works same as before, it will see unsubstituted @placeholders@ and fall back to old behavior.

It seems everyone thinks the xml file is not pretty (myself included!) but I haven't seen proposals to do it more elegantly. Thoughts?

I agree that the single war name is potentially confusing... Either `foo-dev.war` and `foo-prod.war` or just defaulting to prod would work for me.

But I think that is outside the scope of this PR and probably better addressed separately?

deepu105:
I guess some teams might want to have a dev war file and also even if we
always build a prod war how will we set the prod profile as default for
that it will still have the same issue as today and might need the same
solution as this PR

jdubois:
So Deepu you want to merge it?
Who also likes this PR?
I don't want to be the bad guy holding on stuff everyone else wants!
Le 20 avr. 2016 2:27 AM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

> I guess some teams might want to have a dev war file and also even if we
> always build a prod war how will we set the prod profile as default for
> that it will still have the same issue as today and might need the same
> solution as this PR

deepu105:
@jdubois I'm hesitant as well but I don't see a better way to do this. either we need to merge this or let the users always set the profile manually

erikkemperman:
FYI Rebased again to resolve conflicts with commits to master branch.

erikkemperman:
FYI Rebased again... Any news on this PR?

notflorian:
This PR looks good to me. :+1: 
@jdubois @deepu105 : Would you be OK to merge it, or do you think it need more changes?

jdubois:
- I still really don't like reading an XML file. I'm sure this will cause all kinds of issues, and having our own XML-based configuration cannot be a good solution.
- My current opinion is still that we should either generate 1 war file per profile, with the name of the war reflecting the profile chosen (doing "mvn -Pdev package" would generate myapp-dev.war), or only generate production WARs (who need a WAR packaging if it's not for production?). So this change wouldn't be needed

erikkemperman:
@jdubois What kind of issues are you expecting? I've been working with a project with this patch included and for my use cases everything is fine. Note that it has basically zero contact surface with the rest of the code.

The custom XML processing (well, not really, it just uses the standard `Properties.loadFromXml` method) is necessary. Or at least, we can't use Spring's autoconfig voodoo, for reasons discussed several times in this thread.

The thread seems to be getting somewhat circular -- you keep bringing up the naming of the war files and I agree but would again suggest that this is a different topic and out of scope for this PR :-)

Anyway, either way is fine with me. I'm going to have to make a first release of my project this week. It would have been nice if this PR was in the main jhipster repository, or even better if it were part of a minor release... But I guess that's not going to happen, fair enough.

So, feel free to do with this what you like, but I might not be re-basing it all the time anymore to follow jhipster/master.

jdubois:
Sorry for repeating myself, it's just that their are more comments on this thread, and my opinion still hasn't changed from the beginning, so of course I'm always answering the same thing to the same comments.
I still don't think having our own XML file, and reading it from the classpath, is a good idea. Then, yes I agree our current process isn't good, and it should be changed.

erikkemperman:
Well I totally agree it's not pretty -- although I don't expect much trouble other than sore eyes -- and I've certainly been open to suggestions for improvement. It seems most folks agree it is ugly but there haven't been any more elegant proposals, unfortunately.

Not sure what you mean by the "process isn't good"?

Also, I just noticed I had missed a key part in your earlier comment:

> [with proper war file names...] this change wouldn't be needed

Maybe I misunderstand, but you'd still have to do `--spring.active.profiles=...` right? Are we sure we have the same image of what this PR does / was supposed to do?

jdubois:
"process isn't good" -> I agree the way it currently work isn't good, we need a better solution
"with proper war file names" -> my idea here is that by default we package the right profile with each war file. So wouldn't have to add `--spring.active.profiles` with a `-prod.war` file, as it would already be its default profile.
Sorry I'm just lacking time on this part, we have 30 tickets opened and a lot of work, and I'm mostly doing this on my free time.

erikkemperman:
I would be curious to see how setting the default profile would work -- that is pretty much what I was trying to do here -- so I'll keep an eye on this!

No worries, I certainly didn't mean to sound impatient! Your efforts are much appreciated by many here, I'm sure... It's just that this PR has been in limbo for a while now :-)

jdubois:
Thanks, I just need some time to review tickets, but there are so many of them at the moment...

deepu105:
@jdubois @erikkemperman let me step in to clear few things.

@jdubois even with having the profile in war names I dont think there is a clean way to set the default spring profile and thats the very reason for this PR and like you I dont like the XML based solution and had a long discussion with @erikkemperman.

Hence naming pattern of war doesn't have anything to do here, our problem from the very beginning was setting a default spring profile and we always asked the user to do that manually and either we can set it using this PR or continue to ask the user to do it manually like earlier.

I tried to do it with spring default profile property  ut doesnt seem to work, but as per spring documents if we set that in the `application.yml` its suppose to work. But somehow it doesnt.

erikkemperman:
@deepu105 Thanks, I think that's all accurate. Wanted to point out though: it's not a choice anymore, since the war contains only one profile in terms of front end code, the user has to pass the same profile used while making the war.

deepu105:
Yes you are right I guess I didnt articulate that clearly :)

Thanks & Regards,
Deepu

jdubois:
OK I finally had the time to review this.

- We all agree that our current way of working is broken, no problem with that
- I'm still not a big fan of the XML file, but then the idea is to remove the env variable, so indeed we need to store this somewhere - we could have written this in a properties file, but in the end it doesn't change anything, so let's keep it that way
- My own idea was to set this in the WAR name, but as @deepu105 says this isn't working well. I was worried that with your solution people wouldn't know which profile they were using: what I didn't see is that they can still set their own profile. And what I didn't thought of is that the profiles are logged at the very start of the process, so people shouldn't be surprised.

So in the end I agree with you. I'm not merging this, as this could be some kind of a surprise -> @deepu105 if this is fine with you I let you merge this!

Thanks @erikkemperman for having the patience to deal with me :-)

erikkemperman:
@jdubois Thanks for taking a look! One little remark: the only reason I picked .xml instead of .properties is that the former filetype was already set up to be filtered (substituted) by maven/gradle, because of the logback.loglevel property. If you prefer .properties that would be a trivial change.

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -32,19 +32,16 @@ task gulpConstantDev(type: GulpTask) {
...
+    }
     filesMatching('**/logback-spring.xml') {
         filter {
             it.replace('@logback.loglevel@', logbackLoglevel)
         }
     }
 }
 
-<% if (!skipClient) { %>
-setProdProperties.dependsOn gulpConstantDev<% } %>
```
deepu105: this might affect cases when we `bootRun` dev profile after running a prod build, can you check this @erikkemperman gulpConstantDev task should be run may be we need to change this to `bootRun.dependsOn gulpConstantDev` cc @atomfrede

deepu105:
@erikkemperman @jdubois I was thinking and I got this weird idea, and since we are doing a not so pretty solution here anyway we could try this as well.

So lets say we go by the approach here, instead of using `.build-properties.xml` to load the default profile why not have this `spring.profiles.default: @spring.profiles.default@` in our `application.yml` and replace the value there from maven and gradle. Gradle will do it regardless of xml or yml but for maven might need to tinker a bit but remaining aspects of this PR will remain the same. Mean your code will manually read this
```
    private BuildProperties() {
        try (InputStream in = getClass().getClassLoader().getResourceAsStream("application.yml");
                BufferedInputStream bin = new BufferedInputStream(in)) {

            this.loadFromXML(bin); // load from yml instead of xml here
        } catch (IOException e) {
            log.error("Failed to read build properties", e);
        }
    }
```
So in the end we still achieve this without having to have a hidden xml.

What do you guys think?

mraible:
Doesn't this mean you can't change the profile after the war is built? One of the nice things about Spring Boot is you can take a WAR from dev -> test -> prod w/o repackaging it. This seems like it might take that feature away.

generators/server/templates/gradle/_profile_dev.gradle
erikkemperman: @deepu105 I think you're right, I did this because the `setProdProperties` task is no longer there... But now the indirect dependency to `gulpConstantDev` is gone. However, should that not be a switch, based on active profile, to `gulpConstantDev` or `...Prod`?

I am unable to check this out until tomorrow, unfortunately.

erikkemperman:
@deepu105 I actually considered something like that, at the time. But I decided against it because

- we then can't use the built in parse methods from the `Properties` class, 
- my IDE marked yaml values like `@this@` as error,
- I thought it might look to users like they have to substitute this themselves, although of course a comment might help.

erikkemperman:
@mraible This only sets the default, you can still override and set another profile. Having said that, the front-end code in the war will be `dev` or `prod` depending on the active profile at package time. That last bit is unrelated to this PR.

deepu105:
For your 3 points,

1. Spring parses yaml using snakeyml and there should be some class exposed
which we can use to parse it as well. Ill take a look
2. We dont have to stick to @this@ syntax it could be #this# or anything
else that yml can manage, or simply just wrap @this@ in quotes and it will
still work the same
3. as you said a comment would do

erikkemperman:
@deepu105 if we can get maven to substitute `#these#` instead, or some other placeholder syntax other than the `@default@` -- great! We'd have to change the log level placeholders too, of course.

erikkemperman:
And if we're going to be filtering other files than .xml, could be the `BuildProperties.java` directly?

deepu105:
@jdubois @erikkemperman Ok so I did some more digging to see what was causing the original issue in our setup where the below
```
spring:
    profiles:
        active: prod
```
in `application.yml` was not working. and I found the cause, seems like we were setting `addDefaultProfile(app, source);` before the environment was loaded and hence `dev` profile was always taking precedence over `application.yml` see [this order](http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html)

So I guess the solution here is very simple and much cleaner. we can set the below 2 properties in our `application.yml`
```
spring:
    profiles:
        active: #spring.profile.active#
        default: dev
```

and we need to replace `#spring.profile.active#` with `prod` or `dev` based on what profile is used when running maven/gradle. and spring will automatically load the appropriate `application-{profile}.yml`

So this means we wouldnt need an xml and a  `BuildProperties` class and IMO its much cleaner approach and kind of the spring way. 

I tried this on a generated project and seems to work.

I foresee once issue with replacing content using maven/gradle with this PR as well as with my approach above. When we run the build with a specific profile it will replace the placeholder in the file, if we run the build with another profile later there is no more placeholder to replace. @erikkemperman how did you handle that, I dont see anything specific for that in the PR?

erikkemperman:
@deepu105 That sounds good! So you figured out how to call `addDefaultProfile` before the env is loaded?

Note that maven/gradle don't change the file itself, but substitute while copying to `target`/`build`. So that shouldn't be an issue.

Without the properties class, though, how to handle fallback when running without maven (in IDE)?

erikkemperman:
Also, will it still be possible to override the profile after the war is generated (see @mraible's comment above)?

deepu105:
@erikkemperman we dont have to call `addDefaultProfile` as `spring.profile.default` should take care of that.

If maven/gradle doesnt change the source then it should be good, then we have to pass the profile as the `-Dspring.profiles.active` similar to today to make correct profile load during `bootRun` and while running from IDE, since its handled already today we can retain the same and do the replacing of the `@active#` property only during `mvn package` or `gradle bootRepackage`

I think we cant use the dev war for prod and vice versa due to the PR you did to move stuff to the target folder. Anyway we discussed that and was decided that way. So no point going back on it again. I agree it was a nice feature but it was also adding a lot of bloat to the built package

erikkemperman:
@deepu105 Ah, sorry, I had missed you put `profiles.default` in the yaml as well.

Maven/gradle doesn't change the source, at least in my setup (I hiked along with how it replaces `@logback.loglevel@`, which would also have been broken otherwise).

You're right about the front-end implications of `dev` vs `prod` of course, but it was my understanding that there are now additional profiles that affect backend only? This might also be true for user-defined profiles (I know I have those in my project, at least). Also, there seems to be an end-point for getting the active profiles now, so except for how the front-end code is packaged (i.e., concatenated, minified, revisioned or not) there could still be front-end switches on profile.

But as long as the override with `-Dspring.profiles.active` or `--spring.profiles.active` still works this should all work out fine I guess.

About fallback when running in IDE, yes we should then set `--spring.profiles.active` manually if it is not `dev`. That would work. But I was more worried about what spring will do when it tries to load a profile called `#spring.profile.active#` which is what it would I expect it to do when it parses the source `application.yml` rather than the substituted one in `build` or `target` folder?

All in all I absolutely agree that this sounds like a much cleaner way to achieve the desired effect. Kudos! If / when you have a branch that you'd be willing to share, I would love to try it out!

deepu105:
# is comment in yaml so spring takes it as empty :)

BTW seems like the `profiles.default` is not officially supported by spring boot and im tryig to see if there is another way to set a default profile when application.property is not set

erikkemperman:
No need to shout :-) (But I see what happened there, with the hash tag making it H1)

deepu105:
LOL.

Ok found a way to set the default profile elegantly :)

@erikkemperman I can do a PR with the settings to change the way default profiles are done. But im a gradle guy and my maven skills suck, so could you help to do the replacing `#active#` part for maven?

erikkemperman:
@deepu105 Well, I'm not sure how we could override the default delimiters `@` for values to be substituted... The [docs](http://maven.apache.org/shared/maven-filtering/) mention an alternative `${..}` notation but that wouldn't be a comment in yaml.

erikkemperman:
@deepu105 Okay, I think it will require moving how it is done [currently](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/_pom.xml#L696) into a separate [plugin](http://maven.apache.org/plugins/maven-resources-plugin/copy-resources-mojo.html) section, but it looks like we can then specify our own delimiter.

I will try this in a separate project, and will let you know what I find.

jdubois:
This looks much, much better than the XML file :-)
@deepu105 @erikkemperman do you need any help? Looks like you're solving this perfectly!

erikkemperman:
@deepu105 I think something like [this](https://github.com/erikkemperman/jhipster-resource-subs/commit/5b5e4ae58bdb822ab86b6796293f71c23085f823) would do the trick. It assumes that we also change the delimiter for the logback level, so beware that that would have to change in gradle as well. Alternative, we could use both `@` and `#` as delimeters, I suppose/

Either way, I trust this gives you all you need to complete the PR? If not, let me know. 

@jdubois Agreed, if Deepu's approach works it is much better. Thank you guys so much for not giving up on this!!

deepu105:
@erikkemperman I will try today else I can do only next week as im going on a short vacation tomorrow

jdubois:
@erikkemperman yes and I'm very happy :-)
@deepu105 tell me if you need help! We need this for our next version :-)
And don't forget to update the doc accordingly :-)

deepu105:
Ill commit what i did so far in a PR any way, so that you guys can try it out if required. 
@jdubois The `ApplicationWebXml.java` file has a method setting a default profile
```
@Override
    protected SpringApplicationBuilder configure(SpringApplicationBuilder application) {
        return application.profiles(addDefaultProfile())
            .sources(Application.class);
    }
```

But this doesnt seem to be triggered during bootRun or when run from war, when is this trigerred? is this for deploying to tomcat/jboss etc?

jdubois:
@deepu105 to be honnest, I'm not sure if it's being used :-)
I coded this at the very beginning of the project, at that time it was to replace the web.xml file, but as of now it looks like it's not doing anything.... You could try removing it? And that would also clean up the root package, that would be perfect.

deepu105:
Its working without this file, when I run from IDE, bootrun from console or when run as `java -jar` need to check deployment in tomcat

jdubois:
@deepu105 then kill it, KILL IT 💯

erikkemperman:
@deepu105 @jdubois I wondered the same thing, when I was working on this PR, but if I remember correctly this is indeed (only) called by servlet containers as an alternative to web.xml. The logic is completely analogous to the main application class, though, so not really hard to keep both equivalent.

jdubois:
@erikkemperman that's also what I remember... stupid app servers! Then we might need to keep it, and validate that it works well with this new setup

erikkemperman:
The documentation for the superclass `SpringBotoServletInitializer` says 

> Note that a WebApplicationInitializer is only needed if you are building a war file and deploying it. If you prefer to run an embedded container then you won't need this at all.

deepu105:
Ok ill try to check it

deepu105:
First PR is done. we need to do the property replace from maven and gradle. for gradle its very easy as I said earlier. But I wont have time for few days for that

jdubois:
@erikkemperman @deepu105 as work as begun on the property replace, are you OK to close this PR?

erikkemperman:
Yup, I guess that's the way to go. Closing.

PR #3231, opened 2016-03-21
#2244 Add integration tests for elasticsearch

lordlothar99:
- Removed EntityIndexInitializer beans (too slow, and cause of random
errors)
- reset all indexes in IndexReinitializer


geraldhumphries:
Does this still recreate each index with the proper mappings?

deepu105:
@lordlothar99 can you confirm what @geraldhumphries asked, I want to merge this today if possible.

deepu105:
Ok tests are passing so thats good enough. If you want to improve anything we can have it as another PR

PR #3234, opened 2016-03-22
Travis: change baseName and force no insight

pascalgrimaud:
@jdubois : I made some change, so it will be easy for you to filter/track these projects by baseName / rememberMeKey
- change all baseName to `travis*`
- all `"rememberMeKey": "f19f0827c622eb9b81f5227a869ccd44932d78eb"`
- add no-insight json to travis build

We already use [`yo jhipster --force --no-insight`](https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/01-generate-project.sh#L14) but to be sure for JHipster stats

### DIFF
--- a/travis/configstore/insight-generator-jhipster.json
+++ b/travis/configstore/insight-generator-jhipster.json
@@ -1,0 +1,4 @@
+{
+	"clientId": 495049364328,
+	"optOut": true
+}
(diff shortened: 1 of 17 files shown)


jdubois:
Thanks! I see our discussion from yesterday evening made you work all night :-)

PR #3241, opened 2016-03-22
Fix the copy of .gitignore in docker-compose sub generator

pascalgrimaud:
I rename .gitignore -> gitignore
Without this modification, the hidden files will not be present in the npm package -> that means the next release!!
I already met this issue here : https://github.com/jhipster/generator-jhipster-module/issues/1


jdubois:
Good catch!!!

PR #3242, opened 2016-03-23
Added JSR-303 validation rules to saveAccount method. Fix #3042

### LINKED ISSUE #3042: saveAccount REST API doesn't use JSR 303 validation rules on DTO Layer

REST API for saveAccount throws internal errors in case of validation problems (produced by hibernate check of JSR 303 annotations)

AccountResource#saveAccount should have the same validation-rules applied to the given DTO like AccountResource#registerAccount (except password)

##### **JHipster Version(s)** 

2.27.0

##### **Reproduce the error** 

Using api/account with a REST Client & invalid POST body (e.g. missing values)

##### **Suggest a Fix** 

Use different DTOs for registration and changing account settings. Annotate POST Body Entity with @Valid annotation to enable JSR 303 validation rules.

gandrade:
Added JSR-303 validation rules to saveAccount method. Fix #3042

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/dto/_ManagedUserDTO.java
+++ b/generators/server/templates/src/main/java/package/web/rest/dto/_ManagedUserDTO.java
@@ -2,12 +2,19 @@ package <%=packageName%>.web.rest.dto;
 <% if (databaseType == 'mongodb' || databaseType == 'sql') { %>
 import java.time.ZonedDateTime;<% } %>
 
+import java.util.Set;
+
 import <%=packageName%>.domain.User;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 /**
  * A DTO extending the UserDTO, which is meant to be used in the user management UI.
  */
 public class ManagedUserDTO extends UserDTO {
+
+    public static final int PASSWORD_MIN_LENGTH = 4;
+    public static final int PASSWORD_MAX_LENGTH = 100;
 
     private <% if (databaseType == 'mongodb' || databaseType == 'cassandra') { %>String<% } else { %>Long<% } %> id;<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>
 
@@ -16,6 +23,10 @@ public class ManagedUserDTO extends UserDTO {
     private String lastModifiedBy;
 
     private ZonedDateTime lastModifiedDate;<% } %>
+
+    @NotNull
+    @Size(min = PASSWORD_MIN_LENGTH, max = PASSWORD_MAX_LENGTH)
+    private String password;
 
     public ManagedUserDTO() {
     }
@@ -26,6 +37,17 @@ public class ManagedUserDTO extends UserDTO {
         this.createdDate = user.getCreatedDate();
         this.lastModifiedBy = user.getLastModifiedBy();
         this.lastModifiedDate = user.getLastModifiedDate();<% } %>
+        this.password = null;
+    }
+
+    public ManagedUserDTO(<% if (databaseType == 'mongodb' || databaseType == 'cassandra') { %>String<% } else { %>Long<% } %> id, String login, String password, String firstName, String lastName,
+                          String email, boolean activated, String langKey, Set<String> authorities <% if (databaseType == 'mongodb' || databaseType == 'sql') { %>, ZonedDateTime createdDate, String lastModifiedBy, ZonedDateTime lastModifiedDate <% } %>) {
+        super(login, firstName, lastName, email, activated, langKey, authorities);
+        this.id = id;<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>
+        this.createdDate = createdDate;
+        this.lastModifiedBy = lastModifiedBy;
+        this.lastModifiedDate = lastModifiedDate;<% } %>
+        this.password = password;
     }
 
     public <% if (databaseType == 'mongodb' || databaseType == 'cassandra') { %>String<% } else { %>Long<% } %> getId() {
@@ -59,6 +81,10 @@ public class ManagedUserDTO extends UserDTO {
     public void setLastModifiedDate(ZonedDateTime lastModifiedDate) {
         this.lastModifiedDate = lastModifiedDate;
     }<% } %>
+
+    public String getPassword() {
+        return password;
+    }
 
     @Override
     public String toString() {
--- a/generators/server/templates/src/test/java/package/web/rest/_AccountResourceIntTest.java
+++ b/generators/server/templates/src/test/java/package/web/rest/_AccountResourceIntTest.java
@@ -9,6 +9,7 @@ import <%=packageName%>.repository.UserRepository;
 import <%=packageName%>.security.AuthoritiesConstants;
 import <%=packageName%>.service.MailService;
 import <%=packageName%>.service.UserService;
+import <%=packageName%>.web.rest.dto.ManagedUserDTO;
 import <%=packageName%>.web.rest.dto.UserDTO;
 import org.junit.Before;
 import org.junit.Test;
@@ -146,16 +147,20 @@ public class AccountResourceIntTest <% if (databaseType == 'cassandra') { %>exte
     @Test<% if (databaseType == 'sql') { %>
     @Transactional<% } %>
     public void testRegisterValid() throws Exception {
-        UserDTO u = new UserDTO(
+        ManagedUserDTO u = new ManagedUserDTO(
+            null,                   // id
             "joe",                  // login
             "password",             // password
             "Joe",                  // firstName
             "Shmoe",                // lastName
             "joe@example.com",      // e-mail
             true,                   // activated
-            "<%= nativeLanguageShortName %>",                   // langKey
-            new HashSet<>(Arrays.asList(AuthoritiesConstants.USER))
-        );
+            "<%= nativeLanguageShortName %>",               // langKey
+            new HashSet<>(Arrays.asList(AuthoritiesConstants.USER))<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>,
+            null,                   // createdDate
+            null,                   // lastModifiedBy
+            null                    // lastModifiedDate 
+        <% } %>);
 
         restMvc.perform(
             post("/api/register")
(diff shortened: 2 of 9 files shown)


gandrade:
Checking...

gandrade:
@deepu105 please, let me know which changes need to be reviewed.

deepu105:
no its a label so that our team knows that this needs to be reviewed. Dont
worry

Thanks & Regards,
Deepu

deepu105:
@jhipster/developers LGTM can some one take a second look

notflorian:
:+1: The code looks good to me, I will take some time to test this later today.

My only concern is does this mean the password is now added in the json response of the /api/users/ and /api/users/{login} endpoints? If that so, could it become a security issue?

notflorian:
I did the test, and with your PR the encrypted passwords are indeed returned in /api/users/ and /api/users/{login} endpoints, accessible without admin rights. This is in my opinion a big security issue.

Moreover, with your PR I didn't succeded to login with login "user" and default password "user", as it is documented [here](http://jhipster.github.io/security/). I think this is because the password "user" is 4 letters long, and this is too short for the validation rule.

gandrade:
About the encrypted passwords, you are 100% right, this is a big security issue and I'll work on it.
Regarding the password 'user', shall we change the `PASSWORD_MIN_LENGTH` to 4 digits in order to be more compliant with `user` user? As soon as you have decided about password issue, please let me know to work on it too. 

Thanks in advance.
Gustavo Andrade

notflorian:
I will take time this week-end to double check the login issue with "user" to be absolutely certain it is indeed a password size issue.

If it is the case, @jdubois do you think it would be ok to change the default password of the "user" to a 5 letters password, if we document this change in the security page? 
As the passwords are supposed to be changed anyway for security reasons, I think it should not be too impactfull.

gandrade:
OK @notflorian. Keep me posted!

Thanks.

Gustavo Andrade

gandrade:
Folks, I've updated the code to avoid to show encrypted passwords on `/api/users` and `/api/users/{login}` endpoint. Please, let me know any issue about it. 

Thanks in advance.

gandrade:
Just a heads-up... After the last code changes, I've tried to login with `user` user and didn't have any  issues with it.

generators/server/templates/src/main/java/package/web/rest/dto/_ManagedUserDTO.java
```diff
@@ -2,13 +2,20 @@
...
 import <%=packageName%>.domain.User;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
 
 /**
  * A DTO extending the UserDTO, which is meant to be used in the user management UI.
  */
 public class ManagedUserDTO extends UserDTO {
 
+    public static final int PASSWORD_MIN_LENGTH = 5;
```
notflorian: I retested the login with "user" and it seems to be working now, but to avoid backward compatibility issue maybe it would be best to leave de min password length to 4, like it was before in the UserDTO.

generators/server/templates/src/main/java/package/web/rest/dto/_ManagedUserDTO.java
```diff
@@ -60,13 +82,18 @@ public void setLastModifiedDate(ZonedDateTime lastModifiedDate) {
...
+
     @Override
     public String toString() {
         return "ManagedUserDTO{" +
             "id=" + id +<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>
             ", createdDate=" + createdDate +
             ", lastModifiedBy='" + lastModifiedBy + '\'' +
-            ", lastModifiedDate=" + lastModifiedDate +<% } %>
+            ", lastModifiedDate=" + lastModifiedDate + '\'' +<% } %>
+            ", password='" + password +
```
notflorian: I don't think it's a good idea to append the encrypted password in the toString method, it could easily be printed in application logs by mistake.

generators/server/templates/src/test/java/package/web/rest/_AccountResourceIntTest.java
```diff
@@ -146,16 +147,20 @@ public void testGetUnknownAccount() throws Exception {
     @Test<% if (databaseType == 'sql') { %>
     @Transactional<% } %>
     public void testRegisterValid() throws Exception {
-        UserDTO u = new UserDTO(
+        ManagedUserDTO u = new ManagedUserDTO(
```
notflorian: Very small improvement : even if it was already here before, could you take the opportunity of your PR to rename the variable u to user in this file? I found 1 letter variable misleading and hurtful to the code readability.

generators/server/templates/src/test/java/package/web/rest/_AccountResourceIntTest.java
```diff
@@ -243,20 +288,24 @@ public void testRegisterEmailEmpty() throws Exception {
...
+            new HashSet<>(Arrays.asList(AuthoritiesConstants.USER))<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>,
+            null,                   // createdDate
+            null,                   // lastModifiedBy
+            null                    // lastModifiedDate 
+        <% } %>);
 
         // Duplicate login, different e-mail
-        UserDTO dup = new UserDTO(u.getLogin(), u.getPassword(), u.getLogin(), u.getLastName(),
-            "alicejr@example.com", true, u.getLangKey(), u.getAuthorities());
+        ManagedUserDTO dup = new ManagedUserDTO(u.getId(), u.getLogin(), u.getPassword(), u.getLogin(), u.getLastName(),
```
notflorian: Same here, dup means nothing, could you please rename this variable to duplicatedUser or duplicate?

gandrade:
Got your points! :)

generators/server/templates/src/main/java/package/web/rest/dto/_ManagedUserDTO.java
gandrade: @deepu105 and @notflorian Just a quick question... I was looking to `generators/client/templates/src/main/webapp/app/account/register/register.html` file and I've found this snippet:
```
<div class="form-group">
                    <label class="control-label" for="password" translate="global.form.newpassword">New password</label>
                    <input type="password" class="form-control" id="password" name="password" placeholder="{{'global.form.newpassword.placeholder' | translate}}"
                           ng-model="vm.registerAccount.password" ng-minlength=5 ng-maxlength=50 required>
                    <div ng-show="form.password.$dirty && form.password.$invalid">
                        <p class="help-block"
                               ng-show="form.password.$error.required" translate="global.messages.validate.newpassword.required">
                            Your password is required.
                        </p>
                        <p class="help-block"
                               ng-show="form.password.$error.minlength" translate="global.messages.validate.newpassword.minlength">
                            Your password is required to be at least 5 characters.
                        </p>
                        <p class="help-block"
                               ng-show="form.password.$error.maxlength" translate="global.messages.validate.newpassword.maxlength">
                            Your password cannot be longer than 50 characters.
                        </p>
                    </div>
                    <password-strength-bar password-to-check="vm.registerAccount.password"></password-strength-bar>
                </div>
```
In this file, the min length for password field is 5. Is that correct? 
If we leave the min password length to 4, in order to avoid backward compatibility, we will have different validation rules to password field for back-end and front-end Someone could clarify this question?

notflorian: Very well seen, indeed! :+1: 

@mraible You modified this min value in the commit 9460a762c0cc90c557f621347b8849ed52c0005a, do you remember why did you had to make this change? Something with JWT perhaps?

mraible: @notflorian Yes, I changed it because the default user/user authentication didn't work with JWT.

gandrade: So, should we keep the min length password to 5, right?

mraible: The easiest thing to do is change the min-length to 4. If we want to keep it at 5, we should change the default password for `user` to be 5 characters.

gandrade: So, if it is not a problem... I can just change the min-length password to 4.

gandrade: In register.html file

gandrade: In this scenario, what is the better approach? 5 or 4? I can do what you guys prefer.

gandrade: Got your point @mraible. I'll provide the changes! Thanks in advance.

deepu105: Yes change to 4 everywhere yo be consistent

notflorian:
There is a lot of unrelated commit in your PR, could you rebase your branch instead of merging it? Thanks.

gandrade:
@notflorian Just to double check.. Is there a specific `git rebase` command to do this or just a `git rebase -i`?

notflorian:
I usually do in my feature branch: git fetch upstream; git rebase upstream/master; git push -f

gandrade:
I'll try do this and sorry for the mess. Thanks in advance

notflorian:
Thanks @gandrade, your code now looks very good. I'm in favor for merging this PR as soon as the commits are squashed to have a cleaner history. :+1:

gandrade:
@notflorian In this situation, what more I need to do?

gandrade:
@notflorian Are you talking about squash all commits into a single one?

notflorian:
1 or 2, yes : for exemple the first with the message "Added JSR-303 validation rules to saveAccount method" and the second with the message "Adjusted min-length for password field"

gandrade:
@notflorian Build failed but looks weird: https://travis-ci.org/jhipster/generator-jhipster/builds/120081398 How can I proceed to start a new Travis build?

notflorian:
Close & Reopen the PR ;)

gandrade:
All green guys! Thank you!

gandrade:
@notflorian we are good to merge now! :)

deepu105:
@jhipster/developers can some one test and merge this, I wont have much time till weekend

jdubois:
Looks good to me!

Issue #2674, opened 2016-01-23
Email user not null in database model


phouverneyuff:
The jhipster default code should generate @NotNull above the email property in User.java and put nullable="false" in liquibase

I saw in register.html that the field email is required. Is this a bug?

deepu105:
email should be required everywhere can you PR to correct this?

deepu105:
@moifort does this affect social login in anyway?

moifort:
I will take a look. 

Envoyé de mon iPhone

moifort:
It's ok! it's return an error if there is not an email. You can make a PR!

phouverneyuff:
OK. I will.

phouverneyuff:
I see another problem. 
The field "activated" is missing @NotNull too.

```xml
      <createTable tableName="jhi_user">
            <column name="id" type="bigint" autoIncrement="${autoIncrement}">
                <constraints primaryKey="true" nullable="false"/>
            </column>
            <column name="login" type="varchar(<% if (enableSocialSignIn) { %>100<% } else { %>50<% } %>)">
                <constraints unique="true" nullable="false"/>
            </column>
            <column name="password_hash" type="varchar(60)"/>
            <column name="first_name" type="varchar(50)"/>
            <column name="last_name" type="varchar(50)"/>
            <column name="email" type="varchar(100)">
                <constraints unique="true" nullable="false"/>
            </column>
            <column name="activated" type="boolean" valueBoolean="false">
                <constraints nullable="false" />
            </column>
            <column name="lang_key" type="varchar(5)"/>
            <column name="activation_key" type="varchar(20)"/>
            <column name="reset_key" type="varchar(20)"/>
            <column name="created_by" type="varchar(50)">
                <constraints nullable="false"/>
            </column>
            <column name="created_date" type="timestamp" defaultValueDate="${now}">
                <constraints nullable="false"/>
            </column>
            <column name="reset_date" type="timestamp">
                <constraints nullable="true"/>
            </column>
            <column name="last_modified_by" type="varchar(50)"/>
            <column name="last_modified_date" type="timestamp"/>
        </createTable>
```

I'm doing the PR to fix all it.

PR #3250, opened 2016-03-24
add separate changelogs for constraints

### LINKED ISSUE #3067: Separate changelogs for foreign key updates

The entity generator manage every entity changelog independently. So when you have many tables (i have 50, but depends your database layout), you get mass foreign key error from liquibase changelog as many tables doesn't existe while creating/updating the database.

I'm using jhispter-uml, but it's a jhipster issue as the generator must "know" that he is working on an entity list and not a single entity (see suggested fix bellow), nop ?

##### **JHipster Version(s)** 

JHipster 2.27, using jhipster-uml 1.6.1

##### **Suggest a Fix** 

When entity generator get a list of entity to build (from jhipster-uml), constraint management can be separated in a dedicated changelog file, last called by liquibase ?

deepu105:
Currently our entity liquibase changelogs will create tables and constraints in a same changelog file which makes it harder to add many entities together as they have to be created in order to avoid errors.
This PR splits the changelogs into 2
1. current changelog will only have table creation and primary key additions
2. if an entity has one or more foreign key constraint, a new changelog with all constraints of that entity will be created and will be added to the liquibase master.xml

this ensures that all constraints are executed by liquibase after table creation and hence no more liquibase errors if entity order is different. This means we no longer have to worry about entity creation order which is cleaner and thus people can use JDL and entity sub gen much more effectively.

I have tested this with application creation and it works well.

@jhipster/developers Can some one check if this works with the liquibase goal plz? Im bit busy.

FYI I have also refactored some iterations in entity subgen to make things a bit faster.

close #3067


jdubois:
This would be a really great improvement. We need to test it well, and also to update the documentation accordingly, as this removes a lot of our older constraints.
@deepu105 did you check that this is working well with the "liquibase:diff" goal? I've spent a lot of time to ensure that what we generate is correct with that goal, so that after generating several entities, if you type "mvn liquibase:diff", you don't have any diff at all.

deepu105:
I have tested only app deployment and tests, havent test the liquibase goal yet. need help on that.
But looking at the liquibase goal I dont think this should affect it as the goal doesnt rely on file name or the syntax of changelog

deepu105:
ok I checked the liquibase diff goal with gradle and it seems to be working fine, but I would really like to see a confirmation from someone who is very familiar with that goal as I have rarely used it myself

ruddell:
I generated two entities with a one-to-many relationship.  The liquibase changelogs for the entities look great, but when I run 'mvn liquibase:diff' I get the following changelog for User:
`   <changeSet author="ruddell (generated)" id="1458876730446-1">
      <dropNotNullConstraint columnDataType="varchar(100)"
                             columnName="email"
                             tableName="jhi_user"/>
   </changeSet>
</databaseChangeLog>
`
I'm not sure where this is coming from.

deepu105:
the changelog doesnt look related to this. can you try on the master if you
get the same changelog or not?

Thanks & Regards,
Deepu

deepu105:
Are you using H2 DB?

Thanks & Regards,
Deepu

ruddell:
You are correct, this happens for the current master as well.  It does not happen on v2.27.2.  The difference is that the default jhi_user table for v2.27.2 allows a null email.

To generate the project, I am using default options for everything except MySQL in both dev and prod.

ruddell:
I tracked down the commit where it was added (confusing because the file was moved).  It was brought up in Issue #2674 and Pulled in #2770.

deepu105:
Can you open a new issue for that as its unrelated to this PR

ruddell:
Opened a new issue.  Otherwise this PR worked perfectly and would fix some issues people are having in gitter

deepu105:
@jdubois can we mere this?

jdubois:
I have no commuter for the next 3 days so I can't test... This is a very important feature, my main concern is the "liquibase:diff" goal -> if that's fine yes let's merge this.

deepu105:
There seems to be some minor discrepancy in the liquibase diff on current
master,  which is un related to this but if you would like to check
yourself that fine we are not in a hurry for another release anyway :P

MathieuAA:
JHipster UML's code could be simplified thanks to this, great work Deepu!

ruddell:
I fixed the other issue related to mvn liquibase:diff in #3275 so this should be good

deepu105:
Thank you

deepu105:
its rebased and ready now. Im not merging it myself as I want @jdubois or any other @jhipster/developers to verify and merge

deepu105:
@MathieuAA yes now from JDL and JHUML we dont have to worry about entity order or relationship order @cbornet the `--with-entities` logic also can be simplified I guess

deepu105:
and we should update our docs as well as this removes some of the limitations we documented

cbornet:
@deepu105 with this, we can obviously simplify the code of [getExistingEntities](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js#L1030)

deepu105:
@jdubois are you ok with this? can we merge this?

jdubois:
I just don't have the time at the moment -> yes I agree with the idea, so let's merge this

MathieuAA:
It's a bit late, but I have a question for @deepu105.
Thanks to this PR, the entities are generated first and then the relationships when a user generates an entity with JHipster.

What happens if the user wants to change his/her mind about a relationship? 
For instance, say we have two entities A and B:
```
entity A
entity B {
  code String
}

relationship OneToOne {
  A{b(code)} to B{a}
}
```

Here b:B is the injected field in A, and is linked by it's `superImportantCode` attribute.
a:A with it's generated `id` in B.
Nothing fancy here, right?

What if I realize I was completely wrong, and I decide to change my entity to fix my mistake so that the code looks like (code is renamed here):
```
entity A
entity B {
  superImportantCode String
}

relationship OneToOne {
  A{b(superImportantCode)} to B{a}
}
```
The example could be further developed if a user makes a typo in an attribute's name, or if I decide to use another field from A for the joint (I'd have to add the field, and change the relationship).
How does this new system work? _I know how it could_, but I'm wondering about how it is currently designed. I'm not only asking how it behaves with the JDL, but with the Q&As too.

Thanks and sorry for the late question.

deepu105:
For changes in attributes and relationship as long its regeneration the
same files will be overwritten. I need to check and confirm this. Ideally
thats how it should behave so in any combination or case there would be
maximum of two changelog files only.

jdubois:
Indeed, I'm not sure this works well with the Liquibase Maven/Gradle goal.

On Wed, Apr 13, 2016 at 2:41 AM, Deepu K Sasidharan <
notifications@github.com> wrote:

> For changes in attributes and relationship as long its regeneration the
> same files will be overwritten. I need to check and confirm this. Ideally
> thats how it should behave so in any combination or case there would be
> maximum of two changelog files only.
>
> On Wed, 13 Apr 2016 5:12 am Mathieu ABOU-AICHI, <notifications@github.com>
> wrote:
>
> > It's a bit late, but I have a question for @deepu105
> > <https://github.com/deepu105>.
> > Thanks to this PR, the entities are generated first and then the
> > relationships when a user generates an entity with JHipster.
> >
> > What happens if the user wants to change his/her mind about a
> > relationship?
> > For instance, say we have two entities A and B:
> >
> > entity A
> > entity B {
> > code String
> > }
> >
> > relationship OneToOne {
> > A{b(code)} to B{a}
> > }
> >
> > Here b:B is the injected field in A, and is linked by it's
> > superImportantCode attribute.
> > a:A with it's generated id in B.
> > Nothing fancy here, right?
> >
> > What if I realize I was completely wrong, and I decide to change my
> entity
> > to fix my mistake so that the code looks like (code is renamed here):
> >
> > entity A
> > entity B {
> > superImportantCode String
> > }
> >
> > relationship OneToOne {
> > A{b(superImportantCode)} to B{a}
> > }
> >
> > The example could be further developed if a user makes a typo in an
> > attribute's name, or if I decide to use another field from A for the
> joint
> > (I'd have to add the field, and change the relationship).
> > How does this new system work? *I know how it could*, but I'm wondering
> > about how it is currently designed. I'm not only asking how it behaves
> with
> > the JDL, but with the Q&As too.
> >
> > Thanks and sorry for the late question.
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/jhipster/generator-jhipster/pull/3250#issuecomment-209104988
> >
>
> >
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/jhipster/generator-jhipster/pull/3250#issuecomment-209166630>
>



-- 
Julien Dubois

Twitter: @juliendubois <http://twitter.com/#!/juliendubois>
Phone: +33 (0)6 25 02 34 18

deepu105:
Ok so checked this and everything works as expected except one scenario

1. _add new entity with fields and **one to many** | **non owner** relationship:_ only the table creation change-log is created in single file **(this behavior is same as before)**
2. _add new entity with fields and **many to one** | other **owner**  relationship:_ the table creation change-log and the constraints change-log is created in two separate files **(this is new behavior)**
3. _add more fields to existing entity:_ the fields are added to the existing table creation change-log **(this behavior is same as before)**
4. _add new **one to many** or **non owner** relationships to existing entity:_ the existing table creation change-log is updated **(this behavior is same as before)**
5. _add new **many to one** or **owner** relationships:_ the existing table creation change-log and existing constraints change-log is updated **(this behavior is same as before)**
6. _remove fields from existing entity:_ the fields are removed from the existing table creation change-log **(this behavior is same as before)**
7. _remove relationships from existing entity:_ the existing table and constraint change-log files are updated accordingly**(this is new behavior)**
8. _remove all relationships from an existing entity:_ the table changelog is updated properly but the constraints change-log is not removed**(this is new behavior and the only one that fails currently)** 

When I checked I didn't see any difference in behavior of liquibase:diff goal and in theory I dont think this will cause any issue except for the last scenario above. Ill see if I can find a fix for that

PR #3254, opened 2016-03-25
enable swagger via application property

deepu105:
This is as per request by @cbornet earlier.

swagger can be enabled or disabled from `application-*.yml` as well. its disabled by default in production to retain current behavior

so if someone wants swagger in production they can enable it by setting `jhipster.swagger.enabled: true` and then they can use the `no-swagger` switch for finer control at run time

### DIFF
--- a/generators/server/templates/src/test/resources/config/_application.yml
+++ b/generators/server/templates/src/test/resources/config/_application.yml
@@ -14,6 +14,9 @@
 eureka:
     client:
         enabled: false
+    instance:
+        appname: <%= baseName %>
+        instanceId: <%= baseName %>:${spring.application.instance_id:${random.value}}
 
 <%_ } _%>
 spring:
--- a/generators/server/templates/src/test/resources/config/_application.yml
+++ b/generators/server/templates/src/test/resources/config/_application.yml
@@ -17,6 +17,8 @@ eureka:
 
 <%_ } _%>
 spring:
+    application:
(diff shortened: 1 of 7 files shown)


deepu105:
@gmarziou are you ok with this?

generators/server/templates/src/test/resources/config/_application.yml
```diff
@@ -14,9 +14,14 @@
 eureka:
     client:
         enabled: false
+    instance:
```
deepu105: @PierreBesson this was not present in master but the travis builds are passing there but its failing for my PR thats why I have added this. any idea why

generators/server/templates/src/test/resources/config/_application.yml
```diff
@@ -14,9 +14,14 @@
...
+        appname: <%= baseName %>
+        instanceId: <%= baseName %>:${spring.application.instance_id:${random.value}}
 
 <%_ } _%>
 spring:
+    application:
+        name: <%= baseName %>
```
deepu105: same here app name was not present in master but now its complaining!!

generators/server/templates/src/test/resources/config/_application.yml
PierreBesson: I don't understand either.

deepu105:
any way it seems fine now

deepu105:
Ok Im merging this

PR #3277, opened 2016-03-28
Fix auth-expired redirect after one failed login attempt

ruddell:
If a user enters the wrong password on the way to a secured state, the login window reappears but $rootScope.previousStateName changes to 'accessdenied'

This change prevents prevents the 'accessdenied' state from being stored in the previousStateName.  It fixes the flow so that if a user enters an incorrect password but then fixes his mistake on another login attempt, he is correctly redirected to the destination state.


deepu105:
Good one

PR #3286, opened 2016-03-29
Cluster Mode for MongoDB

wmarques:
As the title says, I'm adding a cluster support when the user wants to use a MongoDB database. I used [this guide](https://sebastianvoss.com/docker-mongodb-sharded-cluster.html) to help me configure MongoDB.

### What it actually does :
* Create a replicaset of X node
* Create a server config node
* Add a shard of this replicaset into the Mongo router (`mongos`)

### How to use it :
1. Create an app with MongoDB Cluster Mode (I added a question)
2. Build the image : `docker-compose -f src/main/docker/mongo.yml build`
3. Run the DB : `docker-compose -f src/main/docker/mongo.yml up -d `
4. Scale the mongodb-node : `docker-compose -f src/main/docker/mongo.yml scale appname-mongodb-node=3`
Please note that[ you have to use an odd number of node](https://docs.mongodb.org/manual/core/replication-introduction/).
5. Init the replicaset by running `docker exec -it docker_mongodb-node_1 mongo --eval 'var param=3, folder="docker"' init_replicaset.js`
param is the number of nodes, folder is the folder where the docker-compose YML file is located (I need this to complete the container name in JS file)
6. Init the shard by running `docker exec -it appname-mongodb mongo --eval 'sh.addShard("rs1/docker_mongodb-node_1:27017")'`
7. Start the app : `docker-compose -f src/main/docker/app.yml up -d`

This is a bit hacky but maybe someone will file a cleaner way to do this, for example to get automatically the docker containers name in the init_replicaset.js file.

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -278,6 +278,9 @@ module.exports = JhipsterClientGenerator.extend({
             if (configOptions.languages != null) {
                 this.languages = configOptions.languages;
             }
+            if (configOptions.mongoCluster != null) {
+                this.mongoCluster = configOptions.mongoCluster;
+            }
 
             // Make dist dir available in templates
             if (configOptions.buildTool === 'maven') {
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -233,9 +233,17 @@ module.exports = yeoman.Base.extend({
                     var databaseYaml = jsyaml.load(this.fs.read(path + '/src/main/docker/' + database + '.yml'));
                     var databaseYamlConfig = databaseYaml.services[this.appConfigs[i].baseName.toLowerCase() + '-' + database];
                     delete databaseYamlConfig.ports;
-                    if(this.appConfigs[i].devDatabaseType === 'cassandra') {
+                    if(database === 'cassandra') {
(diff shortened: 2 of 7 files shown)


jdubois:
Excellent!!
Can you also update the documentation? I guess you can just `docker-compose scale` the node?

pascalgrimaud:
Shouldn't we follow what it's done with Cassandra, to be consistent ?
- no more question in main generator, but this question can be added in sub generator
- mongodb.yml => simple mongodb
- mongodb-cluster.yml (or another name) => cluster

pascalgrimaud:
Anyway, it's very good job @wmarques ! :+1:

jdubois:
Yes it should be the same as what we have with Cassandra.
And then we just need to add CitusDB -> https://www.citusdata.com/blog/17-ozgun-erdogan/403-citus-unforks-postgresql-goes-open-source

docker-compose scale everything :-)

deepu105:
Yes i would also prefer it to be same as cassandra without the additional
question

wmarques:
Ok I'll update this and do like Cassandra !

PR #3294, opened 2016-03-29
Cleanup bower js and gulpfile js

andidev:
- exclude bootstrap.js in bower instead of in gulpfile
- dont exclude angular-i18n in wiredep task in gulpfile since there is nothing to exclude


deepu105:
Oh nice, i totally forgot about the bower part when i did that

PR #3296, opened 2016-03-30
remove angular-scenario in bower.js from devDependencies since it is deprecated

andidev:
Not sure that this is okey to do but I cant find where angular-scenario is used and it works without it and it is deprecated since angular 1.2. Though lets see if the CI builds pass first.

Could someone comment on this if this is no problem? If it's a problem please close this issue.


deepu105:
Its for testing and I guess we are not using it at the moment so should be ok to remove

wmarques:
+1

PR #3302, opened 2016-03-30
Add --jhi-prefix option in main generator

wmarques:
Suggested in  #3054 by @deepu105 and @kkorada.
Now the user can pass the --jhi-prefix flag in the main generator : `yo jhipster --jhi-prefix foo` instead of passing it in the client sub-generator.


kkorada:
Is it also added in jhipsterVar of generator:modules ?

wmarques:
No I'm working on it.

PR #3309, opened 2016-03-31
Remove messageformat locales lang.js files since they are deprecated

### LINKED ISSUE #3264: Uncaught TypeError: Cannot set property 'en' of undefined en.js:1

Error messages in console 
```
Uncaught TypeError: Cannot set property 'en' of undefined    en.js:1
Uncaught TypeError: Cannot set property 'fr' of undefined    fr.js:1
```

##### **JHipster Version(s)** 

3.0.0

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.0.0",
    "baseName": "hipster",
    "packageName": "io.example.hipster",
    "packageFolder": "io/example/hipster",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "searchEngine": "no",
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "jwtSecretKey": "6b7351dc7535683df733688a4c69v0563e90178w",
    "useSass": false,
    "testFrameworks": [],
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "fr"
    ]
  }
}
```

Chrome

##### **Suggest a Fix** 

It seems like it is unnecessary to load `lang.js` (i.e `en.js`, `fr.js`) in `index.html`. If I don't import those files, the error messages doesn't show up and no problem with internationalization.

andidev:
fix #3264

### DIFF
--- a/generators/client/templates/_bower.json
+++ b/generators/client/templates/_bower.json
@@ -32,6 +32,7 @@
     "bootstrap-ui-datetime-picker": "2.2.3",
     "jquery": "2.2.2",
     "json3": "3.3.2",
+    "messageformat": "0.3.1",
     "modernizr": "3.3.1",
     "ng-file-upload": "12.0.4",
     "ngInfiniteScroll": "1.2.2",
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -33,35 +33,6 @@ function Generator() {
 util.inherits(Generator, yeoman.Base);
 
 /**
- * A a new message format to the application, in the bower.json file.
- *
- * This is used for internationalization.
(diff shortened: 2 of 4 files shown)


deepu105:
@andidev hope this doesn't affect https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/gulpfile.js#L112 this in anyway

andidev:
Angular locale files are still included. Only message format locale files are removed. So it should not cause any problems there.

PR #3310, opened 2016-04-01
add missing vm prefixes and change register translation

ruddell:
Add more vm's to the app where they are needed.  

Also the register link on the login window was overwritten by the translation which used ui-sref.  This changes the state to register but the login dialog stays open.  I changed the translation to use the vm.register function declared in the LoginController.  This means I had to add a register method to the HomeController so the same translation can be used to get to the register state.  The only other way I could think to do this is copy the translation into the login.json set.

### DIFF
--- a/generators/languages/templates/src/main/webapp/i18n/ca/_global.json
+++ b/generators/languages/templates/src/main/webapp/i18n/ca/_global.json
@@ -49,7 +49,7 @@
         "messages": {
             "info": {
                 "authenticated": "Si vol <a class=\"alert-link\" href=\"\" ng-click=\"vm.login()\">autenticar-se</a>, pot usar els comptes per defect:<br/>- Administrador (nom d'usuari=\"admin\" i contrasenya=\"admin\") <br/>- Usuari (nom d'usuari=\"user\" i contrasenya=\"user\").",
-                "register": "No disposa d'un compte encara? <a class=\"alert-link\" ui-sref=\"register\">Registri un nou compte</a>"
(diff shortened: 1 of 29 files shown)


generators/languages/templates/src/main/webapp/i18n/ca/_global.json
```diff
@@ -49,7 +49,7 @@
         "messages": {
             "info": {
                 "authenticated": "Si vol <a class=\"alert-link\" href=\"\" ng-click=\"vm.login()\">autenticar-se</a>, pot usar els comptes per defect:<br/>- Administrador (nom d'usuari=\"admin\" i contrasenya=\"admin\") <br/>- Usuari (nom d'usuari=\"user\" i contrasenya=\"user\").",
-                "register": "No disposa d'un compte encara? <a class=\"alert-link\" ui-sref=\"register\">Registri un nou compte</a>"
+                "register": "No disposa d'un compte encara? <a class=\"alert-link\" href=\"\" ng-click=\"vm.register()\">Registri un nou compte</a>"
```
deepu105: why is this changed to ng-click? ui-sref is more appropriate here

deepu105:
I dont understand the problem you stated?

ruddell:
If you generate an app with translation enabled, the register link on the login dialog is broken.  Without translation it uses ng-click.

Ui-sref is more appropriate but it leaves the login dialog open, just changing the state in the background.

ruddell:
I'm closing this PR to just open an issue relating to the login dialog's Register link.  I'll separate the small vm fixes into a new one.

PR #2311, opened 2015-11-09
Jwt changes for issue #2100

amit2103:
Only java code changes, will submit the JS changes by this weekend.

PR #3324, opened 2016-04-02
add UAA auth server using springs oauth2 solution for microservice

xetys:
this update decouples the user management from gateway into a JHipster UAA as a separate service (as new selectable application type). This service does all the usual user management stuff (without client) and also acts as oauth2 authorization server as well as resource server for users, accounts etc.

This setup setups two oauth2 clients in memory, one for usage inside the angular app, one for upcomming inter-service communication via client credentials.

This PR contains also needed modification on the client application and docker-compose sub-generator.

The existing tests are changed to these changes and are passing.

This PR is not fully tested on all JHipsters options, and things may be done more efficient, e.g. providing a clientDetailsService instead of in memory, ribbon + hystrix generators using OAuth2RestTemplate etc.


gmarziou:
Wow, thanks for this amazing work.
I'm sure it will take us some time to review it.

xetys:
i am working on the failed travis build :/

deepu105:
Ok ill test this and see how it works

xetys:
i hope my last commit makes travis turn green, then you can test

xetys:
i hope my last commit will turn travis green, then you can test :)

xetys:
Damn it, now it breaks on entity relationships on gateway. Maybe I find time to solve this tonight.

xetys:
there is one test failing on gateway, i will check this tonight if i'll finde more time

jdubois:
@xetys I'm sorry but I'm afraid #3326 might have some consequences on this PR -> from my point of view it should help in the end, so that shouldn't be too annoying.
Of course I'll do my best to help here.

xetys:
hope we can proceed here, if my solution in #3329 are acceptable for you

jdubois:
I can't engage on a date, but yes I want to have this all working together.
I'm just lacking time at the moment, and as this is quite sensitive I want to be sure it's all OK.

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
```diff
@@ -0,0 +1,142 @@
+package <%=packageName%>.config;
+
+import de.stytex.foobar.security.AuthoritiesConstants;
```
jdubois: Your import should be a variable here

generators/app/index.js
```diff
@@ -255,6 +259,11 @@ module.exports = JhipsterGenerator.extend({
                 this.generatorType = 'server';
                 this.skipUserManagement = configOptions.skipUserManagement = true;
             }
+            if (this.applicationType == 'uaa') {
+                this.skipClient = true;
+                this.generatorType = 'server';
+                this.skipUserManagement = false;
```
deepu105: I would prefer `this.skipUserManagement = configOptions.skipUserManagement = false;` for consistency here

generators/client/index.js
```diff
@@ -486,7 +486,7 @@ module.exports = JhipsterClientGenerator.extend({
             this.template(ANGULAR_DIR + 'components/util/_jhi-item-count.directive.js', ANGULAR_DIR + 'components/util/jhi-item-count.directive.js', this, {});
 
             // interceptor code
-            if (this.authenticationType === 'oauth2' || this.authenticationType === 'jwt') {
+            if (this.authenticationType == 'oauth2' || this.authenticationType == 'jwt' || this.applicationType == 'gateway') {
```
deepu105: why is `this.applicationType == 'gateway'` checked here? gateway will anyway have `jwt` and existing check should be sufficient rite? did i miss anything?

generators/client/index.js
```diff
@@ -528,7 +528,7 @@ module.exports = JhipsterClientGenerator.extend({
             this.template(ANGULAR_DIR + 'services/auth/_has-any-authority.directive.js', ANGULAR_DIR + 'services/auth/has-any-authority.directive.js', this, {});
             if (this.authenticationType === 'oauth2') {
                 this.template(ANGULAR_DIR + 'services/auth/_auth.oauth2.service.js', ANGULAR_DIR + 'services/auth/auth.oauth2.service.js', this, {});
-            } else if (this.authenticationType === 'jwt') {
+            } else if (this.authenticationType == 'jwt' || this.applicationType == 'gateway') {
```
deepu105: same as above here

deepu105: also plz use triple equals

generators/client/templates/src/main/webapp/app/blocks/config/_http.config.js
```diff
@@ -19,7 +19,7 @@
         $urlRouterProvider.otherwise('/');
 
         $httpProvider.interceptors.push('errorHandlerInterceptor');
-        $httpProvider.interceptors.push('authExpiredInterceptor');<% if (authenticationType == 'oauth2' || authenticationType == 'jwt') { %>
+        $httpProvider.interceptors.push('authExpiredInterceptor');<% if (authenticationType == 'oauth2' || authenticationType == 'jwt' || applicationType == 'gateway') { %>
```
deepu105: same here `this.applicationType == 'gateway'` check is not required as a gateway will always have `authenticationType == 'jwt'` and hence it will always be true here

generators/client/templates/src/main/webapp/app/blocks/interceptor/_auth-expired.interceptor.js
```diff
@@ -5,7 +5,7 @@
         .module('<%=angularAppName%>')
         .factory('authExpiredInterceptor', authExpiredInterceptor);
 
-    <% if (authenticationType === 'oauth2' || authenticationType === 'jwt') { %>
+    <% if (authenticationType === 'oauth2' || authenticationType === 'jwt' || applicationType == 'gateway') { %>
```
deepu105: here as well

generators/client/templates/src/main/webapp/app/blocks/interceptor/_auth.interceptor.js
```diff
@@ -22,7 +22,7 @@
             if (token && token.expires_at && token.expires_at > new Date().getTime()) {
                 config.headers.Authorization = 'Bearer ' + token.access_token;
             }
-            <% } %><% if (authenticationType == 'jwt') { %>
+            <% } %><% if (authenticationType == 'jwt' || applicationType == 'gateway') { %>
```
deepu105: here too

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
```diff
@@ -32,10 +32,31 @@
             var data = {
                 username: credentials.username,
                 password: credentials.password,
-                rememberMe: credentials.rememberMe
```
deepu105: will remember me work now?

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
```diff
@@ -32,10 +32,31 @@
...
+                'Content-Type': 'application/x-www-form-urlencoded',
+                "Authorization" : "Basic d2ViX2FwcDo="
+            };
+            return $http({
+                url: 'uaa/oauth/token',
+                method: 'post',
+                data: data,
+                headers: headers,
+                transformRequest: function(obj) {
+                var str = [];
```
deepu105: this needs to be intent 4 space

generators/generator-base.js
```diff
@@ -1219,6 +1219,9 @@ Generator.prototype.askModuleName = function (generator, currentQuestion, totalQ
     generator.prompt({
         type: 'input',
         name: 'baseName',
+        when: function (response) {;
+                return generator.applicationType != 'uaa';
```
deepu105: `!==` plz

generators/docker-compose/index.js
```diff
@@ -154,6 +154,10 @@ module.exports = yeoman.Base.extend({
                     var path = this.destinationPath(this.directoryPath + this.appsFolders[i]+'/.yo-rc.json');
                     var fileData = this.fs.readJSON(path);
                     var config = fileData['generator-jhipster'];
+                    //this currently can happen, because the .yo-rc doesn't have a baseName
+                    if(fileData['generator-jhipster'].baseName === undefined) {
```
deepu105: All application type should set the` baseName` and hence this should never happen, for appType `uaa` we should set the basename while generating the app appropriately in server or app generator

generators/docker-compose/index.js
```diff
@@ -89,7 +89,7 @@ module.exports = yeoman.Base.extend({
                             if(file.isDirectory()) {
                                 if(shelljs.test('-f', file.name + '/.yo-rc.json')) {
                                     var fileData = this.fs.readJSON(file.name + '/.yo-rc.json');
-                                    if(fileData['generator-jhipster'].baseName !== undefined) {
+                                    if(fileData['generator-jhipster'].baseName !== undefined || fileData['generator-jhipster'].applicationType == 'uaa') {
```
deepu105: this check is not required if we set a baseName appropriately

generators/server/index.js
```diff
@@ -150,13 +150,13 @@ module.exports = JhipsterServerGenerator.extend({
             }
 
             // force variables unused by microservice applications
-            if (this.applicationType === 'microservice') {
+            if (this.applicationType == 'microservice' || this.applicationType == 'uaa') {
```
deepu105: triple equals plz

generators/server/index.js
```diff
@@ -150,13 +150,13 @@ module.exports = JhipsterServerGenerator.extend({
...
                 this.clusteredHttpSession = 'no';
                 this.websocket = 'no';
             }
 
             var serverConfigFound = this.packageName !== undefined &&
-                this.authenticationType !== undefined &&
+                (this.authenticationType !== undefined || (this.applicationType !== undefined && this.applicationType == 'gateway') )&&
```
deepu105: why is this check needed?

generators/server/index.js
```diff
@@ -205,9 +209,12 @@ module.exports = JhipsterServerGenerator.extend({
     prompting: {
 
         askForModuleName: function () {
-            if (this.baseName) return;
+            //need this for having a concrete name in gateway app to build urls like "/jhipster-uaa/api/account"
+            // @TODO read this application name from gateway during its generation
+            if (this.baseName || this.applicationType == 'uaa') return;
```
deepu105: `this.applicationType == 'uaa'` check seems redundant as its done in the following method already, can we keep it either here or in method only

generators/server/index.js
```diff
@@ -205,9 +209,12 @@ module.exports = JhipsterServerGenerator.extend({
     prompting: {
 
         askForModuleName: function () {
-            if (this.baseName) return;
+            //need this for having a concrete name in gateway app to build urls like "/jhipster-uaa/api/account"
+            // @TODO read this application name from gateway during its generation
```
deepu105: may be you can ask a question to the user to enter the gateways name here when `uaa` is selected

generators/server/index.js
```diff
@@ -219,7 +226,7 @@ module.exports = JhipsterServerGenerator.extend({
             var prompts = [
                 {
                     when: function (response) {
-                        return (applicationType === 'gateway' || applicationType === 'microservice');
+                        return (applicationType == 'gateway' || applicationType == 'microservice' || applicationType == 'uaa');
```
deepu105: `===` plz

generators/server/index.js
```diff
@@ -232,7 +239,7 @@ module.exports = JhipsterServerGenerator.extend({
                             currentQuestion = current;
                         }, applicationType === 'gateway' || applicationType === 'microservice');
                     },
-                    default: applicationType === 'gateway' ? '8080' : '8081'
+                    default: applicationType == 'gateway' ? '8080' : applicationType == 'uaa' ? '9999' : '8081'
```
deepu105: `===`

generators/server/index.js
```diff
@@ -630,7 +637,7 @@ module.exports = JhipsterServerGenerator.extend({
             ];
 
             this.prompt(prompts, function (props) {
-                if (this.applicationType === 'microservice' || this.applicationType === 'gateway') {
+                if (this.applicationType == 'microservice'/* || this.applicationType == 'gateway'*/) {
```
deepu105: plz remove the commented out code

generators/server/index.js
xetys: without this line you are not able to generate a gateway using a ready .yo-rc.json.

generators/server/index.js
```diff
@@ -194,6 +194,10 @@ module.exports = JhipsterServerGenerator.extend({
                     this.languages = ['en', 'fr'];
                 }
 
+                if(this.applicationType == 'gateway') {
```
deepu105: can we do this in `app/index.js` itself similar to how we set it for microservice

generators/server/index.js
```diff
@@ -638,10 +645,20 @@ module.exports = JhipsterServerGenerator.extend({
...
                     this.rememberMeKey = crypto.randomBytes(20).toString('hex');
                 }
-                if (this.authenticationType === 'jwt') {
+
+                if (this.authenticationType == 'jwt' || this.applicationType == 'microservice' || this.applicationType == 'uaa') {
                     this.jwtSecretKey = crypto.randomBytes(20).toString('hex');
                 }
 
+                //this now is job for uaa
+                if(this.applicationType == 'gateway') {
```
deepu105: can we do this in `app/index.js` itself similar to how we set it for microservice

generators/server/index.js
```diff
@@ -936,7 +954,7 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_SecurityUtils.java', javaDir + 'security/SecurityUtils.java', this, {});
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_AuthoritiesConstants.java', javaDir + 'security/AuthoritiesConstants.java', this, {});
 
-            if (this.authenticationType === 'jwt') {
+            if (this.authenticationType == 'jwt' && this.applicationType == 'monolith') {
```
deepu105: check should be reversed `this.applicationType == 'monolith' && this.authenticationType == 'jwt'`

generators/server/index.js
```diff
@@ -956,7 +979,8 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_UserDetailsService.java', javaDir + 'security/UserDetailsService.java', this, {});
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_UserNotActivatedException.java', javaDir + 'security/UserNotActivatedException.java', this, {});
 
-            if (this.authenticationType === 'jwt') {
+
+            if (this.authenticationType == 'jwt' && this.applicationType !== 'uaa') {
```
deepu105: this should be `this.applicationType !== 'uaa' && this.authenticationType == 'jwt'` for readability

generators/server/templates/_build.gradle
```diff
@@ -196,9 +196,13 @@ dependencies {
     <%_ if (authenticationType == 'oauth2') { _%>
     compile group: 'org.springframework.security.oauth', name: 'spring-security-oauth2', version: spring_security_oauth2_version
     <%_ } _%>
-    <%_ if (authenticationType == 'jwt') { _%>
+    <%_ if (authenticationType == 'jwt' && applicationType == 'monolith') { _%>
```
deepu105: swap order plz

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
```diff
@@ -222,7 +222,7 @@ public void setFrom(String from) {
 
         private final RememberMe rememberMe = new RememberMe();
         <%_ } _%>
-        <%_ if (authenticationType == 'oauth2' || authenticationType == 'jwt') { _%>
+        <%_ if (authenticationType == 'oauth2' || authenticationType == 'jwt' || applicationType == 'uaa' || applicationType == 'gateway') { _%>
```
deepu105: here as well `authenticationType == 'jwt'` should suffice for gateway appType unless you are saying that you wont set `authenticationType = 'jwt'` for gateway

test/test-creation.js
```diff
@@ -931,7 +931,7 @@ describe('JHipster generator', function () {
         });
 
         it('creates expected files with the microservice application type', function () {
-            assert.file(expectedFiles.jwt);
+            //assert.file(expectedFiles.jwt); //not anymore
```
deepu105: plz remove the line if not required

generators/server/index.js
deepu105: ok can we reorder it as `(this.applicationType != 'gateway' && this.authenticationType !== undefined)`

deepu105:
sorry for giving lot of comments, hope you will find time to review those :+1:

xetys:
i have never been working on yeoman generators, so its okay :smile:

deepu105:
Thank you, you would love it once you learn it :)
Thanks for bearing with us. This would be a great feature indeed :smile_cat:

deepu105:
@jdubois @xetys I was wondering wouldnt this be kind of a breaking change if existing users try to upgrade? should we offer `uaa` as an option instead of forcing it?

xetys:
perhaps it is, but I believe it worths this, since the current setup has a bottle neck on gateway and is not really microservice like. In other words: The gateways responsibility is to large right now. One more thing is, there is sometimes no need of authentication, or only for some parts of the domain (think about shops wich need auth maximum on checkout)

but I really would love to see this PR beeing merged, so if this approach is inacceptable, I will think about offering the old way of generation also....with gateway with auth and so on

jdubois:
I just add a quick look and indeed, does this code force users to use UAA?
I would rather like to provide the choice with several authentication methods on the gateway, like today with monoliths. And then use JWT for all microservices.

xetys:
how to say..."it forces"...it sets "skipUserManagement" on gateway, assuming you use a UAA for this....similar as you do with jhipster registry, but with more generating capabilities on the UAA as JHipster application itself.

The gateway itself currently does not have any authentication, because this is done by the angular app

what kind of authentication methods you want to provide also? currenty oauth < - > jwt ist the only one

xetys:
I have to write more docs I think

deepu105:
@xetys dont get me wrong. Of course I would love to have this in the generator. My only concern is we shouldnt force it to be the only way if there is another way available. 
What I have in mind is to have one more question in the server sub gen when the `applicationType` is `gateway` as below

```
(x/16) Which *type* of authentication would you like to use?

[ * ] JWT authentication (stateless, with a token)
[   ] UAA authentication server (you need to generate a separate UAA server app using JHipster)

```

this way if JWT is selected we can retain current behavior and if UAA is selected we can skip JWT stuff like in your PR. So the user gets to chose what they want and it wont be a breaking change for current users as well @jdubois what do you think of this approach?

gmarziou:
@deepu105  I like your proposal because it also leaves the door open to propose KeyCloak as an alternative to UUA inthe future if we want to.

xetys:
This is something I could do. The only thing I am worrying about is, that the JWT solution is self crafted and acts as an custom authentication implementation, which have to be maintained over time by JHipster....where as the oauth2 solution is maintained by spring itself

so it would may make sense to expose the JwtTokenConverter from spring-security-oauth2 to manage the current JWT as far as possible. Then all servers would be resource servers, while microservice application don't care, whether its JWT stateles or via OAuth2

deepu105:
@PierreBesson @wmarques could you plz test this with the docker-compose sub gen?

PierreBesson:
OK, I'm testing this now.

PierreBesson:
Hi @xetysn, great work you have done there !
I have fixed some small errors that prevented me to have a working generated uaa server on gradle and maven.
Please review them here: https://github.com/xetys/generator-jhipster/pull/1
I'm now testing to see if it integrates with docker compose.

PierreBesson:
@xetys I'm unable to log in on the gateway. Do you know what might be the problem ?

xetys:
I would look into the console log of your browser...

500 is just the standard forwarding error in zuul, just repeat
4XX is always coming with an error message text...

but I have to pull locally and test, when I find some time

deepu105:
@xetys the current JWT implementation is using JJWT and not spring JWT there was some long discussion regarding that [here](https://github.com/jhipster/generator-jhipster/pull/2311) and we finally choose JJWT over spring so we would like to stick to that and offer options as I suggested earlier above cc @amit2103 as he did some of the earlier JWT stuff here

so it would be awesome if you could make the proposed changes

xetys:
just to understand you correctly, you want the oauth2 communication running with JJWT, or you just want to provide JJWT when the user chooses JWT authentication?

During my development the UAA generates spring-security-tokens were running against the microservices, which were excepting JHipsters JWT tokens, and using the same secret key, the microservice still could verify the token, although it was generated with spring, not with your implementation.

I would not say I know it for sure, but it seems like the only difference between springs JWT and JH JWT, is: the JSON structure, so it was only possible to authoirze requests, but not to extract roles.

If this is true, why whe should provide a manual implemented solution over a spring provided?

Maybe I don't see the difference yet, is there some things I don't know about JJWT

deepu105:
I think the main reason for choosing JJWT over spring JWT was project status, spring JWT seems not actively maintained. Not sure if its still the case

My preference would be to keep current JWT as it is when user chooses option 1 `jwt` and if user chooses option 2 `uaa` then your changes needs to be applied.

@xetys I can help with a PR to your fork if you need @jdubois are you ok with it?

xetys:
there is one more thing: if the JWT mechanism would be done over one provider (JWT or JJWT), we could handle it independent from authenticationType on microservices. 

Since a oauth2 resource server with jwt does exactly the same, as the JWT Token Filter by JHipster. So by choosing one mechanism, the only point to change would be the gateway. 

Handling two authentication types with would force the user to choose this also by generating microservices, or handle both mechanisms. 

If it really have to be JJWT, there is also an option to implement the JWT logic everywhere, and maintain over time. I am speaking about implementing the JwtAccesTokenConverter differntly. 
This of course is an option, but I don't think its good.

Another thing we didn't talked about is, that springs jwt solution also provides encryption with RSA instead of AES (i am sure, JJWT can do that also). This would improve the security, since only the UAA would know the private key and you cannot fake an access token by havin just the public key.

Changing to RSA are about 2 lines of code to change in springs-security....

@deepu105 i am happy about any help I get...but maybe I find some time tonight....

PierreBesson:
@xetys, Someone droped his demo project in the [spring-security-oauth gitter](https://gitter.im/spring-projects/spring-security-oauth) and it seems to do something similar to what we want: https://github.com/kakawait/uaa-behind-zuul-sample.

deepu105:
@jdubois I really need your opinion on this, since you did the JWT implementation. so we will wait for you to come back from holidays may be

amit2103:
@deepu can you pleas provide an overview of whats being done?

xetys:
@PierreBesson yes...but on my first look it seems that with a working and accepted version of #3324 #3329 JHipster will be more modern than this drop :)

I had updated everything, may it's time to take a look over this last little failing test, to make my PR beeing green in travis

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
xetys: should, because the "data" is actually payload for the uaa. the rememberMe flag is set later in services again

generators/docker-compose/index.js
xetys: with uaa, this happens...

generators/server/index.js
xetys: as of new authenticationType 'uaa', this turned back to what it was

generators/server/index.js
xetys: this part contains now logic as of "if gateway with uaa"

xetys:
i am not exactly sure, what caused the failed builds...will check this

currenty the basic separation between uaa gateways and jwt gateways is done.

the simplest, but ugliest solution would be, if whe just ask the "authentication" question on microservice creation as well as on gateway creation. Doing this, it means you have to decide from the very beginning, how the authentication is going to happen, or having fun with regeneration


my vote: use spring jwt instead of jjwt and use it the same way as the jwt gateway uses it now. 
So then the user may decide, wether to do JWT token on gateway or riding the royal flagship of oauth2 servers.....without the need to ever change the microservices....you just could use a system with 2 gateways, JWT and uaa....or latelly switch from one to other....this would be more hipster like :smile:

xetys:
yes...this is murphy, isn't it? I was quite sure, this build would be my first passing on this PR, as it was locally....and travis just wants me to stop contributing to you, by ruining my successfully builds with random errors....unfortunetly this is making me upset :cry:

xetys:
nevertheless, what's now in the box:

I have implemented the "I don't like" solution, as I mentioned above. I still want this discussion to proceed, but for now we got it this way (if you choose uaa on gateway, you have to generate a uaa and choose uaa in microservices, or you go with JWT)

xetys:
omg it....is it really green? :smile:

pascalgrimaud:
@xetys : good job!
I have to find time to look your [jhipster-travis-helper](https://github.com/xetys/jhipster-travis-helper) ! It seems very interesting.

FYI, you can link your forked generator-jhipster to your own travis. For example, my travis -> https://travis-ci.org/pascalgrimaud/generator-jhipster

So you can have your https://travis-ci.org/xetys/generator-jhipster
Then work on your branch until it's green, and Pull Request
With your own travis-ci, you can stop/start your job. No need to wait

xetys:
thank you! Learned something new :smile: 

But it's actully fun when your pc is a monster

generators/docker-compose/index.js
deepu105: It should not happen ideally let me check

deepu105:
lets wait for @jdubois to review this

sdoxsee:
Haven't had a chance to look through the full PR but I have some comments.

1.  I think it makes more sense to use a basic standalone UAA server (lke https://github.com/jhipster/jhipster-openid-connect) rather than a generated jhipster-UAA. I don't think we want to maintain a oauth server generator on top of everything else...it's not jhipster's focus (nor should it be IMO). I think jhipster should provide a basic standalone UAA as default and perhaps add options to other standalone openid connect servers (keycloak, mitreid, doesn't even have to be java...)
1.  I know that there was a discussion (https://github.com/jhipster/generator-jhipster/pull/2311) about JJWT vs spring oauth jwt. I do like that spring oauth jwt is built-in and there's not custom code to maintain...so if the dev chooses JWT-alone without oauth I'd prefer, if it were possible, to build the JWT-alone JWT in such a way that it would work with @EnableResourceServer (on the microservice application level) rather than custom JJWT code.

@xetys have you seen this? https://github.com/sdoxsee/jhipster-openid-connect-microservices

xetys:
the problem I see with the UAA is, that it has no full fledged user management. I was looking for some solution, to both maintain plain old user management and serve them over oauth2. 
The uaa in your repo is provisioning users via shell, so there is initially no db (or I don't see one), as well as no customized registration, activation and so on...this actually is already done by JHipster...the last missing part was communicate this data as an OAuth2 authenticatin server

JHipster is not a install & use software, but a tool for developers, so maintain a UAA inside the generator has to bring the most capabilities of JHipster availiable now, to customize. Since OAuth itself is nothing more than a standard, the developers finally can decide what is better for them, because this is microservices, don't like this solution, swap with something other, what is compatible

I can refer to an situation, were I am replacing a 10 years old monolith with an microservice architecture. Unfortunetely the user management is not that easy as having one central view, were some admins manage them. The app itself has several sources from where to synchronize users into the system.....how could i impmenet this using CloudFoundrys UAA? No idea...With JHipster? Just generate one UAA and add some REST consume logic to my APIs and thats it

So IMO, having the option to customize maximum of the app should be JHipster approach

generators/docker-compose/index.js
```diff
@@ -108,22 +108,10 @@ module.exports = yeoman.Base.extend({
...
                     var path = this.destinationPath(input);
                     if(shelljs.test('-d', path)) {
-                        var files = shelljs.ls('-l',this.destinationPath(input));
-                        this.appsFolders = [];
-
-                        files.forEach(function(file) {
-                            if(file.isDirectory()) {
-                                if(shelljs.test('-f', file.name + '/.yo-rc.json')) {
-                                    var fileData = this.fs.readJSON(file.name + '/.yo-rc.json');
-                                    if(fileData['generator-jhipster'].baseName !== undefined || fileData['generator-jhipster'].applicationType == 'uaa') {
```
pascalgrimaud: Just be carefull. When rebasing, you removed your test : `fileData['generator-jhipster'].applicationType == 'uaa'`
we put this on a new function above: _getAppFolders()

xetys: thanks to this advice :smile:

sdoxsee:
@xetys 
> the problem I see with the UAA is, that it has no full fledged user management.
> The uaa in your repo is provisioning users via shell, so there is initially no db (or I don't see one), as well as no customized registration, activation and so on...this actually is already done by JHipster...

Apparently you can hook it up to a db (mysql or postgres) https://github.com/cloudfoundry/uaa#user-account-data. They also follow an SCIM standard for user management with an API for working with the user data. Also, there IS registration and activation via email (you just have to setup your smtp similar to jhipster)

That said, I agree with you. CLI for user management is not ideal. I like your idea of having a nicer UI for user management. Maybe we can go there (building a user management UI product that uses the SCIM api for UAA) once we have oauth2 working in microservices? However, as you say, I don't know how you hook UAA into existing user databases without exporting data into UAA's db...which isn't always an option for people. In that case, keycloak or mitreid (or others?) would be better options for people as they do allow this (I've done it). 

Out of curiosity, what options would one add to a User Management service generator? 

> JHipster is not a install & use software, but a tool for developers, so maintain a UAA inside the generator has to bring the most capabilities of JHipster availiable now, to customize. Since OAuth itself is nothing more than a standard, the developers finally can decide what is better for them, because this is microservices, don't like this solution, swap with something other, what is compatible

The thing is that there are existing products for OpenID Connect and User Management out there. Why create/maintain a generator for people to implement their own? Would it be better than keycloak? I think I've worked with enough Authorization Servers to know that if you're adding it to an existing monolith then you'll probably need to customize it for your needs (mostly branding and connecting it to existing user data) but, usually, these products provide a way (or ways) to hook into old user databases.

The first agile step IMO is to get jhipster microservices working with a basic standalone UAA.

Thanks for moving oauth2 for microservices forward. I'm appreciating the good discussion

jdubois:
Hi guys,

I'm finishing my holidays and this thread is just too long :-)
There is one thing I want to add about JJWT: I have customers who want some
very fine-grained control on their JWT token, for example on the algorithm.
They are also happy with a "full JWT" solution, so without OAuth2.
For those 2 reasons I'd like to stick to JJWT, at least until I've studied
Spring Security's JWT implementation.

Cheers,

Julien
Le 9 avr. 2016 4:32 PM, "Stephen Doxsee" <notifications@github.com> a
écrit :

xetys:
@sdoxsee First I want to clearify, that I am quite new to all this and maybe less expierienced in well known technology. I'm just dealing a lot with these things for a while.

I am feeling we are talking about little bit different things when talking about UAA. It seems UAA is an well known implementation...at least..there some indeed. For me UAA is a concept. For microservice, this is a service, managing users, and managing authorization and authentication, as other UAAs are doing.

My approach is, to use the already existing user management from JHipster, which will be maintained, since it is used not just for the UAA, and give some UAA cabalilities to it. This can be achieved with a few more dependencies, some more annotations and a bit of bean wiring. So this combines the up-to-date evolution from JHipsters User management with a up-to-date evolution of spring-security oauth2 solution, which they fit to common standards. 

In spring, things like @EnableAuthorization server, comes when they look e.g. at cloudfoundries UAA and say: we need this, spring style wirable and spring boot configurable in our toolchain...

the idea is quite simple: JHipster users already know the user management, so turning it into an UAA inside generator, doesn't force them learn new thinks. If you don't like, you have all the options to change things, from small things into swapping into different solutions. For me, this seems to be an obvious.

But maybe, this is just my lack of knowledge tools like keycloak. What I really don't know, is how difficult is, to prepare external uaas as cloudfoundry or keycloak into JHipsters microservice ecosystem (eureka, zuul...). It will be definitly more work, than generating an JHipster app with some more config as I did. 

Do you think, going with cloudfoundrys uaa/keycloak will make life easier for jhipster users?

deepu105:
@xetys and @sdoxsee both of you guys have valid points and thats the reason we wanted it to be an option so that people can choose what they like, so we could have our current JWT, @xetys new UAA server and tomorrow we could add openid connect, keyclock etc to the options as well.

xetys:
for you to know whats still open right now:

* fixing the base name makes life a bit harder, so I follow @deepu105 's advice and ask for uaa name on gateway genertor for the urls
* looking, how JJWT can be used instead of spring jwt

xetys:
well, next update includes:

* conflict resolving
* you now can choose between JWT and UAA on gateways and microservices
* on gateway creation, JH also asks the name of the uaa application
* some issues fixed when regenerating from saved .yo-rc

**NOTE:** you will need to apply also #3329 into the gateway, to make JWT working

the UAA setup still uses springs JJWT, since this will be a bit more to work, and i had just a little bit of time the last days

so please keep on reviewing it, I hope the quality of this PR slowly reaches the quality of JHipsters codebase :smile:

deepu105:
No worries, we can take our time as we are not in a hurry for a new release
:)

generators/app/index.js
```diff
@@ -255,6 +259,12 @@ module.exports = JhipsterGenerator.extend({
                 this.generatorType = 'server';
                 this.skipUserManagement = configOptions.skipUserManagement = true;
             }
+            if (this.applicationType == 'uaa') {
+                this.skipClient = true;
+                this.generatorType = 'server';
+                this.skipUserManagement = configOptions.skipUserManagement = false;
+                //this.authenticationType = configOptions.authenticationType = 'uaa';
```
deepu105: need to remove this

generators/client/index.js
```diff
@@ -486,7 +490,7 @@ module.exports = JhipsterClientGenerator.extend({
             this.template(ANGULAR_DIR + 'components/util/_jhi-item-count.directive.js', ANGULAR_DIR + 'components/util/jhi-item-count.directive.js', this, {});
 
             // interceptor code
-            if (this.authenticationType === 'oauth2' || this.authenticationType === 'jwt') {
+            if (this.authenticationType === 'oauth2' || this.authenticationType === 'jwt' || this.authenticationType == 'uaa') {
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/blocks/config/_http.config.js
```diff
@@ -19,7 +19,7 @@
         $urlRouterProvider.otherwise('/');
 
         $httpProvider.interceptors.push('errorHandlerInterceptor');
-        $httpProvider.interceptors.push('authExpiredInterceptor');<% if (authenticationType == 'oauth2' || authenticationType == 'jwt') { %>
+        $httpProvider.interceptors.push('authExpiredInterceptor');<% if (authenticationType == 'oauth2' || authenticationType == 'jwt' || authenticationType == 'uaa') { %>
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/blocks/interceptor/_auth.interceptor.js
```diff
@@ -22,7 +22,7 @@
             if (token && token.expires_at && token.expires_at > new Date().getTime()) {
                 config.headers.Authorization = 'Bearer ' + token.access_token;
             }
-            <% } %><% if (authenticationType == 'jwt') { %>
+            <% } %><% if (authenticationType == 'jwt' || authenticationType == 'uaa') { %>
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_account.service.js
```diff
@@ -8,7 +8,7 @@
     Account.$inject = ['$resource'];
 
     function Account ($resource) {
-        var service = $resource('api/account', {}, {
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/account'<%} else { %>'api/account'<% } %>, {}, {
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_activate.service.js
```diff
@@ -8,7 +8,7 @@
     Activate.$inject = ['$resource'];
 
     function Activate ($resource) {
-        var service = $resource('api/activate', {}, {
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/activate'<%} else { %>'api/activate'<% } %>, {}, {
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
```diff
@@ -29,13 +29,45 @@
         }
 
         function login (credentials) {
-            var data = {
+            <% if(authenticationType == 'uaa') { %>
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_auth.session.service.js
```diff
@@ -43,13 +43,18 @@
 
         function logout () {<% if (websocket == 'spring-websocket') { %>
             <%=jhiPrefixCapitalized%>TrackerService.disconnect();<% } %>
+
+            <% if(authenticationType == 'uaa') { %>
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_password-reset-finish.service.js
```diff
@@ -8,7 +8,7 @@
     PasswordResetFinish.$inject = ['$resource'];
 
     function PasswordResetFinish($resource) {
-        var service = $resource('api/account/reset_password/finish', {}, {});
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/account/reset_password/finish'<%} else { %>'api/account/reset_password/finish'<% } %>, {}, {});
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_password-reset-init.service.js
```diff
@@ -8,7 +8,7 @@
     PasswordResetInit.$inject = ['$resource'];
 
     function PasswordResetInit($resource) {
-        var service = $resource('api/account/reset_password/init', {}, {});
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/account/reset_password/init'<%} else { %>'api/account/reset_password/init'<% } %>, {}, {});
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_password.service.js
```diff
@@ -8,7 +8,7 @@
     Password.$inject = ['$resource'];
 
     function Password($resource) {
-        var service = $resource('api/account/change_password', {}, {});
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/account/change_password'<%} else { %>'api/account/change_password'<% } %>, {}, {});
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_register.service.js
```diff
@@ -8,6 +8,6 @@
     Register.$inject = ['$resource'];
 
     function Register ($resource) {
-        return $resource('api/register', {}, {});
+        return $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/register'<%} else { %>'api/register'<% } %>, {}, {});
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/auth/_sessions.service.js
```diff
@@ -8,7 +8,7 @@
     Sessions.$inject = ['$resource'];
 
     function Sessions ($resource) {
-        return $resource('api/account/sessions/:series', {}, {
+        return $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/account/sessions/:series'<%} else { %>'api/account/sessions/:series'<% } %>, {}, {
```
deepu105: `===`

generators/client/templates/src/main/webapp/app/services/user/_user.service.js
```diff
@@ -8,7 +8,7 @@
     User.$inject = ['$resource'];
 
     function User ($resource) {
-        var service = $resource('api/users/:login', {}, {
+        var service = $resource(<% if(authenticationType == 'uaa') { %>'<%= uaaBaseName %>/api/users/:login'<%} else { %>'api/users/:login'<% } %>, {}, {
```
deepu105: `===`

generators/client/templates/src/test/javascript/spec/app/services/auth/_auth.services.spec.js
```diff
@@ -16,9 +16,9 @@ describe('Service Tests', function () {
             sessionStorageService = $sessionStorage;
             authService = Auth;
             spiedAuthServerProvider = AuthServerProvider;
-<%_ if (authenticationType != 'jwt') { _%>
+            <%_ if (authenticationType != 'jwt' && authenticationType != 'uaa') { _%>
```
deepu105: `!==`

generators/docker-compose/index.js
deepu105: set base name while generating the `uaa` app

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
                 },
                 {
                     when: function (response) {
+                        return applicationType === 'gateway' || applicationType == 'microservice';
```
deepu105: `===`

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
...
+                    },
+                    type: 'list',
+                    name: 'authenticationType',
+                    message: function (response) {
+                        return getNumberedQuestion('Which *type* of authentication would you like to use?', currentQuestion, totalQuestions, function (current) {
+                            currentQuestion = current;
+                        }, applicationType === 'monolith');
```
deepu105: condition should be `applicationType === 'gateway' || applicationType === 'microservice'`

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
...
+                    },
+                    choices: [
+                        {
+                            value: 'jwt',
+                            name: 'JWT authentication (stateless, with a token)'
+                        },
+                        {
+                            value: 'uaa',
+                            name: 'Authentication via JHipster UAA (you will have to generate one separately)'
```
deepu105: Authentication via JHipster UAA (you will have to generate app separately)

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
...
+                        return (applicationType === 'gateway' && response.authenticationType === 'uaa');
+                    },
+                    type: 'input',
+                    name: 'uaaBaseName',
+                    validate: function (input) {
+                        if (true) return true; //@TODO check it really
+                        return 'Could not finde UAA application "' + input + '"';
+                    },
+                    message: function (response) {
+                        return getNumberedQuestion('What is the path of your UAA application?.', currentQuestion, totalQuestions, function (current) {
```
deepu105: this should be an app name right like `jhipsterUAA` or `myUaaApp`? similar to the microservice name we use. How are you planning to validate if it exists as in your TODO above?

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
...
+                            currentQuestion = current;
+                        }, true);
```
deepu105: should be `applicationType === 'gateway' && response.authenticationType === 'uaa'` and not true

generators/server/index.js
```diff
@@ -278,6 +282,46 @@ module.exports = JhipsterServerGenerator.extend({
...
+                    },
+                    default: 'uaa'
```
deepu105: `jhi-uaa` or `jhiUaa` may be for consistency?

generators/server/index.js
```diff
@@ -630,18 +674,25 @@ module.exports = JhipsterServerGenerator.extend({
...
                     this.rememberMeKey = crypto.randomBytes(20).toString('hex');
                 }
-                if (this.authenticationType === 'jwt') {
+
+                if (this.authenticationType === 'jwt' || this.authenticationType === 'uaa' || this.applicationType === 'microservice' || this.applicationType === 'uaa') {
                     this.jwtSecretKey = crypto.randomBytes(20).toString('hex');
                 }
 
+                //this now is job for uaa
+                if(this.applicationType == 'gateway' && this.authenticationType == 'uaa') {
```
deepu105: `===`

generators/server/index.js
```diff
@@ -847,7 +901,8 @@ module.exports = JhipsterServerGenerator.extend({
             if (this.searchEngine === 'elasticsearch') {
                 this.template(DOCKER_DIR + '_elasticsearch.yml', DOCKER_DIR + 'elasticsearch.yml', this, {});
             }
-            if (this.applicationType === 'microservice' || this.applicationType === 'gateway') {
+
+            if (this.applicationType == 'microservice' || this.applicationType == 'gateway' || this.applicationType == 'uaa') {
```
deepu105: `===`

generators/server/index.js
```diff
@@ -961,7 +1021,8 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_UserDetailsService.java', javaDir + 'security/UserDetailsService.java', this, {});
             this.template(SERVER_MAIN_SRC_DIR + 'package/security/_UserNotActivatedException.java', javaDir + 'security/UserNotActivatedException.java', this, {});
 
-            if (this.authenticationType === 'jwt') {
+
+            if (this.authenticationType == 'jwt') {
```
deepu105: `===`

generators/server/templates/_build.gradle
```diff
@@ -208,6 +208,10 @@ dependencies {
     <%_ if (authenticationType == 'jwt') { _%>
     compile group: 'io.jsonwebtoken', name: 'jjwt', version: jjwt_version
     <%_ } _%>
+    <%_ if (applicationType == 'uaa' || authenticationType == 'uaa') { _%>
```
deepu105: wouldnt both condition be true always?

generators/server/templates/_pom.xml
```diff
@@ -85,11 +85,11 @@
         <postgresql.version>9.4-1203-jdbc42</postgresql.version>
         <%_ } _%>
         <run.addResources>false</run.addResources>
-        <%_ if (applicationType == 'microservice' || applicationType == 'gateway') { _%>
+        <%_ if (applicationType == 'microservice' || applicationType == 'gateway' || applicationType == 'uaa') { _%>
         <spring-cloud.version>Brixton.RC1</spring-cloud.version>
         <%_ } _%>
         <spring-security.version>4.0.4.RELEASE</spring-security.version>
-        <%_ if (authenticationType == 'oauth2') { _%>
+        <%_ if (authenticationType == 'oauth2' || authenticationType == 'uaa' || applicationType == 'uaa') { _%>
```
deepu105: wouldnt just `authenticationType === 'oauth2' || authenticationType === 'uaa'` do?

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -29,7 +29,11 @@ zuul:
 
 spring:
     application:
+        <%_ if (applicationType == 'uaa') { _%>
+        name: uaa
```
deepu105: should be set as `<%= baseName %>`

deepu105:
@xetys sorry for lot of comments again. PR LGTM now, most of my comments are cosmetic

generators/server/index.js
xetys: what does this note mean?

generators/server/index.js
xetys: The simple way: just check the folder
The good way: check folder, .yo-rc.json and applicationType === 'uaa'

generators/app/index.js
xetys: are you sure?

xetys: or you are speaking about the commented line?

generators/server/templates/_build.gradle
xetys: an uaa application has always authenticationType uaa, but also other applications can have this type.

it's not always true, but authenticationType === 'uaa' is indeed enough

generators/server/templates/src/main/resources/config/_application.yml
xetys: just forgot this place...it was fixed in my early versions

xetys:
"PR LGTM now"...what do you mean with this? :smile: 

I also replaced some more "==" with "===" if it is ok

wmarques:
@xetys 

> Pull Request Looks Good To Me

 :smile: 
Yes we should use triple eq now :)

generators/app/index.js
deepu105: yes the commented line, if its not required plz remove it. But from your code flow I assume its ok to set `authenticationType` type to uaa when `applicationType` is uaa

generators/server/index.js
deepu105: how will you know the folder path? if you are asking folder Path here then how will you know the uaa app name? say for example I can choose a baseName for uaa app like `myAwesomeUaa` and my app can be in a folder `../security/uaaApp` in this case how will you get the app name? are you goann read the `yo-rc.json` in the folder and read base name from that? I guess that would be the only way

generators/server/index.js
```diff
@@ -630,18 +674,25 @@ module.exports = JhipsterServerGenerator.extend({
...
+                if (this.authenticationType === 'jwt' || this.authenticationType === 'uaa' || this.applicationType === 'microservice' || this.applicationType === 'uaa') {
                     this.jwtSecretKey = crypto.randomBytes(20).toString('hex');
                 }
 
+                //this now is job for uaa
+                if(this.applicationType === 'gateway' && this.authenticationType === 'uaa') {
+                    this.skipUserManagement = true;
+                }
+
+                if(applicationType === 'uaa') {
```
deepu105: this should be `this.applicationType ` i guess

deepu105:
I still have some cosmetic issues here, ill do a PR to your fork to correct them

generators/server/index.js
deepu105: Im doing this as well in the PR

generators/server/templates/src/test/resources/config/_application.yml
```diff
@@ -103,24 +103,17 @@ jhipster:
         corePoolSize: 2
         maxPoolSize: 50
         queueCapacity: 10000
+<%_ if (authenticationType == 'oauth2') { _%>
     security:
-<%_ if (authenticationType == 'oauth2' || authenticationType == 'jwt') { _%>
         authentication:
-    <%_ if (authenticationType == 'jwt') { _%>
```
deepu105: is this ok to be removed for all `authenticationType ` ?

deepu105:
I have done a PR https://github.com/xetys/generator-jhipster/pull/5 to fix some stuff. It has commits from master as well but I guess you can safely merge and update this PR

generators/server/index.js
xetys: no, because this is coming from the app generator, the variable applicationType is instanced before

if u use "this" instead, applicationType is undefined

deepu105: this is not a template you cannot get anything in `applicationType` here it would be un defined. You are accessing `this.applicationType ` a line above

deepu105: Im sorry actually both should work coz at line 243 we do this `var applicationType = this.applicationType;`

ddewaele:
Hi,

I was going to log an issue about static roles in the UI (usermgmt controllers) vs dynamic roles in the DB (authorities) when I stumbled upon this post.

I currently have the following microservice architecture that I wanted to share with you guys to see how it aligns where jHipster is going with microservices / oauth2. It was designed and implemented before jHipster 3.0 was out. I've only come to know about jHipster microservices since a couple of days.) 

It offers single sign on / single sign out based on oauth2, and uses the Spring Netflix stack where all resources  (including the auth server) are sitting behind a zuul proxy

The AuthServer is a slightly modified version of the jHipster oauth2 monolith, that is placed behind the zuul gateway and is used to issue tokens and do basic user mgmt. So the uaa / user management is one jHipster app (it required very little modifications to an existing oauth2 app).



Is that the direction you are taking with the oauth2 microservices in jHipster ?

We've put the auth server behind the gateway to make the architecture more portable and more self contained (only 1 URL to worry about on the outside)

Our gateway can do coarse grained authorization checks (acting as kind of a portal landing page, exposing the various resources / UIs and seeing what roles can access what resources).
The apps themselves can do fine-grained authorization as everything is backed by Spring security.

Very interested to see where jHipster is going with all of this....

xetys:
@ddewaele this was exactly my approach. At the moment my PR makes it easy to scaffold an architecture like your image...but without EnableOauth2Sso, wich is done in angularjs in the same way...

@jdubois just to mention, there of course is a way to combine the spring oauth2 implementation and JJWT, by implementing the interfaces and ordering new beans for JwtAccessTokenEnhancer...the default one already is comming with all the authorities and scopes generated by AuthServer + the serialzes userDetails...

deepu105:
This PR LGTM and I think its ready to be merged. Ill wait for @jdubois take a look and confirm

jdubois:
I don't have the time to test right now, but I just read the code quickly and indeed this looks good to me.
The only thing I don't understand is the title of this ticket: it says we migrate from a "manual JWT" to Spring JWT, but from what I've read we still have the former JJWT mechanism.
This is very important to me:

- I'm very happy with the JJWT implementation, and I don't want to change it at the moment, for when JWT is selected
- For OAuth2, I'm OK to use Spring JWT instead, as it's working out of the box

Is this good for you?

xetys:
it's not just good for me...it is the current state of this PR :smile: 

When you choose authentication type "jwt", the JJWT solution is used to manage JWT authorization. When you choose UAA as authentication type, springs default jwt implementation is used.

deepu105:
@jdubois I think this can be merged. btw @xetys you didnt answer my below question

> is this ok to be removed for all authenticationType ?

```
+<%_ if (authenticationType == 'oauth2') { _%>
      security:
 -<%_ if (authenticationType == 'oauth2' || authenticationType == 'jwt') { _%>
          authentication:
 -    <%_ if (authenticationType == 'jwt') { _%>
 -            jwt:
 -                secret: jwt-secret-key
 -                # Token is valid 30 minutes
 -                tokenValidityInSeconds: 1800
 -    <%_ } _%>
 -    <%_ if (authenticationType == 'oauth2') { _%>
              oauth:
                  clientid: <%= baseName %>app
                  secret: mySecretOAuthSecret
                  # Token is valid 30 minutes
                  tokenValidityInSeconds: 1800
 -    <%_ } _%>
```

generators/server/templates/src/test/resources/config/_application.yml
deepu105: @xetys can you answer this?

jdubois:
Thanks @xetys that's what I understood but I wanted to be sure! So this is OK for me.

xetys:
@deepu105 better not. since the uaa authentication uses the same configuration block as jwt....this part should be there on both authenticationTypes.

deepu105:
@xetys but your code is removing the entire block. can you check

xetys:
@deepu105 you were right, this was leading to failing test. fix is pushed

xetys:
the build fail was some travis bug i think...

deepu105:
LGTM now can be merged

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
```diff
@@ -29,13 +29,45 @@
...
-            var data = {
+            <% if(authenticationType === 'uaa') { %>
+              var data = {
+                username: credentials.username,
+                password: credentials.password,
+                grant_type: "password"
+              };
+              var headers = {
+                'Content-Type': 'application/x-www-form-urlencoded',
+                "Authorization" : "Basic d2ViX2FwcDo="
```
sdoxsee: Is this the base 64-encoded "client_id:client_secret"? Shouldn't be hardcoded, should it?

generators/server/templates/_pom.xml
```diff
@@ -118,7 +118,7 @@
         <sonar.tests>${project.basedir}/<%= TEST_DIR %></sonar.tests>
     </properties>
 
-    <%_ if (applicationType == 'microservice' || applicationType == 'gateway') { _%>
+    <%_ if (applicationType === 'microservice' || applicationType === 'gateway') { _%>
     <dependencyManagement>
         <dependencies>
```
sdoxsee: We'll also need dependency management for spring-cloud

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.springframework.cloud</groupId>
                <artifactId>spring-cloud-dependencies</artifactId>
                <version>${spring-cloud.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

generators/client/templates/src/main/webapp/app/services/auth/_auth.jwt.service.js
xetys: you are right, hardcode this makes it unflexible for the future....but encode this each request makes a great overhead. together with a clientDetailsService this should be the first enhancement on this PR.

i had hardcoded it just to make the very plain authorization workin 1:1 as the JJWT do.

xetys:
just tell me, what did i done to your travis, that it is mobbing me about every time i push to JHipster :confused: :smile: 

the build should be alright...bugged again

pascalgrimaud:
Just restarted it. It's green

xetys:
thanks a lot!

btw: this PR just got the most discussed PR on JHipster :smile:

jdubois:
Hi guys, sorry for the delay.
I'm working on merging this today!!!

jdubois:
OK I'm going to merge this.

Some comments, thus:

- It can only be used for testing at the moment, as the secret should not be hard-coded -> this shouldn't be a big change, but at the moment this is why I'm going to write "BETA" on this feature everywhere
- There's a full JHipster app generated for UAA, and I don't think this is necessary.
- There is support for SQL, MongoDB and Cassandra -> are those really useful? As the token are working in-memory, I don't even know if they are really working. I propose to limit to SQL, so there are less stuff to test.
- I don't understand what the "uaaBaseName" is used for? I see only the gateway needs it, but why?

deepu105:
UaaBaseName is required to route the calls to the uaa app url. Thats what i
understood

jdubois:
Thanks @deepu105 indeed, that's logical -> but in that case that should just be a variable used when generating the gateway, we shouldn't have to point to the UAA server's directory. I would like to have our setup working without the need to have the UAA server installed.

xetys:
> It can only be used for testing at the moment, as the secret should not be hard-coded -> this shouldn't be a big change, but at the moment this is why I'm going to write "BETA" on this feature everywhere

the internal "clientId" from my UAA is currently in memory set to "web-app"...as far this is in memory, there is no need to encode base64("web-app:") every time for each request. I am not sure what would be best practice:

* keep clients similar to users in persistance, then we need a (cached) base64 encoding on each request
* render a specific base64 header once (inside a generator e.g.) and set it into ng-constants



> There's a full JHipster app generated for UAA, and I don't think this is necessary.

This is already discussed upper in this thread. There are 2 approaches, one to use something like cloudfoundry uaa and blow it up when needed. The other one (my) is to make JHipster UAA be the same as any other JHipster application, like gateway, mciroservice, registry and so on


> There is support for SQL, MongoDB and Cassandra -> are those really useful? As the token are working in-memory, I don't even know if they are really working. I propose to limit to SQL, so there are less stuff to test.

...depends on how mighty you want the UAA to be


> I don't understand what the "uaaBaseName" is used for? I see only the gateway needs it, but why?

it is the name, which will be visible in the registry (eureka discovery). I first hardcoded it, but using a fixed name for baseName makes some crazy issues in yeoman, so i gave the user the choice of the UAA apps name, but then also the need to specifiy it on gateway generation

xetys:
one more thing...tokens are not stored anywhere as I understand...this is were we need the secret key...to verify a token is valid or not...this is confusing, since you have to implement this in a "TokenStore", but it is directly realized with a JwtAccessTokenEnhancer, which is only checking JWT token validity...

so SQL, Cassandra and Mongo is more for busines logic rather then authorization logic....so the user could implement complex business models over the user domain (authorities, groups, profile, account data....)

imagine a social network with a couple of user domain entities as profile, feed, personal attributes, all in one microservice, which is a UAA as well...

deepu105:
@jdubois it doesnt point to the directory, we ask the user for the directory for uaa then read the app name and save it from the uaa apps `yo-rc.json` look at the filter method for that question

sdoxsee:
It would be better to not use a shared secret for the JWT but rather a
public key (at least in the oauth2 version). I'd recommend asymmetrical
(public/private) over symmetrical (shared) crypto. The authorization server
can sign the token with a private key (that only it knows) and share the
public key with anyone else so that they can verify that the key came from
the authorization server. This can be changed later though.

xetys:
@sdoxsee this is something additional...let me explain

the secret is ONLY for recognizing the client....there is no "client secret" at all, to sign on as the client (don't mix up with user creds!!!). So I choouse "web-app" with no secret as primary web-app client.

But behind this you can still take private/public or symmetrical crypto (as it is now).

so there are currenty 3 lines of defense:
* basic auth for the client
* username + password for the client authenticating the user
* secret key OR private/public key for token validation

and the basic auth is currently hardcoded, since there is only one client...the angular app...

I fear this is confusing a bit :)

sdoxsee:
I was referring to token validation for the JWT.

The client_id/client_secret is for identifying the client app (i.e. angular
SPA in this case). And the PR didn't have a client_secret (which is
fine...since it can't be kept secret anyway in the browser). I think it
shouldn't be hard-coded in the javascript still though...because it's only
for login calls that it would need to be base64-encoded on-the-fly.

xetys:
ok, seems legit, I will PR a JavaScript encoding the upcomming days, and put the client id into ng constants, to make thinks more elegant here.

sdoxsee:
Great. Shouldn't be too different from the monolith oauth2 implementation
in the client:
https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/src/main/webapp/app/services/auth/_auth.oauth2.service.js

PR #3327, opened 2016-04-02
Fix a json syntax error in the user-management pt-br i18n file 

utelemaco:
Fixing a json syntax error in the webapp/i18n/pt-br/user-management.json (a 'd' place in an incorrect place).


ruddell:
@utelemaco it looks like the 'd' should be in the JSON property on the same line.  'createOrEitLabel' should be 'createOrEditLabel'

utelemaco:
@ruddell you are right. Just fix now. Thks

gmarziou:
Thanks for the fix.

PR #3329, opened 2016-04-03
adds a relay token filter to gateway

### LINKED ISSUE #3326: Migration to Spring Cloud RC1 makes JWT token fail in microservices
This is in the master branch, it looks like since we upgraded to the latest Spring Cloud RC1 the JWT tokens are not working anymore (for our 3.0.0 release we were using Spring Cloud M5).

This is because the "Authorization" header is viewed as sensitive (in the ZuulProperties), so it is not forwarded anymore to the microservices.

xetys:
this removes the authorization header from zuuls ignore list

fixes #3326

### DIFF
--- a/generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
+++ b/generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
@@ -7,7 +7,7 @@ import org.springframework.stereotype.Component;
 import java.util.Set;
 
 @Component
-public class RelayTokenFilter extends ZuulFilter {
+public class TokenRelayFilter extends ZuulFilter {
     @Override
     public Object run() {
         RequestContext ctx = RequestContext.getCurrentContext();
--- a/generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
+++ b/generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
@@ -12,7 +12,6 @@ public class TokenRelayFilter extends ZuulFilter {
     public Object run() {
         RequestContext ctx = RequestContext.getCurrentContext();
 
-        // Alter ignored headers as per: https://gitter.im/spring-cloud/spring-cloud?at=56fea31f11ea211749c3ed22
         Set<String> headers = (Set<String>) ctx.get("ignoredHeaders");
         // We need our JWT tokens relayed to resource servers
         headers.remove("authorization");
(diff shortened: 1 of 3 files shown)


xetys:
is this caused by me or some bug occured in travis?

gmarziou:
I don't think so, looks like Travis's hiccup.
Let's close and reopen the PR, this should restart the build.

xetys:
my local test with https://github.com/xetys/jhipster-travis-helper was succeeding :smile:

gmarziou:
This is really great tool, I'll use it.

jdubois:
OK that's simpler than what I wanted to do for #3326 with an "internal JWT token", so that's better.

I'll merge this, but I also found some other minor issues while working on #3326 (mostly formatting or style problem), so I'll do everything at once.

wmarques:
I tested it and it works.

generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Set;
+
+@Component
+public class TokenRelayFilter extends ZuulFilter {
+    @Override
+    public Object run() {
+        RequestContext ctx = RequestContext.getCurrentContext();
+
+        // Alter ignored headers as per: https://gitter.im/spring-cloud/spring-cloud?at=56fea31f11ea211749c3ed22
```
jdubois: I'll change the comment as gitter is not persistent (messages are lost after 15 days)

xetys:
done!

generators/server/templates/src/main/java/package/gateway/_TokenRelayFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Component
+public class TokenRelayFilter extends ZuulFilter {
+    @Override
+    public Object run() {
+        RequestContext ctx = RequestContext.getCurrentContext();
+
+        Set<String> headers = (Set<String>) ctx.get("ignoredHeaders");
+        // We need our JWT tokens relayed to resource servers
+        headers.remove("authorization");
```
jdubois: Shouldn't this be "Authorization"? Is this case-sensitive?

xetys:
it looks like an config value...sinse there is a list of ignored headers.....so this is not EXACTLY the header value

and travis is kidding me again :smile:

PR #3288, opened 2016-04-05
[FEATURE] docker-compose to start Cassandra locally for dev

raphaelbrugier:
Hi,

We want a developer to be able to run `docker-compose -f src/main/docker/somefile.yml build/up`
and then have a Cassandra cluster ready to use locally, with the Keypspace and all the tables + migrations scripts automatically already applied.

##### **Current implementation**
First I want to be sure of my understanding of the generated docker-compose files and the intention behind.
I think it is mainly intended for production:
> You package the app with `./mvnw package -Pprod docker:build`
> Then you run `docker-compose -f src/main/docker/app.yml up`
> And you have the app running in production!

For Cassandra, to have the keyspace and tables ready both in dev and in production, you then run Docker commands to copy the CQL scripts in the container and then execute them.

Our use case is a bit different, and mostly oriented for development first.
We want a developer to be able to run `docker-compose -f src/main/docker/somefile.yml build/up`
and then have a Cassandra cluster ready to use locally, with the Keypspace and all the tables + migrations scripts automatically already applied.

In production, we will run the queries manually.

Currently, it's hard to differentiate the Cassandra docker-compose files for dev and production to achieve this.
- **cassandra.yml** is the default file loaded by *app.ym*l but it uses *cassandra/scripts/init-dev.sh* behind.
- **cassandra-OpsCenter.yml** is never linked, but it uses *cassandra/scripts/init-prod.sh*

#### Modifications:
I have made some changes in the docker/docker-compose files generated for Cassandra by JHipster 3.0 in our app and I'd like to discuss it and see if it's interesting to integrate them in JHipster.
I have created again two different docker-compose files for Cassandra, dev and prod.

The dev docker-compose file will start 3 services:
- **tatami-cassandra**, a container with the Cassandra node contact point
- **tatami-cassandra-node-1**, a container with a second Cassandra node joining the cluster
- **tatami-cassandra-migration** a container to apply the keypsace and schema creation and data migration

The migration service is responsible to apply all the migration scripts from src/main/resources/config/cql in the following order:
1. create-keyspace.cql
2. create-tables.cql
3. all \*_added_entity_\*.cql in alphabetical order
4. all V\*.cql in alphabetical order

Some explanations:
- `create-tables.cql` is the initial bootstrap script. You shouldn't add many database changes after being in production here
- `*_added_entity_*.cql` are the scripts automatically generated by the `yo jhipster:entity` generator
- `V*.cql` are all scripts altering the schemas or data. This is where you will add all the database changes

The idea is to follow incremental database changes principle, where all migration scripts must be applied in a fixed order.
This is why the scripts tend to follow the conventions from [FlywayDB](https://flywaydb.org/documentation/migration/sql.html).

Although there is no tool like FlywayDB or Liquibase for Cassandra, this convention will allow to easily add such tool in the future.

##### I also tried to improve these:
- No failure handling.
  When run manually with `docker exec <container id> init or entities`, if any of the query has a problem then the following queries are still applied.
  This is because cqlsh does not stop on query failure. But the current init script first concat all the CQL files, those making it more dangerous.
  The migration now stops on the fist failing script.
- When adding a new entity model generated with the JHipster entity generator, the CQL script is not added to AbstractCassandraTest
  AbstractCassandraTest now loads all the CQL files
- No need to copy the CQL files in every Cassandra container anymore. Then they can use the base image directly.

##### **JHipster Version(s)**
based on 3.0

##### **Related issues**
It's related to the changes introduced in #3180
I have read the comments but I don't see how it could have covered this need.

##### **Suggest a Fix**
We will use this approach in our project and see how it goes.
But I'd like to know your opinion on how to differentiate docker-compose files for dev and production?
And having the cassandra-dev.yml automatically applying the migration scripts?

See https://github.com/ippontech/tatami/tree/docker-cassandra/src/main/docker for the implementation.


pascalgrimaud:
For me, no objection for this change : cql/changelog/{time}_insert_default_users.cql

@raphaelbrugier : did you transform this issue to a pull request ? I spend a couple of minutes to find this !!

raphaelbrugier:
Yes @pascalgrimaud, I transformed the issue into a pull request!
I didn't want to open a new issue for this with an empty comment just lo link it to this discussion.
Is this unusual?

FYI, this is not something you can do from the Github UI. I had to call their pulls request rest API to do so.

pascalgrimaud:
I didn't know it is possible ! I like it:)

raphaelbrugier:
I reused [this example](http://stackoverflow.com/a/4529172/470107) but had to add a token because I'm using double authentication in GitHub. Worked like a charm!

generators/server/templates/src/main/docker/_cassandra.yml
```diff
@@ -14,10 +12,19 @@ services:
             - 9042:9042
             - 9160:9160
     <%=baseName.toLowerCase()%>-cassandra-node:
-        build:
-            context: .
-            dockerfile: cassandra/Cassandra.Dockerfile
+        image: cassandra:2.2.3
```
pascalgrimaud: it should be 2.2.5

raphaelbrugier: good catch! I've pushed a fix

generators/server/templates/src/main/docker/cassandra/scripts/_autoMigrate.sh
```diff
@@ -0,0 +1,49 @@
...
+}
+
+waitForClusterConnection
+
+log "execute migration scripts"
+
+log "create keyspace and base tables"
+. ./usr/local/bin/init-dev
+
+executeScripts /cql/changelog/*.cql
```
pascalgrimaud: I have to use "executeScripts /cql/changelog/*.cql",
else the executeScripts choose the 1st file in `changelog/*.cql`

generators/server/templates/src/main/docker/cassandra/scripts/_execute-cql.sh
```diff
@@ -0,0 +1,32 @@
...
+log "execute: " $cqlFile
+cqlsh -k $KEYSPACE_NAME -f $cqlFile $CASSANDRA_CONTACT_POINT
+
+if [ $? -ne 0 ]; then
+    log "fail to apply script " $filename
+    log "stop applying database changes"
+    exit 1
+fi
+
+exit 0
```
pascalgrimaud: not sure about this, but I had to remove it too, else the only the 1st cql is executed

pascalgrimaud:
I was ok about this change : `cql/changelog/{time}_insert_default_users.cql`
But does it mean everytime we launch changelog/*.cql, there will be insert_default_users ?

raphaelbrugier:
What do you mean by "launch changelog/*.cql," ?

Here is the implemented behavior
- `cql/changelog/{time}_insert_default_users.cql` will be generated only once, at the initial `yo jhipster`
- In dev: all cql scripts in `cql/changelog/` will be applied everytime you restart the cluster.
- In prod: all cql scripts are applied manually

Does this make sense?

pascalgrimaud:
My steps:
1) generate a new app
2) start cassandra.yml, it will create perfectly all tables!
3) development : add new entities for example (I let cassandra up)
4) launch `docker-compose -f src/main/docker/cassandra.yml up <app>-cassandra-migration` => ̀cql/changelog/{time}_insert_default_users.cql` will be execute again, right ? Just wondering if it can be a problem or not. For example, if the default 'user' is deleted intentionaly by the developer?

In dev: it's not a problem, it's for dev :)
In prod: this `cql/changelog/{time}_insert_default_users.cql` should not be executed. And if it's in `changelog` folder, I'm afraid there can be executed by error

raphaelbrugier:
Ok I see.

Yes you are right, all the changelogs scripts will be applied again.

I didn't think about the step 4) like this.
For me, after generating new entities or adding any cql script either:

- run the cql script manually with CQLsh 

or execute it within the cluster:

- `docker cp src/main/resources/config/cql/newscript.cql <name_of_your_app>-cassandra:/cql/`
- `docker exec -it <name_of_your_app>-cassandra execute-cql /cql/newscript.cql

or restart the cluster:

- `docker-compose -f src/main/docker/cassandra.yml down`
- `docker-compose -f src/main/docker/cassandra.yml up`


To do the step 4) like you describ, I see 2 solutions:
- either all scripts are idempotent
- or implement a changelog tracker system, liquibase-like

You're also right about `cql/changelog/{time}_insert_default_users.cql` being executed in production by mistake. I guess this problem also exists with liquibase, how do you separate the sql changelogs needed only for dev bootstrap?

generators/server/templates/src/main/docker/cassandra/scripts/_execute-cql.sh
raphaelbrugier: I think it is the `exit 0` I have added by mistake when I've extracting this on a separate script. This cause the caller script `autoMigrate.sh` to exit immediately.
I'll fix this and run some more tests.

pascalgrimaud:
liquibase uses the changelog tracker like you said, so old scripts are not executed a 2nd time
For cassandra it's different. My only concern is about this insert_default_users, if it should be in changelog or not. Let me time to think more about it

Faster command to migrate in dev, as the status is "exited", I simply restart it, and I love it!
`docker start jhcassandra-cassandra-migration`

pascalgrimaud:
can you fix npm test too plz?

raphaelbrugier:
I'll fix the npm test.
I'm wondering why I didn't see the failure the first time I opened the PR however.

I'm not sure to understand what you said there:

> Faster command to migrate in dev, as the status is "exited", I simply restart it, and I love it!
> docker start jhcassandra-cassandra-migration

This could _not_ work with Cassandra unless we implement one of the behavior like I described:

> either all scripts are idempotent
> or implement a changelog tracker system, liquibase-like

What would be the "best" way to recommend in the documentation for dev environment?
- whole down/up of the cluster
- or manual migration
- or just start the migration service again (with the risk of running twice the same query)?

pascalgrimaud:
what do you think:
- after executing the cql, copy it somewhere inside the container
- then, next time, before executing the cql, check if it has been copied
- so the cql won't never be executed 2x

raphaelbrugier:
The container is not persistent, has no persistent volume mounted, so you'll lose the modification once it exited.
I could mount the local /tmp/ into this container.
Or we can store in the information in a Cassandra table, but there we are starting to implement a liquibase-like tool with scripts. And I wanted to avoid this for a first iteration on this feature.

pascalgrimaud:
My fault. I'm thinking too much :smile: 
The solution here is already a good improvment! Let's keep this way

raphaelbrugier:
I've pushed two commits to fix the problems you found on the code review yesterday.

On a side note, I find the tests coverage very impressive. The architecture you designed to run the Travis CI on multiple version of the generated apps is very clear and great!

About the migration of new scripts added in dev we were talking, shall we leave it as is?

pascalgrimaud:
Yes leave as this for now

I have to work on JHipster meetup, this week-end. I hope I find time to test and merge.
Thanks for your patience :)

generators/server/templates/src/test/java/package/_AbstractCassandraTest.java
```diff
@@ -29,7 +29,7 @@ public static void startServer() throws InterruptedException, TTransportExceptio
         Session session = cluster.connect();
         CQLDataLoader dataLoader = new CQLDataLoader(session);
         dataLoader.load(new ClassPathCQLDataSet("config/cql/create-tables.cql", true, "cassandra_unit_keyspace"));
-        applyScripts(dataLoader, "config/cql/", "*_added_entity_*.cql");
+        applyScripts(dataLoader, "config/cql/changelog", "*.cql");
```
pascalgrimaud: add / at the end plz:
`config/cql/changelog` => `config/cql/changelog/`

pascalgrimaud:
@raphaelbrugier : theses lines can be removed now https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/05-run.sh#L9-L13
that's why travis failed

raphaelbrugier:
Yes I just realized that.
and there too? https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/03-docker-compose.sh#L10

Also is this still relevant?
"# travis is not stable with docker... need to start container with privileged"
https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/03-docker-compose.sh#L8

I'm afraid I'll have to add a `sleep` after starting the Cassandra cluster with the docker compose command. To wait for the migration to have completed...

pascalgrimaud:
The sleep is not necessary I think, because there is "mvn test" just after, so the containers have time to start

About privileged, I'm not sure, but if I remember well, Travis doesn't allow to use correctly docker exec, without privileged. With your enhancement, no docker exec with dev profile

pascalgrimaud:
The build is green now!
@raphaelbrugier : ok for merging? is it finished?

raphaelbrugier:
Let me try to remove the sleep and the "# travis is not stable with docker... need to start container with privileged" before please.

raphaelbrugier:
Thanks @pascalgrimaud !

pascalgrimaud:
Thank you @raphaelbrugier !

pascalgrimaud:
@raphaelbrugier : I'm testing the cassandra with microservices

I want to discuss with you about this point.
Everytime I want to update the project, by typing `yo jhipster`, there is multiple insert_default_users.cql :
```
ls -l src/main/resources/config/cql/changelog/        
total 12
-rw-r--r-- 1 pgrimaud pgrimaud 1328 avril 15 08:08 20160415060858_insert_default_users.cql
-rw-r--r-- 1 pgrimaud pgrimaud 1328 avril 15 11:51 20160415095136_insert_default_users.cql
-rw-r--r-- 1 pgrimaud pgrimaud 1328 avril 15 14:09 20160415120954_insert_default_users.cql
```

I don't think it's what we want. So what do you think to put back the insert_default_users inside create-tables.cql like before ?

raphaelbrugier:
Ah yes, that's because I used the timestamp of the creation as a prefix.

I don't really like the idea of putting back the queries into `create-tables.sql`
`create-tables.sql` will do more that the name suggest, and having a script into the changelog directory gives an example.

Is it possible to have a fix prefix?
Either store the initial creation of the .yo-rc.json file in a new key or an arbitrary date?

pascalgrimaud:
With 000000000... ?

raphaelbrugier:
What about the Unix Epoch?
19700101000000_insert_default_users

pascalgrimaud:
I proposed 00000... to be consistent with the liquibase changelog: 00000000000000_initial_schema.xml

raphaelbrugier:
Ah ok, I didn't remember the liquibase changeset. It's more consistent to use what you suggested then.

PR #3352, opened 2016-04-06
Npm test for docker compose

pascalgrimaud:
I did some changes on this sub generator:
- if there is no gateway/microservice in the initial folder, exit instead of waiting the correct answer.
I move the validate, because this validate calculated some var which is used further (for not having problem with yeoman mock). Hope I'm clear
- add warning at the end of generator, instead of exiting when no war is found

@wmarques => if you're ok with theses changes, can you test it plz ?

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -291,7 +291,7 @@ module.exports = yeoman.Base.extend({
                 }
                 // Add search engine configuration
                 var searchEngine = this.appConfigs[i].searchEngine;
-                if (searchEngine !== 'no') {
+                if (searchEngine === 'elasticsearch') {
                     var searchEngineYaml = jsyaml.load(this.fs.read(path + '/src/main/docker/' + searchEngine + '.yml'));
                     var searchEngineConfig = searchEngineYaml.services[this.appConfigs[i].baseName.toLowerCase() + '-' + searchEngine];
                     delete searchEngineConfig.ports;
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -78,37 +78,39 @@ module.exports = yeoman.Base.extend({
                 type: 'input',
                 name: 'directoryPath',
                 message: 'Enter the root directory where your gateway(s) and microservices are located',
-                default: '../',
-                validate: function (input) {
-                    var path = this.destinationPath(input);
-                    if(shelljs.test('-d', path)) {
-                        var files = shelljs.ls('-l',this.destinationPath(input));
-                        this.appsFolders = [];
-
-                        files.forEach(function(file) {
-                            if(file.isDirectory()) {
-                                if(shelljs.test('-f', file.name + '/.yo-rc.json')) {
-                                    var fileData = this.fs.readJSON(file.name + '/.yo-rc.json');
-                                    if(fileData['generator-jhipster'].baseName !== undefined) {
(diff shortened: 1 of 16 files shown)


wmarques:
I tested the two "new cases" and it works nicely :+1:

generators/docker-compose/index.js
```diff
@@ -78,37 +78,39 @@ module.exports = yeoman.Base.extend({
                 type: 'input',
                 name: 'directoryPath',
                 message: 'Enter the root directory where your gateway(s) and microservices are located',
-                default: '../',
-                validate: function (input) {
```
deepu105: One disadvantage of moving validate from here is that user will not get a chance to correct his error, inquirer will stay on the question for user to rnter correct path normally when there is a validation error

deepu105:
I have a query. Wouldn't it be ideal to generate the war files at test run time rather than providing static wars?  That way we can test tje entire workflow. Ofcourse it will take more time but our travis runs for more than an hour anyway so this wont add anything more to it as it would be run in parellel.  May be we can have it as the first test in matrix

generators/docker-compose/index.js
pascalgrimaud: yes I agree with you, about this disadvantage, but the yeoman mock bypass the validate. There are some topics about this
Maybe there is a better way...?

pascalgrimaud:
Sorry @deepu105 I dont understand your last message :pensive: 
Are you talking about empty `test/templates/compose/01-gateway/target/docker/jhgate-test.war` ?
If yes, it's to avoid warning message / shelljs.ls error, during npm test

@wmarques : what do you think about replacing the check on *.war by `docker images` instead ?

wmarques:
@pascalgrimaud Agree, I didn't know about this command but it seems more relevant than checking the war

wmarques:
@pascalgrimaud Shall I merge this and we'll do the `docker images `usage in another PR and we do it here ?

pascalgrimaud:
another PR is better
Le 7 avr. 2016 5:01 PM, "William Marques" <notifications@github.com> a
écrit :

wmarques:
@deepu105 do you think there is a trick to permit the user to correct his error without using validate ? I don't really like the idea of leaving instantly the generator, could be frustrating

deepu105:
@wmarques there might be a way let me take a look on the weekend

deepu105:
@pascalgrimaud sorry I read too quickly of course what you have done for the empty war makes sense. I initially thought that the test was on travis and I was thinking in that case we could actually have a travis matrix to run the docker compose flow and test out a gateway and microservice with registry

deepu105:
@pascalgrimaud @wmarques the only way I could make both work was making it an util function. see if its ok in master

@wmarques btw I noticed that if i re run the sub gen on an existing folder the regeneration doesnt happen and I understand some logic is done in promts, but there might be ways to work around that. do you want me to do that or you guys wanted re generation to ask questions?

pascalgrimaud:
@deepu105 : it seems good with function!

wmarques:
@deepu105 I want to have the questions again but I want the subgenerator to store the previous choices I made. I made it like that because sometimes the user will want to add/remove services, remove the monitoring, enable or disable the cluster mode for databases... So instead of updating by hand the docker compose file, he just has to rerun the subgenerator.

deepu105:
Ok

PR #3354, opened 2016-04-06
Fix Protractor tests

mraible:
Fix location of selenium-server JAR and turn off directConnect since it
doesn’t work with Firefox 45.0.1.


mraible:
Closing because too many commits, need to squash.

PR #3360, opened 2016-04-07
Code Enhancement in AccessControlFilter.java

### LINKED ISSUE #3358: [ENHANCEMENT] Re-factor code in AccessControlFilter.java

Unnecessary looping in code when endpoint is found to be not authorized

##### **JHipster Version(s)** 

verison 3.0.0

##### **Suggest a Fix** 

terminate the loop when it is found that endpoint is not authorized

find the code below

`public boolean shouldFilter() {
        String requestUri = RequestContext.getCurrentContext().getRequest().getRequestURI();

        // If the request Uri does not start with the path of the authorized endpoints, we block the request
        for (Route route : routeLocator.getRoutes()) {
            String serviceUrl = route.getFullPath();
            String serviceName = route.getId();

            // If this route correspond to the current request URI
            // We do a substring to remove the "**" at the end of the route URL
            if (requestUri.startsWith(serviceUrl.substring(0, serviceUrl.length() - 2))) {
                if (isAuthorizedRequest(serviceUrl, serviceName, requestUri)) {
                    return false;
                }else{
					return true;
				}
            }
        }
        return true;
    }`

kkorada:
stopped Unnecessary looping in code when endpoint is found to be not authorized

Fix #3358

### DIFF
--- a/generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
+++ b/generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
@@ -53,11 +53,7 @@ public class AccessControlFilter extends ZuulFilter {
             // If this route correspond to the current request URI
             // We do a substring to remove the "**" at the end of the route URL
             if (requestUri.startsWith(serviceUrl.substring(0, serviceUrl.length() - 2))) {
-                if (isAuthorizedRequest(serviceUrl, serviceName, requestUri)) {
-                    return false;
-                }else{
-					return true;
-				}
+				return !isAuthorizedRequest(serviceUrl, serviceName, requestUri);
             }
         }
         return true;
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/gateway/accesscontrol/_AccessControlFilter.java
```diff
@@ -55,7 +55,9 @@ public boolean shouldFilter() {
             if (requestUri.startsWith(serviceUrl.substring(0, serviceUrl.length() - 2))) {
                 if (isAuthorizedRequest(serviceUrl, serviceName, requestUri)) {
                     return false;
-                }
+                }else{
```
notflorian: In that case, instead of doing `if (isAuthorizedRequest(serviceUrl, serviceName, requestUri)) { return false; } else { return true; }` prefer doing ` return !isAuthorizedRequest(serviceUrl, serviceName, requestUri);`

kkorada: yes that's good i will do it.

PierreBesson:
@kkorada Thanks for taking the time to improve my code and submit a PR.

PR #3370, opened 2016-04-09
Added support for two more languages

konstiak:
- Slovak
- Czech


deepu105:
@konstiak can you squash some of the unwanted commits so that the commit history is cleaner. Stuff like `	Fixed small mistakes.` ,  `Finished translation to Slovak` , `Merge remote-tracking branch 'upstream/master'`  etc.

Basically just have the commits which makes sense overall. Will merge this once the commits are cleaned

pascalgrimaud:
@deepu105 : when merging, we can now squash. It's a new feature from github, added recently. Can you test it ?

deepu105:
@pascalgrimaud yes I have done that earlier but it squashes everything to a single commit. In this PRs case @konstiak might want to retain few commits thats makes sense.

pascalgrimaud:
good to know it works!

konstiak:
@pascalgrimaud I think you can squash it to a single commit. (Added Czech and Slovak localization). I can squash it to multiple commits, but it's only translation and I think it won't make more sense than one commit. Thanks.

Issue #3372, opened 2016-04-11
Fix auto focus on password reset UI (need to update reset.request.controller.js)


gtiwari333:
* **Overview of the request**

Auto focus on password reset screen is not working. I'm using jhipster generator version : 3.0.0 .

* **Motivation for or Use Case** 

Its a small bug, better if it fixed in next release.

* **Browsers and Operating System** 

ALL

* **Related issues** 

Did not noticed any existing report to this issue/bug.

* **Suggest a Fix** 

reset.request.controller.js needs to be updated to include  `vm.` around the $timeout logic to match what we have in reset.request.html file  : 

`        $timeout(function (){angular.element('[ng-model="vm.resetAccount.email"]').focus();});
`
The generated code is missing "vm."

deepu105:
Can you do a PR? also can you change to fetch element by ID instead of
attribute?

Thanks & Regards,
Deepu

gtiwari333:
Okay, I will do a PR. 

FYI, the following four .js files has the same issue ( using attribute instead of ID). I will update them all.

- login.controller.js
- register.controller.js
- reset.request.controller.js
- reset.finish.controller.js

gtiwari333:
Submitted pull request https://github.com/jhipster/generator-jhipster/pull/3373

wmarques:
Fixed by #3373 
@gtiwari333 please add "Fix #issuenumber" on the PR next time, it will autoclose the issue when we merge it :)

PR #3373, opened 2016-04-11
Fixing Auto Focus and Using ID instead of attribute - Issue #3372 

gtiwari333:
Issue https://github.com/jhipster/generator-jhipster/issues/3372

### DIFF
--- a/generators/client/templates/src/test/javascript/spec/app/account/reset/finish/_reset.finish.controller.spec.js
+++ b/generators/client/templates/src/test/javascript/spec/app/account/reset/finish/_reset.finish.controller.spec.js
@@ -54,7 +54,7 @@ describe('Controller Tests', function() {
 
             // then
             expect(MockTimeout).toHaveBeenCalledWith(jasmine.any(Function));
-            expect(MockAngular.element).toHaveBeenCalledWith($('#password'));
+            expect(MockAngular.element).toHaveBeenCalledWith($('#password').val());
             expect(MockElement.focus).toHaveBeenCalled();
         });
 
--- a/generators/client/templates/src/test/javascript/spec/app/account/reset/finish/_reset.finish.controller.spec.js
+++ b/generators/client/templates/src/test/javascript/spec/app/account/reset/finish/_reset.finish.controller.spec.js
@@ -54,7 +54,7 @@ describe('Controller Tests', function() {
 
             // then
             expect(MockTimeout).toHaveBeenCalledWith(jasmine.any(Function));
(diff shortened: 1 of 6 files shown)


gtiwari333:
working on fixing the build.

deepu105:
builds are still failing

generators/client/templates/src/test/javascript/spec/app/account/reset/finish/_reset.finish.controller.spec.js
```diff
@@ -54,7 +54,7 @@ describe('Controller Tests', function() {
 
             // then
             expect(MockTimeout).toHaveBeenCalledWith(jasmine.any(Function));
-            expect(MockAngular.element).toHaveBeenCalledWith('[ng-model="vm.resetAccount.password"]');
```
deepu105: This might be causing build failures. I dont think we need to change this

gtiwari333: After I updated  `_reset.finish.controller.js` and `_reset.request.controller.js` to use ID instead of attribute, it caused the test to failed with the following errors : 

 
reset.finish.controller.spec.js :
`Expected spy MockAngular.element to have been called with [ '[ng-model="vm.resetAccount.password"]' ] but actual calls were [ '#password' ].`
 
reset.request.controller.spec.js :
`Expected spy MockAngular.element to have been called with [ '[ng-model="resetAccount.email"]' ] but actual calls were [ '#email' ]. `


I tried to use following instead of `[ '[ng-model="vm.resetAccount.password"]' ]` and `[ ng-model="resetAccount.email"]' ]`   but they didn't work : 

```
$('#password')
$('#password').val()
```

deepu105: did you try this `expect(MockAngular.element).toHaveBeenCalledWith('#password');` ?

gtiwari333: Thanks for suggestion. Just pushed the change. Its building !

gtiwari333: All tests passed !

deepu105:
@wmarques next time use the new squash feature to merge the commits if there are lot of stray commits for small enchancements so that our commit history is clean

PR #3381, opened 2016-04-13
Fix escaped characters in generated docker compose YML

wmarques:
The `&` character where espaced into `&amp;` which caused issues in the  datasource url envirnoment variable (ie: `SPRING_DATASOURCE_URL=jdbc:mysql://mysql:3306/sample?useUnicode=true&characterEncoding=utf8&useSSL=false`)

Do I win the award of the smallest PR in the world ? :smile:


cicoub13:
One character can break the world!

pascalgrimaud:
proud to merge this !!!

PR #3387, opened 2016-04-14
Don't register cors filter if cors property is not set

### LINKED ISSUE #3318: CORS fails in microservice environment
I've enabled CORS on the microservice & on both gateway and microservice but kept getting a 403 error as below. I can successfully access the service via the browser but fails when done via an AJAX based- app. I ran the gateway & microservice directly from the executable WAR file.

```
cors:
allowed-origins: ""
allowed-methods: GET, PUT, POST, DELETE, OPTIONS
allowed-headers: ""
exposed-headers:
allow-credentials: true
max-age: 1800
```

response header for OPTIONS preflight call:

> HTTP/1.1 403 Forbidden
> Server: Apache-Coyote/1.1
> X-Application-Context: HPGateway:dev:8082
> Allow: GET, HEAD, POST, PUT, DELETE, TRACE, OPTIONS, PATCH
> Content-Length: 20
> Date: Fri, 01 Apr 2016 11:38:13 GMT

here's my config:

```
{
"generator-jhipster": {
"jhipsterVersion": "3.0.0",
"baseName": "HPWebServices",
"packageName": "com.company",
"packageFolder": "com/company",
"serverPort": "8084",
"authenticationType": "jet",
"hibernateCache": "no",
"databaseType": "mongodb",
"devDatabaseType": "mongodb",
"searchEngine": "no",
"buildTool": "maven",
"jwtSecretKey": "xxxxxx",
"enableTranslation": true,
"testFrameworks": [
"gatling"
],
"skipClient": true,
"skipUserManagement": true,
"nativeLanguage": "en",
"languages": [
"en",
"fr"
]}}
```

cbornet:
Fix #3318

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
+++ b/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
@@ -10,6 +10,7 @@ import <%=packageName%>.web.filter.CachingHttpHeadersFilter;<% } %>
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.embedded.ConfigurableEmbeddedServletContainer;
 import org.springframework.boot.context.embedded.EmbeddedServletContainerCustomizer;
 import org.springframework.boot.context.embedded.MimeMappings;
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/config/_WebConfigurer.java
```diff
@@ -171,14 +172,14 @@ private void initMetrics(ServletContext servletContext, EnumSet<DispatcherType>
...
     public CorsFilter corsFilter() {
+        log.debug("Registering CORS filter");
         UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
-        CorsConfiguration config = jHipsterProperties.getCors();
-        if (config.getAllowedOrigins() != null && !config.getAllowedOrigins().isEmpty()) {
-            source.registerCorsConfiguration("/api/**", config);
-            source.registerCorsConfiguration("/v2/api-docs", config);
-            source.registerCorsConfiguration("/oauth/**", config);
-        }
+        CorsConfiguration config = props.getCors();
```
wmarques: Seems that props variable is undefined

cbornet: Oops !

PR #3389, opened 2016-04-14
add a length limit for logstash-encoder forwarded stacktraces fix #3364

### LINKED ISSUE #3364: Log statements with stack trace get truncated due to Logstash 8 KB limit

When app logs an exception stack trace and the JSON encoded log statement is larger than 8 KB, Logstash is confused and truncate the statement resulting in invalid JSON fragment.

This limit is the default one in Logstash: https://github.com/elastic/logstash/issues/1505

Logs with exceptions are very important and should contain all required info for investigation while possibly optimizing the volume of data transferred to Logstash.

##### **JHipster Version(s)** 

3.0 master

##### **Reproduce the error** 

Log an error with very long text and exception:

~~~java
log.error("Very log text .....", exception );
~~~

##### **Suggest a Fix** 

logstash-logback-encoder has [few options to limit stack trace depth](https://github.com/logstash/logstash-logback-encoder#customizing-stack-traces).

In particular, the rootFirst option which prints the root cause first could be combined with maxLength in bytes.

PierreBesson:
Fix #3364 issue for stacktraces over 8K sent through an UDP socket.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_LoggingConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_LoggingConfiguration.java
@@ -3,6 +3,7 @@ package <%=packageName%>.config;
 import ch.qos.logback.classic.AsyncAppender;
 import ch.qos.logback.classic.LoggerContext;
 import net.logstash.logback.appender.LogstashSocketAppender;
+import net.logstash.logback.stacktrace.ShortenedThrowableConverter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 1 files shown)


deepu105:
@gmarziou can we merge this?

gmarziou:
I would prefer to wait for Pierre to answer my comments in issue.

gmarziou:
Or we could merge it and then refine it later if Pierre agrees.

PierreBesson:
@deepu105, just doing a last test to see if I need to add one more line to the config or not.

PierreBesson:
@deepu105, @gmarziou, Stacktraces are now written in reverse order when forwarded to logstash. This does not affect regular console logging. Thanks @gmarziou for insisting on this.

gmarziou:
Great work and it's easy to be adjusted by user if needed.

PR #3397, opened 2016-04-15
Store previous state in sessionStorage instead of rootScope in preparat…

### LINKED ISSUE #3395: Store previous state in sessionStorage instead of rootScope

In order to get OAuth2 SSO with different gateway apps we'll need to redirect to the authorization server and leave the SPA. Storing the previousState in localStorage rather than rootScope will enable us to bring the user back to the originally requested (but protected) page.

For example, http://localhost:8080/#/settings is protected--prompting a login (on the external authorization server). If requested directly, the user should be directed back to the requested page (not just "home") following authorization on the external authorization server. Saving the previousState info in localStorage allows us to do this.

This works now and will get us ready for OAuth2

##### **Suggest a Fix** 

PR coming

sdoxsee:
…ion for an external oauth2 authorization server where rootScope state is lost upon redirect. Fix #3395

### DIFF
--- a/generators/client/templates/src/main/webapp/app/blocks/handlers/_state.handler.js
+++ b/generators/client/templates/src/main/webapp/app/blocks/handlers/_state.handler.js
@@ -5,10 +5,10 @@
         .module('<%=angularAppName%>')
         .factory('stateHandler', stateHandler);
 
-    stateHandler.$inject = ['$rootScope', '$state', <% if (enableTranslation) { %>'$translate', '<%=jhiPrefixCapitalized%>LanguageService', 'translationHandler',<% } else { %> '$window', <% } %>
+    stateHandler.$inject = ['$rootScope', '$state', '$localStorage', <% if (enableTranslation) { %>'$translate', '<%=jhiPrefixCapitalized%>LanguageService', 'translationHandler',<% } %> '$window',
         'Auth', 'Principal', 'ENV', 'VERSION'];
 
-    function stateHandler($rootScope, $state, <% if (enableTranslation) { %>$translate, <%=jhiPrefixCapitalized%>LanguageService, translationHandler,<% } else { %> $window, <% } %>
+    function stateHandler($rootScope, $state, $localStorage, <% if (enableTranslation) { %>$translate, <%=jhiPrefixCapitalized%>LanguageService, translationHandler,<% } %> $window,
         Auth, Principal, ENV, VERSION) {
         return {
             initialize: initialize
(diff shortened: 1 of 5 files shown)


deepu105:
@sdoxsee seems like your protractor tests are failing. I restarted few times and still it fails. CAN YOU CHECK IF IT PASSES LOCALLY?

generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
```diff
@@ -48,9 +48,8 @@
 
                 // If we're redirected to login, our
                 // previousState is already set in the authExpiredInterceptor. When login succesful go to stored state
-                if ($rootScope.redirected && $rootScope.previousStateName) {
-                    $state.go($rootScope.previousStateName, $rootScope.previousStateParams);
-                    $rootScope.redirected = false;
```
deepu105: I forgot why I added this but it was done for some reason. Are you sure everything works fine without this?

sdoxsee: Thanks @deepu105. I think the previousStateName being set acts as the redirected flag itself. I've updated the PR to also clear the previousState before redirecting to it. The PR should be passing now but I the build is broken from https://github.com/jhipster/generator-jhipster/commit/72886856a850c61ae4ef81a7c8aaf29c93433768. I'll close and reopen this once the build is fixed.

sdoxsee:
Looks like Travis is happy now. I've altered my original PR to use sessionStorage. Let me know if there are any other suggestions.

generators/client/templates/src/main/webapp/app/blocks/handlers/_state.handler.js
```diff
@@ -62,15 +59,6 @@
                     stateChangeSuccess();
                 }
             });
-
-            function back() {
```
deepu105: isnt this used anywhere?

sdoxsee: Not anymore. Even before my PR, this was orphaned code. See how where it is still in use in 2.27.2:
- https://github.com/jhipster/generator-jhipster/blob/v2.x_maintenance/app/templates/src/main/webapp/scripts/app/_app.js#L64
- https://github.com/jhipster/generator-jhipster/blob/v2.x_maintenance/app/templates/src/main/webapp/scripts/app/account/login/_login.controller.js#L21

The intention for it seems to have been that if a previousState was set due to unauthorized access of a page, the login screen would be shown, followed by a "back()" to restore the previousState. But, there is nothing calling this "back()" function in 3.0 since it is being handled in the login callback.

deepu105: Ok then

PR #3398, opened 2016-04-15
Fix cassandra noi18n

### LINKED ISSUE #3394: Failure when generating with Cassandra and no i18n (master)
In the master, the generator does not work when trying to generate with a Cassandra database and not internationalization

##### **JHipster Version(s)** 
master (a1ed454)

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.0.0",
    "baseName": "i18n",
    "serverPort": "8080",
    "hibernateCache": "no",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "cassandra",
    "searchEngine": "no",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "3e8df6fe47976edc3f9f48f15d0c2cb241c94134",
    "useSass": false,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
```

##### **Reproduce the error** 
run yo:jhipster on the .yo-rc.json above

##### **Suggest a Fix** 
I've tracked in the git history the regression was introduced by commit 8683b49 but I have not found the root cause yet.
Maybe @wmarques have an idea?
Otherwise I'll continue my analysis.

What do you thing to also add a test to in the samples to cover this case?
I don't know how much it cost - in build time - to add a new configuration.

pascalgrimaud:
fix #3394

### DIFF
--- a/generators/util.js
+++ b/generators/util.js
@@ -175,7 +175,7 @@ function geti18nJson(key, _this, template) {
         file = JSON.parse(file);
         return file;
     } catch (err) {
-        this.log('error' + err);
+        _this.log('error' + err);
         // 'Error reading translation file!'
         return undefined;
     }
--- a/test/test-creation.js
+++ b/test/test-creation.js
@@ -639,6 +639,40 @@ describe('JHipster generator', function () {
         });
     });
 
+    describe('cassandra noi18n', function () {
+        beforeEach(function (done) {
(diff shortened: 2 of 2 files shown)


pascalgrimaud:
The 1st commit should fail, because we test the .yo-rc.json : cassandra+noi18n
The 2nd commit should fix

PR #3400, opened 2016-04-15
Add on option to report actuator metrics in the logs

PierreBesson:
The goal is to retrieve useful metrics that are not already reported by Dropwizard metrics. It will be especially useful for retrieve circuit breakers metrics.

I tried to make this as simple as possible. I'm not a big fan of my "exclusion list" but I don't see a better way.

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -1086,6 +1086,7 @@ module.exports = JhipsterServerGenerator.extend({
             this.template(SERVER_MAIN_SRC_DIR + 'package/config/locale/_package-info.java', javaDir + 'config/locale/package-info.java', this, {});
             this.template(SERVER_MAIN_SRC_DIR + 'package/config/locale/_AngularCookieLocaleResolver.java', javaDir + 'config/locale/AngularCookieLocaleResolver.java', this, {});
 
+            this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_LogMetricWriter.java', javaDir + 'config/metrics/LogMetricWriter.java', this, {});
             if (this.databaseType === 'cassandra') {
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_package-info.java', javaDir + 'config/metrics/package-info.java', this, {});
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_JHipsterHealthIndicatorConfiguration.java', javaDir + 'config/metrics/JHipsterHealthIndicatorConfiguration.java', this, {});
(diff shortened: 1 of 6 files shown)


gmarziou:
> I'm not a big fan of my "exclusion list" but I don't see a better way.

I'm a bit confused by these 2 sources of metrics, we have those coming from dropwizard and these coming from actuator, the 2 lists overlap.
Your approach is to build an exclusion list to exclude from actuator the ones that are already reported by dropwizard, shouldn't it be the opposite as actuator is the spring boot standard?

cbornet:
Actually, the spring-cloud metrics should use dropwizard when detected on the classpath but it doesn't seem to be the case.
As for the exclusion, maybe we need to find (write?) a metric reader more appropriate than the MetricsEndpointMetricReader

cbornet:
@PierreBesson Since you initially want metrics from Spectator, can you try replacing MetricsEndpointMetricReader by SpectatorMetricReader ?

cbornet:
And probably you need to annotate the SpectatorMetricReader bean  as @ExportMetricReader

cbornet:
@gmarziou I finally understood ! All dropwizard metrics are exported to public metrics so we just need to deactivate (remove) the dropwizard's slf4jreporter and not filter on the public metrics. This way we get more metrics so :+1: for me

cbornet:
We could also prepare some support for the other exporters available in SB (statsd, redis, openTSDB, JMX) instead of the dropwizard's reporter (graphite, spark)

gmarziou:
@cbornet Awesome!

cbornet:
Arf ! Just saw an issue in this plan : dropwizard's TIMER metrics will be split in multiple metrics so this will generate much more logs...

PierreBesson:
@gmarziou I share your concern for the loss of performance of this solution.

Actually, the issue is that spring boot actuator metrics and dropwizard metrics don't have the same format and don't do the same thing. Dropwizard being the more complete solution, we should not move away from it.
Actuator metrics are are less complete than dropwizard metrics, supporting only COUNTER and GAUGE and not TIMER and HISTOGRAM as those are Dropwizard specific.

For Actuator metrics, logging reporting of the metrics coming from dropwizard is too naive, which result in too many logs. They are not really designed to be reported through the logs. I think they want you to access them through the `/metrics` endpoint with some query to filter them. Hence why they are organised hierarchically.

PierreBesson:
@cbornet Using SpectatorMetricReader is working !
If no one object, I will switch my PR to report only metrics from Spectator Metrics. So it will only be for microservices.
Too bad, we won't have Public Metrics but they are not really needed for us because we already have dropwizard metrics.

@cbornet, Yes spring boot have other exporters but I think that those are not as good as passing through logstash which already have many output plugins.

PierreBesson:
As we have decided not to report all metrics from the endpoint, I'm closing this PR and opening a new one that concerns only Circuit Breakers metrics for microservices and gateways.

Issue #2889, opened 2016-02-15
Use gulp-inject instead of wiredep


deepu105:
Switch to gulp-inject from wiredep

This will give following benefits on top of what wiredep gives

1. we can remove the manual inject code [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/index.js#L473) , [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity/index.js#L1610) and [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js#L35)
2. when in serve mode will automatically add user scripts to index

cc @erikkemperman

erikkemperman:
@deepu105 While I think it's possible to use gulp-inject to replace both wiredep and usemin, it might make sense to only replace gulp-usemin for the moment? We had some talk about this on the other issue, and looked at angular fullstack, which uses a combination of gulp-inject and wiredep.

deepu105:
Im ok either way @erikkemperman do you wanna try this out?

erikkemperman:
I think that would be an interesting experiment, but I'm going to be really busy with some other things the next week, maybe even a bit longer...

erikkemperman:
For anyone wanting to take a stab at this, the [gulpfile](https://github.com/angular-fullstack/generator-angular-fullstack/blob/master/app/templates/gulpfile.babel(gulp).js) of angular fullstack might be provide some inspiration. I especially like their use of `lazypipe` for re-usable gulp blocks.

deepu105:
I looked at it but was a lot of tasks, anyway this is not very important
now can be done later also after 3.0 release

Thanks & Regards,
Deepu

deepu105:
Wiredep is not maintained anymore it seems.

https://github.com/taptapship/wiredep

So we should find an alternative solution soon

PR #3403, opened 2016-04-15
Replaced Wiredep completely with gulp inject

sendilkumarn:
#2889

### DIFF
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -23,7 +23,8 @@ var gulp = require('gulp'),<% if(useSass) { %>
     changed = require('gulp-changed'),
     gulpIf = require('gulp-if'),
     inject = require('gulp-inject'),
-    angularFilesort = require('gulp-angular-filesort');
+    angularFilesort = require('gulp-angular-filesort'),
+    bowerFiles = require('main-bower-files');
 
 var handleErrors = require('./gulp/handleErrors'),
     serve = require('./gulp/serve'),
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -156,7 +156,7 @@ gulp.task('inject:vendor', function () {
 gulp.task('inject:test', function () {
     return gulp.src(config.test + 'karma.conf.js')
         .pipe(plumber({errorHandler: handleErrors}))
-        .pipe(inject(gulp.src(bowerFiles({filter:['**/*.js','!**/ngStorage.js','!**/angular-cache-buster.js']}).concat(['**/ngStorage.js','**/angular-cache-buster.js'])), {
+        .pipe(inject(gulp.src(['**/dist/jquery.js','**/angular.js'].concat(bowerFiles({filter:['**/*.js','!**/jquery.js','!**/angular.js']}))), {
             starttag: '// bower:js',
             endtag: '// endbower',
             transform: function (filepath, file, i, length) {
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -157,6 +157,8 @@ gulp.task('inject:test', function () {
     return gulp.src(config.test + 'karma.conf.js')
         .pipe(plumber({errorHandler: handleErrors}))
         .pipe(inject(gulp.src(bowerFiles(), {read: false}), {
+            starttag: '// bower:js',
+            endtag: '// endbower',
             transform: function (filepath, file, i, length) {
                 return '"' + filepath + '",';
             }
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -15,7 +15,7 @@ var gulp = require('gulp'),<% if(useSass) { %>
     protractor = require('gulp-protractor').protractor,<% } %>
     es = require('event-stream'),
     flatten = require('gulp-flatten'),
-    del = require('del'),    
+    del = require('del'),
     runSequence = require('run-sequence'),
     browserSync = require('browser-sync'),
     KarmaServer = require('karma').Server,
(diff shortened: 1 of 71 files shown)


generators/client/templates/gulpfile.js
```diff
@@ -125,44 +124,40 @@ gulp.task('styles', [<% if(useSass) { %>'sass'<% } %>], function () {
...
 });
 
-gulp.task('wiredep', ['wiredep:test', 'wiredep:app']);
 
-gulp.task('wiredep:app', function () {
+gulp.task('inject:vendor', function () {
     var stream = gulp.src(config.app + 'index.html')
         .pipe(plumber({errorHandler: handleErrors}))
-        .pipe(wiredep())
+        .pipe(inject(gulp.src(bowerFiles()),{
```
deepu105: dont youhave to require main-bower-files?

generators/client/templates/gulpfile.js
```diff
@@ -125,44 +124,40 @@ gulp.task('styles', [<% if(useSass) { %>'sass'<% } %>], function () {
         .pipe(browserSync.reload({stream: true}));
 });
 
-gulp.task('inject', function () {
+gulp.task('inject', ['inject:app','inject:test', 'inject:vendor']);
```
deepu105: i think we should have two tasks for this

`gulp.task('inject:dep', ['inject:test', 'inject:vendor']);` and 
`gulp.task('inject:app', ['inject:app']);`

deepu105:
@sendilkumarn could you check the issues

sendilkumarn:
@deepu105 the ordering in karma conf js is not coming correctly. Can you check gulp inject:test ?

generators/client/templates/gulpfile.js
deepu105: @sendilkumarn can you change this as well please as all these shouldn't be run together and in some tasks these needs to be run in certain sequence along with other tasks

generators/client/templates/gulpfile.js
```diff
@@ -276,20 +274,20 @@ gulp.task('watch', function () {
...
     gulp.watch(<% if(useSass) { %>config.sassSrc<% } else { %>config.app + 'content/css/**/*.css'<% } %>, ['styles']);
     gulp.watch(config.app + 'content/images/**', ['images']);
-    gulp.watch(config.app + 'app/**/*.js', ['inject']);
+    gulp.watch(config.app + 'app/**/*.js', ['inject:app']);
     gulp.watch([config.app + '*.html', config.app + 'app/**', config.app + 'i18n/**']).on('change', browserSync.reload);
 });
 
 gulp.task('install', function () {
-    runSequence(['wiredep', 'ngconstant:dev']<% if(useSass) { %>, 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>, 'inject');
+    runSequence(['inject', 'ngconstant:dev']<% if(useSass) { %>, 'sass'<% } %><% if(enableTranslation) { %>, 'languages'<% } %>, 'inject:app');
```
deepu105: inject shouldnt be run first, only `inject:dep` needs to run first.

test/test-languages.js
```diff
@@ -46,7 +46,7 @@ describe('JHipster generator languages', function () {
                     .inTmpDir(function (dir) {
                         fse.copySync(path.join(__dirname, '../test/templates/default'), dir);
                     })
-                    .withOptions({'skip-wiredep': true})
+                    .withOptions({'skip-inject': true})
```
deepu105: @sendilkumarn no need to change this, let it be `skip-wiredep`

generators/languages/index.js
```diff
@@ -169,11 +169,11 @@ module.exports = LanguagesGenerator.extend({
     },
 
     install: function () {
-        var wiredepAddedBowerOverrides = function () {
-            this.spawnCommand('gulp', ['wiredep']);
+        var injectAddedBowerOverrides = function () {
+            this.spawnCommand('gulp', ['inject']);
```
deepu105: this should be `inject:dep`

deepu105:
@sendilkumarn you need to use something similar to 'gulp-angular-filesort' to get the order right, that was the major challenge in moving to inject from wiredep as wiredep takes care of proper ordering of dependency

generators/client/templates/gulpfile.js
```diff
@@ -125,46 +125,44 @@ gulp.task('styles', [<% if(useSass) { %>'sass'<% } %>], function () {
...
-                js: {
-                    block: /(([\s\t]*)\/\/\s*bower:*(\S*))(\n|\r|.)*?(\/\/\s*endbower)/gi,
-                    detect: {
-                        js: /'(.*\.js)'/gi
-                    },
-                    replace: {
-                        js: '\'src/{{filePath}}\','
-                    }
-                }
+        .pipe(inject(gulp.src(['**/dist/jquery.js','**/angular.js'].concat(bowerFiles({includeDev:true,filter:['**/*.js','!**/jquery.js','!**/angular.js']}))), {
```
deepu105: I dont think this is a generic and nice way to do it. This will also confuse the users. Ideal handling should be using the below code in bower file.
```
  "overrides": {
    "angular" : {
      "dependencies": {
        "jquery": "2.2.2"
      }
    },
    <%_ if (useSass) { _%>
    "bootstrap-sass": {
      "main": [
        "assets/stylesheets/_bootstrap.scss"
      ]
    }
    <%_ } else { _%>
    "bootstrap": {
      "main": [
        "dist/css/bootstrap.css"
      ]
    }
    <%_ } _%>
  },
```

deepu105: the override has jquery as a dependency for angular, we can use this to sort the output of  `bowerFiles()` using a custom function

deepu105: @sendilkumarn check this https://github.com/klei/gulp-inject/issues/65

generators/client/templates/gulpfile.js
```diff
@@ -125,46 +125,44 @@ gulp.task('styles', [<% if(useSass) { %>'sass'<% } %>], function () {
...
+                return '<script src="bower_components' + filepath.replace(config.bower,'') + '"></script>';
+            }
+        }))
         .pipe(gulp.dest(config.app));
-
+        
     return <% if (useSass) { %>es.merge(stream, gulp.src(config.sassSrc)
         .pipe(plumber({errorHandler: handleErrors}))
-        .pipe(wiredep({
+        .pipe(inject(gulp.src(config.bower,{
```
deepu105: are sass files being injected properly? what about css files?

PR #3404, opened 2016-04-15
add monitoring props in monolith docker-compose env variables

PierreBesson:
Make the docker-compose subgen for monolith automatically activate logs and metrics monitoring if the ELK option as been selected. Those are the same props that we set by default for microservices using the config server.

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -317,10 +317,10 @@ module.exports = DockerComposeGenerator.extend({
 
                 // Add monitoring configuration for monolith directly in the docker-compose file as they can't get them from the config server
                 if(this.appConfigs[i].applicationType === 'monolith' && this.useElk) {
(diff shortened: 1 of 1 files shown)


generators/docker-compose/index.js
```diff
@@ -315,6 +315,14 @@ module.exports = DockerComposeGenerator.extend({
                     portIndex++;
                 }
 
+                // Add monitoring configuration for monolith directly in the docker-compose file as they can't get them from the config server
+                if(this.appConfigs[i].applicationType === 'monolith' && this.useElk) {
+                    yamlConfig.environment.push("JHIPSTER_LOGGING_LOGSTASH_ENABLED=true")
```
wmarques: Please replace with single quote, just for code consistency

pascalgrimaud:
It can be interesting to add npm test with monolith, right?

PR #3410, opened 2016-04-15
Define a constant instead of duplicating this literal "user" 4 times.

sgrillon14:
Define a constant instead of duplicating this literal "baseUrl" 3 times.
Define a constant instead of duplicating this literal "user" 4 times.

SONAR: Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences. On the other hand, constants can be referenced from many places, but only need to be updated in a single place.


jdubois:
I'm not a huge fan of this, it adds 2 lines of code and doesn't really improve anything... But let's not be too picky, let's merge this.

PR #3411, opened 2016-04-15
Define a constant instead of duplicating this literal "You have misco…

sgrillon14:
…nfigured your application! " 3 times.

SONAR: Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences. On the other hand, constants can be referenced from many places, but only need to be updated in a single place.

### DIFF
--- a/generators/server/templates/src/main/java/package/_Application.java
+++ b/generators/server/templates/src/main/java/package/_Application.java
@@ -40,6 +40,8 @@ import java.util.Collection;
 public class <%= mainClass %> {
 
     private static final Logger log = LoggerFactory.getLogger(<%= mainClass %>.class);
+    
+    private static final String MIS_CONFIGURED_APPLICATION = "You have misconfigured your application! It should not run with both the '{}' and '{}' profiles at the same time.";
 
     @Inject
     private Environment env;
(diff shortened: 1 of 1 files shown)


jdubois:
C'mon, this is a log message, so there is no risk with this text String. We have 1 more line of code for nothing, and it's harder to understand...

deepu105:
@sgrillon14 was there a warning from sonar on this?

sgrillon14:
I use Sonar on my JHipster application. I take advantage of the information sent by Sonar to cleaning the code generated by the generator. I have often met people who critique this generator by saying that the code generated is not clean. I think that the generated code is already well , I wanted remove errors found by Sonar these pernonnes not criticize the generator.

deepu105:
Ok, I guess since this doesnt cause any issue we could merge

cbornet:
LOL It's because there is a "+" between the log strings to make them multiline. Otherwise the log Strings would have been different...
On the other hand I must say I like to have a 0 sonar issue code.

jdubois:
Having a big laugh at people criticizing the "non clean code" of the generator for this specific use-case: if they can't read code, and can't think for themselves, they should better do another job!!!
I could just modify the String and Sonar wouldn't see anything anymore... That's so easy to trick Sonar, anybody should understand it's just a tool, and without a human with a brain to understand the code, its reports are useless.

deepu105:
@jdubois so what to do with this PR?

jdubois:
let's just change the String, so that everyone is happy in the end...

deepu105:
The log messages are already different, I dont know why this gives Sonar warning in the first place

deepu105:
@sgrillon14 Thanks and Sorry

deepu105:
I changed the string

PR #3415, opened 2016-04-16
hide id field while creating an entity

### LINKED ISSUE #3401: [FEATURE] Hide id text field entity creation dialog
Hey,

I was wondering if the id field in the entity creation dialog could be removed (but stay in the edit dialog). It does not add any meaningful informations as the actual database id has not been assigned yet anyway.

At the moment, I'm reading the JHipster Mini Book (http://www.jhipster-book.com) and as it happens  Matt Raible also removes the id field in his 21-point example. I'm not the only one that had this thought ;)

I use the latest JHipster version (3.0.0).

Thank you :)

Konstantin

konfri:
The id does not provide any meaningful information in the entity creation dialog as the actual database id has not been assigned yet anyway.

fix #3401.

### DIFF
--- a/generators/entity/templates/src/main/webapp/app/entities/_entity-management-dialog.html
+++ b/generators/entity/templates/src/main/webapp/app/entities/_entity-management-dialog.html
@@ -8,7 +8,7 @@
     </div>
     <div class="modal-body">
         <jhi-alert-error></jhi-alert-error>
-        <div class="form-group">
+        <div class="form-group" ng-show="vm.<%=entityInstance %>.id">
             <label for="id" translate="global.field.id">ID</label>
             <input type="text" class="form-control" id="id" name="id"
                     ng-model="vm.<%=entityInstance %>.id" readonly />


generators/entity/templates/src/main/webapp/app/entities/_entity-management-dialog.html
```diff
@@ -8,7 +8,7 @@ <h4 class="modal-title" id="my<%= entityClass %>Label" translate="<%= keyPrefix
     </div>
     <div class="modal-body">
         <jhi-alert-error></jhi-alert-error>
-        <div class="form-group">
+        <div class="form-group" ng-show="showIdValue">
```
deepu105: couldn't this be just a simple `ng-show="vm.<%=entityInstance %>.id"` so that ID will be shown if it exists else will be hidden.

konfri:
As a java backend/rcp developer i did not find the most straightforward way ;)

Thank you for the hint :)

PR #3418, opened 2016-04-17
Disable Karma JS minification to help debugging if it's run with --debug

adam-ho:
The current Karma config minifies JS and adds cover information. However, if it's run from the command line with the --debug option it would help if the developer if the JS files remain readable.

Sample command line start:
```
./node_modules/karma/bin/karma start src/test/javascript/karma.conf.js --debug
# Then navigate to http://localhost:9876/debug.html
```

It might be worthwhile to add this to this page too:
http://jhipster.github.io/running-tests/


deepu105:
can you do a PR to the documentation site as well https://github.com/jhipster/jhipster.github.io

PR #3422, opened 2016-04-19
Add an option to report Circuit Breakers metrics for monitoring

PierreBesson:
Related to this closed PR: https://github.com/jhipster/generator-jhipster/pull/3400

This will allow monitoring of circuit breakers with the JHipster Console.


generators/server/templates/src/main/java/package/config/_MetricsConfiguration.java
```diff
@@ -12,7 +18,11 @@
 import fr.ippon.spark.metrics.SparkReporter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.boot.actuate.autoconfigure.ExportMetricReader;
```
deepu105: these new imports should be only when `<%_ if (applicationType == 'microservice' || applicationType == 'gateway') { _%>`

deepu105: @PierreBesson please update this and merge

PierreBesson: You are right, thanks.

PR #3424, opened 2016-04-19
generate simpler build.gradle.

nobusugi246:


deepu105:
Sorry but this is not how we do our main gradle file and hence is not consistent, I dont see any value in this. cc @atomfrede

atomfrede:
Yes, its not the current case, but I also considered to switch to ``group:name:version``, but I'm not sure how versions are handled from ``gradle.properties`` in that case.

atomfrede:
I'll try to check if we can use the simplified version with dynamic version in the main generator on the weekend.

nobusugi246:
Thanks for this discussion. I think many gradle users feel comfortable this simplified version. In case of dynamic version, please use "(double quotations) instead of '(single quotations) like this,
<pre>
classpath "org.springframework.boot:spring-boot-gradle-plugin:${spring_boot_version}"
</pre>

atomfrede:
@nobusugi246 Thanks for the hint :)
@deepu105 I think I'm going to change the the gradle dependencies to the shorter format: ``compile 'com.fasterxml.jackson.core:jackson-core:2.7.3'`` When adding dependencie from maven central you can just copy and paste the dependency. Or do you have any objections?

deepu105:
@atomfrede I dont have any objections please go ahead

Issue #3338, opened 2016-04-05
REST API user management, regex on login too restrictive


flam182:
##### **Overview of the issue**

If you want to edit the **anonymousUser** (default user) you have this error : 
`No message available`

It's because the regex in the rest api is too restrictive on the login

`    @RequestMapping(value = "/users/{login:[_'.@a-z0-9-]+}",`

Angular can't get the user information

##### **JHipster version** 

Jhipster 3.0 version

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory** 

No entity

##### **Browsers and Operating System** 

Chrome

##### **Suggest a Fix** 
A correction would be in **UserRessource.java** file
`  @RequestMapping(value = "/users/{login:[_'.@a-zA-Z0-9-]+}",`

But in the front, they must have a similar regex control

Or another solution is to remove the regex on the login in the REST API

ruddell:
The regex for a User's login is also set in UserDTO

wmarques:
What's the point of doing a regex for this ? Shouldn't we only do a regex to validate the username in the registration ?

jdubois:
The regexp is because there is an issue with Spring MVC if you use a "." -> that's because if you call "user.json", Spring MVC by default will send you the user in JSON format. The issue is when your username is "julien.dubois"....
At the moment the regexp validates also that the username is in lowercase, but indeed this is isn't useful.
So this looks good to me, can you do a PR?

wmarques:
I did this [here](https://github.com/wmarques/generator-jhipster/commit/771c96fcb7237448bca189483a3f08cfaacf5b5b) but I've got a problem : when I try to login with a capslocked user, like USERTEST, it throws me a 401 error bad credentials but I'm sure that I'm using the right ones and i couldn't find where this come from...

wmarques:
@jhipster/developers if someone has an idea I can work on it :)

ruddell:
@wmarques I tested your branch with the development database set to mysql and it worked, but it doesn't work with h2disk.  I looked into it and it's because H2's select statements are case-sensitive and we get the user from the database by lowercase login instead of the original ([code link](https://github.com/jhipster/generator-jhipster/blob/6db557c0aab25d859dc63eae15db80d5b9783fb6/generators/server/templates/src/main/java/package/security/_UserDetailsService.java#L34)).  I removed the .toLowerCase() and it works.


As a side note, all of the register form translations will need to be updated.

[...]

PR #3425, opened 2016-04-19
Add anonymousUser as an exception un User's login regexp

wmarques:
Related to #3338 
We will be able to view, edit or delete the anonymousUser created by default.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/account/register/register.html
+++ b/generators/client/templates/src/main/webapp/app/account/register/register.html
@@ -33,7 +33,7 @@
                 <div class="form-group">
                     <label class="control-label" for="login" translate="global.form.username">Username</label>
                     <input type="text" class="form-control" id="login" name="login" placeholder="{{'global.form.username.placeholder' | translate}}"
-                           ng-model="vm.registerAccount.login" ng-minlength=1 ng-maxlength=50 ng-pattern="/^[a-z0-9]*$/" required>
+                           ng-model="vm.registerAccount.login" ng-minlength=1 ng-maxlength=50 ng-pattern="/^[_'.@A-Za-z0-9-]*$/" required>
                     <div ng-show="form.login.$dirty && form.login.$invalid">
                         <p class="help-block"
                                ng-show="form.login.$error.required" translate="register.messages.validate.login.required">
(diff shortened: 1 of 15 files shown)


jd-carroll:
Doesn't this change also prevent users from creating logins with `_'.-` ?  

Shouldn't the regex be: `^[_'.@a-z0-9-]*$|anonymousUser`

notflorian:
Isn't the anonymousUser mandatory for Spring Security? Is it a good idea to allow to delete it?

wmarques:
@jd-carroll when we try to register a user, it tells 

> Your username can only contain lower-case letters and digits.

So I changed the regex according to this

@notflorian Oh yes, maybe we shouldn't change the regex for the delete method

deepu105:
@wmarques I thinks its better we lowercase the login and save it and also lowercase both sides in query when matching. I would prefer lowercasing when saving as thats how normally its done everywhere. Also I prefer username to contain `_` and `.` as well so it should be `^[_'.@A-Za-z0-9-]*$` so that people can create user without worrying about case. Can you check if `anonymousUser` works if its lowercased as `anonymoususer`?

deepu105:
as @notflorian said we shouldnt let people delete the `anonymousUser` in the first place, so may be we shouldnt show it in the user listing at all, its just an internal user for spring security

wmarques:
@deepu105 seems to work with lowercased `anonymoususer`

wmarques:
Should work now

wmarques:
I removed the anonymous from the result received in user management controller, it's working

notflorian:
As anonymoususer is now fully in lowercase, do you still need to add it in the regex?

wmarques:
I think I can remove it yes, thanks.
So I should put Locale.ENGLISH ?

notflorian:
Yes Locale.ENGLISH should be good because thanks to the regex, the username cannot have special characters with accents.

deepu105:
@wmarques A test for social login is still failing. Also since we use the regexp in many places in java files, shouldn't we keep it in any of the constants file and use it as static import?

wmarques:
@deepu105 like a variable in the generator ?

deepu105:
Nope as Java constant String

Thanks & Regards,
Deepu

wmarques:
Seems that Travis is failing randomly :disappointed: The test passed [here](https://travis-ci.org/wmarques/generator-jhipster/builds/124691907)

deepu105:
@wmarques you can restart just the failed build alone within travis :) open the failed build and look for the restart icon on top right

wmarques:
@deepu105 Yes i know but I committed right after so let's wait for this one ! :)

PR #3429, opened 2016-04-20
Group KeyAndPasswordDTO.java generation with user-management since on…

sdoxsee:
Right now, KeyAndPasswordDTO.java is only used by AccountResource.java (which is only used in user-management). It would make sense to only generate KeyAndPasswordDTO.java only when user-management is selected.


jdubois:
Indeed! I don't why Travis failed but this should be merged.

PR #3432, opened 2016-04-20
Better microservice and gateway config for Heroku profile

jkutner:
For microservice applications:

* Set the `eureka.instance.hostname` and `non-secure-port` automatically
* Set `eureka.instance.prefer-ip-address` to false 

For gateway applications:

* Add `-Xmx256m` to the default `Procfile`

I've found that the gateway uses very little heap and a lot of off-heap memory (probably for buffers and streams). Setting max heap lower will allow more room for off-heap as to not exceed the quota on a free dyno (i.e. prevents [R14 errors](https://devcenter.heroku.com/articles/error-codes#r14-memory-quota-exceeded))


deepu105:
Thanks @jkutner

PR #3434, opened 2016-04-20
Allow --skip-user-management when using client and server subgenerators

sdoxsee:
Currently the --skip-user-management flag isn't available when generating client-only or server-only. Please review as this might not be the correct way to add this option.

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -92,8 +92,15 @@ module.exports = JhipsterClientGenerator.extend({
             defaults: 'jhi'
         });
 
+        // This adds support for a `--skip-user-management` flag
+        this.option('skip-user-management', {
+            desc: 'Skip the user management module during app generation',
+            type: Boolean,
+            defaults: false
+        });
+
         this.skipServer = configOptions.skipServer || this.config.get('skipServer');
-        this.skipUserManagement = configOptions.skipUserManagement || this.config.get('skipUserManagement');
+        this.skipUserManagement = configOptions.skipUserManagement = this.options['skip-user-management'] || this.config.get('skipUserManagement');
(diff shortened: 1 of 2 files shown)


generators/client/index.js
```diff
@@ -92,8 +92,15 @@ module.exports = JhipsterClientGenerator.extend({
...
+        // This adds support for a `--skip-user-management` flag
+        this.option('skip-user-management', {
+            desc: 'Skip the user management module during app generation',
+            type: Boolean,
+            defaults: false
+        });
+
         this.skipServer = configOptions.skipServer || this.config.get('skipServer');
-        this.skipUserManagement = configOptions.skipUserManagement || this.config.get('skipUserManagement');
+        this.skipUserManagement = configOptions.skipUserManagement = this.options['skip-user-management'] || this.config.get('skipUserManagement');
```
deepu105: Its should be `||` instead of second `=`

generators/server/index.js
```diff
@@ -66,8 +66,15 @@ module.exports = JhipsterServerGenerator.extend({
...
+        // This adds support for a `--skip-user-management` flag
+        this.option('skip-user-management', {
+            desc: 'Skip the user management module during app generation',
+            type: Boolean,
+            defaults: false
+        });
+
         this.skipClient = !this.options['client-hook'] || configOptions.skipClient || this.config.get('skipClient');
-        this.skipUserManagement = configOptions.skipUserManagement || this.config.get('skipUserManagement');
+        this.skipUserManagement = configOptions.skipUserManagement = this.options['skip-user-management'] || this.config.get('skipUserManagement');
```
deepu105: Same here

Issue #3431, opened 2016-04-20
[FEATURE] Maven multi module project issue with resolving path during startup


ykushch:
##### **Overview of the issue**

We have a large project with many module. When I've created module with Jhipster it is not running correctly due to inability to resolve path:

    File root;
    if (env.acceptsProfiles(Constants.SPRING_PROFILE_PRODUCTION)) {
        root = new File("target/www/"); // this is problem place
    } else {
        root = new File("src/main/webapp/"); // this is problem place
    }
    if (root.exists() && root.isDirectory()) {
        container.setDocumentRoot(root);
    }

##### **Motivation for or Use Case** 

It is not possible to run it as submodule of large number of projects. 

##### **JHipster Version(s)** 

Reproduces on all versions

##### **Browsers and Operating System** 

On every system it could be reproduced

##### **Reproduce the error** 

Create a maven structure that looks like this for example:

    MainModule
    |
    |__ SubModule1
    |   |__ JhipsterProject
    |
    |__ SubModule2

##### **Related issues** 

Found that you are not dealing with multi module projects (dividing core and frontend) but it is not the same. It is about including jhipster module as the separate one but in current structure.

##### **Suggest a Fix** 

I am working on fix for this issue. Could contribute as well.

jdubois:
I'm sorry but we don't generate multi module projects... You mean Maven multi-modules? Or with your IDE?

ykushch:
It seems that I did not describe the issue / feature nicely :) 
I have a project that has many maven-modules with different submodules already. And I am just adding Jhipster project to this hierarchy. And during startup the only issue that prevents it to be successful is inability to resolve the path like I've described above. 
I know that possible it is not the case but anyway - it seems like it is more the feature than bug.

ykushch:
It seems that I've managed to implement / fix this. The key thing here is that instead of looking into directly `JhipsterProject` (from the tree representation above) it looked into `MainModule` and cannot find `src/main/webapp/` path of course. Should test it now on Linux and Mac.

deepu105:
@ykushch if the fix is generic and applicable for default monolith project then plz do a PR

ykushch:
@deepu105 sure, I need to perform some additional tests. Tested it already on mac os - works fine with standard and non-standard layout too.

deepu105:
closed via #3447

PR #3447, opened 2016-04-22
Fix Maven multi module project issue with resolving path during startup #3431

ykushch:
[Feature] Maven multi module project issue with resolving path during startup #3431 

All is need in multi module project is to resolve intermediate path. Originally jhipster will look into core module only.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
+++ b/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
@@ -144,7 +144,7 @@ public class WebConfigurer implements ServletContextInitializer, EmbeddedServlet
         String rootPath = Paths.get(".").toUri().normalize().getPath();
         String extractedPath = fullExecutablePath.replace(rootPath, "");
 
-        int extractionEndIndex = extractedPath.indexOf("target/");
+        int extractionEndIndex = extractedPath.indexOf("<% if (buildTool == 'gradle') { %>build/<% } else { %>target/<% } %>");
         if(extractionEndIndex <= 0) {
             return "";
         }
--- a/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
+++ b/generators/server/templates/src/main/java/package/config/_WebConfigurer.java
@@ -22,7 +22,8 @@ import org.springframework.web.cors.CorsConfiguration;
 import org.springframework.web.cors.UrlBasedCorsConfigurationSource;
 import org.springframework.web.filter.CorsFilter;
 <% if (!skipClient) { %>
-import java.io.File;<% } %>
+import java.io.File;
+import java.nio.file.Paths;<% } %>
 import java.util.*;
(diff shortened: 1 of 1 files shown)


deepu105:
@erikkemperman since the original implementation from you are you ok with this change? looks ok to me

erikkemperman:
I think this might not work with Gradle, because of the literal `"target"`?

ykushch:
@erikkemperman you are right. `target` directory is purely maven solution. For gradle we should check instead for `build` directory instead. Thanks for pointing on this! Will do another PR.

erikkemperman:
You should maybe use the `<%= ... %>` pattern instead of a literal? The generator script will fill these in appropriately depending on the chosen build tool, so the template classes don't have to care.

Also I was wondering if the prefix should be the same for `dev` and `prod`, like they are with this patch? If not I would suggest making a helper function to return the `root` file instead of the prefix, taking account of the active profile?

ykushch:
@erikkemperman I've tried to run with `maven` `prod` profile on separate machine with `tomcat 8` and it worked fine with this. Made a new change taking into account `gradle`. 
As for variables - found that there are no exact these variables: `build/` and `target/`. They are appending as literal.

erikkemperman:
Ah yes, maybe I should have made a separate constant for ` build` or `target`.

Things might look all right even if they're not really, depending on how you run the app.

The setDocumentRoot thing was tricky to get right for all of the following scenarios:

- running the main class (eg in IDE) with `--spring.profiles.active=dev`
- `mvn -Pdev spring-boot:run`
- `mvn -Pdev package && java -jar target/foo.war --spring.profiles.active=dev`
- copying the war elsewhere and `java -jar` there.
- deploying the war to servlet containers
- deploying the war to cloud infrastructure
- all of the above with `prod` profile
- all of the above with gradle

ykushch:
@erikkemperman thanks for the complete list of scenarios! 
I've tried building it and then pick `war` file and manually deployed to tomcat instance on separate machine.

ykushch:
BTW if not to use this custom mechanism (documentRoot) it also worked from default folders that Spring Boot provides.

deepu105:
@ykushch is it ready, have you tested with the various combinations?

ykushch:
@deepu105 tested on mac, windows and remote tomcat instance (with linux on board). 
Want to perform some more tests.

deepu105:
@ykushch let us know once your testing is complete

ykushch:
@deepu105 I've finished with tests and they are working fine. Deployed to Pivotal too - works fine. 
But protractor tests started to fail after merge today. I've noticed that default behaviour was changed. When you are opening the application - user sees popup window with sign in at once.

pascalgrimaud:
@ykushch : it should be fixed on master. Can you rebase plz ?

ykushch:
@pascalgrimaud done that. Now all checks Are passing :) thanks.

PR #3451, opened 2016-04-23
Removing POM version overrides relative to spring-boot-dependencies.

adam-ho:
These are already defined here: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-dependencies/pom.xml
In some cases we were using the same version anyway, and in some cases +-1 minor version away.
Spring 1.3.3 defines these, let's use the same version to avoid conflict/confusion during runtime.


deepu105:
if the versions are same then fine. We need to check for gradle as well

deepu105:
cc @atomfrede

adam-ho:
The current setup is the example that it's not really fine - there is a reason Maven considers this to be a warning: whenever you upgrade the parent POM you would need to remember somehow to upgrade some of the dependency versions. However, looking at the POM it's really hard to tell which ones should be updated when the parent is updated.
In our case spring-social-security is already out of sync - to avoid these problems it is recommended to remove the overrides.

deepu105:
@adam-ho I meant if versions are the same i'm fine with removing. I think you misunderstood me. For some dependencies we override it for a reason and these particular dependencies were overwritten earlier for some reason. but with latest spring boot it might not be necessary.

atomfrede:
I've update some dependencies in my PR https://github.com/jhipster/generator-jhipster/pull/3452 but not yet removed some dependency overwrites

adam-ho:
Same:
spring-social-twitter
spring-social-facebook
httpclient
postgresql
jackson-datatype-hibernate4

Different:
spring-social-security (ours: 1.1.2.RELEASE, Spring Dependency: 1.1.4.RELEASE)

Looks like only 1 is different with no apparent reason.
Can we remove the version overrides and use the same / newer versions?

deepu105:
I think we can remove for these.

deepu105:
@adam-ho can you update Gradle as well please?

generators/server/templates/_build.gradle
```diff
@@ -134,7 +134,7 @@ dependencies {
     compile "com.fasterxml.jackson.datatype:jackson-datatype-json-org:${jackson_version}"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-hppc:${jackson_version}"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson_version}"
-    compile "com.fasterxml.jackson.datatype:jackson-datatype-hibernate4:${jackson_version}"
+    compile "com.fasterxml.jackson.datatype:jackson-datatype-hibernate4"
```
deepu105: @atomfrede is this ok to remove only for one dependency? would this cause any conflict later?

atomfrede: It is fine. Thats the way to go with spring boot:

> The spring-boot-gradle-plugin is also available and provides tasks to create executable jars and run projects from source. It also provides **dependency management that, among other capabilities, allows you to omit the version number for any dependencies that are managed by Spring Boot**

PR #3452, opened 2016-04-23
use simplified gradle dependency definition

atomfrede:
Instead of defining seperate ``group``, ``module`` and ``version`` for each dependency we can now use the simplified version ``"group:module:version"``. With that change it is easy to copy and paste dependency from maven central.
 
Also see https://github.com/jhipster/generator-jhipster/pull/3424

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -2,19 +2,19 @@ buildscript {
     repositories {
         mavenLocal()
         jcenter()
-        maven { url 'http://repo.spring.io/plugins-release' }
-        maven { url 'http://repo.spring.io/milestone' }
+        maven { url "http://repo.spring.io/plugins-release" }
+        maven { url "http://repo.spring.io/milestone" }
         maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
-        classpath 'org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:1.2'
-        classpath 'net.ltgt.gradle:gradle-apt-plugin:0.6'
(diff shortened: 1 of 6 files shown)


generators/server/templates/_build.gradle
```diff
@@ -113,138 +113,138 @@ repositories {
...
-    compile group: 'io.dropwizard.metrics', name: 'metrics-core'<% if (hibernateCache == 'ehcache') { %>
-    compile group: 'io.dropwizard.metrics', name: 'metrics-annotation', version: dropwizard_metrics_version
-    compile group: 'io.dropwizard.metrics', name: 'metrics-ehcache', version: dropwizard_metrics_version<% } %>
-    compile group: 'io.dropwizard.metrics', name: 'metrics-graphite', version: dropwizard_metrics_version
-    compile group: 'io.dropwizard.metrics', name: 'metrics-healthchecks', version: dropwizard_metrics_version
-    compile group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: dropwizard_metrics_version
-    compile group: 'io.dropwizard.metrics', name: 'metrics-servlet', version: dropwizard_metrics_version
-    compile group: 'io.dropwizard.metrics', name: 'metrics-json', version: dropwizard_metrics_version
-    compile (group: 'io.dropwizard.metrics', name: 'metrics-servlets', version: dropwizard_metrics_version) {
+    compile 'io.dropwizard.metrics:metrics-core'<% if (hibernateCache == 'ehcache') { %>
```
deepu105: You missed double quotes here

PR #3453, opened 2016-04-24
Added feature required to entity relationship

dornala:
Change Log:
-Added additional question's asking user for more specific details.
-Added "Required" when Entity is choosen with relationship Many to one.
-Add required in client field (To save an value of entity only when relationship is selected)


deepu105:
This doesnt have any commits? ?

dornala:
Ooooops.  Forgot to commit... ;)

Issue #3455, opened 2016-04-24
Invalid HTML is generated on details page - HR tag is not allowed in DL tags


adam-ho:
##### **Overview of the issue**

The generator creates the entity detail HTML with HR tags within DL tags. This is not allowed according to the HTML standard. To achieve the same visual a small CSS and generator template modification is required.

##### **Motivation for or Use Case** 

W3C validation fails.

##### **JHipster Version(s)** 

master

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

n/a

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory** 

n/a

##### **Browsers and Operating System** 

n/a

##### **Reproduce the error** 

Generate entity entity and hav a look at the *-detail.html page.`<hr>` tags are embedded in the `<dl>` tag.

##### **Related issues** 

n/a

##### **Suggest a Fix** 

```
Remove <hr> from _entity-management-detail.html on ln29 and ln73 
Add the following to the main.css before ln370:

.dl-horizontal>dd {
	margin-bottom: 15px;
}

@media screen and (min-width: 768px) {
	.dl-horizontal>dt {
		margin-bottom: 15px;
	}
	.dl-horizontal>dd {
		border-bottom: 1px solid #eee;
		padding-left: 180px;
		margin-left: 0;
	}
}
```

deepu105:
Can you do a PR please?

PR #3456, opened 2016-04-24
Fix 3455 Removing illegal HR tags from details view.

adam-ho:
fix https://github.com/jhipster/generator-jhipster/issues/3455

### DIFF
--- a/generators/client/templates/src/main/webapp/content/css/main.css
+++ b/generators/client/templates/src/main/webapp/content/css/main.css
@@ -291,6 +291,22 @@ Social Buttons
     cursor: pointer;
 }
 
+.dl-horizontal.jh-entity-details > dd {
+    margin-bottom: 15px;
+}
+
+@media screen and (min-width: 768px) {
+    .dl-horizontal.jh-entity-details > dt {
+        margin-bottom: 15px;
+    }
+
+    .dl-horizontal.jh-entity-details > dd {
+        border-bottom: 1px solid #eee;
+        padding-left: 180px;
+        margin-left: 0;
+    }
+}
+
 /* end of entity tables helpers */
 
 /* ui bootstrap tweaks */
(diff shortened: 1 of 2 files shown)


deepu105:
Please add the css in the sass file as well. Also yser management detail might have same issue.

Also the css tweak shouldn't be generuc, ot should be only for the dl used for entity detail. May be add a specific class so that all dl are not affected

adam-ho:
- User management doesn't use HR, so that's okay
- This style isn't generic, only applied to .dl-horizontal, so as a result user management detail view is not consistent with entities detail view. I think it's a good thing. However, if someone wants to use different styles for .dl-horizontal across the project, they can customise the CSS (which they do anyway).
- Which sass file do you refer to?

adam-ho:
@deepu105 Looks like there is some flake with the test : https://travis-ci.org/jhipster/generator-jhipster/jobs/125316929
`Timed out waiting for the WebDriver server at http://127.0.0.1:54793/hub`
Have you seen this before?

adam-ho:
Actually I thought about it and modifying the style isn't a good idea as we are modifying Bootstrap style for no good reason. The new style may or may not look better than the base style and people would need to figure out why and how we are different from Bootstrap. We are a generator framework so shouldn't be too opinionated about style.

So all in all, I would remove the HR because it generates invalid HTML but keep the original DL DT DD Bootstrap style in place and let the users customise it if they want to.

What do you think?

deepu105:
My point is I don't want you to override the bootstrap style but on the detail page I do want the bottom borders by default. So we should add a specific class the the parent dl `<dl class="dl-horizontal entity-detail">` and add custom style to that class. SO if someone doesn't want that they can just remove the class

adam-ho:
Yepp, I can live with that - added the jh-entity-details class.

deepu105:
you missed the sass file. I have update it anyways

Issue #3454, opened 2016-04-24
Login redirect is flakey (we are removing the previous state from sessionStorage before redirect)


adam-ho:
##### **Overview of the issue**

In Firefox and Safari (IE not tested) the redirect to page after login is flakey. 
It means that clicking a link when I'm currently not logged in will not always correctly redirect to the page but leave me on the ugly unauthorised screen.

##### **Motivation for or Use Case** 

It's a common use case to click a link then log in to see the content.

##### **JHipster Version(s)** 

master branch.

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

n/a

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory** 

n/a

##### **Browsers and Operating System** 

Mac OS El Capitan, Firefox latest, Safari latest. Chrome is NOT affected.

##### **Reproduce the error** 

Create any entity in the system, go to the details screen and note the URL (eg.: http://localhost:8080/#/candidate/1). Logout and paste in the URL which will pop up the login screen.
In the affected browsers after login the user is stuck on the non-authorised screen.

##### **Related issues** 

n/a

##### **Suggest a Fix** 

EDIT: the actual issue is that we call Auth.logout() on auth-expired.interceptor.js:25. This will destroy previous state data right away, however, this method is called after the login screen pops up.
The Auth.logout() should only execute anything if we aren't already logged out.

deepu105:
Could raise an issue on ngStorage plugin as well please?

adam-ho:
@deepu105  I've isolated the issue and I believe it's not ngStorage after all, but us calling logout(). I believe it might be a race condition and that's why it's working on Chrome and not on Firefox reliably.

deepu105:
May be we should do it in a promise

deepu105:
closed via #3459

PR #3459, opened 2016-04-25
Fix 3454 Clean up Auth JS services and remove previous state erase in Auth.logout()

adam-ho:
https://github.com/jhipster/generator-jhipster/issues/3454

Looks like the Auth.logout() is not actually required:
- If we are already logged out the previous state is the one we want to get back to - do not erase it.
- If we are just hitting logout button, the sessionStorage is already empty anyway, nothing to erase.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
+++ b/generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
@@ -51,9 +51,9 @@
                 // previousState was set in the authExpiredInterceptor before being redirected to login modal.
                 // since login is succesful, go to stored previousState and clear previousState
                 if (Auth.getPreviousState()) {
-                    var ps = Auth.getPreviousState();
+                    var previousState = Auth.getPreviousState();
                     Auth.resetPreviousState();
-                    $state.go(ps.previousStateName, ps.previousStateParams);
+                    $state.go(previousState.name, previousState.params);
                 }
             }).catch(function () {
                 vm.authenticationError = true;
--- a/generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
+++ b/generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
@@ -5,9 +5,9 @@
         .module('<%=angularAppName%>')
         .controller('LoginController', LoginController);
 
-    LoginController.$inject = ['$rootScope', '$state', '$sessionStorage', '$timeout', 'Auth', '$uibModalInstance'];
+    LoginController.$inject = ['$rootScope', '$state', '$timeout', 'Auth', '$uibModalInstance'];
 
-    function LoginController ($rootScope, $state, $sessionStorage, $timeout, Auth, $uibModalInstance) {
+    function LoginController ($rootScope, $state, $timeout, Auth, $uibModalInstance) {
         var vm = this;
 
         vm.authenticationError = false;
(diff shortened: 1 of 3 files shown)


deepu105:
Session storage may not be automatically resetting in oauth and jwt can you please check that as well

adam-ho:
I haven't changed the authenticationToken handling - I presume that is what you are after in JWT/Oauth case.

generators/client/templates/src/main/webapp/app/services/auth/_auth.service.js
```diff
@@ -182,5 +178,26 @@
...
+
+        function getPreviousState() {
+            var previousStateName = $sessionStorage.previousStateName;
+            var previousStateParams = $sessionStorage.previousStateParams;
+
+            if(!previousStateName) {
+                return null;
+            }
+
+            return { "previousStateName": previousStateName, "previousStateParams": previousStateParams };
```
deepu105: can we have `{ "name": previousStateName, "params": previousStateParams };`

generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
```diff
@@ -50,12 +50,10 @@
...
                 // previousState was set in the authExpiredInterceptor before being redirected to login modal.
                 // since login is succesful, go to stored previousState and clear previousState
-                if ($sessionStorage.previousStateName) {
-                    var previousStateName = $sessionStorage.previousStateName;
-                    var previousStateParams = $sessionStorage.previousStateParams;
-                    delete $sessionStorage.previousStateName;
-                    delete $sessionStorage.previousStateParams;
-                    $state.go(previousStateName, previousStateParams);
+                if (Auth.getPreviousState()) {
+                    var ps = Auth.getPreviousState();
```
deepu105: can we have `var previousState = Auth.getPreviousState();`

generators/client/templates/src/main/webapp/app/components/login/_login.controller.js
```diff
@@ -50,12 +50,10 @@
...
+                    Auth.resetPreviousState();
+                    $state.go(ps.previousStateName, ps.previousStateParams);
```
deepu105: ` $state.go(previousState.name, previousState.params);`

generators/client/templates/src/main/webapp/app/services/auth/_auth.service.js
```diff
@@ -182,5 +178,26 @@
...
+        }
+
+        function resetPreviousState() {
+            delete $sessionStorage.previousStateName;
+            delete $sessionStorage.previousStateParams;
+        }
+
+        function storePreviousState(previousStateName, previousStateParams) {
+            $sessionStorage.previousStateName = previousStateName;
```
deepu105: save its as object `$sessionStorage.previousState = { "name": previousStateName, "params": previousStateParams };`

generators/client/templates/src/main/webapp/app/services/auth/_auth.service.js
```diff
@@ -145,10 +145,6 @@
         function logout () {
             AuthServerProvider.logout();
             Principal.authenticate(null);
-
-            // Reset state memory
-            delete $sessionStorage.previousStateName;
```
deepu105: I was asking about this, would this affect OAuth or JWT?

adam-ho: Not sure how would this be different with OAuth or JWT? It's just an after login state (url), which is stored in the session storage (as opposed to local storage) and wiped right after actually logging in. 
For instance auth.session.service logout() method only wipes the authenticationToken but doesn't deal with previousStates.

deepu105:
Protractor tests are failing for Gradle. I have restarted it again. Please check locally if it fails again.

adam-ho:
Test completed, looks like it's flakey :(

deepu105:
LGTM, @jhipster/developers does any one have any concerns with this? if not we can merge this

jdubois:
Is it tested with all 4 authent mechanisms? Otherwise OK for me.

pascalgrimaud:
Protractor is not activated for all auth mechanisms, maybe it should be

deepu105:
I wont have enough time to test all so any help is welcome

deepu105:
@adam-ho can you test with Oauth and JWT as well please?

adam-ho:
I'm not too familiar with either - however, if we don't trust the unit tests for it I'd recommend adding more. 
Either way, the redirect bug does exist in the generator at the moment.

jdubois:
Sorry I closed this by error!!!!!!

jdubois:
OK I'm merging this and testing afterwards, my bad :-(

PR #3463, opened 2016-04-25
missing dependency with gradle and cassandra

raphaelbrugier:
On the master, a dependency is missing when generating the app for gradle and cassandra

```
Could not resolve all dependencies for configuration ':testCompileClasspath'.
> Could not find org.cassandraunit'cassandra-unit-spring:2.1.3.1:.
  Required by:
      com.mycompany.myapp:mono-gradle-cass:0.0.1-SNAPSHOT
```

It's a typo in the group name declaration.

The first commit add a travis test case.
I'll add a second commit with the fix after the CI shows the failure.

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -235,7 +235,7 @@ dependencies {
     compile "org.springframework.social:spring-social-twitter:${spring_social_twitter_version}"<% } %>
     testCompile "com.jayway.awaitility:awaitility:${awaility_version}"
     testCompile "com.jayway.jsonpath:json-path"<% if (databaseType == 'cassandra') { %>
-    testCompile("org.cassandraunit'cassandra-unit-spring:${cassandra_unit_spring_version}") {
+    testCompile("org.cassandraunit:cassandra-unit-spring:${cassandra_unit_spring_version}") {
         exclude(module: 'org.slf4j')
     }<% } %><% if (testFrameworks.indexOf('cucumber') != -1) { %>
     testCompile "info.cukes:cucumber-junit:${cucumber_version}"


raphaelbrugier:
Failed for the wrong reason, I didn't realized gradle was not automatically picked over maven in travis scripts

raphaelbrugier:
For some reason, the build is not passing with gradle.
@pascalgrimaud Any idea if I forgot something when configuring the conditions to use gradle instead of maven, or some other configuration needed for gradle samples?

atomfrede:
Looks all fine to me. Do you have the same compile error when trying to generate the sample on your local machine?

raphaelbrugier:
Yes, I do have the same error, after adding the same entity.
It's not related to my changes, it is failing for an other reason.
I'm still investigating it, but it looks generating entities with gradle/cassandra with LocalDate field doesn't work anymore.

raphaelbrugier:
Ok I found, it was related to my changes.
I had a typo in a conditional test, that led to copy too much entities when testing :(

pascalgrimaud:
Fixing the typo is ok

Although I love Cassandra, I'm a bit reluctant to add a travis build for cassandra + gradle. It's because Cassandra is the less used database. So I prefer to keep these new builds to another options which are more used

raphaelbrugier:
Ah, so do you want to remove it from the samples/covered cases?

pascalgrimaud:
For this project, yes please. If I remember well, few people use cassandra, that's the main reason.

But I have in mind another project to test more combinaisons so we ll be able to cover gradle+cassandra

raphaelbrugier:
Ok, so I'll remove the commit from the branch and force push.
I'm interested with what you've in mind!
Because Cassandra option is not well covered and I break a thing with my work on the docker-compose, see my other PR.

atomfrede:
Thanks @raphaelbrugier

Issue #3368, opened 2016-04-08
[FEATURE] Replace static dev ribbon image with a dynamic CSS based non-prod ribbon


ddewaele:
Context : 

- Currently jHipster has a static development ribbon image that is displayed when initially running with the dev profile
- When building with the -Pprod profile, at build time Gulp ensures that the development ribbon is removed.
- (doubt) I think that at that point the ```app.constants.js``` is modified and can never be generated to get the development ribbon back (you need the -Pdev profile to change it back, but -Pdev will not run gulp).
- The docs always talk about 2 profiles, but in some environments where there are multiple non-prod environments, having jHipster (optionally) show a banner to clearly visualise what non-prod environment we're on (and if we have the correct spring profile on a particular environment) is can be useful.

Proposal of this feature : 

- Remove the static development ribbon image with a [CSS based ribbon](https://github.com/simonwhitaker/github-fork-ribbon-css) so that is more dynamic.
- When having multiple non-prod environments, (dev, tst, uat, ....), we could show appropriate banner to indicate with what spring profile the app has been started.  (can be optional)
- For the prod profile we can remove the banner

Thinking about implementing this for our projects, if you feel this could be useful for jHipster I'd be happy to create a PR

gmarziou:
Good idea.

Another benefit of using CSS is that we could translate it in all languages we support.

Could you clarify how you would switch between profiles? I imagine doing it server-side maybe using Thymeleaf for injecting profile as a template variable.

ddewaele:
Possibly, need to look at the specifics. It's only on the index page and that's only loaded once normally ...  will draft a proposal here this weekend. But need some r&r now :)

deepu105:
I also did this in one of my project, I had a simple rest controller which
would return active profiles and this endpoint was not protected so that it
is available all the time, and had a directive and service which would use
this to set the ribbon text. But i didn't consider it a very clean solution
and that's why didn't add it to the generator itself.

gmarziou:
Yes this may be the ricky parts and also what to display when there are multiple profiles active at same time.

Anyway, even with only the CSS paft, we would benefit from i18n.

ddewaele:
Been playing around with this a bit with one of my projects.

**jHipster ribbon config option (application.yml)**

Instructs jHipster to show the ribbon for the provided profiles :

```
ribbon:
   displayOnActiveProfiles: dev
```

If users have additional envs all they need to do is this : 

```
ribbon:
[...]

PR #3474, opened 2016-04-26
CSS based dev ribbon based on spring profiles

ddewaele:
Fix for #3368 

Please review. Will add unit tests later today.

1 doubt : 

- ```debugInfoEnabled``` : configured in the angular app config. profile service is not available there. Can we move this somewhere else ?


generators/client/templates/src/main/webapp/app/blocks/config/_compile.config.js
```diff
@@ -5,13 +5,15 @@
...
+    compileServiceConfig.$inject = ['$compileProvider'];
 
-    function compileServiceConfig($compileProvider, ENV) {
+    function compileServiceConfig($compileProvider) {
         // disable debug data on prod profile to improve performance
-        if(ENV === 'prod'){
-            $compileProvider.debugInfoEnabled(false);
-        }
+        
+        // TODO: Howto fetch the dynamic profile in this config.... can this be done elsewhere ?
```
deepu105: You can Inject the `ActiveProfiles` service here and check from it, what is the issue in that?

generators/client/templates/src/main/webapp/app/services/profiles/_active-profiles.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+    angular
+        .module('<%=angularAppName%>')
+        .service('ActiveProfiles', ActiveProfiles);
+
+    ActiveProfiles.$inject = ['$q','$resource'];
+
+    function ActiveProfiles ($q, $resource) {
+
+        var _activeProfiles,_ribbonEnv,_inProduction,_response;
```
deepu105: formatting is wrong, space between variables and indentation

generators/client/templates/src/main/webapp/app/services/profiles/_active-profiles.service.js
```diff
@@ -0,0 +1,69 @@
...
+        function inProduction() {
+        	return _inProduction;
+        }
+        
+        function ribbonEnv() {
+        	return _ribbonEnv;
+        }
+        
+        function activeProfiles() {
+        	return _activeProfiles;
```
deepu105: check if `_activeProfiles` exists, else do a fetch and return promise

generators/client/templates/src/main/webapp/app/services/profiles/_active-profiles.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+    function ActiveProfiles ($q, $resource) {
+
+        var _activeProfiles,_ribbonEnv,_inProduction,_response;
+
+	    var service = {
+	    	activeProfiles: activeProfiles,	    		
+	    	ribbonEnv: ribbonEnv,
+	        inProduction: inProduction,
+	        fetch:fetch
```
deepu105: do not expose fetch, always use `activeProfiles` so that user doesnt have to know if its fetched from server or from service cache

generators/client/templates/src/main/webapp/app/layouts/navbar/_navbar.controller.js
```diff
@@ -5,14 +5,19 @@
...
 
-    function NavbarController ($state, Auth, Principal, ENV, LoginService) {
+    function NavbarController ($state, Auth, Principal, ActiveProfiles, LoginService) {
         var vm = this;
 
         vm.isNavbarCollapsed = true;
         vm.isAuthenticated = Principal.isAuthenticated;
-        vm.inProduction = ENV === 'prod';
+        
+        ActiveProfiles.fetch().then(function(response) {
```
deepu105: use `activeProfiles` call

generators/client/templates/src/main/webapp/app/services/profiles/_page-ribbon.directive.js
```diff
@@ -0,0 +1,34 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('<%=angularAppName%>')
+        .directive('pageRibbon', pageRibbon);
+
+    function pageRibbon(ActiveProfiles,$rootScope<% if (enableTranslation) { %>, $translate<% } %>) {
+        var directive = {
+    		replace: true,
```
deepu105: formating

generators/client/templates/src/main/webapp/app/services/profiles/_page-ribbon.directive.js
```diff
@@ -0,0 +1,34 @@
...
+            restrict: 'AE',
+            template: '<div class="ribbon hidden"><a href="#">{{ribbonTitle}}</a></div>',
+            link: linkFunc
+        };
+
+        return directive;
+
+        function linkFunc(scope, element, attrs) {
+            ActiveProfiles.fetch().then(function(response) {
```
deepu105: use `activeProfiles`

generators/client/templates/src/main/webapp/app/blocks/config/_compile.config.js
deepu105: why do you say its not available here?

deepu105: worst cae you can use `var Principal = $injector.get('Principal');` like approach as well, but services should be available in configuration blocks

ddewaele: Tried the ```injector.get``` but didn't work. Don't think you're supposed to inject services during app config. [Angular docs](https://docs.angularjs.org/guide/module) say : 

_Configuration blocks - get executed during the provider registrations and configuration phase. Only providers and constants can be injected into configuration blocks. This is to prevent accidental instantiation of services before they have been fully configured._

deepu105: You are right. Thats a bummer we need to find a way to work around this :(

deepu105: can you try to do it in a `app/blocks/handlers/* .handler` and call it in the run block? I'm not sure if thats a good way to do it :(

generators/client/templates/src/main/webapp/app/services/profiles/_active-profiles.service.js
```diff
@@ -0,0 +1,57 @@
...
+        .factory('ActiveProfiles', ActiveProfiles);
+
+    ActiveProfiles.$inject = ['$q', '$resource'];
+
+    function ActiveProfiles($q, $resource) {
+
+        var _activeProfiles;
+
+        var service = {
+            activeProfiles : activeProfiles
```
deepu105: should we call it `getActiveProfiles` ? for aesthetics also we should call the service `ProfileService` rather than `ActiveProfiles`

generators/client/templates/src/main/webapp/app/blocks/config/_compile.config.js
ddewaele: What  about in the run block ? (or is that too late ?)

Run Blocks

_Run blocks are the closest thing in Angular to the main method. A run block is the code which needs to run to kickstart the application. It is executed after all of the service have been configured and the injector has been created. Run blocks typically contain code which is hard to unit-test, and for this reason should be declared in isolated modules, so that they can be ignored in the unit-tests._

generators/client/templates/src/main/webapp/app/services/profiles/_active-profiles.service.js
ddewaele: ok for ProfileService. 

Instead of exposing activeProfiles, should it return 1 response object containing

- activeProfiles
- ribbonEnv (will rename this to ribbonProfile I think)
- inProduction

or do you want individual methods for that on the ProfileService ?

deepu105: I think can be seperate methods

generators/client/templates/src/main/webapp/app/blocks/config/_compile.config.js
ddewaele: @deepu105  : The compileProvider can only be accessed in config block.

You can pull the compileProvider out of the config block and use it in the run method.

```
(function() {
    'use strict';

    var provider = {};

    angular
        .module('ribbon8App', [
            'ngStorage', 
            'ngResource',
            'ngCookies',
            'ngAria',
            'ngCacheBuster',
            'ngFileUpload',
            'ui.bootstrap',
            'ui.bootstrap.datetimepicker',
            'ui.router',
            'infinite-scroll',
            // jhipster-needle-angularjs-add-module JHipster will add new module here
            'angular-loading-bar'
        ])
        .config(config)
        .run(run);

    config.$inject = ['$compileProvider'];

    function config($compileProvider) {
        angular.copy($compileProvider, provider);
    }

    run.$inject = ['stateHandler','ActiveProfiles'];

    function run(stateHandler,ActiveProfiles) {
        stateHandler.initialize();
        
        ActiveProfiles.activeProfiles().then(function(response) {
            provider.debugInfoEnabled(!response.inProduction);
        });        
    }
})();
```

deepu105: Thats a hacky solution and will attract angular nazi :( @mraible any idea
on how to do this?
On 28 Apr 2016 00:46, "Davy De Waele" <notifications@github.com> wrote:

In
generators/client/templates/src/main/webapp/app/blocks/config/_compile.config.js
<https://github.com/jhipster/generator-jhipster/pull/3474#discussion_r61293501>
:

>          // disable debug data on prod profile to improve performance
> -        if(ENV === 'prod'){
> -            $compileProvider.debugInfoEnabled(false);
> -        }
> +
> +        // TODO: Howto fetch the dynamic profile in this config.... can this be done elsewhere ?

@deepu105 <https://github.com/deepu105> : The compileProvider can only be
accessed in config block.

You can pull the compileProvider out of the config block and use it in the
run method.

(function() {
    'use strict';

    var provider = {};

    angular
        .module('ribbon8App', [
            'ngStorage',
            'ngResource',
            'ngCookies',
            'ngAria',
            'ngCacheBuster',
            'ngFileUpload',
            'ui.bootstrap',
            'ui.bootstrap.datetimepicker',
            'ui.router',
            'infinite-scroll',
            // jhipster-needle-angularjs-add-module JHipster will add
new module here
            'angular-loading-bar'
        ])
        .config(config)
        .run(run);

    config.$inject = ['$compileProvider'];

    function config($compileProvider) {
        angular.copy($compileProvider, provider);
    }

    run.$inject = ['stateHandler','ActiveProfiles'];

    function run(stateHandler,ActiveProfiles) {
        stateHandler.initialize();

        ActiveProfiles.activeProfiles().then(function(response) {
            provider.debugInfoEnabled(!response.inProduction);
        });
    }
})();

—

Reply to this email directly or view it on GitHub
<https://github.com/jhipster/generator-jhipster/pull/3474/files/bfb301cbc2770722e89419177c2e373c21d1c1ab#r61293501>

deepu105: @ddewaele take a look at this thread http://stackoverflow.com/questions/15937267/inject-service-in-app-config

ddewaele: @deepu105 :  I did look at that, but I assumed providers allowed you to "configure" services at config time, before they become actual services that you inject.

In this case, we would need to do a service invocation (to get the profile info) in the config. I felt it would "hack" the provider mechanism to actually invoke the service itself. Also, I don't think you can even access $resource, $q outside of the get function of the provider. And in this proposal, we would need it outside the get function.

An alternative could be to just set the debugInfo flag statically during build time (using gulp like the ENV was done). 

I still think there's a big different between the deugInfo flag (more static in nature) and the actual environment (more dynamic).

deepu105: Ya in that case lets use ENV constant for that and if we were able to
update the ENV constant with actual profiles once the ProfileService is
invoked it would have been even better but I guess since they are constants
the value cannot be updated.

One issue currently is that if you build using prod profile the ENV will be
updated with `prod` and unless you run `gulp ngconstant:dev` manullay it
willnot be reset to `dev` when running dev profile.

Thanks & Regards,
Deepu

ddewaele: Exactly, that was one of the reasons why I wanted to remove that ENV constant. It's only updated when building with the -Pprod, and there's no obvious way to reset it in a CI scenario.

I would even rename that constant to clearly indicate that it will only be used for the debugInfo.

I don't like the idea of hardcoding an ENV into the artifact. We should be able to do everything with spring profiles.

ddewaele:
Made some changes... 

- passing DEBUG_INFO_ENABLED fromt app.constants (set their by gulp.)
- Renaming to ActiveProfiles to ProfileService (exposes profileInfo() that returns 1 response object containing properties like activeProfiles / inProduction / ribbonEnv. It makes the service easier to read and easier to cache. )
- eslint passed
- travis checks passed

deepu105:
I made some changes to this while merging

Issue #733, opened 2014-11-04
Mongeez requires admin privileges


jdubois:
Mongeez (used to populate MongoDB) does not work on cloud providers like https://mongolab.com/ as it requires admin privileges.
See https://github.com/secondmarket/mongeez/issues/27

We might consider another option soon, as this is a blocker issue with CloudFoundry + Mongodb, which would be a great combo!

d0x:
Do you have a plan to move to another solution?

jdubois:
Yes I want to use another solution. I just haven't worked on it yet -> if you are interested you can take the lead on this!

franciscocpg:
@jdubois
I think the solution here is not to use another solution, but to patch Mongeez because the problem here is just the following  code at [MongeezDao.java](https://github.com/secondmarket/mongeez/blob/master/src/main/java/org/mongeez/dao/MongeezDao.java):
```java
public void runScript(String code) {
        db.eval(code);
}
```
Besides the fact that eval requires admin privileges, it was deprecated in version 3.0 and future releases will remove it according to this thread:
https://jira.mongodb.org/browse/SERVER-17453
I see two options at this moment:
1. Instead of writing javascript codes in mongeez changesets xml file, should write java code and execute it using [bsh](http://www.beanshell.org/).
2. Parse javascript code and create a wrapper for java using mongodb java driver.
For example:
The code
```javascript
db.test.insert({
   "name" : "MongoDB",
   "type" : "database",
   "count" : 1,
   "info" : {
               x : 203,
               y : 102
             }
})
```
should execute
```java
DBCollection testCollection = db.getCollection("test");
BasicDBObject doc = new BasicDBObject("name", "MongoDB")
        .append("type", "database")
        .append("count", 1)
        .append("info", new BasicDBObject("x", 203).append("y", 102));
testCollection.insert(doc);
```
For this option, maybe some project like [jongo](http://jongo.org/) could help.

rreitmann:
Is there any workaround at the moment? I also want to deploy to cloudfoundry (Pivotal Web Services) using mongodb...

jdubois:
The only workaround is to update the database manually.

rreitmann:
:-) This is exactly what I am doing currently.

rreitmann:
I guess I will have a look at this https://github.com/goodeggs/grunt-mongo-migrations and manage the mongo migrations with mongoose...

jdubois:
[...]

PR #3479, opened 2016-04-27
Migrate Mongeez to Mongobee

albertmatyi:
# Why?

Mongeez uses features of Mongodb (`eval`) that don't work in some environments. Mostly cloud environments where this feature is disabled.

* [SO issue 1] (http://stackoverflow.com/questions/31014405/deploying-a-jhipster-mongodb-application-to-heroku)
* [SO issue 2] (http://stackoverflow.com/questions/28506034/jhipster-mongodb-connection-authorization/36887953#36887953)
* [jhipster issue #733](https://github.com/jhipster/generator-jhipster/issues/733)

# Therefore

This PR is a working copy of the current Jhipster generator that uses [Mongobee](https://github.com/mongobee/mongobee/wiki/How-to-use-mongobee) instead of Mongeez to implement database migrations logic.

# Moreover

This approach provides a programmatic toolset for setting up / migrating your database, which might come more handy in certain situations. In this scenario I prefer Java-based config to xml-based one. (personal opinion, but fits with my spring config approach :smile: )

# Guidline

In this case I am taking the approach of putting all migrations  in the `app.package.name.config.dbmigrations` package with the name `{SomeName}Migration.java`.

Note: We might as well stick to a naming convention of  `{SomeName}Migration{NNNN}.java` so the files themselves could be visibly indexed (easier lookup after migration number)

# Further ideas

1. Make migration scripts environment dependent 

# Edits

1. Based on @pascalgrimaud 's comment, this should be done in a module in order not to affect the functionality and stability of the core generator

I am open to suggestions for improvements.

Cheers,
Matyas


pascalgrimaud:
It should be discussed first
About the project, only 34 stars, some issues opened since 1 year, not very active

But what I see from your commits, if you think it can be usefull, maybe do it as a module ?

albertmatyi:
Hey @pascalgrimaud ,

Thanks for the reply. I understand your points, 
And I haven't analyzed in depth:

* the 19 open issues on Mongeez since 2011 
* the 6 issues on Mongobee since 2014


I am no expert on either of the tools. I cannot vouch for their stability. 
I can only say up until now Mongobee satisfied my needs and I've noticed other people are strugling with similar problems.

Your module suggestion is right, thanks, but I don't have now the resources to learn how to create one, so from my POV I'll leave this changeset here. And anyone wishing might create a module out of it.

Or if someone takes the time to thoroughly compare mongeez with mongobee to see which one fits the generic usecase better than we might continue the discussion of updating / merging the PR.

Have a nice day :)
Matyas

jdubois:
We've been stuck for so long with Mongeez, I welcome any migration away from it!!!!!!!!

Indeed, I am worried that the project isn't very people, but neither is Mongeez. In fact, they look both unmaintained and unpopular to me, so at least if Mongobee is solving our issue, I'm all for merging this commit.

-> do you confirm this works with Cloud Providers, like Heroku and CloudFoundry? Did you test it?

If yes, I'd like to merge this in the master branch.

albertmatyi:
@jdubois I've tested it on **heroku & mlab** and it is **working without any issues**.

deepu105:
LGTM

pascalgrimaud:
LGTM, but 1 comment:
- mongeez uses "mongeez" document
- mongobee uses "dbchangelog" document
That means people who update project will be impacted by this. Anyway, it's very minor impact

jdubois:
Let's do it then.
This is an important change, but I've been looking for a solution with Mongeez for far too long.

albertmatyi:
You guys are awesome. Thank you :)

Cheers,
m

PierreBesson:
Thanks for contributing !

jdubois:
Thanks! There still needs some work, for example to make it work with Heroku (ping @jkutner - I'm currently testing it)

jdubois:
OK I can't test it as I need to enter my credit card to have a MongoLab instance... I'm commiting my code now, it's probably working (or not far from working) and I will link my commit to this ticket.
@albertmatyi @jkutner will you be able to test it?
In the meantime I will leave MongoDB on Heroku as unsupported.

jkutner:
@albertmatyi did you test with a [Heroku MongoDB add-on](https://elements.heroku.com/addons/mongolab), or an instance you ran yourself?

@jdubois the Heroku generator does not currently detect for mongo use. Maybe that's the problem you've seen? It should detect for mongo, and configure the free add-on just like Postgres or MySQL. I will open that in a separate PR.

jkutner:
@jdubois testing it now

jdubois:
@jkutner yes have a look at my commit https://github.com/jhipster/generator-jhipster/commit/92ed130151c7ec4b53c22423b3408be27ee510c6
It's just that I'm afraid to add my credit card: as you granted me free access to "private spaces", and out there the bill is huge :-)

albertmatyi:
@jkutner As mentioned in a previous comment I have tested the setup on **heroku** & **mlab**. The default deployment is up and running @ http://shaerecore.herokuapp.com/ (free instance)

@PierreBesson Yes I agree. On the other hand there was already such an issue, which I also referenced in the PR ( https://github.com/jhipster/generator-jhipster/issues/733 )

jkutner:
@albertmatyi thanks! can you share your YML config for mongobee? I'm getting a connection timeout with just the [config in the generator](https://github.com/jhipster/generator-jhipster/blob/master/generators/heroku/templates/_application-heroku.yml#L29-L31).

albertmatyi:
I did not use the heroku generator, I have a jenkins instance that:

1. makes a production build, using the production profile, with the correct Mongo URL
1. Deploys the jar file using the [heroku jar deployer](https://devcenter.heroku.com/articles/deploying-executable-jar-files) along with a `Procfile` (heroku start script) with the following contents:

    web: java -jar build/libs/*.war --spring.profiles.active=prod --server.port=$PORT --jhipster.metrics.jmx.enabled=false --spring.datasource.maximumPoolSize=5


    

And I deploy

albertmatyi:
@jkutner I just realized I have something extra in the `yml` file and that is: 

    mongodb:
            uri: mongodb://{THE_COPY_PASTE_URL_FROM_HEROKU_SETTINGS}?authMode=scram-sha1
            database: {DB_NAME}

Note the `?authMode=scram-sha1` - without it I could not connect to the Mlab mongodb instance

jkutner:
@albertmatyi ah great. thank you for sharing this. I will update the heroku generator to do the same.

albertmatyi:
@jkutner To support my comment: http://docs.mlab.com/db-version-requirements/ - connections to new mlab instances have to be authenticated via scram-sha1 method

jkutner:
I'm having trouble with the Heroku generator because Mongobee incorrectly requires both the Mongo URI and database name. The latter is already included in the URI, but there is not a good way (that I know of) to parse the db name out of the URI in the YML config. So I get the error:

```
java.lang.IllegalArgumentException: Database name must not be empty
```

I know this can be done in XML config with `URI.getPath()` like this:

```xml
<bean class="java.net.URI" id="dbUrl">
    <constructor-arg value="#{systemEnvironment['DATABASE_URL']}"/>
</bean>

<bean id="dataSource" class="org.apache.commons.dbcp.BasicDataSource">
    <property name="databaseName" value="#{@dbUrl.getPath()}"/>
   ...
</bean>
```

Is there a way to do something similar in YML?

jdubois:
Yes you can also put it in the YAML, next to the URI, see https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/_application-prod.yml#L101 for example.
Then I hope you have the database name available as an env variable!

jkutner:
@jdubois the db name is not available as an env var. And it can't be hard-coded because it could be changed by the platform. I may have to back-out to Java code for this.

jdubois:
@jkutner bad news :-(

jkutner:
I'm sure I can fix this, but might take some time. We can leave Heroku/Mongo as unsupported for now.

Issue #3485, opened 2016-04-28
Update Node.js (and npm) version of generated .travis.yml file


rodrigomaio:
`.travis.yml` file is currently being generated with Node.js version 0.12, which is outdated and does not build correctly. Current LTS version is 4.4.3

deepu105:
No it is with latest LTS which branch are you looking at?

rodrigomaio:
The file on the project itself is updated, but the generated file from the template is not.

atomfrede:
As the PR is merged, I'll close this issue. Thanks @rodrigomaio

PR #3486, opened 2016-04-28
Update Node.js version on generated .travis.yml

rodrigomaio:
#3485 
.travis.yml file is currently being generated with Node.js version
0.12, which is outdated and does not build correctly. Current LTS
version is 4.4.3


deepu105:
Oh ok the template. Yes this should be updated

deepu105:
Will merge once tests pass

PR #3494, opened 2016-04-29
gulp build issue fixed

simon-kunz:
Fix for a small issue with the gulp build


deepu105:
Its actually valid in es6, so you will have issue only when you use node version less than 4.x. ANyway ill merge this

PR #3495, opened 2016-04-29
Fix some uaa bugs

sdoxsee:
When using uaa authentication:
* MicroserviceSecurityConfiguration needs to be generated for gateways (otherwise requests to "health", "configprops", etc. are unauthorized. The gateway is also a resource server (like microservice applications) in this case.
* Add token for swagger requests (authorization 'uaa' was left out)
* Call uaa server for audits (since audits are on the uaa server, requests must go to /uaa/api/audits instead of api/audits)


deepu105:
never mind the travis failure, its not related to this PR its some random travis glitch with the gradle protractor tests

sdoxsee:
Phew. Couldn't think of what I did to make gradle fail :P Thanks

PR #3501, opened 2016-04-30
Fix api profile info

pascalgrimaud:
it should fix protractor tests on master


deepu105:
Thank you Pascal, my bad sorry

PR #3502, opened 2016-04-30
Travis: remove protractor tests for gradle

pascalgrimaud:


deepu105:
Couldn't we increase the waiting time?

pascalgrimaud:
I will try something else.
Let's close this

PR #3503, opened 2016-05-01
Added "Required" option for relationships.

dornala:
**Added required option for relationships:**

- Adds option to relationships.
- Adds validation to client relationship field.


dornala:
@deepu105 can you please see log and say why the test failed. I can't figure out. Thank you.

pascalgrimaud:
@dornala: try to rebase with master. Travis failed a lot with gradle, it should be now fixed on master

generators/entity/index.js
```diff
@@ -273,7 +273,14 @@ module.exports = EntityGenerator.extend({
         if (this.relationships.length > 0) {
             this.log(chalk.white('Relationships'));
             this.relationships.forEach(function (relationship) {
-                this.log(chalk.red(relationship.relationshipName) + ' ' + chalk.white('(' + _.upperFirst(relationship.otherEntityName) + ')') + ' ' + chalk.cyan(relationship.relationshipType));
+                var validationDetails = '';
+                var relationshipValidate = relationship.relationshipValidateRules && relationship.relationshipValidateRules.length >= 1;
+                if (relationshipValidate === true) {
```
deepu105: can you simplify this? I guess i gave this comment in your old PR as well. can you please update same PR in future instead of opening new ones as its really inconvenient to review.

```
if (relationship.relationshipValidateRules && relationship.relationshipValidateRules.indexOf('required') !== -1)
```
only this condition would be sufficient to replace three lines of your code

generators/entity/index.js
```diff
@@ -1566,6 +1598,14 @@ module.exports = EntityGenerator.extend({
                     this.fieldsContainManyToOne = true;
                 }
 
+                if (_.isArray(relationship.relationshipValidateRules) && relationship.relationshipValidateRules.length >= 1) {
```
deepu105: change to `if (relationship.relationshipValidateRules)` it will always be array

generators/entity/index.js
```diff
@@ -1566,6 +1598,14 @@ module.exports = EntityGenerator.extend({
...
+                    relationship.relationshipValidate = true;
```
deepu105: move `this.validation = true;` here and avoid the additional check below

generators/entity/templates/src/main/java/package/domain/_Entity.java
```diff
@@ -142,6 +144,9 @@
 
     <%_ } else if (relationshipType == 'many-to-one') { _%>
     @ManyToOne
+    <%_ if (relationshipValidate == true && relationshipType == 'many-to-one' && relationshipValidateRules.indexOf('required') != -1) { _%>
```
deepu105: I think you dont have to check the relationship types again as you will set `relationshipValidate ` only when thse conditions are met so you can simplify to `<%_ if (relationshipValidate) { %>`

generators/entity/templates/src/main/java/package/domain/_Entity.java
```diff
@@ -154,6 +159,9 @@
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
     <%_     }
             if (ownerSide == true) { _%>
+    <%_ if (relationshipValidate == true && relationshipValidateRules.indexOf('required') != -1) { _%>
```
deepu105: here as well, same comment as above

generators/entity/templates/src/main/java/package/domain/_Entity.java
```diff
@@ -163,6 +171,9 @@
     <%_ } else { _%>
     <%_     if (ownerSide) { _%>
     @OneToOne
+    <%_ if (relationshipValidate == true && relationshipType == 'one-to-one' && relationshipValidateRules.indexOf('required') != -1) { _%>
```
deepu105: same here just `<%_ if (relationshipValidate) { %>` is sufficient

dornala:
Sorry Deepu. I don't know that we can update the same PR. I followed same code guidelines as from fields. I will make those changes and submit. Thank you.. :)

deepu105:
Just push new commits to your pr branch and it will be updated
automatically

generators/entity/index.js
```diff
@@ -933,6 +937,30 @@ module.exports = EntityGenerator.extend({
...
                 },
                 default: 'id'
+            },
+            {
+                when: function (response) {
+                    return (response.relationshipAdd === true && (response.relationshipType === 'many-to-one' || (response.relationshipType === 'many-to-many' && response.ownerSide === true) || (response.relationshipType === 'one-to-one' && response.ownerSide === true)));
+                },
+                type: 'confirm',
+                name: 'relationshipValidate',
+                message: 'Do you want add any validation rules to this relationship?',
```
deepu105: "Do you want to add any validation rules to this relationship?" missing `to`

generators/entity/index.js
```diff
@@ -1566,6 +1596,18 @@ module.exports = EntityGenerator.extend({
                     this.fieldsContainManyToOne = true;
                 }
 
+                if (relationship.relationshipValidateRules && relationship.relationshipValidateRules.length >= 1) {
```
deepu105: the three if conditions below can be simplified to just one like this
```
if (relationship.relationshipValidateRules && relationship.relationshipValidateRules.length >= 1) {
    relationship.relationshipValidate = this.validation = this.relationshipRequired = true;
}

File: generators/entity/templates/src/main/java/package/domain/_Entity.java [hunk 9]
```diff
@@ -154,6 +159,9 @@
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
     <%_     }
             if (ownerSide == true) { _%>
+    <%_ if (relationshipRequired) { _%>
```
deepu105: better check `<%_ if (relationshipValidate ) { _%>` alone here to future proof so that other validations can also be added if required, you are already checking the required condition in the ejs file so thats sufficient

generators/entity/index.js
```diff
@@ -1427,6 +1456,7 @@ module.exports = EntityGenerator.extend({
             this.fieldsContainBigDecimal = false;
             this.fieldsContainBlob = false;
             this.validation = false;
+            this.relationshipRequired = false;
```
deepu105: `relationshipRequired ` should at each relationship level so should be`relationship.relationshipRequired `

deepu105:
you also have eslint errors

dornala:
I committed it within my repository deepu that is also pulled as PR. I am still working on it. :(

deepu105:
Its ok nevermind, im changing it and merging so no worries

dornala:
Thank you..

deepu105:
@dornala you can check my modification on the commit above its merged here https://github.com/jhipster/generator-jhipster/commit/38e8de23f636c3c42f7c99ce9b2ece9400ff427a

PR #3506, opened 2016-05-01
Travis retry before protractor tests

pascalgrimaud:
Travis will try to curl the application 10 times.
If it failed, Travis should exit without error => the protractor tests won't be launched.


deepu105:
Awesome

PR #3507, opened 2016-05-01
Add sort function to user management page

sgrillon14:


deepu105:
Could you merge into a single PR instead of opening two PR for same stuff?

deepu105:
Im talking about #3505

pascalgrimaud:
And if I remember well, Cassandra doesnt have sort/pagination options => so need to confirm this PR won't break user management page with Cassandra

sgrillon14:
I offered the same functionality on user management as the entities (same code for sort). If there is a problem here with Cassandra there will also on entities.

pascalgrimaud:
Just checked right now.
On a Cassandra project, when using "yo jhipster:entity foo", there is no option about sort/pagination, so there is no sort/pagination on entities

What I wanted to say, was "Cassandra option" need to be tested also

deepu105:
It might be a bug on entities as well

PR #3518, opened 2016-05-02
Use authorization server's public key on resource servers

sdoxsee:
# Motivation

A shared secret with which to sign JWTs is great when there is no single service responsible for all JWT creation. However, in the case of OAuth2, the authorization server creates and usually signs the token with its private key. The authorization server also exposes its public key to other services so that they can verify the JWT and grant/deny access accordingly.

Some advantages of asymmetric (i.e. public/private key) cryptography vs symmetric (i.e. shared secret) cryptography are:

1. Simpler key management. Asymmetric cryptography only needs to keep the private key safe (on the authorization server) while the public key can be freely exposed. Furthermore, resource servers can dynamically ask the authorization server for its public key. Symmetric cryptography means that the secret must be distributed to all communicating parties (increasing the danger of it being exposed).
2. Non-repudiation: Since the private key of the authorization server is never shared, you can be assured that the JWT was signed by the authorization server and only the authorization server. Symmetric key messages could have come from anyone with the shared secret key.

# Notes

With this PR,

1. A single default keypair is generated on the authorization server (UAA) when it is created. There should be different keypairs for dev/prod. I could use feedback on how we'd like to generate these (e.g. 'silently', 'interactively', is there a proper yeoman way to do this?)
1. Currently, the jhipster-registry assumes JHipsterProperties has a shared secret property on all jhipster apps it's coordinating. I'd like to delete references to the shared secret in jhipster apps that use UAA but can't without changing jhipster-registry. Suggestions?
1. We could have relied on default resource server .yml configuration for the `security.oauth2.resource.jwt.keyUri` property except that is can't handle "discoverable" URIs yet (see
http://stackoverflow.com/questions/36900650/using-security-oauth2-resource-jwt-keyuri-with-a-discoverable-service-name). I've had to manually configure this on initialization. Still, the URL be moved to JHipsterProperties configuration...especially to handle differences in dev/production for http vs https?

I don't imagine this PR is ready to merge but I'd appreciate any feedback to get it closer (ping @xetys)

I should also say that this would take us a step close to being able to do Single Sign On with OAuth2. There are still many permutations of OAuth2 configurations though (including SSO)...and JHipster can't support all of them. It might be worthwhile discussion which one(s) it should support out-of-the-box)

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -473,6 +473,33 @@ module.exports = JhipsterServerGenerator.extend({
                     this.template(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000001_insert_default_users.cql', this, {});
                 }
             }
+
+            if (this.applicationType === 'uaa') {
+                const keyStoreFile = SERVER_MAIN_RES_DIR + 'keystore.jks';
+                if (this.fs.exists(keyStoreFile)) {
+                    this.log(chalk.cyan('\nKeyStore \'' + keyStoreFile + '\' already exists. Leaving unchanged.\n'));
+                } else {
+                    shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
+                    var parent = this;
+                    shelljs.exec('keytool '+
+                        '-genkey ' +
+                        '-noprompt ' +
+                        '-keyalg RSA ' +
+                        '-alias selfsigned ' +
+                        '-keystore ' + keyStoreFile + ' ' +
+                        '-storepass password ' +
+                        '-keypass password ' +
+                        '-keysize 2048 ' +
+                        '-dname "CN=Java Hipster, OU=Development, O=' + this.packageName + ', L=, ST=, C="'
+                    , function(code) {
+                        if (code !== 0) {
+                            parent.env.error(chalk.red(`\nFailed to create a KeyStore with \'keytool\'`), code);
+                        } else {
+                            parent.log(chalk.green('\nKeyStore \'' + keyStoreFile + '\' generated successfully.\n'));
+                        }
+                    });
+                }
+            }
         },
 
         writeServerPropertyFiles: function () {
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -7,8 +7,7 @@ var util = require('util'),
     scriptBase = require('../generator-base'),
     packagejs = require('../../package.json'),
     crypto = require('crypto'),
-    mkdirp = require('mkdirp'),
-    shelljs = require('shelljs');
+    mkdirp = require('mkdirp');
 
 var JhipsterServerGenerator = generators.Base.extend({});
 
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -476,30 +476,7 @@ module.exports = JhipsterServerGenerator.extend({
             }
 
             if (this.applicationType === 'uaa') {
-                const keyStoreFile = SERVER_MAIN_RES_DIR + 'keystore.jks';
-                if (this.fs.exists(keyStoreFile)) {
-                    this.log(chalk.cyan('\nKeyStore \'' + keyStoreFile + '\' already exists. Leaving unchanged.\n'));
-                } else {
-                    shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
-                    var parent = this;
-                    shelljs.exec('keytool '+
-                        '-genkey ' +
-                        '-noprompt ' +
-                        '-keyalg RSA ' +
-                        '-alias selfsigned ' +
-                        '-keystore ' + keyStoreFile + ' ' +
-                        '-storepass password ' +
-                        '-keypass password ' +
-                        '-keysize 2048 ' +
-                        '-dname "CN=Java Hipster, OU=Development, O=' + this.packageName + ', L=, ST=, C="'
-                    , function(code) {
-                        if (code !== 0) {
-                            parent.env.error(chalk.red(`\nFailed to create a KeyStore with \'keytool\'`), code);
-                        } else {
-                            parent.log(chalk.green('\nKeyStore \'' + keyStoreFile + '\' generated successfully.\n'));
-                        }
-                    });
-                }
+                this.generateKeyStore();
             }
         },
 
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -7,7 +7,8 @@ var util = require('util'),
     scriptBase = require('../generator-base'),
     packagejs = require('../../package.json'),
     crypto = require('crypto'),
-    mkdirp = require('mkdirp');
+    mkdirp = require('mkdirp'),
+    shelljs = require('shelljs');
 
 var JhipsterServerGenerator = generators.Base.extend({});
 
--- a/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
@@ -133,15 +133,14 @@ package <%=packageName%>.config;
       }
 
       @Bean
-      @Qualifier("LoadBalancedRestTemplate")
-      public RestTemplate restTemplate(RestTemplateCustomizer customizer) {
+      public RestTemplate loadBalancedRestTemplate(RestTemplateCustomizer customizer) {
           RestTemplate restTemplate = new RestTemplate();
           customizer.customize(restTemplate);
           return restTemplate;
       }
 
       @Inject
-      @Qualifier("LoadBalancedRestTemplate")
+      @Qualifier("loadBalancedRestTemplate")
       private RestTemplate keyUriRestTemplate;
 
       private String getKeyFromAuthorizationServer() {
--- a/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
@@ -72,8 +72,13 @@ package <%=packageName%>.config;
 <% } %>
 <% if(authenticationType == 'uaa') { %>
   import <%=packageName%>.security.AuthoritiesConstants;
+  import org.springframework.beans.factory.annotation.Qualifier;
+  import org.springframework.cloud.client.loadbalancer.RestTemplateCustomizer;
   import org.springframework.context.annotation.Bean;
   import org.springframework.context.annotation.Configuration;
+  import org.springframework.http.HttpEntity;
+  import org.springframework.http.HttpHeaders;
+  import org.springframework.http.HttpMethod;
   import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
   import org.springframework.security.config.annotation.web.builders.HttpSecurity;
   import org.springframework.security.config.http.SessionCreationPolicy;
@@ -82,6 +87,9 @@ package <%=packageName%>.config;
   import org.springframework.security.oauth2.provider.token.TokenStore;
   import org.springframework.security.oauth2.provider.token.store.JwtAccessTokenConverter;
   import org.springframework.security.oauth2.provider.token.store.JwtTokenStore;
+  import org.springframework.web.client.RestTemplate;
+
+  import java.util.Map;
 
   import javax.inject.Inject;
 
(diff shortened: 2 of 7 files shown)


generators/server/index.js
```diff
@@ -1021,6 +1021,16 @@ module.exports = JhipsterServerGenerator.extend({
                     this.copy(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000000_insert_default_users.cql');
                 }
             }
+
+            if (this.applicationType === 'uaa' && !this.fs.exists(SERVER_MAIN_RES_DIR + 'keystore.jks')) {
+                shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
+                if (shelljs.exec('keytool -genkey -noprompt -keyalg RSA -alias selfsigned -keystore ' + SERVER_MAIN_RES_DIR + 'keystore.jks -storepass password -keypass password -keysize 2048 -dname "CN=Java Hipster, OU=Development, O=' + this.packageName + ', L=, ST=, C="').code !== 0) {
```
jdubois: you need to check first if keytool is installed

sdoxsee: Thanks @jdubois 

It should be on the path if `java` is since it's in the `bin` director of `JAVA_HOME` for the JDK. If `keytool` isn't installed, the generator will display:
```
/bin/sh: keytool: command not found
Error: keytool failed
```
Is there another particular check you have in mind?

xetys:
I was going to write some documentation about how to do this manually....since my crap of notebook is back from repair :angry: 

but actually this is the way I also prefer

generators/server/index.js
```diff
@@ -340,7 +340,7 @@ module.exports = JhipsterServerGenerator.extend({
                     message: function (response) {
                         return getNumberedQuestion('What is the folder path of your UAA application?.', currentQuestion, totalQuestions, function (current) {
                             currentQuestion = current;
-                        }, applicationType === 'gateway' && response.authenticationType === 'uaa');
+                        }, (applicationType === 'gateway' || applicationType === 'application') && response.authenticationType === 'uaa');
```
deepu105: we dont have an `applicationType ` called `application` is this a typo? also if you are changing the condition here then the same has to be updated for the when: block as well for the question

generators/server/index.js
deepu105: You could use the yeomans inbuilt spawn command method for this alternatively http://yeoman.io/generator/actions_spawn_command.html

sdoxsee:
Updated the PR with some good suggestions. Thanks! Any other changes or thoughts with regard to...

1. key generation (especially for production)
2. jhipster-registry requiring that JHipsterProperties.java have a JWT secret (even though this wouldn't use it)?

generators/server/index.js
```diff
@@ -1021,6 +1021,37 @@ module.exports = JhipsterServerGenerator.extend({
...
+                    ])
+                    .on('close', (code) => {
+                        if (code === 0) {
+                            this.log(chalk.green('\nKeyStore \'' + keyStoreFile + '\' generated successfully.\n'));
+                        } else {
+                            this.log(chalk.red(`\nFailed to create a KeyStore with \'keytool\'`));
+                            if (code === -2) {
+                                this.log(chalk.red('Is \'keytool\' available on the path along with your JDK installation? (i.e. JAVA_HOME/bin)'));
+                            }
+                            shelljs.exit(code);
```
deepu105: why shelljs.exit?

generators/server/index.js
```diff
@@ -1021,6 +1021,37 @@ module.exports = JhipsterServerGenerator.extend({
...
+                        }
+                    })
+                    .on('error', (err) => {}); // must handle error if spawn failed but 'close' is also triggered.
```
deepu105: can we remove if not needed?

sdoxsee: The yeoman spawnCommand will trigger 'error' if keytool is not installed. It also triggers 'close' as well though. If I don't have anything for 'error' then it throws an 'unhandled' exception. But 'error' is not enough because exit code 1 (which happens if keytool runs but doesn't generate anything) does not trigger 'error'...only 'close'. :P maybe shelljs would be cleaner?

deepu105: I suggested it as an option, no pressure to use, you can use shellJs if you think its more apt for the purpose.

sdoxsee: Thanks @deepu105. I might go back to shelljs. I appreciate the suggestion though as I didn't even know what yeoman had build-in!

sdoxsee:
I've gone back to using shelljs for keytool.

generators/server/index.js
```diff
@@ -1024,6 +1024,31 @@ module.exports = JhipsterServerGenerator.extend({
...
+                        '-alias selfsigned ' +
+                        '-keystore ' + keyStoreFile + ' ' +
+                        '-storepass password ' +
+                        '-keypass password ' +
+                        '-keysize 2048 ' +
+                        '-dname "CN=Java Hipster, OU=Development, O=' + this.packageName + ', L=, ST=, C="'
+                    ).code;
+                    if (exitCode !== 0) {
+                        this.log(chalk.red(`\nFailed to create a KeyStore with \'keytool\'`));
+                        shelljs.exit(exitCode);
```
deepu105: does this exit the generator itself? if so its better we use `this.error()` to avoid any process being affected

generators/server/index.js
```diff
@@ -1024,6 +1024,31 @@ module.exports = JhipsterServerGenerator.extend({
...
                 }
             }
+
+            if (this.applicationType === 'uaa') {
+                const keyStoreFile = SERVER_MAIN_RES_DIR + 'keystore.jks';
+                if (this.fs.exists(keyStoreFile)) {
+                    this.log(chalk.cyan('\nKeyStore \'' + keyStoreFile + '\' already exists. Leaving unchanged.\n'));
+                } else {
+                    shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
+                    var exitCode = shelljs.exec('keytool '+
```
deepu105: couldn't this be done asynchronously? I mean we need not really wait for this to complete right?

deepu105:
Sorry I introduced some merge conflicts here due to moving prompts to a separate file.

sdoxsee:
Merge conflicts have been resolved

deepu105:
thank you

Thanks & Regards,
Deepu

generators/server/index.js
```diff
@@ -471,6 +472,33 @@ module.exports = JhipsterServerGenerator.extend({
                     this.template(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000001_insert_default_users.cql', this, {});
                 }
             }
+
+            if (this.applicationType === 'uaa') {
+                const keyStoreFile = SERVER_MAIN_RES_DIR + 'keystore.jks';
```
deepu105: can we move this entire block of generating keystore to a function in `generator-base.js` and call that from here?

deepu105:
LGTM  :+1:

sdoxsee:
I've finally moved the code block out of server/index.js.

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -121,9 +129,27 @@ public TokenStore tokenStore() {
...
       public JwtAccessTokenConverter jwtAccessTokenConverter() {
           JwtAccessTokenConverter converter = new JwtAccessTokenConverter();
-          converter.setSigningKey(jHipsterProperties.getSecurity().getAuthentication().getJwt().getSecret());
-
+          converter.setVerifierKey(getKeyFromAuthorizationServer());
           return converter;
       }
+
+      @Bean
+      @Qualifier("LoadBalancedRestTemplate")
```
jdubois: by default Spring Bean names start with a lower case, can you rename this "loadBalancedRestTemplate" ?

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -121,9 +129,27 @@ public TokenStore tokenStore() {
...
+      public RestTemplate restTemplate(RestTemplateCustomizer customizer) {
+          RestTemplate restTemplate = new RestTemplate();
+          customizer.customize(restTemplate);
+          return restTemplate;
+      }
+
+      @Inject
+      @Qualifier("LoadBalancedRestTemplate")
+      private RestTemplate keyUriRestTemplate;
```
jdubois: I don't understand, you create a Bean just on line 137, to inject it only here? I think this could be easier

jdubois:
I added a few comments, but is this only for the UAA option? In that case, that's your option @sdoxsee , and that's still in *beta*, so let's merge this!

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
cbornet: I alsso think you can remove the @Qualifier if you call the method "public RestTemplate  loadBalancedRestTemplate(RestTemplateCustomizer customizer)"

jdubois: yes indeed!

sdoxsee:
Thanks everyone for the coding suggestions. I wasn't able to figure out how to use the service name "uaa" in the url to obtain the public key without having to use the rather awkward creation of a bean and then injecting it into the same class. If you have suggestions for getting around this, please let me know. Otherwise, I'm happy to have it merged in "as-is" and let it be improved later.

deepu105:
@xetys are you ok with this?

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
```diff
@@ -87,9 +91,8 @@ public void configure(ClientDetailsServiceConfigurer clients) throws Exception {
          */
         clients.inMemory()
             .withClient("web_app")
-            .scopes("web-app")
+            .scopes("openid")
```
xetys: this change will currently make the client app fail to communicate with the server. 
In angular client there is a "hack" line, where the Basic Auth header is set incode, rather then using Base64. This is done because I initally wanted to avoid base64 encoding on client side. But as the basic auth header is done once for token request, you should apply it there with base64 encode. I would also suggest to set "openid" as angular.constant, so this is easily configurable, if the developer wants to use a different clientId

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -120,9 +128,26 @@ public TokenStore tokenStore() {
...
+          return restTemplate;
+      }
+
+      @Inject
+      @Qualifier("loadBalancedRestTemplate")
+      private RestTemplate keyUriRestTemplate;
+
+      private String getKeyFromAuthorizationServer() {
+          HttpEntity<Void> request = new HttpEntity<Void>(new HttpHeaders());
+          return (String) this.keyUriRestTemplate
```
xetys: is this done once or for every request? Currently i am keeping a "public.cert" in the resource servers resource path. So they have no need to communicate with the authserver to be able to verify the key. This is an security enhancement for me, as well as less request overhead

sdoxsee: This is only part of the initial setup. Only happens when the app starts up...not every request. This is using the auth server's published public key so that each resource server doesn't need to hard-code it. Google publishes theirs here: https://www.googleapis.com/oauth2/v3/certs

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
sdoxsee: I don't think this breaks anything. I believe the former "web-app" scope wasn't being used at all in the client. Base64 is for the client_id and client_secret. I do agree that the client_id and client_secret should be encoded via JS rather than hardcoded...but that's a different matter.

xetys: It is used, when you generate an gateway app using uaa option. Just take a look at the token request. But you can keep it this way. I can fix this also, if you want

deepu105:
@xetys and @sdoxsee let mw know once both of you guys are ok with this and then ill merge it.

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
sdoxsee: @xetys sorry, I still can't see it. If there's something to fix then you're welcome to do it or send me a line number link and a description of what needs to change. I'm pretty sure the token request uses the client_id--not the scope.

xetys: ugh.....i read the false line! I thought you renamed the clientId...

well, i still wouldn't use "openid" as scope, because this actually has not much to do with openid....we do it like openid, without strict openID....but this are nerdy details not really changing the working behaviour...

xetys:
LGTM!

PR #3520, opened 2016-05-03
Fixes mongodb URI with Heroku generator

jkutner:
By parsing the `MONGODB_URI` in the `Procfile`, we can dynamically set the database name.


albertmatyi:
sweet :) great job

PR #3524, opened 2016-05-03
Added additional dir for processor-plugin

sanitar4eg:
This plugin adds a default, but without this option, you can not add other plugins that generate sources

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -865,6 +865,11 @@
                     <defaultOutputDirectory>
                         ${project.build.directory}/generated-sources
                     </defaultOutputDirectory>
+                    <additionalSourceDirectories>
+                        <additionalSourceDirectory>
+                            ${project.build.directory}/generated-sources
+                        </additionalSourceDirectory>
+                    </additionalSourceDirectories>
                     <processors>
                         <processor>org.mapstruct.ap.MappingProcessor</processor>
                     </processors>


deepu105:
What is this for, can you explain what you are trying to do, I dont see why this is required. @jdubois does it make sense to you?

sanitar4eg:
I tried adding querydsl and com.mysema.maven:maven-apt-plugin to the project, and processor-plugin throwed many exceptions.
Maybe exists the other plugins, that have a problem with that

deepu105:
we already have mapstruct which is a pre processor and it works properly so
I guess your issue is something else, probably IDE misconfiguration

Thanks & Regards,
Deepu

sanitar4eg:
Sorry, but it is not IDE misconfiguration.
In my project i have links to querydsl's classes and when MappingProcessor trying analyze project, there missed this classes.
Qurydsl is not like the mapstruct, its additional pre-processor for mapping queries to databases.
I fond many questions on SO related with this error.
Thank you for feedback.

deepu105:
Im sorry but we fo not provide querydsl out of the box and hence I dont think we should have this, may be you can add this to the querydsl tip on our documentation page https://github.com/jhipster/jhipster.github.io

deepu105:
@jdubois plz take a look

jdubois:
This change looks logical, but I'm worried to add more stuff to the pom.xml file, which isn't necessary by our "normal" setup. That would be against our usual philosophy.

I don't even think a tip is needed here: QueryDSL should have a documentation stating how you should configure it work with Maven. It's a QueryDSL documentation issue, not a JHipster documentation issue.

PR #3525, opened 2016-05-03
Fix #3519 ui blob fields show extra commas

### LINKED ISSUE #3519: UI shows extra comma on image/binary blob fields

There is an extra comma shown in the UI for image and binary fields:

screen shot 2016-05-02 at 5 19 53 pm

##### **JHipster Version(s)** 
v3.1

.jhipster/Task.json 
```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "title",
            "fieldType": "String"
        },
        {
            "fieldName": "content",
            "fieldType": "byte[]",
            "fieldTypeBlobContent": "text"
        },
        {
            "fieldName": "image",
            "fieldType": "byte[]",
            "fieldTypeBlobContent": "image"
        },
        {
            "fieldName": "binContent",
            "fieldType": "byte[]",
            "fieldTypeBlobContent": "any"
        }
    ],
    "changelogDate": "20160502204230",
    "dto": "no",
    "service": "no",
    "entityTableName": "task",
    "pagination": "no"
}
```

##### **Suggest a Fix** 

The problem comes from this code:
```
<div class="help-block clearfix">
                    <span class="pull-left">{{vm.task.imageContentType}}, {{byteSize(vm.task.image)}}</span>
                    <button ng-click="vm.task.image=null;vm.task.imageContentType=null;"
                            ng-if="vm.task.image"
                            class="btn btn-default btn-xs pull-right">
                        <span class="glyphicon glyphicon-remove"></span>
                    </button>
                </div>
```

Also, after uploading an image and then trying to edit the entity, the directive `{{byteSize(vm.task.image)}}` doesn't show anything.

I can try to fix these but I don't know how it worked before?
The help-block div doesn't exists for the other field types, so I'm wondering if it was forget after a refactoring.

raphaelbrugier:
Fix #3519

I've also added a translation for "open" the blob and add links around the images.

I'm not happy with the style of the links.
Because there is no "href" attribute, the style to display a pointer cursor is not inherited on the <a> element.
So I had to explicitly add `style="cursor: pointer;"`
Any better idea?


deepu105:
dont know why you had to do that as we have this css by default and it should work
```
/* ==========================================================================
make sure browsers use the pointer cursor for anchors, even with no href
========================================================================== */
a:hover {
  cursor: pointer;
}

.hand {
    cursor: pointer;
}
```

generators/languages/templates/src/main/webapp/i18n/ca/_global.json
```diff
@@ -89,7 +89,8 @@
             "cancel": "Anul.la",
             "delete": "Elimina",
             "edit": "Edita",
-            "save": "Desa",
+            "open": "Open",
+			"save": "Desa",
```
deepu105: formatting seems incorrect

raphaelbrugier: I've used tab for the formatting.
I'll force push a correction

raphaelbrugier:
@deepu105 I've tested again and packaged the app with the prod profile and the cursor style doesn't work.

For some reasons, the style you quoted from main.scss doesn't show up in the generated css file.
I'm using the master fyi.

raphaelbrugier:
Got it, I'm not using SaaS in my test project and the style you mentioned is missing in the plain css file.
I'll add it.

raphaelbrugier:
My mistake, I've probably generated my test project initially from the v3.1.0 tag, because the style does exists in main.css and is working fine in a fresh project.

deepu105:
In master both files have these. So i assume its ok now?

raphaelbrugier:
@deepu105 yes both files being in the master, it's ok. Sorry for the confusion, thanks for the merge!

PR #3529, opened 2016-05-04
Fix lifecycle mapping in Eclipse/STS

erikkemperman:
Since the move to `frontend-maven-plugin` was merged, I noticed Eclipse/STS shows an error. I just investigated a bit and it turns out that the lifecycle mapping metadata section of `pom.xml` must be slightly changed to stop the complaints.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -49,6 +49,9 @@
         <%_ if (testFrameworks.indexOf('cucumber') != -1) { _%>
         <cucumber.version>1.2.4</cucumber.version>
         <%_ } _%>
+        <%_ if (!skipClient) { _%>
+        <frontend-maven-plugin.version>1.0</frontend-maven-plugin.version>
+        <%_ } _%>
         <%_ if (testFrameworks.indexOf('gatling') != -1) { _%>
         <gatling.version>2.1.7</gatling.version>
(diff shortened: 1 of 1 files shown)


jdubois:
I'm a bit worried that the dependency version isn't fixed, do you know which version is used for real in the end?

erikkemperman:
@jdubois Currently there is only one matching version, [link](http://mvnrepository.com/artifact/com.github.eirslett/frontend-maven-plugin).

So you'd prefer a closed range, like `<versionRange>1.0</versionRange>`?

I don't mind either way, that will still get rid of the error I am seeing. For what it's worth though, there is already an open-ended version range for the `jacoco-maven-plugin` plugin...

jdubois:
I'm just afraid that a new version might break everything... Each time we've done this we had issues, and they always happen at the wrong moment, of course!
And for Jacoco that's another error!

erikkemperman:
All right, so I will amend the commit to use a fixed version for frontend-maven-plugin. Should I do the same for jacoco? I just found that there are more recent releases not yet reflected in the current JH pom, too.

deepu105:
Some of the existing plugins in our pom are also having versionRange i guess

erikkemperman:
The PR is adapted, now matches a single version for both plugins, but still uses `versionRange` instead of `version` to keep STS happy.

PR #3530, opened 2016-05-04
add a default logo

deepu105:
add a default logo in navbar, so that users can just replace it with their own logo.


deepu105:
@jdubois please merge if you are ok with this

jdubois:
It looks like the same code than in the JHipster Registry -> yes OK for the idea, but I just changed it in the Registry, as it wasn't displaying correctly, so I need to test before

deepu105:
what happend in registry, it was working when I tested!

deepu105:
Ok you moved the logo to the menu, but I dont think thats the proper place as it would be hidden in mobile, I would rather have the logo and App name in its old location and just remove the home menu instead.

deepu105:
This is what I meant
image

This looks much better and prominent IMO
image

deepu105:
and in desktop view 
image

jdubois:
Agreed, but on my screen (with Chrome), it was displaying like this:
capture d ecran 2016-05-04 12 24 48

deepu105:
Oh thats normal even without the logo image, since the menu items are more there is no space so bootstarp kicks it down. If you remove home, it would be ok in your screen, but in smaller screen it will look that way and IMO thats normal bootstrap

deepu105:
may be we should try to group few menu items together. also we can remove the text for signout, the icon is kind of standard

atomfrede:
Maybe just remove ``Jhipster``, if we have the logo in the navbar we could just have ``Registry v1.1.0`` there. I think the Java Hipster is already well known :)

deepu105:
Yup agree it will be cooler :)
also remove home menu.
remove signout text, i guess icon would be sufficient

so together it will use less space and should look ok in your display. Ill
later add some responsive util classes to limit breakpoints

Thanks & Regards,
Deepu

jdubois:
OK I'm trusting you on this -> and this needs to be copied to the Registry too (or start with the Registry, I find it easier)

deepu105:
This PR is good to go :+1:

deepu105:
@jdubois can i merge this?

jdubois:
Yes

PR #3543, opened 2016-05-05
add default troubleshooting steps fix #3541

### LINKED ISSUE #3541: [Feature] Have a default "blank page" for when generation went wrong
We have many people complaining about a "blank page" when generation fails.

Usually this could be solved by looking at the logs, running again "jhispster --force", etc... There are already lots of tickets about that.

So my proposal is that our default page shows an error explaining those common causes, where to look at, etc. And it would be modified at the end of the generation.

deepu105:
@jdubois please see if the steps are ok. @jhipster/developers let me know if any more steps needs to be added.

The gulp task is required only if we dont want the content in html at all, else angular takes care of not showing the content if gulp install runs successfully

fix #3541

### DIFF
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -137,7 +137,6 @@ gulp.task('inject:app', function () {
         .pipe(gulp.dest(config.app));
 });
 
-
 gulp.task('inject:vendor', function () {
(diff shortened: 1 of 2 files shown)


jdubois:
Perfect!!
I'll probably modify the text a little bit, but the most important thing is that we have a placeholder for this, and then we can improve it

PR #3544, opened 2016-05-05
schema management for Cassandra

raphaelbrugier:
This is a follow-up of the discussion with @pascalgrimaud and @jdubois  in #3288

This pull request adds a table to track the cql scripts executed in cassandra and avoid run them twice, similar to what is done by liquibase/flyway.

The `auto-migrate.sh` and `execute-cql.sh` scripts use the `schema_version` table to only run cql scripts from the changelog folder once.
If a cql script is changed after being executed, an error will be reported.

#### Usage:
1. Start the docker container for cassandra with docker-compose:
 - `docker-compose -f src/main/docker/cassandra.yml up -d`
2. Add some cql scripts in the changelog folder
3. Execute the cql scripts (see below)

There are now 3 ways to execute the cql scripts without restarting the cassandra cluster:
- Run the automatic migration from a docker container: 
 - `docker-compose -f src/main/docker/cassandra.yml up <app>-cassandra-migration`
- Run the automatic migration from locally:
 - `src/main/docker/cassandra/scripts/autoMigrate.sh src/main/resources/config/cql/changelog/`
- Execute only the new script from locally:
 - `src/main/docker/cassandra/scripts/execute-cql.sh src/main/resources/config/cql/changelog/<script>.cql`

To run locally there are some dependencies:
- define the `CASSANDRA_CONTACT_POINT` environment variable:
 - `export CASSANDRA_CONTACT_POINT=\`docker-machine ip default\``
- install: bash >4 and md5sum
- have the cqlsh command in the classpath

I'm opening the PR for discussion.
When improvement I'd like to do is to have the `create-tables.cql` script moved to the changelog folder to be managed as well.

### DIFF
--- a/generators/server/templates/src/main/docker/cassandra/scripts/_autoMigrate.sh
+++ b/generators/server/templates/src/main/docker/cassandra/scripts/_autoMigrate.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Orchestrate the automatic execution of all the cql migration scripts when starting the cluster
 # for local development
@@ -10,12 +10,16 @@ function log {
     echo "[$(date)]: $*"
 }
 
+function logDebug {
+    ((DEBUG_LOG)) && echo "[DEBUG][$(date)]: $*"
+}
+
 function waitForClusterConnection() {
     retryCount=0
     maxRetry=20
     cqlsh -e "Describe KEYSPACES;" $CASSANDRA_CONTACT_POINT &>/dev/null
     while [ $? -ne 0 ] && [ "$retryCount" -ne "$maxRetry" ]; do
-        log 'cassandra not reachable yet. sleep and retry. retryCount =' $retryCount
+        logDebug 'cassandra not reachable yet. sleep and retry. retryCount =' $retryCount
         sleep 5
         ((retryCount+=1))
         cqlsh -e "Describe KEYSPACES;" $CASSANDRA_CONTACT_POINT &>/dev/null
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
@raphaelbrugier : good job ! seems good
I dont think I have time to review and test it (:fr: c'est férié aujourd'hui + long weekend! ;-)
I'll try to do my best

raphaelbrugier:
No hurries!

I'll try to push the move of the `create-table.cql` script in the meantine.

I forgot it was off today :)

generators/server/templates/src/main/docker/cassandra/scripts/_autoMigrate.sh
```diff
@@ -10,12 +10,16 @@ function log {
...
+    ((DEBUG_LOG)) && echo "[DEBUG][$(date)]: $*"
+}
+
 function waitForClusterConnection() {
     retryCount=0
     maxRetry=20
     cqlsh -e "Describe KEYSPACES;" $CASSANDRA_CONTACT_POINT &>/dev/null
     while [ $? -ne 0 ] && [ "$retryCount" -ne "$maxRetry" ]; do
-        log 'cassandra not reachable yet. sleep and retry. retryCount =' $retryCount
+        logDebug 'cassandra not reachable yet. sleep and retry. retryCount =' $retryCount
```
pascalgrimaud: on local machine, if user doesn't have cqlsh, the script won't work and there is no message to warn user
=> anyway, it's not so important, it's supposed to be used with docker, right ? :D

pascalgrimaud:
it works perfectly !

raphaelbrugier:
Thx for the merge @pascalgrimaud !
Yes, I could add a message if cqlsh is not in the classpath.

Also, what about the create_table.cql scripts which is not tracked. What do you think to move it in the changelog folder?

PR #3562, opened 2016-05-07
notify updated version availability

### LINKED ISSUE #3555: Notify user about an updated version if available when generator is run
Notify user about an updated JHipster version if available when generator is run.

Similar to yeoman and some other tools. Its easy to do using request module
```

     _-----_
    |       |    .------------------------------------------.
    |--(o)--|    | Update available: 1.7.1 (current: 1.7.0) |
   `---------´   |     Run npm install -g yo to update.     |
    ( _´U`_ )    '------------------------------------------'
    /___A___\
     |  ~  |
   __'.___.'__
 ´   `  |° ´ Y `
```

deepu105:
This shows a nice notification to user when there is an updated JHipster version available via npm

image

fix #3555

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -58,7 +58,7 @@ module.exports = JhipsterGenerator.extend({
 
         // This adds support for a `--[no-]check-install` flag
         this.option('check-install', {
-            desc: 'Check the installation',
+            desc: 'Check the status of the required tools',
             type: Boolean,
             defaults: true
         });
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -17,6 +17,7 @@ var path = require('path'),
 const JHIPSTER_CONFIG_DIR = '.jhipster';
 const MODULES_HOOK_FILE = JHIPSTER_CONFIG_DIR + '/modules/jhi-hooks.json';
 const WORD_WRAP_WIDTH = 80;
+const GENERATOR_JHIPSTER = 'generator-jhipster';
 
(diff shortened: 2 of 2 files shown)


jdubois:
Great idea!
One related issue, due to the yeoman-maven-plugin : if you want to update the version of JHipster for a specific project, you need to delete the node_modules folder -> do you have a better solution for this? and if the folder isn't deleted, I don't think your code here will work, as the project will still use the older generator.

deepu105:
Yes if they run jhipster on an existing project it will use the generator
which was installed by its initial creation. I dont understand the yeoman
maven plugin connection here.

But if the have an old generator and if they do a fresh project then this
notification will work.

Any solution we do will not work for current installations anyway. This
notification will also work the next time they update so, may be we can add
a line there asking people to clear node modules and do a fresh run if they
upgrading an existing project.

IMO we shouldn't worry much about this scenario

jdubois:
Yes, let's merge this anyway, and then improve it if possible

pascalgrimaud:
I think the message should be `npm update -g generator-jhipster` instead of `npm install -g generator-jhipster`

deepu105:
Npm recommends npm install command for update as well these days :)

deepu105:
Btw you can test the feature by changing the version in the package json of
your local npm linked jhipster codebase. Change it temporary to 3.1.0 or
sonething and run yo jhipster somewhere

jdubois:
Oh I didn't know you could test it, cool!

PR #3564, opened 2016-05-07
replace ng constant plugin

atomfrede:
Replace the ng constant fork with the original upstream plugin as the fork didn't work with node 6 (arch linux).

@deepu105 Sass is working fine for me


deepu105:
:+1:

PR #3569, opened 2016-05-09
Fix npm test for uaa

pascalgrimaud:


deepu105:
Thank you I forgot to change this

Issue #3548, opened 2016-05-06
JSR310 support for Cassandra


cbornet:
I would like to add JSR310 dates support for Cassandra. The 3.0 version of the datastax driver has built-in support for this so it would be quite easy if we upgrade the driver.
To note is that ZonedDateTime is mapped to CQL ```tuple<timestamp,varchar>``` to keep the timezone information and it is Instant that is mapped to CQL timestamp.

Actually, for datetime we currently use ZonedDateTime but I think it would be better to use Instant instead since we don't store the timezone information and we always convert to UTC before storage.

So some questions before implementing:
- should we drop ZonedDateTime and use Instant (breaking change) ?
- or should we have both ZonedDateTime and Instant in parallel (kinda redundant) ?
- or should we use ZonedDateTime also for Cassandra ? In this case, do we use the datastax codec or do we write our own that maps to a simple timestamp (more compact than tuple<timestamp,varchar>) ?

cbornet:
@jhipster/developers opinion is welcome :smile: !

pascalgrimaud:
ping @raphaelbrugier too

pascalgrimaud:
By upgrading to datastax driver 3.0, we should be able to use Cassandra 3.0, right?
And users should be able to use older version of Cassandra, 2.0, 2.1 etc => http://datastax.github.io/java-driver/manual/native_protocol/

cbornet:
Yes but currently, I'm having a hard time with spring-data-cassandra as it doesn't support C* 3.0.
I'm not sure why we have the dependency on spring-data-cassandra since we use the driver directly. I will have a look.

pascalgrimaud:
Strange! And our documentation http://jhipster.github.io/using-cassandra/ said :
> The Cassandra CQL driver is used to access the database. We don't use Spring Data Cassandra, as we prefer to use the driver directly. As a result, the repositories contain quite a lot of code

cbornet:
Yes, I don't know why we don't use spring-data-cassandra. @jdubois ?

raphaelbrugier:
Yes, the C* 3.0 driver will support read/write to a 2.x version.

I haven't use the C* 3.0 driver, so I don't know much about the new extra library.
For reference, the javadoc for the codecs you mentioned @cbornet is [here](http://docs.datastax.com/en/drivers/java/3.0/index.html?com/datastax/driver/extras/codecs/jdk8/LocalDateCodec.html).

[...]

PR #3469, opened 2016-04-25
cassandra warning about netty's native library

raphaelbrugier:
fix #3428

Issue #3428, opened 2016-04-19
Cassandra driver error when starting Cassandra microsevice with docker-compose file


raphaelbrugier:
##### **Overview of the issue**

When starting a Cassandra microservice with the docker-compose generated files, the following warning and stack trace appears in the log:

```
2016-04-19 14:16:59.249  INFO 7 --- [           main] c.j.test.config.MetricsConfiguration     : Initializing Metrics Log reporting
2016-04-19 14:17:05.551  INFO 7 --- [ost-startStop-1] com.mycompany.myapp.config.WebConfigurer   : Web application configuration, using profiles: [prod]
2016-04-19 14:17:05.579  INFO 7 --- [ost-startStop-1] com.mycompany.myapp.config.WebConfigurer   : Web application fully configured
2016-04-19 14:17:07.081  INFO 7 --- [ost-startStop-1] ccom.mycompany.myapp.CasstestApp            : Running with Spring profile(s) : [prod]
2016-04-19 14:17:08.003  INFO 7 --- [           main] c.j.test.config.LoggingConfiguration     : Initializing Logstash logging
2016-04-19 14:17:09.065  WARN 7 --- [           main] com.datastax.driver.core.NettyUtil       : Found Netty's native epoll transport in the classpath, but epoll is not available. Using NIO instead.
java.lang.UnsatisfiedLinkError: no netty-transport-native-epoll in java.library.path
at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1864)
at java.lang.Runtime.loadLibrary0(Runtime.java:870)
at java.lang.System.loadLibrary(System.java:1122)
at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:168)
at io.netty.channel.epoll.Native.<clinit>(Native.java:49)
at io.netty.channel.epoll.Epoll.<clinit>(Epoll.java:30)
at java.lang.Class.forName0(Native Method)
at java.lang.Class.forName(Class.java:264)
at com.datastax.driver.core.NettyUtil.<clinit>(NettyUtil.java:68)
at com.datastax.driver.core.NettyOptions.eventLoopGroup(NettyOptions.java:101)
at com.datastax.driver.core.Connection$Factory.<init>(Connection.java:695)
at com.datastax.driver.core.Cluster$Manager.init(Cluster.java:1286)
at com.datastax.driver.core.Cluster.init(Cluster.java:159)
[...]

PR #3570, opened 2016-05-09
Remove spring-data-cassandra and upgrade to datastax driver v3

cbornet:
See #3548 

Also add protocolVersion to cluster settings
Also export the driver metrics.

### DIFF
--- a/generators/entity/templates/src/main/java/package/repository/_EntityRepository.java
+++ b/generators/entity/templates/src/main/java/package/repository/_EntityRepository.java
@@ -72,7 +72,8 @@ public class <%= entityClass %>Repository {
                     var fieldNameUnderscored = fields[idx].fieldNameUnderscored;
                     if (fields[idx].fieldType == 'Integer') { %>
                 <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getInt("<%= fieldName %>"));<% } else if (fields[idx].fieldType == 'BigDecimal') { %>
-                <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getDecimal("<%= fieldName %>"));<% } else if (fields[idx].fieldType == 'Boolean') { %>
+                <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getDecimal("<%= fieldName %>"));<% } else if (fields[idx].fieldType == 'Date') { %>
+                <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getTimestamp("<%= fieldName %>"));<% } else if (fields[idx].fieldType == 'Boolean') { %>
                 <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getBool("<%= fieldName %>"));<% } else if (fields[idx].fieldType == 'Text') { %>
                 <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getString("<%= fieldName %>"));<% } else if (fields[idx].fieldType === 'ByteBuffer') { %>
                 <%= entityInstance %>.set<%= fieldInJavaBeanMethod %>(row.getBytes("<%= fieldName %>"));
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -217,7 +217,7 @@ module.exports = JhipsterServerGenerator.extend({
 
     _getUaaAppName : function (input) {
         if (!input) return false;
-        
+
         input = input.trim();
         var fromPath = '';
         if(path.isAbsolute(input)) {
@@ -1195,6 +1195,10 @@ module.exports = JhipsterServerGenerator.extend({
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_package-info.java', javaDir + 'config/metrics/package-info.java', this, {});
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_JHipsterHealthIndicatorConfiguration.java', javaDir + 'config/metrics/JHipsterHealthIndicatorConfiguration.java', this, {});
                 this.template(SERVER_MAIN_SRC_DIR + 'package/config/metrics/_CassandraHealthIndicator.java', javaDir + 'config/metrics/CassandraHealthIndicator.java', this, {});
+            }
+
+            if (this.databaseType === 'cassandra' || this.applicationType === 'gateway') {
+                this.template(SERVER_MAIN_SRC_DIR + 'package/config/_CassandraConfiguration.java', javaDir + 'config/CassandraConfiguration.java', this, {});
             }
 
             if (this.hibernateCache === 'hazelcast') {
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -187,10 +187,10 @@ dependencies {
     compile "org.springframework.boot:spring-boot-starter-web"<% if (websocket == 'spring-websocket') { %>
     compile "org.springframework.boot:spring-boot-starter-websocket"<% } %>
     compile "org.springframework.boot:spring-boot-starter-thymeleaf"<% if (databaseType == 'cassandra' || applicationType == 'gateway') { %>
-    compile "org.springframework.data:spring-data-cassandra"
     compile("com.datastax.cassandra:cassandra-driver-core:${datastax_driver_version}") {
         exclude module: 'com.codahale.metrics'
     }
+    compile "com.datastax.cassandra:cassandra-driver-extras:${datastax_driver_version}"
     compile "com.datastax.cassandra:cassandra-driver-mapping:${datastax_driver_version}"<% } %>
     <%_ if (applicationType == 'gateway') { _%>
     compile "org.springframework.cloud:spring-cloud-starter-zuul"
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -4,12 +4,12 @@ assertj_core_version=3.3.0
 awaility_version=1.7.0
 commons_lang_version=2.6
 commons_io_version=2.4<% if (databaseType == 'cassandra') { %>
-cassandra_unit_spring_version=2.1.3.1<% } %><% if (testFrameworks.indexOf('cucumber') != -1) { %>
+cassandra_unit_spring_version=3.0.0.1<% } %><% if (testFrameworks.indexOf('cucumber') != -1) { %>
 cucumber_version=1.2.4<% } %>
 dropwizard_metrics_version=3.1.2
 logstash_logback_encoder_version=4.6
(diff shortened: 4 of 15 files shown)


pascalgrimaud:
@cbornet : can you rebase, for npm test with uaa working

cbornet:
Can someone review. Especially the microservice part.

jdubois:
I'm having a look at it.
@cbornet @raphaelbrugier can you have a look at also updating http://jhipster.github.io/using-cassandra/

raphaelbrugier:
Great work @cbornet!
Yes I'll update the page to explains about the auto-migration.

jdubois:
OK and I'll merge the documentation when the release is done (so the documentation matches the code)

jdubois:
@cbornet I'm still have `protocolVersion: V2` in application-dev.yml and application-prod.yml, is that normal?

jdubois:
@raphaelbrugier I've also seen http://jhipster.github.io/docker-compose/#4 but I'm unable to have this working with Docker, and anyway it should work out of the box with the default config, so there's something to improve here

cbornet:
Yes, I didn't upgrade the docker cassandra.
I was thinking of doing that in a second step

jdubois:
Oh sorry that's logical then!

raphaelbrugier:
@jdubois When you said `it should work out of the box with the default config` you were using this branch code right? It should work out of the box with the v3.2.1
Let me know if it doesn't.

jdubois:
@raphaelbrugier yes I'm using your code, but I started Cassandra with Docker and as @cbornet said this was not working yet...
I've installed it manually and:

- I had to switch the protocol to "V4" (see my earlier comment on that)
- The schema isn't automatically created (it is with Docker, but you could also run the create_keyspace.cql script automatically)

jdubois:
@cbornet this is also causing issues with microservices, this is a classpath error as both Cassandra and Netflix Eureka use the same Apache Commons lib:

    java.lang.NoSuchMethodError: org.apache.commons.codec.binary.Base64.<init>(I)V

cbornet:
Regarding Datastax Enterprise, do we still try to follow it ?
In that case, maybe we should use:
* cassandra-driver-dse:3.0.0.rc1
* cassandra v2.1.14

Note that currently, our docker config spawns a cassandra 2.2

Regarding Cassandra v3, you indeed need to use protocol v4

As for the lib conflict, aouch ! @jdubois any idea to overcome this issue ? Excluding org.apache.commons from one of the libs ?

jdubois:
- yes the Docker config should also be updated
- I would use Apache Cassandra, but the same version as the latest DataStax version
- For the version conflict, we must check if there is a version working for both...

jdubois:
For the version conflict, you must remove commons-codec from Cassandra Unit:

    <dependency>
            <groupId>org.cassandraunit</groupId>
            <artifactId>cassandra-unit-spring</artifactId>
            <version>3.0.0.1</version>
            <scope>test</scope>
            <exclusions>
                <exclusion>
                    <groupId>org.slf4j</groupId>
                    <artifactId>slf4j-log4j12</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>commons-codec</groupId>
                    <artifactId>commons-codec</artifactId>
                </exclusion>
            </exclusions>
        </dependency>

Still there are 3 big issues here:

- This dependency has the "test" scope, so this shouldn't happen at all
- Now the microservice get registered into the Registry, but it has the "DOWN" status, so something else is wrong
- When I generate the front-end code on the gateway, it doesn't work: it seems that it is because we don't have pagination working. Anyway nothing is generated, so that's totally buggy.

cbornet:
@jdubois would you share your gateway and microservice config ? I'll have a look.

pascalgrimaud:
Bad news!!
I thought the problem with "Down" https://github.com/jhipster/generator-jhipster/issues/3428 was resolved here 
https://github.com/jhipster/generator-jhipster/pull/3469

cbornet:
The driver does indeed use netty so it may be related.
@jdubois do you get the same error as in #3428 ?

As for the pagination issue, you mean pagination on a cassandra entity ? This shouldn't be possible.

jdubois:
@cbornet no I don't think it's the same error as #3428
For the pagination, I haven't checked recently, but originally I coded pagination with "pager" and "infinite scroll" for Cassandra (as you shouldn't make count(), you can't have the "pagination links"). So yes this is supposed to work, but it's currently broken.
You can test my sample project [here](https://www.dropbox.com/s/sw692j6xzm3072u/tmp-microservices2.zip?dl=0)

jdubois:
@raphaelbrugier I haven't checked your code but I have a big doubt that your changelogs can work correctly:

- Liquibase works because it does a transaction and locks a table, which is something you can't do with Cassandra (maybe with Paxos??)
- The problem is that you could launch 10 concurrent instances of one service, so they will all try to update the database schema: with Liquibase, the first one which gets the lock does the update, and the others wait, and then don't need to do the update... I don't think this could work for Cassandra, at least not easily: have you checked this use-case?

raphaelbrugier:
#3469 and #3428 were not related to the DOWN flag in the registry but to the warning message displayed in the console when starting the service. At that point the service ran well, I don't remember seeing something wrong in the registry.

@jdubois:
- yes this is __not__ transactional, but the row in the schema_version table is only inserted after the script from the changelog folder is run with success.
 - So we can assume in worst case the row in schema_version is not inserted and the cql script will be run again next time.
- I'm not following you here. The service (the app) is not doing the actual migration. It's done in a separated docker container, dedicated to run the migration script.
 - good thing is this ensure there are not multiple concurrent migrations launched 
 - bad thing is it only works with docker or by manually launching the script

But it's common to have the migration launched by the deployment tool rather than the app itself.

jdubois:
@raphaelbrugier OK I need to check in detail how it works. It's OK for me if it only works with Docker and/or manually, it's more complex to handle than for an SQL database.
The idea with JHipster is that it's all automatic, that's how it works with Liquibase: not everybody will like this, of course, but it's a lot easier and faster

raphaelbrugier:
We agree then.
As you said, I don't think we can have the same level of automation as for an SQL database.
But relying on docker really simplifies the problem.
You can even launch the docker container locally but pointing to a remote (prod?) cluster to execute the migration.

pascalgrimaud:
The automigration didn't work for cassandra microservice.
I saw it with this log:
```
mscassandra-cassandra-migration    | <stdin>:1:InvalidRequest: code=2200 [Invalid query] message="unconfigured table schema_version"
```
It's because we use 'copy' instead of 'template', so the create-table.cql was empty

cbornet:
I have a hard time testing a full microservice setup on my lame HW. Can someone check the DOWN issue ?
@jdubois if you rollback the driver version, does it work ?

pascalgrimaud:
I confirm the "down" issue on the current master. It's not related to your
PR, @cbornet
Le 11 mai 2016 8:34 AM, "Christophe Bornet" <notifications@github.com> a
écrit :

cbornet:
@pascalgrimaud can you open an issue for the down problem ?

pascalgrimaud:
Yes, as soon as I retry and get the logs, I will open this issue

jdubois:
I'm testing with the current master, and there is currently a bug -> the Docker image doesn't match the Spring Data Cassandra provided driver version.
So both the marster and this PR seem broken :-(

@raphaelbrugier do you have time to validate this PR? I would like it to work:

- on monoliths
- gateways -> have a look at the rate limiting feature
- microservices

I'm going to work on the "DOWN" issue

jdubois:
- I corrected the "commons-codec" issue, and as the current master is broken with Docker, I'm merging this -> as long as it's better than the current code, let's merge!
- @raphaelbrugier can you add some documentation on http://jhipster.github.io/using-cassandra/ ? This isn't obvious, I tried to write something but that's not enough
- @raphaelbrugier with Docker there are lots of error by default because of the 2 nodes -> could you look for a solution? At least catch the errors and document. For a new user the first impression wouldn't be good.
- I'll work on the "down" issue

raphaelbrugier:
Hi,

Yes @jdubois I will work on that:
- update the doc
- prepare a PR to move the create-tables.cql file to the changelog
- test all of this with mono/micro app

@jdubois could you please open an issue or mail me the error message about the problem with a 2nd node?
I've ran into something that sounds similar on my initial setup for tatami and it was because of Docker running inside a VM on OS/X. But since I've configured my laptop I don't see any error anymore with multiple nodes on a Cassandra>Docker cluster.

jdubois:
@raphaelbrugier the error is because the second node has an internal Docker IP:

    2016-05-12 16:47:54.238 DEBUG 6660 --- [uster1-worker-0] com.datastax.driver.core.Host.STATES     : [/172.18.0.3:9042] starting reconnection attempt
    2016-05-12 16:47:54.240  WARN 6660 --- [r1-nio-worker-2] com.datastax.driver.core.Session         : Error creating pool to /172.18.0.3:9042
      com.datastax.driver.core.exceptions.ConnectionException: [/172.18.0.3] Pool was closed during initialization
	at com.datastax.driver.core.HostConnectionPool$2.onSuccess(HostConnectionPool.java:149)

raphaelbrugier:
@jdubois Yes it's the same issue then.
I've documented it on the [docker-compose page](https://github.com/jhipster/jhipster.github.io/blob/master/pages/docker_compose.md#note-for-non-linux-users):

> Cassandra nodes : Because Cassandra nodes are also hosted in the Virtual machine, the Cassandra Java driver will receive an error when trying to contact them after receiving their address from the contact point. To workaround this, you can add a routing rule to your routing table, (source).
> 
> Assuming the containers running the Cassandra nodes have IP address 172.18.0.x:
> 
>  `sudo route -n add 172.18.0.0/16 `docker-machine ip default`

Could you please try this and tell me if it works for you?

Do you think it will be more visible if I move this to the documentation page dedicated to C* ?

jdubois:
Sorry @raphaelbrugier I won't have the time to test, but I'm pretty sure this is the issue, and you can easily reproduce it.

- Yes I've seen your documentation, but it's a bit hidden
- My main problem is that everything is done with JHipster so that it works "out of the box" -> isn't there a solution for this? Like forbidding to connect to that node? To be honest I haven't understood why you need it, yet. If there is no other solution, you could also catch the exception and print a specific log message, pointing to the documentation.

raphaelbrugier:
Ok, np.

I prefer to open a new issue so we can discuss the options there.

cbornet:
@jdubois as for pagination, I confirm that it is currently disabled for cassandra : https://github.com/jhipster/generator-jhipster/blob/master/generators/entity/index.js#L1315

jdubois:
yes I noticed this @cbornet -> it's probably just me lacking time, because that was meant to work. The only option that should not be implemented is the "pagination links" as you shouldn't do a count() with Cassandra, but the other options should be OK.

PR #3576, opened 2016-05-10
Cassandra polish cql files

pascalgrimaud:
related to https://github.com/jhipster/generator-jhipster/pull/3570#issuecomment-218275438


generators/server/index.js
```diff
@@ -1018,7 +1018,7 @@ module.exports = JhipsterServerGenerator.extend({
                 this.template(SERVER_MAIN_RES_DIR + 'config/cql/_create-keyspace-prod.cql', SERVER_MAIN_RES_DIR + 'config/cql/create-keyspace-prod.cql', this, {});
                 this.template(SERVER_MAIN_RES_DIR + 'config/cql/_create-keyspace.cql', SERVER_MAIN_RES_DIR + 'config/cql/create-keyspace.cql', this, {});
                 this.template(SERVER_MAIN_RES_DIR + 'config/cql/_drop-keyspace.cql', SERVER_MAIN_RES_DIR + 'config/cql/drop-keyspace.cql', this, {});
-                this.copy(SERVER_MAIN_RES_DIR + 'config/cql/create-tables.cql', SERVER_MAIN_RES_DIR + 'config/cql/create-tables.cql');
```
jdubois: If it's a template, can you also prefix the file with "_" like the other templates? It's just so we know it's template

pascalgrimaud: yes I will change it!

raphaelbrugier:
+1 for using the template method instead of the copy.

This was already wrong before the introduction of the automigration, so I'm wondering how this could have worked before.
And why the `create-table.cql` script generated was not empty and it worked when generating for a monolith app?

I'm moving this file in the changelo folder in an upcoming PR anyway.
I'll test for the microservice case.

generators/server/index.js
```diff
@@ -1018,7 +1018,7 @@ module.exports = JhipsterServerGenerator.extend({
...
+                this.template(SERVER_MAIN_RES_DIR + 'config/cql/_create-tables.cql', SERVER_MAIN_RES_DIR + 'config/cql/create-tables.cql');
```
raphaelbrugier: Are you not supposed to pass `this, {}` in parameters as well?

generators/server/index.js
```diff
@@ -1018,7 +1018,7 @@ module.exports = JhipsterServerGenerator.extend({
...
                 this.copy(SERVER_MAIN_RES_DIR + 'config/cql/changelog/README.md', SERVER_MAIN_RES_DIR + 'config/cql/changelog/README.md');
                 if (this.applicationType !== 'microservice') {
                     this.copy(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000000_insert_default_users.cql');
```
raphaelbrugier: This should use the template method also.

generators/server/index.js
pascalgrimaud: thx!

generators/server/index.js
pascalgrimaud: I kept copy, as it's not a template. But I renamed the file to `insert_default_users.cql`

raphaelbrugier:
@pascalgrimaud Are you sure it's not a template file?
Because there are directives in that [file](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/cql/changelog/_insert_default_users.cql#L20)

I though this kind of directives `<%= nativeLanguage %>` were processed by the yeoman template engine.

pascalgrimaud:
Oh my bad! I want to work too fast!!!

raphaelbrugier:
Ok :) One moment I though there was another template engine I wasn't aware :)

deepu105:
By the way both `template` and `copy` method will process ejs templates in
them so they wont cause any issues, but aesthetically it might be better to
use template

Thanks & Regards,
Deepu

raphaelbrugier:
Thanks @deepu105, this explains why it was working _fine_ 
So I don't think this will fix the issue.
@pascalgrimaud Could you describe more the setup which caused a problem with the migration script please?

pascalgrimaud:
Very strange.
Current master:
- monolithic => create table.cql is ok
- microservice => create table.cql is empty, the new schema table isnt
created, so the automigration cant work

I dont investigate to understand why the cql is empty or not, but by
replacing copy by template seems to resolve the problem => in this PR

Dont hesitate to tell me if u need more information to try to reproduce the
issue
Le 11 mai 2016 6:05 PM, "Raphaël Brugier" <notifications@github.com> a
écrit :

raphaelbrugier:
I couldn't reproduce the issue.
using master rev 4f7478e4c4a8f91b0f3d70fdd955f1fb569f6028

Generating a microservice with cassandra produced the following `create-table.cql`:

```
CREATE TABLE IF NOT EXISTS schema_version (
    script_name text,
    checksum text,
    executed_by text,
    executed_on timestamp,
    execution_time int,
    status text,
    PRIMARY KEY(script_name)
);
```

with .yo-rc.json:

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.2.1",
    "baseName": "cassapp",
    "serverPort": "8081",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "databaseType": "cassandra",
    "searchEngine": "no",
    "buildTool": "maven",
    "jwtSecretKey": "755cd143449114d29cb9ccd3741085108839ef1d",
    "enableTranslation": true,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true,
    "skipUserManagement": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ],
    "clusteredHttpSession": "no",
    "websocket": "no",
    "enableSocialSignIn": false
  }
}
```

pascalgrimaud:
@raphaelbrugier : I'm sorry, I think I'm too tired :tired_face: (I was probably pointing on older branch, before your PR with schema_version)
I retry on master and everything works fine, so this PR is just for polishing

raphaelbrugier:
ahah this happen! :)
If you're ok you can discard this PR as I'm also doing it on another PR

PR #3585, opened 2016-05-12
Set default profile

deepu105:
This is based on the discussion in #3224 

This sets a default profile properly when no profile is passed as `--spring.profiles.active` or `-Dspring.profiles.active` set in `JAVA_OPTS` or as `-Pprod` in build.

I have tested this from IDE, as executable war, via boot run and deployed in tomcat. This seems to work properly.

I have also set the placeholder in `application.yml` so that it can be replaced by the build tool to set an active profile while packaging and that also works in all the above mentioned methods.

Now additional work needs to be done to do the replacement of the value in the file moved to `target|build` folder during `mvn package` or `gradle bootRepackage | gradle war` But I guess that can be done as another PR.

@erikkemperman since ill be away for few days you can give it a shot if you have time.

@jdubois please review if ok

PR #3587, opened 2016-05-12
Use Maven/Gradle profile(s) as default Spring profile(s)

erikkemperman:
As discussed over in the main repository, do the actual subs from Maven/Gradle.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -693,22 +693,6 @@
     </dependencies>
     <build>
         <defaultGoal>spring-boot:run</defaultGoal>
-        <resources>
-            <resource>
-                <directory><%= SERVER_MAIN_RES_DIR %></directory>
-                <filtering>true</filtering>
-                <includes>
-                    <include>**/*.xml</include>
-                </includes>
-            </resource>
-            <resource>
-                <directory><%= SERVER_MAIN_RES_DIR %></directory>
-                <filtering>false</filtering>
-                <excludes>
-                    <exclude>**/*.xml</exclude>
-                </excludes>
-            </resource>
-        </resources>
         <%_ if (testFrameworks.indexOf('cucumber') != -1) { _%>
         <testResources>
             <testResource>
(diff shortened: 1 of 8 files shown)


erikkemperman:
This adds onto @deepu105 's PR [here](https://github.com/jhipster/generator-jhipster/pull/3585) but I accidentally aimed my PR to the main repo instead of Deepu's branch. Sorry about that!

generators/server/templates/_pom.xml
```diff
@@ -941,9 +963,6 @@
                     Enable the line below to have remote debugging of your application on port 5005
                     <jvmArguments>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005</jvmArguments>
                     -->
-                    <arguments>
-                        <argument>--spring.profiles.active=dev</argument>
```
deepu105: As we discussed dont we need this so that correct profiles are passed for `mvn spring-boot:run`?

generators/server/templates/_pom.xml
```diff
@@ -1154,16 +1175,15 @@
                         <artifactId>spring-boot-maven-plugin</artifactId>
                         <configuration>
                             <executable>true</executable>
-                            <arguments>
-                                <argument>--spring.profiles.active=prod</argument>
```
deepu105: same here

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -21,7 +21,7 @@ if (project.hasProperty('no-cache')) {
 }
 
 bootRun {
-    args = ["--spring.profiles.active=" + profiles]
```
deepu105: similarly this might be needed for `gradlew bootRun -Pprod`,

deepu105:
ill test and merge this next week

generators/server/templates/_pom.xml
erikkemperman: Not when running via Maven/Gradle -- because in those cases the value in application.yml will have been filled in. What we discussed (I thought) was only for 

- overriding the default on an existing war file
- running in IDE (when the unsubstituted value will be read)

I've confirmed these cases to work. Sorry if I misunderstood anything?

erikkemperman:
All right, have great holiday man!

generators/server/templates/_pom.xml
deepu105: Oh ok I guess I didnt think it out properly, in dev profile the default will kick in anyway and in prod profile the resources are served from target folder which will have the value replaced file am I right?

generators/server/templates/src/main/resources/_logback-spring.xml
```diff
@@ -61,7 +61,7 @@
         <resetJUL>true</resetJUL>
     </contextListener>
 
-    <root level="@logback.loglevel@">
+    <root level="#logback.loglevel#">
```
deepu105: we need to confirm, the log dynamic setting from the GUI is not affected by any of this. Technically it shouldn't be affected, but im not sure what kind of voodo spring does here

generators/server/templates/_pom.xml
erikkemperman: I'm not sure what you mean... None of the static assets (which are served from different location depending on profile) contain any placeholders themselves, if that's what you were getting at?

generators/server/templates/src/main/resources/_logback-spring.xml
erikkemperman: The logback config only determines its initial state, the GUI changes this dynamically (but doesn't write back to the config file, obviously. I haven't made a point of testing it explicitly, though.

jdubois:
I'm not sure this should be merged now, so I'm waiting for your go @deepu105 - nothing urgent anyway!

erikkemperman:
Definitely not urgent, I am very happy this is now moving forward -- a great improvement over what I had initially proposed!

I just hope there will be no unexpected consequences of the changes which now make it possible to simply use `application.yml` for this purpose.

generators/server/templates/_pom.xml
deepu105: I checked, every scenario except passing switches `./mvnw  "-Drun.profiles=no-liquibase,no-swagger"` like this works. So im gonna put setting of this argument back in while merging so that the above also works as per what we have in docs

erikkemperman: Hm, I haven't seen that `-Drun.profiles` option before... So I didn't try that, indeed. Good catch.

But according to [this](http://docs.spring.io/spring-boot/docs/current/maven-plugin/examples/run-profiles.html) it might not be exactly equivalent to the `<arguments>` block we were talking about here? Instead maybe there should be a `<profiles>` block?

deepu105:
@erikkemperman have made some changes in the commits mentioned above so that all scenarios are working properly

generators/server/templates/_pom.xml
deepu105: Ya may be we can improve it later. but since this was working I didnt try that

deepu105:
@erikkemperman could you update the documentation as well?

PR #3590, opened 2016-05-13
Stylistic Changes to Password Strength Meter

teh48th:
This is a stylistic change to make the Strength Meter to be more presentable. This is not tied to any open issues.

**Old**

password_strength_old

**New**

password_strength_new

### DIFF
--- a/generators/client/templates/src/main/webapp/app/account/password/_password-strength-bar.directive.js
+++ b/generators/client/templates/src/main/webapp/app/account/password/_password-strength-bar.directive.js
@@ -88,6 +88,10 @@
                         .css({ 'background-color': '#DDD' })
                         .slice(0, c.idx)
                         .css({ 'background-color': c.col });
+                }else {
+                    if (!iElement.hasClass('ng-hide')){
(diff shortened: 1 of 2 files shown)


gtiwari333:
@teh48th The Strength Meter in my app was showing like the one you mentioned as "**Old**" . It turns out somehow main.css was missing from my Index.html page.

deepu105:
This needs to be checked, it was already working like what you show as new.  Ill check next week

generators/client/templates/src/main/webapp/app/account/password/_password-strength-bar.directive.js
```diff
@@ -88,6 +88,10 @@
                         .css({ 'background-color': '#DDD' })
                         .slice(0, c.idx)
                         .css({ 'background-color': c.col });
+                }else {
```
jdubois: formatting is not good, a space misses

jdubois:
Might depend on your browser??
I'm sorry but it already works as expected for me, so I'm closing it.

PR #3592, opened 2016-05-13
PaginationUtil: if baseUrl already contains request-params, no second '?' should be added

tomvanrossom:
I had an issue using the `PaginationUtil`

In my case, I provided a baseUrl already containing request-params: `"/api/reports?filter=true"`

Then the old implementations added `page` and `size` parameters without checking, resulting in:  `"/api/reports?filter=true?page=1&size=20"`

This result contains 2 `?`

This fix uses Spring `UriComponentsBuilder` for handling such cases.

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/util/_PaginationUtil.java
+++ b/generators/server/templates/src/main/java/package/web/rest/util/_PaginationUtil.java
@@ -2,6 +2,7 @@ package <%=packageName%>.web.rest.util;
 
 import org.springframework.data.domain.Page;
 import org.springframework.http.HttpHeaders;
+import org.springframework.web.util.UriComponentsBuilder;
 
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/web/rest/util/_PaginationUtil.java
```diff
@@ -22,22 +23,29 @@ public static HttpHeaders generatePaginationHttpHeaders(Page<?> page, String bas
...
         return headers;
-    }<% if (searchEngine == 'elasticsearch') { %>
+    }
+
+    private static String generateUri(String baseUrl, int page, int size) throws URISyntaxException {
+        return UriComponentsBuilder.fromUriString(baseUrl).queryParam("page", page).queryParam("size", size).toUriString();
+    }
+
+
+    <% if (searchEngine == 'elasticsearch') { %>
```
deepu105: use `<%_ if (searchEngine === 'elasticsearch') { _%>`

jdubois:
I guess this has a small perf overhead, but indeed it's cleaner and will solve this kind of use case.
I've re-launched the Travis build, if the tests pass I will merge this.

pascalgrimaud:
@tomvanrossom : can you rebase with master ? I made some change on our current Travis build and there is something strange: your PR uses the old .travis.yml but new scripts SH I updated

Issue #3427, opened 2016-04-19
cassandra microservice app fails to start with docker-compose generated conf 


raphaelbrugier:
##### **Overview of the issue**
When adding a microservice app with a Cassandra database, the app fails to start when launched with the configuration generated by the docker-compose generator.
The reason is the Cassandra cluster is not yet initialized (no keyspace or tables created) when the app starts. The setup of the cluster has to be done manually.

##### **JHipster Version(s)** 
master

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

3 directories: 
- jhigate
- casstest (cassandra microservice)
- compose (docker-compose config)

**gateway:**
```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.0.0",
    "baseName": "jhigate",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "hazelcast",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": "no",
    "buildTool": "maven",
    "jwtSecretKey": "080bce90a1e6565afd977f09ce38b41dc315b080",
    "useSass": false,
    "applicationType": "gateway",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ]
  }
}
```

**cassandra microservice:**
```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.0.0",
    "baseName": "casstest",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8085",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "databaseType": "cassandra",
    "devDatabaseType": "cassandra",
    "prodDatabaseType": "cassandra",
    "searchEngine": "no",
    "buildTool": "maven",
    "jwtSecretKey": "a310878c074bd02c67adc51c484c420978c039aa",
    "enableTranslation": false,
    "applicationType": "microservice",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true,
    "skipUserManagement": true,
[...]

Issue #3588, opened 2016-05-12
Application could not connect to Cassandra node-x running on non-linux Docker


raphaelbrugier:
##### **Overview of the issue**
When the Cassandra cluster is launched within Docker with multiple nodes, the application could not reach the extra nodes when running on non-linux OS.
This is because Docker is running inside a vagrant VM and those the driver could not access directly to the cluster nodes after reaching the contact point.

Error message:
```
2016-05-12 16:47:54.238 DEBUG 6660 --- [uster1-worker-0] com.datastax.driver.core.Host.STATES     : [/172.18.0.3:9042] starting reconnection attempt
2016-05-12 16:47:54.240  WARN 6660 --- [r1-nio-worker-2] com.datastax.driver.core.Session         : Error creating pool to /172.18.0.3:9042
  com.datastax.driver.core.exceptions.ConnectionException: [/172.18.0.3] Pool was closed during initialization
at com.datastax.driver.core.HostConnectionPool$2.onSuccess(HostConnectionPool.java:149)
```

We started discussing this with @jdubois on this [comment](https://github.com/jhipster/generator-jhipster/pull/3570#issuecomment-218781078).

##### **Browsers and Operating System** 

All but linux

##### **Suggest a Fix** 
The current fix is to add a new route to your iptable pointing to the Vragrant ip as explains in [this post](http://krasserm.github.io/2015/07/13/chaos-testing-with-docker-and-cassandra/#port-mapping):
`sudo route -n add 172.18.0.0/16docker-machine ip default`

This is documented in JHipster [docker-compose page](https://github.com/jhipster/jhipster.github.io/blob/master/pages/docker_compose.md#note-for-non-linux-users) but a bit hidden.

@jdubois proposal:
> My main problem is that everything is done with JHipster so that it works "out of the box" -> isn't there a solution for this? Like forbidding to connect to that node? To be honest I haven't understood why you need it, yet. If there is no other solution, you could also catch the exception and print a specific log message, pointing to the documentation

Possible options:
- prevent the driver to ping the other nodes (I don't know if it's possible and a bit against C* philosophy anyway)
- catch the exception and display a better message pointing to the documentation
- do nothing and hope every non-linux users migrate to [native docker for mac](https://blog.docker.com/2016/03/docker-for-mac-windows-beta/) :)

I think I'll implement something by catching the exception unless someone as a better idea on how to handle this. cc @cbornet @pascalgrimaud

jdubois:
[...]

PR #3593, opened 2016-05-13
Cassandra migration tool

raphaelbrugier:
This PR is a work in progress on a migration tool for Cassandra after #3544
It should close issue #3427.
See also  #3588

## Todos:

- [x] Move the create-tables.cql to the changelog folder
- [x] Delete opscenter option
- [x] delay the launch of the application in app.yml
- [ ] manual tests:
 - [x] monolith
 - [x] gateway
 - [x] microservice app
 - [x] docker-compose generator 
 - [ ] relaunch the migration service after an entity generation, without stopping the running cluster
- [ ] update/add travis tests?
- [ ] update documentation


raphaelbrugier:
My main concern is how to have the application started in one container wait for the migration, launched in another container, to be completed.

I'll implement something blocking the application startup till an exception for missing tables is thrown, but better ideas are welcome!

pascalgrimaud:
@raphaelbrugier : don't worry about that... It's an issue about docker-compose. They tried to add the "depend" in their docker-compose file but it's not enought.

A very simple solution (not the best but the easiest) is to ask the container application to wait XX sec before launching JHipster application, see my PR https://github.com/jhipster/generator-jhipster/pull/3412
In fact, I did it for Cassandra!!! => so you can simply add in docker-compose file JHIPSTER_SLEEP to 30 (for example)

raphaelbrugier:
Ah, awesome! Thx @pascalgrimaud

raphaelbrugier:
If you want to test @pascalgrimaud, I've pushed some commits.

The microservices both gateway and app, should now work out of the box with the docker-compose file generated with jhipster:docker-compose 

The monolith app works out-of the box with the app.yml docker-compose file

pascalgrimaud:
I mode some changes yesterday on Travis Build. That's why it's red.
Can you rebase your PR plz?

Thanks for your work, I will test it!

raphaelbrugier:
Just rebased the branch, let's see!

pascalgrimaud:
@raphaelbrugier : the `npm test` failed
See the logs:
```
  146 passing (43s)
  1 failing

  1) JHipster Docker Compose Sub Generator gateway and 1 microservice, with cassandra cluster "before each" hook for "creates expected default files":
     Error: timeout of 20000ms exceeded. Ensure the done() callback is being called in this test.
  
```

generators/server/templates/src/main/resources/config/cql/_create-keyspace-prod.cql
```diff
@@ -1,3 +1,13 @@
 CREATE KEYSPACE IF NOT EXISTS <%= baseName %>
     WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 3 }
     AND DURABLE_WRITES = true;
+
+CREATE TABLE IF NOT EXISTS <%= baseName %>.schema_version (
```
pascalgrimaud: My Unit test failed for microservice cassandra.
Maybe because there is no cql file, so no table, so no keyspace created.
What about moving this CREATE TABLE into 0000_create_table like before and remove <%= baseName %>. of course?

generators/server/templates/src/main/resources/config/cql/_create-keyspace.cql
```diff
@@ -1,3 +1,13 @@
 CREATE KEYSPACE IF NOT EXISTS <%= baseName %>
     WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }
     AND DURABLE_WRITES = true;
+
+CREATE TABLE IF NOT EXISTS <%= baseName %>.schema_version (
```
pascalgrimaud: same comment

generators/server/templates/src/main/resources/config/cql/changelog/_create-tables.cql
```diff
@@ -1,12 +1,3 @@
-CREATE TABLE IF NOT EXISTS schema_version (
```
pascalgrimaud: same comment: maybe keep here, in this file (but this file has been moved to changelog/)

generators/server/templates/src/main/resources/config/cql/_create-keyspace-prod.cql
pascalgrimaud: oh I understand why you move this table here! We need schema_version before scanning changelog/* and executing all cql files.
So forget my comment => but we need to find something to resolve unit tests

generators/server/index.js
```diff
@@ -1052,13 +1046,13 @@ module.exports = JhipsterServerGenerator.extend({
...
-                this.copy(SERVER_MAIN_RES_DIR + 'config/cql/create-tables.cql', SERVER_MAIN_RES_DIR + 'config/cql/create-tables.cql');
                 this.copy(SERVER_MAIN_RES_DIR + 'config/cql/changelog/README.md', SERVER_MAIN_RES_DIR + 'config/cql/changelog/README.md');
 
                 /* Skip the code below for --skip-user-management */
                 if (this.skipUserManagement) return;
-                if (this.databaseType === 'cassandra' ) {
-                    this.template(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000000_insert_default_users.cql');
+                if (this.applicationType !== 'microservice') {
+                    this.template(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_create-tables.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000000_create-tables.cql', this, {});
+                    this.template(SERVER_MAIN_RES_DIR + 'config/cql/changelog/_insert_default_users.cql', SERVER_MAIN_RES_DIR + 'config/cql/changelog/00000000000001_insert_default_users.cql', this, {});
```
pascalgrimaud: when creating a gateway mysql, there shouldn't be this "insert_default_users", only when database is cassandra

pascalgrimaud:
I made some tests. 2 issues left:
- gateway + mysql database => insert_users
- unit tests failed with microservice cassandra

raphaelbrugier:
Thanks for the tests @pascalgrimaud !
Yes I got trapped by the unit tests with the microservice app.

The keyspace is supposed to be created with the first script run, but since there is no default CQL file for the microservice => No keyspace creation.
And, as you discovered, I couldn't move the schema_version table creation into the create-table.cql script because it has to be existing before iterating on the the changelog scripts.

I'll create the keyspace in its own query.

generators/server/index.js
raphaelbrugier: Good catch.

Btw, I didn't touch the other if statement above because I don't understand the reason:
https://github.com/raphaelbrugier/generator-jhipster/blob/e9a36b319bba58ab94c63aabd7036c5f0eff88e8/generators/server/index.js#L1045

Are the cql files written everytime a gateway is selected??

pascalgrimaud:
@raphaelbrugier : what do you think about creating table schema_version` in :
- create-keyspace.cql
- create-keyspace-prod.cql
- 00000000000000_create-tables.cql

1) for Unit Tests, it will use 00000000000000_create-tables.cql => so the keyspace is created
2) when starting app the 1st time, the create table schema_version will be executed 2 times, but it's not important as the table is already created

raphaelbrugier:
@pascalgrimaud I'd prefer not to duplicate the query in both the keyspace creation and the create-tables script.
Also, creating the cassandra_unit_keyspace explictly allow to remove a workaround on the scripts iteration to create it only on the first script.
See : https://github.com/jhipster/generator-jhipster/pull/3593/commits/4d1c8d0d02d78f1b4e7fff67efe37148276dde83#diff-98c1d01972751ca157ded4c13436d308L50

pascalgrimaud:
You're right, I just tried to suggest something :)
Your solution is better than mine!

raphaelbrugier:
Np! just wanted to point the other problem I ran into (create the keyspace when iterating on the scripts)

raphaelbrugier:
Also, any idea if it is needed and how to add new automated tests?
It's getting complicated to test everything manually.

pascalgrimaud:
I will explain it to you, as soon as I test this PR ;-)

pascalgrimaud:
@raphaelbrugier : it's not the right place to discuss about this, but I'll share you my idea

Travis can't test everything, there's too much combination.
My idea is to build a Docker image, when mounting volume to a `.yo-rc.json`, will scan the json and launch appropriate tests. So I can prepare a lot of `.yo-rc.json`, and just choose what I want to test => it's for local test, not by Travis
I have an alpha version, but at the moment, I don't have time to finish this...

Anyway, thanks so much for your PR and improving this docker-compose with cassandra

raphaelbrugier:
Thanks for the merge @pascalgrimaud! looking forward to see more on your idea.
Now I have to work on the doc, most probably Monday.

PR #3596, opened 2016-05-13
Updater for jhipster applications

### LINKED ISSUE #3594: Updater for jhipster applications

Propose a standard updater when new Jhipster versions are available

Each time a new Jhipster version is published, developpers have to regenerate the whole application from scratch, and then re-apply their changes ; they can forgot some changes, and have to check every file in a manual way

##### **Suggest a Fix** 

I wrote a tiny shell script which avoided me to lose a lot of time ; it's based on Git :
* it switches to a "update_jhipster" branch, which won't ever contain anything but the purely generated code
* it generates the application on this branch
* then upgrades jhipster version
* regenerates the app
* switch back to default branch (which holds modified files due to developments)
* starts the merge with update_jhipster branch

The fact is, every time there is a new jhipster version, merge is easy cause git knows that only some changes are about to be applied

lordlothar99:
Fixed #3594


deepu105:
Just one suggestion. Can we do this as a js sub generator rather than a shell script. I would prefer the javascript sub gen as it will fit the generator well and will work on windows/unix as well. Majority of developers and companies still use windows you see :)
I can help you with sub gen if required. Also this will be lot simpler as a sub gen imo

cbornet:
Maybe add a "git init" at the beginning if needed ?

jdubois:
Yes Deepu is right, it would better as a yeoman sub-generator, probably
with something like http://www.nodegit.org so it is cross platform and
consistent with the rest of the project.
Le 14 mai 2016 10:32 AM, "Christophe Bornet" <notifications@github.com> a
écrit :

lordlothar99:
ok I will have a look at moving to a subgenerator soon ;:)

jdubois:
It shouldn't be very difficult and we'll help you!
Le 14 mai 2016 11:00 AM, "François Lecomte" <notifications@github.com> a
écrit :

lordlothar99:
thx @deepu105 & @jdubois ! I'll make baby steps and push frequently, so u'll be able to get a close look

lordlothar99:
I'm facing a "/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found" issue when trying to use NodeGit 0.13.0... obviously I can fix this on my machine, but I'm not very confident about NodeGit working on every computer :(
do you prefer to keep using nodeGit or simply execute git commands instead ?

jdubois:
There are some Git wrappers for node.js, can you have a look at them?
Considering what we do, we can expect users to have Git correctely
installed.
Le 14 mai 2016 12:08 PM, "François Lecomte" <notifications@github.com> a
écrit :

deepu105:
I have used git wrappers in gulp and they were good. I guees we can use a
wrapper or just use shelljs. Ill help you with this next week

lordlothar99:
ok guys, I just finished a 1st version, I let you have a look at this ;)

jdubois:
Great! Our review might take some time, @deep105 and me are both on holidays.

lordlothar99:
No pb!  I'm in holidays in 1 week...

Le sam. 14 mai 2016 18:41, Julien Dubois <notifications@github.com> a
écrit :

generators/server/index.js
```diff
@@ -902,6 +903,7 @@ module.exports = JhipsterServerGenerator.extend({
             this.copy('gitattributes', '.gitattributes');
             this.copy('editorconfig', '.editorconfig');
             this.template('_travis.yml', '.travis.yml', this, {});
+            this.copy(JHIPSTER_CONFIG_DIR + 'upgrade.sh', JHIPSTER_CONFIG_DIR + 'upgrade.sh');
         },
```
deepu105: So these changes are not required right?

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+                if (err != null) throw new Error(err);
+                this.log('Checked out branch ' + branch);
+                done();
+            }.bind(this));
+        };
+    },
+
+    initializing: {
+        displayLogo: function () {
+            this.printJHipsterLogo();
```
deepu105: I dont think we need to show the full logo here

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+        }
+    },
+
+    configuring: {
+        assertGitPresent: function() {
+            var done = this.async();
+            exec('git --version', function (err) {
+                if (err) {
+                    this.log(chalk.yellow.bold('WARNING!') + ' git is not found on your computer.\n',
+                        ' Install git: ' + chalk.yellow('http://git-scm.com/')
```
deepu105: Use `this.error` method so that generator exists here

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+                    );
+                    return;
+                }
+                done();
+            }.bind(this));
+        },
+
+        assertGitRepository: function() {
+            if (! fs.existsSync('.git')) {
+                this.log(chalk.yellow.bold('WARNING!') + ' no git repository found.\n',
```
deepu105: Use `this.error` method

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+                return;
+            }
+        },
+
+        assertNoLocalChanges: function() {
+            var done = this.async();
+            this.git.exec('status', {}, ['--porcelain'], function(err, msg) {
+                if (err != null) throw new Error(err);
+                if (msg != null && msg !== '') {
+                    this.log(chalk.yellow.bold('WARNING!') + ' local changes found.\n',
```
deepu105: Use `this.error` method

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+
+        checkLatestVersion: function() {
+            this.log('Looking for latest ' + GENERATOR_JHIPSTER + ' version...');
+            var done = this.async();
+            shelljs.exec('npm show ' + GENERATOR_JHIPSTER + ' version', {silent:true}, function (code, stdout, stderr) {
+                this.latestVersion = stdout.replace('\n','');
+                if (this.force || semver.lt(this.currentVersion, this.latestVersion)) {
+                    this.log(chalk.green('New ' + GENERATOR_JHIPSTER + ' version found: ' + this.latestVersion));
+                } else {
+                    this.log(chalk.green('No update available'));
```
deepu105: Use `this.error` method. Also the message should say that we are existing as the application is already generated using latest jhipster

generators/upgrade/index.js
deepu105: Or even better we can do a git init here and continue after logging a message

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+    default: {
+        insight: function () {
+            var insight = this.insight();
+            insight.trackWithEvent('generator', 'upgrade');
+        },
+
+        updateJhipster: function() {
+            this.log('Updating ' + GENERATOR_JHIPSTER + '...');
+            var done = this.async();
+            shelljs.exec('npm uninstall -g ' + GENERATOR_JHIPSTER, {silent:true}, function (code, stdout, stderr) {
```
deepu105: Do we need to uninstall? Npm install shoud upgrade just fine

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+            }.bind(this));
+            shelljs.exec('npm install -g ' + GENERATOR_JHIPSTER, {silent:true}, function (code, stdout, stderr) {
+                done();
+            }.bind(this));
+            this.log(chalk.green('Updated ' + GENERATOR_JHIPSTER + ' to version ' + this.latestVersion));
+        },
```
deepu105: This should be printed in the shelljs exec success

generators/upgrade/index.js
```diff
@@ -0,0 +1,180 @@
...
+
+        cleanUp: function() {
+            var done = this.async();
+            this.git.exec('rm', {q: true}, ['-rf', '--ignore-unmatch', '*'], function(err, msg) {
+                if (err != null) throw new Error(err);
+            }.bind(this));
+            shelljs.exec('rm -rf *', {silent:true}, function (code, stdout, stderr) {
```
deepu105: Woulnt this remove .git, node_modules, .yo-rc.json etc as well, not sure about osx but it removes in windows, may be we should specify folders, files to remove

generators/upgrade/index.js
lordlothar99: yes u r right ; the fact is I quite remember I had trouble with simple `npm update` a while ago... maybe that was just me ; let's use `update` here

generators/upgrade/index.js
lordlothar99: yes u r right ; in my first script I used to remove everything but `.*` files ; obviously we only need to retain `.yo-rc.json` file, so let's code it explicitely ; I used to remove `node_modules` by the way, what do u think about that ?

generators/upgrade/index.js
lordlothar99: I had to create a new `error` function as I was unable to find one... :/
I put it in generator-base, hope this will fit requirements

lordlothar99:
@deepu105 I think I fixed everything :)

generators/upgrade/index.js
deepu105: Just use `npm install -g` here not updtae. Install will automatically update to latest

generators/upgrade/index.js
deepu105: Yeoman has an inbuit error method. Could you check our sub generator where its used. Eg the entity generator when validations fail

generators/generator-base.js
```diff
@@ -1415,4 +1415,12 @@ Generator.prototype.filterNumber = function (input, isSigned, isDecimal) {
     return NaN;
 };
 
+Generator.prototype.error = function(msg) {
+    return this.env.error(chalk.red.bold('ERROR! ') + msg);
+};
+
+Generator.prototype.warning = function(msg) {
+    return this.env.error(chalk.yellow.bold('WARNING! ') + msg);
```
deepu105: This would exit the generator. Do we want to exit for warnings?  Imo we should use this.log for warning and only for error we should exit

generators/upgrade/index.js
```diff
@@ -0,0 +1,208 @@
...
+                }.bind(this));
+            }.bind(this);
+            this.git.exec('add', {}, ['-A'], function(err, msg) {
+                if (err != null) return this.error('Unable to add resources in git:\n' + err);
+                commit();
+            }.bind(this));
+        };
+
+        // This adds support for a `--force` flag
+        this.option('force', {
```
deepu105: I think --force is already available in yeoman so need to define again here. You can use it directly using `this.options['force']`

deepu105:
Im gonna make some changes and merge this

generators/server/index.js
deepu105: @lordlothar99 can you remove the unwanted files and changes related to the sh file?

deepu105:
I did a PR to your branch @lordlothar99 can you take a look and test?

deepu105:
I have did another PR to your branch can you merge that. I have replaced the git wrapper with a custom util method using shellJS as the git wrapper project doesnt seem very active and shellJS gives us more control

lordlothar99:
Sure. I merged it, I will test the whole thing today

Le mer. 18 mai 2016 07:06, Deepu K Sasidharan <notifications@github.com> a
écrit :

deepu105:
Ok ill merge this once you confirm in unix. Its all ok in windows

lordlothar99:
OK thx !  Will do this evening

Le mer. 18 mai 2016 10:34, Deepu K Sasidharan <notifications@github.com> a
écrit :

lordlothar99:
all tests ok !! ready for merge !

deepu105:
@lordlothar99 could you kindly add some documentation for this in our site https://github.com/jhipster/jhipster.github.io

deepu105:
I think it should be a sub menu under `Core Jhipster Tasks` menu. Somthing like `Upgrading an application`. what do you think @jdubois

lordlothar99:
OK for the documentation, I ll try to do that today before holidays

jdubois:
Yes this needs its own documentation page.
BTW use Markdown, not HTML. When I have time I will migrate everything to Markdown, it's so much easier.

lordlothar99:
guys, I'm sorry I won't have time to write the documentation in the next 2 weeks.. I wrote (very) few usage instructions in USAGE file ; I don't know what you prefer : either someone writes the doc, either I do it in 2 weeks ?

jdubois:
thanks @lordlothar99 - there's no rush!! It's already great to have this merged! We now need some time to get used to it, and then document it.

lordlothar99:
hi guys! I just saw u already wrote the doc ! thanks !

Issue #3579, opened 2016-05-11
Microservice Cassandra : status Down


pascalgrimaud:
##### **Overview of the issue**

When generating a microservice with Cassandra database, the microservice starts but his status is down

##### **JHipster Version(s)** 

Current master (and probably since 3.0.0 but I'm not sure)

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

```yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "3.2.1",
    "baseName": "mscassandra",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8081",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "databaseType": "cassandra",
    "devDatabaseType": "cassandra",
    "prodDatabaseType": "cassandra",
    "searchEngine": "no",
    "buildTool": "maven",
    "jwtSecretKey": "fc9dbd0c2435ef0b0b08ba229829b84e3e6757fe",
    "enableTranslation": true,
    "applicationType": "microservice",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true,
    "skipUserManagement": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "fr"
    ],
    "clusteredHttpSession": "no",
    "websocket": "no",
    "enableSocialSignIn": false
  }
}
```

##### **Reproduce the error** 

1) generate a microservice project with Database (see .yo-rc.json below)
2) start registry
3) start database cassandra
4) start microservice

![mscassandra](https://cloud.githubusercontent.com/assets/9156882/15182511/2f79854a-178d-11e6-8491-bb74a38f3ab8.png)

Here the log
```log
2016-05-11 15:11:53.853  INFO 9326 --- [           main] com.mycompany.myapp.MscassandraApp       : The following profiles are active: prod
2016-05-11 15:11:57.806  INFO 9326 --- [ost-startStop-1] c.mycompany.myapp.config.WebConfigurer   : Web application configuration, using profiles: [prod]
2016-05-11 15:11:57.813  INFO 9326 --- [ost-startStop-1] c.mycompany.myapp.config.WebConfigurer   : Web application fully configured
2016-05-11 15:11:58.200  INFO 9326 --- [ost-startStop-1] com.mycompany.myapp.MscassandraApp       : Running with Spring profile(s) : [prod]
[...]

PR #3601, opened 2016-05-16
Mail configuration

pascalgrimaud:
- enable health mail for **monolith** application - related to https://github.com/jhipster/generator-jhipster/issues/3579
- add empty properties for smtp configuration

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-dev.yml
+++ b/generators/server/templates/src/main/resources/config/_application-dev.yml
@@ -129,12 +129,6 @@ spring:
         port: 25
         username:
         password:
-        protocol: smtp
-        # tls:
-        # properties.mail.smtp:
-            # auth:
-            # starttls.enable:
-            # ssl.trust:
     messages:
         cache-seconds: 1
     thymeleaf:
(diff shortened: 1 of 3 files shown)


jdubois:
I'm not sure about having the healthcheck option different with monoliths and microservices/gateways -> I'd rather have something consistent here

pascalgrimaud:
Ok let's be consistent about `health.mail.enabled: false`
Do I keep the empty configuration for smtp configuration ?

jdubois:
Good question.
I wanted to have it, but now I remember that we always use each technology's "normal" configuration, and then point to each technology's documentation. So basically, this should be a Spring Boot issue, not a JHipster issue.
If we do this for each configuration, then we will end up with the full default Spring Boot configuration file, which is of course the opposite to what Spring Boot is trying to achieve (convention over configuration). So I know many people struggle with this, but I don't think this is a good idea. What do you think?

pascalgrimaud:
You're right, it's well documented in our site and in official spring boot docs, so no need to put configuration in comments.

So this PR is to fix the indentation and change the `application-dev.yml` to add `port: 25`, to be consistent with `application-prod.yml`
=> I can start easily a SMTP server with `docker run -d -p 1080:80 -p 25:25 -t djfarrelly/maildev`

PR #3612, opened 2016-05-18
move prompts to separate files

### LINKED ISSUE #3577: [Internal] externalize the prompts into a file
Currently our promts for server and entity are so long and complex, it would be very easy to maintain if they are moved to its own files within the respective generators.

I propose to move the prompts into a `prompts.js` file along side the `index.js` of each generator and then just import it int the `prompting` section of `index.js`

@jhipster/developers what do you think? if you guys have no concerns, then ill do it when i have free time

deepu105:
fix #3577 

will merge once travis passes.


pascalgrimaud:
@deepu105 : there is a problem with the last yeoman 1.8.2, so our Travis build will fail, until it is resolved.
Ping @wmarques too, for your PR

We can force the build, by using the version 1.8.1 instead https://github.com/jhipster/generator-jhipster/blob/master/travis/install/02-install-jhipster-stack.sh#L10
Or simply wait...

wmarques:
@pascalgrimaud I updated yeoman on my computer and it asks if we want to install an autocompletion plugin (for zsh) after the installation. I think that's why Travis is crashing

pascalgrimaud:
I didn't test the update on my machine. I don't want to break my install :smile:
I tried to install inside an empty Ubuntu container:
- node 4.4.4
- update npm to 3.9.0
- install yeoman => fail

wmarques:
@pascalgrimaud https://github.com/yeoman/yo/releases seems that autocompletion is breaking things :(

pascalgrimaud:
In my opinion, the best thing is to use an older version of node, until it is resolved or we find another way to install yeoman by skipping the autocompletion.
If we keep like this too long, every PR to our project can't be tested

wmarques:
Agree with you, so we use the 1.8.1 ?

deepu105:
may be we should do `npm i -g yo@1.8.1` for now

deepu105:
Ill push a temp fix

deepu105:
done here https://github.com/jhipster/generator-jhipster/commit/392bde89d665a8dd81c4220631f673ecf63b6460

Issue #3618, opened 2016-05-19
Wrong URL to access the Registry at startup


jdubois:
When I start my gateway and/or microservices, using the current master branch, I get this first line:

    Could not locate PropertySource: I/O error on GET request for "http://localhost:8888/gateway/default": Connection refused; nested exception is java.net.ConnectException: Connection refused

The registry is running on port 8761, not 8888, and this is normally correctly configured in the application-*.yml files, so I don't know where this port number comes from, but it's wrong.

gmarziou:
Might be different but I usually get this when I use a profile that does not have any matching bootstrap-*.yml so registry url is not specified.

jdubois:
Maybe it is because of the latest changes by @deepu105 -> maybe we don't have the dev profile anymore at boot time, and it is added afterwards.

russomi:
I am seeing this on my side, too.

I am trying to run inside of Idea with a spring boot run configuration and my service cannot connect to the repo.  

If I run the service via `mvn` from the command line, no error and I see the Config Server: 	Connected to the JHipster Registry config server! message.

i thought it could be related to the working directory specified in the run configuration, but that doesn't seem to solve it.

This one is annoying me.

russomi:
@jdubois I believe you are headed in the right direction... I was able to resolve it by expliciting activating the 'dev' profile when running the service rather than allowing it to be defaulted.

deepu105:
I need to take a look. Maybe the default is not used at boot

deepu105:
@jdubois Imtrying to debug this, could you tell me how you started the registry and microservice? using docker or just using `mvn spring-boot:run` which profile for microservice?

jdubois:
@deepu105 on the current master:

- create a gateway (probably the same with a microservice)
- start a registry
- start the gateway with `./mvnw` -> it works!!!
- start the gateway with IDEA -> you'll have the issue

Now let me describe precisely so I'm sure you could reproduce:

- in your gateway project, run "idea ." to open up IDEA and open up the project
- the "GatewayApp" main class should be automatically detected as the Main class of the project
- Click "Run" on that class

deepu105:
Ok got it. But one more doubt. Before the change I did was this working with IDE? coz I thought we always had to set `--spring.profiles.active=prod` explicitly in IDE as the `-Pprod` was only in the gradle/maven build? and the dev profile was default anyway.

Btw which profile you used for the above?

[...]

PR #3620, opened 2016-05-19
Uniform profile handling maven/gradle

erikkemperman:
As discussed [here](https://github.com/jhipster/generator-jhipster/commit/d3133965df11716c3271f674d79c286480864cd1#commitcomment-17518049), this makes the handling of profiles (more) uniform across Maven and Gradle builds.

PS @deepu105 You [asked](https://github.com/jhipster/generator-jhipster/pull/3587#issuecomment-219926663) me to update the documentation -- I'll gladly do that, of course, but perhaps let's first deal with this PR, so I don't have to update the docs twice?

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -115,6 +115,9 @@
         <sonar.sources>${project.basedir}/<%= MAIN_DIR %></sonar.sources>
         <sonar.surefire.reportsPath>${project.testresult.directory}/surefire-reports</sonar.surefire.reportsPath>
         <sonar.tests>${project.basedir}/<%= TEST_DIR %></sonar.tests>
+        <!-- These remain empty unless the corresponding profile is active -->
+        <profile.no-liquibase></profile.no-liquibase>
+        <profile.no-swagger></profile.no-swagger>
     </properties>
 
     <%_ if (applicationType === 'microservice' || applicationType === 'gateway' || applicationType === 'uaa') { _%>
@@ -974,9 +977,6 @@
                     Enable the line below to have remote debugging of your application on port 5005
                     <jvmArguments>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005</jvmArguments>
                     -->
-                    <arguments>
-                        <argument>--spring.profiles.active=dev</argument>
-                    </arguments>
                 </configuration>
             </plugin>
             <plugin>
@@ -1058,6 +1058,18 @@
 
     <profiles>
         <profile>
+            <id>no-liquibase</id>
+            <properties>
+                <profile.no-liquibase>,no-liquibase</profile.no-liquibase>
+            </properties>
+        </profile>
+        <profile>
+            <id>no-swagger</id>
+            <properties>
+                <profile.no-swagger>,no-swagger</profile.no-swagger>
+            </properties>
+        </profile>
+        <profile>
             <id>dev</id>
             <activation>
                 <activeByDefault>true</activeByDefault>
@@ -1089,7 +1101,7 @@
                 <!-- log configuration -->
                 <logback.loglevel>DEBUG</logback.loglevel>
                 <!-- default Spring profiles -->
-                <spring.profiles.active>dev</spring.profiles.active>
+                <spring.profiles.active>dev${profile.no-liquibase}${profile.no-swagger}</spring.profiles.active>
             </properties>
         </profile>
         <profile>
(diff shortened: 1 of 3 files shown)


generators/server/templates/_pom.xml
```diff
@@ -1200,7 +1209,7 @@
                 <!-- log configuration -->
                 <logback.loglevel>INFO</logback.loglevel>
                 <!-- default Spring profiles -->
-                <spring.profiles.active>prod</spring.profiles.active>
+                <spring.profiles.active>prod${profile.no-liquibase}</spring.profiles.active>
```
deepu105: please add `${profile.no-swagger}` as well coz, people can enable swagger for production in application.yml and then can choose to disable temporarily using the profile switch

erikkemperman: Ah! Good call! I read in the docs that no-swagger has no effect in prod, so I excluded it here. But now I see that gradle also includes it for prod.

I will add it when I get to campus. When I adapt the documentation, should I mention this option there as well?

deepu105: Yes, you can Im currently testing your PR

Thanks & Regards,
Deepu

generators/server/templates/_pom.xml
```diff
@@ -974,9 +977,6 @@
                     Enable the line below to have remote debugging of your application on port 5005
                     <jvmArguments>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005</jvmArguments>
                     -->
-                    <arguments>
```
deepu105: This seems to be still required(atleast for prod profile). As there are weird issues without this during bootrun. Seems hikariCP is not loading prod profile when this is not set when using `-Pprod`. SInce having this doesnt cause any harm its better we leave it as it is and may be change it to `--spring.profiles.active=dev${profile.no-liquibase}${profile.no-swagger}`. same goes for prod profile and Gradle build.

deepu105: To test the issue I said above generate an app with H2 as dev DB and mysql as prod DB. Dont start mySQL so that prod should fail. Now run `gradlew -Pprod` or maven equivalent and app starts up and log says prod profile but DB seems to work somehow. and ofcourse if you check there is a H2 DB running. Now add `--spring.profiles.active=prod` as argument to maven/gradle build and try again to boot prod profile. App fails to start as mySql is not running.

deepu105: Seems like same behavior when we package the war as well using prod profile.SO the issue might be there even without this PR for war files. so somehow the profile set in `application.yml` is not used 100%, atleast its not used for loading DB, might be related to https://github.com/jhipster/generator-jhipster/issues/3618

erikkemperman:
@deepu105 I just added the no-swagger switch to the prod profile. I'll take a look at the other issue you linked, but at a glance it looks like a more general problem, indeed, not related directly to this Pr.

generators/server/templates/_pom.xml
erikkemperman: @deepu105 I am finding the same thing. I didn't notice earlier, because we use H2 for production as well.

You are right that adding 
```xml
<arguments>
    <argument>spring.profiles.active=${spring.profiles.active}</argument>
</arguments>
```
to the spring-boot plugin configuration "works". But it seems to me that this is just suppressing the symptom of a deeper problem. 

It looks to me like this is a chicken/egg situation around the order in which:

-- Spring looks to VM arguments for explicit profiles set to active
-- Database config looks for properties from active profiles
-- Spring looks at application.yml for active profiles

I saw similar weirdness when working on the original default profile stuff, which was why my original effort ended up using a static initializer block (in BuildProperties.java) to make sure that the default profile was set before anything else happened.

I am curious if this is because the DB configuration uses annotations like `@ConfigurationProperties` and `@Profile`. Would it maybe help to just check for profile dynamically instead?

deepu105: I found the root cause for this. The problem is that when we set a `
spring.profiles.default` spring boot loads up `application.yml` and
`application.dev.yml`, when we set `spring.profiles.active`to prod spring
boot loads up `application.yml`, `application.prod.yml` and
`application.dev.yml` and hence some of the stuff like DB ends up reading
from the dev properties.

I'm trying to find a fix using a listner with
ApplicationListener<ApplicationEvent>

Ill post my findings

Thanks & Regards,
Deepu

jdubois:
It looks like this isn't ready to merge so I'm a adding a specific tag for that

deepu105:
This can be merged after we merge https://github.com/jhipster/generator-jhipster/pull/3624

erikkemperman:
Thanks for review && merge, I will update docs and make a PR for that as well.

deepu105:
Thank you

Thanks & Regards,
Deepu

PR #3622, opened 2016-05-19
Move authority data from entity sub-state to parent entity state

wmarques:
As [child states inherits `data` from parent state](https://github.com/angular-ui/ui-router/wiki/Nested-States-%26-Nested-Views#inherited-custom-data), it's useless to declare the `data` object in each entities substates.
We just need to declare this object in the abstract `entity` state.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/entities/_entity.state.js
+++ b/generators/client/templates/src/main/webapp/app/entities/_entity.state.js
@@ -10,7 +10,10 @@
     function stateConfig($stateProvider) {
         $stateProvider.state('entity', {
             abstract: true,
-            parent: 'app'
+            parent: 'app',
+            data: {
+                authorities: ['ROLE_USER']
+            }
         });
     }
 })();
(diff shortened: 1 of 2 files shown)


deepu105:
If people want to have some fine grained control then it would not be easier if we do this so im not very sure about this.

wmarques:
Agree with that, maybe we can add comment on the sub-entity state to tell the user that he can override the authorities in each state ?

deepu105:
@wmarques in that case can we leave `authorities: []` as an empty array in the child states? so that user can fill required role if they want

ruddell:
@deepu105 if the authorities array is empty in the child state, it will open the state up to anyone (the child overrides the parent authorities).

deepu105:
Ok i didn't actually try it out it was just a guess

wmarques:
@deepu105 yep putting another array erase the array of the parent :(

deepu105:
Ya I didnt think enough

Thanks & Regards,
Deepu

wmarques:
Let's keep it the way it was, it's more clear for the user :)

PR #3624, opened 2016-05-20
workaround for default profile issue

deepu105:
Ok so I tried many ways to set a default profile in a clean way and seems like there might be a spring boot bug which is causing an issue see https://github.com/spring-projects/spring-boot/issues/5998

So for the time being I have done a workaround using a Util file, its a bit similar to what @erikkemperman was doing earlier

@jdubois are you ok with this? If spring fixes the above mentioned issue then we can remove the Util file.

Ill wait for spring guys to evaluate the issue and if its a bug ill see if I can a do a PR to fix it.

fix https://github.com/jhipster/generator-jhipster/issues/3618

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1252,7 +1252,7 @@ Generator.prototype.printJHipsterLogo = function () {
     );
     this.log(chalk.white.bold('                            http://jhipster.github.io\n'));
     if (this.checkInstall) this.checkForNewVersion();
-    this.log(chalk.white('Welcome to the JHipster Generator test') + chalk.yellow('v' + packagejs.version));
+    this.log(chalk.white('Welcome to the JHipster Generator ') + chalk.yellow('v' + packagejs.version));
     this.log(chalk.white('Application files will be generated in folder: ' + chalk.yellow(process.cwd())));
 };
 
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1252,7 +1252,7 @@ Generator.prototype.printJHipsterLogo = function () {
     );
     this.log(chalk.white.bold('                            http://jhipster.github.io\n'));
     if (this.checkInstall) this.checkForNewVersion();
-    this.log(chalk.white('Welcome to the JHipster Generator ') + chalk.yellow('v' + packagejs.version));
+    this.log(chalk.white('Welcome to the JHipster Generator test') + chalk.yellow('v' + packagejs.version));
     this.log(chalk.white('Application files will be generated in folder: ' + chalk.yellow(process.cwd())));
 };
 
--- a/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
+++ b/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
@@ -25,7 +25,7 @@ public final class DefaultProfileUtil {
     private static final Properties BUILD_PROPERTIES = readProperties();
 
     /**
-     * get a default profile from application.yml
+     * Get a default profile from <code>application.yml</code>.
      */
     public static String getDefaultActiveProfiles(){
         if (BUILD_PROPERTIES != null) {
@@ -41,19 +41,19 @@ public final class DefaultProfileUtil {
     }
 
     /**
-     * set a default to use when no profile is configured.
+     * Set a default to use when no profile is configured.
      */
     public static void addDefaultProfile(SpringApplication app) {
         /*
         * The default profile to use when no other profiles are defined
-        * This cannot be set in the `application.yml` file.
+        * This cannot be set in the <code>application.yml</code> file.
         * See https://github.com/spring-projects/spring-boot/issues/1219
         */
         app.setDefaultProperties(Collections.singletonMap(SPRING_PROFILE_ACTIVE, getDefaultActiveProfiles()));
     }
 
     /**
-     * load application.yml from classpath
+     * Load application.yml from classpath.
      */
     private static Properties readProperties() {
         try {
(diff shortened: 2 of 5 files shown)


erikkemperman:
@deepu105 ... And we're back to a helper class. This whole thing is like deja vu all over again :-)

Well, at least there's no extra hidden file. I'm very curious how the guys at Spring-Boot will classify your issue, but in the mean time I agree this is probably better than the current behaviour of master.

deepu105:
@jdubois could you plz merge if you are ok with this. This should fix https://github.com/jhipster/generator-jhipster/issues/3618 as well

jdubois:
Let me fist get back home from Spring IO :-)

deepu105:
Oh ok, hope you had a great time :)

Thanks & Regards,
Deepu

deepu105:
Spring boot guys acknowledged this as a bug so may be we can merge this as a temporaray solution and can remove it when the issue is fixed in spring boot

generators/generator-base.js
```diff
@@ -1252,7 +1252,7 @@ Generator.prototype.printJHipsterLogo = function () {
     );
     this.log(chalk.white.bold('                            http://jhipster.github.io\n'));
     if (this.checkInstall) this.checkForNewVersion();
-    this.log(chalk.white('Welcome to the JHipster Generator ') + chalk.yellow('v' + packagejs.version));
+    this.log(chalk.white('Welcome to the JHipster Generator test') + chalk.yellow('v' + packagejs.version));
```
jdubois: why is this a "test"?

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
+import org.springframework.boot.SpringApplication;
+import org.springframework.core.io.ClassPathResource;
+
+import java.util.Collections;
+import java.util.Properties;
+
+/**
+ * Utility class to load a spring profile to be used as default
```
jdubois: spring -> Spring

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger log = LoggerFactory.getLogger(DefaultProfileUtil.class);
+
+    private static final String SPRING_PROFILE_ACTIVE = "spring.profiles.active";
+
+    private static final Properties BUILD_PROPERTIES = readProperties();
+
+    /**
+     * get a default profile from application.yml
```
jdubois: get -> Get

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -0,0 +1,68 @@
...
+                    activeProfile.contains(Constants.SPRING_PROFILE_PRODUCTION))) {
+                return activeProfile;
+            }
+        }
+        log.warn("No Spring profile configured, running with default profile: {}", Constants.SPRING_PROFILE_DEVELOPMENT);
+        return Constants.SPRING_PROFILE_DEVELOPMENT;
+    }
+
+    /**
+     * set a default to use when no profile is configured.
```
jdubois: set -> Set

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -0,0 +1,68 @@
...
+        /*
+        * The default profile to use when no other profiles are defined
+        * This cannot be set in the `application.yml` file.
+        * See https://github.com/spring-projects/spring-boot/issues/1219
+        */
+        app.setDefaultProperties(Collections.singletonMap(SPRING_PROFILE_ACTIVE, getDefaultActiveProfiles()));
+    }
+
+    /**
+     * load application.yml from classpath
```
jdubois: load -> Load , and you need a "." at the end or it's not valid javadoc

generators/generator-base.js
deepu105: Oops, my bad, I forgot to change that after debugging. Ill correct it asap

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
deepu105: Ok

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
deepu105: Ok missed it here :D

deepu105:
updated all comments

erikkemperman:
@deepu105 I was reading the Spring-Boot guys' response on your issue there, and thought they were suggesting that `spring.profiles.default` ought to work. But you had tried that, obviously, it's in the sample code you gave them. Am I missing something? Where did they say this is a bug on their part?

Just curious. The main thing is it gets resolved somehow, for the moment I guess your workaround is fine.

deepu105:
I meant the original issue with setting active snicoll kind of said thats
its working weird and they would discuss about fixing it so bug or no bug
ill be happy to have it working

erikkemperman:
Agreed! Still, he was pretty sure that `spring.profiles.default` should have worked, right? I was wondering if there's anything specific about JH's DatabaseConfiguration class -- you mentioned seeing this mostly from the wrong database being chosen -- because the profile voodoo there is done with annotations, which I'm not sure during which phase of initialization these are evaluated.

deepu105:
I dont think there is an issue with annotation, which is to load from
external property file. I tried after removing the annotation but same
behaviour

Thanks & Regards,
Deepu

erikkemperman:
All right, it was just a guess...

I just rebased my project on the output from latest JH master, and it all works great!

PR #3630, opened 2016-05-21
hide password reset error messages on form submission

ruddell:
The form error messages should be cleared when submitting the password reset form.  

Fixes: If you enter non-matching passwords then submit the form correctly, the error messages remains displayed under the success message.


pascalgrimaud:
@Ruddell: there is an issue again with yeoman, that s why the build failed
I m investigating and will probably revert to a stable version

pascalgrimaud:
It's done here, with yeoman@1.8.1 => https://github.com/jhipster/generator-jhipster/pull/3631
Can you rebase plz ?

ruddell:
Thanks for fixing the build.  I rebased and it looks like it's passing.

PR #3643, opened 2016-05-25
update bower dependencies

deepu105:


atomfrede:
Looks good, guess there should be no breaking changes.

deepu105:
There is no major version changes So hoping there wont be any issue.

jdubois:
I'm merging this, let's test :-)

PR #3648, opened 2016-05-26
fix showing wrong pages of component jhi-item-count

wongdao:
Issue: the component jhiItemCount(jhi-item-count.directive.js) will shows the wrong range when the number of entity exceeds to 20, for example:

just plus 1 to fix it.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/components/util/_jhi-item-count.directive.js
+++ b/generators/client/templates/src/main/webapp/app/components/util/_jhi-item-count.directive.js
@@ -3,7 +3,7 @@
 
     var jhiItemCount = {
         template: '<div class="info">' +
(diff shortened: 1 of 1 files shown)


deepu105:
@wongdao what was the issue? could you put a description on the PR plz?

wongdao:
@deepu105 the description was added, please reviewing.

jdubois:
Thanks, looks good, I'm merging this!

aquaglia:
The code contains the magic number 20. If you change the page size in pagination.constants.js it does not work any more. Shall I open another bug?

wongdao:
@aquaglia please check this commit [remove hard code in component jhi-item-count #3655](https://github.com/jhipster/generator-jhipster/pull/3655), in which the magic number 20 has already got removed.

aquaglia:
@wongdao Many thanks.

PR #3650, opened 2016-05-26
By default an entity extends AbstractAuditingEntity

mleneveut:
Hi,

I think it could be great that each generated entity extends by default the AbstractAuditingEntity, to gain AOP benefits of automatic createdBy, createdDate , lastModifiedBy and lastModifiedDate fields.


cbornet:
There is already a [module](http://jhipster.github.io/modules/marketplace/#/details/generator-jhipster-entity-audit) for this.

deepu105:
Sorry it wouldnt work just by doing this. you need to have corresponding columns in your DB as well. It was already discussed and we didnt want to do it here thats why the entity-auditing module was created. I suggest you take a look at that. It does what you want and more if needed.

Thanks for the PR but sorry that we cannot merge it.

mleneveut:
Great module, didn't know it as I only used v2 of Jhipster.

Regards.

PR #3657, opened 2016-05-27
fix infinite-scroll template (sent multiple requests on page load)

aBnormaLz:
If i enabled infinite scroll on a page, the angular sent lots and lots of requests to the server because vm.links.last was not declared


jdubois:
Good catch!!

PR #3662, opened 2016-05-29
enables support for feign clients using OAuth2 client credentials grant

xetys:
so feign clients can be added declarative to. This is a requirement for the next step of a JHipster feign client subgenerator.

## contents of this PR

* adds dependencies to microservice applications
* extends JHipster properties to configure client credentials for internal communications
* replaces hardcoded internal client with above configuration in UAA
* configures a OAuth2ProtectedRessourceDetails based on new configuration and declares a OAuth2RequestInterceptor bean
* enables with ```@EnableFeignClients``` the usage of feign clients in microservice applications

so far this gets merged, I am going to proceed with the feign client generators


PierreBesson:
I think this is the right approach for inter-service communication. :+1: @xetys.
However shouldn't this be also enabled for gateway apps ?

xetys:
after hours of total frustration and realizing, there is no offical support to loadbalance the access token uri before acquiring access tokens....here comes an (maybe not fully elegant update), which solves the problem in its root...

the last 2 commits from this comment contain a workaround, so the URL is balanced with eureka and ribbon everytime something is asking for the url. This should only happen, when FeignClients are used.

from this point this PR is complete for me, please review as soon as possible, so I can proceed with a feign-client generator

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,57 @@
...
+            String newUrl;
+            try {
+                newUrl = loadBalancerClient.reconstructURI(
+                    loadBalancerClient.choose(serviceName),
+                    new URI(super.getAccessTokenUri())
+                ).toString();
+
+                return newUrl;
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
```
PierreBesson: Please use a logger call here instead of printStackTrace.

PierreBesson:
I think this PR can be merged after someone else than me have reviewed it. It concerns only microservices with UAA.
@xetys, I would appreciate if you could upload a sample project, where we can test secured Feign calls, circuit breakers, load balancing, etc...

deepu105:
A bit busy right now. Ill take a look as soon as I have time

Thanks & Regards,
Deepu

xetys:
same here

@PierreBesson basicly you can checkout https://github.com/xetys/jhipster-uaa-setup/commits/master
its not clean yet and is containing some experiments in app1, app2 is purely generated using my branch..

the failing build i will checkout as soon as i find time

xetys:
Ok now I found some time to provide slightly more information

This PR allows you to do the following out of the box:
In a example micrroservice with services "app1, serving /api/foos" and "app2, serving /api/baars" you want app2 to have a client for resolving "foos" from app1.

During configuration, you decided to give the UAA the name "uaa" and port 9999 (so you can keep the default configuration unchanged)

in app2 you add a Foo.class 

``` java 

public class Foo {
    private long id;

    private String value;

    public String getValue() {
        return value;
    }

    public void setValue(String value) {
        this.value = value;
    }

    public long getId() {
        return id;
    }

    public void setId(long id) {
        this.id = id;
    }
}
```

and FooClient.java 

``` java 

@FeignClient("http://app1")
public interface FooClient {
    @RequestMapping("/api/foos")
    List<Foo> getFoos();
}
```

to enable the client. You may use it to serve in an app2 ```RestController``` for testing purpose like this:

``` java

@RestController
@RequestMapping("/api/clients/foo")
public class FooClientResource {

    private FooClient fooClient;

    @Autowired
    public FooClientResource(FooClient fooClient) {
        this.fooClient = fooClient;
    }

    @RequestMapping(method = RequestMethod.GET)
    public ResponseEntity<List<Foo>> getFoos() {
        return ResponseEntity.ok(fooClient.getFoos());
    }
}
```

app2 will now consume foos from app1, with an seperate "session" for OAuth2 client internal, rather then the users session. Additionally both "app1" and "uaa" are load balanced using ribbon and eureka client. 

Note, that there is no official solution how to make spring-cloud-feign resovle the UAA via ribbon, so this PR provides a workaround (approved by Dave Syer :smile: ), to let feign also resolve your JHipster UAA (or any other). To apply this configuration in detail, you modify ```application.yml``` (or -dev / -prod)

``` yml

jhipster:
       # ...
        clientAuthorization:
            tokenUrl: http://MyUaa:9999/oauth/token #your custom uaa
            tokenServiceId: MyUaa # eurekas service name
            clientId: internal
            clientSecret: internal

```

you can view this example [here](https://github.com/xetys/jhipster-uaa-setup)

hope that helps you to test....I will wrap all this is merged into a documentation and article as well

generators/server/index.js
```diff
@@ -578,6 +578,9 @@ module.exports = JhipsterServerGenerator.extend({
             if (this.applicationType !== 'microservice' && !(this.applicationType === 'gateway' && this.authenticationType === 'uaa')) return;
 
             this.template(SERVER_MAIN_SRC_DIR + 'package/config/_MicroserviceSecurityConfiguration.java', javaDir + 'config/MicroserviceSecurityConfiguration.java', this, {});
+            if(this.applicationType === 'microservice' && this.authenticationType === 'uaa') {
```
deepu105: space between `if (`

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.stereotype.Component;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+
+@Component
+public class LoadBalancedResourceDetails extends ClientCredentialsResourceDetails {
+    
+    Logger log = LoggerFactory.getLogger(this.getClass());
```
deepu105: change to `private static final Logger log = LoggerFactory.getLogger(LoadBalancedResourceDetails.class);` for uniformity

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,61 @@
...
+    @Autowired(required = false)
+    public void setLoadBalancerClient(LoadBalancerClient loadBalancerClient) {
+        this.loadBalancerClient = loadBalancerClient;
+    }
+
+    @Override
+    public String getAccessTokenUri() {
+        String serviceName = jHipsterProperties.getSecurity().getClientAuthorization().getTokenServiceId();
+        if (loadBalancerClient != null && !serviceName.isEmpty()) {
+            String newUrl;
```
deepu105: simplify to below, no need to assign to `newUrl` return directly
```
           try {
                return loadBalancerClient.reconstructURI(
                    loadBalancerClient.choose(serviceName),
                    new URI(super.getAccessTokenUri())
                ).toString();
            }
```

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
```diff
@@ -83,7 +83,7 @@ public void configure(ResourceServerSecurityConfigurer resources) throws Excepti
     @Override
     public void configure(ClientDetailsServiceConfigurer clients) throws Exception {
         /*
-        @TODO this should be done by a ClientDetailsService (similar to UserDetailsService) with an consumable resource
+        for a better client design, this should be done by a ClientDetailsService (similar to UserDetailsService)
```
deepu105: For

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -183,6 +183,14 @@ jhipster:
                 # Token is valid 24 hours
                 tokenValidityInSeconds: 86400
     <%_ } _%>
+    <%_ if ((applicationType === 'microservice' || applicationType === 'uaa') && authenticationType === 'uaa') { _%>
+        clientAuthorization:
+            # change this depending on your authorization server
+            tokenUrl: http://uaa:9999/oauth/token
```
deepu105: could the app name here be generated dynamically based on the uaa app name the user has choosen?

deepu105:
This looks good. I have some minor comments, mostly aesthetic. I guess its good to merge once those comments are taken care of

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
xetys: whats wrong here?

generators/server/templates/src/main/resources/config/_application-dev.yml
xetys: this would add a new question to microservice generator, similar to gateway...but yes, it could be done

generators/server/templates/src/main/java/package/config/_UaaConfiguration.java
deepu105: Nothing wrong, just capitalize the first word

generators/server/templates/src/main/resources/config/_application-dev.yml
deepu105: Hmmm, I thought we already ask the question when `uaa` is selected as authentication. But if we could make this more seamless the better

generators/server/templates/_build.gradle
```diff
@@ -207,6 +207,10 @@ dependencies {
     compile "org.springframework.cloud:spring-cloud-starter-config"
     compile "org.springframework.retry:spring-retry"
     <%_ } _%>
+    <%_ if (applicationType === 'microservice' && authenticationType === 'uaa') { _%>
```
cbornet: Feign clients are interesting even without uaa so I would the uaa check

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
```diff
@@ -335,6 +349,53 @@ public void setTokenValidityInSecondsForRememberMe(long tokenValidityInSecondsFo
             <%_ } _%>
         }
         <%_ } _%>
+
+        <%_ if((applicationType === 'microservice' || applicationType === 'uaa') && authenticationType === 'uaa') { _%>
+        public static class ClientAuthorization {
+
```
cbornet: It would be better to extend BaseOAuth2ProtectedResourceDetails

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,61 @@
...
+import java.net.URISyntaxException;
+
+
+@Component
+public class LoadBalancedResourceDetails extends ClientCredentialsResourceDetails {
+    
+    Logger log = LoggerFactory.getLogger(this.getClass());
+
+    @Autowired
+    public LoadBalancedResourceDetails(JHipsterProperties jHipsterProperties) {
```
cbornet: Use ```@ConfigurationProperties``` to automatically load from application.properties

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
cbornet: Actually remove this and use LoadBalancedResourceDetails in JHipsterProperties

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
```diff
@@ -228,22 +228,36 @@ public void setFrom(String from) {
 
         private final RememberMe rememberMe = new RememberMe();
         <%_ } _%>
-        <%_ if (authenticationType == 'oauth2' || authenticationType == 'jwt' || authenticationType == 'uaa') { _%>
+
+        <%_ if((applicationType === 'microservice' || applicationType === 'uaa') && authenticationType === 'uaa') { _%>
+
+        private final ClientAuthorization clientAuthorization = new ClientAuthorization();
+
```
cbornet: Use LoadBalancedResourceDetails directly

xetys:
@cbornet I like your idea using ```LoadBalancedResourceDetails``` directly as ```@ConfigurationProperties```, but this just doesn't work, because the getAccessTokenUri() is where the load balancing happens. So spring uses the getter during configuration resolution, and gots into an infinite cycle of trying to resolve a an access token uri, while setting it 

This is maybe not the "elegant" way, though the guys from spring cloud said, it's a good solution. But actually there is no way to inject a access token resolver, since they are instanciated hard inside the code. Just if someone asks...


But I extended ClientDetailsResource on ClientAuthoirzation configuration in JHipsterProperties instead.

generators/server/templates/src/main/resources/config/_application-dev.yml
xetys: we ask only for gateways, but i apply it to microservices in general

generators/server/templates/src/main/java/package/config/_FeignConfiguration.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.springframework.cloud.netflix.feign.EnableFeignClients;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+
+@Configuration
+@Profile("!test")
+@EnableFeignClients(basePackages = "<%=packageName%>.client")
+public class FeignConfiguration {
```
cbornet: Shouldn't this be put directly on the main application instead of pulling a configuration class ?

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -1,129 +1,147 @@
...
+    public JwtAccessTokenConverter jwtAccessTokenConverter() {
+        JwtAccessTokenConverter converter = new JwtAccessTokenConverter();
+        converter.setSigningKey(jHipsterProperties.getSecurity().getAuthentication().getJwt().getSecret());
+
+        return converter;
+    }
+
+    @Bean
+    public RequestInterceptor getOAuth2RequestInterceptor() throws IOException {
+        return new OAuth2FeignRequestInterceptor(new DefaultOAuth2ClientContext(), loadBalancedResourceDetails);
```
cbornet: If I understand well, this will put the OAuth2FeignRequestInterceptor on all FeignClient beans which is not always wanted. Eg : if you have FeignClients to communicate with remote systems.

generators/server/templates/src/main/java/package/config/_FeignConfiguration.java
xetys: no...because

* you can mock a test client this way easier. Running in profile test, no "real" feign clients are constructed
* you can define different non-loadbalanced and non-oauthed feign clients in other packages

this is a fix I applied after working for some weeks with my feign client implementation and stucked on testing this stuff without a faked endpoint for oauth and the other tested service

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
xetys: using ClientCredentialsResourceDetails (which is parent of LoadBalancedResourceDetails...)

generators/server/templates/src/main/java/package/config/_FeignConfiguration.java
cbornet: So @EnableClients needs to be on separate component because of https://github.com/spring-cloud/spring-cloud-netflix/issues/1016

xetys:
ok, this my final state for now...in short

* fixes the error from last commit
* invention of `@AuthorizedFeignClient`

## About `@AuthorizedFeignClient`

In the previous version of this PR, every feign client in the project would be intercepted
with the OAuth2RequestInterceptor. Removing this behaviour makes it possible, to configure
any kind of feign client. To fill the lack of a preconfigured feign client, able to cleanely
communicate with JHipsters UAA, the developer may use `@AuthorizedFeignClient` instead.

*Note:* due to a annotation bug, only `@AuthorizedFeignClient(name = "otherapp")` works, but not
`@AuthorizedFeignClient("otherapp")`

## Why a FeignConfiguration instead of `@EnableFeignClients` on app?

At the moment this is the cleanest way of easily disable feign clients for testing purposes. There would be other way, but they don't work for several reasons. I want to note, that feign clients are only disabled, when the profile "test" is present. So there **is** a way to test with real feign clients, but also the way to just implement a feign client and mock it, so it will be injected the proper spring way.

## Example code

As always, [my jhipster UAA demo microservice](https://github.com/xetys/jhipster-uaa-setup) is up to date with this code, showing examples how to use the new functions


## Upcoming tasks

As far this is merged, @cbornet and me will work on 2 solutions, so it will be possible to generate feign clients for own JHipster microservices as well as for any application providing a swagger spec.

But nevertheless, any kind of manual feign clients can be implemented with this PR


thanks to @cbornet for patience and advices

generators/server/templates/_pom.xml
```diff
@@ -616,6 +616,16 @@
...
         </dependency>
         <%_ } _%>
+        <%_ if (applicationType === 'microservice' && authenticationType === 'uaa') { _%>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-security</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-feign</artifactId>
```
cbornet: This will probably be needed even without UAA

xetys: I could carry out the feign block to general (since it might be interesting for monoliths also)

but cloud security is really needed when working with uaa

generators/server/index.js
```diff
@@ -579,6 +579,12 @@ module.exports = JhipsterServerGenerator.extend({
             if (this.applicationType !== 'microservice' && !(this.applicationType === 'gateway' && this.authenticationType === 'uaa')) return;
 
             this.template(SERVER_MAIN_SRC_DIR + 'package/config/_MicroserviceSecurityConfiguration.java', javaDir + 'config/MicroserviceSecurityConfiguration.java', this, {});
+            if (this.applicationType === 'microservice' && this.authenticationType === 'uaa') {
+                this.template(SERVER_MAIN_SRC_DIR + 'package/config/_FeignConfiguration.java', javaDir + 'config/FeignConfiguration.java', this, {});
```
cbornet: Probably needed even without uaa

PierreBesson:
:+1: I'm also in favor of putting the annotation on a configuration class rather than on the app's main class.

I think that in the best case this PR should also authorize Feign clients for JWT microservices. This should be much more simple than Oauth2. I've done some experimenting on this but wasn't able to make it work.

Normally something like this should make it work:
```
@Configuration
public class FeignConfiguration {
    @Bean
    public RequestInterceptor getJwtRequestInterceptor() throws IOException {
        return new JwtRequestInterceptor();
    }

    private class JwtRequestInterceptor implements RequestInterceptor {
        @Inject
        private TokenProvider tokenProvider;

        @Override
        public void apply(RequestTemplate requestTemplate) {
            String jwt = tokenProvider.createToken(SecurityContextHolder.getContext().getAuthentication(), false);
            requestTemplate.header(JWTConfigurer.AUTHORIZATION_HEADER, "Bearer " + jwt);
        }
    }
}
```
ping @pascalgrimaud I think could make use of this in your project.

xetys:
well, my annotation is causing more issues than expected right now. At the moment I just cannot achieve the openess of feign clients, its all about "all" or "nothing"....i am working on that

xetys:
update...we got some issue with feign in spring itself....when using feign clients together with eureka and oauth, there is no way of setting individual configuration to the clients. 

@cbornet and me killed some hours to figure out how to reproduce that with basic spring, what was not that easy, but actually worked. I am not supporting them by finding that issue.

until that the currently uploaded solution does work for oauth feign clients

@PierreBesson i didn't manage to try your requestinterceptor yet. maybe you should think about not every request contains a security context, like you expect in your interceptor...but it's just a thought

i suggest to wait for me to move that feign client avaible for all applications, and than to merge...since all i can do to make the multiple feign configuration work is to help the spring team find the bug and fix it

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.security.oauth2.client.token.grant.client.ClientCredentialsResourceDetails;
+import org.springframework.stereotype.Component;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+
+@Component
+public class LoadBalancedResourceDetails extends ClientCredentialsResourceDetails {
+
```
cbornet: As discussed, it would be better to create the LoadBalancedResourceDetails class without @Component nor jhipster dependency and then create the bean in OAuth2InterceptedFeignConfiguration with
```java
@Bean
@Configurationproperties("oauthclient.security.oauth2")
LoadBalancedResourceDetails loadBalancedResourceDetails () { 
    return new LoadBalancedResourceDetails();
}

@Bean
public RequestInterceptor getOAuth2RequestInterceptor() {
    return new OAuth2FeignRequestInterceptor(new DefaultOAuth2ClientContext(), loadBalancedResourceDetails());
}
```
That way there is no risk that LoadBalancedResourceDetails is scanned for something other than the feign configuration.

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -183,6 +183,14 @@ jhipster:
                 # Token is valid 24 hours
                 tokenValidityInSeconds: 86400
     <%_ } _%>
+    <%_ if (applicationType !== 'uaa' && authenticationType === 'uaa') { _%>
+        clientAuthorization:
```
cbornet: I think dashed kebab-case is prefered in application.yml

generators/server/templates/src/main/java/package/config/_JHipsterProperties.java
```diff
@@ -335,6 +350,21 @@ public void setTokenValidityInSecondsForRememberMe(long tokenValidityInSecondsFo
             <%_ } _%>
         }
         <%_ } _%>
+
+        <%_ if((applicationType === 'microservice' || applicationType === 'uaa') && authenticationType === 'uaa') { _%>
+        public static class ClientAuthorization extends ClientCredentialsResourceDetails {
+            private String tokenServiceId;
```
cbornet: Not needed if using ConfigurationProperties

generators/server/templates/src/main/java/package/client/_OAuth2InterceptedFeignConfiguration.java
```diff
@@ -0,0 +1,27 @@
+package <%=packageName%>.client;
+
+import com.mycompany.myapp.config.LoadBalancedResourceDetails;
```
jdubois: You should have the package name variable here

generators/server/templates/src/main/java/package/config/_LoadBalancedResourceDetails.java
```diff
@@ -0,0 +1,59 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.client.loadbalancer.LoadBalancerClient;
+import org.springframework.security.oauth2.client.token.grant.client.ClientCredentialsResourceDetails;
+import org.springframework.stereotype.Component;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+
```
jdubois: there is an extra line here

generators/server/templates/_build.gradle
```diff
@@ -207,6 +207,10 @@ dependencies {
     compile "org.springframework.cloud:spring-cloud-starter-config"
     compile "org.springframework.retry:spring-retry"
     <%_ } _%>
+    <%_ if (applicationType === 'microservice') { _%>
```
jdubois: I thought your code was only for people using UAA?

deepu105:
@xetys could you please check the pending comments please?

xetys:
@deepu105 please forgive me :smile: I am currenty in my exams phase...while inventing scrum at work with prove of concept. So my caps were a bit overleaded the recent weeks. 

I am crying anytime I see something happens here and I've got do wait on this

deepu105:
No worries, this can wait school is more important

jdubois:
hey @xetys I didn't know you were a student: yes your exams are much more important!

xetys:
i just need some official paper for my skills :laughing:

xetys:
ok guys...now where I am finally in vacation, I think I have finished this loooooong PR....I am really sorry for that it took that long, since there is a huge setup to test it properly, and to remember all the things to keep in mind...well, lets walk through quickly:

finally all comment are now resolved, and the biggest one was to make that feign configuration interoperable as possible. The `LoadBalancedRessourceDetails` are now threated like configuration properties, and are cleanely resolving the token as intended, but is no longer a `@Component`
One big issue is, that currently it is not possible to strictly distinct multiple feign clients, using the same service. The details were reported here: https://github.com/spring-cloud/spring-cloud-netflix/issues/1211

In short: if you ever have one service "foo", supporting different authentication modes, you cannot define this using `@FeignClient`, since they intended to deprecate the "serviceId" propety, which is now an alias for name. Internaly all clients using one name are configured the same way.

For all the cases, the feign configurations are now cleanly seperated, using a new JHipster Wide annotation

`@ExcludedFromComponentScan`,

which you can simply apply to any configuration class you like, to exlude it from component scan. This was needed for my specific UAA RequestInterceptor, which now is explicitly mentioned, if you prefer `@AuthorizedFeignClient` over `@FeignClient` with refering to that `OAuth2FeignConfiguration`.


I hope I am now done here, so I can focus on the following upcoming contribution, as:

* finally do some documentation, which had been done very very earlier
* generating `@AuthorizedFeignClient`s using `jhipster:entity <NAME>` in a similar fashion as it works for generating entites for gateway
* making `@AuthorizedFeignClient` work for JWT services also
* resolving all the bugs and questions about UAA

deepu105:
Im adding the needs-review badge as I would like inputs from @jdubois @PierreBesson @pascalgrimaud @sdoxsee @gmarziou etc here

xetys:
documentation already on the way :smile:

generators/server/templates/_pom.xml
```diff
@@ -629,6 +629,18 @@
...
         <%_ } _%>
+        <%_ if (authenticationType === 'uaa') { _%>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-security</artifactId>
+        </dependency>
+        <%_ } _%>
+        <%_ if (applicationType === 'microservice' || applicationType === 'gateway' || applicationType === 'uaa') { _%>
+        <dependency>
+        <groupId>org.springframework.cloud</groupId>
```
deepu105: formatting seems off here

xetys: i'm not sure what's wrong here...could u please explain?

deepu105: GroupId and artifactId should be intended 4 spaces

sdoxsee:
I'm on vacation this week. I'll take a look early next week.

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -187,6 +187,17 @@ jhipster:
                 # Token is valid 24 hours
                 tokenValidityInSeconds: 86400
     <%_ } _%>
+    <%_ if (authenticationType === 'uaa') { _%>
+    security:
+        client-authorization:
+            # change this depending on your authorization server
+            <%_ if (applicationType !== 'uaa') { _%>
+            accessTokenUri: http://<%= uaaBaseName.toLowerCase() %>:9999/oauth/token
```
sdoxsee: should port 9999 be assumed or can we get this another way?

sdoxsee:
@xetys Good work. Do you have a reference for the Syer-approved approach for ribbonizing uaa you mention in https://github.com/jhipster/generator-jhipster/pull/3662#issuecomment-223748920? I had an issue referencing the uaa (authserver) using a discoverable uri for obtaining the public key http://stackoverflow.com/questions/36900650/using-security-oauth2-resource-jwt-keyuri-with-a-discoverable-service-name and was curious if I could use a different approach.

generators/server/templates/src/main/resources/config/_application-dev.yml
xetys: We could do the same move here as with uaaBasename like uaaPort. I will take a look

xetys:
It was some time ago, when the only way to apply load balancing that way I did in LoadBalancedResourceDetails, and I pasted that snippet to gitter, asking for a better solution. Dave Syer told, that this a good way to go, if I want to use it on client credentials only. For the moment, this is exactly what I need. 

Your issue with that url is something I think I tried, too. With the same result. If I remember it the right way, at the moment load balancing is only applied for the final service you are going to call, they just missed to add support for uaa resolving also. I was also on some pr to solve this, but was to busy with this PR and university...

xetys:
@sdoxsee just for your information: i removed the 9999 port in configuration since the url is ribbonized and reconstructed anyway....the only important thing is to have the service name as host alias inside

xetys:
so...is there still something to do on this PR?

deepu105:
LGTM @jdubois @PierreBesson please see if this is good for you

jdubois:
Lots of stuff, I didn't see this coming!!

- I'll need some time to review everything
- I'm seeing a few formatting issues -> I'll correct them myself, let me do the merge

generators/server/templates/src/main/java/package/config/_ExcludedFromComponentScan.java
```diff
@@ -0,0 +1,9 @@
+package <%=packageName%>.config;
+
+/**
+ * a simple annotation, to be used for easy exclusion from component scan
+ *
+ * @author David Steiman
+ */
```
jdubois: I'm not a big fan of this annotation: it should be a standard Spring Boot annotation (need to check if this exists), or we can just exclude classes as usual. And it's only used once here.

It could be a good idea (if it doesn't already exists in Spring Boot), but then we would need to use it elsewhere in our code (we have other exclusions), and that should be another PR as it's a different issue.

xetys: The tools from spring boot actually allow to define, which annotations lead to bean exclusion. Using classes in exclude filter is good, if it's the only component to be excluded. As soon there are more of them, it just looks ugly in my opinion.

I just thought using some annotation like this would be a more elegant way. The only downside may be, you don't see excluded beans by looking on app class (but a quick search helps here...)

xetys: So, how do we do here. Is this OK for all or do I have to change this?

jdubois: I'm removing it -> then it's a good idea, but for a separate PR. I'd also like to see if this kind of annotation doesn't already exist, it looks so practical I'm surprised we need to do it.

generators/server/templates/src/main/java/package/client/_OAuth2InterceptedFeignConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.cloud.security.oauth2.client.feign.OAuth2FeignRequestInterceptor;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.oauth2.client.DefaultOAuth2ClientContext;
+
+import javax.inject.Inject;
+import java.io.IOException;
+
+@Configuration
+@ExcludedFromComponentScan
```
jdubois: In fact I don't understand why you exclude this configuration

jdubois:
@xetys I did a specific branch for this https://github.com/jhipster/generator-jhipster/tree/xetys-feign-clients as I've modified quite a number of things

And I'm removing the "ready to merge" tag, as I've got issues and questions

- I did some re-formatting and merged conflicts, that should be OK
- I don't understand why you want to exclude the OAuth2InterceptedFeignConfiguration configuration -> can you explain this to me?
- I removed the @ExcludedFromComponentScan as we need to check this separately

Otherwise, I've created a new "uaa" server with this branch, and tests don't pass as the Liquibase configuration is wrong... Do you have the same issue?

Here is my `.yo-rc.json`:

    {
      "generator-jhipster": {
        "jhipsterVersion": "3.6.1",
        "baseName": "uaa",
        "packageName": "io.github.jhipster.bbl",
        "packageFolder": "io/github/jhipster/bbl",
        "serverPort": "9999",
        "authenticationType": "uaa",
        "hibernateCache": "no",
        "databaseType": "sql",
        "devDatabaseType": "h2Disk",
        "prodDatabaseType": "mysql",
        "searchEngine": "no",
        "buildTool": "maven",
        "enableTranslation": false,
        "applicationType": "uaa",
        "testFrameworks": [],
        "jhiPrefix": "jhi",
        "skipClient": true
      }
    }

xetys:
> I don't understand why you want to exclude the OAuth2InterceptedFeignConfiguration configuration -> can you explain this to me?

if this is a global configuration, the `RequestInterceptor` would be injected into every feign client. If you are about to use feign clients, more then just on service-to-service communication, having that configuration not excluded will force **every** feign client to get authorized via uaa. This is what @cbornet was mentioning earlier in this PR.


> I removed the @ExcludedFromComponentScan as we need to check this separately

as I said, one way to exclude that `OAuth2InterceptedFeignConfiguration` would be

``` java
@ComponentScan(
    excludeFilters = {
        @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, value = ClassOne.class),
        @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, value = ClassTwo.class),
    }
)
``` 
(AFAIK)

So if we need more classes to exclude for whatever reason, this part would be growing, which would look really messy to me. 

I am just a fan of that annotation, since it looks nice (in my taste) and easy and handy to use, without blowing this block.

the liquibase problem is really strange, it tries to find the YAML configuration instead of the XML one...no idea why...and no idea if this is even UAA related...i am looking

xetys:
and my npm is trolling me (again) this stops me from making more tests on that for now :angry: :angry:

xetys:
well, as soon I cry here, it started to work

that branch is generally broken with liquibase...even with a JWT microservice...

xetys:
found: maybe removing `@ComponentScan` globally was not the best idea :laughing: 

i fixed that in the branch

jdubois:
OK great.
Not sure I'll have the time to test as this is quite big.
If not, it will be in another release, but holidays are over and I'd like
to have more frequent releases, so that won't take long.

Le 7 sept. 2016 6:40 PM, "David Steiman" <notifications@github.com> a
écrit :

found: maybe removing @ComponentScan globally was not the best idea 😆

i fixed that in the branch

—
<https://github.com/jhipster/generator-jhipster/pull/3662#issuecomment-245342089>,
or mute the thread

xetys:
sure, ask if there's something more...

ValerioZ:
Sorry, I cannot find @AuthorizedFeignClient. I cannot perform any request by Feign to a protected resource with jhipster module in the marketplace.
Have you any idea of why?
Thanks

jdubois:
@xetys I am testing on the https://github.com/jhipster/generator-jhipster/tree/xetys-feign-clients and I'm going to merge this if it's all OK.

I just have one issue at the moment: I generated a microservice, with one entity, and its "./mvnw test" do not pass, as it's looking for a UAA server. So you need to mock this, or provide a specific "test" profile to solve this. This is not a blocker for me, at the moment.

jdubois:
@xetys sorry but I put everything in Docker, and both my gateway and microservice failed, with error:

    Caused by: org.springframework.beans.NotWritablePropertyException: Invalid property 'security.client-authorization[accessTokenUri]' of bean class [io.github.jhipster.bbl.config.JHipsterProperties]: Cannot access indexed value in property referenced in indexed property path 'client-authorization[accessTokenUri]'; nested exception is org.springframework.beans.NotReadablePropertyException: Invalid property 'security.client-authorization[accessTokenUri]' of bean class [io.github.jhipster.bbl.config.JHipsterProperties]: Bean property 'security.client-authorization[accessTokenUri]' is not readable or has an invalid getter method: Does the return type of the getter match the parameter type of the setter?

Both of them are configured to connect the UAA server, of course.

Sorry but I can't merge it right now, it's not working :-(

xetys:
I will have time to take a look in one hour. I think there is some file just missing somewhere, which I need to find. 

So what do you think about the Bean exclusion annotation, keep it or use direct type exclusion on component scan instead?

jdubois:
I think the Bean exclusion could be good, but it should be in another PR, so we don't mix ideas/concerns. And I still don't understand why you want to exclude that Bean, but if you do I think you can just exclude it directly by type (I'm not 100% sure here, as I'm not sure of why you do it)

xetys:
As I explained, we want to leave the option open for developers, to use feign clients generally, without the uaa stuff. If we don't exclude, the OAuth request interceptor gets injected to all clients. This behaviour is described in official spring docs.

xetys:
@jdubois I regenerated my [JHipster UAA setup][] using this branch and do not expierienced problems using docker...

I uploaded here my example, so you can try my setup

you can use my docker-compose setup

[JHipster UAA setup]: https://github.com/xetys/jhipster-uaa-setup

jdubois:
@xetys I merged it in https://github.com/jhipster/generator-jhipster/tree/xetys-feign-clients and corrected a few things (there was a compilation error for gateway & uaa). When I run it, I still have a error on the gateway:

    Bean property 'security.client-authorization[accessTokenUri]' is not readable or has an invalid getter method: Does the return type of the getter match the parameter type of the setter? -> [Help 1]

jdubois:
I don't understand what's going on:

- Some of the stuff from this PR looks like being already in the master branch
- And the master branch is broken, both the microservice and gateway fail

Looks like I pushed stuff by mistake in the master branch, I'm totally lost here... That's bad, I wanted to do the 3.7.0 this evening as I won't have much time this week...

jdubois:
OK, I understand, this is because there is #4052 which is in both branches, and which was merged by @deepu105 today!
I might as well merge all this to master:

- I'll close this ticket when this merged (I'm merging the branch on generator-jhipster)
- Still there will be the issue for the gateway

deepu105:
@xetys next time try not to mix up commits from different PRs together. This has a lot of unwanted commits and it messes up the commit history on master

ValerioZ:
@xetys I'm using official 3.6.1 release with beta UAA service. Thanks.

xetys:
@ValerioZ as you can see in this PR, this feature is currently in release progress and is scheduled for 3.7, and is not available to prior releases. Update to 3.7 as soon it is out, to use features like `@AuthorizedFeignClient`

Please note, that your question would be placed better in stackoverflow, rather then in this discussion.

PR #3690, opened 2016-06-07
Typo fix and replacing String with constant value

o:


deepu105:
I think you have a missing import

o:
@deepu105 Fixed thanks

PR #3691, opened 2016-06-07
Cleaned up translation + docker-compose file for email

lrkwz:
I've added a docker-compose file to integrate maildev


PierreBesson:
Hi @lrkwz,
It would have been better if you had made 2 PR. Also the maildev docker-compose should have been discussed as an issue first !

I'm actually using a similar docker-compose file at work, it's also my favorite dev mail server.
I would be in favour to merge it in JHipster as `djfarrelly/maildev` is the official image of [djfarrelly/MailDev](https://github.com/djfarrelly/MailDev) (500 stars on github). However it's not verified by Docker Hub so some people may be concerned with it.

However it would need the following changes:
- rename `garanteasy` to `maildev`
- specify the version number 
```
    maildev:
        container_name: maildev
        image: djfarrelly/maildev:0.13.3
        ports:
            - 1080:80
            - 25:25
```

lrkwz:
You are perfectly right! I did change garanteasy to ```<%= baseName.toLowerCase() %>``` as all the other  docker-compose service you built (well I did it already but somehow I loosed the change).

Coming to the **mail** I quite agree on using maildev.
Well mailhog is nice and small as well (950 stars on github), schickling/mailcatcher is way tooo big. 
I do not think it shouldn't be necessary to fix a version in this case (database is different matter) as smtp i really quite stable isn't it?

generators/server/templates/src/main/docker/_mail.yml
```diff
@@ -0,0 +1,8 @@
+version: '2'
+services:
+    <%= baseName.toLowerCase() %>-mail:
+        container_name: <%= baseName.toLowerCase() %>-mail
+        image: djfarrelly/maildev:0.13.3
```
lrkwz: Can be changed to mailhog/mailhog (in this case ports should be ```25:1025``` and ```1080:8025```)

PierreBesson:
With those changes it looks good to me. On the topic of specifying the version numbers I know it might seems like a good idea to leave it implicit but we have already decided that they are mandatory and who knows if in the future maildev will change it's default port or whatever that would render it incompatible...

deepu105:
Could you please make another PR for the maildev change and remove it from here so that the i18n pr can be merged?

pascalgrimaud:
**Maildev** is a nice project, but I'm reluctant to have this docker-compose file in our generator, as it's not an official Docker image.
That's why I write this tips: [tip_local_smtp_server](http://jhipster.github.io/tips/015_tip_local_smtp_server.html) => users are free to use it or not. They can  use their own smtp server.

deepu105:
I agree with Pascal

lrkwz:
ok I'll try to split the PR.

*=> Luca <http://stackexchange.com/users/239746>*



2016-06-08 8:25 GMT+02:00 Deepu K Sasidharan <notifications@github.com>:

> I agree with Pascal

PR #3700, opened 2016-06-12
fix APT cleanup in Dockerfile

Hades32:
running any clean commands in a separate RUN doesn't have the wanted effect. The files would be deleted, but the image size would stay the same, because they still exist in a lower layer.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,14 @@ RUN \
   npm install -g \
     yo \
     bower \
-    gulp-cli
+    gulp-cli \
+
+  # cleanup
+  apt-get clean && \
+  rm -rf \
+    /var/lib/apt/lists/* \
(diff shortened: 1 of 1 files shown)


PierreBesson:
Nice optimization ! It seems obvious when you know that docker images are constituted of layers for each instruction being run and that those layers must be downloaded in full when pulling the image.

I'm merging this.

PierreBesson:
This really succeeded in reducing the image size: https://hub.docker.com/r/jhipster/jhipster/tags/
We achieved a 40% reduction in size :tada:. Thanks @Hades32.

cbornet:
Can we gain as much in the vagrant devbox ?

PierreBesson:
@cbornet No this was only possible because we were misusing docker !

PR #3709, opened 2016-06-14
fix Dockerfile to properly clean apt repository

ruddell:
From PR #3700, `apt-get clean` is in the same command as `npm install -g` line when it should have `&&` between the two.  It currently installs the npm packages apt-get@0.1.0 and clean@4.0.2


pascalgrimaud:
good catch :eyes: 
thx @ruddell

PR #3712, opened 2016-06-15
set up sorting and pagination for user-management to match entities

ruddell:
Pagination for user-management still uses the older method (refreshing/viewing/deleting a User resets the page back to 1).  This aligns user-management pagination with the way entities do it, offering a much better user experience.

I will open a separate PR for client-2 (let me know if this isn't the right way)

### DIFF
--- a/generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
+++ b/generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
@@ -5,9 +5,9 @@
         .module('<%=angularAppName%>')
         .controller('UserManagementController', UserManagementController);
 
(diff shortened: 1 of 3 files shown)


generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
```diff
@@ -71,8 +50,45 @@
                 lastModifiedBy: null, lastModifiedDate: null, resetDate: null,
                 resetKey: null, authorities: null
             };
-            vm.editForm.$setPristine();
```
ruddell: I didn't see a vm.editForm variable and activating/deactivating a user threw an error in the console saying vm.editForm is undefined, so I removed these.

PR #3713, opened 2016-06-15
Client2: set up sorting and pagination for user-management to match entities

ruddell:
PR for angular2 user-management pagination (See #3712)


deepu105:
Thank you @ruddell

PR #3726, opened 2016-06-16
typo fixes and constant file changes

sendilkumarn:


generators/client-2/templates/src/main/webapp/app/components/alert/_alert.service.ts
```diff
@@ -1,14 +1,18 @@
 export function AlertService () {
     this.toast = false;
     /*jshint validthis: true */
-    this.$get = getService;
+    var service = {  getService: getService }
```
deepu105: @sendilkumarn why did you change this?

sendilkumarn: It was throwing an error that the function to return something.

deepu105: Its a provider and hence that line is required. we need to see how we can migrate this properly

sendilkumarn: probably 

`var service = {  getService: getService }
    this.$get = service;
    this.showAsToast = function(isToast) {
        this.toast = isToast;
    };
    getService.$inject = ['$timeout', '$sce', '$translate'];
    return service;`

we can write this way but why this.$get? is that advisable?

Issue #3719, opened 2016-06-15
User Management with Cassandra Temporarily Broken


ruddell:
##### **Overview of the issue**
From my recent PR #3712, user management does not load because Cassandra does not support sorting.

##### **JHipster Version(s)** 
Master

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 
Any Cassandra project

##### **Suggest a Fix** 
I will fix the issue this in an hour or two.

pascalgrimaud:
closed by #3712

ruddell:
That PR was for client 2.  My internet went out at home so I'll send in my code tonight

pascalgrimaud:
Oh sorry, I reopen this.
Time to sleep for me :D

ruddell:
I still need to PR Client-2's code but I want to wait for #3726 to get pulled before I update user-management.state.js

deepu105:
Its merged

Thanks & regards,
Deepu

jdubois:
If that's merged can we close this?

ruddell:
Just PRed the angular-2 code so now it should be fixed.

PR #3727, opened 2016-06-16
remove pagination when using cassandra to fix user management

ruddell:
Fixes the Client v1 user management (partial fix for #3719, still need to update Client 2 code)


deepu105:
Do we do this checks for entity as well?  I dont remember seeing it. Anyway need to check later

ruddell:
@deepu105 Cassandra forces pagination to be 'no' for entities (no question is asked), and the templates do not include sorting when there is no pagination.

deepu105:
Oh yes you are right i forgot

Thanks & regards,
Deepu

PR #3728, opened 2016-06-17
Constants and typo fixes

sendilkumarn:
Need to replace ngconstant-dev function. currently it writes in angular1 format.


deepu105:
I just did that :+1:

PR #3729, opened 2016-06-17
Simplify Angular 1.x vs. 2.x choice. Upgrade to Angular 2 rc2.

mraible:
RC2 came out yesterday. We should keep up with the latest and greatest. 

Question: Does anyone know what the difference between `core-js` and `es6-shim` is? angular-cli uses es6-shim, but it looks like angular2-seed uses core-js.

* angular-cli: https://github.com/angular/angular-cli/commit/4113f3060d919a1aad2ce47c1609f96c12f2ab99
* angular2-seed: https://github.com/mgechev/angular2-seed/blob/master/package.json

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -133,7 +133,7 @@ function askForClient() {
             return (applicationType !== 'microservice');
         },
         message: function (response) {
-            return getNumberedQuestion('Which *Angular* version would you like to use for the client side?', applicationType !== 'microservice');
+            return getNumberedQuestion('Which *Angular* version would you like to use for the client?', applicationType !== 'microservice');
         },
         choices: [
             {
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -133,16 +133,16 @@ function askForClient() {
             return (applicationType !== 'microservice');
         },
         message: function (response) {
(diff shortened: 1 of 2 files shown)


deepu105:
@mraible I have no idea, there are lots of discrepancies between different documentation from angular2 and its very confusing actually. Hope ng2 team comes up with some better and consistent documentation

generators/app/prompts.js
```diff
@@ -133,16 +133,16 @@ function askForClient() {
...
         choices: [
             {
                 value: 'angular1',
-                name: 'I`ll use AngularJS 1 (stable)'
+                name: 'Angular 1.x (stable)'
             },
             {
                 value: 'angular2',
-                name: 'I want to try AngularJS 2 (Please note that this is still in beta)'
+                name: 'Angular 2.x (rc2)'
```
deepu105: @mraible But some how we need to indicate that this option is in beta within JHipster so that people looking for a stable client can use angular1 what about labeling it `[BETA] Angular 2.x (rc2)`?

mraible: @deepu105 Done on `angular-2` branch with https://github.com/jhipster/generator-jhipster/commit/745ed07ef75e77a0924ee2c5fac16b95dfe46375.

PR #3735, opened 2016-06-17
Services update

sendilkumarn:


generators/client-2/templates/src/main/webapp/app/admin/configuration/_configuration.service.ts
```diff
@@ -1,47 +1,42 @@
...
-                angular.forEach(response.data, function (data) {
-                    properties.push(data);
-                });
-                var orderBy = $filter('orderBy');
-                return orderBy(properties, 'prefix');
+                response.data.forEach(prop => {
+                    properties.push(prop);
+                });                
+
+                return properties.filter(properties => 'prefix');
```
deepu105: Shouldn't this be doing a sort?

generators/client-2/templates/src/main/webapp/app/admin/configuration/_configuration.service.ts
```diff
@@ -1,47 +1,49 @@
...
-                angular.forEach(response.data, function (data) {
-                    properties.push(data);
-                });
-                var orderBy = $filter('orderBy');
-                return orderBy(properties, 'prefix');
+                response.data.forEach(prop => {
+                    properties.push(prop);
+                });                
+
+                return properties.sort( (propertyA, propertyB) => {
```
deepu105: may be we should make orderBy and filter as util services as it would be required often. Can you do a PR later for that?

sendilkumarn: yeah will do that.

deepu105: also when you are converting js files, try to put in the types as well

PR #3741, opened 2016-06-18
Upgrade gulp-angular-templatecache to fix package in prod profile

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/3732


pascalgrimaud:
I'm merging this to fix the current build

deepu105:
@jdubois can we release a patch as this might be affecting all

jdubois:
OK I'll do a release ASAP.
I'm currently travelling, with very limited Internet access, so it will probably be tomorrow.

PR #3750, opened 2016-06-21
stop the entity generator when error instead of continuing to prompts

### LINKED ISSUE #3749: Entity Subgenerator continues prompts after errors
If the entity JSON file has invalid JSON in it (such as a missing comma), the entity subgenerator fails to read the configuration file.  The issue is that it continues on and prompts the user for what action they want to take (regenerate, update, etc).
Confusing to users, if the JSON is invalid it should say that and quit.  Currently it exits silently after going through the prompts.

##### **JHipster Version(s)** 
3.0.0 -> Master

Any

Any - at least one is necessary (I removed a comma in the JSON below to demonstrate the issue)
```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "name",
            "fieldType": "String"
        }
    ],
    "changelogDate": "20160621184950"
    "dto": "mapstruct",
    "service": "serviceClass",
    "entityTableName": "foo",
    "pagination": "pagination"
}
```

Mac

##### **Reproduce the error** 
Create an entity (or copy the above JSON)
Remove a comma
Re-run the entity subgenerator
Answer the prompts and the generator silently exits

##### **Suggest a Fix** 
PR linked below.  Thanks to @fredcooke for pointing the issue out

ruddell:
Fix #3749

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -199,8 +199,7 @@ module.exports = EntityGenerator.extend({
         try {
             this.fileData = this.fs.readJSON(this.fromPath);
         } catch (err) {
-            this.log(chalk.red('\nThe configuration file could not be read!\n'));
-            return;
+            this.error(chalk.red('\nThe entity configuration file could not be read!\n'));
         }
         this.relationships = this.fileData.relationships;
         this.fields = this.fileData.fields;
--- a/travis/samples/.jhipster/CassBankAccount.json
+++ b/travis/samples/.jhipster/CassBankAccount.json
@@ -52,7 +52,7 @@
         },
         {
             "fieldName": "operationsFile",
-            "fieldType": "ByteBuffer",
+            "fieldType": "ByteBuffer"
         }
     ],
     "changelogDate": "20150805124838",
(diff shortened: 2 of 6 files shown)


generators/entity/index.js
```diff
@@ -199,8 +199,8 @@ module.exports = EntityGenerator.extend({
         try {
             this.fileData = this.fs.readJSON(this.fromPath);
         } catch (err) {
-            this.log(chalk.red('\nThe configuration file could not be read!\n'));
-            return;
+            this.log(chalk.red('\nThe entity configuration file could not be read!\n'));
+            throw err;
```
deepu105: Dont throw error. Use `this.error ()` instead of `this.log ()` and yeoman will exit automatically and thats the recommended approach

ruddell:
Thanks for the tip, good to know.

ruddell:
Turns out one of the test entities had invalid JSON which caused the failed build.  I updated it.

Does this need an angular-2 PR? I'm confused if everything needs to have a separate PR for that or just frontend files...

deepu105:
@ruddell not required. As long as the change doesnt affect the `client` sub generator its all good

Issue #3476, opened 2016-04-26
Lombok dependency for LoC reduction


imTachu:
##### **Overview of the issue**

Lombok dependency could be included in order to reduce the generated LoC and to have 'cleaner' java classes. 

##### **Motivation for or Use Case** 

This issue is just a suggestion since everything works fine now... Lombok allows you to annotate variables and classes in order to generate getters, setters, toString, equals and hashcode. An example:

`@Getter`
`@Setter`
`private String name;`

This will generate the getter and setter for the variable name ONLY in the .class... The .java file will be clean and in the variable definition itself you can see if it already has those methods.

`@Getter`
`@Setter`
`public class AnyClass ...`

By annotating a class, you will automatically have getters and setters for everybody.

See also: @Data, @ToString @EqualsAndHashCode 
https://projectlombok.org/

##### **JHipster Version(s)** 

v3.0.0

##### **Browsers and Operating System** 

Chrome 49 over Windows 8.1

mallim:
Does mapstruct able work with lombok?

notflorian:
I'm afraid lombok is not compatible with Java 9 : https://groups.google.com/forum/#!topic/bazel-discuss/ZRWdqJfYEPw

Also, the author of this library wrote this in Reddit a long time ago :

> I'm the author of project lombok.
> It's a total hack. Using non-public API. Presumptuous casting (knowing that an annotation processor running in javac will get an instance of JavacAnnotationProcessor, which is the internal implementation of AnnotationProcessor (an interface), which so happens to have a couple of extra methods that are used to get at the live AST).
> On eclipse, it's arguably worse (and yet more robust) - a java agent is used to inject code into the eclipse grammar and parser class, which is of course entirely non-public API and totally off limits.
> If you could do what lombok does with standard API, I would have done it that way, but you can't. Still, for what its worth, I developed the eclipse plugin for eclipse v3.5 running on java 1.6, and without making any changes it worked on eclipse v3.4 running on java 1.5 as well, so it's not completely fragile.

Cleaner is not always better if it is then harder to debug, or understand the real code executed by the JVM, without having to decompile the class. Also, the java code is no more portable without having to include the lombok dependency, and can cause confusion when sharing code.

[...]

Issue #2734, opened 2016-01-30
Java preprocessor doesn't work in JHipster project


delor:
I have tried to use Lombok and http://immutables.github.io/ with JHipster project but it seams that something in build process is "blocking" preprocessor from generating code.

example source code https://bitbucket.org/delor/test_jhipster_immutables

JHipster Generator v2.27.0

jdubois:
Not a JHipster bug, as we don't support Lombok.
And if look at the tickets history, you'll see I'm strongly against Lombok exactly for this kind of reasons.

delor:
What about other libraries based on preprocessor? I do mention also Immutables library.

jdubois:
We do support other Java annotation processors, like MapStruct or JPA, but we do a specific set up for them, and they don't all work together, this is why we limit them.

Issue #1334, opened 2015-04-03
may project lombok fit here?


rblazquez:
Have you consider include project lombok to reduce the amount of boilerplate code?

I know jHipster claims to use a small set of dependencies and i personally agree with this idea, but in my experience the reduction of boilerplate code achieves a much more clean code, and less maintenance effort.

Also project lombok is small set of annotations that are also optional, is a stable project and integrates very well with IDEs.

jdubois:
Duplicate of #398

Issue #907, opened 2014-12-26
Any plans on including lombok?


theborakompanioni:
Hey there. The one thing I have to do for all of my jhipster-based projects is to include lombok (http://projectlombok.org/) - it just speeds up development while improving the readability of the source. Any thoughts on including it? Would a PR be welcomed?

PR #3751, opened 2016-06-21
Add an entity option to use fluent setters

cbornet:
Because fluent is much nicer !

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -59,6 +59,13 @@ module.exports = EntityGenerator.extend({
             type: String
         });
 
+         // This method adds support for a `--[no-]fluent-methods` flag
+        this.option('fluent-methods', {
+            desc: 'Generate fluent methods in entity beans to allow chained object construction',
+            type: Boolean,
+            defaults: false
+        });
+
         // This adds support for a `--angular-suffix` flag
         this.option('angular-suffix', {
             desc: 'Use a suffix to generate AngularJS routes and files, to avoid name clashes',
@@ -81,6 +88,7 @@ module.exports = EntityGenerator.extend({
         });
 
         this.regenerate = this.options['regenerate'];
+        this.fluentMethods = this.options['fluent-methods'];
         this.entityTableName = this.options['table-name'] || this.name;
         this.entityNameCapitalized = _.upperFirst(this.name);
         this.entityTableName = _.snakeCase(this.entityTableName).toLowerCase();
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -529,6 +529,10 @@ module.exports = EntityGenerator.extend({
                     relationship.otherEntityRelationshipNamePlural = pluralize(relationship.otherEntityRelationshipName);
                 }
 
+                if (_.isUndefined(relationship.otherEntityRelationshipNameCapitalized) && (relationship.relationshipType === 'one-to-many' || (relationship.relationshipType === 'many-to-many' && relationship.ownerSide === false) || (relationship.relationshipType === 'one-to-one'))) {
+                    relationship.otherEntityRelationshipNameCapitalized = _.upperFirst(relationship.otherEntityRelationshipName);
+                }
+
                 if (_.isUndefined(relationship.otherEntityNamePlural)) {
                     relationship.otherEntityNamePlural = pluralize(relationship.otherEntityName);
                 }
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -209,6 +209,7 @@ module.exports = EntityGenerator.extend({
         this.changelogDate = this.fileData.changelogDate;
         this.dto = this.fileData.dto;
         this.service = this.fileData.service;
+        this.fluentMethods = this.fileData.fluentMethods;
         this.pagination = this.fileData.pagination;
         this.javadoc = this.fileData.javadoc;
         this.entityTableName = this.fileData.entityTableName || _.snakeCase(this.name).toLowerCase();
@@ -587,6 +588,7 @@ module.exports = EntityGenerator.extend({
             insight.track('entity/pagination', this.pagination);
             insight.track('entity/dto', this.dto);
             insight.track('entity/service', this.service);
+            insight.track('entity/fluentMethods', this.fluentMethods);
         }
     },
     writing : {
--- a/generators/entity/templates/src/main/java/package/domain/_Entity.java
+++ b/generators/entity/templates/src/main/java/package/domain/_Entity.java
@@ -255,7 +255,8 @@ public class <%= entityClass %> implements Serializable {
         relationshipNameCapitalized = relationships[idx].relationshipNameCapitalized,
         relationshipNameCapitalizedPlural = relationships[idx].relationshipNameCapitalizedPlural,
         otherEntityName = relationships[idx].otherEntityName,
-        otherEntityNamePlural = relationships[idx].otherEntityNamePlural;
+        otherEntityNamePlural = relationships[idx].otherEntityNamePlural,
+        otherEntityRelationshipNameCapitalized = relationships[idx].otherEntityRelationshipNameCapitalized;
     %><% if (relationshipType == 'one-to-many' || relationshipType == 'many-to-many') { %>
     public Set<<%= otherEntityNameCapitalized %>> get<%= relationshipNameCapitalizedPlural %>() {
         return <%= relationshipFieldNamePlural %>;
@@ -264,6 +265,12 @@ public class <%= entityClass %> implements Serializable {
     <%_ if (fluentMethods) { _%>
     public <%= entityClass %> <%= relationshipFieldNamePlural %>(Set<<%= otherEntityNameCapitalized %>> <%= otherEntityNamePlural %>) {
         this.<%= relationshipFieldNamePlural %> = <%= otherEntityNamePlural %>;
+        return this;
+    }
+
+    public <%= entityClass %> add<%= otherEntityNameCapitalized %>(<%= otherEntityNameCapitalized %> <%= otherEntityName %>) {
+        this.<%= relationshipFieldNamePlural %>.add(<%= otherEntityName %>);
+        <%= otherEntityName %>.set<%= otherEntityRelationshipNameCapitalized %>(this);
         return this;
     }
     <%_ } _%>
--- a/generators/entity/templates/src/main/java/package/domain/_Entity.java
+++ b/generators/entity/templates/src/main/java/package/domain/_Entity.java
@@ -211,6 +211,17 @@ public class <%= entityClass %> implements Serializable {
         return <%= fieldName %>;
     }
 
+    <%_ if (fluentMethods) { _%>
+        <%_ if (fieldTypeBlobContent != 'text') { _%>
+    public <%= entityClass %> <%= fieldName %>(<%= fieldType %> <%= fieldName %>) {
+        <%_ } else { _%>
+    public <%= entityClass %> <%= fieldName %>(String <%= fieldName %>) {
+        <%_ } _%>
+        this.<%= fieldName %> = <%= fieldName %>;
+        return this;
+    }
+    <%_ } _%>
+
     <%_ if (fieldTypeBlobContent != 'text') { _%>
     public void set<%= fieldInJavaBeanMethod %>(<%= fieldType %> <%= fieldName %>) {
     <%_ } else { _%>
@@ -223,6 +234,13 @@ public class <%= entityClass %> implements Serializable {
     public String get<%= fieldInJavaBeanMethod %>ContentType() {
         return <%= fieldName %>ContentType;
     }
+
+    <%_ if (fluentMethods) { _%>
+    public <%= entityClass %> <%= fieldName %>ContentType(String <%= fieldName %>ContentType) {
+        this.<%= fieldName %>ContentType = <%= fieldName %>ContentType;
+        return this;
+    }
+    <%_ } _%>
 
     public void set<%= fieldInJavaBeanMethod %>ContentType(String <%= fieldName %>ContentType) {
         this.<%= fieldName %>ContentType = <%= fieldName %>ContentType;
@@ -243,12 +261,26 @@ public class <%= entityClass %> implements Serializable {
         return <%= relationshipFieldNamePlural %>;
     }
 
+    <%_ if (fluentMethods) { _%>
+    public <%= entityClass %> <%= relationshipFieldNamePlural %>(Set<<%= otherEntityNameCapitalized %>> <%= otherEntityNamePlural %>) {
+        this.<%= relationshipFieldNamePlural %> = <%= otherEntityNamePlural %>;
+        return this;
+    }
+    <%_ } _%>
+
     public void set<%= relationshipNameCapitalizedPlural %>(Set<<%= otherEntityNameCapitalized %>> <%= otherEntityNamePlural %>) {
         this.<%= relationshipFieldNamePlural %> = <%= otherEntityNamePlural %>;
     }<% } else { %>
     public <%= otherEntityNameCapitalized %> get<%= relationshipNameCapitalized %>() {
         return <%= relationshipFieldName %>;
     }
+
+    <%_ if (fluentMethods) { _%>
+    public <%= entityClass %> <%= relationshipFieldName %>(<%= otherEntityNameCapitalized %> <%= otherEntityName %>) {
+        this.<%= relationshipFieldName %> = <%= otherEntityName %>;
+        return this;
+    }
+    <%_ } _%>
 
     public void set<%= relationshipNameCapitalized %>(<%= otherEntityNameCapitalized %> <%= otherEntityName %>) {
         this.<%= relationshipFieldName %> = <%= otherEntityName %>;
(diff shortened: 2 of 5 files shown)


jdubois:
The entity one of the core JHipster object, can we discuss this before? I'm not familiar with having a "fluent" setter on a domain object - I see the point, of course, but I've never seen those used on JPA entities, and wonder if there are any drawbacks or issues. Besides, I'm not sure everyone will want those (as I've never seen those used for domain objects), so that would mean dead code for a lot of people. Of course, if there are only advantages, we could also evangelize people to use those more!

deepu105:
Either we should use normal setters or fluent setters, having both might be lot of dead code. I don't see any downside in having fluent setters that follow the same naming convention like below
```
class Foo {

  private String bar;

  public String getBar() {
    return bar;
  }

  public Foo setBar(String bar) {
    this.bar = bar;
    return this;
  }
}
```

cbornet:
I think setBar can lead to issues with JavaBean reflection so bar(String bar) would be better.

cbornet:
I thought hibernate would complain if it didn't have the void setBar() method for its reflection but it seems to be OK so maybe we can remove it.

deepu105:
@cbornet what about mapstruct does it play well with this syntax

deepu105:
btw what issue do you foresee with JavaBean reflection?

cbornet:
I don't know, something I read somewhere...
Anyway, we can test with setBar() and if it works, it's just a matter of preference between setBar() or bar().

cbornet:
As for mapstruct, I didn't test directly but I guess that it was tested OK by CI.

jdubois:
you need `setBar` otherwise you are not following the JavaBean convention. Speaking of this, you should also return `void`, so I'm really not sure we should do this.

cbornet:
That's why I duplicated the method. But on in other hand, I removed the ```void setBar``` method and both Jackson and Hibernate seem to be fine with it.

cbornet:
To show the interest of fluent setters, compare
```java
Foo foo = new Foo();
foo.setMyString("foo");
foo.setMyString2("bar");
fooRepository.save(foo);
```
with
```java
fooRepository.save( new Foo().myString("foo").myString2("bar") );
```
This is a simple example but it gets more and more interesting when structures get complex. Consider
```java
Foo foo = new Foo();
foo.setMyString("test");
fooRepository.save(foo);

Bar bar = new Bar();
bar.setMystr("bar");
bar.setFoo(foo);
barRepository.save(bar);
```
vs
```java
barRepository.save( new Bar()
    .foo(fooRepository.save(new Foo()
        .myString("foo1")
        .myString2("foo2")))
    .mystr("bar"));
```
With the fluent version, it almost reads as YAML !

Note: the repository ```save``` can probably be removed if the entities have CascadeType.ALL on their relationships.
@jdubois Is there a reason why we don't set the relationships as CascadeType.ALL by default ?

jdubois:
- OK the fluent interface looks better, then I've never seen anyone using it -> that's why it worries me, and I wonder if there is a technical issue somewhere
- For CascadeType.ALL this is because of our [Policy 1](https://jhipster.github.io/policies/). Anyway here you have a different issue: you need to `save()` because that's a new entity, so it's non-managed. Otherwise it would work automatically (at least without the fluent interface, it works!!)

cbornet:
I thought CascadeType.ALL (or persist) was used so that the entity is automatically added to the session. I need to check that again.

cbornet:
If there is no JavaBean setter, Hibernate [uses properties directly even if they are private](http://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html_single/) (unlike standard JPA) and Jackson will [both serialize and deserialize](http://www.baeldung.com/jackson-field-serializable-deserializable-or-not) (as long as there is a getter).
Now for the other frameworks that use the domain (mongo, ES, cassandra, liquibase ...) , I don't know the impact of not having a Javabean setter but it seems the CI tests where good.

My opinion would be to have both ```void setBar()``` and ```bar()``` generated to avoid issues with libraries that expect to have the JavaBean setter. JavaBean reflection will just ignore the ```bar()``` method.

cbornet:
In java.beans.Introspector
```java
                       } else if (void.class.equals(resultType) && name.startsWith(SET_PREFIX)) {
                            // Simple setter
                            pd = new PropertyDescriptor(this.beanClass, name.substring(3), null, method);
                            if (throwsException(method, PropertyVetoException.class)) {
                                pd.setConstrained(true);
                            }
                        }
```
so the void result type is checked before adding the method to the PropertyDescriptor list.
Also a non-javabean method will just be ignored so it's safe to add them.

jdubois:
- I'll let you have a look at `CascadeType.ALL` :-)
- Then we are going to have both `setBar()` and `bar()`, and once again maybe it can cause issues (JPA, MapStruct, Jackson... so many stuff use the domain object), and this is going to be dead code for a lot of people, so this is bad for them. At best this could be an option.

cbornet:
I tested that
```
PropertyDescriptor descriptor = new PropertyDescriptor("bar", Foo.class);
```
fails with
```
java.beans.IntrospectionException: Method not found: setBar
```
If the void setBar method is not present

cbornet:
It's not really dead code : it's code that people may not use (even if they should :smile: ) and it's in generated code so it's a bit like a library : you don't really need to go see the internal code often.

But if you prefer to do it as an option or module I understand. Please tell me which one in that case.

cbornet:
This reminds me that I wanted to add adders and removers on relationships Set. So that could be part of the package.

jdubois:
It's just that people keep telling me JHipster generates too much code, and that they don't understand what is being generated.
Of course, you could argue that if they don't understand the code, they shouldn't say there is too much code :-)

cbornet:
So they want awesome features but no code. Like in magic ?

If you're ok, I'll do it as a silent option (no question asked) so that it can be configured in jhipster-uml/JDL

deepu105:
--option might be good. So that people who knows and uses these stuff can
enable them when required

Thanks & regards,
Deepu

generators/entity/index.js
```diff
@@ -207,6 +207,7 @@ module.exports = EntityGenerator.extend({
         this.changelogDate = this.fileData.changelogDate;
         this.dto = this.fileData.dto;
         this.service = this.fileData.service;
+        this.fluentMethods = this.fileData.fluentMethods;
```
deepu105: may be you can add `--fluentMethods` as an option to sub generator as well when creating entities by hand.
dont forget to PR the JDL project as well

cbornet:
Yes, I'll add the command line option.

cbornet:
Done. I'll add it to the JDL once this is merged.

deepu105:
:+1:

jdubois:
Then don't forget to document it on jhipster.github.io -> this is indeed a very nice option if you know what it does, and I'm sure a lot of people will like it
And can you also add "insight" statistics -> if a lot of people use it, we can then convert it to a new question

cbornet:
Yup, will do. As for the insight, I have already added [this](https://github.com/cbornet/generator-jhipster/blob/3a760a472d4fa6e0184dd390005754cd09ae644d/generators/entity/index.js#L601). Is there something missing ?

Also what should be the semantic for the JDL ? ```fluent-methods A,B ``` ?

deepu105:
@cbornet I think it should be `fluentMethods A,B` follow the syntax for `skipClient `

cbornet:
ALse I just tried
```java
        Foo foo = new Foo()
            .myString("test")
            .addBar(new Bar().mystr("bar1"))
            .addBar(new Bar().mystr("bar2"));
        fooRepository.save(foo);
```
With a CascadeType.ALL on the OneToMany and the Bar entities where added to the DB.
If I don't put CascadeType.ALL, they are not.
So it does add the entities automatically without the need to call save.

cbornet:
@deepu105 noted !

jdubois:
For `CascadeType.ALL`-> yes, that's cascading the "many" entities, then that's because those are not already managed

I'd rather keep this with the default: I'd rather people learn how it works, than doing a surprise to people who already knows how it works

cbornet:
No problem, it's not a big work to edit it by hand when you need it anyway.

Note : I should have wrote
```java
        fooRepository.save( new Foo()
            .myString("test")
            .addBar(new Bar().mystr("bar1"))
            .addBar(new Bar().mystr("bar2")));
```
:wink:

ghahramani:
@jdubois Actually I converted all the jhipster setter to fluent style :D with no problems. I like this style it gets code much cleaner

cbornet:
Remove "needs-discussion" label ?

jdubois:
Yes let's do this, I agree with you!

jdubois:
@cbornet maybe you could use those in some generated tests, so that people see how it should works? And if that's good, we could put this option as "true" by default... As we should push best practices to users, and that's definitely a good way to use the entities. My only concern is that we would need a lot of testing, to check if this works with all options (including MapStruct, Cassandra...)

cbornet:
I have tested that javaBean reflection works correctly (both reading/writing) so if those libs use it they should be OK.
At the beginning I had done it in the tests but since this is now an option, I removed it. Now,  I could check the fluentMethods flag and generate the test accordingly.

cbornet:
I have added the tests and now it's cross-tested by CI with Mongo and Cassandra.
I haven't added a cross-test with mapstruct but it is easy to do if needed.

jdubois:
No need to rush -> if we do a release tomorrow, it's a bit early to merge this anyway.
But indeed I'd like to have this tested also with MapStruct. And we need to have MapStruct out of beta at some point :-)

jdubois:
OK, this all looks good to me.

Now, 2 things:

- we need more testing -> but this can be done after merging this
- do we put this by default or not? I like the fact that we provide "best practices" and push them, then I don't know how people will react to this -> I'll do a Twitter survey and mention @cbornet

jdubois:
And the vote has started at https://twitter.com/java_hipster/status/760504585748447232

vladmihalcea:
Although the JPA specification forbids Java Bean conventions violation, Hibernate is not so strict. For instance, you can run [this test](https://github.com/vladmihalcea/high-performance-java-persistence/blob/master/core/src/test/java/com/vladmihalcea/book/hpjp/hibernate/mapping/FluentSettersTest.java) just fine. As long as you rely only on Hibernate, you should be fine. However, if you plan on supporting other JPA providers, the Fluent API might not work, so keep that in mind as well.

jdubois:
Thanks a lot @vladmihalcea !!! I didn't know it was a JPA spec violation, and in that case that's quite a big issue to me. It would also go against our [Policy 1](http://jhipster.github.io/policies/).

vladmihalcea:
It's the **2.2 Persistent Fields and Properties** section of the Java Persistence 2.1 spec that says:

> In this case, for every persistent property _property_ of type T of the entity, there is a getter method, _getProperty_, and setter method _setProperty_. For boolean properties, _isProperty_ may be used as an alternative name for the getter method.[2]
> 
> For single-valued persistent properties, these method signatures are:
> - T getProperty()
> - void setProperty(T t)

Now, the spec needs to be generic in order to allow interoperability. In reality, JPA implementations might relax some requirements to provide a better developer experience. 

Since you want to support any JPA provider, you might want to contact all major implementation and validate that the Fluent API is not breaking anything.

jdubois:
Thanks again @vladmihalcea 
No we don't intend to support all JPA providers (just Hibernate, in fact), but we try to follow the "standard" usage of each technology (and so follow JPA, here).
I'll wait until the vote is over: at the moment we have 70% votes for the "fluent setters"!

cbornet:
The PR doesn't remove the java bean methods and it must not as some frameworks may use java bean reflection. The fluent methods are added methods with different signature so they don't disturb the reflection.

vladmihalcea:
But you cannot have two setter methods whose signature only differ in the return valuue type. How are you going to address that?

cbornet:
They differ in name : ```void setBar()``` vs ```Foo bar()```.

vladmihalcea:
Seems legit.

jdubois:
So the vote is over, 116 votes, 72% for the "fluent setters".
Besides, we had a nice blog entry by @vladmihalcea about (see it [here](https://vladmihalcea.com/2016/08/09/fluent-api-entity-building-with-jpa-and-hibernate/)), and this was even tweeted by the official Java account (see it [here](https://twitter.com/java/status/763035625800949761)). So an overall very clear vote for this.

- let's do this
- let's not put this as an option: it adds more maintenance cost, and there is no big issue in doing this (except extra code for non-hipsters who want to use the classical setters)

cbornet:
@jdubois great ! As you know I'm currently on the beach drinking mojitos without any keyboard nearby so can you take the PR from here ?

jdubois:
Yes of course. I'm the one who broke it anyway!

jdubois:
I've changed my mind after reading the code: I'm keeping the "fluent-method" option, for backward compatibility (and if people are unhappy with this), but I'm putting it at "true" by default.
It's a very simple switch, so no big maintenance cost in the end.

jdubois:
I found a couple of small issues, I'll finish this tomorrow!

jdubois:
I'm going to merge this, just a couple of issues I'd like to share before.

There was an "add*" method for adding an entity to a collection (for one-to-many and many-to-many relationships), so I added also a "remove*" method:

- Those "add*" and "remove*" methods forced me to add a new question in many-to-many relationships, when the entity is the owner -> you need to know the name of the relationship in the other entity. We were already asking this question for other use cases, and this is the "otherEntityRelationshipName" parameter.
- Those 2 methods also kind of force to have bi-directional relationships -> if you don't want the bi-directional relationship, there is a couple of more lines to delete

So for me this needs more testing and practice.

cbornet:
I totally forgot about unidirectional relationships ! We should detect them and not set the remote field for these.

jdubois:
Yes and I broke the build because of this... There must one edge case I missed.

This is quite annoying, I'm wondering if that's the right approach. I'm first correcting the build, but then we must have a look at this, and its consequences.

jdubois:
In fact most of the errors where in our `.json` test files, which where not correct :-)

I've also added [a ticket to JHipster Core](https://github.com/jhipster/jhipster-core/issues/51) so this is included in the JDL!

cbornet:
@jdubois I wanted to have a look at unidirectional relationships and I couldn't see any issue.
For unidir ManyToOne, the method is not generated.
And unidir OneToMany are not allowed in jhipster.
Am I missing something ?

jdubois:
@cbornet I didn't try it after my last comment about it, and the code has changed since, so maybe there isn't any issue after all. If you tried and it works, that's perfect!

cbornet:
Ok. I'll check when I come back.
Also I think there might still be an issue in ManyToMany : in the adder method we should add to the remote collection instead of calling its setter. Can someone have a look ?

cbornet:
Oh ! You've already fixed that @jdubois ! 👍  😄

jdubois:
I don't remember everything I've fixed :-)
But yes there was some serious work on this part! And I'm still not 100% sure it's perfect

dhaalves-uel:
Check project Lombok https://projectlombok.org/features/index.html
This feature and more can be achieved with simple annotations

cm325:
I was actually kind of suprised that jhipster wasn't already using lombok, seems like that's what all the cool kids are using these days ;)  I personally understand the reasons against it, but I've seen it in use in lots of places-

pascalgrimaud:
@dhaalves-uel @cm325 : you can look at older issues like #398 #907 #1334 #2734 #3476

deepu105:
@jdubois @cbornet sorry to beat around the dead horse but just wanted to ask. Since its bean a while that we did this and since no body complains should we make fluent setters the default with `public Bar setFoo()` syntax and remove the duplicate methods? 
I was looking at [this](https://github.com/jhipster/generator-jhipster/commit/2cd31621e76e872a5edf4d44405ec4bbb32ddd64) and thought it would have been nice to have fluent style everywhere

jdubois:
I agree with you, but we don't know what people use: maybe most users still use setters... So, in doubt, I would keep them, even if I agree they should be removed

deepu105:
@jdubois we are tracking it `insight.track('entity/fluentMethods', this.fluentMethods);` so may be we can check if anybody has turned it off and also for user class and other stuff we can start using fluent style.
For those who are using setters already they wouldn't see any difference if we start returning the object in that setter I guess

jdubois:
I just checked, we have 0,01% of people not using fluent methods... If it doesn't do any harm (I haven't checked), let's remove this, OK.
But can we first do the Angular 4 code? This is less important

cbornet:
Note : if we remove the setters then we don't respect the javabean contract anymore. Is it not an issue ? (Hibernate and Jackson don't rely on javabean. I think neither does mapstruct)

cbornet:
What we can do is to have fluent setters by default that you can deactivate with --no-fluent-methods. And refactor the User classes.
@jdubois I can't really help on Angular (me back-end type 😅 ) so I can take this if you want.

cbornet:
We also need to cover the UserMapper : our coverage decreased 😭

jdubois:
@cbornet aren't they already fluent by default?

cbornet:
@jdubois indeed !

deepu105:
ofcourse I wanst suggesting we do it now. I just wanted to mention so that I dont forget. I mean we could create only fluent by default and the flag can be used to create non fluent if people want and for user classes we can have only fluent as they are jhipster domain classes. Anyways its not urgent at all.

PR #3755, opened 2016-06-22
Account module imports, fix console errors

wmarques:
Did some changes in systemJS config to avoid `not found `error.
Imported missing necessary services in `account` module to get the homepage work.
Imported missing controllers (`Home` and `Navbar` ones).

So now we have a homepage working ! without CSS but well, it's a first start haha

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/components/_common.module.ts
+++ b/generators/client-2/templates/src/main/webapp/app/components/_common.module.ts
@@ -2,9 +2,6 @@ import { upgradeAdapter } from '../upgrade_adapter';
 
 import { AlertServiceConfig } from '../blocks/config/alert.config';
 
-import { HomeController } from '../home/home.controller';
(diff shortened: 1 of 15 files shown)


generators/client-2/templates/src/main/webapp/app/components/_common.module.ts
```diff
@@ -21,12 +27,16 @@ angular
         'ui.bootstrap',
         'ui.router'
     ])
-    .config(AlertServiceConfig)
+    // bug 'showAsToast is not a function to fix'
+    //.config(AlertServiceConfig)
+    .controller('HomeController', HomeController)
```
deepu105: Can we move home and navbar to app.module.ts ?

wmarques: done !

PR #3756, opened 2016-06-22
Polish

pascalgrimaud:


deepu105:
you could have pushed this directly since its just polish :wink:

PR #3766, opened 2016-06-24
Make Swagger UI available in prod with a new swagger profile

### LINKED ISSUE #3402: Swagger UI not accessible in prod profile even though swagger is enabled

In prod profile, it is now possible to enable swagger with jhister.swagger.enabled: true. However Swagger UI cannot be accessed.

Currently, in prod profile with swagger enabled, it is lacking both the "API" link in the admin interface and the UI in itself.

capture du 2016-04-15 15-27-03

Enable access to API docs in prod profile. It would be especially nice to have for the gateway as it could then be used as an API Portal.

PierreBesson:
Fix #3402
- Add a gulp task to add swagger-ui assets to the build
- Remove the no-swagger profile
- Add a swagger profile that uses property substitution to selectively enable swagger docs. Moreover this is used in the front-end to display/hide the "API" link.

- The dev profile trigger automatically the swagger profile (using spring.profiles.include)
- Fixes the current behavior where swagger ui was not avaible in dev when not using browsersync


generators/client/index.js
```diff
@@ -315,6 +315,7 @@ module.exports = JhipsterClientGenerator.extend({
         writeSwaggerFiles: function () {
             // Swagger-ui for Jhipster
             this.template(MAIN_SRC_DIR + 'swagger-ui/_index.html', MAIN_SRC_DIR + 'swagger-ui/index.html', this, {});
+            this.template(MAIN_SRC_DIR + 'swagger-ui/_index-prod.html', MAIN_SRC_DIR + 'swagger-ui/index-prod.html', this, {});
```
deepu105: is this same as the `index.html` file or is there any difference? i didnt compare them

generators/server/templates/src/main/resources/config/_application-prod.yml
```diff
@@ -216,5 +216,5 @@ jhipster:
             enabled: false
             # edit spring.metrics.export.delay-millis to set report frequency
     <%_ } _%>
-    swagger: # swagger is disabled. It can be disabled by pasing 'no-swagger' profile at run time as well
-        enabled: false
+    swagger: # swagger is disabled by default in prod, it can be enabled by passing the 'swagger' profile at run time
+        enabled: #swagger.enabled#
```
deepu105: where are you setting this from maven/gradle?

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -224,5 +226,5 @@ jhipster:
             enabled: false
             # edit spring.metrics.export.delay-millis to set report frequency
     <%_ } _%>
-    swagger: # swagger is enabled. It can be disabled by pasing 'no-swagger' profile at run time as well
+    swagger: # swagger is enabled by default in dev
         enabled: true
```
deepu105: if this can be controlled purely by profiles then no point having this we can remove it

generators/client/index.js
PierreBesson: The differences are in the javascript import not passing by bower_components. I agree that this could be improved maybe by refactoring the common js code.

generators/server/templates/src/main/resources/config/_application-prod.yml
PierreBesson: Oh I see, this is not right... Should I be setting that in code triggered by the swagger profile ?

generators/client/index.js
deepu105: may be you could use the `inject` plugin to do this dynamically based on profile.

generators/server/templates/src/main/resources/config/_application-prod.yml
deepu105: if you want this to be set, then you need to add it to gradle and pom in same way the default profiles are set. But since now you can do this purely with profile may be this doesn't have any use

deepu105:
dont forget to PR this to the angular-2 branch for the client-2 sub as well

PierreBesson:
I removed the unneeded property. @deepu105 thank you so much for your feedback

I'm considering using gulp-inject for dealing with swagger-ui's index.html dependencies but it seems that is will overly complicate the current gulpfile just for swagger.

deepu105:
ya you are right

deepu105:
may be we can do some simple regex magic when copying the file for prod

PierreBesson:
@deepu105 just did that with success !
My PR is now complete.

jdubois:
Great - @PierreBesson can you also update the documentation on the website?

PierreBesson:
I need to do this, as well as do a PR for the ng2 branch.

Issue #3761, opened 2016-06-22
Sub generator upgrade: issue with upgrading projects generated prior JHipster 3.4.0


Invizible:
##### **Overview of the issue**

Upgrade sub generator does not work correctly if project is generated with jhipster prior version 3.4.0.
I have a project generated with version 3.0.0. So when I wanted to upgrade it, I updated jhipster

`npm i -g generator-jhipster`

So that upgrade sub generator became available and when I ran it , sub generator created a new branch 'jhipster_upgrade', checked out on this new branch and ran 
`jo jhipster --with-entities`

BUT, since we don't have jhipster installed locally (folder is empty when sub generator checks out on the upgrade branch), yeoman uses global jhipster (3.4.2 for example) and generates project not with 3.0.0 but with current version 3.4.2, for example. Then it upgrades jhipster and generates project again but still with the same version 3.4.2 (because it is the latest). It this situation we don't have a difference between 2 upgrade commits, and there is nothing to merge master branch with. So upgrade does not happen.

##### **Motivation for or Use Case** 

Upgrade sub generator should also work for projects generated with version >=3.0.0

##### **JHipster Version(s)** 

3.0.0.

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.0.0",
    "baseName": "app",
    "packageName": "com.app",
    "packageFolder": "com/app",
    "serverPort": "8090",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "postgresql",
    "searchEngine": "no",
    "buildTool": "gradle",
    "enableSocialSignIn": true,
    "rememberMeKey": "60205dc06b687c6b6afd12dda7cc87dc26c08a39",
    "useSass": false,
    "applicationType": "monolith",
    "testFrameworks": [],
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "ru"
    ]
  }
}
```

##### **Browsers and Operating System** 

Windows 10

##### **Reproduce the error** 

1. generate project with jhipster version prior to 3.4.2 (3.0.0, for example)
2. update jhipster version: `npm i -g generator-jhipster`
3. run: `yo jhipster:upgrade`
[...]

PR #3770, opened 2016-06-24
Fix for issue with upgrading projects generated prior JHipster 3.4.0 #3761

Invizible:
I've added a task which installs jhipster locally right before the app generation, so now when you run 
`yo jhipster:upgrade`
it will:

1. check out on the jhipster_upgrade branch
2. clean up the folder (_cleanUp task)
3. install jhipster with needed versioin
4. and then run _generate task, which will use installed locally jhipster and generate correct app

So now upgrade process is independent of the installed globally version and we can upgrade projects generated before 3.4.0. And it also fixes issues when we have several projects generated with different jhipster versions.

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -158,7 +158,7 @@ module.exports = UpgradeGenerator.extend({
                 this.log('Installing JHipster ' + version + ' locally');
                 shelljs.exec('npm install ' + GENERATOR_JHIPSTER + '@' + version, {silent:true}, function (code, msg, err) {
                     if (code === 0) this.log(chalk.green('Installed ' + GENERATOR_JHIPSTER + '@' + version));
-                    else this.error('Something went wrong while installing generator! ' + msg + ' ' + err);
+                    else this.error('Something went wrong while installing the JHipster generator! ' + msg + ' ' + err);
                     callback();
                 }.bind(this));
             }.bind(this);
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -45,6 +45,15 @@ module.exports = UpgradeGenerator.extend({
         }
     },
 
+    _installJhipsterLocally: function (version, callback) {
+        this.log('Installing JHipster ' + version + ' locally');
+        shelljs.exec('npm install ' + GENERATOR_JHIPSTER + '@' + version, {silent:true}, function (code, msg, err) {
+            if (code === 0) this.log(chalk.green('Installed ' + GENERATOR_JHIPSTER + '@' + version));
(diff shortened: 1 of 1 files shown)


Invizible:
I've tested fix on two projects 3.0.0 and 3.4.0

deepu105:
This makes the sub gen even slower as there will be 2 npm installs. Check my comment on the issue. I dont think we should do this

generators/upgrade/index.js
```diff
@@ -155,12 +154,24 @@ module.exports = UpgradeGenerator.extend({
                 }.bind(this));
             }.bind(this);
 
+            var installJhipsterLocally = function(version, callback) {
+                this.log('Installing JHipster ' + version + ' locally');
+                shelljs.exec('npm install ' + GENERATOR_JHIPSTER + '@' + version, {silent:true}, function (code, msg, err) {
+                    if (code === 0) this.log(chalk.green('Installed ' + GENERATOR_JHIPSTER + '@' + version));
+                    else this.error('Something went wrong while installing generator! ' + msg + ' ' + err);
```
jdubois: "while installing the JHipster generator"

jdubois:
I don't know this part well enough -> @lordlothar99 what do you think about this? I see @deepu105 is against it, and personally I don't mind if the upgrade process doesn't work on pre-3.4.0 versions

lordlothar99:
yeah we briefly talked about this in https://github.com/jhipster/generator-jhipster/issues/3761, I wish I could have come on monday to talk about it... 
 in my opinion this would bring value to the upgrader cause there are plenty of projects generated before 3.4.0, and I know guys who told me they were expecting that ; on the other hand I understand @deepu105 argument about performance --> that's why @Invizible refactored a little the code so it wouldn't waste time for "normal" projects. so imo we should  merge this ! @deepu105 what do u think ?

deepu105:
Im ok as long as it doesn't add too much overhead or install time for all

Thanks & regards,
Deepu

lordlothar99:
Yeah!  :sunglasses:

Invizible:
So, are we good with this?

deepu105:
Ok then if everyone is ok with this then Im ok.

generators/upgrade/index.js
deepu105: @Invizible can you take care of this?

deepu105:
@Invizible i will merge this as soon as you take care of the comment from @jdubois . its a minor wording change

Invizible:
@deepu105  OK, will update PR

Invizible:
@deepu105 done

PR #3772, opened 2016-06-26
move oauth2 client details into database

wongdao:
The previous behavior to config oauth2 client details is just using hardcode. This PR will drop that hardcode and store the oauth2 client details into database.

The details of this PR are:

 1. configure spring security oauth2 with jdbc
 2. configure liquibase to insert oauth2 client details into database

NOTE: this PR **only** modified the sql part, the mongodb part is still using the hardcode.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
@@ -22,17 +22,44 @@ import org.springframework.security.oauth2.config.annotation.web.configurers.Aut
 import org.springframework.security.oauth2.config.annotation.web.configurers.AuthorizationServerSecurityConfigurer;
 
 import org.springframework.security.oauth2.provider.token.TokenStore;<% if (databaseType == 'sql') { %>
+import org.springframework.security.oauth2.config.annotation.web.configurers.ResourceServerSecurityConfigurer;
+import org.springframework.security.oauth2.provider.approval.ApprovalStore;
+import org.springframework.security.oauth2.provider.approval.JdbcApprovalStore;
+import org.springframework.security.oauth2.provider.code.AuthorizationCodeServices;
+import org.springframework.security.oauth2.provider.code.JdbcAuthorizationCodeServices;
 import org.springframework.security.oauth2.provider.token.store.JdbcTokenStore;<% } %>
 
 import javax.inject.Inject;<% if (databaseType == 'sql') { %>
 import javax.sql.DataSource;<% } %>
 
 @Configuration
-public class OAuth2ServerConfiguration {
+public class OAuth2ServerConfiguration {<% if (databaseType == 'sql') { %>
+
+    @Inject
+    private DataSource dataSource;
+
+    @Bean
+    public TokenStore tokenStore() {
+        return new JdbcTokenStore(dataSource);
+    }<% } %><% if (databaseType == 'mongodb') { %>
+
+    @Inject
+    private OAuth2AccessTokenRepository oAuth2AccessTokenRepository;
+
+    @Inject
+    private OAuth2RefreshTokenRepository oAuth2RefreshTokenRepository;
+
+    @Bean
+    public TokenStore tokenStore() {
+        return new MongoDBTokenStore(oAuth2AccessTokenRepository, oAuth2RefreshTokenRepository);
+    }<% } %>
 
     @Configuration
     @EnableResourceServer
-    protected static class ResourceServerConfiguration extends ResourceServerConfigurerAdapter {
+    protected static class ResourceServerConfiguration extends ResourceServerConfigurerAdapter {<% if (databaseType == 'sql') { %>
+
+        @Inject
+        private TokenStore tokenStore;<% } %>
 
         @Inject
         private Http401UnauthorizedEntryPoint authenticationEntryPoint;
@@ -70,6 +97,11 @@ public class OAuth2ServerConfiguration {
                 .antMatchers("/swagger-resources/configuration/ui").permitAll()
                 .antMatchers("/swagger-ui/index.html").hasAuthority(AuthoritiesConstants.ADMIN);
         }
+<% if (databaseType == 'sql') { %>
+        @Override
+        public void configure(ResourceServerSecurityConfigurer resources) throws Exception {
+            resources.resourceId("res_<%= baseName %>").tokenStore(tokenStore);
+        }<% } %>
     }
 
     @Configuration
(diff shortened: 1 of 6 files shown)


jdubois:
I'm not an expert in that part, but this looks like a good idea to me. I'd love to have some feedback from people who know about this, who could help me here?

cbornet:
This looks good. I think it can be tested easily with an external swagger-ui with some tweaks (add the authorization config in the springfox docket, configure the client-id/secret in the index.html, allow the cors, configure the redirect-uri? (not sure)). I'll see if I have time to review it.

jdubois:
@cbornet if you can review it, I'll let you merge it :-)

cbornet:
No problem. Two weeks until holidays but that should do it.

cbornet:
Maybe this could also benefit to UAA config ? @xetys ?

xetys:
i already thought about this, but i am not sure, if this fits also the microservice context.

While in monolith OAuth2 is used as instendend, to serve a mayve variating set of clients, in a microservice Oauth2 is primary used for internal security, so maybe some yml configuration is better for the developer rather than a heavy weight clientDetailsStore. 

For my internal stuff i really use some lines of `Configurer`, which makes thinks lighter, and got a overview, since there are currently no more then 4 clients...

generators/server/templates/src/main/resources/config/liquibase/oauth_client_details.csv
```diff
@@ -0,0 +1,2 @@
+client_id;resource_ids;client_secret;scope;authorized_grant_types;authorities;access_token_validity;refresh_token_validity;additional_information;autoapprove
+<%= baseName %>app;res_<%= baseName %>;my-secret-token-to-change-in-production;read,write;password,refresh_token,authorization_code,implicit;ROLE_ADMIN,ROLE_USER;1800;2000;{};true
```
cbornet: Client details should be removed from application.yml if database is sql

generators/server/templates/src/main/resources/config/liquibase/oauth_client_details.csv
```diff
@@ -0,0 +1,2 @@
+client_id;resource_ids;client_secret;scope;authorized_grant_types;authorities;access_token_validity;refresh_token_validity;additional_information;autoapprove
```
cbornet: Add the web_server_redirect_uri column so that users can easily configure it they want

cbornet:
```<column name="authentication" type="BLOB"/>``` is missing for table oauth_code in liquibase intitial_schema.xml. So the oauth code flow fails.
Otherwise the rest works fine :+1: 

@wongdao can you do the fix and check my comments ? Then this can be merged.

Tip to test the auth code flow with an external swagger:
* use swagger-ui > 2.1.5. It is not the version used currently by http://petstore.swagger.io/ but it shouldn't be long before they upgrade. Meanwhile download the master or tag 2.1.5 and run locally.
* allow the CORS for the swagger-ui server
* add ```your-client-id;res_oauth;your-client-secret-if-required;access;refresh_token,authorization_code,implicit;;ROLE_ADMIN,ROLE_USER;1800;2000;{};false``` to oauth_client_details.csv. These are default swagger-ui client-id/client-secret.
* add a docket with oauth security definitions configured
````java
        OAuth oauth = new OAuth("oauth", newArrayList(new AuthorizationScope("access", "")),
            newArrayList(new AuthorizationCodeGrant(
                new TokenRequestEndpoint("http://localhost:8080/oauth/authorize", "", ""),
                new TokenEndpoint("http://localhost:8080/oauth/token", "")
            )));

        SecurityContext securityContext =  SecurityContext.builder()
                .securityReferences(newArrayList(new SecurityReference("oauth",
                    new AuthorizationScope[]{new AuthorizationScope("access", "")})))
                .forPaths(regex(DEFAULT_INCLUDE_PATTERN))
                .build();

        Docket docket = new Docket(DocumentationType.SWAGGER_2)
            .groupName("oauth")
            .securitySchemes(newArrayList(oauth))
            .securityContexts(newArrayList(securityContext))
            ...
```
Note: the oauth is configured with only 1 scope because of https://github.com/swagger-api/swagger-ui/issues/1598
* In swagger-ui explore http://localhost:8080/v2/api-docs?group=oauth
* Then click on "Authorize" and follow the flow

wongdao:
@cbornet sure, I'll check&fix your comments.

wongdao:
@cbornet all your comments got fixed, please review them:

1. client details got removed in both `_application-dev.yml` and `_application-prod.yml` if database is sql
2. an empty `web_server_redirect_uri` column got inserted in `oauth_client_details.csv`

wongdao:
@cbornet unnecessary database type checking got removed in `_initial_schema.xml`, please review the latest commit.

cbornet:
Column ```<column name="authentication" type="BLOB"/>``` is still missing for table oauth_code in liquibase intitial_schema.xml. Can you add it ?

generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
```diff
@@ -70,40 +97,48 @@ public void configure(HttpSecurity http) throws Exception {
...
-        private OAuth2AccessTokenRepository oAuth2AccessTokenRepository;
-
-        @Inject
-        private OAuth2RefreshTokenRepository oAuth2RefreshTokenRepository;<% } %>
+        private DataSource dataSource;<% } %>
 
         @Inject
         private JHipsterProperties jHipsterProperties;
 
+        @Inject
```
cbornet: JHipsterProperties can be removed if database is sql.

wongdao:
@cbornet done add column `authentication` and JHipsterProperties of sql database type removing, see the latest commit please, also and reviewing.

generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
```diff
@@ -70,40 +97,48 @@ public void configure(HttpSecurity http) throws Exception {
...
-
-        @Inject
-        private OAuth2RefreshTokenRepository oAuth2RefreshTokenRepository;<% } %>
-
+        private DataSource dataSource;<% } %>
+<% if (databaseType != 'sql') { %>
         @Inject
         private JHipsterProperties jHipsterProperties;
-
+}<% } %>
```
cbornet: There is an extra ```}```. It fails the build in mongo + oauth2.

cbornet:
Thanks for the PR @wongdao. I'll do the fix for oauth + mongo in another PR.

Now any volunteer to do the MongoAuthorizationCodeServices/MongoApprovalStore/MongoClientDetailsService ? @jhipster/developers ?

generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
wongdao: sorry about that mistake, I got lost in the sql&jwt&mongo yesterday.
thanks for your patient reviewing.

cbornet:
I have a WIP for mongo. Not sure I'll have time to PR before holidays though...

cbornet:
I have it almost working for mongo. I will PR at my return.

PR #3774, opened 2016-06-26
add hibernate dependencies only for sql databases

### LINKED ISSUE #3768: Too many hibernate dependencies

It looks like when using ``mongo`` some ``hibernate`` dependencies are still added to ``build.gradle`` and ``pom.xml`` 

Nothing seems broken, but I think that is not correct. I have removed all hibernate dependencies from ``build.gradle`` and everything worked fine.

##### **JHipster Version(s)** 

All

##### **Reproduce the error** 

Create an app and select ``mongodb`` as database/

##### **Suggest a Fix** 

Move the ``sql`` check somewhere else. 

```xml
        <%_ } _%>
        <dependency>
            <groupId>org.hibernate</groupId>
            <artifactId>hibernate-envers</artifactId>
        </dependency>
        <dependency>
            <groupId>org.hibernate</groupId>
            <artifactId>hibernate-validator</artifactId>
        </dependency>
```

atomfrede:
additional cleanup in the gradle build file

close #3768

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -140,8 +140,8 @@ dependencies {
     }
     compile "com.fasterxml.jackson.datatype:jackson-datatype-json-org:${jackson_version}"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-hppc:${jackson_version}"
-    compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson_version}"
-    compile "com.fasterxml.jackson.datatype:jackson-datatype-hibernate4"
+    compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson_version}"<% if (databaseType == 'sql') { %>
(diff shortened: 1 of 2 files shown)


ghahramani:
@atomfrede Hi, I it for no-sql database. Would you please explain a little bit why did you do that? Don't we need to use `@Cache`, `@ElementCollection` or other `javax.persistence` annotations?

atomfrede:
Hi, 
javax annotations are provided by core Java. The concrete Implementation should be provided by e.g. Hibernate. In case of mongodb all should be provided by spring data. Do you have found a concrete problem when hibernate is not on the classpath ?

ghahramani:
@atomfrede Thanks for quick answer, Actually I'm not very familiar with mongodb but from my sql experience I always add `@ElementCollection    @MapKeyColumn(name = "name") ` to my property which is `Map<String,Object>` and `@Cache` for association property ( X-To-Y ), but I realize because neither hibernate nor spring-data are in gradle dependencies so I don't have this annotations.

PR #3776, opened 2016-06-26
add edit button to detail page

### LINKED ISSUE #3695: [FEATURE] Modify button in <entity>detail.html page

It would be usefull having a "Modify" button in <entity>-detail.html page.
I've dropped one with a simple

    <button type="submit" ui-sref="address.edit({id:vm.address.id})" class="btn btn-primary btn-sm">
		<span class="glyphicon glyphicon-pencil"></span>
		<span class="hidden-xs hidden-sm" translate="entity.action.edit"></span>
    </button>    

pressing the button displays the correct form in an overlay with the entities list behind.

It would be much more "natural" to navigate data and being able to modify what you are looking at.

##### **JHipster Version(s)** 
3.4.0

atomfrede:
* add new state for detail editing, such that the user can stay on the detail page
* don't use back anymore, because otherwise the edit dialog will open again
* therefore publish the previous state to the detail page (in case not found, the entity list is used)

close #3695

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -685,6 +685,7 @@ module.exports = EntityGenerator.extend({
             if (this.skipClient) {
                 return;
             }
+            
             this.copyHtml(ANGULAR_DIR + 'entities/_entity-management.html', ANGULAR_DIR + 'entities/' + this.entityFolderName + '/' + this.entityPluralFileName + '.html', this, {}, true);
             this.copyHtml(ANGULAR_DIR + 'entities/_entity-management-detail.html', ANGULAR_DIR + 'entities/' + this.entityFolderName + '/' + this.entityFileName + '-detail.html', this, {}, true);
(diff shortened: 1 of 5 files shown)


atomfrede:
Will check the travis errors today :(

jdubois:
I'm sure it would be useful, but my issue is that we're supposed to have a feature freeze on the AngularJS 1 version, while the AngularJS 2 version is being developed.
So the issue is that this will be lost when we migrate to AngularJS 2.... @atomfrede I'm sure your code is good, you can merge it if you want, but that will probably be lost with the next release.

atomfrede:
@jdubois You're right, didn't think about it, used some code I added to one of my projects. Maybe we can add it to the ng2 branch somehow.

jdubois:
it's too early to add new features to the AngularJS 2 branch... So that's really up to you! Personally I'm not for adding more and more features to those pages, they are squeleton pages, and in the end it's going to be annoying for people who want to modify them as there will be too much code.

atomfrede:
@jhipster/developers If no work on the entites for ng2 has started I would like to merge this, if you think it can easily be integrated when the entities are migrated to ng2 (should be as its just an additional state).

deepu105:
entity generator is not migrated yet so I guess this can be in. @jdubois this looks ok to me so if you are ok we can merge this

PR #3789, opened 2016-06-28
fix of SpringBoot property according to official documentation (from …

arnosthavelka:
Hello,

I recently created my app by JHipster generated application when I mentioned (when undermarked by IDE) the incorrect SpringBoot property show SQL. According to SpringBoot docs (see http://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html) the property name should be **show-sql** instead of **show_sql**.

Sincerely Arny

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-dev.yml
+++ b/generators/server/templates/src/main/resources/config/_application-dev.yml
@@ -86,7 +86,7 @@ spring:
         <%_ } else { _%>
         database: H2
         <%_ } _%>
-        show_sql: true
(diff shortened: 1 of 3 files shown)


PierreBesson:
Of note, if you wondered why it still worked: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-Configuration-Binding#relaxed-binding

arnosthavelka:
Yes, I am aware it's working even with the "wrong" property name. However it's nice to have properties as expected by IDE.

PR #3799, opened 2016-07-01
Updated _pom.xml

EdwardBeckett:
Updated Maven Coordinates for Sonar Maven Plugin.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -894,7 +894,7 @@
                 </executions>
             </plugin>
             <plugin>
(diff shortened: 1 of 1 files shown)


jdubois:
Shouldn't we also update the Gradle build?

atomfrede:
Not required. Gradle uses the [latest](http://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner+for+Gradle) sonar [plugin](https://plugins.gradle.org/plugin/org.sonarqube) already

PR #3808, opened 2016-07-07
Fix TypeScript Compilation

wmarques:
Following[ this discussion](https://github.com/angular/angular/issues/7280).
In order to fix the `Cannot find name 'Promise' errors`, we should target `es6`. By doing this, we can remove the exclude property which cleans up a bit our `tsconfig` file.

### DIFF
--- a/generators/client-2/templates/_tsconfig.json
+++ b/generators/client-2/templates/_tsconfig.json
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "es6",
         "module": "system",
         "moduleResolution": "node",
         "sourceMap": true,
@@ -10,9 +10,5 @@
         "noImplicitAny": false,
         "suppressImplicitAnyIndexErrors": true,
         "outDir": "<%= DIST_DIR %>app"
-    },
-    "exclude": [
-        "node_modules",
-        "typings/index.d.ts"
-    ]
+    }
 }


deepu105:
@wmarques wouldnt this cause the generated javascript to target ES6? if so that would cause browser compatibility issues as ES6 isn't fully supported by some browsers yet

wmarques:
Yes maybe... But I couldn't manage to get this working while targeting es5 :(

sendilkumarn:
@wmarques Instead of es6 we can have es6-shim in global dependency, that solves the problem related to promise.

istvanszoboszlai:
@wmarques, do you have a problem only in Inetlij Idea, or is it really a compilation problem?
I have had the same issue, and I managed to fix it with checking the
  Languages & Frameworks -> Typescript -> use tsonconfig.json 
checkbox in Idea preferences.
I hope it helps.

wmarques:
@istvanszoboszlai it seems to be a real compilation problem, the checkbox is already checkedin IntelliJ, thanks for your advice anyway :)
@sendilkumarn I'll try this

sendilkumarn:
I checked it. It was working for promises. Let me know if you have any problem

wmarques:
@sendilkumarn can you PR your changes ? I'll remove this one

sendilkumarn:
Okay then.

PR #3811, opened 2016-07-08
Health Service Refactor

wmarques:
Removed some TypeScript errors and used Response Object instead of Promise.


sendilkumarn:
Can't we inject service via constructor?

Export the service and import it in controller.

wmarques:
For now I just updated the service and not the controller. That's why I still import the service in the old way.

Issue #3800, opened 2016-07-04
Syntax error in import BigDecimal statement of EntityResourceIntTest


HemanthD:
##### **Overview of the issue**

The generated EntityResourceIntTest.java contains a syntax error when the entity includes a BigDecimal attribute.

##### **Motivation for or Use Case** 

Entities created with decimal datatype should have the basic tests run successfully out of the box.

##### **JHipster Version(s)** 

3.3.0

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

[.yo-rc.json.txt](https://github.com/jhipster/generator-jhipster/files/345872/default.yo-rc.json.txt)

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory** 

[TempEntity.json.txt](https://github.com/jhipster/generator-jhipster/files/345875/TempEntity.json.txt)

##### **Browsers and Operating System** 

Chrome on Win 8.1 with MySQL 5.7

##### **Reproduce the error** 

Create an entity with a decimal attribute
Check the generated EntityResourceIntTest.java - it would have a syntax error in the import statement related to BigDecimal

```
import javax.annotation.PostConstruct;
import javax.inject.Inject;
import java.math.BigDecimal;;
import java.util.List;
```

##### **Related issues** 

Did not find any similar cases.

##### **Suggest a Fix** 

The issue could be with the below piece of code
```
<% if (fieldsContainBigDecimal == true) { %>
import java.math.BigDecimal;;<% } %>
```

in **generator-jhipster/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java** which has an extra unwanted ';'

deepu105:
Do you want to do a PR to fix this?


Thanks & Regards,
Deepu

pascalgrimaud:
I dont see any issue here. How can ';;' be a syntax error ?
Anyway, this additional ';' should be removed to have a cleaner code :)

HemanthD:
Sure. I will try to submit a PR by early next week.

PR #3812, opened 2016-07-09
PR to fix a few syntax and typo errors

HemanthD:
This PR fixes the syntax warning/error due to unwanted extra ; in a few templates reported in #3800. The PR also includes fixes to a few typo errors and also formatting corrections (removing extra blank lines, adjusting format).

### DIFF
--- a/generators/server/templates/src/main/java/package/domain/_User.java
+++ b/generators/server/templates/src/main/java/package/domain/_User.java
@@ -28,7 +28,7 @@ import java.util.Set;<% if (databaseType == 'sql' || databaseType == 'mongodb') 
 import java.time.ZonedDateTime;<% } %>
 
 /**
- * A user.
+ * An user.
  */<% if (databaseType == 'sql') { %>
 @Entity
 @Table(name = "jhi_user")<% } %><% if (hibernateCache != 'no' && databaseType == 'sql') { %>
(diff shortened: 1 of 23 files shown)


generators/server/templates/src/main/java/package/domain/_User.java
```diff
@@ -28,7 +28,7 @@
 import java.time.ZonedDateTime;<% } %>
 
 /**
- * A user.
+ * An user.
```
ruddell: `An` should be `A` here.  

When `u` makes the same sound as the `y` in `you`, or `o` makes the same sound as `w` in `won`, then `a` is used.

generators/server/templates/src/main/java/package/security/_UserDetailsService.java
```diff
@@ -16,7 +16,7 @@
 import java.util.stream.Collectors;
 
 /**
- * Authenticate a user from the database.
+ * Authenticate an user from the database.
```
ruddell: `an` should be `a` here.

generators/server/templates/src/main/java/package/domain/_User.java
HemanthD: Thanks for the review. I have reverted the changes from 'An user' to 'A user'

deepu105:
thanks and appreciated :+1:

PR #3818, opened 2016-07-11
add RESERVED_WORDS_JHIPSTER to avoid entity name that can break the project

chamot1111:

When a user generate an entity "account", the jhipster project file AccountRessource.java is modified and the project is not working anymore.

##### Propose fix

I just add RESERVED_WORDS_JHIPSTER. For now, it's contains only 'ACCOUNT' but I suppose that the problem can happen with AuditRessource.java too.


deepu105:
We did away with the reserved words by using the JHI suffix for stuff that JHipster generates so we could do that here as well

chamot1111:
You mean that 'AccountRessource.java' should be 'JHIAccountRessource.java'?

jdubois:
I agree with @chamot1111 - let's do a list of JHipster reserved keywords

deepu105:
ok lets do it then

PR #3820, opened 2016-07-11
fix eslint errors in gulp files

vivekmore:
fix eslint errors in gulp files

- client folder
- client-2 folder


sendilkumarn:
LGTM :)

PR #3823, opened 2016-07-12
fixes to the typing error

sendilkumarn:
@wmarques, check with this. moved the es6-shim from ambient to global does the trick.


deepu105:
@wmarques merge this if you are ok

PR #3825, opened 2016-07-12
Ng 2 css fix

sendilkumarn:
@deepu105 @wmarques 
Just hacked around the hack temporarily. But we have to change this eventually


deepu105:
Looks ok, merge it after merging your other PR once @wmarques reviews that

PR #3834, opened 2016-07-13
Setup JSON pretty print in dev profile, fix #3830

### LINKED ISSUE #3830: Pretty print JSON in dev
In dev profile, we should probably enable ```SerializationFeature.INDENT_OUTPUT``` to increase readability (using JSON is all about human readability after all).
In prod, since we compress the response, pretty-printing would not add much traffic (I've read that it was about 2-3%) and ease debugging but that's debatable.
What do you think ?

PierreBesson:


cbornet:
See #3838 for a more generic solution.

PierreBesson:
Good job @cbornet :+1

PR #3838, opened 2016-07-13
Customize the Jackson2ObjectMapperBuilder instead of replacing it

### LINKED ISSUE #3830: Pretty print JSON in dev

cbornet:
also set INDENT_INPUT to true in dev profile

Fix #3830


generators/server/templates/src/main/java/package/config/_JacksonConfiguration.java
```diff
@@ -1,27 +1,49 @@
...
-        module.addSerializer(OffsetDateTime.class, JSR310DateTimeSerializer.INSTANCE);
-        module.addSerializer(ZonedDateTime.class, JSR310DateTimeSerializer.INSTANCE);
-        module.addSerializer(LocalDateTime.class, JSR310DateTimeSerializer.INSTANCE);
-        module.addSerializer(Instant.class, JSR310DateTimeSerializer.INSTANCE);
-        module.addDeserializer(LocalDate.class, JSR310LocalDateDeserializer.INSTANCE);
-        return new Jackson2ObjectMapperBuilder()
-                .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
-                .findModulesViaServiceLoader(true)
-                .modulesToInstall(module);
+    public ObjectMapper myJacksonObjectMapper() {
```
deepu105: Can you change the name to `jacksonObjectMapper` or `customJacksonObjectMapper` ?

cbornet: Oh, yes !

atomfrede:
Do you remove the old serializers for already created projects?

cbornet:
I only keep the serializer for ZonedDateTime.

atomfrede:
I mean the physical files in ``cleanup.js``. I think we should remove the obsolete serializers in already generated projects (or not?)

cbornet:
Oh, yes ! Good point ! I'll add them !

generators/cleanup.js
```diff
@@ -16,3 +17,13 @@ function cleanupOldFiles(generator, javaDir, testDir) {
         generator.removeFile(ANGULAR_DIR + 'components/form/uib-pagination.config.js');
     }
 }
+
+/**
+ * Removes server files that where generated in previous JHipster versions and therefore need to be removed
+ */
+function cleanupOldServerFiles(generator, javaDir, testDir) {
+    if (generator.isJhipsterVersionLessThan('3.4.3')) {
+        generator.removeFile(javaDir + 'domain/util/JSR310DateTimeSerializer.java');
```
cbornet: I think it would be better to use ```generator.fs.delete()``` but there is currently an issue (https://github.com/yeoman/generator/issues/950) so we can see about that later.

generators/server/index.js
```diff
@@ -343,6 +344,10 @@ module.exports = JhipsterServerGenerator.extend({
 
     writing: {
 
+        cleanupOldServerFiles: function() {
+            cleanup.cleanupOldServerFiles(this, this.javaDir, this.testDir);
```
cbornet: The cleanup made by app/index.js cannot be used because this.javaDir is null.

jdubois:
Lots of deleted code, cool!

PR #3844, opened 2016-07-14
Configure springfox to manage pagination parameters

cbornet:
Also upgrade to springfox v2.5.0

Credit goes to @namkee for his tip in https://github.com/springfox/springfox/issues/755


cbornet:
Arf ! Just saw that ```configuration/ui``` has special security config that must be changed to ```swagger-resources/configuration/ui``` ...
Will PR asap !

PR #3848, opened 2016-07-17
add eureka info into jhipster-registry's application.yml generated in src/main/docker

ruddell:
This line disappeared from app.yml in [this change](
https://github.com/jhipster/generator-jhipster/commit/ed91094d943794f25273e933eaf6ee36b6f33560#diff-0eb6bd7d70ceede187ac70836f037149).  Without it, the gateways and microservices throws the error when running `docker-compose -f src/main/docker/app.yml up` :
```
2016-07-17 02:06:24.622  INFO 8 --- [           main] com.netflix.discovery.DiscoveryClient    : Getting all instance registry info from the eureka server
optionsgateway-app_1            | 2016-07-17 02:06:24.751 ERROR 8 --- [           main] c.n.d.s.t.d.RedirectingEurekaHttpClient  : Request execution error
optionsgateway-app_1            | 
optionsgateway-app_1            | com.sun.jersey.api.client.ClientHandlerException: java.net.ConnectException: Connection refused
```

### DIFF
--- a/generators/server/templates/src/main/docker/central-server-config/application.yml
+++ b/generators/server/templates/src/main/docker/central-server-config/application.yml
@@ -8,3 +8,8 @@ jhipster:
         authentication:
             jwt:
                 secret: my-secret-token-to-change-in-production
+
+eureka:
+    client:
+        serviceUrl:
+            defaultZone: http://admin:admin@registry:8761/eureka/
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -169,7 +169,6 @@ module.exports = DockerComposeGenerator.extend({
 
                 // Re-configure the JHipster Registry if it is already configured
                 var spring_cloud_config_uri_configured = false;
-                var eureka_client_serviceurl_defaultzone_configured = false;
                 yamlConfig.environment.forEach(function (env, index, envArr) {
 
                     if (env.startsWith('SPRING_CLOUD_CONFIG_URI')) {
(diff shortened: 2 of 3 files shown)


PierreBesson:
@rudell, I know this is not documented anywhere. But the idea was that this property should be retrieved from the config server so there is no need to define it here. By default, the docker-compose subgen should configure this correctly.

ruddell:
@PierreBesson The docker-compose subgen puts this property into the Environment section in docker-compose.yml for both microservices and gateways.  

I uploaded a sample gateway/microservice project with the docker-compose folder set up.  The docker-compose.yml shows the property on [line 11](https://github.com/ruddell/jhip-example/blob/master/docker-compose/docker-compose.yml#L11) and [line 27](https://github.com/ruddell/jhip-example/blob/master/docker-compose/docker-compose.yml#L27).  If you remove it, both the gateway and microservice throw the above error.  

Am I missing a step?

PierreBesson:
@rudell I'm sorry I am not able to test because I'm currently in vacations without my computer.

Can you check that this property shows up in the registry UI : https://github.com/ruddell/jhip-example/blob/master/docker-compose/central-server-config/application.yml#L15
Then can you check that your apps can receive properties from the config server and that they overwrite their local ones (just edit the central-config/applications.yml).

ruddell:
@PierreBesson I see what you are saying.  The issue isn't the missing environment variable in src/main/docker/app.yml, but actually the missing config in src/main/docker/central-server-config/application.yml.  I have updated this pull request to fix it in the appropriate spot.

I also removed EUREKA_CLIENT_SERVICEURL_DEFAULTZONE from the docker-compose.yml output by the subgen, as it is configured from the registry's application.yml like you said.

ruddell:
Does this still need discussion?  Summary of the changes:
- Change src/main/docker/central-server-config/application.yml to match the one created by the docker-compose subgenerator (missing property).
- Remove extra EUREKA_CLIENT_SERVICEURL_DEFAULTZONE from docker-compose.yml created by the  docker-compose subgenerator - this property is set by the registry's central application.yml (docker-compose/central-server-config/application.yml)

pascalgrimaud:
LGTM @ruddell I will test and merge this

generators/server/templates/src/main/docker/central-server-config/application.yml
```diff
@@ -8,3 +8,8 @@ jhipster:
         authentication:
             jwt:
                 secret: my-secret-token-to-change-in-production
+
+eureka:
+    client:
+        serviceUrl:
+            defaultZone: http://admin:admin@registry:8761/eureka/
```
jdubois: The password is not a constant, "admin" is just a default value, this is the last question from the docker-compose sub-generator

jdubois:
@pascalgrimaud don't merge yet, have a look at my comment

ruddell:
@jdubois This is just for when running one gateway/microservice via `docker-compose src/main/docker/app.yml up`.  In the file [src/main/docker/jhipster-registry.yml](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/docker/_jhipster-registry.yml#L14) the admin password is set to `admin`.  

The docker-compose subgenerator still works the same and sets the password in its own central-server-config/application.yml.  None of the files in src/main/docker are changed by the subgenerator.

jdubois:
Oh yes you are correct @ruddell ! If this is just for standalone JHipster Registry, then this is "admin" by default, as we don't run the sub-generator.

pascalgrimaud:
Good job @ruddell !
Just tested it, and everything works fine without error

PR #3854, opened 2016-07-19
remove translation on enum type correctly

### LINKED ISSUE #3851: Translation not removed from Enum HTML
When generating an entity with the Enum type, translation is not removed by the [regex via stripHtml](https://github.com/jhipster/generator-jhipster/blob/53afa6de1e6a10dc059c65c58c314784a35eb6d8/generators/generator-base.js#L858).  Translate attributes can be found in `foos.html` and `foo-detail.html` around the entity's Enum value.

`            <span translate="{{'notransApp.Specialty.' + vm.foo.specialty}}">{{vm.foo.specialty}}</span>`

##### **JHipster Version(s)** 
3.0.0+

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.4.2",
    "baseName": "notrans",
    "packageName": "com.mycompany.app",
    "packageFolder": "com/mycompany/app",
    "serverPort": "8080",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": "no",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "f117103c1108a6288bbd7171c1ac2cc52c8c71e4",
    "useSass": false,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
```

```
{
    "relationships": [],
    "fields": [
        {
            "fieldName": "specialty",
            "fieldType": "Specialty",
            "fieldValues": "Cardiology,Immunology,Dermatology"
        }
    ],
    "changelogDate": "20160719000929",
    "dto": "mapstruct",
    "service": "serviceClass",
    "entityTableName": "foo",
    "pagination": "pagination"
}
```
##### **Reproduce the error** 

- Generate a project with translation disabled
- Generate an entity with an Enum field

##### **Suggest a Fix** 
Should the translate regex pick this up, or should we add <% if (enableTranslation) { %> around the translate attribute?

ruddell:
This changes the stripHtml regex to also check for `translate=` attributes containing `\ \+\{\}\'` (spaces, plus signs, brackets, single quotes).   The following line now has translation removed correctly: 
`-            <span translate="{{'notransApp.Specialty.' + vm.foo.specialty}}">{{vm.foo.specialty}}</span>		`
`+            <span>{{vm.foo.specialty}}</span>`
From https://github.com/ruddell/translation-removal-example/commit/2614377c7e26d02082826fb1280f21f01782043d

Fix #3851

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -856,7 +856,7 @@ Generator.prototype.copyTemplate = function (source, dest, action, generator, op
     var regex;
     switch (action) {
     case 'stripHtml' :
-        regex = /( translate\="([a-zA-Z0-9](\.)?)+")|( translate-values\="\{([a-zA-Z]|\d|\:|\{|\}|\[|\]|\-|\'|\s|\.)*?\}")|( translate-compile)|( translate-value-max\="[0-9\{\}\(\)\|]*")/g;
+        regex = /( translate\="([a-zA-Z0-9\ \+\{\}\'](\.)?)+")|( translate-values\="\{([a-zA-Z]|\d|\:|\{|\}|\[|\]|\-|\'|\s|\.)*?\}")|( translate-compile)|( translate-value-max\="[0-9\{\}\(\)\|]*")/g;
             //looks for something like translate="foo.bar.message" and translate-values="{foo: '{{ foo.bar }}'}"
         jhipsterUtils.copyWebResource(source, dest, regex, 'html', _this, _opt, template);
         break;


deepu105:
@ruddell have you verified if there is no impact on already existing translations on other parts?

ruddell:
@deepu105 The linked github repo has two commits; the first is a project generated before the regex change and the second is after the regex change.  The only difference is the enum translation is found by the regex.  I also tested generating an entity with a field of each possible type.  I grep'd the entire project and the word 'translate' does not appear in the generated files.

This change doesn't affect projects that have translation.

jdubois:
LGTM, but indeed that's very hard to test, and could break many applications!

PR #3861, opened 2016-07-21
reduce useless logger display

bohrqiu:
`LogsResource#getList` return too many useless `LoggerDTO`


gmarziou:
@bohrqiu I suppose you use this feature, few times we have discussed removing it and it would make even more sense now as we offer logstash support. So what's yrou use case for this? Do you use it in prod?

bohrqiu:
`Logs` page for human to modify or check which logger that customed,display too many loggers is helpless.

gmarziou:
Regarding you PR, I understand it but with your change it's impossible to set a log level to a logger that has not one already, that's why we display all of them and have a filter.

jdubois:
I agree with @gmarziou -> that's why we have an AngularJS filter. I know it's quite slow, but that's the only way to configure any logger in the application.

bohrqiu:
yes, I see @gmarziou @jdubois

PR #3869, opened 2016-07-22
new subgenerator to display information

pascalgrimaud:
Use: `yo jhipster:information`

I change the order in the ISSUE_TEMPLATE to make the copy/paste easier

### DIFF
--- a/generators/information/index.js
+++ b/generators/information/index.js
@@ -1,0 +1,151 @@
+'use strict';
+var generators = require('yeoman-generator'),
+    chalk = require('chalk'),
+    shelljs = require('shelljs'),
+    util = require('util'),
+    scriptBase = require('../generator-base');
+
+var ReportGenerator = generators.Base.extend({});
+
+util.inherits(ReportGenerator, scriptBase);
+
+module.exports = ReportGenerator.extend({
+    constructor: function () {
+        generators.Base.apply(this, arguments);
+    },
+
+    initializing: {
+        sayHello: function() {
+            this.log(chalk.white('Welcome to the JHipster Information Sub-Generator\n'));
+        },
+
+        checkJHipster: function () {
+            var done = this.async();
+            this.log('##### **JHipster Version(s)**\n');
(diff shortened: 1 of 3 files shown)


deepu105:
@pascalgrimaud @jdubois may be it should be `yo jhipster:info` so its easier to use as most projects use info keyword. what say. since this is not releases we can still change?

jdubois:
Yes let's change it

deepu105:
btw great idea :+1:

deepu105:
ok ill change now

deepu105:
done

generators/information/index.js
```diff
@@ -0,0 +1,151 @@
...
+                    this.log('```');
+                }
+                done();
+            }.bind(this));
+        },
+
+        displayYorc: function () {
+            var done = this.async();
+            this.log('##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**\n');
+            shelljs.exec('cat .yo-rc.json', {silent:true}, function (err, stdout, stderr) {
```
deepu105: I dont think it works in windows though :) but we can improve it later

jdubois: No, good catch! It shouldn't be too hard to output the file in pure JS

pascalgrimaud:
Yes agree, yo jhipster:info is better!!

generators/information/index.js
pascalgrimaud: Oh right, I didn't test with Windows :D

generators/information/index.js
```diff
@@ -0,0 +1,151 @@
...
+                    this.log('```\n');
+                }
+                done();
+            }.bind(this));
+        },
+
+        displayEntities: function () {
+            var done = this.async();
+            this.log('##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**\n');
+            shelljs.exec('for f in `ls .jhipster`; do echo ${f%} ; cat .jhipster/${f%} ; done', {silent:true}, function (err, stdout, stderr) {
```
deepu105: this is failing in windows as well

PR #3881, opened 2016-07-26
Fix constraints when customizing table name

### LINKED ISSUE #3723: Error in addForeignKeyConstraint when using overridden table names

Liquibase generated addForeignKeyConstraint seems to use the default entity-name-derived table name rather than the specified/ overridden tablename (using --table-name).

Consider 2 entities Criteria and Option with one-to-many relationship. Both entities are generated using --table-name option as pa_criteria and pa_option. The foreign key generated in pa_option has referencedTableName="criteria" (default table name of Criteria entity) instead of referencedTableName="pa_criteria" (specified table name).

Liquibase fails while creating the fk.

liquibase.exception.DatabaseException: Cannot add foreign key constraint [Failed SQL: ALTER TABLE atlas.pa_option ADD CONSTRAINT fk_option_criteria_id FOREIGN KEY (criteria_id) REFERENCES atlas.criteria (id)]
	at liquibase.executor.jvm.JdbcExecutor$ExecuteStatementCallback.doInStatement(JdbcExecutor.java:301)

Table names can be overridden using --table-name. However, there seem to be an issue when creating references to such entities - requiring manual intervention.

##### **JHipster Version(s)** 

3.3.0

[.yo-rc.json.txt](https://github.com/jhipster/generator-jhipster/files/317725/default.yo-rc.json.txt)

[Criteria.json.txt](https://github.com/jhipster/generator-jhipster/files/317728/Criteria.json.txt)
[Option.json.txt](https://github.com/jhipster/generator-jhipster/files/317727/Option.json.txt)

Chrome on Win 8.1 with MySQL 5.7

##### **Reproduce the error** 

    save attached Criteria.json and Option.json in .jhipster/
    yo jhipster:entity criteria --table-name pa_criteria --regenerate
    yo jhipster:entity option --table-name pa_option --regenerate
    check the xxxx_added_entity_constraints_Option.xml - it will have the incorrect referencedTableName="criteria"
    launch the app
    Liquibase will throw an error while processing the xxxx_added_entity_constraints_Option.xml
    
##### **Related issues** 

Did not find any similar cases.

##### **Suggest a Fix** 

The issue could be with the below piece of code

    <addForeignKeyConstraint baseColumnNames="<%= getColumnName(relationshipName) %>_id"
                             baseTableName="<%= entityTableName %>"
                             constraintName="<%= constraintName %>"
                             referencedColumnNames="id"
                             referencedTableName="<%
                             var otherEntityTable = getTableName(otherEntityName);
                             if (otherEntityTable == 'user') { %>jhi_user<% } else { %><%=otherEntityTable %><% } %>"/>

in generator-jhipster/generators/entity/templates/src/main/resources/config/liquibase/changelog/_added_entity_constraints.xml which is using _otherEntityName_ for _getTableName()_. Perhaps, it should use the *'entityTableName'* attribute from the _otherEntity.json_

ruddell:
Fix #3723, this is Option 2 from @jdubois's post in the issue.  It also fixes relationships with user when using a customized jhiPrefix.

Only thing left is to add a test using custom table names to the entity generation step.

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -118,7 +118,6 @@ module.exports = EntityGenerator.extend({
             this.languages = this.config.get('languages');
             this.buildTool = this.config.get('buildTool');
             this.testFrameworks = this.config.get('testFrameworks');
-            this.jhiPrefix = this.config.get('jhiPrefix');
             // backward compatibility on testing frameworks
             if (this.testFrameworks === undefined) {
                 this.testFrameworks = ['gatling'];
@@ -349,7 +348,7 @@ module.exports = EntityGenerator.extend({
                 this.service = 'no';
             }
             if (_.isUndefined(this.entityTableName)) {
-                this.warning('entityTableName is missing in .jhipster/' + this.name + '.json, using no as fallback');
+                this.warning('entityTableName is missing in .jhipster/' + this.name + '.json, using entity name as fallback');
                 this.entityTableName = this.getTableName(this.name);
             }
             if (_.isUndefined(this.pagination)) {
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -118,6 +118,7 @@ module.exports = EntityGenerator.extend({
             this.languages = this.config.get('languages');
             this.buildTool = this.config.get('buildTool');
             this.testFrameworks = this.config.get('testFrameworks');
+            this.jhiPrefix = this.config.get('jhiPrefix');
             // backward compatibility on testing frameworks
             if (this.testFrameworks === undefined) {
                 this.testFrameworks = ['gatling'];
@@ -211,7 +212,7 @@ module.exports = EntityGenerator.extend({
         this.service = this.fileData.service;
         this.pagination = this.fileData.pagination;
         this.javadoc = this.fileData.javadoc;
-        this.entityTableName = this.fileData.entityTableName || _.snakeCase(this.name).toLowerCase();
+        this.entityTableName = this.fileData.entityTableName;
         this.fields && this.fields.forEach(function (field) {
             this.fieldNamesUnderscored.push(_.snakeCase(field.fieldName));
             this.fieldNameChoices.push({name: field.fieldName, value: field.fieldName});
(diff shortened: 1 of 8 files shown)


generators/entity/index.js
```diff
@@ -347,6 +348,10 @@ module.exports = EntityGenerator.extend({
                 this.warning('service is missing in .jhipster/' + this.name + '.json, using no as fallback');
                 this.service = 'no';
             }
+            if (_.isUndefined(this.entityTableName)) {
+                this.warning('entityTableName is missing in .jhipster/' + this.name + '.json, using no as fallback');
```
deepu105: can we have message as `using entity name as fallback`?

generators/entity/templates/src/main/resources/config/liquibase/changelog/_added_entity_constraints.xml
```diff
@@ -12,6 +12,14 @@
             relationshipName = relationships[idx].relationshipName,
             ownerSide = relationships[idx].ownerSide,
             otherEntityName = relationships[idx].otherEntityName;
+            if (otherEntityName === 'user') {
+                otherEntityTableName = jhiPrefix + '_user';
```
deepu105: I dont think we should prefix `jhiPrefix ` for user table as the table name for user will always be `jhi_user` see [this](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/liquibase/changelog/_initial_schema.xml#L23)
so if user sets a jhiPrefix then this will fail, it works only with default prefix which is `jhi`

generators/entity/templates/src/main/resources/config/liquibase/changelog/_added_entity_constraints.xml
```diff
@@ -12,6 +12,14 @@
...
+            } else {
+                otherEntityTableName = getEntityProperty(otherEntityName, 'entityTableName');
+                if (otherEntityTableName === undefined) {
```
deepu105: you are returning null always from the method so this will not be undefined. plz change to `if (otherEntityTableName) {` so that null, undefined and empty is checked

deepu105: it should be `if (!otherEntityTableName) `

ruddell:
Thanks for reviewing the changes.  You're right about the prefix, it's only used in the JS.  I will fix these and add the test in the morning.

PR #3891, opened 2016-07-27
use formstate instead of classes for validation

### LINKED ISSUE #3703: "show-validation" directive is not in sync with the input validation
##### Form Validation Issue

This bug is seen for all form validations

##### Appears on all Forms (Dialogs)

Form group unnecessarily turns red even when the input box has correct data

##### All Versions

This is there in all versions of JHipster.

{
  "generator-jhipster": {
    "jhipsterVersion": "3.4.0",
    "baseName": "XXXX",
    "packageName": "XXX",
    "packageFolder": "XXX",
    "serverPort": "8080",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": "no",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "XXX",
    "useSass": false,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}

Default App

Firefox on Windows and Linux

##### **Reproduce the error** 
1. On Home Page, From Menu, Go to Sign In and Forgot Password
2. Enter abc@g. **No error in this value - CORRECT**
3. Enter backspace 
4. Screen will have abc@ and it will show validation errors. But there is no Red Border for Form Group. 
5. Now enter a character "g". Screen should show abc@g. Now the Red border appears - AFTER validation errors disappear. **No Error in this value - RED BORDER INCORRECT**
6. This is a disconnect. Red Border should have appeared along with error message in Step 4 and disappear in Step 5, whereas it appeared freshly in Step 5.
7. This happens in ALL forms ALL fields and is more prominent for other validations.

##### **Related issues** 

Not Applicable

##### **Suggest a Fix** 

**The problem is most probably because of the screen value and model value not being in sync.**

See Code and Screenshots below

This is the code which is causing the problem: src/main/webapp/app/components/form/show-validation.directive.js.

```
                if ($inputs.length > 0) {
                    $inputs.each(function() {
                        var $input = angular.element(this);
                        scope.$watch(function() {
                            return $input.hasClass('ng-invalid') && $input.hasClass('ng-dirty');
                        }, function(isInvalid) {
                            $formGroup.toggleClass('has-error', isInvalid);
                        });
                    });
                }
```

noerror

error

atomfrede:
close #3703

### DIFF
--- a/generators/client/templates/src/main/webapp/app/components/form/_show-validation.directive.js
+++ b/generators/client/templates/src/main/webapp/app/components/form/_show-validation.directive.js
@@ -14,7 +14,7 @@
 
         return directive;
 
-        function linkFunc (scope, element) {
+        function linkFunc (scope, element, attrs, formCtrl) {
             element.find('.form-group').each(function() {
                 var $formGroup = angular.element(this);
                 var $inputs = $formGroup.find('input[ng-model],textarea[ng-model],select[ng-model]');
@@ -22,8 +22,9 @@
                 if ($inputs.length > 0) {
                     $inputs.each(function() {
                         var $input = angular.element(this);
+                        var inputName = $input.attr('name');
                         scope.$watch(function() {
-                            return $input.hasClass('ng-invalid') && $input.hasClass('ng-dirty');
+                            return formCtrl[inputName].$invalid && formCtrl[inputName].$dirty;
                         }, function(isInvalid) {
                             $formGroup.toggleClass('has-error', isInvalid);
                         });


ruddell:
Tested and it works for me. 👍 Nice fix to a longstanding issue

atomfrede:
Then let's Merge it.

jdubois:
Great job! I couldn't find this solution!

gvsrini:
Good job indeed!

PR #3893, opened 2016-07-27
An exception was logged as 'DEBUG'

aBnormaLz:
for example:

2016-07-27 11:05:27.787 DEBUG 8732 --- [nio-8080-exec-3] c.s.backend.aop.logging.LoggingAspect    : Enter: com.wanari.backend.web.rest.errors.ExceptionTranslator.processRuntimeException() with argument[s] = [org.springframework.web.multipart.support.MissingServletRequestPartException: Required request part 'file' is not present]

2016-07-27 11:05:27.793 DEBUG 8732 --- [nio-8080-exec-3] c.s.backend.aop.logging.LoggingAspect    : Exit: com.wanari.backend.web.rest.errors.ExceptionTranslator.processRuntimeException() with result = <500 Internal Server Error,com.wanari.backend.web.rest.errors.ErrorDTO@ca2d2e9,{}>

Internal server error logged as debug


aBnormaLz:
https://github.com/aBnormaLz/jhipster-test

Here is an example project to generate the bug

you should POST to /api/maybebug with herder:
Authorization: Bearer eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJhZG1pbiIsImF1dGgiOiJST0xFX0FETUlOLFJPTEVfVVNFUiIsImV4cCI6MTQ3MjIxMzE5OH0.xJ7Kmaz8jRuz5iPvKY6_3Xf1J-9gZ-KDpPNP_TfRCA3e0MasXQRnVvbHkVUzY5PR-BShzK1Z8yC3dSOFmc0M1Q

https://github.com/aBnormaLz/jhipster-test/blob/master/src/main/java/com/wanari/web/rest/UserResource.java

UserResource.createUserTask() is the function

jdubois:
1. It is normal that this is logged in debug: the Logging aspect logs every method call. This is only used in development mode, to ease development. So you know the method calls, parameters, etc. And this of course includes exceptions. You will not have this in production.
2. Then you should have also the error logged at the error level, with the corresponding stacktrace. I just tested it, and it's working for me. But from my understanding your issue is with the first point.

I'm also not merging your PR as there is no need to log the exception here: it should go up to the upper level of the stack, where it's supposed to be caught, and logged. This is my second point: again, it works for me, but if it doesn't work for you, please add a comment -> in that case that should be solved at the Spring Boot level, as it's supposed to configure an exception translator that takes everything.

PR #3897, opened 2016-07-28
OrderBy and Filter Pipes

wmarques:
As suggested by @sendilkumarn.
I implemented them in both Logs and Config components.


generators/client-2/index.js
```diff
@@ -343,6 +343,8 @@ module.exports = JhipsterClientGenerator.extend({
             this.template(ANGULAR_DIR + 'blocks/config/_compile.config.ts', ANGULAR_DIR + 'blocks/config/compile.config.ts', this, {});
             this.template(ANGULAR_DIR + 'blocks/config/_uib-pager.config.ts', ANGULAR_DIR + 'blocks/config/uib-pager.config.ts', this, {});
             this.template(ANGULAR_DIR + 'blocks/config/_uib-pagination.config.ts', ANGULAR_DIR + 'blocks/config/uib-pagination.config.ts', this, {});
+            this.copyJs(ANGULAR_DIR + 'shared/filter.pipe.ts', ANGULAR_DIR + 'shared/filter.pipe.ts', this, {});
```
deepu105: put the pipes under `components/util` plz

wmarques:
@deepu105 I did this in order to follow the style guide : https://angular.io/docs/ts/latest/guide/style-guide.html#!#04-07

deepu105:
@wmarques you are right. I didnt look at the ng2 style guide I didnt realize that it had difference in structure from ng1 guideline. this is good then, may be we can move the components into the shared folder as well. personally i liked the components name rather than shared

Issue #3894, opened 2016-07-27
[FEATURE] Modify $state.go() reload option to use state name instead of 'true'


danielbrutti:
Hi,

##### **Overview of the issue**

When modifying template and create a states hierarchy for Entities, the reload option does not work and redirects to home page (state) when create/edit/delete. For example, for a Country entity:
~~~
$uibModal.open({
                    templateUrl: 'app/entities/country/country-dialog.html',
                    controller: 'CountryDialogController',
                    controllerAs: 'vm',
                    backdrop: 'static',
                    size: 'lg',
                    resolve: {
                        entity: function () {
                            return {
                                name: null,
                                code: null,
                                id: null
                            };
                        }
                    }
                }).result.then(function() {
                    $state.go('country', null, { reload: true });
                }, function() {
                    $state.go('country');
});
~~~

if `country` state does not have entity as parent, which is my case because I had to modify states hierarchy because of new web app template, it will redirect to the parent state on the hierarchy instead of go to `country` and reload the list.

##### **Motivation for or Use Case**

Even when the issue happens because of the modifications I had to made into app template and states, it could happen to other people. And, the solution for this is really simple as explained on this ui-router [ticket](https://github.com/angular-ui/ui-router/issues/1612) and [solution](https://github.com/angular-ui/ui-router/pull/1809/files) already implemented.

##### **Suggest a Fix**

Modify the entities generator in order to make the reload option a little bit different.. just change the `true` value of reload for the `entity state name`. 

In the example I explained before we need to change this `$state.go('country', null, { reload: true });` by this `$state.go('country', null, { reload: 'country' });` for `.new`, `.edit`, `.delete` states.

##### **JHipster Version(s)**
v3.5.0

If you need help, I can try to clone projects and do the change.

Best,

Daniel

sendilkumarn:
Yes please, give it a shot

ruddell:
Can I please get steps to reproduce the issue?  

If I change the parent state from `entity` to `home`, I don't get redirected to the home on changes to the entity via edit/delete/new.  If I then click refresh while on the entity page, I do get sent home.  I experience the same behavior with your pull request (changing state to home)...

deepu105:
Ya I coulndt understand the difference as well!

deepu105:
Closed via #3898

PR #3898, opened 2016-07-28
Modified reload for .new .edit and .delete states

danielbrutti:
Added the `.new`, `.edit` and `.delete` states reload modifications based on this #3894 issue

Best,

Daniel


deepu105:
@sendilkumarn did you test this?

sendilkumarn:
Not yet. I will try to test over the weekend. Completely occupied 👍

sendilkumarn:
this has better use case in `multiple nested` route and  `customised` version

`true` will reload the entire tree and using the `entity` name here will be more correct

deepu105:
Ok then

PR #3902, opened 2016-07-29
Upgrade Health Page

wmarques:
Health Page refactored. For now it's only the page and not the modal showing some details.
I have some trouble to manage the `resolve` property when opening a modal. I don't know how to get data in child component.

I tried using `@Input` decorator in modal component but it's not working the variable I send (health) is always undefined...
If you have any clues here I'd be glad to migrate the modal also.


generators/client-2/templates/src/main/webapp/app/admin/health/health.html
```diff
@@ -12,19 +12,18 @@ <h2 translate="health.title">Health Checks</h2>
                 <th class="col-md-7" translate="health.table.service">Service Name</th>
                 <th class="col-md-2 text-center" translate="health.table.status">Status</th>
                 <th class="col-md-2 text-center" translate="health.details.details">Details</th>
-
             </tr>
         </thead>
         <tbody>
-            <tr ng-repeat="health in vm.healthData">
-                <td><% if (enableTranslation){ %>{{'health.indicator.' + vm.baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ vm.baseName(health.name) }}</span><% } %> {{vm.subSystemName(health.name)}}</td>
```
deepu105: I think the ejs switches are required here

sendilkumarn:
Great work @wmarques 👍

generators/client-2/templates/src/main/webapp/app/admin/health/health.html
wmarques: For now the translate pipe isn't working, it's showing the key and not the translation that's why I removed it... Should I put a TODO or I put the pipe anyway ?

wmarques:
@deepu105 @sendilkumarn thanks for your feedbacks :)

generators/client-2/templates/src/main/webapp/app/admin/health/health.html
sendilkumarn: can we add a dummy pipe now and put a todo there? So that in future if something comes to fill it here will be a minimal change ?

wmarques: Sure i'll do this as soon as I have some time :+1:

wmarques: Done !

wmarques:
@deepu105 @sendilkumarn Should be fine now :)

PR #3918, opened 2016-08-02
Use JAVA_HOME, if available, to find keytool

### LINKED ISSUE #3903: Keytool command used in generateKeyStore doesnt work in windows
The `keytool` command is used for UAA server option in `generateKeyStore` method of `generator-base.js` its not cross platform as the command needs to be executed from the jdk/jre bin folder. not sure if this is applicable for mac and linux @jhipster/developers can anyone check

@sdoxsee since this is from you could you check?

sdoxsee:
fix #3903

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1164,7 +1164,12 @@ Generator.prototype.generateKeyStore = function() {
     } else {
         shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
         var parent = this;
-        shelljs.exec('keytool '+
+        var javaHome = shelljs.env['JAVA_HOME'];
+        var keytoolPath = '';
+        if (javaHome) {
+            keytoolPath = javaHome + '/bin/';
+        }
+        shelljs.exec(keytoolPath + 'keytool '+
             '-genkey ' +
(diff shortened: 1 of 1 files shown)


generators/generator-base.js
```diff
@@ -1164,7 +1164,12 @@ Generator.prototype.generateKeyStore = function() {
     } else {
         shelljs.mkdir('-p', SERVER_MAIN_RES_DIR);
         var parent = this;
-        shelljs.exec('keytool '+
+        var javaHome = shelljs.env['JAVA_HOME'];
+        var keytoolPath = '';
+        if (javaHome) {
+            keytoolPath = javaHome + '/bin/';
+        }
+        shelljs.exec(keytoolPath + 'keytool '+
```
deepu105: @sdoxsee this still doesnt work in windows, the below works. can you check if its ok for osx?
`shelljs.exec('"' + keytoolPath + 'keytool" '+`

sdoxsee:
@deepu105 yes. your change works here on OSX.

deepu105:
ok ill change it


Thanks & Regards,
Deepu

sdoxsee:
Thanks!

Issue #3933, opened 2016-08-04
Error using swagger on production environment


marcelinobadin:
##### **Overview of the issue**

I am running the application on a microservice architecture. My app and my gateway are running with production profile and also swagger.

The gateway swagger is Ok, but when I change the combo to see the swagger from the app I got this:

com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
@6/P���� ��w�����OO^t���w�O���cx����/#����m���ps4��Kf��|^

(with this strangers characters and then my terminal got broken)

##### **JHipster Version(s)**

generator-jhipster@3.5.1 

##### **Browsers and Operating System**

openjdk version "1.8.0_91"
OpenJDK Runtime Environment (build 1.8.0_91-8u91-b14-0ubuntu4ü16.04.1-b14)
OpenJDK 64-Bit Server VM (build 25.91-b14, mixed mode)

git version 2.7.4

node: v6.2.2

npm: 3.10.5

bower: 1.7.9

gulp:
CLI version 1.2.2
Local version 3.9.1

yeoman: 1.8.4

Docker version 1.11.2, build b9f10c9

docker-compose version 1.6.2, build 4d72027

gmarziou:
Does it work when you access swagger spec directly from your service at /v2/api-docs ?

marcelinobadin:
I am trying via postman, ip:port/app/v2/api-docs then I got 200 OK but nothing on body (using Authorization Bearer <token> authenticated as admin - authentication: jwt). If I try to get ip:port/v2/api-docs I got 200 OK with body containing the swagger of my gateway.

marcelinobadin:
If I get direct from the app port (not passing through the gateway) I got the swagger json as a return

gmarziou:
Well I don't use postman, is there a way to show exactly your request  including all headers. Reporting such problem with a curl command is much easier for us to try to reproduce.

marcelinobadin:
well, I found what it is but I need help to solve.

I removed application/json from the mime-types to be compressed on the application-prod configuration and swagger is working. But I still need the application/json content to be compressed and only swagger to not be (as we can see).

Any thoughts?

gmarziou:
Which app-prod, the one of the service or gateway, or both?
I use compression in both dev and prod on services and gateways and I haven't experienced your issue.

marcelinobadin:
[...]

PR #3940, opened 2016-08-05
Issue #3933 fix

marcelinobadin:
Check if response is GZipped and transform it correctly to string on SwaggerBasePathRewritingFilter filter.


jdubois:
LGTM! Thanks!
I'll wait for Travis to pass before merging.

PR #3946, opened 2016-08-08
Home Page Ng2 Migration

wmarques:


sendilkumarn:
👍  LGTM

generators/client-2/templates/src/main/webapp/app/home/_home.component.ts
```diff
@@ -0,0 +1,31 @@
...
+        this.login = loginService.open;
+    }
+
+    ngOnInit() {
+        //TODO: Remove this line after migrating "Principal" service
+        let that = this;
+        this.principal.identity().then(function (account) {
+            that.account = account;
+            that.isAuthenticated = that.principal.isAuthenticated;
+        });
```
deepu105: you can also do this, anyway not trivial since we need to change this part anyways so im merging it
```
        this.principal.identity().then(function (account) {
             this.account = account;
             this.isAuthenticated = this.principal.isAuthenticated;
         }.bind(this));
```

Issue #3807, opened 2016-07-07
Do we need JhiLanguageController?


marcelinobadin:
Everywhere in the front-end code where we use this controller can be easily moved to use only the JhiLanguageService. The only change on the service is to add a ```changeLanguage``` method.

Also, it will simplify, because everywhere we have to use the controller we need to inject dependencies that already contained on the service.

On my humble opinion, we can remove the controller and keep only the service, filter and constants.

wmarques:
Agree with you, on top of that, we should avoid to use the `ng-controller` and keep controller assignments in routes.
So it seems logic to me to remove the `LanguageController` and put logic in the service.

deepu105:
yes it can be removed. may be lets do it for ng2 directly


Thanks & Regards,
Deepu

deepu105:
Im closing this as we agree that this will be removed for ng2

PR #3948, opened 2016-08-08
Navbar Refactoring

wmarques:
Also removed the Language Controller as suggested in #3807.

The dropdowns are not working for the moment as we didn't implemented the ng2 ui bootstrap, so all of the ng1 directives won't work for now (ui-sref, active-menu, uib-dropdown...).


generators/client-2/templates/src/main/webapp/app/layouts/navbar/navbar.html
```diff
@@ -1,27 +1,27 @@
...
                 <span class="sr-only">Toggle navigation</span>
                 <span class="icon-bar"></span>
                 <span class="icon-bar"></span>
                 <span class="icon-bar"></span>
             </button>
-            <a class="navbar-brand logo" href="#/" ng-click="vm.collapseNavbar()">
+            <a class="navbar-brand logo" href="#/" (click)="collapseNavbar()">
                 <img class="logo-img" src="content/images/logo-jhipster.png"/>
-                <span translate="global.title"><%= capitalizedBaseName %></span> <span class="navbar-version">v{{VERSION}}</span>
+                <span translate="global.title">Ng2jhi</span> <span class="navbar-version">v{{VERSION}}</span>
```
sendilkumarn: `Ng2jhi` ?

generators/client-2/templates/src/main/webapp/app/layouts/navbar/navbar.html
```diff
@@ -95,91 +93,71 @@
...
                         <li ui-sref-active="active">
-                            <a ui-sref="logs" ng-click="vm.collapseNavbar()">
+                            <a ui-sref="logs" (click)="collapseNavbar()">
                                 <span class="glyphicon glyphicon-tasks"></span>&nbsp;
                                 <span translate="global.menu.admin.logs">Logs</span>
                             </a>
                         </li>
-                        <li ng-show="vm.swaggerEnabled" ui-sref-active="active">
-                            <a ui-sref="docs" ng-click="vm.collapseNavbar()">
+                        <li [hidden]="!swaggerEnabled" ui-sref-active="active">
```
sendilkumarn: ngIf please

generators/client-2/templates/src/main/webapp/app/layouts/navbar/navbar.html
```diff
@@ -95,91 +93,71 @@
...
-                        <li ui-sref-active="active">
-                            <a ui-sref="<%=jhiPrefix%>-tracker" ng-click="vm.collapseNavbar()">
-                                <span class="glyphicon glyphicon-eye-open"></span>&nbsp;
-                                <span translate="global.menu.admin.tracker">User tracker</span>
-                            </a>
-                        </li>
-                        <%_ } _%>
                         <li ui-sref-active="active">
-                            <a ui-sref="<%=jhiPrefix%>-metrics" ng-click="vm.collapseNavbar()">
+                            <a ui-sref="jhi-metrics" (click)="collapseNavbar()">
```
sendilkumarn: can we follow the same here ? `<%=jhiPrefix%>`

sendilkumarn:
good one 👍

generators/client-2/templates/src/main/webapp/app/layouts/navbar/navbar.html
wmarques: `ngIf` and `[hidden]` does not do the same thing:
`ngIf` remove the element from the DOM while hidden just add a CSS property to hide the element, that's what `ng-show` does, that's why I put `hidden` and not `ngIf` there, is that ok for you ? :)

deepu105:
I think I merged your other PR and that is causing conflict here, can you rebase plz?

wmarques:
Dunno why tests are failing :(

wmarques:
I don't know why I try to generate an app, Yeoman doesn't copy any files and just stops after prompting questions..

deepu105:
There might be a silent error or an ejs error on one of the files being
copied

Thanks & regards,
Deepu

wmarques:
Trying to check diff to see if there is an error like this

wmarques:
Ok I had one ejs error in common module file... Quite to difficult to find it as Yeoman doesn't give any clue about the crash :(

Let's wait for tests but it should be fine, I hope !

PR #3951, opened 2016-08-09
fix gulp inject task to inject bower dependencies correctly

ruddell:
This fixes `gulp inject` to inject installed bower dependences from bower.json.  

You can reproduce the issue by installing something with bower (for example `bower install underscore --save`) and running `gulp inject`, the  files will not be found in index.html.  

Another way to reproduce this is to remove any of the files between `<!-- bower:js --><!-- endinject -->` and run `gulp inject`.  No files will be injected.

Originally reported on [StackOverflow](http://stackoverflow.com/questions/37816714/gulp-task-injectvendor-does-not-inject-bower-dependencies/37826202#37826202)

### DIFF
--- a/generators/client/templates/gulpfile.js
+++ b/generators/client/templates/gulpfile.js
@@ -45,7 +45,7 @@ gulp.task('clean', function () {
 });
 
 gulp.task('copy', function () {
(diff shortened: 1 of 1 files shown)


deepu105:
Good catch, this was a race condition causes the issue rarely thats why it wasn't reported earlier.

PR #3956, opened 2016-08-09
upgrade docker-engine to fix travis tests

ruddell:


ruddell:
The [mongodb test fails](https://travis-ci.org/ruddell/generator-jhipster/jobs/151012545#L2735) due to the mongodb port already being used.  This also affects the older version of docker-engine we were using ([downgraded docker-engine tests](https://travis-ci.org/ruddell/generator-jhipster/builds/151031605). I'm not sure what's happening - might be another travis change?

Also, app-jwt fails sometimes but not always.

pascalgrimaud:
@ruddell: about mongo, there is maybe the same issue here https://github.com/travis-ci/travis-ci/issues/6443

If it's not resolved, I will have a look this week-end, end of holidays :-)

Issue #3952, opened 2016-08-09
Heroku with Postgresql throws Duplicate @ConfigurationProperties 


ruddell:
##### **Overview of the issue**
When deploying a Postgresql Monolith to Heroku, the error `Duplicate @ConfigurationProperties definition for prefix 'spring.datasource.hikari'` is thrown and the app fails to build.

##### **Reproduce the error**
Generate a postgresql app (with master) and run `yo jhipster:heroku`

##### **Suggest a Fix**
Removing `@ConfigurationProperties` from [DatabaseConfiguration](https://github.com/jhipster/generator-jhipster/blob/a74d244b8a829fae264ad74ac4bb9661986e1b5b/generators/server/templates/src/main/java/package/config/_DatabaseConfiguration.java#L85) or [HerokuDatabaseConfiguration](https://github.com/jhipster/generator-jhipster/blob/6db557c0aab25d859dc63eae15db80d5b9783fb6/generators/heroku/templates/src/main/java/package/config/_HerokuDatabaseConfiguration.java#L22) fixes it, is this property needed anymore? The app runs without it.

##### **JHipster Version(s)**
Master, does not affect v3.5.1 (Maybe started with Spring Boot v1.4 upgrade?)

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**
```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.5.1",
    "baseName": "pgsqltest",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "sql",
    "devDatabaseType": "postgresql",
    "prodDatabaseType": "postgresql",
    "searchEngine": "no",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "jwtSecretKey": "d84fac31892ef2bdfd040cec5455e6754078fee3",
    "useSass": false,
    "applicationType": "monolith",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ]
  }
}
```

sendilkumarn:
In `spring 1.4`, there is change in `@ConfigurationProperties`.

[...]

PR #3957, opened 2016-08-09
Replace HerokuDatabaseConfig.java with yaml config

jkutner:
The `HerokuDatabaseConfig.java` is unnecessarily complicated. It also seems to have problems in some cases. The yaml config is much better.

Do I also need to do something to delete an existing `HerokuDatabaseConfig.java`? What does that look like?


ruddell:
To delete an existing file in future versions, add it to [generators/cleanup.js](https://github.com/jhipster/generator-jhipster/blob/master/generators/cleanup.js)

cbornet:
You also need to remove heroku from the excluded profiles in the regular DatabaseConfiguration.java

jdubois:
This is linked to #3952 -> I'll try to solve both at the same time

jdubois:
And thanks @jkutner that's much better than the current code!

PR #3960, opened 2016-08-10
Add Forms Dependency + Refactor Register Page

wmarques:
To handle form validation in Angular 2 we have two ways : Model Driven and Template Driven.
Template Driven is what AngularJS did, we use HTML5 attributes to set constraints on our inputs and display validation messages depending of the error.
Model Driven is new. We define each rule in Component code, in TypeScript.

I chose to use Template Driven as it's easier to do from NG1 and is easier to understand for my point of view.

Here is an article about those 2 systems : http://blog.angular-university.io/introduction-to-angular-2-forms-template-driven-vs-model-driven/

### DIFF
--- a/generators/client-2/templates/_package.json
+++ b/generators/client-2/templates/_package.json
@@ -11,6 +11,7 @@
     "@angular/common": "2.0.0-rc.4",
     "@angular/compiler": "2.0.0-rc.4",
     "@angular/core": "2.0.0-rc.4",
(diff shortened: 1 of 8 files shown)


deepu105:
:+1:


Thanks & Regards,
Deepu

sendilkumarn:
obviously template driven 

module driven is `super awesomely` confusing

PR #3900, opened 2016-07-29
Upgrade to Spring Boot 1.4

mraible:
Update other dependencies to latest release as well.

PR #3966, opened 2016-08-11
revert liquibase to fix mvn:liquibase diff command

ruddell:
Starting from the liquibase version bump in #3900, running `mvn liquibase:diff` on a project throws the following error:

`[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:3.5.1:diff (default-cli) on project mastertest: Execution default-cli of goal org.liquibase:liquibase-maven-plugin:3.5.1:diff failed: An API incompatibility was encountered while executing org.liquibase:liquibase-maven-plugin:3.5.1:diff: java.lang.NoSuchMethodError: liquibase.structure.core.ForeignKey.addForeignKeyColumn(Lliquibase/structure/core/Column;)V`

According to https://github.com/liquibase/liquibase-hibernate/issues/117, reverting liquibase from 3.5 -> 3.4.2 is the only option to fix this for now.

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -23,7 +23,7 @@ geronimo_javamail_1_4_mail_version=1.8.4<% if (clusteredHttpSession == 'hazelcas
 hazelcast_version=3.6.4<% } %>
 hibernate_entitymanager_version=4.3.11.Final
 HikariCP_version=2.4.6<% if (databaseType == 'sql') { %>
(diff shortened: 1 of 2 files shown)


jdubois:
Good catch! Can you also do it for Gradle?

ruddell:
Gradle was still on liquibase v3.4.2, but liquibase_slf4j was behind the pom.xml so I upgraded it to match.

jdubois:
Great, thanks!

Issue #3574, opened 2016-05-10
[FEATURE] Add git commit data to /info endpoint


PierreBesson:
The [Spring Boot Actuator Docs §45.7.2](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html#production-ready-git-commit-information) mention a way to add git commit and branch information to the /info endpoint.

As all management endpoints are now secured by default, it would be interesting to have this feature. This is making use of [ktoso/maven-git-commit-id-plugin](https://github.com/ktoso/maven-git-commit-id-plugin) for maven and [com.gorylenko.gradle-git-properties](https://plugins.gradle.org/plugin/com.gorylenko.gradle-git-properties) for gradle.

It would be very interesting to have this, especially for microservices. Ideally, this info could be displayed in the JHipster Registry.

PierreBesson:
Any feedback on this ?

Another reason why I would like to add this is to enhance our monitoring (with the JHipster-Console) it would enable to know the which version of an app produced a certain log or metric (This would be achieved by adding an option to add this info to every log). This would be useful for the monitoring of canaries, feature flags, etc...

Moreover I don't see how it could have a negative impact for people that don't want to use it.

gmarziou:
Well git commit ids are quite unreadable, I'd rather have the artifact version as shown in same doc, this would work for released artifacts only though.

notflorian:
Is it possible to add both? Either way, I'm ok with this change. ;)

PierreBesson:
OK, by default, adding those 2 lines to the pom: 
```
            <plugin>
                <groupId>pl.project13.maven</groupId>
                <artifactId>git-commit-id-plugin</artifactId>
            </plugin>
```
Will add this to the /info
```
{
	git: {
		branch: "master",
		commit: {
			id: "191793c",
			time: "2016-05-17T15:49:03+0200"
		}
	}
}
```
@gmarziou, Initially, I also wanted to have the git tags but it's not provided by Spring boot by default so we have to add some code to do it... But actually, it might be cleaner to use the artifact pom as you suggest. It would also support other SCM.

We could try out the #property# maven substitution instead of @property@ that was used by @deepu105 on #3587. If it could work it would be awesome. Just adding something like `info.version: #project.version#` would be enough to have the data show up in `/info`

gmarziou:
[...]

PR #3970, opened 2016-08-12
Add Spring Boot build info to maven and to gradle

gmarziou:
/management/info shows build info like version, date, ...

Use same dependencies for both maven and gradle.
Add mariadb driver to gradle properties.

Ref #3574

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -52,6 +52,7 @@ war {
 springBoot {
     mainClass = '<%= packageName %>.<%= mainClass %>'
     executable = true
+    buildInfo()
 }
 
 bootRun {
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -18,9 +18,6 @@ logstash_logback_encoder_version=4.7
 <%_ if (databaseType === 'cassandra' || applicationType === 'gateway') { _%>
 datastax-driver.version=3.0.1
 <%_ } _%>
-<%_ if (databaseType == 'cassandra' || applicationType == 'gateway') { _%>
-datastax_driver_version=3.0.1
-<%_ } _%>
 javax_inject_version=1
 javax_transaction_version=1.2
 json_path_version=0.9.1
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -6,33 +6,60 @@ awaility_version=1.7.0
 commons_codec_version=1.10
 <%_ } _%>
 commons_lang_version=2.6
-commons_io_version=2.5<% if (databaseType == 'cassandra') { %>
-cassandra_unit_spring_version=3.0.0.1<% } %><% if (testFrameworks.indexOf('cucumber') != -1) { %>
-cucumber_version=1.2.4<% } %>
+commons_io_version=2.5
+<%_ if (databaseType === 'cassandra') { _%>
+cassandra_unit_spring_version=3.0.0.1
+<%_ } _%>
+<%_ if (testFrameworks.indexOf('cucumber') != -1) { _%>
+cucumber_version=1.2.4
+<%_ } _%>
 dropwizard_metrics_version=3.1.2
 logstash_logback_encoder_version=4.7
-<% if (databaseType == 'cassandra' || applicationType == 'gateway') { %>
-datastax_driver_version=3.0.1<% } %>
+<%_ if (databaseType === 'cassandra' || applicationType === 'gateway') { _%>
+datastax-driver.version=3.0.1
+<%_ } _%>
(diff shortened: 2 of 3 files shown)


gmarziou:
Wait before merging.

I have some strange issues on my environment, it seems that `npm link` does not always work as expected, when I run `yo jhipster` it seems I get the globally installed generator rather than my linked clone. Due to this, I'm not sure what I'm testing.

Can someone test it?

ruddell:
Generating a project puts `"generator-jhipster": "3.5.1",` into the package.json and installs it into the local node_modules folder.  After the initial generation, the project uses that version.  What I've been doing is removing it from package.json and node_modules/generator-jhipster in order to use the one installed with `npm link`, if there's a better way someone please let me know because it's kind of a pain.

I get an empty JSON array for any combination of options that I try - I tried maven/gradle for both session/JWT, also tested both monolith/gateway options.

deepu105:
This is how `npm link` works

when you run `npm link` on the actual repository code it symlinks the global node_modules version to point to this folder, so when we run `yo jhipster` node always will use the repository pointed by symlink from the global module.
Now once you generate an app you are installing a local version of JHipster as well and any run on the generated app folder after would use the local version which doesnt know of the symlink.

Now to overcome this you need to run `npm link generator-jhipster` on the generated project folder as well so that the local version is symlinked to the global version which points to our local repo.

Hope thats clear so in short to ensure you are using local repo
1. run `npm link` on the local repo folder
2. run  `npm link generator-jhipster` on the generated app folder (you need to do this for each application you create)

jdubois:
Thanks @deepu105 !! I totally agree this is annoying, or least not intuitive. And I was doing the symlink manually, so thanks for the tip.
I'm updating https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#setup with this info

gmarziou:
Thanks @deepu105, I didn't know this trick either.
Thanks @ruddel for testing, I also get an empty json object while it was working when I first tried with a project that was generated with older JHipster version, I'll investigate further.

gmarziou:
Maven: it does not work with `mvn spring-boot:run`, it works only when the jar or war is created so basically by running `mvn package` before or by running `mvn spring-boot:build-info`. maybe there could be some configuration to attach it to another phase but this is not [what is documented](http://docs.spring.io/spring-boot/docs/1.4.0.RELEASE/reference/htmlsingle/#howto-build-info) 

Gradle: it does not work though I followed the doc but I must admit I know nothing about gradle. So I need help.

PierreBesson:
@gmarziou, on the need to run the build-info phase, I don't think there is something we can do about it. Similar to when you run the Application class in the IDE, there is no clean way that it could refresh the build info details. However the goal of the feature is mostly useful to track the build version of a bundled artifact like a jar or a docker image. For dev, people will have to understand how it work (maybe add some docs about it). I think we can leave it like that for now.

I will try to have a look at gradle.

gmarziou:
Yes it sounds reasonable.

PierreBesson:
@gmarziou, I just had a look and tried to add the buildInfo() to the gradle build but it didn''t work for me either. I must have missed something because it does work in the [spring-boot-sample-actuator](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples/spring-boot-sample-actuator). I didn't manage to have the `build/resource/META-INF/build-info.properties` generated with gradle.

atomfrede:
Do we overwrite the ``processResources`` in gradle?

gmarziou:
As far as I know resources are not involved here but I may be wrong.

The actuator example showed me on thing, is that you can have several `springBoot` blocks in build.gradle, in [my commit I added `buildInfo()` to the existing one](https://github.com/jhipster/generator-jhipster/pull/3970/files#diff-5595eed51dd01e38e024547a86042a81) we have maybe that's a mistake.

gmarziou:
Gradle: I tried spring boot 1.4.1 snapshot, no success

I took a trace in debug mode `(./gradlew -d jar`) and now I think it is related to resources as you suggested @atomfrede 

```
00:29:14.538 [INFO] [org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter] Executing task ':bootBuildInfo' (up-to-date check took 0.044 secs) due to:
  Output file /home/gael/jhipster/infog/build/resources/main/META-INF/build-info.properties has changed.
  Output file /home/gael/jhipster/infog/build/resources/main/META-INF/build-info.properties has been removed.
```

I checked t(he file does not exist at the end of the build.

Here is the full log in case, some gradle expert has some suggestion:

[debug.txt](https://github.com/jhipster/generator-jhipster/files/421481/debug.txt)

gmarziou:
Closing this PR as I can't make it work for gradle.

I have submitted a PR #4010 for maven only.

PR #3976, opened 2016-08-12
Improve user entity relationship question

### LINKED ISSUE #3975: Improve entity relationship question for the User entity
As mentioned in the [managing relationships page](http://jhipster.github.io/managing-relationships/), the `User` entity is specific, and not all relationships work with it.

So when it is selected, we should only allow the working relationships:

- many-to-one
- many-to-many (owner)
- one-to-one (owner)

wmarques:
Fix #3975


deepu105:
LGTM

wmarques:
Strange that I have some "merge" commits appearing...

PR #3979, opened 2016-08-12
Fix some typescripts errors

wmarques:


wmarques:
Seems that there is a missing file..

PR #3983, opened 2016-08-12
fix ng2 tests due to renamed file

ruddell:
This should fix the ng2 tests for the other PRs (#3979, #3972), handleErrors.js is now handle-errors.js, file was renamed https://github.com/jhipster/generator-jhipster/commit/0e69ca9a0a59178fddef4dc23ecff8066aa8007c


ruddell:
I also copied over the rest of @deepu105's polishing and split of gulp copy/inject tasks into their own file to match master

deepu105:
Actually we dont have to copy changes to client 1 from master here coz
newer commits will be retained anyway when we merge so as long as we leave
client 1 in angular branch untouched its easier to merge

Thanks & regards,
Deepu

ruddell:
Good to know. So if I fix a bug in client-1 (like the websockets-uaa PR #3972), don't change client-1 in the angular-2 branch, just the client-2 code?

To fix the tests, I see two options.  The first is to merge master - it's looking for the renamed files but they aren't in the angular-2 branch yet. The other option is to change the client-1 section of test-creation.js back from https://github.com/jhipster/generator-jhipster/commit/3094dab5216afdeec23d620a42defa973c4094ee, but it will just break next time master is merged.  I'll let you decide how to handle this.

PR #3985, opened 2016-08-13
Java main class name backward compatible and avoiding compilation issue

### LINKED ISSUE #3889: Numbers in a project name cause Java compilation issues
If I choose `21points` as my application base name, there's compilation problems. For example:

```
:compileJava
/Users/mraible/dev/21-points/src/main/java/org/jhipster/health/21PointsApp.java:27: error: <identifier> expected
public class 21PointsApp {
```

I'd suggest reverting to the previous behavior, where `Application` is used. I'll create a PR.

gmarziou:
PR #3890 introduced a backward compatibility for projects generated since JHipster 3.x

As discussed in the issue, I propose an alternative fix that preserves backward compatibility, is motre convenient for microservices architecture and fixes #3889 with base names that are not valid java classes.

Fix #3889

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1290,9 +1290,11 @@ Generator.prototype.getAngularAppName = function () {
  * get the java main class name.
  */
 Generator.prototype.getMainClassName = function () {
-    // Don't name by baseName because numbers can cause compilation issues.
-    // https://github.com/jhipster/generator-jhipster/issues/3889
-    return 'Application';
+     
+    var main = _.upperFirst(this.getAngularAppName());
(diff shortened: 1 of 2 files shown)


mraible:
What happens when '21points' is used for the name? Does it use 'Application' instead or does it throw an error?

gmarziou:
Yes

mraible:
LGTM 👍  Did not test.

deepu105:
Lgtm

jdubois:
Good idea! As this wasn't released yet, it's really good to solve this now.
-> I just had the same issue from @saturnism so indeed that's annoying

PR #3987, opened 2016-08-14
switch health-modal.controller.ts and health.modal.controller.js code

ruddell:
Client 1 app is broken on the angular-2 branch, this switches the `ts` and `js` code to be in their correct folders


deepu105:
could you check the conflict here?

ruddell:
Is this even needed or will it be fixed later when you merge?

PR #3997, opened 2016-08-17
Handling exception with null cause

### LINKED ISSUE #3921: Missing cause and exception in LoggingAspect output
When LoggingAspect fires exception, the cause is missing resulting in something like

    2016-08-02 18:52:30.807 ERROR 23781 --- [nio-8080-exec-3] c.garanteasy.aop.logging.LoggingAspect   : Exception in com.garanteasy.web.rest.AccountResource.requestPasswordReset() with cause = {} and exception {}
    
    org.springframework.transaction.TransactionSystemException: Could not commit JPA transaction; nested exception is javax.persistence.RollbackException: Error while committing the transaction

lrkwz:
Exceptions can hold a null cause

> Returns the cause of this throwable or null if the cause is nonexistent or unknown.  (The cause is the throwable that caused this throwable to get thrown.) (javadoc)

so it is better to check this BEFORE accessing in in ```e.getCause().getMessage()```

This was caused by my first attempt to fix #3921


deepu105:
This doesnt have a clean commit history

jdubois:
- I still don't understand the issue: can you open up a ticket, and give an example of the issue, and your solution?
- Travis is failing

lrkwz:
Fixed the error, travis should be ok now.

@jdubois this could be an example for you:

```java
        try{
            int a = 1/0;
        } catch( Exception e ){
            System.out.println( e.getCause() );
        }
```

this outputs ```null``` so, when hooked by the aspect will sadly fail.

lrkwz:
Since @deepu105 told tha there is no clean history I've switched my fork to a "feature branch" versioning schema.
This PR can be closed and replaced by #4046

PR #4002, opened 2016-08-18
use versions from generator constants in kubernetes subgenerator

ruddell:


jdubois:
Oh good catch!

And I missed the JHipster Registry update, as it was already one version behind...

-> this should definitely trigger a new minor release

cc @saturnism

pascalgrimaud:
Yeah, it's my fault.
When I improved the kubernetes subgenerator, I had to find the correct version of mysql, psql, mongo, etc. So I did https://github.com/jhipster/generator-jhipster/pull/3867 but I forgot to change the kubernetes subgenerator

PR #4013, opened 2016-08-21
make searches send full queries to API

### LINKED ISSUE #3805: Send full elasticsearch queries to server instead of just the user's query string

Right now the Elasticsearch support works by sending the user's query to the server, then letting the server turn that into a QueryStringQuery. This works fine for simple searches and gives the user some control over more advanced searches, but makes it very difficult for the developer.

For example, if I wanted to change `default_operator` to `AND` for a specific query I need to insert `+` or `AND` before each search term. The proper way to do this is to set `default_operator` on the ES query itself, but with the current solution that would need to be done server-side then can't be applied on a per-query basis. 

If we instead change it so the client app sends a full ES query to the server, it becomes much easier for the developer to fine-tune individual queries.

A common scenario is searching on a specific field. This can be done with queryStringQuery alone:
```
'lastName:test'
```

What if we need to search on more than one field?
```
'lastName:test firstName:test'
```

What if the user searches by more than one word?
```
'lastName:test1 lastName:test2 firstName:test1 firstName:test2'
```

All search terms need to match, but can match either field?
```
'(lastName:test1 AND firstName:test2) OR (lastName:test2 AND firstName:test1) OR (lastName:test1 AND lastName:test2) ... '
```

QueryStringQuery is powerful but can quickly become too complex to maintain. The other downside is that if the user types reserved characters or otherwise causes a syntax error, the server will return an error. Because of this user queries are generally SimpleQueryStringQuery but then you can't use any advanced search functions like searching by field!

##### **Suggest a Fix** 

Instead of sending the user's query to the client app then using it as a QueryStringQuery directly, we can send a full ES query then use it as a WrapperQuery. This gives developers a LOT more control over ES queries and avoids errors. Ex:
```
EntitySearch.query({
                query: {
                    simple_query_string: {
                        query: $scope.currentSearch,
                        fields: ['firstName', 'lastName'],
                        default_operator: 'AND'
                    }
                },
                page: pagingParams.page - 1,
                size: 20,
                sort: [$scope.predicate + ',' + ($scope.reverse ? 'asc' : 'desc'), 'id']
            }, onSuccess, onError);
```

```
    @RequestMapping(value = "/_search/entities",
            method = RequestMethod.GET,
            produces = MediaType.APPLICATION_JSON_VALUE)
    @Timed
    public ResponseEntity<List<Entity>> searchEntities(@RequestParam String query, Pageable pageable) throws URISyntaxException, UnsupportedEncodingException {
        ...
        Page<Entity> page = entitySearchRepository.search(wrapperQuery(query), pageable);
        ...
    }
```
This is equivalent to the last query above and will NOT throw errors if the user enters incorrect syntax. 

Using this as the query will match current behaviour:
```
                query: {
                    query_string: {
                        query: $scope.currentSearch
                    }
                },
```

By the way, I am using a modified pre-3.0 version of JHipster so there may be discrepancies in my samples like ES syntax or code style.

geraldhumphries:
modify Elasticsearch support to send full queries to the generated API
instead of just the search string for a server-side QueryStringQuery

fix #3805

### DIFF
--- a/generators/entity/templates/src/main/webapp/app/entities/infinite-scroll-template.ejs
+++ b/generators/entity/templates/src/main/webapp/app/entities/infinite-scroll-template.ejs
@@ -29,7 +29,11 @@
         <%_ } else { _%>
             if (vm.currentSearch) {
                 <%= entityClass %>Search.query({
-                    query: vm.currentSearch,
+                    query: {
+                        simple_query_string: {
+                            query: vm.searchQuery
+                        }
+                    },
                     page: vm.page,
                     size: 20,
                     sort: sort()
--- a/generators/entity/templates/src/main/java/package/common/search_stream_template.ejs
+++ b/generators/entity/templates/src/main/java/package/common/search_stream_template.ejs
@@ -3,7 +3,7 @@ var mapper = entityInstance  + 'Mapper';
 var entityToDtoReference = mapper + '::'+ entityInstance +'To' + entityClass + 'DTO'; %>
         <%_ if (!viaService) { _%>
         return StreamSupport
-            .stream(<%= entityInstance %>SearchRepository.search(queryStringQuery(query)).spliterator(), false)<% if (dto == 'mapstruct') { %>
+            .stream(<%= entityInstance %>SearchRepository.search(wrapperQuery(query)).spliterator(), false)<% if (dto == 'mapstruct') { %>
             .map(<%= entityToDtoReference %>)<% } %>
             .collect(Collectors.toList());
         <%_ } else { _%>
--- a/generators/entity/templates/src/main/java/package/common/search_template.ejs
+++ b/generators/entity/templates/src/main/java/package/common/search_template.ejs
@@ -12,7 +12,7 @@
         throws URISyntaxException {
         log.debug("REST request to search for a page of <%= entityClassPlural %> for query {}", query);<% if (viaService) { %>
         Page<<%= instanceType %>> page = <%= entityInstance %>Service.search(query, pageable);<% } else { %>
-        Page<<%= entityClass %>> page = <%= entityInstance %>SearchRepository.search(queryStringQuery(query), pageable);<% } %>
+        Page<<%= entityClass %>> page = <%= entityInstance %>SearchRepository.search(wrapperQuery(query), pageable);<% } %>
         HttpHeaders headers = PaginationUtil.generateSearchPaginationHttpHeaders(query, page, "/api/_search/<%= entityApiUrl %>");
         return new ResponseEntity<>(<% if (!viaService && dto == 'mapstruct') { %><%= entityListToDtoListReference %>(page.getContent())<% } else { %>page.getContent()<% } %>, headers, HttpStatus.OK);
     <% } -%>
--- a/generators/entity/templates/src/main/java/package/service/impl/_EntityServiceImpl.java
+++ b/generators/entity/templates/src/main/java/package/service/impl/_EntityServiceImpl.java
@@ -110,7 +110,7 @@ public class <%= serviceClassName %> <% if (service == 'serviceImpl') { %>implem
         log.debug("Request to search <%= entityClassPlural %> for query {}", query);<%- include('../../common/search_stream_template', {viaService: viaService}); -%>
         <%_ } else { _%>
         log.debug("Request to search for a page of <%= entityClassPlural %> for query {}", query);
-        Page<<%= entityClass %>> result = <%= entityInstance %>SearchRepository.search(queryStringQuery(query), pageable);
+        Page<<%= entityClass %>> result = <%= entityInstance %>SearchRepository.search(wrapperQuery(query), pageable);
             <%_ if (dto == 'mapstruct') { _%>
         return result.map(<%= entityInstance %> -> <%= entityToDto %>(<%= entityInstance%>));
             <%_ } else { _%>
--- a/generators/entity/templates/src/main/webapp/app/entities/no-pagination-template.ejs
+++ b/generators/entity/templates/src/main/webapp/app/entities/no-pagination-template.ejs
@@ -21,7 +21,13 @@
             if (!vm.searchQuery) {
                 return vm.loadAll();
             }
-            <%= entityClass %>Search.query({query: vm.searchQuery}, function(result) {
+            <%= entityClass %>Search.query({
+                query: {
+                    simple_query_string: {
+                        query: vm.searchQuery
+                    }
+                }
+            }, function(result) {
                 vm.<%= entityInstancePlural %> = result;
             });
         }<%_ } _%>
--- a/generators/entity/templates/src/main/webapp/app/entities/pagination-template.ejs
+++ b/generators/entity/templates/src/main/webapp/app/entities/pagination-template.ejs
@@ -27,7 +27,11 @@
         <%_ } else { _%>
             if (pagingParams.search) {
                 <%= entityClass %>Search.query({
-                    query: pagingParams.search,
+                    query: {
+                        simple_query_string: {
+                            query: vm.searchQuery
+                        }
+                    },
                     page: pagingParams.page - 1,
                     size: vm.itemsPerPage,
                     sort: sort()
--- a/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
+++ b/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
@@ -461,8 +461,11 @@ public class <%= entityClass %>ResourceIntTest <% if (databaseType == 'cassandra
         <%= entityInstance %>SearchRepository.save(<%= entityInstance %>);
 <%_ } _%>
 
+        // Build the search query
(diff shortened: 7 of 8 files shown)


geraldhumphries:
Only thing I don't like is how many times the word "query" is used. Maybe the param should be named something else?

jdubois:
You broke the build :-)
Yes all those `query` is annoying, but that also comes from the Elasticsearch API, the name `queryStringQuery` is really bad...

deepu105:
looks like the tests are failing with below
```
Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.438 sec <<< FAILURE! - in com.mycompany.myapp.web.rest.TestServiceImplResourceIntTest
searchTestServiceImpl(com.mycompany.myapp.web.rest.TestServiceImplResourceIntTest)  Time elapsed: 0.04 sec  <<< ERROR!
org.springframework.web.util.NestedServletException: Request processing failed; nested exception is Failed to execute phase [query], all shards failed; shardFailures {[PAKHGJZdSMahhgRRBU6AXQ][testserviceimpl][0]: RemoteTransportException[[Ahura][local[1]][indices:data/read/search[phase/query]]]; nested: SearchParseException[failed to parse search source [{"query":{"wrapper":{"query":"aWQ6MQ=="}}}]]; nested: ElasticsearchParseException[Failed to derive xcontent]; }{[PAKHGJZdSMahhgRRBU6AXQ][testserviceimpl][1]: RemoteTransportException[[Ahura][local[1]][indices:data/read/search[phase/query]]]; nested: SearchParseException[failed to parse search source [{"query":{"wrapper":{"query":"aWQ6MQ=="}}}]]; nested: ElasticsearchParseException[Failed to derive xcontent]; }{[PAKHGJZdSMahhgRRBU6AXQ][testserviceimpl][2]: RemoteTransportException[[Ahura][local[1]][indices:data/read/search[phase/query]]]; nested: SearchParseException[failed to parse search source [{"query":{"wrapper":{"query":"aWQ6MQ=="}}}]]; nested: ElasticsearchParseException[Failed to derive xcontent]; }{[PAKHGJZdSMahhgRRBU6AXQ][testserviceimpl][3]: RemoteTransportException[[Ahura][local[1]][indices:data/read/search[phase/query]]]; nested: SearchParseException[failed to parse search source [{"query":{"wrapper":{"query":"aWQ6MQ=="}}}]]; nested: ElasticsearchParseException[Failed to derive xcontent]; }{[PAKHGJZdSMahhgRRBU6AXQ][testserviceimpl][4]: RemoteTransportException[[Ahura][local[1]][indices:data/read/search[phase/query]]]; nested: SearchParseException[failed to parse search source [{"query":{"wrapper":{"query":"aWQ6MQ=="}}}]]; nested: ElasticsearchParseException[Failed to derive xcontent]; }
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:982)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:861)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:622)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:846)
	at org.springframework.test.web.servlet.TestDispatcherServlet.service(TestDispatcherServlet.java:65)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:729)
	at org.springframework.mock.web.MockFilterChain$ServletFilterProxy.doFilter(MockFilterChain.java:167)
	at org.springframework.mock.web.MockFilterChain.doFilter(MockFilterChain.java:134)
	at org.springframework.test.web.servlet.MockMvc.perform(MockMvc.java:155)
	at com.mycompany.myapp.web.rest.TestServiceImplResourceIntTest.searchTestServiceImpl(TestServiceImplResourceIntTest.java:203)
Caused by: org.elasticsearch.action.search.SearchPhaseExecutionException: all shards failed
	at org.elasticsearch.action.search.AbstractSearchAsyncAction.onFirstPhaseResult(AbstractSearchAsyncAction.java:206)
	at org.elasticsearch.action.search.AbstractSearchAsyncAction$1.onFailure(AbstractSearchAsyncAction.java:152)
	at org.elasticsearch.action.ActionListenerResponseHandler.handleException(ActionListenerResponseHandler.java:46)
	at org.elasticsearch.transport.TransportService$DirectResponseChannel.processException(TransportService.java:855)
	at org.elasticsearch.transport.TransportService$DirectResponseChannel.sendResponse(TransportService.java:833)
	at org.elasticsearch.transport.TransportService$4.onFailure(TransportService.java:387)
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:39)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.elasticsearch.ElasticsearchParseException: Failed to derive xcontent
	at org.elasticsearch.common.xcontent.XContentFactory.xContent(XContentFactory.java:201)
	at org.elasticsearch.common.xcontent.XContentFactory.xContent(XContentFactory.java:192)
	at org.elasticsearch.index.query.WrapperQueryParser.parse(WrapperQueryParser.java:60)
	at org.elasticsearch.index.query.QueryParseContext.parseInnerQuery(QueryParseContext.java:250)
	at org.elasticsearch.index.query.IndexQueryParserService.innerParse(IndexQueryParserService.java:320)
	at org.elasticsearch.index.query.IndexQueryParserService.parse(IndexQueryParserService.java:223)
	at org.elasticsearch.index.query.IndexQueryParserService.parse(IndexQueryParserService.java:218)
	at org.elasticsearch.search.query.QueryParseElement.parse(QueryParseElement.java:33)
	at org.elasticsearch.search.SearchService.parseSource(SearchService.java:838)
	at org.elasticsearch.search.SearchService.createContext(SearchService.java:654)
	at org.elasticsearch.search.SearchService.createAndPutContext(SearchService.java:620)
	at org.elasticsearch.search.SearchService.executeQueryPhase(SearchService.java:371)
	at org.elasticsearch.search.action.SearchServiceTransportAction$SearchQueryTransportHandler.messageReceived(SearchServiceTransportAction.java:368)
	at org.elasticsearch.search.action.SearchServiceTransportAction$SearchQueryTransportHandler.messageReceived(SearchServiceTransportAction.java:365)
	at org.elasticsearch.transport.TransportRequestHandler.messageReceived(TransportRequestHandler.java:33)
	at org.elasticsearch.transport.RequestHandlerRegistry.processMessageReceived(RequestHandlerRegistry.java:75)
	at org.elasticsearch.transport.TransportService$4.doRun(TransportService.java:376)
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:37)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

geraldhumphries:
Oops, that's what I get for 4am commits.

Tests should be fixed now, hopefully. You'll notice that I had to separate the query into its own string then add it as a variable to the URL. This is because the URL builder sees everything between `{}` as a placeholder for variables, so putting the JSON there directly was failing.

generators/entity/templates/src/main/webapp/app/entities/infinite-scroll-template.ejs
```diff
@@ -29,7 +29,11 @@
         <%_ } else { _%>
             if (vm.currentSearch) {
```
deepu105: what is the purpose of `vm.currentSearch`? its always same as `vm.searchQuery` can we remove the variable `currentSearch` altogether and just use `searchQuery` everywhere?

geraldhumphries: Using `searchQuery` here was a mistake, it should be `currentSearch`. I think `currentSearch` is used because without it, if a user changes the text in the search box without performing a search, the search will be performed any time the user sorts the current results or changes pages.

geraldhumphries: In the case of infinite scrolling, if the user typed something in the search box then scrolled down to load the next page, they would get page 2 of the search results for whatever they searched instead of what they were expecting.

geraldhumphries: Now that I look over the controllers for other pagination types I think we should make them more consistent. Maybe `currentSearch` can be removed now that pagingParams exists - testing will need to be done.

geraldhumphries:
I had some second thoughts about this and decided to open a question on SO about it: http://stackoverflow.com/questions/39103547/is-it-safe-to-expose-get-only-elasticsearch-queries-directly-through-your-api

If it turns out to be a security concern then I will close this.

jdubois:
The SO question seems to point to a security issue, `You can still crash Elasticsearch if you know how to do it.`... So unfortunately we need to close this.
But the idea is very good, and this is overall much better than what we currently have -> this is the right direction, but there is still work to make it happen.

PR #4018, opened 2016-08-22
Add git commit data to /info endpoint in prod build

### LINKED ISSUE #3574: [FEATURE] Add git commit data to /info endpoint

PierreBesson:
Finally fixes #3574 
cc @gmarziou
Add the following in the /info endpoint **only when doing a prod build**
```
	"git": {
		"commit": {
			"id": {
				"abbrev": "324b701",
				"describe": "v1-2-324b701"//only for maven as the gradle git-commit plugin doesn't support it yet
			}
		},
		"branch": "master"
	}
```
What it does:
- Set `management.info.git.mode` to `full` this is needed so that every property present in the `git.properties` file will end up in /info. I think this is the natural way to use it since in the default `simple` mode it will only put the commit id and branch in /info.
- Setup the git-commit plugin for both maven and gradle and restrict it to a subset of git properties : `git.commit.abbrev`, `git.branch` and `git.commit.id.describe` (unfortunately not available for gradle).
Another advantage of restricting the properties is that this way there won't be any unwanted properties like commit user name or email in the `git.properties` file that ship with the war.

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -16,6 +16,7 @@ buildscript {
         classpath "com.moowork.gradle:gradle-gulp-plugin:0.12"<% } %>
         classpath "se.transmode.gradle:gradle-docker:1.2"
         classpath "io.spring.gradle:dependency-management-plugin:0.5.6.RELEASE"
+        classpath "gradle.plugin.com.gorylenko.gradle-git-properties:gradle-git-properties:1.4.17"
         //jhipster-needle-gradle-buildscript-dependency - JHipster will add additional gradle build script plugins here
     }
 }
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1225,6 +1225,20 @@
                             </execution>
                         </executions>
                     </plugin>
+                    <plugin>
+                        <groupId>pl.project13.maven</groupId>
+                        <artifactId>git-commit-id-plugin</artifactId>
+                        <version>2.2.1</version>
+                        <configuration>
+                            <verbose>false</verbose>
+                            <failOnNoGitDirectory>false</failOnNoGitDirectory>
+                            <includeOnlyProperties>
+                                <includeOnlyProperty>^git.commit.id.abbrev$</includeOnlyProperty>
+                                <includeOnlyProperty>^git.commit.id.describe$</includeOnlyProperty>
+                                <includeOnlyProperty>^git.branch$</includeOnlyProperty>
+                            </includeOnlyProperties>
+                        </configuration>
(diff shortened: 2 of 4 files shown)


gmarziou:
Looks good to me as it uses plugins referenced by Spring boot doc.

jdubois:
I just tried it and it's not working very well. This adds dependencies on 2 new plugins, so that adds complexity and maintenance, and I'd like first to have a couple of things solved:

- I can't get it to work (with Maven). I'm guessing the data should appear in `management/info` ?
- We have no UI to see what this provides: where should this go? In the `config` screen? Without a screen to see that data, it's rather useless.

PierreBesson:
I agree that the fact that it adds dependencies is an issue. However I think it adds interesting data for microservice deployments especially to do continuous delivery.

And yes it doesn't do anything as long as you have not initialized the git repo of the app, then data will appear in /management/info.

I agree that this should be displayed in the UI but for now client side development is freezed waiting for angular 2. But I would like to display this info in the registry.

Another use I have for this is to add the git version number to the logs. So that when doing blue/green deployment, you can distinguish different versions of the same service.

jdubois:
Yes I did initialize a repo and did a prod package, but my "info" endpoint was still empty.

PierreBesson:
Did you do a commit ?

jdubois:
Oh forget about my comment, it's working now.

But still:

- We need show this somewhere on the UI
- It must be documented

gmarziou:
Don't you think that the spring boot documentation is enough? Pierre did only use what is recommended in their doc including these plugins.

jdubois:
"If a tree falls in a forest and no one is around to hear it, does it make a sound?"

I think we should document non-obvious things, otherwise people will not know about it

pascalgrimaud:
@PierreBesson : do you think this feature can be in a module ?

PierreBesson:
Yes it could but if is not setup by default then I cannot get the version info in the JHipster Console which was the initial reason for developing this.

PR #4019, opened 2016-08-22
upgrade to gradle 3.0

### LINKED ISSUE #3994: upgrade to gradle 3
upgrade to gradle 3.0
https://docs.gradle.org/current/release-notes

the deamon claims to be giving 75% performance boost. may be we should enable it by default?

atomfrede:
I've tested so far the following:

* [x] Monolithic App (with bootRun)
* [x] Gatling Tests
* [x] Docker builds
* [x] Generate production war file
* [x] Microservices App
* [x] Gateway

fix #3994


atomfrede:
I have tried also microservices and the gateway. Everything looks fine. The only quetion is, how to handle the enabled daemon. Using it for local development is no problem and the gradle team also says it should be no problem to have the daemon enabled in a ci environment like jenkins (using travis or gitlab ci should be no problem as long the builds are executed in a dedicated container). 

I suggest we keep it enabled und just put a hint in the generated readme.

deepu105:
Yes we should enable it as it gives better speed. If anybody wants to
disable it they can do so by un commenting the property that we have

Thanks & regards,
Deepu

deepu105:
@atomfrede do you wanna do any further change or is it ready to merge?

atomfrede:
@deepu105 I don't have zny changes at the moment, so let's merge it to get more feedback

deepu105:
I have updated the comments on the property file accordingly

PR #4021, opened 2016-08-22
install aws subgenerator dependencies if they are missing

ruddell:
AWS Subgenerator fails silently because `generator` is set in the `try` after the error occurs.  

I also updated the paths to show the appropriate folder to install aws npm dependencies because monoliths/gateways use the local node_modules.  This should also be updated in the AWS documentation http://jhipster.github.io/aws/

### DIFF
--- a/generators/aws/lib/aws.js
+++ b/generators/aws/lib/aws.js
@@ -1,7 +1,6 @@
 'use strict';
 var S3 = require('./s3.js'),
     Rds = require('./rds.js'),
-    chalk = require('chalk'),
     shelljs = require('shelljs'),
     os = require('os'),
     Eb = require('./eb.js');
@@ -12,17 +11,16 @@ var AwsFactory = module.exports = function AwsFactory(generatorRef, cb) {
     generator = generatorRef;
     try {
         Aws = require('aws-sdk');
-        generator.log("Found AWS dependencies");
-        cb()
+        cb();
     } catch (e) {
-        generator.log("Installing AWS dependencies into your JHipster folder");
+        generator.log('Installing AWS dependencies into your JHipster folder');
         var jhipsterPath;
         var applicationType = generator.config.get('applicationType');
         if (applicationType === 'monolith' || applicationType === 'gateway') {
             if (os.platform() === 'win32') {
-                jhipsterPath = 'node_modules\\generator-jhipster\n';
+                jhipsterPath = 'node_modules\\generator-jhipster';
             } else {
-                jhipsterPath = 'node_modules/generator-jhipster\n';
+                jhipsterPath = 'node_modules/generator-jhipster';
             }
         } else {
             var prefix = shelljs.exec('npm config get prefix', {silent:true}).trim();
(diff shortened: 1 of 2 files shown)


deepu105:
I dont understand why there should be a difference. Monoliths and microservice will use local install if available else will look for global. Am i missing anything here?

campbellmcneill:
I'm still getting the silent problem. Trying to isolate it.

ruddell:
@deepu105 You are correct, apps use local install if available but the issue is that generator-jhipster is only installed locally for monoliths/gateways.

Currently the instructions say to install the AWs dependencies in the global generator-jhipster.  This works for microservices, but does not work for monoliths/gateways which use the local generator-jhipster.  That is why it differs.

To reproduce what I'm saying:
 - Generate a monolith
 - Run `yo jhipster:aws` (currently a silent error, but the docs currently say install in global node_modules/generator-jhipster)
 - Run `npm install aws-sdk progress node-uuid` in the global node_modules/generator-jhipster
 - Go back to the monolith and run `yo jhipster:aws`, same result (error)
 - Go to the local node_modules/generator-jhipster and install the AWS dependencies
 - Go back to the monolith root and run `yo jhipster:aws` and you finally get a prompt.

deepu105:
Ok understood you are correct  :)
may be instead of asking the user we could trigger the install against
correct folder as well if its not found :)

Thanks & regards,
Deepu

generators/aws/lib/aws.js
```diff
@@ -2,32 +2,41 @@
...
-        var windowsPath;
-        var linuxPath;
+        generator.log("Installing AWS dependencies into your JHipster folder");
+        var jhipsterPath;
         var applicationType = generator.config.get('applicationType');
         if (applicationType === 'monolith' || applicationType === 'gateway') {
-            windowsPath = 'cd node_modules\\generator-jhipster\n';
-            linuxPath = 'cd node_modules/generator-jhipster\n';
+            if (os.platform() === 'win32') {
+                jhipsterPath = 'node_modules\\generator-jhipster\n';
```
deepu105: Doesnt shell js handle the path accordingly for the OS?

ruddell:
Thanks for the tip @deepu105, code is a lot cleaner now.  I also changed the check on the applicationType to skipClient as that's the real factor.

generators/aws/lib/aws.js
```diff
@@ -1,25 +1,35 @@
...
-            chalk.green('cd /usr/local/lib/node_modules/generator-jhipster\n' +
-            'npm install aws-sdk progress node-uuid')
-        );
+        generator.log('Installing AWS dependencies into your JHipster folder');
+        var jhipsterPath = 'node_modules/generator-jhipster';
+        var skipClient = generator.config.get('skipClient');
+        // use the global JHipster for apps with no client-side code
+        if (skipClient) {
+            var prefix = shelljs.exec('npm config get prefix', {silent:true}).trim();
+            if (os.platform() === 'win32') {
```
deepu105: Is this required? doesnt shelljs take care of path for appropraie os?

ruddell: The global node_modules path has an extra `/lib/` on Unix compared to Windows.  ShellJS does not handle that.

deepu105: ok

PR #4022, opened 2016-08-22
fix add and remove methods for two relationships with same entity

ruddell:
With fluent methods, the following JDL generates two different `addDevice` and `removeDevice` in DeviceRelationship.java
```
entity Device
entity DeviceRelationship

relationship OneToMany {
  DeviceRelationship{originatingDevice} to Device{originatingDeviceRelationship},
  DeviceRelationship{destinationDevice} to Device{destinationDeviceRelationship}
}
```


cbornet:
We should add a test for this in CI !

PR #4029, opened 2016-08-25
Fix #4028 - AbstractCassandraTest CQL scripts order

### LINKED ISSUE #4028: AbstractCassandraTest does not apply the CQL scripts in expected order
The order of the execution of the Cassandra CQL scripts when the tests are started is not predictable and can cause the tests to fail.

When the AbstractCassandraTest is starting, it automatically applies all the scripts from the cql/changelog directory to prepare the tables.
The scripts are supposed to be applied sorted by their names, hence the date prefix
see [this loop](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/test/java/package/_AbstractCassandraTest.java#L49)
This is working fine on Mac and, I assume, most of the times in other OS.

However, in our Jenkins on linux the scripts I've noticed the scripts are not applied in the expected order, 000001_ is applied before 000000_ and causes a failure.

##### **Suggest a Fix**
Sort the list of files before iterating to apply the scripts.
I'll send a PR with that.

##### **JHipster Version(s)**
master

Linux

raphaelbrugier:
* Sort the scripts by their names before iterating

### DIFF
--- a/generators/server/templates/src/test/java/package/_AbstractCassandraTest.java
+++ b/generators/server/templates/src/test/java/package/_AbstractCassandraTest.java
@@ -17,6 +17,10 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 
 /**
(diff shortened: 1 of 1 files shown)


gmarziou:
Looks good to me, thanks for the PR.

PR #4042, opened 2016-08-29
Add closing tag for textarea

amr:
When generating an entity with a field of type CBlob (Text), we use a `textarea` for representing the field in a form. We generate a self-closing `<textarea ... />` field, but it's invalid to do so[1][2]. In my tests, the issue only came to my attention when running my application as an executable fat WAR with the production profile.

What would happen then is that the self-closing `texarea` would eat all the rest of my page contents, leaving me with many missing fields and no submit buttons.

YMMV. I've no idea why it only broke on production builds.

This pull requests fixes this by adding the closing tag `</textarea>`.

Reference:

1. Per [HTML5 spec](https://www.w3.org/TR/html-markup/syntax.html), void elements that are allowed to self-close are: `area, base, br, col, command, embed, hr, img, input, keygen, link, meta, param, source, track, wbr`
2. Google for "self-closing texarea"


stringaro-leo:
Thx @amr my gulp build is saved now, I don't need to edit form anymore

PR #4044, opened 2016-08-29
Kubernetes: distinct monolith and microservice applications

pascalgrimaud:
As suggested on Twitter by Julien, minor improvements to distinct monolith and microservice applications when using Kubernetes sub generator


jdubois:
Thanks @pascalgrimaud !
We also need to improve the doc on that.

PR #4048, opened 2016-08-31
Upgrade commons-lang

erikkemperman:
I was aligning some versions and noticed that jhipster depends on a [5.5 years old version](https://mvnrepository.com/artifact/commons-lang/commons-lang) of commons-lang :-)

Guess it happened because they sneakily changed the groupId and artifactId.

Anyway, this PR brings this dependency [up to date](https://mvnrepository.com/artifact/org.apache.commons/commons-lang3).


jdubois:
Oh yes, this config come from my own personal template project, when I first created the project! And indeed this is old!

erikkemperman:
Thanks for review and merge!

Issue #3982, opened 2016-08-12
UAA Microservice Entity Tests Fail


ruddell:
##### **Overview of the issue**
`mvn test` on a UAA microservice fails after generating an entity.  The error thrown is `Caused by: java.lang.IllegalStateException: No instances available for uaa` (my UAA folder is 'uaa' and it is running, though this shouldn't affect tests).  The application runs fine with `mvn`

**Note**: On v3.5.1, the error also mentions a liquibase problem, but the original cause is the IllegalStateException.  On master, the error only mentions the IllegalStateException.

[Master error stacktrace](http://pastebin.com/raw/C6SufEpL)
[v3.5.1 error stacktrace](http://pastebin.com/raw/S5fB4nzL)

##### **Reproduce the error**
Generate a UAA
Generate a microservice
Generate an entity in the microservice and run `mvn test`

##### **Suggest a Fix**
I'm new to the UAA so can't help much.

##### **JHipster Version(s)**
Master and v3.5.1

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**
All default options (except auth is UAA)
<details>
  <summary>Basic UAA App (Click to expand)</summary>
```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.5.1",
    "baseName": "uaa",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "9999",
    "authenticationType": "uaa",
    "hibernateCache": "hazelcast",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": "no",
    "buildTool": "maven",
    "enableTranslation": true,
    "applicationType": "uaa",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ],
    "clusteredHttpSession": "no",
    "websocket": "no",
    "enableSocialSignIn": false
  }
}
```
</details>

<details>
  <summary>Basic Microservice using UAA (Click to expand)</summary>
```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.5.1",
    "baseName": "errorservice",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8081",
    "authenticationType": "uaa",
[...]

PR #4052, opened 2016-09-01
adds test profile to resource tests and excludes uaa tokenstore beans on that

xetys:
please review if it is ok to use `@ActiveProfile` in the tests...

tried to build some setup for my current PR and docs on that, noticed that #3982 is present in all services, and added this manually to them in order to make it working


deepu105:
I dont like the idea of mixing test related profile and stuff into actual impl. Isnt there any other cleaner way?

xetys:
The only thing I now have in mind is to still use the profile annotation, but with "cloud" instead of "! Test" and force microservice applications to run using cloud profile. 

I think all these load balance dependence features have to be explicitly enabled, since they make the application not really testable, if beans are depending per default

xetys:
ok I decieded to add a profile "uaa" if we use authenticationType uaa. For all tests, if they are usual, they don't lead to the evil uaa beans get wired...this should fix this (and some more problems)

cbornet:
Maybe you can provide some @Primary beans in test to overridethe ones created for normal operation ?

xetys:
primary beans, in order to exclude actualy beans from main application? I would have to checkout how to manage that

xetys:
ok, i tried a simple approach: add a new configuration class where those evil beans are nulled (one naive but working way of disabling beans) and referring them in `@SpringBootTest(classess = ... )`

if this is a way you think is better then the current approach, I will change that to this. In this case, each test must now refer to that additional config class, but no profile magic anymore in actual implementation

xetys:
so this now works in a way as @cbornet asked

deepu105:
This looks much better. But you seem to have a travis failure also you might want to squash some commits

xetys:
i fixed the travis, so for me this is ready to be merged

PR #4063, opened 2016-09-03
[test-undertow] fix tomcat dependencies for gradle

pascalgrimaud:


generators/server/templates/gradle/_profile_prod.gradle
```diff
@@ -9,7 +9,7 @@ ext {
 }
 
 dependencies {
-    compile "org.springframework.boot:spring-boot-starter-tomcat:${spring_boot_version}"
+    compile "org.springframework.boot:spring-boot-starter-undertow:${spring_boot_version}"
```
atomfrede: What happens if we remove the web container dependencies from the profile gradle files? I think we can remove them, as they are defined in the global dependencies part.

pascalgrimaud: Yes, it works without them. You're right. I will remove these lines

pascalgrimaud: thanks again @atomfrede !

Issue #4058, opened 2016-09-02
@WithMockOAuth2Authentication doesn't work due to a security fix from spring


xetys:
Well, the next situation is quite complex, and needs some decision to go on

During my recent work I invented the above mentioned annotation in order to provide some tool to integration test security configuration, which is a very important feature for me, since it's a must have if you are TDD style... 

There also was a question, wether to PR it to spring, and here is some problem: That annotation, in the way I contributed it to JHipster, won't work at all. The reason why it worked to me in July, I simply forgot. It is the stateless flag in spring `OAuth2AuthenticationProcessingFilter`, which clears the security context, when it is present. This is done to fix some whole, where you can get accidentaly authenticated used some cookies (and I have no idea, how this is working, it is described in incode comments)...in other words, only real tokens lead to real authentication....the `@WithSecurityContext` approach is cutted away by spring itself

But, there ***would*** be a way to make it working: 

```  java

    @Override
    public void configure(ResourceServerSecurityConfigurer resources) throws Exception {
        if (Arrays.asList(environment.getActiveProfiles()).contains("test")) {
            resources.stateless(false);
        }
    }
```

in `MicroserviceSecurityConfiguration`. As of @deepu105 's opinion, this is mixing testing stuff into actual implementation. Disabling stateless mode at all, is a no way, as I will never sleep peacefully anymore if know, I was contributing some insecure hack in order to satisfy some needs for testing...

I'm already looking for a cleaner way to solve this, but at the moment, that is the only way I know which is working...so I need your opinion on this issue :smile:

deepu105:
I would rather skip a test case than opening a security hole :) but thats just my opinion.

xetys:
actually, this would only mean to maintain some hacky code, in your to satisfy both needs.

for the security hole thing I am currently discussing on gitter with Rob and Dave....as it currently seems the actual security considarion still exists, if no cookie based authenticated happen in that application....as we don't use cookies for microservice applications. I think if there is some security problem, then if using classic stateful security and oauth support in one application.
If this holds, spring could just add a configuration property, to make `private boolean stateless = true` in `OAuth2AuthenticationProcessingFilter` be configurable to some option, having default value "true".

Then, it would be just com extra config in applications.yml in test package, what would be totally clean. I am currently trying to argue this in https://gitter.im/spring-projects/spring-security

I will report

xetys:
[...]

PR #4066, opened 2016-09-03
replaces OAuth2 mock annotation with a token mock util

xetys:
which uses a RequestPostProcessor and @MockBean to enable secure testing support

fixes issue #4058


xetys:
everyone ok with this? I would like to have #4058 fixed in 3.7

deepu105:
Looks Ok to me

xetys:
polished it a bit, before @jdubois slaps me for more double lines :laughing: 

I'll merge then...

PR #4068, opened 2016-09-05
Angular 2

vivekmore:
remove unused dependency $q


sendilkumarn:
@vivekmore  same here

vivekmore:
There probably was a problem in the rebasing.
I'll submit another PR

PR #4069, opened 2016-09-05
remove unused dependency $q

vivekmore:
remove unused dependency $q


sendilkumarn:
@vivekmore I think there was some problem with rebasing?  Can you check that please ? It is showing 8 commits, including your previous ones.

vivekmore:
There probably was a problem in the rebasing.
I'll submit another PR

PR #4070, opened 2016-09-05
remove $q from injected dependencies

vivekmore:
remove $q from injected dependencies as it is not used


deepu105:
duplicate

sendilkumarn:
@deepu105 this is to master I suppose. The other was to angular 2

deepu105:
Oh ok I didn't notice

PR #4078, opened 2016-09-06
add GB and MB for disk space display in admin Dashboard

flaviencathala:
Disk space in admin Dashboard printed in GB or MB.

screenshot from 2016-09-06 12 25 55

### DIFF
--- a/generators/client/templates/src/main/webapp/app/admin/health/_health.modal.controller.js
+++ b/generators/client/templates/src/main/webapp/app/admin/health/_health.modal.controller.js
@@ -17,5 +17,14 @@
         function cancel() {
             $uibModalInstance.dismiss('cancel');
         }
+
+        vm.getProperValue = function(value) {
+            if (vm.currentHealth.name == "diskSpace") {
+                if (value / 1073741824 > 1)
+                    return (value / 1073741824).toFixed(2) + " GB";
+                return (value / 1048576, 2).toFixed(2) + " MB";
+            }
+            return value;
+        }
     }
 })();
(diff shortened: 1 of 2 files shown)


generators/client/templates/src/main/webapp/app/admin/health/_health.modal.controller.js
```diff
@@ -17,5 +17,14 @@
         function cancel() {
             $uibModalInstance.dismiss('cancel');
         }
+
```
deepu105: I dont like to have this in the controller. You can write a filter to do this if required, but since we are doing ng2 migration I would wait. Else you need to do this for the ng2 branch as well. In either case ill prefer a generic filter

jdubois: Oh sorry @deepu105 I should have introduced you to @040796 (who definitely has a strange username) - he's my new trainee, I'm going to have a look at this with him

deepu105: Oh ha ha ya he really does have a strange user name.

Thanks & regards,
Deepu

040796: We will wait for the migration and add a filter later.

pascalgrimaud: @040796 welcome ! But it should be better to put your realname ;)

deepu105: and easy for others to do @mentions 😄

jdubois:
OK, so let's focus on AngularJS 2 first -> it's a small change, so if @040796 you want to do it as a filter in the HTML page, feel free to do it (indeed I think it's much prettier that way), otherwise let's keep our focus on AngularJS 2

PR #4092, opened 2016-09-07
Remove custom jackson configuration

### LINKED ISSUE #4060: Remove Jackson configuration bean
The default serialization of ZonedDateTime by Jackson is "yyyy-MM-dd'T'HH:mm:ss.SXXX" where the fractional seconds can go up to 9 digits. We currently replace that by "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'" to ensure serialization in UTC time zone with a fixed 3 fractional seconds digits.
Although the ES5 spec specifies 3 fractional seconds digits, I've tested on Chrome52, Firefox 47 and IE 11 that the string "2016-09-02T22:30:44.168123456+02:00" is correctly parsed.
Also since we convert ZonedDateTime to Date before storage, the nanoseconds are lost anyway.

So I wonder if we can remove our custom jackson config and use jackson-datatype-jsr310's default one as per Policy 1.
It would also benefit to people that want to transmit the time-offset for something or that need to use nanosecond precision for some services (not for entities since we don't store the time-offset).

Or does anyone knows a browser version for which that will cause issue ?

cbornet:
Fix #4060


jdubois:
I think we should merge that: @cbornet tested this, and no one commented... I'm going to update the "remove file" sections and merge.

PR #4099, opened 2016-09-08
Change Protractor to use Firefox by default because tests pass consistently

mraible:
I’m glad to report that after spending numerous hours trying to fix this, it seems the Protractor failure issues [I was experiencing](https://groups.google.com/forum/?hl=en#!topic/jhipster-dev/bIRFQMeEZw0) are caused by using the Chrome driver rather than Firefox. 

* If I use Firefox (version <= 46), all tests pass consistently. 
* If I use Chrome, the first entity test fails because it seems the `afterAll()` in `administration.js` doesn’t fully logout the user. 
* If I use Safari, the first test passes. If I run them again, the user is still logged in and the tests start to fail right away. 

I tried altering the `afterAll()` to use `click().then()` promises and `browser.sleep`, but neither worked. 

Clearly the UI works and Firefox is reporting it best.


deepu105:
But its still failing for latest firefox. and since for most people firefox will auto update. wouldnt that be a problem? may be we can do a force logout.

pascalgrimaud:
What do you think about changing our Travis config to upgrade to latest firefox version. See https://docs.travis-ci.com/user/firefox/

We could add to `.travis.yml` file:
```yaml
addons:
  firefox: "latest"
```

jdubois:
Yes but not now, otherwise our build will be broken and we have no solution

Le 9 sept. 2016 8:24 AM, "Pascal Grimaud" <notifications@github.com> a
écrit :

jdubois:
Not sure if I will have time to test this, and I'm worried it blocks the
next release.

Let's keep Firefox:

- Chrome is going to break a lot of people's tests, and it's not working
better with it anyway
- We could still do the release: it's already broken, so we're not making
this worse
- I would document it, even putting a warning in the generator question

Le 9 sept. 2016 8:37 AM, "Julien Dubois" <julien.dubois@gmail.com> a écrit :

Yes but not now, otherwise our build will be broken and we have no solution

Le 9 sept. 2016 8:24 AM, "Pascal Grimaud" <notifications@github.com> a
écrit :

jdubois:
I just tested with:

- JHipster v3.6.1 and the new "Bento" box (see https://github.com/jhipster/jhipster-devbox/pull/59 ) and it's working fine, using Firefox 48 !!! So this really makes me wonder what's happening here.
- On my Mac the current "master" version with Firefox 48.0.2 -> Firefox doesn't crash anymore, but I have a blank browser and it doesn't work...

So we really need more testing here, but that's looking better

jdubois:
I have merged this change, as indeed it works better on Firefox. Chrome is behaving very strangely on VirtualBox, this is a bad user experience for me.
Anyway, we already chose Firefox because it's easier to use on Linux, which is important as Protractor will be used for continuous integration, which typically runs on Linux.

Concerning the current bug:

- I have no issue whatsoever on Linux (using our DevBox on "master", with the brand new "bento" box). I have tried with JHispter v3.6.1 and our "master" version.
- On Mac it's clear there is an issue with Firefox: it used to crash, and with v48 it's having a blank screen and gets launched twice... For me this is a Firefox issue: if you want it to work, you need to use an older version, or wait a little bit, as it's clear they are working on it.

-> so for me this isn't a blocker issue, let's move on

PR #3748, opened 2016-06-20
Improve generated travis file

pascalgrimaud:
As suggested by @PierreBesson, here a new generated `.travis.yml` with better tests

PR #4103, opened 2016-09-08
Update .travis.yml to run Protractor tests

mraible:


pascalgrimaud:
Let me check this, before merging

generators/server/templates/_travis.yml
```diff
@@ -46,6 +52,18 @@ script:
 <%_ } else if (buildTool === 'gradle') { _%>
   - ./gradlew bootRepackage -Pprod -x test
 <%_ } _%>
+  - docker-compose up -d
```
pascalgrimaud: This command won't work, because there is no `docker-compose.yml` file at the root folder.
It's because we migrate all files related to docker into src/main/docker/

Anyway, this line should be more complex:
- test type of database (mysql, postgresql, etc.)
- test if it's gateway or microservice, if so, need to start the registry too

generators/server/templates/_travis.yml
```diff
@@ -46,6 +52,18 @@ script:
...
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+<%_ if (buildTool === 'maven') { _%>
+  - ./mvnw spring-boot:run &
+<%_ } else if (buildTool === 'gradle') { _%>
+  - ./gradlew bootRun &
+<%_ } _%>
+  - bootPid=$!
+  - sleep 120s
+  - grunt itest
```
pascalgrimaud: grunt is gone!! :-)

we should test if the configuration has protractor or not

generators/server/templates/_travis.yml
```diff
@@ -46,6 +52,18 @@ script:
 <%_ } else if (buildTool === 'gradle') { _%>
   - ./gradlew bootRepackage -Pprod -x test
 <%_ } _%>
+  - docker-compose up -d
+  - sleep 30s
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+<%_ if (buildTool === 'maven') { _%>
+  - ./mvnw spring-boot:run &
```
pascalgrimaud: not sure it will work if the configuration is uaa, so maybe exlcude this config ?

pascalgrimaud:
@mraible : I just added some comments.
I recently updated this file. See https://github.com/jhipster/generator-jhipster/pull/3748

I didn't go deeper for 2 reasons:
- the "run" part is complex, and I didn't have time to test every config
- to be consistent with the Jenkinsfile

As we have very few issue about these generated files, I dont know if people use them, so I didn't invest more time on it.

jdubois:
I'm correcting this:

- I'll change the "grunt" part, etc according to @pascalgrimaud comments
- I'll remove the protractor part, as well as the more complex parts, as indeed I'm not sure people use those, and they are going to cost a lot of time

PR #4104, opened 2016-09-08
adds liquibaseAddMigration task to gradle

xetys:
this is a simple task for liquibase I built for my devs at work, to quickly make new liquibase migration files.

To use this, just `./gradlew liquibaseAddMigration`, and you get a new well formated xml file (which is also added to master.xml).

If you think it is meaningful, and ok to be exclusive to gradle (it uses groovy power maven doesn't have :smile: ), let's add this. If not, its ok, since this PR is a suggestion from me for a better workflow.

### DIFF
--- a/generators/server/templates/gradle/_liquibase.gradle
+++ b/generators/server/templates/gradle/_liquibase.gradle
@@ -23,6 +23,45 @@ task liquibaseDiffChangelog(dependsOn: compileJava, type: JavaExec) {
     <% if (authenticationType == 'oauth2') { %>args "--excludeObjects=oauth_access_token, oauth_approvals, oauth_client_details,   oauth_client_token, oauth_code, oauth_refresh_token"<% } %>
 }
 
+task liquibaseAddMigration << {
+    group = "liquibase"
+
+    def dateString = buildTimestamp()
+    def migration = new File("src/main/resources/config/liquibase/changelog/" + dateString + "_migration.xml")
+
+    def content = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n" +
+        "<databaseChangeLog\n" +
+        "    xmlns=\"http://www.liquibase.org/xml/ns/dbchangelog\"\n" +
+        "    xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\n" +
+        "    xsi:schemaLocation=\"http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd\">\n" +
+        "\n" +
+        "    <property name=\"now\" value=\"now()\" dbms=\"mysql,h2\"/>\n" +
+        "    <property name=\"now\" value=\"current_timestamp\" dbms=\"postgresql\"/>\n" +
+        "    <property name=\"now\" value=\"sysdate\" dbms=\"oracle\"/>\n" +
+        "\n" +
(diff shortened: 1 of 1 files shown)


jdubois:
I find it *very* useful, but we need to have the same experience with Maven & Gradle, so we need the same goal with Maven... And that's going to be more annoying!!

xetys:
i think i will need to dig in more in maven for that...it is not unsolvable

pascalgrimaud:
if it's hard to do it with maven, maybe do it on a sub generator ? I think it will be more easier

jdubois:
Yes, easier, same solution for both tools, I like it better

deepu105:
But then it might be better as a module as not everyone would use it. The
module can also do more such DB utils

Thanks & regards,
Deepu

xetys:
if it should be done using yeoman, I think this should be inside the core jhipster generator. If you develop a real application using JHipster, you just can't use `yo jhipster:entity` since you added more advanced code, because the entity generator will edit the existing one, which will lead into total migration disasters for production databases.

So the only alternatives are: handmade liquibase migration (this is what I was used to do until today), or the liquibase:diff. I am not a fan of the diff, since it would force me to first change the real database in order to make some little changes. With tools like IntelliJ with built in XSD parsing, doing migration is very very easy in the xml files.

I think it makes sense to build some migration generator in the core generator set...

deepu105:
@xetys my only concern is that if we start adding sub generators for every small stuff, we would end up having a lot of sub generators to maintain and the user would have to remember all of them. Another option would be to add a generic util sub gen `yo jhipster:util` which would ask the user what they want to do from a list of items like `generate a migrations file` or `generate a DTO` or `Generate a XXXX` etc

xetys:
now we got two contradicting options to go

## having a migration task

***pros***: 
 
* independent from generator-jhipster (some colleages are working on a JH generated app, without having the generator itself)
* quite elegant in gradle

***contras***

* ugly in maven
* not unified to "have everyting JH related in JH generator"

## providing it via `yo jhipster:util`

***pros***

* no gradle vs maven problem
* easy to tune in for more functionally using more prompts

***contras***

* blowing up the generator

Even if I like my gradle task, I am more likely to go the generator way, as it really seems to be less work with more options for the user. I definitly at it should be part of the core generator, because it is a core developers task, but I would like to hear some more opinions on that...

deepu105:
@xetys dont worry I'm just trying out the new review feature from github :smile:

jdubois:
For me this is either a sub-generator or a module. No need for "jhipster:util", it could be its own sub-generator like "yo jhipster:add-liquibase-changelog".

Anyway:

- I'm closing this as this solution is not the best one
- The idea is good, @xetys you can either do a sub-generator or module, it just depends on what you prefer

jdubois:
Yes that could be a help, then it's a bit like the oh-my-zsh shortcut we already have. In the end we would still have the sub-generators anyway.

xetys:
sure I can provide this as a subgenerator, too....for me it makes no real difference if it is unter jhipster: or jhipster:utils. I don't even think a developer should remember all sub-generators of jhipster, since we just can provide a help, which describes shortly what the feature does. For myself, there are not that much tools, where I know every signle option and know what it does, it's sufficient to know where I can find help.

I think for this and further sub generators we should rely on policy 2...if it really has some benefit, it should be done...

if you are all fine with that, I would prefer to do a yo jhipster:add-liquibase-changelog in a PR

deepu105:
Ok if you guys feel that way then I'm fine

Thanks & regards,
Deepu

PR #4107, opened 2016-09-08
change variable to match registry env variable

ruddell:
The registry looks for GIT_SEARCH_PATHS


pascalgrimaud:
Tips: you can add [ci skip] to your commit, so Travis wont trigger a build :-)

PR #4108, opened 2016-09-09
Improve comment in app.yml to use docker cassandra for rate limiting

pascalgrimaud:
Find by @eoinverling on gitter

When generating a gateway project, in `src/main/docker/app.yml` file, the comment part about using docker cassandra is incorrect

Another minor issue:
- when using the docker-compose sub generator, the cassandra part is not generated in the new `docker-compose.yml` file. It's because we have to detect if it's commented or not. So for the moment, I suggested the user to copy it manually


jdubois:
Thanks!

PR #4112, opened 2016-09-10
Remove workaround to set default profile

### LINKED ISSUE #4059: Revisit the logic for default profile setup
In order to have the default profile used for build in deploy as well `dev|prod` we currently read the `application.yml` and set it manually due to a spring boot bug discussion here https://github.com/jhipster/generator-jhipster/pull/3585
Since spring boot 1.4 has a partial fix https://github.com/spring-projects/spring-boot/issues/5998 so we can revisit and see if we can remove the workaround code 

- [x] Test default bootrun
- [x] Test default bootrun -Pprod
- [x] Test package
- [x] Test package -Pprod
- [x] Test IDE run
- [x] Test IDE run -Pprod

deepu105:
There is a minor glitch, which can be solved by documentation.

When the app is run from an IDE using the Application.java main file only dev profile will be activated by default. the swagger profile would need to be explicitly enabled if required. This is due to the issue [here ](https://github.com/spring-projects/spring-boot/issues/6833#issuecomment-246026401)
This issue does not affect bootRun or running packaged war.
Fix #4059

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1120,7 +1120,7 @@
                 <!-- log configuration -->
                 <logback.loglevel>DEBUG</logback.loglevel>
                 <!-- default Spring profiles -->
-                <spring.profiles.active>dev${profile.no-liquibase}</spring.profiles.active>
+                <spring.profiles.active>dev,swagger${profile.no-liquibase}</spring.profiles.active>
             </properties>
         </profile>
         <profile>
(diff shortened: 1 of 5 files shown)


PierreBesson:
I'm OK with this change. However, as you say we need to add documentation as now the only way to have swagger enabled when running the app from main class is by tweaking the PROFILES_ACTIVE env var in the IDE. Am I right ?

jdubois:
As long as it removes the ugly hack let's do it.

deepu105:
@PierreBesson it can be done using any of the valid `--spring.profiles.active` setup, in IntelliJ you have a field in run config dialog to set a profile, there is a similar option in Eclipse as well I guess. I dont think you need to tweak a lot for it

PR #4118, opened 2016-09-11
Create a profile (called IDE) that includes mapstruct processor

amr:
Create a profile (called IDE) that includes mapstruct processor and any tweaks that should be only applied to the IDE.

I didn't find that Gradle needs such a profile. Please correct me if I'm wrong, I'm no Gradle expert.

Docs being updated in jhipster/jhipster.github.io#307

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1249,5 +1249,15 @@
                 <spring.profiles.active>prod${profile.swagger}${profile.no-liquibase}</spring.profiles.active>
             </properties>
         </profile>
+        <profile>
+            <id>IDE</id>
+            <dependencies>
+                <dependency>
+                    <groupId>org.mapstruct</groupId>
+                    <artifactId>mapstruct-processor</artifactId>
+                    <version>${mapstruct.version}</version>
+                </dependency>
+            </dependencies>
+        </profile>
     </profiles>
 </project>


deepu105:
Any why is this required in the first place?  Mapstruct works in IDE without any issues as the mapstruct processor plugin is included by default.

I do not understand this PR. Could you please explain your issue?

deepu105:
Ok I just saw the issue in the website repo. Ok so if Julien agrees to this then its fine.

jdubois:
Well, to be honnest I'm not a big MapStruct + Eclipse user, and I understand the issue is mostly for people using both, which should be pretty common.
So we need to test this first, before merging.

amr:
@jdubois, actually for me the issue is with MapStruct + IntelliJ

jdubois:
Oh, I need to check then, because it's working for me. Still, I'm sure the current situation could be improved, I remember it wasn't working perfectly.

jdubois:
- I'm still having issues to reproduce :-(
- Note that we should also support Gradle, here is an example: https://github.com/jhipster/generator-jhipster/pull/4190/files

jdubois:
OK I confirm the issue and the fix!!!!
How to reproduce:

- create an empty project, and add an entity with a DTO
- in doubt, do `./mvnw clean` to be sure the target directory is empty
- use IDEA to run the main Spring Boot class -> the mapping bean will be missing, as the postprocessor didn't run

Using the IDE with the dev+IDE  Maven profiles solves this.

My only concern is that you need to select `dev` and `IDE`, which is a bit annoying, but I don't have a better solution as you might also want to have `prod` and `IDE` so it's good to be able to mix profiles.

PR #4133, opened 2016-09-13
Fix Register Page

wmarques:
@flaviencathala pointed to me that the register page isn't displaying correctly when accessing via the URL (the link in the navbar seems broken but it should come from ui-sref...)
So I renamed the file into component, added the prefix to avoid conflict with `register` service and added the component in the ng2module.


generators/client-2/templates/src/main/webapp/app/account/register/_register.state.ts
```diff
@@ -10,7 +10,7 @@ export function RegisterStateConfig($stateProvider) {
         },
         views: {
             'content@': {
-                template: '<register></register>'
+                template: '<<%=jhiPrefix%>-register></<%=jhiPrefix%>-register>'
```
deepu105: the prefix needs to be applied on the component as well

wmarques:
@deepu105 Should be fine now :)

Issue #4129, opened 2016-09-13
Add asynchronous message support, with Kafka


jdubois:
One important feature we currently don't provide is asynchronous messaging. I'm not listing here why it's a good thing :-)

There are a number of implementations to achieve this: with ActiveMQ, RabbitMQ, Kafka, Redis, etc... And of course we need a good API for this.

I've already got something working with Kafka, but I'm using Spring Kafka directly: I'm not satisfied with it, but I know how to do it, and I know it works :-)

My current idea is to:

- Add [Spring Cloud Stream support](https://cloud.spring.io/spring-cloud-stream/) as it's a layer on top of RabbitMQ, Kafka, Redis, etc... So we would not be locked on Kafka. I haven't used it yet, and I'm not sure of the limitations -> if I'm not happy with it, I'll go back to "pure Kafka"
- Use Kafka as an implementation: mostly because I already have it working (including Docker images, etc), but that's also the server for which I see the biggest demand/hype at the moment.

Then that would add a new question on the generator, so my idea is to refactor some existing questions:

- Currently we have "do you want to use Elasticsearch?" "do you want to use WebSockets?" etc...
- For all those tools, we could have instead a multi-check box question with "what additional tools do you want to configure?" which would include Elasticsearch, WebSockets, Kafka...
- Internally I would keep the current way of having booleans for using each technology (and not an Array of technologies, which would be more consistant), as it's much easier to use in switches in our code.

deepu105:
👍 to refactor the questions. But we should keep the yo-rc format same as today.

Do you want to add Kafka to  monoloth as well or just for microservices?

I can help you with refactoring the prompts part if you need.

I would merge the below into a single option with multiselect. 

- enableSocialSignIn
- searchEngine
- clusteredHttpSession (does any one ever use this?)
- websocket

xetys:
I am currently taking a look to Spring Cloud Bus, which is the default messaging solution for spring cloud microservices...so I would like to participate on this part, when I did some experiments with it.

One other hot topic is something called event sourcing, which in short is another way of service-to-service communication, as an allternate way to REST communication

jdubois:
Great!

- Kafka is very good for microservices!! So I want it for monoliths, microservices and gateway. Shouldn't be much different.
[...]

PR #4134, opened 2016-09-13
Restructure server side option prompts

deepu105:
Related to #4129

@jdubois take a look and merge if this looks good to you

The options if not selected will be set as `false` instead of `'no'` have added backward compatibility as well.


jdubois:
I didn't try it yet, but yes that's the idea, totally agree with what you did.

jdubois:
Thanks @deepu105 , let's do this!

jdubois:
@deepu105 it looks OK, but those are not only server-side options, they also generate client-side code -> I will rename the question to just "options"

deepu105:
OK


Thanks & Regards,
Deepu

PR #4135, opened 2016-09-13
Fix console error after first login reload

wmarques:
Angular was trying to read the `account` object before it was send by the server.


deepu105:
please tick the item off in todo

sendilkumarn:
There is something wrong here. @wmarques @deepu105  have you checked? 

you cannot apply `multiple template binding to a single element`. I will check this and add a commit

Issue #4130, opened 2016-09-13
Swagger profile not active by default in dev mode


jdubois:
On JHipster 3.7.0, when the app is run in "dev" mode from an IDE, the "swagger" profile is not active by default, so the API is not available.

PierreBesson:
@deepu105, already notified me of this bug (#4112). Would be nice if we could work around it or if we can't do that add some documentation about it.

deepu105:
I have already added documentation for it here http://jhipster.github.io/profiles/

@jdubois its a glitch only when run from IDE its due to https://github.com/spring-projects/spring-boot/issues/6833#issuecomment-246093943

deepu105:
closed via #4146

PR #4142, opened 2016-09-14
Add Swagger profile by default

jdubois:
This is linked to https://github.com/jhipster/generator-jhipster/issues/4130

-> I don't understand why we don't this anymore (that's pretty close to my original code), as this solves the IDE issue, and I don't see any downsides.

@deepu105 what do you think?

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
+++ b/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
@@ -3,8 +3,6 @@ package <%=packageName%>.config;
 import org.springframework.boot.SpringApplication;
 import org.springframework.core.env.Environment;
 
-import java.util.*;
-
 /**
  * Utility class to load a Spring profile to be used as default
  * when there is no <code>spring.profiles.active</code> set in the environment or as command line argument.
@@ -20,14 +18,7 @@ public final class DefaultProfileUtil {
      * @param app the spring application
      */
     public static void addDefaultProfile(SpringApplication app) {
-        Map<String, Object> defProperties =  new HashMap<>();
-        /*
-        * The default profile to use when no other profiles are defined
-        * This cannot be set in the <code>application.yml</code> file.
-        * See https://github.com/spring-projects/spring-boot/issues/1219
-        */
-        defProperties.put(SPRING_PROFILE_DEFAULT, Constants.SPRING_PROFILE_DEVELOPMENT);
-        app.setDefaultProperties(defProperties);
+        app.setAdditionalProfiles(Constants.SPRING_PROFILE_DEVELOPMENT, Constants.SPRING_PROFILE_SWAGGER);
     }
 
     /**


generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -20,14 +18,7 @@
...
     public static void addDefaultProfile(SpringApplication app) {
-        Map<String, Object> defProperties =  new HashMap<>();
-        /*
-        * The default profile to use when no other profiles are defined
-        * This cannot be set in the <code>application.yml</code> file.
-        * See https://github.com/spring-projects/spring-boot/issues/1219
-        */
-        defProperties.put(SPRING_PROFILE_DEFAULT, Constants.SPRING_PROFILE_DEVELOPMENT);
-        app.setDefaultProperties(defProperties);
+        app.setAdditionalProfiles(Constants.SPRING_PROFILE_DEVELOPMENT, Constants.SPRING_PROFILE_SWAGGER);
```
deepu105: @jdubois If we run by default this would cause a double loading of `dev` profile as it would have been already loaded from the `application.yml` file once. If we run in prod, it will also load `dev` which is very bad. If we decide to do this conditionally at this point we would not have any profile info in the first place to decide if we need to set this or not that is why I went with `app.setDefaultProperties(defProperties);` as spring would disregard it when there is a profile set using any of the spring suggested way

deepu105: take look at the logs from the travis build for any [prod profile](https://travis-ci.org/jhipster/generator-jhipster/jobs/159938444). you can see that its always running both `prod` and `dev` 
```
2016-09-14 17:15:50.253  INFO 7015 --- [           main] c.mycompany.myapp.TravisPsqlEsNoi18NApp  : Starting TravisPsqlEsNoi18NApp v0.0.1-SNAPSHOT on testing-gce-99b1c25b-38fb-4d0c-9d00-e84dbf0bf13c with PID 7015 (/home/travis/app/app.war started by travis in /home/travis/app)
2016-09-14 17:15:50.267  INFO 7015 --- [           main] c.mycompany.myapp.TravisPsqlEsNoi18NApp  : The following profiles are active: dev,swagger,prod
2016-09-14 17:15:59.250  WARN 7015 --- [           main] io.undertow.websockets.jsr               : UT026009: XNIO worker was not set on WebSocketDeploymentInfo, the default worker will be used
2016-09-14 17:15:59.253  WARN 7015 --- [           main] io.undertow.websockets.jsr               : UT026010: Buffer pool was not set on WebSocketDeploymentInfo, the default pool will be used
2016-09-14 17:15:59.882  INFO 7015 --- [           main] c.mycompany.myapp.config.WebConfigurer   : Web application configuration, using profiles: [dev, swagger, prod]
```

also this build has errors but its still marked as passed https://travis-ci.org/jhipster/generator-jhipster/jobs/159938442 I dont know why. @pascalgrimaud FYI

pascalgrimaud: very strange!!! so much error and the build is green
I will investigate

pascalgrimaud: So the error logs come from mysql:5.7.15 -> I made a PR to revert back to 5.7.13
But the application manages to start because mixing of profile dev+prod

jdubois: Oh, I see, I had checks for those but they were removed.... Let me put my checks back again and do a new proposal

jdubois: OK this whole spring.profiles.active and spring.profiles.include is new and not very well documented, but it seems like the way to go. I'm trying that and I'll do a new PR, as it's totally different.

deepu105: There was an issue with include, I had a bug opened in spring and its fixed. So it would work once we upgrade to 1.4.1 but still the other issue with `bootstrap-{profile}.yml` files are there

PR #4038, opened 2016-08-28
Fix mysql version and travis mysql build

pascalgrimaud:
I fix 2 issues here:

### ZonedDateTime problem with MySQL 5.7.14

In https://github.com/jhipster/generator-jhipster/issues/4009, @tsaqova mentionned

> here are my schema, i change mysql version to 5.7.13 as workaround because 5.7.14 can't have 2 timestamp column (from zonedatetime) in one table by default

Here the error log on Travis build:

```
[...]

PR #4144, opened 2016-09-14
Use MySQL 5.7.13 instead of MySQL 5.7.15

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/pull/4038


jdubois:
My fault :-(

PR #4145, opened 2016-09-15
Alert Component Migration

wmarques:
With those changes I'm having a very strange `Cannot read property 'injector' of null`..


generators/client-2/templates/src/main/webapp/app/components/alert/_alert-error.component.ts
```diff
@@ -0,0 +1,88 @@
+import { Component, Inject, OnDestroy } from '@angular/core';
+import { AlertService } from "./alert.service";
+
+@Component({
+    selector: 'jhi-alert-error',
+    template: '<div class="alerts" ng-cloak="">' +
```
deepu105: we can use es6 string literals to avoid the concatenation. I'll merge and update this

Issue #4132, opened 2016-09-13
no cloud config in boostrap phase


xetys:
I was about building our first 3.7 service at office and lost a couple of time by looking, why several settings not work....

if you start a JHipster application from war directly, or using docker....at the very first moment no bootstrap configuration is loaded. So setting SPRING_CLOUD_CONFIG_URI doesn't override anything, and finally a lot of things fail. In my case I don't get any connection to SQL db, because the config never arrives the app.

to reproduce that issue, just start a microservice using `java -jar` or docker, and watch the first line commiting a successful handshake with config server (like `s.n.www.protocol.http.HttpURLConnection  : sun.net.www.MessageHeader@7d3e86556 pairs: {GET /config/UAA/dev/master HTTP/1.1: null}`) is missing...

I guess this may have something to do with work on DefaultProfileUtil, which recently was done...

**edit:** copying bootstrap-prod.yml to bootstrap.yml is a dirty fix to that issue

PierreBesson:
Well, initially, the app can be either in dev profile or in prod, that's why we have a bootstrap-dev and bootstrap-prod. Is there a bug that cause the app not to pick up it's profile correctly.

xetys:
i am currently on it...I report as soon I got some results

xetys:
it seems to work sometimes....it's harder to reproduce than it looked like....however in production something goes heavily wrong because of not contacting cloud config server

xetys:
well, I still can't say what is exactly wrong....
I am missing the config fetch during startup in my production and also when starting a UAA using docker....when I point to spring.cloud.config.uri via ENV and maintain database url inside...this doesn't work....

I will have to spend some more time on it....but it seems for config to be broken somehow...and I hate I have no more accurate description of that issue....maybe this have something to do with the 32 degrees in my office now..

PierreBesson:
@xetys Thanks for investigating the issue. Does your uaa-setup repo allows to reproduce the bug ?

xetys:
The docker setup specifies all important config variables over environment, so they are able to run there. For some reason it happens to me, that since 3.7 the services just do not connect to cloud config, and still don't fail. I share database config over the config server, so I recognized this behaviour. Note, I am not using docker compose in production, and so not provide SPRING_DATASOURCE_URL

PierreBesson:
[...]

PR #4146, opened 2016-09-15
Configure Swagger by default with the dev profile, using the YML file

jdubois:
@deepu105 I know you didn't go that way, but I tested both dev & prod profiles:

- In my IDE
- In Maven
- As an executable WAR file

And I had no issue.
Could you test and give me your feedback?

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-dev.yml
+++ b/generators/server/templates/src/main/resources/config/_application-dev.yml
@@ -25,6 +25,8 @@ eureka:
 <%_ } _%>
 
 spring:
+    profiles:
+        active: dev,swagger
     <%_ if (applicationType == 'gateway' && databaseType != 'cassandra') { _%>
     autoconfigure:
         exclude: org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration
--- a/generators/server/templates/src/main/resources/config/_application-prod.yml
+++ b/generators/server/templates/src/main/resources/config/_application-prod.yml
@@ -25,6 +25,8 @@ eureka:
 <%_ } _%>
 
 spring:
+    profiles:
+        active: prod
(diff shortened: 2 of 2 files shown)


generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -25,6 +25,8 @@ eureka:
 <%_ } _%>
 
 spring:
+    profiles:
+        active: dev,swagger
```
deepu105: I have actually tried this last time but I was getting some issues with this is that `dev` profile will be loaded first from `application.yml` in boot sequence, which would load this file. Now from this file spring would try to load the dev profile again, which would go into an infinite loop. But seems like its working fine now. So yes this could be a solution for https://github.com/jhipster/generator-jhipster/issues/4130 but then again the proper solution would be to use `spring.profiles.include` and with the fix for that [landing in SB 1.4.1](https://github.com/spring-projects/spring-boot/issues/6833) we could very well do that later. 
But I dont think this will help with https://github.com/jhipster/generator-jhipster/issues/4132 which is far serious issue than swagger not enabled when run from IDE

jdubois: @deepu105 so it looks like you are OK to merge this - I'm doing it. And then I'm going to look at the other issue.

generators/server/templates/src/main/resources/config/_application-prod.yml
```diff
@@ -25,6 +25,8 @@ eureka:
 <%_ } _%>
 
 spring:
+    profiles:
+        active: prod
```
deepu105: can we remove this as its not required, this will work from the `application.yml` itself

jdubois: OK

PR #4149, opened 2016-09-15
Add oauth2 client details storage support for mongo

cbornet:
Basically, same as #3772 but this time for mongo.


atomfrede:
:+1: Looks good!

jdubois:
That's such a lot of work!!!!

cbornet:
Nexr : cassandra ? :smile:

jdubois:
there are very few people using OAuth 2, and very few people using Cassandra -> I don't recommend wasting time on this...

cbornet:
Yes, that would probably be not very useful and just for the fun...

PR #4152, opened 2016-09-16
Fix SystemJS config and Alert Service error

wmarques:
I followed ng-bootstrap docs and changelog to fix the `404 traceur not found` issue, it's fixed. As @deepu105 suggested I commented the import in `NotificationInterceptor` so it's partially fixed.

Now the error I get is `Only selectors matching element names are supported, got: [ngbDatepickerMonthView] ` from the UpgradeAdapter.

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
+++ b/generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
@@ -4,9 +4,7 @@ import { BrowserModule } from '@angular/platform-browser';
 import { NgModule } from '@angular/core';
 
 // TODO change this to NgbModule -->  after complete migration
-import {NgbAlertModule} from '@ng-bootstrap/ng-bootstrap/alert/alert.module';
-import {NgbCollapseModule} from '@ng-bootstrap/ng-bootstrap/collapse/collapse.module';
-import {NgbDropdownModule} from '@ng-bootstrap/ng-bootstrap/dropdown/dropdown.module';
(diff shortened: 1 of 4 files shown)


sendilkumarn:
LGTM 👍

wmarques:
@sendilkumarn @deepu105 any clues on the error I get? It seems to come from the ngBootstrap lib that doesn't work well with the upgrade adapter :(

deepu105:
@wmarques not yet. ill check later today to see if I can find anything

deepu105:
may be we shouldn't try very hard to make things work with ng1 code as our end goal is to completely move to ng2, so lets focus on making the ng2 part work first

wmarques:
@deepu105 sure but it's a blocking thing as the app doesn't display anymore.. But yes, I agree that if there is some issue with the upgradeadapter it's not a major issue

generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
```diff
@@ -4,9 +4,7 @@ import { BrowserModule } from '@angular/platform-browser';
 import { NgModule } from '@angular/core';
 
 // TODO change this to NgbModule -->  after complete migration
-import {NgbAlertModule} from '@ng-bootstrap/ng-bootstrap/alert/alert.module';
```
deepu105: @wmarques the issue you got with datepicker thing is from this check https://github.com/angular/angular/issues/11280

deepu105: Im reverting this lines for now

PR #4155, opened 2016-09-16
reverting profile util temporary

xetys:
until some better solution is found, to prevert cloud config failures in production

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
+++ b/generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
@@ -1,9 +1,14 @@
 package <%=packageName%>.config;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
 import org.springframework.boot.SpringApplication;
-import org.springframework.core.env.Environment;
+import org.springframework.core.io.ClassPathResource;
 
-import java.util.*;
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


brunnels:
You should not revert the ```public static String[] getActiveProfiles(Environment env)``` method in ```_DefaultProfileUtil.java``` so you will not need to revert the changes in _ProfileInfoResource.java.

generators/server/templates/src/main/java/package/config/_DefaultProfileUtil.java
```diff
@@ -26,18 +55,23 @@ public static void addDefaultProfile(SpringApplication app) {
...
         app.setDefaultProperties(defProperties);
     }
 
     /**
-     * Get the profiles that are applied else get default profiles.
+     * Load application.yml from classpath.
+     *
+     * @return the YAML Properties
      */
-    public static String[] getActiveProfiles(Environment env) {
```
deepu105: no need to revert this method.

generators/server/templates/src/main/java/package/web/rest/_ProfileInfoResource.java
```diff
@@ -1,18 +1,18 @@
 package <%=packageName%>.web.rest;
 
-import <%=packageName%>.config.DefaultProfileUtil;
```
deepu105: no need to revert this class as well

deepu105:
Closing in favor of https://github.com/jhipster/generator-jhipster/pull/4184#issuecomment-248973925

PR #4169, opened 2016-09-19
skip serverSideOptions prompt if no choices are available

ruddell:
Trying to select an option fails (no options are available) when generating:
 - A UAA microservice with no database
 - A JWT microservice with MongoDB

### DIFF
--- a/generators/server/prompts.js
+++ b/generators/server/prompts.js
@@ -475,23 +475,26 @@ function askForOptionalItems() {
             }
         );
     }
-
-    this.prompt({
-        type: 'checkbox',
-        name: 'serverSideOptions',
-        message: function (response) {
-            return getNumberedQuestion('Which other technologies would you like to use?', true);
-        },
-        choices: choices,
-        default: defaultChoice
-    }).then(function (prompt) {
-        this.serverSideOptions = prompt.serverSideOptions;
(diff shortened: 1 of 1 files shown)


jdubois:
- don't merge this: there will soon be a new option for everyone, Kafka !
- there's a bug somewhere: jwt + MongoDB should have the Elasticsearch
option

Let me have a look.

Le 19 sept. 2016 10:39 PM, "Jon Ruddell" <notifications@github.com> a
écrit :

deepu105:
@jdubois merging this should be ok in either case as it will not impact addition of kafka question. Anyway ill let you decide it

jdubois:
@deepu105 yes I'm more worried that the ES option does not show up for JWT + MongoDB

jdubois:
Turns out we don't support MongoDB + Elasticsearch. Shouldn't be hard to do, I don't know why...
Anyway -> I'm going to have a look at merging this and adding Kafka at the same time, so not sure about what will end up here.

deepu105:
Ya thats something else we need to look at. Let me check the conditions
later

Thanks & regards,
Deepu

jdubois:
- OK so the current check is correct
- I still have an edge case with Kafka where I need this

-> I'm merging it

PR #4178, opened 2016-09-21
Fix sonar quality

tkman59:
my first commit ! (hackathon)

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -256,16 +256,15 @@ public class UserService {
     @Transactional(readOnly = true)
     <%_ } _%>
     public User getUserWithAuthorities() {
-
-      Optional<User> optionalUser = userRepository.findOneByLogin(SecurityUtils.getCurrentUserLogin());
-       User user = null;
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -256,11 +256,15 @@ public User getUserWithAuthorities(<%= pkType %> id) {
...
-        User user = userRepository.findOneByLogin(SecurityUtils.getCurrentUserLogin()).get();
-        <%_ if (databaseType == 'sql') { _%>
-        user.getAuthorities().size(); // eagerly load the association
-        <%_ } _%>
-        return user;
+        Optional<User> optionalUser = userRepository.findOneByLogin(SecurityUtils.getCurrentUserLogin());
+        User user = null;
+        if (optionalUser.isPresent()) {
+          <%_ if (databaseType == 'sql') { _%>
+            user.getAuthorities().size(); // eagerly load the association
```
jdubois: I think user will be `null` here -> maybe do `user = optionalUser.get()` before?

tkman59: yes ;)

PR #4184, opened 2016-09-21
Do the maven active profile substitution in bootstrap.yml for microservices

### LINKED ISSUE #4132: no cloud config in boostrap phase

PierreBesson:
fix #4132

### DIFF
--- a/generators/server/templates/src/main/resources/config/_bootstrap.yml
+++ b/generators/server/templates/src/main/resources/config/_bootstrap.yml
@@ -1,0 +1,8 @@
+spring:
+    application:
+        name: <%= baseName %>
+    profiles:
+        # The commented value for `active` can be replaced with valid spring profiles to load.
+        # Otherwise, it will be filled in by <%= buildTool %> when building the WAR file
+        # Either way, it can be overridden by `--spring.profiles.active` value passed in the commandline or `-Dspring.profiles.active` set in `JAVA_OPTS`
+        active: #spring.profiles.active#
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -631,6 +631,7 @@ module.exports = JhipsterServerGenerator.extend({
         writeServerMicroserviceAndGatewayFiles: function () {
             if (this.applicationType !== 'microservice' && this.applicationType !== 'gateway' && this.applicationType !== 'uaa') return;
 
+            this.template(SERVER_MAIN_RES_DIR + 'config/_bootstrap.yml', SERVER_MAIN_RES_DIR + 'config/bootstrap.yml', this, {});
             this.template(SERVER_MAIN_RES_DIR + 'config/_bootstrap-dev.yml', SERVER_MAIN_RES_DIR + 'config/bootstrap-dev.yml', this, {});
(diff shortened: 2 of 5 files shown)


deepu105:
can you check if https://github.com/jhipster/generator-jhipster/pull/4185 has any impact here

generators/server/templates/src/main/resources/config/_bootstrap.yml
```diff
@@ -0,0 +1,8 @@
+spring:
```
deepu105: you need to add this to the `profile_dev|prod.gradle` file in `processResources` section else value will not be replaced. not sure if anything is required for maven

jdubois:
Shouldn't we just merge #4185 instead?

deepu105:
@jdubois

deepu105:
@jdubois not sure if #4185 alone will help here. @PierreBesson can you check if this is still an issue with #4185 as well

PierreBesson:
This is to fix #4132 which is a different bug than what Deepu is talking about. Waiting for @xetys confirmation that it's also working for him.
@deepu, I will do the modification you requested tomorrow.

xetys:
I am sorry, but testing with gradle led to the issue being still present to me. I tried both, merging this PR to tag v3.7.1 and to the current master, in both regeneration I get a

```
2016-09-22 13:45:28.166  WARN 9873 --- [  restartedMain] c.c.c.ConfigServicePropertySourceLocator : Could not locate PropertySource: I/O error on GET request for "http://localhost:8888/bar/default": Verbindungsaufbau abgelehnt; nested exception is java.net.ConnectException
```
where localhost:8888 is the spring default value, because it actually *is* overwritten in the profile specific bootstrap files. After changing the "#spring.profiles.active#" to "dev", it started to connect properly for gradle, bus was impassible to all kinds of `-Dspring.profiles.active´ or `--spring.profiles.active`

deepu105:
Are you running it from the IDE?

Thanks & regards,
Deepu

xetys:
tried with IntelliJ and `./gradlew` from cmd, both the same

PierreBesson:
I just what @deepu105 told me, @xetys I hope that it will fix your issue.

ruddell:
@PierreBesson I tested the fix and microservices generated with master can now connect to the registry on startup 👍

PierreBesson:
Thank you everyone for helping me test this.

jdubois:
Does this mean we should have a new release?

deepu105:
Yes we can have one

jdubois:
OK I will do a release next week probably... I'm starting a new gig tomorrow, so I won't have time before.

PR #4185, opened 2016-09-21
Upgrade to SB 1.4.1

deepu105:
this also address https://github.com/spring-projects/spring-boot/issues/6833


deepu105:
Ok this seems to be working as expected hence im merging this

jdubois:
Perfect!!
So should we do a quick release for this?

deepu105:
@jdubois i would like to wait for https://github.com/jhipster/generator-jhipster/pull/4184 as well

Issue #4167, opened 2016-09-19
Add Spring Boot remote shell


cbornet:
See http://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-remote-shell.html

This gives some nice features for debugging
```shell
> help
Try one of these commands with the -h or --help switch:                                                                                                                                                                                     
                                                                                                                                                                                                                                            
NAME       DESCRIPTION                                                                                                                                                                                                                      
autoconfig Display auto configuration report from ApplicationContext                                                                                                                                                                        
beans      Display beans in ApplicationContext                                                                                                                                                                                              
cron       manages the cron plugin                                                                                                                                                                                                          
dashboard  a monitoring dashboard                                                                                                                                                                                                           
egrep      search file(s) for lines that match a pattern                                                                                                                                                                                    
endpoint   Invoke actuator endpoints                                                                                                                                                                                                        
env        display the term env                                                                                                                                                                                                             
filter     a filter for a stream of map                                                                                                                                                                                                     
java       various java language commands                                                                                                                                                                                                   
jmx        Java Management Extensions                                                                                                                                                                                                       
jul        java.util.logging commands                                                                                                                                                                                                       
jvm        JVM informations                                                                                                                                                                                                                 
less       opposite of more                                                                                                                                                                                                                 
mail       interact with emails                                                                                                                                                                                                             
man        format and display the on-line manual pages                                                                                                                                                                                      
metrics    Display metrics provided by Spring Boot                                                                                                                                                                                          
shell      shell related command                                                                                                                                                                                                            
sleep      sleep for some time                                                                                                                                                                                                              
sort       sort a map                                                                                                                                                                                                                       
system     vm system properties commands                                                                                                                                                                                                    
thread     JVM thread commands                                                                                                                                                                                                              
help       provides basic help                                                                                                                                                                                                              
repl       list the repl or change the current repl                                                                                                                                                                                         
```
and you can execute groovy code in a REPL way on your live app
```shell
> repl groovy
Using repl groovy
> com.mycompany.myapp.StandardApp.context.getBean("userRepository").findAll()
[...]

PR #4186, opened 2016-09-21
expose application context

cbornet:
It can then be accessed easily by REPL tools to get the beans

See #4167

### DIFF
--- a/generators/server/templates/src/main/java/package/_Application.java
+++ b/generators/server/templates/src/main/java/package/_Application.java
@@ -21,6 +21,7 @@ import org.springframework.cloud.netflix.eureka.EnableEurekaClient;
 <%_ if (applicationType == 'gateway') { _%>
 import org.springframework.cloud.netflix.zuul.EnableZuulProxy;
 <%_ } _%>
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.ComponentScan;
 <% if(authenticationType === 'uaa') { %>import org.springframework.context.annotation.FilterType;
 <%_ } _%>import org.springframework.core.env.Environment;
@@ -54,6 +55,8 @@ public class <%= mainClass %> {
     @Inject
     private Environment env;
 
+    public static ApplicationContext context;
+
     /**
      * Initializes <%= baseName %>.
      * <p>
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/_Application.java
```diff
@@ -54,6 +55,8 @@
     @Inject
     private Environment env;
 
+    public static ApplicationContext context;
```
jdubois: I'm not totally sure we should have the app context in a static variable. For what we currently do, of course there is no issue, but the original reason of not having "true" singletons in Spring (which would be static) is for running several app context in parallel in the same JVM. So maybe that could cause issue for some people. Like if they run several JHipster war in an app server, with a "strange" classloader (like JBoss - which I consider as totally wrong), there is a high risk that their app contexts get mixed up.

deepu105: oh yes there is that.

cbornet: I thought app servers were the past ! :smile: 
My actual need is to access the app context from outside the app (from the remore shell in REPL mode). Do you know another way to get it ?

jdubois: Using an app server is clearly a bad idea, but we can't force people :-)
Not sure about it, but could use JMX to get it?
Otherwise you won't find any "static" in Spring, because of those classloader issues (they also used to cause memory leaks in some app servers, which were not dereferencing correctly static fields)

cbornet:
Closing as per @jdubois comment and since there is another way to get the beans from the remote-shell:
```java
> repl groovy
Using repl groovy
> context.getAttributes().get("spring.beanfactory").getBean("userRepository").findAll()
[User{login='system', firstName='System', lastName='System', email='system@localhost', activated='true', langKey='en', activationKey='null'}, User{login='anonymoususer', firstName='Anonymous', lastName='User', email='anonymous@localhost', activated='true', langKey='en', activationKey='null'}, User{login='admin', firstName='Administrator', lastName='Administrator', email='admin@localhost', activated='true', langKey='en', activationKey='null'}, User{login='user', firstName='User', lastName='User', email='user@localhost', activated='true', langKey='en', activationKey='null'}]
>
```

PR #4187, opened 2016-09-21
Better support for re-running the Heroku generator

jkutner:
This PR makes it easier to run `yo jhipster:heroku` again after it has already been run the first time. It will automatically detect the Heroku app name, instead of asking for it again, and it will skip steps like app creation that aren't needed.

This also means that users are encourage to run `yo jhipster:heroku` every time they deploy their application. The instructions in the output have been updated accordingly.

I've also fixed some output formatting in the Maven build process.

### DIFF
--- a/generators/heroku/index.js
+++ b/generators/heroku/index.js
@@ -31,31 +31,49 @@ module.exports = HerokuGenerator.extend({
         this.angularAppName = this.getAngularAppName();
         this.buildTool = this.config.get('buildTool');
         this.applicationType = this.config.get('applicationType');
+        this.herokuAppName = this.config.get('herokuAppName');
     },
 
     prompting: function () {
         var done = this.async();
 
-        var prompts = [
-            {
-                type: 'input',
-                name: 'herokuDeployedName',
-                message: 'Name to deploy as:',
-                default: this.baseName
-            },
-            {
-                type: 'list',
-                name: 'herokuRegion',
-                message: 'On which region do you want to deploy ?',
-                choices: ['us', 'eu'],
-                default: 0
-            }];
+        if (this.herokuAppName && this.herokuAppName.length != 0) {
+            exec('heroku apps:info --json', function (err, stdout) {
+                if (err) {
+                    this.abort = true;
+                    this.log.error(err);
+                } else {
+                    var json = JSON.parse(stdout);
+                    this.herokuAppName = json['app']['name'];
+                    this.log(`Deploying as existing app: ${chalk.bold(this.herokuAppName)}`);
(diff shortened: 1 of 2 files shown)


generators/heroku/index.js
```diff
@@ -31,31 +31,49 @@ module.exports = HerokuGenerator.extend({
...
-                choices: ['us', 'eu'],
-                default: 0
-            }];
-
-        this.prompt(prompts).then(function (props) {
-            this.herokuDeployedName = _.kebabCase(props.herokuDeployedName);
-            this.herokuRegion = props.herokuRegion;
-            done();
-        }.bind(this));
+        if (this.herokuAppName && this.herokuAppName.length != 0) {
```
deepu105: Length check is not required as `if (this.herokuAppName)` itself checks for undefined, null and empty

generators/heroku/index.js
```diff
@@ -31,31 +31,49 @@ module.exports = HerokuGenerator.extend({
...
+            exec('heroku apps:info --json', function (err, stdout) {
+                if (err) {
+                    this.abort = true;
+                    this.log.error(err);
+                } else {
+                    var json = JSON.parse(stdout);
+                    this.herokuAppName = json['app']['name'];
```
deepu105: Is there a specific reason to obtain the bame this way rather than using the one obtained from this.config?

generators/heroku/index.js
jkutner: fixed

generators/heroku/index.js
jkutner: Good question. The value in `json['app']['name']` is the source of truth if there happens to be a conflict between what's in `.yo-rc` and what's returned by `heroku apps:info`. The only way there would be conflict is if the user changed the Heroku app outside of the `yo jhipster:heroku` workflow, which is unlikely.

Thus, the `this.config.get('herokuAppName')` property could be simplified to a boolean `this.config.get('herokuAppExists')`. But think it might prove useful in the future to have the app name in `.yo-rc` (it certainly doesn't hurt).

Makes sense?

deepu105: It indeed makes sense. But I have one more doubt. When you do `heroku apps:info --json` how do we know we are taking the correct app name if we have more than one app?  I have not used heroku much but I assume its a possibility rite?

jkutner: `heroku apps:info` gets info for a single application (despite the command name being plural). That application is the one set as your Git `heroku` remote. If there is no `heroku` remote, the command will error and so will this generator.

jdubois:
LGTM

PR #4196, opened 2016-09-22
[Sonar] utility classes should not have a public constructor

tusken1024:
[Sonar] utility classes should not have a public constructor.
This commit will  break code of users who have instanciated these classes or extended them

Also, Joshua Bloch said in Effective Java that the best way to create utiility classes is to use an enum. 
But, I didn't go futher than just making default constructor private and making classes final


jdubois:
Yes, this is really bad code from me -> I usually always do the private constructor, I don't know what I've done here :-)
This is really neat, I'm beginning to like Sonar!

PR #4197, opened 2016-09-22
[Sonar] iterate map with entrySet

tusken1024:
[Sonar] Iterate over the "entrySet" instead of the "keySet" when both key and value are needed


jdubois:
Thanks!
I really think the previous code was shorter and easier to read...

PR #4199, opened 2016-09-23
Clean Hikari builds

cbornet:


deepu105:
I remember having some issues when we tried this long back. not sure if we have any issues now. I couldnt recall what exactly it was

jdubois:
Yes I think there was some version issue, but that was long ago. I guess this is OK.

jdubois:
I tested and it looks OK, no error found... So I'm merging

PR #4203, opened 2016-09-23
adds bootstrap to testroot to prevent loading cloud config in unit tests

### LINKED ISSUE #4202: Entity tests fail in microservices due to cloud config lookup
In short, since 3.8.0 it works again to fetch config via cloud config for microservice applications, but JHipster is trying that during tests as well, which leads to fails in CI..

## reproduce

Generate microservice using version 3.8.0 (I used gradle option) and a entity (to have some tests), and run tests. It should fail with connect exceptions

## fix

adding a bootstrap.yml to testroot with

```
spring:
    cloud:
        config:
            enabled: false
```

xetys:
fixes #4202


PierreBesson:
@jdubois, I'm wondering if we shouldn't do a fix release for this as without it tests are failing for microservices or maybe we can just do a new release for Consul.

jdubois:
I won't have much time before Thursday, but if a bug fix release is needed
I can do something quick.

Le 25 sept. 2016 4:16 PM, "Pierre Besson" <notifications@github.com> a
écrit :

PR #4204, opened 2016-09-23
Use scala-maven-plugin for incremental and continuous compilation

### LINKED ISSUE #4164: Continuous compilation with Maven
When using an editor such as VS Code, it's interesting to have a continuous compilation of the sources so that the devtools magic operates seamlessly (kinda like Eclipse's "Build on save" setting).
It's very easy to do with gradle with ```./gradlew compileJava --continuous``` but maven doesn't have it out-of-the-box.
So I used scala-maven-plugin with the following config
```xml
            <plugin>
                <groupId>net.alchim31.maven</groupId>
                <artifactId>scala-maven-plugin</artifactId>
                <version>3.2.2</version>
                <executions>
                    <execution>
                        <id>scala-compile-first</id>
                        <phase>process-resources</phase>
                        <goals>
                            <goal>add-source</goal>
                            <goal>compile</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <recompileMode>incremental</recompileMode>
                    <verbose>true</verbose>
                    <scalaVersion>2.11.7</scalaVersion>
                </configuration>
            </plugin>
```
and then ```mvn scala:cc``` in a terminal to get the continuous compilation (incremental so blazing fast)

Now should I put the plugin in the generated pom.xml or write a tip for this ?

cbornet:
Fix #4164

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -32,6 +32,7 @@
         <%_ if (databaseType === 'sql') { _%>
         <hikaricp.version>2.4.6</hikaricp.version>
         <%_ } _%>
+        <argLine>-Djava.security.egd=file:/dev/./urandom -Xmx256m</argLine>
         <assertj-core.version>3.5.2</assertj-core.version>
         <awaitility.version>1.7.0</awaitility.version>
         <%_ if (databaseType === 'cassandra' || applicationType === 'gateway') { _%>
@@ -912,7 +913,6 @@
                 </executions>
                 <%_ } _%>
                 <configuration>
-                    <argLine>-Djava.security.egd=file:/dev/./urandom -Xmx256m ${surefireArgLine}</argLine>
                     <!-- Force alphabetical order to have a reproducible build -->
                     <runOrder>alphabetical</runOrder>
                 </configuration>
@@ -930,8 +930,6 @@
                         <configuration>
                             <!-- Sets the path to the file which contains the execution data. -->
                             <destFile>${project.testresult.directory}/coverage/jacoco/jacoco.exec</destFile>
-                            <!-- Sets the name of the property containing the settings for JaCoCo runtime agent. -->
-                            <propertyName>surefireArgLine</propertyName>
                         </configuration>
                     </execution>
                     <!-- Ensures that the code coverage report for unit tests is created after unit tests have been run -->
@@ -1197,7 +1195,7 @@
                     <plugin>
                         <groupId>net.alchim31.maven</groupId>
                         <artifactId>scala-maven-plugin</artifactId>
-                        <version>3.2.2</version>
+                        <version>3.2.0</version>
                         <executions>
                             <execution>
                                 <id>compile</id>
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1179,7 +1179,49 @@
                 </dependency>
             </dependencies>
             <build>
-                <plugins><% if(!skipClient) {%>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-compiler-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>default-compile</id>
+                                <phase>none</phase>
+                            </execution>
+                            <execution>
+                                <id>default-testCompile</id>
+                                <phase>none</phase>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>net.alchim31.maven</groupId>
+                        <artifactId>scala-maven-plugin</artifactId>
+                        <version>3.2.2</version>
+                        <executions>
+                            <execution>
+                                <id>compile</id>
+                                <phase>compile</phase>
+                                <goals>
+                                    <goal>add-source</goal>
+                                    <goal>compile</goal>
+                                </goals>
+                            </execution>
+                            <execution>
+                                <id>test-compile</id>
+                                <phase>test-compile</phase>
+                                <goals>
+                                    <goal>add-source</goal>
+                                    <goal>testCompile</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <recompileMode>incremental</recompileMode>
+                            <verbose>true</verbose>
+                            <scalaVersion>2.11.7</scalaVersion>
+                        </configuration>
+                    </plugin><% if(!skipClient) {%>
                     <plugin>
                         <groupId>com.github.eirslett</groupId>
                         <artifactId>frontend-maven-plugin</artifactId>


cbornet:
Now also works for continuous testing (`mvnw scala:cctest`)

jdubois:
@cbornet are you running `./mvnw scala:cc` ?

I'm getting:

    Failed to retrieve plugin descriptor for org.eclipse.m2e:lifecycle-mapping:1.0.0: Plugin org.eclipse.m2e:lifecycle-mapping:1.0.0 or one of its dependencies could not be resolved

I still find it really weird to use a Scala plugin to do this job, but then if it's the only one doing it, let's use it and document it (documentation will be important, because people will be lost here!!)

cbornet:
@jdubois Yes, that's what I run for continuous compilation.
For the eclipse-lifecycle-plugin error, are you sure it's related to this PR ?

jdubois:
Oh, maybe not. There was another issue on Eclipse, so maybe I just messed up stuff. I'll have a look again.

cbornet:
Note that the usual goals compile, package, test, etc.. still work

jdubois:
@cbornet now I understand: you put the plugin in the `prod` profile... Is that intentional?

cbornet:
No, it's an error !! Let me check !

jdubois:
OK, I got it working, on the `dev` profile of course :-)

Now just one issue, when you run the entity sub-generator: classes are generated before the changelog, so the app updates before the changelog is ready, and we have missing database tables in the end... So I think I'm going to change the file generation order!

jdubois:
And of course we will need to update the documentation!!

jdubois:
@cbornet there is an issue that I couldn't solve, when you run the entity sub-generator the tables are not automatically updated:

- the changelogs must be generated before the Java classes -> that's easy, we just need to move the function that generates them before the Java classes generation
- the Maven Scala Plugin must also process the resources -> I couldn't make it copy the resources... It's just handling the Java compilation, so the changelogs are not copied. I couldn't find a solution to this.

cbornet:
You mean in cc mode or also in regular compile mode ?

jdubois:
Only the cc mode, I didn't try anything.

What I'm doing:

- Run `./mvnw`
- Run `gulp`
- Run `./mvnw scala:cc`

Then I create an entity with `yo jhipster:entity Foo`, and the idea is that everything should be updated on the fly.
The only thing that fails is the Liquibase changelog, as it wasn't copied by "scala:cc" to the target directory.

cbornet:
Ah ! So you're trying to have the uber-fun demo where you do ```yo jhipster:entity Foo``` and your browser-sync'd app gets updated magically :smile: !
If this works that will be awesome !

jdubois:
Of course I want the full magic!!!!

jdubois:
@cbornet I can't make it work the way I want for the moment, and the Scala compiler is slowing down the "normal" build.
So I'm going to put it in a special profile, called "cc" for continuous compilation, so it doesn't pollute the rest of the build. And I'm going to try again, of course.

cbornet:
OK. I also think having it in a profile is perfectly fine.

jdubois:
I'm giving up having the resources copied... I just don't understand how this works :-(
So yes it has to be in a profile, as it's not fully working, it can't be our default choice.

cbornet:
Resource mdification monitoring may just not be part of scala:cc capabilities.

jdubois:
It looks like it doesn't, but:

- that would be quite a big issue with the plugin, so that seems weird
- I'm not good enough with Maven plugins, but can't we hook it to the "resources" plugin, so that it gets executed when a compilation is triggered?

cbornet:
Once again, easy to do with gradle : http://docs.spring.io/spring-boot/docs/current/reference/html/build-tool-plugins-gradle-plugin.html#build-tool-plugins-gradle-running-applications

cbornet:
Maybe this:
http://docs.spring.io/spring-boot/docs/1.4.1.RELEASE/maven-plugin/usage.html
```xml
<build>
  ...
  <plugins>
    ...
    <plugin>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-maven-plugin</artifactId>
      <version>1.4.1.RELEASE</version>
      <configuration>
        <addResources>true</addResources>
      </configuration>
    </plugin>
    ...
  </plugins>
  ...
</build>
```

cbornet:
@jdubois the addResources trick did it for me ! Can you check ?
As it seems to have implications if put on the build phase (disables resource filtering), we could have it only for cc profile.

jdubois:
Oh yes, it disables resources filtering, so we have the same issue as with #4256

It does work, but just like 50% of the time... There is still work to be done here, but we're getting there.

jdubois:
I got it working 'OK', I'm going to commit this and link it to this ticket.
I had to modify the gulp serve config, so it only works with monoliths.
So it's still not perfect, but that's working really fine in many scenarios!!!

cbornet:
:+1:

cbornet:
I have added this tasks.json in my .vscode
```
{
    "version": "0.1.0",
    "command": "yo",
    "isShellCommand": true,
    "args": [ "jhipster:import-jdl","${file}", "--force"],
    "showOutput": "always"
}
```
So I can run the task from the palette (CTRL+P)
Unfortunately it's not yet possible to bind a key to a task :disappointed:

jdubois:
@cbornet this is really cool!!!! Maybe we should have a sub-generator or a module for generating VS Code configurations.

deepu105:
No sub generator please, we already have a lot of them already :D A module
might be more appropriate

Thanks & regards,
Deepu

cbornet:
There is also the vscode extension but it doesn't seem possible to bundle tasks in it.

jdubois:
Ah we disagree at last :-)
I'd like to have a lot more small sub-generators, and be able to compose them. That's the NPM/Unix way of having small utilities you combine.
We need to discuss that!

PierreBesson:
We would have an ide-settings subgen that would configure project settings for vscode, idea and eclipse...

deepu105:
Ha ha ya at last :P
But note that I dont disagree fully here :P

I'm ok with sub gens my only worry is that too much of them will make user experience bad unless we can come up with an intuitive way to use them. Thats why I was suggesting `jhipster:util` which will compose all these small sub gens and provide a nice gui to select ehat users want easily.
And yes I understand the omzsh argument (atleast on windows and osx) but just see how many people outside of the core even knows about the `jhipster:service` sub gen? 
People will know and use stuff which we highlight a lot in docs, videos etc. Very few will actually dig through docs to find all options so if we have a easy to remember entry and highlight that then it will be actually useful, behind the scenes it can be any number of sub gens but I dont think we will be able to highlight and drive adoption for all of them directly 
Hope you understand my perspective im coming in from a User experience point

deepu105:
Corrected typos. I meant linux and osx not windows and osx

jdubois:
Yes @deepu105 but then that's the same idea than the oh-my-zsh plugin: we do need to offer an easier-to-use way of running those generators, than just remembering them all.

cbornet:
`yo jhipster:help` ?

PR #4211, opened 2016-09-25
update to typescript 2.0

christopherthielen:
update to typescript 2.0 and @types/* packages
remove dependency on typings package

Typescript 2.0 automatically uses typings from the npm "@types/*" org, such as '@types/angular'.
One downside is that during jhipster's "hybrid ng1/ng2" phase, @types/angular requires that you import `angular` from a package, i.e., `import * as angular from 'angular'`

I removed the dep on the 'typings' package, as well as `typings.json` and replaced the required ones with '@types/*' npm dependencies.


deepu105:
Awesome.  Thank you @christopherthielen our end goal is to have a pure ng2 app without any bower dependencies so this is great in that direction :)

PR #4220, opened 2016-09-27
Use hybrid adapter ui-router-ng1-to-ng2

christopherthielen:
- Use `ng1-to-ng2` adapter
  - Fetch packages angular-ui-router, ui-router-ng2, ui-router-ng1-to-ng2 from npm
- Defer routing bootstrap until angular 1 and angular 2 are done bootstrapping
- Register ng2 Account states by including them in the NgModule
- Route to ng2 component class
- Use ng2 directives: uiSref, uiSrefActivce
- Use transition hooks instead of `$stateChange*` event handlers
- Use ui-router-visualizer in dev mode


generators/client-2/templates/src/main/webapp/app/_app.main.ts
```diff
@@ -1,4 +1,11 @@
...
 import './app.module';
 
-upgradeAdapter.bootstrap(document.body, ['<%=angularAppName%>.app'], {strictDi: true});
+function readyFn(ref: UpgradeAdapterRef) {
+    let $urlRouter = ref.ng1Injector.get('$urlRouter');
+    $urlRouter.listen();
+    $urlRouter.sync();
+}
+
+upgradeAdapter.bootstrap(document.body, ['<%=angularAppName%>.app'], {strictDi: true}).ready(readyFn);
```
christopherthielen: ng1/ng2 bootstrap timing can be tricky.
Elsewhere, we do `$urlRouterProvider.deferIntercept()` to defer the initial url sync. Here, we know ng2 is ready, so we tell the router to sync.

generators/client-2/templates/src/main/webapp/app/_app.module.ts
```diff
@@ -45,7 +46,7 @@ angular
         'ngCacheBuster',
         'ngFileUpload',
         'ui.bootstrap',
-        'ui.router',
+        'ui.router.upgrade',
```
christopherthielen: the ng1-to-ng2 angular1 module `ui.router.upgrade` augments the angular 1 ui-router instance (`$uiRouter`) and provides ng2 capabilities such as observable parameters and ng2 "view factories"

generators/client-2/templates/src/main/webapp/app/account/_account.module.ts
```diff
@@ -49,16 +39,6 @@ angular
...
-    .config(AccountStateConfig)
-    .config(ActivateStateConfig)
-    .config(RegisterStateConfig)
-    .config(PasswordStateConfig)
-    .config(PasswordResetInitStateConfig)
-    .config(PasswordResetFinishStateConfig)
-    .config(SettingsStateConfig)
-    <%_ if (authenticationType === 'session') { _%>
-    .config(SessionsStateConfig)
-    <%_ } _%>
```
christopherthielen: Don't have to register the ng2 states in a ng1 `.config()` block (although you _can_).
Instead of config blocks, the states are collected in an array and added to the `UIRouterModule.forChild()` NgModule, then imported into the root NgModule.

generators/client-2/templates/src/main/webapp/app/account/_account.ng2module.ts
```diff
@@ -17,9 +18,36 @@ import { SessionsComponent } from './sessions/sessions.component';
...
+    sessionsState,
+    <%_ } _%>
+    settingsState,
+];
+
 @NgModule({
     imports: [
-        <%=angular2AppName%>SharedModule
+        <%=angular2AppName%>SharedModule,
+        UIRouterModule.forChild({ states: ACCOUNT_STATES })
```
christopherthielen: By declaring the states on the module, the states will be automatically registered when the module loads.

generators/client-2/templates/src/main/webapp/app/account/activate/_activate.state.ts
```diff
@@ -1,23 +1,20 @@
...
+export const activateState = {
+    name: 'activate',
+    parent: 'account',
+    url: '/activate?key',
+    data: {
+        authorities: [],
+        pageTitle: 'activate.title'
+    },
+    views: {
+        'content@': { component: ActivateComponent }
```
christopherthielen: route directly to an ng2 Component class.

generators/client-2/templates/src/main/webapp/app/account/activate/_activate.state.ts
```diff
@@ -1,23 +1,20 @@
...
+    },
+    resolve: {
```
christopherthielen: resolve syntax for ng2 has changed, but it is supports ng1 style.  
I didn't migrate any resolves since mostly ng1 services are being utilized.

generators/client-2/templates/src/main/webapp/app/admin/user-management/_user-management.state.ts
```diff
@@ -74,7 +74,6 @@ export function UserMgmntStateConfig($stateProvider) {
         }
     })
     .state('user-management.new', {
-        parent: 'user-management',
```
christopherthielen: declaring a parent using `name: 'parent.child'` *AND* having a `parent: ` property is ambiguous, and now throws an error

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -1,29 +1,30 @@
...
+StateHandler.$inject = ['$rootScope', '$transitions', <% if (enableTranslation) { %>'$translate', /*'<%=jhiPrefixCapitalized%>LanguageService',*/ 'TranslationHandler',<% } %> '$window',
+        'Auth', 'Principal', '$uiRouter', '$trace'];
 
-StateHandler.$inject = ['$rootScope', '$state', '$sessionStorage', <% if (enableTranslation) { %>'$translate', /*'<%=jhiPrefixCapitalized%>LanguageService',*/ 'TranslationHandler',<% } %> '$window',
-        'Auth', 'Principal'];
+export function StateHandler($rootScope, $transitions, <% if (enableTranslation) { %>$translate, /*<%=jhiPrefixCapitalized%>LanguageService,*/ TranslationHandler,<% } %> $window,
+    Auth, Principal, $uiRouter, $trace) {
+
+    if (DEBUG_INFO_ENABLED) {
+        $trace.enable('TRANSITION');
```
christopherthielen: enable `$trace` in the console. Also show the ui-router visualizer.

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -1,29 +1,30 @@
...
-    Auth, Principal) {
     return {
         initialize: initialize
     };
 
     function initialize() {
         $rootScope.VERSION = VERSION;
+        let deregistrationFns = [];
 
-        var stateChangeStart = $rootScope.$on('$stateChangeStart', function (event, toState, toStateParams, fromState) {
```
christopherthielen: `$stateChangeStart` and friends are deprecated. They're not included in 1.0.0 by default (although there is a polyfill to re-enable them).

Switched this logic to the far more powerful transition hooks.

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -35,25 +36,27 @@ export function StateHandler($rootScope, $state, $sessionStorage, <% if (enableT
                 $translate.use(language);
             });*/
             <% } %>
-        });
+        }));
 
-        var stateChangeSuccess = $rootScope.$on('$stateChangeSuccess',  function(event, toState, toParams, fromState, fromParams) {
+        // Redirect to a state with an external URL (http://stackoverflow.com/a/30221248/1098564)
+        deregistrationFns.push($transitions.onStart({ to: state => state.external }, (transition: Transition) => {
```
christopherthielen: note how the  `{ to: function(state) { return !!state.external } }` criteria object only runs the hook for matching transitions.

generators/client-2/templates/src/main/webapp/app/layouts/navbar/navbar.html
```diff
@@ -14,14 +14,14 @@
...
-                    <a ui-sref="home" href="#/home" (click)="collapseNavbar()">
+                <li uiSrefActive="active">
+                    <a uiSref="home" (click)="collapseNavbar()">
                         <span class="glyphicon glyphicon-home"></span>
                         <span class="hidden-sm" translate="global.menu.home">Home</span>
                     </a>
                 </li>
                 <!-- jhipster-needle-add-element-to-menu - JHipster will add new menu items here -->
-                <li [ngClass]="{active: $state.includes('entity')}" *ngSwitchCase="true" ngbDropdown class="dropdown pointer">
+                <li uiSrefActive="active" *ngSwitchCase="true" ngbDropdown class="dropdown pointer">
```
christopherthielen: `uiSrefActive` in ng2 can enclose multiple `uiSref`.  It will add the 'active' class if any of those uiSref are active.

generators/client-2/templates/src/main/webapp/system.config.js
```diff
@@ -48,7 +51,6 @@
     ngPackageNames.forEach(setPackageConfig);
 
     var config = {
-        meta: { "ui-router-ng2": { format: "cjs" } },
```
christopherthielen: this is no longer necessary.  systemjs was incorrectly detecting ui-router-ng2.js as es6 module, not commonjs

deepu105:
Awesome and thank you so very much. Especially for the explanations :) you rock

generators/generator-base.js
```diff
@@ -46,8 +46,8 @@ Generator.prototype.addElementToMenu = function (routerName, glyphiconName, enab
             file: fullPath,
             needle: 'jhipster-needle-add-element-to-menu',
             splicable: [
-                '<li ui-sref-active="active" >\n' +
-                '                    <a ui-sref="' + routerName + '" ng-click="vm.collapseNavbar()">\n' +
+                '<li uiSrefActive="active" >\n' +
```
deepu105: These 3 methods are used by ng1 client generator and hence would break those. We need to add new methods or if else blocks for ng2 client. But its required only for entity creation which we havent migrated yet, hence we can change these methods later

generators/client-2/templates/src/main/webapp/app/_app.main.ts
deepu105: I think I never figured this part when i tried and that might have made things problematic

generators/client-2/templates/src/main/webapp/app/account/_account.ng2module.ts
deepu105: This as well I was trying to add a UIRouterModule

sendilkumarn:
@christopherthielen  👏 awesome work and thanks.

Immediately after login, the routing is not happening. But upon refresh it just worked great. Is this because of that `race condition` between ng1 and ng2.

wmarques:
@christopherthielen thank you very much for this work :clap: I'll check this soon !

generators/generator-base.js
deepu105: I think the test fails are from this

christopherthielen: I reverted those lines in the generator. Tests pass now.
Is there a way to run the e2e tests locally (without travis)?

christopherthielen: Also, I wonder if protractor will even work with a pure ng2 app.

mraible: Yes, protractor will work, you just have to use `by.id` and `by.css`. You can see https://github.com/mraible/ng2-demo for examples.

deepu105: @pascalgrimaud might know a way to run travis tests locally I guess

christopherthielen:
> Immediately after login, the routing is not happening.

@sendilkumarn this seems to be caused by the top level `uiSrefActive`(s) when they have multiple `uiSref` (the dropdowns, like 'Entities', 'Account', 'Administration').  
I've logged a ticket here https://github.com/angular-ui/ui-router/issues/3046

In the meantime, those three `uiSrefActive` can be removed and it seems to work.

PR #4223, opened 2016-09-27
fix status and health check urls that are used by eureka

ruddell:
From the [Spring Cloud Netflix docs](http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html), we need to set the management context-path so the urls point to the correct endpoints.
`/health -> /management/health`
`/info -> /management/info`


PierreBesson:
This is because we have actually changed the context-path some time after the 3.0 release. However, those endpoints are not used for eureka healthchecks (those are made by the eureka client embedded in every app).  AFAIK those are only used to create links to the app in Eureka default UI (that we don't use).

jdubois:
Yes @PierreBesson  this probably isn't used at all. I'm still merging this as indeed I'd rather have this configuration right, just in case.

PR #4240, opened 2016-10-01
migrate home states to ui-router-ng2

deepu105:
@christopherthielen I migrated the admin routes to ui-router-ng2 but when i tried the same for the app state and home state it doesnt work. any hints? I tried both `UIRouterModule.forRoot` and `UIRouterModule.forChild` but nothing happens and there is no console error


deepu105:
@christopherthielen could you please take a glance and let me know what im doing wrong here?

PR #4242, opened 2016-10-01
set default serviceDiscoveryType if missing in .yo-rc.json

ruddell:
If you run the generator on an old microservice, `serviceDiscoveryType` could be undefined.  This sets the default to `eureka` so the project can compile.

I also changed all comparisons from 'no' to false to match the value put in the .yo-rc.json.

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -97,6 +97,9 @@ module.exports = DockerComposeGenerator.extend({
             this.useElk = this.config.get('useElk');
             this.useKafka = false;
             this.serviceDiscoveryType = this.config.get('serviceDiscoveryType');
+            if (this.serviceDiscoveryType === undefined) {
+                this.serviceDiscoveryType = 'eureka';
+            }
             this.adminPassword = this.config.get('adminPassword');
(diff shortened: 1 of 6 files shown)


PierreBesson:
Thanks a lot @ruddell ! I didn't know this needed to be done.

ruddell:
@PierreBesson Thanks a lot for the Consul option!

generators/docker-compose/index.js
```diff
@@ -315,7 +318,7 @@ module.exports = DockerComposeGenerator.extend({
             if(this.serviceDiscoveryType === 'consul'){
                 this.template('_consul.yml', 'consul.yml');
             }
-            if(this.serviceDiscoveryType !== 'no'){
+            if(this.serviceDiscoveryType !== false){
```
deepu105: Can we simplify this to `if(this.serviceDiscoveryType) ` since it could only be false or a value

ruddell: Do you want this changed everywhere or just this instance where serviceDiscoveryType is checked alone?

deepu105: Yes please

PR #4248, opened 2016-10-02
ng2: clean up missing references and fix compile errors

christopherthielen:
Also switch to ng2 injection for `$state: StateService`


sendilkumarn:
LGTM 👍 Thanks again @christopherthielen

deepu105:
Thanks a lot for the cleanup. My bad I tested on my generated app and then did the templates and was too confident to regenerate and check again.

Note to self: move that lazy ass and test after templating again. :D

PR #4255, opened 2016-10-03
ng2: fix ui-router-ng2 initialization

christopherthielen:
inject `UrlRouter` from `ui-router-ng2` triggers ng2 initialization

this is apparently necessary because none of the ui-router-ng2 code was injected, so it never got initialized.  Injecting `UrlRouter` triggers DI which in turn initializes the ng1-to-ng2 code, which in turn registers states, etc.

Fell free to squash or whatever. I merged from deepu105/ng2-ui-router

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
+++ b/generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
@@ -13,6 +13,7 @@ import { XSRFStrategyProvider } from './shared/XSRF-strategy.provider';
 import { HomeComponent } from './home/home.component';
 import { NavbarComponent } from './layouts/navbar/navbar.component';
 import { FooterComponent } from './layouts/footer/footer.component';
+import { ErrorComponent } from './layouts/error/error.component';
 <%_ if (enableTranslation){ _%>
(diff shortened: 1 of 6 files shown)


deepu105:
Thank you :)

generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
```diff
@@ -12,14 +13,30 @@ import { XSRFStrategyProvider } from './shared/XSRF-strategy.provider';
...
+        errorState,
+        accessdeniedState
+    ]
+};
+
 @NgModule({
     imports: [
         BrowserModule,
         Ng1ToNg2Module,
+        UIRouterModule.forChild(routerConfig),
```
deepu105: @christopherthielen in what case should we use the forRoot method then?

christopherthielen: @deepu105 forRoot provides a new instance of UIRouter object.  It should not be used when using ng1-to-ng2 because the instance is already provided by angular 1 and Ng1ToNg2Module. Once The app is pure ng2 then you need forRoot.

Issue #4252, opened 2016-10-03
securing consul communication using ACLs


xetys:
Similar to JHipster Registry we need some securing solution for config and service discovery for consul...

I spent some time on this topic, and figured out some solution which is almost equivalent to the way we do it for jhipster-registry. 

to sum up, there is a clean way or the fast way (as usual). For the fast way we initialize consul with a master acl token (which has maxmimum permissions), and use it for accessing services and KV. The clean way is to use the master token only to generate some "JHipster ACL client token", which gets a ***default*** read/write permission to services and read on KV. The details of that clients policies is open for later modification / specification. This flow may be provided in our docker setup, if there is any consideration to use it in production. 

For the spring side, its just easy as setting "consul.token" via configuration. For further work I need to know,
if we prefer the more secure solution, by:

* setting a acl_config.json in consul docker files with a predefined master token (similar to JWT secret, can be generated per project)
* changing the consul-config-loader, so it creates the JHipster client ACL using master token
* setting client token to docker files

or a little more straight foreward and less secure solution, by:

* setting a acl_config.json in consul docker files with a predefined master token (similar to JWT secret, can be generated per project)
* setting master token to docker files

I'll PR as soon I've heard some opinions on that :smile:

PierreBesson:
Great ! Thank you for doing this research. I will need to think more about
it but the most important thing is that it should work out of the box in
dev and in prod. So a master token could be generated by the docker-compose
subgen but for dev maybe we could put a predefined non-secure master token
so that it is already shared by all microservices when run in dev.

Also I like the idea of extending consul-config-loader to setup ACLs.

Anyway I'm not sure I understood everything about ACL tokens so you can go
ahead.

Regards,
Pierre

Le 3 oct. 2016 4:23 PM, "David Steiman" <notifications@github.com> a écrit :

Similar to JHipster Registry we need some securing solution for config and
service discovery for consul...

I spent some time on this topic, and figured out some solution which is
almost equivalent to the way we do it for jhipster-registry.

to sum up, there is a clean way or the fast way (as usual). For the fast
way we initialize consul with a master acl token (which has maxmimum
permissions), and use it for accessing services and KV. The clean way is to
use the master token only to generate some "JHipster ACL client token",
which gets a *default* read/write permission to services and read on KV.
The details of that clients policies is open for later modification /
specification. This flow may be provided in our docker setup, if there is
any consideration to use it in production.

For the spring side, its just easy as setting "consul.token" via
[...]

PR #4259, opened 2016-10-04
added ACL support to consul config and service discovery

xetys:
this goes in conjunction with https://github.com/jhipster/consul-config-loader/pull/1 to solve #4252.

It actually changes some configs of the consul container, starting it now with ACL enabled and a self defined master acl token (to change in acl_config.json), and assuming some present client ACL token named "to-change-in-production-client", which is default by the new consul config loader


generators/docker-compose/templates/_consul.yml
```diff
@@ -6,13 +6,18 @@ services:
...
             - 8500:8500
             - 8600:8600
-        command: consul agent -dev -ui -client 0.0.0.0
+        volumes:
+            - ./consul-conf:/consul-conf
+        command: consul agent -dev -ui -client 0.0.0.0 -config-dir /consul-conf
 
     consul-config-loader:
-        image: pbesson/consul-config-loader
+        image: jhipster/consul-config-loader
```
PierreBesson: Please add the v0.1.0 tag to this image. This must be done before the next release.

generators/docker-compose/templates/_consul.yml
```diff
@@ -11,7 +11,7 @@ services:
         command: consul agent -dev -ui -client 0.0.0.0 -config-dir /consul-conf
 
     consul-config-loader:
-        image: jhipster/consul-config-loader
+        image: jhipster/consul-config-loader:v0.1.0
```
pascalgrimaud: Can we change `jhipster/consul-config-loader:v0.1.0` to a CONST here: [generator-constants.js](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js)

So it will be easier to update Docker images, instead of looking everywhere in the code

deepu105:
@xetys please use the `fix #issue` keyword to close automatically the related issues

Issue #4274, opened 2016-10-06
User-Managment counter


mescuderocuu:
In the user management template is counting 4 users when you start a project but only displays 3 of them.

jdubois:
Yes, we hide the "anonymous" user, which is required by Spring Security.

jdubois:
Fixed by #4275

PR #4275, opened 2016-10-06
#4274 Fixing vm.totalItems

mescuderocuu:
The vm.totalItems count all the users even when you subtract the 'anonymoususer', i add a counter so it count only the ones that are displays.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
+++ b/generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
@@ -53,15 +53,15 @@
 
         function onSuccess(data, headers) {
             //hide anonymous user from user management: it's a required user for Spring Security
-            var x = 0;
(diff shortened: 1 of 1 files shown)


jdubois:
In the commit message you could add a keyword like "Fix" to automatically close a ticket with a commit. Have a look at https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#commit

generators/client/templates/src/main/webapp/app/admin/user-management/_user-management.controller.js
```diff
@@ -53,13 +53,15 @@
 
         function onSuccess(data, headers) {
             //hide anonymous user from user management: it's a required user for Spring Security
+            var x = 0;
```
mescuderocuu: Fix in cbed2d9ff87fef8fb5274941394828019872eade

PR #4298, opened 2016-10-09
Fixed JWT project compiles now. Small fixes.

allahbaksh:
I have fixed some of the compile error when the JWT option is chosen.

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/layouts/_index.ts
+++ b/generators/client-2/templates/src/main/webapp/app/layouts/_index.ts
@@ -2,7 +2,9 @@ export * from './error/error.component';
 export * from './error/error.state';
 export * from './footer/footer.component';
 export * from './navbar/navbar.component';
+<%_ if (enableTranslation){ _%>
 export * from './navbar/active-menu.directive';
+<%_ } _%>
 export * from './profiles/page-ribbon.component';
 export * from './profiles/profile.service';
 export * from './profiles/profile-info.model';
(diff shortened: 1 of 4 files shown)


generators/client-2/templates/src/main/webapp/app/layouts/_index.ts
```diff
@@ -2,7 +2,8 @@ export * from './error/error.component';
 export * from './error/error.state';
 export * from './footer/footer.component';
 export * from './navbar/navbar.component';
-export * from './navbar/active-menu.directive';
+<%_ if (enableTranslation){ _%>export * from './navbar/active-menu.directive';
```
deepu105: The export statement should be in a new line. The templating with  _ automatically trims space and lines

generators/client-2/templates/src/main/webapp/app/shared/auth/_auth.service.ts
```diff
@@ -2,9 +2,11 @@ import { Injectable, Inject } from '@angular/core';
 import { StateService } from 'ui-router-ng2';
 import { LoginService } from "../login/login.service";
 import { Principal } from './principal.service';
-import { AuthServerProvider } from './auth-session.service';
-import { <%=jhiPrefixCapitalized%>TrackerService } from '../tracker/tracker.service';
-
+import { AuthServerProvider } from  <%_ if (authenticationType === 'jwt') { _%> './auth-jwt.service';
```
deepu105: For inline template `<%_` shouldn't be used. Thats for neline templates. Use `<%` for inline

deepu105: Better use newline template with seperate imports, refer how its done in shared/index.ts

deepu105:
thank you merged

PR #4304, opened 2016-10-10
Ng2 translate

sendilkumarn:
Added [ng2-translate](https://github.com/ocombe/ng2-translate)

@deepu105 need to change the logic which loads the constant file with language constants.


sendilkumarn:
check dad8b19 for what has to done for translation.

generators/client-2/templates/src/main/webapp/app/home/home.html
```diff
@@ -4,7 +4,7 @@
             <span class="hipster img-responsive img-rounded"></span>
         </div>
         <div class="col-md-8">
-            <h1 translate="home.title">Welcome, Java Hipster!</h1>
+            <h1 [jhi-translate]="'home.title' | translate">Welcome, Java Hipster!</h1>
```
deepu105: why do we need both directive and pipe? cant we do with one?

generators/client-2/templates/src/main/webapp/app/layouts/navbar/_navbar.component.ts
```diff
@@ -23,6 +27,7 @@ export class NavbarComponent implements OnInit {
         private loginService : LoginService,
         <%_ if (enableTranslation){ _%>
         private languageService: <%=jhiPrefixCapitalized%>LanguageService,
+        private translateService: TranslateService,
```
deepu105: is it used?

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
```diff
@@ -24,12 +34,19 @@ import {
 @NgModule({
     imports: [
         <%=angular2AppName%>SharedLibsModule,
-        <%=angular2AppName%>SharedCommonModule,
+        <%=angular2AppName%>SharedCommonModule<%_ if (enableTranslation){ _%>,
+        TranslateModule.forRoot({
```
deepu105: can we move this to a provider file and call it here? similar to alert provider

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
```diff
@@ -43,15 +60,28 @@ import {
...
+        TranslateModule,
+        <%=jhiPrefixCapitalized%>Translate
     ],
     schemas: [CUSTOM_ELEMENTS_SCHEMA]
 
 })
-export class <%=angular2AppName%>SharedModule {}
+export class <%=angular2AppName%>SharedModule {
+    <%_ if (enableTranslation){ _%>
+     static forRoot(): ModuleWithProviders {
```
deepu105: why is this required?

generators/client-2/templates/src/main/webapp/app/home/home.html
deepu105: cant we use the pipe inside the directive? and also do away with the inner single quotes?

generators/client-2/templates/src/main/webapp/app/shared/language/_jhiMissing.translation.ts
```diff
@@ -0,0 +1,7 @@
+import {MissingTranslationHandler} from 'ng2-translate/ng2-translate';
+
+export class <%=jhiPrefixCapitalized%>MissingTranslationHandler implements MissingTranslationHandler {
```
deepu105: can we rename the file to `jhi-missing-translation-handler.config.ts` ?

generators/client-2/templates/src/main/webapp/app/shared/language/_language.service.ts
```diff
@@ -1,25 +1,37 @@
-import { LANGUAGES } from './language.constants';
 import { Injectable, Inject } from '@angular/core';
 
+import { TranslateService } from 'ng2-translate/ng2-translate';
+
+import { LANGUAGES } from './language.constants';
+import { <%=angular2AppName%>SharedCommonModule } from '../../shared';
```
deepu105: why is the module imported?

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
deepu105: I think this should be in the main AppModule as that is the root

deepu105: Ok I just read the docs, so it can be in a shared module. In that case better add it to the `<%=angular2AppName%>SharedLibsModule` and add the custom translation related directives/pipes in the `SharedCommonModule` as this module imports both those

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
```diff
@@ -24,12 +34,19 @@ import {
 @NgModule({
     imports: [
         <%=angular2AppName%>SharedLibsModule,
-        <%=angular2AppName%>SharedCommonModule,
+        <%=angular2AppName%>SharedCommonModule<%_ if (enableTranslation){ _%>,
+        TranslateModule.forRoot({
+          provide: TranslateLoader,
+          useFactory: (http: Http) => new TranslateStaticLoader(http, '/i18n', '.json'),
```
deepu105: as per docs isnt this the default `TranslateStaticLoader` then do we have to overwrite it?

generators/client-2/templates/src/main/webapp/app/shared/directive/jhi-translate.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(private el: ElementRef, private renderer: Renderer) { }
+
+  @Input()
+  set jhiTranslate(value: string | TemplateRef<any>) {
+      this._jhiTranslate = value;
+      this.setTranslatedValue();
+  }
+
+  setTranslatedValue() {
+    if (this._jhiTranslate ) {
```
deepu105: use service to translate key so that we can avoid the pipe and do `jhi-translate="home.title"` https://github.com/ocombe/ng2-translate#4-use-the-service-or-the-pipe

deepu105:
Good work @sendilkumarn please take a look at my comments

generators/client-2/templates/src/main/webapp/app/home/home.html
sendilkumarn: `pipe` is following ng2-translate.
`directive` manually done.

They have a method inside But I don't want to use that and re-engineer the thing.

The api needs `single-quote`.

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
sendilkumarn: To make `TranslateService` as a singleton.

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
sendilkumarn: no default is `assets/i18n`

generators/client-2/templates/src/main/webapp/app/shared/directive/jhi-translate.ts
sendilkumarn: I somehow feel using this will slow things down. I will check the internals

generators/client-2/templates/src/main/webapp/app/home/home.html
sendilkumarn: This one is tricky. we have to trigger this directive when the language changes, which will be an overhead.

generators/client-2/templates/src/main/webapp/app/layouts/navbar/_navbar.component.ts
sendilkumarn: removed 👍

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
sendilkumarn: This one has some wiring problems. Can you check this part. I don't know, moving this chunk over that file is causing problem.

generators/client-2/templates/src/main/webapp/app/shared/language/_jhiMissing.translation.ts
sendilkumarn: Done 👍

generators/client-2/templates/src/main/webapp/app/shared/language/_language.service.ts
sendilkumarn: Removed 👍

generators/client-2/templates/src/main/webapp/app/shared/directive/jhi-translate.ts
sendilkumarn: Refer my first comment.

generators/client-2/templates/src/main/webapp/app/home/_home.component.ts
```diff
@@ -14,7 +18,12 @@ export class HomeComponent implements OnInit {
 
     constructor(private principal: Principal,
                 private $state: StateService,
-                private loginService : LoginService) {
+                private loginService : LoginService<%_ if (enableTranslation){ _%>,
+                translateService: TranslateService,
+                <%=jhiPrefix%>LangService: <%=jhiPrefixCapitalized%>LanguageService<%_ } _%>) {
+        <%_ if (enableTranslation){ _%>
+        <%=jhiPrefix%>LangService.setLocation('home');
```
sendilkumarn: Done 👍 👏

generators/client-2/templates/src/main/webapp/app/home/home.html
deepu105: Ya that would be there with or without the pipe rite?
I would prefer a simpler directive which does everything. we can use the service inside the directive. Also I think its possible to get the directive evaluated during digest cycles

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
deepu105: ok

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
deepu105: ok

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
deepu105: doc says below

> By default, only the TranslateStaticLoader is available. It will search for files in i18n/*.json, if you want you can customize this behavior by changing the default prefix/suffix:

can you check

generators/client-2/templates/src/main/webapp/app/shared/directive/jhi-translate.ts
deepu105: I dont think there will be much difference here. The pipe would be using the service most probably so we can just use it directly

generators/client-2/index.js
```diff
@@ -501,6 +501,7 @@ module.exports = JhipsterClientGenerator.extend({
             this.template(ANGULAR_DIR + 'shared/directive/maxbytes.directive.ts', ANGULAR_DIR + 'shared/directive/maxbytes.directive.ts', this, {});
             this.template(ANGULAR_DIR + 'shared/directive/minbytes.directive.ts', ANGULAR_DIR + 'shared/directive/minbytes.directive.ts', this, {});
             this.template(ANGULAR_DIR + 'shared/directive/number-of-bytes.ts', ANGULAR_DIR + 'shared/directive/number-of-bytes.ts', this, {});
+            this.template(ANGULAR_DIR + 'shared/directive/jhi-translate.ts', ANGULAR_DIR + 'shared/directive/jhi-translate.ts', this, {});
```
deepu105: move this into the language folder

sendilkumarn: No this has to be in directive moving this will make us to create `language` folder even when there is no translation

generators/client-2/templates/src/main/webapp/app/home/home.html
deepu105: I dont like the markup its redundant. Let me see if there is a way to make it work kust with directive

generators/client-2/index.js
sendilkumarn: to return empty so that it will not replace anything. else everywhere in html we have to add in a check

deepu105: Ok currently we do a regex strip of translate directives and pipes so that if translation is disabled the generated project is clean. We need to do the same here as well. Once migration is complete we can modify the regex to strip these as well. So no need to have the directive added when there is no translation

deepu105:
@sendilkumarn i'm working on this

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
sendilkumarn: Yeah Okay.

generators/client-2/templates/src/main/webapp/app/shared/_shared.ng2module.ts
sendilkumarn: Okay 👍

generators/client-2/templates/src/main/webapp/app/home/home.html
deepu105: @sendilkumarn the solution for this is too simple and stupid, we just need to think from a stupid perspective 😉 
```js
import { Component } from '@angular/core';

@Component({
  selector: '[jhi-translate],[jhiTranslate]',
  template: '{{ key | translate }}',
  inputs: ['key:jhi-translate']
})
export class JhiTranslate {

  private key: string;
  //TODO add support to pass translate-values and for translate-compile

}
```

Issue #4296, opened 2016-10-09
jhipster:upgrade fails on the devbox


cbornet:
##### **Overview of the issue**

When upgrading a jhipster app on the devbox. The process fails and the app isn't upgraded

##### **Reproduce the error**

Download the 3.8.0 devbox or install manually with `sudo npm install -g generator-jhipster@3.8.0`
Generate an app (with `--skip-client` for quick generation)
Run `yo jhipster:upgrade`

##### **Suggest a Fix**

Several issues there:
1/ the default shell for the devbox is zsh and the command [rm -rf !(.yo-rc.json|.git)](https://github.com/jhipster/generator-jhipster/blob/master/generators/upgrade/index.js#L44) fails on this shell
2/ the global npm folder is `/usr/lib/modules` and needs root permissions for writing so the installation of the new version fails.

The workarounds are:
* change to bash shell before launching the upgrade (but most of the time, I forget and I lose a lot of time...)
* in the devbox, change ownership of /usr/lib/node_modules to vagrant (not sure it is secure)

Wouldn't it be better to install the new version locally ? I think there are a lot of reasons that could make the global install fail.

##### **JHipster Version(s)**

3.8.0/3.9.0

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**
```json
{
  "generator-jhipster": {
    "jhipsterVersion": "3.8.0",
    "baseName": "standard",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "47b4e760a1684605304b6a03d26cc6f2496a0914",
    "useSass": false,
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ],
    "serviceDiscoveryType": false,
    "skipClient": true
  }
}
```
##### **Browsers and Operating System**

Devbox : ubuntu 16 amd64

cbornet:
[...]

PR #4306, opened 2016-10-10
Improve upgrade subgen

### LINKED ISSUE #4294: Speedup upgrade by skipping js install (--skip-install)
I think this would seriously improve the jhipster:upgrade execution time since the node_modules folder is gitignored anyway.
And jhipster:upgrade should also recommend to run `npm install` at the end (or run it itself)

cbornet:
* use already installed generator for current version if available
* install new generator locally since global installation can fail
* skip installation during merge phase and do it only at the end
* fix cleanup function

Fix #4294 #4296

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -41,14 +41,17 @@ module.exports = UpgradeGenerator.extend({
     },
 
     _cleanUp: function() {
-        if (shelljs.rm('-rf', '!(.yo-rc.json|.git)').code === 0 ) {
-            this.log('Cleaned up directory');
-        }
+        shelljs.ls('-A').forEach( file => {
+            if(['.yo-rc.json', '.jhipster', 'node_modules', '.git'].indexOf(file) === -1) {
+                shelljs.rm('-rf', file);
+            }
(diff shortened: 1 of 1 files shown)


generators/upgrade/index.js
```diff
@@ -214,7 +224,7 @@ module.exports = UpgradeGenerator.extend({
         updateJhipster: function() {
             this.log(chalk.yellow('Updating ' + GENERATOR_JHIPSTER + '. This might take some time...'));
             var done = this.async();
-            shelljs.exec('npm install -g ' + GENERATOR_JHIPSTER, {silent:true}, function (code, msg, err) {
+            shelljs.exec('npm install ' + GENERATOR_JHIPSTER, {silent:true}, function (code, msg, err) {
```
deepu105: not sure if it should be only installed locally. Why do you think it can fail if installed globally?

cbornet: Eg. on ubuntu you need to install global packages with sudo

cbornet: You could also want to install your global packages with yarn

PR #4314, opened 2016-10-12
add addEntryToEhcache to modules index and generify it

ruddell:
Without `addEntryToEhcache` in modules/index.js, calling addEntityToEhcache from a module throws `Undefined: this.addEntryToEhcache`.

Also, when calling `addEntryToEhcache` from a module, this.packageName is always undefined.  I tried to bind(this) on `jhipsterFunc.addEntryToEhcache` but the module threw an error stating that it couldn't bind to undefined.  By making the method more generic, modules can now correctly add ehcache entries.

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -428,22 +428,20 @@ Generator.prototype.addAngularJsInterceptor = function (interceptorName) {
  */
 Generator.prototype.addEntityToEhcache = function (entityClass, relationships) {
     // Add the entity to ehcache
-    this.addEntryToEhcache(entityClass);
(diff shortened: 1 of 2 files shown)


deepu105:
This would be a breaking change for any modules using this method today and hence would be against semver conventions. I would suggest you add a new method if required to handle the new signature and keep the old one intact.

ruddell:
`addEntryToEhcache` wasn't available to modules until this PR and `addEntityToEhcache` was broken.  Both methods result in the same output as before (using the same input).  What is breaking about this change?

deepu105:
Oh sorry, I didn't notice it wasnt open before

PR #4322, opened 2016-10-12
#4320 : Wrong column name when double uppercase

### LINKED ISSUE #4320: Wrong column name in case of double uppercase

Jhipster generator create a wrong liquibase changelog when we have a double Uppercase name.
And we have an error on unit test:

`org.hibernate.HibernateException: Missing column: socrevision_id in SERVER.PUBLIC.FILE`

I have a relation in my uml : soCRevision.

According to SpringNamingStrategy:

```
protected static String addUnderscores(String name) {
		StringBuilder buf = new StringBuilder( name.replace('.', '_') );
		for (int i=1; i<buf.length()-1; i++) {
			if (
				Character.isLowerCase( buf.charAt(i-1) ) &&
				Character.isUpperCase( buf.charAt(i) ) &&
				Character.isLowerCase( buf.charAt(i+1) )
			) {
				buf.insert(i++, '_');
			}
		}
		return buf.toString().toLowerCase(Locale.ROOT);
	}
```

And in my liquibase:

```
<column name="so_c_revision_id" type="bigint">
                <constraints nullable="true" />
            </column>
```

because unit test doesn't pass. ;o) 

##### **Reproduce the error**

create the project with the following files

##### **Suggest a Fix**
I think the problem is here:
https://github.com/jhipster/generator-jhipster/blob/01c01f63bb3206526cf92a34800475265a5e7e0d/generators/generator-base.js#L1164
```
Generator.prototype.getColumnName = function (value) {
    return _.snakeCase(value).toLowerCase();
};
```
must be replace by the Spring version:
```
protected static String addUnderscores(String name) {
		StringBuilder buf = new StringBuilder( name.replace('.', '_') );
		for (int i=1; i<buf.length()-1; i++) {
			if (
				Character.isLowerCase( buf.charAt(i-1) ) &&
				Character.isUpperCase( buf.charAt(i) ) &&
				Character.isLowerCase( buf.charAt(i+1) )
			) {
				buf.insert(i++, '_');
			}
		}
		return buf.toString().toLowerCase(Locale.ROOT);
	}
```

##### **JHipster Version(s)**

3.8.0

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.8.0",
    "baseName": "server",
    "packageName": "org.duff",
    "packageFolder": "org/duff",
    "serverPort": "8080",
    "hibernateCache": "ehcache",
    "websocket": false,
    "databaseType": "sql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "b74f0cc9342da628bc7064b2af4d6f48786c4f31",
    "useSass": false,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ]
  }
}
```

File.json
```
{
  "relationships": [
    {
      "relationshipType": "many-to-one",
      "relationshipName": "soCRevision",
      "otherEntityName": "revision",
      "otherEntityField": "id"
    }
  ],
  "fields": [
    {
      "fieldName": "description",
      "fieldType": "String"
    },
    {
      "fieldName": "version",
      "fieldType": "String"
    }
  ],
  "changelogDate": "20161004125225",
  "dto": "no",
  "pagination": "no",
  "service": "no",
  "entityTableName": "file"
}
```
Revision.json
```
{
  "relationships": [
    {
      "relationshipType": "one-to-many",
      "relationshipName": "soCRepository",
      "otherEntityName": "file",
      "otherEntityRelationshipName": "soCRevision"
    }
  ],
  "fields": [
   {
      "fieldName": "description",
      "fieldType": "String"
    }
  ],
  "changelogDate": "20161004125239",
  "dto": "no",
  "pagination": "no",
  "service": "no",
  "entityTableName": "revision"
}
```

Windows, Chrome and Firfox

Dufgui:
fix #4320

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1161,7 +1161,26 @@ Generator.prototype.getTableName = function (value) {
  * @param {string} value - table column name string
  */
 Generator.prototype.getColumnName = function (value) {
-    return _.snakeCase(value).toLowerCase();
+    value = value.replace('.', '_');
+    var res = '';
+    if(value && value.length > 0)
+        res = value[0];
+    for (var i = 1, len = value.length-1; i < len; i++) {
+        if(
+            value[i-1] !== value[i-1].toUpperCase() &&
+            value[i] !== value[i].toLowerCase() &&
+            value[i+1] !== value[i+1].toUpperCase()
+        ) {
+            res += '_' + value[i];
+        } else {
+            res += value[i];
+        }
+    }
+
+    if(value && value.length > 0)
+        res += value[value.length -1];
+
+    return res.toLowerCase();
 };
 
 /**
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1152,7 +1152,7 @@ Generator.prototype.gitExec = function (args, options, callback) {
  * @param {string} value - table name string
  */
 Generator.prototype.getTableName = function (value) {
-    return _.snakeCase(value).toLowerCase();
+    return this.hibernateSnakeCase(value);
 };
 
 /**
@@ -1161,26 +1161,34 @@ Generator.prototype.getTableName = function (value) {
  * @param {string} value - table column name string
  */
 Generator.prototype.getColumnName = function (value) {
-    value = value.replace('.', '_');
-    var res = '';
-    if(value && value.length > 0)
+    return this.hibernateSnakeCase(value);
+};
+
+/**
+ * get hibernate SnakeCase in JHipster preferred style.
+ *
+ * @param {string} value - table column name or table name string
+ * @see org.springframework.boot.orm.jpa.hibernate.SpringNamingStrategy
+ */
+Generator.prototype.hibernateSnakeCase = function (value) {
+    let res = '';
+    if (value) {
+        value = value.replace('.', '_');
         res = value[0];
-    for (var i = 1, len = value.length-1; i < len; i++) {
-        if(
-            value[i-1] !== value[i-1].toUpperCase() &&
-            value[i] !== value[i].toLowerCase() &&
-            value[i+1] !== value[i+1].toUpperCase()
-        ) {
-            res += '_' + value[i];
-        } else {
-            res += value[i];
+        for (var i = 1, len = value.length - 1; i < len; i++) {
+            if (value[i-1] !== value[i-1].toUpperCase() &&
+                value[i] !== value[i].toLowerCase() &&
+                value[i+1] !== value[i+1].toUpperCase()
+            ) {
+                res += '_' + value[i];
+            } else {
+                res += value[i];
+            }
         }
-    }
-
-    if(value && value.length > 0)
         res += value[value.length -1];
-
-    return res.toLowerCase();
+        res = res.toLowerCase();
+    }
+    return res;
 };
 
 /**


generators/generator-base.js
```diff
@@ -1161,7 +1161,26 @@ Generator.prototype.getTableName = function (value) {
  * @param {string} value - table column name string
  */
 Generator.prototype.getColumnName = function (value) {
-    return _.snakeCase(value).toLowerCase();
+    value = value.replace('.', '_');
```
deepu105: I dont think it will ever have dot in the passed value

generators/generator-base.js
```diff
@@ -1161,7 +1161,26 @@ Generator.prototype.getTableName = function (value) {
...
+    var res = '';
+    if(value && value.length > 0)
+        res = value[0];
```
deepu105: just return and stop flow here

generators/generator-base.js
```diff
@@ -1161,7 +1161,26 @@ Generator.prototype.getTableName = function (value) {
...
+            value[i] !== value[i].toLowerCase() &&
+            value[i+1] !== value[i+1].toUpperCase()
+        ) {
+            res += '_' + value[i];
+        } else {
+            res += value[i];
+        }
+    }
+
+    if(value && value.length > 0)
```
deepu105: redundant

jdubois:
Oh, thanks for the review @deepu105 !!

deepu105:
I had a closer look and indeed its annoying why cant the spring people do a normal snake case 😠 
Anyways to make the solution global I would prefer the below
Add the below method to `generator-base.js` after line 1492 (private methods)
```
Generator.prototype.hibernateSnakeCase = function (value) {
    let res = '';
    if (value) {
        value = value.replace('.', '_');
        res = value[0];
        for (var i = 1, len = value.length - 1; i < len; i++) {
            if (value[i-1] !== value[i-1].toUpperCase() &&
                value[i] !== value[i].toLowerCase() &&
                value[i+1] !== value[i+1].toUpperCase()
            ) {
                res += '_' + value[i];
            } else {
                res += value[i];
            }
        }
        res += value[value.length -1];
        res = res.toLowerCase();
    }
    return res;
};
```

now change the `getColumnName` and `getTableName` methods in same file to call `this.hibernateSnakeCase(value);` instead of `_.snakeCase(value).toLowerCase();`

Dufgui:
Ok i have checked, the spring naming strategy is the same for column and table name. So ok

Dufgui:
So i have tested it before push and it's working.

sendilkumarn:
@deepu105 no other api level method available to handle this? 

Is this even a standard?

deepu105:
@sendilkumarn i think it comes from either JPA or Hibernate so we have to follow

MathieuAA:
@deepu105 @Dufgui: With this implementation, when passing `EntityA` it returns `entitya`. By the way, where are the tests?

deepu105:
Tests should in test-generator-base.js

Thanks & regards,
Deepu

PR #4328, opened 2016-10-13
Add a generator to export entities to JDL

cbornet:


cbornet:
Writing the tests now

generators/export-jdl/index.js
```diff
@@ -0,0 +1,44 @@
...
+        insight: function () {
+            var insight = this.insight();
+            insight.trackWithEvent('generator', 'export-jdl');
+        },
+
+        parseJson: function () {
+            this.log('Parsing entities from .jhipster dir...');
+            try {
+                let entities = {};
+                this.getExistingEntities().forEach( entity => entities[entity.name] = entity.definition );
```
deepu105: I think you can use the map method for this

generators/export-jdl/index.js
```diff
@@ -0,0 +1,44 @@
...
+            } catch (e) {
+                this.log(e.message || e);
+                this.error(`\nError while parsing entities to JDL\n`);
+            }
+        }
+    },
+
+    writing: function () {
+        let content = '// JDL definition for application \'' + this.baseName + '\' generated with command \'yo jhipster:export-jdl\'\n\n' + this.jdl.toString();
+        this.fs.write(this.jdlFile, content);
```
deepu105: you could add a log message to let the user know that the process finished successfully?

generators/export-jdl/index.js
cbornet: I tried but map returns an array, not an Object

generators/export-jdl/index.js
cbornet: done

PR #4335, opened 2016-10-15
ng2: (i18n) related compile time error when i18n option is not selected

allahbaksh:
During the project setup if i18n is not selected then there are compile time error.

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/_app.state.ts
+++ b/generators/client-2/templates/src/main/webapp/app/_app.state.ts
@@ -1,5 +1,8 @@
 import { NavbarComponent } from './layouts';
-import { AuthService, <%=jhiPrefixCapitalized%>LanguageService } from './shared';
+import { AuthService } from './shared';
+<%_ if (enableTranslation) { _%>
+import { <%=jhiPrefixCapitalized%>LanguageService } from "./shared";
+<%_ } _%>
 
 export const appState = {
     name: 'app',
@@ -12,11 +15,12 @@ export const appState = {
             token: 'authorize',
             deps: [AuthService],
             resolveFn: (auth) => auth.authorize()
-        },
+        }<%_ if (enableTranslation) { _%>,
         {
             token: 'translate',
             deps: [<%=jhiPrefixCapitalized%>LanguageService],
             resolveFn: (languageService) => languageService.setLocations(['login'])
         }
+        <%_ } _%>
     ]
 };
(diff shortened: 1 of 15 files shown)


generators/client-2/templates/src/main/webapp/app/_app.state.ts
```diff
@@ -1,5 +1,8 @@
 import { NavbarComponent } from './layouts';
-import { AuthService, <%=jhiPrefixCapitalized%>LanguageService } from './shared';
+import { AuthService } from './shared';
+<%_ if (enableTranslation) { _%>
```
deepu105: Can we strip this as well using regex?

allahbaksh: Will try it out.

deepu105: It gets applied during generation as the files are passed through this filter. I wouldnt want current regex changed as the function would be used for both ng1 and ng2 so i would add these as additional or conditions to the current regex to strip ng2 resolve blocks. for app.state its ok to have condition.

allahbaksh: @deepu105 What about the import statement. Where do you want to put that? Should this be part of client-2 so that it is applied to only Angular 2 or should we check if it is Angular2 code apply this regex otherwise previous one?

allahbaksh: @deepu105  
Below code has condition checking for regex to be applied. What is best place to remove the import declaration.  (this is part of Generator.prototype.copyTemplate method)
` angularVersion = this.config.get('angularVersion');`
`if (angularVersion==='angular2'){`
 `// This will remove resolve translate service. If there are more services then they should be explicitly removed`
 `       regex = /\,[\s]*resolve:[\s]\[\{[\s]*[a-z]*\:[\s]*\'translate\'\,[\s]*[a-zA-z\s:\[\:\,\(\)\=\>]*\.[a-zA-z\(\)\']*[\s]*\}\]/g;`
 `      } else {`
 `         regex = /\,[\s\n ]*(resolve)\:[\s ]*[\{][\s\n ]*[a-zA-Z]+\:(\s)*\[[ \'a-zA-Z0-9\$\,\(\)\{\}\n\.\<\%\=\-\>\;\s]*\}\][\s\n ]*\}/g;`
` }`
`jhipsterUtils.copyWebResource(source, dest, regex, 'js', _this, _opt, template);`

sendilkumarn: @deepu105 can we have this only for `html` ? 

`state.js` and `import` can we follow what we had so far.

deepu105: What do you mean by what we had so far?

jdubois:
There are also many lines which look wrong:

    resolveFn: (languageService) => languageService.setLocations(['activate'])

@flaviencathala just proposed PR #4375 that does the same thing: I'm going to close this PR, and merge PR #4375 instead

PR #4340, opened 2016-10-17
Ng2 file write refactor

deepu105:
@jhipster/developers Take a look at this. What do you think of doing this instead of writing files one by one. It saves a lot of code and makes it easy to read. Also for tests we can call the files object to test for generated files instead of maintaining that seperateley


generators/client-2/files.js
```diff
@@ -0,0 +1,651 @@
...
+
+const mkdirp = require('mkdirp');
+/* Constants use throughout */
+const constants = require('../generator-constants'),
+    MAIN_SRC_DIR = constants.CLIENT_MAIN_SRC_DIR,
+    TEST_SRC_DIR = constants.CLIENT_TEST_SRC_DIR,
+    ANGULAR_DIR = constants.ANGULAR_DIR;
+
+const files = {
+    common: [
```
deepu105: this keys are just for readability. It could have been a single array as well

jdubois:
Yes that's much easier to read! My only concern is that it's twice as long. But yes +1

generators/client-2/files.js
```diff
@@ -0,0 +1,651 @@
...
+};
+
+function writeFiles() {
+    return {
+        writeSepcial: function () {
+            this.fs.copy(this.templatePath('gulp/_handle-errors.js'), this.destinationPath('gulp/handle-errors.js')); // to avoid interpolate errors
+            mkdirp(MAIN_SRC_DIR);
+        },
+
+        writeFiles: function () {
```
deepu105: this method is generic and hence can be moved to generator-base if we are adopting this. so the write block will reduce to just a single method call

generators/client-2/files.js
```diff
@@ -0,0 +1,651 @@
...
+
+module.exports = {
+    writeFiles,
+    files
+};
+
+function writeFiles() {
+    return {
+        writeSepcial: function () {
+            this.fs.copy(this.templatePath('gulp/_handle-errors.js'), this.destinationPath('gulp/handle-errors.js')); // to avoid interpolate errors
```
deepu105: any special handlings can still be done as usual

deepu105:
Many extra lines are to have a neat object view and readability, if
required it can be further shortened/optimized. I guess I can remove aprox
around 100 lines if needed


Thanks & Regards,
Deepu

sendilkumarn:
+1 deepu

But again same comments, the file is again huge. Can we do something?

gmarziou:
Deepu, could you explain the difference between copy and copyHtml?

What about using file extensions as a convention,? `_upgrade_adapter.ts.tpl` would be a template for `_upgrade_adapter.ts`, _`upgrade_adapter.ts` without tpl extension would just be copied as-is. This is what is sued in Rails.

deepu105:
Hi Gael,

Thats a good idea let me see if we can use that. Copy and copyHtml/copyJs
differs coz the latter filters out translation related stuff from the files
when its disabled. I guess it wouldn't hurt to pass all files through them
but will have a little performance impact due to the un necessary regex
lookup.

I'm gonna try out the extension method you suggested.

Thanks & regards,
Deepu

On 18 Oct 2016 02:47, "Gaël Marziou" <notifications@github.com> wrote:

Deepu, could you explain the difference between copy and copyHtml?

What about using file extensions as a convention,? _upgrade_adapter.ts.tpl
would be a template for _upgrade_adapter.ts, _upgrade_adapter.ts without
tpl extension would just be copied as-is. This is what is sued in Rails.

—
You are receiving this because you authored the thread.
<https://github.com/jhipster/generator-jhipster/pull/4340#issuecomment-254336807>,
or mute the thread

sendilkumarn:
@deepu105 why can't we use the directory instead of files. And have an `ignore` option to make it more flexible.

deepu105:
Ya thought of that but it makes the optional files ambiguous especially to
new contributors. Also the main purpose of this was to make it more
readable so that we know what files goes for what

Thanks & regards,
Deepu

wmarques:
+1 for this, if the code is more readable for me it can be a bit longer than the previous one.

deepu105:
@jhipster/developers I have optimized this to be much smaller. so now its just around 100 lines more than previous (I'm not counting the reusable method)
Let me know if this looks good. @gmarziou I tried with extensions but that would be a lot of work to rename and stuff, so thought a simple object metadata would do.

test/test-creation.js
```diff
@@ -1405,8 +1217,12 @@ describe('JHipster client-2 generator', function () {
         it('creates expected files for default configuration for client-2 generator', function () {
             assert.noFile(expectedFiles.server);
             assert.noFile(expectedFiles.maven);
-            assert.file(expectedFiles.client_2);
-            assert.file(['gulpfile.js']);
+            assert.file(getFilesForOptions(require('../generators/client-2/files').files, {
```
deepu105: @jhipster/developers this is the best part of this change. we need not maintain list of files to check in tests anymore it can be obtained dynamically for any config required 😄

mraible:
@deepu105 I just cancelled this build by accident. I was logging to restart another one for another project and Travis CI shifted the UI really quick and I cancelled this one. I restarted it. Sorry!

deepu105:
No probs :)

Thanks & regards,
Deepu

test/test-creation.js
pascalgrimaud: interesting! I need to look how it works

deepu105:
@jhipster/developers guys are you ok with this? shall I merge it?

wmarques:
LGTM :+1:

deepu105:
Seems no body has a problem with this so I'm gonna merge it

Issue #4309, opened 2016-10-11
too many space and wrap when i use yo jhipster:import-jdl generate entity


anjia0532:
https://jhipster.github.io/jdl/#a-namecommentingjdla-commenting--javadoc support java-doc it's so cool,but something is wrong with the `yo jhipster:import-jdl`

jdl code like this
```
/**
 * blog demo
 */
entity Blog {
   /**
     * author's name
     */
   name String required
}
```

generator Blog.java like this

```java

/**
 * blog demo                                                                          //(about 70 space)
 * 
 */
@ApiModel(description = ""
    + "Class comments.@author The JHipster team.                              "
    + "")
@Entity
@Table(name = "blog")
@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
public class Blog implements Serializable {

    private static final long serialVersionUID = 1L;

    @Id
    @GeneratedValue(strategy = GenerationType.AUTO)
    private Long id;

    /**
     * deptartment's Code                                                      
     * 
     */
    @ApiModelProperty(value = ""
        + "deptartment's Code                                                 "
        + "")
    @NotNull
    @Column(name = "dept_id", nullable = false)
    private String deptId;
```

it has too many space and wrap to manual remove it

deepu105:
can you please open this in the jhipster-core project as the conversion happens there

anjia0532:
OK .sorry

cbornet:
The issue is in the main gen. See https://github.com/jhipster/jhipster-core/issues/73

jdubois:
Closing this as duplicate, then.

deepu105:
@jdubois its an issue here and not in core, the other issue is already closed

jdubois:
OK, sorry!

cbornet:
We need to fix that.

anjia0532:
@cbornet @deepu105 @jdubois 

look at [generator-base.js](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js#L1632) line 1632-1638

```js
Generator.prototype.formatAsApiModel = function (text) {
    return jhipsterUtils.wordwrap(text.replace(/\\/g, '\\\\').replace(/\"/g, '\\\"'), WORD_WRAP_WIDTH - 9, '"\n    + "', true);
};

Generator.prototype.formatAsApiModelProperty = function (text) {
[...]

PR #4341, opened 2016-10-17
fix https://github.com/jhipster/generator-jhipster/issues/4309

anjia0532:
fix [issues/4309](https://github.com/jhipster/generator-jhipster/issues/4309)

the [wordwrap](https://github.com/jhipster/generator-jhipster/blob/master/generators/util.js#L200)
```javascript
function wordwrap (text, width, seperator, keepLF) {
    var wrappedText = '';
    var rows = text.split('\n');
    for (var i = 0; i < rows.length; i++) {
        var row = rows[i];
        if (keepLF === true && i !== 0) {
            wrappedText = wrappedText + '\\n';
        }
        wrappedText = wrappedText + seperator + _.padEnd(row,width) + seperator;
    }
    return wrappedText;
}
```
because of _loadjson(in the [entity/index.js#L197](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity/index.js#L197))

```javascript
 try {
            this.fileData = this.fs.readJSON(this.fromPath);
        } catch (err) {
            this.error(chalk.red('\nThe entity configuration file could not be read!\n'));
        }
```

this.fileData.javadoc  result allways in one line,so ,wordwrap [rows](https://github.com/jhipster/generator-jhipster/blob/master/generators/util.js#L202) allways eq 1,

so `wrappedText = wrappedText + seperator + _.padEnd(row,width) + seperator;` == `wrappedText =seperator + _.padEnd(row,width) + seperator;`

and many space because of _.[padEnd](https://lodash.com/docs/4.16.4#padEnd)(row,[width](https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js#L19))
padding width-row.length space to wrappedText

i don't think need to keep wordwrap function in @ApiOperation.

 [4309/#issuecomment-254116002](https://github.com/jhipster/generator-jhipster/issues/4309/#issuecomment-254116002)

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1630,11 +1630,11 @@ Generator.prototype.formatAsFieldJavadoc = function (text) {
 };
 
 Generator.prototype.formatAsApiModel = function (text) {
-    return jhipsterUtils.wordwrap(text.replace(/\\/g, '\\\\').replace(/\"/g, '\\\"'), WORD_WRAP_WIDTH - 9, '"\n    + "', true);
+    return text;
 };
 
 Generator.prototype.formatAsApiModelProperty = function (text) {
-    return jhipsterUtils.wordwrap(text.replace(/\\/g, '\\\\').replace(/\"/g, '\\\"'), WORD_WRAP_WIDTH - 13, '"\n        + "', true);
+    return text;
 };
 
 Generator.prototype.isNumber = function (input) {


jdubois:
Indeed, but in that case we might as well remove the function.
I'll do it and reference the commit in this PR.

PR #4342, opened 2016-10-17
Add error handler for app injections

octavioamu:
Add error handler for app injections, without this any code mistake is going to fail all the app to injects the js files silently


octavioamu:
This error took me a week to discover why my new routes and code wasn't working until I found this missing code and then I could see the error was introduced in some other service (a duplicate property)

PR #4349, opened 2016-10-18
Add SB cloud connectors starter

### LINKED ISSUE #4346: [FEATURE] Add SpringBoot cloud connectors starter

Add SpringBoot cloud connectors starter
https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-cloud-connectors/1.4.1.RELEASE

It will reduce dependencies in pom.

##### **Suggest a Fix**

I can create a PR.

##### **JHipster Version(s)**

version 3.9.1

nkolosnjaji:
Add spring boot cloud connectors starter for maven and gradle

FIX #4346

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -245,9 +245,7 @@ dependencies {
     <%_ if (applicationType === 'microservice' || applicationType === 'gateway' || applicationType === 'uaa') { _%>
     compile "org.springframework.cloud:spring-cloud-starter-feign"
     <%_ } _%>
-    compile "org.springframework.cloud:spring-cloud-cloudfoundry-connector"
-    compile "org.springframework.cloud:spring-cloud-spring-service-connector"
-    compile "org.springframework.cloud:spring-cloud-localconfig-connector"
+    compile "org.springframework.boot:spring-boot-starter-cloud-connectors"
     compile ("org.springframework:spring-context-support") {
         exclude(module: 'quartz')
     }
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -710,16 +710,8 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-cloudfoundry-connector</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-spring-service-connector</artifactId>
-        </dependency>
-        <dependency>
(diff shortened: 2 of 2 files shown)


jdubois:
I see it also adds the Heroku connector, can you check if this makes a difference for our Heroku sub-generator?
Code is at https://github.com/jhipster/generator-jhipster/tree/master/generators/heroku

nkolosnjaji:
sure, I will give it a try

gmarziou:
I'm not sure but it seems this adds dependencies but to depdencyMaganement section like when importing a pom so does it mean that an app not using Heorku will get heroku jars anyway.

nkolosnjaji:
@gmarziou yes, herorku jars will be included. I think it is not a problem because jhipster includes cloudfondry jars(even if it doesn't use them)  and not everybody is going to deploy apps to CF. Anyway, it's good to include all connectors because once you build the app, you can move it to different cloud providers.

jdubois:
That's basically what we already do, but in fact I'm not happy with that... How big are those JARs? I don't want to have too much useless stuff in our generated app.
We could add them in the pom when we run the sub-generators -> that wasn't the case at the time it was coded, but we now have an API for that.

nkolosnjaji:
I give few tries with heroku generator without luck. I didn't add credit card to my herou account so it must be some weird combination with my account type when generator tries to create db

nkolosnjaji:
heroku cloud connector jar is 7Kb.

https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-heroku-connector/1.2.3.RELEASE

jdubois:
OK, then it would be cleaner to add those JARs with the sub-generator (same thing for Cloud Foundry), but that's so small that we don't really care... Let's do this.

jdubois:
I'm restarting your build because it failed, but this is probably a Travis issue... So I probably won't be able to merge it now, but +1 for me (as long as the build passes!)

pascalgrimaud:
I think it's a Travis issue on the new build uaa. I need to investigate why it failed so often...

jdubois:
I don't understand what happened: GitHub sent an error when I clicked on the merge button, and now the PR isn't merged but the code is in the master... So this looks OK and I'm closing this, but very weird...

pascalgrimaud:
@nkolosnjaji : maybe it's because you didn't configure your local git configuration with appropriate name + email ??
We can see it in: https://github.com/jhipster/generator-jhipster/pull/4349/commits
There is no link to your github account

nkolosnjaji:
oh yes :(

I was using github with my ssh key... Now I added that email to my github acc. Do I need to do anything more ?

pascalgrimaud:
I dont know, maybe try your config in one of your own repositories to confirm

PR #4359, opened 2016-10-20
fix test for relation one-to-many

kpallardy:
The entity generator create the following tests lines in createEntity for one to many relation :

`A.setBs(B);`

But Bs is a Set<B> because it's a one to many relation so we expect to have the same line as many to many relation :
`A.getBs().add(B);`

### DIFF
--- a/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
+++ b/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
@@ -260,7 +260,7 @@ public class <%= entityClass %>ResourceIntTest <% if (databaseType == 'cassandra
         <%= otherEntityNameCapitalized %> <%= relationshipFieldName %> = <%= otherEntityNameCapitalized %>ResourceIntTest.createEntity(em);
         em.persist(<%= relationshipFieldName %>);
         em.flush();
-            <%_ if (relationshipType == 'many-to-many') { _%>
+            <%_ if (relationshipType == 'many-to-many' || relationshipType == 'one-to-many') { _%>
         <%= entityInstance %>.get<%= relationshipNameCapitalizedPlural %>().add(<%= relationshipFieldName %>);
             <%_ } else { _%>
         <%= entityInstance %>.set<%= relationshipNameCapitalized %>(<%= relationshipFieldName %>);


jdubois:
This only occurs for required one-to-many relationships, which is not something we currently support.
How did you get that issue?
Could you follow the guidelines, and open a ticket with your use case, so I can test and see your problem? I'm pretty sure you found something here, as nobody would do this without a good reason, but I'd like to understand and test.

Dufgui:
We create a model with uml designer. So we doesn't have the hand of the json files (in .jhipster) generated.
And the many to one relation is view one to many on the other side. In fact we don"t need this.
It's generated by this line of  code:
https://github.com/jhipster/jhipster-uml/blob/master/lib/editors/umldesigner_parser.js#L255

jdubois:
OK! With the generator this isn't possible, but we don't have the same strict rules with JHipster UML (ping @MathieuAA ).
So for me this is a bug with JHipster UML.

MathieuAA:
@Dufgui when you post the issue, don't forget to correctly give me your UML file. Thanks. @jdubois I think you can close this issue.

Dufgui:
So i re create the issue 
https://github.com/jhipster/jhipster-uml/issues/225

MathieuAA:
@jdubois Are you sure that's my fault? I don't think there's a problem with anything here...

jdubois:
From my understanding:

- We have a "required" + "one-to-many" relationship
- This should not happen when you use the "yo" command line, as we have a validator for that (and that doesn't seem very logical anyway - what do we mean here?? that the collection shouldn't be empty? this doesn't make much sense)
- But @Dufgui created the .json files using JHipster UML: as you don't have the same validation (which is normal, you can't have all the rules, it's just too complex), JHipster UML generates the JSON files with those 2 options
- Then, when the generator runs, it generates some wrong tests, as we have those 2 options, that should not be there together

For me, a better solution would be to test this in JHipster Core -> the "required" relationship should be ignored when the "one-to-many" option is selected. That would ensure we are always correct, whatever method is used to generate the .json files.

MathieuAA:
Indeed, I'll make it forbidden in JUML and in the core project.

cbornet:
cc @colameo @jbadeau for the IDE grammar check

colameo:
I can fix this by adding a validation. But personally I think it's not unusual to have a required one-to-many relation in a data model. Let's assume we have a relation like Person ---* Address. A person must have at least 1 address (but can also have many) in the system. That means it is not allowed to store a person without an address (which may makes sense, right?). For such use cases I would like to use a "required one-to-many" relation.

cbornet:
The thing is you can't enforce that in the model. It's an empty-initialized collection so @NotNull can't be used

MathieuAA:
Yes, it's a valid use case. Anyway, one could always have a field named `address` and a set of many additional addresses...

MathieuAA:
@cbornet yeah, but we're talking about a "minlength", targeting a collection. That's the weird point. It could easily be done by hand without any need for a "minlength" (if it exists).

jdubois:
@colameo thanks for the use-case!

In fact, when we mean "Required", we are currently using @NotNull from Bean Validation. So for a "one-to-many" it doesn't mean anything, as we have a Collection anyway.

Indeed your use-case is correct, but then we would need to test the size of the collection: that's perfectly valid, and we should do this, but then that would be a new feature.

cbornet:
Then it's more a "cardinality" support

MathieuAA:
That could very easily be done in JUML, we wouldn't have anything to do at all =)

MathieuAA:
Cardinality is too precise a thing to add in my opinion... I'd rather let the user make the necessary change in the java code than pollute the JDL as different implementation could be achieved (IllegalState/IllegalArgument throwing, or else...).

Dufgui:
At short term I propose to remove required rules  in JUML in case of one-to-many (on the owner side). It's a simple quick fix. Are you ok with this?  After i let you decide, what is the final goal to implement.

MathieuAA:
@Dufgui I'll do it tonight so as to close the issue here and in JUML.

Dufgui:
ok tks

PR #4365, opened 2016-10-20
fix navigation key 

vivekmore:
- fix the navigation key in the jhipster-needle code
- center the footer to make it look nice


generators/client/templates/src/main/webapp/_index.html
```diff
@@ -76,7 +76,7 @@
         </div>
 
         <div class="footer" ng-cloak>
-            <p data-translate="footer">This is your footer</p>
+            <p class="text-center" data-translate="footer">This is your footer</p>
```
deepu105: Can you remove this as we would like users to style this as they like

vivekmore: I understand that. 
My rationale behind adding it was based on other places in the application jhipster is providing some sort of minimalistic styling hence I thought it would be nice have the footer centered (which is the case in most of the websites)

Nevertheless, I'll remove this.

vivekmore:
I have created a new PR. 
Closing this one.

PR #4366, opened 2016-10-21
fix sass imports not updating the main.css file

octavioamu:
There is an old problem ( http://stackoverflow.com/questions/36411296/gulp-not-refresh-on-changes-to-imported-scss-files )  when using sass if you create some imports inside main.scss and go to one of this imported files and made changes you had to go back to main.scss and save it to the imports changes take effects. 
This problem is related to the "changed" plugin don't letting gulp-sass work like it should.


gmarziou:
I can't test right  now but from code review it looks good to me

PR #4369, opened 2016-10-22
#4372 [FEATURE] Remove entity

Dufgui:
add remove option to entity sub generator

### DIFF
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -1,7 +1,6 @@
 'use strict';
 
-const _ = require('lodash'),
-  chalk = require('chalk');
+const _ = require('lodash');
 
 /* Constants use throughout */
 const constants = require('../generator-constants'),
@@ -65,7 +64,7 @@ function writeFiles(){
                     if (!this.skipClient && this.enableTranslation) {
                         var languages = this.languages || this.getAllInstalledLanguages();
                         languages.forEach(function (language) {
-                            this.fs.delete(CLIENT_MAIN_SRC_DIR + 'i18n/' + language + '/' + _.lowerFirst(fieldType) + '.json');
+                            this.removeEnumI18n(language, _.lowerFirst(fieldType));
                         }, this);
                     }
 
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -13,8 +13,13 @@ const constants = require('../generator-constants'),
     SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
 
 module.exports = {
-    writeFiles
+    writeFiles,
+    removeFiles
 };
+
+function removeFiles(){
+
+}
 
 function writeFiles() {
     return {
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -35,10 +35,18 @@ function writeFiles(){
             if (this.skipServer) return;
 
             if (this.databaseType === 'sql') {
+                this.fs.delete(SERVER_MAIN_RES_DIR + 'config/liquibase/changelog/' + this.changelogDate + '_added_entity_' + this.entityClass + '.xml');
+
+                if (this.fieldsContainOwnerManyToMany || this.fieldsContainOwnerOneToOne || this.fieldsContainManyToOne) {
+                    this.fs.delete(SERVER_MAIN_RES_DIR + 'config/liquibase/changelog/' + this.changelogDate + '_added_entity_constraints_' + this.entityClass + '.xml');
+                    this.removeConstraintsChangelogFromLiquibase(this.changelogDate + '_added_entity_constraints_' + this.entityClass);
+                }
+                this.removeChangelogFromLiquibase(this.changelogDate + '_added_entity_' + this.entityClass);
                 this.warning('Please update manually the liquibase to deal with data before drop the table');
             }
             if (this.databaseType === 'cassandra') {
-                this.warning('Please update manually the cql script to deal with data before drop the column family');
+                this.fs.delete(SERVER_MAIN_RES_DIR + 'config/cql/changelog/' + this.changelogDate + '_added_entity_' + this.entityClass + '.cql');
+                this.warning('Please do a manual update of the cql script to choose what to do with data before dropping the column family');
             }
         },
 
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const _ = require('lodash');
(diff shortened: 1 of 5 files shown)


PierreBesson:
:+1: If sufficiently tested I would be in favor of this feature. I can't see any drawbacks to it and in the case it breaks the project people will be saved by git.

generators/entity/files.js
```diff
@@ -16,9 +17,142 @@ module.exports = {
...
+            this.fs.delete(this.destinationPath(this.jhipsterConfigDirectory + '/' + this.entityNameCapitalized + '.json'));
+        },
+
+        deleteDbFiles: function () {
+            if(!this.remove) return;
+
+            if (this.skipServer) return;
+
+            if (this.databaseType === 'sql') {
+                this.warning('Please update manually the liquibase to deal with data before drop the table');
```
deepu105: Why is this required. I think we can delete liquibase changelog files and remove the entry from master xml as well and then print a warning to ask user to update the DB manually or by running liquibase:diff goal after removing the entity

generators/entity/files.js
```diff
@@ -16,9 +17,142 @@ module.exports = {
...
+            }
+            if (this.databaseType === 'cassandra') {
+                this.warning('Please update manually the cql script to deal with data before drop the column family');
```
deepu105: Messages needs to be corrected for grammer

generators/entity/index.js
```diff
@@ -69,6 +69,14 @@ module.exports = EntityGenerator.extend({
             type: Boolean,
             defaults: false
         });
+
+        // This method adds support for a `--[no-]regenerate` flag
```
deepu105: Correct the comment

generators/generator-base.js
```diff
@@ -163,12 +190,32 @@ Generator.prototype.addAdminElementTranslationKey = function (key, value, langua
 Generator.prototype.addEntityTranslationKey = function (key, value, language) {
     var fullPath = CLIENT_MAIN_SRC_DIR + 'i18n/' + language + '/global.json';
     try {
-        jhipsterUtils.rewriteFile({
+        jhipsterUtils.replaceContent({
```
deepu105: This method shouldn't be updated

pascalgrimaud:
I agree it's a good idea and can help people

But my personnal opinion:
- I'm afraid it will lead into an increasing number of issues, if not well tested, and there are so much use cases...
- as the sub generator entity is already complex, I'd rather prefer it in another sub generator, but it's just me

Dufgui:
@pascalgrimaud in fact, yes i can move my code to a new generator but to be maintained and coherent with the entity generator, i think it's a good idea to have all my code in the same file where the entity files is created.

Dufgui:
So I add a unit test and fix a little bug on remove the entity json file

jdubois:
Per our guidelines, I'd like first to have a ticket with a discussion, before implementing such a feature.

Then, I can already share my thoughts:

Pros:
- A lot of people are requesting this

Cons:
- I don't think it's useful, you should better use Git to rollback your changes
- It adds a lot more code in the generator, and that's going to be annoying to maintain (now we must always write code to "undo" what we do in the generator)

So I'm against such a feature, but I'm not the only one deciding here, and I know a lot of people want this, so we must discuss this.

cbornet:
@jdubois I'm not sure using git to revert an existing entity is always possible: if you create Foo with JH 3.0 then upgrade it later with JH 3.10 then file creation can be scattered across multiple commits some of which will also probably be shared with other entities (the upgrade one). So you cannot just find the commit and revert it.
Now I understand the feeling about maintenance. My impression is that it is easy to maintain the list of files to delete and harder to maintain the files to modify (by regexp matching). So we could just provide the file remover and ask the user to do a `yo jhipster --with-entities` to clean the modified files.
Another idea would be to use the same idea as for `jhipster:upgrade` : in another branch, generate the app with the entity, commit, generate the app without, commit, then merge on the main branch.

Dufgui:
it take 2 hour to doing it and 3h with the unit test, and some dev of my team asked it, so i propose it. Sorry to not create a issue for it.

For your first point:
I am agree with @cbornet , it 's complicated to use git for this because all our entity is created one times by jhipster-uml so to rollback it, we have one commit for all of them....

For your second point:
If you want i can split it in another sub generator, it's quite simple. I do it in the same because I think it's simplier for the evolution and keep it coherent with the entity generator.

jdubois:
OK, @cbornet opinion as very important here, as he's using this for real (which is not my case!), so he knows better than me.

Dufgui:
So I create the issue for the discussion.

pascalgrimaud:
@Dufgui : FYI, you have changed your local git configuration, so in the commit logs https://github.com/jhipster/generator-jhipster/pull/4369/commits Github doesn't know you

deepu105:
If we are doing this its better to be here than another sub gen

Thanks & regards,
Deepu

deepu105:
I agree with the maintenance issue Julien highlighted. We already have a
solution to reduce it. Since im working on refactoring the write blocks to
use metadata rather than instructions it would be very easy to write or
delete files using a simple function so lot less code to maintain.
But then I agree that not many would use this.

Thanks & regards,
Deepu

deepu105:
Yes you are right. This is a tough thing to get right. Especially
relationships as it cannot be easily removed. Then we can only help so
much, its better than nothing

Thanks & regards,
Deepu

Dufgui:
So i close this one in favour of #4403

Chris533:
In version 4+, should be mark that how to remove entity even by manual.

Chris533:
Remove codes from files:
src\main\java\.\config\CacheConfiguration.java
src\main\resources\config\liquibase\master.xml
src\main\webapp\app\entities\entity.module.ts
src\main\webapp\app\layouts\navbar\navbar.component.html
src\main\webapp\i18n\en\global.json

Remove files from folders:
src\main\java\.\domain\*.java
src\main\java\.\repository\*Repository.java
src\main\java\.\web\rest\*Resource.java
src\main\webapp\app\entities\*
src\main\webapp\i18n\.\*.json
src\test\gatling\user-files\simulations\*GatlingTest.scala
src\test\java\jshop\bunlu\com\web\rest\*ResourceIntTest.java
src\test\javascript\e2e\entities\*.spec.ts
src\test\javascript\spec\app\entities\*

Otherwise, remove table from database, remove relations from where you used.

gmarziou:
Or use git as advised previously and generate only on branches.

PR #4375, opened 2016-10-24
exceptions handled if no internationalization support

flaviencathala:


jdubois:
Oh OK, sorry @deepu105

jdubois:
@deepu105 I haven't revert it yet, could you have a look at the changes? This is just adding conditions for i18n: this is not doing anything with the regex, and without this the i18n option fails

deepu105:
Ok ill revert it later. I know it fails but i wanted to do this similar to
what we do for ng1 so that we have less conditions to add and maintain.

Thanks & regards,
Deepu

deepu105:
No problem. I am travelling currently thats why couldn't close these. I
would have time this weekend to close the pending ng2 items.

Thanks & regards,
Deepu

PR #4377, opened 2016-10-24
add jhiPrefixCapitalized to module variables

### LINKED ISSUE #4376: Add jhiPrefixCapitalized to the list of variables available when creating a module
Since `<%=jhiPrefixCapitalized%>` is used in templates, I figured I could use the following in my module to expose this variable:

```js
this.jhiPrefixCapitalized = jhipsterVar.jhiPrefixCapitalized;
```

However, this doesn't work and I have to do the following instead:

```js
this.jhiPrefixCapitalized = _.upperFirst(this.jhiPrefix);
```

ruddell:
Fix #4376

### DIFF
--- a/generators/modules/index.js
+++ b/generators/modules/index.js
@@ -1,6 +1,7 @@
 'use strict';
 var util = require('util'),
     generators = require('yeoman-generator'),
+    _ = require('lodash'),
     chalk = require('chalk'),
     scriptBase = require('../generator-base');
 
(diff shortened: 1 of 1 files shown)


mraible:
LGTM! 👍

pascalgrimaud:
@mraible : if you decide to use this new variable, you have to wait the next release of JHipster, then you can release your module. Otherwise, people who install your module won't be able to use it

PR #4368, opened 2016-10-21
Add Spring CSRF default implementation

nkolosnjaji:

PR #4379, opened 2016-10-24
Add Spring CSRF default implementation [ng2]

nkolosnjaji:
Hi,

This [pull ](https://github.com/jhipster/generator-jhipster/pull/4368) enables default CSRF for spring and angular when session auth. I'm managed to get it work for angular, but bad news is that I couldn't make it to work with websockets (it's working with ng1 and will be enabled from v3.10). For some reason stomp client is ignoring any headers provided in tracker.service.ts, so I'm stucked with it.

Any help/suggestion?


deepu105:
Ill merge this once the PR to master is accepted

PR #4389, opened 2016-10-26
translation in session

sendilkumarn:
Session imports translation even when it is not needed.


deepu105:
@sendilkumarn I will fix this this weekend using the regex I said earleir, I will be reverting the commit from flavien as well

PR #4391, opened 2016-10-26
Revert "exceptions handled if no internationalization support"

deepu105:
Reverts jhipster/generator-jhipster#4375

I will be merging this once I have the regex solution ready


jdubois:
Yes ok

PR #4393, opened 2016-10-27
normalize gitignore

### LINKED ISSUE #4387: [FEATURE] gitignore could have folders and not files

When you make a `git clean -df` some folders like node folders are removed even if they are in _.gitignore_ file

Common development of Jhipster tasks require to use git to control changes and errors because some features are difficult to remove or change without impact. So automated discard of untracked files, results in remove some necessary files accidentally.

##### **Reproduce the error**

- Make a new entity or modify one.
- (new files and folder are created)
- run **gulp install** 
- (new files and folder are created)
- **git status** shows modified files and untracked files
- **git checkout .** to back to previous state.
- (untracked files remain)
- **git clean -df**
- git  removes these files and folders and others also, like node_modules

##### **Suggest a Fix**

when you write some like that: `node_modules/**` in _.gitignore_ file git does not ignore the folder, only the files, and therefore when you execute `git clean -df` git will remove the folders because they are not ignored

So because it is a auto generated file, the solution could be to write only folders and not files:

Change `folder_name/**` by `folder_name`

##### **JHipster Version(s)**

- Jhipster: 3.9.1
- git version 1.9.1
- node: v4.6.1
- npm: 3.10.9
- bower: 1.7.9
- gulp:[11:33:27] CLI version 1.2.2
- yeoman: 1.8.5

O.S:
NAME="Ubuntu"
VERSION="14.04.4 LTS, Trusty Tahr"

dpalomar:
change gitignore path expressions for folders only,
according to https://git-scm.com/docs/gitignore and
https://git-scm.com/docs/git-check-ignore

Fix #4387

### DIFF
--- a/generators/server/templates/gitignore
+++ b/generators/server/templates/gitignore
@@ -7,14 +7,14 @@
 ######################
 # Node
 ######################
-/node/**
-/node_tmp/**
-/node_modules/**
+node/
+node_tmp/
+node_modules/
 
 ######################
 # SASS
 ######################
-.sass-cache/**
+.sass-cache/
 
 ######################
 # Eclipse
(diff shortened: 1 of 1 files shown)


jdubois:
LGTM! Just in time for the 3.10.0 release :-)

Issue #4395, opened 2016-10-28
Generate a custom Docker-Compose configuration without uaa


anjia0532:
I create  uaa application and gateway applicaton

```bash
cd uaa
./mvnw package -Pprod docker:build
cd ../gateway
./mvnw package -Pprod docker:build
cd ../
mkdir docker-compose
cd docker-compose
yo jhipster:docker-compose
```
the output 

```bash
$ yo jhipster:docker-compose
Welcome to the JHipster Docker Compose Sub-Generator 
Files will be generated in folder:  /xxx/xxx/xx/docker-compose ### my directory

Found .yo-rc.json config file...
? Which *type* of application would you like to deploy? Microservice application


? Enter the root directory where your gateway(s) and microservices are located .
./
1 applications found at /xxx/xxx/xx  ### my directory

? Which applications do you want to include in your configuration? (Press <space
> to select)
❯◉ gateway ###only gateway 

###continue

? Do you want the JHipster Console (based on ELK) to monitor your applications? 
Yes
JHipster registry detected as the service discovery and configuration provider used by your apps
? Enter the admin password used to secure the JHipster Registry admin

Checking Docker images in applications' directories...
   create docker-compose.yml
   create jhipster-registry.yml
   create central-server-config/application.yml
   create jhipster-console.yml
   create log-conf/logstash.conf
   create log-data/.gitignore

Docker Compose configuration successfully generated!
You can launch all your infrastructure by running : docker-compose up -d

```
`docker-compose.yml`

```yml
version: '2'
services:
    gateway-app:
        image: gateway
        external_links:
            - gateway-mysql:mysql
            - jhipster-registry:registry
        environment:
            - SPRING_PROFILES_ACTIVE=prod,swagger
            - SPRING_CLOUD_CONFIG_URI=http://admin:admin@registry:8761/config
            - SPRING_DATASOURCE_URL=jdbc:mysql://mysql:3306/gateway?useUnicode=true&characterEncoding=utf8&useSSL=false
            - SPRING_DATA_CASSANDRA_CONTACTPOINTS=cassandra
            - JHIPSTER_SLEEP=30
        ports:
            - 8080:8080
    gateway-mysql:
        container_name: gateway-mysql
        image: mysql:5.7.13
        environment:
            - MYSQL_USER=root
            - MYSQL_ALLOW_EMPTY_PASSWORD=yes
            - MYSQL_DATABASE=gateway
[...]

PR #4401, opened 2016-10-28
docker-compose subgenerator: uaa support

pascalgrimaud:
related to https://github.com/jhipster/generator-jhipster/issues/4395

### DIFF
--- a/generators/docker-compose/prompts.js
+++ b/generators/docker-compose/prompts.js
@@ -274,7 +274,7 @@ function askForAdminPassword() {
     }.bind(this));
 }
 
-function getAppFolders(input, applicationType) {
+function getAppFolders(input, composeApplicationType) {
     var files = shelljs.ls('-l', this.destinationPath(input));
     var appsFolders = [];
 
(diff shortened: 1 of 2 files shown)


generators/docker-compose/prompts.js
```diff
@@ -287,7 +287,8 @@ function getAppFolders(input, applicationType) {
                     if ((fileData['generator-jhipster'].baseName !== undefined)
                         && ((applicationType === undefined)
                             || (applicationType === fileData['generator-jhipster'].applicationType)
-                            || ((applicationType === 'microservice') && ('gateway' === fileData['generator-jhipster'].applicationType)))) {
+                            || ((applicationType === 'microservice') && ('gateway' === fileData['generator-jhipster'].applicationType))
```
xetys: shouldn't it be `applicationType === 'uaa'`?

pascalgrimaud: Oh I see, it's a little bit confusing. This `applicationType` is a parameter in `function getAppFolders(input, applicationType) {`, and his value can be: monolith or microservice, accordingly to the first question: monolith or microservice

pascalgrimaud:
When using 30sec or 60sec, when the gateway start, sometimes, it failed because it seems like the uaa is not completely registered. With 80sec, it always works for me.
I don't know if it's my computer or the mecanism in this architecture, but I prefer letting time and be sure the uaa is up, so that everything works fine. Every users can also reduct this time

generators/docker-compose/prompts.js
xetys: it can be 'uaa', too. This actually is why the generator seems not to generate the yaml

pascalgrimaud: so for me, it's correct, but don't hesitate to test it!

PierreBesson:
Although those sleep delays on initialization are hackish we don't have another way because of the limitations of docker-compose.

However for Kubernetes there is a feature currently in beta called init containers that might solve our problem :
http://kubernetes.io/docs/user-guide/production-pods/#handling-initialization

generators/docker-compose/prompts.js
pascalgrimaud: Using console.log to display variables:

```
╭─pgrimaud@pgrxps  ~/test/03-eureka-uaa/00-compose  
╰─$ jhcompose
Welcome to the JHipster Docker Compose Sub-Generator 
Files will be generated in folder: /home/pgrimaud/test/03-eureka-uaa/00-compose

Found .yo-rc.json config file...
? Which *type* of application would you like to deploy? Microservice application
? Enter the root directory where your gateway(s) and microservices are located (../) applicationType=microservice and fileData['generator-jhipster'].applicationType=uaa
applicationType=microservice and fileData['generator-jhipster'].applicationType=gateway
? Enter the root directory where your gateway(s) and microservices are located ../
applicationType=microservice and fileData['generator-jhipster'].applicationType=uaa
applicationType=microservice and fileData['generator-jhipster'].applicationType=gateway
2 applications found at /home/pgrimaud/test/03-eureka-uaa/

? Which applications do you want to include in your configuration? (Press <space> to select, <a> to toggle all, <i> to inverse selection)
❯◉ 01-uaa
 ◉ 02-jhgateuaa
```

pascalgrimaud: @xetys : I rename the variable, to avoid confusing, is it ok for u?

PR #4403, opened 2016-10-28
Add entity removal subgen

### LINKED ISSUE #4372: [FEATURE] Remove entity

Remove all files of generated for one entity by the entities sub generator

Many people ask to remove entity.
##### **Suggest a Fix**

add an option to the entity generator or create a dedicated sub generator
##### **JHipster Version(s)**

All

cbornet:
Git based entity removal subgen for discussion

TODO:
- [ ] Lots of tests
- [ ] Refactor upgrade subgen to use common methods
- [ ] Move code to entity subgen under "[beta] Remove entity (needs git)"

Fix #4372

### DIFF
--- a/generators/remove-entity/USAGE
+++ b/generators/remove-entity/USAGE
@@ -1,0 +1,6 @@
+Description:
+    Removes an entity
+
+Example:
+
+    yo jhipster:remove-entity Foo
--- a/generators/remove-entity/index.js
+++ b/generators/remove-entity/index.js
@@ -1,0 +1,151 @@
+'use strict';
+var util = require('util'),
+    generators = require('yeoman-generator'),
+    chalk = require('chalk'),
+    scriptBase = require('../generator-base'),
+    jhutil = require('../util'),
+    fs = require('fs'),
+    shelljs = require('shelljs');
+
+var RemoveEntityGenerator = generators.Base.extend({});
+
+util.inherits(RemoveEntityGenerator, scriptBase);
+
+/* Constants used throughout */
+const REMOVE_ENTITY_BRANCH = 'jhipster_remove_entity';
+
+module.exports = RemoveEntityGenerator.extend({
+    constructor: function () {
+        generators.Base.apply(this, arguments);
+        this.force = this.options['force'];
+        this.argument('entityName', { type: String, required: true });
+    },
+
+    initializing: {
+        displayLogo: function () {
+            this.log(chalk.green('Welcome to the JHipster Remove entity Sub-Generator '));
+            this.log(chalk.green('This will remove the entity ' + this.entityName));
+        },
+
+        getCurrentJHVersion: function () {
+            this.currentVersion = this.config.get('jhipsterVersion');
+        }
+    },
+
+    configuring: {
+        assertGitPresent: function() {
+            var done = this.async();
+            this.isGitInstalled(function (code) {
+                if (code !== 0) this.error('Exiting the process.');
+                done();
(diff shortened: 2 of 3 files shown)


deepu105:
@cbornet as said in the other thread Im against having this as a sperate sub gen. It should be just another option when an entity is regenerated.
If you are worried about adding more code there you can have this as a seperate file and call it from the entity sub gen index.js but it doesn't make sense to have this as a seperate sub gen when create and edit is done in entity sub gen. So im strongly against this as a seperate sub gen.
Is there any other reasoning to have it seperate?

deepu105:
Oops very sorry i didn't see your 3rd todo. Ignore my previous comment

cbornet:
No problem :smile:

cbornet:
Could someone test on Windows and Mac ?

gmarziou:
Pathological test: I ran `yo jhipster:remove-entity Foo` while I only had a Book entity. It made some things which looks like a mix between Foo and Book and finally failed. You may want to check that passed entity name exists as a user can still make a typo.

cbornet:
Eventually, the code will be called from the entity subgen (TODO 3) so you will not be able to call it if the entity doesn't exist.

gmarziou:
My env: Windows 10, git version 2.8.1.windows.1,  node: v4.3.1,  npm: 3.10.9

gmarziou:
Minor thing: when you run remove-entity and there's only one entity, there is a message: `ls: no such file or directory: .jhipster\*.json`

gmarziou:
I found that it fails after executing `git diff --name-only --diff-filter U`, it returns code=0, msg="", err="".
The problem is that msg is empty and this causes `msg.trim().split('\n').forEach( file => shelljs.rm(file));` to fail.

cbornet:
Thanks. I'll add a null check.

cbornet:
Or maybe that's a CRLF issue since I split on LF

gmarziou:
Yes it could be as I would expect at least one EOL even when there's an empty list.

pascalgrimaud:
@cbornet : FYI, there are some eslint errors

Dufgui:
@cbornet : I test without a git repo and inside a svn repo. So it corrupt my svn repo. I think after git init we must add ignore on all dot files (except jhipster one) maybe no ? or maybe just .svn

cbornet:
@Dufgui Then you probably also have the issue with jhipster:upgrade, don't you ?

jdubois:
This has been opened for quite a long time, sorry @cbornet I didn't see you were the author, we should have discussed this when we met a couple of days ago!
-> what's the status of this PR, is it ready to be merged (once the conflicts are solved)?

cbornet:
It's not ready to be merged. There are some issues on Windows and it needs to be moved to the entity gen. I'll try to get back on this asap.

jdubois:
This has been stuck for more than 3 weeks - I know many people want this, but this looks harder than expected, and once again this shouldn't happen if you use Git correctly (just commit before/after generating, so you can revert a change).
I would like us to focus on JHipster 4 instead, can we close this?

cbornet:
I'm sorry I have so little time at the moment...
You can close if you want and I'll open a new PR when I have time to fix the issues reported.

jdubois:
@cbornet we all do this on our free time (I'm on paternity leave at the moment), so there's no rush and no obligation to work!!!
But I think we lose time on this, and we should rather focus on JHipster 4. We should deliver it this month!

PR #4408, opened 2016-10-30
Small optimization

MathieuAA:
Used `let` in the for loop (and used the `var` which was equivalent, but still).


MathieuAA:
I'll move it to JHCore right now. You may have the same problem with the camel case though. I've fixed it in the core.

PR #4411, opened 2016-10-30
AuthenticationType value mismatch in gatling test templates

mbergumi:
The EntityGatlingTest template expects an authenticationType value of oauth2 when the jhipster-generators uses "uaa"


jdubois:
I don't agree, the "oauth2" is used by the "oauth2" type. "uaa" was added by @xetys and is still in BETA, so I'm not too surprised it doesn't play well with Gatling.
I guess you need to check for both, but I haven't got the time to check now.

mbergumi:
allright. I made that assumption reading at the java tests, but i get it. I'll update to accept both uaa and oauth2 values then.

mbergumi:
it's corrected, let me know if i have missed anything.
, @jdubois

PR #4412, opened 2016-10-31
Fix for critical bug when social login activated

flaviencathala:
@deepu105 you have moved the import of SocialComponent, and weirdly it causes a critical issue when the social login is activated (the application can't launch). I don't know why it doesn't work directly from './' like other components, but if you do please tell me. Otherwise, merge the PR to fix it. Thanks.


deepu105:
fixed in commit mentioned

PR #4413, opened 2016-10-31
Use of external module ng2-cookies instead of $document for cookies

flaviencathala:


deepu105:
I think myself and @wmarques wanted to use https://github.com/salemdar/angular2-cookie which is more popular and consistent with angular2 way of doing stuff can you cange to use the above mentioned lib please

flaviencathala:
I made an other PR for angular2-cookie.

PR #4414, opened 2016-10-31
[WIP] interceptors migration

sendilkumarn:
@deepu105 @wmarques with this I can intercept the requests.

But instead of emit, we have to move towards observables.


deepu105:
Agree 100% we need to use observable insted of emit/broadcast etc

xetys:
I killed hours of my weekend stuck on interceptors 😆 I need that thing for UAA authentication, too

generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
```diff
@@ -57,7 +60,12 @@ let routerConfig = {
     providers: [
         ProfileService,
         { provide: Window, useValue: window },
-        { provide: Document, useValue: document }
+        { provide: Document, useValue: document },
+        {
+            provide: Http,
```
deepu105: Will this work if we define this provider in the shared libs module?

xetys:
well, I need some more explanation, what is wrong with this PR. To clarify, I need that CustomHttp service to make UAA auth work again. Maybe I can help with the Observable issue, if I know what issue we are talking about

sendilkumarn:
@xetys I have few missing parts to complete.
1. _auth.interceptor.js  ( partial )
2. _auth-expired.interceptor.js

But the problem is I don't have time now. If needed urgently, you can work on top of this.

> Observable

yeah we are using observables here. I was just mentioning, that we are using that.

xetys:
ok, I already moved this topic locally to a solution I will present today or tomorrow. I worked out some solution to decouple the logic from that one (potentially huge growing) interceptor into http interceptable classes (chain of responsibility pattern).

I need some more time to carry my working prototype to a PR on your branch...

xetys:
So here is my current state.

The idea is to use the chain of responsibility pattern to solve the http interception problem in general. The motivation is to prevent the http interceptor growing too big with all our (and possible new) solutions (sessions, jwt, oauth2...) into a hard coupled mess. So now, it only gets an array of `HttpInterceptable`s, to chain process both, request and response. So an example use case is to add a bearer token to each request, if there is one stored in local/session storage.

As I am not that familiar with Angular2s IoT container, the only stable working solution was to wire the dependencies in app.ng2module, like shown in my first implementation. By the way, it should be done on app level, to make the interceptors hook in app wide

If this is ok to you, I suggest to implement all our interceptors (auth expired, notification...) like this.

However with this setup I am now able to login using UAA (like always, [up to date on my sample setup](https://github.com/xetys/jhipster-uaa-setup)) and click through the ng2 frontend.

generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_http.interceptor.ts
```diff
@@ -1,72 +1,87 @@
...
 
-import { AuthServerProvider } from '../../shared/auth/auth-session.service';
 
 @Injectable()
-export class CustomHttp extends Http {
-
- constructor(backend: ConnectionBackend, defaultOptions: RequestOptions, private auth: AuthServerProvider <%_ if (authenticationType === 'oauth') { _%>, private principal: Principal <%_ } _%> ) {
+export class HttpInterceptor extends Http {
+    private firstInterceptor : HttpInterceptable
+ constructor(
```
deepu105: there is some indent issue here

generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_http.interceptor.ts
```diff
@@ -1,72 +1,87 @@
...
-        return options;
+
+        return !this.firstInterceptor ? options : this.firstInterceptor.processRequestInterception(options);
     }
- 
+
     intercept(observable: Observable<Response>): Observable<Response> {
+        return !this.firstInterceptor ? observable : this.firstInterceptor.processResponseInterception(observable);
+    }
+    /* TODO, implement logic for all interceptors in similar fashion as AuthInterceptor
```
deepu105: this is important, to merge this

generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_http.interceptor.ts
xetys: I will later look how to do this....

deepu105:
@sendilkumarn @xetys guys is this ready or does it need more work?

xetys:
there are still more interceptors to do...error, notification...but as the whole branch is alpha I wish a merge, as PRing to sendil to PR to here is somekind suboptimal

For other side I need this merge, as the first interceptors are already making UAA login work, so I could proceed to UAA user management, which is not that hard anymore...


for me: please merge but proceed with more interceptor PRs (maybe opening a issue with subtask for that?)

deepu105:
@xetys @sendilkumarn ok I have merged this. you can do more PRs for remaining

PR #4416, opened 2016-11-01
Remove ui-router for ng1

christopherthielen:
_This is a step towards removal_.  There are still some things to be done.  ui-router ng1 services are still being injected in a few places:

- `$state`: use `StateService`.  it can be injected as `StateService` or `Transition.router.stateService`
- `$stateParams`: 
  - inject the `Transition` and get its `.params()`
  - in a global context (service/interceptor), use `UIRouter.globals.params`

- The auth code could use an overhaul
- Need to eliminate the use of `$rootScope`


generators/client-2/files.js
```diff
@@ -140,6 +140,8 @@ const files = {
                 'layouts/profiles/_profile.service.ts',
                 'layouts/profiles/_profile-info.model.ts',
                 'layouts/profiles/_page-ribbon.component.ts',
+                'layouts/main/_jhipster-main.component.ts',
```
christopherthielen: Added a main component which acts as the main layout. contains the top level  uiviews

generators/client-2/templates/src/main/webapp/_index.html
```diff
@@ -9,6 +9,7 @@
     <meta name="google" value="notranslate">
     <%_ } _%>
     <meta name="viewport" content="width=device-width">
+    <base href="/"/>
```
christopherthielen: Is this OK?

generators/client-2/templates/src/main/webapp/_index.html
```diff
@@ -26,9 +27,9 @@
         <p class="browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
     <![endif]-->
     <page-ribbon></page-ribbon>
-    <div ui-view="navbar" ng-cloak></div>
+    <jhipster-main></jhipster-main>
```
christopherthielen: main component.  this is  a downgraded ng2 component.  This Will have to change once we're out of hybrid mode

generators/client-2/templates/src/main/webapp/app/_app.ng2module.ts
```diff
@@ -27,6 +28,7 @@ import { localStorageConfig } from './blocks/config/localstorage.config';
 localStorageConfig();
 
 let routerConfig = {
+    configClass: JHipsterRouterConfig,
```
christopherthielen: Global configuration is possible by using a config class

generators/client-2/templates/src/main/webapp/app/_router.config.ts
```diff
@@ -0,0 +1,26 @@
...
+
+        router.urlMatcherFactory.type('boolean', {
+            decode: function(val: string): boolean { return val === '1' || val === 'true'; },
+            encode: function(val: boolean) { return val ? '1' : '0'; },
+            equals: function(a, b) { return this.is(a) && a === b; },
+            is: function(val) { return [true,false].indexOf(val) >= 0; },
+            pattern: /0|1|true|false/
+        });
+
+        registerTransitionHooks(router.transitionService);
```
christopherthielen: Global transition hooks are registered here

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -1,66 +1,52 @@
...
-
-export function StateHandler($rootScope, $transitions, <% if (enableTranslation) { %>/*<%=jhiPrefixCapitalized%>LanguageService,*/<% } %> $window,
-    /*Auth, Principal,*/ $uiRouter, $trace) {
-
-    if (DEBUG_INFO_ENABLED) {
-        $trace.enable('TRANSITION');
-        SystemJS.import('ui-router-visualizer').then(vis => vis.visualizer($uiRouter));
-    }
+StateHandler.$inject = ['$rootScope'];
+export function StateHandler($rootScope) {
```
christopherthielen: This doesn't do much anymore, so it should probably be removed

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -1,66 +1,52 @@
...
+        let $rootScope = transition.injector().get('$rootScope');
+        $rootScope.toState = transition.to();
+        $rootScope.toParams = transition.params();
+        $rootScope.fromState = transition.from();
 
-            /*if (Principal.isIdentityResolved()) { //TODO needs to fixed after migration
-                Auth.authorize();
-            }*/
+        /*if (Principal.isIdentityResolved()) { //TODO needs to fixed after migration
+         Auth.authorize();
```
christopherthielen: I can provide assistance on overhauling the authentication code.  Look at the sample app for an example.  

https://github.com/ui-router/sample-app-ng1/blob/master/app/global/auth.service.js - the authentication service

https://github.com/ui-router/sample-app-ng1/blob/master/app/global/requiresAuth.hook.js - the requires authentication hook

https://github.com/ui-router/sample-app-ng1/blob/master/app/main/app.states.js#L44-L83 - the login state and resolve, which determines where to return to after a successful authentication

generators/client-2/templates/src/main/webapp/_index.html
deepu105: Does this enable html5mode for routing?

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
deepu105: Your welcome to help on anything with ng2 we are really struggling as there is a lack of good documentation and samples for complex stuff

generators/client-2/templates/src/main/webapp/app/blocks/handlers/_state.handler.ts
```diff
@@ -1,66 +1,52 @@
...
-        });
-    }
+             $translate.use(language);
+             });*/
+        <% } %>
+    });
+
+    // Redirect to a state with an external URL (http://stackoverflow.com/a/30221248/1098564)
+    $transitions.onStart({ to: state => state['external'] }, (transition: Transition) => {
+        window.open(transition.to().url, '_self');
```
deepu105: I think so as the original idea was to handle external page redirects

generators/client-2/files.js
deepu105: I have renamed the files to be consistent with what we generate

Issue #4402, opened 2016-10-28
ClassNotFoundException when using Kafka + Gatling


raphaelbrugier:
##### **Overview of the issue**

In a microservice architecture, both for a gateway and a service, a class is missing when adding the kafka option.

I've run into this stack trace when starting a gateway with the kafka option and a service too in Kubernetes.
I've reproduced the issue with docker-compose.
##### **Reproduce the error**
- generate a gateway with kafka
- generate a service with kafka
- generate a docker-compose config
- start everything

gateway stacktrace:

```
The application will start in 0s...

        ██  ██    ██  ████████  ███████    ██████  ████████  ████████  ███████
        ██  ██    ██     ██     ██    ██  ██          ██     ██        ██    ██
        ██  ████████     ██     ███████    █████      ██     ██████    ███████
  ██    ██  ██    ██     ██     ██             ██     ██     ██        ██   ██
   ██████   ██    ██  ████████  ██        ██████      ██     ████████  ██    ██

:: JHipster 🤓  :: Running Spring Boot 1.4.1.RELEASE ::
:: http://jhipster.github.io ::

2016-10-28 20:56:56.312  INFO 6 --- [           main] com.mycompany.myapp.JhigateApp           : The following profiles are active: prod
2016-10-28 20:57:13.702  WARN 6 --- [           main] o.s.c.a.ConfigurationClassPostProcessor  : Cannot enhance @Configuration bean definition 'refreshScope' since its singleton instance has been created too early. The typical cause is a non-static @Bean method with a BeanDefinitionRegistryPostProcessor return type: Consider declaring such methods as 'static'.
2016-10-28 20:57:33.691  INFO 6 --- [           main] c.n.d.provider.DiscoveryJerseyProvider   : Using JSON encoding codec LegacyJacksonJson
2016-10-28 20:57:33.702  INFO 6 --- [           main] c.n.d.provider.DiscoveryJerseyProvider   : Using JSON decoding codec LegacyJacksonJson
2016-10-28 20:57:34.804  INFO 6 --- [           main] c.n.d.provider.DiscoveryJerseyProvider   : Using XML encoding codec XStreamXml
2016-10-28 20:57:34.805  INFO 6 --- [           main] c.n.d.provider.DiscoveryJerseyProvider   : Using XML decoding codec XStreamXml
[...]

PR #4420, opened 2016-11-01
Fix microservice with kafka

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/4402


tezcane:
I believe this needs to be fixed on Gradle builds too.

pascalgrimaud:
@tezcane : we should not have this issue anymore as we switch to Camden instead of using Brixton

If you still have an issue, plz open a new ticket and provide all information to reproduce

PR #4429, opened 2016-11-04
Fix translations with links

mraible:
ng2-translate does not process directives in translation keys. This is a workaround.

See https://github.com/ocombe/ng2-translate/issues/223 for more information.

@jhipster/developers If you believe this is a good approach (not many other options at this point), I'll continue working on this PR and fix all the keys with embedded links.


deepu105:
I know this workaround but it requires we change this in all 23 languages :( I was trying to get it working without a work around but I guess if we don't have a solution we have no choice but this

generators/client-2/templates/src/main/webapp/app/account/register/_register.html
```diff
@@ -117,8 +117,9 @@ <h1 translate="register.title">Registration</h1>
                 <button type="submit" [disabled]="registerForm.form.invalid" class="btn btn-primary" translate="register.form.button">Register</button>
             </form>
             <p></p>
-            <div class="alert alert-warning" translate="global.messages.info.authenticated" translate-compile>
-                If you want to <a class="alert-link" (click)="openLogin()">sign in</a>, you can try the default accounts:<br/>- Administrator (login="admin" and password="admin") <br/>- User (login="user" and password="user").
+            <div class="alert alert-warning">
+                <span jhi-translate="global.messages.info.authenticated.prefix">If you want to </span>
+                <a class="alert-link" (click)="openLogin()">sign in</a><span jhi-translate="global.messages.info.authenticated.suffix">, you can try the default accounts:<br/>- Administrator (login="admin" and password="admin") <br/>- User (login="user" and password="user").</span>
```
deepu105: Don't u need to put the sign in text in a span inside the anchor and apply translate on that as well, same in other places

deepu105: Sorry you don't need a span you just need to apply the translate directive

generators/languages/templates/src/main/webapp/i18n/en/_global.json
```diff
@@ -48,8 +48,20 @@
         },
         "messages": {
             "info": {
+                <%_ if (angularVersion == 'angular1') { _%>
                 "authenticated": "If you want to <a class=\"alert-link\" href=\"\" ng-click=\"vm.login()\">sign in</a>, you can try the default accounts:<br/>- Administrator (login=\"admin\" and password=\"admin\") <br/>- User (login=\"user\" and password=\"user\").",
                 "register": "You don't have an account yet? <a class=\"alert-link\" href=\"\" ng-click=\"vm.register()\">Register a new account</a>"
+                <%_ } else if (angularVersion == 'angular2') { _%>
```
deepu105: Can we use === for all

deepu105:
Thanks


Thanks & Regards,
Deepu

mraible:
NOTE: I just discovered that the Japanese version of `reset.json` is not translated.

https://github.com/jhipster/generator-jhipster/blob/master/generators/languages/templates/src/main/webapp/i18n/ja/reset.json

mraible:
@deepu105 Please review again. I've translated all 27 languages (by counting directories under `i18n`).

deepu105:
OK I'll do and merge tomorrow

Thanks & regards,
Deepu

generators/languages/templates/src/main/webapp/i18n/ca/_global.json
```diff
@@ -48,8 +48,20 @@
         },
         "messages": {
             "info": {
+                <%_ if (angularVersion === 'angular1') { _%>
                 "authenticated": "Si vol <a class=\"alert-link\" href=\"\" ng-click=\"vm.login()\">autenticar-se</a>, pot usar els comptes per defect:<br/>- Administrador (nom d'usuari=\"admin\" i contrasenya=\"admin\") <br/>- Usuari (nom d'usuari=\"user\" i contrasenya=\"user\").",
                 "register": "No disposa d'un compte encara? <a class=\"alert-link\" href=\"\" ng-click=\"vm.register()\">Registri un nou compte</a>"
+                <%_ } else if (angularVersion === 'angular2') { _%>
```
deepu105: I think we can just change the ng1 templates as well to this style rather than maintaining two versions here. But it can be done later

Issue #4373, opened 2016-10-24
Tests generated for ZonedDateTime entity attributes are failing


mmaia:
##### **Overview of the issue**

Tests generated for ZonedDateTime entity attributes are failing, basically the generated tests are using ZoneId.systemDefault() which causes the test to fail.

The fix is to use UTC instead as discussed here: http://stackoverflow.com/questions/40191004/integration-tests-with-zoneddatetime-using-spring-boot-1-4-error

So this is how it's generated: 

``` java
private static final ZonedDateTime DEFAULT_DATE = ZonedDateTime.ofInstant(Instant.ofEpochMilli(0L), ZoneId.systemDefault());
private static final String DEFAULT_DATE_STR = DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(DEFAULT_DATE);
```

and this is the fix: 

``` java
private static final ZonedDateTime DEFAULT_DATE = ZonedDateTime.ofInstant(Instant.ofEpochMilli(0L), ZoneOffset.UTC);
private static final String DEFAULT_DATE_STR = DateTimeFormatter.ISO_INSTANT.format(DEFAULT_DATE);
```
##### **Motivation for or Use Case**

When generating a new entity and creating an attribute as a ZonedDateTime the integration tests fail. 
##### **Reproduce the error**

Create a new jhipster project. 
Create a new entity using `yo jhipster:entity MyEntity`
Add a property of type ZonedDateTime to your entity
Run generated tests for this entity

Tests will fail with an error like: 

```
java.lang.AssertionError: JSON path "$.[*].date"
Expected: a collection containing "1970-01-01T01:00:00+01:00"
 but: was "1970-01-01T00:00:00.000Z"
```

I've extensively tested that generating 3 projects and always had the same issue.
##### **Related issues**

issue #815 was a similar one: 
https://github.com/jhipster/generator-jhipster/issues/815
##### **Suggest a Fix**

The fix that worked for me is to stop using system default zone and use UTC instead, issue and solution properly described here: 

http://stackoverflow.com/questions/40191004/integration-tests-with-zoneddatetime-using-spring-boot-1-4-error
##### **JHipster Version(s)**

JHipster Generator v3.9.1
##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

``` json
{
  "generator-jhipster": {
    "jhipsterVersion": "3.9.1",
    "baseName": "zonedDateTime",
    "packageName": "com.mmaia.jhipster",
[...]

PR #4433, opened 2016-11-05
Configure default date format as ISO

cbornet:


jdubois:
You're clearly the boss on this kind of subject @cbornet , but is this solving something like #4373 ? I'm just trying to understand the impacts... Otherwise let's merge this, as you know all this better than anyone else

cbornet:
I don't think it solves #4373. It just configures the converters to use the ISO format for dates by default so you don't have to use `@DateTimeFormat` annotation in the RequestMapping methods (since ISO is our format both in front-end and back-end).

PR #4435, opened 2016-11-06
change @PostConstruct to @Before in tests

### LINKED ISSUE #4249: The mocked service methods of integration tests using the new @SpringBootTest annotation always return null

The new Spring Boot 1.4 @SpringBootTest annotation breaks the use of mocked services in the JHipster integration tests. Any mocked service method call using the @SpringBootTest annotation always returns null.

While the generated integration tests all pass without any issues, any attempt to use mocked services to return pre-specified objects will only return null objects instead. The issue appears to be the new @SpringBootTest annotation as reverting to the previous Spring Boot 1.3 deprecated annotations resolves the issue.
##### **Reproduce the error**

I have a modified fork of the JHipster Sample Gradle App project that shows an integration test failing with the @SpringBootTest annotation but the same test working properly with the deprecated annotations.

The forked project: https://github.com/andrewkrentz/jhipster-sample-app-gradle-springboottest-annotation-issue
- The BankAccountResourceIntTest createEntity test fails as the mocked UserService returns a null object when it should not.
- The WorkingBankAccountResourceIntTest createEntity test succeeds using the deprecated Spring 1.3 annotations as the mocked UserService works as expected.
##### **Related issues**

Not that I have been able to find.
##### **Suggest a Fix**

None, though it could be a Spring Boot/Mockito issue and not a JHipster one.
##### **JHipster Version(s)**

3.8.0

```
{
  "generator-jhipster": {
    "baseName": "jhipsterGradleSampleApplication",
    "packageName": "io.github.jhipster.sample",
    "packageFolder": "io/github/jhipster/sample",
    "hibernateCache": "ehcache",
    "websocket": false,
    "databaseType": "sql",
    "prodDatabaseType": "mysql",
    "useCompass": false,
    "buildTool": "gradle",
    "rememberMeKey": "e48612ba1fd46fa7089fe9f5085d8d164b53ffb2",
    "searchEngine": false,
    "enableTranslation": true,
    "testFrameworks": [
      "gatling",
      "protractor"
    ],
    "languages": [
      "en"
    ],
    "serverPort": 8080,
    "jhipsterVersion": "3.8.0",
    "enableSocialSignIn": false,
    "useSass": false,
    "jhiPrefix": "jhi",
    "messageBroker": false
  }
}
```

N/A

Browser: N/A
OS: Mac OS X El Capitan

mehmet-aslan:
Fix #4249

New instances are created for mocked fields of tests between `@PostConstruct` method and `@Before` method.

This is confusing when using Mockito methods in test methods as these are not affect objects used in setup method.

It could be caused by another issue but changing `@PostConstruct` to `@Before` works well.

Is there any specific reason to use `@PostConstruct`?

### DIFF
--- a/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
+++ b/generators/entity/templates/src/test/java/package/web/rest/_EntityResourceIntTest.java
@@ -34,7 +34,6 @@ import org.springframework.test.web.servlet.setup.MockMvcBuilders;<% if (databas
 import org.springframework.transaction.annotation.Transactional;<% } %><% if (fieldsContainBlob == true) { %>
 import org.springframework.util.Base64Utils;<% } %>
 
-import javax.annotation.PostConstruct;
 import javax.inject.Inject;<% if (databaseType == 'sql') { %>
 import javax.persistence.EntityManager;<% } %><% if (fieldsContainLocalDate == true) { %>
 import java.time.LocalDate;<% } %><% if (fieldsContainZonedDateTime == true) { %>
@@ -204,7 +203,7 @@ public class <%= entityClass %>ResourceIntTest <% if (databaseType == 'cassandra
 
     private <%= entityClass %> <%= entityInstance %>;
 
-    @PostConstruct
+    @Before
     public void setup() {
         MockitoAnnotations.initMocks(this);
         <%= entityClass %>Resource <%= entityInstance %>Resource = new <%= entityClass %>Resource();


xetys:
The `@PostConstruct` is called once before the test, while `@Before` is running before each test method. Things like preparing the mock have no reason to be constructed each time again. When something is failing, then it should be reviewed individually, but currently the tests work great going the current way. 

Maybe considering a split into two methods and put things to run once per test in post construct and repeatable things into before methods?

mehmet-aslan:
JUnit constructs a new instance of the class before each test method so `@PostConstruct` is already running more than once. There is no benefit from `@PostConstruct` while it causes this issue.

jdubois:
I'm the culprit here, and I can't remember why I used `@PostConstruct` instead of a JUnit annotation here.... I guess it was to be sure the beans where correctly initialized: in that case that's me being lazy, as I'm not totally sure of the state of beans with `@Before`, I should work on that (but if you tell me that's OK, it's fine I'm trusting you!).
BTW, and I think that's the issue here between @mehmet-aslan and @xetys , shouldn't we use `@BeforeClass` here? So it only runs once. It should all be stateless anyway, so if this causes an issue, then that's even better.

xetys:
I know `@Before` is meant to be called before each test method. I am not sure if PostConstruct is running multiple times. AFAIK, it does not (I am using it this way in several ways). Or in other words, I never recognized it working more than once, but I could be wrong in this point

However, if  we use `@Before`, we should not reinitialize everything for each test method. I would prefer to split these into two methods (before / beforeClass)

I should take a look at this mocking issue, as I am using Mockito continuously, with last spring boot version without any issues, so this all is somehow confusing for me

mehmet-aslan:
`@BeforeClass` only works with static methods. The setup method here uses instance variables so it cannot be used.

The problem of `@PostConstruct` is JUnit always creates fresh new instance of test class for each test method and because of that `@PostConstruct` runs on newly created object before each test method.

I don't have strong evidence about beans, but I saw `@Before` is called after `@PostConstruct` every time if both are exist and beans are initialized when `@PostConstruct` is changed to `@Before` so I assume it will work with beans.

xetys:
@mehmet-aslan is right...`@PostConstruct` is called per test method. Therefore of course, its better to use `@Before`...so LGTM

we should remember this topic when JUnit 5 is part of JH, so we use `@BeforeAll` for the once per test setup

jdubois:
OK so it looks like we all agree that this is better than my `@PostConstruct`, and anyway I mostly used it because I know Spring better than JUnit :-)
So let's use it!

PR #4439, opened 2016-11-07
Remove "Beta" from entity prompts

mikeyhogarth:
During your (excellent!) Devoxx2016 talk you said you wanted to take this out of beta because it always works.


jdubois:
Thanks a lot!

deepu105:
Thank you

PR #4442, opened 2016-11-07
Enable oauth with live reload

bagnier:
/oauth was not served by the Browsersync live reload. This patch enables oauth with live reload.


jdubois:
Oh, this is my fault, thanks!
I don't understand why you're the first one to notice that!!

PR #4426, opened 2016-11-03
Yarn flag

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/4317

How to use:
`yo jhipster --yarn`

Sorry @jdubois : I have no time for changing this with a new question -> but it can be done later

PR #4444, opened 2016-11-08
Install yarn in Docker image [ci skip]

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/4317
To allow the use of `yo jhipster --yarn` (https://github.com/jhipster/generator-jhipster/pull/4426) with the Docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,11 +31,12 @@ RUN \
   # upgrade npm
   npm install -g npm && \
 
-  # install yeoman bower gulp
+  # install yeoman bower gulp yarn
   npm install -g \
     yo \
     bower \
-    gulp-cli && \
+    gulp-cli \
+    yarn && \
 
   # cleanup
   apt-get clean && \


cbornet:
Yo, bower and gulp-cli could also be installed with yarn

pascalgrimaud:
Yes I agree with you @cbornet but I'm not sure if we should promote completely yarn over npm.
If we do it, we have to be consistent everywhere:
- change the doc https://jhipster.github.io/installation/
- change the travis build: https://github.com/jhipster/generator-jhipster/blob/master/travis/install/02-install-jhipster-stack.sh#L10-L12
- merge your PR in the devbox: https://github.com/jhipster/jhipster-devbox/pull/72
- change the pom: https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/_pom.xml#L1115-L1118 -> look here https://github.com/eirslett/frontend-maven-plugin/pull/499

So I'm not sure if yarn is stable enough to do that.
All the team need to agree before doing all these changes :)

sendilkumarn:
@pascalgrimaud very valid point. 

We should migrate completely, if we have to promote it 👍

cbornet:
Completely agreed ! Sorry for the impatience ! :smile:

PR #4452, opened 2016-11-09
remove useless cookie service import

flaviencathala:


flaviencathala:
@deepu105 As you asked.

PR #4454, opened 2016-11-09
Hibernate 5.2, Ehcache 3, JCache support

henri-tremblay:
This is the current version on my work in Hibernate 5.2 and Ehcache 3.

It includes the following:
- Migrate Hibernate to 5.2
- Migrate Jackson hibernate module to be compatible with Hibernate 5
- Migrate liquibase-hibernate to version 3.6 which support Hibernate 5
- Migrate to Ehcache 3 by using JCache
- Migrate Metrics for Ehcache 3 to use JCache metrics

**What is not working:**
- Hazelcast can probably be configured with JCache to have something more similar to Ehcache. And get JCache Metrics

**Questions I have:**
- CacheConfiguration used to remove metrics. That is weird because it was removing more than Ehcache metrics. So I don't know if it was buggy or should be moved to MetricsConfiguration
- Hibernate has changed its JPA compliant naming format. For instance a FK is now named FKblablabla instead of FK_blablabla. We have to provide our own implementation to fix that

**What is fun:**
- Caching metrics are now coming from JCache so could be used by any JCache implementation
- Because of JCache, there should be much more common code between caching implementation
- It's easier to switch to another caching framework


cbornet:
You should be able to remove JSR310PersistenceConverters.java since Hibernate5 supports JSR310 types out-of-the-box

PierreBesson:
The JCache metrics is a very nice addition.

henri-tremblay:
@cbornet Thanks. I will look at it.

Can someone look at why Hazelcast isn't happy? It seems to come from `com.hazelcast:hazelcast-hibernate52` that isn't really 5.2 compatible.

henri-tremblay:
Oh, I've added a little script to generate a JHipster project from all the samples. So, if you want to see the current failures, just run that. Travis should also return the same result of course.

generators/client/templates/src/main/webapp/app/admin/metrics/_metrics.controller.js
```diff
@@ -26,14 +26,16 @@
                 if (key.indexOf('web.rest') !== -1 || key.indexOf('service') !== -1) {
                     vm.servicesStats[key] = value;
                 }
-                if (key.indexOf('net.sf.ehcache.Cache') !== -1) {
+            });
+            angular.forEach(newValue.gauges, function (value, key) {
```
deepu105: Note to self: Need to update for NG2 version as well cc @sendilkumarn @wmarques

deepu105:
I will take a look at Hazelcast, but I wont have time untill weekend so if anybody wants to take a hit plz do so @jdubois do you have any contacts from hazelcast team?

jdubois:
@deepu105 I need to check, but I'm not too worried about Hazelcast + 2nd level cache for Hibernate.
My main worry is Hazelcast HTTP session clustering, which adds a lot of issues: if we remove this, I'm confident about the rest.

mesutcelik:
Hi @henri-tremblay @cbornet @jdubois ,

I think you have hit the [bug](https://github.com/hazelcast/hazelcast-hibernate5/issues/9) if you are using `hibernate 5.2.3.Final`. We already fixed the problem in v1.1.1

Can you please try [hazelcast-hibernate52-1.1.1](https://github.com/hazelcast/hazelcast-hibernate5/releases/tag/v1.1.1)? 

P.S: hazelcast-hibernate52-1.1.1 was just released so it might take a few hours to be in maven central if you are using maven.

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
```diff
@@ -0,0 +1,76 @@
...
+
+import javax.cache.management.CacheStatisticsMXBean;
+import javax.management.MalformedObjectNameException;
+import javax.management.ObjectInstance;
+import javax.management.ObjectName;
+
+import static com.codahale.metrics.MetricRegistry.*;
+
+/**
+ * MetricSet retriving JCache specific JMX metrics for every configured caches
```
jdubois: You need a dot at the end of the sentence, so it's correct JavaDoc

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String M_BEAN_COORDINATES = "javax.cache:type=CacheStatistics,CacheManager=*,Cache=*";
+
+    private Set<ObjectInstance> objectInstances;
+
+    public JCacheGaugeSet() {
+        try {
+            this.objectInstances = ManagementFactory.getPlatformMBeanServer().queryMBeans(ObjectName.getInstance(M_BEAN_COORDINATES), null);
+        } catch (MalformedObjectNameException e) {
+            log.error("Failed to retrieve JCache statistics", e);
```
jdubois: Shouldn't this be re-thrown? That should make start up fail?

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
```diff
@@ -0,0 +1,76 @@
...
+        return Arrays.stream(c.getMethods())
+            .filter(method -> method.getName().startsWith("get"))
+            .map(method -> method.getName().substring(3))
+            .collect(Collectors.toList());
+    }
+
+    private String toDashCase(String camelCase) {
+        return camelCase.replaceAll("(.)(\\p{Upper})", "$1-$2").toLowerCase();
+    }
+
```
jdubois: You have an extra line here

generators/server/templates/src/main/java/package/config/jcache/_SpringCacheRegionFactory.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hibernate.cache.CacheException;
+import org.hibernate.cache.jcache.JCacheRegionFactory;
+import org.springframework.core.io.DefaultResourceLoader;
+import org.springframework.core.io.Resource;
+
+import java.io.IOException;
+import java.util.Properties;
+
+/**
+ * Special hibernate region factory that will convert a Spring URI (e.g. classpath:ehcache.xml) to a real URI (e.g. file://ehcache.xml)
```
jdubois: dot missing at the end of the sentence, and Hibernate should have a capital H

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
henri-tremblay: Done

generators/server/templates/src/main/java/package/config/jcache/_SpringCacheRegionFactory.java
henri-tremblay: Done

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
henri-tremblay: Done

generators/server/templates/src/main/java/package/config/jcache/_JCacheGaugeSet.java
henri-tremblay: The exception shouldn't happen since I'm providing the query. But I did a bunch of refactoring to make the code simpler and I am now throwing an `InternalError`

jdubois:
@henri-tremblay I would like to do a specific release for this, do you know when your PR will be ready?

henri-tremblay:
I should work on it today. I've pushed the rebase. Now, it's basically a matter of fixing CacheConfiguration I think.

henri-tremblay:
Everything seems to work now. Let's see if the build agrees

jdubois:
I changed very few things, and all my tests worked, so this is merged!!!

cbornet:
:tada: 
Now we need a clear migration tutorial for the liquibase checksum calculation change !

PR #4460, opened 2016-11-11
Removed unnecessary Maven profile

chaodhib:
Removed unnecessary Maven profile after the recent MapStruct related changes

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1369,20 +1369,5 @@
                 <spring.profiles.active>dev,swagger</spring.profiles.active>
             </properties>
         </profile>
-        <profile>
-            <!--
-                Profile for applying IDE-specific configuration.
-                At the moment it only configures MapStruct, which you need when working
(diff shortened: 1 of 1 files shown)


jdubois:
I've done several tests with Eclipse, and in fact I kept it on purpose.

It is not written in the MapStruct documentation (ping @gunnarmorling ) but if we don't have this specific Maven profile to add the `mapstruct-processor` to the classpath, then when you add an entity, the mapper won't be generated.

What I don't understand is that we don't seem to have any issue with the UserDTO (which is here by default, before any entity is added).

So for me, we need to keep this profile.

Ping also @cbornet as he coded this part.

chaodhib:
Oh ok, my apologies. I thought you forgot about it.

jdubois:
@chaodhib no it was on purpose, although I don't really understand why we need it...
Anyway, thanks for reviewing my commit :-)
If you have time to test this, it is really welcome, as I never use Eclipse so I'm not the best tester for this.

PR #4463, opened 2016-11-12
use Hibernate initialization to eagerly load user authorities

hdurix:
use Hibernate.initialize() instead of .size() to load relation which may
confuse readers

NB: an other solution could be to load the association in UserRepository but this commited solution is a quicker one and needs less code changes.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -10,6 +10,9 @@ import <%=packageName%>.security.AuthoritiesConstants;
 import <%=packageName%>.security.SecurityUtils;
 import <%=packageName%>.service.util.RandomUtil;
 import <%=packageName%>.web.rest.vm.ManagedUserVM;
+<%_ if (databaseType == 'sql') { _%>
+import org.hibernate.Hibernate;
+<%_ } _%>
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


jdubois:
Yes but I didn't want to use Hibernate specific APIs.

hdurix:
Alright, do you want me to have a look on the other solution?

I can add 2 new methods in UserRepository: findOneWithAuthorities and findOneByLoginWithAuthorities as it is done with findAllWithAuthorities.

jdubois:
Oh yes, and I didn't recognize you (sorry, I'm using my phone), great to have you contributing!!

pascalgrimaud:
@hdurix : FYI, I think your local git configuration isn't correct. So github didn't recognize you in commit log -> see here https://github.com/jhipster/generator-jhipster/pull/4463/commits

cbornet:
Yes : either eager load or use a DTO.

hdurix:
@jdubois : Ok I'll do this way, it will be much cleaner. No worries & thanks, happy to be onboard!
@pascalgrimaud : Thanks, I'll have a look at it.

jdubois:
So sorry @hdurix but as you understand I'll keep it the way it is:

- OK it's not very readable code, and yours is better
- But the idea is that we stick to the JPA API as much as possible, and try not to use the Hibernate API

hdurix:
Yes, I'll create a new pull request once I have the other solution with JPA done.

PR #4470, opened 2016-11-14
eager load user authorities with JPA

hdurix:
eager load authorities in user repository and stop using .size() in user service (only for sql databases).

Better version of this PR: https://github.com/jhipster/generator-jhipster/pull/4463
It seems more readable and it uses JPA API over Hibernate API.


nkolosnjaji:
Little suggestion...

@Query annotation can be avoided by using JPA graphs. This way it will be a type-safe.
spring-data-jpa also [supports ](http://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.entity-graph) this JPA 2.1 feature.

hdurix:
Yes exactly, it could be better to use JPA graphs.

The problem is sometimes, you need to load authorities eagerly, sometimes you don't. For instance, you search user by login without authorities to delete him, and you load user by id with authorities to be sent to client side.
I'm not sure you'd be able to have those 2 methods in the repository, ie `findOneByLogin` and `findOneWithAuthoritiesByLogin`: both would have to be named `findOneByLogin` to be understood by spring-data-jpa.

Except if the prefix is different, so you [could create 2 methods](http://stackoverflow.com/a/33373905): 

- `findOneByLogin` (without authorities)
- `readOneByLogin` (with authorities thanks to JPA graphs)

Since the prefix is different, it will be OK for spring-data-jpa, but maybe it will also be more confusing than using `@Query`.

What do you think? Am I missing a point?

nkolosnjaji:
@hdurix you can use two methods like this:

- findOneByLogin
- findWithAuthoritiesByLogin

based [on](http://docs.spring.io/spring-data/jpa/docs/1.10.4.RELEASE/reference/html/#repositories.query-methods.query-creation) spring data parser will strip prefix find…By

hdurix:
Great! I can do this way.

delverdev:
Seems like it also resolves [n + 1 problem](http://stackoverflow.com/questions/97197/what-is-the-n1-selects-issue). Is this problem resolved in another entities? I just don't know is it worth to open a new ticket.

hdurix:
I added "ById" in `findOneWithAuthoritiesById`, I reckon I had to.

Moreover, I tried to use the [procedure](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#submitting-a-pull-request) to update my pull request, I had some troubles so I'm not sure it is ok.

geraldhumphries:
@delverdev this does solve the n+1 problem, but if you are trying to use pagination it introduces another one: the result of the query is a cartesian product and hibernate needs to paginate in-memory. This means it will perform the select without a limit then get a subset of that result in-code.

You need to be careful to only apply this solution where appropriate.

nkolosnjaji:
@hdurix UserDetailsService (security package) is also using userRepository to fetch roles.

hdurix:
Oh yes you're right, it does not use .size() to fetch authorities but it has to be remplaced as well!

New commit incoming...

nkolosnjaji:
@geraldhumphries can you explain more in detail about possible problebles with EntityGraphs and Pageable parameter ? 

I test with this code and it works for /#/user-management?page=2 in admin section.

```
    @EntityGraph(attributePaths = {"authorities"})
    Page<User> findAllWithAuthoritiesBy(Pageable pageable);
```

I just replaced **@Query** annotation with **@EntityGraph** and added `By` suffix

geraldhumphries:
@nkolosnjaji when this query is performed, you will see the following warning in your console: `o.h.h.internal.ast.QueryTranslatorImpl   : HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!`

This warning means that it selected every user without applying paging params, then looped through them in code to build the page you requested. This can lead to massive performance issues. 

For example I was experimenting with entity graphs to improve the performance a query on a large entity with ~70,000 rows. I had set up the entity graph to fetch a bunch of relationships. What ended up happening is that all 70,000 rows and their relationships were loaded into memory then partitioned into a page with the 10 results I requested. This took over 5 minutes.

This all happens because the result of a `join` is a cartesian product, and pagination is done with (in Postgres at least) LIMIT and OFFSET.

E.g. if you have ten users and each user has three authorities, then the query `SELECT * FROM user LEFT JOIN authority ...` will have 30 results. If you want a page of the first 5 users, you can't just add LIMIT 5 OFFSET 0 to that query  - that would only get you the first two users with three authorities in the first one and two authorities in the second (this assumes the results are ordered by user). Hibernate's solution is to get all 30 results, put them into your entity objects (Users with a collection of Authorities) then grab the first 5 users.

nkolosnjaji:
@geraldhumphries you are right about `HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!` point, but it's not related to EntityGraphs. The use-case you described is *also* happening with current implementation, `Page<User> findAllWithAuthorities(Pageable pageable)` with @Query annotation never applied limit. Adding @EntityGraph to findAllWithAuthorities with Page return is safe, because it will produce same query like custom @Query.

@hdurix  and @gmarziou can you also test this with current implementation ? Just go to /#/user-management and you will se this error.

delverdev:
@geraldhumphries, so as I understood, in apps with large amount of rows, `Pageable` should be replaced by LIMIT with OFFSET? Because it may be a big 'surprise' in production.

nkolosnjaji:
@delverdev no, Pageable is using LIMIT and OFFSET, this just don't work(limitation) for SQL dabases when Pageing Entity with collection.

geraldhumphries:
@nkolosnjaji it does produce the exact same query, but I would not say it's safe. We should not be using either `@Query` or `@EntityGraph`.

I was unaware that JHipster uses `@Query` in this list because I have a custom implementation of authorities and probably removed the query altogether.

I would say that if we really need to display authorities in this list, we should probably rely on n+1 selects.

geraldhumphries:
We can solve many performance issues by setting `hibernate.default_batch_fetch_size` globally, or on specific collections using `@BatchSize`. This solves the n+1 selects problem because it tells hibernate to initialize collections for multiple entities in a single query. A batch size of 20 on `User.authorities` will mean that hibernate will only perform two queries for each page (size = 20) of users: one for the users, and one for all of their authorities.

nkolosnjaji:
@geraldhumphries then, if page size is 5, and we have users with more than 5 roles, will it display only 5 roles ?

geraldhumphries:
@nkolosnjaji no, because that's taken care of by the in-memory paging. If the user had 5 authorities and we attempted to use `limit` and `offset` with a join (and it's sorted by user with that user first), then yes, we would only get one user with 5 authorities.

I think the best thing to do here is to never combine join fetch/entity graphs with pagination. In other words, on the user management list we should be initializing the user authorities manually.

nkolosnjaji:
@geraldhumphries yes, I agree with you. @jdubois what do you think about this ? Obviously we are talking here about limitation of SQL databases.

I can see two solutions:

1.display only user without roles on users pages
2. like @geraldhumphries suggested, manually "initialize" user authorities.

jdubois:
Sorry guys I haven't got much free time at the moment.
I didn't want to do any of those tricks on the User entity, as it is being
used and modified by our end-users, and we don't know what they'll do.
That's why the relationship isn't eager.
Then, I'm doing .size() to force the collection loading as this should use
the second level cache (but this would need to be tested), so no more
database hit.

Le 18 nov. 2016 7:14 PM, "Nikola Kološnjaji" <notifications@github.com> a
écrit :

geraldhumphries:
@jdubois this is not about changing the User entity itself, but about changing the query on the user management page. Right now it uses `userRepository.findAllWithAuthorities` which has problems with pagination (see discussion above). The suggested solution was to actually force initialization by calling getAuthorities.size() in findAllUsers rather than the join fetch.

ctamisier:
I think this PR should only affect the findOne request (with EntityGraph or Query annotations) to avoid the use of size() to load a single user with its authorities.

For the findAll I do agree that it is not optimal because of the use of Pageable + fetch (and the use of size() to initialize the authorities would make sense)

jdubois:
@geraldhumphries I understand the issue, but doing the .size() is actually my older code -> do you suggest keeping the old code, and not merging this PR?

hdurix:
@jdubois `.size()` was done when getting a single user, @geraldhumphries advises to use it for user management list because of pagination.

But the use of getAuthorities.size() may not be necessary in our case for user management list because user authorities will be initialized by `UserDTO`'s constructor (thanks to `user.getAuthorities().stream()`).
We just have to be in a transaction.

I did the following test:

- use of `userRepository.findAll(pageable)` instead of the custom `@Query`
- add `@Transactional(readOnly = true)` in UserResource (I didn't have the service layer in this test)

The result was:

- no `HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!` warning any more
- 1 query to get all users, 1 query by user to get his authorities (only for user with at least one authority)

What do you think?

hdurix:
And the second level cache is used with this solution.

cbornet:
I also think that the DTO is the best solution but in that case, it should be initialized in a service (no Transactional in resource!)

hdurix:
Yes you're right, and I didn't notice the user service is always created. So you can add a new method in the service instead of calling the repository directly.

ctamisier:
This solution is optimal if we have an app with 'second cache level' enabled. (probably always a good idea but it is an option at the yeoman app generation)

Otherwise would '1 query by user to get his authorities' (n+1 select) be less performant than 'HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!' ? 

(n+1 SQL requests vs memory pagination)

generators/server/templates/src/main/java/package/repository/_UserRepository.java
```diff
@@ -43,10 +44,10 @@
     Optional<User> findOneByLogin(String login);
     <%_ if (databaseType == 'sql') { _%>
 
-    @Query("select user from User user left join fetch user.authorities where user.id = ?1")
-    User findOneWithAuthorities(<%= pkType %> id);
+    @EntityGraph(attributePaths = {"authorities"})
```
delverdev: I think `{"authorities"}` can be replaced by `"authorities"`. Other paths too.

delverdev:
I would suggest a solution for dynamically graphs. I tested it with my app with second level cache and all seem to work. First, create a method in some util service class (I called it `QueryExecutor`):
```
public <T> Optional<T> fetch(Class<T> clazz, Object id, String... relations) {
    EntityGraph<T> graph = entityManager.createEntityGraph(clazz);
    Stream.of(relations).forEach(path -> {
        String[] splitted = path.split("\\.");
        Subgraph<T> root = graph.addSubgraph(splitted[0]);
        for (int i = 1; i < splitted.length; i++)
            root = root.addSubgraph(splitted[i]);
    });
    return Optional.ofNullable(entityManager.find(clazz, id, Collections.singletonMap("javax.persistence.loadgraph", graph)));
}
```
We can delegate it in services:

```
public Optional<User> fetchById(long id, String... relations) {
    return queryExecutor.fetch(User.class, id, relations);
}
```

And use it like this:
```
    userService.fetchById(1L, "authorities");
```

Or even with subgraphs:
```
    userService.fetchById(1L, "authorities", "orders.products"); //authorities, orders and their products will be fetched
```

hdurix:
@delverdev I reckon it's a bit of a shame to do this job manually.

I tested the `@BatchSize` as explained @geraldhumphries, we can add this to the user-authorities relation (with the size of pagination: 20) and let the relation be initialized when creating the DTO. All the accesses to the database will be merge together for the 20 users thanks to the `@BatchSize`.
That way it will create 1 query to get all the users and 1 to get all their authorities.

Things to do:

- remove `findAllWithAuthorities(Pageable pageable)` from user repository
- add `@BatchSize(size = 20)` to user authorities
- move user DTO creation to user service in order to be transactional (instead of MVC resource)

I'll do a commit with these changes.

hdurix:
I'm not very satisfied of my last commit. 
I had to move all the code from user resource to user repository: `ResponseEntity` is now created in the service.
I don't think it should belong to this, but I didn't know how to do it otherwise.

I wanted to only find the users and map them to DTOs in the service but I should have returned 2 results: current page and DTO list. If someone's got a better idea...

Nevertheless, no more `HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!` warning and only 2 requests for each page!

generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -9,9 +9,19 @@
 import <%=packageName%>.security.AuthoritiesConstants;
 import <%=packageName%>.security.SecurityUtils;
 import <%=packageName%>.service.util.RandomUtil;
+<% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
+import <%=packageName%>.web.rest.util.PaginationUtil;
+<%_ } _%>
 import <%=packageName%>.web.rest.vm.ManagedUserVM;
```
jdubois: OK, so first I think we should refactor this part, so the service layer doesn't call the web layer, before we can update this part

hdurix: My last commit has been fixing this. I think it's a lot cleaner now: no class from the `web` layer is called in the service anymore (except the DTO).
I just mapped the page of user to a new one with the DTOs.

cbornet:
@hdurix you should rebase on upstream/master, not merge

hdurix:
Yes no problem I'll stop the merges.

But there is something I don't know how to do. Every time I rebase on upstream/master (when the branch has conflict to be resolved), it seems I have to resolve the conflicts for each commits of the branch (conflicts I have already resolved on previous rebases), is there a way to prevent this?

Do I need to do something for the 2 last commits (merges)?

Thanks!

generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -11,7 +11,10 @@
 import <%=packageName%>.service.util.RandomUtil;
 import <%=packageName%>.web.rest.vm.ManagedUserVM;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;<% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
+import org.slf4j.LoggerFactory;
+<% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
```
jdubois: You should use <%_ text _%> instead of <% text %>, so you don't add an empty line

hdurix: Great, I couldn't figure out what was the difference between them. I'll fix that, thanks!

generators/server/templates/src/main/java/package/repository/_UserRepository.java
```diff
@@ -41,9 +39,11 @@
...
     <%_ if (databaseType == 'sql') { _%>
 
-    @Query(value = "select distinct user from User user left join fetch user.authorities",
-        countQuery = "select count(user) from User user")
-    Page<User> findAllWithAuthorities(Pageable pageable);
+    @EntityGraph(attributePaths = "authorities")
+    User findOneWithAuthoritiesById(<%= pkType %> id);
+
+    @EntityGraph(attributePaths = "authorities")
+    Optional<User> findOneWithAuthoritiesByLogin(String login);
```
delverdev: It's not flexible. Let's add one-to-many  `actions` to `user`. Sometimes it may be needed to find a user only with `authorities`, but sometimes only with `actions`. Now we have to create `findOneWithActionsById` and `findOneWithActionsByLogin`. Furthermore, one day we will need to find a user with `authorities` **and** `actions`. So `findOneWithAuthoritiesAndActionsById` and `findOneWithAuthoritiesAndActionsByLogin` will be created. I don't even speak about 3+ relations or even inner relations like `action.consequences`.

hdurix: Yes you have to create one query by need, in order to fetch only what you need for your purpose. That way it's always optimized and you never grab all the database. Plus, with Spring Data JPA, you have to define every query like `findOneByCode`, `findOneByLabel`, `findOneByName`, ... to get what your want.
For the name, if you need to create a query with several relations, you can named it something like `findOneWithRelationsById` or `findOneForManagementById` to don't have a very long name.
Do you think it can be simplified?

delverdev: I suggested [this solution](#issuecomment-262909520) but now I have troubles with second-level cache. However, I tested your solution and seems like second cache doesn't work with it too. I tested it on my own application, based on JHipster, so now I'm installing a clean version and I will try to simplify it as best as I can. I will report.

delverdev: Now, on clean JHipster 3.12.1 cache doesn't seem to work. Each time I get `/api/account`, Hibernate logs ` select user0_.id as id1_4_,...`.  If it doesn't work on the clean application, our solutions will not work too.

hdurix: You're right, cache works with authority relationship (only loaded once) but not with user entity.

ctamisier: For this, query cache need to be activated because the request is not made on the primary key
`hibernate.cache.use_query_cache: true`

and enable it for the query in the repository
`@QueryHints({@QueryHint(name = "org.hibernate.cacheable", value = "true")})
Optional<User> findOneByLogin(String login);`

delverdev: @hdurix, do you have any idea how to make it work? Because therefore there is no sense in `@Cache` annotations on classes. I would even say that cache becomes useless in that case.
@ctamisier, I think the better solution is to find a way to get entities by primary key. Query cache is using on frequently loaded entities.

jdubois:
I resolved the merge conflicts to test, and I'm sorry because this is a very long thread, but I don't understand why we're doing this.

I tested with both versions: when a user logs in, which is the most important part of the app (admin screen is only used by admins, and probably many people change it or use another mechanism to add users), we still have 2 requests:

- Yes 2 requests is bad, this is because we use the login and not the user id, so we don't use the 2nd level cache -> there would definitely be something to improve here, but this PR does the same job as the current code
- The new request is much more complex: as it doesn't use the `.size()` method, it can't use the 2nd level cache, so it's less efficient (for people using a 2nd level cache, but this is our default and recommended approach)

So the bottom line is that I have the same number requests, and the requests are more complex.

hdurix:
For your first point, I did the same test as you, and I tried to log in on the 2 branches:

- on master, there are 2 queries: one for the user, and one for his authorities
- on this branch, there is only 1 query: for user and authorities (with a left fetch join). Did I miss something?

For your second point, I didn't notice that but yes that's totally true, the new request is less efficient (for example when a user connects twice). If the cache uses the login instead of the id, would the authorities be stored in cache as well (and correct this issue)?
As you said, for those 2 solutions, the query is done many times, but it could be improved in an other PR.

I agree with you, admin screen is not used very often, but the new version now uses the 2nd level cache for authorities (it's the opposite result of your second point, because the authorities are not directly fetched any more).

In a nutshell, how I could list the pros and cons of this PR:

- \+ does not use `.size()` (it can be bit hard to understand)
- \+ no collection fetch applied in memory, you first get the 20 users you want, and then the associated authorities --> see https://github.com/jhipster/generator-jhipster/pull/4470#issuecomment-260806751
- \+ use of `@EntityGraph` instead of `@Query` --> see https://github.com/jhipster/generator-jhipster/pull/4470#issuecomment-260396499
- \+ gives some nice examples to users of how retrieving entity with relations (even if the user management page is not used that often, it can be a good example for managing entities with relations)
- \+ 2nd level cache for authorities is used for admin view
- \- 2nd level cache for authorities is not used when logging in
- \- add notion of `@BatchSize` in `User` class, which needs to be the same as pagination constant `itemsPerPage`

jdubois:
I am not sure for the following points:

- + no collection fetch applied in memory -> I don't think we currently do this
- + use of `@EntityGraph` instead of `@Query` -> the argument is that we are "type safe", but I don't think we currently have this kind of issue at the moment

There are 2 things that are really important for me in this part:

- How other people will be able to extend and use our code, because this will happen a lot -> I have no idea which solution is better.
- The performance of the login process -> Indeed, on the first attempt, your solution has 2 requests, when the current code has 3 requests. But, as we use the 2nd level cache, we quickly have only 2 requests with the current code, and those requests are much simpler. BTW, I'm counting more requests than you, as we have one request for login the user in, and then the same one again when he has logged in to get his user information -> this is the part where the 2nd level cache doesn't work, and that's really a big issue for me.

So yes, in my previous comment I missed to count the first request to populate the authorities cache, and in that case your solution is better.

-> as this looks better, I'm OK to merge this, unless we have comments from people like @geraldhumphries who seems to have a very good opinion on this? Other comments?

jdubois:
I did more testing, for the admin screen I also find this better: there is the batch request for getting the authorities, but after that the request is much simpler, and indeed we don't do the collection fetch in memory (you were right!!).

So I'm merging this.

PR #4476, opened 2016-11-15
Easily build the samples

henri-tremblay:
I found this script really useful when testing my changes.

Seems to work fine but you might have another way to do it.

One thing:
I don't understand why all samples are one layer deep but `app-gateway-yaa` has a child project `uaa`. If it is normal, then it isn't supported by the script yet.


PierreBesson:
+1 It would need some testing by a third party before merging but something like this is badly needed when working on a big feature where you have a lot of tricky Travis test to pass quickly.

@xetys had developped something like this. Do you approve this ?

pascalgrimaud:
I was thinking about something similar in the past but I abandon the idea. I prefer pushing the code on my fork and let travis doing these tests (simply link travis to your forked project, no need to PR)
- so I can work on something else, waiting all Travis tests
- or I focus on specific use cases concerning my PR, which is not covered by Travis

pascalgrimaud:
@henri-tremblay : about UAA, it's because we use Travis Matrix.
Every folder [here](https://github.com/jhipster/generator-jhipster/tree/master/travis/samples) is a specific build. But theses 2 .yo-rc.json [files](https://github.com/jhipster/generator-jhipster/tree/master/travis/samples/app-gateway-uaa) are copied into 2 differents folders before using `yo jhipster` : 
- https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/01-generate-project.sh#L14
- https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/01-generate-project.sh#L21

But you're right, it can confused people to have 2 yo-rc.json in folder+sub folder.
As I coded it, I should split them :-)

deepu105:
Its a good idea

henri-tremblay:
@pascalgrimaud Running them in travis is ok but not enough. Because I found myself tweaking a little thing, then regenerating everything, then loading one of the projects in my IDE. Fixing it. Changing the generator. Running the generation again. And so on and so on.

pascalgrimaud:
@henri-tremblay : sure Travis is not enough. But if there is a same .yo-rc.json configuration which failed on your local but passed on Travis, let me know plz. As I coded a big part on our Travis matrix build, I'd like to know these cases.

Anyway, I'm ok with this script if it can help people for testing.

pascalgrimaud:
As mentionned by @PierreBesson, @xetys coded this: https://github.com/xetys/jhipster-travis-helper
But I'm not sure if it's up-to-date

henri-tremblay:
What @xetys did looks like a more complete version of what I did. So I think one or the other should go directly in generator-jhipster repo.

henri-tremblay:
A bit of code reading is telling me that my bash could possibly be improved by reusing the travis scripts. **But it still is a really good first version.** 

Refactoring `app-gateway-uaa` would also be really helpful. And simplify the travis scripts

Next step would be to adapt the travis scripts and use them.

pascalgrimaud:
@henri-tremblay : I will refactor app-gateway-uaa

pascalgrimaud:
@henri-tremblay : I would prefer to have this script under the travis folder, instead of the root folder (there are already a lot of files). What do you think?

henri-tremblay:
Fine with me.

henri-tremblay:
Done

pascalgrimaud:
Just tested and it works fine.
Some minor comments, it can be done in another PR:
- good way to update all these .yo-rc.json
- missing entities
- use `yarn link generator` + `yo jhipster --force --yarn` -> so much faster :smile: 

As you said, it's already a first good script, which can be improved later

Issue #3550, opened 2016-05-06
Secure actuator endpoints behind a specific /management URL


jdubois:
This is quite urgent, sorry to break stuff at the last minute, but it's for security.

This has already been discussed (ping @cbornet @gmarziou I think you participated), but we have an issue with our actuator endpoints:

- We have a lot of them secured in the SecurityConfiguration, and that makes a long file (and besides it has a small performance cost as Spring Security needs to go thru all of them)
- It's not fully secured as when Spring Boot Actuator adds more endpoints we might forget to update them (and there's probably already a security issue here today!!)
- Microservices are not secured, as they were supposed to be behind the gateway: in fact that's false in several cases, for example with Heroku, where all services are publicly available. So another big security issue.

My proposal, and I'm going to work on this ASAP (as it's security-related):

- Put all endpoints behind "/management" so it's easy to secure them with just one Spring Security rule
- (optional) Add basic auth support for them, so that the JHipster Registry (or any other tool) can access them remotely easily

gmarziou:
I am all for name spacing apis, so +1

PR #4477, opened 2016-11-15
Remove jhipster from the URL under the management part

gzsombor:
and fix pagination urls to correctly point back. 
This 'jhipster' separator doesn't add any functionality to the application, just confuses the thing.

### DIFF
--- a/generators/client/templates/src/main/webapp/app/admin/audits/_audits.service.js
+++ b/generators/client/templates/src/main/webapp/app/admin/audits/_audits.service.js
@@ -8,7 +8,7 @@
     AuditsService.$inject = ['$resource'];
 
     function AuditsService ($resource) {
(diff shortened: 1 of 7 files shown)


deepu105:
@jdubois added the seperator I guess. Not sure what was the reason so lets wait for his review.

pascalgrimaud:
It was done for security -> I find the ticket https://github.com/jhipster/generator-jhipster/issues/3550

gzsombor:
The admin functionality is still under /management/ only the jhipster prefix / or suffix is removed from the audit and the log listing

jdubois:
Thanks everyone, first some background on this:

- All admin features are behind "management" for security
- Then, I used "management/jhipster" for all JHipster endpoints, so you don't fix them with the Spring Boot actuator admin endpoints

So, concerning this specific change:

- Only downside is that we mix the JHipster specific endpoints with the Spring Boot "official" endpoints. I'm not sure it can cause any issue, only risk is that Spring Boot does a new endpoint which is the same as us - this is unlikely, but possible
- And of course I agree with @gzsombor that this is less confusing, and easier.

So +1 from me

PR #2771, opened 2016-02-03
Allow test continue if fail and generate the report

phouverneyuff:
With gradle, when only one test fail, the report isn't generated.
Removing too the println to be less wordy

PR #4484, opened 2016-11-16
Jenkinsfile: report tests results

raphaelbrugier:
This a proposal to improve the Jenkinsfile for Jenkins pipelines.

The tests, java or javascript are currently not reported.

This adds the tests reports but still fails the build as soon as the test stage is terminated.

## Before:
screen shot 2016-11-16 at 5 24 18 pm

## After:
screen shot 2016-11-16 at 5 23 26 pm

screen shot 2016-11-16 at 5 24 02 pm

### DIFF
--- a/generators/server/templates/_Jenkinsfile
+++ b/generators/server/templates/_Jenkinsfile
@@ -31,7 +31,7 @@ node {
         } catch(err) {
             throw err
         } finally {
-          step([$class: 'JUnitResultArchiver', testResults: '**/target/surefire-reports/TEST-*.xml'])
+            step([$class: 'JUnitResultArchiver', testResults: '**/target/surefire-reports/TEST-*.xml'])
         }
     }
 
@@ -46,7 +46,7 @@ node {
         } catch(err) {
             throw err
         } finally {
-            step([$class: 'JUnitResultArchiver', testResults: '**/target/test-results/karma/TESTS-*.xml'])
+            step([$class: 'JUnitResultArchiver', testResults: '**/build/**/TEST-*.xml'])
         }
     }
 
@@ -58,7 +58,11 @@ node {
         } catch(err) {
             throw err
         } finally {
+            <%_ if (buildTool === 'maven') { _%>
             step([$class: 'JUnitResultArchiver', testResults: '**/target/test-results/karma/TESTS-*.xml'])
+            <%_ } else if (buildTool === 'gradle') { _%>
+            step([$class: 'JUnitResultArchiver', testResults: '**/build/test-results/karma/TESTS-*.xml'])
+            <%_ } _%>
         }
     }
 
--- a/generators/server/templates/_Jenkinsfile
+++ b/generators/server/templates/_Jenkinsfile
@@ -26,7 +26,13 @@ node {
     }
 
     stage('backend tests') {
-        sh "./mvnw test"
+        try {
+            sh "./mvnw test"
+        } catch(err) {
+            throw err
+        } finally {
+          step([$class: 'JUnitResultArchiver', testResults: '**/target/surefire-reports/TEST-*.xml'])
+        }
     }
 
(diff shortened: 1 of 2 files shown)


generators/server/templates/_Jenkinsfile
```diff
@@ -35,13 +41,25 @@ node {
     }
 
     stage('backend tests') {
-        sh "./gradlew test"
+        try {
+            sh "./gradlew test"
+        } catch(err) {
+            throw err
+        } finally {
+            step([$class: 'JUnitResultArchiver', testResults: '**/target/test-results/karma/TESTS-*.xml'])
```
deepu105: Are you sure its in target folder?  For gradle we use the build folder

generators/server/templates/_Jenkinsfile
```diff
@@ -35,13 +41,25 @@ node {
...
 <%_ } _%>
 <%_ if (applicationType !== 'microservice') { _%>
     stage('frontend tests') {
-        sh "gulp test"
+        try {
+            sh "gulp test"
+        } catch(err) {
+            throw err
+        } finally {
+            step([$class: 'JUnitResultArchiver', testResults: '**/target/test-results/karma/TESTS-*.xml'])
```
deepu105: Same here if gradle we need to see if its in build

jdubois:
Great!

generators/server/templates/_Jenkinsfile
raphaelbrugier: I have tested it with Gulp, the tests are reported correctly:

screen shot 2016-11-17 at 9 17 35 am

deepu105: when gradle is used as build as well?

pascalgrimaud:
Just tested on my current Maven project, it works well! I didn't test with Gradle
Thanks @raphaelbrugier

raphaelbrugier:
I've pushed the Gradle support.

There is, however, a difference with Maven. When the tests fail this does not stop the build (and the next stages).
This is because the `gradle test` return a `0` status code when the tests are failing.
This is explicitly configured in the [build.gradle](https://github.com/jhipster/jhipster-sample-app-gradle/blob/master/build.gradle#L56).

What was the reason to explicitly not stop the build in case of failure?

atomfrede:
@raphaelbrugier A very good question :) It was introduced here: https://github.com/jhipster/generator-jhipster/commit/c5fef9490175666d5130f1481ce0f6c315cefc3b 

The reason seems to be, that when one test fails no report is generated: https://github.com/jhipster/generator-jhipster/pull/2771

Will have look if that (still) true.

raphaelbrugier:
I tried again with Gradle, but I did not reproduce the problem described in the #2771 issue.

I explicitly set `ignoreFailures false` and ran ./gradlew test with a failing test.
See [here](https://github.com/raphaelbrugier/jhipster-pipeline/tree/eba143601a3d4c6334c1584671011c8150e7ca15)  for my test project.
All reports were created in the `./build/test-results/` folder.

atomfrede:
Removing ignore failures seems good to me.

pascalgrimaud:
I would prefer to comment the line + adding information about this + link to the 2 tickets (this PR and #2771), instead of removing it completely

atomfrede:
+1

pascalgrimaud:
@raphaelbrugier : I found an issue with the Jenkinsfile for microservice. Jenkins2 didn't checkout the project

Can you move this code:
https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/_Jenkinsfile#L14-L16
just before this line https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/_Jenkinsfile#L2 ?

Otherwise, I will do it after this PR is merged

pascalgrimaud:
LGTM

jdubois:
Great, I'm merging this!

PR #4485, opened 2016-11-17
fix #4483 Session token is not persisted in Cassandra

### LINKED ISSUE #4483: [BUG] Cassandra - Session token is not persisted in the persistent_token table

When using Cassandra, the session token is not persisted in the `persistent_token` table.
The table exists and the CustomPersistentRememberMeServices tries to read the token and delete it, but never saves it.
With a SQL database, the token is persisted in the `onLoginSuccess` method.

NB: The session mechanism works, fine. It's just the token is never persisted and those can not be invalidated.

##### **Reproduce the error**

- Create a JHipster monolith application with Cassandra and a "Session" security.
- Log in the app.
- See the `persistent_token` table is still empty.

##### **Suggest a Fix**

I'm not sure what is the expected behavior, having the token working with Cassandra - in that case it should be saved - or not working and JHipster should not generate Token related code.

I could propose a PR based on the expected behavior.

##### **JHipster Version(s)**

3.11

```
{
  "generator-jhipster": {
    "jhipsterVersion": "3.11.0",
    "baseName": "jhisecurity",
    "serverPort": "8080",
    "hibernateCache": "no",
    "websocket": false,
    "databaseType": "cassandra",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "useSass": false,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": false,
    "jwtSecretKey": "9eb2dba9a24dad2315af2d3443d68016af60236c",
    "rememberMeKey": "a1fff3e15cb7761550e69b712b3919bdf470a624"
  }
}
```

raphaelbrugier:
- store the token
- delete the token in both persistent_token and persistent_token_by_user tables

I changed the mapping of the `persistent_token_by_user` table to store multiple `series` per user, I guess this was the intended behavior @jdubois ?

### DIFF
--- a/generators/server/templates/src/main/java/package/repository/_PersistentTokenRepository.java
+++ b/generators/server/templates/src/main/java/package/repository/_PersistentTokenRepository.java
@@ -46,6 +46,8 @@ public class PersistentTokenRepository {
 
     private PreparedStatement insertPersistentTokenStmt;
 
+    private PreparedStatement deletePersistentTokenSeriesByUserIdStmt;
+
     @PostConstruct
     public void init() {
         mapper = new MappingManager(session).mapper(PersistentToken.class);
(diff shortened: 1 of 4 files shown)


jdubois:
Yes, one user could be logged in with several browsers.

jdubois:
I will test next week -> I also want to upgrade our Cassandra version to 3.x, so I'll do everything at the same time

raphaelbrugier:
Sounds good to me, I have a couple of improvements from our current project to propose for Cassandra as well.

PR #4503, opened 2016-11-18
rename Angular 2 HTML component files

### LINKED ISSUE #4499: Rename Angular 2 HTML component files

ATM with the Angular branch, component files are named :

```
├── activate
│   ├── activate.component.ts
│   ├── activate.html
├── password
│   ├── password.component.ts
│   ├── password.html
```

Following the Angular CLI standards, the HTML files should be called the name of the component with an `html` suffix : 

```
├── activate
│   ├── activate.component.ts
│   ├── activate.component.html
├── password
│   ├── password.component.ts
│   ├── password.component.html
```

When using the Angular CLI with the following commands : 

```
ng g component activate
ng g component password
```

We get the following : 

```
├── activate
│   ├── activate.component.ts
│   ├── activate.component.html
├── password
│   ├── password.component.ts
│   ├── password.component.html
```

Sticking to Angular CLI would be better

agoncal:
HTML component files should be called the name of the component with an html suffix

Fix #4499


generators/generator-base.js
```diff
@@ -41,7 +41,7 @@ util.inherits(Generator, yeoman.Base);
  */
 Generator.prototype.addElementToMenu = function (routerName, glyphiconName, enableTranslation) {
     try {
-        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.html';
+        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.component.html';
```
deepu105: this shouldnt be changed now. It will be have to work for both ng1 and ng2 hence will be changed after entity migration

generators/generator-base.js
```diff
@@ -67,7 +67,7 @@ Generator.prototype.addElementToMenu = function (routerName, glyphiconName, enab
  */
 Generator.prototype.addElementToAdminMenu = function (routerName, glyphiconName, enableTranslation) {
     try {
-        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.html';
+        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.component.html';
```
deepu105: same case

generators/generator-base.js
```diff
@@ -529,8 +529,8 @@ Generator.prototype.addColumnToLiquibaseEntityChangeset = function (filePath, co
  */
 Generator.prototype.addSocialButton = function (isUseSass, socialName, socialParameter, buttonColor, buttonHoverColor) {
     var socialServicefullPath = CLIENT_MAIN_SRC_DIR + 'app/account/social/social.service.js';
-    var loginfullPath = CLIENT_MAIN_SRC_DIR + 'app/account/login/login.html';
-    var registerfullPath = CLIENT_MAIN_SRC_DIR + 'app/account/register/register.html';
+    var loginfullPath = CLIENT_MAIN_SRC_DIR + 'app/account/login/login.component.html';
```
deepu105: same case

generators/server/files.js
```diff
@@ -142,7 +142,7 @@ function writeFiles() {
             }
 
             // Thymeleaf templates
-            this.copy(SERVER_MAIN_RES_DIR + 'templates/error.html', SERVER_MAIN_RES_DIR + 'templates/error.html');
+            this.copy(SERVER_MAIN_RES_DIR + 'templates/error.component.html', SERVER_MAIN_RES_DIR + 'templates/error.component.html');
```
deepu105: should not be changed

test/test-creation.js
```diff
@@ -53,7 +53,7 @@ const expectedFiles = {
         SERVER_MAIN_RES_DIR + 'banner.txt',
         SERVER_MAIN_RES_DIR + 'ehcache.xml',
         SERVER_MAIN_RES_DIR + '.h2.server.properties',
-        SERVER_MAIN_RES_DIR + 'templates/error.html',
+        SERVER_MAIN_RES_DIR + 'templates/error.component.html',
```
deepu105: the filenames in this test file should not be changed as they are for ng1

generators/server/files.js
deepu105: the file shouldnt be renamed as well as its a thymeleaf template

generators/generator-base.js
```diff
@@ -92,7 +92,7 @@ Generator.prototype.addElementToAdminMenu = function (routerName, glyphiconName,
  */
 Generator.prototype.addEntityToMenu = function (routerName, enableTranslation) {
     try {
-        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.html';
+        var fullPath = CLIENT_MAIN_SRC_DIR + 'app/layouts/navbar/navbar.component.html';
```
deepu105: same

generators/client-2/files.js
```diff
@@ -240,27 +240,27 @@ const files = {
...
                 'admin/logs/_logs.service.ts',
-                { file: 'admin/logs/_logs.html', method: 'copyHtml' },
+                { file: 'admin/logs/_logs.component.html', method: 'copyHtml' },
                 { file: 'admin/logs/_logs.state.ts', method: 'copyJs' },
                 'admin/metrics/_metrics.component.ts',
                 'admin/metrics/_metrics-modal.component.ts',
                 'admin/metrics/_metrics.service.ts',
-                { file: 'admin/metrics/_metrics.html', method: 'copyHtml', template: true },
+                { file: 'admin/metrics/_metrics.component.html', method: 'copyHtml', template: true },
                 { file: 'admin/metrics/_metrics-modal.html', method: 'copyHtml', template: true },
```
deepu105: this can be renamed as well

generators/client-2/files.js
```diff
@@ -240,27 +240,27 @@ const files = {
...
                 'admin/docs/_docs.component.ts',
-                'admin/docs/_docs.html',
+                'admin/docs/_docs.component.html',
                 'admin/docs/_docs.state.ts',
                 'admin/health/_health.component.ts',
                 'admin/health/_health-modal.component.ts',
                 'admin/health/_health.service.ts',
-                { file: 'admin/health/_health.html', method: 'copyHtml' },
+                { file: 'admin/health/_health.component.html', method: 'copyHtml' },
                 { file: 'admin/health/_health-modal.html', method: 'copyHtml' },
```
deepu105: this as well

generators/generator-base.js
agoncal: Oh, so the idea is to have Angular 1 and 2 both running in JHipster 4.0 ? How can this work in both cases (NG1 / NG2) if we rename the file in NG2 but not in NG1 ? If we want to keep both working, that would mean to rename the files also in NG1, no ?

deepu105: The idea is to support generation of both ng1 and ng2 apps. so there will be question to choose version and based on that either client-1 or client-2 will be used. client 1 is ng1 as is today and wouldnt require file renames. client-2 is ng2 and hence the common methods like these will have an if else or something to support both.

deepu105: just renaming wouldnt work as the code in both version would be slightly different as well, so we would have to do conditions here anyway

deepu105: yes. we have the angularVersion variable on the generator context for that, so while generating an app or entity we will know which angular version is being used. its similar to how we do all other options today. So it will be a simple if else or even a fresh new method for ng2 if required. But this can be done later in any case

agoncal: @deepu105 But just to be sure. Will Angular 1 will be supported in JHipster 4 ? I thought it would disappear (or will it stay but in BETA ?)

deepu105:
It will stay along side angular2 :) atleast till JHipster 5


Thanks & Regards,
Deepu

On Mon, Nov 21, 2016 at 1:17 PM, Antonio Goncalves <notifications@github.com

test/test-creation.js
```diff
@@ -1386,6 +1498,8 @@ describe('JHipster client-2 generator', function () {
 
         it('creates expected files for default configuration for client-2 generator', function () {
             assert.noFile(expectedFiles.server);
+            assert.noFile(expectedFiles.serverNG1);
+            assert.noFile(expectedFiles.serverNG2);
             assert.noFile(expectedFiles.maven);
             assert.file(expectedFiles.i18nJson);
             assert.file(getFilesForOptions(require('../generators/client-2/files').files, {
```
deepu105: @agoncal we dont need to maintain a list of files for ng2 and ng1. for ng2 im building the list on the fly here and soon ill do the same for ng1 as well so for now you dont need to change anything on this test files. it should work as it is

test/test-creation.js
```diff
@@ -187,6 +194,40 @@ const expectedFiles = {
         CLIENT_MAIN_SRC_DIR + 'i18n/fr/user-management.json'
     ],
 
+    clientNG1 : [
```
deepu105: We dont have to do this. see my comment below

generators/server/files.js
```diff
@@ -142,7 +142,11 @@ function writeFiles() {
             }
 
             // Thymeleaf templates
-            this.copy(SERVER_MAIN_RES_DIR + 'templates/error.html', SERVER_MAIN_RES_DIR + 'templates/error.html');
+            if (this.angularVersion === 'angular1') {
+                this.copy(SERVER_MAIN_RES_DIR + 'templates/error.html', SERVER_MAIN_RES_DIR + 'templates/error.html');
+            } else {
+                this.copy(SERVER_MAIN_RES_DIR + 'templates/error.component.html', SERVER_MAIN_RES_DIR + 'templates/error.component.html');
```
agoncal: Sorry @deepu105, so can you quickly explains me what `test-creation.js` and `files.js` are for then ?

deepu105:
@agoncal I'm gonna polish few minor stuff here and merge this hope its ok with you :)

generators/server/files.js
deepu105: test-creation is to verfiy if the files required are generated for an app and files.js actually decides what are the files that needs to be generated for a app config

PR #4509, opened 2016-11-19
Add Prometheus metrics support 

### LINKED ISSUE #4445: [FEATURE] Prometheus support (Metrics)
I think the metrics aspect of JHipster is not put forward enough as this stuff is really cool. Our official monitoring solution is the JHipster Console (of which I'm releasing a v2 and not planning to abandon) but it might not fit everyone use case (for exemple if you already have a log management solution) so it might be interesting to offer an alternative.

In the mean time some new technologies around metrics have come up recently like Influxdb and Prometheus.
It seems that there is already an InfluxDB reporter in dropwizard metrics however it is not yet released although the PR has been merged for more than a year ! However we could still integrate it using a third party lib like [davidB/metrics-influxdb](https://github.com/davidB/metrics-influxdb)

However, what I'm more interested in is Prometheus (https://prometheus.io/) which has been conceived as a monitoring solution for microservices. Main advantages for this solution :
- Fully open source unlike the offering from InfluxData (features like clustering only available in the proprietary version of InfluxDB)
- Easily scalable, retrieve metrics by polling an endpoint on the service rather than receiving data over a socket like most other systems. Also support consul service discovery for finding nodes.
- Good UI
- Alerting
- De facto standard for Kubernetes monitoring

So I'm suggesting that we do the following:
- [x] Remove the Spark Exporter. I'm sure no one uses it and it is not developped anymore by its author (@ahars a collegue at IpponTech)
- [x] Configure a prometheus exporter using the official client ([prometheus/client_java](https://github.com/prometheus/client_java)) that support dropwizard out of the box.
- [x] Add an option to the docker-compose subgen to setup prometheus

See my demo: [PierreBesson/jhipster-prometheus-demo](https://github.com/PierreBesson/jhipster-prometheus-demo)

PierreBesson:
Fixes #4445

- add a property to expose metrics as a prometheus endpoint
- add an option to the docker-compose subgen to setup prometheus monitoring for one or several apps.


package.json
```diff
@@ -40,15 +40,16 @@
...
     "mkdirp": "0.5.1",
-    "shelljs": "0.7.4",
     "pluralize": "3.0.0",
-    "yeoman-generator": "0.24.1",
-    "js-yaml": "3.6.1",
-    "yo": "1.8.5",
     "semver": "5.3.0",
-    "jhipster-core": "1.2.0"
+    "shelljs": "0.7.4",
+    "uuid": "3.0.0",
```
PierreBesson: Here I simply added a dependency on the uuid lib but npm reordered them for me.
I need UUID to setup sample alerting webhook on [http://webhook.site](http://webhook.site).
I will add something similar for ELK soon.

generators/docker-compose/files.js
```diff
@@ -31,11 +31,29 @@ function writeFiles() {
         },
 
         writeElkFiles: function() {
-            if(!this.useElk) return;
+            if(this.monitoring !== 'elk') return;
```
deepu105: what about backward compatibility?

package.json
deepu105: ok no probs

deepu105:
Oh ok


Thanks & Regards,
Deepu

PierreBesson:
@pascalgrimaud I don't understand why Travis is not happy. Can you look into it when you have the time ?

pascalgrimaud:
@PierreBesson : yes I will have a look as soon as I got the logs

ruddell:
Julien fixed the problem https://github.com/jhipster/generator-jhipster/commit/70adc5f414e954b5387487a5e6dca01609b1072f

jdubois:
Yes, that was my fault!!!

PierreBesson:
One last thing. I think I should precise that Prometheus Monitoring works only with apps >= 3.12...

PierreBesson:
I have added some precision in the question for now. Once 3.12 is released maybe I should introduce a version check as well.

PierreBesson:
Don't merge this yet I noticed that the UUID for alerting doesn't actually work... You can't just generate the UUID yourself, you have to actually request the creation of a new UUID endpoint on webhook.site by POSTing on 'http://webhook.site/token'.
So I guess I will need to remove the uuid stuff.

PierreBesson:
I removed the UUID stuff but it is still easy to setup webhook, slack, etc...
Also alerts will show up in the alerting UI :

capture du 2016-11-21 23-21-42

I think this is ready be merged now. The next step will be to write some docs.

pascalgrimaud:
good job @PierreBesson !
Yes, need documentation plz :-)

cbornet:
I think some people may want to use ELK for their logs and Prometheus for the metrics. Shouldn't we support this ?

cbornet:
(Some people would include me)

jdubois:
Yes, for me logging and metrics are two different things

PierreBesson:
@cbornet, well this is already possible... I could add a specific option to setup this with the docker-compose subgen but it will add another question.

PR #4512, opened 2016-11-19
Yarn saved in config

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/4317

Steps:
1) create folder, then, `yarn link generator-jhipster` or `npm link generator-jhipster`
2) generate project: `yo jhipster --yarn` -> the config will save the option in `.yo-rc.json`
3) regenerate project: `yo jhipster --force` -> it will use the config, so yarn

If you choose maven, the `pom.xml` will use the frontend-maven-plugin with yarn
For gradle [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/gradle/_profile_prod.gradle#L55), it will still use npm install because the `com.moowork.node` doesn't support yarn yet

I tested with node-sass too

> Travis official support for Yarn is merged and will be activated on monday

So I will update our Travis build in another PR, next week


jdubois:
Great! Can't wait until Monday, to see the performance difference!!

PR #4516, opened 2016-11-20
fix #4513 Cassandra tests improvements

### LINKED ISSUE #4513: [FEATURE] Cassandra tests improvements
I'd like to propose two improvements for the Cassandra tests.

## 1/ Start Cassandra unit only once for all the tests.
Currently, the cluster is started before each test class.
After starting, all CQL scripts are applied to prepare the tables.
We found in our current project that it was a very expensive operation to do for every test class.

We reduced our tests duration by starting the cassandra-unit cluster only once, using a simple boolean. see [here](https://github.com/raphaelbrugier/jhipster-experiment/blob/cassandra-rdm-port/src/test/java/com/mycompany/myapp/AbstractCassandraTest.java#L37).

## 2/ Start the cassandra-unit embedded cluster on a random port
Currently, it's not possible to run multiple continuous integration jobs in parallel because Cassandra-unit starts on the default port 9042.
Only one job will pass and the others will fail with the error:
`org.apache.cassandra.exceptions.ConfigurationException: /127.0.0.1:7010 is in use by another process.  Change listen_address:storage_port in cassandra.yaml to values that do not conflict with other services`

Cassandra-unit can start on a random port since jsevellec/cassandra-unit#119

To use this feature I needed to do the following:
- copy a custom configuration file for Cassandra-unit because of this bug: jsevellec/cassandra-unit#176
- start Cassandra-unit on a random port before starting the Spring Context. [see here](https://github.com/raphaelbrugier/jhipster-experiment/blob/cassandra-rdm-port/src/test/java/com/mycompany/myapp/AbstractCassandraTest.java#L42)
- add a spring 'Test' profile to override the Cassandra Configuration for the tests. [see here](https://github.com/raphaelbrugier/jhipster-experiment/blob/cassandra-rdm-port/src/main/java/com/mycompany/myapp/config/Constants.java#L11)
- return the random port to instantiate the Cassandra client. [see here](https://github.com/raphaelbrugier/jhipster-experiment/blob/cassandra-rdm-port/src/test/java/com/mycompany/myapp/config/CassandraTestConfiguration.java#L26)

The main problem was to inject the Random port into the Cassandra cluster client configuration. Ideas welcome for a better way ;)

See [this commit](https://github.com/raphaelbrugier/jhipster-experiment/commit/899d8c1bc3bba72508378e7cff9f4cdb38dfcb88) for an overall idea of the changes.

If these features and implementations seem good to you, I'll submit a PR.

raphaelbrugier:
* Start Cassandra unit only once for all the tests.
* Start the cassandra-unit embedded cluster on a random port

I used 0 port as the random port as per Julien's comment

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_Constants.java
+++ b/generators/server/templates/src/main/java/package/config/_Constants.java
@@ -8,6 +8,7 @@ public final class Constants {
     //Regex for acceptable logins
     public static final String LOGIN_REGEX = "^[_'.@A-Za-z0-9-]*$";
     // Spring profile for development and production, see http://jhipster.github.io/profiles/
+    public static final String SPRING_PROFILE_TEST = "test";
     public static final String SPRING_PROFILE_DEVELOPMENT = "dev";
     public static final String SPRING_PROFILE_PRODUCTION = "prod";
     // Spring profile used when deploying with Spring Cloud (used when deploying to CloudFoundry)
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -473,6 +473,8 @@ function writeFiles() {
             if (this.databaseType === 'cassandra') {
                 this.template(SERVER_TEST_SRC_DIR + 'package/_CassandraKeyspaceUnitTest.java', testDir + 'CassandraKeyspaceUnitTest.java', this, {});
                 this.template(SERVER_TEST_SRC_DIR + 'package/_AbstractCassandraTest.java', testDir + 'AbstractCassandraTest.java', this, {});
+                this.template(SERVER_TEST_SRC_DIR + 'package/config/_CassandraTestConfiguration.java', testDir + 'config/CassandraTestConfiguration.java', this, {});
+                this.template(SERVER_TEST_RES_DIR + '_cassandra-rdmport.yml', SERVER_TEST_RES_DIR + 'cassandra-rdmport.yml', this, {});
             }
 
             this.template(SERVER_TEST_SRC_DIR + 'package/web/rest/_TestUtil.java', testDir + 'web/rest/TestUtil.java', this, {});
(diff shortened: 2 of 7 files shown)


generators/server/templates/src/main/java/package/config/_Constants.java
```diff
@@ -8,6 +8,7 @@
     //Regex for acceptable logins
     public static final String LOGIN_REGEX = "^[_'.@A-Za-z0-9-]*$";
     // Spring profile for development and production, see http://jhipster.github.io/profiles/
+    public static final String SPRING_PROFILE_TEST = "test";
```
jdubois: I don't like to have a specific "test" profile, which would be only for Cassandra. Indeed, we could create one, but not just for this specific case.

raphaelbrugier: You cannot set the port in the file, it is randomly picked by cassandra-unit when it starts.
That's the point of this feature, to start on a random port so it does not conflict between multiple CI jobs.

jdubois: I mean putting "0" directly in the `application.yml` file should already work - I don't understand the issue

jdubois: If that's just for the `CassandraTestConfiguration` I wouldn't bother with this: just put the `getPort` method in `CassandraConfiguration`... You have a check for `0` that can only be useful, there's no need to have a more complex configuration

raphaelbrugier: Yes, that was my initial idea, but then the problem is you can not access the `DatabaseDescriptor.getNativeTransportPort();` method.
It's part of the cassandra-all lib, and only visible in the test scope.
Which makes sense because it's pulled by cassandra-unit only for the tests.

I did not find another way to discover the port from the java driver.

The only alternative I can think of is to have AbstractCassandraTest modifying the configuration properties passed before the Spring context is started. But I have no clue how to do this.

jdubois: Indeed... Then maybe it's time to have that profile, I guess many people add it anyway. I'll have a closer look today, and will probably merge this.

jdubois: I'm also going to rename `rdmport` to `random-port` as I didn't understand what it meant at first sight

raphaelbrugier: Sounds good!

PR #4529, opened 2016-11-22
makes the SecurityUtil more generic to Authentication

### LINKED ISSUE #4526: SecurityUtils#isCurrentUserInRole doesn't work in a uaa context

When you use `SecurityUtils.isCurrentUserInRole` in a uaa context the respond will be always false.

##### **Reproduce the error**

Create a microservice and a uaa server, define an behavior for a particular role.
```
if (SecurityUtils.isCurrentUserInRole(AuthoritiesConstants.ADMIN)) {
            return 1;
}
return 0;
```
It will return 0;

##### **Suggest a Fix**

[SecurityUtils#L74](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/security/_SecurityUtils.java#L74)
If the Principal is not an instance of `UserDetails` it will return false. In a UAA context, the principal is an instance of `String` (the user login), so by default the respond is false.

This is the naive fix I use:
```
public static boolean isCurrentUserInRole(String authority) {
        SecurityContext securityContext = SecurityContextHolder.getContext();
        Authentication authentication = securityContext.getAuthentication();
        if (authentication != null) {
            Collection<? extends GrantedAuthority> authorities = authentication.getAuthorities();
            for (GrantedAuthority grantedAuthority : authorities) {
                if (grantedAuthority.getAuthority().equals(authority)) {
                    return true;
                }
            }
        }
        return false;
    }
```

I didn't test it with a classic JWT (or monolith) architecture so I don't know if it can be back-ported.
Otherwise, we need to obtain a UserDetail instance from the `OAuth2Authentication`.

xetys:
so the static function are usable in all kinds of authentications, supported by JHipster

also changed the class to Java 8 Stream API usage instead of explicit loops

fix #4526

### DIFF
--- a/generators/server/templates/src/main/java/package/security/_SecurityUtils.java
+++ b/generators/server/templates/src/main/java/package/security/_SecurityUtils.java
@@ -46,15 +46,8 @@ public final class SecurityUtils {
         SecurityContext securityContext = SecurityContextHolder.getContext();
         Authentication authentication = securityContext.getAuthentication();
         if (authentication != null) {
-            Collection<? extends GrantedAuthority> authorities = authentication.getAuthorities();
-            if (authorities != null) {
-                for (GrantedAuthority authority : authorities) {
-                    if (authority.getAuthority().equals(AuthoritiesConstants.ANONYMOUS)) {
-                        return false;
-                    }
-                }
-            }
-            return true;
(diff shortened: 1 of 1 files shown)


jdubois:
Looks much cleaner and more readable than my original code!!!

PR #4530, opened 2016-11-22
Fixing callback url when deployed behind reverse proxy

lrkwz:
When deployed behind a reverse proxy, the application pushed towards the social (google, facebook, ...) a callback url computed on the request: which normally leads to http://localhost:8080/signin/google behing refused by the social app as non granted.

Both ```ConnectController``` and  ```ProviderSigninController``` accept the applicationUrl parameter which can be declared in Environment as spring.application.url. 
See http://docs.spring.io/spring-social/docs/current/reference/htmlsingle/#creating-connections-with-connectcontroller

### DIFF
--- a/generators/server/templates/src/main/java/package/config/social/_SocialConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/social/_SocialConfiguration.java
@@ -48,7 +48,7 @@ public class SocialConfiguration implements SocialConfigurer {
             ConnectionRepository connectionRepository) {
 
         ConnectController controller = new ConnectController(connectionFactoryLocator, connectionRepository);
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/config/social/_SocialConfiguration.java
```diff
@@ -24,6 +26,7 @@
 import org.springframework.social.twitter.connect.TwitterConnectionFactory;
 // jhipster-needle-add-social-connection-factory-import-package
 
+
```
jdubois: Can you remove this extra line?

lrkwz: of course

jdubois:
Thanks!

Issue #4356, opened 2016-10-19
Many-To-Many dto compile error


beaussart:
##### **Overview of the issue**

When there is a Many-To-Many of an entity with himself, the mapstruct dto fail to compile. The mapper throw a `java.lang.StackOverflowError` when doing so.

##### **Motivation for or Use Case**

I need for a project to have a entity Step requiring previous Steps to be done before and I would like a dto for the rest endpoint.

##### **Reproduce the error**

Just create a entity with many-to-many with himself and with dto.
##### **Suggest a Fix**

I think the problem is comming from the mapstruct dto himself, trying to map a looping graph.
##### **JHipster Version(s)**

```
jhipster-test@0.0.0 /home/beaussan/jhipsterTest
└── generator-jhipster@3.9.1 

```
##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

``` yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "3.9.1",
    "baseName": "jhipsterTest",
    "packageName": "jh.test",
    "packageFolder": "jh/test",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "useSass": false,
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
```
##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

Step.json

``` yaml
{
    "fluentMethods": true,
    "relationships": [
        {
            "relationshipName": "requiredSteps",
            "otherEntityName": "step",
            "relationshipType": "many-to-many",
            "otherEntityField": "id",
            "ownerSide": true,
            "otherEntityRelationshipName": "step"
        }
    ],
    "fields": [
        {
            "fieldName": "isDone",
            "fieldType": "Boolean"
        }
    ],
    "changelogDate": "20161019133829",
    "dto": "mapstruct",
    "service": "no",
    "entityTableName": "step",
    "pagination": "no"
}


```
##### **Browsers and Operating System**

Linux, Arch

jdubois:
[...]

PR #4539, opened 2016-11-23
Upgrading to MapStruct 1.1.0.Final

gunnarmorling:
Updated the versions of MapStruct I found. I hope I didn't miss anything?

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -50,7 +50,7 @@ mariadb_java_client_version=1.4.4
 h2_version=1.4.188
 <% if (testFrameworks.indexOf('gatling') != -1) { %>
 gatling_version=2.2.0<% } %>
-mapstruct_version=1.1.0.CR2<% if (enableSocialSignIn) { %>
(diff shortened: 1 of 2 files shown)


deepu105:
@gunnarmorling thanks. You could also consider to help us to get mapstruct support out of beta :stuck_out_tongue_winking_eye:

gunnarmorling:
I'll try to do the best I can :) I'm not sure though what would be needed
to make it happen. Anything you miss from MapStruct in order to call your
support final?

deepu105:
I guess for now its mostly quirks in IDE and some issues with handling complex relationships @jdubois might know better

jdubois:
Yes we have 2 kinds of issues:

IDE configuration: I've spent a lot of time on this, I have no idea how to do it correctly. You can have a look at https://jhipster.github.io/configuring-ide/ (look for the "IDE profile" at the bottom), and you can contribute changes [here](https://github.com/jhipster/jhipster.github.io).

And we also have mapping issues, like [here](https://github.com/jhipster/generator-jhipster/issues/4356), but this isn't really a MapStruct issue, it's mostly us having trouble to use it correctly, with some corner cases.

PR #4541, opened 2016-11-24
Upgrading to Hibernate Validator 5.3.3.Final

gunnarmorling:
This updates HV to the latest and greatest. I'm not 100% sure about the POM change, in that case you seem to get the HV version from the Spring Boot dependencies POM. I can rework or split it up as needed.

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -27,7 +27,7 @@ liquibase_slf4j_version=2.0.0
 liquibase_core_version=3.4.2
 liquibase_hibernate5_version=3.6<% } %><% if (databaseType == 'mongodb') { %>
 mongobee_version=0.10<% } %>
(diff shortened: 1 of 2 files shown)


jdubois:
Yes, we normally get it from Spring Boot, and don't upgrade unless it solves a specific issue.... Then I don't know why it's set up in Gradle, but not in Maven!

gunnarmorling:
Right, that's what's unclear to me as well.

jdubois:
Then it's good to have this fixed, and having a newer version is probably good as we also upgrade Hibernate in the next release

gunnarmorling:
Sweet, thanks!

gunnarmorling:
Hey, before my change you linked above, JHipster would use the HV as configured by Spring Boot with Maven but use a specific version with Gradle. So I changed it to use the same version with Maven, too. I think you need the variable reference for Gradle with the current kind of set-up.

ctamisier:
ok indeed, so I added:
`compile "org.hibernate:hibernate-validator:${hibernate_validator_version}"` in the PR

Issue #4545, opened 2016-11-24
javadoc generation for test classes failes


kaidohallik:
##### **Overview of the issue**

Error occurs on javadoc generation with test classes (modified build.gradle for that, described later).

Stack trace:
```
.....
:javadoc
C:\projects\kaido-github\jhipster-test-global\src\main\java\com\mycompany\myapp\config\DefaultProfileUtil.java:39: warning: no @param for env
    public static String[] getActiveProfiles(Environment env) {
                           ^
C:\projects\kaido-github\jhipster-test-global\src\main\java\com\mycompany\myapp\config\DefaultProfileUtil.java:39: warning: no @return
    public static String[] getActiveProfiles(Environment env) {
                           ^
javadoc: error - com.sun.tools.doclets.internal.toolkit.util.DocletAbortException: com.sun.tools.doclets.internal.toolkit.util.DocletAbortException: com.sun.tools.do
clets.internal.toolkit.util.DocletAbortException: com.sun.tools.javac.code.Symbol$CompletionFailure: class file for javax.interceptor.InterceptorBinding not found
1 error
2 warnings
:javadoc FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':javadoc'.
> Javadoc generation failed. Generated Javadoc options file (useful for troubleshooting): 'C:\projects\kaido-github\jhipster-test-global\build\tmp\javadoc\javadoc.op
tions'

* Try:
Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':javadoc'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:84)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:61)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:58)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:88)
[...]

PR #4546, opened 2016-11-25
fix test classes javadoc generation

kaidohallik:
fix jhipster/generator-jhipster#4545


jdubois:
I'm just using my phone, but reading your code I don't think this fixes the issue

kaidohallik:
I agree that when reading code changes it seems unbelievable that this can fix the issue, but in fact this fixes.
It was not easy to figure out what causes the problem but after fixing this and pondering over the issue all is logical.

jdubois:
Thanks!!!! I'll have a closer look - let me merge this (I had the same kind
of issue last week, so I should know).

PR #4559, opened 2016-11-26
Travis: remove app from cache

pascalgrimaud:
Plz dont merge yet


pascalgrimaud:
Looking every logs here https://travis-ci.org/jhipster/generator-jhipster/builds/179110047
it seems the gulp test is completely back, so I'm merging this

PR #4561, opened 2016-11-27
fix entity javadoc and API description

kaidohallik:
fix API description error (string in multiple lines) if entity javadoc contains \n
remove empty lines from generated entity javadoc

fix jhipster/generator-jhipster#4309


jdubois:
Needs to be tested, but that's definitely some nice coding in our internal APIs, great work @kaidohallik ! I hope we can merge this soon.

kaidohallik:
I updated PR, fixed problems in function formatAsApiDescription in file generator-base.js:
- fix eslint error: newline not allowed in regexp
- discard empty rows
- don't add space between rows if markup is used

PR #4562, opened 2016-11-27
Implement additional module installation step in the client generator

gzsombor:
So in the app generator, the generator asks for additional jhipster-generator to install, so it's more easy to discover them.

Open questions, how to detect if the version supports the current one?

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -2,12 +2,14 @@
 
 var chalk = require('chalk');
 
+
 module.exports = {
     askForInsightOptIn,
     askForApplicationType,
     askForModuleName,
     askFori18n,
-    askForTestOpts
+    askForTestOpts,
+    askForMoreModules
 };
 
 function askForInsightOptIn() {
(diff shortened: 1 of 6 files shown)


deepu105:
Why is this done on client generator? Shouldn't this be done on the app generator step. You can still use the variable in client gen for package.json

gzsombor:
When the client generator is not called, there wont be any 'package.json', so it seemed to put there the questions.

deepu105:
Yes but we already have a ticket to add package json for even server only
apps thats why and having this in the app generator would be better once we
have angular2 merged to master

Thanks & regards,
Deepu

gzsombor:
I've moved the code into the 'app' generator, but travis doesn't like it

generators/app/prompts.js
```diff
@@ -125,3 +128,75 @@ function askForTestOpts() {
         done();
     }.bind(this));
 }
+
+function httpGet(url, onSuccess, onFail) {
```
deepu105: since this could be used in the future can we move it to generator-base?

generators/app/prompts.js
```diff
@@ -125,3 +128,75 @@ function askForTestOpts() {
...
+            body += chunk;
+        });
+        res.on('end', function() {
+            onSuccess(body);
+        });
+    }).on('error', onFail);
+}
+
+function askModulesToBeInstalled(done, generator) {
+    httpGet('http://npmsearch.com/query?fields=name,keywords,description,author,homepage,version,repository,created&q=keywords:jhipster-module&start=0&size=50',
```
deepu105: we can remove some of the unused params to make payload lighter

generators/app/prompts.js
```diff
@@ -125,3 +128,75 @@ function askForTestOpts() {
...
+                        generator.otherModules.push({name:module.name[0], version:module.version[0]});
+                    });
+                    done();
+                }.bind(generator));
+            } else {
+                done();
+            }
+        },
+        function (error) {
+            this.warning('Unable to contact server for modules: '+error.message);
```
deepu105: can we change to `Unable to contact server to fetch additional modules: ${error.message}` using backtick notation

generators/app/prompts.js
```diff
@@ -125,3 +128,75 @@ function askForTestOpts() {
...
+                    message: 'Which other modules would you like to use?',
+                    choices: choices,
+                    default: []
+                }).then(function (prompt) {
+                    // [ {name: [moduleName], version:[version]}, ...]
+                    generator.otherModules = [];
+                    prompt.otherModules.forEach(function(module) {
+                        generator.otherModules.push({name:module.name[0], version:module.version[0]});
+                    });
+                    done();
```
deepu105: you have to set the other modules here `this.configOptions.otherModules = generator.otherModules;` so that client can see it when composed

gzsombor:
I think, I've implemented all your suggestions

jdubois:
Yes, that looks like a very neat idea!!
Did you check when there are network issues, or when npmsearch is unavailable? If yes, I'm OK to merge this

gzsombor:
I've checked, if I modify the host name - to something not available - it will print a warning, and continues.

jdubois:
Let's do it then :-)

PR #4575, opened 2016-11-29
Fix npm test on angular-2 branch

mraible:
`npm test` fails after last commit. This fixes it.


mraible:
Fixed by https://github.com/jhipster/generator-jhipster/pull/4577.

PR #4586, opened 2016-11-30
Fix Hazelcast versions and add missing JCache dependency

mehmet-aslan:
There are some Hazelcast dependency version errors and a missing JCache dependency at Gradle. Both are correct at Maven.

@jdubois you may want to merge this before releasing a Hibernate 5 / Hazelcast related version update.


jdubois:
Oh, thanks!

PR #4594, opened 2016-12-01
add an imagePullPolicy: IfNotPresent to k8s deployment files

PierreBesson:
Following a [discussion on the mailing list](https://groups.google.com/forum/#!topic/jhipster-dev/W44XX9W9Y4Y), I suggest we add this to be able to deploy to minikube without pushing to an external registry (using `eval $(minikube docker-env)`).

It could also be argued that it is a sensible default and should not cause problems if images are tagged properly.


mraible:
LGTM 👍

PR #4600, opened 2016-12-02
Isolate Jenkins jobs with docker and support CD in heroku

cbornet:
Using docker in Jenkins allows to:
* simplify the setup of Jenkins (eg. no need to install tools on the nodes)
* isolate the builds and start on clean environments

This PR will:
* Use [CloudBees Docker Pipeline Plugin](https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Docker+Pipeline+Plugin) to run builds in docker containers
* Call ArtifactArchiver task at the end of the packaging to archive the generated war.
* Add a SonarQube analysis step (optional, must be uncommented)

Jenkins setup requirements:
* [CloudBees Docker Pipeline Plugin](https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Docker+Pipeline+Plugin) must be installed
* Docker installed in the Jenkins node
* `jenkins` user has permission to use docker (usermod -aG docker jenkins)
* If you use a dockerized Jenkins, run it with `docker run -d --name jenkins2 -p 18080:8080 -p 50000:50000 -v $(which docker):/usr/bin/docker -v jenkins_home:/var/jenkins_home -v /var/run/docker.sock:/var/run/docker.sock --group-add=$(stat -c %g /var/run/docker.sock) jenkinsci/jenkins` so that it can use the host docker
* Optional but recommended : remap host jenkins user to container's root user (for better security and so that files generated during the build belong to jenkins). To do that add a file /etc/docker/daemon.json with content:
```json
{ 
    "userns-remap": "jenkins"
} 
```
and in /etc/subuid add a line `jenkins:106:1` replacing 106 by the uid of the jenkins user (get it with command `id jenkins`)

TODO/Next:
- [x] Create and use an official jhipster ci docker image
- [ ] Update website doc
- [x] Configure cache dirs through env variables
- [x] Find a way to cache the maven/gradle wrapper (currently downloaded at each build)
- [x] Make the heroku subgen generate a CD stage. For the impatient:
```groovy
        stage('package and deploy') {
            sh "git remote add heroku https://git.heroku.com/jhipstersampleapplication.git || true"
            sh "HEROKU_API_KEY=\"xxxx-xxx-xx-xx-xxxxxx\" ./mvnw com.heroku.sdk:heroku-maven-plugin:1.1.1:deploy-war -DskipTests -Pprod -s settings.xml"
            step([$class: 'ArtifactArchiver', artifacts: '**/target/*.war', fingerprint: true])
        }            
```
(a jhipster app is continuously deployed at https://jhipstersampleapplication.herokuapp.com :smile: )

### DIFF
--- a/generators/heroku/templates/_Jenkinsfile
+++ b/generators/heroku/templates/_Jenkinsfile
@@ -5,7 +5,7 @@ node {
         checkout scm
     }
 
-    docker.image('atomfrede/gitlab-ci-jhipster-stack').inside('-u root<% if (buildTool === 'maven') { %> -e MAVEN_OPTS=-Duser.home=./<% } else if (buildTool === 'gradle') { %> -e GRADLE_USER_HOME=.gradle<% } %>') {
+    docker.image('jhipster/jhipster-ci-stack').inside('-u root<% if (buildTool === 'maven') { %> -e MAVEN_OPTS="-Duser.home=./"<% } else if (buildTool === 'gradle') { %> -e GRADLE_USER_HOME=.gradle<% } %>') {
 <%_ if (applicationType !== 'microservice') { _%>
         stage('check tools') {
             sh "node -v"
@@ -54,7 +54,6 @@ node {
 <%_ if (applicationType !== 'microservice') { _%>
         stage('frontend tests') {
             try {
-                sh "echo '{ \"allow_root\": true }' > /root/.bowerrc"
                 sh "gulp test"
             } catch(err) {
                 throw err
@@ -69,10 +68,9 @@ node {
 
 <%_ } _%>
         stage('package and deploy') {
-            sh "git remote add heroku https://git.heroku.com/<%= herokuAppName %>.git || true"
-<%_ if (buildTool === 'maven') { _%
+<%_ if (buildTool === 'maven') { _%>
             // Fill with your Heroku API key (get it at https://dashboard.heroku.com/account)
-            sh "HEROKU_API_KEY=\"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxx\" ./mvnw com.heroku.sdk:heroku-maven-plugin:1.1.1:deploy-war -DskipTests -Pprod"
+            sh "HEROKU_API_KEY=\"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxx\" ./mvnw com.heroku.sdk:heroku-maven-plugin:1.1.1:deploy-war -DskipTests -Pprod -Dheroku.appName=<%= herokuAppName %>"
 <%_ } else if (buildTool === 'gradle' ) { _%>
             sh "./gradlew bootRepackage -Pprod -x test"
             // Fill with your Heroku API key (get it at https://dashboard.heroku.com/account)
--- a/generators/heroku/templates/_Jenkinsfile
+++ b/generators/heroku/templates/_Jenkinsfile
@@ -1,0 +1,96 @@
+#!/usr/bin/env groovy
+
+node {
+    stage('checkout') {
+        checkout scm
+    }
+
+    docker.image('atomfrede/gitlab-ci-jhipster-stack').inside('-u root<% if (buildTool === 'maven') { %> -e MAVEN_OPTS=-Duser.home=./<% } else if (buildTool === 'gradle') { %> -e GRADLE_USER_HOME=.gradle<% } %>') {
+<%_ if (applicationType !== 'microservice') { _%>
+        stage('check tools') {
+            sh "node -v"
+            sh "npm -v"
+            sh "bower -v"
+            sh "gulp -v"
+            sh "java -version"
+        }
+
+        stage('npm install') {
+            sh "npm install"
+        }
+
+<%_ } _%>
+<%_ if (buildTool === 'maven') { _%>
+        stage('clean') {
+            sh "./mvnw clean"
+        }
+
+        stage('backend tests') {
+            try {
+                sh "./mvnw test"
+            } catch(err) {
+                throw err
+            } finally {
+                junit '**/target/surefire-reports/TEST-*.xml'
+            }
+        }
+
+<%_ } else if (buildTool === 'gradle') { _%>
+        stage('clean') {
+            sh "./gradlew clean"
+        }
+
+        stage('backend tests') {
+            try {
+                sh "./gradlew test"
+            } catch(err) {
+                throw err
+            } finally {
+                junit '**/build/**/TEST-*.xml'
+            }
+        }
+
+<%_ } _%>
+<%_ if (applicationType !== 'microservice') { _%>
+        stage('frontend tests') {
+            try {
+                sh "echo '{ \"allow_root\": true }' > /root/.bowerrc"
+                sh "gulp test"
+            } catch(err) {
+                throw err
+            } finally {
+                <%_ if (buildTool === 'maven') { _%>
+                junit '**/target/test-results/karma/TESTS-*.xml'
+                <%_ } else if (buildTool === 'gradle') { _%>
+                junit '**/build/test-results/karma/TESTS-*.xml'
+                <%_ } _%>
+            }
+        }
+
+<%_ } _%>
+        stage('package and deploy') {
+            sh "git remote add heroku https://git.heroku.com/<%= herokuAppName %>.git || true"
+<%_ if (buildTool === 'maven') { _%
+            // Fill with your Heroku API key (get it at https://dashboard.heroku.com/account)
+            sh "HEROKU_API_KEY=\"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxx\" ./mvnw com.heroku.sdk:heroku-maven-plugin:1.1.1:deploy-war -DskipTests -Pprod"
+<%_ } else if (buildTool === 'gradle' ) { _%>
+            sh "./gradlew bootRepackage -Pprod -x test"
+            // Fill with your Heroku API key (get it at https://dashboard.heroku.com/account)
+            sh "HEROKU_API_KEY=\"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxx\" ./gradlew deployHeroku"
+<%_ } _%>
+            archiveArtifacts artifacts: '**/target/*.war', fingerprint: true
+        }
+
+        // Uncomment the following block to add Sonar analysis.
+        /*stage('quality analysis') {
+            withSonarQubeEnv('Sonar Server') {
+<%_ if (buildTool === 'maven') { _%>
+                sh "./mvnw sonar:sonar"
+<%_ } else if (buildTool === 'gradle' ) { _%>
+                sh "./gradlew sonarqube"
+<%_ } _%>
+            }
+        }*/
+
+    }
+}
--- a/generators/heroku/templates/_Jenkinsfile
+++ b/generators/heroku/templates/_Jenkinsfile
@@ -5,26 +5,27 @@ node {
         checkout scm
     }
 
-    docker.image('jhipster/jhipster-ci-stack').inside('-u root<% if (buildTool === 'maven') { %> -e MAVEN_OPTS="-Duser.home=./"<% } else if (buildTool === 'gradle') { %> -e GRADLE_USER_HOME=.gradle<% } %>') {
-<%_ if (applicationType !== 'microservice') { _%>
-        stage('check tools') {
-            sh "node -v"
-            sh "npm -v"
-            sh "bower -v"
-            sh "gulp -v"
+    docker.image('openjdk:8').inside('-u root<% if (buildTool === 'maven') { %> -e MAVEN_OPTS="-Duser.home=./"<% } else if (buildTool === 'gradle') { %> -e GRADLE_USER_HOME=.gradle<% } %>') {
+        stage('check java') {
             sh "java -version"
         }
 
-        stage('npm install') {
-            sh "npm install"
-        }
-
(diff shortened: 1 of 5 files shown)


PierreBesson:
Maybe this can be a 4.0 feature but I'm wondering if it should be in the CI module...

cbornet:
@PierreBesson Jenkins and heroku are currently in the main gen.

cbornet:
But I would totally agree to put all the CI conf we have in a subgen (including the ones from Pascal's module). That would allow to ask questions such as "Do you want to use Sonar ?"

pascalgrimaud:
I will explain my point of view:
- in some companies, you don't have the hand on Jenkins (another team, sysops, devops, etc.), so you can't install additional plugins, you have to ask them
- Docker need to be installed and the Jenkins user must have the permission to use Docker
- they have to configure Docker to access to the base image directly from internet (Docker Hub) or deploy a private registry
- there are a lot of things to do otherwise this Jenkinsfile won't work

My first idea when I coded the initial Jenkinsfile was to write something very simple, which can work almost immediatly

At the beginning, it sounds very good, but now I see the code, I am in the same position as @PierreBesson :
- I wonder if it should be in a module as it is specific use case (I can give you right on my CI module)
- or it can be another file like `Jenkinsfile_DockerNode`, so people can choose

BTW, great work @cbornet ! I will try to use this new Jenkins configuration on my project as I am admin on our Jenkins machine

cbornet:
> in some companies, you don't have the hand on Jenkins (another team, sysops, devops, etc.), so you can't install additional plugins, you have to ask them

Actually using docker means less plugins to install and less configuration of Jenkins ! You just need the Docker Plugin. Look at our current installation guide: we ask users to install the Node Plugin, then ask them to configure it to include bower, gulp, etc... And it's a manual process through a WEB UI whereas a Dockerfile can be put in SCM (where the project team does PRs that are merged by the jenkins admin)

> Docker need to be installed and the Jenkins user must have the permission to use Docker

Yes , you need to install Docker ... and that's all. Otherwise you need to install node, npm, gulp, bower, maven, gradle. And if you have a python project (like I do) , python, pip, virtualenv, tox, ... And I can go on with every tech stack. Docker simplifies that by putting everything in an isolated image. That's a great improvement and I'm talking from experience: we previously had to do a Jenkins slave per tech stack because we didn't want to have dependencies between our build machines (imagine one needs to build C++ for jessie, and the other one for wheezy). Building a docker image is 10 times easier than configuring a Jenkins slave (which for us was already a docker container btw, but that you need to access with ssh, so more config...)

> they have to configure Docker to access to the base image directly from internet (Docker Hub) or deploy a private registry

No, if you're paranoïd (and that's a great quality) you can just build your image on the docker node from a Dockerfile

> there are a lot of things to do otherwise this Jenkinsfile won't work

Yes but far less than when you don't use it


That said I understand your point of view and I agree that docker could be a generator option. But in that case I would rather have the generator in the main project. I already find it curious that we have Travis and Jenkins in the main gen, but Circle and Gitlab in a module...

pascalgrimaud:
@cbornet : you don't need to convince me about Docker, you know I love it :-)

cbornet:
Yes, I know ! Docker rocks 🐋 !
If you're the Jenkins admin in your company, check [this blog](https://jenkins.io/blog/2016/08/08/docker-pipeline-environments/): by building the image from a Dockerfile inside the project itself, you can have an architecture where Jenkins nodes have generic build images for various languages/platforms and projects can add the missing tools they need for their build without disturbing other projects or the Jenkins admin.

cbornet:
Should I also mention that I don't like the Jenkins UI admin interface ? And that I largely prefer the "code as configuration" you get with Jenkinsfile + Dockerfile ?

cbornet:
See also https://fabric8.io/presentations/JenkinsWorld-CD-Jenkins-Docker-Kubernetes.pdf lessons's learnt (and feel free to disagree 😄 )

mraible:
Not related to this issue, but does anyone know how to get Chrome working in the Jenkins Docker image? I'm using the instructions at https://jhipster.github.io/setting-up-ci-jenkins2/ to try and run a pipeline for my [ng2-demo](https://github.com/mraible/ng2-demo) application. However, it gets hung up on the following error:

```
[1m[33mWARNING in ./~/@angular/core/src/linker/system_js_ng_module_factory_loader.js
57:15 Critical dependency: the request of a dependency is an expression[39m[22m
[33m05 12 2016 15:34:38.594:WARN [karma]: [39mPort 9876 in use
[32m05 12 2016 15:34:38.595:INFO [karma]: [39mKarma v1.2.0 server started at http://localhost:9877/
[32m05 12 2016 15:34:38.595:INFO [launcher]: [39mLaunching browser Chrome with unlimited concurrency
[32m05 12 2016 15:34:38.778:INFO [launcher]: [39mStarting browser Chrome
[31m05 12 2016 15:34:38.779:ERROR [launcher]: [39mNo binary for Chrome browser on your platform.
  Please, set "CHROME_BIN" env variable.
```

PierreBesson:
@cbornet I'm reconsidering my initial thoughts about this, the docker approach for CI presents a lot of advantages, I would be OK if this becomes the default but it would be best if all CI configuration was in its own subgen. But I'm afraid not everyone has admin rights to their Jenkins.

@mraible I think it would be quite complicated to get chrome working in docker as you would need an to have X server running inside docker. Although, it does seems that some people have managed to do it : https://github.com/mark-adams/docker-chromium-xvfb

mraible:
@PierreBesson Running Chrome on Docker seems to be essential for getting Angular 2 tests to run with Jenkins + Docker. Any tips on combining the Jenkins Docker image with https://hub.docker.com/r/markadams/chromium-xvfb/?

PierreBesson:
@mraible Sorry I have never done it.

cbornet:
@mraible I think the easiest is to copy the scripts from https://github.com/mark-adams/docker-chromium-xvfb/tree/master/images/base and replace `FROM debian:jessie` by `FROM jenkins`

mraible:
@cbornet I tried creating a `Dockerfile` with the following:

```
FROM jenkins

USER jenkins

RUN apt-get update && apt-get install -y xvfb chromium

ADD xvfb-chromium /usr/bin/xvfb-chromium
RUN ln -s /usr/bin/xvfb-chromium /usr/bin/google-chrome
RUN ln -s /usr/bin/xvfb-chromium /usr/bin/chromium-browser
```

And then tried to create an image using `docker build`. 

```
± docker build -t docker-jenkins .
Sending build context to Docker daemon 329.8 MB
Step 1 : FROM jenkins
 ---> ff6f0851ef57
Step 2 : USER jenkins
 ---> Using cache
 ---> adbc33cef8fe
Step 3 : RUN apt-get update && apt-get install -y xvfb chromium
 ---> Running in f80f85955482
E: List directory /var/lib/apt/lists/partial is missing. - Acquire (13: Permission denied)
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
The command '/bin/sh -c apt-get update && apt-get install -y xvfb chromium' returned a non-zero code: 100
```

Any tips?

cbornet:
Yes, you need to be `USER root` to install with apt.

cbornet:
@mraible did you make make it work ?

mraible:
I haven't had a chance to try this image yet. On flights all day, can try
tomorrow.

cbornet:
I have changed to now use directly the openjdk docker image

cbornet:
@mraible it seems we configure protractor with Firefox by default. So it's probably Firefox that we should install in a docker.

mraible:
@cbornet Yes, because of https://github.com/jhipster/generator-jhipster/pull/4076.

cbornet:
So what should we do from here ?

1/ Don't use docker ? 
* Now that it's not required to setup NodeJS in Jenkins (node is downloaded by maven/gradle), if you install jenkins from the .war and since jenkins already requires you to have jdk8, it makes sense for a minimal setup.

2/ Use docker (this PR) ? 
* If you use the dockerized Jenkins, it's not much to setup (just add user jenkins to the docker group)
* In a multi-project/enterprise Jenkins, it's a great added value
* It's required for e2e/protractor tests (see #4693)

3/ Create a subgen that will ask the question ?
* As a separate module (Pascal's) or in the main gen ? Or a simple one (1/) in the main gen and a complex one in the subgen/module (more maintenance) ?
* We could make a smart one:
    * Do you want to use a Docker build environment ?
    * Do you want to package as a Docker image and publish to a registry ?
    * Heroku|AWS|Cloud Foundry config found, do you want to setup CD ?
    * What additional integrations do you want to use ? (SonarQube/Github/Gitlab/Slack, ...)

jdubois:
For me we don't have much choice, we must go with 3), and I would do it in the main gen as this has to be "official".

I have a lot of customers who don't have Docker -> in that case let's just generate a very simple `mvn test`-like configuration

For everybody with Docker, which is great for CI (as we see with Travis every day), we can go much much farther. And indeed, one of my goals would be to deploy automatically to a cloud like Heroku (ping @jkutner ), as this would really be an awesome experience.

cbornet:
OK to go for 3/. It's more work but in the end a better experience.
As for Heroku CD, you can have a look at https://github.com/jhipster/generator-jhipster/pull/4600/commits/156bedfb4bd1e31746528ede8e9ed91018ccbaa8 : I had done it in this PR... 😞 
As for the subgen, should it be a generic `jhipster:ci` that asks which CI need to be generated in a choice question (Jenkins/Travis/Circle/Gitlab) ? Or one gen per CI (eg. `jhipster:jenkins`)

pascalgrimaud:
@cbornet : if needed, dont hesitate to take my code from my module. I ll try to have a look on this, after holidays :-)

PierreBesson:
Very excited for option 3). Good luck @cbornet. I think I might be able to contribute an automatic deploy to kubernetes as we have done something similar at my job.

jkutner:
Whatever you decide, I'll make sure it works with Heroku (please mention me on any relevant discussions). I think we can skip the `git remote add` step, too, and use `-Dheroku.appName=sushi` in the `mvnw` or `gradlew` command.

cbornet:
Yes that's what I did. You also need to pass the API key.

generators/heroku/templates/_Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+<%_ if (buildTool === 'maven') { _%>
+        stage('clean') {
+            sh "./mvnw clean"
+        }
+
+    <%_ if (applicationType !== 'microservice') { _%>
+        stage('prepare tools') {
+            sh "./mvnw com.github.eirslett:frontend-maven-plugin:install-node-and-npm -DnodeVersion=v6.9.2 -DnpmVersion=3.10.9"
```
pascalgrimaud: it can fail, as the plugin can change when using yarn or npm

generators/heroku/templates/_Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+            sh "echo '{ \"allow_root\": true }' > /root/.bowerrc"
+        }
+
+        stage('npm install') {
+            sh "./mvnw com.github.eirslett:frontend-maven-plugin:npm"
```
pascalgrimaud: same here, it can be 'yarn'

generators/server/templates/_Jenkinsfile
```diff
@@ -1,77 +1,107 @@
...
+        stage('clean') {
+            sh "./mvnw clean"
+        }
 
-    stage('npm install') {
-        sh "npm install"
-    }
+    <%_ if (applicationType !== 'microservice') { _%>
+        stage('prepare tools') {
+            sh "./mvnw com.github.eirslett:frontend-maven-plugin:install-node-and-npm -DnodeVersion=v6.9.2 -DnpmVersion=3.10.9"
```
pascalgrimaud: same here

generators/server/templates/_Jenkinsfile
```diff
@@ -1,77 +1,107 @@
...
+            sh "echo '{ \"allow_root\": true }' > /root/.bowerrc"
+        }
 
-<%_ } _%>
-<%_ if (buildTool === 'maven') { _%>
-    stage('clean') {
-        sh "./mvnw clean"
-    }
+        stage('npm install') {
+            sh "./mvnw com.github.eirslett:frontend-maven-plugin:npm"
```
pascalgrimaud: same here

generators/server/templates/_Jenkinsfile
```diff
@@ -1,77 +1,107 @@
...
-    stage('clean') {
-        sh "./gradlew clean"
-    }
+        stage('clean') {
+            sh "./gradlew clean"
+        }
+    <%_ if (applicationType !== 'microservice') { _%>
+        stage('npm install') {
+            sh "echo '{ \"allow_root\": true }' > /root/.bowerrc"
+            sh "./gradlew npmInstall -PnodeInstall"
```
pascalgrimaud: same here

cbornet:
Although I agree it would be better to support yarn, I don't see why npm would fail. And the current Jenkinsfile is also npm only...

pascalgrimaud:
I was wrong, you're right @cbornet. npm and yarn are compatible each other, we can use both without any problem

pascalgrimaud:
Just tested it, and it works fine!
Good job @cbornet !

cbornet:
Closing as I will do a sub-gen with questions for the options to use (docker, heroku, sonar, protractor, etc...)
I also opened #4896 for improving the simple setup.

PR #4603, opened 2016-12-03
Migration of interceptors to Angular 2

VictorADS:
Migrate interceptors to Angular2.
 I'd like to get @xetys approval cause I changed some of his code to make it work.

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
+++ b/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
@@ -1,8 +1,0 @@
-HttpConfig.$inject = ['$httpProvider', 'httpRequestInterceptorCacheBusterProvider'];
-
-export function HttpConfig($httpProvider, httpRequestInterceptorCacheBusterProvider) {
-    //Cache everything except rest api requests
-    httpRequestInterceptorCacheBusterProvider.setMatchlist([/.*api.*/, /.*protected.*/], true);
-
-    // jhipster-needle-angularjs-add-interceptor JHipster will add new application http interceptor here
-}
--- a/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
+++ b/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
@@ -4,7 +4,6 @@ export function HttpConfig($httpProvider, httpRequestInterceptorCacheBusterProvi
     //Cache everything except rest api requests
     httpRequestInterceptorCacheBusterProvider.setMatchlist([/.*api.*/, /.*protected.*/], true);
 
-    $httpProvider.interceptors.push('ErrorHandlerInterceptor');
     $httpProvider.interceptors.push('NotificationInterceptor');
     // jhipster-needle-angularjs-add-interceptor JHipster will add new application http interceptor here
 }
--- a/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
+++ b/generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
@@ -4,6 +4,5 @@ export function HttpConfig($httpProvider, httpRequestInterceptorCacheBusterProvi
     //Cache everything except rest api requests
     httpRequestInterceptorCacheBusterProvider.setMatchlist([/.*api.*/, /.*protected.*/], true);
 
-    $httpProvider.interceptors.push('NotificationInterceptor');
     // jhipster-needle-angularjs-add-interceptor JHipster will add new application http interceptor here
 }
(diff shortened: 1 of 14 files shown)


generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
```diff
@@ -1,22 +1,31 @@
...
+import { Observable } from 'rxjs/Observable';
 
-NotificationInterceptor.$inject = ['$q'];
+export class NotificationInterceptor extends HttpInterceptable {
 
-export function NotificationInterceptor ($q) {
-    var service = {
-        response: response
-    };
+    constructor() {
```
mraible: Isn't a default constructor that calls super() the same as no constructor at all?

VictorADS: That might be the case but l left it like that because  later we will need to inject the AlertService into this interceptor and l planned to do this within the constructor

mraible: Makes sense.

VictorADS:
Can anyone tell me why there is no Singleton in the loginModalService ? This is causing some issue when an user type the wrong credential when trying to login ( The AuthExpiredInterceptor pops an another loginModal cause ATM it is not a singleton !)

generators/client-2/templates/src/main/webapp/app/blocks/config/_http.config.ts
```diff
@@ -4,7 +4,5 @@ export function HttpConfig($httpProvider, httpRequestInterceptorCacheBusterProvi
     //Cache everything except rest api requests
     httpRequestInterceptorCacheBusterProvider.setMatchlist([/.*api.*/, /.*protected.*/], true);
 
-    $httpProvider.interceptors.push('ErrorHandlerInterceptor');
```
deepu105: I think we can remove this file as it doesnt do anything now

generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_auth-expired.interceptor.ts
```diff
@@ -13,7 +13,8 @@ import {AuthServerProvider} from '../../shared/auth/auth-jwt.service';
 <%_ } if (authenticationType === 'session') { _%>
 import { AuthServerProvider } from '../../shared/auth/auth-session.service';
 import { StateStorageService } from '../../shared/auth/state-storage.service';
-<% } %>
+import { LoginModalService } from '../../shared/login/login-modal.service';
+    <% } %>
```
deepu105: unwanted space

generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_http.interceptor.ts
```diff
@@ -31,7 +31,7 @@ export class HttpInterceptor extends Http {
     }
 
     request(url: string | Request, options?: RequestOptionsArgs): Observable<Response> {
-        return this.intercept(super.request(url, this.getRequestOptionArgs(options)));
+        return super.request(url, this.getRequestOptionArgs(options));
```
deepu105: how does the intercept work then?

VictorADS: Reasoning behind : Everytime we call http.get or smthing, this method (request) was called so we intercepted the same request twice. It doesn't make any error appear but I thought that was kinda useless to do the same thing twice ?

deepu105: Ok but I still dont get how we are doing it twice is there some other place where we call the intercept method? sorry didnt look into the code in detail yet

VictorADS: In fact, I think it's intrinsic to the http class. Each time we try make an http request the appropriate method (get , put, post...) is called and then  the request method is called. So if we intercept it in every method (like before my commit) you intercept the request twice. The first one in the get method and then in the request method for exemple. You can try it by yourself if you add some logs in each method of the http.interceptor and in each interceptor to understand what l'm talking about. Sorry if l was unclear  !

deepu105: ok that makes sense. @xetys are you ok with this?

xetys: thanks @VictorADS you figured this thing out!

xetys:
I didn't test this PR, yet, but took some looks over this PR and I am ok with it. Looks clean. As we are working on alpha, I am ok to merge this

Issue #4602, opened 2016-12-03
Travis: java misconfigured and/or unavailable


pascalgrimaud:
Our current Travis build can't work because of this https://github.com/travis-ci/travis-ci/issues/6928

So all PR will fail until it is fixed.
There are some temporary configuration to made it works. I will try later to do it

PR #4607, opened 2016-12-05
Travis: java misconfigured

pascalgrimaud:
It should fix https://github.com/jhipster/generator-jhipster/issues/4602

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,8 @@ env:
     - JHIPSTER=app-microservice-consul
 
 before_install:
+  - jdk_switcher use oraclejdk8
+  - java -version
   - export TZ=Australia/Canberra
   - date
   - sudo /etc/init.d/postgresql stop
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ before_install:
   - java -version
   - export TZ=Australia/Canberra
   - date
+  - sudo /etc/init.d/mysql stop
   - sudo /etc/init.d/postgresql stop
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start


jdubois:
Great, looks like it's working!!!
I'm merging this, and re-launch the failed builds, so everything should be OK when I arrive at work!

pascalgrimaud:
@jdubois : relaunch the failed builds won't work -> all PR need to fetch + rebase master

jdubois:
Oh yes, my bad...

pascalgrimaud:
FYI, all builds are OK except the OAuth2:
- all builds: https://travis-ci.org/pascalgrimaud/generator-jhipster/builds/181260138
- OAuth2: https://travis-ci.org/pascalgrimaud/generator-jhipster/jobs/181260153

It is clearly not our fault, as the OAuth2 configuration works well.

A quick solution would be to use the old version of Travis, with `group: deprecated` but I think we can wait and test the next release tomorrow morning US/Eastern: https://docs.travis-ci.com/user/build-environment-updates/2016-12-06/

jdubois:
Well, for OAuth2, I don't know. I spent a couple of hours today to try to find out what's happening, and that's really strange.

- I can't reproduce the issue on my laptop
- But if you look at the log it's a Spring misconfiguration (there's a bean that is not configured), so that looks like our fault, and we should be able to reproduce this
- The Maven tests work on Travis, so the Spring bean is configured there - it just makes no sense: the tests pass, but you can't start the app...

pascalgrimaud:
I'm pretty sure it comes from Travis, as I can't reproduce on my machine too.

When using `group: deprecated` which is the old Travis image, the OAuth2 build is OK:
https://travis-ci.org/pascalgrimaud/generator-jhipste

When using the current, OAuth2 build failed.

I wonder if I should PR immediatly this change or wait the new Travis release tomorrow?

jdubois:
In that case Travis is badly broken!!! I can't understand why it would trigger a misconfigured Spring bean in just one specific part of one specific build, when that works everywhere else!
Yes can you PR the change? We have too many broke builds at the moment, and I'm not confident at all with Travis... Then we'll see tomorrow if we can upgrade.

Issue #4407, opened 2016-10-30
Angular2 known issues


deepu105:
Adding a known issues todo list for Angular2 so that we have track of these here
- [x] `Resetting invalid cookie language "en" to prefered language "en"` this is being logged enormous times, thats not the issue, the issue is somewhere the stuff is being run so many times then really needed and its kind of slowing down the page loads. This needs to be investigated and fixed
- [x] There is a console error during first login after a page refresh
- [x] UI bootstrap doesnt seem to work -migrated to ng-bootstrap
- [x] UI router doesnt seem to work completeley - moved to ui-router-ng2 see https://github.com/jhipster/generator-jhipster/pull/4220
- [x] NPM `postinstall` of `typings install` doesn't work in windows (not sure of other os) some weird error. It works when the command is run manually - **works** `typings` needs to be installed with global scope `npm i -g typings` [see](https://github.com/jhipster/generator-jhipster/issues/3664#issuecomment-243690915)
- [x] Health modal needs to be migrated
- [x] Update to RC6
- [x] After RC6 update: This app works for now. But we have an issue. If we use multiple `@NgModule` like how I have defined for now, then we are not able to use the common pipes and directives across these, seems like app doesnt load if I declare a pipe in more than one module. we need to find a solution for this or we need to stick to just a single huge module for the entire app(Which is ugly) - **fixed** [see](https://github.com/jhipster/generator-jhipster/issues/3664#issuecomment-245885129)
- [x] click of any header menu item reloads the page [after ng-bootstrap update] **fixed** added `href="#"` to empty anchors
- [x] shared directives needs to be tested and validated
- [x] error when clicking on the login link in account/register page
- [x] swagger api page doesnt load
- [x] websocket connection is failing
- [x] User Management Detail/Delete Dialogs don't load translation into page title
- [x] User Management user list does not refresh after making a change to the users
- [x] User Management pagination doesnt work properly
- [x] metrics page needs update once #4454 is merged to master
- [x] password strength bar directive doesnt work
- [x] The navbar menu doesnt work in mobile view
- [x] bootstrap show/hide elements based on screensize doesnt work might be a class rename issue
- [x] missing translations in health screen
- [X] user management view details is very slow to load
- [x] websocket sendActivity is not being triggered, need to see if can be triggered on route changes
- [x] tracker page needs to be tested for edge cases
[...]

PR #4609, opened 2016-12-05
Refresh user list  in UserManagement

VictorADS:
Fix User Management user list does not refresh after making a change to the users in
 #4407


generators/client-2/templates/src/main/webapp/app/admin/user-management/_user-management-dialog.component.ts
```diff
@@ -50,7 +53,8 @@ export class UserMgmtDialogComponent implements OnInit {
         }
     }
     private onSaveSuccess (result) {
-        this.isSaving = false;
+    this.$eventManager.broadcast({ name: 'userListModification', content:'OK'});
```
deepu105: formatting: indent

deepu105: formatting: indent

PR #4611, opened 2016-12-05
Travis: use group deprecated

pascalgrimaud:
As discussed here https://github.com/jhipster/generator-jhipster/pull/4607#issuecomment-265000737


jdubois:
It's working!! Travis is definitely badly broken!
Thanks @pascalgrimaud

PR #4619, opened 2016-12-06
Swagger api page doesnt load 

maniacneron:
- Fix swagger api page doesnt load #4407

-  Fix imports for SessionStorageService and LocalStorageService if
authentication type jwt, uua and oauth2


generators/client-2/templates/src/main/webapp/app/blocks/interceptor/_http.provider.ts
```diff
@@ -6,6 +6,9 @@ import { AuthInterceptor } from './auth.interceptor';
 <%_ } if(authenticationType === 'session') { _%>
 import { StateStorageService } from '../../shared/auth/state-storage.service';
 <%_ } _%>
+<%_ if (authenticationType === 'oauth2' || authenticationType === 'jwt' || authenticationType === 'uaa') { _%>
+import {LocalStorageService, SessionStorageService} from "ng2-webstorage/index";
```
deepu105: no need of index just `'ng2-webstorage'` would do. use single quotes, also leave space before and after the braces

PR #4624, opened 2016-12-07
Fix swagger pagination plugin not called

### LINKED ISSUE #4615: Paging and sorting parameters missing from api docs
With the update to springfox 2.6.1 in JH 3.12.0 paging and sorting parameters are missing from the api docs.  If I revert the changes from #4438 the issue is resolved.

cbornet:
Fix #4615


ruddell:
Works great.  This is also needed in UserResource.java getAll

cbornet:
@ruddell You're right ! Done.

PR #4625, opened 2016-12-07
Removing exposed ports: nobody listening there

lrkwz:
Dockerfile exposes ports 8080, 3000 and 3001 but nothing is listening there in the container.
Starting jhipster container as documented

    docker run --name jhipster -v `pwd`:/home/jhipster/app -v ~/.m2:/home/jhipster/.m2 -p 8080:8080 -p 3000:3000 -p 3001:3001 -d -t jhipster/jhipster

blocks these ports so you can run the application only inside the container.

I know that develoers can still run the container without exposing these ports but, since there are other dockerfile definitions (```src/main/docker```) to containerize the application I see no point in exposing these ports here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,5 +67,5 @@ RUN \
 USER jhipster
 WORKDIR "/home/jhipster/app"
 VOLUME ["/home/jhipster/app"]
-EXPOSE 8080 3000 3001
+#EXPOSE 8080 3000 3001
 CMD ["tail", "-f", "/home/jhipster/generator-jhipster/generators/server/templates/src/main/resources/banner-no-color.txt"]


ruddell:
The ports are used by mvn/gulp after generating a project.  I think exposing the ports helps - if you load Kitematic, then it suggests which ports to use.

Wouldn't the same result happen if you excluded `-p 8080:8080 -p 3000:3000 -p 3001:3001` from the command when you call it?

This dockerfile has a different purpose than the docker files in `src/main/docker`: "Please note: this Docker image is for running the JHipster generator inside a container. It’s completely different from the Docker and Docker Compose configurations that JHipster will generate, which goal is to run your generated application inside a container"

jdubois:
Yes 8080 is port used when the application run, and 3000/3001 are the BrowserSync port (but I think they changed to 9000/9001 - which would means nobody every uses it from Docker!!)

jdubois:
After checking:

- We use port `9000` for BrowserSync
- We use port `3001` for the BrowserSync admin UI

I find it quite strange we use those ports instead of (`9000` and `9001`) or (`3000` and `3001`) but it's probably been working like that for some time and I've seen any complaint...

So this PR shouldn't be merged, but we need to update the current opened ports.

Issue #3100, opened 2016-03-07
Test generation failure : regexp pattern constraint not taken into account in test generated code


fraf:
To summarize :

When I generate the entity "typeFichier", having a field named "cdSens", the generated test file is named "TypeFichierResourceIntTest".
Its data members are generated like below :

```
 private static final String DEFAULT_FORMAT = "AAAAA";
    private static final String UPDATED_FORMAT = "BBBBB";
 **   private static final String DEFAULT_CD_SENS = "AAAAA";
    private static final String UPDATED_CD_SENS = "BBBBB";**
    private static final String DEFAULT_LIBELLE_FORMAT = "AAAAA";
    private static final String UPDATED_LIBELLE_FORMAT = "BBBBB";
    private static final String DEFAULT_TYPE = "AAAAA";
    private static final String UPDATED_TYPE = "BBBBB";
    private static final String DEFAULT_NIVEAU = "AAAAA";
    private static final String UPDATED_NIVEAU = "BBBBB";
```

That generated code is wrong ! Because I specified a regexp pattern restriction on field "cdSens" : ([U,D]).
Consequently, tests fails ! I can't package the webapp anymore.

Should be :

```
private static final String DEFAULT_FORMAT = "AAAAA";
    private static final String UPDATED_FORMAT = "BBBBB";
  **  private static final String DEFAULT_CD_SENS = "U";
    private static final String UPDATED_CD_SENS = "D";**
    private static final String DEFAULT_LIBELLE_FORMAT = "AAAAA";
    private static final String UPDATED_LIBELLE_FORMAT = "BBBBB";
    private static final String DEFAULT_TYPE = "AAAAA";
    private static final String UPDATED_TYPE = "BBBBB";
    private static final String DEFAULT_NIVEAU = "AAAAA";
    private static final String UPDATED_NIVEAU = "BBBBB";
```

EDIT : should we help the generator giving it a valid input to generate the code ?

jdubois:
Yes it's a limitation when we generate entities with REGEXP.
It's documented here: http://jhipster.github.io/creating-an-entity/

I'm not happy with this limitation, but we have already worked on this for some time, and couldn't find any good solution, as we don't know what will match your regexp.

The solution, of course, is that you write the correct data that maps your regexp, and then the test will pass.

PR #4628, opened 2016-12-08
fix tests for entity fields with pattern

kaidohallik:
fix jhipster/generator-jhipster#3100

### DIFF
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -1,6 +1,8 @@
 'use strict';
 
-const _ = require('lodash');
+const _ = require('lodash'),
+    randexp = require('randexp'),
+    chalk = require('chalk');
 
 /* Constants use throughout */
 const constants = require('../generator-constants'),
@@ -116,6 +118,7 @@ const serverFiles = {
             path: SERVER_TEST_SRC_DIR,
             templates: [{
                 file: 'package/web/rest/_EntityResourceIntTest.java',
+                options: {'context': {'randexp': randexp, '_': _, 'chalkRed': chalk.red}},
                 renameTo: generator => `${generator.packageFolder}/web/rest/${generator.entityClass}ResourceIntTest.java`
             }]
         },
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -2,7 +2,8 @@
 
 const _ = require('lodash'),
     randexp = require('randexp'),
-    chalk = require('chalk');
+    chalk = require('chalk'),
+    fs = require('fs');
 
 /* Constants use throughout */
 const constants = require('../generator-constants'),
@@ -118,7 +119,7 @@ const serverFiles = {
             path: SERVER_TEST_SRC_DIR,
             templates: [{
                 file: 'package/web/rest/_EntityResourceIntTest.java',
-                options: {'context': {'randexp': randexp, '_': _, 'chalkRed': chalk.red}},
+                options: {'context': {'randexp': randexp, '_': _, 'chalkRed': chalk.red, 'fs': fs, 'SERVER_TEST_SRC_DIR': SERVER_TEST_SRC_DIR}},
                 renameTo: generator => `${generator.packageFolder}/web/rest/${generator.entityClass}ResourceIntTest.java`
             }]
         },
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -485,7 +485,7 @@ module.exports = EntityGenerator.extend({
 
                 if (_.isUndefined(field.fieldValidateRulesPatternJava)) {
                     field.fieldValidateRulesPatternJava = field.fieldValidateRulesPattern ?
-                        field.fieldValidateRulesPattern.replace(/\\/g, '\\\\') : field.fieldValidateRulesPattern;
+                        field.fieldValidateRulesPattern.replace(/\\/g, '\\\\').replace(/"/g, '\\"') : field.fieldValidateRulesPattern;
                 }
 
                 if (_.isArray(field.fieldValidateRules) && field.fieldValidateRules.length >= 1) {
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -144,8 +144,8 @@
           <%_ } _%>
           <%_ if (fields[idx].fieldValidateRules.indexOf('pattern') != -1) { _%>
           <small class="form-text text-danger"
-           [hidden]="!editForm.controls.<%= fieldName %>?.errors?.pattern" jhiTranslate="entity.validation.pattern" translateValues="{ pattern: '<%= fields[idx].fieldValidateRulesPattern %>' }">
-            This field should follow pattern "<%= fields[idx].fieldValidateRulesPattern %>".
+           [hidden]="!editForm.controls.<%= fieldName %>?.errors?.pattern" jhiTranslate="entity.validation.pattern" translateValues="{ pattern: '<%= fields[idx].fieldValidateRulesPattern.replace(/"/g, '&#34;') %>' }">
+            This field should follow pattern "<%- fields[idx].fieldValidateRulesPattern.replace(/&/g, '&amp;').replace(/\{/g, '&#123;').replace(/\}/g, '&#125;') %>".
           </small>
             <%_ } _%>
             <%_ if (fieldType == 'Integer' || fieldType == 'Long' || fieldType == 'Float' || fieldType == 'Double' || fieldType == 'BigDecimal') { _%>
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/ng_validators.ejs
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/ng_validators.ejs
@@ -29,7 +29,7 @@ if (field.fieldValidate == true) {
         validators.push('maxbytes="' + field.fieldValidateRulesMaxbytes +'"');
     }
     if (rules.indexOf('pattern') != -1) {
-        validators.push('pattern="' + field.fieldValidateRulesPattern + '"');
+        validators.push('pattern="' + field.fieldValidateRulesPattern.replace(/"/g, '&#34;') + '"');
     }
 
     result = validators.join(' ');
--- a/generators/entity/templates/client/angularjs/src/main/webapp/app/entities/ng_validators.ejs
+++ b/generators/entity/templates/client/angularjs/src/main/webapp/app/entities/ng_validators.ejs
@@ -29,7 +29,7 @@ if (field.fieldValidate == true) {
         validators.push('maxbytes="' + field.fieldValidateRulesMaxbytes +'"');
     }
     if (rules.indexOf('pattern') != -1) {
-        validators.push('ng-pattern="/' + field.fieldValidateRulesPattern + '/"');
+        validators.push('ng-pattern="/' + field.fieldValidateRulesPattern.replace(/"/g, '&#34;') + '/"');
     }
 
     result = validators.join(' ');
--- a/generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
+++ b/generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
@@ -115,9 +115,53 @@ public class <%= entityClass %>ResourceIntTest <% if (databaseType == 'cassandra
         if (fields[idx].fieldValidateRulesMaxlength < sampleTextLength) {
             sampleTextLength = fields[idx].fieldValidateRulesMaxlength;
         }
-        for (var i = 0; i < sampleTextLength; i++ ) {
+        for (var i = 0; i < sampleTextLength; i++) {
             sampleTextString += "A";
             updatedTextString += "B";
+        }
+        if (!this._.isUndefined(fields[idx].fieldValidateRulesPattern)) {
+            // Generate Strings, using pattern
+            try {
+                var sampleTextHigherLength = 100;
+                var sampleTextLengthLimit = 1000;
+                var patternRegExp = new RegExp(fields[idx].fieldValidateRulesPattern);
+                if (!patternRegExp.test(sampleTextString)) {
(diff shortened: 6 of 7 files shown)


kaidohallik:
I made code cleaning:
* instead of expanding data object with randexp use ejs template context option
* use 1 const keyword instead of 2 const keywords and separate require with comma
* fix spacing
* in undefined check use !== instead of !=
* in maxlength exists check use `fieldValidateRulesMaxlength !== undefined` instead of `fieldValidateRules.indexOf('maxlength') != -1`

and updated PR

kaidohallik:
Improved code:
* add error handling, in case of pattern error default values are generated
* use `_.isUndefined` instead of `!== undefined`
* fix spacing

PR updated.

jdubois:
This is a very nice idea, and I totally agree the end result is better than what we currently have.

BUT, I have an issue here: from my understanding you generate a random test String, so this is going to change each time the generator is used. Many people (including me, or our full test suite) are re-generating their entities quite often: those test Strings will change every time, and add a lot of "noise" in the commits. So, unless I'm mistaken, I'm not a huge fan of doing this.

kaidohallik:
You are right, there will be noise. But in fact that noise already has there.

Current workflow on first generation:
1. generate
2. change tests

Current workflow on regeneration:
1. regenerate
2. review all tests: revert broken pattern texts and leave added logic

Workflow with improved logic on first generation:
1. generate and thats all

Regeneration workflow with improved logic:
1. regenerate, if noise is accepted then all is done
2. if noise is not accepted then review all tests: revert regenerated pattern texts

So my conclusion is that this enhancment simplifies work, it does not add extra work to current workflow.

kaidohallik:
More enhanced solution can be: write originally generated test values to some properties file under .jhipster folder and try to read these valuse from there while regenerating.
If this value is still valid then use it.

jdubois:
Oh yes, +1 for persisting it in the json files, that's neat!

kaidohallik:
So let's discuss possible solution details.
Should it be in the separate file or in the same file where other Entity elements are located?

If the same file like that:
```
{
	"fieldName": "myField",
	"javadoc": "My field.",
	"fieldType": "String",
	"fieldValidateRules": [
		"required",
		"maxlength",
		"pattern"
	],
	"fieldValidateRulesMaxlength": 30,
	"fieldValidateRulesPattern": "^[A-Za-z][A-Za-z0-9_]*$",
	"fieldTestValues": [
		"AFirstTestValue",
		"ASecondTestValue"
	]
}
```
then declaration is compact, all in one file. But drawback is: if regenerating from JDL Studio file then test values are gone.

If separate file then there are 2 possible solutions.

**Solution 1.** Use one common pattern fields test values file where test values for every entity are held. Possible file name: `.patternFieldsTestValues`
Possible file structure:
```
{
    "entities": [
        {
            "entityName": "myEntity",
            "fields": [
                {
                    "fieldName": "myField",
                    "testValues": [
                        "AFirstTestValue",
                        "ASecondTestValue"
                    ]
                },
                ..........
            ]
        },
        ..........
    ]
}
```

**Solution 2.** Use separate file for every entity. Possible file name pattern: `.EntityNameTestValues`
Possible file structure:
```
{
    "fields": [
        {
            "fieldName": "myField",
            "testValues": [
                "AFirstTestValue",
                "ASecondTestValue"
            ]
        },
        ..........
    ]
}
```

jdubois:
Don't worry about the JDL, we are also doing it cc @MathieuAA !
That's quite usual in fact: the main project adds a new property, then it must also be implemented in the JDL.

MathieuAA:
Looking into it tonight or tomorrow, so many problems everywhere =/

jdubois:
@MathieuAA take your time, it's not done yet! And the JDL can be one version behind the main generator...

MathieuAA:
@kaidohallik What do you mean by " But drawback is: if regenerating from JDL Studio file then test values are gone." ? If i understand you correctly, this is a bug in JDL studio.

jdubois:
Yes but it's not coded yet in the generator, so that's not a bug!

MathieuAA:
@jdubois Ahah so JHipster Core and JHipster UML are ahead of you guys for once.

kaidohallik:
To me it seems that it's hard to do in JDL Studio level.
Currently all is entered by user in JDL Studio. If requirement will be that system will generate dynamically some additional part in JDL Studio then there are some problematic places:
* on what event should JDL Studio generate this additional part
* it can be difficult to maintain this dynamically generated part
* for user it can be weird that system is generating some additional part in JDL Studio
* what if user deletes this dynamically generated part

Possible solution is keep test values out from JDL Studio (from .jh file) and to do that invisibly in jhipster:import-jdl
1. parse current entity files in .jhipster folder and remember current test values
2. in entity json files generation assign these old test values

jdubois:
Yes you are correct. It's a bit complex but that's doable.
That's basically what we do in the main generator for the remember me
security token.
Same idea, so same solution.

kaidohallik:
One con for using current entity json files to store test values: it basicly adds unimportant information, trash there.
Maybe better solution is to create separate file in `.jhipster` folder like described in one of my previous comments. File name can be for example `.patternFieldsTestValues` or `.testValues`.

jdubois:
It's the same as the `remember me` in the `.yo-rc.json`, and I don't want to have tons of files here. There's one file to generate an entity, and that's easy to use. The only change to do is that the JDL must understand it, or if it's too difficult to read (which shouldn't be the case), just override it and let's regenerate the attribute again.

kaidohallik:
So decision is to go on with storing test values in the current entity json files and I can start implementing it?

jdubois:
Yes for me, unless this causes too much trouble with @MathieuAA (and I will see him tomorrow evening, so we'll talk about it)

MathieuAA:
As far as I'm concerned, regex validations were possible from (almost) day one (see [here](https://github.com/jhipster/jhipster-core/blob/master/lib/dsl/grammar.txt#L265)).
The writing of this is done [here first](https://github.com/jhipster/jhipster-core/blob/master/lib/dsl/grammar.txt#L265), then [there](https://github.com/jhipster/jhipster-core/blob/master/lib/parser/entity_parser.js#L175) so that the validation `fieldValidateRulesPattern` is written in the final JSON.

kaidohallik:
Quick analysis of jhipster:import-jdl below.

Work is done in project generator-jhipster in parseJDL: https://github.com/jhipster/generator-jhipster/blob/master/generators/import-jdl/index.js#L51
This executes jhipster-core functions:
1. reads data from .jh/.jdl file(s)
2. converts .jh/.jdl file(s) to entities json format
3. writes entities json file(s) to disk

Possible solution is to add new step between current 2. and 3. step: populate new entity descriptions with current fieldTestValues.
Code can go to project jhipster-core module entity_parser.js into new function retainFieldTestValues.

MathieuAA:
Hi! Julien and I will have a little chat about this tomorrow. Thank you very much for your contribution, it's about time this issue is solved.

Le 18 décembre 2016 22:05:39 GMT+01:00, Kaido Hallik <notifications@github.com> a écrit :
>Quick analysis of jhipster:import-jdl below.
>
>Work is done in project generator-jhipster in parseJDL:
>https://github.com/jhipster/generator-jhipster/blob/master/generators/import-jdl/index.js#L51
>This executes jhipster-core functions:
>1. reads data from .jh/.jdl file(s)
>2. converts .jh/.jdl file(s) to entities json format
>3. writes entities json file(s) to disk
>
>Possible solution is to add new step between current 2. and 3. step:
>populate new entity descriptions with current fieldTestValues.
>Code can go to project jhipster-core module entity_parser.js into new
>function retainFieldTestValues.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/jhipster/generator-jhipster/pull/4628#issuecomment-267846550

-- 
Mathieu ABOU-AICHI

MathieuAA:
What do you mean by test values? Are you talking about test files here? 

Le 18 décembre 2016 22:05:39 GMT+01:00, Kaido Hallik <notifications@github.com> a écrit :
>Quick analysis of jhipster:import-jdl below.
>
>Work is done in project generator-jhipster in parseJDL:
>https://github.com/jhipster/generator-jhipster/blob/master/generators/import-jdl/index.js#L51
>This executes jhipster-core functions:
>1. reads data from .jh/.jdl file(s)
>2. converts .jh/.jdl file(s) to entities json format
>3. writes entities json file(s) to disk
>
>Possible solution is to add new step between current 2. and 3. step:
>populate new entity descriptions with current fieldTestValues.
>Code can go to project jhipster-core module entity_parser.js into new
>function retainFieldTestValues.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/jhipster/generator-jhipster/pull/4628#issuecomment-267846550

-- 
Mathieu ABOU-AICHI

kaidohallik:
In [this comment](https://github.com/jhipster/generator-jhipster/pull/4628#issuecomment-267558206) I described possible solutions for storing first time randomly generated patterned field test values.
@jdubois preferred first solution: store pattern field test values in entity json files.
These test values are used in server side test EntityNameResourceIntTest.java
Using same test values for patterned fields avoids test files changes in regeneration.

kaidohallik:
As a side comment: @deepu105 refactored yesterday entity generator. So my current implementation is not more valid. To allow this solution must add context option possibility to Generator.prototype.writeFilesToDisk (currently only interpolate is allowed).

But I think better solution is not to use context option any more as we agreed that random test values will be written into entity json files.
So final solution could be: entity json files are populated with test values before file generation and template is using this data as any other data.

kaidohallik:
I updated this PR, this is now working with current master.
Caution: don't merge this if you don't want to allow other options to templates beside interpolate.
This PR will be overwritten soon and no context option will be needed to template function.

kaidohallik:
@jdubois @MathieuAA Did you make any decision about generating/holding pattern field test values on Monday?

MathieuAA:
Hi, if I have the time look into that this week-end, I will. But don't expect any major breakthrough from me this time of the year. I'll do my best.

kaidohallik:
@MathieuAA No rush from me. I was just curious.

jdubois:
I just tested this with the following pattern: "^[_A-Za-z0-9-\\\\+]+(\\\\.[_A-Za-z0-9-]+)*"

- It generated a very long line of random characters
- Some lines had compilation errors, as they used the "\" character, which is for escaping Strings in Java

As this produces randomly compilation errors because of that  "\" character, we can't merge this like this

kaidohallik:
PR updated with the following changes:
* escaped test values in java
* if standard test value matches pattern then use this
* applied test value length reducing algorithm for randomly generated test values:
    - try standard test value length
    - try maxlength if this is available
    - try reduce to 100 symbols
    - finally truncate to 1000 symbols

jdubois:
Thanks @kaidohallik !
But at the moment you don't save the resulting Strings in the `.jhipster/entity.json` files yet? I think you should also add this, and then @MathieuAA can add this later in the JDL. Otherwise, it means files created with JDL will have this String that changes, it's not really bad if it's only temporary.

kaidohallik:
I have a better idea (why didn't I realize that before): instead of holding the test values in the `.jhipster/entity.json` files I can parse these from the existing old source files with regexp.
If the test value from the current `entityResourceIntTest.java` test file is still valid then use this old value.
In this way we retain also manual modifications made by the end user.
Is it ok that I go on with this new idea?

MathieuAA:
Good idea!

jdubois:
This is something we don't really do at the moment - we do parse some files with our "needles" system, so that's kind of the same idea, but nothing going this far.
It is indeed a great new idea, and something worth experimenting -> if it works well, maybe we could use this idea elsewhere

kaidohallik:
Reading from old source is implemented. I pushed new commit. Should I rebase or not?

jdubois:
I'm sorry but this time I tested validating an email, using:

    ^[_A-Za-z0-9-\\\\+]+(\\\\.[_A-Za-z0-9-]+)*@[A-Za-z0-9-]+(\\\\.[A-Za-z0-9]+)*(\\\\.[A-Za-z]{2,})$;

It did generate a rather long String, which is hard to read, but the tests are not passing. So for me this isn't working at the moment.

kaidohallik:
This pattern causes error also in current master.
For correct pattern must replace `\\\\` with `\\` everywhere.

I added commit where I configured randexp to generate minimalistic samples. Now samples should be readable.

kaidohallik:
@jdubois Sorry, no error in master with this pattern. Just this was not email pattern, to get email pattern must replace `\\\\` with `\\`.

jdubois:
Oh sorry, we have issues with such patterns anyway -> they are not the same in JS and in Java, and we can't do much about it. Let me test this again.

kaidohallik:
@jdubois Seems that java pattern is correctly modified for backslashes [here](https://github.com/jhipster/generator-jhipster/blob/f122021c4f422ce4f06424cfa544d2f4ea50cb79/generators/entity/index.js#L486-L489). But this place needs also escaping for quotes (in case pattern contains quotes), must add `.replace(/"/g, '\\"')` and seems that needs also escaping quote in front-end.

Your previous error could be caused by:
* randomly generated very long test value
* truncated to 1000 symbols
* first 1000 symbols didn't match pattern
Now generator generates small test values and no truncation any more, so this error should be gone.

jdubois:
OK, this looks much better indeed.

I still have one issue, which makes the whole thing fail: I'm using `{2,}` in my regexp, and the `{` character causes issues with Angular... So it should be escaped:

- @kaidohallik if you can escape it it would be perfect. Could you also test with i18n enabled? As this uses `{` also for the `translateValues` attribute
- If you don't have time to do it, then I'll still merge this -> we probably already have this issue (I didn't test, but this looks obvious we already have the issue), so in that case we're not adding more bugs, just staying with the same issues

kaidohallik:
@jdubois `{` issue wasn't related to i18n, this existed with and without i18n. I made new commit with changes in patterns which:
* fixes `"` in back-end, angular 1 and angular 2
* fixes `{` and `}` in angular 2

jdubois:
This is so complex!
OK let's merge this, this has been tested for quite some time now, and I'm not even sure so many people use regexps in their entities!

PR #4630, opened 2016-12-08
sass support for bootstrap 4 and font-awesome

dimeros:
Sass support for bootstrap 4 and font-awesome in angular-2.

### DIFF
--- a/generators/client-2/templates/src/main/webapp/app/_vendor.ts
+++ b/generators/client-2/templates/src/main/webapp/app/_vendor.ts
@@ -6,9 +6,17 @@
 // import 'angular-loading-bar';
 // import 'ng-file-upload';
 // import 'ng-infinite-scroll';
+
+<%_ if(useSass) { _%>
+import '../scss/vendor.scss';
+<%_} else { _%>
 import 'bootstrap/dist/css/bootstrap.min.css';
 import 'font-awesome/css/font-awesome.min.css';
+<%_ } _%>
+
+
+
 import '../content/css/main.css';
 <%_ if(useSass) { _%>
-import '../content/css/vendor.css';
+// import '../content/css/vendor.css'; deprecated???
 <%_ } _%>
--- a/generators/client-2/templates/_package.json
+++ b/generators/client-2/templates/_package.json
@@ -28,6 +28,7 @@
     "reflect-metadata": "0.1.8",
     "rxjs": "5.0.0-rc.4",
     "swagger-ui": "2.2.8",
+    "tether": "1.4.0",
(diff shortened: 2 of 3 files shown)


generators/client-2/files.js
```diff
@@ -40,8 +40,19 @@ const files = {
     sass: [
         {
             condition: generator => generator.useSass,
+            templates: [
+            '_.bootstraprc',
```
deepu105: what does these files do? could you please explain

generators/client-2/templates/src/main/webapp/app/_vendor.ts
```diff
@@ -6,9 +6,18 @@
...
 import 'bootstrap/dist/css/bootstrap.min.css';
 import 'font-awesome/css/font-awesome.min.css';
+<%_ } _%>
+
+
+
 import '../content/css/main.css';
 <%_ if(useSass) { _%>
-import '../content/css/vendor.css';
+// import '../content/css/vendor.css'; deprecated???
```
deepu105: why?

generators/client-2/templates/webpack/_webpack.common.js
```diff
@@ -95,7 +95,21 @@ module.exports = function (options) {
             ]),
             new webpack.ProvidePlugin({
                 $: "jquery",
-                jQuery: "jquery"
+                jQuery: "jquery",
+                "window.jQuery": "jquery",
+                'Tether': 'tether',
+                'window.Tether': 'tether',
+                Tooltip: "exports?Tooltip!bootstrap/js/dist/tooltip",
```
deepu105: we dont need any bootstrap JS stuff. We use only css from here. For js we use ng-bootstrap

dimeros: This configuration is needed because of the issue https://github.com/shakacode/bootstrap-loader/issues/172.  If I didn't include these changes, I should comment the bootstrap scripts from the _.bootstraprc file.

generators/client-2/templates/src/main/webapp/app/_vendor.ts
dimeros: The vendor.css is not generated. .

dimeros:
It is too complicated. I will leave bootstrap-loader. I don't close the pull request as I will commit a new fix until end of day.

deepu105:
yes I dont see why we need bootstrap loader anyway :P


Thanks & Regards,
Deepu

deepu105:
Please open a new PR to the master branch

PR #4632, opened 2016-12-08
Angular 2

deepu105:
Reference PR I will be merging this manually


deepu105:
This is merged but there are some TODOs and probably there will be test fails as well

PR #4634, opened 2016-12-08
Refactor the entity -> ResponseEntity(...) conversion, to remove a co…

gzsombor:
…uple of lines of code, and optimize

 the user updating, to avoid the 2 separate transactions - one for the update, and one for the load.

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/web/rest/_EntityResource.java
+++ b/generators/entity/templates/server/src/main/java/package/web/rest/_EntityResource.java
@@ -16,6 +16,7 @@ import <%=packageName%>.service.dto.<%= entityClass %>DTO;
 import <%=packageName%>.service.mapper.<%= entityClass %>Mapper;
 <%_ } } _%><% if (pagination != 'no') { %>
 import io.swagger.annotations.ApiParam;<% } %>
+import io.github.jhipster.web.util.ResponseUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;<% if (pagination != 'no') { %>
 import org.springframework.data.domain.Page;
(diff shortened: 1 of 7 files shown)


gzsombor:
@deepu105 , @jdubois : what do you think?

deepu105:
I'm ok with it. since @jdubois is refactoring common code to the lib I'll wait for his opinion here

jdubois:
Yes the code look better. I didn't check for the transaction, but I trust you, and that's indeed better.

My main concern is the use for this new utility class - I'm not a big fan of using a utility class everywhere. Anyway, that's was why I was waiting to have #4674 closed before commenting: this should go in the lib, and then that's less of an issue.

gzsombor:
Yes, it's a primary candidate for the lib 👍 :)

deepu105:
then may be we can merge it and then move it to lib

jdubois:
Just wait a little bit, I guess we can already merge the lib, I'm just waiting for some feedback

jdubois:
Now the library is merged, @gzsombor can you refactor your PR?

deepu105:
please hold on refactoring as I'm refactoring the entity folder structure

deepu105:
The source folder structure have changed so you might wanna update them

gzsombor:
I've moved to util class to the jhipster library. Unfortunately this means, that this pull request can't be merged until a new jhipster release is come out !

gzsombor:
@jdubois : could you do a jhipster-lib release?

jdubois:
@gzsombor done! Should be on Maven Central in less than 2 hours.

gzsombor:
Thanks, it seems now it's mergeable !

PR #4636, opened 2016-12-09
sass support for bootstrap 4 and font-awesome

dimeros:
sass support for bootstrap 4 and font-awesome


deepu105:
Looks much better

Issue #4549, opened 2016-11-25
Angular 2 release plan


deepu105:
This ticket is to plan and manage Angular 2 release as per original thread here https://github.com/jhipster/generator-jhipster/issues/3664 (which is locked for core team only)

discussion started https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jhipster-dev/nyworLcz6MQ/4ghzZklkCAAJ

Any further discussion on this matter should happen here

The steps are in order

**Step 1:**

- [x] Migrate everything from the app generator to NG2 (Means we dont have any ng1 controller, directive and service) and this is 98% done we only have few config files and interceptors to finish
- [x] Merge webpack (I'm 90% ok with the PR #4422 needs some polish regarding my comments and also need to open a specific ticket with things to do)
- [x] Merge to master (I would like to do this as there is gonna be quite a lot of conflicts and stuff, I already tried locally once so I know what are the things to expect)
- [x] Refactor from `/client`  and `/client-2` to `/client/templates/angularjs` and `/client/templates/angular` (I will do this, and its very important that we do this architecturally now rather than later)
- [x] Add a error log for entity generator to let people know that it doesn't yet work with ng2 (If you guys are patient to wait for say one more week then I can migrate this as well, but this can be done only after merge with master)
- [x] Fix atleast Karma tests (Personally I wouldn't like to ship a software with broken tests) protractor can wait as its an option
- [x] Freeze all development on 3.x and move it to maintenance mode (It should only get very important update or fix if required)
- [x] Release "JHipster 4.0.0" notice that angular2 support will be marked **BETA**

**Step 2:** We could go with multiple 4.0.x patch releases for this or can do a single minor with all below together 

- [x] Migrate entity sub generator if its not already done for above release
- [x] Remove ng1 dependencies from the generated angular2 app and make it work as a pure ng2 app (This should improve performance etc)
- [x] Migrate to BS4 and improve UI (for alpha the current UI with BS3 would do eventhough it has some glitches)
- [ ] Improve webpack config to add in missing steps from the gulp workflow
- [x] Fix protractor tests if not already done
- [x] Fix tslint warnings
- [ ] Improve performance, fix known/reported bugs and polish
- [ ] finish items from #4794 

**Step 3:** 

- [ ] Remove beta label from angular2 option (when all critical bugs known/reported are fixed, all dependencies updated to atleast RCs)
- [ ] Make angular 2 the default option in the question
[...]

PR #4646, opened 2016-12-11
Fix packaging with maven for Angular1

pascalgrimaud:
Related to my comment https://github.com/jhipster/generator-jhipster/issues/4549#issuecomment-266231078

Plz check closely my modifications before merging.

For angular1:
- avoid npm/yarn install, bower install, and gulp install when using dev profile
- missing bower install, and gulp build+test for prod profile

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1203,6 +1203,106 @@
             </dependencies>
             <build>
                 <plugins>
+                    <%_ if (!skipClient && clientFw === 'angular2') { _%>
+                    <plugin>
+                        <groupId>com.github.eirslett</groupId>
+                        <artifactId>frontend-maven-plugin</artifactId>
+                        <version>${frontend-maven-plugin.version}</version>
+                        <executions>
+                            <execution>
+                                <id>install node and npm</id>
+                                <goals>
+                                    <goal>install-node-and-npm</goal>
+                                </goals>
+                                <configuration>
+                                    <nodeVersion>v6.9.2</nodeVersion>
+                                    <npmVersion>3.10.9</npmVersion>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>npm install</id>
+                                <goals>
+                                    <goal>npm</goal>
+                                </goals>
+                                <configuration>
+                                    <arguments>install</arguments>
+                                </configuration>
+                            </execution>
+                            <execution>
+                                <id>webpack build</id>
+                                <goals>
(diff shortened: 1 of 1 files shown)


deepu105:
@jdubois can you take a look at this please?

generators/server/templates/_pom.xml
```diff
@@ -1203,37 +1203,12 @@
             </dependencies>
             <build>
                 <plugins>
-                    <%_ if (!skipClient) { _%>
+                    <%_ if (!skipClient && clientFw === 'angular2') { _%>
```
jdubois: Why do we have this new `clientFw` variable? There is already an `angularVersion` variable

pascalgrimaud: It has been changed by @deepu105

jdubois: Oh, OK, it's indeed a better name, it doesn't depend on Angular

pascalgrimaud: probably to make the possibility to use another thing than angular in the future

deepu105: yes exactly :)

generators/server/templates/_pom.xml
```diff
@@ -1203,37 +1203,12 @@
...
                     <plugin>
                         <groupId>com.github.eirslett</groupId>
                         <artifactId>frontend-maven-plugin</artifactId>
                         <version>${frontend-maven-plugin.version}</version>
                         <executions>
-                            <%_ if (yarnInstall) { _%>
```
deepu105: @pascalgrimaud does this mean we wont have yarn support when using ng2 in dev?

pascalgrimaud: @deepu105 : maybe I drop too much code here! We should support yarn too for ng2 of course, but I forget to ask you this question: in dev profile, is it mandatory to npm install / yarn for ng2 ?

jdubois: For me we should be able to use one or the other, in both AngularJS 1 and Angular 2. Then, in the future, if Yarn works well enough, I would concentrate on Yarn.

deepu105:
Actually given that you need to run webpack in order to do any development
we need not actually do everything from maven for dev we could just trigger
a webpack build post npm install like we used to do for gulp. let me take
another look at that tomorrow

Thanks & regards,
Deepu

pascalgrimaud:
Just updated my PR :
- yarn support when using ng2 in dev
- add properties for node.version npm.version yarn.version

PR #4649, opened 2016-12-11
Fix ng2 jhiPrefix compile errors

mehmet-aslan:
This only fixes compile errors if jhiPrefix is different from default. There could be more things to do to complete ng2-jhiPrefix support.

Also fixed URL in metrics.service.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
@@ -11,7 +11,7 @@ import {
     FilterPipe,
     OrderByPipe,
     <%_ if (enableTranslation){ _%>
-    <%=jhiPrefixCapitalized%>Translate,
+    JhiTranslate,
     <%=jhiPrefixCapitalized%>MissingTranslationHandler,
     <%=jhiPrefixCapitalized%>LanguageService,
     FindLanguageFromKeyPipe,
@@ -43,7 +43,7 @@ import {
         CapitalizePipe,
         KeysPipe,
         <%_ if (enableTranslation){ _%>
-        <%=jhiPrefixCapitalized%>Translate,
+        JhiTranslate,
         FindLanguageFromKeyPipe,
         <%_ } _%>
         JhiAlertComponent,
@@ -74,7 +74,7 @@ import {
         CapitalizePipe,
         KeysPipe,
         <%_ if (enableTranslation){ _%>
-        <%=jhiPrefixCapitalized%>Translate,
+        JhiTranslate,
         FindLanguageFromKeyPipe,
         <%_ } _%>
         JhiAlertComponent,
--- a/generators/client/templates/angular/src/main/webapp/app/admin/health/_health-modal.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/health/_health-modal.component.ts
@@ -11,7 +11,7 @@ export class <%=jhiPrefixCapitalized%>HealthModalComponent {
 
     currentHealth: any;
 
-    constructor(private healthService:JhiHealthService, public activeModal: NgbActiveModal) {}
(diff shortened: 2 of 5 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/language/_jhi-translate.directive.ts
```diff
@@ -7,7 +7,7 @@ import { Component, Input } from '@angular/core';
     selector: '[jhi-translate]',
     template: '<span [innerHTML]="jhiTranslate | translate:translateValues"></span>'
 })
-export class JhiTranslate {
+export class <%=jhiPrefixCapitalized%>Translate {
```
deepu105: this should be `Jhi` always as the directive will always be `jhi-translate`

generators/client/templates/angular/src/main/webapp/app/shared/component/_jhi-item-count.component.ts
```diff
@@ -9,7 +9,7 @@ import { Component, Input } from '@angular/core';
             of {{total}} items.
         </div>`
 })
-export class JhiItemCountComponent {
+export class <%=jhiPrefixCapitalized%>ItemCountComponent {
```
deepu105: this should be `Jhi` always as the directive will always be `jhi-item-count`

generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert.component.ts
```diff
@@ -10,7 +10,7 @@ import { AlertService } from './alert.service';
             </div>
         </div>`
 })
-export class JhiAlertComponent implements OnInit, OnDestroy {
+export class <%=jhiPrefixCapitalized%>AlertComponent implements OnInit, OnDestroy {
```
deepu105: this should be `Jhi` always as the directive will always be `jhi-alert`

generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
```diff
@@ -16,7 +16,7 @@ import { Subscription } from 'rxjs/Rx';
             </div>
         </div>`
 })
-export class JhiAlertErrorComponent implements OnDestroy {
+export class <%=jhiPrefixCapitalized%>AlertErrorComponent implements OnDestroy {
```
deepu105: this should be `Jhi` always as the directive will always be `jhi-*`

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -20,10 +20,10 @@ import {
     MaxbytesValidator,
     MinbytesValidator,
     ShowValidationDirective,
-    JhiItemCountComponent,
+    <%=jhiPrefixCapitalized%>ItemCountComponent,
```
deepu105: These shoudnt be changed this should be `Jhi` always as the directive will always be `jhi-*`

generators/client/templates/angular/src/main/webapp/app/shared/language/_jhi-translate.directive.ts
mehmet-aslan: This gives compile error at [shared-common.module.ts](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts#L14). This needs to be changed to `JhiTranslate` in this case.

mehmet-aslan: So `JhiMissingTranslationHandler` and `JhiLanguageService` classes could also be changed from `jhiPrefix` to `Jhi` to achieve this.

I updated PR without doing this to only fix compile errors. They could be changed with another PR if needed.

PR #4660, opened 2016-12-12
Spring constructor injection

### LINKED ISSUE #4554: Use constructor-based injection with Spring
Sonar tells us we should use constructor-based injection with Spring.

I've always been a big fan of field-based injection, and for me there was a consensus in the Spring team to say that was the best choice (but now I see newer members of Pivotal would rather do constructor-based injection).

Anyway, I'm not going to recap all the pros/cons for each choice here, just that I'm changing my mind on this for a very specific new reason: Angular 2. I would like us to have something consistent in the back and front technologies, and with Angular 2 we are using constructor-based injection.

Can you vote on this comment with the GitHub emoticons? Vote up if you want constructor-based, and down if you want to keep field-based injection.

jdubois:
Migration to constructor-based injection.

Apart from the usual stuff about dependency injection, a few non-obvious points:

- More homogenous codebase (we had a mix of different styles, now everything is the same)
- This is harder to do than expected in JHipster, as it's quite hard to generate correct constructors when there are lots of options
- Removes CDI!! So @Inject isn't used anymore, it's not even included in the `pom.xml` anymore!

Fix #4554


xetys:
wouldn't this break any code generated with earlier version, which have any kind of customization? In my case, I try to keep to the coding style as the generated project when creating new classes, so using `@Inject` and field injection. 

So for a "clean" JHipster project, without any customs in pom/gradle, only default config with some logic inside the services, which is totally ok for updates (just skip services, force everything else), it is now required to manually merge the buildfile.

This is not a big issue, as users have to manually merge the buildfile anyway, as soon they change it for some reasons. However I am not sure, if removing `@Inject` is a good idea

deepu105:
@xetys yes it might but we are doing a major version release and hence its ok to break code. we could do it only for major versions

xetys:
ok, true point :+1:

jdubois:
Yes that's why I did this so quickly, because I knew we wouldn't have a new major version before a long time!

PR #4664, opened 2016-12-12
added bootstrap js

dimeros:
WIP(4407)
The navbar menu doesnt work in mobile view
bootstrap show/hide elements based on screensize doesnt work.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.html
@@ -1,6 +1,6 @@
 <nav class="navbar navbar-light bg-faded">
     <div class="container">
-        <button class="navbar-toggler hidden-lg-up" type="button" data-toggle="collapse" data-target="#navbarResponsive" aria-controls="navbarResponsive" aria-expanded="false" aria-label="Toggle navigation"></button>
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/src/main/webapp/app/_vendor.ts
```diff
@@ -6,6 +6,7 @@
 // import 'angular-loading-bar';
 // import 'ng-file-upload';
 // import 'ng-infinite-scroll';
+import 'bootstrap/dist/js/bootstrap.min.js';
```
deepu105: is bootstrap using js to show hide stuff instead of css? I would really like to avoid bootstrap js as we already have ng-bootstrap doing the same stuff in angular way. can we find a way to get these working using ng bootstrap directives

dimeros: is bootstrap using js to show hide stuff instead of css?  Yes. And I see that it needs a configuration regarding tether as an error is displayed.

dimeros: Ok I will search for this alternative

dimeros:
Wrong title anymore. Is this change enough for these two issues at the moment?

deepu105:
not sure i'll merge it anyway and check later

PR #4665, opened 2016-12-12
Ng2 social login forms submit fix

flaviencathala:
This is a fix for the forms of social login.
The submit buttons doesn't work without this fix.
This is not very clean, but it works.


generators/client/templates/angular/src/main/webapp/app/shared/social/_social.component.ts
```diff
@@ -34,4 +34,8 @@ export class <%=jhiPrefixCapitalized%>SocialComponent implements OnInit {
         this.csrf = this.csrfService.getCSRF();
     }
 
+    submit()
+    {
+        $('#form' + this.label).submit();
```
gzsombor: Instead of this frightening jqueryism, please add ngNoForm to the form, that will fix the issue - the original problem is that, as Angular2 founds the 'form' tag, it captures it, and adds it's custom javascript magic. The ngNoForm attribute would prevent this.

flaviencathala: @gzsombor Thanks a lot !

flaviencathala:
@deepu105 It's a clean fix now

PR #4666, opened 2016-12-12
Ng2 metrics update

VictorADS:


generators/client/templates/angular/src/main/webapp/app/admin/metrics/_metrics.component.ts
```diff
@@ -31,15 +31,17 @@ export class <%=jhiPrefixCapitalized%>MetricsMonitoringComponent implements OnIn
...
+                    let value = metrics.gauges[key].value;
                     // remove gets or puts
                     let index = key.lastIndexOf('.');
                     let newKey = key.substr(0, index);
 
                     // Keep the name of the domain
-                    index = newKey.lastIndexOf('.');
                     this.cachesStats[newKey] = {
-                        'name': newKey.substr(index + 1),
+                        'name': newKey.substr(18),
```
deepu105: you do have const :smile:

deepu105: btw I would rather do `const JCACHE_KEY = 'jcache.statistics';` and then do `JCACHE_KEY.length` here

PR #4669, opened 2016-12-12
Fix highlighting of the navbar

gzsombor:
It seems, that uiSrefActive should be on the tag with the 'dropdown-item' class, I guess, it's a bootstrap change


gzsombor:
Hope, it wont cause any problem

PR #4671, opened 2016-12-13
remove touch command

### LINKED ISSUE #4670: Dockerfile touch command
Our Dockerfile contains the line `RUN sh -c 'touch /app.war'`.  If you run `docker history <image name>` on an app's docker image, you can see that the command doubles the size of the WAR layer.

```
68c432e8ff1a        5 seconds ago       /bin/sh -c sh -c 'touch /app.war'               83.2 MB             
e172ad022017        7 seconds ago       /bin/sh -c #(nop) ADD file:b447fcbb42ca6c1...   83.2 MB  
```

The only source where I could find the `touch` command is in the Spring Boot Docker [Getting Started](https://spring.io/guides/gs/spring-boot-docker/) guide with the explanation:

> You can use a RUN command to "touch" the jar file so that it has a file modification time (Docker creates all container files in an "unmodified" state by default). This actually isn’t important for the simple app that we wrote, but any static content (e.g. "index.html") would require the file to have a modification time.

The behavior is the same with and without the command.
 - When I sh into the image, the WAR file has a last modified date of the image build time.
 - The last-modified header for the index.html and i18n match the image build time
 - The last-modified header for CSS/JS and images match the image deploy time.  

When updating a deployed docker app, only new layers of the image are downloaded.  Currently, we download the WAR file twice for each update.  With this change, we would only download one layer and docker images built will be 30% smaller (50-80MB).

##### **Reproduce the error**
To see the docker image layers from a sample monolith:
```
docker pull jhipster/jhipster-sample-app
docker history jhipster/jhipster-sample-app
```
To test the docker image without the `touch` command, you can use this app.yml (the other image is the same app but with the `touch` command):
```
version: '2'
services:
    jhipstersampleapplication-app:
        image: ruddell/jhipstersampleapplication
#        image: jhipster/jhipster-sample-app
        external_links:
            - jhipstersampleapplication-mysql:mysql
        environment:
            - SPRING_PROFILES_ACTIVE=prod,swagger
            - SPRING_DATASOURCE_URL=jdbc:mysql://mysql:3306/jhipstersampleapplication?useUnicode=true&characterEncoding=utf8&useSSL=false
            - JHIPSTER_SLEEP=10 # gives time for the database to boot before the application
        ports:
            - 8080:8080
    jhipstersampleapplication-mysql:
        image: mysql:5.7.13
        environment:
            - MYSQL_USER=root
            - MYSQL_ALLOW_EMPTY_PASSWORD=yes
            - MYSQL_DATABASE=jhipstersampleapplication
        command: mysqld --lower_case_table_names=1 --skip-ssl

```
Is there a reason for the `touch` command that I'm missing?

ruddell:
Fix #4670

### DIFF
--- a/generators/server/templates/src/main/docker/_Dockerfile
+++ b/generators/server/templates/src/main/docker/_Dockerfile
@@ -5,7 +5,6 @@ ENV JHIPSTER_SLEEP 0
 # add directly the war
 ADD *.war /app.war
 
-RUN sh -c 'touch /app.war'
 VOLUME /tmp
 EXPOSE <%= serverPort %><% if (hibernateCache == 'hazelcast') { %> 5701/udp<% } %>
 CMD echo "The application will start in ${JHIPSTER_SLEEP}s..." && \


pascalgrimaud:
Can someone test this change on Windows?

pascalgrimaud:
I manage to test it on Windows and everything seems OK.
Good job @ruddell !

PR #4673, opened 2016-12-13
[WIP] Ng2 tracker service

flaviencathala:
@christopherthielen Could you please take a look we still have $rootScope in tracker service.


generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
```diff
@@ -39,7 +37,7 @@ export class <%=jhiPrefixCapitalized%>TrackerService {
     connect () {
         if (this.connectedPromise === null) this.connection = this.createConnection();
         //building absolute path so that websocket doesnt fail when deploying with a context path
-        var loc = this.$window.location;
```
deepu105: we should either continue to use the`Window` object from ng2 or do something like http://juristr.com/blog/2016/09/ng2-get-window-ref/

dimeros:
I see that for some reason, the connection cannot be established. 
Debugging it says that the sockJs is undefined. @flaviencathala is it reproducible to your side?

flaviencathala:
@dimeros Nope I've never seen this error

deepu105:
@flaviencathala are you doing anything here?

generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
deepu105: please change this

deepu105:
@flaviencathala I'm closing this. If you are working on this please open a new PR with cleaner commits

PR #4676, opened 2016-12-14
Migrating our pages to Bootstrap 4

VictorADS:


generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.ts
```diff
@@ -4,6 +4,7 @@ import { DatePipe } from '@angular/common';
 import { Audit } from './audit.model';
 import { AuditsService } from './audits.service';
 import { ParseLinks } from '../../shared';
+import { ITEMS_PER_PAGE } from '../../shared/constants/pagination.constants';
```
deepu105: please import this in the above import as `{ ParseLinks, ITEMS_PER_PAGE } from '../../shared'`

generators/client/templates/angular/src/main/webapp/app/account/settings/_settings.component.html
```diff
@@ -54,19 +54,19 @@ <h2 jhi-translate="settings.title" translate-values="{username: '{{settingsAccou
...
                            *ngIf="emailInput.errors.email" jhi-translate="global.messages.validate.email.invalid">
                             Your e-mail is invalid.
                         </p>
-                        <p class="help-block"
+                        <p class="form-text text-muted"
                            *ngIf="emailInput.errors.minlength" jhi-translate="global.messages.validate.email.minlength">
                             Your e-mail is required to be at least 5 characters.
                         </p>
-                        <p class="help-block"
+                        <p class="form-text text-muted"
```
deepu105: does this still highlight in red?

generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html
```diff
@@ -36,7 +36,8 @@
     </table>
 
     <!-- TODO : use ng-bootstrap to handle pagination -->
```
deepu105: if this is working can you remove the TODO

generators/client/templates/angular/src/main/webapp/app/admin/configuration/_configuration.component.html
```diff
@@ -3,12 +3,12 @@ <h2 jhi-translate="configuration.title">Configuration</h2>
 
     <span jhi-translate="configuration.filter">Filter (by prefix)</span> <input type="text" [(ngModel)]="filter" class="form-control">
     <label>Spring configuration</label>
-    <table class="table table-condensed table-striped table-bordered table-responsive"
+    <table class="table table-sm table-striped table-bordered"
```
deepu105: is tables responsive by default in bs4 now?

generators/client/templates/angular/src/main/webapp/content/css/_main.css
```diff
@@ -344,6 +344,7 @@ ui bootstrap tweaks
 }
 
 
+
```
deepu105: can we remove empty double lines

generators/client/templates/angular/src/main/webapp/app/account/settings/_settings.component.html
deepu105: I think it s much better to use `form-text text-danger` here. And it would be nice if we place it in a `<small> element inside the `<p>` but thats optional

generators/client/templates/angular/src/main/webapp/app/admin/configuration/_configuration.component.html
sendilkumarn: I guess we need `table-responsive` there

VictorADS: This is my bad. I removed table-responsive because it would cause some strange issues (table not taking 100% of the width or overflow in some modals) but I misread the doc and actually I just have to wrap the table inside a table-responsive !

sendilkumarn: `Responsive tables no longer require a wrapping element. Instead, just put the .table-responsive right on the <table>.`

to be exact

VictorADS: It seems so Sendil, thanks ! But it's quite interesting, in some pages I have to wrap the element or it doesn't work as expected ( if you have the time try it in audits.component.html)

generators/client/templates/angular/src/main/webapp/app/account/settings/_settings.component.html
VictorADS: Highlighting doesn't work ATM And i have no idea how to do it, maybe create our own component ?

deepu105: I dont understand what you meant but dont worry. Ill do it

PR #4677, opened 2016-12-14
Add a gradle flag to build using downloaded node/npm

### LINKED ISSUE #4667: Use the gradle node plugin to download node, npm, ...
Just like with the frontend-maven-plugin, we should setup gradle to install node, npm and such from a gradle task.
This way, the only requirement to build a JHipster project is JDK8.

cbornet:
* Add a nodeDl property to use a downloaded node/npm (as with frontend-maven-plugin). Default is deactivated.
* Add a workaround for gradle-node-plugin/issues/134 (only works for npm so even if yarn was used at project generation, it fallsback to npm if nodeDl flag is on)

Fix #4667

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -1,3 +1,4 @@
+import groovy.json.JsonSlurper
 import org.gradle.internal.os.OperatingSystem
 
 buildscript {
@@ -15,10 +16,7 @@ buildscript {
         classpath "org.springframework.boot:spring-boot-gradle-plugin:${spring_boot_version}"
         classpath "org.springframework.build.gradle:propdeps-plugin:0.0.7"
         <%_ if(!skipClient) { _%>
-        classpath "com.moowork.gradle:gradle-node-plugin:0.14"
-            <%_ if(clientFw === 'angular1') { _%>
-        classpath "com.moowork.gradle:gradle-gulp-plugin:0.13"
-            <%_ } _%>
+        classpath "com.moowork.gradle:gradle-node-plugin:1.0.1"
         <%_ } _%>
         classpath "io.spring.gradle:dependency-management-plugin:0.6.1.RELEASE"
         //jhipster-needle-gradle-buildscript-dependency - JHipster will add additional gradle build script plugins here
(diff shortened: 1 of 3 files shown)


cbornet:
With this you can build with a bare openjdk image
```
./gradlew clean
./gradlew npmInstall -PnodeDl
./gradlew test -PnodeDl
./gradlew gulp_test -PnodeDl
./gradlew bootRepackage -x test -Pprod -PnodeDl
```

generators/server/templates/_build.gradle
```diff
@@ -15,10 +16,7 @@ buildscript {
         classpath "org.springframework.boot:spring-boot-gradle-plugin:${spring_boot_version}"
         classpath "org.springframework.build.gradle:propdeps-plugin:0.0.7"
         <%_ if(!skipClient) { _%>
-        classpath "com.moowork.gradle:gradle-node-plugin:0.14"
-            <%_ if(clientFw === 'angular1') { _%>
-        classpath "com.moowork.gradle:gradle-gulp-plugin:0.13"
```
deepu105: is gulp included in the node plugin?

generators/server/templates/gradle/_yeoman.gradle
```diff
@@ -1,12 +1,20 @@
...
 
-task bower(type: Exec) {
-    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
-        commandLine 'cmd', '/c', 'bower', 'install'
-    }else{
-        commandLine 'bower', 'install'
-    }
+task bower(type: NodeTask) {
+    description = "Installs dependencies using Bower"
+    script = file("${project.projectDir}/node_modules/bower/bin/bower")
```
deepu105: I dont think the path is same in windows/unix can you please check?

generators/server/templates/_build.gradle
```diff
@@ -369,7 +373,29 @@ task wrapper(type: Wrapper) {
 
 task stage(dependsOn: 'bootRepackage') {
 }
+<% if(!skipClient) { %>
+if (project.hasProperty('nodeDl')) {
```
deepu105: im not happy with the property name can we have something better may be like `nodeInstall` or `localNode` or something

generators/server/templates/_build.gradle
cbornet: yes

generators/server/templates/_build.gradle
cbornet: OK for nodeInstall

generators/server/templates/gradle/_yeoman.gradle
deepu105: ya but the I saw somebody added a os check on our new webpack task on one of the gradle file, were based on os the bin path was different

cbornet: Yes. On Linux node_modules/.bin/webpack is a symlink to node_modules/webpack/bin/webpack.js. So IMO this platform check was useless and both could use node_modules/webpack/bin/webpack.js directly.

deepu105:
OH ok then we can remove it. DO you wanna do it as part of this PR?


Thanks & Regards,
Deepu

On Wed, Dec 14, 2016 at 7:24 PM, Christophe Bornet <notifications@github.com

cbornet:
Yep.

cbornet:
@deepu105 Is your review OK ? Once this is merge, I will update #4600

PR #4681, opened 2016-12-14
Adding color to output when using Docker image

pascalgrimaud:
...because it's soon :christmas_tree: 
adding some colors when using the Docker image built from application


jdubois:
I had no idea you could do that! I'm eager to try!

PR #4684, opened 2016-12-15
Better display of Spring Profiles when the application is starting

pascalgrimaud:
I always scroll down/up to find the Spring profile(s) between 2 log.info as soon as the application started.


PierreBesson:
Nice, this is also a problem for me 
I also wonder if we could do a refresh of the banner for 4.0.

cbornet:
You want to put an image banner :smile: . It's now directly supported by SB

PR #4687, opened 2016-12-15
bug fix IE11 endsWidth issue in notification interceptor

### LINKED ISSUE #4680: IE11 stop working after upgrade
* **Overview of the request**

I want to report a bug in src/main/webapp/app/blocks/interceptor/notification.interceptor.js This bug is occuring in IE11. Edge and other browsers are working ok

* **Motivation for or Use Case** 

My customers woudl like to use IE11

* **Browsers and Operating System** 
IE11 , Windows 10

Only IE11 

* **Suggest a Fix** 

Here is problematic part

```
var headers = Object.keys(response.headers()).filter(function (header) {
                return header.endsWith('app-alert') || header.endsWith('app-params')
            }).sort();
```

IE11 hangs as javascript failed on **endsWidth** - result is partially loaded page

 I have found fix but not sure if it is best one so please just use it as sample which make it work for me

```
var headers = Object.keys(response.headers()).filter(function (header) {
                return header.indexOf('app-alert', header.length - 'app-alert'.length) !== -1 || header.indexOf('app-params', header.length - 'app-params'.length) !== -1;
            }).sort();
```

rohajda:
endsWidth is not supported by IE11 and Opera. Use indexOf instead

Fix #4680

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
@@ -18,8 +18,7 @@ export class NotificationInterceptor extends HttpInterceptable {
             let headers = [];
             let i;
             for(i = 0; i < arr.length; i++){
-                if(arr[i][0].indexOf('app-alert', arr[i][0].length - 'app-alert'.length) !== -1 ||
-                    arr[i][0].indexOf('app-params', arr[i][0].length - 'app-params'.length) !== -1) {
+                if(arr[i][0].endsWith('app-alert') || arr[i][0].endsWith('app-params')) {
                     headers.push(arr[i][0]);
                 }
             }
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
@@ -26,7 +26,7 @@ export class NotificationInterceptor extends HttpInterceptable {
             headers.sort();
             let alertKey = headers.length >= 1 ? error.headers.get(headers[0]) : null;
             if(typeof alertKey === 'string'){
-                //AlertService.sucnpm linkcess(alertKey, { param : response.headers(headers[1])});
+                //AlertService.success(alertKey, { param : response.headers(headers[1])});
             }
             return Observable.throw(error);
         });
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
@@ -18,14 +18,15 @@ export class NotificationInterceptor extends HttpInterceptable {
             let headers = [];
             let i;
             for(i = 0; i < arr.length; i++){
-                if(arr[i][0].endsWith('app-alert') || arr[i][0].endsWith('app-params')) {
+                if(arr[i][0].indexOf('app-alert', arr[i][0].length - 'app-alert'.length) !== -1 ||
+                    arr[i][0].indexOf('app-params', arr[i][0].length - 'app-params'.length) !== -1) {
                     headers.push(arr[i][0]);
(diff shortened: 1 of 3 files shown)


generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
```diff
@@ -18,14 +18,15 @@ export class NotificationInterceptor extends HttpInterceptable {
...
+                if(arr[i][0].indexOf('app-alert', arr[i][0].length - 'app-alert'.length) !== -1 ||
+                    arr[i][0].indexOf('app-params', arr[i][0].length - 'app-params'.length) !== -1) {
                     headers.push(arr[i][0]);
                 }
             }
             headers.sort();
             let alertKey = headers.length >= 1 ? error.headers.get(headers[0]) : null;
             if(typeof alertKey === 'string'){
-                //AlertService.success(alertKey, { param : response.headers(headers[1])});
+                //AlertService.sucnpm linkcess(alertKey, { param : response.headers(headers[1])});
```
notflorian: I think there is a small typo here, it seems you typed npm link in the middle of AlertService.success ;)

rohajda: ups, it is comment but shoudl not be there. Sorry. I am going to fix it

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
```diff
@@ -18,7 +18,8 @@ export class NotificationInterceptor extends HttpInterceptable {
             let headers = [];
             let i;
             for(i = 0; i < arr.length; i++){
-                if(arr[i][0].endsWith('app-alert') || arr[i][0].endsWith('app-params')) {
+                if(arr[i][0].indexOf('app-alert', arr[i][0].length - 'app-alert'.length) !== -1 ||
```
deepu105: no need to change here as this is typescript. You need to change this in the templates/angularjs folder inside the client which is in JS and will be affected

deepu105: ok you have done that so please remove the changes on this file alone

generators/client/templates/angularjs/src/main/webapp/app/blocks/interceptor/_notification.interceptor.js
```diff
@@ -16,7 +16,7 @@
 
         function response (response) {
             var headers = Object.keys(response.headers()).filter(function (header) {
-                return header.endsWith('app-alert') || header.endsWith('app-params');
```
mehmet-aslan: Very similar line (using `endsWith`) exists at `_alert-error.directive.js`. It could need to be changed as well.

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
rohajda: ok, i will remove that change. How about comment below where there are other places with endsWidth - should i scan angularjs *.js for it and fix?

generators/client/templates/angularjs/src/main/webapp/app/blocks/interceptor/_notification.interceptor.js
rohajda: i have fixed that too. Thanks for hint

deepu105:
@jdubois if you wanna do a patch release for 3.x we need to hotfix this one into that branch but its very minor issue for only people using IE11 so I wouldnt bother actually :smiling_imp:

jdubois:
Depends on the time I have, as always...

Le 15 déc. 2016 5:01 PM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

jdubois:
I cherry-picked the changes in https://github.com/jhipster/generator-jhipster/commit/3eecf5b7968010b9eaa4b463f580c54fcfd83446 on the 3.x maintenance branch.

PR #4698, opened 2016-12-16
[WIP] Ngb pagination

VictorADS:
Hello guys, I'd like to get your help on this one.
Currently the pagination works and we can switch between them.
The problem is that in the URL is encoded and  the comma ','  for the parameter 'sort'  is changed.
Thus the parseAscending method  in PaginationUtil doesn't work.

Should I change PaginationUtil method ? 
Or should I try to have a raw url (ie : not encoded) ?

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
@@ -78,9 +78,9 @@
         </table>
     </div>
     <%_ if (databaseType !== 'cassandra') { _%>
-    <div class="text-xs-center">
+    <div class="text-xs-center" *ngIf="users">
         <jhi-item-count [page]="page" [total]="queryCount" [items-per-page]="itemsPerPage"></jhi-item-count>
-        <ngb-pagination size="sm" [collectionSize]="totalItems" [(page)]="page" [pageSize]="itemsPerPage" (click)="loadPage(page)"></ngb-pagination>
+        <ngb-pagination [collectionSize]="totalItems" [(page)]="page" (pageChange)="loadPage(page)"></ngb-pagination>
     </div>
     <%_ } _%>
 </div>
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
@@ -1,5 +1,6 @@
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { CookieService } from 'angular2-cookie/services/cookies.service';
+import {PaginationConfig} from '../blocks/config/uib-pagination.config';
 
 import {
     <%=angular2AppName%>SharedLibsModule,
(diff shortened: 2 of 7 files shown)


deepu105:
@geraldhumphries I think you did the sorting stuff rite? do you have an opinion here?
@VictorADS if its easy to modify the util method to accept encoded URL then go for it

generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
```diff
@@ -58,7 +57,7 @@ import {
         <%_ if (enableSocialSignIn) { _%>
         SocialService,
         <%_ } _%>
-        AuthService
+        AuthService,
```
deepu105: stray comma here

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
```diff
@@ -78,9 +78,9 @@ <h2 jhi-translate="userManagement.home.title">Users</h2>
         </table>
     </div>
     <%_ if (databaseType !== 'cassandra') { _%>
-    <div class="text-xs-center">
+    <div class="text-xs-center" *ngIf="users">
         <jhi-item-count [page]="page" [total]="queryCount" [items-per-page]="itemsPerPage"></jhi-item-count>
-        <ngb-pagination size="sm" [collectionSize]="totalItems" [(page)]="page" [pageSize]="itemsPerPage" (click)="loadPage(page)"></ngb-pagination>
```
geraldhumphries: `size="sm"` is removed here, which is a change to the UI. Unless this is an obsolete/unused attribute I would leave it in.

generators/client/templates/angular/src/main/webapp/app/shared/service/_pagination-util.service.ts
```diff
@@ -8,7 +8,10 @@ export class PaginationUtil  {
         if (sortArray.length > 1){
             return sort.split(',').slice(-1)[0] === 'asc';
         } else {
-            // default to true if no sort defined
```
geraldhumphries: I would readd this comment above `return true;`

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
VictorADS: Now it is set via the PaginationConfig file

generators/client/templates/angular/src/main/webapp/app/shared/service/_pagination-util.service.ts
```diff
@@ -8,7 +8,10 @@ export class PaginationUtil  {
...
+            sortArray = sort.split('%2C');
+            if (sortArray.length > 1){
+                return sort.split('%2C').slice(-1)[0] === 'asc';
+            }
```
geraldhumphries: This code is essentially a duplicate of the rest of the code in this function. It should be refactored to be cleaner. Also, return `sortArray.slice(-1)[0] === 'asc';` unless there's a reason for it to split the string twice (I know that line was preexisting).

geraldhumphries:
I never considered encoded URLs so if this fixes a bug involving that then I'm all for it.

PR #4706, opened 2016-12-18
Save clientPackageManager in yo-rc.json file

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/4317#issuecomment-267764276

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -176,6 +176,14 @@ module.exports = JhipsterGenerator.extend({
                     this.enableTranslation = true;
                 }
             }
+            this.clientPackageManager = this.config.get('clientPackageManager');
+            if (!this.clientPackageManager) {
+                if (this.yarnInstall) {
+                    this.clientPackageManager = 'yarn';
+                } else {
+                    this.clientPackageManager = 'npm';
+                }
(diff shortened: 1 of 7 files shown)


deepu105:
Take a look at my comment on the ticket. You dont have to rename it everywhere. You could use an in memory variable to retain yarnInstall/npmInstall as it would look better on switches

pascalgrimaud:
@deepu105 : yes I agree it would be better for switches, but I was not happy with the variable name `yarnInstall` which I choose (because for me, `yarnInstall` is like a test if yarn is installed or not). I'd prefer to have a test with:
- `clientPackageManager === 'yarn'` instead of `yarnInstall`
- `clientPackageManager === 'npm'` instead of `!yarnInstall`

And these tests are in pom.xml and gradle files, not everywhere. So if you're ok, I'd prefer keep this new version.

What I did in this PR:
- I rebase with master
- I add npm test for clientFramework

deepu105:
ok


Thanks & Regards,
Deepu

jdubois:
Yes I agree with @pascalgrimaud as it's easier to understand for people who want to contribute

Issue #4685, opened 2016-12-15
Incorrect ordering properties/libs in pom.xml


pascalgrimaud:
By adding a lot of librairies, the current pom.xml is not correctly ordered.

Someone in your project could do a `mvn clean install` and commit it.
Then, upgrading to a newer JHipster version is a pain, specially for the pom.xml file, because there is a lot of changes.

Step to reproduce:
1) generate a Maven projet
2) git init
3) git add .
4) git commit -m "init"
5) mvn clean install
6) git status

So you will see a lot of changes.

I would like to reorder everything. Are you ok with that?

jdubois:
Yes but there is some impact with my current "lib". I'm pretty sure we will merge it, so can you do this in `migration-to-jhipster-library`branch?

pascalgrimaud:
Sure, if all agree, I can begin to work on it
But it's not urgent, it is just something I notice when using mvn clean install

jdubois:
Closing as the PR has been merged, and properties are now re-ordered correctly.

pascalgrimaud:
In fact, it's not finished.
All dependencies are not ordered correctly. Launching `mvn install` will still change the pom.xml.

I started to work on it, but it's a huge work, so I think I will do it later

jdubois:
OK, that's not very important...

PR #4721, opened 2016-12-20
Reorder the properties in pom.xml file

pascalgrimaud:
1st step for this ticket https://github.com/jhipster/generator-jhipster/issues/4685
I prefer to make little PR, so it will be easier

Note: I added a new properties `jhipster.version`, are you ok with that @jdubois ?

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -183,7 +183,7 @@ repositories {
 }
 
 dependencies {
-    compile "io.github.jhipster:jhipster:${jhipster.server.version}"
+    compile "io.github.jhipster:jhipster:${jhipster_server_version}"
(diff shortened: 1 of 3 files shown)


jdubois:
Yes for `jhipster.version` but:

- You need to make the same property for Gradle
- As this is the server lib, I would call the property `jhipster.server.version` so it's clear it's not the generator itself
- We need to agree on the versioning scheme. Here what I intend to do: for JHipster `v4.0.0`, I will release the JHipster lib `4.0.0` (note there is no `v` as with Maven people usually only use numbers). If there is an issue with the lib (as it's going to be released before the generator, for testing, etc.), it will have a 4th digit: `4.0.0.1`, `4.0.0.2`, etc. I didn't check, but I guess it follows Semver.

PR #4683, opened 2016-12-15
possible fix of hanging gradlew in Windows

kaidohallik:
console output doesn't contain broken texts any more

fix jhipster/generator-jhipster#4675

Issue #4675, opened 2016-12-14
code compilation gets stuck after running gradlew


AlienWareSumant:
I have done everything as the jhipster installation guide suggests. But when I try to run the code using gradlew, it works for sometime and then gets stuck at gulpconstantdev. No change was done in any file after running "yo jhipster" from cmd. Please find the image attached below and let me know what the resolution should be.
![jhipsterissue](https://cloud.githubusercontent.com/assets/13749340/21182336/876d98e0-c228-11e6-9dec-fbdb144de5c4.JPG)

jdubois:
This ticket has been closed as the guidelines are not followed.

Tickets must follow our [Guidelines](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#submitting-an-issue), as mentioned in:

1. our [Readme file on the front page of the project](https://github.com/jhipster/generator-jhipster/blob/master/README.md), 
2. the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) and
3. our [Help page](http://jhipster.github.io/help/)

We have also created a template on the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) to help you follow those guidelines.

kaidohallik:
I have the same issue. I have 2 workarounds:
1. if using `gradlew --info` then everything works again
2. closing current command window and opening new command window, then `gradlew` works for a while in this new command window

##### **JHipster Version(s)**

```
jhipster@0.0.0 c:\projects\jhipster-test
`-- generator-jhipster@3.12.1

```


##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

```yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "3.12.1",
    "baseName": "jhipster",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "buildTool": "gradle",
    "enableSocialSignIn": true,
    "rememberMeKey": "65c1391bc15928034869154f58a65c34bd70e7f5",
    "useSass": true,
    "applicationType": "monolith",
[...]

PR #4730, opened 2016-12-21
fix random error/hanging of gradlew after executing gulp plugin in Windows

kaidohallik:
This PR is enhanced version of fix jhipster/generator-jhipster#4683 where gulp bufferOutput=true is assigned only in Windows.
Opening new PR because GitHub doesn't allow updates on closed PR-s.

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -86,6 +86,13 @@ if (OperatingSystem.current().isWindows()) {
             classpath = files("$buildDir/classes/main", "$buildDir/resources/main", pathingJar.archivePath)
         }
     }
+    <%_ if(!skipClient && clientFramework === 'angular1') { _%>
+
+    // avoid random error and hanging in windows, see https://github.com/jhipster/generator-jhipster/pull/4683
+    gulp {
+        bufferOutput = true
+    }
+    <%_ } _%>
 } else {
     bootRun {
         addResources = false


generators/server/templates/_build.gradle
```diff
@@ -86,6 +86,13 @@ if (OperatingSystem.current().isWindows()) {
             classpath = files("$buildDir/classes/main", "$buildDir/resources/main", pathingJar.archivePath)
         }
     }
+    <%_ if(!skipClient && clientFw === 'angular1') { _%>
```
pascalgrimaud: clientFw -> clientFramework

kaidohallik:
@pascalgrimaud thanks for pointing out clientFw error, I updated the PR now

kaidohallik:
I encounter this problem sometimes also in `yarn_install` task.
For gulp tasks current bufferOutput=true solution is uncomfortable for longer gulp tasks in prod profile - do not get feedback instantly.

I propose solution:
* I write new PR for tip `Avoid gradlew random hanging and error in Windows` in [https://jhipster.github.io/tips/](https://jhipster.github.io/tips/)
* I write PR which reverts changes from this PR and instead outputs information at the end of every node task in Windows: `In case of error or hanging check https://jhipster.github.io/tips/0NN_tip_avoid_gradlew_random_hanging_and_error_in_windows.html`

kaidohallik:
Possible tip content below.

# Avoid gradlew random hanging and error in Windows

Sometimes `gradlew` hangs or gives following error in Windows at the end of the node tasks:
```

* What went wrong:
Unable to process incoming event 'ProgressComplete ' (ProgressCompleteEvent)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

No hanging and no error if instead of `gradlew` running:
* `gradlew --info`
* `gradlew --debug`
* `gradlew --console plain`

Other sometimes working solutions:
* increase command line window Screen Buffer Size, for example to maximum possible value
* open new command line window and use this new window for executing gradle tasks

If problem occurs at the end of the gulp task you can avoid this problem by uncommenting
```
//    gulp {
//        bufferOutput = true
//    }
```
in `build.gradle`.
But drawback is: not getting immediate feedback from gulp tasks any more. In prod profile it's uncomfortable.

If you want to use gulp `bufferOutput = true` only in dev profile then put following code into `profile_dev.gradle`:
```
if (OperatingSystem.current().isWindows()) {
    gulp {
        bufferOutput = true
    }
}
```

Related issues and PR-s:
* https://github.com/jhipster/generator-jhipster/issues/4675
* https://github.com/jhipster/generator-jhipster/pull/4683
* https://github.com/jhipster/generator-jhipster/pull/4730

kaidohallik:
Instead of tip maybe better solution is:
* create issue with content of previous comment
* in executing `gradlew` output link to this new issue at the end of every node task in Windows

kaidohallik:
After some testing I found that after adding
```
    doLast {
        println()
    }
```
to gulp (without `bufferOutput=true`) and node tasks no more hanging in my computer.
`yarn_install`, `gulpConstantDev` and `gulpBuildWithOpts` working fine after this change.

PR #4731, opened 2016-12-21
fix the order of the generator questions

### LINKED ISSUE #4729: The order of the generator questions is not correct
* **Overview of the request**

The order of the generator questions is not correct (see image below).

jhipsterquestionsorder

* **Browsers and Operating System** 

ANY

* **Suggest a Fix** 

I have a fix, I will make a pull request

Akuka:
fix #4729 

After the change the order is correct:
after


jdubois:
I don't understand how that's supposed to work :-(
Anyway this isn't really working, the last question is now "15/14", so that's still wrong... That's why I'm closing this.

jdubois:
Sorry, now I understand what you did :-)
Hang on, I'm correcting!

PR #4735, opened 2016-12-21
[WIP] Migration of entity

VictorADS:
We finished the templates (infinite scroll, pagination, pager..) 
We fixed enum and LocalDate Fields 

There are however some little bugs on that we will fix tomorrow with @flaviencathala


deepu105:
please fix the eslint errors

deepu105:
also make sure tests are green

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-delete-dialog.component.html
```diff
@@ -1,4 +1,4 @@
-<%_ var keyPrefix = angularAppName + '.' + entityTranslationKey + '.'; _%>
+<%_ var keyPrefix = angularAppName + '.' + entityTranslationKey; _%>
```
deepu105: whay is that dot removed? i think it should be there

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.state.ts
```diff
@@ -137,6 +143,15 @@ export const <%= entityInstance %>EditState = {
         let <%= entityInstance %>Service: <%= entityClass %>Service = trans.injector().get(<%= entityClass %>Service);
         let id = trans.params()['id'];
         <%= entityInstance %>Service.find(id).subscribe(<%= entityInstance %> => {
+            <%_ if(hasDate) { _%>
+            <% for (idx in fields) { if (fields[idx].fieldType == 'LocalDate') { %>
+            <%= entityInstance %>.<%=fields[idx].fieldName%> = {
```
deepu105: can we do this in the entity.service?

generators/generator-base.js
```diff
@@ -1927,3 +1927,21 @@ Generator.prototype.httpGet = function(url, onSuccess, onFail) {
         });
     }).on('error', onFail);
 };
+
+/**
+ * Function to print a proper array with simple quoted strings
+ *
+ *  @param {array} array - the array to print
+ */
+Generator.prototype.printProperArray = function(array) {
```
deepu105: rename to `toArrayString`
also instead of looping may be we could do ```array.join(`', '`)``` and then wrap the output in `[' ']` give it a try

deepu105: so i think it will be something like ```return `['${array.join('\', \'')}']`;```

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-delete-dialog.component.html
VictorADS: I think its just that when we use it in the jhi-translate directive we already have a "." right after keyPrefix so it was causing two dots to be printed

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.state.ts
VictorADS: I can do this, but I don't know if it's a good idea ! Anybody can use the find method to get an entityInstance but it's only in the modal that we want to format the date so that it works with NgbDatePicker.

Ps : NgbDatepIcker only accepts objects with this format  { year: xxx, month: xxx, day: xxx}, and what we receive from find is a Date object (I hope I was clear enough)

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-delete-dialog.component.html
deepu105: oh ok

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.state.ts
deepu105: ok, let it be here then

deepu105: but we need to find a proper solution may be add a todo here as well

generators/entity/files.js
```diff
@@ -213,7 +213,7 @@ const angularFiles = {
             templates: [
                 {
                     file: 'entities/_entity-management.component.html', method: 'copyHtml', template: true,
-                    renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.component.html`
+                    renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityPluralFileName}.component.html`
```
deepu105: lets keep it as `entityFileName` and fix the tests so that this is compatible to angular CLI

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.state.ts
```diff
@@ -61,7 +67,7 @@ export const <%= entityInstance %>State = {
         {
             token: 'translate',
             deps: [JhiLanguageService],
-            resolveFn: (languageService) => languageService.setLocations(['<%= i18nToLoad %>'])
+            resolveFn: (languageService) => languageService.setLocations(<%- toArrayString(i18nToLoad) %>)
```
deepu105: why `<%-` instead of `<%=`?

VictorADS: @flaviencathala  is there any reason behind this ?

flaviencathala: Yes there is if you don't put the '-', the quotes in the array will be "&34" something instead of "'" (a proper quote).

deepu105: oh ya it needs to be unescaped

generators/entity/files.js
deepu105: @VictorADS @flaviencathala I'm ok with the PR so once you change the file name here you can merge this

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
```diff
@@ -36,7 +36,7 @@ <h4 class="modal-title" id="my<%= entityClass %>Label" jhi-translate="<%= keyPre
     <div class="form-group"<% if ((fieldType == 'byte[]' || fieldType === 'ByteBuffer') && fieldTypeBlobContent != 'text') { %> ngf-drop ngf-change="set<%=fieldNameCapitalized %>($file, <%= entityInstance %>)"<% if (fieldTypeBlobContent == 'image') { %> ngf-pattern="'image/*'"<% } %><% } %>>
     <label class="form-control-label" jhi-translate="<%= translationKey %>" for="field_<%= fieldName %>"><%= fieldNameHumanized %></label>
     <%_ if (fields[idx].fieldIsEnum) { _%>
-    <select class="form-control" name="<%= fieldName %>" [(ngModel)]="<%= entityInstance %>.<%= fieldName %>" id="field_<%= fieldName %>" #<%= fieldName %>="ngModel"<% if (fields[idx].fieldValidate == true && fields[idx].fieldValidateRules.indexOf('required') != -1) { %> required<% } %>>
```
erkieh: Was the objective of removing these variable names from the controls to access the same controls through the form object in validation messages?

Issue #4724, opened 2016-12-20
[FEATURE] Move maven dependencies and plugins in a parent pom


dacloutier:
##### **Overview of the issue**
Currently, the jhipster generator creates a maven pom.xml that defines springboot-starter-parent as the parent pom:

```
    <parent>
        <artifactId>spring-boot-starter-parent</artifactId>
        <groupId>org.springframework.boot</groupId>
        <version>1.4.2.RELEASE</version>
        <relativePath/>
    </parent>
```
Then goes to define all the properties, dependencies and plugins required for the jhipster application.

##### **Motivation for or Use Case**
If you then modify the pom.xml for your application, adding properties, dependencies or plugins, these changes are at risk of being overridden when you update the jhipster version in the future.

Also, these poms are quite repetitive between all your generated apps. This can lead to confusion and minor discrepancies are hard to detect.

Also, I tried to do this myself, and created a custom parent pom for all my projects, but then I upgraded from 3.10 to 3.12 and the upgrade process changed my pom from top to bottom.

##### **Suggest a Fix**
If the JHipster team created a public parent pom and deployed it on maven central then the pom.xml generated by the jhipster generator could be a lot simpler and easy to maintain as the app ages.

Example:

```
<?xml version="1.0" encoding="UTF-8"?>
<project ......   >
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>io.github.jhipster</groupId>
        <artifactId>jhipster-parent</artifactId>
        <version>3.12.2</version>
        <relativePath/>
    </parent>
    <groupId>app.package</groupId>
    <artifactId>app.name</artifactId>
    <version>0.0.1-SNAPOSHOT</version>
    <name>Your App</name>
    <prerequisites>
        <maven>3.0.0</maven>
    </prerequisites>

    <dependencies>
        <!-- custom dependencies based on the choices entered in the app generator -->
        <dependency>
            <groupId>com.h2database</groupId>
            <artifactId>h2</artifactId>
        </dependency>
        <dependency>
            <groupId>mysql</groupId>
            <artifactId>mysql-connector-java</artifactId>
        </dependency>
        <dependency>
            <groupId>org.ehcache</groupId>
            <artifactId>ehcache</artifactId>
        </dependency>
        <!-- jhipster-needle-maven-add-dependency -->
    </dependencies>
</project>
```

So, the yo jhipster:upgrade could just change the version of the parent pom and it's job would be done.

[...]

PR #4741, opened 2016-12-22
factorize maven version

ctamisier:
re-push, I thought you merged it so I removed it. sorry.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -55,9 +55,8 @@
         <hazelcast-hibernate.version>1.1.1</hazelcast-hibernate.version>
         <%_ } _%>
         <%_ if (databaseType === 'sql') { _%>
+        <hibernate-validator.version>5.3.3.Final</hibernate-validator.version>
         <hibernate.version>5.2.4.Final</hibernate.version>
-        <%_ } _%>
-        <%_ if (databaseType === 'sql') { _%>
         <hikaricp.version>2.4.6</hikaricp.version>
         <%_ } _%>
         <%_ if (enableSocialSignIn) { _%>
@@ -70,6 +69,9 @@
         <jcache.version>1.0.0</jcache.version>
         <%_ } _%>
         <jhipster.server.version>0.0.7</jhipster.server.version>
+        <%_ if (authenticationType === 'jwt') { _%>
+        <jjwt.version>0.7.0</jjwt.version>
+        <%_ } _%>
         <%_ if (searchEngine === 'elasticsearch') { _%>
         <jna.version>4.2.2</jna.version>
         <%_ } _%>
@@ -477,7 +479,7 @@
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
-            <version>5.3.3.Final</version>
+            <version>${hibernate-validator.version}</version>
         </dependency>
         <dependency>
             <groupId>org.liquibase</groupId>
@@ -630,7 +632,7 @@
         <dependency>
             <groupId>io.jsonwebtoken</groupId>
             <artifactId>jjwt</artifactId>
-            <version>0.7.0</version>
+            <version>${jjwt.version}</version>
         </dependency>
         <%_ } _%>
         <%_ if (authenticationType === 'uaa') { _%>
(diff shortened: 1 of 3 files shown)


generators/server/templates/_pom.xml
```diff
@@ -56,8 +56,7 @@
         <%_ } _%>
         <%_ if (databaseType === 'sql') { _%>
         <hibernate.version>5.2.4.Final</hibernate.version>
-        <%_ } _%>
-        <%_ if (databaseType === 'sql') { _%>
+        <hibernate-validator.version>5.3.3.Final</hibernate-validator.version>
```
pascalgrimaud: Must be before hibernate.version

generators/server/templates/_pom.xml
```diff
@@ -73,6 +72,9 @@
         <%_ if (searchEngine === 'elasticsearch') { _%>
         <jna.version>4.2.2</jna.version>
         <%_ } _%>
+        <%_ if (authenticationType === 'jwt') { _%>
+        <jjwt.version>0.7.0</jjwt.version>
```
pascalgrimaud: Must be before jna.version

jdubois:
And do we have the same properties for Gradle? Both builds should be similar.

I understand the use of variables, but I didn't put any there because the version number is only used once, so it doesn't help a lot.

jdubois:
Sorry for my last comment: of course this will become interesting when we do #4724

ctamisier:
I'm just used to do this to have like an overview of the versions used for the dependencies..
Ok for the order. (I made the changes)
For gradle, jjwt version is already a variable.
And for hibernate validator I realize that it is inherited.
For maven the defined version in the parent is 5.2.4.Final and we override it to 5.3.3.Final.
I guess it should be the same for gradle. do I apply the change ?

ctamisier:
Or maybe better to not override it. Were there some reasons to update the version ?

ctamisier:
ok in gradle there is hibernate_validator_version=5.3.3.Final but the variable is not used. I guess I used it and that's it ? or we keep the spring boot parent one ?

jdubois:
Have a look at the commit which made the Hibernate Validator version override - I think there is a good a reason for it, we don't override a Spring Boot-provided variable by random

ctamisier:
https://github.com/jhipster/generator-jhipster/commit/631445b3c2517369d63020b62f570a57a4b7ca13

no special indication here... I will try to find something related in the issues.

jdubois:
OK but the author is @gunnarmorling and he's the spec lead for Bean Validation, so I'm sure he knows this better than anyone else! I would keep the version number.

ctamisier:
ok for the version.
I don't if hibernate_validator_version is used when the dependency is declared without the variable: 
`compile "org.hibernate:hibernate-validator"`

I am wondering if this is required:
`compile "org.hibernate:hibernate-validator:${hibernate_validator_version}"`

I don't know gradle and neither if hibernate_validator_version is declared in a "parent" (my mind is maven-oriented..)

ctamisier:
For maven in dependency:tree we have
`[INFO] +- org.springframework.boot:spring-boot-starter-web:jar:1.4.2.RELEASE:compile`
`[INFO] |  +- org.hibernate:hibernate-validator:jar:5.3.3.Final:compile`

so we don't even need to declare:
`<dependency>`
`<groupId>org.hibernate</groupId>`
`<artifactId>hibernate-validator</artifactId>`
`<version>${hibernate-validator.version}</version>`
`</dependency>`

We can just have
`<properties>`
`...`
`<hibernate-validator.version>5.3.3.Final</hibernate-validator.version>`
`...`
`</properties>`

ctamisier:
For Gradle I tried and 
`compile "org.hibernate:hibernate-validator:${hibernate_validator_version}"` is required.

For maven I removed the dependency because already imported by spring-boot-starter-web.
I just factorized the version in the properties

gunnarmorling:
> For maven I removed the dependency because already imported by spring-boot-starter-web.

Ah, so you are overriding the property originally declared in `spring-boot-dependencies`. That should work, just let's hope the variable name doesn't get changed upstream :)

ctamisier:
hmm I see so maybe it's not really a good idea to remove it.
And this variable being overriden in maven was because it was like this in gradle (but actually not used). The need to upgrade to HV to 5.3.3.Final wasn't wanted from what I understand in this comment https://github.com/jhipster/generator-jhipster/pull/4541#issuecomment-268892810 ?
Should we just use the one from spring-boot-dependencies ?

gunnarmorling:
> The need to upgrade to HV to 5.3.3.Final wasn't wanted from what I understand in this comment 

It was wanted (sorry, had put my other comment to #4541 accidentally instead of this PR). HV 5.3 has many advantages over 5.2, hence I sent PR #4541 to pull 5.3 into JHipster apps.

ctamisier:
okay.
I revert the change about the dependency (you're right, we can't rely on the variable name in the parent and override it).

Issue #4725, opened 2016-12-20
Refactor Hibernate ID generation


jdubois:
This comes from the discussion at #4690

For our entity IDs generation, we currently use the "Auto" generator from JPA. It's standard, and then Hibernate has to use an implementation working with the current database.

This has worked well since the beginning, I coded the original implementation and I don't think we had any issue with it.

However, with JHipster 4 I'd like to make some breaking changes here.

I want to use the new `SequenceStyleGenerator` from Hibernate. In fact, we currently have a warning because with "Auto" we use the older Sequence generator (which is kind of stupid, with "Auto" it shouldn't select a deprecated generator...). This new generator is supposed to be `optimized` (I looked at the code and this also looks quite stupid, I don't see where there is an optimization here), and can automatically fall back to using a table if your database doesn't support sequences (that's for MySQL, which is used by 50% of our users - and yes that's also stupid to call "Sequence" a generator that can use a table...).

PROs:

- No more warning when Hibernate starts because we use the older Sequence
- Do not use "identity" anymore with MySQL: this costs a few more requests, but in the end it's worth it as we can now use JPA batches and table per class inheritance. So a lot more features, with a very tiny performance impact.

CONs:

- The configuration is a overly complex. But as we generate it, it shouldn't be a big issue.
- We can be backward compatible with JHipster 3, but that requires a few tweaks (basically starting the sequence with the higher number than any previously existing ID)

Thoughts, comments?

ctamisier:
If we want to be backward compatible, the already existing liquibase changeset must not be changed (liquibase checksums already stored in DB). And then only new migrations can be added altering tables and/or sequences.

What I see from this is to have many migrations even for a brand new starting project. (may confuse the new jhipster user). But still, as a jhipster 3 user, I want to upgrade to jhipster 4... (add a new option in the yeoman generator ? :( )

jdubois:
I'm not too worried for migrations: it's still using the same sequence (for databases that use sequences), and for MySQL then you need a new changeset to add the table. That doesn't look too complex for me.

gzsombor:
I'm a bit confused, for MySQL, do you suggest to not use auto-increment fields anymore? I don't understand why this could be an improvement :(

jdubois:
@gzsombor yes, because auto-increment (aka "identity") has two drawbacks: it doesn't support batching, and table per class inheritance.

gzsombor:
[...]

PR #4742, opened 2016-12-22
Refactor hibernate id generator

jdubois:


jdubois:
Fix #4725

MathieuLorber:
I have a problem on current master (a409b6), on an empty project, the app stops to launch as soon as I had a (simple) entity.
It logs "Caused by: org.h2.jdbc.JdbcSQLException: Database is already closed". Suggested solutions to the problem don't work.
When I revert this patch, it's ok again => an empty project + 1 entity launchs (I still have a ng2 state problem, I can't find the entity, but it's another problem).

pascalgrimaud:
@MathieuLorber : can you open a new issue with the same description here + full details plz? (your .yo-rc.json, your configuration, etc.)
And if possible, the exact use case (with JDL or with sub generator entity)
So it will be easier for us to try to reproduce the issue

jdubois:
Did you do a 'link' on generator-jhipster to use the master branch when generating your entities ?

MathieuLorber:
=> https://github.com/jhipster/generator-jhipster/issues/4759

@jdubois yep

PR #4750, opened 2016-12-23
Upgrade to Spring Boot 1.4.3

pascalgrimaud:
See: https://spring.io/blog/2016/12/23/spring-boot-1-4-3-available-now

To be consistent, I updated:
- the registry: https://github.com/jhipster/jhipster-registry/pull/93
- jhipster server lib: https://github.com/jhipster/jhipster/pull/3

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -37,7 +37,7 @@ postgresql_version=9.4.1212<% } %><% if (authenticationType == 'oauth2') { %>
 spring_security_oauth2_version=2.0.12.RELEASE<% } %>
 spring_security_version=4.1.3.RELEASE
 springfox_version=2.6.1
-spring_boot_version=1.4.2.RELEASE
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
@atomfrede : I didnt forget gradle! ;-)

jdubois:
On my machine it makes the app fail to load, I have a 404 error on the main page... As soon as I downgrade it works again...
I'm looking for a solution.

MathieuLorber:
@jdubois + on an empty project, the app stop to launch as soon as I had a (simple) entity. I think this is this patch (and not https://github.com/jhipster/generator-jhipster/pull/4742 - is there a way to be sure of which version(commit) of jhipster has been used ?)
It logs "Caused by: org.h2.jdbc.JdbcSQLException: Database is already closed". Suggested solutions to the problem don't work.

sylvainlecoy:
It makes my "main" profile fail as well (404 error) not only dev.

PR #4755, opened 2016-12-23
removes client framework prompt when generating UAA server

xetys:


deepu105:
Why?

xetys:
The uaa server is just a specific microservice application, which doesn't
have a frontend at all, similar to usual microservice applications.

Deepu K Sasidharan <notifications@github.com> schrieb am Fr., 23. Dez.
2016, 18:27:

deepu105:
Ahh i forgot that it doesnt have a frontend at all :) then wouldnt a better solution be to skip this question altogether when skipClient is enabled?

generators/app/prompts.js
```diff
@@ -139,7 +139,7 @@ function askForClient() {
         type: 'list',
         name: 'clientFramework',
         when: function (response) {
-            return (applicationType !== 'microservice');
+            return (applicationType !== 'microservice' && applicationType !== 'uaa');
         },
         message: function (response) {
             return getNumberedQuestion('Which *Framework* would you like to use for the client?', applicationType !== 'microservice');
```
deepu105: you need to add the condition here as well

xetys:
Yes, I think we should trigger it to skipClient, rather than to nonDRY
maintain this condition in all these places.

I will do it, as soon as I reach my note!

Deepu K Sasidharan <notifications@github.com> schrieb am Fr., 23. Dez.
2016, 18:50:

PR #4758, opened 2016-12-23
Migration of Entities + Fix for our translation service

VictorADS:
I would like to get your review on this, don't merge if it's good because I need to fix the detail page when there is a ZonedDateTime field !


generators/client/templates/angular/src/main/webapp/app/shared/language/_translate-partial-loader.provider.ts
```diff
@@ -31,8 +31,12 @@ export class TranslatePartialLoader implements TranslateLoader {
         return Observable.create(observer => {
             this.http.get(`${this.prefix}/${lang}/${part}${this.suffix}`).subscribe((res) => {
                 let responseObj = res.json();
-                Object.keys(responseObj).forEach(key => {
-                    combinedObject[key] = responseObj[key];
+                Object.keys(responseObj).forEach(key=>{
+                    if(!combinedObject[key]) {
+                        combinedObject[key] = responseObj[key];
+                    }else{
+                        $.extend(true,combinedObject[key], combinedObject[key], responseObj[key])
```
deepu105: can we try to do this without jquery? btw what is the issue?

VictorADS: The issue was when we were loading multiple part files with the same key, it would overwrite the objet and some fields where not translated (for example in entities when you chose enum fields  you had 2 différent file with the same key one for the entitity itself and one for the enum)

deepu105: oh ok I get it. I think I didn't test the scenario. OK but still i would like to do it without jquery if possible. could you try to use `Object.assign` https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

deepu105: wait a minute this should already work as we merge observers in `getPartFile` method and already every state loads 2 files. we need to debug this to why its failing

VictorADS: We merge the observers, but the translations are in combinedObject. And  when we receive the response for the http request we overwrite the combinedObject[key] field. The problem is not when there are multiple files, but when they have the same key like :
File1 contains : { "test" : { a : 1}}
File 2 contains : { "test" : { b : 2}}

We want the merged object to be like this : { "test" : { a : 1, b : 2 }} am I right ?
And what we are doing right is now is just overwriting and we only get {test : {a : 1 }} or {test : { b :2 }} depending on which file is parsed last

VictorADS: If you want to test it by yourself, create an entity with an enum field and replace the 24th line  by:
`oldObsevers = oldObsevers.merge(newObserver).do( x => console.log(x));`

deepu105: oh ok I get it

deepu105:
@VictorADS @flaviencathala you guys have marked this WIP. should I merge this or do you wanna do more stuff here?

VictorADS:
We are still working on this, i'll warn you when we are done !

flaviencathala:
@deepu105 It's ok for us, we will need to find an alternative to ng-file-upload for blob fields.

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.model.ts
```diff
@@ -25,7 +25,7 @@ for (idx in fields) {
     if (fields[idx].fieldIsEnum) {
         tsType = fieldType;
     } else if (fieldType == 'ZonedDateTime') {
-        tsType = 'Date';
+        tsType = 'any';
```
wmarques: Why ?

VictorADS: Because I need to switch between Date and Object type to show it in ngbootstrap

Issue #4302, opened 2016-10-10
Angular 2 Protractor & Karma migration


deepu105:
The protractor and Karma test cases needs to be migrated in order to be compatible with Angular2. The migration needs to be done in the client-2 subgen
Related to #3664 
@mraible hoping you would be able to help?

TODO:

- [ ] migrate karma tests for core
- [x] migrate karma tests for entity
- [x] migrate protractor tests for core
- [x] migrate protractor tests for entity

mraible:
@deepu105 Yes, I should be able to help. Probably not until later this week though.

deepu105:
Sure


Thanks & Regards,
Deepu

deepu105:
@jhipster/developers guys can anyone take up this?
also we need to decide if we need to keep the same folder structure or move the tests into the same folders as the src similar to angular style guide and angular CLI this will be important if we are gonna support CLI later

jdubois:
- If nobody takes this, I have a new trainee arriving in the middle of November. But that means it won't be finished before several weeks.
- For the folder structure, we need to keep the "Maven" folder structure or everyone will be lost: that would just be the root of the folder structure, then we should follow AngularJS style guide and Angular CLI. I remember @jmirc did a trick for that, so we could support the Angular Yeoman generator with JHipster 2, but I can't find the code anymore...

mraible:
Is it possible to disable the UI-Router state viewer? I think it might be causing issues with Protractor tests.

deepu105:
Should be possible I guess. @christopherthielen

christopherthielen:
Can you toggle debug mode for the tests?

https://github.com/jhipster/generator-jhipster/blob/angular-2/generators/client-2/templates/src/main/webapp/app/blocks/config/_router.config.ts#L11-L14

```
        if (DEBUG_INFO_ENABLED) {
            trace.enable(Category.TRANSITION);
            SystemJS.import('ui-router-visualizer').then(vis => vis.visualizer(router));
        }
```

mohamed2a:
I'll work on Protractor migration, I'll upgrade it to version 4 to make it compatible with Angular2

mraible:
I spent an hour today trying to setup Karma and Webpack to run a simple hello world test. Unfortunately, I failed. I used https://github.com/AngularClass/angular2-webpack-starter as a guide.

[...]

PR #4760, opened 2016-12-23
Setup Karma and TypeScript testing infrastructure

mraible:
Helps with https://github.com/jhipster/generator-jhipster/issues/4302.


generators/client/templates/angular/_package.json
```diff
@@ -77,12 +79,13 @@
...
     "raw-loader": "0.5.1",
     "run-sequence": "1.2.2",
+    "sourcemap-istanbul-instrumenter-loader": "0.2.0",
     "string-replace-webpack-plugin": "0.0.4",
     "style-loader": "0.13.1",
     "tslint": "4.0.2",
     "tslint-loader": "3.2.1",
     "typescript": "2.0.10",
-    "webpack": "2.1.0-beta.25",
+    "webpack": "2.2.0-rc.2",
```
deepu105: I think this will be breaking our current webpack setup. Need to check

sendilkumarn:
LGTM 👍

PR #4766, opened 2016-12-25
Sort by directive

sendilkumarn:
yet to migrate `$state.transitionTo`'s ng2 equivalent. 

@deepu105 can you look into it ?


generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -11,19 +11,21 @@ import {
...
     KeysPipe,
     MaxbytesValidator,
     MinbytesValidator,
     ShowValidationDirective,
-    JhiItemCountComponent,
+    <%=jhiPrefixCapitalized%>ItemCountComponent,
     alertServiceProvider,
-    JhiAlertComponent,
-    JhiAlertErrorComponent,
+    <%=jhiPrefixCapitalized%>AlertComponent,
```
deepu105: these as well

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -11,19 +11,21 @@ import {
     FilterPipe,
     OrderByPipe,
     <%_ if (enableTranslation) { _%>
-    JhiTranslate,
-    JhiMissingTranslationHandler,
-    JhiLanguageService,
+    <%=jhiPrefixCapitalized%>Translate,
```
sendilkumarn: Okay that makes sense. I will change them. Only Translation related right?

PR #4772, opened 2016-12-27
Add Social profile photo

ramzimaalej:
Feature:
* Shows social profile photo when it's available
* Defaults to "Account" when profile photo is not available.


generators/client/templates/angularjs/src/test/javascript/spec/app/account/settings/_settings.controller.spec.js
```diff
@@ -35,7 +35,8 @@ describe('Controller Tests', function() {
                 activated: true,
                 email: "john.doe@mail.com",
                 langKey: "<%= nativeLanguage %>",
-                login: "john"
+                login: "john",
```
deepu105: this needs to be done only for angular not for angular1 hence please remove changes in the `angularjs` folder

gzsombor:
Why only for angular 2? As looking on the current state of the angular2 ecosystem, it seems it's a bit far from usable. Otherwise, a neat idea! :+1:  You could generate gravatar image url for users, which doesn't have a social login :wink:

deepu105:
@gzsombor else we would keep porting and maintaining changes over to ng1 as well and thats a real burden. I think it would be easier for us if we restrict backports to only security fixes and any bug fixes

gzsombor:
Sure, no one wants to force you, or anyone else to work on a task, which he doesn't like :)  I just wanted to say, that if someone implements this feature for angular 1, we could merge it

deepu105:
Of course


Thanks & Regards,
Deepu

ramzimaalej:
Guys, can you help understand what's wrong with the build?

ruddell:
@ramzimaalej I don't think it relates to your changes.  I restarted the build and it should hopefully pass.

Just `app-mariadb-kafka` failed and it was due to gulp not being found.

ramzimaalej:
Thanks @ruddell !

generators/client/templates/angularjs/src/test/javascript/spec/app/account/settings/_settings.controller.spec.js
ramzimaalej: I removed it.

generators/client/templates/angular/src/main/webapp/app/shared/auth/_principal.service.ts
```diff
@@ -92,4 +92,8 @@ export class Principal {
     getAuthenticationState(): Observable<any> {
         return this.authenticationState.asObservable();
     }
+
+    getImageUrl (): String {
```
jdubois: you have an extra space before ()

jdubois:
- Did you test it with Cassandra and MongoDB ?
- This adds one more field to the User object, so there is a small performance impact
- I don't want to have a bloated user object with tons of features, which most people will end up deleting because they don't need them. This will also add issues to people who want to upgrade from JHipster 3, as they will have to do a specific changelog. I know this was an idea from @deepu105 , and now you have coded it we might as well merge it, but I just wanted to say I wasn't a huge fan of the idea in the first place.

deepu105:
May be we could add the field only when social login is used


Thanks & Regards,
Deepu

jdubois:
No let's merge this, the code is really good, and we need to have an "upgrade" changelog for Liquibase anyway.

generators/client/templates/angular/src/main/webapp/app/shared/auth/_principal.service.ts
ramzimaalej: It's been removed now.

ramzimaalej:
@jdubois to be honest, I have not tested this feature neither with Cassandra nor Mongo. 

I would have created a UserProfile class and made a 1-to-1 relationship with User, to avoid boilerplate code in User. But, then I went with the simplest solution.

deepu105:
@jdubois are you ok with this or does it need more work?

jdubois:
This needs to be tested with MongoDB and Cassandra, then yes OK

ramzimaalej:
I tested this with Mongo. It works fine! However, When I select Cassandra the generator doesn't ask me to add social login. is that normal or I should file a bug ?

jdubois:
Not sure Cassandra is supported, and that would be a very very minor option anyway (less than 0,5% of people use Cassandra).
In that case you could remove your support (it would be dead code)

ramzimaalej:
you mean just remove this snippet or remove the whole code when the selected database is Cassandra ?

'@Field("image_url")<% } %><% if (databaseType == 'cassandra') { %>'

jdubois:
Remove the part with Social login + Cassandra -> as this isn't supported, that means dead code in the end.
And anyway, very very people use it.

ramzimaalej:
It's been removed, please review and let me know if there is anything else.

deepu105:
could you please fix the errors in ng2 protractor tests

ramzimaalej:
fixed.

deepu105:
@jdubois LGTM please merge if its ok for you

jdubois:
Let's merge this, this has been waiting for too long :-)

PR #4774, opened 2016-12-27
remove unused method

ctamisier:


jdubois:
I know it's not used anymore, but when I code a project I usually use this quite often. Maybe I'm the only one?

ctamisier:
Yes, you're right it's useful when we start dealing with permissions, and when we need to do more than @Secured. let's reject this PR ?

deepu105:
Its a util and useful so lets keep it

PR #4775, opened 2016-12-27
Password Strength Bar Revamp + Karma tests

sendilkumarn:
Password strength bar rewrite.


generators/client/templates/angular/src/main/webapp/app/account/password/_password-strength-bar.component.ts
```diff
@@ -15,70 +15,70 @@ import {Component, Input, ElementRef} from '@angular/core';
...
-                } else if (s <= 40) {
-                    idx = 3;
-                } else {
-                    idx = 4;
-                }
-                return {idx: idx + 1, col: this.colors[idx]};
-            }
-        };
-    }
+    constructor(private renderer: Renderer, private elementRef: ElementRef) { }
```
deepu105: can we move the constructor up?

jdubois:
Is this for the Karma tests on Angular 2? Could you rename the ticket so we know what it's about?

Issue #3575, opened 2016-05-10
CORS microservices : Gateway - UAA - OPTIONS pre flight failure on authentication


walternate:
##### **Overview of the issue**

I know you love these CORS issues so I have decided to add one of my very own. When setting up a gateway and uaa microservice environment, OPTIONS pre-flight requests during authentication are causing CORS issues.

##### **JHipster Version(s)** 

3.2.1

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder** 

Gateway .yo-rc

{
  "generator-jhipster": {
    "jhipsterVersion": "3.2.1",
    "baseName": "gateway",
    "packageName": "com.application.server",
    "packageFolder": "com/application/server",
    "serverPort": "8080",
    "authenticationType": "uaa",
    "uaaBaseName": "uaa",
    "hibernateCache": "no",
    "clusteredHttpSession": "no",
    "websocket": "no",
    "databaseType": "mongodb",
    "devDatabaseType": "mongodb",
    "prodDatabaseType": "mongodb",
    "searchEngine": "no",
    "buildTool": "gradle",
    "jwtSecretKey": "c3089601a32274412f570ed2e177fdee1316e9b1",
    "enableTranslation": false,
    "applicationType": "gateway",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true
  }
}

UAA yo-rc

{
  "generator-jhipster": {
    "jhipsterVersion": "3.2.1",
    "baseName": "uaa",
    "packageName": "com.application.server",
    "packageFolder": "com/application/server",
    "serverPort": "9999",
    "authenticationType": "uaa",
    "hibernateCache": "no",
    "databaseType": "mongodb",
    "devDatabaseType": "mongodb",
    "prodDatabaseType": "mongodb",
    "searchEngine": "no",
    "buildTool": "gradle",
    "jwtSecretKey": "fbf2b10db6eb8e00737ee95a80175562c8523b07",
    "enableTranslation": false,
    "applicationType": "uaa",
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "skipClient": true
  }
}

##### **Browsers and Operating System** 

Mac OSX El Capitan, Chrome 50.0.2661.94

##### **Reference** 

#3318, #3237

##### **Reproduce the error** 

Uncommenting or commenting any combination of the CORS settings in application.yml for both microservices offers no solution and I get:

[...]

PR #4778, opened 2016-12-27
configure corsFilter in the apps security config

### LINKED ISSUE #4704: CORS Headers Blocked by 401
The CorsFilter (when enabled) is added after the UsernamePasswordAuthenticationFilter, meaning any requests that result in 401 do not have the expected CORS headers.  The CorsFilter bean is setup in WebConfigurer.java

Anything that uses CORS, such as other frontends/apps using a JHipster backend.

##### **Reproduce the error**
Generate any app, enable CORS in application.yml, start it, and run the following CURL (origin/referrer headers to cause Access Control Headers to appear).
```
curl http://localhost:8080/api/account -H 'Origin: http://192.168.99.100:8100' -H 'Referrer: http://192.168.99.100:8100/' -v
curl http://localhost:8080/api/profile-info -H 'Origin: http://192.168.99.100:8100' -H 'Referrer: http://192.168.99.100:8100/' -v
```
The first request won't have the CORS headers due to the 401 error, the second request will have the CORS headers.

##### **Suggest a Fix**
Similar to the code in JWTConfigurer.java, the below code fixes the headers.
`http.addFilterBefore(corsFilter, UsernamePasswordAuthenticationFilter.class);`

I'm not sure where this should go (it works as a quick fix in the last line of `SecurityConfiguration.configure()`), and it is dependent on CORS being enabled.

##### **Related Issues**
#4035 (this issue is actually mentioned at the end)

##### **JHipster Version(s)**
v3 and master

Any

ruddell:
Fix #4704

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
@@ -13,7 +13,6 @@ import io.github.jhipster.security.Http401UnauthorizedEntryPoint;
 import io.github.jhipster.security.AjaxLogoutSuccessHandler;
 
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.HttpMethod;
@@ -68,15 +67,15 @@ public class OAuth2ServerConfiguration {<% if (databaseType == 'sql') { %>
 
         private final AjaxLogoutSuccessHandler ajaxLogoutSuccessHandler;
 
-        @Autowired(required = false)
-        private CorsFilter corsFilter;
+        private final CorsFilter corsFilter;
 
         public ResourceServerConfiguration(TokenStore tokenStore, Http401UnauthorizedEntryPoint http401UnauthorizedEntryPoint,
-            AjaxLogoutSuccessHandler ajaxLogoutSuccessHandler) {
+            AjaxLogoutSuccessHandler ajaxLogoutSuccessHandler, CorsFilter corsFilter) {
 
             this.tokenStore = tokenStore;
             this.http401UnauthorizedEntryPoint = http401UnauthorizedEntryPoint;
             this.ajaxLogoutSuccessHandler = ajaxLogoutSuccessHandler;
+            this.corsFilter = corsFilter;
         }
 
         @Override
@@ -91,8 +90,7 @@ public class OAuth2ServerConfiguration {<% if (databaseType == 'sql') { %>
             .and()
                 .csrf()
                 .disable()
-                // By default CORS is disabled. Uncomment here and in application.yml to enable.
-                // .addFilterBefore(corsFilter, UsernamePasswordAuthenticationFilter.class)
+                .addFilterBefore(corsFilter, UsernamePasswordAuthenticationFilter.class)
                 .headers()
                 .frameOptions().disable()
             .and()<% if (!websocket) { %>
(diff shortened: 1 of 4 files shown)


generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
```diff
@@ -65,6 +68,9 @@ public TokenStore tokenStore(OAuth2AccessTokenRepository oAuth2AccessTokenReposi
 
         private final AjaxLogoutSuccessHandler ajaxLogoutSuccessHandler;
 
+        @Autowired(required = false)
```
jdubois: We don't do field injection anymore -> can you do a setter instead?

generators/server/templates/src/main/java/package/config/_OAuth2ServerConfiguration.java
```diff
@@ -85,6 +91,8 @@ public void configure(HttpSecurity http) throws Exception {
             .and()
                 .csrf()
                 .disable()
+                // By default CORS is disabled. Uncomment here and in application.yml to enable.
+                // .addFilterBefore(corsFilter, UsernamePasswordAuthenticationFilter.class)
```
jdubois: So the corsFilter might not be used at all? Then I don't like to have an unused variable around

ruddell: Two options I see are:
 - Always register the filter, just don't apply CORS headers if not enabled.  It was made conditional due to a microservice issue in #3318 but now all CORS config is handled by the gateway since #3575.
 - Remove all the code except the comment and add a note for the user to add CorsFilter to the constructor (easier)

jdubois: We have very often the question to add CORS, so this should be easy to use. Isn't there a way to enable it only when it's configured? For example you test if the variable is null or not, and only add it when it's not.
I'll have a look if I have time this evening

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -123,7 +123,7 @@ jhipster:
         corePoolSize: 2
         maxPoolSize: 50
         queueCapacity: 10000
-    #cors: #By default CORS are not enabled. Uncomment to enable.
+    #cors: #By default CORS is disabled. Uncomment to enable.
```
jdubois: We could have this all uncommented, and add an "enable" property, like we have already for many properties -> this would be easier to use, and more consistent with the rest of the code. Then, for your check, you just need to look if it's enabled or not.

ruddell: This will need to be changed in the jhipster mvn library (and would need a release), I can do a PR later tonight if you prefer that way.

jdubois: Yes I think it's better, I'll have a look and do the library part if that's good for you

ruddell: Thanks!

jdubois: Argh! I didn't know we were in fact using `org.springframework.web.cors.CorsConfiguration`, so we can't modify it and add the `enable` property...

jdubois: OK, so we have a `CorsConfiguration` bean created anyway, I didn't know this -> in that case you can inject it in the constructor (it's mandatory), and only add it if it's configured? That was basically your original code, only difference is that it's created all the time

jdubois: I'm going to try to merge your PR, I need to learn better how CORS is working anyway (I'm not the author of this part of the code, and I realize I don't know it well enough)

jdubois: OK, ignore my comments, this is all good :-)

PR #4780, opened 2016-12-28
AWS Subgenerator Polish

ruddell:
- The npm package `node-uuid` was deprecated in favor of `uuid`
 - Microservices now have a local generator-jhipster so the config is very simple
 - We install the npm packages for the users, so the error messages should just say what went wrong

### DIFF
--- a/generators/aws/lib/aws.js
+++ b/generators/aws/lib/aws.js
@@ -2,7 +2,6 @@
 var S3 = require('./s3.js'),
     Rds = require('./rds.js'),
     shelljs = require('shelljs'),
-    os = require('os'),
     Eb = require('./eb.js');
 
 var Aws, generator;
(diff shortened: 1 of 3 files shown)


generators/aws/lib/aws.js
```diff
@@ -14,18 +13,7 @@ var AwsFactory = module.exports = function AwsFactory(generatorRef, cb) {
...
-        if (skipClient) {
-            var prefix = shelljs.exec('npm config get prefix', {silent:true}).trim();
-            if (os.platform() === 'win32') {
-                jhipsterPath = prefix + '/' + jhipsterPath;
-            } else {
-                jhipsterPath = prefix + '/lib/' + jhipsterPath;
-            }
-        }
-        shelljs.exec('npm install aws-sdk progress node-uuid --prefix ' + jhipsterPath, {silent:true}, function (code, msg, err) {
+        shelljs.exec('npm install aws-sdk progress uuid --prefix node_modules/generator-jhipster', {silent: true}, function (code, msg, err) {
```
cbornet: I think this does not play well with yarn, doesn't it ?

jdubois: Indeed, there should be a check for this. And as Yarn is now our default, we should use Yarn by default, and switch to NPM if it is configured in the project

ruddell: Good catch, didn't notice because I have both npm and yarn.  Fixed it in a new PR

PR #4781, opened 2016-12-28
Migration to BS4 + Translation issues

VictorADS:
Hello, fixing issues in #4407

- bootstrap show/hide elements based on screensize doesnt work might be a class rename issue

- missing translations in health screen

Some explanations for bootstrap :
Right now we are hidding elements only in a specific media size but in bootstrap4 we cannot hide in specific media size anymore so it had to go. I didn't replace it because we are now toggling our navbar when we  reach md size (`<div class="collapse navbar-toggleable-md` in navbar.component.html) while with in ng1 we were toggling the navbar at xs size so theres no need to hide anything.

Correct me if I'm wrong !
@deepu105  Btw you can also update 

-  User Management pagination doesnt work properly

- metrics page needs update once #4454 is merged to master

They are done.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.html
@@ -19,7 +19,7 @@
                     <!-- TODO: Add translation for service name -->
                     <td><% if (enableTranslation) { %>{{'health.indicator.' + baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ baseName(health.name) }}</span><% } %> {{subSystemName(health.name)}}</td>
                     <td class="text-xs-center">
-                        <span class="tag" [ngClass]="getTagClass(health.status)" jhi-translate="'health.status.' + health.status">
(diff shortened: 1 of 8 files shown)


FireWolf2007:
commit 5d4b19424d42b1ae0a90eb1375f586f7a93b8fd4
"hidden-sm-down" from Bootstrap 4 replaces "hidden-xs hidden-sm" from Bootstrap 3, but angularjs uses Bootstrap 3 and this prevent minimize view/eidt/delete buttons .

deepu105:
@FireWolf2007 yes indeed. Would you like to PR a fix

FireWolf2007:
@deepu105 May be, but I'm not expert in Bootstrap 3/4 yet :-)
I'm reading https://getbootstrap.com/docs/4.0/migration/#responsive-utilities
```hidden-sm-down``` will be removed too (or already removed in alphas).
Some more information:
https://stackoverflow.com/questions/45666656/bootstrap-4-responsive-utilities-visible-hidden-xs-sm-lg-not-working

Is anyone an expert in Bootstrap 4?

PR #4779, opened 2016-12-27
Ng Jhipster refactor

deepu105:
Refactor angular2 code to external lib

PR #4784, opened 2016-12-28
Add PipeTest

g-boy05:
the unit test for pipe folder

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -397,6 +397,8 @@ const files = {
                 '_karma.conf.js',
                 'spec/_entry.ts',
                 'spec/app/admin/health/_health.component.spec.ts',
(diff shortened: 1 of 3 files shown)


jdubois:
Congrats on your first PR @g-boy05 !
And this isn't any easy one :-)

A couple of comments:

- Have a look at the lint output: you have some missing whitespaces, I think you will get warnings
- For your files to be processed, you need to add them to `generators/client/files-angular.js`

g-boy05:
I already added my files on my local version, I will do the pull request now.

deepu105:
Since these files are being moved to the new ng-jhipster library it would be nice if you can do this PR there instead of here

deepu105:
see this https://github.com/jhipster/generator-jhipster/pull/4779

g-boy05:
ok deepu
I will do it now

deepu105:
ok then lest close this

PR #4786, opened 2016-12-29
Use baseUrl in tsconfig.json for shorter imports

gmarziou:
~~~ts
import {JhiHealthCheckComponent} from '../../../../../../main/webapp/app/admin/health/health.component'
import {JhiHealthService} from '../../../../../../main/webapp/app/admin/health/health.service';
~~~

can be replaced by

~~~ts
import {JhiHealthCheckComponent} from 'admin/health/health.component'
import {JhiHealthService} from 'admin/health/health.service';
~~~


deepu105:
assuming you opened 2 PR by mistake

PR #4787, opened 2016-12-29
Use baseUrl in tsconfig.json for shorter imports

gmarziou:
Same as #4786 which was for entity-ng2 branch.

~~~ts
import {JhiHealthCheckComponent} from '../../../../../../main/webapp/app/admin/health/health.component'
import {JhiHealthService} from '../../../../../../main/webapp/app/admin/health/health.service';
~~~

can be replaced by

~~~ts
import {JhiHealthCheckComponent} from 'admin/health/health.component'
import {JhiHealthService} from 'admin/health/health.service';
~~~

### DIFF
--- a/generators/client/templates/angular/_tsconfig.json
+++ b/generators/client/templates/angular/_tsconfig.json
@@ -9,7 +9,7 @@
         "removeComments": false,
         "noImplicitAny": false,
         "suppressImplicitAnyIndexErrors": true,
-        "baseUrl": "<%= CLIENT_MAIN_SRC_DIR %>",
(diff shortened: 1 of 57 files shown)


gmarziou:
Damn failing on a constant.

deepu105:
May be we can do similar import in app source as well instaed of doing relative import. what do you think?

gmarziou:
I'll try and let you know

gmarziou:
Yes it works but it raises some questions too which in fact are very comparable to java packaging.
I'll put more comments on dev ML.

gmarziou:
I'll submit a new PR using 'app' as import prefix.

PR #4789, opened 2016-12-29
Add Pipe Test

g-boy05:
Move PR from master to ng-jhipter-refactor


deepu105:
@g-boy05 I think you got me wrong. You need to move this to the https://github.com/jhipster/ng-jhipster library

PR #4790, opened 2016-12-29
Travis: activate protractor for ng2 projects

pascalgrimaud:


deepu105:
I think no i18n builds are failing. to be frank I didnt test those options

deepu105:
this should work once #4779 is merged

PR #4791, opened 2016-12-30
Fix uiSrefActive issue

VictorADS:
Menu and sub menu are now correctly highlighted 

Edit : Sometimes it throws ObjectUnsubscribedError when trying to go to an another state. It seems to be related with this https://github.com/angular-ui/ui-router/issues/3046 
@christopherthielen  Can you help ? The problem only happens at startup and once you refresh the page it dissapear


deepu105:
@VictorADS so can we merge this or do you still need help?

VictorADS:
The highlightings works but I think it's ui-router that doesn't work. I'm trying to see what causes the error but with the current master branch I get some error so I'm taking a bit more time. Sorry.

christopherthielen:
I'm releasing angular-ui-router (for ng1) rc.1 as soon as possible.  Then I'll work on releasing a new ui-router-ng2 with the fix included.

PR #4792, opened 2016-12-30
Fix critical issue with no i18n

VictorADS:


generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.html
```diff
@@ -19,8 +19,8 @@
                     <!-- TODO: Add translation for service name -->
                     <td><% if (enableTranslation) { %>{{'health.indicator.' + baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ baseName(health.name) }}</span><% } %> {{subSystemName(health.name)}}</td>
                     <td class="text-xs-center">
-                        <span class="tag" [ngClass]="getTagClass(health.status)" jhi-translate="{{'health.status.' + health.status}}">
-                            {{health.status}}
+                        <span class="tag" [ngClass]="getTagClass(health.status)"<% if (enableTranslation) { %> jhi-translate="{{'health.status.' + health.status}}"<% } else { %> jhi-translate="'health.status.' + health.status"<% } %>>
```
deepu105: wait a minute if there is no translation the directive should not exist at all so the actual problem seems to be in the regex stripping the directives

deepu105:
@VictorADS I found the real issue and im gonna fix it so closing this

generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.html
VictorADS: Do you mean there is a regex deleting every occurrence of jhi-translate directives  ? Because right now in every template even without i18n the directive is present

deepu105:
yes the regex is not working


Thanks & Regards,
Deepu

PR #4795, opened 2017-01-01
fix health karma test

mohamed2a:
By applying this changes, you can run karma tests.


generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.html
```diff
@@ -19,8 +19,8 @@
                     <!-- TODO: Add translation for service name -->
                     <td><% if (enableTranslation) { %>{{'health.indicator.' + baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ baseName(health.name) }}</span><% } %> {{subSystemName(health.name)}}</td>
                     <td class="text-xs-center">
-                        <span class="tag" [ngClass]="getTagClass(health.status)" jhiTranslate="{{'health.status.' + health.status}}">
-                            {{health.status}}
+                        <span class="tag" [ngClass]="getTagClass(health.status)">
```
deepu105: I'm not ok with this fix as it will not work well when i18n is disabled. The real problem to solve is to figure out how to use the `jhiTranslate` directive properly here else we will have same issues when more tests are added and our users will face same issue when they add more tests.

mohamed2a: Ok, I'm working on how to use `jhiTranslate `directive

PR #4800, opened 2017-01-02
Fix social component with i18n

gzsombor:


deepu105:
damn I always miss the social files :+1:

PR #4597, opened 2016-12-02
Add generator-jhipster as an npm/yarn dependency if the app is a microservice

PierreBesson:
Fix #4218
Add a minimal package.json and npm/yarn install step after generation for skipClient apps. 

TODO
- [x] more testing
- [x] a flag to disable the feature
- [x] checking that it works with the additional module installation question

PR #4805, opened 2017-01-02
Microservices: fix generator-jhipster version in package json

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/pull/4597

When generating a microservice application, the package.json is

```yaml
{
  "devDependencies": {
    "generator-jhipster": ""
  }
}
```

It should be instead:

```yaml
{
  "devDependencies": {
    "generator-jhipster": "3.12.1"
  }
}
```

Another thing: https://github.com/jhipster/generator-jhipster/blob/master/generators/app/index.js#L340-L355

The goal of skip install is to skip `npm install` or `yarn install`, with the `package.json` created.

In the case of microservices, the `package.json` is not created. Is it intentional @PierreBesson ?

I should write npm test too, but I don't have time tonight. It can be done later

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -159,6 +159,9 @@ module.exports = JhipsterGenerator.extend({
             }
             this.baseName = this.config.get('baseName');
             this.jhipsterVersion = this.config.get('jhipsterVersion');
+            if (this.jhipsterVersion === undefined) {
+                this.jhipsterVersion = packagejs.version;
+            }
             this.clientFramework = this.config.get('clientFramework');
             if (!this.clientFramework) {
                 /* for backward compatibility */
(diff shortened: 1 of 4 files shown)


PierreBesson:
1/ I think this was actually working correctly, the empty jhipsterVersion happen only in development and disappear as soon as you do `yarn/npm link generator-jhipster`so it should not happen for users, but I guess it's not bad to double check and set it to the current versionlike you do.

2/ You are right you should still generate the package.json, can move the this.template line before the skipInstall check in this PR.

pascalgrimaud:
1/ ok, let's keep like in this PR, so it's the same behavior than monolith

2/ ok, I'm doing it in this PR

PR #4809, opened 2017-01-03
fix gradle error/hanging in windows

kaidohallik:
reverts #4683 and replaces with better solution for gulp
resolves error/hanging on executing npmInstall and yarn_install
my comment in open gradle issue: https://github.com/gradle/gradle/issues/882#issuecomment-269996978

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -86,11 +86,23 @@ if (OperatingSystem.current().isWindows()) {
             classpath = files("$buildDir/classes/main", "$buildDir/resources/main", pathingJar.archivePath)
         }
     }
-    <%_ if (!skipClient && clientFramework === 'angular1') { _%>
+    <%_ if (!skipClient) { _%>
 
-    // avoid random error and hanging in windows, see https://github.com/jhipster/generator-jhipster/pull/4683
-    gulp {
-        bufferOutput = true
+    // avoid random error and hanging in windows, see https://github.com/jhipster/generator-jhipster/pull/4809
+    tasks.withType(NodeTask) {
+        doLast {
+            println()
+        }
+    }
+    tasks.withType(com.moowork.gradle.node.npm.NpmInstallTask) {
+        doLast {
+            println()
+        }
+    }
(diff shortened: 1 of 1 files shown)


kaidohallik:
I updated PR, now avoiding webpack error/hanging in Windows: changed `GulpTask` to `NodeTask` and use this in both cases: angular1 and angular2.

If gradle issue will be resolved then can remove this part from `build.gradle`

deepu105:
this is really hacky and I feel we shouldnt be doing this as its not an issue with what we do or our code and to me its not a serious issue affecting everyone so I would rather have it as a tip for windows users and keep our builds clean from such stuff @jhipster/developers what do you think?

pascalgrimaud:
I agree with you @deepu105
I didn't follow closely all these tickets, but it seems like a problem from gradle.

@kaidohallik : is there a problem with an older version of gradle ? If no, I think we should use an older version and wait for https://github.com/gradle/gradle/issues/882

mehmet-aslan:
I normally work with larger screen buffer height (5000 - 9000) at Windows command prompt and clear output periodically. I didn't encounter this issue at this workflow but I could reproduce it with screen buffer height 300. `gradlew build` command is either gives error or just hangs at `npmInstall/yarn_install` and `webpack` tasks even if I clear output before.

This hack solves the issue but I prefer a tip about using larger screen buffer height and clearing unnecessary output before buffer exceeds.

kaidohallik:
@pascalgrimaud There are inputs from gradle 2.x version about that problem and current version 3.2.1 is also affected.

It maybe can be fixed in gradle side but maybe this is some strange Windows problem what cannot be fixed in gradle.

Tip is ok for me. But gulp older fix is already taken into source. So I will make revert PR for #4730

kaidohallik:
I made PR-s for tip and for revert gulp hack.
If tip will be accepted, will you @mehmet-aslan update tip with your suggestions to avoid error/hanging?

mehmet-aslan:
@kaidohallik I checked your tip and it seems enough for me. You already wrote about increasing buffer height so I could only add clearing output when I have time.

kaidohallik:
Tip is merged and old gulp hack is reverted so I'm closing this.

PR #4817, opened 2017-01-03
reorganized styles, what allows styleUrls and fixes SASS option

### LINKED ISSUE #4814: incomplete/failing styles setup for CSS and SASS
I was playing around the last days with the styles, as I need to kick off a JHipster 4 project on sass, and it was not really successfull, until I spent some hours with @arelav (my frontend guru from Viber :grinning: ) and finally made it working...

## for CSS and SCSS

as of many guides for ng2 and demo code + angular cli code, there is a practice of including component specific stylesheets (scss or css), which is bound to the component. 
This is not working with current JHipster state, as there is no to-string-loader, which is needed for both, css and sass. 

I am definitely convinced that this is a practice we must offer to our users, as providing so much component style support, but still messing up the styles in a separate folder is not consistent to our codebase, as well as to other guides. Furthermore, the view encapsulation allows the dev not to think about ambiguous css selectors, as these files are only working for their components, and not for the whole project.

Next point, both main and vendor style are collected inside vendor.ts without any distinction in a way, what is meant to be used for vendor styles. It is rendered to disk, rather then to memory in dev mode, so it is a bit less peformant.

## SASS

Due to a misconfiguration, there is a main.css generated by JHipster, which is collected by webpack, but also put on gitignore by our generator. Additionally, there is no configuration of compiling main.scss to main.css, so the whole generated project will fail on other systems (main.css will be mssing as of gitignore, sass not rendering a new one)! Obviously there is no real sass support at all.

## my proposed solution

as in my current [demo project](https://github.com/xetys/jhipster-4-sass-material) I am doing the following things:

- enabling to-string-loader, so components can load their own styles
- moving main.(s)css from content/css | scss into app/layouts/main, so it is the main specific style, and is now loaded to memory
- disabling ViewEncapsulation on main component, so the main.css is affecting the whole project (and works like it did before)
- using 2 css testers with style-loader for vendor.scss, and to-string-loader for the rest

This steps were needed to be done manually, in order to make the brand new angular-material work with our angular2 app, enabling custom themes via sass.

My current reference implementation can be viewed on the link above, and @arelav is admitting, this way of doing this is more clean, than we do. It also shows, how angular material can be set up to use own theming, and demonstrates the combined power of bootstrap and material!

I will start a PR on this. Maybe even this evening.

xetys:
thanks to @arelav for supporting...

close #4814


generators/client/templates/angular/_package.json
```diff
@@ -84,6 +84,7 @@
     "sourcemap-istanbul-instrumenter-loader": "0.2.0",
     "string-replace-webpack-plugin": "0.0.4",
     "style-loader": "0.13.1",
+    "to-string-loader": "^1.1.5",
```
gmarziou: Should be exact version

pascalgrimaud:
@xetys : I put to **on-hold** as asked by @deepu105 in the ticket https://github.com/jhipster/generator-jhipster/issues/4814

deepu105:
I'm gonna add my comments on the ticket

xetys:
@deepu105 would you please check if this is ok now?

generators/client/templates/angular/src/main/webapp/app/layouts/main/_main.component.ts
```diff
@@ -2,6 +2,6 @@ import { Component } from '@angular/core';
 
 @Component({
     selector: '<%=jhiPrefix%>-main',
-    templateUrl: './main.component.html'
+    templateUrl: './main.component.html',
```
deepu105: Unwanted comma

generators/client/templates/angular/src/main/webapp/scss/_global.scss
```diff
@@ -105,37 +29,12 @@ Logo styles
             width: 70px;
             display: inline-block;
             vertical-align: middle;
-            background: url("../images/logo-jhipster.png") no-repeat center center;
+            background: url("../content/images/logo-jhipster.png") no-repeat center center;
```
deepu105: Why path change here?  Shouldn't it be same as the css as this is output in same directory?

xetys: no it's generated inside scss folder as well for me. So I had to fix this path to point to the images properly.

deepu105: can you see if its possible to malke it generate inside the content folder like other css?

xetys: Are there any concerns about moving the file to content/scss ? This would obviously solve this

deepu105:
I dont have any concern as long as scss source files doesn't get moved to
the target/www folder

Thanks & regards,
Deepu

xetys:
all change request should be resolved now

Issue #4810, opened 2017-01-03
@AuthorizedFeignClient invoke other services throw timeout


anjia0532:
1.  jhipster-registry v2.5.7
1. uaa v3.12.2
1. microservices1(e.g. named app1) v3.12.2
1. microservices2(e.g. named app2) v3.12.2

when the first time app1 invoked the app2 thow time-out exception,but since the second is successful.

```java
Exception in com.shunneng.app1.web.rest.App1Resource.getApp1s() with cause = 'java.util.concurrent.TimeoutException' and exception = 'invokeWebService timed-out and no fallback available.'

com.netflix.hystrix.exception.HystrixRuntimeException: invokeWebService timed-out and no fallback available.
	at com.netflix.hystrix.AbstractCommand$21.call(AbstractCommand.java:783)
	at com.netflix.hystrix.AbstractCommand$21.call(AbstractCommand.java:768)
	at rx.internal.operators.OperatorOnErrorResumeNextViaFunction$4.onError(OperatorOnErrorResumeNextViaFunction.java:140)
	at rx.internal.operators.OnSubscribeDoOnEach$DoOnEachSubscriber.onError(OnSubscribeDoOnEach.java:87)
	at rx.internal.operators.OnSubscribeDoOnEach$DoOnEachSubscriber.onError(OnSubscribeDoOnEach.java:87)
	at com.netflix.hystrix.AbstractCommand$DeprecatedOnFallbackHookApplication$1.onError(AbstractCommand.java:1448)
	at com.netflix.hystrix.AbstractCommand$FallbackHookApplication$1.onError(AbstractCommand.java:1373)
	at rx.internal.operators.OnSubscribeDoOnEach$DoOnEachSubscriber.onError(OnSubscribeDoOnEach.java:87)
	at rx.observers.Subscribers$5.onError(Subscribers.java:230)
	at rx.internal.operators.OnSubscribeThrow.call(OnSubscribeThrow.java:44)
	at rx.internal.operators.OnSubscribeThrow.call(OnSubscribeThrow.java:28)
	at rx.Observable.unsafeSubscribe(Observable.java:10211)
	at rx.internal.operators.OnSubscribeDefer.call(OnSubscribeDefer.java:51)
	at rx.internal.operators.OnSubscribeDefer.call(OnSubscribeDefer.java:35)
	at rx.Observable.unsafeSubscribe(Observable.java:10211)
[...]

PR #4818, opened 2017-01-03
Add hystrix configuration commented for microservices

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/4810

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application.yml
+++ b/generators/server/templates/src/main/resources/config/_application.yml
@@ -25,7 +25,7 @@ ribbon:
     eureka:
         enabled: true
 <%_ } _%>
-<%_ if (applicationType == 'gateway') { _%>
+<%_ if (applicationType === 'gateway' || applicationType === 'microservice') { _%>
 zuul:
     host:
         connect-timeout-millis: 5000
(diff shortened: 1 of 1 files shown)


PierreBesson:
I don't understand how the zuul properties would be useful for microservices as I thought zuul was not embedded in microservices, only the gateway.

pascalgrimaud:
Oh you're right @PierreBesson!
Only the part on Hystrix should be enough so

gmarziou:
isn't ribbon being used also on a microservice calling another one?

pascalgrimaud:
You make be doubt!
I need to check the config we use on my project :)

pascalgrimaud:
The config with hystrix should be enough.

But remember the Policy 1: technologies used by JHipster have their default configuration used as much as possible, I should not put this additional config for microservices as it's very specific -> only when there are communication between microservices. There are already some tickets about that, so the information can easily be found.

Let's close this!

PierreBesson:
@pascalgrimmaud, but maybe we should still leave a commented out hystrix property with a link to the docs for discovery purposes...

xetys:
+1 to providing more docs on this

PR #4819, opened 2017-01-03
WebSocket/stomp fixes

gzsombor:
Fix the tracker using webstomp-client(which is a more recent) and sockjs-client, and fix the typescript warnings

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
@@ -1,7 +1,6 @@
-declare var SockJS;
-declare var Stomp;
 import { Injectable, Inject } from '@angular/core';
-import { Observable, Observer } from 'rxjs/Rx';
+import { Observable, Observer, Subscription } from 'rxjs/Rx';
+import { UIRouter } from 'ui-router-ng2';
 <%_ if (authenticationType === 'oauth2') { _%>
 import { LocalStorageService } from 'ng2-webstorage';
 <%_ } _%>
@@ -10,6 +9,9 @@ import { CSRFService } from '../auth/csrf.service';
 <%_ if (authenticationType === 'jwt' || authenticationType === 'uaa') { _%>
 import { AuthServerProvider } from '../auth/auth-jwt.service';
 <%_ } _%>
+
+import SockJS = require('sockjs-client');
+import Stomp = require('webstomp-client');
 
 @Injectable()
 export class <%=jhiPrefixCapitalized%>TrackerService {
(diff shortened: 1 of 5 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
```diff
@@ -11,6 +10,9 @@ import { CSRFService } from '../auth/csrf.service';
 import { AuthServerProvider } from '../auth/auth-jwt.service';
 <%_ } _%>
 
+import SockJS = require('sockjs-client');
```
deepu105: why not `import { SockJs } from 'sockjs-client'` or `import * as SockJs from 'sockjs-client'` or something similar?

gzsombor: I've tried that ones, but it didn't worked. Unfortunately sockjs-client exports only a constructor function, not an object with various functions on it. 
In plain javascript:
var sockjs = required("sockjs-client");
var s = new sockjs();
Instead of 
var s = new sockjs.SockJS();

deepu105: thats strange as per the examples on this page https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import it should be possible to import functions as `import { SockJs } from 'sockjs-client'` or `import SockJs  from 'sockjs-client'`

gzsombor: In theory, the theory and the practice is the same, but in practice ... :wink: 
If you write:

```
import SockJS from 'sockjs-client';
const socket = new SockJS(url);
```
It will translated to
```
var sockjs_client_1 = __webpack_require__(1056);
var socket = new sockjs_client_1.default(url);
```
If you write:
```
import { SockJS } from 'sockjs-client';
const socket = new SockJS(url);
```
it is translated to :
```
var sockjs_client_1 = __webpack_require__(1056);
var socket = new sockjs_client_1.SockJS(url);
```

and of course, neither is ok. So if you know a better option, I'm open to it, but unfortunately I haven't found one.

gzsombor:
I'm going to merge it, without this, when websocket is enabled, it is impossible to use the application. If later we figure out a better way to import sockjs we can fix it.

generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
dancancro: Hi @gzsombor I was wondering why sockjs and stomp were chosen for this rather than [Observable.websocket](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-webSocket). Thanks.

gzsombor: It worked fine for Angular 1.x, so we stayed with it. What are the advantages of Observable.websocket ?

PR #4821, opened 2017-01-05
Added curley braces for all the single statement if block; IMHO it is better being less error prone.

lrkwz:
(I hope my branch is cleaned up again now).


delverdev:
I think you should put one space after `if(...)`.

PR #4822, opened 2017-01-06
Use pureFilter instead of filter pipe

VictorADS:


deepu105:
you also need to update to the latest ng-jhipster lib

deepu105:
ok i have updated the lib just rebase to latest master

PR #4829, opened 2017-01-08
Update webpack to 2.2.0-rc.3

gmarziou:


deepu105:
@gmarziou could you also update this for ng-jhipster lib if you have time

PR #4831, opened 2017-01-09
Remove const restriction for "url" variable for tracker service.

romaindequidt:
Stateless authentication has to add extra parameters to the "url" variable of the tracker service.
You can reprodude this issue by choosing stateless authentication (JWT, OAuth2...) and spring websocker when generating project.


gzsombor:
Make sense, I haven't tested it originally with stateless authentication.

romaindequidt:
@deepu105 did you re-launch Travis build job #6834.3. It failed the first time and now it's Ok.
Is that caused by Travis CI instability?

deepu105:
yes I suppose


Thanks & Regards,
Deepu

pascalgrimaud:
Travis failed randomly quite often since the migration from firefox to chrome. I need to investigate it...

PR #4834, opened 2017-01-09
Git ignore for Java annotation processor in Eclipse

nkolosnjaji:
`.factorypath` also needs to be excluded from git for Eclipse users.

More @ https://www.gitignore.io/api/eclipse

### DIFF
--- a/generators/server/templates/gitignore
+++ b/generators/server/templates/gitignore
@@ -32,6 +32,7 @@ local.properties
 .classpath
 .settings/
 .loadpath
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Little tip @nkolosnjaji : you can add [ci skip] to your commit message, so Travis wont trigger a build (as there is no need here to test the generator) :)

nkolosnjaji:
@pascalgrimaud thanks, I will keep that in mind for next time

PR #4844, opened 2017-01-10
Allow setting server port for monolith apps

gmarziou:
On my PC, I have a service running at port 8080, so each time I generate a monolith, I have to manually change the port to a different value in several files.

This allows me to change port in `.yo-rc.json` and re-generate.

I only changed the templated files, there are probably other files to modify too but I'd like to get feedback before going further.


jdubois:
+1 of course.
BTW, we had this joke on Twitter on using port 1337 by default, what do you think of this?

PR #4845, opened 2017-01-11
Correct jhiHasAuthority Directive and make use of it

pvlastaridis:


jdubois:
Really happy to see such contribution on our unreleased Angular 2 code!

pvlastaridis:
I am not sure the cousin of this directive jhiHasAnyAuthority and if it is of any of use.

deepu105:
@pvlastaridis could you do the same for `jhiHasAnyAuthority`?

pvlastaridis:
@deepu105 `jhiHasAnyAuthority` is working fine but it's just not being used anywhere.

I believe we should leave it as it is, since it's a useful directive.

PR #4846, opened 2017-01-11
Adds Ng2-Router ( angular )

sendilkumarn:
This brings angular router to our application, the basic app works fine with these changes.

Two major changes,
I have moved the translate resolvers into the component itself (rather than placing them in state), this actually speeds up the application a little bit. Added to that I think angular team also suggests, we should have minimal of items in resolve.

Have a look at create / edit / delete options of user management, they are event driven and rather than via routes.

TODOs
- [x] Entity Migration
- [x] Way to handle setting browser titles
- [x] ~~Remove ui-router completely~~ ( In a diff PR )
- [x] Normalize user management popup route to have same naming strategy as that of entity routes
- [x] ~~rename from state to route once all are done.~~ ( In a diff PR )


jdubois:
I didn't know you were working on this... @VictorADS I know you're doing the same thing, can you have a look?

@deepu105 are you OK to migrate to NG Router? I know it's not cool for UI Router, but it's not really our fault if the Angular 2 team copied their work, and is pushing to use their own solution: it's clear that our users mostly want NG Router, as it's "standard", and that's also what I see at all my client's projects.

So +1 for me, but I think the Angular 2 team isn't playing nicely with others, and that's not a very good spirit... Then again, this isn't our fault.

wmarques:
I agree with what @jdubois said and moreover, we would be more compatible with ng cli which would be great and asked by our users (hi @agoncal :smile: ).
Moving to NG Router is more accurate about our philosophy to follow Angular standards.

VictorADS:
LGTM. I have some remarks though :

- I think you can get rid of the useHash property because we planned to use html5 mode as url strategy

- I can' see where you are redirecting the user when he enters a wrong URL ? It should be something like `  { path: '**', component: HomeComponent } `

- For the user-detail component you are using queryParams to load the correct user. Can't the login be part of the url itself ?

sendilkumarn:
> useHash property 
Without hash, the reload is not working properly. That is the main reason, probably we can change after we finish the complete migration to ng-router.

> { path: '**', component: HomeComponent }

Yeah. I will add this. 


> For the user-detail component you are using queryParams to load the correct user. Can't the login be part of the url itself ?

can't get this.

VictorADS:
In user-management.component.html you redirect to the detail page using this :
`<td><a routerLink="/user" [queryParams]="{ login: user.login }">{{user.id}}</a></td>`

Correct me if i'm wrong but wouldn't it redirect to something like : `localhost:8080/user?login=admin` ?
Because I think what we want is more something like `localhost:8080/user/admin`

gzsombor:
I think it's better to put the page dependencies in the resolver section, so rendering is not started until everything is loaded from the backend. Because this way, the user wont see a broken page, and a second render could be skipped (which would happen after each of the resources are loaded).
 Using events, instead of states is a step backward, it's better to have bookmarkable, and easily locatable pages - you will feel the pain, when you have to debug a page/component which you can reach only after 2 or 3 clicks, instead of a quick re-load. 
 As there was considerable work to fix all the routing problems in the recent weeks, I would oppose to start it over again, as I can't see too much advantages of ng-router over ui-router. And I bet, it's more easy to convince the ui-router team to fix something, than to convince the angular team.

sendilkumarn:
@victorADS yeah correct. I too thought it was odd.

deepu105:
For the record I'm not a fan of angular router and I fully agree with @gzsombor The only reason to migrate is to support angular CLI [which is stupid coz the angular team shouldn't force people to use their router] I really like ui router more but if everyone wants go this way I'm ok with it as I dont want to be a bottleneck (
I'm really sorry @christopherthielen hope you would understand). But as I would at least like to have same level of support we had. with states for edit/delete etc would be one of that.

I would rather do this after the release so that we don't delay it gain trying to fix issues from the new router

mraible:
Angular CLI doesn't support generating routes, so why does migrating to Angular router help with supporting Angular CLI?

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -2,21 +2,23 @@
 import { AuthInterceptor } from './blocks/interceptor/auth.interceptor';
 <%_ } %>
 import { NgModule } from '@angular/core';
+import {RouterModule} from '@angular/router';
```
deepu105: spacing

generators/client/files-angular.js
```diff
@@ -123,11 +121,13 @@ const files = {
                 { file: 'home/_home.component.html', method: 'copyHtml' },
                 // layouts
                 'layouts/_index.ts',
+                'layouts/_layout-routing.module.ts',
                 'layouts/profiles/_profile.service.ts',
                 'layouts/profiles/_profile-info.model.ts',
                 'layouts/profiles/_page-ribbon.component.ts',
                 'layouts/main/_main.component.ts',
                 'layouts/main/_main.component.html',
+                'layouts/navbar/_navbar.state.ts',
```
deepu105: why a state for this?

generators/client/templates/angular/_package.json
```diff
@@ -14,6 +14,7 @@
     "@angular/http": "2.4.1",
     "@angular/platform-browser": "2.4.1",
     "@angular/platform-browser-dynamic": "2.4.1",
+    "@angular/router": "^3.4.2",
```
deepu105: fix version

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -2,21 +2,23 @@
...
 import { BrowserModule } from '@angular/platform-browser';
-import { UIRouterModule } from 'ui-router-ng2';
 import { Ng2Webstorage } from 'ng2-webstorage';
 
+import {JhiLanguageService} from 'ng-jhipster';
```
deepu105: spacing

generators/client/templates/angular/src/main/webapp/app/account/_account.module.ts
```diff
@@ -1,5 +1,5 @@
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
-import { UIRouterModule } from 'ui-router-ng2';
+import {RouterModule} from '@angular/router';
```
deepu105: can you check the spacing everywhere please

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.state.ts
```diff
@@ -1,21 +1,9 @@
...
-import { ActivateComponent } from './activate.component';
+import {Routes} from '@angular/router';
+import {ActivateComponent} from './activate.component';
 
-export const activateState: Ng2StateDeclaration = {
-    name: 'activate',
-    parent: 'account',
-    url: '/activate?key',
-    data: {
-        authorities: [],
```
deepu105: how are we handling authorities and title?

sendilkumarn: Via canactivate

deepu105:
@mraible wait a minute what!!!!!!!! I havent tried it yet but that was the whole discussion regarding migrating to something inferior than ui-router

christopherthielen:
> hope you would understand

Of course I understand if you choose the Angular Router instead of UI-Router, but appreciate the sentiment :)

It's also worth noting that (back when the CLI was still generating routes), Zackary Chapple (one of the the Angular CLI people) said he's open to making the CLI compatible with UI-Router routes.

sendilkumarn:
Page title as I said need to find a way.

For changing to routes from events, it is quite complicated. Since we are opening in pop-ups it is not possible to load them ( as far as I know ). Probably we need to find a way. 

Activating routes based on can activate is really very good. It really simplifies things. Moving forward making roles dynamic, we can use this.

sendilkumarn:
They have disabled it for time being.

VictorADS:
@sendilkumarn  For the PageTitle can  you try [this](http://stackoverflow.com/questions/33520043/how-to-detect-route-change-in-angular-2) ?

For the popup I think its possible if we add router-outlet with a name in the template of user-mgmt and in the state definition we add a children property  which would activate the router-outlet
Something like :
```{
    path: 'user-management',
     component: UserMgmtComponent,
     resolve: {
       'pagingParams': UserResolvePagingParams
      }
     children: [
       {path: 'new', component: UserMgmtDialogComponent, outlet: 'popup'},
       {path: 'edit/:login', component: UserMgmtDialogComponent, outlet: 'popup},
,`

**Comment by jdubois** [comment]
Yes I'm really sorry @christopherthielen about this... I really don't like the way the Angular 2 team is handling this. I have everybody asking me to follow their lead, without any experience or knowledge on what is provided, just because it's "the standard way".
And concerning Angular CLI: at the moment the router is disabled, but they should add NG Router at some point, and I haven't seen any plan to add UI Router as an option.

**Comment by gzsombor** [comment]
So ng-router is not able to load the "popups" ? The deficiencies of this product is quickly growing ...

**Comment by wmarques** [comment]
I didn't use ng router in a real project so I might be wrong, I'll try to check deeper about what we could do with NG Router. 
Of course it we can't manage authorities and modal states as easily as UI Router does, it's gonna be a problem.

**Comment by gmarziou** [comment]
@sendilkumarn why not using route's data property to define authorities and have only one generic CanActivate rather these many similar implementations ? It would have the benefit to use same design pattern as what we have today with ui-router.

**Comment by gmarziou** [comment]
Here is how we could write this generic CanActivate

~~~java
@Injectable()
export class UserResolve implements CanActivate {

    constructor(private principal: Principal) {
    }

    canActivate(route: ActivatedRouteSnapshot) {
        let authorities = route.data['authorities'];
        return this.principal.identity().then(account => this.principal.hasAnyAuthority(authorities));
    }
}
~~~

and the route definition:

~~~java
    {
        path: 'user-management',
        component: UserMgmtComponent,
        data: {
            authorities: ['ROLE_ADMIN']
        },
        resolve: {
            'pagingParams': UserResolvePagingParams
        },
        canActivate: [UserResolve]
    }
~~~

**Comment by gmarziou** [comment]
pageTitle could probably be implemented using same definition as with ui-router and maybe using resolve or another guard (though it could be an abuse).

**Comment by gmarziou** [comment]
Minor thing: generated code does not compile for JWT authentication, due to missing SessionResolve.

**Comment by sendilkumarn** [comment]
@VictorADS  👍 Thanks for the suggestion.  I have used something similar for page title. 

But for Popup, with separate router, we can show that in view but cannot load that as a popup. check out the [original mail-app from victor savkin](https://github.com/vsavkin/router_mailapp), the `popup:compose` link is not  a popup rather loaded in aside. :( So, definitely need a more exploration and help on this. 

@gmarziou 👍  Thanks for the suggestion. Incorporated 😄

**Comment by sendilkumarn** [comment]
@christopherthielen  sorry again, personally I like `ui-router` But as @jdubois said we have lot of people asking / suggesting to use this one.

**Comment by gmarziou** [comment]
>  rename from state to route once all are done.

We could also decide to keep state for ui-router and use route for ng router.
This way we could offer both.

File: generators/client/templates/angular/src/main/webapp/app/shared/language/_language.helper.ts [hunk 7]
```diff
@@ -32,12 +20,13 @@ export class JhiLanguageHelper {
...
      */
     updateTitle(titleKey?: string) {
-        // TODO : Proper way to handle this.
-        // if (!titleKey && this.$state.current.data && this.$state.current.data.pageTitle) {
-        //     titleKey = this.$state.current.data.pageTitle;
-        // }
-        // this.translateService.get(titleKey || 'global.title').subscribe(title => {
-        //     window.document.title = title;
-        // });
+
```
gmarziou: It's no longer needed to use `window` to set title, we can use the `TitleService`

https://angular.io/docs/ts/latest/cookbook/set-document-title.html

gmarziou:
I thought we had a progress bar or a spinner but could not find it, if we don't this is a good candidate to be implemented using events.

sendilkumarn:
was loading bar working before ?

deepu105:
we didnt migrate loading bar to ng2 as i couldnt find an equivalent plugin
for ng2


Thanks & Regards,
Deepu

gmarziou:
Right this was https://github.com/chieffancypants/angular-loading-bar which used an $http interceptor. I think we can anything not specific to angular (pure css) as we just have to use events and depend on route resolved data rather than $http.

deepu105:
if we can write it as a generic solution without having to do it in each
route/component then fine else I guess it might be overkill



Thanks & Regards,
Deepu

sendilkumarn:
we can have the generic solution with `router events`. It is much easier to write, but we can wait for any module, to check if it is coming, else will do that.

sendilkumarn:
Is it okay ? Can we start doing for entities ?

deepu105:
Lets get it right for app first

Thanks & regards,
Deepu

sendilkumarn:
For app it is working...

deepu105:
ok

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/_password-reset-finish.component.ts
```diff
@@ -18,16 +18,28 @@ export class PasswordResetFinishComponent implements OnInit {
     resetAccount: any;
     success: string;
     modalRef: NgbModalRef;
+    key: string;
 
-    constructor(private passwordResetFinish: PasswordResetFinish,
+    constructor(
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
```
deepu105: since this would be on every component we could do it the same way we did earlier for states using copyJs method and by updating the strip regex

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-modal.service.ts
```diff
@@ -0,0 +1,42 @@
+import { Injectable } from '@angular/core';
+import { NgbModal, NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
+
+import { UserMgmtDialogComponent } from './user-management-dialog.component';
+import { User } from './user.model';
+import { UserService } from './user.service';
+
+@Injectable()
+export class UserModalService {
+    private isOpen = false;
```
deepu105: this is lot of duplicated code cant we try to use  a generic service instead of one per modal?

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.state.ts
```diff
@@ -0,0 +1,32 @@
...
+//
+//   constructor(private authService: AuthService) {}
+//
+//   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+//       return this.authService.authorize();
+//   }
+//
+// }
+
+export const navbarRoute: Routes = [
```
deepu105: its still weird to have navbar as a state. may be this should be done as app.state or main-state

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-modal.service.ts
sendilkumarn: we can have it one per entity ... ( considering what we currently did for the other health / login ... )

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.state.ts
sendilkumarn: Wanted it to be separate route. We actually have two dependencies for this that is why wanted to add that as a separate one.

gmarziou:
Here is a screenshot of Augury's Router Tree of default app using ng router.

2017-01-15_00h11_19

Few things to note:
- as a chrome extension it is not intrusive
- it is purely static so it does not have same features as ui-router visualizer
- it does not show route's data, I opened an issue for this: https://github.com/rangle/augury/issues/914

Regarding our code, it shows that we don't use component-less routes (or parent states).
For me it would make sense to set required authorities on parent so for user management, we would have pseudo code):

~~~ts
const adminRoutes: Routes = [
  {
    path: 'admin',
    component: AdminComponent,
    canActivate: [AuthGuard],
    data: {
      authorities: ['ROLE_ADMIN']
    },
    canActivate: [UserRouteAccessService],
    children: [
      {
        path: 'user-management',
        component: UserMgmtComponent,
        resolve: {
          'pagingParams': UserResolvePagingParams
        },
        data: {
          pageTitle: 'userManagement.home.title'
        }
      }, {
        path: 'user/:login',
        component: UserMgmtDetailComponent,
        data: {
          pageTitle: 'userManagement.detail.title'
        }
      }
    ]
  }
];
~~~

In fact for user management states, it would be probably better to have an intermediate parent route so that user-management could be a self-contained module.

I think same pattern could be applied by entity generator and I tend to think that we should consider renaming routes to something more expressive like: `home`, `list`, `details`, etc... In fact route names could probably follow translation keys for page titles in `user-management.json`: `userManagement.home`, `userManagement.detail`, ....

sendilkumarn:
@gmarziou The problem with adding a child route for user management is that we end up in creating multiple dummy components like `admin component` ( or the parent component ).

gmarziou:
In fact I don't think so, parent should be a [component-less route](https://angular.io/docs/ts/latest/guide/router.html#!#component-less-route) and that was a mistake in my pseudo code: there shouldn't be any AdminComponent.

sendilkumarn:
@gmarziou done :)

jdubois:
@sendilkumarn for me this is the last "big" thing that is missing for JHipster 4: do you need help? When do you think this could be merged?

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -2,21 +2,23 @@
...
 <%_ } %>
 import { NgModule } from '@angular/core';
-import { BrowserModule } from '@angular/platform-browser';
-import { UIRouterModule } from 'ui-router-ng2';
+import { RouterModule } from '@angular/router';
+import { BrowserModule, Title } from '@angular/platform-browser';
 import { Ng2Webstorage } from 'ng2-webstorage';
 
-import { <%=angular2AppName%>SharedModule } from './shared';
+import { JhiLanguageService } from 'ng-jhipster';
```
deepu105: required only when i18n is used either use an ejs condition block or pass the file through copyJs method when defining in files.
Same goes for every other file importing `JhiLanguageService`.

generators/client/files-angular.js
```diff
@@ -305,6 +305,7 @@ const files = {
                 'admin/user-management/_user-management-detail.component.ts',
                 'admin/user-management/_user-management-dialog.component.ts',
                 'admin/user-management/_user-management-delete-dialog.component.ts',
+                'admin/user-management/_user-modal.service.ts',
                 'admin/user-management/_user.model.ts',
                 'admin/user-management/_user.service.ts',
                 { file: 'admin/user-management/_user-management.state.ts', method: 'copyJs' },
```
deepu105: state files no longer needs to be passed through `copyJs` as there is no i18n stuff to strip anymore so this can just be a single line file name entry. Same for every other state/route files

deepu105: the `file:.., method:..` pattern now must be applied to the component files handling the translation part

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -2,21 +2,23 @@
...
 
-import { appState } from './app.state';
-import { HomeComponent, homeState } from './home';
-import { <%=jhiPrefixCapitalized%>RouterConfig } from './blocks/config/router.config';
+import { LayoutRoutingModule } from './layouts';
+import { HomeComponent } from './home';
 import { customHttpProvider } from './blocks/interceptor/http.provider';
 import { PaginationConfig } from './blocks/config/uib-pagination.config';
 
+
```
deepu105: extra line detected

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -68,7 +58,10 @@ let routerConfig = {
         { provide: Window, useValue: window },
         { provide: Document, useValue: document },
         customHttpProvider(),
-        PaginationConfig
+        PaginationConfig,
+        JhiLanguageService,
```
deepu105: This service is already provided by the `ng-jhipster` lib so it shouldnt be provided again here. it will cause issues

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -68,7 +58,10 @@ let routerConfig = {
...
+        UserRouteAccessService,
```
deepu105: this service should be provided by shared/common module and not here

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -68,7 +58,10 @@ let routerConfig = {
...
+        Title
```
deepu105: this also should be provided by common or shared/common module

generators/client/templates/angular/src/main/webapp/app/_app.state.ts
```diff
@@ -1,24 +1,35 @@
...
+export class AuthorizeResolve implements Resolve<any> {
+
+  constructor(private authService: AuthService) {}
+
+  resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+    return this.authService.authorize();
+  }
+
+  canActivate() {
+    return true;
```
deepu105: should this be true or should this check something?

generators/client/files-angular.js
```diff
@@ -129,11 +127,13 @@ const files = {
                 { file: 'home/_home.component.html', method: 'copyHtml' },
                 // layouts
                 'layouts/_index.ts',
+                'layouts/_layout-routing.module.ts',
                 'layouts/profiles/_profile.service.ts',
                 'layouts/profiles/_profile-info.model.ts',
                 'layouts/profiles/_page-ribbon.component.ts',
                 'layouts/main/_main.component.ts',
                 'layouts/main/_main.component.html',
+                'layouts/navbar/_navbar.state.ts',
```
deepu105: I think you missed to remove this and add `app.state.ts` back

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.component.ts
```diff
@@ -16,18 +17,27 @@ export class ActivateComponent implements OnInit {
     modalRef: NgbModalRef;
 
     constructor(
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
```
deepu105: update the regex in https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base.js#L1094 to strip this line and pass components through `copyJs` method

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.component.ts
```diff
@@ -16,18 +17,27 @@ export class ActivateComponent implements OnInit {
...
+        <%_ } _%>
         private activate: Activate,
         private loginModalService: LoginModalService,
-        private trans: Transition
-    ) {}
+        private route: ActivatedRoute
+    ) {
+        <%_ if (enableTranslation) { _%>
+        this.jhiLanguageService.setLocations(['activate']);
```
deepu105: this also can be stripped via same regex. remove unwanted regex for the ui router part. let me know if you need help with that

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.state.ts
deepu105: ok

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.component.ts
deepu105: same for all components

generators/client/templates/angular/src/main/webapp/app/account/sessions/_sessions.component.ts
```diff
@@ -15,7 +16,16 @@ export class SessionsComponent implements OnInit {
     success: string;
     sessions: Session[];
 
-    constructor(private sessionsService: SessionsService, private principal: Principal) {}
+    constructor(
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
+        <%_ } _%>
+        private sessionsService: SessionsService,
+        private principal: Principal) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/account/register/_register.component.ts
```diff
@@ -28,6 +28,9 @@ export class RegisterComponent implements OnInit {
         private registerService: Register,
         private elementRef: ElementRef,
         private renderer: Renderer) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/account/social/_social-register.component.ts
```diff
@@ -10,7 +10,15 @@ export class SocialRegisterComponent implements OnInit  {
     @Input() provider: string;
     providerLabel: string;
 
-    constructor (@Inject('$stateParams') private $stateParams) {}
+    constructor (
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
+        <%_ } _%>
+        @Inject('$stateParams') private $stateParams) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/account/social/_social.state.ts
```diff
@@ -1,43 +1,29 @@
...
-    name: 'social-register',
-    parent: 'account',
-    url: '/social-register/:provider?{success:boolean}',
+export const socialRegisterRoute: Route = {
+    url: 'social-register/:provider?{success:boolean}',
+    component: SocialRegisterComponent,
     data: {
-        authorities: [],
-        pageTitle: 'social.register.title'
+      authorities: [],
```
deepu105: incorrect indentation

deepu105: on entire file

generators/client/templates/angular/src/main/webapp/app/admin/docs/_docs.component.ts
```diff
@@ -1,7 +1,14 @@
 import { Component } from '@angular/core';
+import { JhiLanguageService } from 'ng-jhipster';
 
 @Component({
     selector: '<%=jhiPrefix%>-docs',
     templateUrl: './docs.component.html'
 })
-export class <%=jhiPrefixCapitalized%>DocsComponent {}
+export class <%=jhiPrefixCapitalized%>DocsComponent {
+    <%_ if (enableTranslation) { _%>
```
deepu105: this can be ejs block as its different from other places

generators/client/templates/angular/src/main/webapp/app/admin/gateway/_gateway.component.ts
```diff
@@ -13,7 +14,15 @@ export class <%=jhiPrefixCapitalized%>GatewayComponent implements OnInit {
     gatewayRoutes: GatewayRoute[];
     updatingRoutes: Boolean;
 
-    constructor(private gatewayRoutesService: GatewayRoutesService) { }
+    constructor(
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
+        <%_ } _%>
+        private gatewayRoutesService: GatewayRoutesService) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.ts
```diff
@@ -11,7 +13,17 @@ export class <%=jhiPrefixCapitalized%>HealthCheckComponent implements OnInit {
     healthData: any;
     updatingHealth: boolean;
 
-    constructor(private modalService: NgbModal, private healthService: <%=jhiPrefixCapitalized%>HealthService) {}
+    constructor(
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
+        <%_ } _%>
+        private modalService: NgbModal, 
+        private healthService: <%=jhiPrefixCapitalized%>HealthService) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/admin/logs/_logs.component.ts
```diff
@@ -14,10 +15,17 @@ export class LogsComponent implements OnInit {
     orderProp: string;
     reverse: boolean;
 
-    constructor ( private logsService: LogsService ) {
+    constructor (
+        <%_ if (enableTranslation) { _%>
+        private jhiLanguageService: JhiLanguageService,
+        <%_ } _%>
+        private logsService: LogsService ) {
```
deepu105: `) {` can be on new line similar to other places

generators/client/templates/angular/src/main/webapp/app/admin/tracker/_tracker.state.ts
```diff
@@ -1,29 +1,20 @@
...
-    }
+export const trackerRoute: Route = {
+  path: 'jhi-tracker',
+  component: <%=jhiPrefixCapitalized%>TrackerComponent,
+  data: {
+    pageTitle: 'tracker.title'
+  }
 };
+
+// onEnter: (trans: Transition) => {
```
deepu105: we can move this onInit and onDestroy of the component

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
```diff
@@ -1,7 +1,7 @@
 <div>
     <h2>
         <span jhiTranslate="userManagement.home.title">Users</span>
-        <button class="btn btn-primary float-xs-right" uiSref="user-management.new">
+        <button class="btn btn-primary float-xs-right" (click)="createUser()">
```
deepu105: why not use the route?

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
```diff
@@ -30,31 +36,40 @@ export class UserMgmtComponent implements OnInit {
...
+        private router: Router,
+        private userModalService: UserModalService
     ) {
         <%_ if (databaseType !== 'cassandra') { _%>
-        this.itemsPerPage = ITEMS_PER_PAGE;
-        this.page = paginationUtil.parsePage($state.params['page']);
-        this.previousPage = this.page;
-        this.reverse = paginationUtil.parseAscending($state.params['sort']);
-        this.predicate = 'id';
+        this.activatedRoute.data.subscribe(data => {
```
deepu105: add destroy event

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
```diff
@@ -132,7 +144,20 @@ export class UserMgmtComponent implements OnInit {
         this.users = data;
     }
 
-    private onError (error) {
+    private onError(error) {
         this.alertService.error(error.error, error.message, null);
     }
+
+    createUser() {
+        this.modalRef = this.userModalService.open(UserMgmtDialogComponent);
```
deepu105: so we dont have solution to make these work with routes?

deepu105: may be we can use a route to call a component and open the model from onInit or similar event of the component? we can have that as a thin component on the same file as the model component?

generators/client/templates/angular/src/main/webapp/app/layouts/_index.ts
```diff
@@ -3,9 +3,11 @@ export * from './error/error.state';
 export * from './main/main.component';
 export * from './footer/footer.component';
 export * from './navbar/navbar.component';
+export * from './navbar/navbar.state';
```
deepu105: I guess file is removed

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.state.ts
```diff
@@ -0,0 +1,30 @@
+import { Route } from '@angular/router';
+
+import { UserRouteAccessService } from '../../shared';
+import { NavbarComponent } from './navbar.component';
+import { AuthService } from '../../shared';
```
deepu105: is this still needed?

gmarziou:
> for me this is the last "big" thing that is missing for JHipster 4: do you need help? When do you think this could be merged?

How can we proceed? I mean if we merge it on master, will the entity sub generator (on ui-router) still work?

Do we want to replace ui-router by ng router? Or should we try to support both?

What about renaming *.state.ts files to *.route.ts before merging, this would give us an opportunity to try to support both.

deepu105:
IMO this could be part of our final and we can proceed without this for
Beta as master is more stable now and merging this would again make it bit
more unstable


Thanks & Regards,
Deepu

jdubois:
- yes for renaming state.ts to route.ts
- for me, the Angular team forces us to use their router, so we will need to support NG Router, even if we don't like it
- supporting both NG Router and UI Router looks like a lot of work, and I'm already totally fed up with this migration....
- so if we end up with NG Router only, I think it's very bad if we have UI Router for the beta

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.component.ts
sendilkumarn: can help with this ?

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
sendilkumarn: That was exactly what I had in mind. (tried too)
This in turn makes us to create that thin component for every logical entity / modal we need to replicate this. This needs some exploration too.

sendilkumarn:
converted the popup as a `route` 👍

sendilkumarn:
guys, is ng2 entity generation working? 

Whatever I do, it is still pointing to js ( only in entity ). `npm | yarn link` is enough right ? am I missing any thing ?

VictorADS:
@sendilkumarn  Try `yarn link generator-jhipster` in your generated project ?

deepu105:
yes you need run yarn link generator-jhipster each time you create a new
project or delete node_modules


Thanks & Regards,
Deepu

wmarques:
Wow, very good job @sendilkumarn ! :+1: 
Btw I agree with what @jdubois said, let's move on NG Router, maintaining two routers is not worth and will require too much work :disappointed:

sendilkumarn:
Basic Entity works with this code currently. But we have to test a lot of different combinations.

VictorADS:
Is there a reason why we keep the pagingParamResolve ?
As far as I understand we are now using PaginationUtil to parse the query parameters and we are not using the 'pagingParams' token.

deepu105:
We can remove it if not used at all

Thanks & regards,
Deepu

sendilkumarn:
@deepu105 which one ?

deepu105:
That was to @VictorADS regarding his question

sendilkumarn:
Okay 👍  

Let me know if anyother things are missed out.

deepu105:
why are the tests failing?


Thanks & Regards,
Deepu

jdubois:
@deepu105 what do you think about merging this now? I know it's not perfect & that the tests are failing, but for me this is the last thing missing for JHipster 4, so let's merge it and put all our efforts on this. If we combine all our forces on this, I'm sure it will be finished in a couple of days, and then we can do the code freeze before the release.

deepu105:
Let me take a final look please. If ok ill merge it today or tomorrow


Thanks & Regards,
Deepu

gmarziou:
I read @mraible's [great article on Eclipse newsletter](https://www.eclipse.org/community/eclipse_newsletter/2017/january/article3.php) it seems to me that the article itself isn't impacted by this PR, but your [source repo](https://github.com/mraible/jhipster4-demo) will become obsolete.

Shouldn't we wait for Matt to update his repo on a branch? Or if he is not available to do so, I could try to do it.

mraible:
I'd be happy to update my repo once master is close to stable. Or maybe we could create a tag from current master and I can update my article for that?

deepu105:
Ok i'm gonna fix some issues and merge this

deepu105:
I have created a new branch `ng2-ui-router` from current master so that we can refer back to stable branch easily if required. we can delete it once we get ng2 out of beta

deepu105:
I still feel that ui-router was far more cleaner, efficient and simpler compared to ng2 router :(
I wish the angular-cli had support for it

gmarziou:
Few issues related to pagination, I'm working on it.

I also think that each entity should have its own module, my `entity.module.ts` with only 11 entities is 300 lines long. Generation should be simpler too.

jdubois:
Yes +1 for a module per entity

gmarziou:
I open an issue to track it: #4918

generators/client/files-angular.js
VictorADS: Can you explain why the change ? Right now, without i18n, the titles are like `home.title` because we don't change the `pageTitle` field in the state files.

PR #4851, opened 2017-01-12
ng2: fix sass build with postcss config

pascalgrimaud:
It should fix the build with ng2 + sass options.
I just follow the instructions on README here https://github.com/postcss/postcss-loader

@deepu105 : feel free to close this PR if you find another / better solution, as I don't really understand what I'm doing... :clown_face:


jdubois:
I also have no idea what you're doing, but it solves the issue, and I'm fed up with this broken build -> I'm merging it, @deepu105 you can revert it if you don't like it

Always ask for permission after doing :-)

Issue #4849, opened 2017-01-12
liquibase:diff and required relation with entity generator


ctamisier:
##### **Overview of the issue**
When generating an entity with 'yo jhipster:entity' and specifying a required relation.
If we run 'mvn clean compile liquibase:diff', the changelog contains a dropNotNullConstraint.

the generator produces in the java class:
`@ManyToOne
    @NotNull
    private User user;`

##### **Motivation for or Use Case**

It is a required relation so we do not want a dropNotNullConstraint
##### **Suggest a Fix**
the generator should produce:
`  @ManyToOne(optional = false)
    @NotNull
    private User user;`

and with this, liquibase does not generate the dropNotNullConstraint

##### **JHipster Version(s)**
jhipster 3.12.1 (maybe it is already fixed in master branch but I don't know yet where I can check this)

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

> {
>     "fluentMethods": true,
>     "relationships": [
>         {
>             "relationshipName": "freeMailType",
>             "otherEntityName": "freeMailType",
>             "relationshipType": "many-to-one",
>             "relationshipValidateRules": [
>                 "required"
>             ],
>             "otherEntityField": "code"
>         },
>         {
>             "relationshipName": "user",
>             "otherEntityName": "user",
>             "relationshipType": "many-to-one",
>             "relationshipValidateRules": [
>                 "required"
>             ],
>             "otherEntityField": "login",
>             "ownerSide": true
>         }
>     ],
>     "fields": [
>         {
>             "fieldName": "dayDate",
>             "fieldType": "LocalDate",
>             "fieldValidateRules": [
>                 "required"
>             ]
>         }
>     ],
>     "changelogDate": "20170112170527",
>     "dto": "mapstruct",
>     "service": "serviceClass",
>     "entityTableName": "free_mail_day",
>     "pagination": "pagination"
> }

gmarziou:
Please include your .yo-rc.json as it could be related to your database or other options

jdubois:
Without your `.yo-rc.json` we can't test it, this highly depends on the database you use.

Then, you also need to check if it's a Liquibase bug or not, this might not be our fault at all.

jdubois:
This ticket has been closed as the guidelines are not followed.

Tickets must follow our [Guidelines](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#submitting-an-issue), as mentioned in:

[...]

PR #4854, opened 2017-01-12
add optional=false on a required many-to-one relationship

ctamisier:
https://github.com/jhipster/generator-jhipster/issues/4849


jdubois:
Some interesting information on this: http://stackoverflow.com/a/3331954/150868
And I learned something today, I didn't know this difference!

ctamisier:
Ok nice, I didn't know either.

So using `@JoinColumn(nullable=false)` is the annotation that means exactly what we want about the database schema -> a not null column.

But liquibase is smart enough to understand that optional=false means a not null column as well. So it is sufficient.

jdubois:
Yes, and it's shorter, and might be better with Hibernate, cool!

Issue #4794, opened 2016-12-31
Angular 2 TODO for next


deepu105:
This ticket is to track TODO items after the angular2 release as per #4549 

Items with **- to verify first** needs to be checked/tested first to see if already implemented

**Angular**
- [x] fix any commented out and FIXME/TODO in code  **- to verify first** @JulienMrgrd 
- [x] Support AOT where possible  **- to verify first**
- [x] polish templates
- [x] check [styleguide](https://angular.io/docs/ts/latest/guide/style-guide.html) compliance  **- to verify first** #5879 
- [x] check and optimize the way how files are served from target|build folder  **- to verify first**
- [x] Check #5825 to upgrade dependencies, including ng-bootstrap. There are 2 issues with ng-jhipster, and tests must be upgraded
- [x] Validate that `./mvnw -Pprod` uses a prod build (and do the same with Gradle) [ SK It is working as intended ]
- [x] Put all prod assets under `/app` (for the Angular files) and `/content`(for the static assets) and validate that there is a hash for all of them (they will get cache headers from the server-side) -[sk]
 hash applicable only for `prod` #5878 
- [x] correct `AlertService` and `InterceptableHttp` which have warnings `This will become an error in Angular v5.x` @JulienMrgrd 

**webpack**
- [x] Make HMR possible with webpack
- [x] ng annotate/ng template (if at all required in ng2)  **- to verify first** - Can close this as we don't have any uses now
- [x] sourcemaps for prod and dev [seems like there is no sourcemap]
- [x] error notifications during build (it was cool to have using gulp notify)
- [x] get app version from maven/gradle for prod @deepu105
- [x] verify and optimize prod webpack build  #5530
- [x] vendor.dll.js created in prod is not hashed and does not have expire header @sendilkumarn 
- [x] Reduce number of compilation warnings in browser console #4794
- [x] Make TSLint to fail build when there are lint errors

**General**
- [ ] Test all scenarios and make sure there is no bugs, scenarios listed in #4301
- [x] Remove beta label from angular2 option (when all critical bugs known/reported are fixed, all dependencies updated to atleast RCs)
- [x] Make angular 4 the default option in the question

VictorADS:
@deepu105  I got HMR working but there is a problem. When we change angular 2 files like html or ts the whole angular 2 module is refreshed causing a blink in the screen. I thought it was i18n loading but in reality I'm pretty sure it's ui-router that causes it because I tried angular/router and everything works fine.

[...]

PR #4858, opened 2017-01-13
Add HMR and sourcemap in Webpack

VictorADS:
I also added the DLL plugin which allows to build the vendor chunk in a separate process and use it in the main config without having to recompile it

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -96,6 +96,7 @@
     "webpack": "2.2.0-rc.3",
     "webpack-dev-server": "2.2.0-rc.0",
     "webpack-merge": "2.0.0",
+    "write-file-webpack-plugin": "3.4.2",
     <%_ if (buildTool == 'maven') { _%>
     "xml2js": "0.4.17",
     <%_ } _%>
@@ -121,9 +122,8 @@
     "tsc:w": "tsc -w",
     "lint": "tslint src/main/webapp/app/**/*.ts --force",
     "lint:fix": "tslint src/main/webapp/app/**/*.ts --fix --force",
-    "webpack:hmr": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --content-base src/ --port=8090",
     "webpack:build": "webpack --config webpack/webpack.dev.js",
-    "webpack:dev": "webpack --watch --config webpack/webpack.dev.js",
+    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --content-base src/ --port=8090",
     "webpack:prod": "webpack -p --config webpack/webpack.prod.js"
   }
 }
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -127,7 +127,7 @@
     "lint": "tslint src/main/webapp/app/**/*.ts --force",
     "lint:fix": "tslint src/main/webapp/app/**/*.ts --fix --force",
     "webpack:build": "webpack --config webpack/webpack.vendor.js && webpack --config webpack/webpack.dev.js",
-    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9002",
+    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9060",
     "webpack:prod": "webpack -p --config webpack/webpack.vendor.js && webpack -p --config webpack/webpack.prod.js"
   }
 }
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -44,6 +44,7 @@
     <%_ if (protractorTests) { _%>
     "@types/selenium-webdriver": "2.53.33",
     <%_ } _%>
+    "add-asset-html-webpack-plugin": "1.0.2",
     "angular2-infinite-scroll": "0.2.8",
     "angular-cli": "1.0.0-beta.24",
     "angular2-template-loader": "0.6.0",
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -96,6 +96,7 @@
     "webpack": "2.2.0-rc.3",
     "webpack-dev-server": "2.2.0-rc.0",
     "webpack-merge": "2.0.0",
+    "webpack-visualizer-plugin": "0.1.9",
     "write-file-webpack-plugin": "3.4.2",
     <%_ if (buildTool == 'maven') { _%>
     "xml2js": "0.4.17",
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -40,6 +40,7 @@
     "@types/angular-animate": "1.5.6",
     "@types/angular-mocks": "1.5.8",
     "@types/jasmine": "2.5.38",
+    "@types/node": "6.0.45",
     <%_ if (protractorTests) { _%>
     "@types/selenium-webdriver": "2.53.33",
     <%_ } _%>
@@ -93,6 +94,7 @@
     "tslint-loader": "3.3.0",
     "typescript": "2.1.4",
     "webpack": "2.2.0-rc.3",
+    "webpack-dev-server": "2.1.0-beta.12",
     "webpack-merge": "2.0.0",
     <%_ if (buildTool == 'maven') { _%>
     "xml2js": "0.4.17",
@@ -119,6 +121,7 @@
     "tsc:w": "tsc -w",
     "lint": "tslint src/main/webapp/app/**/*.ts --force",
     "lint:fix": "tslint src/main/webapp/app/**/*.ts --fix --force",
+    "webpack:hmr": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --content-base src/ --port=8090",
     "webpack:build": "webpack --config webpack/webpack.dev.js",
     "webpack:dev": "webpack --watch --config webpack/webpack.dev.js",
     "webpack:prod": "webpack -p --config webpack/webpack.prod.js"
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -94,7 +94,7 @@
     "tslint-loader": "3.3.0",
     "typescript": "2.1.4",
     "webpack": "2.2.0-rc.3",
-    "webpack-dev-server": "2.1.0-beta.12",
+    "webpack-dev-server": "2.2.0-rc.0",
     "webpack-merge": "2.0.0",
     <%_ if (buildTool == 'maven') { _%>
     "xml2js": "0.4.17",
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -117,15 +117,17 @@
     "test": "karma start src/test/javascript/karma.conf.js",
     "test:watch": "karma start --watch",
     <%_ if (protractorTests) { _%>
-    "postinstall": "webdriver-manager update",
+    "postinstall": "webdriver-manager update && npm run webpack:build",
     "e2e": "protractor <%= TEST_SRC_DIR %>protractor.conf.js",
+    <%_ } else { _%>
+    "postinstall": "npm run webpack:build",
     <%_ } _%>
     "tsc": "tsc",
     "tsc:w": "tsc -w",
     "lint": "tslint src/main/webapp/app/**/*.ts --force",
     "lint:fix": "tslint src/main/webapp/app/**/*.ts --fix --force",
-    "webpack:build": "webpack --config webpack/webpack.vendor.js",
-    "webpack:dev": "npm run webpack:build && webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --content-base src/ --port=8090",
-    "webpack:prod": "webpack -p --config webpack/webpack.prod.js"
+    "webpack:build": "webpack --config webpack/webpack.vendor.js && webpack --config webpack/webpack.dev.js",
+    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9002",
+    "webpack:prod": "webpack -p --config webpack/webpack.vendor.js && webpack -p --config webpack/webpack.prod.js"
   }
 }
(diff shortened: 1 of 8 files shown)


generators/client/templates/angular/_package.json
```diff
@@ -119,8 +124,8 @@
     "tsc:w": "tsc -w",
     "lint": "tslint src/main/webapp/app/**/*.ts --force",
     "lint:fix": "tslint src/main/webapp/app/**/*.ts --fix --force",
-    "webpack:build": "webpack --config webpack/webpack.dev.js",
-    "webpack:dev": "webpack --watch --config webpack/webpack.dev.js",
+    "webpack:build": "webpack --config webpack/webpack.vendor.js",
```
deepu105: can we make sure build will compile the entire project instead of just vendor?

generators/client/templates/angular/webpack/_webpack.common.js
```diff
@@ -144,6 +107,9 @@ module.exports = function (options) {
                 inject: 'body',
                 data: DATAS
             }),
+            new AddAssetHtmlPlugin([
+                { filepath: path.resolve('./target/www/vendor.dll.js'), includeSourcemap: false }
```
deepu105: path should be build specific do `/<%= BUILD_DIR %>/www/...` everywhere

generators/client/templates/angular/webpack/_webpack.common.js
```diff
@@ -125,7 +84,11 @@ module.exports = function (options) {
         },
         plugins: [
             new CommonsChunkPlugin({
-                names: ['manifest', 'polyfills', 'vendor'].reverse()
+                names: ['manifest', 'polyfills'].reverse()
+            }),
+            new webpack.DllReferencePlugin({
+                context: './',
+                manifest: require(path.resolve('./target/www/vendor.json')),
```
deepu105: use BUILD_DIR

generators/client/templates/angular/webpack/_webpack.dev.js
```diff
@@ -1,11 +1,34 @@
...
+                '/swagger-resources',
+                '/v2/api-docs',
+                '/h2-console'
+            ],
+            target: 'http://127.0.0.1:<%= serverPort %>',
+            secure: false
+        }]
+    },
+    output: {
+        path: path.resolve(<% if (buildTool === 'gradle') { %>'./build/www')<% } else { %>'./target/www')<% } %>,
```
deepu105: use BUILD_DIR

generators/client/templates/angular/webpack/_webpack.prod.js
```diff
@@ -2,14 +2,18 @@ const commonConfig = require('./webpack.common.js');
...
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
+const Visualizer = require('webpack-visualizer-plugin');
 const ENV = 'prod';
 
 module.exports = webpackMerge(commonConfig({env: ENV}), {
-    output: {
+        devtool: 'cheap-module-source-map',
+        output: {
+        path: <% if (buildTool === 'gradle') { %>'./build/www'<% } else { %>'./target/www'<% } %>,
```
deepu105: use BUILD_DIR

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -1,3 +1,4 @@
+import './vendor.ts';
```
deepu105: why is this imported again inst it specified as entry in vendor build conf

generators/client/templates/angular/webpack/_webpack.vendor.js
```diff
@@ -0,0 +1,66 @@
...
+                loaders: [
+                    'file-loader?hash=sha512&digest=hex&name=[hash].[ext]',
+                    'image-webpack-loader?bypassOnDebug&optimizationLevel=7&interlaced=false'
+                ]
+            },
+        ]
+    },
+    output: {
+        filename: '[name].dll.js',
+        path: './target/www',
```
deepu105: use BUILD_DIR

generators/client/templates/angular/webpack/_webpack.vendor.js
```diff
@@ -0,0 +1,66 @@
...
+        library: '[name]',
+    },
+    plugins: [
+        new webpack.DllPlugin({
+            name: '[name]',
+            path: './target/www/[name].json',
```
deepu105: use BUILD_DIR

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
VictorADS: If I don't import this, the style tags for bootstrap and fa aren't injected in index.html. It might be because we build them out of the html webpack process. I didnt find a better solution so if you have a better idea

deepu105:
ok


Thanks & Regards,
Deepu

jdubois:
@VictorADS your build is failing, and I have the same error locally

VictorADS:
@jdubois are you running `yarn run webpack : build && yarn run webpack : dev`  ?

jdubois:
Oh no, only `yarn run webpack:dev`, and I guess it's the same issue with Travis

deepu105:
@VictorADS it shouldnt require 2 commands IMO can we add the watch on dev alone and let it build automatically

VictorADS:
@deepu105 what do you mean by 

> add the watch on dev alone and let it build automatically

jdubois:
Yes we should only have one command to build & run at the same time. In fact, it should be the default command, like running `gulp` in JHipster 3 -> it should be very easy to have everything running in just one step.

I tried it this morning:

- The speed is really good. Maybe we could remove the lint, to make it even faster.
- I hope that moving to NG Router (see #4846 ) will make it even faster (no more vizualizer -> it's great but it slows things down), and hopefully true hot reload in the browser (at the moment my browser refreshed everything)

Anyway, that's definitely some great improvement!!!!

VictorADS:
Can't we have like one command that build everything up (vendor + commonconfig) and keep the `yarn run webpack:dev` as it is ?

Because we only need to compile vendor stuff once to make it work

gmarziou:
+1 to remove lint

cbornet:
Is it possible to make lint a command line option ?

jdubois:
So, here is how I would like it to be:

- By default: we do build + dev, but not lint -> the default command should be easy, and work out of the box, so that it's great for new users
- There is an option to have just "dev", so people can speed up their build easily
- There is an option for "lint", for people who want it also

jdubois:
And when you run Maven:

- by default you have the dev profile, which runs the default  Webpack task (build + dev) - probably there needs to be a forked process for dev
- in "prod" mode, it does build + lint + karma tests

gmarziou:
About maven, I currently have an issue, the frontend-maven-plugin takes forever to download node, I thought it's was a temporary nodejs.org issue.

As I have node installed globally,I looked for a way to run  only spring-boot in one window and yarn webpack:dev in another. So far, the only way I know is to run the main app which I find a bit confusing for a beginner. Is there a way of using maven to run only the java part and not the frontend-maven-plugin?

gmarziou:
I find it weird that our default maven goal does not reflect the workflow we want to promote in dev : hot reloading with webpack frontend server and sprint boot backend server.

ideally, user should have to run a single command that would spawn the 2 servers, I don't know is it is feasible with maven or gradle.

jdubois:
@gmarziou yes, in fact we have 2 main use cases:

- Running from Maven (same for Gradle, of course): you just run `./mvnw` and you should have the Java server running, as well as Webpack building everything. It would be super-cool to have also Webpack do the hot reload, but: 
  - it could be complex to achieve (we probably need to fork the process: there is one process running Maven and another one running Webpack),
  - and that's not how it works today (you run Maven and Gulp in 2 different terminals today, so that's the same as running Maven and Webpack in the next release).
This is mostly used for first-time users: it just runs on the terminal, everybody should be able to have this running quickly.
- Running in the IDE: you run your Java server as a main class (and have hot-reload with Spring Boot, debugging, etc...), and you run Webpack in a terminal (might be an embedded terminal in your IDE, of course), and you have hot reload with Webpack. I expect IDEs to have Webpack support, so this will be improved automatically. This is the best possible configuration, as you have hot reload in Java and TypeScript, but it's a bit more complex to set up than just running Maven.

deepu105:
Regarding the webpack build flow. I guess Victor you might have
misunderstood it a little but no worries. Just keep it as it is in your PR
ill improve it later.

Thanks & regards,
Deepu

deepu105:
guys I have started a thread to discuss the build flow #4868

generators/client/templates/angular/_package.json
```diff
@@ -117,15 +117,17 @@
     "test": "karma start src/test/javascript/karma.conf.js",
     "test:watch": "karma start --watch",
     <%_ if (protractorTests) { _%>
-    "postinstall": "webdriver-manager update",
+    "postinstall": "webdriver-manager update && npm run webpack:build",
```
jdubois: not sure we can do the `&&` -> does this work on Windows?

deepu105: it should work as its executed by node but its always good to check

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
deepu105: but does it get double imported then?

generators/client/templates/angular/webpack/_webpack.dev.js
```diff
@@ -1,11 +1,41 @@
...
+const execSync = require('child_process').execSync;
+const fs = require('fs');
+const ddlPath = './<%= BUILD_DIR %>www/vendor.json';
+
+if (!fs.existsSync(ddlPath)) {
+    execSync('webpack --config webpack/webpack.vendor.js');
+}
 
 module.exports = webpackMerge(commonConfig({env: ENV}), {
+    devtool: 'cheap-module-source-map',
```
deepu105: inline-source-map or cheap-module-eval-source-map might be better for dev as they would be faster to build. https://webpack.github.io/docs/configuration.html#devtool

generators/client/templates/angular/webpack/_webpack.dev.js
```diff
@@ -19,9 +49,13 @@ module.exports = webpackMerge(commonConfig({env: ENV}), {
         new BrowserSyncPlugin({
             host: 'localhost',
             port: 9000,
-            proxy: 'http://localhost:<%= serverPort %>'
+            proxy: 'http://localhost:9002'
```
deepu105: can we use some port number which has less chances of conflict

generators/client/templates/angular/webpack/_webpack.prod.js
```diff
@@ -2,14 +2,18 @@ const commonConfig = require('./webpack.common.js');
 const webpackMerge = require('webpack-merge');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
+const Visualizer = require('webpack-visualizer-plugin');
 const ENV = 'prod';
 
 module.exports = webpackMerge(commonConfig({env: ENV}), {
-    output: {
+        devtool: 'cheap-module-source-map',
```
deepu105: For prod we can do `source-map` as its not build always

deepu105:
could you rebase and see if builds are passing

generators/client/templates/angular/webpack/_webpack.dev.js
gmarziou: Why don't we keep `<%= serverPort %>`? This is a must for gateways and also useful for #4844

gmarziou:
Just a side comment, when webpack compiles ts files, it may skip files that are not required. I just experienced it today, `yarn tsc` was failing on one file that was not compiled by webpack. I think we should add `yarn tsc` as part of our travis builds.

generators/client/templates/angular/webpack/_webpack.dev.js
deepu105: I think this one is proxying through webpack dev server so a fixed port might not be an issue its not the actual server port

gmarziou: OK sorry.

deepu105:
@gmarziou agree. It might also be a good idea to invoke it for the prod webpack build build

gmarziou:
Good point for prod build.

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
VictorADS: No, webpack seems to be smart enough to not import it twice. Btw I added a visualizer for our bundles. When you build for prod, look at the "stats.html" file in target/www its really great !

deepu105:
ok nice


Thanks & Regards,
Deepu

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
deepu105: ok nice :clap:

VictorADS:
Do you guys  want me to add something more ?

generators/client/templates/angular/webpack/_webpack.common.js
deepu105: I think you forgot to update `<%= BUILD_DIR %>` here can you update it please

deepu105:
great work ill merge it as soon as travis is happy :)

deepu105:
@VictorADS one of the ng2 build seems to fail consistently could you try to check it locally?

VictorADS:
Thats odd, i tried locally and it succeeded

deepu105:
did you run protractor on the same yo-rc from sass-noi18n travis build?

deepu105:
ok it passed

gmarziou:
Great work @VictorADS, I noticed a minor thing by looking at production [hash].main.bundle.js: HTML templates are not minified, they include unneeded white spaces. In angular 1, I think these spaces were stripped by our build process, maybe there's a similar option to enable?

~~~
{t.exports=n.p+"a30deb26b4eb1521433021e326cbcc2c.png"},function(t,e){t.exports='<div>\n    <div
 class="row">\n        <div class="col-md-8 offset-md-2">\n            <h1
 jhiTranslate="activate.title">Activation</h1>\n\n            <div class="alert alert-success" 
*ngIf="success">\n                <span jhiTranslate="activate.messages.success">\n
                    <strong>Your user has been activated.</strong> Please\n
                    <a class="alert-link" (click)="login()" 
~~~

deepu105:
@gmarziou yes thats one of the todo items in the #4794

gmarziou:
ok, thanks

PR #4861, opened 2017-01-14
Fix health karma test

mohamed2a:
I override the template of `HealthCheckController` to test health component and service


mohamed2a:
related to #4302

PR #4863, opened 2017-01-14
some fixes in entity generator

ctamisier:
- Prompt 'required' validation when target Entity is a User
- add optional=false on a required one-to-one relationship


jdubois:
LGTM

PR #4865, opened 2017-01-14
No need to use the JavaMailSenderImpl, enough to get the interface.

gzsombor:
Make the UserResource.getUser call similar to getAllUser - ManagedUserVM is received consistently

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_MailService.java
+++ b/generators/server/templates/src/main/java/package/service/_MailService.java
@@ -8,7 +8,7 @@ import org.apache.commons.lang3.CharEncoding;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.MessageSource;
-import org.springframework.mail.javamail.JavaMailSenderImpl;
+import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.MimeMessageHelper;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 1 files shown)


jdubois:
Yes for the "MailSender", but here you also modify some other code, including using the ManagedUserVM on the service layer, which you shouldn't do (package tangling)

gzsombor:
There are already methods in UserService which take or return ManagedUserVM : 
[createUser](https://github.com/gzsombor/generator-jhipster/blob/c92d46400f0715611f01b29ca4b5028bb9577d9f/generators/server/templates/src/main/java/package/service/_UserService.java#L165)
[updateUser](https://github.com/gzsombor/generator-jhipster/blob/c92d46400f0715611f01b29ca4b5028bb9577d9f/generators/server/templates/src/main/java/package/service/_UserService.java#L217)
[getAllManagedUsers](https://github.com/gzsombor/generator-jhipster/blob/c92d46400f0715611f01b29ca4b5028bb9577d9f/generators/server/templates/src/main/java/package/service/_UserService.java#L278)

I've noticed because I've added a field to ManagedUserVM which initialized from a lazy loaded field of user.
I don't know why the travis build is failed

jdubois:
No package in the service layer should call the Web layer.
At some point we had a huge mess here, and I refactored it, maybe there are some this stuff still missing there.
They should be removed before JHipster 4.

jdubois:
I'm creating a specific ticket for this

gzsombor:
The build failed with some angular2 errors, which is surely unrelated to this change

PR #4869, opened 2017-01-15
Eliminate a bunch of warnings about the incorrect " characters used

gzsombor:
The build is complaining about it, it's easy to fix.


jdubois:
Yes those were bugging me too!!! We need to enforce linting before commiting...
Thanks !!!

PR #4870, opened 2017-01-15
priority to "url: '/{login}/edit'" vs "url: '/user/:login'"

ctamisier:
fix the edition of the user with 'user' login.
The side effect is that we can't have a user called 'edit'.

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.state.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.state.js
@@ -58,27 +58,6 @@
             }
             <%_ } _%>
         })
-        .state('user-management-detail', {
-            parent: 'user-management',
-            url: '/user/:login',
-            data: {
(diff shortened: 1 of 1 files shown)


ctamisier:
This issue does not occur on jhipster 3.12.1 (occurs in master branch), so I guess this PR would be just a workaround..
So ignore this PR I would say, but the problem is here..

jdubois:
OK, I confirm the issue, I didn't know that (please create a ticket before, explaining the issue, so we know what you are correcting)

ctamisier:
ok I will create an issue with more details, actually the problem is also in jhipster 3.12.1 but it does not occur the same way..

jdubois:
No, I mean do it next time -> here I understand the issue, I'm going to work on it, no worries

jdubois:
Oh, you just moved the route, that's easy :-)
OK I'm merging this without even testing :-)

jdubois:
yes we'll need to disable that, but that's not a huge priority

PR #4872, opened 2017-01-15
Fix infinite scroll with elasticsearch

gzsombor:
Further testing needed for the other paging options, I'm going to do that as well.

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
@@ -5,7 +5,7 @@
     }
 _%>
 import { Injectable } from '@angular/core';
-import { Http, Response, URLSearchParams } from '@angular/http';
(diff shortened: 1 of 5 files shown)


jdubois:
Great!!! I didn't think we could have that ready for JHipster 4!

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
```diff
@@ -59,46 +60,62 @@ export class <%= entityClass %>Service {
         });
     }
 
-    query(req?: any): Observable<Response> {
-        let options: any = {};
+<%_ if(hasDate) { _%>
+    convertResponse(res: any): any {
```
deepu105: can we make these methods private?

gzsombor: Good idea! We can have private methods in TypeScript - I always forget it :smile:

Issue #4867, opened 2017-01-15
Remove the use of the Web layer in the service layer


jdubois:
As discussed in #4865

We have some Web/REST objects being called from the service layer, which is an architecture error, and causes package tangling (this should be seen by Sonar, I need to check).

Typically, the `ManagedUserVM` shouldn't be used in the service layer, that's why we have a `UserDTO`-> the `VM` is for "View Model", that's used for displaying in the Angular/HTML views, and the `DTO` is for "Data Transfer Object", that's used for transferring the object from one layer to another (if the entity isn't enough). Of course a DTO can be used as a VM, but the opposite isn't true.

gzsombor:
In my opinion, the DTO-s should be returned from the service layer - and calling the service layer from the web layer should use DTO-s as well - so the web layer doesn't depend on the back-end domain model. This has the following advantages, at least:
* You can create and ship a separate API modul - which just contains the necessary DTO's, which can be used in any java based rest client, without exposing the domain entities
*  No more lazy loading problems during creating the DTO's, as the DTO is created _inside_ the service layer, _inside_ a transaction
* No more lazy loading problems during the JSON creation, as the DTO's doesn't have any lazy loaded, hibernate managed variables
* You can add various restrictions and validations to the DTO's, even with annotations, which can be easily verified in the call site, as opposed to calling the service with domain entities, where you can't specify easily, which constraints is needs to fulfilled.

jdubois:
@gzsombor we totally agree for the DTO -> the VMs are different, they are only needed for displaying on the Web layer, something specific.

So we have:

- Entities by default: that's the easiest to use and manage
- DTOs as an option, served by the service layer: better for everything you just said, including lazy loading and business rules
- VMs as an option, served by the Web/REST layer: only for displaying something specific on the view layer

And you can have entities + DTO + VMs all at the same time, that's a bit complex, but if you have complex business requirements, that doesn't shock me.

jdubois:
@gzsombor I'm pretty sure we agree here, there is just a small misunderstanding: I'm going to refactor this in a branch, and let you have a look.

jdubois:
@gzsombor please have a look at #4877 and tell me what you think

PR #4873, opened 2017-01-15
fix many-to-many relationship when User is the target Entity

ctamisier:
When creating a many-to-many relationship to a User using the generator:

- UserDTO needs the id field to make the JPA relation. I moved it from ManagedUserVM.
- UserDTO needs the login (to show in the table when there is no pagination)

I had to remove `@Mapping(target = "id", ignore = true)`, was it necessary before for some reasons ?

As the User is the target entity it is a **unidirectional** many-to-many, too hard or risky to change the User class right ?

So after the generation we have:
`public EntityA addUser(User user) {`
       ` this.users.add(user);`
        `user.getEntityAS().add(this);` //does not compile and need to be remove manually (so far)
        `return this;`
`    }`

The strange thing about this PR is that we have the 'id' field in UserDTO which is mainly use in the subclass ManagedUserVM. 
The id field in UserDTO is only used to solve this 'many-to-many relationship to a User' use case...

this use case using the generator creating an EntityA (EntityA.json) :
{
    "fluentMethods": true,
    "relationships": [
        {
            "relationshipName": "user",
            "otherEntityName": "user",
            "relationshipType": "many-to-many",
            "otherEntityField": "login",
            "ownerSide": true,
            "otherEntityRelationshipName": "entityA"
        }
    ],
    "fields": [],
    "changelogDate": "20170115151136",
    "dto": "mapstruct",
    "service": "serviceClass",
    "entityTableName": "entitya",
    "pagination": "no"
}

### DIFF
--- a/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
+++ b/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
@@ -109,6 +109,10 @@ public class UserDTO {
         return login;
     }
 
+    public void setLogin(String login) {
+        this.login = login;
+    }
+
     public String getFirstName() {
         return firstName;
     }
--- a/generators/server/templates/src/main/java/package/service/mapper/_UserMapper.java
+++ b/generators/server/templates/src/main/java/package/service/mapper/_UserMapper.java
@@ -29,7 +29,6 @@ public interface UserMapper {
     <%_ if ((authenticationType == 'session') && (databaseType == 'sql')) { _%>
     @Mapping(target = "persistentTokens", ignore = true)
     <%_ } } _%>
-    @Mapping(target = "id", ignore = true)
     @Mapping(target = "activationKey", ignore = true)
     @Mapping(target = "resetKey", ignore = true)
     @Mapping(target = "resetDate", ignore = true)


jdubois:
Hold on, this is linked to #4867 let me have a look before merging

jdubois:
So I'm going to totally break your PR with #4877

If my PR gets merged, the "id" will indeed be moved from the "VM" to the "DTO", which should make your code easier.

Now for your questions:

- We try not to touch the User class, as people might rely on it, or modify it, so it should be as simple as possible
- I guess the "id" is ignored in the mapping for security reasons: you normally only get users with the login, not the id. I'm not sure if this is relevant: is that an issue for you?

ctamisier:
Ok for the PR breaking changes, I will refactor this PR accordingly to the new code.

To make the many-to-many working with the User as the target, the 'id' is required in UserDTO so I do not ignore it anymore in the mapping...

jdubois:
Yes I think you can remove the "id" mapping, when people are touching the `User` entity there are supposed to know what they are doing

jdubois:
My PR #4877 is merged, can you have a look @ctamisier ?

ctamisier:
Ok I pushed the updates

generators/server/templates/src/main/java/package/service/mapper/_UserMapper.java
```diff
@@ -54,7 +53,7 @@ default User userFromId(<% if (databaseType == 'mongodb' || databaseType == 'cas
     }
 
     default Set<Authority> authoritiesFromStrings(Set<String> strings) {
-        return strings.stream().map(string -> {
+        return strings == null ? null : strings.stream().map(string -> {
```
jdubois: Nobody should be allowed to send a `null` parameter to the method, this shouldn't happen

generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
```diff
@@ -109,6 +109,10 @@ public String getLogin() {
         return login;
     }
 
+    public void setLogin(String login) {
```
jdubois: This means the object isn't immutable anymore, why do you want to be able to change the user login? I don't think this is currently needed (the method is not called)

ctamisier: The data from UserDTO class is used to send back to the front-end.
The login of the user is used by default to be displayed (in the table without pager).
So the login in UserDTO need to be set from the User (mapstruct mechanism, if there is not setter, the value is not 'copied')

generators/server/templates/src/main/java/package/service/mapper/_UserMapper.java
ctamisier: A null value is send from the front-end to the back-end (only the id and the login(with this PR) have values, according to the setters in UserDTO), when we save the entity that has the many-to-many relationship to the User.
And I think it should stay null because we don't do anything with the authorities at this level. We just want to save the relations (in the joint table).

jdubois: Not sure about this: if the front end does send a null value, why don't we have an NPE? Anyway I don't like to have null values hanging around

ctamisier: I guess it's the first time we make a such relation to a User:
EntityA <-*------*-> User

If we save EntityA, there is an NPE.. (So far we don't have this use case I suppose).

If we don't want to have a null value, it means that we populate the UserDTO with all the data but only the 'id' will be used (to persist the relation between EntityA and the User).

jdubois: I'm sorry but I don't understand the use case. How do you save EntityA? Just doing a `.persist()`? Or you add it to an Hibernate-managed User?

ctamisier: Sorry I forgot to say that I'm using the entity generator.
I'm creating an EntityA that has a many-to-many relationship to a User.
And I want to be able to create an EntityA and link it to several using the generated backend and frontend code.

ctamisier: I updated the first post in the conversation adding the content of my EntityA.json

jdubois: Still, that feels wrong to me to have a `null` hanging around here -> I need to check this. Could you split your PR in 2? We could merge the rest of this PR, and then have a specific PR for this "null" issue, and discuss this more properly. Besides, I'm not sure Sonar will like it, and we work hard to have the best Sonar rating at https://sonarqube.com/dashboard/?id=1054756

ctamisier:
I pushed the changes about:

> Still, that feels wrong to me to have a null hanging around here -> I need to check this. Could you split your PR in 2? We could merge the rest of this PR, and then have a specific PR for this "null" issue, and discuss this more properly. Besides, I'm not sure Sonar will like it, and we work hard to have the best Sonar rating at https://sonarqube.com/dashboard/?id=1054756

Maybe someone should try without this PR (first) to understand what's going on about all this..

jdubois:
Yes, sorry @ctamisier but I didn't have the time to check... JHipster 4 is coming, and it's going to take most of my free time in the next days/weeks

ctamisier:
sure, no problem !, I will probably create another issue about the null, and we'll check this later.

PR #4874, opened 2017-01-16
Fix trackBy systax warning

tientq:
Ref: https://angular.io/docs/ts/latest/api/common/index/NgFor-directive.html


gmarziou:
Thanks

PR #4877, opened 2017-01-16
Remove the use of the Web layer in the service layer

### LINKED ISSUE #4867: Remove the use of the Web layer in the service layer

jdubois:
Fix #4867

### DIFF
--- a/generators/server/templates/src/main/java/package/domain/_User.java
+++ b/generators/server/templates/src/main/java/package/domain/_User.java
@@ -96,11 +96,13 @@ public class User<% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
     @Field("lang_key")<% } %><% if (databaseType == 'cassandra') { %>
     @Column(name = "lang_key")<% } %>
     private String langKey;
+    <%_ if (databaseType == 'mongodb' || databaseType == 'sql') { _%>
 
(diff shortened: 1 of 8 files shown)


gzsombor:
Makes sense !

If I were you, I would introduce "imageUrlSupported" flag, and "pkType" variable in the server generator, instead of the lot's of repetitive if-s :)

jdubois:
So happy you like it :-)
Yes my code is repetitive but I find it clearer that way, for newbies. Anyway, at some point I hope that we will support this in Cassandra, so they should be totally deleted in the end.

deepu105:
I think we already have the `pkType` flag in server generator

gzsombor:
Just in the entity generator

deepu105:
Oh ya sorry

Thanks & regards,
Deepu

PR #4880, opened 2017-01-16
Add settings spec to Karma

ramzimaalej:


generators/client/files-angular.js
```diff
@@ -246,6 +246,20 @@ const files = {
                 'account/password/_password-strength-bar.css'
             ]
         },
+        {
```
deepu105: why is this duplicated again?

deepu105:
lint failure as well


Thanks & Regards,
Deepu

generators/client/files-angular.js
ramzimaalej: darn, didn't notice that when I merged, let me fix it.

PR #4883, opened 2017-01-16
clean code

ctamisier:


deepu105:
@ctamisier please add a short description on what you are doing and if its related to any issue mention that as well

ctamisier:
- Remove nullable = true, it is already the default value.
- use curly braces (like in the rest of the file and in the entity generator) when defining variables in angular routes.
There is no issue associated to this.

(I will know for the next time to put a real description)

jdubois:
OK, and next time please do 2 commits: those are different issues. It's useful for us, for reverting or cherry picking a change for example.

ctamisier:
yeah totally right...

PR #4886, opened 2017-01-17
Update _mysql.yml set mysql charset

anjia0532:
yo jhipster:docker-compose generate docker-compose.yml . and not set mysql charset , although  [application-prod.ym](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/_application-prod.yml#L64) set the jdbc string use utf-8 but it was still throw exception when insert the unicode character( e.g. chinese )


anjia0532:
this config not support emoji . emoji need utf8mb4 not utf8

jdubois:
Didn't test, but makes a lot of sense, so I'm merging, thanks!

PR #4891, opened 2017-01-18
Update README for angular2 and yarn

gmarziou:
I gave an example on how to add a dependency, maybe this should be on web site or both.


pascalgrimaud:
There are some problems to git clone with Travis... Anyway, it's just a README

gmarziou:
I should have put skip CI

PR #4896, opened 2017-01-18
Jenkinsfile improvements

cbornet:
* Only depend on JDK8 (which is already required for jenkins)
* Support yarn
* Archive artifacts
* Support SonarQube


cbornet:
I'll update the docs now

PR #4898, opened 2017-01-18
Treat reset.json as template.

tan9:


gmarziou:
Could you explain what problem you solve ?

tan9:
@gmarziou, reset.json contains template syntax introduced by the inclusion of angular2 feature. However, generated resource in locale other than English will be as-is to the template, becoming an invalid JSON.

gmarziou:
OK, so it should be renamed as _reset.json as the other templates.

gmarziou:
OK thanks. Please fix the conflict.

tan9:
@gmarziou , conflict resolved :)

PR #4902, opened 2017-01-19
fix uaa cors with 401

ruddell:
Related to https://github.com/jhipster/generator-jhipster/issues/4704, I forgot to fix it for UAA.  401s now have the correct CORS headers (when enabled).


ruddell:
Travis failed one of the builds due to an ng2 protractor problem - not bothering to restart it

PR #4903, opened 2017-01-19
Add a sub generator for various CI/CD pipelines

cbornet:
Following discussion on #4600 
This PR is for a subgen to generate CI/CD config files with options

Currently supported pipelines and integrations are:

Feature | Jenkins | Gitlab | Circle |
| ----- |:-------:|:------:|:------:|
Yarn|x (Maven)|x (Maven)|
Gitlab status|x|N/A|
SonarQube|x|
Build in Docker|x|x|N/A|
Gatling tests| |x (for basic apps)
Deploy to Heroku|x|x|x|

TODO/Next/ideas:
* Support NG2+
* Travis CI
* Gatling / Protractor / Cucumber tests
* Github
* AWS / CloudFoundry / Kubernetes deployment
* Build a docker image and publish to a docker registry
* Publish to Artifactory
* Git flow (master/develop/MR/PR/tags)
* Generator tests
* jhipster.github.io doc

Notes:
* I reused templates from generator-jhipster-ci. Kudos to @pascalgrimaud 
* I haven't tested enough Gitlab and Circle generated files

### DIFF
--- a/generators/ci-cd/index.js
+++ b/generators/ci-cd/index.js
@@ -2,6 +2,7 @@
 var util = require('util'),
     chalk = require('chalk'),
     generators = require('yeoman-generator'),
+    prompts = require('./prompts'),
     scriptBase = require('../generator-base');
 
 var PipelineGenerator = generators.Base.extend({});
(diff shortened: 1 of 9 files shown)


generators/ci-cd/index.js
```diff
@@ -0,0 +1,131 @@
...
+            this.abort = false;
+            if(this.clientFramework === 'angular2') {
+                this.log(chalk.red('Angular > 1 is currently not supported, exiting...'));
+                this.abort = true;
+            }
+        }
+    },
+
+    prompting: {
+        askPipelines: function () {
```
deepu105: can you put prompts in its own file like we do in other parts

generators/heroku/index.js
```diff
@@ -277,10 +277,19 @@ module.exports = HerokuGenerator.extend({
             this.template('_bootstrap-heroku.yml', constants.SERVER_MAIN_RES_DIR + '/config/bootstrap-heroku.yml');
             this.template('_application-heroku.yml', constants.SERVER_MAIN_RES_DIR + '/config/application-heroku.yml');
             this.template('_Procfile', 'Procfile');
+            if (this.buildTool === 'gradle') {
+                this.template('_heroku.gradle', 'gradle/heroku.gradle');
```
deepu105: do we have this file?

cbornet: oops

generators/ci-cd/index.js
cbornet: yes, I can.
I thought it was only for generators with a lots of questions

jdubois:
This is totally awesome!!! Notifying @jkutner as this automatically deploys to Heroku, so he should love this - Joe, we need tons of free credits on Heroku to test :-)
I'm merging this as soon as the build passes.

jkutner:
Wow, this is great. Did the `-Dheroku.appName=sushi` option not work for Gradle? I think that would eliminate the need for the `heroku.gradle` file (although, I somewhat like the idea of using that if we ever what to add some of the advanced configuration like `includes` or `processTypes`).

I also wonder if I should switch the Heroku generator to use the `mvnw`/`gradlew` for deployment instead of the CLI (they use the same code under the hood, but I think it would simplify the generator).

Good work @cbornet !

cbornet:
@jkutner I needed the heroku.gradle anyway to be able to add the `apply plugin: 'com.heroku.sdk.heroku-gradle'`. But maybe this could be done by adding a needle and a applyGradlePlugin function.

cbornet:
Also the advantage in dev of the heroku toolbelt is that you have already logged in so you don't need to pass credentials anymore.

jkutner:
@cbornet that's true. with the CLI we avoid passing the Heroku API token around locally, which puts us at less risk of exposing it in the wrong way.

I think the `heroku.gradle` file is probably easier to manage than a needle. I think it's fine that way.

PR #4908, opened 2017-01-19
Fixed cannot logout issue.

tan9:
When logging out, the subscription === undefined, and rose following error:

> Cannot read property 'unsubscribe' of undefined.


jdubois:
Are you using our WebSocket support with Angular 2? I thought it wasn't working at all?? If you have some feedback I'd be happy to read it

gzsombor:
It's working since #4819 :) At least, you could login, and see the activities happening

jdubois:
@gzsombor oh you're working too hard, I can't follow :-)
Can you check the items on https://github.com/jhipster/generator-jhipster/issues/4407 ?

Issue #4884, opened 2017-01-16
Travis tests are flaky


deepu105:
Our travis builds are acting very flaky recently. We had to restart several jobs to get it pass.
Here is an example https://travis-ci.org/jhipster/generator-jhipster/jobs/192443191

Most of the times the failure is from Protractor and sometimes JWT or cassandra builds

Its really annoying and time wasting

cc @pascalgrimaud @jdubois

pascalgrimaud:
Yes I notice that since we migrate from firefox to chrome. I try to analyse the reason but it's really hard because it's totally random.

So sorry @deepu105 but be sure as soon as it's possible, I ll try to reduce these random failures.

mraible:
One of the reasons I switched to Firefox last year was because I saw issues with Chrome and logging out. Sometimes it worked, sometimes it didn't. It always worked with Firefox. However, it looks like there's still issues with Protractor and the latest Firefox. https://github.com/angular/protractor/issues/3750

jdubois:
Most failures I had recently were legitimate, I didn't notice that.
I'll be more careful about it.

deepu105:
@pascalgrimaud since its mostly protractor can we try to catch if there is a failure and retry once again just the protractor step?
I know its not a solution but I'll rather have that then to go and restart builds manually

pascalgrimaud:
@deepu105 : it's sound ok for me. I will give a try

PR #4913, opened 2017-01-20
Travis: protractor tests retry

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/4884

### DIFF
--- a/travis/scripts/05-run.sh
+++ b/travis/scripts/05-run.sh
@@ -23,14 +23,27 @@ launchCurlOrProtractor() {
         exit 1
     fi
 
-    if [ "$PROTRACTOR" == 1 ]; then
-        echo "[$(date)] Connected to application. Start protractor tests."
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Perfect! The build number 3 (mysql in prod + protractor) failed 2 times:
https://travis-ci.org/jhipster/generator-jhipster/jobs/193681707

So we can see the retry.

deepu105:
Damn is it failing even after 3 retries?

jdubois:
There is definitely a problem on this build.....

pascalgrimaud:
In fact, the commit message was wrong, there is only 1 retry, so 2 launchs of protractor.
We can increase the number of retries if you want

deepu105:
but if it requires more than 2 retry then we are doing something wrong I
guess


Thanks & Regards,
Deepu

jdubois:
At least it passes, so definitely a Travis issue...

PR #4919, opened 2017-01-21
Migrate tracker to angular router.

tan9:


generators/client/templates/angular/src/main/webapp/app/admin/_admin.module.ts
```diff
@@ -33,7 +36,7 @@ import {
     <%_ } _%>
     <%_ if (websocket === 'spring-websocket') { _%>
     <%=jhiPrefixCapitalized%>TrackerComponent,
-    TrackerResolve,
+    trackerRoute,
```
sendilkumarn: `trackerRoute` is not needed to import here.

generators/client/templates/angular/src/main/webapp/app/admin/_admin.module.ts
```diff
@@ -1,6 +1,9 @@
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { RouterModule } from '@angular/router';
 import { ParseLinks } from 'ng-jhipster';
+<%_ if (websocket === 'spring-websocket') { _%>
+import { JhiTrackerService } from './../shared/tracker/tracker.service';
```
sendilkumarn: change this to `<%=jhiPrefixCapitalized%>TrackerService`

generators/client/templates/angular/src/main/webapp/app/admin/_admin.module.ts
```diff
@@ -91,8 +94,8 @@ import {
         <%_ } _%>
         LogsService,
         <%_ if (websocket === 'spring-websocket') { _%>
-        TrackerResolve,
-        <%_ } _%>        
+        JhiTrackerService,
```
sendilkumarn: same here as above

generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
```diff
@@ -1,6 +1,6 @@
 import { Injectable, Inject } from '@angular/core';
 import { Observable, Observer, Subscription } from 'rxjs/Rx';
-import { UIRouter } from 'ui-router-ng2';
+import { Router, NavigationEnd } from '@angular/router';
```
sendilkumarn: can you just move this above? such that we will group `@angular` imports together

tan9:
@sendilkumarn , commit revised according to your comments. Thank you.

PR #4932, opened 2017-01-23
Fix pagination

VictorADS:


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.state.ts
```diff
@@ -31,13 +31,12 @@ export class <%= entityAngularJSName %>ResolvePagingParams implements Resolve<an
...
   resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
-
-    return {
-        page: this.paginationUtil.parsePage('1'),
-        sort: 'asc',
-        search: null,
-        predicate: this.paginationUtil.parsePredicate('id,asc'),
-        ascending: this.paginationUtil.parseAscending('id,asc')
+      let page = route.params['page'] ? route.params['page'] : '1';
+      let sort = route.params['sort'] ? route.params['sort'] : 'id,asc';
```
gmarziou: Maybe the default page could be set directly paginationUtil.parsePage, so that we don't have to care about it in calling code.

~~~ts
   parsePage (page = '1') {
        return parseInt(page);
    }
~~~

PR #4933, opened 2017-01-23
Fix protractor tests for ng2 entities

pascalgrimaud:
- fix : an invalid or illegal selector was specified


generators/entity/templates/client/angular/src/test/javascript/e2e/entities/_entity.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('<%= entityClass %> e2e test', () => {
     });
 
     it('should load create <%= entityClass %> dialog', function () {
-        element(by.css('button.create-<%= entityUrl %>"]')).click().then(() => {
+        element(by.css('button.create-<%= entityUrl %>')).click().then(() => {
```
deepu105: Oops my bad and I was wondering why its failing :( may be I was too tired

PR #4937, opened 2017-01-23
ng2: fix gateway microservice

pascalgrimaud:
With this fix, the gateway with ng2 can start normally.
But there is still an issue in configuration page


gmarziou:
LGTM

PR #4947, opened 2017-01-24
Change oauth_client_details column lengths to match Oracle 11g limits

bonhamcm:
The fields in question are:

* additional_information
* autoapprove


jdubois:
Oracle definitely sucks....

PR #4948, opened 2017-01-24
Add transition hook

### LINKED ISSUE #4930: Ng2 router should store destination route
To avoid forgetting it as this is the last issue to solve before being able to remove ui-router completely from package.json

`register-transition-hooks.ts` used to configure ui-router , it had 3 responsibilities on state transitions:

1. store destination state in case to bring user to it later in case she has to login first
2. support for external URLs in ui-router
3. set page title

Only first one is not yet covered and relates to `StateStorageService` that should probably be renamed to `RouteStorageService`.

~~~js
    $transitions.onStart({}, (transition: Transition) => {
        let $storageService = transition.injector().get(StateStorageService);
        $storageService.storeDestinationState(transition.to(), transition.params(), transition.from());
        let principal = transition.injector().get(Principal);
        let auth = transition.injector().get(AuthService);
        if (principal.isIdentityResolved()) {
            auth.authorize();
        }

        // Update the language //FIXME not sure if this is required, its causing some weird error as well
        /*let languageService = transition.injector().get(JhiLanguageService);
        languageService.getCurrent().then(current => {
            languageService.changeLanguage(current);
        });*/
        <%_ } _%>
    });
~~~

VictorADS:
Should fix #4930 
Ps: I didn't find a way to support external URLs


jdubois:
Renamed the issue, we don't have a "transition hooker" in JHipster :-)

generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
```diff
@@ -119,7 +119,14 @@ export class UserMgmtComponent implements OnInit, OnDestroy {
     }
 
     transition () {
-        this.router.navigate(['/user-management', {page: this.page, sort: this.predicate + ',' + (this.reverse ? 'asc' : 'desc')}]);
+        this.router.navigate(['/user-management'],
+            { queryParams:
+                {
+                page: this.page,
```
deepu105: indent is wrong

generators/client/templates/angular/src/main/webapp/app/shared/auth/_state-storage.service.ts
```diff
@@ -24,21 +24,20 @@ export class StateStorageService {
...
+        let destinationInfo = {
+            'destination': {
+                'name': destinationState.name,
+                'data': destinationState.data,
+               // 'parent': destinationState.parent
+            },
+            'params': destinationStateParams,
+            'from': {
+                'name': fromState.name,
+                 // 'data': fromState.data,
```
deepu105: if there is something TODO here acn you add it as TODO comment?

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-detail.component.html
```diff
@@ -68,9 +68,13 @@
             </span>
             <%_ } else { _%>
             <%_ if (dto == 'no') { _%>
-            <a [routerLink]="['<%= otherEntityStateName %>-detail', <%= entityInstance + '.' + relationshipFieldName + '?.id' %>]">{{<%= entityInstance + "." + relationshipFieldName + "?." + otherEntityField %>}}</a>
+            <div *ngIf="<%= entityInstance + '.' + relationshipFieldName %>">
```
deepu105: do the same for user-management as well

VictorADS: We don't need it in user-management as the login is required so there's no way it could be null

deepu105:
can you add the todo comments for commented items and look at test failure

deepu105:
Why did you close this

VictorADS:
My bad, could you squash the commits when merging ?

deepu105:
Ok

generators/client/templates/angular/src/main/webapp/app/shared/auth/_state-storage.service.ts
jdubois: As we don't use it, I would remove the comments

PR #4950, opened 2017-01-24
Use a property for setting the JHipster Registry password

PierreBesson:
This `central-server-config/` dir was mistakenly left there in my Consul, so I'm removing it because it is not used.

There needs to be a new release of the JHipster lib for this to work, see : https://github.com/jhipster/jhipster/pull/5


jdubois:
@PierreBesson JHipster lib 0.0.10 release is under way, it should be replicated on Maven Central in less than 2 hours

generators/server/templates/src/main/docker/_app.yml
```diff
@@ -55,7 +55,7 @@ services:
             - SPRING_CLOUD_CONSUL_PORT=8500
                   <%_ } _%>
                   <%_ if (serviceDiscoveryType == 'eureka') { _%>
-            - SPRING_CLOUD_CONFIG_URI=http://admin:admin@registry:8761/config
+            - SPRING_CLOUD_CONFIG_URI=http://admin:$${jhipster.registry.password}@registry:8761/config
```
pascalgrimaud: Double $$ here.
Do u need money? :P

PierreBesson: This is to bypass docker-compose env var substitution : https://docs.docker.com/compose/compose-file/#/variable-substitution (read carefully as it is not straightforward 😄 )
I actually spent quite a bit of time to understand why it didn't work with just one $.

PR #4953, opened 2017-01-24
[ng2] Fix social buttons on login modal

nkolosnjaji:
Finally I found it! ... and it is so easy fix :)

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
@@ -49,9 +49,9 @@
         <%_ if (enableSocialSignIn) { _%>
         <div class="col-md-6">
             <br/>
-            <jh-social provider="google"></jh-social>
-            <jh-social provider="facebook"></jh-social>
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
```diff
@@ -49,9 +49,9 @@ <h1 jhiTranslate="login.title">Sign in</h1>
         <%_ if (enableSocialSignIn) { _%>
         <div class="col-md-6">
             <br/>
-            <jh-social provider="google"></jh-social>
-            <jh-social provider="facebook"></jh-social>
-            <jh-social provider="twitter"></jh-social>
+            <jhi-social provider="google"></jhi-social>
```
ruddell: `jhi` should be `<%= jhiPrefix %>` to match [the component selector](https://github.com/nkolosnjaji/generator-jhipster/blob/0c1820a1de30af6a721c70ea02b31b174322fab3/generators/client/templates/angular/src/main/webapp/app/shared/social/_social.component.ts#L7)

deepu105:
Please squash and merge once tests are done

PR #4956, opened 2017-01-24
shorten foreign key names to fit oracle length limit of 30

### LINKED ISSUE #4954: Generated application can not start properly with OAuth and Oracle Database

Generated application can not start properly with OAuth and Oracle Database because foreign key called "fk_oauth_client_token_user_name" defined in 00000000000000_initial_schema.xml is too long (31 chars) for an Oracle object name. Oracle object names can be 30 chars length at most.

Generated application can't be started.

##### **Reproduce the error**

Just generate an aplication with OAuth and Oracle Database and simply try to run it.

##### **Related issues**

No.

##### **Suggest a Fix**

Change the foreign key name from "fk_oauth_client_token_**user_name**" to "fk_oauth_client_token_**username**" which is one char less.

##### **JHipster Version(s)**

generator-jhipster@3.12.2

{
  "generator-jhipster": {
    "jhipsterVersion": "3.12.2",
    "baseName": "GCPlatform",
    "packageName": "com.gcoban.platform",
    "packageFolder": "com/gcoban/platform",
    "serverPort": "8080",
    "hibernateCache": "ehcache",
    "databaseType": "sql",
    "devDatabaseType": "oracle",
    "prodDatabaseType": "oracle",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "useSass": true,
    "testFrameworks": [
      "gatling"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "tr"
    ]
  }
}

No entity defined yet.

Chrome on MacOS Sierra. The problem is irrelevant with Browser or OS.

ruddell:
Fix #4954

### DIFF
--- a/generators/server/templates/src/main/resources/config/liquibase/changelog/_initial_schema.xml
+++ b/generators/server/templates/src/main/resources/config/liquibase/changelog/_initial_schema.xml
@@ -268,7 +268,7 @@
 
         <addForeignKeyConstraint baseColumnNames="user_name"
             baseTableName="oauth_client_token"
-            constraintName="fk_oauth_client_token_user_name"
+            constraintName="fk_oauth_client_token_username"
             referencedColumnNames="login"
             referencedTableName="jhi_user"/>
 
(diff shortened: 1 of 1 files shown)


jdubois:
I don't really like putting `username` instead of `user_name` as this won't be coherent with the rest of what we generate (we generally truncate the table, so that would be `user_nam`, and it is even be more horrible!!).
Then only very few people use OAuth2, and I don't think anybody cares about the foreign key names, so let's do this...

Issue #4923, opened 2017-01-22
Inconsistent route paths for composed entity names


gmarziou:
The route paths generated for an entity with camel cased name like TournamentType are not consistent:
- the normal routes (list and detail) use dash case: tournament-type
- the popup routes use camel case: tournamentType 

~~~
   path: 'tournament-type',
   path: 'tournament-type/:id',
   path: 'tournamentType-new',
   path: 'tournamentType-edit/:id',
   path: 'tournamentType-delete/:id',
~~~

The addEntityToMenu() function which is used to add entries to the navbar use dash case.

I think we should be consistent.
If we use dashcase then we may have problems with some entity names containing New, Edit or Delete like what we have with user management and :login.

Maybe we should define a better route scheme:

~~~
   path: 'tournament-type/list',
   path: 'tournament-type/detail/:id',
   path: 'tournament-type/new',
   path: 'tournament-type/edit/:id',
   path: 'tournament-type/delete/:id',
~~~

Maybe we can find some good practices.

deepu105:
I think it might be an overlooked mistake. I guess its better to have dash
case for urls as it is better for SEO and is common practice

Thanks & regards,
Deepu

jdubois:
Yes, definitely dash!

gmarziou:
OK, but what do you think of my proposal to use dash AND to separate the entity name from the "action"? @sendilkumarn any comment?

sendilkumarn:
@gmarziou yeah dashed definitely. I guess it is my mistake.

deepu105:
@gmarziou im gonn a change this to your proposal as im fixing tests

gmarziou:
Thanks, I will review and test yourPR

deepu105:
I'm modifying it a bit to follow what we had for ng1 its more standard I guess but we can ofcourse change if there is a better way

gmarziou:
I guess that our ng1 conventions were highly driven by the notion of state, ng2 routes may lead to slightly different conventions. In partiicular I think we could have a componentless route to be able to manage authorities globally for the entity or to add a resolve property. This would be the equivalent of ui-router abstract state.

You can do it with my suggestion to use anaction.

```
tournament-type   Componentless
tournament-type/<action>
```

deepu105:
i'm talking about the ng1 urls, which IMO follows REST url conventions closeley

```
   path: 'tournament-type',
   path: 'tournament-type/:id',
   path: 'tournament-type/new',
[...]

PR #4959, opened 2017-01-25
fixes ID links for entity and user-management components

ruddell:
When an entity contains more than one word, the link on the ID in the entity table was broken.  For example: `logEntry` should be `log-entry`

User management ID also pointed to the wrong place


gmarziou:
I thought this was already fixed by Deepu for #4923

ruddell:
The "View Entity" button worked correctly, but the link on the entity and user IDs still used the old format

deepu105:
Ah ok I missed them

PR #4974, opened 2017-01-26
Fix Backslashes in validation patterns are not escaped in 'translateValues'

### LINKED ISSUE #4966: Backslashes in validation patterns are not escaped in `translateValues`

Backslashes in validation patterns are not escaped in the `translateValues` causing an unrecoverable client-side error.

Add a valid pattern to a field, here a "non blank" pattern.

##### **Reproduce the error**

1. Generate an entity X with the validation pattern `^.*[\\S]+.*$` (see attachment below)
2. Open the `Create a new X` dialog
3. A JavaScript error is raised and all translated texts are blank 

##### **Related issues**

n/a

##### **Suggest a Fix**

The associated input field has a generated `translateValues="{ pattern: '^.*[\S]+.*$' }"` which is wrong: it should be `translateValues="{ pattern: '^.*[\\S]+.*$' }"` with 2 backslashes.

##### **JHipster Version(s)**

Master branch, commit 16e2184748a39d300bae2eaeb57856e4650d0370

[yo-rc.json](https://github.com/jhipster/generator-jhipster/files/730281/yo-rc.json.zip)

[Platform.json](https://github.com/jhipster/generator-jhipster/files/730286/Platform.json.zip)

Chrome

ddidier:
Fix #4966

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -144,7 +144,7 @@
           <%_ } _%>
           <%_ if (fields[idx].fieldValidateRules.indexOf('pattern') != -1) { _%>
           <small class="form-text text-danger"
-           [hidden]="!editForm.controls.<%= fieldName %>?.errors?.pattern" jhiTranslate="entity.validation.pattern" translateValues="{ pattern: '<%= fields[idx].fieldValidateRulesPattern.replace(/"/g, '&#34;') %>' }">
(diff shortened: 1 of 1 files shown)


jdubois:
Oh thanks @ddidier !!
I'm merging this now, as I was just testing the issue, so it's going to be easy for me to validate

jdubois:
@ddidier not sure it works.... Now I can write anything in my field, like "toto". Can you check?

jdubois:
@ddidier in fact `^.*[\\S]+.*$` forces me to write an `S`, and I think what you meant is that no whitespace should be validated.

ddidier:
@jdubois not sure I understand. My goal is to force the field to be NOT blank, so yes, `toto` is valid.
The double backslash `^.*[\\S]+.*$` is only for the translation parameter, not the validation itself which remains single `^.*[\S]+.*$`. Thank you.

jdubois:
Oh OK that should be fine, thanks @ddidier

PR #4987, opened 2017-01-27
Fix skip server flag issues

VictorADS:
Bad imports are removed

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -99,7 +99,6 @@ const files = {
                 'blocks/config/_prod.config.ts',
                 'blocks/config/_uib-pagination.config.ts',
                 //interceptors
-                'blocks/interceptor/_auth-expired.interceptor.ts',
                 'blocks/interceptor/_errorhandler.interceptor.ts',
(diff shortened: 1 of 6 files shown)


generators/client/files-angular.js
```diff
@@ -111,6 +110,13 @@ const files = {
             templates: [
                 'blocks/interceptor/_auth.interceptor.ts'
             ]
+        },
+        {
+            condition: generator => !generator.skipServer,
```
deepu105: Actually I wouldnt worry much about this as if you are skipping server then you need to make sure you are providing endpoints for all the REST apis else everything will fail so its not really important to validate this. but since you have done it its ok

deepu105: But if there are mores stuff that would require such check lets not bother and lets just document that if you skip server its your responsibility to provide equivalent API

VictorADS: So what do you want me to do ? With these changes the app compiles with no error and run as expected

deepu105: Its fine as its only for  a single file

PR #4990, opened 2017-01-27
footer css property in global.css

btelintelo:
In the footer component, there is

`<div class="footer">`

So in global css we need .footer rather than footer


gmarziou:
Thanks, it should be done also in scss and angular1.

PR #4992, opened 2017-01-27
Each entity has now its own angular module

### LINKED ISSUE #4918: Each entity should have its own module
While working on fixing entity pagination with ng router, I found that entity.module.ts was 300 lines long for only 11 entities and also it was difficult to expose `<Entity>ResolvePagingParams` as a provider in this file without putting lot of logic in generator.

This is why I propose that each entity should have its own module.

gmarziou:
Close #4918


generators/client/files-angular.js
```diff
@@ -123,9 +123,6 @@ const files = {
         {
             path: ANGULAR_DIR,
             templates: [
-                // entities
-                'entities/_index.ts',
-                'entities/_entity.module.ts',
```
deepu105: I think we should still keep the entity module and import every other entity here so that the app.module remains small and cleaner

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -9,8 +9,8 @@ import { Ng2Webstorage } from 'ng2-webstorage';
 
 import { <%=angular2AppName%>SharedModule, UserRouteAccessService } from './shared';
 import { <%=angular2AppName%>AdminModule } from './admin/admin.module';
-import { <%=angular2AppName%>EntityModule } from './entities/entity.module';
 import { <%=angular2AppName%>AccountModule } from './account/account.module';
+/* jhipster-needle-add-entity-module-import - JHipster will add entity modules here */
```
deepu105: this will make this file huge

generators/client/files-angular.js
deepu105: It will also help to add stuff common to all entities in one place

generators/entity/files.js
```diff
@@ -353,7 +361,7 @@ function writeFiles() {
         },
 
         writeClientFiles: function () {
-            if (this.skipClient) return;
+            if (this.skipClient) return;
```
deepu105: extra space

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -16,8 +16,7 @@ import { EventManager, AlertService<% if (enableTranslation) { %>, JhiLanguageSe
 import { <%= entityClass %> } from './<%= entityFileName %>.model';
 import { <%= entityClass %>PopupService } from './<%= entityFileName %>-popup.service';
 import { <%= entityClass %>Service } from './<%= entityFileName %>.service';
-<%- include('model-class-import-template.ejs'); -%>
```
deepu105: didnt we need this?

generators/entity/templates/client/angular/src/main/webapp/app/entities/_index.ts
```diff
@@ -0,0 +1,11 @@
...
+export * from './<%= entityFileName %>-popup.service';
+export * from './<%= entityFileName %>.service';
+export * from './<%= entityFileName %>-dialog.component';
+export * from './<%= entityFileName %>-delete-dialog.component';
+export * from './<%= entityFileName %>-detail.component';
+export * from './<%= entityFileName %>.component';
+export * from './<%= entityFileName %>.state';
+export * from './<%= entityFileName %>.module';
+
+
```
deepu105: extra line

generators/generator-base.js
```diff
@@ -162,112 +162,30 @@ Generator.prototype.addEntityToModule = function (entityInstance, entityClass, e
         if (clientFramework === 'angular1') {
             return;
         }
-        var entityPath = CLIENT_MAIN_SRC_DIR + 'app/entities/entity.module.ts';
```
deepu105: nice :clap:

generators/client/files-angular.js
gmarziou: Fully agree with entity.module.ts, not so sure about index.ts because it's going to be huge.

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
gmarziou: OK I will do it on Saturday or Sunday.

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
gmarziou: I thought it is simpler to import always the model and the service like this:

import { TournamentType, TournamentTypeService } from '../tournament-type/';
import { RankingList, RankingListService } from '../ranking-list/';

generators/client/files-angular.js
deepu105: we can have an index.ts for each entity in its own folder and a parent index.ts here which exports the other index.ts files if required

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
deepu105: ok

deepu105:
@gmarziou let me know if you need any help on this

gmarziou:
Thanks, I'm working on it right now.

gmarziou:
Ready @deepu105

gmarziou:
I'm fixing the bugs detected by Travis and which are related to User relationships.

jdubois:
The Travis built fails, looks like this doesn't work with Protractor (I already re-started it, it failed again)

deepu105:
Its important to get protractor working with this

gmarziou:
I'm not sure it's related to my changes, when I run `yarn run e2e` on my PC, it fails also and I can see this error in brower's page:

> Access-Control-Allow-Origin: *
>Error : no method specified  in uri /

gmarziou:
It fails the same on master, has it worked recently?

deepu105:
The last I checked it worked. Let me check again with current master

Thanks & regards,
Deepu

deepu105:
master builds for protractor are passing fine. there was issues if you use a custom prefix I have fixed it otherwise its fine

deepu105:
can you rebase and try again

gmarziou:
I rebased and now I have an undefined in imports.
I may have found why, in sample app we have a have a one-to-many relationship between `BankAccount` and `Operation`, my (naive) implementation resulted in a circular dependency between both modules using their `imports` property.

PR #4993, opened 2017-01-27
Prepare for AOT

gmarziou:
When using angular-cli on a JHipster project, the AOT compilation fails on `http.provider.ts` because of `useFactory` property  using a lambda rather than a named function.

See https://medium.com/@isaacplmann/making-your-angular-2-library-statically-analyzable-for-aot-e1c6f3ebedd5#.hylx41wiy


jdubois:
Great work !
Turns out AOT isn't as magical as in the ads 😀

PR #5012, opened 2017-01-29
Add separator for JWT token in Angular 2

### LINKED ISSUE #5003: JWT doesn't work with Swagger "try it out"
This comes from our JHipster 4 tests, see https://github.com/jhipster/generator-jhipster/issues/4301#issuecomment-275880677 
With JWT authentication, testing the API with Swagger fails:

- Select an API to test
- click on "try it out", there will be an authentication error
- It is not authenticated as the JWT token isn't sent

cicoub13:
fix #5003

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/swagger-ui/_index.html
+++ b/generators/client/templates/angularjs/src/main/webapp/swagger-ui/_index.html
@@ -99,7 +99,7 @@
                     window.swaggerUi.api.clientAuthorizations.add("key", apiKeyAuth);
                 <%_ } _%>
                 <%_ if (authenticationType === 'jwt' || authenticationType === 'uaa') { _%>
-                    var authToken = JSON.parse(localStorage.getItem("jhi-authenticationToken") || sessionStorage.getItem("jhi-authenticationToken"));
+                    var authToken = JSON.parse(localStorage.getItem("jhi-authenticationtoken") || sessionStorage.getItem("jhi-authenticationtoken"));
                     var apiKeyAuth = new SwaggerClient.ApiKeyAuthorization("Authorization", "Bearer " + authToken, "header");
                     window.swaggerUi.api.clientAuthorizations.add("bearer", apiKeyAuth);
                 <%_ } _%>
--- a/generators/client/templates/angular/src/main/webapp/app/_app.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_app.module.ts
@@ -34,7 +34,7 @@ import {
     imports: [
         BrowserModule,
         LayoutRoutingModule,
-        Ng2Webstorage.forRoot({ prefix: 'jhi'}),
+        Ng2Webstorage.forRoot({ prefix: 'jhi', separator: '-'}),
         <%=angular2AppName%>SharedModule,
         <%=angular2AppName%>AdminModule,
         <%=angular2AppName%>EntityModule,


jdubois:
I was just beginning to have a look at this!! This community is incredible :-)

cicoub13:
I saw your tweet, ran through the bugs. I had already worked on this topic on version 3, so I try to help 😄

jdubois:
Unfortunately @cicoub13 it looks like this is not enough, I now have 403 errors with BrowserSync (need to check why...), and still 404 when I use port 8080.

jdubois:
Sorry I cleaned up my browser cache, and just got:

```
{
  "timestamp": "2017-01-29T16:15:49.404+0000",
  "status": 500,
  "error": "Internal Server Error",
  "exception": "io.jsonwebtoken.MalformedJwtException",
  "message": "JWT strings must contain exactly 2 period characters. Found: 0",
  "path": "/api/account"
}
```

cicoub13:
OK, i check this

jdubois:
Oh, I get it, this because we don't have the "sessionStorage" anymore, now we switched to Angular 2.

cicoub13:
In fact, jHipster use **ng2-webstorage** to store token. And token isn't "jhi-authenticationToken" but "jhi-authenticationtoken" (caused by line 20 in https://github.com/PillowPillow/ng2-webstorage/blob/master/lib/helpers/keyStorage.ts).

I can correct this in index.html of swagger (only for angular2) but it also impact uaa authentication and i can't test.

jdubois:
Yes can you correct this in the index.html of Swagger? I just had a look, now I understand the issue, quite tricky!!
Don't worry for UAA - it's still in BETA, and the focus in JWT as it's the stable option. It's also already broken, so this will be a specific work anyway.

jdubois:
oh it's "jhi-authenticationtoken" now??? It probably also breaks Gatling... I'll have a look

jdubois:
OK that's fine with Gatling...

cicoub13:
This should be OK with 8080 and 9060 port. Always a 403 with 9000 (host returned by /api-docs is "localhost:9060" and cross-origin HTTP requests is forbidden)

jdubois:
Thanks - merged it but I still get: `"message": "JWT strings must contain exactly 2 period characters. Found: 0"` when I'm using the API, so I need more testing

jdubois:
@cicoub13 sorry I still have the `2 period characters` error, so I reopened the issue

generators/client/templates/angularjs/src/main/webapp/swagger-ui/_index.html
```diff
@@ -99,7 +99,7 @@
                     window.swaggerUi.api.clientAuthorizations.add("key", apiKeyAuth);
                 <%_ } _%>
                 <%_ if (authenticationType === 'jwt' || authenticationType === 'uaa') { _%>
-                    var authToken = JSON.parse(localStorage.getItem("jhi-authenticationToken") || sessionStorage.getItem("jhi-authenticationToken"));
+                    var authToken = JSON.parse(localStorage.getItem("jhi-authenticationtoken") || sessionStorage.getItem("jhi-authenticationtoken"));
```
pascalgrimaud: @jdubois @cicoub13 :
- I think this file is correct and should not be changed -> angularjs
- you should edit the file inside angular folder

jdubois: oh @pascalgrimaud you are correct, this is the wrong file!!

cicoub13: Sorry, I should not code when I'm tired. Do I open a new merge request ?

jdubois: @cicoub13 yes if you have the time -> it's your code, you should be the author of the solution :-)

PR #5017, opened 2017-01-29
Correction of name of JWT Token in Swagger

cicoub13:
Correction of name of JWT Token in Swagger (only for Angular2)
caused by ng2-webstorage (lowercase)


jdubois:
Yes it works :-)

pascalgrimaud:
Sorry but I think we need to change back the original angularjs, otherwise, jwt wont work for ng1 -> but I am not sure, as I didnt test your modifications (no computer)

ruddell:
@pascalgrimaud It's taken care of https://github.com/jhipster/generator-jhipster/commit/6476e9817e331d4d8b31328c295506a4e656d6d5

Issue #4985, opened 2017-01-27
Travis: improvement for NG2


pascalgrimaud:
##### **Overview of the issue**

As NG1 is frozen, our Travis builds should cover NG2

The project from Hipster Labs https://github.com/hipster-labs/jhipster-travis-build will launch a nightly build : https://travis-ci.org/hipster-labs/jhipster-travis-build/ 

##### **Related issues**

Following discussion here: https://github.com/jhipster/generator-jhipster/issues/4301

##### **Suggest a Fix**

Some improvements which can be done:
- [ ] there should not be duplicate builds between main generator-jhipster and hipster-labs -> don't hesitate to propose a full list
- [X] missing build for NG2: with prod profile
- [X] missing build for NG2: elasticsearch
- [ ] protractor : default behavior like now. When merged into master, protractor for all app!
- [X] logs: I would like to add `-Dspring.jpa.show-sql=false` when launching tests to reduce the logs. What do you think?
- [X] Sonar analysis: switch to NG2 ?

deepu105:
:+1: for reducing logs as we are only concerned about error logs there. we could also try to specify log levels so we dont see all the info and debug logs I guess

jdubois:
yes error logs are a huge problem for me, as most of the time I read them on my phone...

pascalgrimaud:
@jdubois : it's funny because I do the same, scroll, scroll, scroll and boring about Hibernate logs!

pascalgrimaud:
About Sonar, I don't see a Typescript profile: https://sonarqube.com/profiles
Do you confirm me you want to switch to NG2 for Sonar analysis ?

jdubois:
Yes, everybody is going to switch to NG2 anyway... This is like the rest of the build, in fact.
Probably Sonar will add one soon (we have it at work, so I guess SonarQube will update their config soon)

pascalgrimaud:
The 1st nightly build : https://travis-ci.org/hipster-labs/jhipster-travis-build/builds/195992165
The Cron Jobs from Travis work well !!

jdubois:
I'm removing angular 2 and v4 tags, as this is not directly related. This is not a blocker issue for releasing JHipster 4, anyway.

pascalgrimaud:
Yes, total agree

deepu105:
Im getting a protractor failure in travis for the ng2-default app but when
I run it locally it works fine I have no idea whats going on there. Note
[...]

PR #5020, opened 2017-01-29
Travis: improvements for ng2 logs

pascalgrimaud:
Close https://github.com/jhipster/generator-jhipster/issues/4985
I won't do protractor for all apps immediatly, I have other ideas for the hipster-labs project that will cover more things

### DIFF
--- a/travis/scripts/04-tests.sh
+++ b/travis/scripts/04-tests.sh
@@ -23,12 +23,12 @@ fi
 cd "$HOME"/app
 if [ -f "mvnw" ]; then
     ./mvnw test \
-        -Dlogging.level.io.github.jhipster.sample=INFO \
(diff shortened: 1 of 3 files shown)


travis/scripts/04-tests.sh
```diff
@@ -22,9 +22,13 @@ fi
 #-------------------------------------------------------------------------------
 cd "$HOME"/app
 if [ -f "mvnw" ]; then
-    ./mvnw test -Dspring.jpa.show-sql=false
+    ./mvnw test \
+        -Dlogging.level.io.github.jhipster.sample=INFO \
```
deepu105: I think we could even set to ERROR as thats what we would be interested in wont we?

pascalgrimaud: I'm ok, let reduce more!

Issue #5019, opened 2017-01-29
ng2 app not displayed when using specific jhi-prefix with entities


pascalgrimaud:
##### **Overview of the issue**

The current [build](https://github.com/jhipster/generator-jhipster/blob/master/travis/samples/app-ng2-default/.yo-rc.json) which use a specific jhi-prefix failed:
https://travis-ci.org/jhipster/generator-jhipster/jobs/196399903

See in the `.yo-rc.json` file:  `"jhiPrefix": "custom"` : https://github.com/jhipster/generator-jhipster/blob/master/travis/samples/app-ng2-default/.yo-rc.json#L25

Related to this dicussion https://github.com/jhipster/generator-jhipster/issues/4985#issuecomment-275939086
Ping @deepu105 

##### **Reproduce the error**

Simply go into http://localhost:8080
Nothing is displayed

![jhi-prefix](https://cloud.githubusercontent.com/assets/9156882/22408071/a9dd82dc-e672-11e6-9a90-9109540c141a.png)

##### **Suggest a Fix**

In some entity.html files

Instead of:
```html
<tr customSort [(predicate)]="predicate" [(ascending)]="reverse" [callback]="transition.bind(this)">
<th customSortBy="id"><span jhiTranslate="global.field.id">ID</span> <span class="fa fa-sort"></span></th>
<th customSortBy="label"><span jhiTranslate="travisNg2App.label.label">Label</span> <span class="fa fa-sort"></span></th>
```

I use:
```html
<tr jhiSort [(predicate)]="predicate" [(ascending)]="reverse" [callback]="transition.bind(this)">
<th jhiSortBy="id"><span jhiTranslate="global.field.id">ID</span> <span class="fa fa-sort"></span></th>
<th jhiSortBy="label"><span jhiTranslate="travisNg2App.label.label">Label</span> <span class="fa fa-sort"></span></th>
```

And it works.

So probably something to do here with jhiPrefix:
https://github.com/jhipster/generator-jhipster/blob/master/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.component.html#L39-L49

##### **JHipster Version(s)**

```
travis-ng-2@0.0.0 /home/pgrimaud/tmp/18-toto
└── generator-jhipster@3.12.2  -> /home/pgrimaud/projects/jhipster/generator-jhipster

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

```yaml
{
  "generator-jhipster": {
[...]

PR #5021, opened 2017-01-29
Fix the SortDirective by using jhiSort from ng-jhipster

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5019


ruddell:
user-management.component.html has `<%=jhiPrefix%>SortBy` as well

pascalgrimaud:
@ruddell : mhhhh I don't see it lol !
https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html

ruddell:
You're right haha, my master is a few hours old

deepu105:
Oh ok I checked again and see that I indeed fixed it but I commited only user management part and the entity part is still on my local thats why it works locally for me. damn i'm tired :) thanks @pascalgrimaud

PR #5023, opened 2017-01-29
Module per entity

### LINKED ISSUE #4918: Each entity should have its own module

gmarziou:
I rebased so here is a new PR.
It fixes all known issues about entity modules and works with protractor.

Close #4918


jdubois:
This is huge!!!!!!

gmarziou:
Well, on one hand it adds some structure to our entities but on the other hand lazy loading entity modules separately would be a difficult task because of the dependencies between entities when they have relationships.

So I believe that lazy loading modules would be more useful at feature level like our admin module. I have few ideas now about what could be done, I'll add them to after release issue.

jdubois:
Yes, for me lazy loading entity modules is more a business decision - you need to know what your users are going to do, or what components are going to use. So we should help people doing this easily, but we can't generate that for them.

generators/client/templates/angular/src/main/webapp/app/admin/_index.ts
```diff
@@ -41,3 +41,4 @@ export * from './user-management/user.model';
 export * from './user-management/user-modal.service';
 <%_ } _%>
 export * from './admin.state';
+export * from './admin.module';
```
deepu105: @gmarziou was this really required? I removed module from exports as it was causing circular dependencies as most of the above components will be imported into the module as well.

generators/entity/index.js
```diff
@@ -571,6 +571,15 @@ module.exports = EntityGenerator.extend({
                 if (_.isUndefined(relationship.otherEntityStateName)) {
                     relationship.otherEntityStateName = _.trim(_.kebabCase(relationship.otherEntityName), '-') + this.entityAngularJSSuffix;
                 }
+                if (_.isUndefined(relationship.otherEntityModuleName)) {
+                    if (relationship.otherEntityNameCapitalized !== 'User') {
+                        relationship.otherEntityModuleName = this.angular2AppName + relationship.otherEntityNameCapitalized + 'Module';
+                        relationship.otherEntityModulePath = 'entities/' + relationship.otherEntityStateName;
+                    } else {
+                        relationship.otherEntityModuleName = this.angular2AppName + 'AdminModule';
```
deepu105: what is this about?

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.module.ts
```diff
@@ -0,0 +1,68 @@
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { RouterModule } from '@angular/router';
+
+import { InfiniteScrollModule } from 'angular2-infinite-scroll';
```
deepu105: shouldnt this be imported only when entity has infinite scroll enabled?

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.module.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import { <%= angular2AppName %>SharedModule } from '../../shared';
+<%_ for (var rel of differentRelationships) { 
+       if (rel.otherEntityNameCapitalized === 'User') { _%> 
+import { <%= angular2AppName + 'AdminModule' %> } from '../../admin/admin.module';
```
deepu105: What component are we using from AdminModule? I would like to avoid such dependencies when possible. If something is shared it should be moved to shared module else this will be a recipe for disaster

generators/entity/templates/client/angular/src/main/webapp/app/entities/_index.ts
```diff
@@ -0,0 +1,9 @@
+export * from './<%= entityFileName %>.model';
+export * from './<%= entityFileName %>-popup.service';
+export * from './<%= entityFileName %>.service';
+export * from './<%= entityFileName %>-dialog.component';
+export * from './<%= entityFileName %>-delete-dialog.component';
+export * from './<%= entityFileName %>-detail.component';
+export * from './<%= entityFileName %>.component';
+export * from './<%= entityFileName %>.state';
+export * from './<%= entityFileName %>.module';
```
deepu105: same here can we avoid exporting the module?

deepu105:
@gmarziou  I know its merged and I didnt get to review it soon enough but can you take a look at the comments I left. I'm specifically not comfortable with importing admin module into every entity
cc @jdubois

jdubois:
Yes I merged it as this is kind of becoming urgent, but this should be improved

deepu105:
@jdubois no worries I understand

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.module.ts
gmarziou: Yes I fully agree with you. This is required for entities that have relationships with User, I think we should consider extracting User and UserService from AdminModule, maybe User should be considered as an entity

gmarziou: anyway before isolating User in its own module I can probably simplify the template

deepu105:
@gmarziou I have moved user-service into shared

PR #5025, opened 2017-01-30
fix uaa 404 on port 9000 and user management listings for uaa

ruddell:
The user management api endpoint was missing the UAA condition and the webpack proxy was missing the UAA endpoint.  I also simplified the condition in user.service.ts


jdubois:
Awesome !!

xetys:
Thanks a lot @ruddel 

I am currently on my long awaited vacation in the mountains while preparing the workshop at microxchng with @atomfrede, so it's hard to keep track on these issues. I hope I will manage to take a look at all the uaa parts before release!

Issue #5008, opened 2017-01-29
Many-to-many relationships links do not work


jdubois:
This comes from our JHipster 4 tests at https://github.com/jhipster/generator-jhipster/issues/4301#issuecomment-275901699

When a many-to-many relationship is created, for example with this JDL:

```
entity Alice {
	aaa String,
    bbb Integer,
    ccc Long,
    ddd BigDecimal,
    eee Float,
    fff Double,
    ggg Boolean,
    hhh LocalDate,
    iii ZonedDateTime
}

entity Bob {
   jjj String required
}

relationship ManyToOne {
	Bob{alice(aaa)} to Alice{bob}
}

relationship ManyToMany {
    Alice{manybob} to Bob{manyalice}
}
```

When the many-to-many relationship is created:

- It appears in the link with commas, and there is one comma at the end that should be removed
- The links cannot be clicked

See the attached screenshot:

<img width="240" alt="screenshot 2017-01-29 14 53 31" src="https://cloud.githubusercontent.com/assets/316835/22404525/bdd08e32-e632-11e6-8774-72c3d0040a2e.png">

jdubois:
Closing as the PR is merged

PR #5028, opened 2017-01-30
Use correct route and last variable of ngFor

VictorADS:
Fix for #5008


jdubois:
Tested successfully, I'm merging :-)

PR #5032, opened 2017-01-30
ng2: fix ribbon by using hidden

pascalgrimaud:
@deepu105 @gmarziou @jdubois : are you ok with that ?
Related to https://github.com/jhipster/generator-jhipster/commit/c6dec85235ac0531b091af703e5bc24c6abdb213#commitcomment-20669303

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
@@ -4,7 +4,7 @@ import { ProfileInfo } from './profile-info.model';
 
 @Component({
     selector: '<%=jhiPrefix%>-page-ribbon',
-    template: `<div class="ribbon {{cssClass}}"><a href=""<% if (enableTranslation) { %> jhiTranslate="global.ribbon.{{ribbonEnv}}"<% } %>>{{ribbonEnv}}</a></div>`,
+    template: `<div class="ribbon" *ngIf="!inProduction"><a href=""<% if (enableTranslation) { %> jhiTranslate="global.ribbon.{{ribbonEnv}}"<% } %>>{{ribbonEnv}}</a></div>`,
     styleUrls: [
         <%_ if (useSass) { _%>
         'page-ribbon.scss'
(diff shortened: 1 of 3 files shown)


generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
```diff
@@ -4,7 +4,7 @@ import { ProfileInfo } from './profile-info.model';
 
 @Component({
     selector: '<%=jhiPrefix%>-page-ribbon',
-    template: `<div class="ribbon {{cssClass}}"><a href=""<% if (enableTranslation) { %> jhiTranslate="global.ribbon.{{ribbonEnv}}"<% } %>>{{ribbonEnv}}</a></div>`,
+    template: `<div class="ribbon" [hidden]="!ribbonEnv"><a href=""<% if (enableTranslation) { %> jhiTranslate="global.ribbon.{{ribbonEnv}}"<% } %>>{{ribbonEnv}}</a></div>`,
```
gmarziou: Maybe you could use `*ngIf` rather than `[hidden]` because it removes completely the ribbon from the DOM rather than just hiding it.

`<div class="ribbon" *ngIf="ribbonEnv != null">`

pascalgrimaud: you're right @gmarziou !
done and thanks you

gmarziou:
LGTM thanks

jdubois:
Yes but does it hide the banner in prod?

pascalgrimaud:
Yes, same behavior than ng1
- in dev : red banner with dev
- in prod : nothing

jdubois:
Yes but you don't use the 'isInProduction" method

Le 31 janv. 2017 6:51 AM, "Pascal Grimaud" <notifications@github.com> a
écrit :

jdubois:
The use case is: always show the banner, unless you are in prod mode.
People use test, pre-prod profiles, and need the banner.

Le 31 janv. 2017 7:08 AM, "Julien Dubois" <julien.dubois@gmail.com> a
écrit :

pascalgrimaud:
Oh yes you're right
I think we should make a test on `inProduction`, let me made this change

pascalgrimaud:
I made the change, and squash my modifications

jdubois:
LGTM

Le 31 janv. 2017 7:41 AM, "Pascal Grimaud" <notifications@github.com> a
écrit :

deepu105:
So in short i guess using ribbonEnv variable will be more flexible so that user has absolute control on this else there is no purpose of doing this as a rest endpoint as we could just check the constant for prod env and apply ribbon otherwise. So i would prefer the previous version

jdubois:
OK with you Deepu... Then the "isInProduction" method should also be
removed.

Le 31 janv. 2017 8:37 AM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

deepu105:
I think we can have it as its a useful info and was added for some reason
which I forgot :D need to loom at the PR which did this

Thanks & regards,
Deepu

pascalgrimaud:
Don't hesitate to update / change / close this PR!
I'm at work soon, so I won't be able to update immediatly this PR, until lunch time...

deepu105:
Ok ill try to do it

Thanks & regards,
Deepu

jdubois:
Same for me, I'm on the train, network is too awful to test complex stuff
...

Le 31 janv. 2017 8:53 AM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

deepu105:
im fixing it

PR #5043, opened 2017-01-31
Tuning tsc

gmarziou:
By default, tsc includes **/* and our tsconfig.json excludes node_modules, I found it's better to include source folders explicitly (white list versus black list approach).

By defining includes explicitly, compilation time went from 12.80s to 11.15s with some variations of course.


gmarziou:
Travis errors are nto related to this change as it fails in ng1 about gulp tool.

Issue #5044, opened 2017-02-01
Account Register component not working properly


nicoraynaud:
##### **Overview of the issue**

While working with latest master branch, with angular 2, i noticed that the register form is not displaying errors when server side web service returns an HTTP 400 response.

##### **Motivation for or Use Case**

I feel this is just a ng2 migration issue, and could be easily fixed before the release coming soon.

##### **Reproduce the error**

Install latest generator-jhipster (as of today), create an app, run it and 
- go to http://localhost:9000/#/register
- set "admin" as username (this is to get the "login already exists" error)
- Fill in the other fields without errors.
- Click "subscribe

=> The form does not show the related error's div while the query has been made the server returned the HTTP 400.

##### **Suggest a Fix**

in **/app/account/register/register.component.ts** : i moved the content of "processError" private method as a lambda directly following the service call. This way, the changes made on the components data gets picked up by the template. 
I also had to change the check on the content of the response. It was checking whether response.data === "the string" while I had to check against **reponse._body**

Note that I am a js and ng2 beginner, so i'm really not sure i got this the proper way !

```
register() {
        if (this.registerAccount.password !== this.confirmPassword) {
            this.doNotMatch = 'ERROR';
        } else {
            this.doNotMatch = null;
            this.error = null;
            this.errorUserExists = null;
            this.errorEmailExists = null;
            this.languageService.getCurrent().then(key => {
                this.registerAccount.langKey = key;
                this.registerService.save(this.registerAccount).subscribe(() => {
                    this.success = true;
                }, error => {
                    this.success = null;
                    if (error.status === 400 && error._body === 'login already in use') {
                        this.errorUserExists = 'ERROR';
                    } else if (error.status === 400 && error._body === 'e-mail address already in use') {
                        this.errorEmailExists = 'ERROR';
                    } else {
                        this.error = 'ERROR';
                    }
                });
            });
        }
    }
```

Note that if i were to suggest more changes to this component, i would change the error variables from "string" to "boolean" data type, to make it more readable. (right now, we set "ERROR" when there is an error and null otherwise => this seems to me like a boolean type being disguised :D ).

##### **JHipster Version(s)**

JHipster v.3.12.2

##### **JHipster configuration**
.yo-rc.json

```
{                                                              
  "generator-jhipster": {                                      
    "jhipsterVersion": "3.12.2",                               
    "baseName": "optone",                                      
    "packageName": "nc.opt.template",                          
[...]

PR #5046, opened 2017-02-01
Fix for websockets using webpack-dev-server

flaviencathala:
Here is a fix for websockets in dev mod. But websockets are broken in the last versions of webpack-dev-server, see: https://github.com/chimurai/http-proxy-middleware/issues/143, so it still doesn't work.

Until they fix it or they find a way to fix it (the only known way to fix it, by changing spdy options, doesn't work, I don't know why), we can keep my fix.


flaviencathala:
I also fixed a bug in register (https://github.com/jhipster/generator-jhipster/issues/5044).
I forgot to create a new branch to make a new PR...

Issue #4989, opened 2017-01-27
Images can't be loaded with Angular2


asanzdiego:
- [x] For bug reports run `yo jhipster:info` in your project folder and copy paste the info after the **Suggest a Fix** section or fill the below template manually.
- [ ] For feature requests, label the title with [FEATURE] and explain your use case and ideas clearly below, you can remove sections which are not relevant.
- [x] Remove sensitive info if any
- [x] Make sure the first checkbox is followed(checked) for bug reports else the issue will be closed :smiling_imp:

##### **Overview of the issue**

Images can't be loaded with Angular2

##### **Motivation for or Use Case**

With AngularJS woks fine, but  not with Angular2.

##### **Reproduce the error**

```
? (1/15) Which *type* of application would you like to create? Monolithic application (recommended for simple projects)
? (2/15) Which *Framework* would you like to use for the client? [BETA] Angular 2.x
? (3/15) What is the base name of your application? dgise
? (4/15) Would you like to install other generators from the JHipster Marketplace? No
? (5/15) What is your default Java package name? dgise
? (6/15) Which *type* of authentication would you like to use? JWT authentication (stateless, with a token)
? (7/15) Which *type* of database would you like to use? SQL (H2, MySQL, MariaDB, PostgreSQL, Oracle, MSSQL)
? (8/15) Which *production* database would you like to use? MariaDB
? (9/15) Which *development* database would you like to use? H2 with in-memory persistence
? (10/15) Do you want to use Hibernate 2nd level cache? Yes, with ehcache (local cache, for a single node)
? (11/15) Would you like to use Maven or Gradle for building the backend? Maven
? (12/15) Which other technologies would you like to use? Search engine using ElasticSearch
? (13/15) Would you like to use the LibSass stylesheet preprocessor for your CSS? No
? (14/15) Would you like to enable internationalization support? Yes
? Please choose the native language of the application? Spanish
? Please choose additional languages to install English
? (15/15) Besides JUnit and Karma, which testing frameworks would you like to use? (Press <space> to select, <a> to toggle all, <i> to inverse selection) No one selected.
```

Afer:

`yo jhipster:import-jdl jhipster-jdl.jh`

With jhipster-jdl.jh:

```
entity Actuacion {
	actuacionNombre String required,
    actuacionDescripcion String required
}

entity Informe {
	informeDescripcion String required,
    informeFoto ImageBlob required
}

relationship OneToMany {
	Actuacion{informe} to Informe
}

[...]

PR #5048, opened 2017-02-01
Migrate file BLOB field for entity

VictorADS:
Related to #4989

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
@@ -11,7 +11,7 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { Response } from '@angular/http';
 
 import { NgbActiveModal, NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
-import { EventManager, AlertService<% if (enableTranslation) { %>, JhiLanguageService<% } %> } from 'ng-jhipster';
+import { EventManager, AlertService<% if (enableTranslation) { %>, JhiLanguageService<% } %><% if (fieldsContainBlob) { %>, DataUtils<% } %> } from 'ng-jhipster';
 
 import { <%= entityClass %> } from './<%= entityFileName %>.model';
 import { <%= entityClass %>PopupService } from './<%= entityFileName %>-popup.service';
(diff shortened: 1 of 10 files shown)


deepu105:
This looks good enough to start with. I dont want to implement the boilerplate for a full fledged file uploader here as there are lots of libraries which does that so this is good to start if required later we can consider some library

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -90,7 +93,32 @@ export class <%= entityAngularJSName %>DialogComponent implements OnInit {
...
+        if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { _%>
+    set<%= fields[idx].fieldNameCapitalized %>($event, <%= entityInstance %>) {
+        if ($event.target.files && $event.target.files[0]) {
+            let $file = $event.target.files[0];
+            <%_ if (fields[idx].fieldTypeBlobContent === 'image') { _%>
+                if (!/^image\//.test($file.type)) {
+                    return;
+                }
+                <%_ } _%>
+            this.dataUtils.toBase64($file, function(base64Data) {
```
deepu105: use arrow function

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -90,7 +93,32 @@ export class <%= entityAngularJSName %>DialogComponent implements OnInit {
...
+    }
 
+    openFile(contentType, field) {
+        return this.dataUtils.openFile(contentType, field);
+    }
+    <%_ } _%>
+    <%_ for (idx in fields) {
+        if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { _%>
+    set<%= fields[idx].fieldNameCapitalized %>($event, <%= entityInstance %>) {
+        if ($event.target.files && $event.target.files[0]) {
```
deepu105: for multiple fields this will be repeated can we move the common boilerplate code into a private helper method and pass the filed specific stuff below as a callback?

I would even consider writing a generic function like below as I hate to see duplicated code and pass the filedName and a boolean if its image as below
```
setFileData($event, <%= entityInstance %>, field, isImage) {
  if ($event.target.files && $event.target.files[0]) {
    let $file = $event.target.files[0];
    if (isImage && !/^image\//.test($file.type)) {
      return;
    }
    this.dataUtils.toBase64($file, (base64Data) => {
      <%= entityInstance %>[field] = base64Data;
      <%= entityInstance %>[`${field}ContentType`] = $file.type;
    }
  }
}
```

deepu105: But this should actually be written as a directive

Issue #5065, opened 2017-02-02
Node 6.9.0 should be required


jvence:
Welcome to the JHipster Generator v4.0.0
Documentation for creating an application: https://jhipster.github.io/creating-an-app/
Application files will be generated in folder: /Users/alain/dev/project/myapp/HPWebServices2.1/gateway
WARNING! yarn is not found on your computer.

? (1/15) Which *type* of application would you like to create? Microservice gateway
? (2/15) Which *Framework* would you like to use for the client? [BETA] Angular 2.x
? (3/15) What is the base name of your application? hpgateway
? (4/15) Would you like to install other generators from the JHipster Marketplace? No
? (5/15) As you are running in a microservice architecture, on which port would like your server to run? It should be
 unique to avoid port conflicts. 8082
? (6/15) What is your default Java package name? com.myapp.gateway
? (7/15) Which *type* of authentication would you like to use? JWT authentication (stateless, with a token)
? (8/15) Which Service Discovery and Configuration solution would you like to use? JHipster Registry (using Eureka an
d Spring Cloud Config)
? (9/15) Which *type* of database would you like to use? MongoDB
? (10/15) Would you like to use Maven or Gradle for building the backend? Maven
? (11/15) Which other technologies would you like to use? (Press <space> to select)
? (12/15) Would you like to use the LibSass stylesheet preprocessor for your CSS? No
? (13/15) Would you like to enable internationalization support? No
? (14/15) Besides JUnit and Karma, which testing frameworks would you like to use? Gatling
events.js:141
      throw er; // Unhandled 'error' event
      ^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode in /usr/local/lib/node_modules/generator-jhipster/generators/server/templates/_README.md while compiling ejs
    at Function (native)
    at Object.Template.compile (/usr/local/lib/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:495:12)
    at Object.compile (/usr/local/lib/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:306:16)
    at handleCache (/usr/local/lib/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:169:18)
    at Object.exports.render (/usr/local/lib/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:332:10)
[...]

PR #5082, opened 2017-02-03
Check node version when using the generator. Need at least 6.9.0+

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/5065
I don't put `fix` or `close` as @gmarziou suggested another solution too

```
Welcome to the JHipster Generator v4.0.0
Documentation for creating an application: https://jhipster.github.io/creating-an-app/
Application files will be generated in folder: /home/pgrimaud/tmp/23-toto
WARNING! Your node version is too old (v4.5.0). You should use at least Node v6.9.0+
? (1/15) Which *type* of application would you like to create? (Use arrow keys)
❯ Monolithic application (recommended for simple projects) 
  Microservice application 
  Microservice gateway 
  [BETA] JHipster UAA server (for microservice OAuth2 authentication) 
```

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -108,6 +108,24 @@ module.exports = JhipsterGenerator.extend({
             }.bind(this));
         },
 
+        checkNode: function () {
+            if (this.skipChecks || this.skipServer) return;
+            var done = this.async();
+            exec('node -v', function (err, stdout, stderr) {
+                if (err) {
+                    this.warning('Node is not found on your computer.');
+                } else {
+                    var nodeVersionMajor = stdout.split('.')[0].replace(/v/g, '');
+                    var nodeVersionMinor = stdout.split('.')[1];
+                    var nodeVersion = stdout.replace(/\n/g, '');
+                    if (nodeVersionMajor < 6 || (nodeVersionMajor >= 6 && nodeVersionMinor < 9)) {
+                        this.warning('Your node version is too old (' + nodeVersion + '). You should use at least Node ' + chalk.bold('v6.9.0+'));
+                    }
+                }
+                done();
+            }.bind(this));
+        },
+
         checkGit: function () {
             if (this.skipChecks || this.skipClient) return;
             var done = this.async();


pascalgrimaud:
After reading the stackoverflow post, this solution is more permissive (a warning) and more consistent with other checks version we already have, so I think we can close #5065 ?

jdubois:
Not sure if the warning is enough, as people will probably ignore it, but then that's indeed more consistent.... And who knows, maybe it could work with some older versions...

deepu105:
@pascalgrimaud @jdubois I dont think we should do it this way. This would also require manually maintaining the version in the prompts. The below in package should have worked so I would investigate that first and even if we are doing a check we should check against whats defined as engine instead of a hardcoded check
```
  "engines": {
    "node": ">=6.9.0"
  },
```

pascalgrimaud:
Agree with you @deepu105

jdubois:
Yes agreed @deepu105 - and I really don't get why the "engines" part of the package.json doesn't work.

deepu105:
@jdubois I think it works but it just prints a warning so may be people dont read that warning :)

jdubois:
Oh... If that's the case then I'll be quite angry with all those bug reports.... Anyway it's clearly written in our documentation.

deepu105:
Oh wait but then we are using yarn so we need to check if Yarn prints warnings as well. But NPM warnings are quite noisy so easy for people to miss it actually. They shouldnt have removed the strictEngine keyword

jdubois:
And we got a bug on the new check: https://twitter.com/ecukalla/status/827848436364013568

generators/app/index.js
```diff
@@ -108,6 +108,24 @@ module.exports = JhipsterGenerator.extend({
...
+            if (this.skipChecks || this.skipServer) return;
+            var done = this.async();
+            exec('node -v', function (err, stdout, stderr) {
+                if (err) {
+                    this.warning('Node is not found on your computer.');
+                } else {
+                    var nodeVersionMajor = stdout.split('.')[0].replace(/v/g, '');
+                    var nodeVersionMinor = stdout.split('.')[1];
+                    var nodeVersion = stdout.replace(/\n/g, '');
+                    if (nodeVersionMajor < 6 || (nodeVersionMajor >= 6 && nodeVersionMinor < 9)) {
```
deepu105: Btw if we are checking this then better to use the semver library to do that we already use that  in `isJhipsterVersionLessThan` method in `generator-base.js`

anjia0532: my node version is v7.5.0 is least version but `yo jhipster` waring `). You should use at least Node v6.9.0+v7.5.0`
**(nodeVersionMajor >= 6 && nodeVersionMinor < 9)** is not ok

pascalgrimaud:
Oh sorry, I coded it too fast... :(

pascalgrimaud:
It should be == instead of >=...

anjia0532:
@pascalgrimaud  yes ,should be ==, my PR https://github.com/jhipster/generator-jhipster/pull/5090

Issue #5088, opened 2017-02-04
yo jhispter wrongly warns for old version of npm


ecukalla:
Greetings and thanks for jhipster 4 release!!!

##### **Overview of the issue**
I have node 7.5.0, while yo jhipster, warns that I should have at least 6.9.0+

    WARNING! Your node version is too old (v7.5.0). You should use at least Node v6.9.0+

<img width="562" alt="screen shot 2017-02-04 at 12 55 47" src="https://cloud.githubusercontent.com/assets/775064/22618608/0f18e336-eae1-11e6-8450-57b7f1744398.png">

##### **Motivation for or Use Case**

It seems to me like a bug, wrongly reporting a minor version of 6.9.0

##### **Reproduce the error**

Just run yo jhipster.

##### **JHipster configuration**

    yo jhipster:info
    Welcome to the JHipster Information Sub-Generator
    
    ##### **JHipster Version(s)**
    
    ```
    /Users/edmond/Documents/ws/smth
    └── (empty)
    
    ```
    
    cat: no such file or directory: .yo-rc.json
    
    ##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**
    
    ```yaml
    
    ```
    
    
    ##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**
    
    ls: no such file or directory: .jhipster/*.json
    
    ##### **Browsers and Operating System**
    
    java version "1.8.0_101"
    Java(TM) SE Runtime Environment (build 1.8.0_101-b13)
    Java HotSpot(TM) 64-Bit Server VM (build 25.101-b13, mixed mode)
    
    git version 2.10.1 (Apple Git-78)
    
    node: v7.5.0
    
    npm: 4.1.2
    
    bower: 1.8.0
    
    gulp:
    [13:50:01] CLI version 1.2.2
    
    yeoman: 1.8.5
    
    yarn: 0.17.10
    
    Docker version 1.13.0, build 49bf474
    
    docker-compose version 1.10.0, build 4bd6f1a

pascalgrimaud:
@ecukalla : I code it too fast yesterday sorry. The condition should be with == instead of >=

alachebbi:
project doesn't work with jhipster 4.0.1 
error jhipster-core@1.2.8: The engine "node" is incompatible with this module. Expected version ">=6.9.5".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
[...]

PR #5093, opened 2017-02-04
Use semver to check node version

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5088 with the solution proposed by @deepu105 here : https://github.com/jhipster/generator-jhipster/pull/5082#discussion_r99466122

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -6,7 +6,8 @@ var util = require('util'),
     cleanup = require('../cleanup'),
     prompts = require('./prompts'),
     packagejs = require('../../package.json'),
-    exec = require('child_process').exec;
+    exec = require('child_process').exec,
+    semver = require('semver');
 
 var JhipsterGenerator = generators.Base.extend({});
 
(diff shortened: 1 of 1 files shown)


generators/app/index.js
```diff
@@ -115,11 +116,9 @@ module.exports = JhipsterGenerator.extend({
                 if (err) {
                     this.warning('Node is not found on your computer.');
```
deepu105: should be `NodeJS is not found on your system`

generators/app/index.js
```diff
@@ -115,11 +116,9 @@ module.exports = JhipsterGenerator.extend({
...
                 } else {
-                    var nodeVersionMajor = stdout.split('.')[0].replace(/v/g, '');
-                    var nodeVersionMinor = stdout.split('.')[1];
-                    var nodeVersion = stdout.replace(/\n/g, '');
-                    if (nodeVersionMajor < 6 || (nodeVersionMajor >= 6 && nodeVersionMinor < 9)) {
-                        this.warning('Your node version is too old (' + nodeVersion + '). You should use at least Node ' + chalk.bold('v6.9.0+'));
+                    var nodeVersion = semver.clean(stdout);
+                    if (semver.lt(nodeVersion, '6.9.0')) {
```
deepu105: can we put the supported version in the constants and use that here and in the package.json templates please? so that we dont have to change in multiple places

pascalgrimaud: Do you prefer using constants or the value inside the package.json:
```json
  "engines": {
    "node": ">=6.9.0",
    "npm": ">=2.14.2"
  }
```

pascalgrimaud: I choose the value from package.json
Don't hesitate to tell me if you prefer a constants

generators/app/index.js
jdubois: Done

PR #5097, opened 2017-02-04
fix autoincrement

ctamisier:
In `_Entity.java` and in `_User.java`,
When using a database other than `mysql` and `mariadb`: the `hibernate_sequence` is used according to the following code:
`<% if (databaseType == 'sql') { %>`
`@Id`
`<%_ if (prodDatabaseType == 'mysql' || prodDatabaseType == 'mariadb') { _%>`
`@GeneratedValue(strategy = GenerationType.IDENTITY)`
`<%_ }  else { _%>`
`@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "sequenceGenerator")`
`@SequenceGenerator(name = "sequenceGenerator")`
`<%_ } _%>`

In `_added_entity.xml`, the `autoIncrement` variable is set to `true` for `postgresql` (and h2,oracle,...),
which creates a sequence (with liquibase) for this entity **which is actually not used** because of the use of sequenceGenerator described above. (This is my case and I'm using postgresql 9.6):
`<property name="autoIncrement" value="true" dbms="mysql,h2,postgresql,oracle,mssql"/>`

In _initial_schema.xml, everything is fine for the User entity and `autoincrement` is set to `true` only for `mariadb` and `mysql`, which doesn't create a sequence in database.

So I would say that `autoIncrement` should be set to `true` only for `mysql` and `mariadb` ?

### DIFF
--- a/generators/entity/templates/server/src/main/resources/config/liquibase/changelog/_added_entity.xml
+++ b/generators/entity/templates/server/src/main/resources/config/liquibase/changelog/_added_entity.xml
@@ -11,7 +11,7 @@
     <property name="now" value="sysdate" dbms="oracle"/>
     <property name="now" value="GETDATE()" dbms="mssql"/>
 
-    <property name="autoIncrement" value="true" dbms="mysql,h2,postgresql,oracle,mssql"/>
+    <property name="autoIncrement" value="true" dbms="mysql,mariadb"/>
 
     <property name="floatType" value="float4" dbms="postgresql, h2"/>
     <property name="floatType" value="float" dbms="mysql, oracle, mssql"/>


jdubois:
You're probably the only one who reviewed this code (and I'm the author of the code, btw), so we should discuss this.

First discussion, MySQL and MariaDB should be exactly the same:

- for this part you can should be correct
- but I just tested with MariaDB, and we do have auto increment working
- and as MariaDB is a MySQL fork, I've looked (quickly) at the Liquibase documentation, and it seems they consider those databases are the same

-> so in the end, the current works, and the "mariadb" keyword is probably ignored. But we could add it, in case it becomes useful in a future release.

For the other databases, they should not support auto increment, and that's why we use a sequence 

- in that case the "auto increment" should not work
- so I guess Liquibase is just ignoring them, as there is no error

-> so for me it doesn't cause any error, even if this variable is total non-sense, of course

So in the end I'm OK with your change, even if the current code doesn't cause any issue at all. Then, the issue is that this needs to be tested on all databases, did you do it? It's not that hard, using Docker, even for Oracle and MS SQL Server

jdubois:
oh sorry, this should be like in the "initial_schema.xml" file!!! I'll do it.

jdubois:
I'm linking my commit to here so you have the reference

PR #5099, opened 2017-02-04
remove Reflect.ts warnings from console also in Windows

kaidohallik:
Without this change 144 warnings from `Reflect.ts` are shown in Windows:
* in command line window on dev server start and on every HMR
* also in web browser console on load and on every HMR


gmarziou:
I tested it successfully on Windows. Great.

PR #5107, opened 2017-02-05
entity generation taking angular suffix into account

baskeboler:
entity generation was not generating correctly when an angular suffix was supplied. jhipster:import-jdl failed with example from https://jhipster.github.io/jdl-studio/


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-delete-dialog.component.ts
```diff
@@ -12,23 +12,23 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { NgbActiveModal, NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import { EventManager<% if (enableTranslation) { %>, JhiLanguageService<% } %> } from 'ng-jhipster';
 
-import { <%= entityClass %> } from './<%= entityFileName %>.model';
-import { <%= entityClass %>PopupService } from './<%= entityFileName %>-popup.service';
-import { <%= entityClass %>Service } from './<%= entityFileName %>.service';
+import { <%= entityAngularJSName %> } from './<%= entityFileName %>.model';
```
mraible: #ItsJustAngular so please rename from `AngularJS` to `Angular`

baskeboler: sure, I was reusing an existing property.

deepu105:
This needs some testing so please lets us know if you have tested the below

- [ ] with suffix
- [ ] without suffix
- [ ] with different relationships (oneToMany, manyToMany, OneToOne, ManyToOne)
- [ ] relationship with user

baskeboler:
@deepu105 thanks for the suggestion, I did some more testing and this need some more work .. 
this is what i found:

* [x] with suffix
  * works fine
* [x] without suffix
* [ ] with different relationships (oneToMany, manyToMany, OneToOne, ManyToOne)
  * works fine as long as the other entity has the same angular suffix, otherwise breaks
  * we need to figure out the other entity's suffix, any suggestions? Should I read the other entity's json file?
* [ ] relationship with user
  * works fine with no suffix, breaks with suffix, same problem as above.

deepu105:
we already read the otherEntity json for tableName and stuff so you could also get the suffix when we read it

deepu105:
@pascalgrimaud can we add these into any of our existing tests

pascalgrimaud:
@deepu105 : yes sure. In fact, our entities [here](https://github.com/jhipster/generator-jhipster/tree/master/travis/samples/.jhipster) need to be updated / improved / completed, because there are some use cases we don't test

generators/entity/index.js
```diff
@@ -238,6 +238,38 @@ module.exports = EntityGenerator.extend({
             this.skipServer = true;
         }
     },
+
+    _loadOtherEntityJson: function(entityName) {
```
deepu105: As I said we already read it for tableName somewhere hence please try to reuse that or extarct both into this common function so that we dont read the same file twice. If you cant find where its read let me know I'll point you to there as I need to look it up as well. But if im not wrong its done in the liquibase changelog template I guess

baskeboler: yes, i just realized that. i just removed that and used getEntityProperty from the base generator.

deepu105:
ok hold on. I guess again we end up reading the file twice so lets try to read the file once and get all the properties we are interested in so do you mind waiting so that I can change existing function so that you can later update your PR to just use the property?

baskeboler:
@deepu105 I don't see where the json files are being read for relationships ... can you give me the location?

deepu105:
Ill reafctor the existing code this weekend and will merge this manually

baskeboler:
sounds good :)

deepu105:
@baskeboler its done so you can rebase and ammend your PR to read the required data from `otherEntityData` variable at line 556 of entity/index.js its done here https://github.com/jhipster/generator-jhipster/commit/d6e018a0f7ba7a5f0bcd16e2df7e75783b65fdd0

baskeboler:
Thanks, I will do this tonight.

generators/entity/templates/client/angularjs/src/main/webapp/app/entities/_entity-management.controller.js
```diff
@@ -7,7 +7,7 @@
 
     <%= entityAngularName %>Controller.$inject = [<% if (fieldsContainBlob) { %>'DataUtils', <% } %>'<%= entityClass %>'<% if (searchEngine == 'elasticsearch') { %>, '<%= entityClass %>Search'<% } %><% if (pagination != 'no') { %>, 'ParseLinks', 'AlertService', 'paginationConstants'<% } %> <%_ if (pagination == 'pager' || pagination == 'pagination'){ %>, 'pagingParams'<% } %>];
 
-    function <%= entityAngularName %>Controller (<% if (fieldsContainBlob) { %>DataUtils, <% } %><%= entityClass %><% if (searchEngine == 'elasticsearch') { %>, <%= entityClass %>Search<% } %><% if (pagination != 'no') { %>, ParseLinks, AlertService, paginationConstants<% } %> <%_ if (pagination == 'pager' || pagination == 'pagination'){ %>, pagingParams<% } %>) {
+    function <%= entityAngularName %>Controller(<% if (fieldsContainBlob) { %>DataUtils, <% } %><%= entityClass %><% if (searchEngine == 'elasticsearch') { %>, <%= entityClass %>Search<% } %><% if (pagination != 'no') { %>, ParseLinks, AlertService, paginationConstants<% } %> <%_ if (pagination == 'pager' || pagination == 'pagination'){ %>, pagingParams<% } %>)
```
deepu105: a open curly brace seems to be missing here?

generators/entity/index.js
```diff
@@ -428,8 +428,8 @@ module.exports = EntityGenerator.extend({
             this.entityPluralFileName = entityNamePluralizedAndSpinalCased + this.entityAngularJSSuffix;
             this.entityServiceFileName = entityNameSpinalCased + this.entityAngularJSSuffix;
             this.entityAngularName = this.entityClass + _.upperFirst(_.camelCase(this.entityAngularJSSuffix));
-            this.entityStateName = entityNameSpinalCased + this.entityAngularJSSuffix;
-            this.entityUrl = entityNameSpinalCased + this.entityAngularJSSuffix;
+            this.entityStateName = _.kebabCase(this.entityAngularName);
```
deepu105: does the change produce the same result?

baskeboler: its slightly different, if we have angularSuffix="MySuffix" and an entity "MyEntity", our angularName would be "MyEntityMySuffix" and our entityStateName would be "my-entity-my-suffix" where previously it was "my-entity-MySuffix", i think.

generators/entity/templates/client/angularjs/src/main/webapp/app/entities/_entity-management.controller.js
baskeboler: oops, my bad :(

deepu105:
you still have 15 commits here. You need to rebase to master and squash while rebasing it check the contribution guide on the repo root

Issue #5110, opened 2017-02-05
Health Refresh (ng2) should map data in case its 503 error (service down)


angusoid:
##### **Overview of the issue**

When doing health check and there is a certain service down, the server will return 503 with body response. 

Current code (ng2) for refreshing health is not covering that scenario as  shown below.

_health.component.ts_

    refresh() {
        this.updatingHealth = true;

        this.healthService.checkHealth().subscribe(health => {
            this.healthData = this.healthService.transformHealthData(health);
            this.updatingHealth = false;
        });
    }

Example from ng1 client
```
        vm.updatingHealth = true;
            <%=jhiPrefixCapitalized%>HealthService.checkHealth().then(function (response) {
                vm.healthData = <%=jhiPrefixCapitalized%>HealthService.transformHealthData(response);
                vm.updatingHealth = false;
            }, function (response) {
                vm.healthData =  <%=jhiPrefixCapitalized%>HealthService.transformHealthData(response.data);
                vm.updatingHealth = false;
            });
```
##### **Suggest a Fix**

_health.component.ts_
```
    refresh() {
        this.updatingHealth = true;

        this.healthService.checkHealth().subscribe(health => {
            this.healthData = this.healthService.transformHealthData(health);
            this.updatingHealth = false;
        }, error=> {
            if (error.status == 503) {
                this.healthData = this.healthService.transformHealthData(error.json());
                this.updatingHealth = false;
            }
        });
    }
```

##### **JHipster Version(s)**

Latest

jdubois:
Looks like you for the solution -> could you do a PR?

angusoid:
Sure

jdubois:
Thanks for the PR, closing this as it's merged

PR #5111, opened 2017-02-05
Fix ng2 health check display when status is 503(service down)

angusoid:
Fix for #5110


generators/client/templates/angular/src/main/webapp/app/admin/health/_health.component.ts
```diff
@@ -44,6 +44,11 @@ export class <%=jhiPrefixCapitalized%>HealthCheckComponent implements OnInit {
         this.healthService.checkHealth().subscribe(health => {
             this.healthData = this.healthService.transformHealthData(health);
             this.updatingHealth = false;
+        }, error => {
+            if (error.status == 503) {
```
gmarziou: this line produces a tslint warning:

WARNING in ./src/main/webapp/app/admin/health/health.component.ts
[48, 30]: == should be ===

jdubois: done

PR #5121, opened 2017-02-05
enable 'no database' option for uaa+gateway combination, polish prompts

ruddell:
Gateways that use UAA don't require a database.

Also, I simplified some of the conditions for databaseType prompt:
 - Don't prompt if a databaseType is already selected
 - Removed database prompt that appear when social login is enabled because that can't be selected until afterwards.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
@@ -42,10 +42,12 @@
                     <div *ngFor="let authority of user.authorities">
                         <span class="tag tag-info">{{ authority }}</span>
                     </div>
-                </td><% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
+                </td>
+                <%_ if (databaseType !== 'cassandra') { _%>
                 <td>{{user.createdDate | date:'dd/MM/yy HH:mm'}}</td>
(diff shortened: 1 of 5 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
```diff
@@ -42,10 +42,12 @@
                     <div *ngFor="let authority of user.authorities">
                         <span class="tag tag-info">{{ authority }}</span>
                     </div>
-                </td><% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
```
ruddell: This was to match the condition in the table header - the headers appear without the cells

jdubois:
I can't believe you can have a gateway without a database! Something I really didn't expected!

Then 2 things to check:

- You should still be able to use Cassandra for rate limiting -> there shouldn't be any change here, so that should still work
- I have no idea how the entity sub-generator will work -> but this works OK for the microservices with no database, so that shouldn't be a big deal

ruddell:
Cassandra Rate Limiting works the same (you need to create the appname_ratelimiting table manually if you don't use Cassandra as the database type).  I used Docker to test this.

For microservices, you can't generate an entity without a database.  When a gateway has no database, it will automatically ask for the microservice from which to generate the entity.

jdubois:
Perfect then! I just restarted your build as there was a failure

PR #5126, opened 2017-02-06
fix bold

ctamisier:
based on the english settings.json


jdubois:
Not sure we should put HTML tags in the i18n files... Of course this is in all the other i18n files, so I'm merging this so it all looks the same, but this is a bad practice

ctamisier:
Yeah I was wondering about this.. it seems that they do it in rubyonrails (and even adding an _html suffix on keys has a meaning..) : http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

PR #5127, opened 2017-02-06
Define a module for home

gmarziou:
Home page should have its own module to keep app.module simple and high level and because it gives some structure to users for modifying it.

Remove layout block dependency on home route as it deals only with layout not content.

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -126,6 +126,7 @@ const files = {
                 'entities/_entity.module.ts',
                 // home module
                 'home/_index.ts',
+                { file: 'home/_home.module.ts', method: 'copyJs' },
                 { file: 'home/_home.route.ts', method: 'copyJs' },
                 { file: 'home/_home.component.ts', method: 'copyJs' },
                 { file: 'home/_home.component.html', method: 'copyHtml' },
--- a/generators/client/templates/angular/src/main/webapp/app/home/_home.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/home/_home.module.ts
@@ -1,0 +1,24 @@
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { RouterModule } from '@angular/router';
+import { ParseLinks } from 'ng-jhipster';
+
+import { ParattSharedModule } from '../shared';
+
+import { HOME_ROUTES, HomeComponent } from './';
+
+
+@NgModule({
+    imports: [
+        ParattSharedModule,
+        RouterModule.forRoot(HOME_ROUTES, { useHash: true })
+    ],
+    declarations: [
+        HomeComponent,
+    ],
+    entryComponents: [
+    ],
+    providers: [
+    ],
+    schemas: [CUSTOM_ELEMENTS_SCHEMA]
+})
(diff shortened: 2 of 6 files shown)


generators/client/templates/angular/src/main/webapp/app/home/_home.module.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ParseLinks } from 'ng-jhipster';
+
+import { ParattSharedModule } from '../shared';
+
+import { HOME_ROUTES, HomeComponent } from './';
+
+
+@NgModule({
+    imports: [
+        ParattSharedModule,
```
deepu105: Missing template here

deepu105: AppName need to be in template

gmarziou:
Travis errors unrelated to code change

jdubois:
@gmarziou I'm restarting Travis -> you don't have permission to do it? You should, as part of the core dev team, but I have no idea where this is configured!

generators/client/files-angular.js
```diff
@@ -126,6 +126,7 @@ const files = {
                 'entities/_entity.module.ts',
                 // home module
                 'home/_index.ts',
+                { file: 'home/_home.module.ts', method: 'copyJs' },
```
deepu105: no need for copyJs as its required only if we have any i18n regex to strip. may be I should rename the method names to be more meaningful

generators/client/templates/angular/src/main/webapp/app/home/_home.module.ts
```diff
@@ -0,0 +1,24 @@
+import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { RouterModule } from '@angular/router';
+import { ParseLinks } from 'ng-jhipster';
```
deepu105: I dont think we are using parseLinks here?

generators/client/templates/angular/src/main/webapp/app/home/_home.route.ts
```diff
@@ -1,14 +1,16 @@
-import { Route } from '@angular/router';
+import { Routes } from '@angular/router';
```
deepu105: any reason to change to Route array?

gmarziou: That's what I've seen in some angular examples and it looked nicer to read but I can change if you prefer the array.

generators/client/templates/angular/src/main/webapp/app/home/_home.module.ts
gmarziou: Correct, I'll remove it

generators/client/files-angular.js
gmarziou: OK, I'll have a look.

generators/client/templates/angular/src/main/webapp/app/home/_home.route.ts
deepu105: we are using Route everywhere except when we have multiple routes in same file

gmarziou:
Closing to replace by corrected one.

PR #5131, opened 2017-02-07
Ng2 translations in webpack bundles

flaviencathala:
DO NOT MERGE until ng-jhipster 0.1.7 is released with this PR (https://github.com/jhipster/ng-jhipster/pull/28) merged.
Need some tests once next ng-jhipster will be released.

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -60,6 +60,8 @@
     "jasmine-core": "2.5.2",
     <%_ if (protractorTests) { _%>
     "jasmine-reporters": "2.2.0",
+    <%_ } if (enableTranslation) { _%>
+    "json-loader": "0.5.4",
     <%_ } _%>
     "karma": "1.4.0",
     "karma-chrome-launcher": "2.0.0",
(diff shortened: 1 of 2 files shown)


deepu105:
See my comments on the PR in ng-jhipster. We need to discuss this

jdubois:
@deepu105 so what do we do here?

deepu105:
Lets wait to see how it goes. Im personally against current i18n support in ng2 and i dont think we can support that practically so if the ngx-translate is killed in favor of that I would rather fork and use it as it works well for us.

jdubois:
@deepu105 any decision here? What can I do to help this move forward?

deepu105:
@jdubois I'm ok to bundle the files but then both this PR and https://github.com/jhipster/ng-jhipster/pull/28 will need change as we dont need a partial loader if we are bundling the files we can use a static loder to load the final concatenated json once. So it will need a rewrite of the current i18n implementation. Since its a not an urgent item I can do it later or @flaviencathala can proceed with the changes and I can review and guide him

flaviencathala:
@deepu105 I believe it's better to load only useful translation files to not overcharge the memory with useless files. It's not a big deal anyway.

deepu105:
@flaviencathala if we bundle it with webpack you will be loading the entire file into memory anyway so it doesn't matter actually. Thats why I said i like current way of loading partials over network :D

gzsombor:
Sorry, I've just saw the other issue, and the discussion alone. Currently as we call the 'setLocations' in every page, it force reloading all the json-s, for every page change. At least, locally, in my dev environment. So it's a bit far from optimal :)

deepu105:
closed in favour of PR #5689

PR #5132, opened 2017-02-07
remove redundant search translation key if elasticsearch is not enabled

kaidohallik:


kaidohallik:
If I was reviewing generated entity i18n files then I was wondering why is this key there, is this used somewhere, must I correct this, can I remove this.  
After analyzing generated code and generator-jhipster source code I know now that I can remove these lines.

I think that pros of this change are:
* generated entity i18n files reviewer has no questions any more for what is this key
* there is a minor decrease of i18n file sizes (if reviewer left these keys in previously generated project)

and cons are:
* there is additional condition in generator-jhipster code what makes code more complex
* maybe someone is using this key for custom entity search

If you think that cons win then feel free to close this PR.

deepu105:
I'll leave it to @jdubois to decide

jdubois:
Usually I do this, as it's more important to have "perfect generated code" for end-users. This key is also confusing to people not using Elasticsearch.
Honestly I wouldn't have done it myself, as I don't think it's worth all that effort, but now you did it, let's merge it.

kaidohallik:
Actually this task was perfect for IntelliJ IDEA Replace All functionality. Sharing what I did, maybe someone can benefit:
1. In Project Explorer right click on folder `generators/entity/templates/client/i18n` and choose `Replace in path...`
2. Check option `Regular expression`
3. Text to find: `,\n                "search": "(.*)"\n`
4. Replace with: `<% if (searchEngine === 'elasticsearch') { %>,\n                "search": "$1"<% } %>\n`
5. Click `Find` and then click `All files`

PR #5109, opened 2017-02-05
Also include login and home in languageService locations

eelcobrolman:
Fixes #5102

PR #5134, opened 2017-02-07
Fix compilation error for social login and no i18n

gmarziou:


deepu105:
@gmarziou its not required to add conditions for this. we need to make this file copied using the `copyJs` method in the files.js so that the regex removes this. Thats how we do this for all the components

gmarziou:
I did not know, I just found that one social component was failing because it was setting i18n location in constyructor while not having the jhiLanguageService injected. I'll see if I can simplify the template.

gmarziou:
@deepu105 the regexp for removing i18n locations is able to match single argument, this is not enough since  #5104 and #5109

~~~js
            /(this\.[a-zA-Z0-9]*(L|l)anguageService\.setLocations\(\[[\'|\"][a-zA-Z0-9-_]*[\'|\"]\]\)\;[\s]*)/,// jhiLanguageService invocations
~~~

I propose to simplify it and to make it work for multiple arguments:

~~~js
            /(this\.[a-zA-Z0-9]*(L|l)anguageService\.setLocations\(\[[\'\"a-zA-Z0-9-_,\s]+\]\)\;[\s]*)/,// jhiLanguageService invocations
~~~

deepu105:
yes ofcourse we should do that. Also if we are going with mwerging i18n
json into the js with webpack then we wouldnt need this at all as all the
json keys will be available but thats another change all together


Thanks & Regards,
Deepu

Issue #5128, opened 2017-02-07
Docker 1.13


pascalgrimaud:
Docker 1.13 has been released few weeks ago, so I would like to update:
- [X] the documentation, to use the new Docker API
- [X] the docker-compose files: it was implemented at v1! Since the v1, Docker Network has been improved a lot. So we should remove all `links` or `external_links` for using DNS alias instead
- [X] update compose files at https://github.com/jhipster/jhipster-docker-hub

Then, it need to be tested, as I know it's something all of you use in demo, specially with MicroServices.
Are you ok with that ?

PierreBesson:
I was going to add this feature request as well.

I think we should remove all links, external-links, container-name and move to docker-compose v3. This will let us deploy on swarm mode from the docker-compose file (At last !).

That would mean we require docker >= 1.13.

jdubois:
Yes +1 of course
I would also like to be compatible with http://rancher.com/rancher-os/ (they have their own .yml format, and a load balancer to configure), it shouldn't be very hard. And if you want to play with it, I have my own cluster at Ippon :-)

pascalgrimaud:
@PierreBesson : I already finish to remove all links and external_links on my computer and begin to test everything since this morning. I need to remove container name too, as suggested.
As soon as it's ready, I will PR to let you test it

@jdubois : never tested rancher, but let me have a look on this!

PierreBesson:
@pascalgrimaud Amazing ! I will try it on a swarm mode cluster. It really easy to do thanks to docker-machine.
@jdubois, For rancher, if the format differs from docker-compose, this would be a specific sub-generator. I'm looking forward to try it.

pascalgrimaud:
Agree with you @PierreBesson !
@jdubois : should we open a new ticket for Rancher ?

jdubois:
Yes new ticket for Rancher! You should be able to run it on your laptop, but I have "true" cluster at Ippon, so I will be able to do some much better tests.
Just so you know my cunning plan: I'm preparing the new "onboarding" at Ippon for new hires, and I want them to deploy to RancherOS an application that can scale automatically. Sounds like fun :-)

pascalgrimaud:
ahaha sounds very fun!
It's incredible to work for JHipster. After Angular2, there are still so much ideas to improve the project!

jdubois:
Now that the Docker PR are merged, I'm closing this.

PR #5135, opened 2017-02-07
Improve compose files with Docker 1.13

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/5128
- remove all container_name, links and external_links to use DNS alias instead
- add npm test

@PierreBesson : if you have time to review it

@jdubois / @mraible : I know you do a lot of demos, so I hope you can test this. I don't want to break something!

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -229,7 +229,6 @@ module.exports = DockerComposeGenerator.extend({
                         cassandraMigrationConfig.build.context = relativePath;
                         var createKeyspaceScript = cassandraClusterYaml.services[databaseServiceName + '-migration'].environment[0];
                         cassandraMigrationConfig.environment.push(createKeyspaceScript);
-                        cassandraMigrationConfig['links'] = cassandraClusterYaml.services[databaseServiceName + '-migration'].links;
                         var cqlFilesRelativePath = pathjs.relative(this.destinationRoot(), path + '/src/main/resources/config/cql');
                         cassandraMigrationConfig['volumes'][0] = cqlFilesRelativePath + ':/cql:ro';
 
--- a/generators/docker-compose/templates/_jhipster-registry.yml
+++ b/generators/docker-compose/templates/_jhipster-registry.yml
@@ -1,7 +1,6 @@
 version: '2'
 services:
     jhipster-registry:
-        container_name: jhipster-registry
         image: <%= DOCKER_JHIPSTER_REGISTRY %>
         volumes:
             - ./central-server-config:/central-config
--- a/generators/docker-compose/templates/_kafka.yml
+++ b/generators/docker-compose/templates/_kafka.yml
@@ -1,10 +1,8 @@
 version: '2'
 services:
     zookeeper:
-        container_name: zookeeper
         image: <%= DOCKER_ZOOKEEPER %>
     kafka:
-        container_name: kafka
         image: <%= DOCKER_KAFKA %>
         environment:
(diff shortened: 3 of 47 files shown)


PierreBesson:
@pascalgrimaud I have notice a new docker-compose feature that could replace our JHipster Sleep by using docker healthchecks and depends_on : https://docs.docker.com/compose/compose-file/#/dependson

Also big news, I have managed to deploy a docker-compose with a full microservice stack to a swarm, with scaling working correctly. So when I have the time I will finally be able to add the option to generate a docker-compose for "prod".

romaindequidt:
May I propose you smething like this?
https://github.com/romaindequidt/generator-jhipster/commit/72038a4588cebbe3cca8a7a5ebeb6587398925e1

pascalgrimaud:
@PierreBesson : interesting! so we need to upgrade the docker-compose file version to 2.1+. I think it can be done on another PR to be clearer

@romaindequidt : sure, it looks better than what I did:
- you can PR directly on my branch (so this PR will be updated)
- or you can wait this PR to be merged, then you make a PR, so you get credits for your work

romaindequidt:
@pascalgrimaud I wanted to PR our PR but i do not if i can (i'm not jhipster member and I did find your branch :/)
Sorry to ask you that but, may you explain me the procedure?

pascalgrimaud:
@romaindequidt : so if you want to PR on my branch, just go to this link https://github.com/pascalgrimaud/generator-jhipster/compare/docker-1.13...romaindequidt:docker-1.13

jdubois:
LGTM so @pascalgrimaud I let you merge this when the PR on your PR is also merged :-)

romaindequidt:
@jdubois @pascalgrimaud Are you sure to force everyone to upgrade their Docker client version?
For exemple, some AWS service are only available for 1.12 version:
[Amazon ECS Container Agent Versions](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html)
Should we add a subgenerator option/parameter to choose one of the versions?

jdubois:
Thanks @romaindequidt I didn't notice that!!!
No we can't support multiple versions, that would be a lot of work for something quite temporary. Our options are:

- Stay on the newer version, and if people want to use it, they need to upgrade (or stay on an old JHipster release otherwise)
- Switch back to the older version, and in that case people won't benefit from the newer code

I would stay with 1.13, people need to upgrade, or use an up-to-date cloud. @pascalgrimaud what's your opinion?

pascalgrimaud:
In fact, this PR removes `links` and `external_links`, to use DNS alias instead.
It's not concerned by Docker 1.13 -> the title was wrong, it should be: `Improve compose files by using DNS alias`
The version in Docker compose files is still v2, not v3 (Docker 1.13), so don't worry @romaindequidt it is still compatible :)

But I made this PR https://github.com/jhipster/jhipster.github.io/pull/386 to update the Docker client API to 1.13

romaindequidt:
@pascalgrimaud Ok, if it's not this PR it will be the next one ;)
I 'd like to work on the "healthchecks" and "depends_on" parameters proposed by @PierreBesson , so in that case we go forward and we do not manage "backward compatibility"

pascalgrimaud:
Yes plz @romaindequidt !

PierreBesson:
@romaindequidt Yes a PR would be very much appreciated ! Do not worry about "background compatibility" as for us docker-compose is first and foremost a "dev" tool (no meaningful way to manage your "prod" with it before the last version).

PR #5139, opened 2017-02-07
sync gradle dev profile with maven dev profile in Angular 2

kaidohallik:
Currently in gradle there are 2 quite inconvenient differences from maven:
1. In `dev` profile maven doesn't execute any frontend tasks but gradle executes `yarn/npm install` (it also triggers `webpack` from `postInstall` in `package.json`) and `webpack` second time. It makes every app startup with gradle very slow.
As a side note: in code https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/gradle/_profile_dev.gradle#L67 I can see that there is tried to achieve the same behaviour as with maven but in gradle all tasks are also project properties, so `webpack.onlyIf { project.hasProperty('webpack') }` always returns true.
2. In `webpack` profile maven doesn't execute `npm/yarn install` but gradle executes. Beside install steps it causes double `webpack` compilation when using gradle (`postInstall` command from `package.json` runs also `webpack`).

This PR removes these differences, makes gradle behave similarily with maven.

After this change:  
`gradlew` is starting app with dev profile without any frontend task  
`gradlew -Pwebpack` runs webpack and after that starts app (no need to specify `-Pdev -Pwebpack` profiles like is needed with maven: `-Pdev,webpack`)

### DIFF
--- a/generators/server/templates/gradle/_profile_dev.gradle
+++ b/generators/server/templates/gradle/_profile_dev.gradle
@@ -31,7 +31,7 @@ task gulpConstantDev(type: GulpTask) {
     args = ["ngconstant:dev", "--no-notification"]
 }
     <%_ } else { _%>
-task webpack(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task, dependsOn: '<%= clientPackageManager %>_install') {
+task webpackBuildDev(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task) {
     args = ["run", "webpack:build:dev"]
 }
     <%_ } _%>
@@ -63,7 +63,7 @@ processResources {
     <%_ if (clientFramework === 'angular1') { _%>
 processResources.dependsOn gulpConstantDev
     <%_ } else { _%>
-processResources.dependsOn webpack
-webpack.onlyIf { project.hasProperty('webpack') }
+processResources.dependsOn webpackBuildDev
+webpackBuildDev.onlyIf { project.hasProperty('webpack') }
     <%_ } _%>
 <%_ } _%>


jdubois:
@atomfrede as you are our Gradle maintainer and expert can you do a review?

atomfrede:
I will check this evening too.

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -31,7 +31,7 @@ task gulpConstantDev(type: GulpTask) {
     args = ["ngconstant:dev", "--no-notification"]
 }
     <%_ } else { _%>
-task webpack(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task, dependsOn: '<%= clientPackageManager %>_install') {
+task webpackTask(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task) {
```
atomfrede: I would name the task according to what webpack is doing. So maybe something like ``webpackBuildDev``

kaidohallik: Thanks for the suggestion, task name changed.

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -31,7 +31,7 @@ task gulpConstantDev(type: GulpTask) {
     args = ["ngconstant:dev", "--no-notification"]
 }
     <%_ } else { _%>
-task webpack(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task, dependsOn: '<%= clientPackageManager %>_install') {
+task webpackBuildDev(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task) {
```
deepu105: In maven the webpack profile installs local node and yarn|npm so instead of depending on npmInstall|yarnInstall we should see if we can depend on a task to make sure local versions are installed if possible otherwise I'm fine with it

generators/server/templates/gradle/_profile_dev.gradle
```diff
@@ -63,7 +63,8 @@ processResources {
     <%_ if (clientFramework === 'angular1') { _%>
 processResources.dependsOn gulpConstantDev
     <%_ } else { _%>
-processResources.dependsOn webpack
-webpack.onlyIf { project.hasProperty('webpack') }
+processResources.dependsOn webpackBuildDev
+// tasks are properties of the project, so it's important to have different task and profile names
```
deepu105: I dont think this comment is required

generators/server/templates/gradle/_profile_dev.gradle
atomfrede: With the current change `webpackBuildDev` does not depend on any task. Should we remove the dependency for `yarn install` (`npm install`) we currently have?

kaidohallik: @deepu105 In JHipster gradle projects local versions can be installed with `-PnodeInstall` flag.  
Possible solution is to force local `node/npm/yarn` install also if `-Pwebpack` flag is specified.  
But to me it seems wrong. In what use case is someone using `-Pwebpack` while not `node` and `npm` or `yarn` installed?  
The purpose of `-Pwebpack` seems to be: if developer doesn't want to run development server with `yarn/npm start`, but changes front-end, then front-end can be compiled by `-Pwebpack` flag and run only `gradle/mvn` or am I missing something?  
One real world example for `-Pwebpack` use case is: running app after entity generation.

@atomfrede I didn't find these dependencies, can you give more information what did you mean with that?

kaidohallik:
@deepu105 I removed comment and rebased.

generators/server/templates/gradle/_profile_dev.gradle
atomfrede: I think before the webpack task also did a `yarn install` (via `dependsOn: 'yarn_install'`) it looks like it is missing now, but not sure if it is required.

kaidohallik: Maven `webpack` profile doesn't execute `yarn_install/npm_install`, so now gradle behaviour is similar to maven behaviour.

I think developers normal behaviour is to execute `yarn install` only after modifying `package.json`.
And they do it manually, it's normal. This is also written in generated `README.md`.

Currently `yarn_install` made webpack task slow: unnecessary install checks and double `webpack` compilation (`postInstall` from `package.json` also triggered `webpack`).

One concern can be: maybe this unexpected all time `webpack` execution and therefore `yarn_install` execution was used somewere, for example in `Jenkinsfile` or `.gitlab-ci.yml`.
But I see that there is all correct, commands are in the form:
```
./gradlew npmInstall -PnodeInstall
```
`npmInstall` parameter is correctly added there.

atomfrede: Ah, sorry I didn't check maven :( So everything is fine ✔️

PR #5144, opened 2017-02-08
fix app starting instruction in Windows

kaidohallik:


kaidohallik:
@deepu105 Thanks for your valuable comments. Replaced `os.platform()` specific solution with `path.sep` specific solution. PR updated.

generators/server/index.js
```diff
@@ -402,11 +403,11 @@ module.exports = JhipsterServerGenerator.extend({
 
         let logMsg =
             'Run your Spring Boot application:' +
-            '\n ' + chalk.yellow.bold('./mvnw');
+            '\n ' + chalk.yellow.bold('.' + path.sep + 'mvnw');
```
deepu105: @kaidohallik so on windows this would be `.mvnw` and on *nix `./mvnw` can you check if `.mvnw` works in windows as I dont have a windows machine now to check

kaidohallik: My test results where this was working:
* in Windows cmd.exe produced `.\mvnw` and this works
* in Windows PowerShell produced `.\mvnw` and this works (in PowerShell also `./mvnw` works)
* in Windows Ubuntu bash produced `./mvnw` and this works
* in Ubuntu produced `./mvnw` and this works

And in this case this was not working:
* in Windows git bash produced `.\mvnw` and this is not working, in Windows git bash are working `./mvnw` and `.\\mvnw`

So this solution is incomplete. Currently I have no idea how to achieve correct result. Possible solutions seem to be:
1. close this PR in favor to Windows git bash
2. merge this PR in favor to Windows cmd.exe

deepu105:
Or we can just log something like below
"Run './mvnw' ('mvnw' if using windows cmd)"

Thanks & regards,
Deepu

kaidohallik:
Yes, good idea, change done.

jdubois:
- Couldn't we generate the command depending on the current OS?
- Personally 'cmd' isn't clear to me -> "command line" maybe?

kaidohallik:
Now showing starting instruction comment only in Windows. Changed cmd to cmd.exe.

kaidohallik:
changed "cmd.exe" to "Windows Command Prompt" and rebased

Issue #5045, opened 2017-02-01
Spring Boot 1.5.1 released


cbornet:
https://spring.io/blog/2017/01/30/spring-boot-1-5-1-released

cbornet:
with Kafka autoconfiguration

jdubois:
Yes, but we'll do it after the 4.0 release, as it might break stuff.

- For Kafka I think they just included Spring Kafka, so it shouldn't change
much for us
- For the logger endpoint, it looks like they copied us (again !), so we'll
need to refactor and remove our code


Le 1 févr. 2017 9:08 AM, "Christophe Bornet" <notifications@github.com> a
écrit :

cbornet:
I'm glad when they copy us : less is more 😄

gzsombor:
It would be nice to have a more advanced log endpoint, where you could actually see the logs :) Sure, there are log servers, with all the bells and whistles - but sometimes you don't need that much, just to look into log, to know what happened 10 minutes ago :)

jdubois:
@gzsombor  Yes, we would need some kind of in-memory data store for that. We could also have the same thing for the security audits, to improve log in performance. I would love to have some kind of simple hashmap or list, which would just store the last 10,000 events, for example.

erikkemperman:
For what it's worth -- I'm playing around with a branch that has SB1.5 (and some other upgrades). It's actually not that big of a deal, as far as I can see so far.

Background is I was trying to update my JH app to 3.12.2 (after ~0.5 year hiatus) but ran into some difficulties, which seem to be due to slightly unaligned versions of hibernate components, resulting in a helper class (FieldInterceptionHelper) not being available. It seems this is only a problem when running from a war file, and then only with Tomcat. But unfortunately I really need precisely that combination.

I've found some issues on SB which look like they are the same thing and apparently 1.5 makes this go away.

Currently I'm having some trouble with hibernate ("missing sequence") after  generating some simple entities -- but that also seems to be happening with current master? I'll see if I can get that resolved quickly, or will make a separate ticket for it.

Anyway, if anyone is interested to go SB1.5 in the near future let me know, I'd be happy to submit what I have here.

jdubois:
No, you should not have any "missing sequence" with master.

Le 1 févr. 2017 1:06 PM, "Erik Kemperman" <notifications@github.com> a
écrit :

erikkemperman:
[...]

PR #5146, opened 2017-02-08
[WIP] Spring Boot 1.5.1

erikkemperman:
As per https://github.com/jhipster/generator-jhipster/issues/5045#issuecomment-278023167 -- please see questions/remarks in that thread also.

Some additions:

- There is a counterpart to this PR over at the serverside JHipster lib, [here](https://github.com/jhipster/jhipster/pull/7). I was wondering if it might make sense to deduplicate some currently redundant stuff between the generator and the lib, for instance by having the lib as parent of the generated project. But I'm not completely sure how that might work with gradle; I suppose one might look into the spring-boot-gradle plugin, which [advertises](https://github.com/spring-gradle-plugins/dependency-management-plugin) that it can import POMs (of course that's how it interacts with `spring-boot-dependencies` too!). If something like that works out, I guess it might even be possible to have all version properties in a single place, instead of duplicated for gradle and maven.

- I was able to get rid of the awkward dependency override that I mentioned earlier, on `spring-cloud-netflix`, by going forward and including `spring-cloud` Camden SR5. It's probably a good idea to keep an eye on `spring-cloud-stream` Chelsea, which is currently at M1.

- Thanks to @cbornet who pointed me to an earlier complication with `netty-handler` which I was now able to remove.

- I found kind of a cute [trick](https://github.com/jhipster/generator-jhipster/pull/5146/files#diff-5595eed51dd01e38e024547a86042a81R199) to effectively import version strings from the `spring-boot-dependencies` POM into the gradle build. But after discovering that I didn't double back to see if it could be applied more widely.

- In general I was able to remove some version strings from JH code. Mostly this is to align with `spring-boot-dependencies` implicitly, but also I found out that -- provided we use the same property names --  it is sufficient to override the version property and we don't have to include the version with the dependency itself. As an example, see what I did with `assertj` [here](https://github.com/jhipster/generator-jhipster/pull/5146/files#diff-4ac12e0abae2ba3bd72788629c00c4abR40) and [here](https://github.com/jhipster/generator-jhipster/pull/5146/files#diff-4ac12e0abae2ba3bd72788629c00c4abL458).

Still to do:

- [ ] There's a failing test (app-ng2-jwt)  that I wasn't able to quickly fix (or spend a lot of time on).
- [ ] SB1.5 has autoconfiguration for Kafka, but I don't know much about it and haven't looked into it now.

### DIFF
--- a/generators/server/templates/src/test/java/package/cucumber/stepdefs/_StepDefs.java
+++ b/generators/server/templates/src/test/java/package/cucumber/stepdefs/_StepDefs.java
@@ -2,13 +2,15 @@ package <%=packageName%>.cucumber.stepdefs;
 
 import <%=packageName%>.<%= mainClass %>;
 
-import org.springframework.boot.test.SpringApplicationContextLoader;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.web.WebAppConfiguration;
 import org.springframework.test.web.servlet.ResultActions;
 
+import org.springframework.boot.test.context.SpringBootTest;
+
 @WebAppConfiguration
-@ContextConfiguration(classes = <%= mainClass %>.class, loader = SpringApplicationContextLoader.class)
+@SpringBootTest
+@ContextConfiguration(classes = <%= mainClass %>.class)
 public abstract class StepDefs {
 
     protected ResultActions actions;
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -44,9 +44,7 @@ dependencyManagement {
     mavenBom 'org.springframework.cloud:spring-cloud-stream-dependencies:' + spring_cloud_stream_version
     <%_ } _%>
     <%_ if (serviceDiscoveryType) { _%>
-    mavenBom 'org.springframework.cloud:spring-cloud-dependencies:' + spring_cloud_version {
-        bomProperty 'spring-cloud-netflix-core.version' '1.2.5.RELEASE'
-    }
+    mavenBom 'org.springframework.cloud:spring-cloud-dependencies:' + spring_cloud_version
     <%_ } _%>
   }
 }
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -246,7 +246,7 @@ dependencies {
     <%_ } _%>
     <%_ if (databaseType === 'sql') { _%>
     compile "org.hibernate:hibernate-envers"
-    compile "org.hibernate:hibernate-validator:${hibernate_validator_version}"
+    compile "org.hibernate:hibernate-validator"
     compile ("org.liquibase:liquibase-core") {
         exclude(module: 'jetty-servlet')
     }
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -392,9 +392,6 @@ dependencies {
     testCompile "org.hamcrest:hamcrest-library"
     <%_ if (gatlingTests) { _%>
     testCompile "io.gatling.highcharts:gatling-charts-highcharts:${gatling_version}"
-    <%_ if (databaseType === 'cassandra') { _%>
-    compile "io.netty:netty-handler:${netty_handler_version}"
-    <%_ } _%>
     <%_ } _%>
     <%_ if (devDatabaseType != 'h2Disk' && devDatabaseType != 'h2Memory') { _%>
     testCompile "com.h2database:h2"
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -44,7 +44,9 @@ dependencyManagement {
     mavenBom 'org.springframework.cloud:spring-cloud-stream-dependencies:' + spring_cloud_stream_version
     <%_ } _%>
     <%_ if (serviceDiscoveryType) { _%>
-    mavenBom 'org.springframework.cloud:spring-cloud-dependencies:' + spring_cloud_version
+    mavenBom 'org.springframework.cloud:spring-cloud-dependencies:' + spring_cloud_version {
+        bomProperty 'spring-cloud-netflix-core.version' '1.2.5.RELEASE'
+    }
     <%_ } _%>
   }
 }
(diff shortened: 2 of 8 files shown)


PierreBesson:
* having the lib as parent of the generated project : This was already suggested in #4724

erikkemperman:
Ah, thanks,  I had not noticed that issue before. I might play around with lib as parent, or as sibling with dependency management, when I find some more time.

It seems that some more Travis tests are failing than with the exact same code (at 50a1585) on my own fork... For what it's worth, there is only one failing test here: https://travis-ci.org/erikkemperman/generator-jhipster/builds/199545726

erikkemperman:
I added another commit to get rid of an unused dependency (I needed it in my app but turns out JH doesn't use `@Inject` anymore). While I was at it, I rebased onto the current tip of master. I'll leave it alone now, so as not to get in the way of proper review :-)

erikkemperman:
FYI I have rebased to resolve merge conflicts. Has anyone else noticed that yarnpkg.com is unreliable and causing Travis builds to fail, or is it just me?

jdubois:
@erikkemperman yes Yarn installation sometimes fail

generators/server/templates/src/test/java/package/cucumber/stepdefs/_StepDefs.java
```diff
@@ -2,13 +2,15 @@
...
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.web.WebAppConfiguration;
 import org.springframework.test.web.servlet.ResultActions;
 
+import org.springframework.boot.test.context.SpringBootTest;
+
 @WebAppConfiguration
-@ContextConfiguration(classes = <%= mainClass %>.class, loader = SpringApplicationContextLoader.class)
+@SpringBootTest
+@ContextConfiguration(classes = <%= mainClass %>.class)
```
tan9: Or we can simply use `@SpringBootTest(classes = <%= mainClass %>.class)` ?

erikkemperman: @tan9 I tried this, or at least if I understand your suggestion correctly, but I get NPEs from `mvn test` if I omit the `@ContextConfiguration` annotation. You're right that we could move the `classes` argument to `@SpringBootTest` as per [the docs](http://docs.spring.io/spring-boot/docs/1.5.1.RELEASE/api/org/springframework/boot/test/context/SpringBootTest.html#classes--) but would it actually matter if we have to have both annotations anyway?

jdubois:
Thanks for all the hard work here, I'll review this tomorrow

erikkemperman:
@jdubois I might find some time to rebase and possibly finetune this a bit today. Ideally I'd like to get that one failing test out of the way. Well, the consistently failing test, flaky `yarn` seems to spoil tests randomly quite often... I think I'd prefer to keep `npm` as default, but maybe that's just me.

generators/server/templates/src/test/java/package/cucumber/stepdefs/_StepDefs.java
erikkemperman: @tan9 Looking a bit further into this, I noticed that a lot of other `@SpringBootTest`-annotated classes do _not_ have a `@ContextConfiguration`. It turns out that it is required for this class, though, due to the `@WebAppConfiguration`, see [here](http://docs.spring.io/spring/docs/4.3.6.RELEASE/javadoc-api/org/springframework/test/context/web/WebAppConfiguration.html).

erikkemperman:
Rebased onto latest master, going to have to leave the failing test (`app-ng2-jwt`) in place due to lack of time... Sorry about that!

generators/server/templates/src/test/java/package/cucumber/stepdefs/_StepDefs.java
tan9: @erikkemperman This is a missing feature of Cucumber, we can keep tracking this issue: https://github.com/cucumber/cucumber-jvm/issues/1061

jdubois:
The JWT issue looks like a Travis glitch, I'm re-starting it

erikkemperman:
It might have been Travis (well, yarnpkg) during the last build, but actually the app-ng2-jwt sample has been failing consistently.

jdubois:
I'm going to give this a try, but without the other library upgrades - let's to this in several steps, it will be easier to migrate

jdubois:
Sorry @erikkemperman I'm starting to understand you modified our variable names to match those from Spring Boot -> I was wrong in my comments!!!
Then, what I'd like to do is:

- First, upgrade de Spring Boot, Spring Cloud, Spring Cloud Stream dependencies
- Then, see how we can upgrade all other third-party libs one by one, as each of them has its own issues, and we need to be able to test (and revert) those changes easily

I'm having a try at the first step

erikkemperman:
@jdubois Thanks for taking a look Julien! Sure, whatever you think is the best approach. For what it's worth, though, the dependency upgrades that are not aligned with SB should be all minor (patch-)releases, if I remember correctly.

I notice that the branch needs a rebase and won't merge at the moment, but I think it's probably better to leave this is as it is just for the moment to not interfere with review. Let me know if you'd like me to rebase onto the tip of master!

jdubois:
Thanks @erikkemperman - in the meantime I discovered some new issue with Spring Data Cassandra, which we don't use, and that they have "upgraded". I have no idea what they did, but they somehow broke the `CassandraAutoConfiguration`, which I coded myself (and contributed to Spring Boot). I'm already fed up with this :-(

jdubois:
OK I got everything "working". I had so many issues, I wonder what else we will find!!!!
@erikkemperman thank you, thank you, thank you! I'm going to do my own commit as I don't want to have the other libraries upgrades, but of course we'll do it eventually.

erikkemperman:
FYI In a separate branch, I have rebased onto master to resolve merge conflicts, and squashed my changes in a single commit, [here](https://github.com/erikkemperman/generator-jhipster/tree/springboot151-rb). Travis results are [here](https://travis-ci.org/erikkemperman/generator-jhipster/builds/203445104).

jdubois:
Argh sorry, I just commited :-(

erikkemperman:
@jdubois Oops, looks like we're cross-posting. Never mind, disregard that last post :-) Thanks for review!!

jdubois:
@erikkemperman I credited you in the commit ( https://github.com/jhipster/generator-jhipster/commit/014422d5410349e607f7b2bafad9bb0f27c81df4 ) but as you can see I'm only doing something as simple as possible, just migrating. Then there is still the full job of upgrading the other libraries.
But there are so many side effects, we can only do that in small steps, so it's easy to track and revert, if needed.

erikkemperman:
@jdubois Sure, that makes a lot of sense. I guess you could take a look at that branch I linked just now to see what newer lib versions are available, but it will require some work to make a bunch of distinct PRs out of it... I'm not going to have time for that any time soon I think.

PR #5150, opened 2017-02-08
update jacoco to latest version

atomfrede:
Was there a specific reason why we didn't update jacoco for quite a long time (especially in gradle)?


erikkemperman:
@atomfrede I bumped this to .8 in https://github.com/jhipster/generator-jhipster/pull/5146, but I guess they've made a .9 release since I looked :-)

jdubois:
No reason, just nobody worked on this

PR #5162, opened 2017-02-09
Better fix for #5134

gmarziou:
As suggested by @deepu105


jdubois:
That's going to be hard to test!! I hate those regexps :-)

gmarziou:
:-) and I simplified it.

deepu105:
Not that hard if you use something like this http://regexr.com/

Issue #4955, opened 2017-01-24
Add support for the JHipster Registry in monolith applications


jdubois:
The JHipster Registry is only available for microservices, but there is some interest in using it in a monolith:

- It allows to scale the monolith if Hazelcast is used as a second level cache (that's the only way for the cache to find the other instances on-the-fly)
- It can help to configure the monolith (as it comes with a Spring Cloud Config server

I personally need this for the option (Hazelcast clustering when the app scales with Docker), and it's probably very easy to do (as it already works for the gateway)

pascalgrimaud:
I like the idea, but for me, it's a lot of work, depending how far you want to go.
Do you want this for JHipster 4?

jdubois:
No I'm respecting my own "code freeze" for JHipster 4, so it would be for a later version.

PierreBesson:
It is not a lot of work at all. When working on Consul, I have actually already organised the code to be able to do that in the future...
I will try to do it when I have the time.

pascalgrimaud:
Do you mean :
1) rework the current jhipster-registry project, for being able to use with monolith
2) or add additional question to generate a fresh registry on generator-jhipster

I thought about the 2 so I think I misunderstood :smile:

PierreBesson:
@pascalgrimaud I think Julien mistyped "using it in a monolith" he certainly meant "using it **with** a monolith"

pascalgrimaud:
@PierreBesson : I was dreaming about a JHipster Registry with Angular2 :stuck_out_tongue:

jdubois:
I just want to connect the monolith to the existing Registry, like we do
for microservices.

Only way to scale with Hazelcast !

Le 25 janv. 2017 11:21 AM, "Pascal Grimaud" <notifications@github.com> a
écrit :

@PierreBesson <https://github.com/PierreBesson> : I was dreaming about a
JHipster Registry with Angular2 😛

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jhipster/generator-jhipster/issues/4955#issuecomment-275071035>,
or mute the thread

deepu105:
@jdubois why do you say its the only way? doesnt hazelcast automatically create and join clusters if you have app nodes running in the same IP segment, you could also blacklist/whitelist to specific IPs. I have used that and it works very well and out of the box previously. It worked while running using inbuilt Tomcat and when deployed to JBoss containers as well when I used

[...]

PR #5193, opened 2017-02-13
skip docker-compose registry config for monoliths without service discovery

ruddell:
When a monolith's serviceDiscoveryType was false or missing, the docker-compose generator would assume you needed a registry.  This fixes it for the kubernetes subgen as well since it uses the same prompts.

This is the issue discussed at the bottom in https://github.com/jhipster/generator-jhipster/issues/4955#issuecomment-278998120 @PierreBesson @jdubois

### DIFF
--- a/generators/docker-compose/prompts.js
+++ b/generators/docker-compose/prompts.js
@@ -204,18 +204,19 @@ function askForMonitoring() {
 }
 
 function askForServiceDiscovery() {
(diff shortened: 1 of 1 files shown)


jdubois:
I still don't understand exactly how you get that error, but I see the point and of course I trust you on this.

PierreBesson:
Thanks @rudell. This was bothering me...

Issue #5196, opened 2017-02-14
build error:The system cannot find the path specified.


guayeguaye:
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

##### **Overview of the issue**
Running `yo jhipster`  but build failed
The message is 
`WARNING in ./src/main/webapp/content/images/hipster2x.png
DEPRECATED. Configure gifsicle's interlaced option in it's own options. (gifsicl
e.interlaced)
 @ ./~/css-loader!./src/main/webapp/app/home/home.css 6:896-941
 @ ./src/main/webapp/app/home/home.css
 @ ./src/main/webapp/app/home/home.component.ts
 @ ./src/main/webapp/app/home/index.ts
 @ ./src/main/webapp/app/home/home.module.ts
 @ ./src/main/webapp/app/app.module.ts
 @ ./src/main/webapp/app/app.main.ts

WARNING in ./src/main/webapp/content/images/hipster2x.png
DEPRECATED. Configure optipng's optimizationLevel option in it's own options. (o
ptipng.optimizationLevel)
 @ ./~/css-loader!./src/main/webapp/app/home/home.css 6:896-941
 @ ./src/main/webapp/app/home/home.css
 @ ./src/main/webapp/app/home/home.component.ts
 @ ./src/main/webapp/app/home/index.ts
 @ ./src/main/webapp/app/home/home.module.ts
 @ ./src/main/webapp/app/app.module.ts
 @ ./src/main/webapp/app/app.main.ts

WARNING in ./src/main/webapp/content/images/hipster.png
DEPRECATED. Configure gifsicle's interlaced option in it's own options. (gifsicl
e.interlaced)
 @ ./~/css-loader!./src/main/webapp/app/home/home.css 6:307-350
 @ ./src/main/webapp/app/home/home.css
 @ ./src/main/webapp/app/home/home.component.ts
 @ ./src/main/webapp/app/home/index.ts
 @ ./src/main/webapp/app/home/home.module.ts
 @ ./src/main/webapp/app/app.module.ts
 @ ./src/main/webapp/app/app.main.ts

WARNING in ./src/main/webapp/content/images/hipster.png
DEPRECATED. Configure optipng's optimizationLevel option in it's own options. (o
ptipng.optimizationLevel)
 @ ./~/css-loader!./src/main/webapp/app/home/home.css 6:307-350
 @ ./src/main/webapp/app/home/home.css
[...]

PR #5200, opened 2017-02-14
img-loader downgrade to fix warnings

flaviencathala:
Fix for warnings (see https://github.com/jhipster/generator-jhipster/issues/5196)

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -56,7 +56,7 @@
     "file-loader": "0.10.0",
     "generator-jhipster": "<%= packagejs.version %>",
     "html-webpack-plugin": "2.28.0",
-    "image-webpack-loader": "3.2.0",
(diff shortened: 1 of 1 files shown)


deepu105:
@flaviencathala downgrading cannot be a permanent fix. as we have to keep libs upto date so I would prefer to have latest lib and fix the deprecation as per the libs suggestion. We can temporarily downgrade if there is some bug but not for deprecation. Hence please revert this

deepu105:
here is what should be done to fix it https://github.com/tcoopman/image-webpack-loader/issues/68

jdubois:
@deepu105 I agree for this specific issue, but what about the Windows issue? If downgrading solves the Windows issue, we don't have another solution for the moment

deepu105:
@jdubois I'm not sure if its related but let me test this in windows later today

Issue #5052, opened 2017-02-01
Anonymous user is send whereas it should not


nicoraynaud:
##### **Overview of the issue**

The pagination elements generated for entities are not working (tested only with ng2)

##### **Reproduce the error**

- Create an app
- Generate an entity with pagination "entity"
- create many elements (or change pagination default size)
- try to use the pagination buttons in the generated component "entity.component.html"

=> The buttons don't work as expected.

In comparison, the pagination elements in the Audit admin view work ok.

##### **Suggest a Fix**

I think the paginated elements should be changed in the entity generator from

```
<div class="text-xs-center" *ngIf="jobs && jobs.length">
            <jhi-item-count [page]="page" [total]="queryCount" [itemsPerPage]="itemsPerPage"></jhi-item-count>
            <ngb-pagination [collectionSize]="totalItems" [(page)]="page" (pageChange)="loadPage(page)"></ngb-pagination>
        </div>
```
to
```
<div class="text-xs-center" *ngIf="jobs && jobs.length">
            <jhi-item-count [page]="page" [total]="queryCount" [itemsPerPage]="itemsPerPage"></jhi-item-count>
            <ngb-pagination [collectionSize]="totalItems" [(page)]="page" [pageSize]="itemsPerPage"  (pageChange)="loadPage(page)"></ngb-pagination>
        </div>
```

##### **JHipster Version(s)**

JHipster v.3.12.2

##### **JHipster configuration**
.yo-rc.json

```
{                                                              
  "generator-jhipster": {                                      
    "jhipsterVersion": "3.12.2",                               
    "baseName": "optone",                                      
    "packageName": "nc.opt.template",                          
    "packageFolder": "nc/opt/template",                        
    "serverPort": "8080",                                      
    "authenticationType": "jwt",                               
    "hibernateCache": "ehcache",                               
    "clusteredHttpSession": false,                             
    "websocket": false,                                        
    "databaseType": "sql",                                     
    "devDatabaseType": "h2Disk",                               
    "prodDatabaseType": "postgresql",                          
    "searchEngine": false,                                     
    "messageBroker": false,                                    
    "buildTool": "maven",                                      
    "enableSocialSignIn": false,                               
    "jwtSecretKey": "70a20309af8c06c2652019d9bf7570c92df7de9e",
    "useSass": true,                                           
    "clientPackageManager": "yarn",                            
    "applicationType": "monolith",                             
    "clientFramework": "angular2",                             
[...]

PR #5203, opened 2017-02-14
Remove anonymous user filter in user management

flaviencathala:
https://github.com/jhipster/generator-jhipster/issues/5052

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.ts
@@ -129,17 +129,9 @@ export class UserMgmtComponent implements OnInit, OnDestroy {
 
     <%_ } _%>
     private onSuccess(data, headers) {
-        // hide anonymous user from user management: it's a required user for Spring Security
-        let hiddenUsersSize = 0;
-        for (let i in data) {
-            if (data[i]['login'] === 'anonymoususer') {
-                data.splice(i, 1);
-                hiddenUsersSize++;
-            }
-        }
         <%_ if (databaseType !== 'cassandra') { _%>
         this.links = this.parseLinks.parse(headers.get('link'));
-        this.totalItems = headers.get('X-Total-Count') - hiddenUsersSize;
+        this.totalItems = headers.get('X-Total-Count');
         this.queryCount = this.totalItems;
         <%_ } _%>
         this.users = data;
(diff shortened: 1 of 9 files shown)


deepu105:
what about filtering it on the backend?

flaviencathala:
@deepu105 I don't really know how to make a clean fix for this on the backend, maybe @jdubois have an idea on how to do a clean filter ?

jdubois:
Yes we need to filter it in the backend -> this is just a "where" clause to remove the "anonymous" user when you do a "findAll", it shouldn't be very complex

hdurix:
@flaviencathala I can do it on the backend this weekend if you want.

jdubois:
@hdurix +1 yes do it, @flaviencathala is more a front-end person, that will be more efficient

hdurix:
I have added a PR on your [branch](https://github.com/flaviencathala/generator-jhipster/pull/4) @flaviencathala.

However, the anonymous user is not filtered with MongoDB. There is a case-sensitive issue with this database, both on master and on my PR: the anonymous user is created with an uppercase letter (only with MongoDB) and the current application works only with lowercase.

You can see on master branch:

- The check for hiding anonymous user is done with `anonymoususer` (see https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.controller.js#L59)
- The anonymous user is created with `anonymousUser` (see https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/dbmigrations/_InitialSetupMigration.java#L72)
- The audit event is not inserted if the user login is `anonymoususer` (see https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/repository/_CustomAuditEventRepository.java#L26)

I can do a quick fix on this PR by ignoring case (something like `findAllByLoginIgnoreCaseNot` instead of `findAllByLoginNot`) but I reckon the problem is deeper: some unwanted audit events might be inserted in database.
Moreover I think it's dangerous to just change the case in the initial setup migration script because it could break JHipster upgrade mechanism (I don't know if there is some hash file check like liquibase `databaseChangeLog`).

Anyway, maybe it would be better to create a new issue for that?

jdubois:
@hdurix the correct name is `anonymoususer` (all lower case), as this is a required name for Spring Security.
So the solution for me is to change the set up script in MongoBee (the one you listed, at https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/dbmigrations/_InitialSetupMigration.java#L72 ). I don't think there is a checksum here (there is one with Liquibase, I don't think MongoBee does the same), but anyway it's a bug so let's change this (and if people have existing databases, it will be annoying but they will need to migrate them).

-> @hdurix  can you do it? And then I'll check with @flaviencathala to merge everything tomorrow

hdurix:
Yes no problem I'll do it today.

hdurix:
I added a commit on the branch to correct this.

flaviencathala:
I think this is ready to be merged !

PR #5206, opened 2017-02-14
Add HTTP/2 Support and instructions for enabling TLS

mraible:
After this patch, if you enable TLS[1], HTTP/2 will be enabled for requests. If you don't enable TLS, HTTP/1.1 will still be used.

NOTE: I tried to use Apache with mod_proxy and mod_ssl (pointing to 8080), and found it doesn’t work. Deploying to Cloud Foundry or Heroku (and enabling SSL) doesn't work either.

[1] See [Wikipedia](https://en.wikipedia.org/wiki/Transport_Layer_Security) to see why I call it TLS instead of SSL.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -164,6 +164,7 @@
         <spring-social-twitter.version>1.1.2.RELEASE</spring-social-twitter.version>
         <%_ } _%>
         <springfox.version>2.6.1</springfox.version>
+        <undertow.version>1.4.10.Final</undertow.version>
         <%_ if (clientPackageManager === 'yarn') { _%>
         <yarn.version>v0.19.1</yarn.version>
         <%_ } _%>
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -90,6 +90,7 @@ mapstruct_version=1.1.0.Final
 <%_ if (enableSocialSignIn) { _%>
 spring_social_google_version=1.0.0.RELEASE
 <%_ } _%>
+undertow_version=1.4.10.Final
 
 ## below are some of the gradle performance improvement settings that can be used as required, these are not enabled by default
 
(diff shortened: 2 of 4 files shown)


mraible:
NOTE: Hard-coding an upgrade of Undertow can likely be removed when https://github.com/jhipster/generator-jhipster/pull/5146 is added.

Also - proof it works is below.

screen shot 2017-02-14 at 10 57 30 am

deepu105:
So with this patch does it work on CF, apache and heroku?

Thanks & regards,
Deepu

mraible:
@deepu105 If you enable HTTPS for Spring Boot, yes. 

Since Apache, CF, and Heroku all use web servers (or proxies) in front of Spring Boot - they have to support HTTP/2.

deepu105:
Ok Its fine as long as this doesn't break anything in any container or
cloud

Thanks & regards,
Deepu

On 14 Feb 2017 11:53 p.m., "Matt Raible" <notifications@github.com> wrote:

@deepu105 <https://github.com/deepu105> If you enable HTTPS for Spring
Boot, yes.

Since Apache, CF, and Heroku all use web servers (or proxies) in front of
Spring Boot - they have to support HTTP/2.

—

<https://github.com/jhipster/generator-jhipster/pull/5206#issuecomment-279862711>,
or mute the thread

jdubois:
Yes that's great and that's the future, but we must also :

- check how this works with Zuul
- document it on the website (not sure where!)

Le 15 févr. 2017 8:37 AM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

> Ok Its fine as long as this doesn't break anything in any container or
> cloud
>
> Thanks & regards,
> Deepu
>

generators/server/templates/_pom.xml
```diff
@@ -164,6 +164,7 @@
         <spring-social-twitter.version>1.1.2.RELEASE</spring-social-twitter.version>
         <%_ } _%>
         <springfox.version>2.6.1</springfox.version>
+        <undertow.version>1.4.10.Final</undertow.version>
```
mraible: Added.

Issue #5172, opened 2017-02-11
Kubernetes generator generates YAML file w/ `imagePullPolicy: IfNotPresent`


saturnism:
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

##### **Overview of the issue**

Kubernetes generator currently generates YAML file with `imagePullPolicy: IfNotPresent`. This may be great for the first deployment w/o having to pull from a remote registry. However, if you rebuild the image, it won't automatically be repulled from the registry either.

If we remove this - then every time you redeploy, it'll try to pull the latest image. This is also not desired in production.

The better practice is to tag the image w/ a version (potentially githash, or a build number, but not the artifact version), and deploy the tagged images instead.

##### **Motivation for or Use Case**

To avoid unnecessary image pulls, but also don't want to pull for latest image every time.

##### **Reproduce the error**

`yo jhipster:kubernetes`

##### **Related issues**
N/A

##### **Suggest a Fix**

Tag the image w/ a version (potentially githash, or a build number, but not the artifact version), and deploy the tagged images instead.

##### **JHipster Version(s)**

v4.0.2

##### **JHipster configuration**
N/A

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**
Irrelevant

##### **Browsers and Operating System**
All browsers/OS.

- [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
Thanks @saturnism for reviewing our code!

For this ticket, I'm working on it:
- adding a new question to ask the tag
- remove imagePullPolicy

jdubois:
@pascalgrimaud I see you referenced this in a closed PR, should we also close this ticket?

pascalgrimaud:
@jdubois : I proposed a very simple solution which resolved this ticket but @PierreBesson wasn't agree with my solution. We discussed about it but I dont know if Pierre worked on it since the last time we met.

pascalgrimaud:
@jdubois : in details, my solution added a new question to the user, and asked him which tag hz wants:
- if he let the default answer: the generator will use latest, exactly same as currently
- if he choose to put a specific tag, for example v1.2.3, it would be the same tag for all microservices, but no repull from Docker Hub if the image already exists, as suggested by Ray. Then, for me, the user can edit these configuration, as it's hard to code something to ask every version for every microservices

PierreBesson:
[...]

Issue #5174, opened 2017-02-11
Kubernetes generator should expose Monolith and API gateway on port 80, or use Ingress


saturnism:
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

##### **Overview of the issue**

When the service is created for the API gateway, it's exposed via port 8080.

##### **Motivation for or Use Case**
Usually users expect the web frontend to be on port 80.

##### **Reproduce the error**
`yo jhipster:kubernetes`

##### **Related issues**
N/A

##### **Suggest a Fix**
In the service YAML, add both port and targetPort, so that port = 80, and targetPort = 8080.
Alternatively, add an Ingress yaml to create a HTTP ingress.

##### **JHipster Version(s)**
v4.0.2

##### **JHipster configuration**
N/A

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**
Irrelevant

##### **Browsers and Operating System**
Irrelevant

- [x] Checking this box is mandatory (this is just to show you read everything)

PierreBesson:
I think we should do both : exposing on port 80 as default and add a question for ingress.

pascalgrimaud:
@PierreBesson : It's already done for port 80 on my local, I'm currently testing it to confirm it works well. I will let you add ingress if you want

deepu105:
Guys any update on this?

pascalgrimaud:
@deepu105 : it was already done for port 80 (instead of ingress) in this PR https://github.com/jhipster/generator-jhipster/pull/5210 (which resolved 2 tickets)
but was closed because of this comment https://github.com/jhipster/generator-jhipster/issues/5172#issuecomment-297818067

jdubois:
I'm not totally sure this is done, but this has been stucked for a very long time, with no work being done, so I'm closing this.

xetys:
I would suggest adding an ingress question. I'm always copy-pasting that from other sources, as we don't provide that option.

Is it ok, if I re-open that with that update?

pascalgrimaud:
I'm ok, if someone works on it

PR #5210, opened 2017-02-15
Kubernetes improvements

pascalgrimaud:
- use Docker tag instead of imagePullPolicy -> fix https://github.com/jhipster/generator-jhipster/issues/5172
- add port and targetPort -> related to https://github.com/jhipster/generator-jhipster/issues/5174, I don't close this as @PierreBesson will probably work on adding ingress

### DIFF
--- a/generators/kubernetes/index.js
+++ b/generators/kubernetes/index.js
@@ -94,6 +94,7 @@ module.exports = KubernetesGenerator.extend({
             this.adminPassword = this.config.get('adminPassword');
             this.jwtSecretKey = this.config.get('jwtSecretKey');
             this.dockerRepositoryName = this.config.get('dockerRepositoryName');
+            this.dockerTag = this.config.get('dockerTag');
             this.dockerPushCommand = this.config.get('dockerPushCommand');
             this.kubernetesNamespace = this.config.get('kubernetesNamespace');
 
@@ -148,6 +149,7 @@ module.exports = KubernetesGenerator.extend({
         askForAdminPassword: prompts.askForAdminPassword,
         askForKubernetesNamespace: prompts.askForKubernetesNamespace,
         askForDockerRepositoryName: prompts.askForDockerRepositoryName,
+        askForDockerTag: prompts.askForDockerTag,
         askForDockerPushCommand: prompts.askForDockerPushCommand
     },
 
@@ -209,6 +211,7 @@ module.exports = KubernetesGenerator.extend({
             this.config.set('serviceDiscoveryType', this.serviceDiscoveryType);
             this.config.set('jwtSecretKey', this.jwtSecretKey);
             this.config.set('dockerRepositoryName', this.dockerRepositoryName);
+            this.config.set('dockerTag', this.dockerTag);
             this.config.set('dockerPushCommand', this.dockerPushCommand);
             this.config.set('kubernetesNamespace', this.kubernetesNamespace);
         }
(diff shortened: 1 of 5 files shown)


PierreBesson:
@pascalgrimaud I don't think it makes much sense to ask the tag in a question like you do. Are you sure about this ? What would you do if different microservices have different version numbers ?

What we could do is configure the maven/gradle docker build to set the version from the artifact version or git commit. Then when generating the k8s configuration we ask for the tag for each microservices.

pascalgrimaud:
@PierreBesson : good point about microservices with different version numbers. I'm not sure about adding a question about the tag, it was simply the easiest and fastest solution.

I didn't use the artifact version because of the comment from Saturnism:
> Tag the image w/ a version (potentially githash, or a build number, but not the artifact version), and deploy the tagged images instead.

The idea was to generate some Kubernetes files and people would custom with what they want/need.
If you have another idea, don't hesitate @PierreBesson !

PierreBesson:
I think it would be OK if the tag question was asked for every microservice. Can you do this for now ?

pascalgrimaud:
right now -> I can't, because of day work :)
I try to do it tonight or tomorrow morning

pascalgrimaud:
After 1 day thinking about this, here my opinion:
- in real production, the user will probably use `latest` as tag version, and the developer will update their image to the registry. So the K8s cluster will always pull the latest image. I can't imagine an architecture with 20 microservices and the DevOps updating every yaml file, to adapt to every version, every new release. This full configuration should stay stable with `latest`

- on demo, specially with Minikube, we should use tag (`demo` or `test`), to avoid pulling the image from a registry. So we can build, push, tag directly to minikube (using `eval $(minikube docker-env)`). I don't think we will put a specific version to every microservices

So for me, this PR is simple: ask tag version and use the same for all.
It will be overkill to ask a question tag for every microservices.
@PierreBesson : what do you think?

PierreBesson:
@pascal As someone who is part of team actually managing a production kubernetes cluster with over 20 microservices I'm the right person to answer you !

Using the "latest" tag in prod is certainly a bad idea. What you usually do is that you have a separate kubernetes clusters for staging (or at least a specific namespace) where you test different combinations between versions. You run your integration tests and once it's passing you can promote those tags to production.
In the ideal world this happens automatically.

When you say "updating every yaml file" this is basically what you need to do. Yes this is tedious but this is very important so you that you can upgrade or rollback to specific versions.

Of course you can use templating of yaml files or something like helm/kpm to manage versions in manifest files but really "latest" should not be used.

So for me either we ask the tag question for every microservice or we put the maven artifact version as the defaut tag and extract the tag from properties.

pascalgrimaud:
Ok, I understood and agree your point of view about not using latest

> When you say "updating every yaml file" this is basically what you need to do. Yes this is tedious but this 
is very important so you that you can upgrade or rollback to specific versions.

I wonder if it will be usefull to code and maintain our k8s sub generator to ask a tag for every microservice ?
- Will the DevOps team use our sub generator to update / changing their version -> I'm not sure
- Or will they simply use their own tools to edit the generated yaml files ? -> I think for this options, as the initial yaml files we provide will probably be customized

@PierreBesson : your turn :smile:

pascalgrimaud:
@PierreBesson : so, after spending 2 hours on looping for prompting a question for every microservices, it's harder than expected... as I'm not a NodeJS developer :(
Should we spend more time, for something the user can simply edit ?

@jdubois / @deepu105 : do you have an opinion about this?

PierreBesson:
@pascalgrimmaud,

Don't worry asking a question for each microservice. I'm having an idea now. That we can put thz version as a default docker tag and extract it from the build.properties file in the target/build directory. Just give me some time to experiment with it.

mraible:
I'd love to demo how this works in a screencast. I was able to do a successful deploy to Google Cloud today and would love to document/record a screencast. Running on port 80 seems like an essential feature.

pascalgrimaud:
rebase done to keep this PR up-to-date, after the refactor by Deepu

pascalgrimaud:
@PierreBesson : this PR has been opened for a long time. What should we do ? 
- close this and you take the lead on the opened ticket ? 
- or merge this, as for me, it's better than what we currently have

PierreBesson:
@pascal If we ask the docker tag question we should ask it for every service, then that would be ok. Would you agree to do this change ?

pascalgrimaud:
As already said, I tried to code this but it's not as easy, and I really think it's not worth doing this, because it can be done manually very fast: it's our policy 2.

So I'm closing this, let stay like this for now -> with latest tag

Issue #5164, opened 2017-02-10
bootRun fails when home directory contains spaces (Windows only)


cmoine:
##### **Overview of the issue**

When I run gradlew, I obtain the following error:
```
:bootRun
java.lang.NoClassDefFoundError: org/springframework/core/env/Environment
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
        at java.lang.Class.privateGetMethodRecursive(Class.java:3048)
        at java.lang.Class.getMethod0(Class.java:3018)
        at java.lang.Class.getMethod(Class.java:1784)
        at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:544)
        at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:526)
Caused by: java.lang.ClassNotFoundException: org.springframework.core.env.Environment
        at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        ... 7 more
```

##### **Motivation for or Use Case**

The Web App simply do not execute (blocking ?)

##### **Reproduce the error**

After an analysis, the gradle directory must contains spaces (hence 99% of the time the Home directory) to reproduce

##### **Related issues**

No

##### **Suggest a Fix**

I fixed the build.gradle pathingJar (shouldn't be patching by the way ?) by adding ".replaceAll(' ','%20'):
```
doFirst {
            manifest {
                attributes 'Class-Path': configurations.runtime.files.collect {
                    it.toURL().toString().replaceFirst(/file:\/+/, '/') // .replaceAll(' ','%20')
                }.join(' ')
            }
        }
```

##### **JHipster Version(s)**

```
j-hipster-app-2@0.0.0 C:\tmp\jHipsterApp2
`-- generator-jhipster@4.0.2

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

```yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "4.0.2",
    "baseName": "jHipsterApp2",
    "packageName": "com.itemis.jhipster",
    "packageFolder": "com/itemis/jhipster",
    "serverPort": "8080",
[...]

PR #5213, opened 2017-02-15
fix ticket #5164 by patching correctly the URLs in the MANIFEST.MF when

cmoine:
fix ticket #5164 by patching correctly the URLs in the MANIFEST.MF when  …
they contains spaces


jdubois:
I have no idea how this works, so thank you so much for contributing this :-)

cmoine:
Trust me ;-)

It comes from how the URL should be presented in the MANIFEST.MF:
C:\Users\Christophe%20M
instead of
C:\Users\Christophe M

jdubois:
I do, it's already merged :-)

cmoine:
Yes I saw that, thank you !

----- Original Message -----

Issue #5207, opened 2017-02-14
Top navbar has rounded corners


mraible:
##### **Overview of the issue**

The top navbar has rounded corners. They're subtle, but they're there.

![screen shot 2017-02-14 at 11 08 53 am](https://cloud.githubusercontent.com/assets/17892/22947671/34e2dc0e-f2b9-11e6-9291-b4fd78b4e674.png)

##### **Motivation for or Use Case**

I like it when UIs are perfect. ;)

##### **Reproduce the error**

Generate an app, checkout the top nav.

##### **Suggest a Fix**

Restructure HTML or add CSS. Not sure.

##### **JHipster Version(s)**

4.0.3

##### **Browsers and Operating System**

OS name: "mac os x", version: "10.12.3", arch: "x86_64", family: "mac"
Chrome Version 56.0.2924.87 (64-bit)

- [x] Checking this box is mandatory (this is just to show you read everything)

deepu105:
Ya I noticed it too but its from bootstrap but we can override that
ofcourse

Thanks & regards,
Deepu

On 14 Feb 2017 9:28 p.m., "Matt Raible" <notifications@github.com> wrote:

*Overview of the issue*

The top navbar has rounded corners. They're subtle, but they're there.

[image: screen shot 2017-02-14 at 11 08 53 am]
<https://cloud.githubusercontent.com/assets/17892/22947671/34e2dc0e-f2b9-11e6-9291-b4fd78b4e674.png>
*Motivation for or Use Case*

I like it when UIs are perfect. ;)
*Reproduce the error*

Generate an app, checkout the top nav.
*Suggest a Fix*

Restructure HTML or add CSS. Not sure.
*JHipster Version(s)*

4.0.3
*Browsers and Operating System*

OS name: "mac os x", version: "10.12.3", arch: "x86_64", family: "mac"
Chrome Version 56.0.2924.87 (64-bit)

   - Checking this box is mandatory (this is just to show you read
   everything)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/jhipster/generator-jhipster/issues/5207>, or mute the
thread

JulienMrgrd:
We can override @deepu105.
We just have to add "! important" after each css line to override.

like this :
![capture d ecran de 2017-02-15 10-08-36](https://cloud.githubusercontent.com/assets/11616517/22967450/d23a5fa6-f366-11e6-885a-014af84905e7.png)

[...]

PR #5216, opened 2017-02-15
Fix for Top navbar has rounded corners #5207

JulienMrgrd:
Tests OK with and without Sass.

capture d ecran de 2017-02-15 15-05-49

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
@@ -10,6 +10,8 @@ Navbar
 .jh-navbar {
     background-color: #353d47;
     padding: .2em 1em;
+    border-radius: 0; /* Hide the navbar rounded corners */
+	-moz-border-radius: 0;
     .profile-image {
         margin: -10px 0px;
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
```diff
@@ -10,6 +10,8 @@ Navbar
 .jh-navbar {
     background-color: #353d47;
     padding: .2em 1em;
+    border-radius: 0; /* Hide the navbar rounded corners */
```
gmarziou: Why not set $enable-rounded to false?

JulienMrgrd: The `$enable-rounded` not only affects the navbar, but few elements of Bootstrap.
[http://v4-alpha.getbootstrap.com/getting-started/options/](url)
> $enable-rounded : Enables predefined border-radius styles on various components.

gmarziou: OK, let's wat and see if we get other round corners issues.

PR #5221, opened 2017-02-15
Add service registration metadata info

PierreBesson:
Displays nicely thanks to @wildloop's PR !

screen shot 2017-02-15 at 18 28 01

Tested to work with Eureka and Consul.


PierreBesson:
I also found this very interesting project that lets you use eureka metadata to customize Ribbon routing : https://github.com/jmnarloch/ribbon-discovery-filter-spring-cloud-starter

I'm thinking that we could do a Zuul Filter that will let you specify in the config specific metadata constraints related to which version of a microservice you want to route to.

PR #5224, opened 2017-02-15
jhiHasAnyAuthority is now a structural directive

### LINKED ISSUE #5063: Convert has-authority directives to structural directives
JHipster 4.0.0

The "Administration" menu doesn't disappear for non-admin users. It just doesn't render - so there's a large gap in the menu.

screen shot 2017-02-02 at 7 44 40 am

gmarziou:
*Warning*: breaking change with new template syntax. You can pass a single string or an array of strings as for `ngClass`.

```html
<some-element *jhiHasAnyAuthority="'ROLE_ADMIN'">...</some-element>

<some-element *jhiHasAnyAuthority="['ROLE_ADMIN', 'ROLE_USER']">...</some-element>
```

- new implementation is adding or removing content rather than hiding or showing it
- remove `jhiHasAuthority` because all can be done from `jhiHasAnyAuthority`
- fix `Principal` methods: `hasAuthority` and `hasAnyAuthority` to always return a `Promise`
- fix `Principal. `authenticate()'` so that it updates `authenticationState`  to notify observers on sign out.

Close #5063


generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.html
```diff
@@ -31,7 +31,7 @@
                     <!-- jhipster-needle-add-entity-to-menu - JHipster will add entities to the menu here -->
                 </ul>
             </li>
-            <li *ngSwitchCase="true" <%=jhiPrefix%>HasAuthority="ROLE_ADMIN" ngbDropdown class="nav-item dropdown pointer">
```
deepu105: what about the ngSwitch?

generators/client/templates/angular/src/main/webapp/app/shared/auth/_has-any-authority.directive.ts
```diff
@@ -1,41 +1,42 @@
-import { Directive, ElementRef, Input, Renderer, OnInit } from '@angular/core';
+import { Directive, ElementRef, Input, TemplateRef, ViewContainerRef } from '@angular/core';
 import { Principal } from './principal.service';
 
+/**
+ * @whatItDoes Conditionally includes an HTML element if current user has any
```
deepu105: nice :clap: we need to add this kind of docs in more places

generators/client/templates/angular/src/main/webapp/app/shared/auth/_principal.service.ts
```diff
@@ -9,7 +9,7 @@ import { <%=jhiPrefixCapitalized%>TrackerService } from '../tracker/tracker.serv
 @Injectable()
 export class Principal {
     private _identity: any;
-    private authenticated = false;
+    private authenticated: boolean = false;
```
deepu105: type not required as its inferred from the value and this will also create a tslint warning

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.html
gmarziou: You cannot have 2 structural diirectives on same element and in this case the *ngStwitch="true" means user is authenticated which is already covered by HasAnyAuthority.

generators/client/templates/angular/src/main/webapp/app/shared/auth/_has-any-authority.directive.ts
gmarziou: Yes I copied it from angular code

generators/client/templates/angular/src/main/webapp/app/shared/auth/_principal.service.ts
gmarziou: Ok, I'll remove it

PR #5228, opened 2017-02-16
open health endpoint for all apps

ruddell:
This lets developers set up simple unauthenticated health checks (I wanted it for Rancher). We already open it up for Consul.

While admins get a full report, users and anonymous users get:
```
curl 'http://localhost:8080/management/health' 
{
  "description" : "Spring Cloud Eureka Discovery Client",    # Not there for monoliths
  "status" : "UP"
}
```
> Shows application health information (when the application is secure, a simple ‘status’ when accessed over an unauthenticated connection or full message details when authenticated).
http://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html


jdubois:
We removed this for security reasons, but indeed that should be fine. It's like a 200 response code from the index page.

PR #5232, opened 2017-02-16
Polish

sendilkumarn:


generators/client/templates/angular/src/main/webapp/app/shared/auth/_principal.service.ts
```diff
@@ -9,8 +9,8 @@ import { <%=jhiPrefixCapitalized%>TrackerService } from '../tracker/tracker.serv
 @Injectable()
 export class Principal {
     private _identity: any;
-    private authenticated = false;
-    private authenticationState = new Subject<any>();
+    private _authenticated = false;
```
deepu105: Can we remove the underscores everywhere as it was originally to give as a feel as its a private var but now we have the private keyword so underscore is useless actually

pascalgrimaud:
@sendilkumarn : there are some fails in Travis build. Here an extract of logs:
```
ERROR in [at-loader] src/main/webapp/app/account/password/password.component.ts:28:24 
    TS2341: Property 'identity' is private and only accessible within class 'Principal'.
ERROR in [at-loader] src/main/webapp/app/account/settings/settings.component.ts:26:24 
    TS2341: Property 'identity' is private and only accessible within class 'Principal'.
ERROR in [at-loader] src/main/webapp/app/account/settings/settings.component.ts:38:28 
    TS2341: Property 'identity' is private and only accessible within class 'Principal'.
ERROR in [at-loader] src/main/webapp/app/admin/user-management/user-management.component.ts:51:24 
    TS2341: Property 'identity' is private and only accessible within class 'Principal'.
ERROR in [at-loader] src/main/webapp/app/admin/user-management/user-management.route.ts:21:27 
    TS2341: Property 'identity' is private and only accessible within class 'Principal'.
```

sendilkumarn:
@pascalgrimaud thanks 👍 

Awww.. there is another method with the same name.

PR #5234, opened 2017-02-16
@EnableConfigurationProperties -> ApplicationProperties.class

ctamisier:
ApplicationProperties class is defined but not loaded.
I'm thinking to load this class to be available for injection, instead of letting the developer to do it.

JHipsterProperties class is loaded by spring.factories file (JHipster core library).


jdubois:
Good catch, I totally forgot you needed to reference it, I thought it was automatic? Let me check

jdubois:
Tested, and indeed this is not automatic! Thanks a lot!!!!

ctamisier:
you're welcome, thanks for this great project!

PR #5244, opened 2017-02-18
Use ehcache-prod instead of ehcache-dev in application-prod.xml

nghib:
Fixed typos issue


jdubois:
Oh I really made a stupid error here, thanks!!!
Next time can you do 2 PRs, one for each concern?

nghib:
Understood. I will do 2 PRs next time

PR #5252, opened 2017-02-19
Audit Todo 

sendilkumarn:


generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.ts
```diff
@@ -21,22 +21,19 @@ export class AuditsComponent implements OnInit {
...
-    datePipe: DatePipe;
 
     constructor(
         private jhiLanguageService: JhiLanguageService,
         private auditsService: AuditsService,
         private parseLinks: ParseLinks,
-        @Inject(LOCALE_ID) private locale: string,
-        private paginationConfig: PaginationConfig
+        private paginationConfig: PaginationConfig,
+        private datePipe: DatePipe
```
deepu105: is it working without setting Locale? i was failing earlier

sendilkumarn: yeah.

Issue #5253, opened 2017-02-19
JHipster 4.0.5 does not include spring-cloud-starter-dependencies when serviceDiscoveryType=false


jander99:
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

##### **Overview of the issue**

generator-jhipster 4.0.5 dev profile is unable to run after yeoman generation.  

##### **Motivation for or Use Case**

Project is broken.  

##### **Reproduce the error**

Generate a new JHipster Microservice using generator-jhipster 4.0.5 with the following options

```
? (1/15) Which *type* of application would you like to create? Microservice application
? (2/15) What is the base name of your application? jhmstest
? (3/15) Would you like to install other generators from the JHipster Marketplace? No
? (4/15) As you are running in a microservice architecture, on which port would like your server to run? It should be unique to avoid port conflicts. 8081
? (5/15) What is your default Java package name? com.github.jander99.jhms
? (6/15) Which *type* of authentication would you like to use? JWT authentication (stateless, with a token)
? (7/15) Which Service Discovery and Configuration solution would you like to use? No Service Discovery and Configuration
? (8/15) Which *type* of database would you like to use? SQL (H2, MySQL, MariaDB, PostgreSQL, Oracle)
? (9/15) Which *production* database would you like to use? MySQL
? (10/15) Which *development* database would you like to use? H2 with disk-based persistence
? (11/15) Do you want to use Hibernate 2nd level cache? Yes, with HazelCast (distributed cache, for multiple nodes)
? (12/15) Would you like to use Maven or Gradle for building the backend? Maven
? (13/15) Which other technologies would you like to use? (Press <space> to select, <a> to toggle all, <i> to inverse selection)
? (14/15) Would you like to enable internationalization support? Yes
? Please choose the native language of the application? English
? Please choose additional languages to install French, Spanish
? (15/15) Besides JUnit and Karma, which testing frameworks would you like to use? Gatling, Cucumber
```

After generation, run maven in dev profile

```
λ mvnw -Pdev
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.springframework.cloud:spring-cloud-starter:jar is missing. @ line 379, column 21
[ERROR] 'dependencies.dependency.version' for org.springframework.cloud:spring-cloud-starter-ribbon:jar is missing. @ line 383, column 21
[...]

PR #5255, opened 2017-02-19
Fix dependencies when choosing no service discovery

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5253


PierreBesson:
Are you sure it does not break service discovery enabled monolith ?

PierreBesson:
I would use the following condition : "if serviceDiscovery || type=microservice/gateway/uaa"

And this might be needed in several places.

PR #5257, opened 2017-02-19
configure websocket CORS and enable CORS in dev

### LINKED ISSUE #5005: "dev" mode with Websocket isn't user-friendly
This comes from our Angular 2 test cases - see https://github.com/jhipster/generator-jhipster/issues/4301#issuecomment-275883927

With Websockets selected:

- it works in "prod" mode
- in "dev" mode there is no way to use it correctly: with just HMR (port 9060) or with BrowserSync + HMR (port 9000) the socket can't be opened, so it's impossible to develop easily. You need to restart everything, as far as I can tell.

I'm not sure this is easy to solve because of HMR, but we should provide a work-around so people can work effectively with Websockets

ruddell:
Fix #5005

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-dev.yml
+++ b/generators/server/templates/src/main/resources/config/_application-dev.yml
@@ -236,6 +236,15 @@ jhipster:
             time-to-live-seconds: 3600
             backup-count: 1
     <%_ } _%>
+   <%_ if (websocket === 'spring-websocket') { _%>
+    cors:
+        allowed-origins: "*"
+        allowed-methods: GET, PUT, POST, DELETE, OPTIONS
+        allowed-headers: "*"
+        exposed-headers:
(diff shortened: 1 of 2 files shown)


jdubois:
+1 for the idea, of course.
Then I'm wondering if we should better just configure this in the `application-dev.yml`, it would be simpler, and not hard-coded. What do you think?

ruddell:
Agreed and updated

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -236,6 +236,13 @@ jhipster:
             time-to-live-seconds: 3600
             backup-count: 1
     <%_ } _%>
+    cors:
```
jdubois: I would only only this if "websockets" is selected

PR #5258, opened 2017-02-19
fixing header change on lang change

sendilkumarn:
This fix adds change of header text when the lang is changed.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/language/_language.helper.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/language/_language.helper.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { Title }     from '@angular/platform-browser';
+import { Router, ActivatedRouteSnapshot } from '@angular/router';
 import { TranslateService, TranslationChangeEvent, LangChangeEvent } from 'ng2-translate/ng2-translate';
 
 import { LANGUAGES } from './language.constants';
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/language/_language.helper.ts
```diff
@@ -23,25 +24,30 @@ export class JhiLanguageHelper {
...
+            this.updateTitle();
+        });
+
+        this.translateService.onLangChange.subscribe((event: LangChangeEvent) => {
+            this.updateTitle();
+        });
+    }
+
+    private getPageTitle(routeSnapshot: ActivatedRouteSnapshot) {
+        let title: string = (routeSnapshot.data && routeSnapshot.data['pageTitle']) ? routeSnapshot.data['pageTitle'] : 'ang4App';
```
jdubois: I'll modify this: if there no title, let's not put anything (better than a wrong name)

sendilkumarn: my bad. Sorry 👍

deepu105: We could set the app name as default title right

jdubois: I put an empty String, at least nous one will complain it's wrong!

sendilkumarn: I fixed it via [this](https://github.com/jhipster/generator-jhipster/commit/11495631ff45c1f51f5083f2b6e8b9f970972467)

PR #5276, opened 2017-02-21
Angular 2 Internet Explorer problem #5186

### LINKED ISSUE #5186: Angular 2 Internet Explorer problem
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

Angular 2 monolithic application home page is not loading , getting error for IE 10-11.. There is no problem in chrome, edge and firefox.. 

IE 10 
------
SCRIPT5009: 'Map' is undefined
vendor.dll.js (45115,9)

IE 11
-----
SCRIPT5007: Unable to get property 'apply' of undefined or null reference
vendor.dll.js (54101,5)

##### **Reproduce the error**
 yarn start, page is not loading in IE 10-11.

##### **JHipster Version(s)**
4.0.3, 4.0.2

Angular 2 monolithic application

##### **JHipster Version(s)**

```
flow@0.0.0 D:\yeoman\jhipster\4.0.2\flow
`-- generator-jhipster@4.0.3

```

```yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "4.0.3",
    "baseName": "flow",
    "packageName": "com.liquidity.flow",
    "packageFolder": "com/liquidity/flow",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "databaseType": "sql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "jwtSecretKey": "45f8a07be79784ded6f6dc1bb868afeb9ecdf6a7",
    "useSass": false,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false,
    "clientFramework": "angular2"
  }
}
```

ls: no such file or directory: .jhipster/*.json

java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)

git version 2.11.1.windows.1

node: v6.9.5

npm: 4.1.2

bower: 1.7.9

gulp:
[14:40:40] CLI version 3.9.1

yeoman: 1.8.5

yarn: 0.19.1

Windows 10 Home 64 Bit, Internet Explorer 11 - 10

serkaneren78:
fix #5186

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
@@ -1,6 +1,5 @@
 /* tslint:disable */
 // Polyfills
-import 'ie-shim'; // Internet Explorer 9 support
 // import 'core-js/es6';
 // Added parts of es6 which are necessary for your project or your browser support requirements.
 import 'core-js/es6/symbol';
--- a/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
@@ -1,8 +1,6 @@
 /* tslint:disable */
 // Polyfills
 import 'reflect-metadata/Reflect';
-// import 'core-js/es6';
-// Added parts of es6 which are necessary for your project or your browser support requirements.
 import 'core-js/es6/symbol';
 import 'core-js/es6/object';
 import 'core-js/es6/function';
@@ -24,4 +22,3 @@ import 'core-js/es6/reflect';
 import 'core-js/es7/reflect';
 
 import 'zone.js/dist/zone';
-import 'zone.js/dist/long-stack-trace-zone';
--- a/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
@@ -1,5 +1,6 @@
 /* tslint:disable */
 // Polyfills
+import 'reflect-metadata/Reflect';
 // import 'core-js/es6';
 // Added parts of es6 which are necessary for your project or your browser support requirements.
 import 'core-js/es6/symbol';
--- a/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
@@ -1,3 +1,27 @@
 /* tslint:disable */
-import 'reflect-metadata/Reflect';
+// Polyfills
+import 'ie-shim'; // Internet Explorer 9 support
+// import 'core-js/es6';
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
```diff
@@ -1,3 +1,27 @@
 /* tslint:disable */
+// Polyfills
 import 'reflect-metadata/Reflect';
+// import 'core-js/es6';
+// Added parts of es6 which are necessary for your project or your browser support requirements.
+import 'core-js/es6/symbol';
```
deepu105: This would bundle this for all browsers which os unnecessary overhead. May be we should add the polyfill to index.html directly using the IE conditional directive

deepu105: Also we are generating es5 code with webpack so most of this shouldn't be needed at all in the first place. I think this needs further investigation

generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
```diff
@@ -1,3 +1,27 @@
 /* tslint:disable */
+// Polyfills
 import 'reflect-metadata/Reflect';
+// import 'core-js/es6';
```
jdubois: we shouldn't have this commented line

generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
```diff
@@ -1,3 +1,27 @@
...
+import 'core-js/es6/set';
+import 'core-js/es6/weak-map';
+import 'core-js/es6/weak-set';
+import 'core-js/es6/typed';
+import 'core-js/es6/reflect';
+
+import 'core-js/es7/reflect';
+
 import 'zone.js/dist/zone';
+import 'zone.js/dist/long-stack-trace-zone';
```
jdubois: Isn't this just for debugging? If this isn't directly linked to the ticket, let's remove this

generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
serkaneren78: all features of es5 moved to es6 namespace. Should i close the pull request?

deepu105:
As I said in my previous review may be we should think of a diferent solution rather than bundling all these extra js code to all browsers. One way would be to add the IE shim from a CDN directly to the `index.html` using the IE conditional comments. But technically as we transpile typescript to ES5 this should have worked so we need to check why the transpiled code still has ES6 code in it.

PR #5278, opened 2017-02-22
Karma tests

sendilkumarn:


generators/client/templates/angular/src/test/javascript/spec/app/account/register/_register.component.spec.ts
```diff
@@ -1,12 +1,11 @@
 import { ComponentFixture, TestBed, async, inject, tick, fakeAsync } from '@angular/core/testing';
-import { MockBackend } from '@angular/http/testing';
-import { Http, BaseRequestOptions } from '@angular/http';
 import { Renderer, ElementRef } from '@angular/core';
 import { Observable } from 'rxjs/Rx';
 <%_ if (enableTranslation) { _%>
```
deepu105: i think you missed to remove the i18n imports here

sendilkumarn: Actually it is needed. we are using it below. 
It takes an instance from TestModule's langService though

deepu105: ok

PR #5286, opened 2017-02-22
Ehcache programmatic configuration (no xml)

henri-tremblay:
It think it makes the code simpler and cleaner. It also adds type safety to cache declaration. Some comments:
* `build-samples.sh` was buggy. So I fixed it
* The Ehcache properties are now normal JHipsterProperties which is much nicer I think
* Much less boilerplate `application.yml` stuff
* Hibernate was upgraded to get [HHH-11471](https://hibernate.atlassian.net/browse/HHH-11471)
* `SocialUserConnection` package in `ehcache.xml` was wrong. Moved to `CacheConfiguration` and fixed
* Caches for users and authority are now created only when used
* Cache declarations are now type safe (the domain entity must exist)


henri-tremblay:
The build is failing because jhipster 1.0.4 doesn't currently contain my code. I'm not sure how I should fix this.

jdubois:
@henri-tremblay yes I'll have a look at it, no problem!
Thanks for your PR, this is exactly what I wanted to do, but it was too complex for me :-)

travis/build-samples.sh
```diff
@@ -20,8 +20,8 @@ function buildProject() {
     dir=$1
     echo "*********************** Building $dir"
     pushd "$mydir/samples/$dir"
-    npm link generator-jhipster
     yo jhipster --force
+    npm link generator-jhipster
```
jdubois: I wouldn't link after generation -> I will revert this line

jdubois:
I did a few fixes ( @henri-tremblay if you have the time to check...) but that looks all good to me.

- Now we have programmatic configuration, this is much better!!!!
- I will push the commit in 1 or 2 hours, the time for the JHipster server lib to be synchronized in Maven central

travis/build-samples.sh
henri-tremblay: Why? Linking is quite useful because we can jump into the sample and try things like entity generation right after having built it.

henri-tremblay:
All fine with your changes

jdubois:
@henri-tremblay looks like this is making the build fail: https://travis-ci.org/jhipster/generator-jhipster/jobs/204607026

Not sure about it, but I don't have the time to check right now

pascalgrimaud:
I will restart the build, as it concerns protractor

pascalgrimaud:
Its ok! The build is green

jdubois:
Great, thanks @pascalgrimaud !

henri-tremblay:
Excellent news!

PR #5295, opened 2017-02-23
fix typos

apixandru:


jdubois:
Incredible, we had those for years!!!! Thanks a lot !

PR #5301, opened 2017-02-25
add Armenian language support

mikron:
add Armenian language support


jdubois:
Thanks Mikron! Can you also do a PR on https://github.com/jhipster/jhipster.github.io/blob/master/pages/installing_new_languages.md so it's documented too?

PR #5304, opened 2017-02-25
prevent deactivate himself, ng1

ctamisier:
Like for the user deletion, we can not deactivate our own account in the user management page.
If this is accepted, I can apply it on ng2.

We still can edit our own account with the dialog and deactivate it but it 'requires more clicks'.
or should we disable the deactivation in the user edition as well ?
or nothing of all of this ?


jdubois:
We usually don't go too deep in the "business logic" in JHipster, but here that looks logical, so +1 from me.
And you also need to do it on the server-side :-)

ctamisier:
ok I see, it needs to be done on server-side for user deletion as well.
So far the protection is managed by the front end.

PR #5306, opened 2017-02-25
change url from /user/{login} to /{login} avoiding url conflict when …

ctamisier:
…login='user'

Similar that the url used for ng2.

Otherwise when login='user' there is a url conflict:
url: 'user/{login}' 
url: '/{login}/delete'

You can click on 'Supprimer' for the user 'user' to see the problem.

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.state.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.state.js
@@ -113,7 +113,7 @@
         })
         .state('user-management-detail', {
             parent: 'user-management',
-            url: '/user/{login}',
+            url: '/{login}',
             data: {
                 authorities: ['ROLE_ADMIN'],
                 pageTitle: 'user-management.detail.title'


gmarziou:
I'm confused.
With your change, wouldn't we have even more conflicts with other states?
For instance, if a user is named "home".

ctamisier:
> I'm confused.
With your change, wouldn't we have even more conflicts with other states?
For instance, if a user is named "home", "docs", ...

There is still 'user-management' url in the parent state:
#/user-management
#/user-management/{login}
#/user-management/{login}/edit
#/user-management/{login}/delete

actually the problem now is if we create a user with the login : 'new'
#/user-management/new => (will conflict with #/user-management/{login})...

I tried to define #/user-management-new instead of #/user-management/new
but I have some issues.. The list of users behind the user creation popup disappears.

gmarziou:
OK, I forgot about angular-ui router parent states but as you said, this solves an issue and creates another one. The only solution that works 100% is to have same depth on all routes.

ctamisier:
we can have the user creation on `#/user-management-new` to avoid this new conflict (like in ng2 if i'm not wrong) by doing:

```
 .state('user-management.new', {
            url: '-new',
```
instead of
```
 .state('user-management.new', {
            url: '/new',          
```

but the '-new' looks a bit hacky...

jdubois:
I understand the issue, but indeed now instead of having an issue with "user", you have one with "new"... So this is not really solving the issue.
I don't have a good solution here - I also don't like the "-new" URL! For AngularJS 1 I'm not worrying too much: this code has been in production for a long time, and will probably be deprecated in the future (AngularJS 1 will die in the end, anyway...).
So I'm closing this because the solution isn't great (it's causing another issue), but still we have this bad design in our URLs.

ctamisier:
It's not only about the url design, we can see an error (visually) in the app when we try to remove the user 'user'. An empty alert is showed and then disappear (but we still can delete the user).
I don't how it is in angular 2 (I will check..)

PR #5307, opened 2017-02-26
add the hand on the sortable header

ctamisier:
Add the hand on the sortable header

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.html
+++ b/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management.html
@@ -9,7 +9,7 @@
       </div>
     </div>
     <div class="table-responsive">
-        <table class="table table-striped">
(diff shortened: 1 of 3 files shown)


gmarziou:
While this works, it should be better to define a CSS class for all these elements that should have this behavior rather than doing it in code because this way it can be overwritten by anyone  without coding.

deepu105:
We already have a `hand` class so you just need to add it where needed

ctamisier:
actually it might be better to let only the sort icon running the sort instead of the whole header...

deepu105:
@ctamisier I would prefer the header to act as sort click are as it was before. From a usability perspective having it only on the icon makes it harder to click and hence not so user friendly. So unless you have a strong reason to make the change please revert back your last commit

ctamisier:
sure!

ctamisier:
I added `jh-table` class in the user management table (to be the same that the entity management table). This adds a style on the sorting icon when we pass the mouse over.

PR #5321, opened 2017-02-27
Properly catch exception on validateToken

ramzimaalej:


generators/server/templates/src/main/java/package/security/jwt/_TokenProvider.java
```diff
@@ -90,7 +90,15 @@ public boolean validateToken(String authToken) {
             return true;
         } catch (SignatureException e) {
             log.info("Invalid JWT signature: " + e.getMessage());
```
gmarziou: Why not logging full exception:?

    log.info("Invalid JWT signature", e);

ramzimaalej: I just kept the initial logging. But, definitely, I can log the exception as requested

cbornet:
With tests ! Very nice !

jdubois:
Yes it's better, but I don't think this solve the "500" error

ramzimaalej:
The fix prevents the "500" error as the exception is caught and a false is returned thus denying access to the requested resource. I'll try to add an integration test to capture that.

jdubois:
Thanks, I'm on my phone so it's probably just me having trouble to figure out how this works. I'll test it today.

PR #5329, opened 2017-02-28
to avoid conflict of grabbing websocket port

jhcao23:
both the spring-websocket and Browsersync use websocket so they will be grabbing the websocket port. the best way is to separate the Browsersync websocket port to a different one otherwise the spring-websocket will not work in the way as expected.

One typical error would refer to this ticket https://github.com/BrowserSync/browser-sync/issues/625.

this change requires BrowserSync version 2.8.1+


ruddell:
Confirmed it works, thanks!  This was bothering me but I wasn't sure where it came from.

Can you please change it so that the comma is at the end of the `target` line?  Also because we generate with browser-sync 2.18.7, I don't think the comment is needed

            proxy: {
                target: 'http://localhost:9060'<% if (websocket === 'spring-websocket') { %>,
                ws: true<% } %>
            }

PR #5337, opened 2017-03-01
Support Thai language

chegola:
Please consider adding Thai language.


generators/languages/templates/src/main/webapp/i18n/th/metrics.json
```diff
@@ -0,0 +1,101 @@
+{
+    "metrics": {
```
deepu105: I think you missed this file

chegola: @deepu105  I intended to leave it as most of them are technical term which we normally use in English.

deepu105:
@chegola could you please add this to documentation website by doing a PR to https://github.com/jhipster/jhipster.github.io repo

chegola:
@deepu105 It's done.

PR #5353, opened 2017-03-03
change raw-loader to html-loader to enable sourcing images directly f…

### LINKED ISSUE #5333: Images referred in HTML templates are not found

I generated a project using jH 4.0.3 version. I copied all my actual navbar content into navbar.html. The logo in navbar.html is referring to `../../../content/images/my_logo.png`

All the images in default project are referred in css files and not from HTML files. I can't figure how to get this working with images referred directly from HTML. When I start the application, the image is not found. Also, I checked in www folder under build. There is no generated hash file name for `my_logo.png`.

Component HTML must be able to refer to images in `.../content/images/...` directly without involving CSS

##### **Reproduce the error**

Add a new image in `/content/images` folder and refer to that image directly using ` `

##### **JHipster Version(s)**

4.0.3

##### **Other**

node: v7.5.0
npm: 4.2.0
bower: 1.8.0
gulp: 3.9.1
yarn: 0.19.1

santoshkt:
…rom HTML files.

Fix #5333

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -55,6 +55,7 @@
     "extract-text-webpack-plugin": "2.0.0-beta.5",
     "file-loader": "0.10.0",
     "generator-jhipster": "<%= packagejs.version %>",
+    "html-loader": "0.4.5",
     "html-webpack-plugin": "2.28.0",
     "image-webpack-loader": "3.2.0",
     "jasmine-core": "2.5.2",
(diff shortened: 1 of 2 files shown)


xsalefter:
Okay. I got the same problems: The image problem as described in #5333 and production problem as @ruddell described in the comment. I confirm that using `html-loader` fix the static images problem, and to solve production problem, you could add this in webpack.common.js

```                
{
  test: /\.html$/,
  loader: 'html-loader',
  options: {
    minimize: false,
    removeComments: false,
    collapseWhitespace: false
  },
  exclude: ['./src/main/webapp/index.html']
}
,

But I'm not sure that this is what we want.

**Comment by deepu105** [comment]
I'm fixing this

**Comment by deepu105** [comment]
The correct fix was to add 
```
                    options: {
                        minimize: true,
                        caseSensitive: true,
                        removeAttributeQuotes:false,
                        minifyJS:false,
                        minifyCSS:false
                    },
```

PR #5355, opened 2017-03-04
Docker image: use Yarn install instead of NPM [ci skip]

pascalgrimaud:
To be consistent with the Devbox.
If someone can review, test and confirm it works fine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,16 +13,16 @@ RUN \
 
   # install utilities
   apt-get install -y \
-     wget \
-     curl \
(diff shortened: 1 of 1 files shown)


romaindequidt:
LGTM (after reviewing https://github.com/pascalgrimaud/generator-jhipster/pull/9)

jdubois:
I merged it, so with our automatic Docker image builds this should be created soon at https://hub.docker.com/r/jhipster/jhipster/builds/ -> can you check the end result?

PR #5356, opened 2017-03-04
Upgrade Node, NPM and Yarn

pascalgrimaud:
- Node: 6.10.0 LTS
- NPM: 4.3.0
- Yarn 0.21.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - docker
 language: node_js
 node_js:
-  - "6.9.5"
+  - "6.10.0"
 jdk:
   - oraclejdk8
 cache:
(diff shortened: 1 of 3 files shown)


kaidohallik:
Maybe upgrade also in ci-cd subgenerator?

pascalgrimaud:
You're right @kaidohallik ! Completely forgot it

But I think we should a global var like for the version of Docker image: https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js
It will be easier for maintain

kaidohallik:
Yarn and Npm versions can be eliminated from ci-cd subgenerator by adding new profile for ci-cd in Maven.  
Gradle in ci-cd already uses Yarn and Npm versions from `gradle.properties` through -PnodeInstall.

But then still Node version is referenced in many places and yarn-npm versions in two places: pom.xml and gradle.properties.

So adding Node, Yarn and Npm versions as global constants sounds good idea to me.

PR #5358, opened 2017-03-04
Add links to project files and folders in README.md

hdurix:
Add links to relative paths in project README file. 

For example on Github, or Gitlab, it creates real links on your project, see [there](https://github.com/hdurix/jhipster-readme-with-links).

### DIFF
--- a/generators/server/templates/_README.md
+++ b/generators/server/templates/_README.md
@@ -41,7 +41,7 @@ Before you can build this project, you must install and configure the following 
 <%_ } _%>
 
 After installing Node, you should be able to run the following command to install development tools.
-You will only need to run this command when dependencies change in `package.json`.
+You will only need to run this command when dependencies change in [package.json](package.json).
 
     <%= clientPackageManager %> install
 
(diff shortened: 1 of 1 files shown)


generators/server/templates/_README.md
```diff
@@ -126,7 +126,7 @@ To optimize the <%= baseName %> application for production, run:
     ./gradlew -Pprod clean bootRepackage<% } %>
 
 <%_ if(!skipClient) { _%>
-This will concatenate and minify the client CSS and JavaScript files. It will also modify `index.html` so it references these new files.
+This will concatenate and minify the client CSS and JavaScript files. It will also modify [index.html](index.html) so it references these new files.
```
pascalgrimaud: This link doesn't exist. It should be `src/main/webapp/index.html` instead
Can you do another PR with [ci skip] in your commit message? So it won't trigger a new Travis build :construction_worker_man: :construction_worker_woman:

gmarziou: Sorry I did not catch it in my review, thanks Pascal. I'm wondering  whether we should even have a link here as the file may be built by webpack so only exists in target/www folder.

pascalgrimaud: np :)
I agree, it's not mandatory to have it in README

hdurix: Ok I remove the link for `index.html` in a new PR.

PR #5359, opened 2017-03-04
remove index href in README [ci skip]

hdurix:
As discussed [there](https://github.com/jhipster/generator-jhipster/pull/5358#discussion_r104285923), remove `index.html` link as it is malformed and handled by webpack.


gmarziou:
Thanks

PR #5361, opened 2017-03-04
Add an option to ci-cd subgen to publish to a docker registry

cbornet:


generators/ci-cd/templates/_Jenkinsfile
```diff
@@ -156,6 +156,27 @@ node {
 <%_ if (jenkinsIntegrations.includes('docker')) { _%>
 <%= gitLabIndent %>    }
 <%_ } _%>
+<%_ if (jenkinsIntegrations.includes('publishDocker')) { _%>
+
+<%= gitLabIndent %>    def dockerImage
+<%= gitLabIndent %>    stage('build docker') {
+    <%_ if (buildTool === 'maven') { _%>
+<%= gitLabIndent %>        sh "cp -R src/main/docker target/"
```
pascalgrimaud: why not using directly the maven goal docker:build ?
is It for using the pluging with docker.build('....') ?

cbornet: That's mainly because when you build in docker you don't have docker in the build container. And I would want to avoid docker-in-docker and the complications that go with it.
Now we could use docker:build when not building in docker. That would just be a little more work.

pascalgrimaud: yes, docker in docker is always hard to maintain

cbornet: Now it's true that since Jenkins is built in Java, we could use maven/gradle to build the docker image.
As for the publish phase, it seems easier to use the docker pipeline plugin but I guess that could be doable with maven/gradle (with more work) if we want to avoid the dependency on the plugin.

pascalgrimaud: For me, both solutions are correct. So I'm ok for both, as with Gradle, we don't use a plugin

For Maven, see the doc https://github.com/spotify/docker-maven-plugin#usage
we can use: `./mvnw docker:build -DpushImage` too, but I don't know how we will manage the credentials

cbornet: Yes. So l propose to use the Jenkins plugin for now and using maven/gradle can be done later as an improvement.

pascalgrimaud:
@cbornet :can you update the documentation too plz ?

cbornet:
yup !

PR #5364, opened 2017-03-04
Slightly improved the code style

MathieuAA:


MathieuAA:
@deepu105 tagging you as you're the one in charge of this branch. Awesome work by the way :)

generators/import-jdl/index.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
-const util = require('util');
-const shelljs = require('shelljs');
-const generator = require('yeoman-generator');
-const chalk = require('chalk');
-const jhiCore = require('jhipster-core');
-const scriptBase = require('../generator-base');
+
+const util = require('util'),
```
deepu105: This was changed on purpose as its the most commonly used code style to have const and let defined as separate lines. Its also the default in most of the eslint style templates. So i guess we should also stick to it. I was gonna updade eslint rules for this as well

MathieuAA: This must be the stupidest "convention" ever! I'll update the PR right away...

PR #5379, opened 2017-03-06
Fix some warnings in TS compilation

pascalgrimaud:
- trailing whitespace
- Exceeds maximum line length of 140


gmarziou:
Personally I think the rule of max line length should be disabled as it can be easily exceeded for long entity name. For instance, I have RatingMovement and TournamentGallery which result in RatingMovementComponent and TournamentGalleryComponent, there's not much we can do about it.

pascalgrimaud:
yes why not.. it can be discussed I think

gmarziou:
OK let's merge this and I can open another PR to discuss the max lne length rule.

PR #5380, opened 2017-03-07
Change some entities and fields names to avoid duplications in Sonar

pascalgrimaud:
Simply using some firstnames. Hope you like them all ;-)


jdubois:
you should have used the names of the team members :-)

PR #5384, opened 2017-03-07
Initialize Mongo using MongoTemplate

cbornet:
Because static types are always better :smile: 

This has been tested on regular Mongo, heroku and CF

### DIFF
--- a/generators/server/templates/src/main/java/package/domain/_Authority.java
+++ b/generators/server/templates/src/main/java/package/domain/_Authority.java
@@ -29,6 +29,13 @@ public class Authority implements Serializable {
     @Column(length = 50)<% } %>
     private String name;
 
+    public Authority() {
+    }
+
+    public Authority(String name) {
+        this.name = name;
+    }
+
     public String getName() {
         return name;
     }
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -34,7 +34,7 @@ liquibase_slf4j_version=2.0.0
 liquibase_hibernate5_version=3.6
 <%_ } _%>
 <%_ if (databaseType === 'mongodb') { _%>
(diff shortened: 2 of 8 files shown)


cbornet:
It also removes a bunch of Sonar warnings for Mongo (duplicate strings and false-positive hard-coded password)

jdubois:
Impressive!! I'm trusting your tests, so I'm merging this directly.

generators/server/templates/src/main/java/package/domain/_Authority.java
```diff
@@ -29,6 +29,13 @@
     @Column(length = 50)<% } %>
     private String name;
 
+    public Authority() {
```
pascalgrimaud: is this constructor needed ?
As it created a bug in our Sonar Analyze so we are now BAA. I see it after your tweet @jdubois sorry :-(

jdubois: For me the error is with the other constructor, as we need an empty one anyway, and it's the other one that lacks the validation.... I'm correcting the Sonar error, but indeed it might not be needed at all - this was added by @cbornet so he probably knows better

jdubois: OK this is really annoying, I'm removing the constructor

cbornet: If you remove this one then the empty constructor can also be removed.
I added this constructor for convenience but of course null values should be handled.
Note that when you call `new Authority()`, name is also not set as there is no default value for it so it's not really better...

cbornet: I had kept the empty constructor because Jackson needs it (or you need to specify the constructor explicitly). I don't know if Jackson is used here so maybe that was useless.

PR #5389, opened 2017-03-08
add breadcrumb feature on client-side

stevehouel:
add breadcrumb display on main component. By default it display pageTitle text on each breadcrumb node.
It s also possible to force node to be hidden by adding breadcrumb property on route.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.html
@@ -1,0 +1,14 @@
+<ol class="breadcrumb">
+    <li class="breadcrumb-item">
+        <a [hidden]="breadcrumbs.length === 0" routerLink=""><i class="fa fa-home"></i></a>
+        <span [hidden]="breadcrumbs.length > 0"><i class="fa fa-home"></i></span>
+    </li>
+    <li class="breadcrumb-item" [ngClass]="{ 'active': last }" *ngFor="let breadcrumb of breadcrumbs;  let last = last">
+        <a [hidden]="last" [routerLink]="[breadcrumb.url]">{{breadcrumb.label | translate}}</a>
+        <%_ if (enableTranslation) { _%>
+        <span [hidden]="!last">{{breadcrumb.label | translate}}</span>
+        <%_ } else { _%>
+        <span [hidden]="!last">{{breadcrumb.label}}</span>
+        <%_ } _%>
+    </li>
+</ol>
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.ts
@@ -1,0 +1,75 @@
+import { Component, OnInit } from '@angular/core';
+import { Router, ActivatedRoute, NavigationEnd, Params, PRIMARY_OUTLET } from '@angular/router';
+import 'rxjs/add/operator/filter';
+
+interface IBreadcrumb {
+  label: string;
+  url: string;
+}
+
+@Component({
+  selector: '<%=jhiPrefix%>-breadcrumb',
+  templateUrl: './breadcrumb.component.html'
+})
+export class BreadcrumbComponent implements OnInit {
+
+  public breadcrumbs: IBreadcrumb[];
+
+  constructor(
+    private activatedRoute: ActivatedRoute,
+    private router: Router
+  ) {
+    this.breadcrumbs = [];
+  }
+
+  ngOnInit() {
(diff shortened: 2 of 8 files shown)


generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.ts
```diff
@@ -0,0 +1,75 @@
+import { Component, OnInit } from '@angular/core';
+import { Router, ActivatedRoute, NavigationEnd, Params, PRIMARY_OUTLET } from '@angular/router';
+import 'rxjs/add/operator/filter';
+
+interface IBreadcrumb {
```
mraible: We don't generally use `I` prefixes for interfaces.

stevehouel: @mraible What do you use instead ?

mraible: Nothing at all, just name it `Breadcrumb`.

generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.html
```diff
@@ -0,0 +1,14 @@
+<ol class="breadcrumb">
+    <li class="breadcrumb-item">
+        <a [hidden]="breadcrumbs.length === 0" routerLink=""><i class="fa fa-home"></i></a>
+        <span [hidden]="breadcrumbs.length > 0"><i class="fa fa-home"></i></span>
+    </li>
+    <li class="breadcrumb-item" [ngClass]="{ 'active': last }" *ngFor="let breadcrumb of breadcrumbs;  let last = last">
+        <a [hidden]="last" [routerLink]="[breadcrumb.url]">{{breadcrumb.label | translate}}</a>
+        <%_ if (enableTranslation) { _%>
```
sendilkumarn: this is not needed here. The `| translate` will be stripped away (if needed) with regex.

generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.html
```diff
@@ -0,0 +1,14 @@
+<ol class="breadcrumb">
+    <li class="breadcrumb-item">
+        <a [hidden]="breadcrumbs.length === 0" routerLink=""><i class="fa fa-home"></i></a>
+        <span [hidden]="breadcrumbs.length > 0"><i class="fa fa-home"></i></span>
+    </li>
+    <li class="breadcrumb-item" [ngClass]="{ 'active': last }" *ngFor="let breadcrumb of breadcrumbs;  let last = last">
```
sendilkumarn: can we use `routerLinkActive` here instead of `ngClass`

generators/client/templates/angular/src/main/webapp/app/layouts/breadcrumb/_breadcrumb.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    // iterate over each children
+    if (children.length > 0) {
+        for (let child of children) {
+        // verify primary route
+        if (child.outlet !== PRIMARY_OUTLET) {
+            continue;
+        }
+
+        // get the route's URL segment
+        let routeURL: string = child.snapshot.url.map(segment => segment.path).join('/');
```
sendilkumarn: can we use activatedRoute's url instead of snapshot's?

sendilkumarn: similarly below,

deepu105:
This needs to be discussed first. I'm not very sure if we need this feature. @jdubois what do you think?

atomfrede:
I don't like breadcrumbs. From my experience they are often misleading and
don't bring much benefit.

Am 08.03.2017 21:35 schrieb "Deepu K Sasidharan" <notifications@github.com>:

jdubois:
Yes, sorry @stevehouel but this should be discussed in a ticket first, and for example have screenshots, etc. This is going to impact everyone.
I'm a bit reluctant to make those pages even more complex: I already think they are far too complex, and that we shouldn't have made them this way. At the beginning JHipster was just a "skeleton", so people could improve it, and now people have to remove tons of code to have something working, and I think we already went too far, and that we should instead simplify what we already have, not add new features.

xetys:
I agree with @jdubois. It is so much pain today to upgrade a generated project to the last version because of so many changes in code, one just don't need (in his particular case)

I think this is case for a module.

stevehouel:
Sure @jdubois I understand the purpose and the desire to only make a "skeleton". So remove this PR. I will maybe do a module as @xetys said.

PR #5392, opened 2017-03-09
Added missing { in template file

toolforger:


toolforger:
It would be nice if automatic tests caught this kind of mistake.
They'd have to test whether the input parameters actually show up in the generated text.
(I don't know about the ramifications though, and I don't know how to write such a test - feel free to open an Issue if you think it's a good idea to to this.)

jdubois:
Thanks, that's a great ticket as we don't test this, indeed.
This could be tested with Protractor + testing all the languages, but this would be too much for our poor build infrastructure. It's a matter of cost, mostly, as we only use the free tier from Travis. There is some work on this currently, but it can't be shared yet (not our fault, it has to be private for some very good reasons at the moment), and anyway it won't solve everything, as once again our ressources are very limited.

toolforger:
Thanks.
A quick grep revealed that this particular error does not occur in other places, so I guess automated testing is more a nice-to-have feature rather than a dealbreaker.

pascalgrimaud:
I can be done during `yarn test` I think, during the same tests in [test-langages.js](https://github.com/jhipster/generator-jhipster/blob/master/test/test-languages.js)

The idea would be:
1 - check if all json are valid
2 - checking some regexps like `{{username}}`, `{{ param }}` etc.

I tried once to do the point 1) but as I sux in nodeJs, I didn't go further...

PR #5405, opened 2017-03-12
Travis: some minor improvements

pascalgrimaud:
- rename app folder with gradle and psql
- try to optimize order to speedup the global time
- simply use -DskipTests without true value


deepu105:
@pascalgrimaud we still have a lot of INFO and DEBUG logs on the travis builds which makes it very difficult to scroll through to find errors. Is it possible to set some environment variable in travis script so that only WARNING and above logs are printed for everything? cc @jdubois

deepu105:
may be something like this http://stackoverflow.com/a/6307666/1273312

pascalgrimaud:
I think we can reduce more. 
Anyway, the log info or debug are not important, right ? only the final result is important IMO

jdubois:
Yes I just changed the log level for tests today -> it was wrong anyway, the `logback-test.xml` which was supposed to work according to the Spring doc only worked when running tests in an IDE (and were not working with Maven!).
So that should be OK now.

PR #5414, opened 2017-03-14
Polish [ci skip]

pascalgrimaud:


deepu105:
@pascalgrimaud the builds `app-ng2-default` and `app-default-from-scratch` are almost the same, the only difference is one uses a different jhiPrefix may be we can merge those and use that slot to build some other combination. what do you think?

pascalgrimaud:
Maybe but be carefull, there are 2 major differences:
- app-default-from-scratch is used on our Sonar -> https://sonarqube.com/dashboard?id=io.github.jhipster.sample%3Ajhipster-sample-application
- app-default-from-scratch has RUN_APP=0, so the application won't be launched

deepu105:
What if we remove RUN_APP=0  and it can stiil do sonar


Thanks & Regards,
Deepu

pascalgrimaud:
yes can be done

PR #5421, opened 2017-03-15
Add webpack dashboard plugin for ng2

### LINKED ISSUE #5420: Add the cool webpack dev server dashboard
Integrate https://github.com/FormidableLabs/webpack-dashboard for Angular 2

image

deepu105:
fix #5420 

@jhipster/developers plz try it out on ng2 apps

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -91,6 +91,7 @@
     "tslint-loader": "3.4.3",
     "typescript": "2.2.1",
     "webpack": "2.2.1",
+    "webpack-dashboard": "0.3.0",
     "webpack-dev-server": "2.4.2",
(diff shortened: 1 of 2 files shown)


deepu105:
Damn i think it breaks the travis builds :(
Need to check if there is a way to enable this only when doing `yarn start`

jdubois:
It looks definitely cool but it doesn't work:

- In my terminal (iTerm2 3.0.15 with lots of modifications) it's a bit weird, and the scrolling is bad, but that could be usable
- In Intellij IDEA it's not working well at all

So, this is cool, but for me lots of trouble for just some graphical "wow" effect.

I'm closing this as it's been open for some time, and for me we have better things to do at the moment.

deepu105:
Ya agree, i was about to close it myself

PR #5422, opened 2017-03-15
Add a docker-registry compose file when choosing publishDocker in ci-cd

PierreBesson:
As suggested by @cbornet.


pascalgrimaud:
LGTM

PR #5423, opened 2017-03-15
adds needle for proxying microservice API urls through webpack

### LINKED ISSUE #5417: webpack proxy doesn't include requests to gateway forwarded APIs
- Please follow the issue template below for bug reports and feature requests.
- Tickets opened without any of these informations will be **closed** without any explanation.

in #5263 there was mentioned a issue I found already some time ago. When using our current microservice setup there is an issue, were requests to zuul routed microservice APIs don't work, as the paths are missing in webpack config

##### **Reproduce the error**

just generate a microservice application and try to open the UI on entity pages. All requests to `localhost:9000/<service-name>` are running into a 404.

##### **Related issues**

was mentioned in #5263 

##### **Suggest a Fix**

when creating entities, the sub-generator should add the paths here: https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/webpack/_webpack.dev.js#L24

##### **JHipster Version(s)**

4.0.8

I can PR a solution for this

xetys:
so this happens now for every entity generation

fix #5417


xetys:
this is solved by yeoman already....I tried my bookstore example, which uses several entities per microservice and it worked perfectly

but something is reaally wrong with travis...and it's not my fault, I hope 😆

pascalgrimaud:
Just tested it, it works well
LGTM

jdubois:
I'm quite surprised this is solved by Yeoman, and that you don't get the same line several times

deepu105:
Not by Yeoman. Its solved by us. We have a regex check to skip writing if the content is already present :)

Issue #5435, opened 2017-03-19
Duplicate administrator navbar entry


thomas-burko:
##### **Overview of the issue**

When I go to the user settings logged in as admin and click on save, it duplicates the Administrator li entry in the navbar. It repeats every time I click on save.

##### **Motivation for or Use Case**

It should not create another li administrator entry.

##### **Reproduce the error**

Create a new JHipster App (everything default, Angular 2)
Log in as admin
Go to Account -> Settings
Click on Save
Watch the navbar

##### **Suggest a Fix**

I'm not sure, but I think the problem caused in the file **has-any-authority.directive.ts**.
When I add the line _this.viewContainerRef.clear();_ in the method updateView() in the if query the problem is resolved. (Between Line 35 and 36). 

##### **JHipster Version(s)**

4.0.8

##### **JHipster Version(s)**

```
demo@0.0.0 /Users/Thomas/Downloads/demo
├── UNMET PEER DEPENDENCY @angular/compiler@2.4.7
├── UNMET PEER DEPENDENCY @angular/core@2.4.7
└── generator-jhipster@4.0.8 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

```yaml
{
  "generator-jhipster": {
    "jhipsterVersion": "4.0.8",
    "baseName": "demo",
    "packageName": "",
    "packageFolder": "",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "",
    "clientFramework": "angular2",
    "useSass": false,
    "clientPackageManager": "npm",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ]
  }
}
```

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

ls: no such file or directory: .jhipster/*.json

##### **Browsers and Operating System**

java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
[...]

PR #5436, opened 2017-03-19
Fix duplicate administrator navbar entry

thomas-burko:
see #5435

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/auth/_has-any-authority.directive.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/auth/_has-any-authority.directive.ts
@@ -32,10 +32,9 @@ export class HasAnyAuthorityDirective {
 
     private updateView(): void {
         this.principal.hasAnyAuthority(this.authorities).then(result => {
+            this.viewContainerRef.clear();
             if (result) {
                 this.viewContainerRef.createEmbeddedView(this.templateRef);
-            } else {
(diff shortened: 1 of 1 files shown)


gmarziou:
I'm away from home and cannot test right now but this fix makes sense.

generators/client/templates/angular/src/main/webapp/app/shared/auth/_has-any-authority.directive.ts
```diff
@@ -33,6 +33,7 @@ export class HasAnyAuthorityDirective {
     private updateView(): void {
         this.principal.hasAnyAuthority(this.authorities).then(result => {
             if (result) {
+                this.viewContainerRef.clear();
```
deepu105: @thomas-burko could you move this line outside the if and remove the else part

thomas-burko:
moved statement outside and removed else part

tezcane:
If I open two different browsers and login one as admin and one as user.  Then the "Administration" menu item shows up on the user view and also disappears from the admin view.

So this does not work:
```
this.viewContainerRef.createEmbeddedView(this.templateRef);
this.viewContainerRef.clear();
```

However, the original seems to cover all the cases I tested:
```
if (result) {
    this.viewContainerRef.clear();
    this.viewContainerRef.createEmbeddedView(this.templateRef);
} else {
    this.viewContainerRef.clear();
}
```

PR #5437, opened 2017-03-19
Export clientPackageManager for module [ci skip]

pascalgrimaud:


deepu105:
While writing the React module I realized that we actually dont need the modules sub gen to use JHipster API at all https://github.com/hipster-labs/generator-jhipster-react/blob/master/generators/app/index.js#L4 :)

pascalgrimaud:
Oh, good to know!
But I think we need to keep it, for all the current module, and document the way to use JHipster API

deepu105:
Yes


Thanks & Regards,
Deepu

PR #5452, opened 2017-03-20
fix imageUrl not updated from /account REST call, remove UserService.java's extra whitespace

tezcane:
AccountResource.saveAccount()->UserService.updateUser(): imageUrl is now passed so it can be updated
UserService.java: "@Transactional(readOnly = true)    " to "@Transactional(readOnly = true)"

new issues

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -215,13 +215,19 @@ public class UserService {
      * @param lastName last name of user
      * @param email email id of user
      * @param langKey language key
+     <%_ if (databaseType == 'mongodb' || databaseType == 'sql') { _%>
+     * @param imageUrl image URL of user
+     <%_ } _%>
      */
-    public void updateUser(String firstName, String lastName, String email, String langKey) {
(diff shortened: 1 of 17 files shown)


generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -263,6 +269,9 @@ public void updateUser(String firstName, String lastName, String email, String l
                 <%_ if (databaseType == 'mongodb' || databaseType == 'cassandra') { _%>
                 userRepository.save(user);
                 <%_ } _%>
+                <%_ if (searchEngine == 'elasticsearch') { _%>
+                // userSearchRepository.save(user); // left out by design
```
deepu105: I dont see why this is required here. can you please rebase this branch with latest master and then remove this?

tezcane: Sure.  Can you please help me understand why it is not needed and/or recommended?

Only the admin is calling userUpdate(UserDTO user) currently, does that have something to do with it?

It had resolved "User managment" page not getting updates in my code, though I really have my doubts now (will revisit).

deepu105:
Plz rebase and squash

tezcane:
OK, working on it now.

PR #5455, opened 2017-03-21
Removing throws clause to resolve "sonar" bug

sreekrishna:
Removing unnecessary throws clause to resolve sonar reported bug(major) "Define and throw a dedicated exception instead of using a generic one".


jdubois:
Thanks @sreekrishna - unfortunately Sonar only gives us one "free" project, at https://sonarqube.com/dashboard?id=io.github.jhipster.sample%3Ajhipster-sample-application so we can't test all options :-(

PR #5458, opened 2017-03-21
properly template oauth_client_details.csv

ruddell:
Oauth2 apps are currently broken as the database has `<%= baseName %>` instead of the actual baseName.  This started happening in v4.1.0 (it's fine in v4.0.8)


pascalgrimaud:
@ruddell : what do you mean by broken? The application start but we can't connect to it, isn't it ?

ruddell:
@pascalgrimaud The application starts and appears to work, but you can't login because the clientId/clientSecret on the frontend don't match the database.

Database contains:
`<%= baseName %>app;res_<%= baseName %>;my-secret-token-to-change-in-production;read,write;password,refresh_token,authorization_code,implicit;;ROLE_ADMIN,ROLE_USER;1800;2000;{};true`

pascalgrimaud:
Ok thanks for the information @ruddell ! That means we should put more protractor tests...
I will do it on hipster-labs builds

PR #5465, opened 2017-03-22
Added missing _enum_nl.json file for the Dutch language

tomdesair:
There is no `_enum_nl.json` file. This breaks the JDL parser (and probably also the entity generator) when adding an enum with i18n support for Dutch.


pascalgrimaud:
Thanks @tomdesair 
By counting the file, there are 2 others missing file: gl and ro

But it's the same file for all languages! I wonder if we can do something better than this.

PR #5469, opened 2017-03-23
fix imageUrl not updated from /account REST call, remove UserService.…

tezcane:
Wasn't successful on rebase and squash of #5452 so I made a fresh branch.

Also, in this commit I removed the previously added userRepository.save(user), per @deepu105's instructions.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -219,13 +219,13 @@ public class UserService {
      * @param imageUrl image URL of user
      <%_ } _%>
      */
-    public void updateUser(String firstName, String lastName, String email, String langKey<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>, String imageUrl<% } %>) {
+    public void updateUser(String firstName, String lastName, String email, String langKey<% if (databaseType === 'mongodb' || databaseType === 'sql') { %>, String imageUrl<% } %>) {
         userRepository.findOneByLogin(SecurityUtils.getCurrentUserLogin()).ifPresent(user -> {
             user.setFirstName(firstName);
             user.setLastName(lastName);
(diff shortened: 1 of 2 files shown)


tezcane:
Two failures:
 7833.10 JHIPSTER=app-ng2-cassandra
 7833.11  JHIPSTER=app-ng2-mongodb

Both fail the same tests:
  UserResourceIntTest.updateUser:313 expected:<"j[hipsterFirstName]"> but was:<"j[ohn]">
  UserResourceIntTest.updateUserLogin:350 expected:<"j[hipster]"> but was:<"j[ohndoe]">

I looked through the code and didn't see how my changes would effect these particular tests- there logic also looks sound.

tezcane:
Found issue, submitted commit.

generators/server/templates/src/main/java/package/service/_UserService.java
```diff
@@ -215,13 +215,19 @@ public User createUser(UserDTO userDTO) {
      * @param lastName last name of user
      * @param email email id of user
      * @param langKey language key
+     <%_ if (databaseType == 'mongodb' || databaseType == 'sql') { _%>
+     * @param imageUrl image URL of user
+     <%_ } _%>
      */
-    public void updateUser(String firstName, String lastName, String email, String langKey) {
+    public void updateUser(String firstName, String lastName, String email, String langKey<% if (databaseType == 'mongodb' || databaseType == 'sql') { %>, String imageUrl<% } %>) {
```
deepu105: please use `===` and `!==` for conditions. We might have old conditions without it but we are trying to change that

tezcane: OK, done.

tezcane:
It appears the Travis CI build has already failed several tests after I made the "==" to "===" changes.  I believe this is due to something else though. Please advise.

PR #5470, opened 2017-03-23
Fix LoggingConfiguration log enriching

PierreBesson:
Logstash log enriching with application name, port, etc was broken by the migration to constructor injection.

This caused things like this in Kibana:

screen shot 2017-03-23 at 12 28 57

This PR fixes the issue.


jdubois:
Excellent, thank you so much!!!!

- I did noticed it was broken and did not know what to do
- I also thought the wrong injection, and wanted to remove it, but didn't understand to was linked

Perfectly in time for the next release, scheduled this afternoon :-)

pascalgrimaud:
thanks @PierreBesson !!! 
I need this fix in my project, so I will test it soon

PR #5483, opened 2017-03-24
add polyfill to avoid some IE errors

kaidohallik:
related to #5186 and #5276

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/_index.html
+++ b/generators/client/templates/angular/src/main/webapp/_index.html
@@ -15,10 +15,10 @@
     <![endif]-->
     <<%= jhiPrefix %>-main></<%= jhiPrefix %>-main>
     <script>
-        // load shim for IE 11
+        // load shim for IE
         if (window.navigator.userAgent.indexOf('Trident/') > 0) {
             var script = document.createElement('script');
-            script.src = 'https://cdnjs.cloudflare.com/ajax/libs/es6-shim/0.35.3/es6-shim.min.js';
+            script.src = 'core-js-shim.min.js';
             script.type = 'text/javascript';
             document.getElementsByTagName('head')[0].appendChild(script)
         }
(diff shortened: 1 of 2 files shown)


kaidohallik:
Alternative solution to `core-js/client/shim` is to use es6-shim, named by @deepu105 in https://github.com/jhipster/generator-jhipster/issues/5186#issuecomment-288990778
1. add dev dependency to package.json:
```
	"es6-shim": "0.35.3",
```
2. add this line to polyfills.ts:
```
import 'es6-shim/es6-shim';
```

kaidohallik:
page with core-js/client/shim is 600 kB bigger  
page with es6-shim/es6-shim is 350 kB bigger

kaidohallik:
I made shim conditional to IE 11 only with javascript hack in index.html. PR updated.

generators/client/templates/angular/src/main/webapp/_index.html
```diff
@@ -1,6 +1,15 @@
 <!doctype html>
 <html class="no-js">
 <head>
+    <script>
```
deepu105: can we move this to right after the `<jhi-main></jhi-main>` in this file please.

Btw did this work in IE?
@holgerdunst does this solve your issue?

kaidohallik:
Script moved after the `<jhi-main></jhi-main>`. PR updated.  
This solves IE 11 error.  
I didn't test this with IE 10, for IE 10 this needs additional changes, for example `userAgent` string is different in IE 10.

sendilkumarn:
@kaidohallik `Trident` does include IE10 right ?

Can we have more specific IE11 check in that case as in [this](http://stackoverflow.com/a/17447718)

kaidohallik:
@sendilkumarn My previous comment about IE 10 can be incorrect. It's possible that this fix works also in IE 10. `Trident/` is also part of IE 10 user agent string: https://msdn.microsoft.com/en-us/library/ms537503(v=vs.85).aspx

kaidohallik:
@sendilkumarn Now I think that this shim can be applied to all IE versions, so no need to detect IE 11.  
Feature detection is possible, but then logic goes maybe too complex. Question is: what feature exactly to detect...

PR #5489, opened 2017-03-26
Modify ParameterizedErrorVM to support client-side interpolation of multiple params

l30thelion:
Modify params attribute of ParameterizedErrorVM  template to be key/value pair (Map) so that the client-side is able to perform string interpolation of multiple parameters within a message.  

See `alert-error.component.ts` and notice that calls to the function addErrorAlert, which performs  interpolation, use key/value pairs (json object) to specify interpolation parameters.  Using an array will not work for case where there are multiple placeholders (params).

Current implementation uses a "params" array which translates to json as, for example:
```
{
  "message" : "error.myCustomError",
  "params" : ["param 1 value", "param 2 value"]
  }
} 
```
A map is required to support the feature correctly:
```
{
  "message" : "error.myCustomError",
  "params" : {
    "param0" : "Program 4",
    "param1" : "another param"
  }
}
```

Custom, parameterized exception, which can be translated on the client side. For example:

`throw new CustomParameterizedException("myCustomError", "hello", "world");
   
Can be translated with:
   `"error.myCustomError" :  "The server says {{param0}} to {{param1}}"`

or

Custom, parameterized exception, which can be translated on the client side. For example:
```
final HashMap<String, String> params = new HashMap<>();
params.put("customParamName", "hi");
params.put("anotherParamName", "you");
throw new CustomParameterizedException("myCustomError", params.);

```
Can be translated with:
`"error.myCustomError" : "The server says {{customParamName}} to {{anotherParamName}}"`

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
@@ -62,7 +62,7 @@ export class <%=jhiPrefixCapitalized%>AlertErrorComponent implements OnDestroy {
                                 'Field ' + fieldName + ' cannot be empty', 'error.' + fieldError.message, {fieldName: fieldName});
                         }
                     } else if (httpResponse.text() !== '' && httpResponse.json() && httpResponse.json().message) {
-                        this.addErrorAlert(httpResponse.json().message, httpResponse.json().message, httpResponse.json());
+                        this.addErrorAlert(httpResponse.json().message, httpResponse.json().message, httpResponse.json().params);
                     } else {
                         this.addErrorAlert(httpResponse.text());
                     }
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
@l30thelion : in your last commit, you tried to fix the 'upgrade' sub generator ? It should be better to do another PR than putting all fixes in the same

l30thelion:
@pascalgrimaud: last commit wasn't meant to be pushed, sorry, removed it, thx!

pascalgrimaud:
thanks to you ! :)

jdubois:
@l30thelion can you check Sonar: https://sonarqube.com/component_issues?id=io.github.jhipster.sample%3Ajhipster-sample-application#resolved=false|types=CODE_SMELL|sinceLeakPeriod=true

It doesn't like that your field isn't final (and there's also an empty line missing above it)

jdubois:
@l30thelion I'm modifying the code to make Sonar happy -> not a big fan of what I'm doing, I'll link my commit here, can you have a look?

l30thelion:
@jdubois I see, looks good.

Ooops sorry, I also see that for consistency, the "this" prefix should be used: 

`paramMap.put(PARAM + i, params[i]);`

shoud be:

`this.paramMap.put(PARAM + i, params[i]);`

PR #5495, opened 2017-03-27
Compiling with AOT

sendilkumarn:
Adding AOT


sendilkumarn:
Yeah, since if we want to use that compiled version, we have to change the main.ts and then recompile. That makes many things easy.

deepu105:
Protractor tests seems to fail. Can you try running them locally

sendilkumarn:
Good to see things are in green atlast.

generators/server/templates/gitignore
```diff
@@ -140,3 +140,8 @@ Desktop.ini
 # ESLint
 ######################
 .eslintcache
+
+######################
+# AOT
+######################
+/aot/**/*
```
jdubois: why do you have a new "aot" folder? Shouldn't this be in "target/aot"?

jdubois:
@sendilkumarn congrats! I just made a simple comment, otherwise is it ready to be merged? I can't wait to test this?

generators/client/templates/angular/_package.json
```diff
@@ -112,16 +114,18 @@
...
   },
   "scripts": {
-    "lint": "tslint 'src/main/webapp/app/**/*.ts' --force",
-    "lint:fix": "tslint 'src/main/webapp/app/**/*.ts' --fix --force",
+    "lint": "tslint '<%= MAIN_SRC_DIR%>app/**/*.ts' --force",
+    "lint:fix": "tslint '<%= MAIN_SRC_DIR%>app/**/*.ts' --fix --force",
     "tsc": "tsc",
     "tsc:w": "tsc -w",
+    "ngc": "ngc -p tsconfig-aot.json",
+    "cleanup": "rimraf '<%= MAIN_SRC_DIR%>app/**/*.js' '<%= MAIN_SRC_DIR%>app/**/*.js.map' <%= BUILD_DIR %>",
```
deepu105: Is this required after moving AOT files into target?

generators/client/templates/angular/_package.json
```diff
@@ -112,16 +114,18 @@
...
     "start": "<%= clientPackageManager %> run webpack:dev",
-    "webpack:build": "webpack --config webpack/webpack.vendor.js && webpack --config webpack/webpack.dev.js",
+    "webpack:build": "<%= clientPackageManager %> run ngc && webpack --config webpack/webpack.vendor.js && webpack --config webpack/webpack.dev.js",
     "webpack:build:dev": "webpack --config webpack/webpack.dev.js",
     "webpack:build:vendor": "webpack --config webpack/webpack.vendor.js",
     "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9060",
-    "webpack:prod": "<%= clientPackageManager %> test && webpack -p --config webpack/webpack.vendor.js && webpack -p --config webpack/webpack.prod.js",
+    "webpack:prod": "<%= clientPackageManager %> run test && <%= clientPackageManager %> run ngc && webpack -p --config webpack/webpack.vendor.js && webpack -p --config webpack/webpack.prod.js",
```
deepu105: can you put ngc first and test next?

generators/client/templates/angular/src/main/webapp/app/shared/_shared-libs.module.ts
```diff
@@ -5,6 +5,9 @@ import { CommonModule } from '@angular/common';
 import { NgbModule } from '@ng-bootstrap/ng-bootstrap';
 import { NgJhipsterModule } from 'ng-jhipster';
 import { InfiniteScrollModule } from 'angular2-infinite-scroll';
+<%_ if (websocket === 'spring-websocket') { _%>
+import { WindowRef } from './tracker/window.service';
```
deepu105: can you move this to shared-common.module as this file should only have external libs. same goes for provider def below

generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
```diff
@@ -11,8 +12,8 @@ import { AuthServerProvider } from '../auth/auth-oauth2.service';
 <%_ } _%>
 
 <%_ // TODO find a better way to import these libs here _%>
```
deepu105: remove this comment if the below import works. I tried it once but it didnt work. Are you sure it works now?

deepu105:
Also dont we have to make some changes to the prod build so that it uses the AOT generated files instead of normal typescript transpiled?

generators/client/templates/angular/_package.json
sendilkumarn: yeah, these files are created in these folders. Apparently we haven't configured it anywhere. strange.

generators/client/templates/angular/src/main/webapp/app/shared/tracker/_tracker.service.ts
sendilkumarn: It works.

generators/client/templates/angular/_package.json
deepu105: But havent we configured the AOT files to be generated in the target folder then why are these files being created here?

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -6,6 +6,10 @@ import { TranslateService } from 'ng2-translate';
 <%_ } _%>
 import { AlertService } from 'ng-jhipster';
 
+<%_ if (websocket === 'spring-websocket') { _%>
+import { WindowRef } from './tracker/window.service';
```
deepu105: Ah sorry I missed this. Window service is common right. That means we should be using this instead of direct window reference everywhere. That means we can actually move the window service into ng-jhipster lib and use it here. what do you think? I think its used for csrf-service as well

sendilkumarn: Csrf ? Is that in ng jhipster ?

sendilkumarn:
@deepu105 that was my question too.

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
deepu105: Ok seems its already removed so its only used here. But i think this could still be moved to the lib. what do you think?

deepu105:
Seems like we need to change the bootstrap to use as per [this](https://angular.io/docs/ts/latest/cookbook/aot-compiler.html)
`platformBrowser().bootstrapModuleFactory(AppModuleNgFactory);` in prod mode so that AOT is used. 
[Here](https://medium.com/@laco0416/aot-compilation-with-webpack-359ac9f4916f) is an article on doing AOT with webpack

So in order to have AOT working I think we need to do this as well. This PR will make our code AOT compatible but wont do actual AOT

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
sendilkumarn: I dont want to add it in lib because majority of people wont even use websockets. So it is unnecessary overhead. Let our lib be simple and we add what is used across. IMO

sendilkumarn:
That is another discussion. I have tried both and they both have problems. Like the build is not happening. Rollup (as suggested by angular ) and ngtools. 

This is my plan.

Add an main file specific to aot
Use ngtools in webpack for bundling
Resolve errors ( there are plenty )

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
deepu105: that was the point of the lib right move stuff thats trivial. but its fine we can do that later. So will merge this now and open another ticket to change prod build to use AOT

pascalgrimaud:
@deepu105 @sendilkumarn : I have some questions for you. It's just to understand clearly what has been done here and if it's the correct behavior:

- a lot of the new files *.js and *.js.map are generated into `src/main/webapp` after every compile: is it correct?
- every minor change result into 3 modified files, instead of one:
```shell
╭─pgrimaud@xps  ~/projects/devoxx2017/demo2  ‹master› 
╰─$ gst
Sur la branche master
Modifications qui ne seront pas validées :
  (utilisez "git add <fichier>..." pour mettre à jour ce qui sera validé)
  (utilisez "git checkout -- <fichier>..." pour annuler les modifications dans la copie de travail)

	modifié :         src/main/webapp/app/layouts/navbar/navbar.component.js
	modifié :         src/main/webapp/app/layouts/navbar/navbar.component.js.map
	modifié :         src/main/webapp/app/layouts/navbar/navbar.component.ts
```

- should we commit them to git ? or should they be in gitignore ?
- I test `yarn run cleanup`and I got:
```shell
╭─pgrimaud@xps  ~/projects/devoxx2017/demo2  ‹master*› 
╰─$ gst
Sur la branche master
Modifications qui ne seront pas validées :
  (utilisez "git add/rm <fichier>..." pour mettre à jour ce qui sera validé)
  (utilisez "git checkout -- <fichier>..." pour annuler les modifications dans la copie de travail)

	supprimé :        src/main/webapp/app/account/account.module.js
	supprimé :        src/main/webapp/app/account/account.module.js.map
	supprimé :        src/main/webapp/app/account/account.route.js
	supprimé :        src/main/webapp/app/account/account.route.js.map
	supprimé :        src/main/webapp/app/account/activate/activate.component.js
	supprimé :        src/main/webapp/app/account/activate/activate.component.js.map
	supprimé :        src/main/webapp/app/account/activate/activate.route.js
	supprimé :        src/main/webapp/app/account/activate/activate.route.js.map
	supprimé :        src/main/webapp/app/account/activate/activate.service.js
	supprimé :        src/main/webapp/app/account/activate/activate.service.js.map
	supprimé :        src/main/webapp/app/account/index.js
	supprimé :        src/main/webapp/app/account/index.js.map
(...)
```

The current result is that our official [Sonar Analysis](https://sonarqube.com/dashboard?id=io.github.jhipster.sample%3Ajhipster-sample-application) check all these new generated files, so we lost the 80%+ and AAA -> ping @jdubois so you are aware

jdubois:
Indeed, that's not correct:

- For me the generated files (JS, map...) should go inside the `target` directory. They are "compiled" files, like `.class` files in Java
- Also, putting them in `target/` should solve the Sonar analysis, which is a very, very big issue for me (a lot of people are watching this)

sendilkumarn:
@pascalgrimaud you are correct. It should not be generated there. Added a fix for that, it will be placed under target folder now.

pascalgrimaud:
thanks @sendilkumarn !

Issue #5490, opened 2017-03-27
Upgrade postgres docker image from 9.5.4 to 9.6.2


tezcane:
##### **Overview of the issue**

Upgrade postgres docker image from 9.5.4 to 9.6.2

##### **Motivation for or Use Case**

It has been out for some time and has been stable in my testing with JHipster.

##### **Reproduce the error**

N/A

##### **Related issues**

N/A

##### **Suggest a Fix**

In postgresql.yml change:  
   image: postgres:9.5.4 
to 
  image: postgres:9.6.2

I can do PR for this.

##### **JHipster Version(s)**

4.1.1

##### **JHipster configuration**

N/A

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

N/A

##### **Browsers and Operating System**

N/A

- [ X ] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
if it's a LTS version, yes, let's upgrade. 
I can see in Docker Hub some alpine version, so maybe we can switch to alpine too

tezcane:
It will be supported until September 2021
https://www.postgresql.org/support/versioning/

FYI, there is a new 9.5.6 version too (still newer than 9.5.4)

jdubois:
If it's LTS, and if you tested it without any issue, go ahead and do a PR, we'll merge it

tezcane:
I tested the alpine image too, but shied away from it due to Murphy.  Which version would you guys want?

From https://hub.docker.com/_/postgres/ :

postgres:alpine
This image is based on the popular Alpine Linux project, available in the alpine official image. Alpine Linux is much smaller than most distribution base images (~5MB), and thus leads to much slimmer images in general.

This variant is highly recommended when final image size being as small as possible is desired. The main caveat to note is that it does use musl libc instead of glibc and friends, so certain software might run into issues depending on the depth of their libc requirements. However, most software doesn't have an issue with this, so this variant is usually a very safe choice. See [this Hacker News comment thread](https://news.ycombinator.com/item?id=10782897) for more discussion of the issues that might arise and some pro/con comparisons of using Alpine-based images.

pascalgrimaud:
I prefer stability rather than gaining time for first download... I am very happy with classic image, nether had issue. So in that case, maybe we should stay like this for database (without alpine)

tezcane:
I prefer the full image as well since it should have better support for advanced use cases and not have corner cases too.

[...]

PR #5506, opened 2017-03-29
Update Postgres 9.5.4 to latest 9.6.2 (Support to Sep. 2021)

tezcane:
Update Postgres 9.5.4 to latest 9.6.2 (Support to Sep. 2021)

Update for #5490


pascalgrimaud:
FYI, you need to put 'fix #5490'  or 'close #5490' instead of 'update for #5490'
With that, the ticket will be auto closed as soon as it is merged

tezcane:
Thanks for the tip @pascalgrimaud, will do for the future.

Why did half the Travis builds fail?  The errors look genuine but probably not due to this update.

jdubois:
@tezcane there was another error in the build, not your fault!

PR #5512, opened 2017-03-29
config i18n in test context

ctamisier:
adding the path for message*.properties when **running the tests**.

Otherwise this kind of call :
`messageSource.getMessage("error.subtitle", null, Locale.FRENCH);`

throw an exception because it can't find the key
`org.springframework.context.NoSuchMessageException: No message found under code 'error.subtitle' for locale 'fr'.`


ctamisier:
it's already fixed actually...

PR #5520, opened 2017-03-30
Dependencies cleanup

erikkemperman:
`metrics-spring` no longer depends on `net.codahale.metrics`:
http://mvnrepository.com/artifact/com.ryantenney.metrics/metrics-spring/3.1.3

same for `cassandra-driver-core`:
http://mvnrepository.com/artifact/com.datastax.cassandra/cassandra-driver-core/3.1.4

`hikaricp` no longer depends on `sun.tools`:
http://mvnrepository.com/artifact/com.zaxxer/HikariCP/2.6.0

`liquibase-core` no longer depends on `jetty-servlet`:
http://mvnrepository.com/artifact/org.liquibase/liquibase-core/3.5.3

And some minor maven/gradle alignment for good measure.

### DIFF
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -242,9 +242,7 @@ dependencies {
     <%_ if (databaseType === 'sql') { _%>
     compile "org.hibernate:hibernate-envers"
     compile "org.hibernate:hibernate-validator"
-    compile ("org.liquibase:liquibase-core") {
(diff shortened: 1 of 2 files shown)


erikkemperman:
Thanks for review and merge Julien!

pascalgrimaud:
Thanks for this PR @erikkemperman !
The alignment between maven and gradle is very important: there should not be difference between them.

PR #5521, opened 2017-03-30
Use metrics-jcache

henri-tremblay:
Upgrading metrics was needed to get metrics-jcache.

I wasn't able to test locally so we will see how travis goes.

This PR is linked with jhipster/jhipster#12

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -13,6 +13,8 @@ cassandra_unit_spring_version=3.0.0.1
 <%_ if (cucumberTests) { _%>
 cucumber_version=1.2.4
 <%_ } _%>
+# Overridden to get metrics-jcache
+dropwizard_metrics_version=3.2.2
 logstash_logback_encoder_version=4.8
 javax_transaction_version=1.2
 json_path_version=0.9.1
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -190,6 +190,9 @@ repositories {
 dependencies {
     compile "io.github.jhipster:jhipster:${jhipster_server_version}"
     compile "io.dropwizard.metrics:metrics-core"
+    <%_ if (hibernateCache == 'ehcache') { _%>
+    compile "io.dropwizard.metrics:metrics-jcache:" + dependencyManagement.managedVersions["io.dropwizard.metrics:metrics-core"]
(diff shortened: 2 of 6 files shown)


PierreBesson:
Nice to see new releases of Dropwizard metrics.
This will fail as long as we don't merge your PR on jhipster/jhipster and release the lib. cc @jdubois.

henri-tremblay:
@PierreBesson No. In fact you don't need the jhipster PR. I'm just removing useless code from it. But Travis is failing. I will fix that.

generators/server/templates/_gradle.properties
```diff
@@ -13,6 +13,8 @@ cassandra_unit_spring_version=3.0.0.1
 <%_ if (cucumberTests) { _%>
 cucumber_version=1.2.4
 <%_ } _%>
+# Overridden to get metrics-jcache
+dropwizard-metrics.version=3.2.2
```
pascalgrimaud: I think it need to be rename to: `dropwizard_metrics_version=3.2.2`

And this [file](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/_build.gradle#L196) needs this:
```groovy
    <%_ if (hibernateCache === 'ehcache') { _%>
    compile "io.dropwizard.metrics:metrics-jcache:${dropwizard_metrics_version}"
    <%_ } _%>
```

henri-tremblay:
Ok, I'm stuck and tired of searching. I can't find the name to use to overload the dropwizard-metrics version in Gradle. Two seconds with Maven, and have to read the entire spring documentation to find nothing for Gradle. Grrr...

generators/server/templates/_gradle.properties
henri-tremblay: I tried. Doesn't work. However, I'm using the dependencyManagement version of metrics-core instead of the variable. Also, I've decided to always add metrics-jcache. That was on purpose. But I can indeed add it only when needed. Let me change it.

jdubois:
oh sorry I merged this without reading your comment @henri-tremblay - and I'm like you, and can't understand how Gradle configurations work. I'll push a quick (and ugly!) fix, but I'm not sure you can do better.

henri-tremblay:
Thanks! I'm starting to wonder if we can do better... At least it works and now we have a nice off-the-shelf jcache metrics

PR #5529, opened 2017-04-01
Fix cassandra audits

pascalgrimaud:


deepu105:
doesnt audits work in Cassandra?

pascalgrimaud:
No backend, it has never been coded
That's why I removed the front

deepu105:
Wow after all these years. Are there even people using cassandra at all

jdubois:
Stats show less than 0,5% of projects, and that's probably me doing tests :-)
So yes very disappointing, that's also why I'm removing it from the gateway.
At some point we will need to remove Cassandra support, probably.

sendilkumarn:
and me 😉 doing tests 😛

deepu105:
Ya probably its not worth the maintenance effort :(

PR #5533, opened 2017-04-02
This stops generating js & js.map in the same folder

sendilkumarn:
It took a while to figure this out, since we already specify a path below. 

It fixes [this](https://github.com/jhipster/generator-jhipster/pull/5495#issuecomment-291016939)

@jdubois @pascalgrimaud

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -119,7 +119,7 @@
     "tsc": "tsc",
     "tsc:w": "tsc -w",
     "ngc": "ngc -p tsconfig-aot.json",
-    "cleanup": "rimraf '<%= MAIN_SRC_DIR%>app/**/*.js' '<%= MAIN_SRC_DIR%>app/**/*.js.map' <%= BUILD_DIR %>",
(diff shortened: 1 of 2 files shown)


jdubois:
I don't understand your commit, but still I'm merging and trying because I trust you :-)

sendilkumarn:
back to AAA @jdubois thanks 👍

deepu105:
@sendilkumarn good. I thought you copied the original tsconfig for this and added the AOT section but just saw that the `tsconfig.json` and `tsconfig-aot.json` are not in sync. They should be in sync to avoid issues in dev and prod.

Could you please make sure they are in sync?

I think ideally you should just copy the `tsconfig.json` and then add below to it. 
Also module should be commonjs and not es2015.

I'm gonna do the change in a PR can you do some tests?

```
  "files": [
    "<%= MAIN_SRC_DIR %>app/app.module.ts",
    "<%= MAIN_SRC_DIR %>app/app.main.ts"
  ],

  "angularCompilerOptions": {
   "genDir": "<%= BUILD_DIR %>aot",
   "skipMetadataEmit" : true
 }
```

PR #5534, opened 2017-04-03
fix tsconfig-aot.json

deepu105:
Related to #5533

@sendilkumarn can you test this

### DIFF
--- a/generators/client/templates/angular/_tsconfig-aot.json
+++ b/generators/client/templates/angular/_tsconfig-aot.json
@@ -1,23 +1,28 @@
 {
-  "compilerOptions": {
-    "target": "es5",
-    "module": "es2015",
-    "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


jdubois:
So, are you both OK to merge this? I'm going to do full tests this morning, so it's really the right time for me if you want this in our 4.2.0 release

sendilkumarn:
@jdubois can we park this for a while ?

sendilkumarn:
@deepu105 It is not strictly asking you to check that per say.

jdubois:
Yes, no problem, it's just if you want it included in the next release

deepu105:
@sendilkumarn if things are working fine with commonjs lets go with that. we dont need tree shaking at the moment so when we do if it causes issues we can reconsider

jdubois:
Done :-)

PR #5535, opened 2017-04-03
Fixed missing 'one-to-many' relationship case for required relation

mmoscher:
Currently a **required**  `one-to-many` relationship yields to a compilation error in the appropriate *ResourceIntTest.java class.

```
jhipster-test/src/test/java/de/test/web/rest/AResourceIntTest.java:[82,22] incompatible types: de.test.domain.B cannot be converted to java.util.Set<de.test.domain.B>
```
It fails since a one-to-many relations is handled wrong: trying to access a `set` method which is expecting a set of b's.

Reproducible with the following JDL used in a fresh project:

```
entity A {}
entity B {}

relationship OneToMany {
  A{b required} to B
}
```

As the relation to B is many A only has a set of B's. Thus we also need to add one item to the set, similar to the many-to-many case.

Cheer's,
Marco


jdubois:
Wow, first time contribution to the project, and you go into the most difficult part of our code!!!! Your build failed, I'm having a look.

jdubois:
Looks like a Travis glitch, so I'm merging this

jdubois:
And kudos for your first contribution!

mmoscher:
Thanks :) - hope it was worth it  :mag: :v:

PR #5536, opened 2017-04-04
pass packageName and packageFolder to addEntityToEhcache

ruddell:
This change allows modules to use addEntityToEhcache and addEntryToEhcache.  Currently (from a module), addEntryToEhcache fails with the below error:
```
Unable to add undefined.domain.EntityAuditEvent.class.getName() to CacheConfiguration.java file.
	/private/tmp/jhi/mono/src/main/java/undefined/config/CacheConfiguration.java doesn't exist
```

### DIFF
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -340,7 +340,7 @@ function writeFiles() {
                 this.addChangelogToLiquibase(`${this.changelogDate}_added_entity_${this.entityClass}`);
 
                 if (this.hibernateCache === 'ehcache') {
-                    this.addEntityToEhcache(this.entityClass, this.relationships);
+                    this.addEntityToEhcache(this.entityClass, this.relationships, this.packageName, this.packageFolder);
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


ruddell:
Travis's protractor tests are failing due to invalid chrome version

pascalgrimaud:
@ruddell : I suppose the restart won't work, right?
I don't know if it's related to Travis or a transitive dependency.

jdubois:
- Thanks @ruddell for the fix - I already saw the issue but didn't have time to create the ticket, so that's perfect :-)
- Yes the build is broken because of Chrome - no idea why! Probably a transitive dependancy, indeed

ruddell:
@pascalgrimaud A restart didn't fix it, it was working a few hours ago though

jdubois:
Yes I think it comes from Selenium.

- no idea why it's only on one project, and not all
- I'm relunctant to manually upgrade Chrome in Travis, it will slow down
the build, and the default version should be good enough for what we do

jdubois:
I'm merging as Travis fails for a totally different reason

jdubois:
Oh so the issue is that the latest Selenium release requires a Chrome version that is newer than what Travis provides!!! I'm creating a specific issue for that

PR #5540, opened 2017-04-04
Generate criteria class and backend for filtering

gzsombor:
Add automatically the static metamodel generation/JPASpecificationExecution, and generate criteria classes and filtering code for JPA (with spring's JPA Specifications)

This is a first version, only SQL is checked, and with 'service' class generation. And the REST controller is not modified yet.

What do you think?

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -213,6 +213,7 @@ module.exports = EntityGenerator.extend({
                 this.validation = false;
                 this.dto = 'no';
                 this.service = 'no';
+                this.jpaMetamodelFiltering = false;
             } else {
                 // existing entity reading values from file
                 this.log(`\nThe entity ${this.name} is being updated.\n`);
@@ -256,6 +257,7 @@ module.exports = EntityGenerator.extend({
         this.searchEngine = this.fileData.searchEngine || this.searchEngine;
         this.javadoc = this.fileData.javadoc;
         this.entityTableName = this.fileData.entityTableName;
+        this.copyFilteringFlag(this.fileData, this);
         if (_.isUndefined(this.entityTableName)) {
             this.warning(`entityTableName is missing in .jhipster/${this.name}.json, using entity name as fallback`);
             this.entityTableName = this.getTableName(this.name);
@@ -302,6 +304,7 @@ module.exports = EntityGenerator.extend({
         askForTableName: prompts.askForTableName,
         askForDTO: prompts.askForDTO,
         askForService: prompts.askForService,
+        askForFiltering: prompts.askForFiltering,
         askForPagination: prompts.askForPagination
     },
 
@@ -401,6 +404,10 @@ module.exports = EntityGenerator.extend({
                 this.warning(`service is missing in .jhipster/${this.name}.json, using no as fallback`);
                 this.service = 'no';
             }
+            if (_.isUndefined(this.jpaMetamodelFiltering)) {
+                this.warning(`jpaMetamodelFiltering is missing in .jhipster/${this.name}.json, using 'no' as fallback`);
+                this.jpaMetamodelFiltering = false;
+            }
             if (_.isUndefined(this.pagination)) {
                 if (this.databaseType === 'sql' || this.databaseType === 'mongodb') {
                     this.warning(`pagination is missing in .jhipster/${this.name}.json, using no as fallback`);
@@ -427,6 +434,7 @@ module.exports = EntityGenerator.extend({
             this.data.dto = this.dto;
             this.data.service = this.service;
             this.data.entityTableName = this.entityTableName;
+            this.copyFilteringFlag(this, this.data);
             if (this.databaseType === 'sql' || this.databaseType === 'mongodb') {
                 this.data.pagination = this.pagination;
             } else {
@@ -744,6 +752,7 @@ module.exports = EntityGenerator.extend({
                         fieldsContainBigDecimal: this.fieldsContainBigDecimal,
                         fieldsContainBlob: this.fieldsContainBlob,
                         fieldsContainImageBlob: this.fieldsContainImageBlob,
+                        jpaMetamodelFiltering: this.jpaMetamodelFiltering,
                         pkType: this.pkType,
                         entityApiUrl: this.entityApiUrl,
                         entityClass: this.entityClass,
--- a/generators/entity/prompts.js
+++ b/generators/entity/prompts.js
@@ -32,6 +32,7 @@ module.exports = {
     askForTableName,
     askForDTO,
     askForService,
+    askForFiltering,
     askForPagination
 };
 
@@ -272,6 +273,36 @@ function askForTableName() {
         if (props.entityTableName !== this.entityTableName) {
             this.entityTableName = _.snakeCase(props.entityTableName).toLowerCase();
         }
+        done();
+    });
+}
+
+function askForFiltering() {
+    // don't prompt if server is skipped, or the backend is not sql, or no service requested
+    if (this.useConfigurationFile || this.skipServer || this.databaseType !== 'sql' || this.service === 'no') {
+        return;
+    }
+    const done = this.async();
+    const prompts = [
+        {
+            type: 'list',
+            name: 'filtering',
+            message: 'Do you want to add filtering?',
+            choices: [
+                {
+                    value: 'no',
+                    name: 'Not needed'
+                },
+                {
+                    name: 'Dynamic filtering for the entities with JPA Static metamodel',
+                    value: 'jpaMetamodel'
+                }
+            ],
+            default: 0
+        }
+    ];
+    this.prompt(prompts).then((props) => {
+        this.jpaMetamodelFiltering = props.filtering === 'jpaMetamodel';
         done();
     });
 }
--- a/generators/entity/templates/server/src/main/java/package/repository/_EntityRepository.java
+++ b/generators/entity/templates/server/src/main/java/package/repository/_EntityRepository.java
@@ -64,7 +64,7 @@ import java.util.UUID;<% } %>
 <%_ } if (databaseType === 'sql' || databaseType === 'mongodb') { _%>
 @SuppressWarnings("unused")
 @Repository
-public interface <%=entityClass%>Repository extends <% if (databaseType === 'sql') { %>JpaRepository<% } %><% if (databaseType === 'mongodb') { %>MongoRepository<% } %><<%=entityClass%>,<%= pkType %>> {
+public interface <%=entityClass%>Repository extends <% if (databaseType === 'sql') { %>JpaRepository<% } %><% if (databaseType === 'mongodb') { %>MongoRepository<% } %><<%=entityClass%>,<%= pkType %>><% if (jpaMetamodelFiltering) { %>, JpaSpecificationExecutor<<%=entityClass%>><% } %> {
     <%_ for (idx in relationships) {
         if (relationships[idx].relationshipType === 'many-to-one' && relationships[idx].otherEntityName === 'user') { _%>
 
--- a/generators/entity/templates/server/src/main/java/package/service/_EntityQueryService.java
+++ b/generators/entity/templates/server/src/main/java/package/service/_EntityQueryService.java
@@ -1,0 +1,139 @@
+<%#
+ Copyright 2013-2017 the original author or authors.
+
+ This file is part of the JHipster project, see https://jhipster.github.io/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+package <%=packageName%>.service;
+<%_ const serviceClassName = entityClass + 'QueryService';
+    const instanceType = (dto === 'mapstruct') ? entityClass + 'DTO' : entityClass;
+    const instanceName = (dto === 'mapstruct') ? entityInstance + 'DTO' : entityInstance;
+    const mapper = entityInstance  + 'Mapper';
+    const dtoToEntity = mapper + '.'+ 'toEntity';
+    const entityToDto = mapper + '.'+ 'toDto';
+    const entityListToDto = mapper + '.' + 'toDto';
+    const entityToDtoReference = mapper + '::'+ 'toDto';
+    const repository = entityInstance  + 'Repository';
+    const criteria = entityClass + 'Criteria';
+
+    if (fieldsContainLocalDate === true) { _%>
+import java.time.LocalDate;<% } %><% if (fieldsContainZonedDateTime === true) { %>
+import java.time.ZonedDateTime;<% } if (fieldsContainBigDecimal === true) { %>
+import java.math.BigDecimal;<% } %>
+
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.domain.Specifications;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import io.github.jhipster.service.QueryService;
+
+import <%=packageName%>.domain.<%= entityClass %>;
+import <%=packageName%>.domain.*; // for static metamodels
+import <%=packageName%>.repository.<%= entityClass %>Repository;<% if (searchEngine === 'elasticsearch') { %>
+import <%=packageName%>.repository.search.<%= entityClass %>SearchRepository;<% } %>
+import <%=packageName%>.service.dto.<%= entityClass %>Criteria;
+<% if (dto === 'mapstruct') { %>
+import <%=packageName%>.service.dto.<%= entityClass %>DTO;
+import <%=packageName%>.service.mapper.<%= entityClass %>Mapper;<% } %>
+<%_ for (idx in fields) { if (fields[idx].fieldIsEnum === true) { _%>
+import <%=packageName%>.domain.enumeration.<%= fields[idx].fieldType %>;
+<%_ } } _%>
+
+/**
+ * Service for executing complex queries for <%= entityClass %> entities in the database.
+ * The main input is a {@link <%= entityClass %>Criteria} which get's converted to {@link Specifications},
+ * in a way that all the filters must apply.
+ * It returns a {@link List} of {%link <%= instanceType %>} or a {@link Page} of {%link <%= instanceType %>} which fullfills the criterias
+ */
+@Service<% if (databaseType === 'sql') { %>
+@Transactional(readOnly = true)<% } %>
+public class <%= serviceClassName %> extends QueryService<<%= entityClass %>> {
+
+    private final Logger log = LoggerFactory.getLogger(<%= serviceClassName %>.class);
+
+<%- include('../common/inject_template', {viaService: false, constructorName: serviceClassName, queryService: false}); -%>
+
+    /**
+     * Return a {@link List} of {%link <%= instanceType %>} which matches the criteria from the database
+     * @param criteria The object which holds all the filters, which the entities should match.
+     * @return the matching entities.
+     */
+    @Transactional(readOnly = true)
+    public List<<%= instanceType %>> findByCriteria(<%= criteria %> criteria) {
+        log.debug("find by criteria : {}", criteria);
+        final Specifications<<%= entityClass %>> specification = createSpecification(criteria);
+    <%_ if (dto === 'mapstruct') { _%>
+        return <%= entityListToDto %>(<%= repository %>.findAll(specification));
+    <%_ } else { _%>
+        return <%= repository %>.findAll(specification);
+    <%_ } _%>
+    }
+
+    /**
+     * Return a {@link Page} of {%link <%= instanceType %>} which matches the criteria from the database
+     * @param criteria The object which holds all the filters, which the entities should match.
+     * @param page The page, which should be returned.
+     * @return the matching entities.
+     */
+    @Transactional(readOnly = true)
+    public Page<<%= instanceType %>> findByCriteria(<%= criteria %> criteria, Pageable page) {
+        log.debug("find by criteria : {}, page: {}", criteria, page);
+        final Specifications<<%= entityClass %>> specification = createSpecification(criteria);
+    <%_ if (dto === 'mapstruct') { _%>
+        final Page<<%= entityClass %>> result = <%= repository %>.findAll(specification, page);
+        return result.map(<%= entityToDtoReference %>);
+    <%_ } else { _%>
+        return <%= repository %>.findAll(specification, page);
+    <%_ } _%>
+    }
+
+    /**
+     * Function to convert <%= criteria %> to a {@link Specifications}
+     */
+    private Specifications<<%= entityClass %>> createSpecification(<%= criteria %> criteria) {
+        Specifications<<%= entityClass %>> specification = Specifications.where(null);
+        if (criteria != null) {
+            if (criteria.getId() != null) {
+                specification = specification.and(buildSpecification(criteria.getId(), <%= entityClass %>_.id));
+            }
+            <%_
+            fields.forEach((field) => {
+                if (isFilterableType(field.fieldType)) { _%>
+            if (criteria.get<%= field.fieldInJavaBeanMethod %>() != null) {
+                specification = specification.and(<%= getSpecificationBuilder(field.fieldType) %>(criteria.get<%= field.fieldInJavaBeanMethod %>(), <%= entityClass %>_.<%= field.fieldName %>));
+            }
+            <%_ }
+            });
+
+            relationships.forEach((relationship) => {
+                const relationshipType = relationship.relationshipType;
+                if (relationshipType == 'many-to-one' || relationshipType == 'one-to-one') { _%>
+            if (criteria.get<%= relationship.relationshipNameCapitalized %>Id() != null) {
+                specification = specification.and(buildReferringEntitySpecification(criteria.get<%= relationship.relationshipNameCapitalized %>Id(), <%= entityClass %>_.<%= relationship.relationshipFieldName %>, <%= relationship.otherEntityNameCapitalized %>_.id));
+            }
+            <%_    } // if relationshipType=...
+            }); // forEach
+        _%>
+        }
+        return specification;
+    }
+
+}
--- a/generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
+++ b/generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
@@ -1,0 +1,102 @@
+package <%= packageName %>.service.dto;
+
+import java.io.Serializable;<% if (fieldsContainBigDecimal === true) { %>
+import java.math.BigDecimal;<% } %>
+<%_ for (idx in fields) { if (fields[idx].fieldIsEnum === true) { _%>
+import <%= packageName %>.domain.enumeration.<%= fields[idx].fieldType %>;
+<%_ } } _%>
+import io.github.jhipster.service.filter.Filter;
+import io.github.jhipster.service.filter.IntegerFilter;
+import io.github.jhipster.service.filter.LongFilter;
+import io.github.jhipster.service.filter.StringFilter;
+import io.github.jhipster.service.filter.RangeFilter;
+<%_ if (fieldsContainInstant === true) { _%>
+import io.github.jhipster.service.filter.InstantFilter;<% } %>
+<%_ if (fieldsContainLocalDate === true) { _%>
+import io.github.jhipster.service.filter.LocalDateFilter;<% } %>
+<%_ if (fieldsContainZonedDateTime === true) { _%>
+import io.github.jhipster.service.filter.ZonedDateTimeFilter;<% } %>
+
+<%_
+  const referenceFilterType = '' + pkType + 'Filter';
+  var filterVariables = [{name:'id', type: pkType, filterType:referenceFilterType,fieldInJavaBeanMethod:'Id' } ];
+  var extraFilters = {};
+  fields.forEach((field) => {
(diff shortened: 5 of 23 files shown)


deepu105:
I like the idea of having generated methods for this, but do we have enough use cases to make this available by default?

gzsombor:
We can create a angular component to list entities based on a filter. We can put this component to the entity list page - with a minimal UI for the filters, and even into the entity detail page, to display one-to-many relations.

gmarziou:
This looks great and having a UI for the filter would be awesome.

xetys:
I also like this idea, as filtering is the second thing you do with literally every non-trivial entity, and would appreciate to have a "best practice" way on that.

deepu105:
I like the idea but my only concern is that it adds a lot more code for everyone by default may be this should have a opt-in or opt-out flag in the entity subgen so that people can choose not to have it if they want. One complaint I keep hearing is that we generate a lot of code and some of it becomes noise for people who dont use it

jdubois:
Totally agree with @deepu105 , I also have a lot of people telling me we generate a lot of useless code - of course those people usually have no idea why we generate that code, and it takes them a few weeks to understand why that's interesting, but still that gives a bad first impression.
So this should be at least an option. Maybe a module?
Then yes, the criteria API is great, and generating this is a great idea, of course.

nkolosnjaji:
Hi all,

First of all great work @gzsombor. Here are some of my toughts. JPA Criteria API is good for jHipster and I think most of modern dynamic web apps need some kind of filtering data on server side. On other hand, I think that [querydsl](http://www.querydsl.com/) is better fit here, because it does support MongoDB aswell and it gives us abbility to use awesome [spring-data-commons](http://docs.spring.io/spring-data/commons/docs/current/reference/html/#core.web.type-safe) web support. It really saves a lot of time on server side.

Nikola

gzsombor:
I don't want to add an extra dependency - querydsl - when the built in solutions 'hibernate + spring data' is perfectly capable to handle : the rdbms case. If someone needs to execute complex queries, it's better to not use MongoDB :-)
 So the current state of the patch: 
- supports SQL databases
- you can build fairly complex filters over the REST interface, so for example: 
  /api/students?name.equals=John&age.greaterThan=12&description.contains=soccer

I think, the mapping and the implementation is pretty enough - even tests are generated for the filters 😉 
 There are some field types, where this doesn't work correctly - the binary types, and with the date types. I hope, the later could be fixed somehow.

nkolosnjaji:
I can agreen that about complex queries and MongoDB, but sometimes you want to build dynamic quries (witout using elasticsearch flag) and filter that from GUI. With SpringData + QueryDSL there is no new extra generated classes because they alredy did specificaltion(Predicate) builder. With Criteria there will be need to maintain this extra classes and maintain unit tests, where SpringData team already did it.

gzsombor:
Querydsl needs a custom annotation processor tool to generate classes on compilation time, even for MongoDB ( http://www.querydsl.com/static/querydsl/latest/reference/html/ch02s07.html ), so I don't understand what you want to say.  The generated classes are not a problem, from my point of view, for validating all the queries which is executed by the application, this is needed.
 Back to the current patch : I figured out, what went wrong with the date fields - unfortunately Spring is unable to handle the conversion from the GET parameters, into fields which types are generic, so I had to add custom Filter classes. However, I could add a bunch of tests for the filters, which I've found a big relief 😄

nkolosnjaji:
By extra generated classes I don't mean classes generated by maven/gradle (static metamodel or queryDSL classes). I was meaning classes generated by jHipster generator - classes added by this PR.

There is now 10 new/extra classes with this PR. I don't have anything against this, but just saying, that there is no such thing when using QueryDSL + Spring data. No test classes + no differences between SQL and Mongo.

gzsombor:
To enable Spring Data + static metamodels, you only need 4 lines : https://github.com/jhipster/generator-jhipster/pull/5540/files#diff-4ac12e0abae2ba3bd72788629c00c4abR857
 Sure, you can add the appropriate QueryDSL maven/gradle plugin as well, but alone, it's not a big help.  The rest is for the REST interface (pun intended 😄) - with lot's of tests.
Yes, there are a 5 generated class, which can be added to https://github.com/jhipster/jhipster, I've added here just for convenience - it's faster to iterate inside the generator, without every time creating a new release, when it is accepted.

gzsombor:
With moving the mostly static generated code to jhipster/jhipster, I was able to shrink the patch further. But this will need a new release from the server library - which is not yet available, so the travis tests are failing. 
 The code coverage is pretty good, I believe, as for every entity field, we generate 3-5 test cases into the *EntityResourceIntTest, so everyone could see, how neat the URLs, which are accepted.
 For example : 
* /api/myEntity?field1.equals=Something
* /api/myEntity?intField.greaterThan=100
* /api/myEntity?enumField.in=YES,NO

What do you think, is the generated code small enough to worth the addition ? :smile:

jdubois:
I'm sorry @gzsombor I know I need to review this, but at the moment the focus is on Angular 4 - I spent the evening on it and I'm totally lost, so I really don't have the time for this at the moment. I'll have a look after the next minor release.

jdubois:
I'm so sorry @gzsombor I really don't have enough time - I haven't look at all at this PR, and it's 2 months old...

jdubois:
I finally got some time, so here are my thoughts:

- For complex applications, this is indeed very good. As long as it's an option (as people ALWAYS complain we generate too many things), that's definitely a great addition to JHipster. Have you got an idea of the amount of generated code for one entity? Is this really big?
- There's a lot of code and I'm bit afraid of the maintenance. But if you have good test coverage, that should be fine.
- No problem to add stuff into the JHipster server library, however I have 2 questions: why do we need so much boilerplate code, isn't JPA or Spring Data already enough? I'm not sure of the package name, maybe something with JPA in it?
- I know you only focus on JPA, but don't go further at the moment: it's by far the most widely used option, so let's not add even more code now.
- For the MVC endpoints: yes the service layer should be used (otherwise you generate unused methods, which is weird), but as we have AngularJS + Angular (and soon React) that's going to add a lot of work. So let's not do it now: it's optional, and then people need to manually code it - anyway that's going to be used when some real business logic is needed, and so that's hard to generate.

gzsombor:
About the boilerplate : Initially I've tried implementing a "fat" criteria class, where each field from the entity has a couple of field in the criteria class for various filtering, for example :
From a `String name` 
```
String nameEqualsTo;
String nameContains;
Boolean nameIsNull;
```

etc ...
But it resulted in a huge generated service and criteria implementations. Then I had the idea, to create a Filter<TYPE> class, which contains the 'equalsTo' / 'isNull' / etc attribute. With this, the generated and repeating code is dramatically reduced - so far, so good. Unfortunately, I've noticed that Spring is unable to convert properly the generic types. So in :
```
class EntityCriteria {
  Filter<Integer> age;
  Filter<LocalDate> startDate;
}

```
And code, something like in the REST controller:

public void searchEntity(EntityCriteria criteria) {
   Integer age = criteria.getAge().getEqualsTo()
}

This would compile fine - because of the Java generics - but on runtime, it results in an ugly cast cast expressions, because Spring decides that EntityCriteria.age is a Field<?>, so it can put a String, without any conversions. However, if we create a IntegerFilter extends Filter<Integer> - then it suddenly start working. That's why we had to generate that add simple Filter classes. And there were also an other problem, with the date formats too. But in the end, the generated API is pretty neat, I believe. I hope, some frontend ninja could add some generated panels for filtering :wink: 
So we can merge the server side library improvements, create a new library release, and then I can polish this patch to be merged.

jdubois:
@gzsombor I see you merged your code in the server lib:

- I'm doing some reformatting/documentation on it -> you need to add more documentation there, because it's really complex for people who don't now JPA
- I'm still not sure about the package name, that doesn't look very clear to me
- I'm going to do a release very soon as there is another change I want to have (upgrade to Swagger 2.7.0 as we need this for Spring Boot 2.0.0)

gzsombor:
Thanks, for creating a new release with the filter classes! 
There is always a possibility to improve the documentation, I agree :) However, deep inside me, I hoped, that when the developers would see the generated test methods, they will just experience a big _wow_, and _aha_, so no further explanation would be needed :wink: 
I'm going to work on it, to make this patch as simple, as possible, and improve the javadocs here and there

generators/entity/templates/server/src/main/java/package/repository/_EntityRepository.java
```diff
@@ -64,7 +64,7 @@
 <%_ } if (databaseType === 'sql' || databaseType === 'mongodb') { _%>
 @SuppressWarnings("unused")
 @Repository
-public interface <%=entityClass%>Repository extends <% if (databaseType === 'sql') { %>JpaRepository<% } %><% if (databaseType === 'mongodb') { %>MongoRepository<% } %><<%=entityClass%>,<%= pkType %>> {
+public interface <%=entityClass%>Repository extends <% if (databaseType === 'sql') { %>JpaRepository<% } %><% if (databaseType === 'mongodb') { %>MongoRepository<% } %><<%=entityClass%>,<%= pkType %>><% if (databaseType === 'sql') { %>, JpaSpecificationExecutor<<%=entityClass%>> <% } %> {
```
deepu105: shouldnt this be added only when filtering is enabled?

generators/entity/index.js
```diff
@@ -244,6 +245,11 @@ module.exports = EntityGenerator.extend({
         this.searchEngine = this.fileData.searchEngine || this.searchEngine;
         this.javadoc = this.fileData.javadoc;
         this.entityTableName = this.fileData.entityTableName;
+        if (this.databaseType === 'sql') {
```
deepu105: can you move this into a method in `generator-base-private` so that its can be reused below?

generators/entity/index.js
```diff
@@ -415,6 +426,11 @@ module.exports = EntityGenerator.extend({
             this.data.dto = this.dto;
             this.data.service = this.service;
             this.data.entityTableName = this.entityTableName;
+            if (this.databaseType === 'sql') {
```
deepu105: reuse from method

generators/entity/prompts.js
```diff
@@ -276,6 +277,36 @@ function askForTableName() {
...
+                {
+                    name: 'Dynamic filtering for the entities with JPA Static metamodel',
+                    value: 'jpaMetamodel'
+                }
+            ],
+            default: 0
+        }
+    ];
+    this.prompt(prompts).then((props) => {
+        this.filtering = props.filtering;
```
deepu105: I would like to make checking for stuff simpler, so may be can we also add a property called `this.jpaMetamodelFiltering = true` so that we can do `<% if (jpaMetamodelFiltering) {  %>` everywhere. I was thinking about refactoring the generator when I find time to follow the pattern of `this.options.elasticSearch`, `this.options.websocket` everywhere so that our EJS code becomes easier to maintain and refactor. This is how most other generators do options and I find it elegant.

generators/entity/templates/server/src/main/java/package/service/_EntityQueryService.java
```diff
@@ -0,0 +1,163 @@
...
+import <%=packageName%>.repository.search.<%= entityClass %>SearchRepository;<% } %>
+import <%=packageName%>.service.dto.<%= entityClass %>Criteria;
+<% if (dto === 'mapstruct') { %>
+import <%=packageName%>.service.dto.<%= entityClass %>DTO;
+import <%=packageName%>.service.mapper.<%= entityClass %>Mapper;<% } %>
+<%_ for (idx in fields) { if (fields[idx].fieldIsEnum === true) { _%>
+import <%=packageName%>.domain.enumeration.<%= fields[idx].fieldType %>;
+<%_ } } _%>
+<%_
+  function getSpecificationBuilder(fieldType) {
```
deepu105: can you move the functions to `generator-base-private` we need to keep our templates simpler

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+import io.github.jhipster.service.filter.StringFilter;
+import io.github.jhipster.service.filter.RangeFilter;
+_ if (fieldsContainInstant === true) { _%>
+import io.github.jhipster.service.filter.InstantFilter;<% } %><%_ if (fieldsContainLocalDate === true) { _%>
+import io.github.jhipster.service.filter.LocalDateFilter;<% } %><% if (fieldsContainZonedDateTime === true) { %>
+import io.github.jhipster.service.filter.ZonedDateTimeFilter;<% } %>
+
+<%_
+  const referenceFilterType = '' + pkType + 'Filter';
+  function filterableType(fieldType) {
```
deepu105: please move to `generator-base-private` and reuse

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -35,7 +35,10 @@
 import <%=packageName%>.repository.search.<%= entityClass %>SearchRepository;<% } if (dto === 'mapstruct') { %>
 import <%=packageName%>.service.dto.<%= entityClass %>DTO;
 import <%=packageName%>.service.mapper.<%= entityClass %>Mapper;<% } %>
-import <%=packageName%>.web.rest.errors.ExceptionTranslator;
+import <%=packageName%>.web.rest.errors.ExceptionTranslator;<% if (filtering === 'jpaMetamodel') { %>
```
deepu105: use `<%_ _%> ` on new line as its more readable

generators/entity/templates/server/src/main/java/package/service/_EntityQueryService.java
```diff
@@ -0,0 +1,163 @@
...
+     * Function to convert <%= criteria %> to a {@link Specifications}
+     */
+    private Specifications<<%= entityClass %>> createSpecification(<%= criteria %> criteria) {
+        Specifications<<%= entityClass %>> specification = Specifications.where(null);
+        if (criteria != null) {
+            if (criteria.getId() != null) {
+                specification = specification.and(buildSpecification(criteria.getId(), <%= entityClass %>_.id));
+            }
+            <%_
+            for (idx in fields) {
```
deepu105: `fields.forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/_EntityQueryService.java
```diff
@@ -0,0 +1,163 @@
...
+                const fieldInJavaBeanMethod = fields[idx].fieldInJavaBeanMethod;
+                const fieldValidateRules = fields[idx].fieldValidateRules;
+                if (filterableType(fieldType)) { _%>
+            if (criteria.get<%= fieldInJavaBeanMethod %>() != null) {
+                specification = specification.and(<%= getSpecificationBuilder(fieldType) %>(criteria.get<%= fieldInJavaBeanMethod %>(), <%= entityClass %>_.<%= fieldName %>));
+            }
+            <%_ }
+            }
+
+            for (idx in relationships) {
```
deepu105: `relationships.forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+      return fieldType !== 'byte[]' && fieldType !== 'ByteBuffer';
+  }
+  var filterVariables = [{name:'id', type: pkType, filterType:referenceFilterType,fieldInJavaBeanMethod:'Id' } ];
+  var extraFilters = {};
+  for (idx in fields) {
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        filterType = 'Filter<' + fieldType + '>';
+      }
+      filterVariables.push( { filterType : filterType,
+            name: fields[idx].fieldName,
+            type: fieldType,
+            fieldInJavaBeanMethod: fields[idx].fieldInJavaBeanMethod });
+    }
+  }
+  for (idx in relationships) {
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+
+<%_ for (idx in filterVariables) { _%>
+
+    private <%- filterVariables[idx].filterType %> <%= filterVariables[idx].name %>;
+<%_ } _%>
+
+    public <%= entityClass %>Criteria() {
+    }
+
+<%_ for (idx in filterVariables) { _%>
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Criteria class for the <%= entityClass %> entity. This class is used in <%= entityClass %>Resource to
+ * receive all the possible filtering options from the Http GET request parameters.
+ * For example the following could be a valid requests:
+ * <code> /<%= entityApiUrl %>?id.greaterThan=5&amp;attr1.contains=something&amp;attr2.specified=false</code>
+ * As Spring is unable to properly convert the types, unless specific {@link Filter} class are used, we need to use
+ * fix type specific filters.
+ */
+public class <%= entityClass %>Criteria implements Serializable {
+<%_ for (idx in extraFilters) { _%>
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Class for filtering <%= idx %>
+     */
+    public static class <%= extraFilters[idx].type %> extends <%- extraFilters[idx].superType %> {
+    }
+
+<%_ } _%>
+    private static final long serialVersionUID = 1L;
+
+<%_ for (idx in filterVariables) { _%>
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public void set<%= filterVariables[idx].fieldInJavaBeanMethod %>(<%- filterVariables[idx].filterType %> <%= filterVariables[idx].name %>) {
+        this.<%= filterVariables[idx].name %> = <%= filterVariables[idx].name %>;
+    }
+
+<%_ } _%>
+    @Override
+    public String toString() {
+        return "<%= entityClass %>Criteria{" +
+<%_ for (idx in filterVariables) {
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +460,111 @@ public void initTest() {
             <%_ } _%>
             .andExpect(jsonPath("$.<%=fields[idx].fieldName%>").value(<% if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { %>Base64Utils.encodeToString(<% } else if (fields[idx].fieldType === 'ZonedDateTime') { %>sameInstant(<% } %><%='DEFAULT_' + fields[idx].fieldNameUnderscored.toUpperCase()%><% if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { %><% if (databaseType === 'cassandra') { %>.array()<% } %>)<% } else if (fields[idx].fieldType === 'Integer') { %><% } else if (fields[idx].fieldType === 'Long') { %>.intValue()<% } else if (fields[idx].fieldType === 'Float' || fields[idx].fieldType === 'Double') { %>.doubleValue()<% } else if (fields[idx].fieldType === 'BigDecimal') { %>.intValue()<% } else if (fields[idx].fieldType === 'Boolean') { %>.booleanValue()<% } else if (fields[idx].fieldType === 'ZonedDateTime') { %>)<% } else { %>.toString()<% } %>))<% } %>;
     }
+<%_ if (filtering === 'jpaMetamodel') {
+        for (jdx in fields) {
```
deepu105: `forEach` would be more readable

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +460,111 @@ public void initTest() {
...
+        // Get all the <%= entityInstance %>List where <%= searchBy.fieldName %> is not null
+        default<%= entityClass %>ShouldBeFound("<%= searchBy.fieldName %>.specified=true");
+
+        // Get all the <%= entityInstance %>List where <%= searchBy.fieldName %> is null
+        default<%= entityClass %>ShouldNotBeFound("<%= searchBy.fieldName %>.specified=false");
+    }
+<%_
+            }
+            // the range criterias
+            if (['Byte', 'Short', 'Integer', 'Long', 'BigDecimal', 'Double', 'LocalDate', 'ZonedDateTime'].indexOf(searchBy.fieldType) !== -1) {  _%>
```
deepu105: use `includes` instead of indexOf

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +460,111 @@ public void initTest() {
...
     }
+<%_ if (filtering === 'jpaMetamodel') {
+        for (jdx in fields) {
+            const searchBy = fields[jdx];
+            // we can't filter by all the fields.
+            if (['byte[]', 'ByteBuffer'].indexOf(searchBy.fieldType) === -1) {
```
deepu105: use `includes` instead of `indexOf`

generators/server/templates/_pom.xml
```diff
@@ -919,6 +919,14 @@
                             <artifactId>mapstruct-processor</artifactId>
                             <version>${mapstruct.version}</version>
                         </path>
+<%_ if (databaseType === 'sql') { _%>
```
deepu105: shouldn't filtering also be checked?

generators/server/templates/gradle/_mapstruct.gradle
```diff
@@ -20,4 +20,7 @@ apply plugin: "net.ltgt.apt"
 
 dependencies {
     apt "org.mapstruct:mapstruct-processor:${mapstruct_version}"
+<%_ if (databaseType === 'sql') { _%>
```
deepu105: shouldn't filtering also be checked?

deepu105:
@gzsombor looks great. I have some comments for improvements and few queries

deepu105:
@gzsombor could you please take a look at the review comments

gzsombor:
Sorry, forgot to finish the changes properly, and push it. Hopefully, it works now

generators/server/templates/_pom.xml
gzsombor: As with the mapstruct processor, it's much simpler, to unconditionally generate all the JPA static metamodel classes, in the server generator, instead of adding, when the first entity is created with filtering.
 We can fix this, but I think, it became a bit too complicated.

deepu105: ahh, ok

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,108 @@
...
+  var filterVariables = [{name:'id', type: pkType, filterType:referenceFilterType,fieldInJavaBeanMethod:'Id' } ];
+  var extraFilters = {};
+  fields.forEach((field) => {
+    const fieldType = field.fieldType;
+    if (isFilterableType(fieldType)) {
+      var filterType;
+      if (field.fieldIsEnum == true) {
+        filterType = fieldType + 'Filter';
+        extraFilters[fieldType] = {type : filterType, superType: 'Filter<' + fieldType + '>', fieldType:fieldType};
+      } else if (fieldType === 'LocalDate' || fieldType === 'ZonedDateTime' || fieldType === 'Instant') {
```
deepu105: `['LocalDate', 'ZonedDateTime', 'Instant'].includes(fieldType)` would look nicer

generators/entity/templates/server/src/main/java/package/service/dto/_EntityCriteria.java
```diff
@@ -0,0 +1,108 @@
...
+        filterType = fieldType + 'Filter';
+      } else if (fieldType === 'String') {
+        filterType = 'StringFilter';
+      } else if (fieldType === 'Long') {
+        filterType = 'LongFilter';
+      } else if (fieldType === 'Integer') {
+        filterType = 'IntegerFilter';
+      } else if (fieldType === 'Float' || fieldType === 'Double' || fieldType === 'BigDecimal') {
```
deepu105: `['LocalDate', 'ZonedDateTime', 'Instant'].includes(fieldType)` pattern would look nicer

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +463,118 @@ public void initTest() {
             <%_ } _%>
             .andExpect(jsonPath("$.<%=fields[idx].fieldName%>").value(<% if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { %>Base64Utils.encodeToString(<% } else if (fields[idx].fieldType === 'ZonedDateTime') { %>sameInstant(<% } %><%='DEFAULT_' + fields[idx].fieldNameUnderscored.toUpperCase()%><% if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { %><% if (databaseType === 'cassandra') { %>.array()<% } %>)<% } else if (fields[idx].fieldType === 'Integer') { %><% } else if (fields[idx].fieldType === 'Long') { %>.intValue()<% } else if (fields[idx].fieldType === 'Float' || fields[idx].fieldType === 'Double') { %>.doubleValue()<% } else if (fields[idx].fieldType === 'BigDecimal') { %>.intValue()<% } else if (fields[idx].fieldType === 'Boolean') { %>.booleanValue()<% } else if (fields[idx].fieldType === 'ZonedDateTime') { %>)<% } else { %>.toString()<% } %>))<% } %>;
     }
+<%_ if (jpaMetamodelFiltering) {
+        for (jdx in fields) {
```
deepu105: forEach

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +463,118 @@ public void initTest() {
...
+        // Get all the <%= entityInstance %>List where <%= searchBy.fieldName %> is null
+        default<%= entityClass %>ShouldNotBeFound("<%= searchBy.fieldName %>.specified=false");
+    }
+<%_
+            }
+            // the range criterias
+            if (['Byte', 'Short', 'Integer', 'Long', 'LocalDate', 'ZonedDateTime'].includes(searchBy.fieldType)) { 
+              var defaultValue = 'DEFAULT_' + searchBy.fieldNameUnderscored.toUpperCase();
+              var biggerValue = 'UPDATED_' + searchBy.fieldNameUnderscored.toUpperCase();
+              if (searchBy.fieldValidate === true && searchBy.fieldValidateRules.indexOf('max') !== -1) {
```
deepu105: includes instead of indexOf

generators/entity/templates/server/src/test/java/package/web/rest/_EntityResourceIntTest.java
```diff
@@ -454,6 +463,118 @@ public void initTest() {
...
+<%_     } _%>
+
+    /**
+     * Executes the search, and checks that the default entity is returned
+     */
+    private void default<%= entityClass %>ShouldBeFound(String filter) throws Exception {
+        rest<%= entityClass %>MockMvc.perform(get("/api/<%= entityApiUrl %>?sort=id,desc&" + filter))
+            .andExpect(status().isOk())
+            .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8_VALUE))
+            .andExpect(jsonPath("$.[*].id").value(hasItem(<%= entityInstance %>.getId().intValue())))<% for (idx in fields) {%>
```
deepu105: forEach

generators/generator-base-private.js
```diff
@@ -540,4 +540,31 @@ module.exports = class extends Generator {
         }
         return jdl;
     }
+
+    /**
+     * Return the method name which converts the filter to specification
+     * @param {string} fieldType
+     */
+    getSpecificationBuilder(fieldType) {
+        if (fieldType === 'Integer' || fieldType === 'Long' || fieldType === 'Float' || fieldType === 'Double' || fieldType === 'BigDecimal' || fieldType === 'LocalDate' || fieldType === 'ZonedDateTime' || fieldType === 'Instant') {
```
deepu105: `['LocalDate', 'ZonedDateTime', 'Instant'].includes(fieldType)` pattern would look nicer

generators/server/templates/gradle/_mapstruct.gradle
deepu105: NA

gzsombor:
I think, I've fixed everything.

jdubois:
Yes I want to have this merged, but I won't have time until early September -> you could do it without me, I don't think there's any issue here

deepu105:
@gzsombor I can try to merge this, but can you look at why travis fails?

deepu105:
I restarted build, lets see

gzsombor:
sure, I will check

deepu105:
I think its a lint fail for js

gzsombor:
It was a major whitespace error :wink:

jdubois:
This will be in our next release - @gzsombor could you add some documentation in jhipster.github.io ? This is really not trivial stuff.

gzsombor:
Sure ! I'm trying to setup the jekyl stuff to generate the pages, but currently I can only see the 'Hyde' part - neither work on Ubuntu, nor the docker image works :disappointed:

jdubois:
Thanks @gzsombor -> don't worry for testing the pages, just write the content, I will test and correct

PR #5541, opened 2017-04-04
fix social sign in redirect in Angular 2

kaidohallik:
in AngularJs 1 both `/#/home` and `/#/` are working (route `home` is defined with path `/`)
in Angular 2 (4) only `/#/` is working (no route `home` exists, home component path is empty string)
so changing redirect to `/#/` which works both in AngularJs 1 and Angular 2 (4)


deepu105:
Sounds logical

jdubois:
OK then I'm merging this if the build works (I just restarted it)

PR #5543, opened 2017-04-05
Fix missing parameters on default docker-compose for application

stevehouel:
Add missing parameters for eureka service url based on docker-compose service name. By default it localhost, but in case of many instances, this will not work anymore. Also with that, we will have same configuration for config server and eureka server


PierreBesson:
Yes I also noticed the bug yesterday. Thanks for fixing this.

PR #5562, opened 2017-04-09
Route activate checks improvements

xiasenhai:
The principal data may lost when refreshing the page


xiasenhai:
i have an OAuth2 based UAA server, one service app , and a gateway.  after login when i refresh the page, the login modal popup out. refreshing home page seems ok.

Issue #5563, opened 2017-04-09
Can't  install other generators from the JHipster Marketplace + blank page for client project 


p4535992:
##### **Overview of the issue**

When i start to use `yo jhipster` and i decide to use the marketplace i get the following stacktrace exception, is strange because the first project i made has been built without a problem like in the video tuorial of the site, i put the image of the exception:

![jhipster_exception_use_add_generator](https://cloud.githubusercontent.com/assets/5201916/24836033/c56230ae-1d11-11e7-86fe-1c3ce050ce43.png)

if i choose to no use the marketplace the app is build fine but return on client (`mvnw` if maven on localhost:8080) a blank page wihtout  exception,  on server side (`yarn start` on localhost:9090) seems work fine instead

##### **Motivation for or Use Case**

Is a bug  (on my opinion) because is pure code "create a application" of jhipster i don't touch anything, is the default project i created with jhipster it's not working.

##### **Reproduce the error**

My steps on the creation of a application:

1. Install JHipster yarn global add generator-jhipster
2. Create a new directory and go into it mkdir myApp && cd myApp
3. Run JHipster and follow instructions on screen yo jhipster
4. Select Monolithic Application and enter
5. Select the name of my application and enter
6. Select  yes or no on the question "if  i like to install generators from the jhipster marketplace" and enter
   6.1 if i choose yes => **[ERROR]** Exception i showed in the image at the "Overview of issue" paragraph.
   6.2 if i choose no everything seems fine
7. Select other choice like database , version of angular i want to use , ecc.
..............
8. Start the application with mvnw (if maven)
9. **[ERROR]** blank page in all the browser 

I put a image with the all choices i made it for the project when i choose to no install some generator from the  marketplace:

![jhipster_exception_choice_propject](https://cloud.githubusercontent.com/assets/5201916/24836038/d3dd0b54-1d11-11e7-9819-2626ea00d073.png)

NOTE: I made a checkout to the database MySQL i'm using and seems all fine for confirmed that i put a image with the result of the run of the spring boot project with the ide intellijidea:
![jhipster_build_idea](https://cloud.githubusercontent.com/assets/5201916/24836116/c3676466-1d13-11e7-9d61-1798166f702c.jpg)

##### **Related issues**
For the exception launched to the select "yes i want to use the marketplace":

- can't find no one previuos issue related to this problem.

For the blanck page issue i found this reference (i read them but the solutions suggesteds not seems the case):

[...]

PR #5565, opened 2017-04-09
Fix marketplace when there is an empty author

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5563


deepu105:
fix seems ok but do we really need to show the auther name here? To me just name and description might be sufficient

pascalgrimaud:
Author name is not needed, so I agree with you. 
Do you want me to remove it ?

deepu105:
Lets just remove and keep it simple

pascalgrimaud:
Done!

PR #5574, opened 2017-04-10
allow state change when no authorities are required

### LINKED ISSUE #5542: Unable to register with angular 4

When trying to register a new account you can't access the page, because your are not allowed to open the register page

No registration possible

##### **Reproduce the error**

Generate a new monolithic applications using jwt and angular 4

##### **Related issues**

Nothing found.

##### **Suggest a Fix**

I think the problem is in the `checkLogin` in `UserRouteAccessService.ts`. When the method `hasAnyAuthority` is called on the current principal this method always returns `false` as there is no current principal yet. So from my point of view, when a route has empty authorities (`[ ]`) like `register` it should return always true

##### **JHipster Version(s)**

Latest master.

##### **JHipster Version(s)**

```
matetracker@0.0.0 /home/fred/git/matetracker
├── UNMET PEER DEPENDENCY @angular/common@4.0.0
├── UNMET PEER DEPENDENCY @angular/core@4.0.0
├── generator-jhipster@4.1.1 
└── UNMET PEER DEPENDENCY zone.js@0.8.2

```

```yaml
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.gitlab.atomfrede.matetracker",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.1.1",
    "baseName": "matetracker",
    "packageName": "com.gitlab.atomfrede.matetracker",
    "packageFolder": "com/gitlab/atomfrede/matetracker",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "jwtSecretKey": "cde6ceb8bd1abd7260f20b43a6a1005b95960a7f",
    "clientFramework": "angular2",
    "useSass": false,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "de"
    ]
  }
}
```

ls: no such file or directory: .jhipster/*.json

java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

git version 2.12.2

node: v6.9.5

npm: 3.10.10

bower: 1.6.5

gulp:
[07:39:55] CLI version 3.9.0

yeoman: 1.8.5

yarn: 0.19.1

Docker version 17.03.1-ce, build c6d412e329

docker-compose version 1.11.2, build dfed245b

No entities generated.

Linux with latest Firefox

Of course I can provide a PR.

atomfrede:
* When the state doesn't require any authority `true` is returned
* Someone should check if the login modal opens multiple times (sometimes I hat that, sometimes not, I looks like it depends on the requested, protected page/route)
* I am open for better suggestions
close #5542

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
@@ -16,6 +16,11 @@ export class UserRouteAccessService implements CanActivate {
     }
 
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Promise<boolean> {
+
+        if (route.data['authorities'].length == 0) {
+            return true;
+        }
+
         return this.checkLogin(route.data['authorities'], state.url);
     }
 
--- a/generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
@@ -17,11 +17,12 @@ export class UserRouteAccessService implements CanActivate {
 
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Promise<boolean> {
 
-        if (route.data['authorities'].length == 0) {
+        const authorities = route.data['authorities'];
+        if (authorities && authorities.length == 0) {
             return true;
         }
 
-        return this.checkLogin(route.data['authorities'], state.url);
+        return this.checkLogin(authorities, state.url);
     }
 
     checkLogin(authorities: string[], url: string): Promise<boolean> {


generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
```diff
@@ -16,6 +16,11 @@ export class UserRouteAccessService implements CanActivate {
     }
 
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Promise<boolean> {
+
+        if (route.data['authorities'].length == 0) {
```
gmarziou: Could we have the case with `route.data['authorities']` being undefined?

xiasenhai: we definitely could have!
`route.data['authorities'] && route.data['authorities'].length == 0`

delverdev: It would be better if a variable `authorities` would be created (3 usages).

generators/client/templates/angular/src/main/webapp/app/shared/auth/_user-route-access-service.ts
```diff
@@ -16,7 +16,13 @@ export class UserRouteAccessService implements CanActivate {
     }
 
     canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Promise<boolean> {
-        return this.checkLogin(route.data['authorities'], state.url);
+
+        const authorities = route.data['authorities'];
+        if (authorities && authorities.length == 0) {
```
ruddell: If authorities is undefined, we should return true as well.  Otherwise, another error will be thrown later in `Principal.hasAnyAuthority` when looping the array.  Maybe something like this?
```
if (!authorities || authorities.length == 0) {
```

atomfrede: You are absolutely right.

jdubois: @atomfrede if that's fine with you I'll do the change while merging

jdubois:
I'm going to merge this -> is everyone OK that this solves #5542 ? Sorry I didn't follow this closely.

Then I'll probably be able to do a release "soon"

Issue #5485, opened 2017-03-25
jhipster:upgrade failed


ndywicki:
##### **Overview of the issue**

Hello,

I've a little bug with new jhipster upgrade.
During the upgrade the package.json is modified and uncommited in the current branch -> upgrade failed.

I've commited manually package.json and relaunch the upgrade command but the  process was always blocking after 2 hours to "Updating generator-jhipster to 4.1.1 . This might take some time...".

I've stopped the process, and relaunched the upgrade. Finally terminaned successfully in few secondes.
3 files was in conflicts. I resolved the conflicts and now all seem ok.

Note: May be it's normal but I notice also that :
`yo jhipster:upgrade` not upgrade the generator-jhipster npm global package :
```
λ npm list -g | grep jhipster
+-- generator-jhipster@4.0.8
| +-- jhipster-core@1.2.8
```


Full output : 
[upgrade_jhispter.txt](https://github.com/jhipster/generator-jhipster/files/870154/upgrade_jhispter.txt)



##### **Motivation for or Use Case**

Use upgrade tool to make Jhipster upgrade more easly.

##### **Reproduce the error**

Use `yo jhipster:upgrade` on jhipster v4.0.8.

##### **Related issues**

N/A

##### **Suggest a Fix**

Update script to commit package.json before check of the unversionning files

##### **JHipster Version(s)**

Project with Jihipster v4.0.8

`λ yo jhipster:info                                                                                               
Welcome to the JHipster Information Sub-Generator                                                                
                                                                                                                 
##### **JHipster Version(s)**                                                                                    
                                                                                                                 
```                                                                                                              
testupgrade@0.1.0 D:\workspaces\jhipster\testupgrade                                                             
+-- UNMET PEER DEPENDENCY @angular/compiler@2.4.9                                                                
+-- UNMET PEER DEPENDENCY @angular/core@2.4.9                                                                    
+-- generator-jhipster@4.1.1                                                                                     
`-- UNMET PEER DEPENDENCY zone.js@0.8.2                                                                          
                                                                                                                 
```                                                                                                              
                                                                                                                 
                                                                                                                 
##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**                              
                                                                                                                 
```yaml                                                                                                          
{                                                                                                                
  "generator-jhipster": {                                                                                        
    "jhipsterVersion": "4.1.1",                                                                                  
    "baseName": "testupgrade",                                                                                   
[...]

PR #5583, opened 2017-04-11
improve upgrade subgenerator

ruddell:
Adds two command line flags:
 - Verbose option to log everything that's going on during the upgrade process
    - `--verbose` 
 - Allow the user to specify the target version to upgrade to.  Currently it always upgrades to the max version which can be problematic for older versions. 
    - `--target-version=4.2.0`  

I can PR the docs if everyone is ok with the names of the flags.  Hopefully this helps with #5485


jdubois:
Excellent, thanks a lot for this!! I'm worried this sub-generator isn't working as good as it should.

PR #5588, opened 2017-04-12
return bad request on PUT if ID is not provided in entity

ctamisier:
I would suggest to return a bad request when an update of an entity is performed when the ID of the entity is not provided.

Currently if the ID is null, the call is redirect to an entity creation.
I think if the client perform a PUT and does not provide an ID, that means there is a mistake somewhere and this action shouldn't be considered as an entity creation.

What do you think?


ctamisier:
i reread about the PUT and it could be used for creation as well so we can ignore this PR..

PR #5591, opened 2017-04-13
Angular dependency update

deepu105:
Also has some lint fixes


jdubois:
I shouldn't have merged this, the build is broken :-(

jdubois:
The bug seems to be caused by Angular 4.0.2, I'm downgrading it

deepu105:
I was waiting for build to pass

PR #5595, opened 2017-04-14
add validation to forbid relationship's first letter to be capital

### LINKED ISSUE #5594: Entity generates wrong mappedBy attribute

When creating relationships with a first letter capitalized name, the entity generator generates a value for mappedBy and a value for the bean attribute on the other side which differ from one another, resulting in an exception as hibernate cannot match to two.

Here is such an exception : `Caused by: org.hibernate.AnnotationException: mappedBy reference an unknown target entity property: com.mycompany.myapp.domain.Book.TheAuthor in com.mycompany.myapp.domain.Author.hisBooks
`

Here are the configuration files for such a case : 
```json
{
    "fluentMethods": true,
    "relationships": [
        {
            "relationshipName": "HisBooks",
            "otherEntityName": "book",
            "relationshipType": "one-to-many",
            "otherEntityRelationshipName": "TheAuthor"
        }
    ],
    "fields": [
        {
            "fieldName": "name",
            "fieldType": "String"
        },
        {
            "fieldName": "birtDate",
            "fieldType": "LocalDate"
        }
    ],
    "changelogDate": "20170414080213",
    "dto": "no",
    "service": "no",
    "entityTableName": "author",
    "pagination": "no"
}
{
    "fluentMethods": true,
    "relationships": [
        {
            "relationshipName": "TheAuthor",
            "otherEntityName": "author",
            "relationshipType": "many-to-one",
            "otherEntityField": "name"
        }
    ],
    "fields": [
        {
            "fieldName": "title",
            "fieldType": "String"
        },
        {
            "fieldName": "description",
            "fieldType": "String"
        },
        {
            "fieldName": "publicationDate",
            "fieldType": "LocalDate"
        },
        {
            "fieldName": "price",
            "fieldType": "BigDecimal"
        }
    ],
    "changelogDate": "20170414080357",
    "dto": "no",
    "service": "no",
    "entityTableName": "book",
    "pagination": "no"
}
```

This configuration results in :
`@OneToMany(mappedBy = "TheAuthor")` on the `Author.java` side and `private Author theAuthor;` on the `Book.java` side.

By simply answering the prompt questions, I get an application that won't build. The fix is really easy but it should be automatic (set the first letter of the mappedBy attribute to lowercase).

##### **Reproduce the error**

1. Generate an application
2. Generate the entity Author using above information (or copy json and regenerate)
3. Generate the entity Book using above information (or copy json and regenerate)
4. `mvn test`

##### **Related issues**

I believe "MappedBy reference an unknown target entity property (bidirecional) #1528" is about the same problem but the author didn't provide enough information.

##### **Suggest a Fix**

I suggest preventing the user from providing a relationship name with a first letter capital at the validation step, exactly as for fields.
I think this is the best solution as : 
1. Entity generator put the first letter lowercase for the bean attribute which respects java convention.
2. It's consistent to the way the entity generator proceed with the fields name.
3. It's really easy to do.

Another solution would be to use another variable in the template (the same variable as for the java bean I guess).

**Tell me if one suits you and I make the PR**

##### **JHipster Version(s)**

I found the bug on 4.2 and reproduced it on 4.3.

AdrienHorgnies:
also fix a typo

closes #5594

### DIFF
--- a/generators/entity/prompts.js
+++ b/generators/entity/prompts.js
@@ -415,7 +415,7 @@ function askForField(done) {
                 } else if (input === '') {
                     return 'Your field name cannot be empty';
                 } else if (input.charAt(0) === input.charAt(0).toUpperCase()) {
-                    return 'Your field name cannot start with a upper case letter';
+                    return 'Your field name cannot start with an upper case letter';
                 } else if (input === 'id' || fieldNamesUnderscored.indexOf(_.snakeCase(input)) !== -1) {
                     return 'Your field name cannot use an already existing field name';
                 } else if (!skipServer && jhiCore.isReservedFieldName(input)) {
@@ -916,6 +916,8 @@ function askForRelationship(done) {
                     return 'Your relationship cannot contain special characters';
                 } else if (input === '') {
                     return 'Your relationship cannot be empty';
+                } else if (input.charAt(0) === input.charAt(0).toUpperCase()) {
+                    return 'Your relationship cannot start with an upper case letter';
                 } else if (input === 'id' || fieldNamesUnderscored.indexOf(_.snakeCase(input)) !== -1) {
                     return 'Your relationship cannot use an already existing field name';
                 } else if (jhiCore.isReservedKeyword(input, 'JAVA')) {


jdubois:
Thanks & kudos for your 1st contribution to the project!

AdrienHorgnies:
Technically not the first one :-D Well, the first one was really cheap : https://github.com/jhipster/generator-jhipster/commit/0b9dd6b2253a07723af343c68897c9b30ebc0657 .

PR #5606, opened 2017-04-16
log exception with stacktrace

ctamisier:
Logs the exception with the stacktrace when exception is handled by `@ExceptionHandler(Exception.class)`.

In production, a RuntimeException that leads for the client to a `ErrorConstants.ERR_INTERNAL_SERVER_ERROR` should be logged server side (message and stacktrace) to know what is wrong

WDYT?

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/errors/_ExceptionTranslator.java
+++ b/generators/server/templates/src/main/java/package/web/rest/errors/_ExceptionTranslator.java
@@ -20,6 +20,8 @@ package <%=packageName%>.web.rest.errors;
 
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.AnnotationUtils;
 <%_ if (databaseType != 'no' && databaseType != 'cassandra') { _%>
 import org.springframework.dao.ConcurrencyFailureException;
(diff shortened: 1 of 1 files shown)


gmarziou:
Logging seems a good idea, however method should not be named `processRuntimeException` as they are not `RuntimeException` and also I wuold log some details from request:

~~~java
     @ExceptionHandler(Exception.class)
      public ResponseEntity<ErrorVM> processException(Exception ex, HttpServletRequest request) {		      
          log.error("Error while processing {}", request.getRequestURL(), ex);
~~~

Logging the HTTP verb would make sense also.

I'm surprised we have to do so, I would have expected the default exception handler to do so but I haven't checked.

ctamisier:
what is logged so far is:
`2017-04-16 18:38:24.317  WARN 1909 --- [  XNIO-2 task-1] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by Handler execution: java.lang.RuntimeException: uhoh`

when this is thrown:
`throw new RuntimeException("uhoh");`

For me I would like to have more information about where this is thrown. Because being an unhandled exception (not handled by ExceptionTranslator), I want to know where in the code this is happening to be able to "debug it". That's why I'm thinking to log the stacktrace to target the bug.

ctamisier:
Another example:

If I have a NullPointerException with the `prod` profile, we just have this in the logs:
`2017-04-16 18:50:12.881  WARN 1996 --- [  XNIO-2 task-1] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by Handler execution: java.lang.NullPointerException
`

No problem with the `dev` profile because LoggingAspect logs the stacktrace.

gmarziou:
You don't have to convince me. Of course, this is useful.
My point was that this is the default behaviour in spring boot and I did not understand why we did not use it, now I understand that this is required to fill up ErrorVM

PR #5609, opened 2017-04-17
user validation in WS layer

ctamisier:
Validates the `managedUserVM` in the WS layer (like in AccountResource)

- [] Tests are added where necessary, **supplementary tests could be written when having a null or empty login for example but it might not be worth the amount of code for this.., let me know**

### DIFF
--- a/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
+++ b/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
@@ -43,6 +43,7 @@ public class UserDTO {
         if (enableSocialSignIn) {
             columnMax = 100;
         } _%>
+    @NotNull
     @Pattern(regexp = Constants.LOGIN_REGEX)
     @Size(min = 1, max = <%=columnMax %>)
     private String login;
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
@ctamisier : Sonar is not happy. https://sonarqube.com/component_issues?id=io.github.jhipster.sample%3Ajhipster-sample-application#resolved=false|types=BUG

Can you have a look plz ?

sendilkumarn:
Can we remove @NotNull from the DTO

ctamisier:
instead, we can use `@NotBlank` from `hibernate-validator` which is more accurate (not null neither empty neither white spaces), and sonar doesn't complain with it (tried with IntelliJ sonar lint)
would it be ok for you ?

if we remove `@NotNull`, we'll have a NPE here : `managedUserVM.getLogin().toLowerCase()`, and I think this kind of validation should occur as soon as possible by using the `@Valid` on the WS.

cbornet:
It feels weird to change the annotation just to workaround Sonar.
It's not the first time we get this issue. Is this Sonar rule really relevant ? We generally don't want validation at object construction (which is often done by a mapper such as mapstruct of jackson) but later when it's checked by @Valid.

cbornet:
Other point : when referencing an entity in a relationship, you often only care about its ID/PK.
You would do for instance
```java
User user = new User();
user.setId(userId);
foo.setUser(user);
fooRepository.save(foo);
```
Of course user has some fields that are non-null in DB, but you're not going to fetch the DB just to fill the class !

PR #5622, opened 2017-04-18
Disable angular-cli webpack commands

gmarziou:
Currently in generated project, users can still run "ng build", "ng serve" and "ng test" which can bring confusion as these commands do nt have the required options for JHipster (api proxy, ...).

This change adds `"ejected": true` in `.angular-cli.json` so that users can no longer use angular-cli build process but still use other commands like "ng generate".

If a user tries "ng build" on an ejected project, an error is thrown: "An ejected project cannot use the build command anymore."

See https://github.com/angular/angular-cli/wiki/eject#ng-eject

### DIFF
--- a/generators/client/templates/angular/_.angular-cli.json
+++ b/generators/client/templates/angular/_.angular-cli.json
@@ -17,9 +17,10 @@
  limitations under the License.
 -%>
 {
+    "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
     "project": {
-        "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
-        "name": "<%= dasherizedBaseName %>"
+        "name": "<%= dasherizedBaseName %>",
+        "ejected": true
     },
     "apps": [{
         "root": "<%= MAIN_SRC_DIR %>",


stevehouel:
@gmarziou I think that Angular Cli is required and can do a lot of tasks. But as you say, having both AngularCli and Webpack is not a good thing and can make confusion for users. Maybe should we remove Webpack work and only use AngularCli for build/serve/test tasks. I am currently working on it

pascalgrimaud:
@stevehouel : plz hold on about "Maybe should we remove Webpack work and only use AngularCli for build/serve/test tasks. I am currently working on it" -> I think it should be discussed first

stevehouel:
Yes no worries @pascalgrimaud I am just working on it on my side for my personnal use. I will no do any PR before everyone validate that.

gmarziou:
@stevehouel of course if we could rely on webpack embedded in angular-cli that would be perfect, less work for us with still the possibility for our user to customize it further once they'd use "ng eject". In fact this is what is done in https://github.com/giovannicandido/generator-spring-spa

deepu105:
@stevehouel yes if we can get it to work perfectly for our setup it would be great. Can you open a separate ticket and make your proposal there. If you can get Angular cli to do all of our webpack stuff then we can definitely switch to that.

stevehouel:
@deepu105 Ok i will open a ticket for that and try to list all current webpack work.

gmarziou:
Let's close it as I'm confident @stevehouel wil succeed :)

PR #5628, opened 2017-04-19
use @NotBlank instead of @NotNull for the login

ctamisier:
fix for sonar warning about @NotNull

https://github.com/jhipster/generator-jhipster/pull/5609

### DIFF
--- a/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
+++ b/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
@@ -24,6 +24,7 @@ import <%=packageName%>.domain.Authority;<% } %>
 import <%=packageName%>.domain.User;
 
 import org.hibernate.validator.constraints.Email;
+import org.hibernate.validator.constraints.NotBlank;
 
 import javax.validation.constraints.*;
 <%_ if (databaseType == 'mongodb' || databaseType == 'sql') { _%>
@@ -43,7 +44,7 @@ public class UserDTO {
         if (enableSocialSignIn) {
             columnMax = 100;
         } _%>
-    @NotNull
+    @NotBlank
     @Pattern(regexp = Constants.LOGIN_REGEX)
     @Size(min = 1, max = <%=columnMax %>)
     private String login;


ctamisier:
This PR conflicts with https://github.com/jhipster/generator-jhipster/pull/5627

sendilkumarn:
I closed the other.

cbornet:
@pascalgrimaud did you see https://github.com/jhipster/generator-jhipster/pull/5609#issuecomment-295587193 ?

pascalgrimaud:
Sorry, I didn't read it closely.
@cbornet : so do you think we should stay as before, without @NotBlank and without @NotNull ?

cbornet:
No, keep @NotNull and exclude the sonar rule.
If we consider the sonar rule valid, we shouldn't replace NotNull by NotBlank just because it's not detected.

pascalgrimaud:
Can you PR your proposal @cbornet plz ? but do you think everybody would be ok for excluding a Sonar rule ?

As you suggested in the Mailing List, It would be good to have a sonar analyze before merging PR, to try to  keep our Sonar Analyze as best as possible: https://sonarqube.com/dashboard?id=io.github.jhipster.sample%3Ajhipster-sample-application
But it's hard to find time for everyone...

cbornet:
I can PR but I would like to have others opinion before.
@jhipster/developers

jdubois:
Unfortunately it should work with Sonar "out of the box", otherwise people will have all those "bugs" after generation.

cbornet:
Well, we already exclude some sonar rules if I'm not mistaken

jdubois:
I know one, but that was a bug in Sonar, so I sent them a bug report and discussed the issue with the Sonar team on their mailing list : so this is supposed to be temporary, until they fix it.

cbornet:
The thing is : it seems weird to me to use NotBlank instead of NotNull just to avoid being detected by Sonar.
The issue still exists, it's just that Sonar doesn't detect it because nobody wrote the rule yet. And it could change tomorrow...

Tcharl:
I'm not for the NotBlank: it's a good practice to stick to the reference instead of vendor specific stuff

Issue #5618, opened 2017-04-18
Removing IntelliJ configuration file while using yo jhipster:upgrade


Ckram:
##### **Overview of the issue**

When using yo jhipster:upgrade for the first time (when no jhipster_upgrade branch exists yet) it deletes the IDE (IntelliJ IDEA) configuration files

##### **Motivation for or Use Case**

It should not delete them because  they should be ignored.

##### **Reproduce the error**

run yo jhipster:upgrade
Then IntelliJ does not recognize the project

 I think its done in the cleanUp() of the subGenerator.

{
  "generator-jhipster": {
    "jhipsterVersion": "4.3.0",
    "baseName": "name",
    "packageName": "com.package.name",
    "packageFolder": "com/package/name",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "rememberMeKey": "",
    "clientFramework": "angular1",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [
      "gatling",
      "cucumber",
      "protractor"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "fr",
    "languages": [
      "fr",
      "en"
    ]
  }
}

##### **Browsers and Operating System**

java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

git version 2.10.0

node: v6.10.0

npm: 4.3.0

bower: 1.8.0

gulp:
[10:25:16] CLI version 1.2.2
[10:25:16] Local version 3.9.1

yeoman: 1.8.5

yarn: 0.21.3

- [x] Checking this box is mandatory (this is just to show you read everything)

Skit5:
Have you tried to ignore them like explained here ?
https://jhipster.github.io/configuring-ide-idea/

Ckram:
Yes I had to do it again after the upgrade to let idea recognize my files

pascalgrimaud:
[...]

PR #5638, opened 2017-04-22
Added .idea to the ignored files for jhipster-upgrade

Ckram:
Pull request for the issue https://github.com/jhipster/generator-jhipster/issues/5618

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -46,7 +46,7 @@ module.exports = UpgradeGenerator.extend({
 
     _cleanUp() {
         shelljs.ls('-A').forEach((file) => {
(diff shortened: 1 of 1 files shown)


cbornet:
We should find a more generic solution for this. Your IDE could be Eclipse or VSCode and the files to ignore would be different.

Ckram:
Maybe we can do it with each of them, Or ignore all the files/directories starting with a dot ?

deepu105:
What about reading the .gitignore file of the project and ignoring them along with the current ones?

jdubois:
Those other solutions would be very hard to do... Ignoring just a few files like this looks a lot easier to me, and at least that fixes the issue right now, so I'm happy with it.

PR #5640, opened 2017-04-22
Reduce code duplication on entity save response handling

gzsombor:
The same code is repeated twice for updating and inserting entity - this can be refactored to a separate function

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
@@ -28,6 +28,7 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Response } from '@angular/http';
 
+import { Observable } from 'rxjs/Rx';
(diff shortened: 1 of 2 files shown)


gzsombor:
Not clear, why the travis builds are failing

pascalgrimaud:
@gzsombor : the failures come from protractor. Travis doesnt like when there are too much login/logout... nothing related to your PR
I will try to improve that the next days

deepu105:
@pascalgrimaud may instead of login/logout per entity we can do it once for all entities together

pascalgrimaud:
@deepu105 : so 1 file for protractor, with entities, and we could add a needle in it:
- login
- test entity 1
- test entity 2
- (...)
- logout

Sounds good!

deepu105:
Im not talking about 1 file. I would still like 1 file per entity otherwise
its difficult to maintain for real use case but somehow if we can do the
login logout outsude of each file it will be faster and better. Ideally if
we could login/logout once per run that would be great

pascalgrimaud:
Sure @deepu105 there are different solutions and we can improve the protractor tests!

gzsombor:
Anyway, so there is no opposition agains merging this patch ? 😄

Issue #5586, opened 2017-04-12
[Suggestion] Introduce generic method names for MapStruct Object Mappers


jCube1980:
### Motivation for or Use Case

The interfaces generated by jhipster that are annotated with `org.mapstruct.Mapper`
have methods of the form

```
<%= entityClass %>DTO <%= entityInstance %>To<%= entityClass %>DTO(<%= entityClass %> <%= entityInstance %>);

<%= entityClass %> <%= entityInstance %>DTOTo<%= entityClass %>(<%= entityClass %>DTO <%= entityInstance %>DTO);
```

examples

```
CustomerDTO customerToCustomerDTO(Customer customer);
Customer customerDTOToCustomer(CustomerDTO customerDTO);
```

The proposal/suggestion is to use a generic base interface so that method names can be **standardized**. 

```
public interface EntityMapper <D, E> {
	E toEntity(D dto); //suggested method names. Feel free to change.
	D toDto(E entity);
	List <E> toEntity(List<D> dtoList);
	List <D> toDto(List<E> entityList);
}

@Mapper(componentModel = "spring", uses = {})
public interface CustomerMapper extends EntityMapper<CustomerDTO, Customer> {}
```

### Advantages

1) Standardized method names always lead to a better developer experience. 
2) First up, developers need to remember fewer methods!
3) Developers don't have to visually select the method when using content assist. (They can select using memory)
4) Simplifies jhipster core implementation.

gmarziou:
Are you sure this is technically possible with MapStruct? 

As far as I know MapStruct uses method names to generate code similarly to Spring JPA repository `findByNameAndEmail()`.

Have you made a proof of concept?

jCube1980:
Yes sir, it will work. The code sample pasted above is based on what I tried out. MapStruct and spring-data-JPA are markedly different in their implementation approaches. MapStruct generates code at compile time. Spring data JPA does not generate code and the proxy-ing (obviously) happens at run time.

A detailed explanation of how spring-data-jpa works behind the scene can be found [here](http://stackoverflow.com/questions/38509882/how-are-spring-data-repositories-actually-implemented)

MapStruct actually generates code during compilation phase and for object mapping, it need not look beyond the parameter(s) and return type of the interface methods. The name can be any valid Java method name.

If the concern is about how generated methods resolve other mapping methods internally, the logic is explained [here](http://mapstruct.org/documentation/dev/reference/html/#mapping-method-resolution). Here also, method names are not used.

[...]

PR #5642, opened 2017-04-22
change mapper interface method names to generic names

jCube1980:
closes https://github.com/jhipster/generator-jhipster/issues/5586

Travis [build](https://travis-ci.org/jCube1980/generator-jhipster/builds/224716153) is successful and generated apps work fine.

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
+++ b/generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
@@ -21,16 +21,18 @@ package <%=packageName%>.service.mapper;
 import java.util.List;
 
 /**
- * Contract for a generic entity to dto mapper.
+ * Contract for a generic dto to entity mapper.
+ @param <DTO> - DTO type parameter.
+ @param <ENTITY> - Entity type parameter.
  */
 
-public interface EntityMapper <D, E> {
+public interface EntityMapper <DTO, ENTITY> {
 
-    public E toEntity(D dto);
+    public ENTITY toEntity(DTO dto);
 
-    public D toDto(E entity);
+    public DTO toDto(ENTITY entity);
 
-    public List <E> toEntity(List<D> dtoList);
+    public List <ENTITY> toEntity(List<DTO> dtoList);
 
-    public List <D> toDto(List<E> entityList);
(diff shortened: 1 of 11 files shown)


generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
```diff
@@ -0,0 +1,36 @@
...
+-%>
+package <%=packageName%>.service.mapper;
+
+import java.util.List;
+
+/**
+ * Contract for a generic entity to dto mapper.
+ */
+
+public interface EntityMapper <D, E> {
```
gzsombor: I prefer to have a readable type arguments, eg : EntityMapper<DTO, ENTITY>

gzsombor:
Other than this one minor issue, it's LGTM +1

generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
jCube1980: I fully agree with the intention of keeping the type names readable/descriptive. 

I followed the Sun/Oracle [convention](http://stackoverflow.com/questions/2900881/generic-type-parameter-naming-convention-for-java-with-multiple-chars) which recommends "single, uppercase letters for type parameters". This helps developers effortlessly distinguish the difference between a type variable and an ordinary class or interface name. Also, this convention guarantees consistency. 

Another option I came across is adding details in the javadoc comments section. This approach follows the convention (thereby removing the possible confusion) and adds the detail you wanted. 

```
/**
 Contract for a generic dto to entity mapper.
 @param <D> - DTO
 @param <E> - Entity
 */
public interface EntityMapper <D, E> {
```

If it doesn't look good, I can change it as per your suggestion.

jCube1980:
Hi @gzsombor, I have changed the generic type parameter names as you had suggested. However the Travis build is failing due to a timeout issue. Any chance to re-trigger just the two steps that are failing? I re-triggered the failing steps in my fork and it completed successfully.

pascalgrimaud:
I'm restarting the 2 jobs

jCube1980:
The 2 jobs failed again.. :(

PR #5645, opened 2017-04-23
update to swagger ui v3

deepu105:

@jhipster/developers I was able to update to latest swagger UI v3 but the documentation isnt that great and I'm not sure on how to get the Auth working for JWT/OAUTH it works for Session.
Alos I think we need to write a plugin to have the dropdown instead of the input for url

If any of you know how to solve the auth issue plz help out

image


generators/client/templates/angular/src/main/webapp/swagger-ui/_index.html
```diff
@@ -17,80 +17,118 @@
...
+                        oauth2RedirectUrl: './dist/oauth2-redirect.html',
+                        validatorUrl: data.validatorUrl,
+                        supportedSubmitMethods: data.supportedSubmitMethods,
+                        docExpansion: data.docExpansion,
+                        apisSorter: data.apisSorter,
+                        showRequestHeaders: data.showRequestHeaders
+                    });
+                }
+
+                // TODO
```
deepu105: These needs to be fixed

PierreBesson:
I'm sad about this because the new UI is really ugly but then this is not out fault and we will need this to move to swagger 3...

gmarziou:
Shouldn't e wait fro Springfox to support it?

PierreBesson:
@gmarziou I thibk it's not a problem since this version of the swagger UI is fully compatible with Swagger (OpennAPI) version 2.

cbornet:
See https://github.com/swagger-api/swagger-ui/issues/2915 . It might be too early to migrate to Swagger UI 3

deepu105:
You are right lets wait then.


Thanks & Regards,
Deepu

PierreBesson:
I did review the swagger UI v3 when it came out a few weeks ago and apart from having terrible UI. It seems that it has a number of regressions : https://github.com/swagger-api/swagger-ui/labels/3.x
Also it looks to be a complete rewrite by some junior devs at SmartBear which explains why it looks a bit ugly for now (However big 👍  for them if they are reading this, this is still good work).
Still it is going in the right direction as this new architecture will be easier to integrate, so I suggest we wait for a few month before moving to it.

Also I had some projects to improve the Swagger UI integration in out gateway, maybe it will be the right time to do this as well.

MathieuAA:
@PierreBesson don't worry, the v2 code looked as ugly as it currently is ;)

jdubois:
@deepu105 are you confident in merging this? Did you check that it works with the Gateway, and that the microservices APIs are available from it?

deepu105:
lets not do this now. lets wait for it to become stable

PR #5647, opened 2017-04-23
Fix "sign in" left align when langage different from english

sabrinapayet:
When language selected is french, "Sign in" left-align is not well displayed.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
@@ -23,7 +23,7 @@
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-4">
+        <div class="col-md-4 offset-md-2">
             <h1 jhiTranslate="login.title">Sign in</h1>
         </div>
         <div class="col-md-8 offset-md-2">


generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
```diff
@@ -23,7 +23,7 @@ <h4 class="modal-title" jhiTranslate="login.title">Sign in</h4>
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-4">
+        <div class="col-md-4 offset-md-2">
```
kaidohallik: bootstrap grid size is 12, but 4+2*2=8 <> 12  
maybe better is `class="col-md-6 offset-md-3"`

pascalgrimaud: Not really beautiful with `class="col-md-6 offset-md-3"`. I only tested it in french

kaidohallik: OK. Seems that `class="col-md-8 offset-md-2"` is better solution. With this English, French and Estonian looks good.

kaidohallik: I'll create new PR to fix longer multi word title like Estonian

PR #5655, opened 2017-04-24
improve sign in window

kaidohallik:
discussed in #5647

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
@@ -23,10 +23,11 @@
 </div>
 <div class="modal-body">
     <div class="row">
+        <%_ if (enableSocialSignIn) { _%>
+        <div class="col-12">
+        <%_ } else { _%>
         <div class="col-md-8 offset-md-2">
-            <h1 jhiTranslate="login.title">Sign in</h1>
-        </div>
-        <div class="col-md-8 offset-md-2">
+        <%_ } _%>
             <div class="alert alert-danger" *ngIf="authenticationError" jhiTranslate="login.messages.error.authentication">
                 <strong>Failed to sign in!</strong> Please check your credentials and try again.
             </div>
@@ -60,7 +61,7 @@
                 <a class="alert-link" (click)="requestResetPassword()" jhiTranslate="login.password.forgot">Did you forget your password?</a>
             </div>
             <div class="alert alert-warning">
-            	<span jhiTranslate="global.messages.info.register.noaccount">You don't have an account yet?</span>
+                <span jhiTranslate="global.messages.info.register.noaccount">You don't have an account yet?</span>
                 <a class="alert-link" (click)="register()" jhiTranslate="global.messages.info.register.link">Register a new account</a>
             </div>
         </div>
--- a/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
@@ -23,7 +23,7 @@
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-2">
+        <div class="col-md-8 offset-md-2">
             <h1 jhiTranslate="login.title">Sign in</h1>
         </div>
         <div class="col-md-8 offset-md-2">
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
```diff
@@ -23,7 +23,7 @@ <h4 class="modal-title" jhiTranslate="login.title">Sign in</h4>
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-2">
+        <div class="col-md-8 offset-md-2">
```
ramzimaalej: Bootstrap grid has 12 columns, what you did adds up to 16.

kaidohallik: How do you calculate 16?

sabrinapayet: In fact, is this title usefull ? Because it appears two times.

kaidohallik: Good point. I tested without this h1 title and looked good to me.

sabrinapayet: And like this it's more consistent with the other modals in entities.

kaidohallik: I'll change the title of this PR and I'll do new commit here.

deepu105:
Guys wait for me to take a look

generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
```diff
@@ -23,10 +23,11 @@ <h4 class="modal-title" jhiTranslate="login.title">Sign in</h4>
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-2">
-            <h1 jhiTranslate="login.title">Sign in</h1>
-        </div>
+        <%_ if (enableSocialSignIn) { _%>
+        <div class="col-12">
+        <%_ } else { _%>
         <div class="col-md-8 offset-md-2">
```
deepu105: Th idea of this seems like its was for centering the input boxes.
I think it might be better to use `<div class="col-xs-12">` with a custom class which does `text-align: center` or use the bootstrap option https://v4-alpha.getbootstrap.com/utilities/spacing/#horizontal-centering and get rid of the offset

generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
```diff
@@ -23,10 +23,11 @@ <h4 class="modal-title" jhiTranslate="login.title">Sign in</h4>
 </div>
 <div class="modal-body">
     <div class="row">
-        <div class="col-md-4 offset-md-2">
-            <h1 jhiTranslate="login.title">Sign in</h1>
-        </div>
+        <%_ if (enableSocialSignIn) { _%>
+        <div class="col-12">
```
deepu105: this should be col-xs-6 as the social buttons will take another 6 column

kaidohallik: No, this is error message div. If social sign in is enabled then currently error message div is narrower than main area, which is not beatiful. This fixes that.

generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
kaidohallik: If this h1 title remains then yes, this is correct solution.  
Question is: is this good that there is twice title "Sign in" or not?

generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
deepu105: oh ok i missed the code below

generators/client/templates/angular/src/main/webapp/app/shared/login/_login.component.html
deepu105: im ok with removing the extra title, but again centering without offset is preferrable

deepu105:
@kaidohallik can you fix failing tests?

kaidohallik:
Locally tests are passed.  
Error is:
```
Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
```
This is random Travis error. Restarting Travis build can help.

PR #5657, opened 2017-04-25
update to eureka client register and service url env variables

srinivasa-vasu:
update to eureka client register and service url env variables to reflect the right env variables

change serviceurl to service_url (EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE)
change EUREKA_REGISTER_WITH_EUREKA to EUREKA_CLIENT_REGISTER_WITH_EUREKA

### DIFF
--- a/generators/kubernetes/templates/registry/_jhipster-registry.yml
+++ b/generators/kubernetes/templates/registry/_jhipster-registry.yml
@@ -93,7 +93,7 @@ spec:
           value: my-secret-token-to-change-in-production
         - name: EUREKA_CLIENT_FETCH_REGISTRY
           value: 'true'
-        - name: EUREKA_REGISTER_WITH_EUREKA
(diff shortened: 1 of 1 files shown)


PierreBesson:
@xetys, this might be the fix for the registry issue you noticed.

deepu105:
@srinivasa-vasu thanks. Interested in adding openshift support? we had it a while ago but the removed it when RH changed the API and we didn't have anyone to maintain it. Btw the heroku and Kubernetes support are official so if you do the openshift part it can be considered official as well I guess :smiley_cat:

deepu105:
Here is the old version which had the support https://github.com/jhipster/generator-jhipster/tree/v1.x_maintenance/openshift

deepu105:
@jdubois @PierreBesson for some context about above message, Srini is a good friend of mine and works for Red Hat

PR #5662, opened 2017-04-25
Add ResponseWrapper which can hold the parsed and converted response,…

gzsombor:
Add ResponseWrapper which can hold the parsed and converted response, and
 refactor the conversion methods, so one implementation is called in all the code
 paths: after get/create/update/search/query, for converting the server response to javascript objects. And the dubious res.json().data = jsonResponse is won't needed

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_index.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_index.ts
@@ -48,6 +48,7 @@ export * from './user/account.model';
 export * from './user/user.model';
 export * from './user/user.service';
 <%_ } _%>
+export * from './model/response-wrapper.model';
 <%_ if (enableSocialSignIn) { _%>
 export * from './social/social.service';
 export * from './social/social.component';
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -366,6 +366,7 @@ const files = {
                 'shared/_shared-common.module.ts',
                 'shared/constants/_pagination.constants.ts',
                 // models
+                'shared/model/_response-wrapper.model.ts',
                 'shared/user/_account.model.ts',
                 // login
                 'shared/login/_login.component.ts',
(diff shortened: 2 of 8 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/_index.ts
```diff
@@ -49,6 +49,7 @@ export * from './user/account.model';
 export * from './user/user.model';
 export * from './user/user.service';
 <%_ } _%>
+export * from './services/response-wrapper.model';
```
deepu105: I think it can be in a model folder rather than services

generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
```diff
@@ -44,6 +44,7 @@ import {
     <%=jhiPrefixCapitalized%>SocialComponent,
     SocialService,
 <%_ } _%>
+    ResponseWrapper,
```
deepu105: I dont think this is needed here

generators/client/templates/angular/src/main/webapp/app/shared/services/_response-wrapper.model.ts
```diff
@@ -0,0 +1,9 @@
+import {  Response, Headers } from '@angular/http';
```
deepu105: Response is not used

generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
deepu105: can you remove this import its not used @gzsombor

gzsombor: Oops, I forgot to do it!

gzsombor:
Done, all green! :smile:

PR #5672, opened 2017-04-27
adds $JAVA_OPTS to Dockerfile

xetys:
this allows the user to specify options like memory limit by
providing environment variables.

## Motivation

after running a workload of ~10 services (with replicas on top) in kubernetes, I run into heavy memory issues, were entire nodes were crashing as running Java in docker with default settings allocates nearly all the memory. This PR **does not change that**!! But now the user can provide 

```
ENV JAVA_OPTS = " -Xmx300m -Xms300m" 
```

in any way, like from CMD, in docker-compose, rancher-compose or kubernetes deployment, to **really** limit the JVM memory usage.

I can add docs to this for kubernetes

I decided not do set any options, in favor of not breaking change existing setups. I personally run my containers with 300m limits and have noticed no performance issues in any way.

### DIFF
--- a/generators/server/templates/src/main/docker/_Dockerfile
+++ b/generators/server/templates/src/main/docker/_Dockerfile
@@ -20,6 +20,8 @@ FROM <%= DOCKER_JAVA_JRE %>
 
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JHIPSTER_SLEEP=0
+ENV JAVA_OPTS=""
+
 
 # add directly the war
 ADD *.war /app.war
@@ -27,4 +29,5 @@ ADD *.war /app.war
 EXPOSE <%= serverPort %><% if (hibernateCache == 'hazelcast') { %> 5701/udp<% } %>
 CMD echo "The application will start in ${JHIPSTER_SLEEP}s..." && \
     sleep ${JHIPSTER_SLEEP} && \
-    java -Djava.security.egd=file:/dev/./urandom -jar /app.war
+    java $JAVA_OPTS -Djava.security.egd=file:/dev/./urandom -jar /app.war
+


PierreBesson:
Well, we already document this : https://jhipster.github.io/docker-compose/#8
Is tour change really necessary to make it work ?

xetys:
I was following [this guide](https://developers.redhat.com/blog/2017/03/14/java-inside-docker/) and there JAVA_OPTIONS are passed directly in CMD, what was not the case with our container. I'm not totally sure if it would work without that change...but this docker file is already tested this morning, and I saved 66% of my memory usage...

PierreBesson:
After looking it up online it seems that on the contrary to what I expected, you do need to explicitly add the java_opts to the java command. I recall that it's also what is setup at my work now I understand why. So I'm merging this.

generators/server/templates/src/main/docker/_Dockerfile
```diff
@@ -20,11 +20,14 @@ FROM <%= DOCKER_JAVA_JRE %>
 
 ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS \
     JHIPSTER_SLEEP=0
+ENV JAVA_OPTS=""
```
pascalgrimaud: it should be in the same line ENV above, so it won't add additional Docker layer

generators/server/templates/src/main/docker/_Dockerfile
```diff
@@ -20,11 +20,14 @@ FROM <%= DOCKER_JAVA_JRE %>
...
+
 
 # add directly the war
 ADD *.war /app.war
 
 EXPOSE <%= serverPort %><% if (hibernateCache == 'hazelcast') { %> 5701/udp<% } %>
 CMD echo "The application will start in ${JHIPSTER_SLEEP}s..." && \
     sleep ${JHIPSTER_SLEEP} && \
-    java -Djava.security.egd=file:/dev/./urandom -jar /app.war
+    java $JAVA_OPTS -Djava.security.egd=file:/dev/./urandom -jar /app.war
```
pascalgrimaud: I prefer to have ${JAVA_OPTS} so no problem when using double quote in var

jdubois:
Sorry @pascalgrimaud I already merged this, can you do a PR with your 2 changes?

Issue #5664, opened 2017-04-26
Zuul proxies do not load balance routes correctly


jdubois:
This was working perfectly fine on older releases of JHipster, but it looks like Spring Cloud modified the way they do load balancing (and ping) for Zuul routes.
Please note this is only for the Zuul routes that are automatically created when a service is registered in Eureka, if routes are configured manually it's an even harder issue (but that's another issue).
From what I experience:

- The Zuul route always uses the first instance, and not the others
- When the instance goes down, after the Gateway gets synced with Eureka, the route "quickly" uses another instance

So you have no "real" load balancing, and downtime when an instance gets killed.

stevehouel:
@jdubois If you want we can work on it together. You can also use my code which test the loadbalancing (the one you show on Devoxx). You can find it on https://github.com/stevehouel/jhipster-story

jdubois:
@stevehouel do you have time to work on this today?

My current experiment is by doing:


    @Configuration
    public static class RibbonConfiguration {

        @Bean
        public IRule ribbonRule(IClientConfig config) {
            WeightedResponseTimeRule rule = new WeightedResponseTimeRule();
            rule.initWithNiwsConfig(config);
            return rule;
        }

        @Bean
        public IPing ribbonPing() {
            return new NIWSDiscoveryPing();
        }
    }

But it doesn't really work at the moment :-(

stevehouel:
@jdubois I can test it today if you want, I will update my jhipster-story with this configuration and deploy it (maybe on my rancher cluster). 

What I understand is that you want to use the DiscoveryPing and not the default Ribbon Eureka impl. I will have a look on algo for the WeightResponseTimeRule but it looks better than default RoundRobinRule

stevehouel:
Another way for LB rule is BestAvailableRule or AvailabilityFilteringRule

jdubois:
@stevehouel from my understanding, the current implementations are just `ZoneAvoidanceRule` and `NoOpPing` see http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html - and that would explain why it doesn't work anymore

stevehouel:
Yes I agress, NoOpPing disable check of each server using ping request (which was the one used before the update). Now it's only based on eureka server list and status. ZoneAvoidanceRule is for big cloud infrastructure with many Availability Zone. But for us this will not make any distinction between our instances.

jdubois:
[...]

PR #5682, opened 2017-04-28
Update Ribbon configuration for Zuul proxy

stevehouel:
Use another rule based on Response weight for Ribbon configuration which avoid to be Zone dependent.

Related to issue #5664

- [ X] Tests are added where necessary
- [ X] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application.yml
+++ b/generators/server/templates/src/main/resources/config/_application.yml
@@ -19,7 +19,7 @@
 # ===================================================================
 # Spring Boot configuration.
 #
-# This configuration will be overridden by the Spring profile you use,
+# This configuration will be overriden by the Spring profile you use,
 # for example application-dev.yml if you use the "dev" profile.
 #
 # More information on profiles: https://jhipster.github.io/profiles/
@@ -50,21 +50,18 @@ eureka:
         status-page-url-path: ${management.context-path}/info
         health-check-url-path: ${management.context-path}/health
         metadata-map:
-            zone: primary # This is needed for the load balancer
             profile: ${spring.profiles.active}
             version: ${info.project.version}
 ribbon:
     eureka:
         enabled: true
+    NFLoadBalancerRuleClassName: com.netflix.loadbalancer.WeightedResponseTimeRule
 <%_ } _%>
 <%_ if (applicationType === 'gateway') { _%>
-# See http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html
-zuul: # those values must be configured depending on the application specific needs
+zuul:
     host:
-        max-total-connections: 1000
-        max-per-route-connections: 100
-    semaphore:
-        max-semaphores: 500
+        connect-timeout-millis: 5000
+        socket-timeout-millis: 10000
     <%_ if (serviceDiscoveryType == 'consul') { _%>
     ignoredServices: 'consul,<%= baseName %>' # Prevent Consul and the "<%= baseName %>" app itself from having a Zuul route automatically created for them
     <%_ } _%>
--- a/generators/server/templates/src/main/resources/config/_application.yml
+++ b/generators/server/templates/src/main/resources/config/_application.yml
@@ -19,7 +19,7 @@
 # ===================================================================
 # Spring Boot configuration.
 #
-# This configuration will be overriden by the Spring profile you use,
+# This configuration will be overridden by the Spring profile you use,
 # for example application-dev.yml if you use the "dev" profile.
 #
 # More information on profiles: https://jhipster.github.io/profiles/
@@ -58,10 +58,13 @@ ribbon:
     NFLoadBalancerRuleClassName: com.netflix.loadbalancer.WeightedResponseTimeRule
 <%_ } _%>
 <%_ if (applicationType === 'gateway') { _%>
-zuul:
+# See http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html
+zuul: # those values must be configured depending on the application specific needs
     host:
-        connect-timeout-millis: 5000
-        socket-timeout-millis: 10000
+        max-total-connections: 1000
+        max-per-route-connections: 100
+    semaphore:
+        max-semaphores: 500
     <%_ if (serviceDiscoveryType == 'consul') { _%>
     ignoredServices: 'consul,<%= baseName %>' # Prevent Consul and the "<%= baseName %>" app itself from having a Zuul route automatically created for them
     <%_ } _%>


generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -19,7 +19,7 @@
 # ===================================================================
 # Spring Boot configuration.
 #
-# This configuration will be overridden by the Spring profile you use,
+# This configuration will be overriden by the Spring profile you use,
```
jdubois: nope, it's `overridden`

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -50,21 +50,18 @@ eureka:
...
     eureka:
         enabled: true
+    NFLoadBalancerRuleClassName: com.netflix.loadbalancer.WeightedResponseTimeRule
 <%_ } _%>
 <%_ if (applicationType === 'gateway') { _%>
-# See http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html
-zuul: # those values must be configured depending on the application specific needs
+zuul:
     host:
-        max-total-connections: 1000
```
jdubois: don't change those 3 values, I modified those on purpose, those have changed (and we tested them with higher values than the default)

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -50,21 +50,18 @@ eureka:
         status-page-url-path: ${management.context-path}/info
         health-check-url-path: ${management.context-path}/health
         metadata-map:
-            zone: primary # This is needed for the load balancer
             profile: ${spring.profiles.active}
             version: ${info.project.version}
 ribbon:
     eureka:
         enabled: true
+    NFLoadBalancerRuleClassName: com.netflix.loadbalancer.WeightedResponseTimeRule
```
jdubois: oh that was the trick I was looking for

generators/server/templates/src/main/resources/config/_application.yml
stevehouel: my bad I guess that I did not get the last version on master. Because I never change those values

generators/server/templates/src/main/resources/config/_application.yml
stevehouel: Yes exactly. Simple and easy :)

jdubois:
I just did some tests in Docker, and the load balancer is failing like before :-(

- I start 3 nodes
- I do some "curl" through Zuul to get data from one node (probably the first one)
- I kill the first node
- All my "curl" commands fail for something like 30 seconds, until the node is removed by Eureka

-> no load balancing is happening at all :-(

stevehouel:
Thats weird @jdubois because on my side i did the same thing but from a
call on gui and lb was working. If you want i can commit my code for my
jhipster story repo and test on your side
Le ven. 28 avr. 2017 à 21:53, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
I know @stevehouel - I'm very surprised and disappointed!! Then I didn't check if the configuration was right - are you sure you tested exactly with what's in your PR's `application.yml`? I need to debug that to see what's happening (but probably not during this week end)

stevehouel:
I will check it again if you want. The only thing i did not have is your
zull special configuration ( errors on my commit) i will check it tomorrow
and verify that everything work fine
Le ven. 28 avr. 2017 à 21:59, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
Thanks @stevehouel - then hopefully this is due to my the Zuul specific configuration, I guess it's not picking up your load balancer configuration (that should be easy to see with a debugger)

stevehouel:
Sure @jdubois. I will check it tomorrow
Le ven. 28 avr. 2017 à 22:04, Julien Dubois <notifications@github.com> a
écrit :

stevehouel:
So @jdubois I have done all my checks and this is my conclusion: It s working.

So I check both with my script which is doing a call every seconds and by using a curl manually like you.  the rule we choose is optimized to have a better response time (it will redirect request to the lowest charge server). But in your case with a manual curl your are not doing enough request to use that function so all request are redirect to the same server. I also test the FilterAvailibilityRule and we have a better result on it but I guess that we need to choose. For me the Weight system is optimized for production env (in my case shutting down 3 servers in the same time should never happend in production otherwise they have other several issues). And the availibility rule is the best for our test. So I guess we should let the Weight rule for production and if you want to test availibility you should switch to it. I will do the documentation once PR is merged. I also commit my code for my jhipster-story case if you want to see LB working go to graph page and watch Network queries.

jdubois:
@stevehouel I just ran my debugger and I'm still using the ZoneAwareLoadBalancer, so I have something wrong in my configuration

stevehouel:
@jdubois do you still have your ribbon zone parameter ??
Le dim. 30 avr. 2017 à 10:52, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
No, I *think* I have your configuration... I'm checking everything

stevehouel:
I will try to generate again from scratch my components and i will check.
Le dim. 30 avr. 2017 à 10:55, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
@stevehouel at least I found my initial error - I didn't read the documentation good enough!!!

So with the default LB (the ZoneAwareLoadBalancer), you need to set the same zone both in the gateway (= the client) and the microservices.

So with having everywhere:

    eureka:
        instance:
            metadata-map:
                zone: primary

-> I have result that works in fact quite well. The LB works correctly, and quickly removes the bad instance.

stevehouel:
@jdubois is that working with my configuration or your ?
Le dim. 30 avr. 2017 à 11:55, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
With an updated version of mine: I needed to also set up the gateway, not just the microservice

stevehouel:
So we need to choose between also those rules and which one we would like
to provide by default.
Le dim. 30 avr. 2017 à 12:40, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
At the moment, I only have one configuration working, so it's quite obvious... I'm going to commit that -> we are going to rework the whole microservices documentation next week, so we'll be able to document other options

jdubois:
I'm closing this as:

- The "zone" configuration indeed works!
- The load balancer configuration is not getting used - I don't know why

@stevehouel I am going to rework our microservices documentation, and we'll have a specific page for the gateway -> we'll follow [Policy 1](https://jhipster.github.io/policies/) and keep the default config (with the "zone" stuff), but we have seen this needs to be fine tuned for many people, so if you can have a good configuration for the load balancer, that would be great to document it there.

stevehouel:
Sure @jdubois I will make a paragraph  of all possible configuration for
the LB and with pros and cons. I agree with you about letting the default
configuration.

Le mar. 2 mai 2017 à 09:02, Julien Dubois <notifications@github.com> a
écrit :

PR #5684, opened 2017-04-28
Add JSR310's Instant data type

cbornet:

When you want a datetime type, 99% of the time the Java type you want to use is Instant and not ZonedDateTime.
Instant is independent from the timezone rules and should be used each time you need a timestamp (eg : logs, creation datetime, event datetime, etc ...)
ZonedDateTime depends on the timezone rules and these rules change often (eg. a country changes the days for DST, drops or adopts DST). So it should be used for calendar appointment (eg : your appointment is at 9AM in a week. If the TZ rules change your appointment is still at 9AM but the POSIX time has changed). 

Notes:
* JDBC/Hibernate don't support storing the data timezone even for DBs that have `timestamp with timezone` types
* Our mongo mapping doesn't support storing the TZ either
* With this PR, the TZ will be stored in Cassandra as it uses the datastax codec 
* We currently don't support TZ at the REST layer either (and there's no standard for the zone ids)
* If you want to be sure that your users don't miss their flight, you need to ensure that the tz rules are up-to-date with a tool such as http://www.oracle.com/technetwork/java/javase/tzupdater-readme-136440.html


jdubois:
@cbornet can you also add this on the documentation page at https://jhipster.github.io/creating-an-entity/ ?

cbornet:
@jdubois  done. https://github.com/jhipster/jhipster.github.io/pull/432

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
```diff
@@ -18,7 +18,7 @@
 -%>
 <%_
     let hasDate = false;
-    if (fieldsContainZonedDateTime || fieldsContainLocalDate) {
+    if (fieldsContainInstant || fieldsContainZonedDateTime) {
```
pascalgrimaud: I think we should keep `fieldsContainLocalDate` in the condition too
I will made a PR for that

cbornet: Indeed... 😅

PR #5689, opened 2017-04-30
Merge all the translations into a language specific json

gzsombor:
Instead of lot's of json files, which gets reloaded on every page change, use only one, called 'all'

### DIFF
--- a/generators/client/templates/angular/webpack/_webpack.common.js
+++ b/generators/client/templates/angular/webpack/_webpack.common.js
@@ -23,6 +23,7 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const StringReplacePlugin = require('string-replace-webpack-plugin');
 const AddAssetHtmlPlugin = require('add-asset-html-webpack-plugin');
 const WebpackNotifierPlugin = require('webpack-notifier');
+const MergeJsonWebpackPlugin = require("merge-jsons-webpack-plugin")
 const path = require('path');
 
 module.exports = function (options) {
@@ -123,6 +124,15 @@ module.exports = function (options) {
             new webpack.ProvidePlugin({
                 $: "jquery",
                 jQuery: "jquery"
+            }),
+            new MergeJsonWebpackPlugin({
+                output:{
+                    groupBy:[
+<%_ for (language in languages) { _%>
+                        { pattern:"./src/main/webapp/i18n/<%= languages[language] %>/*.json", fileName:"./<%= BUILD_DIR %>www/i18n/<%= languages[language] %>/all.json" },
+<%_ } _%>
+                    ]
+                }
             }),
             new HtmlWebpackPlugin({
                 template: './src/main/webapp/index.html',
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -115,6 +115,7 @@
     "webpack-merge": "4.1.0",
     "webpack-notifier": "1.5.0",
     "webpack-visualizer-plugin": "0.1.11",
+    "merge-jsons-webpack-plugin": "1.0.5",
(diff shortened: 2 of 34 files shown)


gmarziou:
I see this uses merge-jsons-webpack-plugin, @stevehouel do you see some way to make it work with your PR #5661?

gzsombor:
One test is failed, with "Parameter 2 of method consulRegistration in org.springframework.cloud.consul.serviceregistry.ConsulAutoServiceRegistrationAutoConfiguration required a bean of type 'javax.servlet.ServletContext' that could not be found." - I'm pretty sure, that it's not this commits fault. 
 Webpack is not hipster anymore ? These javascript tools do not have a long life time :)

stevehouel:
Yes we are migrating all webpack configuration to be fully compatible with
angular cli stuffs. And no I am not sure that we can do that merge of all
json files into one with angularci. It s using the default angular i18n
functions so files are not the same. I will check it anyway to be sure :)

generators/client/templates/angular/webpack/_webpack.common.js
```diff
@@ -124,6 +125,15 @@ module.exports = function (options) {
                 $: "jquery",
                 jQuery: "jquery"
             }),
+            new MergeJsonWebpackPlugin({
+                output:{
+                    groupBy:[
+<%_ for (language in languages) { _%>
+                        { pattern:"./src/main/webapp/i18n/<%= languages[language] %>/*.json", fileName:"./src/main/webapp/i18n/<%= languages[language] %>/all.json" },
```
deepu105: shouldn't the output be created in the target/build folder instead of the src?

deepu105:
@stevehouel @gzsombor @gmarziou the angular-cli only approach is an experiment we havent yet decided if we will do it. we still need to validate and test it. There are pros and cons of that approach. Ill talk about that in the PR for it, so we need to worry about that for now

stevehouel:
Sure sure no worries I agree :)

generators/client/templates/angular/webpack/_webpack.common.js
gzsombor: Good question. My original thought was the same, however in that case, I couldn't figure out, how to run the server from my IDE, which is serving content from src/... I thought about adding a dynamic service, which could merge all the jsons automatically during runtime in dev mode. But it's seemed to be a bigger task than I would like to handle :smile:  
I've ended up commiting the merged jsons, which is sub-optimal solution.

deepu105: Content is always served from `build/target` folder for angular 2+ as we compile TS to js into the `target|build/www` folder so you shouldnt have any difference at all, but I guess the problem here would be that the json are not bundled into the app at compile time there is always a one time http request getting the `all.json`. So we need to bundle the json with the main app bundles. But if you think it can be done later then i'm ok with your current solution

gzsombor: Indeed, you are right ! Thanks :)

gzsombor:
I guess, it's just the random test failure, nothing serious ...

deepu105:
@stevehouel FYI

jdubois:
This is currently discussed on the dev mailing list at https://groups.google.com/forum/?hl=en#!topic/jhipster-dev/Jl711fWRr9Y

-> it looks like this causing an error, and as I would like to do a release soon, this is blocking us - if we don't find a solution soon, I'll revert this commit for the release

gzsombor:
I think, this will fix that problem: https://github.com/jhipster/ng-jhipster/pull/34

Or, we could remove the translate-partial-loader altogether.

jdubois:
@gzsombor thank you so much!!!

PR #5690, opened 2017-04-30
Ask the server about the list of authorities

gzsombor:
The user management dialog only shows the 2 builtin roles, instead of asking the server for a list of roles. After this commit, this will work correctly.


pascalgrimaud:
Is it possible to add test for the new method `getAuthorities` in `UserResource.java` ?

It would be nice to keep 100% coverage for UserResource: https://sonarqube.com/component_measures/metric/coverage/tree?id=io.github.jhipster.sample%3Ajhipster-sample-application

gzsombor:
Sure, I will create one

PR #5700, opened 2017-05-02
Store JWT in authentication object

cbornet:

This is especially useful for aggregate micro-services where you want to relay the token to downstream services.

### DIFF
--- a/generators/server/templates/src/main/java/package/security/_SecurityUtils.java
+++ b/generators/server/templates/src/main/java/package/security/_SecurityUtils.java
@@ -50,6 +50,22 @@ public final class SecurityUtils {
         }
         return userName;
     }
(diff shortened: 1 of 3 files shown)


jdubois:
@cbornet what is the use case for having access to the token? For me that's adding overhead to each thread (as it's stored in thread local), and I don't understand the need

cbornet:
@jdubois it is useful when you want to relay the token from a micro-service to another. Typically when you aggregate micro-services. Suppose user calls MS A which in turns calls MS B, you want to have the token used for the call to A to be used by A to call B.
See http://cloud.spring.io/spring-cloud-security/spring-cloud-security.html#_token_relay for the way to do it with OAuth . But we also need something equivalent for JWT.

jdubois:
Oh yes, I totally agree!

PR #5707, opened 2017-05-04
update to prodDatabaseType check while generating final k8s files

srinivasa-vasu:


generators/kubernetes/files.js
```diff
@@ -11,8 +11,8 @@ function writeFiles() {
                 this.app = this.appConfigs[i];
                 this.template('_deployment.yml', `${appName}/${appName}-deployment.yml`);
                 this.template('_service.yml', `${appName}/${appName}-service.yml`);
-
-                if (this.app.prodDatabaseType) {
+                // If we choose microservice with no DB, it is trying to move _no.yml as prodDatabaseType is getting tagged as 'string' type
+                if (this.app.prodDatabaseType !== 'no') {
```
deepu105: I think it might be better to & this with the original as we dont want to move the file when the property is undefined as well. @PierreBesson am I right or is there no way for this to be undefined?

srinivasa-vasu: I thought so, but it gets defined every time, either with the chosen db or 'no'.

deepu105: Ok then

PR #5726, opened 2017-05-07
JHipster CLI

deepu105:

@jhipster/developers this is the first version of the JHipster CLI

Please test it and provide feedback to me.

To test this locally
1.  Pull this branch
```
git checkout -b deepu105-jh-cli
git pull https://github.com/deepu105/generator-jhipster.git jh-cli
```
2. Run `npm install` or yarn equivalent
3. Run `npm install -g` or yarn equivalent
4. Run `npm link` or yarn equivalent
5. The CLI should be installed in path now. Create a new folder somewhere and run below commands to test (please note to disable the jhipster zsh/fisherman plugin if you have it setup)

- The initial version basically maps all our existing `yo jhipster:*` commands to `jh *`. All flags supported by each subgen can be passed in the same way for eg: `jh entity testEntity --skip-client`. the `--help` flag lists all possible arguments and flags
- A `-d` flag can be passed to enable debug logs
- `jh --help` prints all available commands and options
- `jh` runs the default command equivalent to `yo jhipster` supports all the flags present
- `jh --version` prints version


test/test-cli.js
```diff
@@ -0,0 +1,45 @@
...
+
+        exec(`${cmd} junkcmd`, (error, stdout, stderr) => {
+            assert(error);
+            assert.equal(error.code, 1);
+            assert.equal(stderr.includes('is not a known command'), true);
+            done();
+        });
+    });
+
+    // TODO figure out a way to get this test working
```
deepu105: Let me know if anyone knows a way to get this working

deepu105:
@jhipster/developers guys can anyone test this please?

Our travis fails with below error, probably comes from the new dependencies added
```
error hawk@0.10.2: The engine "node" is incompatible with this module. Expected version "0.8.x".
error Found incompatible module
```

deepu105:
Apparently the failure is coz of yarn. Npm doesnt fail on this

pascalgrimaud:
huge work @deepu105 as always
I should be able to test it tomorrow, I will let u know

ruddell:
I went through each command from `jh --help` and tested.  Note if you use the Zsh Jhipster plugin you need to disable it before this will work correctly
***
When I run `npm link`, the mode on the cli.js file changes:
```
diff --git a/cli/jh.js b/cli/jh.js
old mode 100644
new mode 100755
```
`jh aws` print out the "executing" line twice and loops through the prompts twice:
```
Executing jhipster:aws 
Executing jhipster:aws 
? Application name: (test) 
```
`jh service <ServiceName>` prints out the "executing" line twice, service name should be required (it seems to be required with the --debug flag)
`jh export-jdl` is broken, but works with the debug flag
`jh languages` is broken, but works with the debug flag
`jh import-jdl <JDL Filename>` is broken, but works with the debug flag

deepu105:
Thanks guys, John those are weird issues, the executing twice and stuff. Im
on ubuntu what os are you on? Btw guys dont hesitate to fix any issue you
find :) i can push the branch to jhipster repo if it makes it easy for you
guys

ruddell:
I'm on Mac.  The project I tested with was all default options. I experienced the same thing with both Node v6.10.0 and v7.7.1

deepu105:
ok ill check it out.

atomfrede:
I can test it with fish shell. I guess I should remove my jhipster fish plugin first?

deepu105:
yes as the 'jh` command will conflict. You can remap the fish plugin to
something like jhi as well


Thanks & Regards,
Deepu

pascalgrimaud:
I would prefer to not have `yarn --ignore-engines` but without this ignore, we can't install new dependencies.
The real problem comes from:
```
$ npm ls hawk
generator-jhipster@4.3.0 /home/pgrimaud/projects/jhipster/generator-jhipster
├─┬ autocmdr@0.0.8
│ └─┬ winston@0.7.3
│   └─┬ request@2.16.6
│     └── hawk@0.10.2 
└─┬ insight@0.8.4
  └─┬ request@2.81.0
    └── hawk@3.1.3 
```

pascalgrimaud:
I had the same issues, listed by @ruddell
I'm on Ubuntu 16.10

pascalgrimaud:
The double print comes from `command.js` file, when the name and alias are the same.
```
    aws: {
        alias: 'aws',
        desc: 'Initializes a AWS app and generates a WAR file that is ready to push to AWS.'
    },
```
If you change alias: 'aws' to alias: 'aws2', no double print.

Quick tests this morning, I can't go further. Time to real work :)

deepu105:
@pascalgrimaud @ruddell I fixed the issues mentioned by you guys can you check again?

ruddell:
Everything works for me 👍

pascalgrimaud:
Same, everything works :+1: 

@deepu105 : one last point for me is the use of: `yarn --ignore-engines` instead of `yarn`
Currently, I don't see another way to avoid that.

deepu105:
Ok ill look into them

deepu105:
@pascalgrimaud yes i dont see a way to avoid it as well. I can try to raise an issue to autocmdr which uses the older dependencies.

Anyway lets merge this so that more people will test it in the master branch.

jdubois:
Same problem here with `--ignore-engines`, and then I can't have the `jhipster` command in my path... I'll have a look, but we need to solve those before releasing this officially (anyway, it could come as a "alpha" feature in the next release, to have some feedback, in case we break too many things)

deepu105:
@jdubois ok I will remove the autocmdr dependency soon. That should solve the `--ignore-engines` issue but then I dont understand why you couldnt have it in the path. doing `npm install -g` normally adds it to the node bin and hence make it available on path. can you try with NPM and see if it works?

I think having it as alpha will affect people adopting it so I would rather fix it properly and then release, anyway I'm sure we can fix it before next release

jdubois:
I got it working with `npm`, but not with `yarn` (even with the `--ignore-engines` - it's just not doing anything)

deepu105:
Ok i'll look into it


Thanks & Regards,
Deepu

PR #5740, opened 2017-05-11
Change audit event date type from LocalDateTime to Instant

### LINKED ISSUE #5738: Use Instant in PersistentAuditEvent instead of LocalDateTime
I think Instant would be more relevant than LocalDateTime for auditEventDate field of PersistentAuditEvent.
If agree, I can PR.

cbornet:

Also cover CustomAuditEventRepository with an integration test

Fix #5738

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -565,6 +565,7 @@ function writeFiles() {
             }
 
             if (this.databaseType === 'sql' || this.databaseType === 'mongodb') {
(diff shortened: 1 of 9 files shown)


cbornet:
Unfortunately, we were saving the events with the system timezone and not UTC (https://github.com/jhipster/generator-jhipster/pull/5740/commits/db2f30c231b31542b190a62f9f2f46b283a633f9#diff-722e9c6188977ef4fa34404587b4310aL94). So users who have a TZ different than UTC on their server (bad practice btw) and existing events in their DB will probably get an offset in the event dates.
Now I think it's healthier to have timestamp stored in UTC (which will be forced by this PR) since you never know what will be the tz of the db reader (eg. if you change your server tz, all your dates are offset)

PR #5755, opened 2017-05-12
Center elasticsearch field [ci skip]

pascalgrimaud:
Before:
elasticsearch-right

After:
elasticsearch-center

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.component.html
@@ -30,7 +30,7 @@
     <jhi-alert></jhi-alert>
     <div class="row">
         <%_ if (searchEngine == 'elasticsearch') { _%>
-        <div class="col-sm-12 no-padding-right">
+        <div class="col-sm-12">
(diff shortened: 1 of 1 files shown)


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management.component.html
```diff
@@ -30,7 +30,7 @@
     <jhi-alert></jhi-alert>
     <div class="row">
         <%_ if (searchEngine == 'elasticsearch') { _%>
-        <div class="col-sm-12 no-padding-right">
+        <div class="col-sm-12 no-padding-center">
```
deepu105: does bootstrap have such a class?

pascalgrimaud: oh let me test it on the default app and not on mine!

pascalgrimaud: indeed, it doesn't exist lol :stuck_out_tongue_closed_eyes:
I will simply remove no-padding-right for the same result

PR #5772, opened 2017-05-16
cache is fail in brower, in prod profile

gemron:


deepu105:
Ok that indeed is a good plan. Just let me know if you need me to do
anything


Thanks & Regards,
Deepu

ruddell:
@jdubois That's true for AngularJS.  In production with Angular (2+), all fonts, images, and js files are located on the base path

jdubois:
@ruddell yes I know, this is the main reason why Angular 4 is still in beta, the prod build is not yet finished. I don't think we have hashes everywhere, so they shouldn't be cached, or people won't have their JS updated when a new version is released.

ruddell:
All JS bundles, images, and fonts are hashed during the prod build, with the exception of `vendor.dll.js`

gemron:
thanks! I hope jhipster  can fix it . 
another issure is:    jhipster 's client in prod is so big ,brower init vendor and main js is so slowly.I hope it  jhipster 's client  support lazy load.

PR #5775, opened 2017-05-16
remove zipkin from jhipster-console for monoliths

srinivasa-vasu:


generators/docker-compose/templates/_docker-compose.yml
```diff
@@ -61,10 +61,12 @@ services:
         extends:
             file: jhipster-console.yml
             service: jhipster-console
+    <%_ if (composeApplicationType === 'microservice') { _%>
```
PierreBesson: I'm not sure about this `composeApplicationType`. Where is it defined ? Shouldn't it be microservice or gateway ?
[EDIT] Never mind it is in prompt.js. No problem.

srinivasa-vasu: @PierreBesson - This is defined in docker-compose (/docker-compose/prompt.js) - prompt to identify whether micro or monoliths. Gateway is not required, anyway it comes under microservices and the 'composeApplicationType' holds either microservice or monolith.

PR #5789, opened 2017-05-18
Fix various sonar code smells

cbornet:

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
+++ b/generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
@@ -22,17 +22,17 @@ import java.util.List;
 
 /**
  * Contract for a generic dto to entity mapper.
- @param <DTO> - DTO type parameter.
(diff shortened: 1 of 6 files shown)


generators/entity/templates/server/src/main/java/package/service/mapper/_BaseEntityMapper.java
```diff
@@ -22,17 +22,17 @@
 
 /**
  * Contract for a generic dto to entity mapper.
- @param <DTO> - DTO type parameter.
- @param <ENTITY> - Entity type parameter.
+ @param <D> - DTO type parameter.
+ @param <E> - Entity type parameter.
```
gzsombor: In my opinion, this sonar rule is silly - when a type has more than one generic parameters, especially when there is no other restrictions on the type - it's simply confusing, and makes harder to understand what happens here.  Just read the original interface definition : it's crystal clear that, what each method needs and provides. 
 Otherwise, I'm suprised that Sonar didn't flagged that the public modifiers are not necessary on all the interface methods :wink:

Issue #5752, opened 2017-05-12
Access/Refresh Token Handling for UAA Authentication


Metavirulent:
* **Overview of the request**

I want to propose a pull request improving token handling for microservice gateways with `authenticationType==='uaa'` fixing critical security issues with the current implementation. The proposed changes are:
- store the refresh token along with the access token
- keep the tokens as http-only cookies rather than in the sessionStorage/localStorage (for security reasons, and to enable SSO, s.b.)
- implement a filter in the gateway that automatically sends a refresh_token grant to UAA when the access token is about to expire/expired, seamlessly setting new cookies.

This fixes a number of security and usability problems as described below.

* **Motivation for or Use Case** 

Currently, for `authenticationType==='uaa'`, only the access token is used, the refresh token is discarded. Furthermore, the access token is stored in the localStorage/sessionStorage and thus accessible from within Javascript.

This has the following security problems:
- storing access tokens in a way that they can be retrieved from within Javascript is considered to be insecure (see https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage#so-whats-the-difference, https://www.owasp.org/index.php/HTML5_Security_Cheat_Sheet#Local_Storage and https://www.whitehatsec.com/blog/web-storage-security/).
- due to the lack of refresh token handling, access tokens validity must be set to a much longer duration which further aggravates the problem. Normally, access tokens are supposed to be short lived so if they get compromised, the window of attack is fairly small.

For security-sensitive applications that require an external audit, these are show-stoppers.

In addition to these security issues, there are a few additional problems:
- once the access token expires, the user is kicked out of his "session", even if he was active mere seconds ago. This can be quite confusing and isn't very good from a usability point of view. He should only be kicked out if he is inactive for too long and his refresh token expires, not because the access token he obtained quite a while ago suddenly goes poof.
- tokens cannot be shared across gateways. Most microservices will probably feature multiple gateways and they shall all share the same tokens, governed by a central UAA. That way, you could have some kind of SSO across gateways. It's common to assign subdomains to individual apps (like Google does with maps.google.com, accounts.google.com, etc.) and once you are logged in to one of them, you want to reuse that token with the next one. This is currently not possible.

* **Browsers and Operating System** 

n/a

* **Related issues** 

n/a

* **Suggest a Fix** 

I'd prepare a pull request that
- modifies the way UAA gateways authenticate, setting and clearing cookies on the top level domain
[...]

PR #5812, opened 2017-05-22
Refresh token grant

Metavirulent:
I've implemented #5752, so the security model for UAA microservices is now a bit different.

Here's a list of changes:
- access/refresh tokens are now stored as http-only cookies on the top-level domain. If you want them to be set on another domain, then you can change that via application properties (`oauth2.web-client-configuration.cookie-domain`).
- gateways will automatically use refresh tokens to get a new token pair if the access token is expired or is about to expire. This should be transparent to the web application.
- access token validity for the web_app client can be configured on UAA via `uaa.web-client-configuration.access-token-validity-in-seconds`. Don't use too long durations.
- "remember me" means your refresh token is kept persistent so it will still be around when you close the browser. Duration of remember me can be configured on UAA via property `uaa.web-client-configuration.refresh-token-validity-in-seconds-for-remember-me`.
- if "remember me" is not checked, then there's an optional session expiration timer that will terminate the session after a configurable period of time. This is a common security requirement. The duration can be configured at the gateway via `oauth2.web-client-configuration.session-timeout-in-seconds`. A negative value will disable this. Please note that this is only checked when the access token expires, so having a long access token validity will delay the timeout.
- the "web_app" client used to lack a password since it was used on the browser and setting a password there is moot. The client ID/secret is no longer used by the angular client and thus it now has a proper password that can be configured. The UAA and the gateways must have the same password configured or they won't be able to talk to each other. The properties are uaa.web-client-configuration.secret (on UAA) and oauth2.web-client-configuration.secret (gateway). Please change in production!
- the keystore used to sign JWTs is now configurable via `uaa.key-store.name`. The password and alias can be set via `uaa.key-store.password` and `uaa.key-store.alias`, respectively. These values must be changed in production, you don't want to sign tokens with JHipster's default store in prod.
- the public key configured on UAA must be fetched by all microservices (and gateways) so they can verify JWT tokens. When the microservice could not reach UAA, then it failed to start up. This is now different, microservices will continue to startup and fetch the key later. There are two new properties `uaa.signature-verification.ttl` controls how long the key lives before it is fetched again, `uaa.signature-verification.public-key-refresh-rate-limit` limits requests to UAA to avoid spamming it. You will usually leave them at their default values. In any case, if verification fails, then the microservice will check if there's a new key. That way, keys can be replaced on the UAA and the services will catch up.
- signature verification and the OAuth2 grants are now accessed via interfaces. There's an UAA implementation if your OAuth2 provider is UAA but this can be replaced by a different implementation, e.g. for keycloak. There's a proof of concept that also implements these interfaces for keycloak. This is by far not a full solution, as other areas on the UI are also tied to UAA, but it abstracts the OAuth2 parts nicely.


generators/server/templates/src/main/java/package/config/oauth2/_OAuth2JwtAccessTokenConverter.java
```diff
@@ -0,0 +1,98 @@
+<%#
+ Copyright 2013-2017 the original author or authors.
```
jdubois: Can you update the copyright, have a look at the latest files, we've added "from the JHipster project" here. The issue is that we have people copy/pasting this, in some other projects, without respecting the license, and so in their Git history they are the "original author". And as you see we are taking care of everybody's work...

jdubois:
- Thanks for the HUGE work, I don't know where to start here!!
- I would really like to have a review from @xetys here, as he's the UAA boss :-)
- I don't know how this fits with Keycloak -> I really, really want to have a good Keycloak integration, can you update me on this?

Then for the really annoying part:

- Our focus for the next release is having Angular 4 out of beta. Your code is so huge, I don't think we should merge this now, or we would lose our focus on Angular 4.
- This means we have a couple of weeks before merging your code, which gives time for a review. But my fear is that we have merge conflicts, which is often the case with big changes. I hope that's OK with you.

xetys:
It looks like a huge work! At least, what I could observe from smartphone during my travel back.

Of course, I will do a review on that stuff.
About the 'out of beta' @jdubois: we might do two hits at once? I am not that involved in the angular stuff, do I can spend my time here. Just a suggestion

jdubois:
Sorry @xetys but I can't. I also want to review this code, and that takes time.

lukasw44:
Any progress with this task ?

jdubois:
Sorry, but our next release is still not out, and we have a huge work for Angular 4. This has been postponed for too long, at one point we need to focus and finish having Angular 4 out of beta.
Any help on Angular 4 is highly welcome, of course.

deepu105:
Since this PR is stuck for some time I propose this
@xetys can you review and give comments if any
@Metavirulent can you please rebase to current master

@xetys @jdubois lets tackle getting UAA out of beta separately, lets just review and merge this for now

xetys:
@deepu105 @Metavirulent @jdubois I add this on my schedule this week. I was waiting as @jdubois wanted to focus more on angular 4.

Metavirulent:
I rebased to master.
For the review, here are a few pointers.

## Refresh Tokens

The general flow for refreshing access tokens is:
- authentication is now done via `AuthResource` calling `OAuth2AuthenticationService`'s authenticate which will set Cookies.
- for each request, the `RefreshTokenFilter` (installed by `RefreshTokenFilterConfigurer`) checks whether the access token is expired and whether it has a valid refresh token
- if so, then it triggers the refresh process via `OAuth2AuthenticationService` refreshToken.
- this uses the OAuth2TokenEndpointClient interface to send a refresh token grant to the OAuth2 server of choice, in our case UAA (via UaaTokenEndpointClient). Here, other implementations can be substituted, e.g. one for Keycloak. This is not part of the PR.
- the result of the refresh grant is then used downstreams as new cookies and set upstream (to the browser) as new cookies.

## Signature Verification
- the `OAuth2JwtAccessTokenConverter` fetches the public key, or actually a `SignatureVerifier` from the OAuth2 implementation when decoding fails or the key does not exist/is expired.
- this is covered by the interface `OAuth2SignatureVerifierClient` and its UAA implementation `UaaSignatureVerifierClient`. Other OAuth2 implementations need to replace this by another implementation.
- the SignatureVerifier is then used as usual to verify JWT token signatures for validity.

There's a lot of interest in a Keycloak integration. To be clear, this PR is *not* about Keycloak integration but the whole refresh token handling can be easily made Keycloak-ready by replacing UaaSignatureVerifierClient and UaaTokenEndpointClient by Keycloak-specific implementations. This handles JWT verification and authentication/token refreshing. I've included Keycloak implementations of these two interfaces.

However, it is not a full keycloak handling as this is not the purpose of the PR (which is already fairly big as it stands). At least a UserInfo endpoint must be abstracted, so requests to the AccountResource end up at keycloak (these requests are directly sent to UAA right now).

xetys:
Hey,

I started my tests today using [my demo microservice generator](https://github.com/xetys/jhipster-demo-generator) and here are my first insights. First of all: huge work! Thanks a lot for this contribution. One of the first very cool improvements is that it's now possible to start UAA microservices without depending on the UAA to be registered and running. Here are the first bugs:

* creating an entity, not inside UAA is causing `Could not verify the provided CSRF token because your session was not found.`
* when remember me is not active, token expire doesn't trigger a reload to 'access denied' page, just silently getting 401 errors

I was testing directly in docker. You can generate a project using my generator to have the same testing conditions

xetys:
I was thinking about how to solve the CSRF issue. It is failing obviously as we are having at least 2 (or even more) sources of CSRF token repositories running. The user receives the XSRF cookie from the gateway, while an underlying service has its own repository enabled. As of default, Spring tries to save the token to the current session (according to https://docs.spring.io/spring-security/site/docs/current/reference/html/csrf.html), what also makes no sense, as this PR provides a "fake session". We exploit the HTTP only cookie behavior to let the browser send JWTs automatically.  But this fact makes it important, to have a CSRF protection enabled.

To make this possible, we need some token distribution strategy, which doesn't depend on a state, which is then hard to scale and will cause performance impacts, when requests go through several microservices. To be more precise, the gateway should provide some token in an XSRF cookie, which can be verified by any other service.

I suggest taking the most simple way to go and use the access token for that purpose. That way we enforce the client to add the access token for critical requests, but still allow normal requests pass, keeping the simplicity of using HTTP only cookies remain. If an attacker will prepare an evil form, he must know the access token anyway. A CSRF attack is well protected by using the access token as CSRF and is simple to verify on all microservices using a custom `RequestMatcher` in `CsrfFilter` without unnecessary overhead.

jdubois:
I was going to review this, but 250+ commits, 416 changed files??? I think there's something wrong in your PR

lukasw44:
@xetys I think this is a good idea using access token for CSRF Does anyone see security issue with this solution ?

gzsombor:
You should rebase and squash all the commits to one changeset, it's really hard to see what happened

xetys:
this is a lot better now, but I'm still confused with the list of your commits. There is some kind of 4 "periods" of the same commits (respecting the commit order) on different dates.

deepu105:
Yes the commits look weird, could you please squash them and reduce into bunch of more logical commits

deepu105:
@xetys Is this ok for you? if so we can squash and merge this as there are too many commits here.
@Metavirulent if you don't want to lose your genuine commits I suggest you clean up else we would have to squash resulting in single commit credit for you

Metavirulent:
I'm currently on vacation and have no laptop (thanks to the wood fires in southern France). Feel free to squash them as I won't be able to look into it soon...

deepu105:
@xetys is this ok to merge? if so I'll merge them manually after cleaning up the commits

xetys:
actually, there is still a problem with the CSRF, what prevents writing operations with the current state...

how are we going to proceed here? It's a wonderful work but not ready to be merged to master.

Metavirulent:
Not sure what you mean with the CSRF problem. Since tokens are now kept in cookies, we had to switch on CSRF protection.

xetys:
I did a test and it is obviously still present. If you create an architecture with at least UAA and another microservice, you won't be able to add new entities. This happens because you enabled CSRF and every microservice delivers its own CSRF token. The request takes the UAA CSRF token, which only leads to the effort if the endpoint is on the UAA. 

The only way I see how to work around this scalable is to use the access token as CSRF token. Otherwise, we have to find a way to generate statelessly but session specific CSRF token, which is identical to all services.

Another workaround is moving the CSRF responsibility to the gateway and disables CSRF on the microservices, like before. If the user wants to make it super secure, s/he can apply a security role permitting calls coming from a certain IP or host, to only let the gateway access the microservice endpoints. However, I think this is unnecessary, as usually, users cannot access microservices directly. Enabling classic CSRF on gateway includes making it stateful, what impacts its scalability.

Merging this would break any entity communication from the client to the endpoints currently

lukasw44:
Is there any reason why access token cannot be used as a csrf token ?

Metavirulent:
@xetys now I see what you mean. That's actually unintended, there's no point in enabling CSRF protection on anything but a gateway, I updated the PR accordingly. Good find!

xetys:
Your fix breaks the SecurityBuilder in MicroserviceSecurityConfiguration in microservices. After fixing everything the CSRF issue is resolved now.

I also tested the refresh token grant, it works really good. One exception is when remember me is active, you run out of token expire and then logout. The gateway does another token grant and instead of logout, the user gets its session extended.

Metavirulent:
I could not reproduce the problem you described in the second paragraph. There's a window where the refresh token cookie will still be there even though the token it contains has already expired and in that case, it will indeed do a new refresh grant but the cookies still get deleted in the response to the client which should invalidate the session.

If the refresh cookie is already gone, then the server will respond with a 401 which also invalidates the session, so either way the session will end if you log out. I can't seem to be able to extend my session during a logout.

deepu105:
@xetys once you are satisfied with the PR please dont forget to squash the commits as the commit history somehow seems to be corrupted with a lot of duplicate entries

cbornet:
Maybe @Metavirulent could rebase and squash

xetys:
LGTM

xiasenhai:
According to  #6361 , does that mean option **> JHipster UAA server** will be replaced/removed?



@Metavirulent @xetys thanks!

xetys:
nope...AFAIK this PR affects the general OAuth2 support we have. I'm willing to say this PR is moving UAA miles forward towards OutOfBeta

PR #5827, opened 2017-05-24
Fix gradle clean

joshhill:


atomfrede:
The delete of target folder is on purpose as the h2 writes it files to target folder. At least when this was coded it deleted target folder in addition to default gradle build folder. Let me check what's going on now.

joshhill:
Ah OK, thanks for the history. I'll close PR.

Issue #5587, opened 2017-04-12
jhiAlertError not showing alerts after hot module reloading


santoshkt:
##### **Overview of the issue**

jhiAlertError intercepts response and displays error messages after I start the application. But when I make any changes, hot module replacing happens and after that it cannot display error messages in the selector. I don't have any custom changes in my webpack.

##### **Motivation for or Use Case**

HMR should work with jhiAlertError so it will be easy in development

##### **Reproduce the error**

Try to reproduce a scenario which will show errors in jhiAlertError. 
Make some changes in the code to trigger HMR
Try reproducing the same error again

##### **JHipster Version(s)**
```
{
  "generator-jhipster": {
    "jhipsterVersion": "4.0.8",
    "baseName": "jhipster",
    "packageName": "com.github.santoskt.ngxpipes",
    "packageFolder": "com/github/santoskt/ngxpipes",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "jwtSecretKey": "2d99b780856b9651aecf8fe9c51ed01e38455852",
    "clientFramework": "angular2",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}```

- [x] Checking this box is mandatory (this is just to show you read everything)

deepu105:
Fixed via https://github.com/jhipster/ng-jhipster/pull/33
Lib needs to be released and upgraded here hence keeping this open

jdubois:
@deepu105 can you do the ng-jhipster release soon? I'd like to do a generator-jhipster release during the week-end

deepu105:
Sure ill do it today

deepu105:
closing as per comment by @pascalgrimaud in #5829

PR #5829, opened 2017-05-24
Fix alert for entities and users: created, updated, deleted

pascalgrimaud:
I wanted to have a look at https://github.com/jhipster/generator-jhipster/issues/5587 but I realize that the alert system doesn't work.
This PR should fix the alert system for entities, and I can't reproduce https://github.com/jhipster/generator-jhipster/issues/5587 so it should already be fixed in ng-jhipster (let's close https://github.com/jhipster/generator-jhipster/issues/5587)

I'm working on the same for user-management but it's a little bit more complex for me, so it will be in another PR. Give me time.

Plz review closely, I'm not a front developer :)

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/user/_user.service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/user/_user.service.ts
@@ -30,12 +30,14 @@ export class UserService {
 
     constructor(private http: Http) { }
 
-    create(user: User): Observable<Response> {
-        return this.http.post(this.resourceUrl, user);
+    create(user: User): Observable<ResponseWrapper> {
+        return this.http.post(this.resourceUrl, user)
+            .map((res: Response) => this.convertResponse(res));
     }
 
-    update(user: User): Observable<Response> {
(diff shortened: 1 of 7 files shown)


pascalgrimaud:
It doesn't work when `"enableTranslation": false`.
I need to update this PR

pascalgrimaud:
Fixed alert for user-management too.

The tests I did:
- default with translation / without translation
- mongodb with translation / without translation
- cassandra with translation / without translation

generators/client/templates/angular/src/main/webapp/app/shared/user/_user.service.ts
```diff
@@ -30,12 +30,14 @@ export class UserService {
 
     constructor(private http: Http) { }
 
-    create(user: User): Observable<Response> {
-        return this.http.post(this.resourceUrl, user);
+    create(user: User): Observable<ResponseWrapper> {
```
pascalgrimaud: I need to return the user object and the response too.
So I used the ResponseWrapper here

PR #5837, opened 2017-05-26
if i use mongodb this entity's id is "string" type.

gemron:

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
@@ -80,7 +80,7 @@ export class <%= entityAngularName %>Service {
         });
     }
 
-    find(id: number): Observable<<%= entityAngularName %>> {
+    find(id: string|number): Observable<<%= entityAngularName %>> {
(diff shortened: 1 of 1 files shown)


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity.service.ts
```diff
@@ -80,7 +80,7 @@ export class <%= entityAngularName %>Service {
         });
     }
 
-    find(id: number): Observable<<%= entityAngularName %>> {
+    find(id: string|number): Observable<<%= entityAngularName %>> {
```
deepu105: This is better to be handled with ejs so that people dont get confused so you need to do 
`<% if (databaseType == 'sql') { %>string<% } else { %>number<% } %>`

jdubois: Yes +1

jdubois:
I was going to correct the PR, but it's a bit more complex -> I'm doing this in a separate commit, so I'm closing this, and I will link my commit to this ticket. Thanks @gemron for seeing this issue!

gemron:
my pleasure，😄。I will be actively feedback on the problems found in use and hope Jhipster is getting better and better

Issue #5805, opened 2017-05-22
Gateway generation : Unit test failed on windows


hamark:
##### **Overview of the issue**

After generating a gateway via Jhipster 4.5.1, I run the following command on a windows system

`mvnw package`

Unfortunately I have the result  : 

~~~
Failed tests:
  WebConfigurerTest.testCustomizeServletContainer:114 expected:<"target[/]www"> but was:<"target[\]www">
~~~
##### **Suggest a Fix**

Using File.separator instead of "/" in the unit test `WebConfigurerTest.testCustomizeServletContainer`

```
       String targetPath= "target" + File.separator +"www";
       assertThat(container.getDocumentRoot().getPath()).isEqualTo(targetPath);
```

##### **JHipster Version(s)**

4.5.1

##### **JHipster configuration**

```
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.xxx"
    },
    "jhipsterVersion": "4.5.1",
    "baseName": "gatewayGp",
    "packageName": "com.xxx",
    "packageFolder": "com/xxx",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "hazelcast",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "jwtSecretKey": "...",
    "clientFramework": "angular2",
    "useSass": true,
    "clientPackageManager": "yarn",
    "applicationType": "gateway",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
```

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

N/A

##### **Browsers and Operating System**

Windows 7

jdubois:
That really depends on the options you selected, could you **please** follow the guidelines and post your `.yo-rc.json`

hamark:
I have updated the issue

tezcane:
I had done the same exact as mentioned in the suggested fix and can confirm it works on Windows 10 now.

deepu105:
could you please submit a PR?

jdubois:
Yes that would be great, as we don't have Windows!

tezcane:
[...]

PR #5839, opened 2017-05-26
use build folder based on build tool

deepu105:

Make h2 and elasticsearch use build folder according the build system instead of using target always as its a bit confusing

@jdubois do you think its ok? coz I dont remember why we didnt do this originally


atomfrede:
I can't remember a specific reason. So I think it was to have one configuration for both build systems.

jdubois:
Yes it's OK.

deepu105:
need to check why the builds are failing

generators/server/templates/src/test/java/package/config/_WebConfigurerTest.java
```diff
@@ -159,7 +159,7 @@ public void testCustomizeServletContainer() {
         assertThat(container.getMimeMappings().get("json")).isEqualTo("text/html;charset=utf-8");
         <%_ if (!skipClient) { _%>
         if (container.getDocumentRoot() != null) {
-            assertThat(container.getDocumentRoot().getPath()).isEqualTo("target/www");
+            assertThat(container.getDocumentRoot().getPath()).isEqualTo("<%= BUILD_DIR %>www");
```
pascalgrimaud: Related to https://github.com/jhipster/generator-jhipster/issues/5805, I'm not sure it will work on Windows

deepu105: yes it needs to be fixed properly anyway

PR #5846, opened 2017-05-28
Added --history-api-fallback

mystdeim:
Added param for webpack-dev-server. It prevents 404 error in development mode during browser refresh or internal redirect

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -148,7 +148,7 @@
     "webpack:build": "<%= clientPackageManager %> run ngc && webpack --config webpack/webpack.vendor.js && webpack --config webpack/webpack.dev.js",
     "webpack:build:dev": "webpack --config webpack/webpack.dev.js",
     "webpack:build:vendor": "webpack --config webpack/webpack.vendor.js",
-    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9060",
+    "webpack:dev": "webpack-dev-server --config webpack/webpack.dev.js  --progress --inline --hot --profile --port=9060 --history-api-fallback",
     "webpack:test": "<%= clientPackageManager %> run ngc && <%= clientPackageManager %> run test",
     "webpack:prod": "<%= clientPackageManager %> run ngc && webpack -p --config webpack/webpack.vendor.js && webpack -p --config webpack/webpack.prod.js",
     "test": "<%= clientPackageManager %> run lint && karma start src/test/javascript/karma.conf.js",


jdubois:
I haven't tried it but it looks like it would solve some errors I currently have some time, so +1 for merging this

deepu105:
Sorry I didn't get it are you talking about 404 errors on the console? Could you please explain what this is solving?

mystdeim:
For example: you go on the page http://localhost:9000/user-management and refresh your browser after it you get "Cannot GET /user-management" because this route doesn't exist.
If we use "--history-api-fallback", we intercept that request

jdubois:
But this route isn't correct (it misses the hashbang), unless you're using HTML 5 pushstate? Please note we want to support that in the future.

mystdeim:
I created project with Angular 4, it doesn't use hashbang.

deepu105:
Thats strange this needs to be properly checked first

jdubois:
- I just checked and we do have a hashbang
- I'm also pretty sure we didn't code that

-> @mystdeim this is definitely very strange!!! How did you create your application?

For me this is an error, at least for the moment, so we shouldn't merge this.

mystdeim:
@jdubois  I create new application and you right: jhipster uses hashbang. 
Maybe I broke something when updated application from 4.3 to 4.5. But I could find what's wrong. How to return using hashbang? :)

P.S. Angular doc (https://angular.io/docs/ts/latest/guide/router.html) says that they use html 5 style by default.

> HTML5 style navigation is the router default. In the LocationStrategy and browser URL styles Appendix, learn why HTML5 style is preferred, how to adjust its behavior, and how to switch to the older hash (#) style, if necessary.

deepu105:
Yes we switched to hash strategy as getting html5 mode needs more backend
works and is more complicated to setup as app wouldn't know which are
backend routes and which are front end routes resulting in 404 when you try
to access a url directly


Thanks & Regards,
Deepu

mystdeim:
@deepu105 How to return to hash strategy? Because after update my project is in the inconsistent state: html5 strategy on the frontend (I couldn't find how to turn it off), but backend doesn't understand direct route, therefore I became use --history-api-fallback. I didn't face any problem with this option.

jdubois:
- Indeed we'll move to HTML5 pushState, I'm pretty sure it's simple and won't have much issues. It's just that we have a lot of work for the next release, not sure this can be done.
- So your PR might be good for the next release -> let's keep it open, just in case
- Now for returning to the hash strategy, see https://github.com/jhipster/generator-jhipster/search?utf8=%E2%9C%93&q=useHash&type=

jdubois:
I'm closing this because this has been opened for a long time, and I don't think we'll remove the hashbang anytime soon.
Current work is on adding ReactJS support: this totally depends whether we use also the hashbang for React or not.

PR #5853, opened 2017-05-30
modify interaction with Heroku CLI for plugin install and new Heroku account

ggotti:
Fixes two issues with the Heroku CLI invocation on MacOS (10.12.5)
- Heroku generator fails if the plugin has already been installed; the install command returns a -1 error code, causing the program to abort. This fix checks to see if the plugin is already available. If it is, do not install it again.
- If your Heroku account has not been verified, then you cannot install AddOns (such as the Database AddOn). Previous behaviour silently swallows error.

@jkutner would you mind having a look at this?


jkutner:
LGTM! Thanks @ggotti

PR #5864, opened 2017-06-01
Pool max size not set correctly

henri-tremblay:
It needs to be under `hikari` to work. I've also reduce it to 8 because the maximum allowed (at least for JawsDB) is 10. So, when the app in launched, it is impossible to connect with any other client to the database. Which is super annoying. 8 seems to be a good compromise.


jdubois:
LGTM
ping @jkutner as this is his code

jkutner:
LGTM. Thanks @henri-tremblay

Issue #5620, opened 2017-04-18
JHipster behind reverse proxy


Skit5:
##### **Overview of the issue**

The UI doesn't display anything when it's loaded behind a reverse proxy on my remote server (no other port than port 80 opened so I use a proxy_pass with nginx). It's perfectly reachable but it seems that webpack-angular 2 has some trouble with the path http://my-domain/my-path

> vendor.dll.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.manifest.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.global.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.main.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.polyfills.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.main.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)
> c9977571b3904794d148.global.bundle.js Failed to load resource: the server responded with a status of 404 (Not Found)

##### **Motivation for or Use Case**

This is a recurrent problem I encounter every time I run it behind a reverse proxy. Unfortunately, I couldn't find any solution or documentation on the topic.

**It seems that the /*.bundle.js are retrieved from http://my-domain/*.bundle.js instead of http://my-domain/my-path/*.bundle.js**

##### **Reproduce the error**

Just add a: location /my-path/ { proxy_pass: localhost:port; }
in the nginx configuration file (here using Ubuntu 16.04 )

##### **Related issues**

No solution or similar issue found; no documentation at all.

##### **Suggest a Fix**

My current investigation lead me to webpack.dev.js where there might be configurations for the proxy though I couldn't point the real issue or the configuration to make.

##### **JHipster Version(s)**

JHipster Generator v4.1.1

##### **JHipster configuration**
```

{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "*",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.1.1",
    "baseName": "*",
    "packageName": "*",
    "packageFolder": "*",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "no",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
[...]

PR #5871, opened 2017-06-02
use relative base instead of absolute path for context paths and proxies

ruddell:
Related to #5620, this allows a user to deploy the app to any context path without needing to change the base href.  When deploying to the root path, it works the same as in the past.


pascalgrimaud:
Completely forgot to do it, after our discussion on Gitter, some time ago.

I tested the deployment of the application.war.original with tomcat-8 and it works well!
Good job @ruddell !

PR #5878, opened 2017-06-05
Ang 4 moving files to app and content + trimming load + PWA enhancements

sendilkumarn:


sendilkumarn:
@deepu105 FYI

generators/client/templates/angular/_package.json
```diff
@@ -157,7 +157,7 @@
     "e2e": "protractor <%= TEST_SRC_DIR %>protractor.conf.js",
     "postinstall": "webdriver-manager update && node node_modules/phantomjs-prebuilt/install.js && <%= clientPackageManager %> run webpack:build"
     <%_ } else { _%>
-    "postinstall": "node node_modules/phantomjs-prebuilt/install.js && <%= clientPackageManager %> run webpack:build"
+    "postinstall": "node node_modules/phantomjs-prebuilt/install.js"
```
deepu105: why removed?

sendilkumarn:
@deepu105 same here. we can merge this

generators/client/templates/angular/_package.json
deepu105: @sendilkumarn you didnt say why this is removed?

sendilkumarn: @deepu105 will merge this. This will make build to run only once

sendilkumarn:
`angular 1`

ng1


`angular2`

ng2



After we merge this PR, this is where we will end up.

For Best Practices, it is more about HTTP2 push.

We have to work on pay load reduction in ng2 for sure and also make sure both ng1 and ng2 are reduce the load time.

generators/client/templates/angular/_package.json
deepu105: Yes but if someone tries to launch app without running yarn start then it would not work

deepu105: Yes but if someone tries to launch app without running yarn start then it would not work

deepu105: I guess best solution would be to check if build is already run and avoid running bit again from script. Like how its done for vendor script

mraible:
@sendilkumarn Can you attach a PDF of the report or screenshot so we can see what's missing to get to 100 with Angular?

sendilkumarn:
@mraible 

[Angular1.pdf](https://github.com/jhipster/generator-jhipster/files/1062669/Angular1.pdf)
[Angular2.pdf](https://github.com/jhipster/generator-jhipster/files/1062670/Angular2.pdf)

PR #5880, opened 2017-06-05
Making JHipster app PWA 

sendilkumarn:
This adds PWA basic things to the application

- [x] webapp manifest
- [x] chrome browser url bar color change

May be we can add
- [x] basic service worker to it (let people customise it if they want)

I don't want to add app shell architecture.

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -101,7 +101,8 @@ const files = {
                 { file: '_favicon.ico', method: 'copy' },
                 '_robots.txt',
                 '_404.html',
-                '_index.html'
+                '_index.html',
+                '_manifest.webapp'
             ]
         }
     ],
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -102,7 +102,8 @@ const files = {
                 '_robots.txt',
                 '_404.html',
                 '_index.html',
-                '_manifest.webapp'
+                '_manifest.webapp',
+                '_sw.js'
             ]
         }
     ],
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -103,7 +103,6 @@ const files = {
                 '_404.html',
                 '_index.html',
                 '_manifest.webapp',
-                '_sw-loader.js',
                 '_sw.js'
             ]
         }
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -103,6 +103,7 @@ const files = {
                 '_404.html',
                 '_index.html',
                 '_manifest.webapp',
+                '_sw-loader.js',
                 '_sw.js'
             ]
         }
--- a/generators/client/templates/angular/src/main/webapp/_index.html
+++ b/generators/client/templates/angular/src/main/webapp/_index.html
@@ -44,7 +44,15 @@
             document.getElementsByTagName('head')[0].appendChild(script)
         }
     </script>
-    <!-- <script src="sw-loader.js" />-->
+    <!-- uncomment this for adding service worker
+        <script>
+            if ('serviceWorker' in navigator) {
+                 navigator.serviceWorker
+                    .register('./sw.js')
+                    .then(function() { console.log('Service Worker Registered'); });
+            }
+        </script>
+    -->
     <!-- Google Analytics: uncomment and change UA-XXXXX-X to be your site's ID.
     <script>
         (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
--- a/generators/client/templates/angular/src/main/webapp/_index.html
+++ b/generators/client/templates/angular/src/main/webapp/_index.html
@@ -27,6 +27,7 @@
     <meta name="google" value="notranslate"><% } %>
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <link rel="shortcut icon" href="favicon.ico" />
+    <link rel="manifest" href="manifest.webapp">
 </head>
 <body>
     <!--[if lt IE 9]>
--- a/generators/client/templates/angular/src/main/webapp/_index.html
+++ b/generators/client/templates/angular/src/main/webapp/_index.html
@@ -43,12 +43,8 @@
             script.type = 'text/javascript';
             document.getElementsByTagName('head')[0].appendChild(script)
         }
-        /*if ('serviceWorker' in navigator) {
-            navigator.serviceWorker
-                .register('./sw.js')
-                .then(function() { console.log('Service Worker Registered'); });
-        }*/
     </script>
+    <!-- <script src="sw-loader.js" />-->
     <!-- Google Analytics: uncomment and change UA-XXXXX-X to be your site's ID.
     <script>
         (function(b,o,i,l,e,r){b.GoogleAnalyticsObject=l;b[l]||(b[l]=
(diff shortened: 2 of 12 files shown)


deepu105:
@sendilkumarn can we do this after getting ngx out of Beta?

sendilkumarn:
I think we can add this. Service worker can be added later. This change is really very minimal and related to ui alone

deepu105:
ok


Thanks & Regards,
Deepu

generators/client/files-angular.js
```diff
@@ -101,7 +101,8 @@ const files = {
                 { file: '_favicon.ico', method: 'copy' },
                 '_robots.txt',
                 '_404.html',
-                '_index.html'
+                '_index.html',
+                '_manifest.webapp'
```
mraible: I would recommend renaming this file to `manifest.json` since it is a JSON file. Also, Ionic uses convention.

generators/client/templates/angular/src/main/webapp/_manifest.webapp
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "<%=angular2AppName%>",
+  "short_name": "<%=angular2AppName%>",
+  "icons": [
+        {
+            "src": "/content/images/logo-jhipster.png",
+            "sizes": "192x192",
+            "type": "image/png",
+            "density": 4
```
mraible: Lighthouse docks points if you don't have a 512px icon in its 2.0.0 release.

generators/client/files-angular.js
sendilkumarn: our webpack build creates a `manifest.json` file. I don't want to contradict this with that. 
Also `angular mobilekit` preached  ( 😉 ) to use this name.

generators/client/templates/angular/src/main/webapp/_manifest.webapp
sendilkumarn: yeah I know that part. But I don't want to overcomplicate things here. But yes I do agree we can have those 5 combinations which satisfies all the needs.

sendilkumarn:
@mraible  Thanks for the suggestion added service worker and it is commented by default. If people wants they can use that. 

With that for a sample app we are getting above 75% ( without https ) in light house. that is an improvement from 23%

generators/client/templates/angular/src/main/webapp/_index.html
```diff
@@ -41,6 +43,11 @@
             script.type = 'text/javascript';
             document.getElementsByTagName('head')[0].appendChild(script)
         }
+        /*if ('serviceWorker' in navigator) {
```
jdubois: This is going to create "code smells" in Sonar (we already have the issue for the Google Analytics code which is commented), could this be done another way?
Otherwise it's not really important, once minified this won't make any difference

sendilkumarn: There is no harm in uncommenting it 🤔, but really give this options to users such that they can use this, if they want to...

deepu105: may be put this in a separate <script> block and comment that like how we do for GA below

generators/client/templates/angular/src/main/webapp/_index.html
```diff
@@ -42,6 +44,7 @@
             document.getElementsByTagName('head')[0].appendChild(script)
         }
     </script>
+    <!-- <script src="sw-loader.js" />-->
```
deepu105: it need not be separate file, I think it can just be script inside a <script> tag like the google analytics code below. Sorry for not being clear enough earlier

generators/client/templates/angular/webpack/_webpack.common.js
```diff
@@ -120,6 +120,9 @@ module.exports = function (options) {
                 { from: './node_modules/swagger-ui/dist', to: 'swagger-ui/dist' },
                 { from: './src/main/webapp/swagger-ui/', to: 'swagger-ui' },
                 { from: './src/main/webapp/favicon.ico', to: 'favicon.ico' },
+                { from: './src/main/webapp/manifest.webapp', to: 'manifest.webapp' },
+                // { from: './src/main/webapp/main.js', to: 'main.js' },
```
deepu105: did you accidentally comment this?

generators/client/templates/angularjs/src/main/webapp/_index.html
```diff
@@ -119,5 +129,13 @@
     <!-- endinject -->
     <!-- endbuild -->
 
+    <!-- <script>
```
deepu105: duplicate

robertofd1995:
Hello, I have a doubt first of all, thanks for the awesome work you do.

Second, I have uncommented the line of the service worker, I am in running in develop mode but when I open the chrome toolbox, and I go to see if the service worker has been detected, no service worker appears. Also when I check the offline load the page doesn't load.

captura de pantalla 2017-07-04 a las 18 27 41

So there is something else to do to make it work?

Regards Roberto.

deepu105:
ping @sendilkumarn . Can you add some docs to enable PWA in our website please?

deepu105:
@robertofd1995 you need to uncooment the line in webpack.common.js and in index.html

sendilkumarn:
Yeah we have to add in docs. Let me assign myself to that. 👍

robertofd1995:
Hello, @sendilkumarn.

Just for your information, I uncomment also in webpack, but now when I try the offline load, it gives me now and exception "uncaught (in Promise) TypeError: Failed to fetch" in line 1 on the default sw.js.

I am new to PWA and also to JHipster, but if service worker is working, and in the sw.js we declare to put in cache the index.html, shouldn't at least load the HTML of the page?

sendilkumarn:
@robertofd1995 You need to add your generated js / styling files in cache in order to completely load. we will `progressively` enhance them

rakeshkalra1983:
Hi @sendilkumarn, in Prod mode the name of js and styling files are different each time we deploy application, example: app/manifest.e567849023fa94ad663a.bundle.js

How can I add this such files with dynamic names to sw.js?

sendilkumarn:
@rakeshkalra1983 sorry for the delayed response. 

There are many webpack plugins available for this. We need to choose the one among them.

rakeshkalra1983:
@sendilkumarn Thanks for your earlier reply. 

I am facing one problem after switching on the Service Worker: After every new release, my site stops to respond when I call its URL from my browser. To fix this I have to clear the local cache data, and then it starts working.

Since I am naive on UI side, my guess is that we are saying in manifest.mf to cache index.html, and index.html have enteries for all css and java-scripts with some random number in its name. And since this random-numbers change with every release, hence site stops responding. 

I will really appreciate your help.

sendilkumarn:
@rakeshkalra1983 yeah it changes, and we have to generate this file dynamically. Currently I have done it for angular 1. This will be added in JHipster soon. 

For ngx you can try using angular/service-worker and ng-sw-tools webpack plugin 👍

mraible:
I like how Angular CLI treats service workers in that it only enables it when you compile with "prod" mode. This means you can develop normally in development (w/o worrying about caching), but when you do a prod build, the SW is enabled.

gmarziou:
@mraible It makes sense as this is already what we do for server code.

sendilkumarn:
ofcourse that is the plan and using the above tools will surely enable that by default. I am not finding enough time now  to do that. I will try to do it later. In between if you guys can do a PR I can review it with pleasure 😄

caffeinecoders:
Hello, thanks for adding PWA support. I am using Jhipster 4.7.2 version with Angular 2. 
There is one issue I am facing, with PWA turned on, whenever I am making new release with changes in scripts, css or index.html, I can't see those changes on my browser. It seems I am seeing old website only.

Can you please help.

gmarziou:
@caffeinecoders please don't comment on a merged pull request. Open a new issue with all required details.

Issue #5867, opened 2017-06-02
Missed update in Elasticsearch after update user via service/UserService.updateUser


FireWolf2007:
##### **Overview of the issue**

Missed update in Elasticsearch after update user via service/UserService.updateUser

##### **Motivation for or Use Case**

Update user from REST doesn't update it's elasticsearch record.

##### **Reproduce the error**

Update, for example, user name and search user in elasticsearch repository - old name will be returned.

##### **Suggest a Fix**
Add save into updateUser method if used ElasticSearch engine in application.

Our patch for service/UserService.java.java updateUser method:
```
--- a/service/UserService.java
+++ b/service/UserService.java
@@ -190,6 +190,7 @@ public class UserService {
                 userDTO.getAuthorities().stream()
                     .map(authorityRepository::findOne)
                     .forEach(managedAuthorities::add);
+                userSearchRepository.save(user);
                 log.debug("Changed Information for User: {}", user);
                 return user;
             })
```

Patched method:
```
    /**
     * Update all information for a specific user, and return the modified user.
     *
     * @param userDTO user to update
     * @return updated user
     */
    public Optional<UserDTO> updateUser(UserDTO userDTO) {
        return Optional.of(userRepository
            .findOne(userDTO.getId()))
            .map(user -> {
                user.setLogin(userDTO.getLogin());
                user.setFirstName(userDTO.getFirstName());
                user.setLastName(userDTO.getLastName());
                user.setEmail(userDTO.getEmail());
                user.setImageUrl(userDTO.getImageUrl());
                user.setActivated(userDTO.isActivated());
                user.setLangKey(userDTO.getLangKey());
                Set<Authority> managedAuthorities = user.getAuthorities();
                managedAuthorities.clear();
                userDTO.getAuthorities().stream()
                    .map(authorityRepository::findOne)
                    .forEach(managedAuthorities::add);
                userSearchRepository.save(user);
                log.debug("Changed Information for User: {}", user);
                return user;
            })
            .map(UserDTO::new);
    }
```

##### **JHipster Version(s)**

```
telestat@0.0.0 /
├── generator-jhipster@4.5.2 
└─┬ generator-jhipster-elasticsearch-reindexer@0.3.0
  └── generator-jhipster@4.5.1 

```

##### **Browsers and Operating System**

java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
[...]

Issue #5441, opened 2017-03-20
missing imageUrl and elasticsearch saves, various other updates with pending pull request


tezcane:
##### **Overview of the issue**

Bug/Issues:

Issue 1: UserService.updateUser() and AccountResource.saveAccount()
imageUrl user update from /account REST layer is missing

Issue 2: UserService.updateUser(UserDTO userDTO)
elasticsearch save missing here, so "Admin Panel->User management" items aren't updated, query is getting stale data.

Issue 3: creationEmail.html
Email's link is for password reset key, not new account activation

Issue 4: User.java
email min size of 5 is missing from the Size annotation

Issue 5: AccountResourceIntTest.testRegisterDuplicateLogin():
"getLogin()" called twice instead of "getLastName()"

Issue 6: AccountResourceIntTest.testGetExistingAccount():
Add missing langKey insertion and check

SMALLER ITEMS - Code Cleanup Items

7. User.equals(): replace if else block with "return login.equals(user.login)"
8. activate.json: "Your user has been activated." to "Your user account has been activated."
9. activate.json: "Please" to "Please sign in."
10. comment updates: "succesful" to "successful" and "doesnt" to "doesn't"1
11. ManagedUserVM.java: ",  imageUrl" to ", imageUrl"
12. pom.xml: " </plugins" to "</plugins" (one extra space above </pluginManagement) 
13. ProfileInfoResourceIntTest.java: remove unused import ...beans.factory.annotation.Autowired;
14. RateLimiting.java:  remove unused import ... javax.servlet.http.HttpServletResponse;
15. user-management.components.ts: constructor alignment fixed
16. user-management.components.html: "<th  " to "<th "
17. UserResourceIntTest.java: rename "autorities" to "authorities" across file
18. UserResourceIntTest.updateUserExistingEmail(): remove unused variable "databaseSizeBeforeUpdate"
19. UserResourceIntTest.updateUserExistingLogin(): remove unused variable "databaseSizeBeforeUpdate"
20. UserService.java: "@Transactional(readOnly = true)    " to "@Transactional(readOnly = true)"

##### **Motivation for or Use Case**

These are issues and code clean up that is needed by the community.

##### **Reproduce the error**

For elasticsearch issue:
1. Use Elasticsearch enabled project
2. Update a setting from a normal user, e.g. change the email
3. Go to admin panel - User management and see the update has not come

##### **Related issues**

No

##### **Suggest a Fix**

[...]

PR #5885, opened 2017-06-07
Fix missing Elasticsearch reindex after updating User

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/5867

___

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -287,6 +287,9 @@ public class UserService {
                 <%_ if (databaseType == 'mongodb' || databaseType == 'cassandra') { _%>
                 userRepository.save(user);
                 <%_ } _%>
+                <%_ if (searchEngine === 'elasticsearch') { _%>
+                userSearchRepository.save(user);
+                <%_ } _%>
                 log.debug("Changed Information for User: {}", user);
                 return user;
             })


cbornet:
We should add a test for that

pascalgrimaud:
Indeed, I forget to check the coverage for that.
But I'm pretty sure this new line will be covered, as the UserService is at 95.9% (thanks to your hard work!!)

cbornet:
Yes the coverage will be the same. But adding an assertion will strengthen the test (code coverage is a lure if you don't verify anything)

pascalgrimaud:
Again I totally agree with you @cbornet, but I think it should be in another ticket, because today, when choosing ElasticSearch, we don't have tests on api/_search in UserResourceIntTest, either Entities.

tezcane:
I have reported this a few times, e.g. in #5541 #5542 and one other related bug which I can't find right now.  I wonder why it was ignored when I tried to fix it @jdubois

jdubois:
Sorry @tezcane but I don't understand... Don't take it bad, we're doing this mostly on our free time (on my phone while commuting, at the moment), so we can miss stuff.

tezcane:
@jdubois I don't take it bad, no worries.  I was always questioning myself why this fix was not needed and how it was meant by design like I was told... I brought it up so we can spark some thoughts on not losing out on bug fixes in the future.  When the bug was rejected I tried adding a comment to the fact that it was left out by design (as I was told), but that was also rejected.

@pascalgrimaud Sorry, I put in wrong bug links.  I fixed it now.

pascalgrimaud:
@tezcane : ok, your initial ticket contained a lot of item and we simply miss it sorry. That's why we always ask to open 1 ticket for 1 bug, with all description to reproduce. Better, you can to directly a PR if you are sure about the fix.

So reading this ticket https://github.com/jhipster/generator-jhipster/issues/5441 : is everything fixed ?

tezcane:
Agreed, my first commits were a train wreck... I misread generator contributor documentation and thought PR's were preferred over opening issues first.  

#5441 everything is fixed, yes.

It is not just mentioned in #5441 though, this issue was highlighted in a few places (other bugs, PR comments, etc.). I should have followed up more maybe, but was too trusting and felt like I may be pestering my heros (yes you all are many developers heros :) ) by asking too many questions on a dead/concluded topic.

jdubois:
@tezcane you expect too much from us :-)
we're just normal guys, doing mistakes and not totally reading their emails while commuting...

Issue #5858, opened 2017-05-30
Error double notification


mtdx:
##### **Overview of the issue**

https://gyazo.com/61dd0762032221da5c277a73f8679f45
https://gyazo.com/1a2f3005425e1d6cabaacd8052aa877f

There are 2 small issues in those gifs, first when you're in a popup and some error happens, it also shows in the back (behind the popup), in this case the problem is the image being to big, but is not the first time I see it.

Also the "Field Image cannot be empty" isn't that descriptive for a max bytes restriction. 

##### **Reproduce the error**

Generate a simple entity, with an image field, with a max bytes size restriction and try to upload a bigger size image.

##### **JHipster Version(s)**

freshly generated app, JHipster 4.5.2

##### **JHipster configuration**

java version "1.8.0_131"
node: v6.10.3
npm: 4.2.0
yeoman: 1.8.5
yarn: 0.24.6

##### **Browsers and Operating System**

macOS, Chrome browser

JulienMrgrd:
I can take this issue

deepu105:
You are still not following the guidelines. We need your yo-rc.json and entity.json. can you plz run jhipster info on your app and post the results

mtdx:
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.xxx.xxx",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.5.2",
    "baseName": "xxx",
    "packageName": "com.xxx.xxx",
    "packageFolder": "com/xxx/xxx",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": "spring-websocket",
    "databaseType": "sql",
    "devDatabaseType": "postgresql",
    "prodDatabaseType": "postgresql",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": true,
    "rememberMeKey": "506073c2654996412640f902f76e6c270b46a403",
    "clientFramework": "angular2",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "de"
    ]
  }
}


{
    "fluentMethods": true,
[...]

PR #5886, opened 2017-06-07
Some fixes related to issue #5858

JulienMrgrd:

--------------------------------

In case a new version comes out, here are some corrections of #5858, but it still lacks 2 tasks:

- [ ] Do not display the same error several times when clicking the save button (without having modified anything). http://imgur.com/a/cz0zH

- [ ] The "Server Error notification"

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
@@ -28,7 +28,9 @@ import { Subscription } from 'rxjs/Rx';
     template: `
         <div class="alerts" role="alert">
             <div *ngFor="let alert of alerts"  [ngClass]="{\'alert.position\': true, \'toast\': alert.toast}">
-                <ngb-alert type="{{alert.type}}" close="alert.close(alerts)"><pre [innerHTML]="alert.msg"></pre></ngb-alert>
+                <ngb-alert *ngIf="alert && alert.type && alert.msg" [type]="alert.type" (close)="alert.close(alerts)">
+                    <pre [innerHTML]="alert.msg"></pre>
+                </ngb-alert>
             </div>
         </div>`
 })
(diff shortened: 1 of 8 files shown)


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -122,7 +126,24 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
             });
         }
     }
-   <%_ } _%>
+
+    <%_ if (fieldsContainImageBlob) { _%>
+    clearInputImage(field: string, fieldContentType: string, idInput: string) {
```
deepu105: since this method is common may be this can be moved to data-util file and be called from here, may be even pass elementRef

generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts
```diff
@@ -77,7 +79,7 @@ export class <%=jhiPrefixCapitalized%>AlertErrorComponent implements OnDestroy {
                                 fieldError.objectName + '.' + convertedField)<% } else { %>convertedField.charAt(0).toUpperCase() +
                                 convertedField.slice(1)<% } %>;
                             this.addErrorAlert(
-                                'Field ' + fieldName + ' cannot be empty', 'error.' + fieldError.message, { fieldName });
+                                'Error on field "' + fieldName + '"', 'error.' + fieldError.message, { fieldName });
```
sendilkumarn: can we make it template string ?

JulienMrgrd: All those texts in the alert error component are in the Typescript (by default), and passed to the addAlertError function.
As it's used as a directive, I don't know exactly where do you want to put this default text. With some inputs in the directive ?

JulienMrgrd:
(Don't pay attention to the above git history)

@deepu105 : I've applied your review on the dialog.component, with the new "Move common method" commit (on this PR) and with a new ng-jhipster PR (jhipster/ng-jhipster#38)

deepu105:
@JulienMrgrd I have made a new release 0.1.18 for ng-jhipster you can update the version here and travis should pass

PR #5887, opened 2017-06-08
Change Eureka Environment variable to valid character

ggotti:
The “EUREKA_CLIENT_SERVICE-URL_DEFAULTZONE” environment variable specified in the _app.yml file contains an invalid hyphen(-) character. When the container is started, this variable is not set. This change replaces the hyphen with an underscore, which Spring Boot can still read due to its relaxed binding rules. This change aligns the _app.xml file it to the Kubernetes+Openshift versions of this template.

### DIFF
--- a/generators/server/templates/src/main/docker/_app.yml
+++ b/generators/server/templates/src/main/docker/_app.yml
@@ -27,7 +27,7 @@ services:
             - SPRING_CLOUD_CONSUL_PORT=8500
             <%_ } _%>
             <%_ if (serviceDiscoveryType == 'eureka') { _%>
-            - EUREKA_CLIENT_SERVICE-URL_DEFAULTZONE=http://admin:$${jhipster.registry.password}@jhipster-registry:8761/eureka
+            - EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=http://admin:$${jhipster.registry.password}@jhipster-registry:8761/eureka
             - SPRING_CLOUD_CONFIG_URI=http://admin:$${jhipster.registry.password}@jhipster-registry:8761/config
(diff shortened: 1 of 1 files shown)


jdubois:
I'm *very* surprised this variable is not set, we use this all the time. Are you sure about this?

PierreBesson:
It also surprise me that it causes problems but anyway this is the correct way to write it and tests are passing so I'm merging it.

pascalgrimaud:
@PierreBesson : FYI, there are no Travis tests on this part, but it seems correct, so OK for merging.

Anyway, I have some comments:
- this property `EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE` should not be there, as it should be provided by Spring Cloud Config from `application.yml`
- the config for the jhipster-registry in the generated `app.yml` is not really consistent with the generated files from docker-compose, kubernetes or openshift -> we need to check that

PierreBesson:
@pascalgrimaud Oh I didn't notice that this is the `app.yml` file. We don't use this very often as we mostly use the docker subgen.
Indeed, the property should not be here and be provided by the config server.

jdubois:
OK, that would explain why we don't have any issue, as it's always overridden by the config server.

Then I don't see where it's configured, for example it's not in our [sample config](https://github.com/jhipster/jhipster-registry-sample-config/blob/master/application.yml) - there's something wrong here.

And if that's not used at all, we should as well remove the variable.

jdubois:
And reading https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html I'm pretty sure a "-" is bad, but does the same job as a "_"

Issue #5754, opened 2017-05-12
Feature request: deploying outside of the WAR


yfain:
I like JHipster, but would like to have more deployments options.

##### **Overview of the issue**

Currently, JHipster packages Angular app inside the WAR file:
https://jhipster.github.io/production/

##### **Motivation for or Use Case**

An Angular app is a static resource and I don't think that every user's request should go to the Spring Boot or any app server.  I'd rather deploy an Angular app under a high-performance Web server like NGINX, and one or more Spring Boot instances would run behind it. This would allow configuring proxies on NGINX so some of the requests wouldn't even hit Spring Boot. I created a diagram showing what I'd like to be able to do.

<img width="750" alt="ng" src="https://cloud.githubusercontent.com/assets/2254842/26018033/9ebd88a2-373a-11e7-9605-40adb21bb015.png">

Thank you!

Vaelyr:
There is a microservice architecture already developed just for that purpose.

https://jhipster.github.io/microservices-architecture/

deepu105:
yes please use the microservice architecture for such use cases for monolith we want it to be simple as possible

jdubois:
- For me this is totally valid to do this with monoliths. I know people doing the same kind of thing using Varnish, for example.
- From a performance point of view, I'm not sure it's such a good idea : serving static files with Spring Boot is very fast, and adding one layer on top of each REST request will slow them a little bit. So I'm not sure here.
- From a scalability point of view this is a good idea, and something I would like to push, at least for monoliths (otherwise you can't scale them, and parts of the work we did with Hazelcast is useless)
- This is in fact a pattern I wanted to have with AngularJS 1 (not sure how it will work with Angular 4, but shouldn't be very different). I also wanted to be able to serve static assets through a CDN. We should be able to document it, as basically all the front end is in one folder, so you just need to move it around. You could also do the opposite: all RESR requests start with "/api/" so that should be easy to filter.

jdubois:
Re-opening, at least so we document this.
We could also generate a Docker Compose file that configures Nginx for a monolith.

yfain:
@jdubois >serving static files with Spring Boot is very fast, and adding one layer on top of each REST request will slow them a little bit.

Nginx is blazing fast but it's not only about speed. Some third party services may offer private API  keys that are not safe to use in the browsers. We could store them in the Nginx layer, which will be adding these keys while proxying requests to such services.

stevehouel:
[...]

PR #5890, opened 2017-06-08
initial commit - infinispan changes

srinivasa-vasu:
infinispan initial commit

- for entity L2 cache
- jcache/spring abstractions and 
- clustered sessions

won't run on kubernetes/openshift because of lack of hibernate support for 9.x Infinispan releases

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/domain/_Entity.java
+++ b/generators/entity/templates/server/src/main/java/package/domain/_Entity.java
@@ -181,9 +181,8 @@ public class <%= entityClass %> implements Serializable {
 
     <%_ } else if (relationshipType === 'many-to-many') { _%>
     @ManyToMany<% if (ownerSide === false) { %>(mappedBy = "<%= otherEntityRelationshipNamePlural %>")
-    @JsonIgnore
-    <%_     } else { _%>
-    <%_     } if (hibernateCache !== 'no') { if (hibernateCache === 'infinispan') { _%>
+    @JsonIgnore<% } %>
+    <%_      if (hibernateCache !== 'no') { if (hibernateCache === 'infinispan') { _%>
     @Cache(usage = CacheConcurrencyStrategy.READ_WRITE) <%_     } else { _%>
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
     <%_     } }
--- a/generators/entity/templates/server/src/main/java/package/domain/_Entity.java
+++ b/generators/entity/templates/server/src/main/java/package/domain/_Entity.java
@@ -21,50 +21,79 @@ let importApiModelProperty = false;
 let importJsonignore = false;
 let importSet = false;
 const uniqueEnums = {}; %><%- include imports -%>
-<% if (databaseType === 'cassandra') { %>
-import com.datastax.driver.mapping.annotations.*;<% } %><% if (importJsonignore === true) { %>
-import com.fasterxml.jackson.annotation.JsonIgnore;<% } %><% if (typeof javadoc != 'undefined') { %>
-import io.swagger.annotations.ApiModel;<% } %><% if (importApiModelProperty === true) { %>
-import io.swagger.annotations.ApiModelProperty;<% } %><% if (hibernateCache !== 'no') { %>
+
+<%_ if (databaseType === 'cassandra') { _%>
+import com.datastax.driver.mapping.annotations.*;
+<%_ } if (importJsonignore === true) { _%>
+import com.fasterxml.jackson.annotation.JsonIgnore;
+<%_ } if (typeof javadoc != 'undefined') { _%>
+import io.swagger.annotations.ApiModel;
+<%_ } if (importApiModelProperty === true) { _%>
+import io.swagger.annotations.ApiModelProperty;
+<%_ } if (hibernateCache !== 'no') { _%>
 import org.hibernate.annotations.Cache;
-import org.hibernate.annotations.CacheConcurrencyStrategy;<% } %><% if (databaseType === 'mongodb') { %>
+import org.hibernate.annotations.CacheConcurrencyStrategy;
+<%_ } if (databaseType === 'mongodb') { _%>
 import org.springframework.data.annotation.Id;
 import org.springframework.data.mongodb.core.mapping.Document;
-import org.springframework.data.mongodb.core.mapping.Field;<% } %><% if (searchEngine === 'elasticsearch') { %>
-import org.springframework.data.elasticsearch.annotations.Document;<% } %>
-<% if (databaseType === 'sql') { %>
-import javax.persistence.*;<% } %><% if (validation) { %>
-import javax.validation.constraints.*;<% } %>
-import java.io.Serializable;<% if (fieldsContainBigDecimal === true) { %>
-import java.math.BigDecimal;<% } %><% if (fieldsContainBlob && databaseType === 'cassandra') { %>
-import java.nio.ByteBuffer;<% } %><% if (fieldsContainInstant === true) { %>
-import java.time.Instant;<% } %><% if (fieldsContainLocalDate === true) { %>
-import java.time.LocalDate;<% } %><% if (fieldsContainZonedDateTime === true) { %>
-import java.time.ZonedDateTime;<% } %><% if (importSet === true) { %>
+import org.springframework.data.mongodb.core.mapping.Field;
+<%_ } if (searchEngine === 'elasticsearch') { _%>
+import org.springframework.data.elasticsearch.annotations.Document;
+<%_ } if (databaseType === 'sql') { _%>
+
+import javax.persistence.*;
+<%_ } if (validation) { _%>
+import javax.validation.constraints.*;
+<%_ } _%>
+import java.io.Serializable;
+<%_ if (fieldsContainBigDecimal === true) { _%>
+import java.math.BigDecimal;
+<%_ } if (fieldsContainBlob && databaseType === 'cassandra') { _%>
+import java.nio.ByteBuffer;
+<%_ } if (fieldsContainInstant === true) { _%>
+import java.time.Instant;
+<%_ } if (fieldsContainLocalDate === true) { _%>
+import java.time.LocalDate;
+<%_ } if (fieldsContainZonedDateTime === true) { _%>
+import java.time.ZonedDateTime;
+<%_ } if (importSet === true) { _%>
 import java.util.HashSet;
-import java.util.Set;<% } %>
-import java.util.Objects;<% if (databaseType === 'cassandra') { %>
-import java.util.UUID;<% } %><% Object.keys(uniqueEnums).forEach(function(element) { %>
+import java.util.Set;
+<%_ } _%>
+import java.util.Objects;
+<%_ if (databaseType === 'cassandra') { _%>
+import java.util.UUID;
+<%_ }
+Object.keys(uniqueEnums).forEach(function(element) { _%>
 
-import <%=packageName%>.domain.enumeration.<%= element %>;<% }); %>
+import <%=packageName%>.domain.enumeration.<%= element %>;
+<%_ }); _%>
 
-<% if (typeof javadoc == 'undefined') { -%>
+<%_ if (typeof javadoc == 'undefined') { _%>
 /**
  * A <%= entityClass %>.
  */
-<% } else { -%>
+<%_ } else { _%>
 <%- formatAsClassJavadoc(javadoc) %>
 @ApiModel(description = "<%- formatAsApiDescription(javadoc) %>")
-<% } -%>
-<% if (databaseType === 'sql') { -%>
+<%_ } _%>
+<%_ if (databaseType === 'sql') { _%>
 @Entity
 @Table(name = "<%= entityTableName %>")
-<%_ if (hibernateCache !== 'no') { if (hibernateCache === 'infinispan') { _%>
-@Cache(usage = CacheConcurrencyStrategy.READ_WRITE) <%_ } else { _%>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<%_ } } _%><% } %><% if (databaseType === 'mongodb') { %>
-@Document(collection = "<%= entityTableName %>")<% } %><% if (databaseType === 'cassandra') { %>
-@Table(name = "<%= entityInstance %>")<% } %><% if (searchEngine === 'elasticsearch') { %>
-@Document(indexName = "<%= entityInstance.toLowerCase() %>")<% } %>
+<%_     if (hibernateCache !== 'no') {
+            if (hibernateCache === 'infinispan') { _%>
+@Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
+<%_         } else { _%>
+@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
+<%_         }
+        }
+} if (databaseType === 'mongodb') { _%>
+@Document(collection = "<%= entityTableName %>")
+<%_ } if (databaseType === 'cassandra') { _%>
+@Table(name = "<%= entityInstance %>")
+<%_ } if (searchEngine === 'elasticsearch') { _%>
+@Document(indexName = "<%= entityInstance.toLowerCase() %>")
+<%_ } _%>
 public class <%= entityClass %> implements Serializable {
 
     private static final long serialVersionUID = 1L;
@@ -76,11 +105,14 @@ public class <%= entityClass %> implements Serializable {
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "sequenceGenerator")
     @SequenceGenerator(name = "sequenceGenerator")
     <%_ } _%>
-    private Long id;<% } %><% if (databaseType === 'mongodb') { %>
+    private Long id;
+<%_ } if (databaseType === 'mongodb') { _%>
     @Id
-    private String id;<% } %><% if (databaseType === 'cassandra') { %>
+    private String id;
+<%_ } if (databaseType === 'cassandra') { _%>
     @PartitionKey
-    private UUID id;<% } %>
+    private UUID id;
+<%_ } _%>
 
 <%_ for (idx in fields) {
     if (typeof fields[idx].javadoc !== 'undefined') { _%>
@@ -117,8 +149,8 @@ public class <%= entityClass %> implements Serializable {
     @Column(name = "<%-fieldNameAsDatabaseColumn %>", precision=10, scale=2<% if (required) { %>, nullable = false<% } %>)
         <%_ } else { _%>
     @Column(name = "<%-fieldNameAsDatabaseColumn %>"<% if (fieldValidate === true) { %><% if (fieldValidateRules.indexOf('maxlength') !== -1) { %>, length = <%= fieldValidateRulesMaxlength %><% } %><% if (required) { %>, nullable = false<% } %><% } %>)
-    <%_     }
-        } _%>
+        <%_ }
+    } _%>
     <%_ if (databaseType === 'mongodb') { _%>
     @Field("<%=fieldNameUnderscored %>")
     <%_ } _%>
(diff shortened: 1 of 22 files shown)


generators/entity/templates/server/src/main/java/package/domain/_Entity.java
```diff
@@ -58,8 +58,9 @@
 <% } -%>
 <% if (databaseType == 'sql') { -%>
 @Entity
-@Table(name = "<%= entityTableName %>")<% if (hibernateCache != 'no') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType == 'mongodb') { %>
+@Table(name = "<%= entityTableName %>")
+<% if (hibernateCache != 'no') { if (hibernateCache == 'infinispan') { %>@Cache(usage = CacheConcurrencyStrategy.READ_WRITE) <% } else { %>
```
deepu105: since this part is becoming complex may be we can switch to using `<%_` to have newline slurped
so
```
<%_ if (hibernateCache != 'no') { if (hibernateCache == 'infinispan') { _%>
@Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
<%_ } else { _%>
<%_ } } 
} if (databaseType == 'mongodb') { _%>
```

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
```diff
@@ -58,8 +58,9 @@
...
+@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } } %><% } %><% if (databaseType == 'mongodb') { %>
```
deepu105: doesnt infinispan support `NONSTRICT_READ_WRITE` ?

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
deepu105: can we also use `===` and `!==` on conditions. I know existing code has `==` but its old code and we need to change it all so atleast when we add new stuff we need to use `===`

generators/server/templates/src/main/java/package/config/_CacheConfiguration.java
```diff
@@ -25,8 +25,8 @@
 import org.ehcache.expiry.Duration;
 import org.ehcache.expiry.Expirations;
 import org.ehcache.jsr107.Eh107Configuration;
-
 import java.util.concurrent.TimeUnit;
+import org.springframework.boot.autoconfigure.cache.JCacheManagerCustomizer;
```
deepu105: is this needed for hazelcast?

generators/server/templates/src/main/java/package/config/_CacheConfiguration.java
```diff
@@ -277,4 +291,60 @@ public SessionRegistry sessionRegistry() {
     }
     <%_ } _%>
     <%_ } _%>
+
+    <%_ if (hibernateCache == 'infinispan' || clusteredHttpSession == 'infinispan') { _%>
+    private final Logger log = LoggerFactory.getLogger(CacheConfiguration.class);
+
+    /* for customizing session cookie key; default is 'session'
+    @Bean
+    public org.springframework.session.web.http.CookieSerializer cookieSerializer() {
```
deepu105: we dont normally qualify classes like this. Wouldnt the IDE import it when the code is un-commented? anyway I'm not sure of it. @jdubois what do you think?

generators/server/templates/src/main/java/package/config/_SecurityConfiguration.java
```diff
@@ -170,6 +188,10 @@ protected void configure(HttpSecurity http) throws Exception {
             .sessionManagement()
             .maximumSessions(32) // maximum number of concurrent sessions for one user
             .sessionRegistry(sessionRegistry)
+            .and().and()<% } %><% if (clusteredHttpSession == 'infinispan') { %>
+            .sessionManagement()
+            .maximumSessions(32) // maximum number of concurrent sessions for one user
+            .sessionRegistry(sessionRegistry())
             .and().and()<% } %>
```
deepu105: @jhipster/developers why do we use `and().and()` is it by mistake or was there a reason to do this? looks weird

generators/server/templates/src/main/java/package/domain/_Authority.java
```diff
@@ -34,8 +34,10 @@
  * An authority (a security role) used by Spring Security.
  */<% if (databaseType == 'sql') { %>
 @Entity
-@Table(name = "jhi_authority")<% if (hibernateCache != 'no') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType == 'mongodb') { %>
+@Table(name = "jhi_authority")<% if (hibernateCache != 'no') {
```
deepu105: we can use slurping template `<%_ _%>` here as well to make this more readable

generators/server/templates/src/main/java/package/domain/_PersistentToken.java
```diff
@@ -40,7 +40,8 @@
  */<% if (databaseType == 'sql') { %>
 @Entity
 @Table(name = "jhi_persistent_token")<% } %><% if (hibernateCache != 'no' && databaseType == 'sql') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% if (databaseType == 'mongodb') { %>
+<% if (hibernateCache == 'infinispan') { %>@Cache(usage = CacheConcurrencyStrategy.READ_WRITE)<% } else {%>
```
deepu105: `<%_ _%>`

generators/server/templates/src/main/java/package/domain/_SocialUserConnection.java
```diff
@@ -44,8 +44,10 @@
  * A Social user.
  */<% if (databaseType == 'sql') { %>
 @Entity
-@Table(name = "jhi_social_user_connection")<% if (hibernateCache != 'no') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType == 'mongodb') { %>
+@Table(name = "jhi_social_user_connection")<% if (hibernateCache != 'no') {
```
deepu105: `<%_ _%>`

generators/server/templates/src/main/java/package/domain/_User.java
```diff
@@ -53,7 +53,8 @@
  */<% if (databaseType == 'sql') { %>
 @Entity
 @Table(name = "jhi_user")<% } %><% if (hibernateCache != 'no' && databaseType == 'sql') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% if (databaseType == 'mongodb') { %>
+<% if (hibernateCache == 'infinispan') { %>@Cache(usage = CacheConcurrencyStrategy.READ_WRITE)<% } else {%>
```
deepu105: `<%_ _%>`

generators/server/templates/src/main/java/package/domain/_User.java
```diff
@@ -148,15 +149,19 @@
     @JoinTable(
         name = "jhi_user_authority",
         joinColumns = {@JoinColumn(name = "user_id", referencedColumnName = "id")},
-        inverseJoinColumns = {@JoinColumn(name = "authority_name", referencedColumnName = "name")})<% if (hibernateCache != 'no') { %>
-    @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% if (databaseType == 'sql') { %>
+        inverseJoinColumns = {@JoinColumn(name = "authority_name", referencedColumnName = "name")})
+    <% if (hibernateCache != 'no') {
```
deepu105: `<%_ _%>`

generators/server/templates/src/main/java/package/domain/_User.java
```diff
@@ -148,15 +149,19 @@
...
+    @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } } %><% if (databaseType == 'sql') { %>
     @BatchSize(size = 20)<% } %><% } %><% if (databaseType == 'sql' || databaseType == 'mongodb') { %>
     private Set<Authority> authorities = new HashSet<>();<% } %><% if (databaseType == 'cassandra') { %>
     private Set<String> authorities = new HashSet<>();<% } %><% if (authenticationType == 'session' && databaseType == 'sql') { %>
 
     @JsonIgnore
-    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "user")<% if (hibernateCache != 'no') { %>
-    @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %>
+    @OneToMany(cascade = CascadeType.ALL, orphanRemoval = true, mappedBy = "user")
+    <% if (hibernateCache != 'no') {
```
deepu105: `<%_ _%>`

generators/server/templates/src/main/resources/config/_application-dev.yml
```diff
@@ -256,6 +263,23 @@ jhipster:
             time-to-live-seconds: 3600 # By default objects stay 1 hour in the cache
             max-entries: 100 # Number of objects in each cache entry
         <%_ } _%>
+        <%_ if (hibernateCache == 'infinispan') { _%>
+        infinispan:
+            config-file: default-configs/default-jgroups-udp.xml
```
deepu105: is the file provided by infinispan?

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
srinivasa-vasu: It does, but the default cache mode for L2 is 'invalidation' mode. Though it is similar to replication, but it differs by,
A write operation would update the cache in the owner node and remove the stale values from the other nodes' memory (will not do an immediate replication). Benefit of this, network traffic is minimised as the invalidation messages are minimal compared to replicating the updated cache entity and the other nodes will look up the updated data in a lazy manner when needed. This mode is the recommended for L2 over replicated and distributed mode and it works only with READ_WRITE. For [more info](http://planet.jboss.org/post/hibernate_second_level_cache_improvements)

generators/server/templates/src/main/java/package/config/_CacheConfiguration.java
srinivasa-vasu: I think this is used for ehcache. I just changed the import order. I believe it was hanging separately in a if condition.

generators/server/templates/src/main/java/package/config/_CacheConfiguration.java
srinivasa-vasu: Yes it does. If I am not wrong, in eclipse u can do cntrl+alt+o, but in Intellij, you have to delete few characters backwards to activate intellisense completion. I just included it so that users can import the write class when they uncomment.

generators/server/templates/src/main/java/package/config/_SecurityConfiguration.java
srinivasa-vasu: That's how Spring fluent API is designed.

generators/server/templates/src/main/resources/config/_application-dev.yml
srinivasa-vasu: Yes (comes from jar) - default implementation. Users can customise and include local file as well by specifying the path here.

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
```diff
@@ -58,8 +58,10 @@
 <% } -%>
 <% if (databaseType == 'sql') { -%>
 @Entity
-@Table(name = "<%= entityTableName %>")<% if (hibernateCache != 'no') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType == 'mongodb') { %>
+@Table(name = "<%= entityTableName %>")
+<%_ if (hibernateCache !== 'no') { if (hibernateCache === 'infinispan') { _%>
+@Cache(usage = CacheConcurrencyStrategy.READ_WRITE) <%_ } else { _%>
```
deepu105: Using slurp scriptlet should always be in newline else it will create bad formatted output during generation

generators/server/templates/src/main/java/package/config/_SecurityConfiguration.java
deepu105: Wow that's weird

deepu105:
@pascalgrimaud we need to add this to travis tests

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
srinivasa-vasu: Changed and checked the format as well. Its good.

pascalgrimaud:
@deepu105 : yes of course ! will add a new build

deepu105:
LGTM @jdubois what do you think?

jdubois:
Overall that's of course very good work, and an excellent new option.

I'll do a more detailed code review later, as I saw a few small things to improve (nothing important), but I have 2 bigger questions, from an architecture point of view:

- It's cool that you wanted to add clustered HTTP session, but to be honest this is a lot of trouble for very little gain. This adds a lot of work to test/validate, and very few people use it (and most people use it for the wrong reason, each time I talked with someone about this, he thought it was doing different). For me this is an option we should remove globally. Anyway, I don't think you should support that.
- For clustering, I understand there is a file provided by Infinispan, which is not configured by default. I'm pretty sure that this uses multicast to find other nodes (that's what's everybody is doing by default). Unfortunately, this won't work on most cloud providers. Did you have a look at our Hazelcast code for this? We use Eureka (=the JHipster Registry) to get the other instances, and create a cluster on-the-fly using the Hazelcast API. Could you do the same with Infinispan? For me this is a must-have feature.

srinivasa-vasu:
_Clustered HTTP session_
- Same implementation is being inherited in JEE containers like Wildfly/EAP (JGroups) which is tested, proven and reliable
- Activating is very simple with spring boot support added in 9.x releases - just an annotation to be added 
- Session gets managed in a similar way cache gets managed.
I agree with you - Though the evangelism and focus is more on cloud native stateless applications, it is very difficult to imagine a complete business service without a state. Of course state can be handled in multiple ways (even if we implement a pure 12f cloud native app), but still lot of enterprises are pretty much used to stateful applications. Even same traction is visibly seen of late with the likes of Kubernetes/OpenShift where they have added support for stateful applications through persistent storage and statefulsets in the recent releases. Cloudfoundry as well is doing something in that area though they favoured 12f to the core. 
You would really see the magic in Kubernetes/OpenShift while doing a zero downtime deployment where the session gets managed seamlessly when rolling out updated versions of apps/services. It just works like a magic as far as whatever I have experienced.

_Clustering & Discovery_ 
JGroups supports multiple protocols and discovery mechanisms. Though UDP's default discovery is based on MPING multicast, it supports TCP, Tunnel etc., with alternative discovery mechanisms like TCPPING, TCPGOSSIP, JDBC_PING etc. It has separate config file for UDP, TCP and as well for AWS(default-jgroups-ec2.xml), GCP(default-jgroups-google.xml) and Azure (jdbc-ping and gossip) where discovery through multicast may be blocked or not supported.
In a similar way, it supports KUBE-PING and as well with a separate configuration for Kubernetes/OpenShift. Support for Kubernetes is added in 9.x releases. This is a limitation now, as hibernate-infinispan support is not yet released for 9.x. We can't run infinispan powered apps on Kubernetes/OpenShift because of this limitation.

Let me have a look at Hazelcast-Eureka discovery mechanism as well.

deepu105:
@srinivasa-vasu one clarification. I'm pretty sure that when @jdubois said that session clustering is not important its from the perspective of the Jhipster community as from our Google analytics usage stats it is evident that session clustering is used very less and most people who raise issues regarding it end up realising that they choose the option by accident. 
Also in the past we had a lot of issues arising from the session clustering option thats why we were thinking of removing it. But these days it seems quite stable, thanks to the recent rewrite of our implementation by both ehcache and hazelcast teams.

jdubois:
@srinivasa-vasu did you know I was the original author of the JBoss HA and clustering documentation? Too bad the JBoss team removed my name :-(
Anyway, just to tell you I know how this works, so as @deepu105 says let me clarify a few things:

- Indeed I'm pushing for 12F apps, and that's the original goal of JHipster.
- Our stats do show that the great majority of users are not using clustering, but the most important thing is that when I talk to people who chose clustering, they usually have no idea what that means. So I really see this as a pitfall.
- I also want to push sticky sessions: for me this is enough for most people using clustering, and also much more performant (obviously), and not pushing consistency issues (if you want clustering + performance, you are probably doing this asynchronously, so you'll have consistency issues). This is one of the reasons why #5754 makes sense to me.
- If people do want state, they can also use Spring Cache -> that's why this needs to work also with Infinispan. This is another way of solving the state issue, with more control for the developer.

Last but not least: yes, please have a look at how we use Eureka+Infinispan, see [here](https://github.com/jhipster/generator-jhipster/blob/v4.5.3/generators/server/templates/src/main/java/package/config/_CacheConfiguration.java#L191). The idea is that we have the same architecture for everything - and Eureka is the single source of truth of our service discovery (that's what Zuul uses, for example), so it's important to be consistent with it. Besides, we know this works very well, as we've been doing this for a long time.

srinivasa-vasu:
@jdubois 
1. I have removed session related changes.
2. I have included registry+infinispan support. Whenever registry is defined, hosts details will be received from registry and cluster formation will be through TCP. I have also retained native discovery mechanism as well. When registry is not defined, still discovery of nodes will happen through the native mechanism based on the underlying platform. Preference will be for registry if exists, if not fallback discovery would be used for forming the cluster.

jdubois:
Looks awesome !! I'll review this as soon as possible

deepu105:
@srinivasa-vasu can you merge https://github.com/srinivasa-vasu/generator-jhipster/pull/1 into this please

generators/server/templates/src/main/java/package/config/_CacheConfiguration.java
```diff
@@ -278,4 +313,212 @@ public SessionRegistry sessionRegistry() {
...
+    public static EmbeddedCacheManager getCacheManager(){
+        return cacheManager;
+    }
+
+    public static void setCacheManager(EmbeddedCacheManager cacheManager) {
+        CacheConfiguration.cacheManager = cacheManager;
+    }
+
+    /**
+     * <p>
```
jdubois: Note to self: I will rewrite this, so we have the same style as the rest (don't worry @srinivasa-vasu this is very minor)

generators/server/templates/src/main/java/package/domain/_Authority.java
```diff
@@ -34,8 +34,10 @@
  * An authority (a security role) used by Spring Security.
  */<% if (databaseType === 'sql') { %>
 @Entity
-@Table(name = "jhi_authority")<% if (hibernateCache !== 'no') { %>
-@Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)<% } %><% } %><% if (databaseType === 'mongodb') { %>
+@Table(name = "jhi_authority")
+<%_ if (hibernateCache !== 'no') { if (hibernateCache === 'infinispan') { _%>
+@Cache(usage = CacheConcurrencyStrategy.READ_WRITE) <%_ } else { _%>
```
jdubois: Why don't use "NONSTRICT_READ_WRITE" like we usually do? I know it's a tradeoff between speed and consistency, but at least we should have the same with every option

jdubois:
This is so big: I'll try to find at least 4 hours this week to do all tests on this

jdubois:
I'm going to merge this:

- I didn't test everything, but that's still in beta
- I modified quite a lot of documentation, @srinivasa-vasu can you have a look?
- Cache statistics do not show up in the "metrics" page, @srinivasa-vasu is that normal? Can you look how it works with Ehcache? That's supposed to use JCache, so that should work.

pascalgrimaud:
I will add a new Travis build soon

jdubois:
Thanks @pascalgrimaud - Travis is broken at the moment, I don't think this is our fault, thus

srinivasa-vasu:
@jdubois: Sorry, I was at some offsite location for the past couple of days where I had little or no n/w. Will have a look at it today.

deepu105:
No worries @srinivasa-vasu there are no commitments. We all do this on our free time in the spirit of OSS so take a look when you can

jdubois:
Yes, the important things are:

- This is merged
- This is in BETA, so nobody expects this to be perfect yet (and anyway, that's OSS, so nobody can complain)

generators/server/templates/src/main/java/package/domain/_Authority.java
srinivasa-vasu: It does support NONSTRICT_READ_WRITE, but the default cache mode for L2 is 'invalidation' mode. 

Though it is similar to replication, but it differs by:
As you might know, a write operation would update the cache in the owner node and remove the stale values from the other nodes' memory (will not do an immediate replication). Benefit of this, network traffic is minimised as the invalidation messages are minimal compared to replicating the updated cache entity and the other nodes will look up the updated data in a lazy manner when needed. This mode is recommended for L2 over replicated and distributed mode and it works only with READ_WRITE. For [more info](http://planet.jboss.org/post/hibernate_second_level_cache_improvements)

srinivasa-vasu:
@jdubois : Cache statistics will not get populated, because L2 is implemented using the DefaultCacheManager, not through JCacheManager like the way ehcache has done it. All the stats will be under org.infinispan MXB, but not under javax.cache. If cache stats need to be shown, either the stats data needs to be pulled in from the registered MXB or JCacheManager needs to be defined  (just to populate the stats info but still the lifecycle will be managed by the underlying layer) .

srinivasa-vasu:
@jdubois - Added [code](https://github.com/jhipster/generator-jhipster/pull/5971) to display jcache stats

PR #5891, opened 2017-06-08
Add copyright to Apache license

mraible:
ci-skip


jdubois:
I'm sorry but this is the example text of the license, I think it was good before.

Have a look at https://github.com/apache/cassandra/blob/trunk/LICENSE.txt#L190 for example, that's supposed to stay that way.

-> I'm reverting the change as I'm pretty sure this is a mistake, and as I'll probably do a release soon, but we can discuss this, of course

PR #5892, opened 2017-06-08
add the management/* endpoints to the default swagger includePattern

PierreBesson:

screen shot 2017-06-07 at 11 02 36

@jdubois do you agree with this change ? I know it clutters the docs view but I think it serves its purpose.

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application.yml
+++ b/generators/server/templates/src/main/resources/config/_application.yml
@@ -198,7 +198,7 @@ jhipster:
     mail:
         from: <%= baseName %>@localhost
     swagger:
-        default-include-pattern: /api/.*
+        default-include-pattern: /(api|management)/.*
         title: <%=baseName%> API
         description: <%=baseName%> API documentation
         version: 0.0.1


pascalgrimaud:
As it comes from the Endpoints provided by Spring Boot Actuator, it's indeed interesting.
But I'm really not sure it is usefull during development phase: the developer uses swagger to test their own API

I would prefer having this line in comment + adding docs, so people can decide to activate it or not:
```yaml
default-include-pattern: /api/.*
# default-include-pattern: /(api|management)/.*
```

But it's just my personnal opinion :-)

PierreBesson:
I just realized that it the value should be: `default-include-pattern: (/api|${management.context-path})/.*` so that it supports changing `/management` to something else.

jdubois:
Like @pascalgrimaud I'm not sure the developers will need this, and:

- This will clutter their documentation
- This will slow down things a little (Swagger can take a long time to start - there's even a specific check for this, because for people with 100s of entities it takes too long)

I'm also not a fan of commenting stuff:

- It also clutters the code
- Sonar doesn't like it
- It's trivial to do that change anyway

xetys:
As of developer experience, in particular, to those who are not familiar with all wrapped technologies and configurations, this Swagger extension seems useful to me. Like if someone is new to Actuator and doesn't know which endpoints do exist. In that case, it's easier for the user to throw something away if he wants to optimize things, rather then add something he doesn't know.

jdubois:
I tested it and:

- Indeed I understand the interest for some people. On the other hand it makes it a lot harder for people who just want to work on their API, for example when you are doing a simple microservice.
- Performance looks OK, at least it's not measurable (should be less than 10ms)

I'm really not sure of what to do here.

PierreBesson:
@jdubois I have thought about it, maybe we could expose those endpoint on another swagger resource using Springfox Swagger Groups. I just did a quick test and it works well only by changing the jhipster lib.
screen shot 2017-06-23 at 12 26 30

However, it would require some tweaking on the zuulRewriteFilter (for microservices) but it would not break anything and this can be done later.

jdubois:
Oh, this is what groups are for??? cc @JulienMrgrd @JoffreyDcu we were struggling with that!!

Then this is used by the gateway, when we have several microservices, so that means this should be a "unique" name. And it should also work for microservices exposed through the gateway.

PierreBesson:
@jdubois I have done the work on the related JHipster lib PR.
For the gateway, I've done some tests to modify the zuul filter to take into account the management swagger resources and it does work. But then I realized that it's mostly useless to have, especially in prod if there is more than one instance of each microservice then the calls will get load balanced which we don't want.

However it could be possible to add a swagger UI on the registry so that you can access the full management API from there, I saw that you already coded something so that you can proxy calls to a specific instance.

I'm closing this here as only the jhipster lib change is needed for this.

PR #5919, opened 2017-06-13
Removed TODO

JulienMrgrd:
As it's always the good way in ng2/4.
(https://angular-2-training-book.rangle.io/handout/http/catching-rejections/catch_and_release.html)

Moreover, here there are no service or forEach that we can use.
This way is normally OK. Do you agree @xetys ? (as you are the writter of this TODO)

-------------------------------------


deepu105:
@JulienMrgrd @xetys its not a bad way so, if you guys want to make it better its fine but its good enough for now

xetys:
That TODO was more about how we check the router path, not the Observable.catch

Issue #5603, opened 2017-04-15
kubernetes improvements


xetys:
I was deploying JHipster microservice to a production like kubernetes cluster and noticed a few thing we could do better, and I want to PR that. Here are the points:

- [x] improving namespace integration, so it works better with non-default namespace
- [x] adding livenessProbe to /health, so kubernetes recognizes failing pods earlier
- [x] adding readinessProbe, so rollout kills pods on real readiness, instead of as container is running
- [x] adding JHipster Console as option, with the corresponding adjustment to env / application config
- [x] add a "service exposure selection" to let the user choose between "LoadBalancer", "NodePort" or Ingress
- [x] adding Prometheus as option (using the Prometheus operator for kubernetes)
- [x] adding README.md with first step instructions (e.g. what is printed by sub-gen after run)
- [ ] adding Zipkin as option
- [x] _adding consul (already implemented earlier)_

feel free to add points or discuss the mentioned. I will maybe need some support of @PierreBesson, especially on Zipkin, as I didn't everything, yet.

jdubois:
Do you have the Registry working well? We're doing a big work to improve it, so it's my current priority

xetys:
Hard to say. It's working, so far. There are some strange things happening like random unavailability of one replica, or the fact that currently one replica replies within 50ms, while the other is taking 5 seconds. But can't say what the cause is.

In general I have no big issues​ with the registry.

PierreBesson:
@xetys, you have basically listed the things I had in mind for improvements to our k8s subgen. I would really like to work on it but I have no time right now. If you can help that would be awesome. Once all of this is done then we can take k8s out of beta.

Also, I thought I already did consul for k8s. Did you try it ?

xetys:
to be honest, I first tried that, what's there. Currently, it is only JHipster Registry, so I started with that. I already tried Consul some time ago in a pure docker based environment. 

Once I generated these k8s configs, I started to change and tune them, by adding Prometheus operator, JHipster console, ingresses, and more. But for completeness a PR would only make sense, if I do also the stuff, I didn't try, yet. Like Consul and Zipkin.

xetys:
@jdubois I fear the registry is not really ready for kubernetes, yet. For a demo setup, there are no issues. But when something bad happens, like a full crash of one minion, the registry doesn't respawn working. I also experience different issues, when starting, while a couple of microservices are trying to connect to the registry...

for the moment, that is my short list:

[...]

PR #5927, opened 2017-06-15
Kubernetes improvements

xetys:
This PR is a collection to solve the points in #5603 


xetys:
side note 1: I added resource limits as well. These are tested in my JHI microservice application while driving replica counts too high numbers to provoke cluster failure. With these setting a had the best performance and respawn capabilities.

side note 2: using readiness probe enables the option of a zero-downtime deployment using RollingUpdate strategy, which is enabled by default for deployments. However, this really works if the deployment has at least 2 replicas. Kubernetes kills a pod to start a new one. It stops killing other pods if the newly started are not ready. 

I suggest adding a new question for "enable zero downtime deployment?", which leads to setting every deployment's replica count to 2. Another option is not to ask the user, but display a note in the generator log. I can do both. I want to user to be aware, that zero downtime is cool for production, but also needs double resources, as all deployments are running in 2 or more replicas.

PierreBesson:
Great work @xetys ! +1 for the namespace.

For the replica count let's put some info in the docs about it, no need to add another question.

pascalgrimaud:
Is it possible to add yarn test [here](https://github.com/jhipster/generator-jhipster/blob/master/test/test-kubernetes.js) ?

jdubois:
This is so big - who can do a code review here?

PierreBesson:
I will do the review this week. If it works well enough. I will consider removing the beta status for the kubernetes subgen.

sendilkumarn:
@xetys awesome work 👏 
I tried this with some basic config.
It is working awesome 👍

pascalgrimaud:
As https://github.com/jhipster/generator-jhipster/pull/6094 has been merged, @xetys can you rebase and resolve conflicts in this PR plz ?

xetys:
done

generators/kubernetes/templates/_README-KUBERNETES.md
```diff
@@ -0,0 +1,154 @@
...
+<%_ } _%>
+<%_ } _%>
+
+<%_ if (app.serviceDiscoveryType === 'eureka') { _%>
+## JHipster registry
+
+The registry is deployed using a headless service in kubernetes, so the primary service has no IP address, and cannot get a node port. You can create a secondary service for any type, using:
+
+```
+$ kubectl kubectl expose service jhipster-registry --type=NodePort --name=exposed-registry<%= kubernetesNamespace === 'default' ? '' : ` -n ${kubernetesNamespace}` %>
```
srinivasa-vasu: Pls remove the additional kubectl. It is repeated twice.

generators/kubernetes/templates/_service.yml
```diff
@@ -20,11 +20,19 @@ apiVersion: v1
 kind: Service
```
srinivasa-vasu: Do you want to move this object to _deployment.yml. If you look at other templates kafka, mysql, registry etc both the deployment and service objects in the same file. To make it consistent.

generators/kubernetes/templates/_README-KUBERNETES.md
```diff
@@ -0,0 +1,154 @@
+# JHipster generated kubernetes configuration
```
srinivasa-vasu: Do you want to add a statement that this generator needs to be run from a sub-folder to have a clean generation. If we run from the base folder it prompts for replacing the app generated README.md file. I know first time if we run from the base will miss the message, but at least from the subsequent runs it can be avoided.

generators/kubernetes/templates/_service.yml
xetys: I'd rather say split the other specs into separate files. When you change and apply the files, you sometimes don't want to apply all specs from a file, if you just update the deployment. It also stops of overwriting a service, changed by other, when updating a deployment.

Should I split the other files for consistency, too?

srinivasa-vasu: I think it is OK to have it in the same file. In a real time case, say in production if something needs to be changed, then we most likely tend to apply the change by patching it through cli or the console. I mean you wouldn't really delete a service object and recreate it in production right? For eg: if we want to change the selector or service type we more likely prefer to use kubectl cli. And moreover service object is a logical wrapper to reach to the deployed service instances. To me logically grouping these objects makes sense as it all together represent a single app entity.

deepu105: @xetys let not try to split everything now. Lets try to merge this in first following same convention we have now and later we can discuss about splitting or not splitting

generators/kubernetes/files.js
```diff
@@ -38,6 +38,37 @@ function writeFiles() {
...
+                    this.template('_ingress.yml', `${appName}/${appName}-ingress.yml`);
+                }
+                if (this.prometheusOperator) {
+                    this.template('_prometheus.yml', `${appName}/${appName}-prometheus.yml`);
+                }
+            }
+        },
+
+        writeReadme() {
+            this.template('_README-KUBERNETES.md', 'README.md');
```
PierreBesson: Never mind I'm doing this myself in another PR...

PR #5932, opened 2017-06-15
some UI fixes in Angular 4

ctamisier:
**- prevent deactivate himself, angular**
we already have this in angular 1, this PR do the same on angular 4...

**- use badge instead of tag (like in _user-management-detail)**
tag class does not work (does not belong to bootstrap 4 I think), this PR uses badge instead, like in the user page.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
@@ -58,7 +58,7 @@
                 <% if (enableTranslation) { %><td>{{user.langKey}}</td><% } %>
                 <td>
                     <div *ngFor="let authority of user.authorities">
-                        <span class="tag tag-info">{{ authority }}</span>
+                        <span class="badge badge-info">{{ authority }}</span>
                     </div>
                 </td>
                 <%_ if (databaseType !== 'cassandra') { _%>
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management.component.html
```diff
@@ -50,15 +50,15 @@
...
                 <td>{{user.email}}</td>
                 <td>
-                    <span class="badge badge-danger hand" (click)="setActive(user, true)" *ngIf="!user.activated"
-                          jhiTranslate="userManagement.deactivated">Deactivated</span>
-                    <span class="badge badge-success hand" (click)="setActive(user, false)" *ngIf="user.activated"
-                          jhiTranslate="userManagement.activated">Activated</span>
+                    <button class="btn btn-danger btn-sm" (click)="setActive(user, true)" *ngIf="!user.activated"
+                            jhiTranslate="userManagement.deactivated">Deactivated</button>
+                    <button class="btn btn-success btn-sm" (click)="setActive(user, false)" *ngIf="user.activated"
+                            [disabled]="currentAccount.login==user.login" jhiTranslate="userManagement.activated">Activated</button>
```
deepu105: Use triple equals with proper spacing

ctamisier:
use of '===' done.
Applied as well on _user-management.html of angular 1.

PR #5933, opened 2017-06-16
use navbar-inverse for angular 1

ctamisier:
use navbar-inverse for angular 1 (like in angular 4).
It actually makes the app less grey.


deepu105:
we wanted the BS4 version to look different and thats why used the inverse class. Hence we wouldnt want to do this for BS3. Its a trivial change for anybody to do themselves so sorry

ctamisier:
Ok I see.
Indeed it's an easy change but for me it gives a 'better style' (my opinion) right after having generated the app. Specially for a demo including the generation and then showing the app.
Having this black menu gives a better impression (was my case).
But anyway soon we'll all be using Angular 4 :).

PR #5959, opened 2017-06-20
Removed TODO and unnecessary code

JulienMrgrd:
Related to https://github.com/jhipster/generator-jhipster/issues/4794#issuecomment-307340461 (7th item)

Yarn tests are OK.
Tested with and without translation.

--------------------------------

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
@@ -23,8 +23,11 @@ import { Observable } from 'rxjs/Observable';
 
 export class NotificationInterceptor extends JhiHttpInterceptor {
 
+    private alertService: JhiAlertService;
+
     constructor(private injector: Injector) {
         super();
+        setTimeout(() => this.alertService = injector.get(JhiAlertService));
     }
 
     requestIntercept(options?: RequestOptionsArgs): RequestOptionsArgs {
@@ -43,7 +46,6 @@ export class NotificationInterceptor extends JhiHttpInterceptor {
                 headers.sort();
                 const alertKey = response.headers.get(headers[ 0 ]);
                 if (typeof alertKey === 'string') {
-                    const alertService: JhiAlertService = this.injector.get(JhiAlertService);
                     if (alertService) {
                         <%_ if (enableTranslation) { _%>
                         const alertParam = headers.length >= 2 ? response.headers.get(headers[ 1 ]) : null;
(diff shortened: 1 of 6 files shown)


deepu105:
@JulienMrgrd good work. you need to remove the manual code from entity dialog and entity delete dialog as well I guess

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
```diff
@@ -31,22 +32,31 @@ export class NotificationInterceptor extends JhiHttpInterceptor {
...
+            response.headers.forEach((value, name) => {
+                if (name.toLowerCase().endsWith('app-alert') || name.toLowerCase().endsWith('app-params')) {
+                    headers.push(name);
+                }
+            });
+            if (headers.length > 1) {
+                headers.sort();
+                const alertKey = response.headers.get(headers[ 0 ]);
+                if (typeof alertKey === 'string') {
+                    const alertService: JhiAlertService = this.injector.get(JhiAlertService);
```
deepu105: may be we should get the service and set it to a local variable in constructor instead of calling injector on every alert

pascalgrimaud:
Time to vote for @JulienMrgrd ?

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts
JulienMrgrd: I cannot inject directly the JhiAlertService in this component, without having an error "Cannot instantiate cyclic dependency! Http ("[ERROR ->]"): in NgModule TestInterceptorAppModule in ./TestInterceptorAppModule@-1:-1"

I've used the same way than the AuthExpiredInterceptor, using the injector.

deepu105: I'm not talking about direct injection. Use the injector but instaed of `const alertService: JhiAlertService = this.injector.get(JhiAlertService)` here do `this.alertService = this.injector.get(JhiAlertService)` in constructor

JulienMrgrd: Now it's ok :)

However, as I got a `Maximum call stack size exceeded` error, probably due to a cyclic dependency (that I couldn't find yet), I added in the constructor a call to setTimeout (following https://stackoverflow.com/questions/40525850/circular-dependency-injection-angular-2), that works perfectly. But it's maybe a workaround...

deepu105:
Any idea what was causing the circular dependency?

PR #5961, opened 2017-06-20
Farsi translation & RTL support

SidMorad:
According to Wikipedia(https://en.wikipedia.org/wiki/Persian_language)
There are approximately 110 million Farsi speakers worldwide, with the language
holding official status in Iran, Afghanistan, and Tajikistan.

Note: Farsi is a Right to Left (RTL) language, therefore we need proper CSS for displaying text correctly in html pages.
and here is `rtl.css` example that needs to be included(dynamically) if current language is a RTL language:
```
html {
  direction: rtl;
}

.col-xs-1, .col-xs-2, .col-xs-3, .col-xs-4, .col-xs-5, .col-xs-6, .col-xs-7, .col-xs-8, .col-xs-9, .col-xs-10, .col-xs-11, .col-xs-12 {
  float: right;
}

.modal-footer {
  text-align: left;
}

.dropdown-menu {
  text-align: right;
}

th {
  text-align: right;
}

@media (min-width: 768px) {
  .dl-horizontal dt {
    float:right;
  }
}
```
and here is a condition example in `index.html`
`    <link rel="stylesheet" ng-href="{{langKey === 'fa' ? 'content/css/rtl.css?v=2' : ''}}">`
 

This commit assumes you will take care of HTML/RTL issues yourself, and
will show translations in correct format when text direction is set to rtl.

Please let me know if something is missing? or any improvement needed? 

Thanks!


deepu105:
Thanks for the translation @SidMorad
This is our first RTL language and hence I think we need make sure it works well like other languages. Let me take a look and think about a solution

jdubois:
Awesome !!!! That's really an important feature.

SidMorad:
I would like to point out, Ionic-team done really good job recently(as of version 3.4.2) regards to [RTL support](http://ionicframework.com/docs/theming/rtl-support/). It seems their solution is heavily depends to using SASS, but you may find some useful information at [ionic-team/ionic/issues/11211](https://github.com/ionic-team/ionic/issues/11211).

@deepu105 I look forward to see your solution, in time between, I think adding RTL support and adding Frasi translation better to be two independent pull requests, don't you think? plus suggested solution must work with ease for next RTL language too, which means no dependency between specific language and RTL support.

@jdubois Thanks for saying supporting RTL is an important feature. As a RTL developer, I get used to the fact that supporting RTL is not a high priority task! ؛-)

Thanks for creating JHipster! JHipsters!

deepu105:
@SidMorad since you already have a solution and it looks good could you make a PR with it. We need to support it in botg ng1 and ng2± first before we support RTL languages

SidMorad:
@deepu105 My solution is pretty simple and is kind of hard-coded, let's look at this line:

`<link rel="stylesheet" ng-href="{{ langKey === 'fa' ? 'content/css/rtl.css?v=2' : '' }}">`

And for supporting next RTL language, let's say is Hebrew language, then we need to change it to:

`<link rel="stylesheet" ng-href="{{ langKey === 'fa' || langKey === 'he' ? 'content/css/rtl.css?v=2' : '' }}">`

Plus If a RTL language is not between selected languages, then (I assume) majority of developers don't expect/want to see such a line in their `index.html` or a `rtl.css` file in their `content/css` directory. am I wrong?

So I suggest someone who is more familiar with code base, work on this, and provides better/clever way to support RTL when is necessary.

deepu105:
@SidMorad that's easy to solve in the generator base we need to add another attribute `rtl: true` to the language object with RTL and then use that in the language constant in the generated code as well that way it's easier to implement this and we need to add the line only when there is at least 1 RTL language selected. You give it a shot if you are struck ill help you

SidMorad:
@deepu105 thanks for your support, sure I will give a shot this weekend. but no promise for contribute in next week, because we are suppose to release version 1 of our product by end of next week.

Note: In Iran Thursday and Fridays are weekend days! plus we use Persian calendar, another source of troubles for us as local developers! :-)  sometimes I think of working in another country and tada! all problems solved!: no deal with RTL support, different calendar, unfriendly government!, etc! :-)

SidMorad:
@deepu105 Regards to RTL support, I went to slightly different path that we talk about initially. please have a look and let me know what do you think? thanks.

Note: I did test 4 different scenarios:
 - Monolithic app + ng1 + en, fa
 - Monolithic app + ng1 +  en, de
 - Monolithic app + ng1 + sass + en, fa 
 - Monolithic app + ng1 + sass + en, de

SidMorad:
JHipster + Farsi Language + RTL support _screenshot_:

screenshot from 2017-06-22 16-43-48

deepu105:
Thanks ill take a look

generators/client/templates/angularjs/src/main/webapp/app/blocks/handlers/_translation.handler.js
```diff
@@ -44,6 +47,24 @@
...
+      <%_ if (enableRTLSupport) { _%>
+        // Update language and direction in index.html, e.g. <html dir="ltr" or <html dir="rtl"
+        function updatePageDirection() {
+          var currentLang = $translate.proposedLanguage() || $translate.use();
+          angular.element('html').attr('lang', currentLang);
+          angular.element('html').attr('dir', isRTL(currentLang) ? 'rtl' : 'ltr');
+        }
+
+        // Returns true if passed language key is a Right-to-Left language key
+        function isRTL(langKey) {
```
deepu105: I would like to get this from the existing language filter. See my comment below

generators/client/templates/angularjs/src/main/webapp/app/components/language/_language.filter.js
```diff
@@ -36,6 +36,7 @@
                 'en': 'English',
                 'es': 'Español',
                 'et': 'Eesti',
+                'fa': 'فارسی',
```
deepu105: It would be better to change the pipe to below and then use it in the above service. 
```
(function() {
     'use strict';
 
     angular
         .filter('findLanguageFromKey', findLanguageFromKey);
         .filter('findLanguageRtlFromKey', findLanguageRtlFromKey);
     var languages = {
        'ca': { name: 'Català' },
        ....
        'fa': { name: 'فارسی', rtl: true },
        ...
        'zh-tw': { name: '繁體中文' }
     }
     function findLanguageFromKey() {
         return findLanguageFromKeyFilter;
 
         function findLanguageFromKeyFilter(lang) {
             return languages[lang].name;
         }
     }
     function findLanguageRtlFromKey() {
         return findLanguageRtlFromKeyFilter;
 
         function findLanguageRtlFromKeyFilter(lang) {
             return languages[lang].rtl;
         }
     }
 })();
```

generators/generator-base.js
```diff
@@ -437,6 +452,7 @@ module.exports = class extends Generator {
             { name: 'Dutch', value: 'nl' },
             { name: 'English', value: 'en' },
             { name: 'Estonian', value: 'et' },
+            { name: 'Farsi', value: 'fa' },
```
deepu105: add a new attribute `{ name: 'Farsi', value: 'fa', rtl: true },` and use that in `isRTLSupportNecessary` instead of hardcoding the values

SidMorad:
@deepu105 Thanks for your solutions, I did apply your requested changes for AngularJs(1.x) module. Please let me know if something is missing?

For Angular(4.x) module I did nothing, and as I said earlier I am kind of busy in this week, so I can be just a minor helper if someone else want to continue to work on this, thanks.

deepu105:
Ok then, ill merge this and add angular support

SidMorad:
@deepu105 I notice my mind is with this issue, while I am working on other issues, so I did add some progress for angular(4.x) module(which they pass initial tests). do you want me to push them too?

deepu105:
If you have done yes please

SidMorad:
I did test following scenarios as of revision [08dfb9a](https://github.com/jhipster/generator-jhipster/pull/5961/commits/08dfb9a61be0ec9a5902b0123cc9657700aaf05b):
- Monolithic app + ng2 + en, fa
- Monolithic app + ng2 + en, de
- Monolithic app + ng2 + no i18n

Please let me know if something is missing? or can be done better? thanks.

deepu105:
Thanks, i'm gonna do a little polish and merge this


Thanks & Regards,
Deepu

deepu105:
merged manually

SidMorad:
I start using JHipster v4.6.0 and it's RTL support in one of my projects, and I notice we can have some improvements regards to RTL support. and here I'am going to document it for future developers who may need this.

First I would like to point out, according to [this comment](https://github.com/twbs/bootstrap/issues/17279#issuecomment-181151047), bootstrap project may add RTL support officially as of version 4.2.0 or later. So if bootstrap 4.2.0 is out and has RTL support, then please ignore this comment, and follow bootstrap guide in this regards.

Bootstrap 4.0.0-alpha.6 is used by JHipster v4.6.0 and our primary technique for adding RTL support was watching `<html dir="rtl">` attribute and apply necessary CSS depends to it's value. for example:
```
html[dir=rtl] th {
    text-align: right;
}
```
[bright/bootstrap-rtl](https://github.com/bright/bootstrap-rtl) project is using the same technique(with supporting almost all differences between LTR & RTL). and as I am writing this, it is compatible with bootstrap 3.x version. so I think if you are looking for a complete RTL support, then following file can be your guide:
[https://github.com/bright/bootstrap-rtl/blob/master/dist/css/bootstrap-rtl-ondemand.css](https://github.com/bright/bootstrap-rtl/blob/master/dist/css/bootstrap-rtl-ondemand.css)

and in my opinion upgrading above file to bootstrap 4.x is not that much complex, you only need to search for differences and replace them with updated syntax. for example search for 
```
.col-md-offset-
```
and replace it with
```
.offset-md-
```
Have fun RTLing! ؛-)

p.s. If you find a better way for adding RTL support, please let me know, thanks.

PR #5964, opened 2017-06-21
set login required

ctamisier:
The login of a user should be required in the html template.
(like in Angular 4)

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management-dialog.html
+++ b/generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management-dialog.html
@@ -35,7 +35,7 @@
         <div class="form-group">
             <label class="control-label" data-translate="userManagement.login">Login</label>
             <input type="text" class="form-control" name="login"
-                   ng-model="vm.user.login" ng-required="vm.user.id == null" ng-minlength=1 ng-maxlength=50 ng-pattern="/^[_'.@A-Za-z0-9-]*$/">
+                   ng-model="vm.user.login" required ng-minlength=1 ng-maxlength=50 ng-pattern="/^[_'.@A-Za-z0-9-]*$/">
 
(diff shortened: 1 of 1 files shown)


generators/client/templates/angularjs/src/main/webapp/app/admin/user-management/_user-management-dialog.html
```diff
@@ -35,7 +35,7 @@ <h4 class="modal-title" id="myUserLabel" data-translate="userManagement.home.cre
         <div class="form-group">
             <label class="control-label" data-translate="userManagement.login">Login</label>
             <input type="text" class="form-control" name="login"
-                   ng-model="vm.user.login" ng-required="vm.user.id == null" ng-minlength=1 ng-maxlength=50 ng-pattern="/^[_'.@A-Za-z0-9-]*$/">
```
deepu105: I think it was implemented this way for a specific reason may be something to do with social login, so I wont change it without checking why

ctamisier: I investigated a bit and it was like this in angular 2 as well until commit 81e27a4478a1d8cdfb0830bb252d6bb1ab3bfb66 that changed it to `required`...

deepu105: @ruddell seems like you did the change, any idea why?

ruddell: That change is when I converted User Management from NG1 -> NG2.  I can't remember a specific reason, but the login is always required and annotated as `@NotNull` in User.java so it makes sense

deepu105: ok

PR #5966, opened 2017-06-21
WIP: Fix automatic upgrade

### LINKED ISSUE #5883: "jhipster upgrade" not working properly
As we have basically 4 duplicate tickets on the JHipster upgrade sub-generator, I'm closing #5881 #5869 #5811 #5649 in favor of this new "meta ticket".

It's basically all the same issue: on some systems, the upgrade sub-generator cannot update the JHipster version. Typically, this is because it doesn't have the right to do. Then, there are various other issues depending on how it was installed.

This is why this failing for a lot of people, with different error messages depending on their own configuration and how they installed JHipster. And we can't always reproduce as it depends on our own setup.

This was mostly triggered by the use of Yarn: this adds one more installation mechanism, and it depends on how Yarn was installed.

There is also the issue of people upgrading their JHipster version globally, and then the sub-generator gets stuck with that global version.

-> I'm not sure this is true for everyone, so here is how to test this: say you are using JHipster 4.5.1, do a "jhipster upgrade". You will have a commit telling the application was re-generated with JHipster 4.5.1 (in the commit message). Have a look in the commit, if it's in fact generated with JHipster 4.5.2 (you can see it in `package.json`), then you have the issue of a global JHipster installation that you can't modify

This is so complex, and doing an upgrade isn't such a big deal to do manually, that I'm considering removing this sub-generator (cc @lordlothar99 as you coded this originally).

We just don't have the time to work on this -> would someone be interested in taking this ticket?

tientq:
Fix #5883

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -32,6 +32,7 @@ util.inherits(UpgradeGenerator, BaseGenerator);
 const GENERATOR_JHIPSTER = 'generator-jhipster';
 const UPGRADE_BRANCH = 'jhipster_upgrade';
 const GIT_VERSION_NOT_ALLOW_MERGE_UNRELATED_HISTORIES = '2.9.0';
+const GENERATOR_JHIPSTER_CLI_VERSION = "4.5.1"
 
 module.exports = UpgradeGenerator.extend({
     constructor: function (...args) { // eslint-disable-line object-shorthand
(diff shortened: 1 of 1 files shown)


tientq:
Tested on:
[x] Upgrade from 4.5.2 to 4.5.5
[x] Upgrade from 4.4.0 to 4.5.5
[ ] Upgrade project that has conflict in `package.json`

jdubois:
That issue is the biggest one we had in the recent months, and nobody could fix yet (including me!). If you solved this, many people will thank you!!!

jdubois:
@tientq have a look at https://travis-ci.org/jhipster/generator-jhipster/jobs/245617623 you have some lint issues

tientq:
Yes, thank for your remind.

generators/upgrade/index.js
```diff
@@ -32,6 +32,7 @@ util.inherits(UpgradeGenerator, BaseGenerator);
 const GENERATOR_JHIPSTER = 'generator-jhipster';
 const UPGRADE_BRANCH = 'jhipster_upgrade';
 const GIT_VERSION_NOT_ALLOW_MERGE_UNRELATED_HISTORIES = '2.9.0';
+const GENERATOR_JHIPSTER_CLI_VERSION = '4.5.1';
```
jdubois: this is the current JHipster version, it's already in `package.json` and should be available (I'll have a look)

jdubois: Oh sorry, I didn't understood correctly, this is good

jdubois:
I just tried it:

- It worked!!!! Yeah!!!!!!!!
- Only issue is that my `yarn.lock` wasn't commited, but that's a very minor point, and it would be easy to solve
- I didn't test any edge cases, just did from v4.5.2 to the current v4.5.5

So I'm merging this: whatever problems we might encounter, it's still much better than the current code.

THANK YOU SO MUCH @tientq !!!!!!!!

tientq:
You are welcome.

PR #5967, opened 2017-06-21
There should be no </p> in javadoc

henri-tremblay:


jdubois:
Thanks, and I did put some in other places, I need to check.

PR #5971, opened 2017-06-22
initial commit: jcache stats

srinivasa-vasu:


generators/generator-base.js
```diff
@@ -713,15 +713,16 @@ module.exports = class extends Generator {
      * @param {array} relationships - the relationships of this entity
      * @param {string} packageName - the Java package name
      * @param {string} packageFolder - the Java package folder
+     * @param {string} cacheProvider - the cache provider
      */
-    addEntityToEhcache(entityClass, relationships, packageName, packageFolder) {
+    addEntityToCache(entityClass, relationships, packageName, packageFolder, cacheProvider) {
```
deepu105: Can we create a passthrough alias method with the old name. As this is part of the public module API it would break plugins using this method

deepu105: Also for the new param, set ehcache as the default value in argument

pascalgrimaud: very good catch about not breaking module API

srinivasa-vasu: done.

PR #5974, opened 2017-06-23
use the real fs when reading entity JSON files

gimoh:
Normally `this.fs` is yeoman's in-memory filesystem (mem-fs) so once
file is read from disk it's not re-read.

Using jhipster-core function `readEntityJSON` means we always read from
disk and have most up to date and consistent state.

This came up when implementing a change in `import-jdl` generator where
entities needed to be re-read after they've been changed by the import.

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -29,6 +29,7 @@ const exec = require('child_process').exec;
 const os = require('os');
 const https = require('https');
 const pluralize = require('pluralize');
+const jhiCore = require('jhipster-core');
 const packagejs = require('../package.json');
 const jhipsterUtils = require('./util');
 const constants = require('./generator-constants');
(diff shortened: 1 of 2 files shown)


gimoh:
BTW, I couldn't find any unit tests for the generator-base, are there any?

deepu105:
@gimoh yes they are under tests directory

deepu105:
filename `test-generator-base.js`

gimoh:
Hmm, travis is failing randomly, I'm going to rebase and try again.

generators/generator-base.js
```diff
@@ -1438,14 +1439,21 @@ module.exports = class extends Generator {
...
 
-        return entities.sort(isBefore);
+        return shelljs.ls(
+            path.join(JHIPSTER_CONFIG_DIR, '*.json')
+        ).reduce((acc, file) => {
+            try {
+                const definition = jhiCore.readEntityJSON(file);
+                acc.push({ name: path.basename(file, '.json'), definition });
+            } catch (error) {
+                // not an entity file / malformed?
```
jdubois: shouldn't we at least log the error?

deepu105: We could but its shouldn't matter much as its a read-only operation

deepu105: any way ill add a warning log

gimoh: this is based on [jhipster-core lib/reader/json_reader.js#51](https://github.com/jhipster/jhipster-core/blob/master/lib/reader/json_reader.js#L51) which doesn't log at the moment so I assumed there could be other files / directories there that could trigger it too frequently

PR #5977, opened 2017-06-23
fixes registry  issues on k8s and oc

### LINKED ISSUE #5975: jhipster-registry beyond 2.5.8 not working on k8s

After several updates to JHipster registry and the kubernetes sub-generator, the recent versions of the registry don't work properly. I had the first contact with that issue with version 2.6.0 and mentioned it in #5603. The cluster setup fails, and services effectively are registering to 2 distinct registries.

I want to use the cool brand new registry in kubernetes, too!

##### **Reproduce the error**

Just generate microservices and a corresponding kubernetes setup and run it against a recent version of kubernetes (1.6.4 in my case). The "replicas" page will every time show an empty array and listing localhost, while reloading of applications will show 2 different zones.

##### **Suggest a Fix**

I already have an (upcoming) fix for that in #5927, which has a working configuration, tested on a self-hosted CoreOS kubernetes cluster.

The first is that self-registering Eureka to itself when clustering, is a bad idea. It was wrong configured, until https://github.com/jhipster/generator-jhipster/commit/23d746696547ab9478bc8b5c0483959f1975b3ac. After the fix, the configuration was correct and the misbehavior started. The same commit provides a correction for `EUREKA_CLIENT_SERVICEURL_DEFAULTZONE` to `EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE`, which should be the correct convention for that attribute. As far as I can say, there still is that issue that the 'wrong' version of the variables works, while to correct does not. This is crucial for Eureka to connect in a cluster.

Another point is, that the current version of JHipster Registry doesn't preloads the `jhipster.registry.password` attribute from config server, while connection to the other nodes, so clustering fails, even with the correct variable. Passing JHIPSTER_REGISTRY_PASSWORD in env file solves that trivially.

##### **JHipster Version(s)**

current upstream and JHipster Registry 2.6.0 and later

xetys:
fix #5975 

was tested on my bigger cluster several seconds after 3.0.3 was available on hub.docker.com :smile: 

### DIFF
--- a/generators/kubernetes/templates/registry/_jhipster-registry.yml
+++ b/generators/kubernetes/templates/registry/_jhipster-registry.yml
@@ -84,13 +84,14 @@ spec:
         # Registry configuration
         - name: SPRING_PROFILES_ACTIVE
           value: prod,native
-        - name: SECURITY_USER_PASSWORD
+        - name: JHIPSTER_REGISTRY_PASSWORD
           valueFrom:
             secretKeyRef:
               name: registry-secret
               key: registry-admin-password
         - name: JHIPSTER_SECURITY_AUTHENTICATION_JWT_SECRET
           value: my-secret-token-to-change-in-production
+        # Eureka specific tweaking
         - name: EUREKA_INSTANCE_LEASE_RENEWAL_INTERVAL_IN_SECONDS
           value: '15'
         - name: EUREKA_INSTANCE_LEASE_EXPIRATION_DURATION_IN_SECONDS
@@ -109,12 +110,14 @@ spec:
           value: '2000'
         - name: EUREKA_CLIENT_FETCH_REGISTRY
           value: 'true'
-        - name: EUREKA_CLIENT_REGISTER_WITH_EUREKA
-          value: 'true'
         - name: EUREKA_CLIENT_REGISTRY_FETCH_INTERVAL_SECONDS
           value: '15'
         - name: EUREKA_CLIENT_INITIAL_INSTANE_INFO_REPLICATION_INTERVAL_SECONDS
           value: '15'
+        - name: EUREKA_SERVER_REGISTRY_SYNC_RETRY_WAIT_MS
+          value: '500'
+        - name: EUREKA_SERVER_A_SGCACHE_EXPIRY_TIMEOUT_MS
+          value: '60000'
         # Example "prod" configuration
         # - name: SPRING_PROFILES_ACTIVE
         #   value: prod
@@ -134,16 +137,13 @@ spec:
             LAST_POD_INDEX=$((${CLUSTER_SIZE} - 1))
             REPLICAS=""
             for i in $(seq 0 $LAST_POD_INDEX); do
-              if [ ${STATEFULSET_NAME}-${i} != ${HOSTNAME} ]; then
-                if [ -z ${REPLICAS} ]; then
-                  REPLICAS="http://admin:${SECURITY_USER_PASSWORD}@${STATEFULSET_NAME}-${i}.jhipster-registry.${STATEFULSET_NAMESPACE}.svc.cluster.local:8761/eureka/"
-                else
-                  REPLICAS="${REPLICAS},http://admin:${SECURITY_USER_PASSWORD}@${STATEFULSET_NAME}-${i}.jhipster-registry.${STATEFULSET_NAMESPACE}.svc.cluster.local:8761/eureka/"
-                fi
+              REPLICAS="${REPLICAS}http://admin:${JHIPSTER_REGISTRY_PASSWORD}@${STATEFULSET_NAME}-${i}.jhipster-registry.${STATEFULSET_NAMESPACE}.svc.cluster.local:8761/eureka/"
+              if [ $i -lt $LAST_POD_INDEX ]; then
+                REPLICAS="${REPLICAS},"
               fi
             done
-            export EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=$REPLICAS
-            echo "EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=${REPLICAS}"
+            export EUREKA_CLIENT_SERVICEURL_DEFAULTZONE=$REPLICAS
+            echo "EUREKA_CLIENT_SERVICEURL_DEFAULTZONE=${REPLICAS}"
 
             java -jar /jhipster-registry.war --spring.cloud.config.server.git.uri=${GIT_URI} --spring.cloud.config.server.git.search-paths=${GIT_SEARCH_PATHS} -Djava.security.egd=file:/dev/./urandom
         volumeMounts:
--- a/generators/kubernetes/templates/registry/_jhipster-registry.yml
+++ b/generators/kubernetes/templates/registry/_jhipster-registry.yml
@@ -91,21 +91,30 @@ spec:
               key: registry-admin-password
         - name: JHIPSTER_SECURITY_AUTHENTICATION_JWT_SECRET
           value: my-secret-token-to-change-in-production
+        - name: EUREKA_INSTANCE_LEASE_RENEWAL_INTERVAL_IN_SECONDS
+          value: '15'
+        - name: EUREKA_INSTANCE_LEASE_EXPIRATION_DURATION_IN_SECONDS
+          value: '60'
+        - name: EUREKA_SERVER_PEER_EUREKA_NODES_UPDATE_INTERVAL_MS
+          value: '15000'
+        - name: EUREKA_SERVER_RENAWAL_THRESHOLD_UPDATE_INTERVAL_MS
+          value: '15000'
+        - name: EUREKA_SERVER_REGISTRY_SYNC_RETRIES
+          value: '3'
+        - name: EUREKA_SERVER_EVICTION_INTERVAL_TIMER_IN_MS
+          value: '30000'
+        - name: EUREKA_SERVER_ENABLE_SELF_PRESERVATION
+          value: 'false'
+        - name: EUREKA_SERVER_PEER_NODE_CONNECT_TIMEOUT_MS
+          value: '2000'
         - name: EUREKA_CLIENT_FETCH_REGISTRY
           value: 'true'
         - name: EUREKA_CLIENT_REGISTER_WITH_EUREKA
           value: 'true'
-        # Eureka specific tweaking
-        - name: EUREKA_SERVER_REGISTRY_SYNC_RETRY_WAIT_MS
-          value: '500'
-        - name: EUREKA_SERVER_A_SGCACHE_EXPIRY_TIMEOUT_MS
-          value: '60000'
-        - name: EUREKA_SERVER_EVICTION_INTERVAL_TIMER_IN_MS
-          value: '30000'
-        - name: EUREKA_SERVER_PEER_EUREKA_NODES_UPDATE_INTERVAL_MS
-          value: '30000'
-        - name: EUREKA_SERVER_RENAWAL_THRESHOLD_UPDATE_INTERVAL_MS
-          value: '15000'
+        - name: EUREKA_CLIENT_REGISTRY_FETCH_INTERVAL_SECONDS
+          value: '15'
+        - name: EUREKA_CLIENT_INITIAL_INSTANE_INFO_REPLICATION_INTERVAL_SECONDS
+          value: '15'
         # Example "prod" configuration
         # - name: SPRING_PROFILES_ACTIVE
         #   value: prod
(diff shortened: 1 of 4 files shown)


generators/kubernetes/templates/registry/_jhipster-registry.yml
```diff
@@ -84,17 +84,13 @@ spec:
...
           valueFrom:
             secretKeyRef:
               name: registry-secret
               key: registry-admin-password
         - name: JHIPSTER_SECURITY_AUTHENTICATION_JWT_SECRET
           value: my-secret-token-to-change-in-production
-        - name: EUREKA_CLIENT_FETCH_REGISTRY
-          value: 'true'
-        - name: EUREKA_CLIENT_REGISTER_WITH_EUREKA
-          value: 'true'
```
srinivasa-vasu: I think these properties are required and need to be set 'true'. This is required to avoid DiscoveryComposite - eureka 'Unknown' in the console. Though you have defined the individual instances to be discovered in the 'command' tag, without these properties set you will not get the status 'UP' in the registry console. I had the same issue when I tested it last time in K8s couple of months ago. Did you manage to check the console for status after removing these properties?
[More info](https://github.com/jhipster/jhipster-registry/issues/69) can be found in this thread

generators/kubernetes/templates/registry/_jhipster-registry.yml
```diff
@@ -125,13 +121,13 @@ spec:
...
-              REPLICAS="${REPLICAS}http://admin:${SECURITY_USER_PASSWORD}@${STATEFULSET_NAME}-${i}.jhipster-registry.${STATEFULSET_NAMESPACE}.svc.cluster.local:8761/eureka/"
+              REPLICAS="${REPLICAS}http://admin:${JHIPSTER_REGISTRY_PASSWORD}@${STATEFULSET_NAME}-${i}.jhipster-registry.${STATEFULSET_NAMESPACE}.svc.cluster.local:8761/eureka/"
               if [ $i -lt $LAST_POD_INDEX ]; then
                 REPLICAS="${REPLICAS},"
               fi
             done
-            export EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=$REPLICAS
-            echo "EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=${REPLICAS}"
+            export EUREKA_CLIENT_SERVICEURL_DEFAULTZONE=$REPLICAS
+            echo "EUREKA_CLIENT_SERVICEURL_DEFAULTZONE=${REPLICAS}"
```
srinivasa-vasu: Again I had the same issue with EUREKA_CLIENT_SERVICEURL_DEFAULTZONE that the instances didn't discover them properly. I managed to get it working after changing it to EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE. [More info](https://github.com/jhipster/generator-jhipster/pull/5657)

Also, we can remove this initialisation altogether from the 'command' tag. Anyway we have defined the instances to be discovered in 'application-config' configMap. Ideally it should work without this as the client service-url will be picked up from configMap. I tested this in OpenShift by removing the entire initialisation code and it worked without any issue. Pls check that out. Let me also try to test it in K8s.

generators/kubernetes/templates/registry/_jhipster-registry.yml
xetys: When keeping this fields, the registry registers itself as a discoverable service, which IMHO is not the intention of Eureka. Tested with 3.0.1 - 3.0.3

generators/kubernetes/templates/registry/_jhipster-registry.yml
xetys: I didn't build that initialisation block but tried to figure out what was causing my Registries failure  past my last time of successful usage of 2.5.8

If you have a better replica initialization, let me know, please. 

I was testing the current state with deploying up to 50 spring boot microservices under heavy conditions to provoke failure and pick the configuration with the best behavior.

I will check one thing and report again

generators/kubernetes/templates/registry/_jhipster-registry.yml
xetys: when resetting that, the UNKNOWN label turns to UP, but the registry is listed as a service.

the clustering itself work in both cases....how do we proceed here, then? @jdubois @PierreBesson

generators/kubernetes/templates/registry/_jhipster-registry.yml
xetys: ok, after a LOT of struggling hours, suddenly that worked for me in 3.0.3 with EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE.

I also did a local test by rebuilding that init block to just get the next peer address, rather then a list of all peers. But I have no idea how to get rid completely of it, as eureka needs to know at least the next peer to start in peer-aware mode.

However, I will rethink this PR after some tests.....Eureka is one of my greatest microservice time killer :tired_face:

generators/kubernetes/templates/registry/_jhipster-registry.yml
srinivasa-vasu: Isn't it the default in Peer-aware mode? In Peer-aware mode, a Eureka server also acts as a Eureka client and registers to a peer on a given URL. EUREKA_CLIENT_REGISTER_WITH_EUREKA needs to be true in this case. In standalone mode, this can be set to false.

xetys: I am not sure here. The [official documentation](http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html) doesn't say this is needed. 
Furthermore, you definitely see it is not running in peer-aware when reloading Eureka gives distinct view to 2 registries, rather than a consistent list of instances. In contraposition, if you reload instances and always get the same list, it means the Eureka instances are replicated properly.

So the only thing "register with Eureka" is doing, is the self-registration, which is IMHO not correct.

generators/kubernetes/templates/registry/_jhipster-registry.yml
xetys: I will revert these changes, as they are indeed working with this convention. I head an issue with headless services while I was testing, and thought this was the issue here.

PierreBesson: @srinivasa-vasu The reason I added the initialization script at the time was that with it you can scale the registry without downtime:
1) increase your stateful set replica count by one -> this create a new replica with the correct cluster size
2) delete the old replicas one by one so they can find the new instance
3) finally update the config map so that your applications will know about the new replica.

I thought this way you can "scale" eureka like your other apps even though I agree Eureka is complex and things might not wor well all the time.

generators/kubernetes/templates/registry/_jhipster-registry.yml
srinivasa-vasu: I referred [this](http://blog.abhijitsarkar.org/technical/netflix-eureka/). I have changed couple of things in this one, will give a pull request with my changes. I can confirm that peer-aware mode is working fine with the changes. I have done multiple rounds of testing in OpenShift with plenty of combinations that I could think of. Will test K8s tonight. If possible, you may pls try out various combinations with the changes as well.

But by the way, sync between the instances will take few seconds as it is determined by the parameter EUREKA_SERVER_PEER_EUREKA_NODES_UPDATE_INTERVAL_MS. Generally to avoid too many updates, it will be kept somewhere between 15 and 45s.

generators/kubernetes/templates/registry/_jhipster-registry.yml
srinivasa-vasu: Yeah, I understand that. More than 2 instances actually makes it complicated as it involves few manual steps and also takes few seconds for all the instances to be in sync. At this point of time, there is no way to directly refer 'replicas' field in the template to enable auto configuration.

srinivasa-vasu: @xetys : I have pushed [this](https://github.com/jhipster/generator-jhipster/pull/5998). I have tested in both K8s and OC. It looks good. If possible, pls do test with this version.

jdubois: I just had a look, it's indeed `EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE`, so with the "_" as the name of the property is "serviceUrl".
What really misleading is `defaultZone` which becomes `DEFAULTZONE` (with no `_`) but that's inside a map of properties, so I guess it doesn't work the same as a real property

jdubois:
I'm a bit lost here, is this ready to be merged?

srinivasa-vasu:
@jdubois Not yet. I have tested this, but yet to get the confirmation from @xetys

xetys:
@jdubois we are having 2 PRs for one problem. The PR of @srinivasa-vasu is more explicitly configuring Eureka for operating inside OpenShift and k8s, but we still have problems to scale the StatefulSet after deployment without running in inconsistency.

deepu105:
May be we can use best of both PR?


Thanks & Regards,
Deepu

xetys:
So I did @deepu105 . I rebased @srinivasa-vasu 's branch into mine, so his work is now in this PR. I tested that configuration with my cluster and scaled up to more replicas, which worked without any restarts and nearly without any inconsistency. With `fetchRegistry` the new replicas immediately fetch the current state and does not loosing the leases, as they do with `register-with-eureka`

I suggest merging this PR, which includes #5998 to merge the best of the 2 states. Feel free to check that state here

deepu105:
@srinivasa-vasu are you ok with this? can we close the other PR then?

deepu105:
@srinivasa-vasu @xetys you guys know this way better than any of us, so if both of you agree please go ahead and merge it :)

generators/kubernetes/templates/registry/_jhipster-registry.yml
```diff
@@ -125,13 +137,13 @@ spec:
...
```
deepu105: @xetys seems like its `EUREKA_CLIENT_SERVICEURL_DEFAULTZONE` without `_` again. I'm confused now which is the correct one

xetys: Thanks, this got wrong by rebasing. I'll correct that asap!

srinivasa-vasu: @xetys - Yeah, it needs to be changed. Since you have the same info in application.xml, it takes preference and hence it figures out the replicas. (if you remove the service-url entry from application.yml, then this would fail).

I am testing with this branch (xetys/fix-eureka-k8s), but still I see issues when you increase the replica count from the default one(2).  Try running with registerWithEureka 'true', you would be able to see the inconsistency.

This is what I noticed in the logs abt the replicas,
**Instance 1 and 2:**

> Individual node >>>>com.netflix.eureka.cluster.PeerEurekaNode@472d7c37
> Individual node >>>>http://admin:admin@jhipster-registry-0.jhipster-registry.default.svc.cluster.local:8761/eureka/
> Individual node >>>>com.netflix.eureka.cluster.PeerEurekaNode@d41a9356
> Individual node >>>>http://admin:admin@jhipster-registry-1.jhipster-registry.default.svc.cluster.local:8761/eureka/

**Instance 3:**

> Individual node >>>>com.netflix.eureka.cluster.PeerEurekaNode@6107aa75
> Individual node >>>>http://admin:admin@jhipster-registry-2.jhipster-registry.default.svc.cluster.local:8761/eureka/
> Individual node >>>>com.netflix.eureka.cluster.PeerEurekaNode@472d7c37
> Individual node >>>>http://admin:admin@jhipster-registry-0.jhipster-registry.default.svc.cluster.local:8761/eureka/
> Individual node >>>>com.netflix.eureka.cluster.PeerEurekaNode@d41a9356
> Individual node >>>>http://admin:admin@jhipster-registry-1.jhipster-registry.default.svc.cluster.local:8761/eureka/

After we increase the replica count, instance 1 and 2 are not detecting the new instances as replicas. Changing details in application.yml will get reflected only in new instances, but not in already running ones. 
Though the configuration `(org.springframework.cloud.netflix.eureka.EurekaClientAutoConfiguration$RefreshableEurekaClientConfiguration)` uses  `@RefreshScope`, it needs to be triggered after the property change for the existing instances to get refreshed. 

I am still not convinced with this. Can we just keep it open for 1 or 2 days? I am thinking of writing something in jhipster-registry to make this configMap refresh happens automatically.

PierreBesson: @srinivasa-vasu About spring cloud config refreshes I have wanted to add a `refresh config` button on the registry for a long time. I don't know if it's a good idea to make the refresh happen automatically, it will look like dark magic and be annoying to troubleshoot.

xetys: @srinivasa-vasu I'm not in hurry with this. As JHipster registry is our default solution it's crucial for me it works like a charm in kubernetes. 

This situation is very strange, as it really seems to fail under opposite conditions. For my setup adding `registerWithEurka` lead to new replicas loose all leases. Maybe something different is causing that. We should find that.

srinivasa-vasu: @PierreBesson @xetys - Have created [this class](https://github.com/jhipster/jhipster-registry/pull/163) to detect configMap changes to auto refresh @Refreshed scope beans. This would refresh peer url as eurekaClient is @Refresh scoped. With this, we can remove the stateful set initialisation block and most of the manual steps can be avoided. Only manual step that is required is, whenever we scale the registry corresponding url needs to be updated in jhipster-registry.xml configMap file.

jdubois:
@xetys @srinivasa-vasu you are both the experts here: should we merge this PR and close #5998 ?

srinivasa-vasu:
@jdubois : Actually we can close this and merge #5998. #5998 is the updated one and it is dependent on https://github.com/jhipster/jhipster-registry/pull/163

deepu105:
closing as the changes here are present on the other PR

PR #5993, opened 2017-06-26
add zipkin to rancher-compose subgen

ruddell:
This adds Zipkin to the rancher-compose subgen when it's chosen.  I copied the config from the docker-compose subgen.  

### DIFF
--- a/generators/rancher-compose/index.js
+++ b/generators/rancher-compose/index.js
@@ -68,6 +68,7 @@ module.exports = RancherGenerator.extend({
             this.DOCKER_PROMETHEUS = constants.DOCKER_PROMETHEUS;
             this.DOCKER_PROMETHEUS_ALERTMANAGER = constants.DOCKER_PROMETHEUS_ALERTMANAGER;
             this.DOCKER_GRAFANA = constants.DOCKER_GRAFANA;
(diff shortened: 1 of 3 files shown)


jdubois:
I guess it's the same with Docker Swarm and Kubernetes, but couldn't this be optional? I think @PierreBesson did something on that part

- by default Zipkin doesn't work (as you need a special profile), so this new container isn't used by default
- this eats a lot of memory, now with Docker I need 10 Go to run my stack, and that's a lot more than before

I would love to have a specific question in the 3 sub-generators, that either turns Zipkin totally on (putting the profile & adding the image), or not running it at all. BTW, I think @stevehouel is working on refactoring the code to have common code between the 3 sub-generator, so this needs to be discussed with him.

stevehouel:
Yes @jdubois I m working on those 3 sub gen but first I will only do the
refactoring to merge common code and then we should work on that behavior
Le sam. 1 juil. 2017 à 08:04, Julien Dubois <notifications@github.com> a
écrit :

ruddell:
@jdubois I commented out the Zipkin sections so it is optional until the user enables it.  This should be good until the prompts are changed.

When configuring any Java container (such as Zipkin), it's important to set memory limits like in our [docker-compose docs](https://jhipster.github.io/docker-compose/#8) or else you'll end up using much more than needed (you may already be doing this, just trying to help).

PR #5998, opened 2017-06-27
registry fix k8s and oc

srinivasa-vasu:


xetys:
ok, for now, I have tested the following:

## Initial deployment
this seems to work well with this setup. As discussed in #5977 the registry itself appears as discoverable service. Disabling configuration `registerWithEureka` does not affect the clustering, which works properly. I see an only sense in the `fetchRegistry` attribute.

## postponed scaling
If you increase the replica count after deployments, this leads to a break in the peer-aware mode, as reloading the registry shows just a part of the instances available at the newly deployed instances. Over time, neither the old instances get the new replica under /replicas, nor the new instance list a full list of app instances. 

I tried my setup by adding `fetchRegistry` to `true`, while still not enabling `registerWithEureka`. Scaling the StatefulSet doesn't break the peer-aware mode as well as propagating the new eureka instances to all nodes and resulting in a working state.


So I cannot say exactly why this PR is failing to scale, and why the other work. However I think we have to combine both changes into something working

srinivasa-vasu:
@xetys - As per [this](https://github.com/spring-cloud/spring-cloud-netflix/issues/2070), registerWithEureka must be set 'true' while running Eureka in a peer-aware cluster mode. It is logical to have 1 instance to be a client to the other server.

I tested increasing the replica count, I managed to see the consistency between the nodes. It was not immediate, but eventual consistency because of the env variables timing that I had. I could see the same state within a min in all the instances. Hope, after increasing the replica count, you would have manually changed the 'cluster_size' in the initialisation block as well.

In terms of changes:
1. I have removed the 'self' as one of the replica instances in the initialisation block. So replica instances will be all other nodes except the 'self'
2. I have moved eureka - service-url from application.yml to <app-name>.yml. Reason being this property again overrides the env variable of the eureka stateful set (which should not be the desired state) as registry uses it while it starts.

xetys:
OK, I will check that again.... I partly agree now the self-registration part, despite it's still working without that for me (and that for >2 months in production).

Sorry If I am slowing down all this here....I just want our next Eureka to cause less headache on k8s 😄

srinivasa-vasu:
Sure, no issues 👍

xetys:
I cannot submit the issue is away for me :cry: 

Here is what I am doing:

1. checking out the latest commit of your branch
2. generating a complete setup using [this generator](https://github.com/xetys/jhipster-demo-generator)
3. running all kubernetes deployments as they are
4. wait, until all services are registered. I am using 5 replicas of each.
5. edit the StatefulSet and increasing the replica value to 3, as well as edit the application config map and add the new Eureka host to default Zone

Using this branch, I get the following picture:

1. the app does register somehow to the first 2 replicas, and they keep there. The replica page point to the other registry (0 to 1, 1 to 0)
2. the new replica is performing a registry fetch and fetches the complete state
3. it does NOT list itself, as the other 2 registries
4. and looses all replicas over time

so postponed scaling over time does not set up the third replica properly. You can observe this by reloading the registry UI a lot of times, to get load balanced to the third replica.

Using the other PR, repeatedly also is not consistent immediately. But after 2 or 3 minutes all 3 registries are consistent.

More facts:

I am running my tests against an application deployed to a non-default namespace (but corrected across all files properly on a self-hosted CoreOS cluster running kubernetes 1.6.4

Maybe I am doing something wrong?

srinivasa-vasu:
Steps look ok. But did u restart replica 0 and 1 after you spun up the third instance. Those instances need to be restarted (either one by one or at one go) to have the third instance as the additional replica. So in the end you will have 0 -> 1 & 2 as replica, 1 -> 0 & 2 as replica and 2 -> 0 & 1 as replica. This is the behaviour in Kubernetes. If you restart the existing replicas as and when you add the new one, state remains in sync. Also, I observed adding 'self' to the existing list is not causing any impact as well. Strangely, in OpenShift the whole setup works without any restart.

Also 'service-url ' configMap entry can be changed as 
`defaultZone: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/` - This is to avoid the manual activity whenever a new replica gets added. Instead we shall have the look-up through the headless service.

srinivasa-vasu:
@xetys Any update on this?

xetys:
I still have a pending review about the config map.

I didn't try to test my setup to kill pods after scaling the StateFul set.

How about taking the initializer from my PR in #5977 as the k8s version, and your one for the OpenShift parts, to make sure, scaling works without any further action like pod killing?

srinivasa-vasu:
In the mean time, we shall proceed that way. In the initialization part, service_url needs to be updated. Please check that. I will also see, if we can still further fine tune it, I see some scope for further improvements.

deepu105:
@srinivasa-vasu @xetys did you guys come to an agreement here and in the other related PR #5977 can we merge any of this?

srinivasa-vasu:
@deepu105 - This has some dependency on https://github.com/jhipster/jhipster-registry/pull/163(https://github.com/jhipster/jhipster-registry/pull/163). I have asked the opinion of @xetys and @PierreBesson

srinivasa-vasu:
@jdubois This PR is in-line with this change https://github.com/jhipster/jhipster-registry/pull/163. If @xetys has no comments, then this can be merged.

deepu105:
@xetys are you ok to merge this?

xetys:
hey,

I'm right back from summer break and will check it out the next days

jdubois:
Hi guys, can we merge this?

generators/kubernetes/templates/registry/_application-configmap.yml
```diff
@@ -30,15 +30,32 @@ data:
...
-  #  configserver:
-  #    status: Connected in prod
+  # app specific configuration
+  <%_ if (serviceDiscoveryType === 'eureka') {  _%>
+  <%= app.baseName.toLowerCase() %>.yml: |-
+    eureka:
+      client:
+        service-url:
+          # This must contain a list of all Eureka server replicas for registry HA to work correctly
+          defaultZone: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/
```
jdubois: Is `.svc.cluster.local` a standard domain name suffix in Kubernetes?

srinivasa-vasu: Yes for service objects

xetys:
I still have a pending question in the review section

srinivasa-vasu:
@xetys : pls refer this https://github.com/jhipster/jhipster-registry/pull/163. K8s config (beans annotated with @RefreshScope) gets refreshed whenever configMap gets updated, so that newly added instances are automatically added to the registry cluster.

xetys:
I didn't test this as deep as I did before, but I was concerning details that don't affect the awesome work to much. If I find something later, I will open another issue. So LGTM

generators/kubernetes/templates/registry/_application-configmap.yml
```diff
@@ -30,15 +30,22 @@ data:
...
       security:
         authentication:
           jwt:
             secret: <%= jwtSecretKey %>
-  #application-prod.yml: |-
-  #  configserver:
-  #    status: Connected in prod
+  # app specific configuration
+  <%_ if (serviceDiscoveryType === 'eureka') {  _%>
+  <%= app.baseName.toLowerCase() %>.yml: |-
```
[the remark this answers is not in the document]
srinivasa-vasu: @xetys : Thanks for highlighting this. Anyway, this entry per application will not change as the registration will be through one of the available registry instances. Scaling the registry servers doesn't need to be updated in the per app entry and this is not required any more in config-map and hence can be moved to app deployment xml.

jdubois:
Merging as this was approved by @xetys

PR #6001, opened 2017-06-29
Update field_validators.ejs

FireWolf2007:
Entity generation  (from JDL)
number Long required min(70000000000) max(79999999999)

 produces
```
@Min(value = 70000000000)
@Max(value = 79999999999L)
private Long number;
```
This change will add 'L' to both limits for type Long, I don't sure about old comparison, saved for now.

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/common/field_validators.ejs
+++ b/generators/entity/templates/server/src/main/java/package/common/field_validators.ejs
@@ -51,14 +51,15 @@ if (field.fieldValidate === true) {
         if (field.fieldType === 'Float' || field.fieldType === 'Double' || field.fieldType === 'BigDecimal') {
             validators.push('@DecimalMin(value = "' + field.fieldValidateRulesMin + '")');
         } else {
+            const isLong = (field.fieldValidateRulesMin > MAX_VALUE || field.fieldType == 'Long') ? 'L' : '';
             validators.push('@Min(value = ' + field.fieldValidateRulesMin + ')');
         }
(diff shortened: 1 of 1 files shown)


generators/entity/templates/server/src/main/java/package/common/field_validators.ejs
```diff
@@ -51,14 +51,15 @@ if (field.fieldValidate === true) {
         if (field.fieldType === 'Float' || field.fieldType === 'Double' || field.fieldType === 'BigDecimal') {
             validators.push('@DecimalMin(value = "' + field.fieldValidateRulesMin + '")');
         } else {
+            const isLong = (field.fieldValidateRulesMin > MAX_VALUE || field.fieldType == 'Long') ? 'L' : '';
```
deepu105: where is this used?

FireWolf2007: Example above:
>Entity generation (from JDL)
>number Long required min(70000000000) max(79999999999)

deepu105: No I meant I dont find any usage for `isLong` within this else block. Dont you have to append it to line below

FireWolf2007: Oh, sorry. It's my fault. Fixed.

FireWolf2007:
I think, this change require my example as simple test application for CI.

PR #6008, opened 2017-06-29
enable management endpoints CORS when using the dev profile

PierreBesson:

This is needed to make call from Swagger UI to the management endpoints work when proxying from browsersync.
Linked to https://github.com/jhipster/jhipster/pull/19

### DIFF
--- a/generators/server/templates/src/main/resources/config/_application-dev.yml
+++ b/generators/server/templates/src/main/resources/config/_application-dev.yml
@@ -213,6 +213,15 @@ liquibase:
     contexts: dev
 <%_ } _%>
 
+# CORS is only enabled with the "dev" profile for endpoints, so BrowserSync can access the Management API
+endpoints:
+    cors:
+        allowed-origins: "*"
+        allowed-methods: GET, PUT, POST, DELETE, OPTIONS
+        allowed-headers: "*"
+        allow-credentials: true
+        max-age: 1800
+
 # ===================================================================
 # To enable SSL, generate a certificate using:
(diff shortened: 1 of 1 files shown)


PierreBesson:
@jdubois when testing this: https://github.com/jhipster/jhipster/pull/19, I needed to add this to make the management calls work from swagger UI but you are right that it's fishy that this is needed.

jdubois:
@PierreBesson one of them is bad, that's for sure

jdubois:
OK, I get it: this wasn't exposed by Spring Boot, so we were doing it ourselves. Now it's a standard Spring Boot property, let's use it.

I'm merging this and doing the necessary changes.

jdubois:
So I merged this and changed everything, and now that I'm testing: I have no idea why, but our previous code was working, and this one isn't!!!!!
There's definitely something that was done in our previous code that was solving the issue, so:

- I'm rollbacking everything
- This ticket remains closed as this was already working

PierreBesson:
@jdubois I don't think the spring boot `endpoint.cors` property is doing the same thing as the jhipster cors properties that are used to initialize the CorsFilter bean, see [src/main/java/package/config/_WebConfigurer.java##L242](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/_WebConfigurer.java#L242).

So I think you did the right thing by reverting, it seems using the corsFilter explicitly is better because it lets you customize which endpoints to apply CORS to.

cbornet:
`endpoints.cors` is only for actuator endpoints.

jdubois:
Oh thanks @cbornet

PierreBesson:
@cbornet Oh Yes I relever now that was why I wanted to enable it because of : https://github.com/jhipster/jhipster/pull/19

PR #6015, opened 2017-07-01
set liquibase xsd file to 3.5

ctamisier:
liquibase 3.5.3 is used, so this xsd file version in the xml can be set to 3.5 as well.

And being able to see the context attribute valid like in the following migration:
`<include context="!test" file="classpath:config/liquibase/changelog/20170630140601_partial_index.sql" relativeToChangelogFile="false"/>`


ctamisier:
oops..., done

PR #6022, opened 2017-07-03
Up-to-date Gatling

JulienMrgrd:
This PR is intended to update Gatling (versions, structure, ...).

Concerning the structure, in particular the new "user-files" dir, I've follow the official documentation (http://gatling.io/docs/current/general/bundle_structure/)

In the pom.xml, we can use default args to give to all gatling tests (http://gatling.io/docs/current/cookbook/passing_parameters/).
I also added a dir, "target/gatling/results/${maven.build.timestamp}/" in pom.xml, to have different folders each time you launch Gatling tests.

In addition, if a BigDecimal field of an entity is required, when gatling tries to create the entity, the call fails.
So I added a case when an entity field is a BigDecimal. 

Maven and Yarn tests are OK.

---------------------------------------------------

### DIFF
--- a/generators/entity/files.js
+++ b/generators/entity/files.js
@@ -149,9 +149,9 @@ const serverFiles = {
             condition: generator => generator.gatlingTests,
             path: TEST_DIR,
             templates: [{
-                file: 'gatling/simulations/_EntityGatlingTest.scala',
+                file: 'gatling/user-files/simulations/_EntityGatlingTest.scala',
                 options: { interpolate: INTERPOLATE_REGEX },
-                renameTo: generator => `gatling/simulations/${generator.entityClass}GatlingTest.scala`
+                renameTo: generator => `gatling/user-files/simulations/${generator.entityClass}GatlingTest.scala`
             }]
         }
     ]
--- a/generators/entity/templates/server/src/test/gatling/user-files/simulations/_EntityGatlingTest.scala
+++ b/generators/entity/templates/server/src/test/gatling/user-files/simulations/_EntityGatlingTest.scala
@@ -135,7 +135,7 @@ class <%= entityClass %>GatlingTest extends Simulation {
             .exec(http("Create new <%= entityInstance %>")
             .post("<% if (applicationType === 'microservice') {%>/<%= baseName.toLowerCase() %><% } %>/api/<%= entityApiUrl %>")
             .headers(headers_http_authenticated)
-            .body(StringBody("""{"id":null<% for (idx in fields) { %>, "<%= fields[idx].fieldName %>":<% if (fields[idx].fieldType === 'String') { %>"SAMPLE_TEXT"<% } else if (fields[idx].fieldType === 'Integer' || fields[idx].fieldType === 'BigDecimal') { %>"0"<% } else if (fields[idx].fieldType === 'Instant' || fields[idx].fieldType === 'ZonedDateTime' || fields[idx].fieldType === 'LocalDate') { %>"2020-01-01T00:00:00.000Z"<% } else { %>null<% } } %>}""")).asJSON
+            .body(StringBody("""{"id":null<% for (idx in fields) { %>, "<%= fields[idx].fieldName %>":<% if (fields[idx].fieldType === 'String') { %>"SAMPLE_TEXT"<% } else if (['Integer', 'BigDecimal'].includes(fields[idx].fieldType)) { %>"0"<% } else if (['Instant', 'ZonedDateTime', 'LocalDate'].includes(fields[idx].fieldType)) { %>"2020-01-01T00:00:00.000Z"<% } else { %>null<% } } %>}""")).asJSON
             .check(status.is(201))
             .check(headerRegex("Location", "(.*)").saveAs("new_<%= entityInstance %>_url"))).exitHereIfFailed
             .pause(10)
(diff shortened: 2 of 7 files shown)


generators/entity/templates/server/src/test/gatling/user-files/simulations/_EntityGatlingTest.scala
```diff
@@ -135,7 +135,7 @@ class <%= entityClass %>GatlingTest extends Simulation {
             .exec(http("Create new <%= entityInstance %>")
             .post("<% if (applicationType === 'microservice') {%>/<%= baseName.toLowerCase() %><% } %>/api/<%= entityApiUrl %>")
             .headers(headers_http_authenticated)
-            .body(StringBody("""{"id":null<% for (idx in fields) { %>, "<%= fields[idx].fieldName %>":<% if (fields[idx].fieldType === 'String') { %>"SAMPLE_TEXT"<% } else if (fields[idx].fieldType === 'Integer') { %>"0"<% } else if (fields[idx].fieldType === 'Instant' || fields[idx].fieldType === 'ZonedDateTime' || fields[idx].fieldType === 'LocalDate') { %>"2020-01-01T00:00:00.000Z"<% } else { %>null<% } } %>}""")).asJSON
+            .body(StringBody("""{"id":null<% for (idx in fields) { %>, "<%= fields[idx].fieldName %>":<% if (fields[idx].fieldType === 'String') { %>"SAMPLE_TEXT"<% } else if (fields[idx].fieldType === 'Integer' || fields[idx].fieldType === 'BigDecimal') { %>"0"<% } else if (fields[idx].fieldType === 'Instant' || fields[idx].fieldType === 'ZonedDateTime' || fields[idx].fieldType === 'LocalDate') { %>"2020-01-01T00:00:00.000Z"<% } else { %>null<% } } %>}""")).asJSON
```
deepu105: since we are touching this line lets simplify it like below
```
.body(StringBody("""{"id":null<% for (idx in fields) { %>, "<%= fields[idx].fieldName %>":<% if (fields[idx].fieldType === 'String') { %>"SAMPLE_TEXT"<% } else if (['Integer', 'BigDecimal'].includes(fields[idx].fieldType)) { %>"0"<% } else if (['Instant', 'ZonedDateTime', 'LocalDate'].includes(fields[idx].fieldType)) { %>"2020-01-01T00:00:00.000Z"<% } else { %>null<% } } %>}""")).asJSON
  
```

generators/entity/files.js
```diff
@@ -149,9 +149,9 @@ const serverFiles = {
             condition: generator => generator.gatlingTests,
             path: TEST_DIR,
             templates: [{
-                file: 'gatling/simulations/_EntityGatlingTest.scala',
+                file: 'gatling/user-files/simulations/_EntityGatlingTest.scala',
```
deepu105: why this change?

test/test-creation.js
```diff
@@ -973,8 +973,8 @@ describe('JHipster generator', () => {
                 `${TEST_DIR}features/user/user.feature`
             ]);
             assert.noFile([
-                `${TEST_DIR}gatling/gatling.conf`,
-                `${TEST_DIR}gatling/logback.xml`
+                `${TEST_DIR}gatling/conf/gatling.conf`,
```
deepu105: I dont see the renamed files in this PR

JulienMrgrd: Sorry, indeed I forgot to move the gatling.conf & the logback.xml inside this new folder.

generators/entity/files.js
JulienMrgrd: To follow the new Gatling structure (http://gatling.io/docs/current/general/bundle_structure/)

generators/entity/templates/server/src/test/gatling/user-files/simulations/_EntityGatlingTest.scala
JulienMrgrd: Ok nice, I will do PR now

test/test-creation.js
JulienMrgrd: Apparently, Github do not show when we move files. But I correctly moved the two files in generator-jhipster/generators/server/templates/src/test/gatling/conf/.

atomfrede:
Another thing, why don't we put the gatling files into `src/gatling`? @JulienMrgrd I can understand the rational behind using `user-files` but it seems strange to me.

JulienMrgrd:
Indeed @atomfrede.
But since all Gatling files were in the "test" folder before my PR, I did not want to move everything.
Moreover, as Gatling is a stress tool to test his server, it can be understood that it's in the "test" folder.

test/test-creation.js
deepu105: Wasnt the files already in conf folder? may be the tests were wrong

deepu105:
I think this should be in `src/test`. But having `user-files` folder is strange, I prefer then old structure. If it's not a requirement by Gatling to have `user-files` folder then I would vote to change to `src/test/gatling/simulations`

test/test-creation.js
JulienMrgrd: Yes indeed, tests were wrong. I thought I had moved them, but no I only made a refactoring of these tests during my PR.
Anyway, now it's ok

JulienMrgrd:
It's not mandatory by Gatling to use a `user-files` folder, just a strange indication indeed.
So I can revert these changes concerning this folder.

PR #6024, opened 2017-07-03
cacheBuster with build timestamp for translation json files

ctamisier:
When I deploy the app with a war file to tomcat or using the jhipster deployment script to heroku,

the translation json files (e.g. global.json, home.json, ...) are not necessarily reloaded by $translateProvider. This causes that the updated translations are not loaded and I have to manually delete the browser cache..

I have most of the time a `200  (from disk cache)` with chrome or a `200 cached` with firefox. 

I firstly investigate in `WebConfigurer` through `initCachingHttpHeadersFilter` and add `cachingHttpHeadersFilter.addMappingForUrlPatterns(disps, true, "/i18n/*");`, thinking that the caching rules are not applied to i18n url => but I still have the same problem..

Then I tried `$translateProvider.useLoaderCache(false);` => same problem..

So through this PR, I added a "cacheBuster" using the build timestamp (a new constant) on the `urlTemplate`,

so the urls look like this having necessarily a `200` with the new files the first time and then `from cache` on next requests: 
- global.json?build=1499106805822
- home.json?build=1499106805822

I didn't try on Angular 4 yet, but for me this mechanism fixes those caching browser issues in a general way.

WDYT

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/blocks/config/_translation.config.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/blocks/config/_translation.config.js
@@ -23,12 +23,12 @@
         .module('<%=angularAppName%>')
         .config(translationConfig);
 
-    translationConfig.$inject = ['$translateProvider', 'tmhDynamicLocaleProvider'];
+    translationConfig.$inject = ['$translateProvider', 'tmhDynamicLocaleProvider', 'BUILD_TIMESTAMP'];
 
-    function translationConfig($translateProvider, tmhDynamicLocaleProvider) {
+    function translationConfig($translateProvider, tmhDynamicLocaleProvider, BUILD_TIMESTAMP) {
         // Initialize angular-translate
         $translateProvider.useLoader('$translatePartialLoader', {
-            urlTemplate: 'i18n/{lang}/{part}.json'
+            urlTemplate: 'i18n/{lang}/{part}.json' + (BUILD_TIMESTAMP ? '?build=' + BUILD_TIMESTAMP : '')
         });
 
         $translateProvider.preferredLanguage('<%= nativeLanguage %>');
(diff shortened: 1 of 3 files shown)


generators/client/templates/angularjs/src/main/webapp/app/blocks/config/_translation.config.js
```diff
@@ -23,12 +23,12 @@
...
 
-    translationConfig.$inject = ['$translateProvider', 'tmhDynamicLocaleProvider'];
+    translationConfig.$inject = ['$translateProvider', 'tmhDynamicLocaleProvider', 'BUILD_TIMESTAMP'];
 
-    function translationConfig($translateProvider, tmhDynamicLocaleProvider) {
+    function translationConfig($translateProvider, tmhDynamicLocaleProvider, BUILD_TIMESTAMP) {
         // Initialize angular-translate
         $translateProvider.useLoader('$translatePartialLoader', {
-            urlTemplate: 'i18n/{lang}/{part}.json'
+            urlTemplate: 'i18n/{lang}/{part}.json' + (BUILD_TIMESTAMP ? '?build=' + BUILD_TIMESTAMP : '')
```
deepu105: why not just send current time? why build time?

ctamisier: To hit the cache and avoiding reload the same file. The translation files don't change for the same deployment (unlike db data for example)..

ctamisier:
And I don't have this problem with angular 4

generators/client/templates/angularjs/src/main/webapp/app/blocks/config/_translation.config.js
deepu105: ok

deepu105:
We normally dont update angular 1 part, but if this is an issue then its fine I guess

PR #6028, opened 2017-07-04
Do not use build timestamp for dev

ctamisier:
- Remove  build timestamp for dev build because it updates 'app.constants.js' everytime we run 'gulp' which could be annoying (and useless).

- Change to double quotes in `.constant('BUILD_TIMESTAMP', "");` as gulp produces double quote rather than simple ones...

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/_app.constants.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/_app.constants.js
@@ -23,5 +23,5 @@
         .module('<%=angularAppName%>')
         .constant('VERSION', '0.0.1-SNAPSHOT')
         .constant('DEBUG_INFO_ENABLED', true)
-        .constant('BUILD_TIMESTAMP', '');
+        .constant('BUILD_TIMESTAMP', "");
 })();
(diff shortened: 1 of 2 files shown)


deepu105:
@ctamisier are you sure that this is the last time you PR a change for this :smiling_imp:

generators/client/templates/angularjs/src/main/webapp/app/_app.constants.js
```diff
@@ -23,5 +23,5 @@
         .module('<%=angularAppName%>')
         .constant('VERSION', '0.0.1-SNAPSHOT')
         .constant('DEBUG_INFO_ENABLED', true)
-        .constant('BUILD_TIMESTAMP', '');
+        .constant('BUILD_TIMESTAMP', "");
```
deepu105: Did you check if Lint is ok with this?

ctamisier:
yes !

generators/client/templates/angularjs/src/main/webapp/app/_app.constants.js
ctamisier: it is ok (for the eslint config we have) and gulp produces those double quotes

ctamisier:
yes last PR for this

PR #6042, opened 2017-07-06
add a default locale provider

xiasenhai:
By adding a default local provider, the Date/Number string will format and display as local style not US style.

already answered by this link:
https://stackoverflow.com/questions/34904683/how-to-set-locale-in-datepipe-in-angular2

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import { NgModule } from '@angular/core';
+import { NgModule, LOCALE_ID } from '@angular/core';
 import { Title } from '@angular/platform-browser';
 
 <%_ if (websocket === 'spring-websocket') { _%>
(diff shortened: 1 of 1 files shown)


deepu105:
But I thought we already do this with date pipe. Im on my phone cam someone test this please

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -51,7 +51,11 @@ import {
         <%_ } if (websocket === 'spring-websocket') { _%>
         WindowRef,
         <%_ } _%>
-        Title
+        Title,
+        {
+            provide: LOCALE_ID,
+            useValue: '<%= nativeLanguage %>'
+        },
```
deepu105: remove the comma, it was there by mistake

aleris:
I can confirm that without it all the dates are displayed in US format, so the locale is not used.
With it:
- it display correctly the date with the angular pipe;
- it display correctly the localized date in the input of the LocalDate once the date was picked, however the picker UI is not localized;
- it does not correctly display the localized date in the input of ZonedDateTime or the UI for it.

sendilkumarn:
Let us merge this. I have few other lints to do. I will add this with that. 👍

PR #6046, opened 2017-07-06
use hand cursor on sortable header

ctamisier:
(like in angular js)


deepu105:
thanks

deepu105:
please add [ci skip] to you commit for such trivial changes that doesnt affect builds

tientq:
You should add [ci skip] to contribute guidelines, don't you?

deepu105:
@tientq nope as its easy for people not to read it properly and end updating ci skip for stuff that might break build.

tientq:
Ok, you are right.

PR #6066, opened 2017-07-09
fix ID header 

ctamisier:
1) fix indent for the ID header (align with the other `<th>` in the html template)
2) use <% %> instead of <%_ _%> (which let the extra space between the 2 span tags, like in the other `<th>` headers and like in `_entity-management.component.html`)

This PR in mainly for the 2) which 'disjoin' the ID header with its sortable icon


ctamisier:
I forgot to put [ci skip]..

Issue #6052, opened 2017-07-07
Getting "TypeError: The header content contains invalid characters" when searching with non ASCII text


aleris:
##### **Overview of the issue**

With default generated code, when using Elasticsearch and Angular 2, getting "TypeError: The header content contains invalid characters" when searching with non ASCII characthers, for example "ș". Below the client side exception:

```
TypeError: The header content contains invalid characters
    at ServerResponse.setHeader (_http_outgoing.js:374:11)
    at setHeader (D:\Dev\openjob10\node_modules\http-proxy-middleware\node_modules\http-proxy\lib\http-proxy\passes\web-outgoing.js:94:15)
    at D:\Dev\openjob10\node_modules\http-proxy-middleware\node_modules\http-proxy\lib\http-proxy\passes\web-outgoing.js:116:7
    at Array.forEach (native)
    at Array.writeHeaders (D:\Dev\openjob10\node_modules\http-proxy-middleware\node_modules\http-proxy\lib\http-proxy\passes\web-outgoing.
js:111:35)
    at ClientRequest.<anonymous> (D:\Dev\openjob10\node_modules\http-proxy-middleware\node_modules\http-proxy\lib\http-proxy\passes\web-in
coming.js:166:20)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:191:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:522:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
    at Socket.socketOnData (_http_client.js:411:20)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:191:7)
    at readableAddChunk (_stream_readable.js:178:18)
    at Socket.Readable.push (_stream_readable.js:136:10)
    at TCP.onread (net.js:561:20)
```

##### **Motivation for or Use Case**

Bug?

##### **Reproduce the error**

In the search field of an entity put "școala" for example and hit enter. 

##### **Suggest a Fix**

1. In request-util.ts wrap `req.query` with `encodeURIComponent`:

`params.set('query', encodeURIComponent(req.query));`

2. Add a new `QueryUtil` class in \web\rest\util and an utility static method which is using `URLDecoder.decode` to decode the query as encoded by `encodeURIComponent` on client.

3. 
Change `queryStringQuery(query)` to use `QueryUtil `decode method in:
```
[...]

PR #6070, opened 2017-07-10
Fix for invalid characters when searching with non ASCII text #6052 

aleris:
Fix for "TypeError: The header content contains invalid characters" when searching with non ASCII text, see #6052.


gzsombor:
Can you solve with this https://stackoverflow.com/a/34836332 ?
That way, it should work with other parameters as well.

aleris:
Not the character encoding is the problem but url escaping (MIME url encoding). 

However, I did a little more research and the solution proposed above was more like a workaround than a solution. The actual problem was with `PaginationUtil.generateSearchPaginationHttpHeaders` which was not encoding the query string back. The solution worked because the query was encoded twice. 

In the mean time I see that the function [was updated already](https://github.com/jhipster/generator-jhipster/commit/3b7d696ebae3dec4dda084dab9cfa813df257dd5) to do a proper url escaping so this PR can be canceled as the problem is already fixed.

PR #6073, opened 2017-07-10
Enhance and separate RTL support

sohibegit:
Separated it as I want to override some of bootstrap CSS

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -58,6 +58,13 @@ const files = {
                 'content/css/_vendor.css',
                 'content/css/_documentation.css'
             ]
+        },
+        {
+            condition: generator => !generator.useSass && generator.enableI18nRTL,
+            path: MAIN_SRC_DIR,
+            templates: [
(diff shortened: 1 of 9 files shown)


generators/client/files-angular.js
```diff
@@ -58,6 +58,13 @@ const files = {
                 'content/css/_vendor.css',
                 'content/css/_documentation.css'
             ]
+        },
+        {
+            condition: generator => !generator.useSass && generator.enableI18nRTL,
```
gmarziou: Does it mean that enableI18nRTL is a global option for the whole app? 
If so, can we build an app with both a mix of LTR and RTL languages?

deepu105: yes, the styles are applied based on the language choosen at run time

gmarziou: Cool :)

deepu105:
@sohibegit could you please also add a test case for this in the test-languages.js under test directory?

sohibegit:
@deepu105 sure

PR #6079, opened 2017-07-11
Changed role strings in angular 2 client to typed properties of a new AuthoritiesConstants class

aleris:
Added a new AuthoritiesConstants class on angular 2 client side in auth folder, similar with the java server one.
All "hard-coded" references of form 'ROLE_ADMIN', 'ROLE_USER' are now changed to AuthoritiesConstants.ADMIN, AuthoritiesConstants.USER.
Please note that I also removed an unused 'authorities' property in _entity-management-dialog.component.ts

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
@@ -21,6 +21,7 @@ import { Router } from '@angular/router';
 import { NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import { JhiLanguageService } from 'ng-jhipster';
 
+import { AuthoritiesConstants } from '../../shared';
 import { ProfileService } from '../profiles/profile.service';
 import { <% if (enableTranslation) { %>JhiLanguageHelper, <% } %>Principal, LoginModalService, LoginService } from '../../shared';
 
@@ -39,6 +40,7 @@ import { VERSION, DEBUG_INFO_ENABLED } from '../../app.constants';
 })
 export class NavbarComponent implements OnInit {
 
+    AuthoritiesConstants = AuthoritiesConstants;
     inProduction: boolean;
     isNavbarCollapsed: boolean;
     languages: any[];
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
@@ -40,7 +40,7 @@ import { VERSION, DEBUG_INFO_ENABLED } from '../../app.constants';
 })
 export class NavbarComponent implements OnInit {
 
-    AuthoritiesConstants = AuthoritiesConstants;
+    authoritiesConstants = AuthoritiesConstants;
     inProduction: boolean;
     isNavbarCollapsed: boolean;
     languages: any[];
--- a/generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
@@ -1,0 +1,7 @@
+export class AuthoritiesConstants {
+    public static readonly ADMIN = 'ROLE_ADMIN';
+
+    public static readonly USER = 'ROLE_USER';
+
+    public static readonly ANONYMOUS = 'ROLE_ANONYMOUS';
+}
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
@@ -62,7 +62,6 @@ import { ResponseWrapper } from '../../shared';
 export class <%= entityAngularName %>DialogComponent implements OnInit {
 
     <%= entityInstance %>: <%= entityAngularName %>;
-    authorities: any[];
     isSaving: boolean;
     <%_
     const query = generateEntityQueries(relationships, entityInstance, dto);
@@ -100,7 +99,6 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
 
     ngOnInit() {
         this.isSaving = false;
-        this.authorities = ['ROLE_USER', 'ROLE_ADMIN'];
         <%_ for (idx in queries) { _%>
         <%- queries[idx] %>
         <%_ } _%>
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -406,6 +406,7 @@ const files = {
         {
             path: ANGULAR_DIR,
             templates: [
+                'shared/auth/_authorities-constants.ts',
                 'shared/auth/_csrf.service.ts',
                 'shared/auth/_state-storage.service.ts',
                 'shared/auth/_principal.service.ts',
--- a/generators/client/templates/angular/src/main/webapp/app/account/password/_password.route.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/account/password/_password.route.ts
@@ -18,14 +18,14 @@
 -%>
 import { Route } from '@angular/router';
 
-import { UserRouteAccessService } from '../../shared';
+import { UserRouteAccessService, AuthoritiesConstants } from '../../shared';
 import { PasswordComponent } from './password.component';
 
 export const passwordRoute: Route = {
     path: 'password',
     component: PasswordComponent,
     data: {
-        authorities: ['ROLE_USER'],
+        authorities: [AuthoritiesConstants.USER],
         pageTitle: 'global.menu.account.password'
     },
     canActivate: [UserRouteAccessService]
--- a/generators/client/templates/angular/src/main/webapp/app/account/sessions/_sessions.route.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/account/sessions/_sessions.route.ts
@@ -18,14 +18,14 @@
 -%>
 import { Route } from '@angular/router';
 
-import { UserRouteAccessService } from '../../shared';
+import { UserRouteAccessService, AuthoritiesConstants } from '../../shared';
 import { SessionsComponent } from './sessions.component';
 
 export const sessionsRoute: Route = {
     path: 'sessions',
     component: SessionsComponent,
     data: {
-        authorities: ['ROLE_USER'],
+        authorities: [AuthoritiesConstants.USER],
         pageTitle: 'global.menu.account.sessions'
     },
     canActivate: [UserRouteAccessService]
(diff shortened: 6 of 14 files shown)


jdubois:
This is definitely a good idea!

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
```diff
@@ -39,6 +40,7 @@ import { VERSION, DEBUG_INFO_ENABLED } from '../../app.constants';
 })
 export class NavbarComponent implements OnInit {
 
+    AuthoritiesConstants = AuthoritiesConstants;
```
deepu105: make instances start with lowercase `authoritiesConstants`

generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
```diff
@@ -0,0 +1,7 @@
+export class AuthoritiesConstants {
```
deepu105: Maybe singular might be better? `AuthoritiesConstant`

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -100,7 +99,6 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
 
     ngOnInit() {
         this.isSaving = false;
-        this.authorities = ['ROLE_USER', 'ROLE_ADMIN'];
```
deepu105: wasn't this used? even in html?

generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
gzsombor: I always wondered, why can't we use enums for this on the client side?

deepu105: Typescript does support enums, so technically we can

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.component.ts
aleris: ok will change

generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
aleris: Regarding singular, I kept the same name as server side (see generators\server\templates\src\main\java\package\security\_AuthoritiesConstants.java). 
Regarding enum suport, unfortunately for typescript version <2.4 enums do not have string literals. 
In 2.4 are/will be supported, [see blog announcement](https://blogs.msdn.microsoft.com/typescript/2017/06/12/announcing-typescript-2-4-rc/#user-content-string-enums), but I guess will have to wait for the version upgrade.

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
aleris: nope, double checked

generators/client/templates/angular/src/main/webapp/app/shared/auth/_authorities-constants.ts
deepu105: ok

deepu105:
@aleris this is breaking the AOT production build with latest angular 4.3.0
```

$ node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js --config webpack/webpack.prod.js --progress --profile
Starting compilation using the angular compiler.
Angular compilation done, starting webpack bundling.
Error: Illegal state: symbol without members expected, but got {"filePath":"/home/deepu/workspace/temp/jh4test/src/main/webapp/app/shared/auth/authorities-constants.ts","name":"AuthoritiesConstants","members":["ADMIN"]}.
    at StaticSymbol.assertNoMembers (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:700:19)
    at AotSummaryResolver.resolveSummary (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:25670:22)
    at CompileMetadataResolver._loadSummary (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15110:66)
    at CompileMetadataResolver._getEntryComponentMetadata (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15927:49)
    at /home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15909:48
    at Array.forEach (native)
    at CompileMetadataResolver._getEntryComponentsFromProvider (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15908:30)
    at /home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15872:83
    at Array.forEach (native)
    at CompileMetadataResolver._getProvidersMetadata (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15835:19)
    at /home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15837:23
    at Array.forEach (native)
    at CompileMetadataResolver._getProvidersMetadata (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15835:19)
    at /home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15410:63
    at Array.forEach (native)
    at CompileMetadataResolver.getNgModuleMetadata (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:15401:49)
    at addNgModule (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:24242:58)
    at /home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:24253:14
    at Array.forEach (native)
    at _createNgModules (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:24252:26)
    at analyzeNgModules (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:24127:14)
    at analyzeAndValidateNgModules (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:24137:35)
    at AotCompiler.analyzeModulesAsync (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler/bundles/compiler.umd.js:23773:46)
    at CodeGenerator.codegen (/home/deepu/workspace/temp/jh4test/node_modules/@angular/compiler-cli/src/codegen.js:32:14)
    at /home/deepu/workspace/temp/jh4test/node_modules/ngc-webpack/src/main.js:11:103
    at Object.main (/home/deepu/workspace/temp/jh4test/node_modules/@angular/tsc-wrapped/src/main.js:106:16)
    at Object.runInternal (/home/deepu/workspace/temp/jh4test/node_modules/ngc-webpack/src/main.js:15:27)
    at /home/deepu/workspace/temp/jh4test/node_modules/ngc-webpack/src/plugin.js:53:48
    at process._tickCallback (internal/process/next_tick.js:109:7)
    at Module.runMain (module.js:606:11)
error Command failed with exit code 1.

```

deepu105:
@jdubois this will block us from upgrading to angular 4.3.0 so either we should revert this or find a solution. @aleris can you try to find a solution for this?

jdubois:
AOT compatibility is such a pain!!!!! I really find this change is a great idea, but we should indeed revert it before people use it (as soon as this is released, going back will be very annoying)

deepu105:
@aleris I tried many workarounds like using a constant instead of static class using ENUM etc but the issue remains so I'm reverting this for now. If you find way to make it work with angular 4.3 please do a new PR. I'm really sorry It was indeed a great improvement but as @jdubois said working with AOT is a pain in the ass as you cannot do many of the simple JS/TS stuff there

gzsombor:
If here, the enum cause problems with AOP compilation, how come that in the generated entities, where there are enums as well, they are working fine? AOP is a strange black magic ... :disappointed:

deepu105:
I have no idea, I think AOT has issues only when we use enums or constants
in route declaration files that too it works in version 4.2 but fails in
4.3 as they added a new assertion

aleris:
Sorry for delayed response. It certainly looks like a bug in AOT.  I tried different versions also and can confirm that nothing that exports a constant member inside anything else (class, enum, doesn't matter) does not work. What's worse is that even using a export default const variable does not work, because it does not accept any usage construct like `a: [B.c]`. So the only solution I found is:

1. in \main\webapp\app\shared\auth\authorities-constants.ts use:
```
export const ADMIN = 'ROLE_ADMIN';
export const USER = 'ROLE_USER';
export const ANONYMOUS = 'ROLE_ANONYMOUS';
```

2. import directly from file (skip index.ts import and reexport):

`import * as AuthoritiesConstants from '../../shared/auth/authorities-constants';`

3. use a local 

```
const USER = AuthoritiesConstants.USER;
...
authorities: [USER],
```

It is a little convoluted, let me know what you think.

deepu105:
@aleris thats too much lines of code just to use a constant. May be we should create an issue in Angular. Since you might have better info on this would you like to do that?

deepu105:
@manekinekko do you think this could be an AOT bug or is it by design?

aleris:
Agree, is too much, maybe is better to wait for the fix. I will try to make a minimal reproducible scenario to submit to angular. 

I do not think is by design, as it worked ok in 4.2.6.

manekinekko:
@deepu105 haven't heard about an issue related to AOT and exported enums. 
But, I've built a little demo app (4.3) using both Enums and your AuthoritiesConstants class, and I can compile it to AOT without errors. Do you have a demo app with this bug I can reproduce?

AOT is all about taking your app and compiling it during built time in order to catch errors ahead of time. there is no black magic involved. The source code is here: https://github.com/angular/angular/tree/master/packages/compiler-cli

aleris:
I submitted [an issue in angular](https://github.com/angular/angular/issues/18170). I attached there a zip with a minimal project to reproduce the problem.

I also found another simpler workaround.

If we use:
`import { AuthoritiesConstants } from '../../shared/auth/authorities-constants';`

Instead of
`import { AuthoritiesConstants } from '../../shared;`

The problem does not manifest any more. What do you think?

deepu105:
@manekinekko thanks for the clarification. @aleris was faster to give a sample :)
This was working with Angular 4.2.6 but started to throw error with Angular 4.3.0
In general is it a good idea to use an index barrels file at folder root with all members exported? I believe I followed an older ng2 style guide while doing this but I see that current style guide doesn't have any mention of barrels

manekinekko:
the barrels are considered to be a good practice with SystemJS since they make the configuration much easier. But with webpack this is not mandatory, however, I consider it to be a good practice to follow (especially for big projects).

deepu105:
@manekinekko thanks for clearing that up :smile_cat:

PR #6083, opened 2017-07-11
Fix jhiItemCount zh-cn translation

iconben:
This is a small patch to fix item counter translation in pagination.


generators/client/templates/angularjs/src/main/webapp/app/components/util/_jhi-item-count.directive.js
```diff
@@ -20,8 +20,11 @@
     'use strict';
 
     var jhiItemCount = {
-        template: '<div class="info">' +
-                    'Showing {{(($ctrl.page - 1) * $ctrl.itemsPerPage) === 0 ? 1 : (($ctrl.page - 1) * $ctrl.itemsPerPage + 1)}} - ' +
+        template: '<div class="info" data-translate="global.item-count" ' +
+        'translate-value-first="{{(($ctrl.page - 1) * $ctrl.itemsPerPage) == 0 ? 1 : (($ctrl.page - 1) * $ctrl.itemsPerPage + 1)}}" ' +
```
deepu105: please align it properly, the indentation here is wrong

PR #6094, opened 2017-07-12
Refactor docker modules

stevehouel:
Refactor and Merge common code on "docker" modules which are listed below:
- Kubernetes
- Rancher compose
- Openshift
- Docker compose


deepu105:
Great work @stevehouel I'll review it soon

deepu105:
I had quick first look, few questions/comments

You moved sub generators into a folder, and modified cli to run them, now what about the below commands? we still support yeoman CLI as well for legacy reason. Can you check `yo jhipster:docker-compose`, `yo jhipster:kubernetes` etc?

What was the reason to move everything inside a docker folder? Agree that aesthetically they look better but yeoman convention is to have sub generators in folders at same level as main generator

We could have moved all common code into two files called `docker-base.js`, `docker-base-promts.js` along side the `generator-base.js` as well and kept the old file structure for all the sub generators

deepu105:
@stevehouel I dont have an issue with the structure if it works with both cli and yeoman

stevehouel:
So @deepu105 I rework on those folder structure and move everything to root generators folder. That was working on my computer the previous structure but for your knowledge (on my side I did not know), even if you use npm ou yarn link if you call a module from yeoman who does not exist on current generator version, it will look on all your versions and then call it. So that was working but not on the linked generators.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "generator-jhipster",
-  "version": "4.6.1",
+  "version": "4.6.2",
```
deepu105: This shouldnt be changed

deepu105: its causing build failure as well

PR #6101, opened 2017-07-12
Add support for arabic language

sohibegit:


generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -54,7 +54,11 @@ import {
         Title,
         {
             provide: LOCALE_ID,
+        <%_ if (languages.includes('ar-ly')) { _%>
```
deepu105: Why?

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -54,7 +54,11 @@ import {
         Title,
         {
             provide: LOCALE_ID,
+        <%_ if (nativeLanguage === 'ar-ly') { _%>
+            useValue: 'en'
```
sohibegit: This line will Force angular to use en locale as angular Arabic locale will result into Arabic numbers like (۰۱۲۳٤۵) and those are not wanted in most of Arab countries also Arabic numbers will result into Spring exceptions when trying to parse them

sohibegit:
> Could you also add a specific test case to check rtl files are included when language is rtl?

@deepu105 I need a global property inside .yo-rc.json to make the test for example:
`"containsRTLLang": true,`

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
deepu105: ok, but I dont like hardcoding this for a language. Not sure if we will have same issue for other languages which might have non standard numbers. Anyway dont worry I'll change it later
@jhipster/developers any idea about other languages that might have similar issue here so that I can exclude them as well?

deepu105:
@sohibegit that shouldnt be needed. Anyway i'll add the test

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
sohibegit: @deepu105 anyway, this line `useValue: 'en'` will never care about changing the language on run-time.
You can try to use `useValue: 'ar-ly'` in one of your apps after that even if you change to any other language at run-time it will always show Arabic numbers especially Dates you can check Created date at User management.

Issue #6051, opened 2017-07-07
Upgrade to Typescript 2.4.x


deepu105:
Please note that TS 2.4.x breaks our builds due to following issues (Actually this version of TS has better type inference and stricter type checks which surface some of the hidden issues in our code), hence we need to fix those as well

1.  RxJS has some issues https://github.com/ReactiveX/rxjs/issues/2539
2. Our entity component code has issues

deepu105:
@pascalgrimaud I found something weird with our travis prod build while working on this issue.
https://travis-ci.org/jhipster/generator-jhipster/jobs/253976729#L3315
The build either failed or stalled but its reported as passed. Any idea?

deepu105:
@pascalgrimaud not just one most of our recent prod builds are stalled at that point so technically we are never testing the prod AOT builds https://travis-ci.org/jhipster/generator-jhipster/jobs/253721138#L3322

pascalgrimaud:
I will have a look tomorrow @deepu105

deepu105:
@pascalgrimaud I think it might have something to do with the logging from the maven frontend plugin. The webpack part is written as ERROR instead of INFO
```
[ERROR]  10% building modules 1/4 modules 3 active .../main/webapp/content/scss/global.scssWarning: The 'no-use-before-declare' rule requires type infomation.
[INFO] 
[INFO] [at-loader] Using typescript@2.4.1 from typescript and "tsconfig.json" from /home/deepu/workspace/temp/jh4test/tsconfig-aot.json.
[INFO] 
24231ms building modules                                                                 
[ERROR] 47ms sealing
[ERROR] 0ms optimizing 
[ERROR] 0ms basic module optimization 
[ERROR] 217ms module optimization
[ERROR] 0ms advanced module optimization 
[ERROR] 21ms basic chunk optimization       
[ERROR] 0ms chunk optimization 
[ERROR] 2ms advanced chunk optimization 
[ERROR] 1ms module and chunk tree optimization 
[ERROR] 1ms chunk modules optimization 
[ERROR] 0ms advanced chunk modules optimization 
[ERROR] 10ms module reviving
[ERROR] 0ms module order optimization 
[ERROR] 12ms module id optimization
[ERROR] 3ms chunk reviving 
[ERROR] 1ms chunk order optimization 
[ERROR] 26ms chunk id optimization
[ERROR] 259ms hashing
[ERROR] 0ms module assets processing 
[...]

PR #6125, opened 2017-07-16
Travis: remove travis_wait

pascalgrimaud:
Following the @deepu105's comments [here](https://github.com/jhipster/generator-jhipster/issues/6051#issuecomment-315560699)

> The build either failed or stalled but its reported as passed. Any idea?

I have the same when running in my computer. As @sendilkumarn said, it's more a WARNING than an ERROR and the build finished with 0 code. So for me, it's ok.
Looking our code [here](https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/05-run.sh#L72-L75), we should exit the build when there are error

> @pascalgrimaud not just one most of our recent prod builds are stalled at that point so technically we are never testing the prod AOT builds https://travis-ci.org/jhipster/generator-jhipster/jobs/253721138#L3322

The code [here](https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/05-run.sh#L62-L67) use profile `prod`, for 2 builds:
- [app-default-from-scratch](https://github.com/jhipster/generator-jhipster/blob/master/.travis.yml#L36)
- [app-ng2-psql-es-noi18n](https://github.com/jhipster/generator-jhipster/blob/master/.travis.yml#L38)

> @pascalgrimaud I think it might have something to do with the logging from the maven frontend plugin. The webpack part is written as ERROR instead of INFO

Indeed, the stdout is different. I think the problem comes from the use of `travis_wait`.
I removed it, as it's not necessary. The build takes less than 10min now, as you guys did so a good job :)
See https://travis-ci.org/pascalgrimaud/generator-jhipster/jobs/254089816#L6155
It should be the same output than in local computer
_____

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ install:
 script:
   - $JHIPSTER_SCRIPTS/03-docker-compose.sh
   - $JHIPSTER_SCRIPTS/04-tests.sh
-  - travis_wait 30 $JHIPSTER_SCRIPTS/05-run.sh
+  - $JHIPSTER_SCRIPTS/05-run.sh
   - $JHIPSTER_SCRIPTS/06-sonar.sh
 notifications:
   webhooks:


sendilkumarn:
but even in local we get this warning (in the form of error)

Issue #6124, opened 2017-07-16
Travis production builds are broken


deepu105:
It seems like our Travis production builds are somehow broken. Production builds do work locally but some how they are stalled and terminated in Travis but builds still pass so we never notice any issues.

Some recent samples 

https://travis-ci.org/jhipster/generator-jhipster/jobs/253976729#L3315
 
https://travis-ci.org/jhipster/generator-jhipster/jobs/253721138#L3322

This might or might not have something to do with how the logs are written from the maven tasks, the following writes webpack output as ERROR which is incorrect
```
[ERROR]  10% building modules 1/4 modules 3 active .../main/webapp/content/scss/global.scssWarning: The 'no-use-before-declare' rule requires type infomation.
[INFO] 
[INFO] [at-loader] Using typescript@2.4.1 from typescript and "tsconfig.json" from /home/deepu/workspace/temp/jh4test/tsconfig-aot.json.
[INFO] 
24231ms building modules                                                                 
[ERROR] 47ms sealing
[ERROR] 0ms optimizing 
[ERROR] 0ms basic module optimization 
[ERROR] 217ms module optimization
[ERROR] 0ms advanced module optimization 
[ERROR] 21ms basic chunk optimization       
[ERROR] 0ms chunk optimization 
[ERROR] 2ms advanced chunk optimization 
[ERROR] 1ms module and chunk tree optimization 
[ERROR] 1ms chunk modules optimization 
[ERROR] 0ms advanced chunk modules optimization 
[ERROR] 10ms module reviving
[ERROR] 0ms module order optimization 
[ERROR] 12ms module id optimization
[ERROR] 3ms chunk reviving 
[ERROR] 1ms chunk order optimization 
[ERROR] 26ms chunk id optimization
[ERROR] 259ms hashing
[ERROR] 0ms module assets processing 
[ERROR] 473ms chunk assets processing
[ERROR] 210ms additional chunk assets processing
[ERROR] 0ms recording 
[ERROR] 17048ms additional asset processing
[ERROR] 0ms chunk asset optimization 
[INFO] 
[INFO] [at-loader] Checking started in a separate process...
[ERROR]  94% asset optimization(node:21575) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.
[INFO] 
[INFO] [at-loader] Ok, 1.226 sec.
[INFO] MergetJsonsWebpackPlugin emit...
[...]

PR #6128, opened 2017-07-16
fixing build warning (error :P) - in angularx 

sendilkumarn:
#6124 

### DIFF
--- a/generators/client/templates/angular/_tslint.json
+++ b/generators/client/templates/angular/_tslint.json
@@ -65,7 +65,6 @@
         "no-switch-case-fall-through": true,
         "no-trailing-whitespace": true,
         "no-unused-expression": true,
-        "no-use-before-declare": true,
         "no-var-keyword": true,
         "object-literal-sort-keys": false,
         "one-line": [


deepu105:
I have already fixed it by enabling type check so this is not required anymore

sendilkumarn:
The warning below is in our master 
screen shot 2017-07-16 at 10 13 01 pm

That is fixed by this commit

The warning is 
screen shot 2017-07-16 at 10 12 54 pm

deepu105:
We need `no-use-before-declare` so removing it is not the solution issue is that it needs type checking, and I enabled it and it was working when running `yarn lint` now I added it only for yarn lint step but not for ts-lint loader so I think we need to enable it for ts-lint loader as well

deepu105:
so we need to add `options: { typeCheck: true }` to 
```
{
            test: /\.ts$/,
            enforce: 'pre',
            loaders: 'tslint-loader',
            options: { typeCheck: true },
            exclude: ['node_modules', new RegExp('reflect-metadata\\' + path.sep + 'Reflect\\.ts')]
        }
```

deepu105:
can you try it?

deepu105:
Also since we run tslint on both prod and dev the rule can be moved to common file I guess

sendilkumarn:
Let me have a look

PR #6130, opened 2017-07-16
Typecheck

sendilkumarn:


deepu105:
You might have to add the `skipLibCheck: true` to the tsconfig.aot file for the failure, if that doesnt solve it then we could remove tslint phase from prod build as lint will be done during test phase anyway

sendilkumarn:
Interesting another pain in aot 👎

deepu105:
@sendilkumarn this is making the dev build much slower, I think i'm gonna remove the typecheck on tslint loader

deepu105:
anyway its should be ok as we removed tslint loader from prod

PR #6149, opened 2017-07-19
#6131 Use failsafe for integration tests

### LINKED ISSUE #6131: Enhancement: moving actual backend  end-to end tests to src/it

Actual jhipster generated tests are not unit tests: they are actually integration ones as they are hitting the database and trying to use rest interface.

The Maven convention to these tests is then to:
* Put them in a src/it/java|ts|resource folder
* Prefix them by IT
* Use the maven failsafe plugin instead of the surefire one (so launching them during the integration-test phase)
* Use the integration-test goal of jacoco and reporting them in the parent artefact (that's ok: jhipster is mono module), but **appending** them in an jacoco-it.exec file.
* Using sonar.jacoco.itReportPath variable
* Using the build-helper-maven-plugin to add a new source folder within IDEs

We could generate fully mocked ServiceImpl unit tests in the src/test/xxx folder and End to end test within src/it/...
Also we can configure surefire to avoid scanning any class but the service ones as it should be where all the business logic should be.

We could then have testing coverage for IT and UT tests!

##### **Reproduce the error**

generate a new jhipster project: no src/it folder! 

##### **Related issues**

Not really related, but #6117 speaks about test layout!

##### **Suggest a Fix**

As always, I can make a PR

##### **JHipster Version(s)**

4.6.1

Any

Tcharl:
Fix #6131 


deepu105:
Frankly the old naming convention looks clearer to me. Please look at the comments on the original issue. I'm still not convinced that this is required change

Tcharl:
I thought your +1 smiley means that you were inline with this.

The main added value here is that you can execute 'mvn package' to execute your unit tests so that it's pretty quick to be sure that your 'ServiceImpl' are good enough, while executing 'mvn install' will launch slower tests (IT one) in order to ensure that your End to End stack is working well.

I can always rename the Integration tests name with '<Test>IntTest', but its more a matter of taste: once you're familiar with, prefixing your tests with 'IT' is shorter than suffixing them by 'IntTest' ;-)

deepu105:
@Tcharl lets take the discussion back to the ticket

deepu105:
@Tcharl could you please change the naming convention back to it was as per discussion

Tcharl:
Sure

deepu105:
@Tcharl I think you need to rebase your branch as it shows unrelated old commits from master here.
Also is there a change in behavior with the fail safe plugin? Does `mvn package` and `mvn test` run all tests as earlier?

Tcharl:
Hi @deepu105 , yes, I should rebase that stuff, I sometimes mess with git and its comands...

With this version, 'mvn test' will execute unit tests, while 'mvn verify' will execute integration ones.
I don't think I can configure the behavior to have the integration tests during the 'test' phase:  jacoco coverage would have two agents executed at the same time (the 'unit test' and 'integration test' one).

deepu105:
Then I think this needs thought. Our users are so much used to the
behaviour of running all tests with mvn test, the same happens in our
gradle build as well. While I understand the point not sure if we should
change it. @jdubois what do you think

Tcharl:
As discussed in the [corresponding issue ticket](https://github.com/jhipster/generator-jhipster/issues/6131), @jdubois don't want to disturb actual users: as you say, it will change the behavior of the 'usual' users. On the opposite, it could be considered as an improvement for advanced ones as unit tests can be executed more often (no need to execute integration tests every time), so that we could use IDE TDD driven tools like Infinitest and so on.
I thought to implement a module for this feature in order to avoid disturbing the ones who just want to 'mvn test' everytime.

Concerning the fact that Jhipster don't want to change users' habits, even if I understand the pros (grabbing a larger community, don't wasting time reading the doc to understand how it works, ...), my vision diverge for this aspect: IMHO JHipster should be targeted to 'Hipsters', so the ones who want to have the best framework ever ;-) (even I Jhipster already is that one^^). But again, it's a matter on vision and strategy, so I'm totally with your POV guys :-).

deepu105:
@Tcharl even though we started out as something for Hipsters, now our majority of users are normal people who just want to accelerate development and use complex integrations without spending too much time worrying about the implementation details. Hence we make sure that while still add hip technologies we also keep it simple and usable by normal people. Anyways hipsters will always find stuff not hip enough :smiling_imp: 
Sorry that you spend a lot of time on this and sorry if we weren't helpful. I guess a module would be the best solution here.  Would you be ok to close this then?

Tcharl:
I don't lost my time, as this will be the basis for the module ;-), also it helped the team to choose wether or not using failsafe within the base framework, so no loss at all. No problem on closing this, this is a pleasure to learn the yo internals and Jhipster always saved me more time than coding a simple PR :p.
Please close the PR without any regret, and feel free to reopen it once the users will be Hipster enough ;-).

It's always a pleasure to help such wonderful projects with such smart guys.

PR #6164, opened 2017-07-24
Fixing error when create an model from jdl file when selecting Indonesia Language

mirzaakhena:
After release v4.6.2 i found that jhipster fail to generate jdl entity because it cannot find the _entity_id.json file.

```
Error jhipster:import-jdl model.jdl 

ERROR! Copying template client/i18n/_entity_id.json failed. [Error: ENOENT: no such file or directory, open '/MY_PROJECT/node_modules/generator-jhipster/generators/entity/templates/client/i18n/_entity_id.json']
```


pascalgrimaud:
Thanks, there is some missing tests, I'm working on it

PR #6175, opened 2017-07-27
Upgrade the JHipster Console to v2.2.1

PierreBesson:
Lots of work here. I would really appreciate if someone could review it. Kibana and Zipkin should work out of the box which is no small feat (remember to package the app with the zipkin profile).


pascalgrimaud:
@PierreBesson : I try to do my best to review and test it, but no promise :smile:

pascalgrimaud:
LGTM

Issue #6072, opened 2017-07-10
Timestamp changes after save an entity (from "12:** AM" to "12:** PM")


Akuka:
##### **Overview of the issue**
After saving an entity (from the UI) with a timestamp, if the time is "12:** **AM**", after saving the entity the time will be "12:** **PM**", even if you changed other filed of the entity.

** ****EDIT**** **
The issue is more general:
every timestamp which presents as "PM" becomes "AM" when you try to edit the entity. once you will click on save (even if you didn't touched the time), the new time (as AM) will persist to the database.

##### **Reproduce the error**
Generate an entity with a ZonedDateTime filed, and one more other filed.
Run the web and create a new entity with timestamp of "12:** **AM**" ( 00:** ).
Edit the entity, change something in the other filed (do not change the timestamp), save the entity.
The timestamp will change to "12:** **PM**"

##### **JHipster Version(s)**

```
customer@0.0.0 /Users/xxx/dev/jhipster/xxx/xxx
└── generator-jhipster@4.6.0 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.xxx.customer"
    },
    "jhipsterVersion": "4.6.0",
    "baseName": "customer",
    "packageName": "com.xxx.customer",
    "packageFolder": "com/xxx/customer",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "clientFramework": "angularX",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
</pre>
</details>

##### **Environment and Tools**

java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

[...]

PR #6181, opened 2017-07-30
Fix issue #6072 - Fix the wrong date pattern

Akuka:
The problem is the wrong pattern that used for set up the time ("hh" instead "HH")
In addition, today there is no ability to edit the seconds, so I will also add "ss" to the pattern.
This is how the pattern should look like: yyyy-MM-ddTHH:mm:ss


Akuka:
I submitted also a PR on the ng-jhipster project that will parse the seconds part.
https://github.com/jhipster/ng-jhipster/pull/43

deepu105:
could you please explain the use case and what exactly you are trying to do? may be with an example?

Akuka:
Just try to edit and save (even without change) any entity with a ZonedDateTime field which the timestamp that is "PM" (For example: 14:14:14, which displayed as 2:14:14 PM), and see yourself how it will change to "AM" (For Example: 02:14:14, which displayed as 2:14:14 AM).

Akuka:
You can find the patterns on the Angular website:
https://docs.angularjs.org/api/ng/filter/date

> 'HH': Hour in day, padded (00-23)
> 'hh': Hour in AM/PM, padded (01-12)

The pattern was wrong.

Issue #6159, opened 2017-07-22
Angular 4 - prod build performance


jdubois:
Doing a perf audit shows our Angular 4 prod build is slower and bigger than our AngularJS 1 prod build.
This comes from our main js bundle:
- it's very big (500kb) without GZipping (gzip works really well on it, which shows there's a problem !!)
- time spent executing it is too long (and that's because it's really big)

Opening it shows our compiled components in a definitely non efficient format: it works, but it's big and slow.

I guess this comes from our AOT configuration.

deepu105:
Julien ofcourse the bundle size would be more than ng1 as we have lot more
js code than ng1 in ng2, and the libraries now are huge as well. but
shouldn't be very drastically huge. When i checked last time it was descent
and loading was pretty fast. Ill check again

delverdev:
Did you have an idea to separate user and admin`s parts? Yes, perhaps there is no much code in the admin part, but some projects can have libraries that will be used only by admins (charts libraries, datatables etc.). It can slow users that don't need them. Also, some admin panel would be more convenient and productive (dropdowns take time, a vertical menu would be great). *-management pages have much code and html that can slow a loading too.

jdubois:
Yes, but this is another issue

deepu105:
@jdubois lets track this in #5833 I'll add a line item there

deepu105:
@jdubois also what did you mean by `Opening it shows our compiled components in a definitely non efficient format: it works, but it's big and slow.`?

jdubois:
If you open up the "main.js", you'll see the generated code is really verbose, here's a screenshot below.
I guess there's an issue in the aot configuration.
This is why the file is so big, gets compressed so well (lots of duplicate stuff in it), and why the "scripting" performance of our prod build is so bad.

<img width="1440" alt="screenshot 2017-07-21 14 50 50" src="https://user-images.githubusercontent.com/316835/28747501-3d8660a6-74a0-11e7-9e5a-fb9986dcf07f.png">

deepu105:
When I tested the prod profile I didn't find its performance bad, and yes
the generated code is verbose due to AOT, but its minified. I assume your
browser formatted it. Anyway let me check again

jdubois:
yes I clicked on the `{}` icon to format it, still it isn't minified correctly, and it's really bad. It should be 20-30kb only

deepu105:
Ok ill check this tomorrow

atomfrede:
[...]

PR #6185, opened 2017-07-31
webpack optimizations

deepu105:
For #6159

This change improves the bundle size greatly and I think this is the best that we can get out of this. I tried many different combinations. but there is one strange problem. The home page image is not generated but the logo image is generated. I couldn't figure out why, or I'm just sleepy. SO if someone can figure that out then we can merge this @sendilkumarn wanna give it a shot?

cc @jdubois 

image

image

image


deepu105:
This also improved the build speed, saves atleast 10 seconds due to parallel uglify

sendilkumarn:
I will assign it to me. I will review it.

sendilkumarn:
I will assign it to me. I will review it.

deepu105:
@sendilkumarn please note I moved the uglify related change to master as the issue causing part is only ngc-loader in this PR

generators/client/templates/angular/webpack/_webpack.prod.js
```diff
@@ -65,6 +65,19 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
             ],
             exclude: ['node_modules/generator-jhipster']
         },
+        {
+            test: /\.ts$/,
+            use: [
+                {
+                    loader: 'ngc-webpack',
+                    options: {
+                        disable: false,
```
sendilkumarn: this is not needed. This will run the `ngcwebpack` in ghost mode. Remove it.

generators/client/templates/angular/webpack/_webpack.prod.js
```diff
@@ -65,6 +65,19 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
             ],
             exclude: ['node_modules/generator-jhipster']
         },
+        {
+            test: /\.ts$/,
```
sendilkumarn: Actually this entire thing is not needed.

sendilkumarn: Remove this.

generators/client/templates/angular/webpack/_webpack.prod.js
sendilkumarn: remove disabled & resource override in below plugin.

sendilkumarn:
@jdubois `500kb` you are talking about the entire bundle or main.js alone ?

jdubois:
It was the main bundle

generators/client/templates/angular/webpack/_webpack.prod.js
deepu105: disabled: true will run it in ghost mode (Ie it wont do anything, here disabled is false so should run loader)
Similarly resource override is to remove html strings as with AOT they are not required.

generators/client/templates/angular/webpack/_webpack.prod.js
deepu105: This was added to optimize the resource further. This loader removes the TS annotations and stuff which are not required for AOT. This loader produces a main bundle size of 388kb for below config, without this the size would be around 480 or something
```
{
  "generator-jhipster": {
    "promptValues": {
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.6.2",
    "baseName": "jh4test",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": true,
    "jwtSecretKey": "afc8cb1c9375cae7b67dac3ba1abc4599d9e2d35",
    "useSass": true,
    "testFrameworks": [
      "protractor"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "zh-tw",
      "fa",
      "ar-ly",
      "nl",
      "et",
      "fr",
      "hi"
    ]
  },
  "generator-jhipster-entity-audit": {
    "auditFramework": "custom"
  }
}
```

deepu105:
@sendilkumarn read my comments on #6159

generators/client/templates/angular/webpack/_webpack.prod.js
sendilkumarn: and of course it is `disabled` not `disable`

sendilkumarn:
``` js 
/**
   * If false the plugin is a ghost, it will not perform any action.
   * This property can be used to trigger AOT on/off depending on your build target (prod, staging etc...)
   *
   * The state can not change after initializing the plugin.
   * @default true
   */
```

from ngcwebpack plugin

sendilkumarn: Added to that for `resource override`, we have given empty string which again results in false. So we are not using resource override currently.

sendilkumarn:
@deepu105 I saw that already. I don't see any optimisation. with or without this bundle size is still at 401kb.

generators/client/templates/angular/webpack/_webpack.prod.js
sendilkumarn:
```js
/*
 * ### resourceOverride and assets
   * If you reference assets in your styles/html that are not inlined and you expect a loader (e.g. url-loader)
   * to copy them, don't use the `resourceOverride` feature as it does not support this feature at the moment.
*/
```

Since we are using styles and html that are not inlined, this feature cannot be used.

deepu105:
Ok let me check


Thanks & Regards,
Deepu

generators/client/templates/angular/webpack/_webpack.prod.js
deepu105: For the loader it is `disable` not `disabled` for the plugin its `disabled` https://github.com/shlomiassaf/ngc-webpack/blob/master/src/aot-clean-transformer/loader/text-based-loader/loader.ts#L197

deepu105: and yes we are not using resource override for the same reason as we have image reference in CSS

deepu105:
I spend some time on this again but couldn't solve the issue. It's still mysterious that `logo-jhipster.png` file gets copied correctly while the other images are skipped. All images are in fact referred from sass files in the same manner so I don't understand why it works for one file but fails for another.
So i'm giving up on this. This saves around 120KB on the main js bundle though.
@sendilkumarn if you wanna try to get the image issue working go ahead else lets close this

jdubois:
Maybe this comes from the image format? Don't try to reduce the image sizes, just copy them: this is OS specific (so it will cause lots of bug reports) and for me images should already be optimized here (for me the original format is PSD, SVG or similar, not PNG or JPEG)

generators/client/templates/angular/webpack/_webpack.prod.js
sendilkumarn: @deepu105 But we have given `resourceoverride`in plugin with an empty value. we can remove it.

sendilkumarn:
currently, we are getting 401kb and I guess you were getting around 388kb without image, It is more or less same. We can close this...

deepu105:
Sendil no, currently we get 500± kb without this, try with the same yo rc i
used

generators/client/templates/angular/webpack/_webpack.prod.js
deepu105: It doesn't matter for this issue

deepu105:
@jdubois we already removed image optimizations. So i dont think its related to that. I tried different formats it doesn't work

jdubois:
I just tested and I confirm that this code from @deepu105 does work, and improve both the main bundle size and the performance.
I think I can further improve it, if I use the latest Angular CLI configuration -> let me try this, and please @sendilkumarn @deepu105 have a look at my final code

jdubois:
Quite surprisingly, this works better (smaller size, better performance) than using the build generated by Angular CLI.
-> so for me we should merge this
-> @sendilkumarn I see you don't agree everywhere, but I can guarantee you this works
-> also, I'll upgrade TypeScript, Webpack, etc so we have the latest versions (everything runs fine on my box)

jdubois:
@deepu105 I have images working for:

- CSS build in dev and prod mode
- SaSS build in dev and prod mode

By "images", I mean: the big image on the home page, the logo on the top left navbar, and the fontawesome icons.

So that looks OK to me.

deepu105:
@jdubois thats great now I'm curious, did you upgrade any libs? coz the big image on home page was not working when i did the PR

jdubois:
Yes that's the commit I did just after... But now I found some other issue with session-based authent in prod mode, I'm on it...

sendilkumarn:
wow this is interesting. let see what was changed in the libs that made this happen

jdubois:
To be honest I forgot about the comment, and it just worked out of the box :-)
(and I know session based authent fails, I'm currently fixing it)

Issue #5346, opened 2017-03-02
On error during entity update, save button is not enabled back


rakeshkalra1983:
##### **Overview of the issue**

When editing an entity, on pressing save button, if error is received from the server save button is not enabled back.
I believe this happens only when enableTranslation is false, as I don't see this issue in Jhipster Sample application from git.

##### **JHipster Version(s)**

Jhipster Version 4.0.7

##### **JHipster configuration**

{
  "generator-jhipster": {
    "jhipsterVersion": "4.0.7",
    "baseName": "test",
    "packageName": "com.test",
    "packageFolder": "com/test",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": true,
    "jwtSecretKey": "",
    "clientFramework": "angular2",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}

##### **Browsers and Operating System**

Browser: Chrome & Firefox
OS: Mac OS

- [x] Checking this box is mandatory (this is just to show you read everything)

mraible:
FWIW, Dennis and I saw this in our workshop last week too.

wmarques:
@rakeshkalra1983 
Thanks for the report !
Can you do a PR ? (same question for your other issues :smile: )

ruddell:
One cause of this is that Angular doesn't have support for min/max validation on the number input...

wmarques:
Save button isn't enabled because when we catch the error, we do a `res.json()`, but sometimes it's just a text message (for example when we didn't start spring boot), so we should catch the exception raised by res.json() and then do a `res.text()` instead.
I don't know if we should do this in the service or in the onSave method... What do you think guys ?

PR #6207, opened 2017-08-07
Remove redundant and broken error handling when saving entities

stieler-it:
Error handling seems to be unnecessary at this point:
1. There is a global HTTP error handling routine in alert-error.component.ts
2. The local error handling seems to be broken because if the error response contains json (which is the regular case), response.message stays undefined, so there is no message to show
If there is no specific reason for save error handling in the dialog component, I would propose to remove the code for now.

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
@@ -148,7 +148,7 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
 
     private subscribeToSaveResponse(result: Observable<<%= entityAngularName %>>) {
         result.subscribe((res: <%= entityAngularName %>) =>
-            this.onSaveSuccess(res), (res: Response) => this.onSaveError(res));
+            this.onSaveSuccess(res), (res: Response) => this.onSaveError());
     }
 
     private onSaveSuccess(result: <%= entityAngularName %>) {
@@ -157,17 +157,11 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
         this.activeModal.dismiss(result);
     }
 
-    private onSaveError(error) {
-        try {
-            error.json();
-        } catch (exception) {
-            error.message = error.text();
-        }
+    private onSaveError() {
         this.isSaving = false;
-        this.onError(error);
     }
 
-    private onError(error) {
+    private onError(error: any) {
         this.alertService.error(error.message, null, null);
     }
     <%_


ruddell:
This was originally added because of https://github.com/jhipster/generator-jhipster/issues/5346

I don't normally use Angular so I'm not sure if that issue is still present

stieler-it:
Hi @ruddell, my PR keeps the failure handler for this reason and sets `this.isSaving = false;` which will re-enable the save button. I suggest to remove the acutal error handling because it does not seem to be used anyway.

ruddell:
I was just giving a link to why it was originally added. Again, I'm not the best reviewer for this as I don't use Angular

With this change, is `onError` ever called?

stieler-it:
Yes, it is. However, it's not obvious since the relevant code is created in [another file](https://github.com/jhipster/generator-jhipster/blob/41875056faa000c20a8712a7bfd92f30e50f2a5b/generators/generator-base-private.js#L527).

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -157,17 +157,11 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
         this.activeModal.dismiss(result);
     }
 
-    private onSaveError(error) {
-        try {
-            error.json();
-        } catch (exception) {
-            error.message = error.text();
```
deepu105: I think this is to handle some cases where the message is text instead of json so that when the error is intercepted in global handle the `error.message` attribute has some value if we are removing this then we need to make sure similar handling is done in global error handler

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
```diff
@@ -157,17 +157,11 @@ export class <%= entityAngularName %>DialogComponent implements OnInit {
...
-        }
+    private onSaveError() {
         this.isSaving = false;
-        this.onError(error);
```
deepu105: Are you seeing error message in the screen after removing this? I need to find time to test this

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
stieler-it: The broken thing is.. error is of type Response, which does not even have a .message property. So the usual case - server sends JSON response - will call the alertService with "undefined" instead of a message. So this code would actually just work in the irregular case, when there is no JSON object in the server response.

generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.ts
stieler-it: Yes, it is shown by the global error handler - this code would never show it (see comment above). I am currently not sure about the "non JSON" case but in either case it should be handled at the same place.

jdubois:
This indeed makes sense to me... And we should focus on the JSON response, we're supposed to do REST API calls.
I'm testing this and will merge if it's OK

jdubois:
LGTM

PR #6213, opened 2017-08-09
Console log forwarding improvements

PierreBesson:

Some updates for the integration with the JHipster Console:

- Fix bad InstanceId forwarded to the Console (see https://github.com/jhipster/jhipster-console/issues/55)
- Add artifact version to logs forwarded to the console (for visibility when doing canaries)
- Add JvmAttributeGaugeSet to Dropwizard so that we can have the "uptime" metrics (planned for the registry).

And most important feature:
- When metrics logs are enabled, they are now filtered from the console and file appenders and only sent to logstash !
I'm sure this was a blocker for a lot of people.

### DIFF
--- a/generators/server/templates/gradle/_profile_dev.gradle
+++ b/generators/server/templates/gradle/_profile_dev.gradle
@@ -69,19 +69,23 @@ processResources {
             it.replace('#logback.loglevel#', logbackLoglevel)
         }
     }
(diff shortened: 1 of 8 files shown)


PierreBesson:
I've finally fixed the gradle tests. Can someone review it. It would be great if this could be released at the same time as the new console.

PR #6217, opened 2017-08-10
Add empty option for dynamically populated select boxes

### LINKED ISSUE #6148: It is not possible to select the first entry of a ManyToOne relationship in Internet Explorer 11

When creating an entity A that has a *required* ManyToOne relationship to entity B, it is not possible to select an instance of type B while creating a new entity A - or at least not directly (see reproduction).

In case 1 this is a blocker, in case 2 it degrades the usability.

##### **Reproduce the error**

1. If there is only one entity of type B, IE shows this one as selected when the dialog is opened. You can open the select box and select it again, but it does not update the model, so the required field is still empty. *It is not possible to create an entity of type A in this case!*
2. If there are more entity, it is not possible to select the first (selected) one directly. However, you can select another one, and then switch back to the first one to finally select it.

Case 1 (Validation fails although item is selected)
ie-selected-required

##### **Related issues**

I did not find a related issue.

##### **Suggest a Fix**

A possible fix would be to *always* include an empty option element, even if the relationship is required. It is still not possible to save an invalid entity. This seems to be valid since the current initial state (an option is selected that does not even exist), could be wrong anyway.

##### **JHipster Version(s)**

4.5.3

[jhipster-info.out.txt](https://github.com/jhipster/generator-jhipster/files/1158932/jhipster-info.out.txt)

Seems to be already included in the info file.

Windows, IE11. The same application works on Chrome and Firefox.

stieler-it:
Workaround for an Angular / IE11 issue. Empty options are removed as soon as a real option has been selected.

fixes #6148

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -203,6 +203,8 @@ _%>
             <select class="form-control" id="field_<%= relationshipName %>" name="<%= relationshipName %>" [(ngModel)]="<%= entityInstance %>.<%=relationshipFieldName %>" <% if (relationshipRequired) { %> required<% } %>>
                 <%_ if (!relationshipRequired) { _%>
                 <option [ngValue]="null"></option>
+                <%_ } else { _%>
+                <option *ngIf="!editForm.value.<%= relationshipName %>" [ngValue]="null" selected></option>
                 <%_ } _%>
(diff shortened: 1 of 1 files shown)


stieler-it:
Travis tests are failing. I tried to reproduce it locally with the ng2-default app - however, the line I changed seems not to occur in the generated sample. Re-running the tests leads to a different result. Any advice?

deepu105:
@jhipster/developers can some one test this is chrome and FF

sendilkumarn:
It is working 👍

sendilkumarn:
Thanks @stieler-it

PR #6219, opened 2017-08-11
JHipster CLI now runs project local module

### LINKED ISSUE #6208: CLI should run local generator-jhipster rather than global one

`jhipster` command runs globally installed generator-jhipster rather than project locally installed generator-jhipster. `yo jhipster` works as expected.

I tested only on Windows 10, not sure this is the same for other OS.

I'm not sure this happens also for npm.

The steps to reproduce are rather easy.

Maybe this is something specific to my environment but it explains why I could not get rid of the tabtab error.

This could impact 2 kinds of users:
- JHipster developers or advanced users who use master and `yarn link`.
- users who run different versions of JHipster in different projects

##### **Reproduce the error**

- Install generator-jhipster 4.6.1 globally `yarn global add generator-jhipster@4.6.1`
- in a project, install generator-jhipster 4.6.2 locally with `yarn add generator-jhipster@4.6.2`
- execute `jhipster`, it will print 4.6.1 
- execute `yo jhipster`, it will print 4.6.2 

In my environment, `jhipster.CMD` is located in `C:\Users\Gaël\AppData\Local\Yarn\config\global\node_modules\.bin` folder.

It runs `C:\Users\Gaël\AppData\Local\Yarn\config\global\node_modules\generator-jhipster\cli\jhipster.js`

It should run the project's local version of generator-jhipster, in my case `e:\projets\issues\clarity\node_modules\generator-jhipster\cli\jhipster.js`

##### **Suggest a Fix**

jhipster command points to `C:\Users\Gaël\AppData\Local\Yarn\bin\jhipster.CMD`
~~~
@"%~dp0\..\config\global\node_modules\.bin\jhipster.cmd"   %*
~~~
which points to `C:\Users\Gaël\AppData\Local\Yarn\config\global\node_modules\.bin\jhipster.cmd`
~~~
@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe"  "%~dp0\..\generator-jhipster\cli\jhipster.js" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "%~dp0\..\generator-jhipster\cli\jhipster.js" %*
)
~~~

`%~dp0` always point to global directory.

The test on node.exe existence will never be true for yarn on Windows.

All the node commands I checked are written the same way, so this is quite strange.

##### **JHipster Version(s)**

I suppose since CLI was introduced.

java version "1.8.0_141"
Java(TM) SE Runtime Environment (build 1.8.0_141-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.141-b15, mixed mode)

git version 2.12.0.windows.1

node: v6.11.1

npm: 3.10.10

gulp:
[01:11:10] CLI version 1.4.0

yeoman: 2.0.0

yarn: 0.27.5

Docker version 17.06.0-ce, build 02c1d87

docker-compose version 1.14.0, build c7bdf9e3

Windows 10 Pro.

gmarziou:
`cli/jhipster.js` now detects whether there's a project local version of `cli/cli.js`so that it is loaded in priority over globally installed version.

Fix #6208

### DIFF
--- a/cli/jhipster.js
+++ b/cli/jhipster.js
@@ -21,6 +21,7 @@
 const semver = require('semver');
 const packageJson = require('../package.json');
 const logger = require('./utils').logger;
+const path = require('path');
 
 const currentNodeVersion = process.versions.node;
 const minimumNodeVersion = packageJson.engines.node;
@@ -33,4 +34,28 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
     /* eslint-enable  */
     process.exit(1);
 }
-require('./cli');
+
+requireCLI();
+
+/*
+ * Require cli.js giving priority to local version over global one if it exists.
+ */
+function requireCLI() {
+    /* eslint-disable global-require */
+    try {
+        const localCLI = require.resolve(path.join(process.cwd(), 'node_modules', 'generator-jhipster', 'cli', 'cli.js'));
+        if (__dirname !== path.dirname(localCLI)) {
+            // load local version
+            /* eslint-disable import/no-dynamic-require */
+            logger.info('Using JHipster locally installed in current project\'s node_modules');
+            require(localCLI);
+            return;
+        }
+    } catch (e) {
+        // Unable to find local version, so global one will be loaded anyway
+    }
+    // load global version
+    logger.info('Using JHipster globally installed');
+    require('./cli');
+    /* eslint-enable  */
+}


deepu105:
Similar solution can be used for composing Jhipster modules as well i guess so people dont have to install them locally

gmarziou:
> May be we should print a log here specifying which cli is being used. 

OK

> Would this affect when doing npm link?

yarn link works fine, it's only a link so resolve() does not see the difference  with a "real" local module.
So npm link should be fine also.

> Is it possible to add a test for this?

It sounds complex because we must install jhipster globally, so I guess using a docker container could work.

gmarziou:
It should be fine now.

PR #6223, opened 2017-08-13
jhipster upgrade should load global version

gmarziou:
This is a follow-up of #6219.

Using local version of upgrade.js could have bugs that have been fixed in more recent version, this is why global version is better for upgrading.

Ref #6208

### DIFF
--- a/cli/jhipster.js
+++ b/cli/jhipster.js
@@ -35,24 +35,33 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
     process.exit(1);
 }
 
-requireCLI();
+let preferGlobal = false;
+if (process.argv.length >= 3) {
+    if (process.argv[2] === 'upgrade') {
+        // Prefer global version for `jhipster upgrade` to get most recent code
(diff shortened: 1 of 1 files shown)


cli/jhipster.js
```diff
@@ -35,24 +35,33 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
     process.exit(1);
 }
 
-requireCLI();
+let preferGlobal = false;
+if (process.argv.length >= 3) {
```
deepu105: I think since we use commander it already would have all arguments passed so may be you could use the commander api for this. Im on my phone so i cant point to the exact api now

gmarziou: Yes I saw commander use in cli.js and thought about extracting some common code but it could make things even more complex because whatever file I put this common code in I might have to choose between local and global version. So it might be simpler to have jhipster.js depends as little as possible on other files.

deepu105: oh ok, makes sense

PR #6229, opened 2017-08-14
Support for API-first dev using swagger-codegen

cbornet:

To use/test:
* write a swagger spec in `src/main/resources/api.yml` (eg : copy [the swagger petstore](https://raw.githubusercontent.com/swagger-api/swagger-codegen/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml))
* compile or generate-sources
* implement the interfaces and override the default methods. eg:
```java
@RestController
@RequestMapping("/api/petstore")
public class PetStore implements PetApi {

    @Override
    public ResponseEntity<List<Pet>> findPetsByStatus(@RequestParam List<String> status) {
        return new ResponseEntity<>(
            status.stream()
                .distinct()
                .map(Pet.StatusEnum::fromValue)
                .map(statusEnum -> new Pet().id(RandomUtils.nextLong()).status(statusEnum))
                .collect(Collectors.toList()),
            HttpStatus.OK);
    }
}
```

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -154,6 +154,11 @@ module.exports = JhipsterServerGenerator.extend({
                 this.messageBroker = false;
             }
 
+            this.apiFirst = this.config.get('apiFirst') === 'no' ? false : this.config.get('apiFirst');
+            if (this.apiFirst === undefined) {
+                this.apiFirst = false;
+            }
+
             this.serviceDiscoveryType = this.config.get('serviceDiscoveryType') === 'no' ? false : this.config.get('serviceDiscoveryType');
             if (this.serviceDiscoveryType === undefined) {
                 this.serviceDiscoveryType = false;
@@ -259,6 +264,11 @@ module.exports = JhipsterServerGenerator.extend({
                     this.serviceDiscoveryType = false;
                 }
 
+                // If the service discovery is not defined, it is disabled by default
+                if (this.apiFirst === undefined) {
+                    this.apiFirst = false;
+                }
+
                 // If translation is not defined, it is enabled by default
                 if (this.enableTranslation === undefined) {
                     this.enableTranslation = true;
@@ -304,6 +314,7 @@ module.exports = JhipsterServerGenerator.extend({
             this.configOptions.serviceDiscoveryType = this.serviceDiscoveryType;
             this.configOptions.buildTool = this.buildTool;
             this.configOptions.enableSocialSignIn = this.enableSocialSignIn;
+            this.configOptions.apiFirst = this.apiFirst;
             this.configOptions.authenticationType = this.authenticationType;
             this.configOptions.uaaBaseName = this.uaaBaseName;
             this.configOptions.serverPort = this.serverPort;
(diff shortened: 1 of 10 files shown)


deepu105:
@cbornet I suggest to make this opt in by adding it as an option in question `Which other technologies would you like to use?` else this will just add more bloat and I'm sure there will be a lot of people who will not use it all the time

cbornet:
Of course, no problem !

cbornet:
It's now an option

deepu105:
Great

generators/server/index.js
```diff
@@ -154,6 +154,11 @@ module.exports = JhipsterServerGenerator.extend({
                 this.messageBroker = false;
             }
 
+            this.apiFirst = this.config.get('apiFirst') === 'no' ? false : this.config.get('apiFirst');
```
deepu105: can you use a specific option name something like `enableSwaggerCodeGen` and make it boolean instead of string so that we could just say `swaggerCodeGen: true` and it will be much easier and cleaner especially in EJS

generators/server/index.js
```diff
@@ -259,6 +264,11 @@ module.exports = JhipsterServerGenerator.extend({
                     this.serviceDiscoveryType = false;
                 }
 
+                // If the service discovery is not defined, it is disabled by default
+                if (this.apiFirst === undefined) {
```
deepu105: wouldn't this be set already above seems redundant and by making it boolean such defaulting will be unnecessary anyway

generators/server/index.js
cbornet: The idea is to have it extensible for future possible technologies. Like we have for searchEngine, messageBroker, ...
Swagger has the most hype ATM, but that could change.

generators/server/index.js
cbornet: I copy/pasted the behavior from serviceDiscoveryType (even the comment, I realize, ...) but you're probably right.

generators/server/index.js
deepu105: Even if extensible in future its still more manageable to have one flag per option as it keeps ejs more readable and easy to manage. I wanted to change existing options this way and already did that for test frameworks

cbornet: Just to be sure : do I keep a string option in `.yo-rc.json` that is transformed to a boolean flag for the templates ?

deepu105: that might be more complex so for the moment lets just have it as boolean similar to searchEngine option. If we ever add another apiFirst stuff we could reconsider or just add another boolean. In any case booleans are leaner and easy to maintain

jdubois:
This will be in our next release, but this is not easy to understand for most people: @cbornet could you document this on jhipster.github.io ?

cbornet:
I know I have to document this but I'm AFK until September 30th :sweat_smile: . Can it wait until then ?

deepu105:
May be we should add one more checkbox in PR template for docs :P


Thanks & Regards,
Deepu

On Tue, Aug 22, 2017 at 7:14 PM, Christophe Bornet <notifications@github.com

PierreBesson:
I will write the docs on it getting inspiration from @cbornet's module readme.
Agree to add the checkbox for documentation.

jdubois:
Great, thanks (and no hurry!!)

tezcane:
Was the swagger-codegen doc made yet?   I couldn't find it.  I am considering for my next project.

atomfrede:
It is here: http://www.jhipster.tech/doing-api-first-development/

PR #6235, opened 2017-08-15
Replace directive with translate pipe to fix IE rendering issue

### LINKED ISSUE #6147: Select Option values not shown in Internet Explorer 11

When creating an entity that references an Enum, the option values are not shown in IE.
ie-listen-leer

It makes our application unusable in IE which will be a main browser in our user base.

##### **Reproduce the error**

Generate an application using a JDL containing an Enum (see below).

Open the applicatoin in IE11, Entities -> JobHistory, create a new. The select box shows two empty entries. The same application works in Chrome and FF.

##### **Related issues**

Nothing found for jhipster, however it could to be related to:
https://stackoverflow.com/questions/12942681/how-to-fix-ie-select-issue-when-dynamically-changing-options

##### **Suggest a Fix**

See SO ticket for some hacks, also using Angular's translate pipe instead of the JHipster component seems to fix the issue.

Does not work:
`<option value="M" jhiTranslate="babsappApp.Geschlecht.M">M</option>`
Works:
`<option value="M">{{'babsappApp.Geschlecht.M' | translate}}</option>`

Surprisingly (ok, actually not so much for IE), it is enough to change the *first* option to the pipe variant. Even if there is a second option still using the jhiTranslate, it is still fixed.

##### **JHipster Version(s)**

JHipster 4.5.3 but I was also able to reproduce it with the latest version (from last week).

[jhipster-info.out.txt](https://github.com/jhipster/generator-jhipster/files/1158801/jhipster-info.out.txt)

The critial part seems to be the enum, e.g.:
```
entity JobHistory {
	startDate ZonedDateTime,
	endDate ZonedDateTime,
	language Language
}

enum Language {
    FRENCH, ENGLISH, SPANISH
}
```

Windows, IE 11

stieler-it:
Internet Explorer 11 seems to have a problem with dynamically filled select boxes. The jhiTranslate directive seems to be evaluated too late in the lifecycle, so IE shows only blank options. It works with Angular's translate pipe, though.

Fixes #6147

### DIFF
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -62,7 +62,7 @@ _%>
                 const values = fields[idx].fieldValues.replace(/\s/g, '').split(',');
                 for (key in values) {
                     const value = values[key]; _%>
-                <option value="<%= value %>">{{'<%=enumPrefix%>.<%=value%>' | translate}}</option>
+                <option value="<%= value %>"><% if (enableTranslation) { %>{{'<%=enumPrefix%>.<%=value%>' | translate}}<% } else { %><%=value%><% } %></option>
                 <%_ } _%>
             </select>
             <%_ } else { _%>
--- a/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -62,7 +62,7 @@ _%>
                 const values = fields[idx].fieldValues.replace(/\s/g, '').split(',');
                 for (key in values) {
                     const value = values[key]; _%>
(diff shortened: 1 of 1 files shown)


generators/entity/templates/client/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
```diff
@@ -62,7 +62,7 @@ <h4 class="modal-title" id="my<%= entityClass %>Label" jhiTranslate="<%= keyPref
                 const values = fields[idx].fieldValues.replace(/\s/g, '').split(',');
                 for (key in values) {
                     const value = values[key]; _%>
-                <option value="<%= value %>" jhiTranslate="<%=enumPrefix%>.<%=value%>"><%= value %></option>
+                <option value="<%= value %>">{{'<%=enumPrefix%>.<%=value%>' | translate}}</option>
```
deepu105: you need to handle case when translation is disabled as well. For that `<%= value %>` alone is required

stieler-it: Fixed in new commit

PR #6249, opened 2017-08-18
Improve output of jhipster upgrade

### LINKED ISSUE #6237: jhipster upgrade should be verbose by default

Currently, there is a `--verbose` option for `jhipster upgrade`, I think it should be the default mode because it gives more information to the user about what is being done to her git repository.
Also, some operations like npm install can take long time to complete and when output is muted user has no idea whether things are running smoothly or process is hung for some reason and may never complete.
In case of failure, user may have a clue on what went wrong and maybe fix it by herself.

So, I propose:
- to set verbose mode by default
- to remove `--verbose` option and maybe to add `--silent` or `--quiet` 
- to display the exact command that is currently being executed (git, npm, yarn, ...)

gmarziou:
This will ease debugging `jhipster upgrade` issues.

- Use log.ok() standard method in success() logging method
- Skip install if there are conflicts in `package.json` file
- List the conflicting files  at the end if any
- Set verbose output by default
- Add a `--silent`option

Fix #6237

### DIFF
--- a/cli/jhipster.js
+++ b/cli/jhipster.js
@@ -17,7 +17,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 const semver = require('semver');
 const packageJson = require('../package.json');
 const logger = require('./utils').logger;
@@ -35,21 +34,23 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
     process.exit(1);
 }
 
-let preferGlobal = false;
-if (process.argv.length >= 3) {
-    if (process.argv[2] === 'upgrade') {
-        // Prefer global version for `jhipster upgrade` to get most recent code
-        preferGlobal = true;
-    }
+let preferLocal = true;
+
+// Don't use commander for parsing command line to avoid polluting it in cli.js 
+// --prefer-local: Always resolve node modules locally (useful when using linked module)
+if (process.argv.includes('upgrade') && !process.argv.includes('--prefer-local')) {
+    // Prefer global version for `jhipster upgrade` to get most recent code
+    preferLocal = false;
 }
-requireCLI(preferGlobal);
+
+requireCLI(preferLocal);
 
 /*
  * Require cli.js giving priority to local version over global one if it exists.
  */
-function requireCLI(preferGlobal) {
+function requireCLI(preferLocal) {
     /* eslint-disable global-require */
-    if (!preferGlobal) {
+    if (preferLocal) {
         try {
             const localCLI = require.resolve(path.join(process.cwd(), 'node_modules', 'generator-jhipster', 'cli', 'cli.js'));
             if (__dirname !== path.dirname(localCLI)) {
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1575,12 +1575,15 @@ module.exports = class extends PrivateBase {
         }
         if (options.async === undefined) options.async = true;
         if (options.silent === undefined) options.silent = true;
+        if (options.trace === undefined) options.trace = true;
 
         if (!Array.isArray(args)) {
             args = [args];
         }
         const command = `git ${args.join(' ')}`;
-        this.info(command);
+        if (options.trace) {
+            this.info(command);
+        }
         shelljs.exec(command, options, callback);
     }
 
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1573,13 +1573,14 @@ module.exports = class extends PrivateBase {
         if (arguments.length < 3) {
             options = {};
         }
-        options.async = true;
-        options.silent = true;
+        if (options.async === undefined) options.async = true;
+        if (options.silent === undefined) options.silent = true;
 
         if (!Array.isArray(args)) {
             args = [args];
         }
         const command = `git ${args.join(' ')}`;
+        this.info(command);
         shelljs.exec(command, options, callback);
     }
 
@@ -1689,6 +1690,24 @@ module.exports = class extends PrivateBase {
      */
     warning(msg) {
         this.log(`${chalk.yellow.bold('WARNING!')} ${msg}`);
+    }
+
+    /**
+     * Print an info message.
+     *
+     * @param {string} value - message to print
+     */
+    info(msg) {
+        this.log.info(msg);
+    }
+
+    /**
+     * Print a success message.
+     *
+     * @param {string} value - message to print
+     */
+    success(msg) {
+        this.log.ok(msg);
     }
 
     /**
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -53,26 +53,28 @@ module.exports = UpgradeGenerator.extend({
             this.clientPackageManager = this.config.get('clientPackageManager');
             this.clientFramework = this.config.get('clientFramework');
             this.skipInstall = this.options['skip-install'];
-            this.silent = !this.options.verbose;
+            this.silent = this.options.silent;
             this.targetVersion = this.options['target-version'];
         }
     },
 
     _gitCheckout(branch, callback) {
-        this.gitExec(['checkout', '-q', branch], (code, msg, err) => {
+        this.gitExec(['checkout', '-q', branch], { silent: this.silent }, (code, msg, err) => {
             if (code !== 0) this.error(`Unable to checkout branch ${branch}:\n${err}`);
-            this.log(`Checked out branch "${branch}"`);
+            this.success(`Checked out branch "${branch}"`);
             callback();
         });
     },
 
     _cleanUp() {
+        const filesToKeep = ['.yo-rc.json', '.jhipster', 'node_modules', '.git', '.idea', '.mvn'];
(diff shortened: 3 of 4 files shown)


tientq:
That's much better, awesome.

cli/jhipster.js
```diff
@@ -35,6 +35,10 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
     process.exit(1);
 }
 
+/* WARNING: due to this code, it's a bit more difficult to test when developing
+ * the upgrade generator, workaround consists in running it this way:
+ * node generator-jhipster/cli/cli.js upgrade
+ */
```
deepu105: should we add a flag to control this behavior like `--use-local` and `--use-global` which can override these

generators/upgrade/index.js
```diff
@@ -82,22 +84,24 @@ module.exports = UpgradeGenerator.extend({
             const generatorDir = this.clientPackageManager === 'yarn' ? shelljs.exec('yarn bin', { silent: this.silent }).stdout : shelljs.exec('npm bin', { silent: this.silent }).stdout;
             generatorCommand = `${generatorDir.replace('\n', '')}/jhipster`;
         }
-        shelljs.exec(`${generatorCommand} --with-entities --force --skip-install`, { silent: this.silent }, (code, msg, err) => {
-            if (code === 0) this.log(chalk.green(`Successfully regenerated application with JHipster ${version}`));
+        const regenerateCmd = `${generatorCommand} --with-entities --force --skip-install`;
+        this.log.info(regenerateCmd);
```
deepu105: shouldnt this be `this.info`?

generators/upgrade/index.js
```diff
@@ -106,12 +110,15 @@ module.exports = UpgradeGenerator.extend({
     _regenerate(version, callback) {
         this._generate(version, () => {
             if (this.clientFramework === 'angular1' && version === this.latestVersion) {
+                this.log.info('bower install');
```
deepu105: same here

generators/upgrade/index.js
```diff
@@ -106,12 +110,15 @@ module.exports = UpgradeGenerator.extend({
...
                 const result = this.spawnCommandSync('bower', ['install']);
                 if (result.status !== 0) {
                     this.error('bower install failed.');
                 }
             }
-            shelljs.rm('-Rf', `${SERVER_MAIN_RES_DIR}keystore.jks`);
+            const keystore = `${SERVER_MAIN_RES_DIR}keystore.jks`;
+            this.log.info(`Removing ${keystore}`);
```
deepu105: here as well

generators/upgrade/index.js
```diff
@@ -225,8 +232,10 @@ module.exports = UpgradeGenerator.extend({
             const installJhipsterLocally = (version, callback) => {
                 this.log(`Installing JHipster ${version} locally`);
                 const commandPrefix = this.clientPackageManager === 'yarn' ? 'yarn add' : 'npm install';
-                shelljs.exec(`${commandPrefix} ${GENERATOR_JHIPSTER}@${version} --dev --no-lockfile --ignore-scripts`, { silent: this.silent }, (code, msg, err) => {
-                    if (code === 0) this.log(chalk.green(`Installed ${GENERATOR_JHIPSTER}@${version}`));
+                const generatorCommand = `${commandPrefix} ${GENERATOR_JHIPSTER}@${version} --dev --no-lockfile --ignore-scripts`;
+                this.log.info(generatorCommand);
```
deepu105: here

generators/upgrade/index.js
```diff
@@ -295,24 +306,40 @@ module.exports = UpgradeGenerator.extend({
...
+            done();
+        });
+    },
+
     install() {
         const done = this.async();
         if (!this.skipInstall) {
-            shelljs.rm('-rf', 'node_modules');
             this.log('Installing dependencies, please wait...');
+            this.log.info('Removing the node_modules directory');
```
deepu105: here

generators/generator-base.js
```diff
@@ -1683,6 +1684,24 @@ module.exports = class extends PrivateBase {
     }
 
     /**
+     * Print an info message.
+     *
+     * @param {string} value - message to print
+     */
+    info(msg) {
```
deepu105: the logging related methods are kind of spread around. There is debug in base-private. may be we should bring them all in this file also may be make it more clearer by making a wrapper object so that we can do
```
this.log.info()
this.log.debug()
this.log.error()
```
and so on. WDYT?

deepu105: since log is already defined by yeoman may be we should use `logger` namespace

generators/upgrade/index.js
gmarziou: No because this.log.info() provides 2 things I like here: indentation and grey info prefix like you can see in [screenshot](https://github.com/jhipster/generator-jhipster/issues/6237#issuecomment-323235768).

cli/jhipster.js
gmarziou: Yes but I must test that it addresses this particular use case when developing the generator because unfortunately, yarn does not handle binaries well when using link.
I'll give feedback on this.

generators/upgrade/index.js
deepu105: Yes but your info method also calls the same `this.log.info` rite?
```
+    info(msg) {
 +        this.log.info(msg);
 +    }
```

generators/generator-base.js
gmarziou: Yes I thought about it too but I did not want to impact too much code. From the log statuses provided by yeoman, I like `ok()` and `info()` and `maybe error()` (I don't remember it well) but I'm not sure the others are that useful, for instance I tried to use `invoke()` for all git calls but finally preferred `info()`. 

Also I must admit that I did not understand how their indentation system works with [emitted events](https://github.com/yeoman/environment/blob/master/lib/util/log.js#L51-L52).

generators/upgrade/index.js
gmarziou: Right, I don't remember, I probably forgot about it. I'll have a look.

generators/generator-base.js
gmarziou: and I agree that we should have a uniform way to log success, error, info and debug rather than using all kind of colors with chalk. Yeoman's doc does not provide much guidance

generators/upgrade/index.js
gmarziou: OK just checked, you're right. I replaced all calls to `this.log.info()` by `this.info()`

generators/upgrade/index.js
gmarziou: BTW, here I don't think it really makes sense to delete `node_modules`, yarn is supposed to manage deps better and probably npm5 too. IMO, it just slows down the upgrade process.

gmarziou:
I rebased and made following changes:
- replaced this.log.info() by this.info()
- used commander for parsing new option `--prefer-local` which is useful when developing the upgrade generator itself with yarn link.

cli/jhipster.js
```diff
@@ -35,21 +35,31 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
...
-if (process.argv.length >= 3) {
-    if (process.argv[2] === 'upgrade') {
-        // Prefer global version for `jhipster upgrade` to get most recent code
-        preferGlobal = true;
-    }
-}
-requireCLI(preferGlobal);
+let preferLocal = true;
+
+program.command('upgrade')
```
deepu105: doesnt this interfere with the original config in the cli file?

gmarziou: I don't think so because I tested it (not all generators but upgrade and app are fine) and because program is defined as a const in both jhipster.js and cli.js so if they were interfering I would expect cli to raise an error for assigning twice a constant. I will check under debugger anyway.

gmarziou: I was wrong: under debugger I was able to see that `program` is a global variable and in cl.js it is still holding the options set in jhipster.js. In fact, only the ref to the object is a const. I will rollback to previous code that did not use commander.

deepu105: ya I think that is safer

cli/jhipster.js
```diff
@@ -35,21 +34,23 @@ if (!semver.satisfies(currentNodeVersion, minimumNodeVersion)) {
...
-if (process.argv.length >= 3) {
-    if (process.argv[2] === 'upgrade') {
-        // Prefer global version for `jhipster upgrade` to get most recent code
-        preferGlobal = true;
-    }
+let preferLocal = true;
+
+// Don't use commander for parsing command line to avoid polluting it in cli.js 
+// --prefer-local: Always resolve node modules locally (useful when using linked module)
+if ((process.argv.indexOf('upgrade') > -1) && (process.argv.indexOf('--prefer-local') < 0)) {
```
deepu105: you could also do `process.argv.includes('upgrade') && process.argv.includes('--prefer-local'))` which seems cleaner.

gmarziou: Good idea, will do.

PR #6251, opened 2017-08-18
Update _UserServiceIntTest.java

cmyker:
Vanilla jHipster installation is failing integration test when you run whole integration tests package. 
The reason is that another test class (AccountResourceIntTest) is running before and deleting the data that this test relies on.
Proposed solution - load whole new context for this test.

### DIFF
--- a/generators/server/templates/src/test/java/package/service/_UserServiceIntTest.java
+++ b/generators/server/templates/src/test/java/package/service/_UserServiceIntTest.java
@@ -30,7 +30,8 @@ import <%=packageName%>.service.util.RandomUtil;<% } %>
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.test.context.SpringBootTest;<% if (databaseType === 'sql') { %>
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;<% if (databaseType === 'sql') { %>
 import org.springframework.transaction.annotation.Transactional;<% } %>
 import org.springframework.test.context.junit4.SpringRunner;
 <% if (databaseType === 'sql' || databaseType === 'mongodb') { %>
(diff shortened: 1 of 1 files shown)


gmarziou:
Isn't it because `AccountResourceIntTest` is not annotated as` @Transactional` ?

cmyker:
Might be. But keep in mind that other tests could break this one.

gmarziou:
Well all tests that update the database should be annotated `@Transactional`.

`@DirtiesContext` is more used to [reset stateful beans in the `ApplicationContext`](https://docs.spring.io/spring/docs/current/spring-framework-reference/html/integration-testing.html#__dirtiescontext), in the case of integration tests, doing so resets the H2 database because it is in memory, it would not work if H2 was in file mode or if you were using an external database.

Another drawback of `@DirtiesContext` is that by resetting the ApplicationContext, you may slow down your tests.

I don't think we have many stateful beans in JHipster apps, caching is disabled for tests if I'm correct so
I could be wrong but I don't think we need `@DirtiesContext`

cmyker:
I've tried your suggestion with `@Transactional`, but the test fails anyway

cmyker:
Also marking tests with `@Transactional` makes them produce "java.lang.IllegalStateException"

cmyker:
Should I close this PR and reports the bug instead?

gmarziou:
Please open an issue and provide details about your setup (`jhipster info`) so that we can reproduce. You can keep this PR linked to the issue or add what you tried here to the issue comments.

jdubois:
Reeading your configuration at #6269 you are using MongoDB, so no @Transactional for you :-)

cbornet:
DirtiesContext slows down the tests too much. I've had the same kind if issue in the generator-jhipster-grpc. My solution was to have a test suite that cleans the table and creates a user for each test.

cbornet:
See https://github.com/cbornet/generator-jhipster-grpc/blob/master/generators/app/templates/_UserServiceIntTest.java

gmarziou:
Couldn't we just use `@Transactional` when database is SQL and maybe something else for noSQL?

jdubois:
`@Transactional` is only for SQL, because, they support transactions. This just doesn't work for MongoDB or Cassandra.
This is also why you can't have automatic rollback for MongoDB and Cassandra (who said NoSQL was easier or less verbose???)

jdubois:
Closing this as:

- I still can't reproduce #6269
- What is being discussed here is wrong anyway: there is not `@Transactional` for non-relational databases, so this isn't the solution to anything

Issue #6263, opened 2017-08-21
Access control policy is not working


lchayoun:
##### **Overview of the issue**

When trying to use AccessControlPolicy, the return code is 200, instead of 403, and the request is getting routed to the service

##### **Motivation for or Use Case**

This is unexpected since the request shouldn't be routed

##### **Reproduce the error**

1) configure AccessControlPolicy
2) access the relevant uri

##### **Related issues**

Not found

##### **Suggest a Fix**

remove the if statement on the run function, need to call ctx.setSendZuulResponse(false) always

##### **JHipster Version(s)**

latest

##### **JHipster configuration**

jhipster.gateway.authorizedMicroservicesEndpoints.serviceA: ['/api/pathA']

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

none

##### **Browsers and Operating System**

Not relevant

- [X] Checking this box is mandatory (this is just to show you read everything)

deepu105:
This ticket has been closed as the guidelines are not followed.

Tickets must follow our [Guidelines](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#submitting-an-issue), as mentioned in:

1. our [Readme file on the front page of the project](https://github.com/jhipster/generator-jhipster/blob/master/README.md), 
2. the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) and
3. our [Help page](http://jhipster.github.io/help/)

We have also created a template on the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) to help you follow those guidelines.

Open a new issue with proper details or update this issue with all details and request to reopen.
Issues opened without proper details will be closed without explanation.

If you have a question please use Stack Overflow for that. Tag the question with [jhipster](http://stackoverflow.com/questions/tagged/jhipster) and someone will help you out there.

This helps us to keep our issue tracker clean. Also, Stack Overflow will give your query a larger audience.

PR #6264, opened 2017-08-21
generator-jhipster-6263 Access control policy is not working - remove…

lchayoun:
…d enclosing if statement


lchayoun:
for #6263

lchayoun:
@jdubois maybe I don't follow, but in order to make swagger work (because it is blocked by the AccessControl) you added a check that all gzipped responses will be allowed?

HttpStatus.FORBIDDEN.value() is not enough, you can check RibbonRoutingFilter and SimpleHostRoutingFilter, the shouldFilter() function will return false on several cases, one of them is that sendZuulResponse is true.

in case we decided to deny a request based on access control, we want to stop the routing.

gmarziou:
@marcelinobadin you were the one who wrote #3940 to fix issue #3933, maybe you can join the discussion?

jdubois:
Please note that I didn't write this filter, so maybe there's something I don't understand - but I just tested and I totally agree with you.
I really don't know what happened here - so thanks for testing, and I'm merging this!

PR #6274, opened 2017-08-22
Spring Boot 2.0.0.M3

JoffreyDcu:
I did the following:
- updated spring boot dependency
- updated classes which used renamed or deprecated spring classes or functions

The binding of properties fails with Travis, but works fine locally.
ElasticSearch does not start a node by default anymore (use docker?).

### DIFF
--- a/generators/entity/templates/server/src/main/java/package/web/rest/_EntityResource.java
+++ b/generators/entity/templates/server/src/main/java/package/web/rest/_EntityResource.java
@@ -26,6 +26,9 @@ import <%=packageName%>.domain.<%= entityClass %>;
 import <%=packageName%>.service.<%= entityClass %>Service;<% } else { %>
 import <%=packageName%>.repository.<%= entityClass %>Repository;<% if (searchEngine === 'elasticsearch') { %>
 import <%=packageName%>.repository.search.<%= entityClass %>SearchRepository;<% }} %>
+<%_ if (reactive) { _%>
+import <%=packageName%>.web.rest.util.AsyncUtil;
+<%_ } _%>
 import <%=packageName%>.web.rest.util.HeaderUtil;<% if (pagination !== 'no') { %>
 import <%=packageName%>.web.rest.util.PaginationUtil;<% } %>
 <%_ if (dto === 'mapstruct') { _%>
@@ -51,6 +54,10 @@ import org.springframework.http.HttpStatus;
 <%_ } _%>
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
+<% if (reactive) { _%>
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+<%_ } _%>
 <% if (validation) { %>
 import javax.validation.Valid;<% } %>
 import java.net.URI;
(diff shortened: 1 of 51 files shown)


PierreBesson:
Maybe this could be merged on a new branch named `jhipster-v5` or `reactive` as Spring Boot 2.0 is only scheduled for the end of november (at best !), see https://github.com/spring-projects/spring-boot/milestones. Then @JoffreyDcu and others would be free to work on reactive support on this branch.

deepu105:
@PierreBesson yes these should be worked on a new branch similar to jh-react

deepu105:
use this branch https://github.com/jhipster/generator-jhipster/tree/jh-reactive-spring

cbornet:
Starting JDK9 support would also be nice !

cbornet:
So we move to spring-data-cassandra ?

jdubois:
@cbornet sorry yes, let me explain.

- Less than 2% of users use Cassandra, so my own custom code brings a lot of maintenance for too few people
- Spring Data Cassandra has been totally re-worked: it's supposed to be much better, but I haven't looked at it in detail
- This is going to help us for Spring Webflux, as they already have reactive interfaces (which is not the case with my current code)

balbler:
I am also in the process of converting an older jhipster based app to Spring boot 2.0, so I will try to help with my limited knowledge.
Two Questions:
->ES 5.0 introduces new field types by splitting the former string type into a text and a keyword type (https://www.elastic.co/blog/strings-are-dead-long-live-strings) . Does it make sense to pull this  differentiation into jhipster?

-> on my Application I do have an issue where there seems to be an issue with AbstractRememberMeServices causing NullPointerExceptions in the loginSuccess Method. Seems the Spring proxy is used there resulting in Null fields. Does not happen with older spring boot/ spring versions. I can reproduce this by also changing the jhipster sample app. Anybody else seeing this?

deepu105:
@JoffreyDcu I suggest doing the Spring boot update and Reactive support in 2 different PR so that its easy to review and merge
Also i'm not sure if adding Cassandra support is worth the effort for Reactive support

generators/server/templates/src/main/java/package/service/_SocialService.java
```diff
@@ -122,7 +123,10 @@ private User createUserIfNotExist(UserProfile userProfile, String langKey, Strin
         String login = getLoginDependingOnProviderId(userProfile, providerId);
         String encryptedPassword = passwordEncoder.encode(RandomStringUtils.random(10));
         Set<Authority> authorities = new HashSet<>(1);
-        authorities.add(authorityRepository.findOne("ROLE_USER"));
+
+        Optional<Authority> authorityOptional = authorityRepository.findById("ROLE_USER");
+        if (!authorityOptional.isPresent()) throw new NoSuchElementException();
```
cbornet: use orElseThrow

agoncal:
FYI I started working on integrating HibernateSearch in JHipster. But if we want to have all the nice Spring Boot Starter integration, we will have to wait for Spring Boot 2 : https://github.com/snowdrop/spring-boot-hibernate-search-booster/issues/2#issuecomment-347933057

generators/entity/templates/server/src/main/java/package/web/rest/_EntityResource.java
```diff
@@ -130,7 +147,7 @@
      * @return the ResponseEntity with status 200 (OK) and the list of <%= entityInstancePlural %> in body
      */
     @GetMapping("/<%= entityApiUrl %>")
-    @Timed<%- include('../../common/get_all_template', {viaService: viaService}); -%>
+    @Timed<%- include('../../common/get_all_template', {viaService: viaService}) -%>
```
cbornet: We should probably also have an endpoint producing `application/stream+json` to prevent blocking in large JSON serialization/deserialization

jdubois:
Hi everyone, as @JoffreyDcu went back to school, I don't think he will have time to split this PR into 3, as @deepu105 suggested (which would probably have been the best solution, indeed).
I'm going to merge this in a specific branch, and will start working on it - we need to upgrade to the latest Spring Boot 2.0 RC1 release, and use our new jhipster-dependencies project, so there's quite a lot of work.

jdubois:
Arrgh, according to GitHub there was no merge conflict, which was very surprising - and doing this manually shows a *lot* of conflicts, in fact!!!
I'm not sure this can be merged easily -> I'll first do an upgrade to Spring Boot 2 without the reactive option, as @deepu105 suggested. The main work here is to have the reactive option, I hope we can still get it without too much trouble.

gmarziou:
SB 2.0 release is planned for Feb, 20th that's quite a long time to keep this branch.

jdubois:
@gmarziou that's why I would do a branch with just the Spring Boot 2 upgrade, not the webflux stuff -> this is what is causing all the conflicts, as it's scattered all over the entites
The Spring Boot 2 upgrade alone shouldn't be a big change, so hopefully this can be maintained easily

deepu105:
closing since the work is now in another branch #7061

PR #6279, opened 2017-08-23
Exporting entity enum

agoncal:
When generating an Entity containing an enumeration, this enumeration should be exported in the Angular model.

See discussion : https://stackoverflow.com/questions/45615044/entity-enum-is-not-exported/45616603#45616603


gmarziou:
I restarted the failed build as I'm not sure it's related to your change.

PR #6328, opened 2017-09-07
Use RFC7807 Problem error format

cbornet:

Following https://tools.ietf.org/html/rfc7807


jdubois:
Looks weird: you delete some new error messages, and you have 'org.zalando' everywhere.... I'm on my phone so I'm not sure, Christophe can you re-check?

PierreBesson:
This looks cool. We need to make sure it doesn't cause problems but it seems like a good idea.
@jdubois, this is the library @cbornet wants to include https://github.com/zalando/problem-spring-web/blob/master/README.md

cbornet:
I have moved some of the http errors to `error.http.4xx` standard codes because `error.methodNotSupported` didn't give more information than error.http.405 for instance.
Yes the lib I use to implement rfc7807 is Zalando's one. Great lib : check their code, really nice !

cbornet:
Once/if the PR is merged, we should provide error descriptions at
http://www.jhipster.tech/problem/problem-with-message
http://www.jhipster.tech/problem/contraint-violation
http://www.jhipster.tech/problem/parameterized
as per rfc7807 recommendation
I can change the URLs if the path doesn't look good

cbornet:
You can check [ExceptionTranslatorIntTest](https://github.com/jhipster/generator-jhipster/pull/6328/files#diff-48ad94b020fc55935f4e4e290664cc5e), you will see that the fields aren't changed except `description` which becomes `detail` but this field is not used in the front end.

This makes me think that it would be great to have CDC tests between the front and the back using something such as [pact](https://docs.pact.io/).

jdubois:
Thanks @cbornet - I have more time today (and a computer!) to have a look at this.

First of all let me ping @whiskeysierra who seems to be the main contributor for https://github.com/zalando/problem-spring-web (Hi Willi, please have a look at JHipster, as if we push your library then you'll end up with lots of downloads - and issues! - from us)

What I like:

- Less code
- Cleaner solution
- Probably does a better job at handling exceptions

What I don't like:

- This is pushing https://github.com/zalando/problem-spring-web to everybody, and this isn't a widely-used library - but that also depend on @whiskeysierra

Then, I wanted to do a release early next week with our improved error handling code, and this is changing everything, so we should decide this very quickly.

-> can you vote (thumb up if you like, thumb down if you don't like) on this comment? I'll also ask on twitter.

whiskeysierra:
> as if we push your library then you'll end up with lots of downloads - and issues! - from us

No problem there. I'm up for the challenge!

> and this isn't a widely-used library

True, but you have the power to change that, I guess.

jdubois:
Thanks @whiskeysierra !! Yes indeed you should see the "JHipster effect" in your downloads, and also in your tickets... So I'd rather ask upfront, because that might add some work for you (I know about it, I got something like 50-100 emails per day on JHipster...)

atomfrede:
We use zalando problem in our products and it works fine :) So +1 from me.

RobWin:
We us the library in some of our services and it works. +1

Dufgui:
Never use it but I like the concept. +1 for me.

jdubois:
OK, looks like everybody loves this: congrats @cbornet for your always excellent work!!!
I'll merge this and put it in the next release

jdubois:
I'm merging this -> @deepu105 I think this is OK, otherwise @cbornet can you provide us more insight?
I'm going to test this, but I have first work to add the new error pages on the JHipster website

cbornet:
Great ! See my [previous comment](https://github.com/jhipster/generator-jhipster/pull/6328#issuecomment-327947992) about backward compatibility. The biggest change was in message translation keys which is handled by the PR (and was unrelated to Problem)

PR #6334, opened 2017-09-11
Fix Travis protractor builds by fixing yarn version to pre 1.X

PierreBesson:
Yarn 1.X introduces breaking changes: https://github.com/yarnpkg/yarn/releases/tag/v1.0.0

I have some ideas in how to fix those problems and upgrade to yarn 1.X but it can be done later as it's not straightforward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
   # Repo for Yarn
-  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 0.27.5
   - export PATH=$HOME/.yarn/bin:$PATH
   - yarn global add yo bower gulp-cli
 install:


jdubois:
Oh thanks... I'm trying to fix this...
Of course that's good to fix the Travis build, but if the end-user project doesn't work that's an even bigger issue. I'm testing everything I can.

jdubois:
- This looks OK for the generated projects... In fact I already updated without noticing, when I did a `brew update`
- I can't find what the issue is in Travis, looks like a corrupted cache, but it still fails when you remove the cache... This looks like a very buggy release to me, as one of the highlights of Yarn was that it's supposed to validate the downloaded dependencies!! And the Yarn documentation doesn't give any information - when you have such a widely-used tool, which is automatically updated on most computers, at least you should document your breaking changes correctly! And provide good error messages! Anyway, if your fix works, I'll merge this, because I'm already fed up with this. Maybe we should consider NPM 5 now.

PierreBesson:
@jdubois the yarn 1.0 release states the breaking changes ! Also this is their first major release so I would not be so harsh on them. 
Note that yarn 1.0 works very well for me in local, I think the problem is https://github.com/eirslett/frontend-maven-plugin/issues/647
Anyway I have this mostly figured this out, so I think we will be able to upgrade to yarn 1.0 once the maven plugin is updated.

pascalgrimaud:
I was so busy today at work, as you can see, so very good job @PierreBesson !
I don't know how we can manage this. Does it mean people will have different behavior between Yarn < 1.0 and Yarn >= 1.0 ?

jdubois:
On my machine, I had to delete all my `node_modules` folders and do a new `yarn install`, otherwise I have tons of errors telling me I have corrupted files, like we had on Travis. Doesn't look very good to me, but it works.

PierreBesson:
@pascalgrimaud, no problem to manage this. I think that using the front-end maven plugin, you can manage the node/yarn version easily (there is a similar plugin for gradle that we plan to add).
I think this was bound to happen as we were not fixing the yarn version used by Travis.

pascalgrimaud:
I applied this fix to [hipster-labs/jhipster-travis-build](https://github.com/hipster-labs/jhipster-travis-build)
All builds should be green now: https://travis-ci.org/hipster-labs/jhipster-travis-build/branches

So it should be OK for the release tomorrow!

PR #6338, opened 2017-09-12
Include required polyfills in bundles

### LINKED ISSUE #6337: Application does not show before hitting refresh in IE11

After deploying our JHipster application on Oracle WebLogic container in prod profile, it showed up a blank white page in IE11. The application loaded correctly when pressing F5 / Reload.

The application is not accessible at all on first visit.

##### **Reproduce the error**

- Build application with prod profile
- Deploy to container
- Access the application with IE11
- Most of the time the page shows up blank, Javascript console shows the error `Zone.js has detected that ZoneAwarePromise (window|global).Promise has been overwritten. Most likely cause is that a Promise polyfill has been loaded after Zone.js (Polyfilling Promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)`

##### **Related issues**

n/a

##### **Suggest a Fix**

The problem is that core-js is inserted on-demand which seems to lead to an unpredictable include order. Interestingly, it works after pressing F5. However, the suggested fix would be to *always* include polyfills for supported browsers, i.e. not dependent on the user-agent. We have already fixed this issue using polyfills.ts in the same way Angular CLI template does it. I will open a PR to discuss.

Additional benefit: The application loads a lot fast in IE11, I could not see other browsers getting slower.

##### **JHipster Version(s)**

4.6.2

.yo-rc.json

```
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "XXX",
      "nativeLanguage": "de"
    },
    "jhipsterVersion": "4.6.2",
    "baseName": "XXX",
    "packageName": "XXX",
    "packageFolder": "XXX",
    "serverPort": "8080",
    "hibernateCache": "no",
    "websocket": false,
    "databaseType": "sql",
    "prodDatabaseType": "oracle",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "useSass": true,
    "testFrameworks": [
      "protractor"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "de",
    "languages": [
      "de"
    ]
  }
}

```

The problem is only reproducible on IE (11).

stieler-it:
Dynamic loading seems to be problematic with IE11 in prod profile.

fixes #6337

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
@@ -16,8 +16,77 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-/* tslint:disable */
-import 'reflect-metadata/Reflect';
-import 'zone.js/dist/zone';
+/**
+ * This file includes polyfills needed by Angular and is loaded before the app.
+ * You can add your own extra polyfills to this file.
+ *
+ * This file is divided into 2 sections:
+ *   1. Browser polyfills. These are applied before loading ZoneJS and are sorted by browsers.
+ *   2. Application imports. Files imported after ZoneJS that should be loaded before your main
+ *      file.
+ *
+ * The current setup is for so-called "evergreen" browsers; the last versions of browsers that
+ * automatically update themselves. This includes Safari >= 10, Chrome >= 55 (including Opera),
+ * Edge >= 13 on the desktop, and iOS 10 and Chrome on mobile.
+ *
+ * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
+ */
+
+/***************************************************************************************************
+ * BROWSER POLYFILLS
+ */
+
+/** IE9, IE10 and IE11 requires all of the following polyfills. **/
+import 'core-js/es6/symbol';
+import 'core-js/es6/object';
+import 'core-js/es6/function';
+import 'core-js/es6/parse-int';
+import 'core-js/es6/parse-float';
+import 'core-js/es6/number';
+import 'core-js/es6/math';
+import 'core-js/es6/string';
+import 'core-js/es6/date';
+import 'core-js/es6/array';
+import 'core-js/es6/regexp';
+import 'core-js/es6/map';
+import 'core-js/es6/weak-map';
+import 'core-js/es6/set';
+
+/** IE10 and IE11 requires the following for NgClass support on SVG elements */
+// import 'classlist.js';  // Run `npm install --save classlist.js`.
+
+/** Evergreen browsers require these. **/
+import 'core-js/es6/reflect';
+import 'core-js/es7/reflect';
+
+
+/**
+ * Required to support Web Animations `@angular/animation`.
+ * Needed for: All but Chrome, Firefox and Opera. http://caniuse.com/#feat=web-animation
+ **/
+// import 'web-animations-js';  // Run `npm install --save web-animations-js`.
+
+
+
+/***************************************************************************************************
+ * Zone JS is required by Angular itself.
+ */
+import 'zone.js/dist/zone';  // Included with Angular CLI.
+
+
+
+/***************************************************************************************************
+ * APPLICATION IMPORTS
+ */
+
+/**
+ * Date, currency, decimal and percent pipes.
+ * Needed for: All but Chrome, Firefox, Edge, IE11 and Safari 10
+ */
+// import 'intl';  // Run `npm install --save intl`.
+/**
+ * Need to import at least one locale-data with intl.
+ */
+// import 'intl/locale-data/jsonp/en';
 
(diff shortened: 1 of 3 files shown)


generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
```diff
@@ -16,8 +16,77 @@
...
+ *
+ * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
+ */
+
+/***************************************************************************************************
+ * BROWSER POLYFILLS
+ */
+
+/** IE9, IE10 and IE11 requires all of the following polyfills. **/
+import 'core-js/es6/symbol';
```
deepu105: I have a problem with this. This forces all these additional JS on every browser and increases bundle size just to support IE, not everyone might support IE
We officially do not support IE9, 10 so no need of those polyfills by default anyway.

As a compromise what I can accept is loading polyfills required by IE11 alone and have the remaining commented out with clear instruction to enable if need to support older IE versions.

stieler-it:
I am not sure about the line `import 'reflect-metadata/Reflect';` - is that actually needed?

deepu105:
It was added to fix some issue in the past not sure if its still applicable

generators/client/templates/angular/src/main/webapp/app/_polyfills.ts
stieler-it: We also don't need to support older version than IE11. However, following the comments of the Angular CLI template I was not able to find out which of them are not necessary in this case. I will investigate further.

stieler-it: According to Angular Docs (https://angular.io/guide/browser-support) IE11 requires "ES6" polyfills My PR follows the Angular CLI example which does not include the whole core-js package but picks some polyfills. It looks to me like all those would be necessary for IE11. Do you see any indication that some might not be necessary?

deepu105: take a look at this https://polyfill.io/v2/docs/features/

stieler-it: Thanks. Looking at the information there, I think it would be possible to remove ``core-js/es6/date`` and ``core-js/es6/regexp``. The other polyfills seem to be required even for IE11. The page also shows that the difference of the required polyfills between IE10 and IE11 is not more than 400 bytes (before compression).

I would suggest to keep the polyfills as suggested by Angular CLI to try not to diverge too much from the standard.

If you don't agree, I would suggest the following patch to this PR:

```
...
- import 'core-js/es6/date';
+ // import 'core-js/es6/date'; // Already supported by IE11
  import 'core-js/es6/array';
- import 'core-js/es6/regexp';
+ // import 'core-js/es6/regexp'; // Already supported by IE11
...
```

deepu105:
@stieler-it ok then can you do a comparison of devtools network tab in prod profile with and with out this to see what is the overall increase in bundle size with this? if its is not too much we can merge it as it is

stieler-it:
It should be enough to compare polyfills.xxx.bundle.js, right?

Broken version: 52.9 KB
Fixed version: 92.9 KB (including date and reg.exp)

That means an increase of 40 KB.

Please note that:
- Vendor.xxx.bundle.js has about 1000 KB for our application
- for IE we are saving core-js-shim.min.js which is 77.7 KB

deepu105:
did you see any difference in size of the vendor.xxx.bundle.js ?

stieler-it:
Yes but I thought that would be unrelated to my change. I will make a clean comparison tomorrow. Also, I have to remove some empty lines before this should be merged.

stieler-it:
I now have now done a comparison on our currently developed application:

Before:
```
        46 global.67160f753b403b9fd5d3.bundle.js
 1.947.878 main.67160f753b403b9fd5d3.bundle.js
     1.739 manifest.67160f753b403b9fd5d3.bundle.js
    53.740 polyfills.67160f753b403b9fd5d3.bundle.js
 1.266.467 vendor.67160f753b403b9fd5d3.bundle.js
```

After:
```
        46 global.adec2453a1b651db8793.bundle.js
 1.948.465 main.adec2453a1b651db8793.bundle.js
     1.739 manifest.adec2453a1b651db8793.bundle.js
    94.732 polyfills.adec2453a1b651db8793.bundle.js
 1.268.843 vendor.adec2453a1b651db8793.bundle.js
```

So around ~40KB more in polyfills and ~2KB more in vendor, which is roughly 1 percent increase for our application. (Again, for non-IE browsers. For IE it's actually an improval.)

deepu105:
Ok then, I hate IE :rage:

deepu105:
Will merge once build is green

Issue #6281, opened 2017-08-23
Migrate to the new HttpClientModule from Angular 4.3


agoncal:
##### **Overview of the issue**

Angular 4.3 brings a new module for HTTP communication. This new module has "_updated API's with support for progress events, json deserialization by default, Interceptors and many other great features_".

##### **Motivation for or Use Case**

JHispter 4.7 has upgraded Angular from 4.2.6 to 4.3.2. From Angular 4.3 a new HTTP module has arrived and it should be the one to use (the "old" one might get deprecated). 

I've liste a few external references below on how to migrate and so on. Basically, the "old" `HttpModule` located under `@angular/http`, is replaced by the new `HttpClientModule` from `@angular/common/http`. 

##### **External references**

* [HTTPClient](https://angular.io/guide/http)
* [The Angular HTTP Client - Quickstart Guide](http://blog.angular-university.io/angular-http/)
* [Migrating to HttpClient](http://brianflove.com/2017/07/21/migrating-to-http-client/)
* [Migrating from Angular HttpModule to new Angular 4.3 HttpClientModule](https://devops.datenkollektiv.de/migrating-from-angular-httpmodule-to-new-angular-43-httpclientmodule.html)
* [Difference between HTTP and HTTPClient in angular 4?](https://stackoverflow.com/questions/45129790/difference-between-http-and-httpclient-in-angular-4)
* [what's the difference between HttpModule and HttpClientModule in Angular2?](https://stackoverflow.com/questions/45405593/whats-the-difference-between-httpmodule-and-httpclientmodule-in-angular2)
* [Angular - How to use HttpClientModule?](http://blog.ninja-squad.com/2017/07/17/http-client-module/)

##### **JHipster Version(s)**

JHipster 4.7

Thisen:
I'm currently working on a migration to HttpClient in the Angular template. I could be assigned to this, if you don't mind.

jdubois:
Oh that's nice @Thisen -> I'm sending you an invite as you need to be a member of the project to have a task assigned (that's how GitHub works, I always find it not user friendly...)

Thisen:
I have accepted the invite now @jdubois.

Thisen:
Just a quick update: I haven't forgot about this issue, but holiday, moving project at work and sickness have stalled me. I believe I will get back to working on it in the weekend.

chegola:
@Thisen  I am interesting on HttpClientModule, pls let me if I can help on something.

[...]

PR #6359, opened 2017-09-14
Ensure that tokens are only sent with internal requests

dancancro:

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_auth.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_auth.interceptor.ts
@@ -31,6 +31,8 @@ export class AuthInterceptor extends JhiHttpInterceptor {
     }
 
     requestIntercept(options?: RequestOptionsArgs): RequestOptionsArgs {
+        if (!options || !options.url || /^http/.test(options.url)) return options;
+
         const token = this.localStorage.retrieve('authenticationToken') || this.sessionStorage.retrieve('authenticationToken');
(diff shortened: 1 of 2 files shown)


gmarziou:
Thanks Dan, maybe you could have described a little the context, I will as I was in touch with you about it.

Current implementation of AuthInterceptor adds the JWT token as the Authorization header to all HTTP requests, there are 2 problems with this approach:
- it can make some external APIs to reject requests
- it exposes user's tokens to other servers than the intended JHipster app which is a potential security issue

gmarziou:
@jhipster/developers Dan's fix solves the problem in a simple way but I think we should consider reviewing our interceptors and our token storage solution, maybe by using a key that stores the domain name that emitted the token. Also I'm not sure how this would work with external auth providers like uaa, okta, keycloak, ...

gmarziou:
I think it should also be considered in #6281

Metavirulent:
In the context of UAA, we're about to get rid of auth.interceptor since access tokens will be kept in cookies and thus governed by the browser. Since cookies are bound to domains, they won't be sent  to other domains.

TL;DR: with PR #5812 merged, this will not have an impact on UAA-based microservices.

ruddell:
Tested the AngularJS change and it works well.

The current behavior is that authorization token will not be sent unless your site is explicitly added to the remote site's Allowed Origins CORS headers.  Credentials will not be sent if the server has `*` for Allowed Origins ([MDN source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Allow-Origin)).

xetys:
LGTM

jdubois:
Thanks!!

PR #6369, opened 2017-09-15
Email address check should be case insensitive - fixed only for SQL apps

gzsombor:

Currently it is possible to register twice with the same email address, but different case - example@Example.com and example@example.com - which is confusing. And when you ask for a password remainder, currently the user needs to enter with the same case. AFAIK the host names are case-insensitives by some RFC, and the user names by conventions (I haven't seen an email provider, which allowed to register duplicate accounts with different cases.

 This patch fixes the problem only for relational dbs, where it can be fixed easily by query.

### DIFF
--- a/generators/server/templates/src/main/java/package/repository/_UserRepository.java
+++ b/generators/server/templates/src/main/java/package/repository/_UserRepository.java
@@ -80,7 +80,7 @@ public interface UserRepository extends <% if (databaseType === 'sql') { %>JpaRe
 
     Optional<User> findOneByResetKey(String resetKey);
 
-    Optional<User> findOneByEmail(String email);
+    Optional<User> findOneByEmailIgnoreCase(String email);
 
     Optional<User> findOneByLogin(String login);
     <%_ if (databaseType === 'sql') { _%>
@@ -225,9 +225,9 @@ public class UserRepository {
         return findOneFromIndex(stmt);
     }
 
-    public Optional<User> findOneByEmail(String email) {
+    public Optional<User> findOneByEmailIgnoreCase(String email) {
         BoundStatement stmt = findOneByEmailStmt.bind();
-        stmt.setString("email", email);
+        stmt.setString("email", email.toLowerCase());
         return findOneFromIndex(stmt);
     }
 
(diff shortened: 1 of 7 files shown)


generators/server/templates/src/main/java/package/repository/_UserRepository.java
```diff
@@ -80,7 +82,12 @@
 
     Optional<User> findOneByResetKey(String resetKey);
 
+    <%_ if (databaseType === 'sql') { _%>
+    @Query("from User u where lower(u.email) = lower(:email)")
+    Optional<User> findOneByEmail(@Param("email") String email);
```
ruddell: This can be done without an annotation by using the JPA query `findOneByEmailIgnoreCase`

That would work for Mongo but I'm not really sure how Cassandra works

jdubois:
Good catch - I'm also pretty sure this can be abused and could be a security issue.

generators/server/templates/src/main/java/package/repository/_UserRepository.java
gzsombor: I didn't want to change the method name conditionally based on databaseType - it is called from more than 20 places - it would mean a lot's of if's.

gzsombor:
The other solution is to convert to lower case, before persisting.
Pro:
* it would work on cassandra and mongo too
Con:
* if someone manually upgrade, they need manual db changes. With liquibase it's not that difficult, but with other ... :scream: :wink:

generators/server/templates/src/main/java/package/repository/_UserRepository.java
deepu105: why not use `findOneByEmailIgnoreCase` as the method name for all DB?

deepu105: and for cassandra since its a method hand written we could do something like
```
    public Optional<User> findOneByEmailIgnoreCase(String email) {
        BoundStatement stmt = findOneByEmailStmt.bind();
        stmt.setString("email", email);
        return findOneFromIndex(stmt);
    }
```
and in the save method we can convert the email to lowercase

gzsombor:
I've implemented your suggestions, thank you @deepu105  and @ruddell , it seems to be working :smile:

PR #6373, opened 2017-09-17
fix csrf + cors filter chain order

ruddell:
CorsFilter needs to be before CsrfFilter or else a CSRF issue blocks the CORS headers from being applied.  Very similar to https://github.com/jhipster/generator-jhipster/issues/4704, related [StackOverflow answer](https://stackoverflow.com/a/46269353/3737815) with more explanation.

This affects Session apps and UAA gateways (now using CSRF) receiving requests from a client that requires CORS headers.

Instructions to reproduce:
 - Generate an app with session auth
 - Start it in `dev` at localhost:8080, `dev` has CORS on by default
 - Go to http://jruddell.com/examples/test.html and open the console.  It's the example from the StackOverflow linked above, hosted so that it requires CORS headers.   It makes a POST request to localhost:8080/api/authenticate to login as user (you can view the source). 
 - The request fails from CSRF, but the console complains `No 'Access-Control-Allow-Origin' header is present`.  If you try to view the response to the request in the network tab, it is blank.

Solution:
 - Add the CorsFilter before the CsrfFilter like in this PR and restart the app
 - You can now view the response in the network tab and see the real error message which is CSRF related.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
@@ -99,9 +99,6 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.cloud.client.loadbalancer.RestTemplateCustomizer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


jdubois:
I'm totally trusting you here, thanks!

Issue #6252, opened 2017-08-18
Epic: Enhance Gradle Support


oehme:
Following up on #5949, @atomfrede, @eriwen and I had a great chat this week on how to make JHipster's Gradle support awesome. I'd like to summarize our initial feedback here. We can break it down into smaller stories later on.

# High level observations

Let's start with a few things that can't be expressed in terms of "there's a problem at line 5" :)

## Verbose build scripts

The generator creates a lot of script content, which is probably overwhelming for the average user and also hard to evolve on JHipster upgrades. Most of the logic are conventions or defaults that should live inside a plugin (or small set of plugins) instead. A `jhipster` Gradle plugin could pre-configure the project with all your conventions and defaults and the user can still customize what they want. But mostly they won't have to and the build scripts will be reduced to a small fraction of their current size.

@atomfrede agreed to start writing a plugin and we will meet again in three weeks from now to talk about his experience and how to proceed.

I think it would make sense to break out plugins for each optional aspect (like gatling). That way the build script would mirror the generator config pretty much one for one. Here's what that would look like:

```
apply plugin: 'jhipster'
apply plugin: 'jhipster-gatling'
apply plugin: 'jhipster-mapstruct'

bootRepackage {
   mainClass = 'io.github.jhipster.application.JhipsterGradleSampleApplicationApp'
}
```

## Tasks are always out-of-date

Gradle has great incremental build support, but unfortunately the JHipster project doesn't benefit from it for several reasons:

- the Spring Boot build info put into the JAR contains a ms timestamp, making the JAR always out-of-date
- the Spring Boot `bootRepackage` task overwrites the WAR file, which makes the war task out-of-date (which then overwrites it again and so on...)
- the build script contains `processResources.dependsOn cleanResources`. Gradle is actually very good at detecting and cleaning stale output files, so this might be a historic artifact

The Spring team is aware of the issues and we will stay in touch with them to ensure they are resolved in the new version of the Boot plugin.

## Dependency resolution is slow

The Spring Dependency Management plugin re-parses BOMs on every build, a very costly operation, which adds about 500ms to every build. 

After hacking around the issues above I could reduce incremental build times from 5s to 200ms. @atomfrede said the build takes 43s on his machine. Getting a [build scan](https://scans.gradle.com) of that would be helpful.

Again, nothing much you can do here, but we are in touch with the Spring team. 

## Overuse of `-P` properties

[...]

PR #6380, opened 2017-09-19
WIP: separate testing tasks

atomfrede:
* Seperate UnitTests and SB Integrationtests and BDD tests
* all tests tasks can be executed indepent from each other
* Need to execute `check` instead of just test now, therefore update the travis script

update #6252 


travis/scripts/04-tests.sh
```diff
@@ -28,7 +28,7 @@ if [ -f "mvnw" ]; then
         -Dlogging.level.io.github.jhipster.sample=ERROR \
         -Dlogging.level.io.github.jhipster.travis=ERROR
 elif [ -f "gradlew" ]; then
-    ./gradlew test \
+    ./gradlew check \
```
deepu105: needs update to docs as well, also this is a breaking change, which is not nice for a non major release

atomfrede: Both your points are valid. Doing a 4.9.0 would be the way to go I guess. Same as updating the maven docker plugin when it was changed from `docker:build` to `dockerfile:build`

atomfrede:
Note: ci-cd generator should use `check` or separate tasks when using gradle

travis/scripts/04-tests.sh
deepu105: hmmm

atomfrede:
Update ci templates to use check instead of test. Removed the obsolete 
```
clean {
    delete "target"
}
```
 from `build.gradle` which was used to clean h2 db files,which are now also under `build` directory.

travis/scripts/04-tests.sh
jdubois: I'm also reluctant here: this will annoy a lot of people, and "test" is also the task name we use for Maven and Yarn

deepu105: I tend to agree with @jdubois 
I know its not the correct convention but I guess many of us are so used to running only the test tasks and since the test task still exists it will result it in we still running it and forgetting about the integration tests until something breaks in a CI

atomfrede: Okay then. Let's keep test execute all unit and spring integration tests. What about cucumber? I would execute it via check and not test.

deepu105: that sounds fine

atomfrede:
Will try to clean it up during the weekend.

atomfrede:
Update the testing stuff. Java tests (Unit and Spring Integration) are executed with `test`. BDD Tests via cucumber are executed during `check` phase.

generators/ci-cd/templates/_.gitlab-ci.yml
```diff
@@ -58,8 +58,8 @@ before_script:
 gradle-build:
     stage: build
     script:
-        - ./gradlew compileJava -x check -PnodeInstall --no-daemon
-gradle-test:
+        - ./gradlew compileJava -x test -PnodeInstall --no-daemon
+gradle-check:
```
deepu105: may be this can remain hradle-test as it is executing the test task

atomfrede:
@deepu105 No changes left anymore. From my point of view we can merge this

PR #6383, opened 2017-09-20
Update css and styles to prevent white on white

henri-tremblay:
I'm not a CSS expert but some styles need to be updated to bootstrap 4. So at least now it is usable.


deepu105:
can you paste a screenshot of how it looks now(I dont have time to run and check now)

henri-tremblay:
Old look (badge-default isn't a valid tag anymore):
screen shot 2017-09-20 at 07 25 14

New look:
screen shot 2017-09-20 at 07 24 24

jdubois:
Thanks! I will also make this change on the JHipster Registry, which has the same screen (I'll link my commit here)

PR #6395, opened 2017-09-21
Fix to use the proper entity class on the client side

gzsombor:
not just a fake object, which has similar structure.

 This is useful, if the user adds methods to the entity class later.
The current implementation is confusing, because the user expects, that if he adds some code to the entity.model, then he can use it from the various components


deepu105:
LGTM, I didn't test it, hope you tested with and without date fileds

jdubois:
Also can't test this at the moment -> I'm merging this because it's so big we should see quickly if there's an issue

gzsombor:
I've tested it with the default sample app, it works fine.

PR #6405, opened 2017-09-25
[do not merge] Use violations instead of FieldErrorVM

henri-tremblay:
I decided to do a PR instead of a bug because I have 95% of the answer. So, I'm pretty sure `handleMethodArgumentNotValid` should return a `ConstraintViolationProblem`.  Because otherwise, all the `instanceof` in `process` are useless because it will never happen.

The only problem is that `Violation` doesn't contain and `ObjectName`. So the code in `alert-error.component.ts` currently doesn't work. I don't know how to fix it.


generators/server/templates/src/main/java/package/web/rest/errors/_ExceptionTranslator.java
```diff
@@ -88,18 +88,7 @@
     @Override
     public ResponseEntity<Problem> handleMethodArgumentNotValid(MethodArgumentNotValidException ex, @Nonnull NativeWebRequest request) {
         BindingResult result = ex.getBindingResult();
```
cbornet: You can just remove the overriden method completely

cbornet:
That's why I didn't use `violations` in the first place.

generators/server/templates/src/main/java/package/web/rest/errors/_ExceptionTranslator.java
henri-tremblay: Right...

henri-tremblay:
I guessed that. :-) My current thoughts are: Zalando seems happy with it so we probably can get rid of it. Right now it is only used for localization if I am right. Can't the objectName be passed to the `JhiAlertComponent`?

jdubois:
LGTM - @cbornet can this be merged or will that mess up with your PR #6411 ?

deepu105:
@jdubois but this creates a bug in the alert service so IMO this PR is not complete. I'll rather stick to current solution until there is a way to pass custom params in violation

generators/server/templates/src/test/java/package/web/rest/errors/_ExceptionTranslatorIntTest.java
```diff
@@ -91,9 +91,8 @@ public void testMethodArgumentNotValid() throws Exception {
              .andExpect(status().isBadRequest())
              .andExpect(content().contentType(MediaTypes.PROBLEM))
              .andExpect(jsonPath("$.message").value(ErrorConstants.ERR_VALIDATION))
-             .andExpect(jsonPath("$.fieldErrors.[0].objectName").value("testDTO"))
```
deepu105: This was used in i18n and I dont see the benefit of this PR if it removes existing functionality

cbornet:
Agreed. We cannot use `violations` yet. We should first PR to Zalando to at least add the error `code` that we use for the translations.

henri-tremblay:
Yes, @cbornet  and @deepu105  are right. The only thing I am wondering is: Isn't it possible to pass the objectName to the `JhiAlertComponent` used in the page? Or whatever translation context tag? It will remove the need of having it in the `Problem`.

deepu105:
@henri-tremblay if you can get it working that way then we can do it. Since the alert is from an interceptor I think there was no other clear way of passing these values

henri-tremblay:
That's the part I don't know. I'm only guessing about that since my knowledge of Typescript and Angular is quite limited.

cbornet:
I opened a PR https://github.com/zalando/problem-spring-web/pull/118 to add the BindingResult `code` field.

whiskeysierra:
I commented on the PR on our side also, bust just for reference. The constraint violation problem is defined as

> It’s **not** meant to be used for end-user input validation, but for client developer convenience. Any constraint violation problem happening in production should be considered a bug.
> 
> http://zalando.github.io/problem/constraint-violation/

I therefore believe that it's not the right thing for your use case (unless we change its semantics).

jdubois:
I'm closing this as I believe @whiskeysierra knows this better than all of us

Issue #6397, opened 2017-09-22
[Feature] Integrate Traefik into the microservice architecture


jdubois:
I am very satisfied by the Netflix stack, but it is not going fast enough for me: we've been waiting for Eureka 2 and Zuul 2 for months, and I don't see them being released anytime soon

Of course we already support Consul. What we also need is support for a Zuul competitor, so we can have a full Eureka + Zuul alternative.

Our architecture would change a bit: the current gateway would be replaced, and become a "front-end" that is served by the new gateway. Not a big change, in fact, just moving a box.

I see several alternatives: Apache (which I use for https://start.jhipster.tech ), [HAProxy](http://www.haproxy.org/) (I've never used it), [Fabio](https://github.com/fabiolb/fabio) (which is used by my current client), and [Traeffic](https://github.com/containous/traefik).

Traeffic seems to have a lot of traction today, and I have some interesting news about it:

- There's a team at Ippon who is ready to do that support on Tuesday next week
- I'm speaking with @emilevauge (the Traeffic lead dev) on Monday next week

-> So my goal is to have this support quite soon, is this fine with everyone?

PierreBesson:
I have been thinking of Traefik support for some time as well. For me Traefik is great as it can be complementary to Zuul. For example you can use Traefik eureka back-end to load balance between several zuul gateways.

However I'm not ready to move away from Zuul completely as Zuul offer many powerful API management features. Then it can be an interesting option to have Traefik powered gateway. So :+1:

xetys:
I had already the first contact with traefik as k8s ingress solution and I
like the light weight of it. I'm not a big fan of TOML, but if it really
works well with Eureka and consul backend, it is a competitive alternative
to eureka + Zuul, as both have a high resource consumption and bootup time,
compared to consul + traefik.

So +1 from me

gmarziou:
The things I like in Eureka: I know it :) and it"s written in Java integrated in Spring Cloud. What I don't like: slow startup, memory leaks that require to restart the registry every month or so.

Consul is great, starts fast, integrated in Spring Cloud but written in Go. Commercial support is expensive.

The fact that the gateway is a java app does matter a lot to me as I have been able to customize or add some endpoints, filters.

xetys:
And as of #5812 our gateway is also handling JWT over http cookies, what
enables a great experience with uaa in general. Another point is the rate
limiting. So if there will be traefik support, that solution will lack of
some features our gateway currently has

[...]

PR #6412, opened 2017-09-26
[WIP] Add Traefik support

jdubois:
This work was done with @ptavares @Yannoo @latourna @Gregoriz @djahanshahi

This is a work in progress:
- When Traefik starts faster than Consul, it doesn't sync with it (and needs to be restarted to work)
- The gateway doesn't work well, as it's not on the "/" path anymore -> the easiest solution would be to map it to the "/" path, but we haven't found how

See #6397

### DIFF
--- a/generators/client/templates/angular/webpack/_webpack.common.js
+++ b/generators/client/templates/angular/webpack/_webpack.common.js
@@ -33,7 +33,7 @@ module.exports = (options) => {
         // The root URL for API calls, ending with a '/' - for example: `"http://api.jhipster.tech:8081/"`
         // In `prod` mode, you will need to enable CORS on your API server
         // (see the `jhipster.cors` common JHipster property in the `application-*.yml` configurations)
-        SERVER_API_URL: `"/"`<% } %>
+        SERVER_API_URL: `""`<% } %>
     };
     return {
         resolve: {
--- a/generators/client/templates/angular/webpack/_webpack.common.js
+++ b/generators/client/templates/angular/webpack/_webpack.common.js
@@ -30,8 +30,9 @@ module.exports = (options) => {
     const DATAS = {
         VERSION: `'${utils.parseVersion()}'`,
         DEBUG_INFO_ENABLED: options.env === 'development'<% if (authenticationType !== 'uaa') { %>,
-        // The root URL for API calls, ending with a '/' - for example: `"http://api.jhipster.tech:8081/"`
-        // In `prod` mode, you will need to enable CORS on your API server
+        // The root URL for API calls, ending with a '/' - for example: `"http://www.jhipster.tech:8081/myservice/"`.
+        // If this URL is left empty (""), then it will be relative to the current context.
+        // If you use an API server, in `prod` mode, you will need to enable CORS  
         // (see the `jhipster.cors` common JHipster property in the `application-*.yml` configurations)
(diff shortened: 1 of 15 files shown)


pascalgrimaud:
yeah @ptavares ! finally, 1st contribution :)

Gregoriz:
About gateway's URL problem, traefik provides configuration for URL rewrite:

http://docs.traefik.io/configuration/entrypoints/#rewriting-url

One solution could be to redirect the `/` endpoint to `/gateway`

jdubois:
Thanks @Gregoriz ! I would love to have something like:
```
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    regex = "^http://localhost/gateway/(.*)"
    replacement = "http://localhost/$1"
```
However I have 2 issues:
- I'm not sure of the format here, as I don't know what hostname to write
- I just can't test: Docker4Mac crashes so bad on my Mac, that I need to reinstall it completely several times per day... That's probably a trick from Docker, Inc to increase their number of pulls :-)

Gregoriz:
The configuration you provided seems to take any URL starting with `http://localhost/gateway/` and remove the `gateway` part.

I was more thinhink about something like that:

```
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    regex = "^http://localhost[/]?$"
    replacement = "http://localhost/gateway/"
```
This would redirect `http://localhost/`  or `http://localhost` to `http://localhost/gateway`.
However, as soon as a service is called, then no redirection would be applied.
Ex:  `http://localhost/micro1` would not be redirected to `http://localhost/gateway`

I don't have enough knowledge in JHipster to think about any side effects which could be caused by such a redirection rule.

jdubois:
I just found a much simpler solution (I wonder why I didn't think of this!!!!):

- On the "JHipster gateway", use the new `SERVER_API_URL` (configured in Webpack), so it is just "" (currently it is "/", so it doesn't use relative URLs -> I'm the one who coded that, and I have no idea why!!!!!). With this change, URLs are relatives, so the whole `http://127.0.0.1/gateway/` application works!
- As the "JHipster gateway" uses its internal Zuul to route microservices URLs, you then need:
  - Either keep the current config, so it uses `http://127.0.0.1/gateway/microservice1/` -> it goes through Traefik and then trough Zuul. This could be interesting for rate limiting, for example.
  - Or configure the Angular service for each entity to use an absolute URL (just add `/` again), so it will use `http://127.0.0.1/microservice1/` -> this looks better to me

So in the end it's just a simple work on having a `/` or not, this shouldn't be hard to configure in a template!

jdubois:
I've just added 2 new commits corresponding to my previous comment:

- The URL prefix for all apps is now "", so it's relative: it should work everywhere, and should also allow to deploy monoliths to a non-root context.
- For microservices, there is now an absolute API starting with "/", as this is how they are routed by Zuul or Traefik. For example, a "micro1" microservice is available through the "/micro1" route.

jdubois:
I'm merging this as it has only very limited impact on the rest of the code (only the relative URLs are a big change, but I think it's better for everyone anyway).

TODOs:

- Documentation (I'll do it today probably)
- Maybe add this feature to Kubernetes (cc @saturnism ) and Rancher (cc @stevehouel )? Shouldn't be hard to

PR #6421, opened 2017-09-27
Fixing the auth-interceptor

gzsombor:
Fixing the js problem "ERROR TypeError: _this.authServerProvider.logout is not a function
    at CatchSubscriber.eval [as selector] (VM8171 auth-expired.interceptor.ts:42)
    at CatchSubscriber.error (VM7476 catch.js:104)..."

Which caused by wrong dependency declarations - which cause that in AuthExpiredInterceptor
 the loginServiceModal is null, and authServerProvider is actually a JhiEventManager.
Unfortunately injecting the AuthServerProvider would cause a cyclic dependency, so the Injector is used,
and the AuthServerProvider is acquired lazily

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_auth-expired.interceptor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_auth-expired.interceptor.ts
@@ -17,9 +17,7 @@
  limitations under the License.
 -%>
 import { JhiHttpInterceptor } from 'ng-jhipster';
-<%_ if (authenticationType === 'oauth2' || authenticationType === 'jwt' || authenticationType === 'uaa') { _%>
 import { Injector } from '@angular/core';
-<%_ } _%>
 import { RequestOptionsArgs, Response } from '@angular/http';
 import { Observable } from 'rxjs/Observable';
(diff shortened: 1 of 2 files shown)


gzsombor:
I haven't checked, I suspect that it's never worked, as using the AuthServerProvider in that interceptor would created a cyclic dependendency, which would created a noticeable failure at app startup :-)

deepu105:
I'm surprised. If it caused a failure I'm pretty sure at least some of us
would have noticed as it seems to be for all auth types


Thanks & Regards,
Deepu

gzsombor:
With the original code, the error is only triggered, when the browser receives a 403 error from other than the api/account page. This is not a very common thing, and in that case, nothing highly visible happened. Just a stack trace in the console - without logging out, and showing up the login dialog - but I'm not sure, that login dialog is an always appropriate solution. 
 I've found it by accident, because I've played with this logout/login stuff in different windows - sure, it's a good recipe for finding bugs :-).
 First, I've tried to fix the dependency, to require AuthServerProvider, then I've noticed, that this cause a serious problem, with the cyclic reference error. Which is a much more noticeable thing. That's why I guess, when this was coded, and the error encountered, only the dependency is removed - and as it compiled fine, it seemed a good fix.

PR #6423, opened 2017-09-28
Update _messages_zh_cn.properties

gemron:
complete the remaining Chinese Translation


generators/generator-constants.js
```diff
@@ -94,7 +94,7 @@ const LANGUAGES = [
     },
     { name: 'Armenian', dispName: 'Հայերեն', value: 'hy' },
     { name: 'Catalan', dispName: 'Català', value: 'ca' },
-    { name: 'Chinese (Simplified)', dispName: '中文（简体）', value: 'zh-cn' },
+    { name: 'Chinese (Simplified)', dispName: '中文（简体）', value: 'zh-CN' },
```
PierreBesson: I dont think this line should be changed. This will cause problem when people upgrade their application.

gemron: yes , i seen client have a i18n floder.

deepu105: please revert this line

deepu105: Also please revert all file renames to uppercase

generators/languages/templates/src/main/resources/i18n/_messages_zh_CN.properties
```diff
@@ -30,18 +30,18 @@ email.activation.text2=\u795D\u60A8\u4F7F\u7528\u6109\u5FEB,
 email.signature=<%= baseName %> \u56e2\u961f.
 
 # Creation email
-email.creation.text1=Your <%= baseName %> account has been created, please click on the URL below to access it:
+email.creation.text1=\u60a8\u7684 <%= baseName %> \u0020\u0020\u8d26\u6237\u5df2\u7ecf\u521b\u5efa\uff0c\u0020\u8bf7\u70b9\u51fb\u4ee5\u540e\u94fe\u63a5\u521b\u5efa\uff1a
```
deepu105: You can use the actual characters here as this looks very cryptic and is not user friendly

generators/generator-constants.js
gemron: so  modify   this line   https://github.com/jhipster/generator-jhipster/blob/efc1e35ef1c0dae087c613760cf2c475447349d7/generators/generator-base-private.js#L91 
ok?

gemron: this is  JDK 8 and JRE 8 Supported Locales
http://www.oracle.com/technetwork/java/javase/java8locales-2095355.html

deepu105: No. Why are you changing this in the first place?

gemron: so we can do a transfer,ok ?

deepu105:
It is still not acceptable, why can't you just use standard Chinese chars like in other files and you need to rename the file back to lowercase as well

gemron:
in locale  cn is CN, so the i18n  file message_zh_cn.properties is not work.  when i change this file name to message_zh_CN.properties,it is working

gemron:
but when i change this file's name ,this line will error.   https://github.com/jhipster/generator-jhipster/blob/efc1e35ef1c0dae087c613760cf2c475447349d7/generators/generator-base-private.js#L91

deepu105:
Oh, ok. This is why we ask contributors to explain clearly what they are doing. @PierreBesson you seem to get this so would you like to review and take care of this

jdubois:
Yes I think you need the upper case for CN, but I don't know the reason why.
And you are correct, @PierreBesson is a man of many skills, and speaks Chinese fluently, so he should be able to help :-)

gemron:
this is JDK 8 and JRE 8 Supported Locales  ,http://www.oracle.com/technetwork/java/javase/java8locales-2095355.html 
 I think the jhipster's language value on the server side should be converted to the same value as the locale of Java

PierreBesson:
@deepu105 I'm on it.

I did some tests today and indeed there is something not working here with email i18n. Also @gemron is correct that it should be `zh_CN` as this is the value in the `java.util.Locale enum (which we should be using by the way).

There are some improvements we can do to this part as there is currently no tests on i18n emails and they seem broken currently. But it is not simple to manage both client and server i18n as java and JavaScript don't use the same conventions...

PierreBesson:
I just did some more tests during my lunch break and indeed our code works fine for simple locales like "en" and "fr" but is broken for composed locales such as "zh_CN" or "pt_BR". `Locale.forLanguageTag()` is not able to convert those string to java locales.
I will open an issue to discuss migrating our server localisation handling from `String` to `Java.util.Locale`.
@jdubois The issue should only affect a some of our supported languages that uses composed codes : chinese, arabic and portuguese, I think.

@gemron Even though it doesn't work completely yet, please rename the file to the original name (lowercase `cn`). I will squash and merge your PR to have the translated strings.

jdubois:
Thanks @PierreBesson !

gemron:
@PierreBesson ok.

Issue #6362, opened 2017-09-15
Upgrade to Yarn 1.0


jdubois:
Yarn 1.0 was just released, see https://github.com/yarnpkg/yarn/releases/tag/v1.0.0

Our current build works with the 0.27.5, but fails with 1.0.0. This needs to be fixed.

jdubois:
And maybe it's just me, but for developing JHipster, Yarn 1.0.0 is totally broken... `yarn link generator-jhipster` doesn't work correctly anymore:

- After running it, I have errors in my build (saying there are some errors in generator-jhipster, like `error An unexpected error occurred: "ENOENT: no such file or directory`)
- To solve this, I need to run `yarn install` in generator-jhipster
- And after that, it is like `yarn link` didn't work, I have a non-development release of JHipster

Time to test NPM 5 again!!! Maybe move totally to NPM 5? It will ease installation, too.

gmarziou:
npm5 works fine except that it's not packaged by default with node 6.x

gmarziou:
yarn link has some open issues with binaries, so you might be better not using the cli but using yo.

jdubois:
Yes, but I'm using Node 8 without any trouble for some time now... I now we always recommend LTS, but maybe it's time to change. The main benefit for me is that there is one less thing to install for end-users - and for most Java developers, all this Node stuff is quite frightening.

Then, I would like to test what we can do with NPM 5. I have one big issue with our current build: it takes a long time to generate because it downloads some non-cacheable assets like Sass. Most of those are useless (unless you select Sass, of course), and come as transitive dependencies (Sass comes from Angular CLI), so I would like to be able to exclude them. I haven't found a way to do it yet, and this would speed up and ease the build for a lot of people. Also, I would love to be able to do an "offline build", that would be so helpful for demos and conferences where you have no Wifi.

PierreBesson:
Anyway, we still need to wait for a release of the front-end-maven plugin, so no hurry. Also personally, I have no problem building a JHipster app with my locally installed yarn 1.0.1.

gmarziou:
[Node 8 will be LTS in October](https://github.com/nodejs/LTS#lts-schedule1), so I'd prefer we wait for so that our LTS version recommendation stays clear.

pascalgrimaud:
New release of frontend-maven-plugin: https://github.com/eirslett/frontend-maven-plugin/releases/tag/frontend-plugins-1.6
[...]

PR #6424, opened 2017-09-28
Upgrade Yarn, NodeJS, NPM, frontend-maven-plugin

pascalgrimaud:
- Yarn 1.0.2
- NodeJS 6.11.3
- NPM 5.4.2
- frontend-maven-plugin 1.6
- remove cache, and use global node 6.11.3
- use yarn global add file, instead of yarn link

Fix https://github.com/jhipster/generator-jhipster/issues/6362

_____

- [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green -> https://travis-ci.org/pascalgrimaud/generator-jhipster/builds/280601366

### DIFF
--- a/travis/scripts/02-generate-project.sh
+++ b/travis/scripts/02-generate-project.sh
@@ -24,6 +24,5 @@ fi
 mkdir -p "$APP_FOLDER"
 cp -f "$JHIPSTER_SAMPLES"/"$JHIPSTER"/.yo-rc.json "$APP_FOLDER"/
 cd "$APP_FOLDER"
-yarn link generator-jhipster
 jhipster --force --no-insight --skip-checks --with-entities
 ls -al "$APP_FOLDER"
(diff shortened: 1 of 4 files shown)


travis/scripts/02-generate-project.sh
```diff
@@ -16,14 +16,12 @@ if [ "$JHIPSTER" == "app-ng2-gateway-uaa" ]; then
     mkdir -p "$UAA_APP_FOLDER"
     cp -f "$JHIPSTER_SAMPLES"/uaa/.yo-rc.json "$UAA_APP_FOLDER"/
     cd "$UAA_APP_FOLDER"
-    yarn link generator-jhipster
```
deepu105: how does this link happen now? if yarn link doesnt work we should just use `npm link` here, they both can co exist without issues

jdubois:
Thanks @pascalgrimaud

travis/scripts/02-generate-project.sh
pascalgrimaud: In fact, in Travis, I don't use `yarn link`, but use this instead: `yarn global add file:"$TRAVIS_BUILD_DIR"`, so it will install globally the version from this folder (the master version)

`yarn link` seems to be totally broken with Travis, and sometimes in my local too. When it happens, I need to clear some folders in `~/.config/yarn` and reinstall everything.

pascalgrimaud: Here the new line: https://github.com/jhipster/generator-jhipster/blob/master/.travis.yml#L72

deepu105: Just use `npm link` it works fine right

deepu105: the only command where Yarn shines is when you install stuff for everything else npm commands are perfectly fine as well

PR #6435, opened 2017-09-29
OIDC support for Microservices architecture

danielpetisme:


danielpetisme:
As @mraible explained #6435 , I started a microservices support for Oauth2.
This is a global status.

So far, I succeed to generates a Gateway with OAuth2Sso enabled (redirects the user to Keycloak for authentication and successfully load principal and authorities) and ResourceServer (API's gateways's endpoints are accessible via a http client js or curl or whatever). In order combine both features I had to define 2 mutually excluding configurations). The gateway also acts as a token relay to provide jwt to the backend microservices.
I have one known issue, the swagger api pages can't be loaded...
The microservices have a bearer-only oauth2 client configuration meaning that microservice only verifies that the token is valid but can't generate one. The user has to get it before. Microservices also load principal and authorities so the dev can finedly manage the microservice enpoints permissions.

In monolith applications when the authenticationType is oauth2 a set of user management files was generated. Currently in a microservice architecture I skip the user related files (everything is delagated to the IdP).

Regarding the code itself, it does work and the existing generator-jhipster test work. I didn't added new ones so far. Regarding the generated apps I didn't run the tests suite or generate all the combination.

jdubois:
Awesome! We need to find a solution for the Swagger doc, but this is a minor point compared to the rest (for a first release, that's not a blocker issue)

danielpetisme:
If we follow UAA pattern, none of te gateway or services components expose user-management endpoint (/account, /register, etc).

For me there is 3 options: 
1. Change the front app code with the corresponding provider endpoints. I don't like this solutions, 
2. Play with Zuul to reverse proxy the requests (ie change `/account` into `http://localhost:9080/auth/realms/jhipster/protocol/openid-connect/auth`). The benefit is that there is no code to maintain but seems it's purely declarative I don't know how far we can go.
3. Add a intermediate layer on server side (actually on the gateway it would make sense). We coud use generic properties like `security.oauth2.resource.userInfoUri` to have something quite generic. 

At the moment I think option#3 it's the cleaner from an architecture stand point, it's coherent with the gateway orchestration responsability... but it adds responsibility ro this component.

What do you think?

jdubois:
- I'm very interested by this feature - unfortunately I don't have much time this month, but I'll try to help as much as possible
- To simplify things, maybe some of those classes shouldn't be generated, and be in https://github.com/jhipster/jhipster instead
- Unfortunately the Travis build fails -> is it related to your latest comment? I'm not expert here, but indeed point 3 looks good - maybe @mraible you could help on this?

-> outside of those comments, how can I help you?

sdoxsee:
@danielpetisme not entirely sure if I understand but here are some thoughts. Option 2 and/or 3 sound best to me. Do you have to choose? Here's an example gateway configuration where it uses [zuul](https://github.com/kbastani/spring-cloud-event-sourcing-example/blob/master/online-store-web/src/main/resources/application.yml#L17) AND [security.oauth2.resource.userInfoUri](https://github.com/kbastani/spring-cloud-event-sourcing-example/blob/master/online-store-web/src/main/resources/application.yml#L22).

danielpetisme:
@sdoxsee you're right I can actually use both approaches but I'm not sure it's consistent
First of all, I don't want to build a uaa-proxy thing so only the api gateway should discuss with the oauth provider (keycloak or okta or whatever).
I think it's important to keep the frontend application agnostic of the oauth provider. So I'm targeting to keep the following endpoints:

capture d ecran 2017-10-06 a 23 31 23

That means that I could use to Zuul to map some urls. For instance a Zuul rule to map /api/account to `security.oauth2.resource.userInfoUri` but when it comes to password management the Jhipster process is not excatly mappable to the keycloak one, or least from what I have seen.
Rather than spreading some part in a config file and the other part in code I prefer to centralize everything.

I'm not a Zuul expert so feel free to correct me.

NOTE: Another option could simply be to say that User management isn't a Jhipster concern anymore and it's up to the dev to integrate it's oauth provider into the generated app. It's not the proposition I prefer but it's a pramatic (radical) option.

danielpetisme:
@jdubois 
Most of all I need some reviews to validate the architecture:
- Security
  - Are the clients definitions and their usage between the gateway and microservices correct?
  - Is the Token relaying made by the gateway safe (the communication between the gateway and services is made via HTTP so the token isn't encrypted... almost all the gw-services communicatoins are made in a private trusted network but security matters...)
- General questions
  - To be able to use `@EnableOauthSso` and `@EnableResourceServer` I had to use a silly hack to split the api access and the classic HTMLs assets. They are mutually excluded. I'm not comfortable with this snippet but it's the best solution I found, so I would definitively welcome enhancement remarks
  - Rather than hardcoding the principal and authority extraction I've added a default implementation `Principal and Authorities Extractor` based on a configurable property. It's one more property to document and support but I think it would ease the integration or further Oauth provider. What you think about this. 

Regarding the Travis CI status, I don't understand the issues (it failed on `app-gradle` for instance but I didn't touched anything related to gradle). As former Jenkins Administrators it hurts me to say that but... it works on my machine 😄 

I'm currently running the test suite... we'll see

sdoxsee:
I'd say that with oidc, user management is not longer a jhipster concern since many oidc providers also have user management but feel free to push back on that...as long as jhipster can handle the role and permissions from the IdP.

danielpetisme:
The CI fails on some functional testing. It fails to set property a date in the date time fields. 
Example of failing snippet:
```
operationDialogPage.setDateInput(12310020012301);
expect(operationDialogPage.getDateInput()).toMatch('2001-12-31T02:30');
```
I actually didn't touched those files.
If I compare the `sample-apps` hosted on github and the project generate by Travis script I've got a difference on the `protractor.conf.js` file.
Travis generated
```
capabilities: {
        'browserName': 'chrome',
        'phantomjs.binary.path': require('phantomjs-prebuilt').path,
        'phantomjs.ghostdriver.cli.args': ['--loglevel=DEBUG']
    },
```
Sample App
```
capabilities: {
        'browserName': 'chrome',
        chromeOptions: {
            args: [ "--headless", "--disable-gpu", "--window-size=800,600" ]
        }
    },
```
If I use the sample app settings the tests succeed. With the travis config a browser pops up.

Do you have any clue where this difference comes from? Did you faced it?

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -183,3 +183,127 @@ public RestTemplate vanillaRestTemplate() {
...
+
+import java.util.Map;
+import java.util.Optional;
+
+@Configuration
+@EnableResourceServer
+@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
+public class MicroserviceSecurityConfiguration extends ResourceServerConfigurerAdapter {
+
+
```
mraible: Remove extra line break.

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -183,3 +183,127 @@ public RestTemplate vanillaRestTemplate() {
...
+<%_ if(applicationType === 'gateway') { _%>
+    @Bean
+    @Qualifier("authorizationHeaderRequestMatcher")
+    public RequestMatcher authorizationHeaderRequestMatcher() {
+        return new RequestHeaderRequestMatcher("Authorization");
+    }
+<%_ } _%>
+
+
```
mraible: Remove extra line break.

generators/server/templates/src/main/java/package/config/_MicroserviceSecurityConfiguration.java
```diff
@@ -183,3 +183,127 @@ public RestTemplate vanillaRestTemplate() {
...
+        http
+            .csrf()
+            .disable()
+            .headers()
+            .frameOptions()
+            .disable()
+            .and()
+            .sessionManagement()
+            .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
+            .and()
```
mraible: I'd recommend out-denting the `.and`s in this code block, like https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/_SecurityConfiguration.java#L191.

generators/server/prompts.js
```diff
@@ -422,15 +426,16 @@ function askForServerSideOpts(meta) {
...
+        // user-management will be handled by UAA app, oauth expects users to be managed in IpP
+        if ((this.applicationType === 'gateway' && this.authenticationType === 'uaa') || this.authenticationType === 'oauth2') {
             this.skipUserManagement = true;
         }
 
         if (this.applicationType === 'uaa') {
             this.authenticationType = 'uaa';
         }
 
+
```
mraible: Remove extra line break.

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -186,19 +189,31 @@ security:
...
+            clientSecret: internal
+            authenticationScheme: header
+            clientAuthenticationScheme: header
+            <%_ } _%>
         resource:
             userInfoUri: ${oauth2.issuer}/protocol/openid-connect/userinfo
             tokenInfoUri: ${oauth2.issuer}/protocol/openid-connect/token/introspectr
             preferTokenInfo: false
+            jwt:
+                keyUri: ${oauth2.issuer}
```
mraible: These lines aren't needed for a monolith, so should only be included for microservice architecture.

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -186,19 +189,31 @@ security:
...
         resource:
             filter-order: 3
-    <%_ } else if (authenticationType === 'oauth2') { _%>
+    <%_ } _%>
+    <%_ if (authenticationType === 'oauth2') {  _%>
     oauth2:
         client:
             accessTokenUri: ${oauth2.issuer}/protocol/openid-connect/token
             userAuthorizationUri: ${oauth2.issuer}/protocol/openid-connect/auth
+            <%_ if (applicationType === 'gateway') { _%>
```
mraible: `clientId`, `clientSecret`, `clientAuthenticationScheme`, and `scope` should also be included for a monolith, but `tokenName` is not needed.

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -177,7 +177,10 @@ spring:
 # export SECURITY_OAUTH2_CLIENT_CLIENT_SECRET="38561a05"
 # ===================================================================
 
-oauth2.issuer: http://localhost:9080/auth/realms/jhipster
+oauth2:
+    issuer: http://localhost:9080/auth/realms/jhipster
+    principal-attribute: preferred_username
+    authorities-attribute: roles
```
mraible: `principal-attribute` and `authorities-attribute` not needed for monolith.

generators/server/templates/src/main/resources/config/_application.yml
```diff
@@ -186,19 +189,31 @@ security:
...
+            <%_ } _%>
+            <%_ if (applicationType === 'microservice') { _%>
+            clientId: internal
+            clientSecret: internal
+            authenticationScheme: header
+            clientAuthenticationScheme: header
+            <%_ } _%>
         resource:
             userInfoUri: ${oauth2.issuer}/protocol/openid-connect/userinfo
             tokenInfoUri: ${oauth2.issuer}/protocol/openid-connect/token/introspectr
```
mraible: According to http://localhost:9080/auth/realms/jhipster/.well-known/openid-configuration, the `tokenInfoUri` should be `http://localhost:9080/auth/realms/jhipster/protocol/openid-connect/token/introspect`. This is a mistake I made in my PR. However,, it doesn't seem to matter what the value is.

generators/server/templates/_build.gradle
```diff
@@ -360,6 +360,7 @@ dependencies {
     <%_ } _%>
     <%_ if (authenticationType === 'oauth2') { _%>
     compile "org.springframework.security.oauth:spring-security-oauth2:${spring_security_oauth2_version}"
+    compile "org.springframework.security:spring-security-jwt"
```
mraible: This isn't needed when using OAuth 2.0 and a monolith.

generators/server/templates/_pom.xml
```diff
@@ -708,6 +708,10 @@
             <groupId>org.springframework.security.oauth</groupId>
             <artifactId>spring-security-oauth2</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-jwt</artifactId>
+        </dependency>
```
mraible: This isn't needed when using OAuth 2.0 and a monolith.

danielpetisme:
The test about oauth2 are not failing anymore now it's eureka. It's strange because it fails on things like:
```
2017-10-10 07:28:50.867 ERROR 8527 --- [freshExecutor-0] com.netflix.discovery.DiscoveryClient    : DiscoveryClient_TRAVISDEFAULT/travisDefault:a534e8834a2417a16ad7c40bdec5fa69 - was unable to refresh its cache! status = Cannot execute request on any known server
```
```
* Rebuilt URL to: http://localhost:8080/
* Hostname was NOT found in DNS cache
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1...
* connect to 127.0.0.1 port 8080 failed: Connection refused
* Failed to connect to localhost port 8080: Connection refused
* Closing connection 0
```

I'm not 100% sure it's related to the code itself.

mraible:
Just noticed a regression. When selecting `oauth2` and monolith, AngularJS should not be listed as a choice. @deepu105 implemented this.

screen shot 2017-10-09 at 3 45 10 pm

mraible:
More testing... when creating a microservice gateway, shouldn't we prompt for JHipster Registry vs. Consul? The answers don't seem to line up:

```
? (5/16) Which service discovery server do you want to use? (Use arrow keys)
  Yes
  Consul
  No service discovery
```

jdubois:
@mraible maybe you're not on the right branch, or had an issue with `yarn link` (it happens all the time to me!) -> the current code looks correct for the `Which service discovery server do you want to use?` question

I need all this for my current client: I'm going to do a quick review

jdubois:
- The build failed because there was a JVM crash, I restarted it and it works OK
- @danielpetisme I'm not an expert and I can't answer all questions here: yes, there is no problem in exchanging JWT token over HTTP inside the "private" network. 1) that's supposed to be secured at the network level 2) the JWT token are signed and 3) you can easily add HTTPS

danielpetisme:
@jdubois What I propose in a first release is to remove all the user-management related files (as proposed by @sdoxsee). That means no more account settings page, account creation or users settings and user. The end user will have to go to the IdP to create an account or deal with his settings.
This config would be the equivalent of a future `--skip-user-management` settings of oauth2.

In a second shot, we can work on a Jhipster support of the account/users UI + IdP API orchestration in the gateway.

What do you think guys?

jdubois:
@danielpetisme yes for me all the user account management is handled by Keycloak/Okta, that's one of the main points of using this method. Like we do in JHipster UAA. So yes you should "skip user management".

danielpetisme:
The uaa deals with user management but since it's a jhipster app it's registered in the registry and accessible via the gateaway and visible from the gateway UI.

I'll update the code tonight to not generate the user related files.

jdubois:
Oh yes, let me detail:

- The most common use case for me: users will be created elsewhere, and for example they will be stored in an LDAP server or a database -> the main role of Keycloak and Okta is to aggregate those sources of users, and give authentication tokens to the applications

Now for Keycloak only, as it can be used as a Docker image:
- Keycloak port should be exposed by default in Docker-Compose -> so people can create a user, etc... directly using it
- If you use Traefik, Keycloak should also be routed through it (and that would be very close to what you have with a gateway + JHipster UAA)

mraible:
I agree that users should be created elsewhere. However, in the monolith implementation, I keep some of the user-related classes. My reason is so there can be a `User` entity that people can create relationships with. How I do this is I sync the User from the IdP with the local database's user when `/api/account` is called. See the logic [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java#L131). FWIW, I've sent this code to the Spring Security folks for a code review.

jdubois:
I've reviewed the code in more detail, and my main issue is this configuration from @mraible 

```
oauth2:
    issuer: http://localhost:9080/auth/realms/jhipster
    principal-attribute: preferred_username
    authorities-attribute: roles
```

This configuration is not type safe, and to be consistent with the rest of the code: it should be stored in https://github.com/jhipster/jhipster/blob/master/src/main/java/io/github/jhipster/config/JHipsterProperties.java using a `jhipster` prefix -> let me do that, and do a new release of jhipster/jhipster so you can use it.

jdubois:
I just released https://github.com/jhipster/jhipster/releases/tag/v1.1.12 that includes the new OAuth2 configuration (see https://github.com/jhipster/jhipster/commit/6ff3a2259edee3581f4b42c693da8f84141ef88f )

-> you need to migrate the configuration to the new `jhipster.security.authentication.oauth2` prefix, and @mraible the documentation should be changed accordingly
-> @danielpetisme I can help on this part, don't hesitate to ping me

mraible:
@jdubois `oauth2.issuer` is not used by any code, it's only a shortcut for the other OAuth-related properties that Spring Security OAuth uses. The other two properties (`principal-attribute` and `authorities-attribute`) are only used by the microservices code at the moment.

jdubois:
Yes Matt I understand that - it's a specific case we haven't had before.
But it took me some time to understand: if it's coded and documented like
the rest, and uses a 'jhipster' prefix, I think it would be easier to
understand.
You would also have automatic completion with Idea: when this doesn't work
I'm always looking at what's wrong.

danielpetisme:
@jdubois @mraible I've mixed feelings for. Right now `oauth2.issuer` acts as a variable for the others oauth2 properties. I think we could simply duplicate this value, we're talking about 4 consecutives lines...

If we plan to pilot the IdP provider (registering for instance) from JHipster code, then this variable could be useful but it will raise more issues (It requires an implementation per provider).

Regarding the `principal-attribute` and `authorities-attribute` they're used in the extractor pattern I proposed. Matt had a code generation oriented strategy hardcoding the snippet where I tried to have a framework oriented approach using the underneath Spring piping. 
Don't know if we should merge the 2 approaches, use one or the other.

The OIDC monolith and microservices implementation are not 100% aligned  but for the sake of speed I propose to treat this point as a dedicated issue.

@mraible sounds good to you if we stay as-is in a first release?

jdubois:
yes this is just a variable used 4 times @danielpetisme - could we just remove it? And then probably just copy/paste it, that's OK in the YAML file (and not really our fault, this is just the Spring Security OAuth config that works this way). If that's fine, then I'll remove my config in the jhipster/jhipster library.

danielpetisme:
+1 for me

mraible:
@jdubois If we move it, we'll need to change [_app.yml](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/docker/_app.yml) from:

```
- OAUTH2_ISSUER=http://keycloak:9080/auth/realms/jhipster
```

To:

```
- SECURITY_OAUTH2_CLIENT_ACCESS_TOKEN_URI = http://keycloak:9080/auth/realms/jhipster/protocol/openid-connect/token
- SECURITY_OAUTH2_CLIENT_USER_AUTHORIZATION_URI = http://keycloak:9080/auth/realms/jhipster/protocol/openid-connect/auth
- SECURITY_OAUTH2_RESOURCE_USER_AUTHORIZATION_URI = http://keycloak:9080/auth/realms/jhipster/protocol/openid-connect/userinfo
- SECURITY_OAUTH2_RESOURCE_TOKEN_INFO_URI = http://keycloak:9080/auth/realms/jhipster/protocol/openid-connect/introspect
```

We might need to update some other documentation too.

mraible:
@danielpetisme Can you please provide the steps you're using the ensure this PR works? I want to compare your methodology with the one I'm using. Here's what I expect to be able to do:

1. Create a gateway with OAuth
2. Create a microservices app with OAuth
3. Create an entity in the microservices app
4. Create the UI for that entity in the gateway
5. Clone JHipster Registry and run it
6. Start Keycloak using docker compose (from the Gateway app's directory)
7. Start the gateway and the microservices app and login successfully

I think it should also be possible to use the docker-compose and kubernetes sub-generators to create a system that'll start all of the above services.

danielpetisme:
I'm not 100% done. If you want to try something simply 
1&2- generate a gateway and a microservice as usual (using oauth2 of course).
3- Create an entity as you do usually
4-5 Do as usual
6- The keycloak docker-compose file is generated with in the gateway or the microservice
7- If you start everything you'll see that the login redirects to the IdP but the gatewat UI is not updated (ie. you see the sign in link again and you seems to be logged in). 
Actually you are logged in but the roles are not loaded in the front app. The explication comes from the lack of user-management code in the backend. The front tries to retrieve `/api/account`.
You defined for monolith a version and the`uaa` project also.
If you want to make it work you must provide an implementation that's one :
Add a `AccountResource.java` with the following method
```
public ResponseEntity<UserWithAuthorities> getAccount(Principal principal) {
        return Optional.ofNullable(principal)
            .filter(it -> it instanceof OAuth2Authentication)
            .map(it -> ((OAuth2Authentication) it).getUserAuthentication())
            .map(authentication ->
                new UserWithAuthorities(
                    authentication.getName(),
                    authentication.getAuthorities().stream().map(it -> it.getAuthority()).collect(Collectors.toSet())
                )
            )
            .map(user -> new ResponseEntity<>(user, HttpStatus.OK))
            .orElse(new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR));
    }
```
and create the according `UserWithAuthorities`
```
public class UserWithAuthorities {
    private final String login;
    private final Set<String> authorities;

    public UserWithAuthorities(String login, Set<String> authorities) {
        this.login = login;
        this.authorities = authorities;
    }

    public Set<String> getAuthorities() {
        return authorities;
    }

    public String getLogin() {
        return login;
    }
}
```
It will work. What bothers me is to have user-related code in the backend which only acts as a JWT decoded actually. All the information we are looking for are stored in the token answered by the IdP (roles, mail, etc).
What I would like to do is to remove this dependency to the backend and only use the jwt.
By decoding the JWT Token in the angular app will retrieve the roles, lang, etc. That means the token is at least in the session storage (cf. the discussion we had about token storage).
The hard work is to understand how AccountService, Principal and AuthExpired work in the front application to intercept to have something elegant (for instance the `Principal.identity` method should reload the jwt token rather than requesting `/api/account`.

That was for the UI. Beside that the gateway-microservice and inter-microservice secured communication work (each request is trusted by the authorization server and the roles restrictions are properly applied). You can for instance
1- Create a gateway
3- Create a service  `World` which exposes a `/api/world` service returning `"World"` only for admins
2 - Create a service `Hello`which exposes a service `/api/hello` service which invoke the `World` service via Feign and returns the string `Hello` + response of the `World` service for authenticated users
3- Run a curl request against the gateway
```
curl -v --request GET  --url http://localhost:8080/hello/api/hello \
  --header "Authorization: Bearer $(curl --silent --request POST --url http://localhost:9080/auth/realms/jhipster/protocol/openid-connect/token --header 'content-type: application/x-www-form-urlencoded' --data 'grant_type=password&client_id=web_app&&username=admin&password=admin' | jq .access_token | tr -d '"')"
```
Try with the user and admin account.
You'll verify that the gateway acts as a Resource Server and also as JWT forwarder to underneath microservices. You'll also see that each request is secured and apply the roles defined).

jdubois:
Great work @danielpetisme !!
- On a personal note: I just checked and saw you were a student, and that you are French - that's incredible! At the end of this week there's a talk at your company about "keeping developers happy", don't hesitate to reach the speaker from Société Générale, I work with him one day per week (and they are big JHipster + OpenID Connect users).
- Concerning `OAUTH2_ISSUER` -> yes let's copy it 4 times, we already do this in other parts of the code. That's easier to understand, even if it's a bit cumbersome. I'll revert my changes in `jhipster/jhipster` (side note: I also removed the older OAuth configuration there, I will still keep it removed, as it's not used anymore)
- Concerning the UI part: yes @danielpetisme this would be a lot better to do this in the UI, but that's more work. I'd rather have a first release soon, and then we can improve it. The current code is also easier to maintain for us, and also that was for people who want to encrypt their JWT token (then the UI can't decrypt it, you need to go through the server to get the roles - I have a client doing this).

danielpetisme:
@jdubois 🤣 I think there is a mistake in a bio somewhere on the Internet. I'm not student anymore (I just turned 30 this year. BTW, I'm Cyril & Romain's contact. I asked them during Devoxx to come to the Lava Java User Group and  and take the opportunity to visit my company. You're also welcome whenever you want 😉 (I sent you a mail months ago so I propose to switch to the mail to continue the discussion).

I'm building a version with code in the backend, stay tuned.

mraible:
@jdubois I was hoping to do a blog post and screencast of the OIDC support today, and do a lightning talk on it tonight at Denver's JUG. 

What's the status of a 4.10.0 release with the refactored OIDC support? Do you want to wait on this PR's microservices support before releasing?

I don't want to do a screencast too soon, and then have things change. If you want me to get rid of `OAUTH2_ISSUER`, I can do that and update the documentation.

jdubois:
@danielpetisme oh I understand everything now :-)
@mraible yes we should do a 4.10.0 release soon (before merging Couchbase support, or the release would be too big), and I also badly need this for one of my clients. Unfortunately I really have too little time to test -> once this PR is merged, are you confident everything is fine? Anyway OIDC would still be marked "BETA". In that case I could do a release tomorrow evening/night.

jdubois:
I just pushed `jhipster/jhipster` v1.1.13 with no more "OAuth2" configuration, as we discussed this would be copy/pasted everywhere

-> @danielpetisme @mraible I should have some time tomorrow evening to review (and correct some parts, so don't hesitate to give me some work), and do a release. And if there's too much work, I'll do it over the week-end.

mraible:
@jdubois I'll have some time to work on this tomorrow as well. I can hold my blog post and screencast until the release.

danielpetisme:
I juste pushed the 2 files required to have fullfill the `/api/account` endpoint. Now it you generate an application the UX is ok.

@jdubois The code has to be cleaned. There is an annonying between skipUserManagement and the historic oauth2/monolith implementation (Eg. `<%_ if (!skipUserManagement || authenticationType === 'oauth2') { _%>`). Right now during the bootstrap of the app an exception is generated because (Liquibase|Mongobee|other) try to create the basic users (user, admin, etc.)  but the `users.csv` is skipped.  If the legacy oauth2 support is truly history It would worth it to review users files and only use the `skipUserManagementFlag`).

@mraible I read your blog post about `angular-oauth2-oidc` and tried to integrate it and get stuck on the hash fragments... just as you in the original [PR](https://github.com/gmarziou/jhipster-keycloak/issues/2). What do you think of keeping the authentication in the UI and the backend app just verifies the token but can't produce one?

danielpetisme:
I juste pushed the 2 files required to have fullfill the `/api/account` endpoint. Now it you generate an application the UX is ok.

@jdubois The code has to be cleaned. There is an annonying between skipUserManagement and the historic oauth2/monolith implementation (Eg. `<%_ if (!skipUserManagement || authenticationType === 'oauth2') { _%>`). Right now during the bootstrap of the app an exception is generated because (Liquibase|Mongobee|other) try to create the basic users (user, admin, etc.)  but the `users.csv` is skipped.  If the legacy oauth2 support is truly history It would worth it to review users files and only use the `skipUserManagementFlag`).

@mraible I read your blog post about `angular-oauth2-oidc` and tried to integrate it and get stuck on the hash fragments... just as you in the original [PR](https://github.com/gmarziou/jhipster-keycloak/issues/2). What do you think of keeping the authentication in the UI and the backend app just verifies the token but can't produce one?

mraible:
@danielpetisme The `<%_ if (!skipUserManagement || authenticationType === 'oauth2') { _%>` is something I added. Basically, I don't want to skip user management entirely with the OAuth impl (for monolith) because I added in user syncing. The code I wrote doesn't allow updating the user from the UI (or in resource/server classes), but it does [sync the the user from the IdP with the user in the database](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java#L211). The reason I implement it this way is so you can create relationships with the `User` entity. Maybe we need a new flag called `syncUser` or something like that? Without this feature, you won't be able to create relationships with users, which I think is an important feature.

As far as using [angular-oauth2-oidc](https://www.npmjs.com/package/angular-oauth2-oidc), I like the current implementation because it's all server-side and you can set environment variables to override properties and change IdPs. I documented how to do this at the bottom of http://www.jhipster.tech/security/.

If you move it to the client, you'll have to edit TypeScript and recompile to change IdPs. 

We [discussed this a fair bit](https://github.com/gmarziou/jhipster-keycloak/issues/2#issuecomment-327832758) when I was implementing it and came to the conclusion that server-side was better.

danielpetisme:
@mraible Indeed is confusing to have a `skipUserManagement` flag and still manage some user files.  Since your code provision the user into Jhipster db for sync purpose can we skip the initial user creation?

danielpetisme:
@jdubois How did you build and deploy the `jhipster` lib? I can't find the new properties. I retrieved the jar and disassemble it. I can't see an `oauth2` property (only `jwt`)

mraible:
> @mraible Indeed is confusing to have a `skipUserManagement` flag and still manage some user files. Since your code provision the user into Jhipster db for sync purpose can we skip the initial user creation?

I could skip the loading of `users.csv`, but we'd still need to load `authorities.csv` and `users_authorities.csv` for the default admin/admin and user/user to work. Does not loading `users.csv` buy us anything?

mraible:
@danielpetisme I created a PR against master to remove `oauth2.issuer` https://github.com/jhipster/generator-jhipster/pull/6510

jdubois:
@danielpetisme I removed the old "oauth" code, and as we said we wouldn't use this property (as it's copy/pasted), there is no specific JHipster configuration here - that's why you don't see anything. If you want to use the lib, just clone its repo, and you can do a `mvn install` -> that will install your version in your local `.m2` repo, so you will use your version instead of the official one.

jdubois:
@mraible oh I didn't see you were syncing users from OIDC:

- that's a great feature, as I often have questions on using users from a microservice
- but at the moment we don't do this with JWT authentication -> I don't want this feature to block us from merging this PR, we can still do it later - I'm not totally sure, is this causing an issue here?
- and I'm also not sure of the solution: this means each microservice will get a "user" table, which is being copy/pasted from OIDC, and which is not always in sync. I don't have a better idea, but that looks weird to me.

jdubois:
- I tried to fix the merge conflict
- I'm testing this -> this is becoming so big, I'd love to merge this tonight, but it depends on what I find

jdubois:
Oops.... So sorry everyone, I merged this by mistake!!! I'm doing a JHipster training and I was trying to show a few tricks, and of course I messed up badly :-(
This is totally my fault, and I'm not available to clean up (as I'm doing the training...) - hopefully this isn't too bad, and anyway I wanted to merge this during the week-end.

danielpetisme:
I fixed the bug you mentioned @jdubois + I reuse @mraible user sync as-is for microservice architecture (only for gateways).
https://github.com/danielpetisme/generator-jhipster/tree/oidc-support-microservices

I think it's good enough for a first release. 

Regarding the anticipated merge. Do you want me to create a new PR? I can create one to fix the bug and another for the user sync if you want.

jdubois:
Yes please ... really sorry

danielpetisme:
1 or 2 pr ?

jdubois:
Oh sorry - the easiest for you, so 1 I guess

adnansenyurt:
I'm using generator v4.9.0 but I don't see keycloak.yml created under docker directory.
Also, is this ready for cloud deployment, or dev profile only? Shall I go back to 4.8.2 for now?

jdubois:
@adnansenyurt this isn't released yet, this why this ticket is not yet linked to a release version. You should use the `master` branch, please read the contributing doc.

PR #6411, opened 2017-09-26
Refactor AccountResource, UserResource and EntityResource to use Problem

cbornet:
- Please make sure the below checklist is followed for Pull Requests.

- [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

Throw `Problem` exceptions that will be treated by the `ExceptionTranslator` instead of returning `ResponseEntity` with an unchecked body
See #6404

PR #6437, opened 2017-09-29
Method with 6-7 String parameters is not the best pattern to follow

gzsombor:
It's better to pass the values as an object, so it's simpler to understand

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -39,6 +39,7 @@ import <%=packageName%>.security.SecurityUtils;
 import <%=packageName%>.service.util.RandomUtil;
 <%_ } _%>
 import <%=packageName%>.service.dto.UserDTO;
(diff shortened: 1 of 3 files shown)


cbornet:
Is it OK to have a dependency from service layer to REST layer ?
I think a method `User registerUser(UserDTO userDTO, String password)` would be cleaner.

jdubois:
Good catch @cbornet -> there should not be a service -> REST dependency. I cleaned them up some time ago, that was really bad.

cbornet:
OK. I'll do the change after #6411

gzsombor:
Please, do not re-use the UserDTO, again. We can have a new class, for registration. Classes are cheap ;-)
 We can have a RegistrationDTO or UserRegistrationDTO, and an UserUpdateDTO, and an UserSelfViewDTO and PublicUserDTO. 
 Currently it's very confusing, who can see what, and which field is mandatory in different places.

cbornet:
That would mean moving ManagedUserVM to the service layer (and become ManagedUserDTO). No problem for me.

PR #6444, opened 2017-10-01
Fix Align Left Account Menu

carpanese:
In the Bootstrap update, the login screen, account and password for example are getting off the screen. A scroll bar is generated below.

This improvement leaves everything correct. I ran the tests on the new generator and it worked perfectly.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
@@ -57,6 +57,9 @@ Navbar
             color: #fff;
         }
     }
+	.dropdown-menu-right {
+        left: auto !important;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/src/main/webapp/app/layouts/navbar/_navbar.scss
```diff
@@ -57,6 +57,9 @@ Navbar
             color: #fff;
         }
     }
+	.dropdown-menu-right {
```
sendilkumarn: I don't think this is necessary. Adding `.ml-auto` to the element will fix it.

deepu105:
@carpanese can you make the requested change?

deepu105:
@sendilkumarn if there is no response could you update this?

sendilkumarn:
closing this due to inactivity

PR #6446, opened 2017-10-01
Use jhiPrefix to prefix JHipster tables

### LINKED ISSUE #6004: Use jhiPrefix to prefix JHipster tables

As stated in the [documentation](https://jhipster.github.io/creating-an-app/#3) : 

> --jhi-prefix - Add prefix before services, components and state/route names (Default: jhi)

I think the `jhiPrefix` could also be used to change the name of the JHipster tables : 

* `jhi_user`
* `jhi_authority`
* `jhi_user_authority`
* `jhi_persistent_token`

These names are hard-coded in the [Liquibase initial schema](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/liquibase/changelog/_initial_schema.xml#L60) or in the entity [@Table annotation](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/domain/_User.java#L55). And this way, we would have : 

* `custom_user`
* `custom_authority`
* `custom_user_authority`
* `custom_persistent_token`

WDYT ?

When having several microservices applications, you end up with a few different databases, and all have a `jhi_user` table. It would be better to have `reg_user`, `order_user`.... depending on the prefix.

##### **JHipster Version(s)**

4.5.6

duderoot:
Fix #6004

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -219,13 +219,14 @@ module.exports = class extends BaseGenerator {
             validateTableName() {
                 const prodDatabaseType = this.prodDatabaseType;
                 const entityTableName = this.entityTableName;
+                const jhiPrefix = this.jhiPrefix;
                 if (!(/^([a-zA-Z0-9_]*)$/.test(entityTableName))) {
                     this.error(chalk.red('The table name cannot contain special characters'));
                 } else if (entityTableName === '') {
                     this.error(chalk.red('The table name cannot be empty'));
                 } else if (jhiCore.isReservedTableName(entityTableName, prodDatabaseType)) {
-                    this.warning(chalk.red(`The table name cannot contain the '${entityTableName.toUpperCase()}' reserved keyword, so it will be prefixed with 'jhi_'`));
-                    this.entityTableName = `jhi_${entityTableName}`;
+                    this.warning(chalk.red(`The table name cannot contain the '${entityTableName.toUpperCase()}' reserved keyword, so it will be prefixed with '${jhiPrefix}_'`));
+                    this.entityTableName = `${jhiPrefix}_${entityTableName}`;
                 } else if (prodDatabaseType === 'oracle' && entityTableName.length > 26) {
                     this.error(chalk.red('The table name is too long for Oracle, try a shorter name'));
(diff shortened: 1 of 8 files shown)


generators/server/index.js
```diff
@@ -148,6 +148,7 @@ module.exports = class extends BaseGenerator {
                 if (this.searchEngine === undefined) {
                     this.searchEngine = false;
                 }
+                this.jhiPrefix = this.configOptions.jhiPrefix || this.config.get('jhiPrefix') || this.options['jhi-prefix'];
```
deepu105: there is no jhi-prefix option on this sub gen hence this line should be
`this.jhiPrefix = this.configOptions.jhiPrefix || this.config.get('jhiPrefix');`

generators/entity/index.js
```diff
@@ -219,13 +219,14 @@ module.exports = class extends BaseGenerator {
             validateTableName() {
                 const prodDatabaseType = this.prodDatabaseType;
                 const entityTableName = this.entityTableName;
+                const jhiPrefix = this.jhiPrefix;
```
deepu105: make sure the prefix is in table name format hence introduce a new param 
```
this.jhiTablePrefix = this.getTableName(this.jhiPrefix);`
```
per sub generator and then use that instead of ` this.jhiPrefix`

generators/server/index.js
duderoot: 10x for the hints

generators/entity/index.js
deepu105: You didn't do this yet?

duderoot: I make the requested changes and commit them. Unfrotunately I overide the first commit by ammending the last commit and now the history of the PR is not clean anymore. Sorry for that. If there is something else for this PR to do let me know. Tank you!

duderoot:
@deepu105 sorry to bother, is there something that I can do for this PR? I am asking because, this PR was a little mess up because of overwriting the history 😄

jdubois:
LGTM -> @duderoot let's just wait for @deepu105 to do a last review, and that will be merged

deepu105:
My original comment was to introduce an instance level variable derived using table name. I'll  make the change and merge this manually later

Issue #6440, opened 2017-09-30
[Feature] Create new, small sub-generators for Spring Boot


jdubois:
Our "jhipster entity" sub-generator generates a complete CRUD entity: that's good for a lot of reasons, but sometimes you just want to create a simple REST service. Not everything is a CRUD :-)

If you want smaller components:
- We already have the "jhipster service" sub-generator, which generates a Spring service bean
- On the client-side, you can use Angular CLI to generate components, routers, etc...

I would like to:
- Add a new "jhipster spring-controller" sub-generator, which generates a simple Spring MVC REST controller
- Rename the "jhipster service" to "jhipster spring-service"
- Think about other use-cases: a "jhipster jpa-repository"?

This would be an easy work, which shouldn't have much impact elsewhere, but which could help a lot of people.

atomfrede:
Could be a good thing for hacktoberfest

deepu105:
@jdubois how about containing everything in a parent sub gen (Of course code wise each will be separate and we will just compose it from the parent) so that users don't have to remember every sub generator and it will be easier to advertise and drive adoption. 
I propose `jhipster component` which will ask the below question
```
What would you like to do?
1. Generate an entity
2. Generate a Spring service
3. Generate a Spring controller
```
Based on the answer we will just compose the appropriate sub gen and it will take it from there asking additional questions if any and generating the stuff

**Advantages of this**

- Easier for people to remember
- Easier adoption otherwise the option goes unnoticed like our Service generator which is rarely used
- Possible to add more complex options like `Generate a Spring controller with service` for which we will just compose the controller sub gen followed by the service sub gen
-

jdubois:
@deepu105 Yes but that's just another sub-generator which is a "wrapper" of the others - no big deal here!
For my personal use-case, I'd rather have "jhipster spring-controller", I find it easier and quicker to use (and that would still be possible, so no problem).

ramzimaalej:
I like the idea, this is similar to what Grails has and offers a good level of granularity of what will be generated. It happened to me to delete all the UI part for a given entity because it was not needed. 
I would prefer a different command than answering questions like @jdubois mentioned.

jdubois:
Yes: at the beginning Grails (and RoR!!) where heavy sources of inspiration, we are going back to roots :-)

clarenced:
Hi @jdubois I am interested in this feature !!!

clarenced:
#6502

[...]

PR #6451, opened 2017-10-02
add spring-controller sub-generator

xetys:
for #6440 

based on service and renamed service subgen to spring-service

I would like to add both, travis tests and integration tests for the controller...

additionally, it might be interesting for the interactive mode to add controller actions like fields in entity generator, where the user can choose the HTTP method and the action's name. WDYT? 


cli/commands.js
```diff
@@ -1,62 +1,66 @@
 module.exports = {
-    app: {
+    'app': {
```
MathieuAA: Why did you add these? They actually make the code a bit harder to read and are unnecessary (except for `spring-controller` where it's needed).

xetys:
No, there are several other, as 'import-jdl', 'docker-compose' etc. I found
it inconsistent to read so I unified it... I can take it back if it's needed

Am Mo., 2. Okt. 2017, 08:35 schrieb Mathieu ABOU-AICHI <
notifications@github.com>:

MathieuAA:
@xetys that's the trade-off =/ it's generally the linter that warns you (more of an error-warning :)) about it.
It seems, in our case, that the tests fail because of that.

cli/commands.js
```diff
@@ -1,62 +1,66 @@
...
     },
     'rancher-compose': {
         desc: 'Deploy the current application to Rancher'
     },
-    server: {
+    'server': {
         desc: 'Create a new JHipster server-side application'
     },
-    service: {
+    'spring-service': {
```
deepu105: add `alias : 'service'` here so that old command works as well, else this will be breaking change

xetys:
@deepu105 seriously...ok... I will change that

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * <%= controllerClass %> controller
+ */
+@RestController
+@RequestMapping("/api/<%= apiPrefix %>")
+public class <%= controllerClass %>Controller {
+
+    private final Logger log = LoggerFactory.getLogger(<%= controllerClass %>Controller.class);
+
```
jdubois: Maybe we should also add a simple "GET" method? Like a "helloworld()" method that returns a String?

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
```diff
@@ -0,0 +1,35 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * <%= controllerClass %> controller
+ */
+@RestController
+@RequestMapping("/api/<%= apiPrefix %>")
```
jdubois: Maybe we should also generate a unique URL? Here you would get error when you generate 2 controllers in a row. How about "/api/<%= apiPrefix %>-<%=controllerClass%>"

jdubois: Maybe use Lodash to use a dasherized version of "controllerClass"

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
xetys: How about a similar process like the field adding in entity sub-generator? And if no method added, then add a hello-world method

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
xetys: apiPrefix = kebabCase(controllerClass)

xetys:
my todos left:

- [ ]: yeoman tests
- [ ]: Spring integration tests
- [ ]: (in discusson) method generator
- [ ]: default action (if no other action exists)

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
deepu105: That would be much better otherwise the class we generate wouldnt have much value over the one you could create in an IDE

jdubois: yes +1 - but keep the question simple

generators/spring-controller/templates/src/main/java/package/web/rest/_Controller.java
jdubois: the issue with KebabCase is that you would keep upper case letters -> maybe do a lower case after?

xetys: no, it doesn't, I tried already...

```
/**
 * AnotherTest controller
 */
@RestController
@RequestMapping("/api/another-test")
public class AnotherTestController {
    //...
}
```

jdubois: I'm quite surprised, good news then, let's do it!

xetys:
so far, here it is

generators/spring-controller/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+        this.controllerActions.forEach((action) => {
+            action.actionPath = _.kebabCase(action.actionName);
+            action.actionNameUF = _.upperFirst(action.actionName);
+            this.log(chalk.green(`adding ${action.actionMethod} action '${action.actionName}' for /api/${this.apiPrefix}/${action.actionPath}`));
+        });
+
+        this.template(
+            `${SERVER_MAIN_SRC_DIR}package/web/rest/_Controller.java`,
+            `${SERVER_MAIN_SRC_DIR + this.packageFolder}/web/rest/${this.controllerClass}Controller.java`
```
deepu105: this could be `${SERVER_MAIN_SRC_DIR}${this.packageFolder}/web/rest/${this.controllerClass}Controller.java`
same in other usage below

generators/spring-controller/prompts.js
```diff
@@ -0,0 +1,102 @@
...
+    askForControllerActions
+};
+
+function askForControllerActions() {
+    const done = this.async();
+
+    askForControllerAction.call(this, done);
+}
+
+function askForControllerAction(done) {
```
deepu105: You can move this method inside the `askForControllerActions` and convert this into a const arrow like below so that you dont have to bind `this`
```
function askForControllerActions() {
  const askForControllerAction = done => {
    ...
    this.prompt(prompts).then((props) => {
      if (props.actionAdd) {
        ...
        askForControllerAction(done);
      }
    });
  }
  const done = this.async();
  askForControllerAction(done);
}
```

generators/spring-controller/prompts.js
```diff
@@ -0,0 +1,102 @@
...
+                },
+                {
+                    name: 'PUT',
+                    value: 'Put'
+                },
+                {
+                    name: 'DELETE',
+                    value: 'Delete'
+                }
+
```
jdubois: Please remove the empty spaces

generators/spring-controller/templates/src/test/java/package/web/rest/_ControllerIntTest.java
```diff
@@ -0,0 +1,68 @@
...
+import <%= mockRequestImports[idx] %>;
+<%_ } _%>
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+/**
+ * Test class for the <%= controllerClass %> REST controller.
+ *
+ * @see <%= controllerClass %>Controller
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = <%= mainClass %>.class)
```
jdubois: Oh you even generate tests on your controller!!! Big +1!

generators/spring-controller/prompts.js
```diff
@@ -0,0 +1,99 @@
...
+                type: 'input',
+                name: 'actionName',
+                validate: (input) => {
+                    if (!(/^([a-zA-Z0-9_]*)$/.test(input))) {
+                        return 'Your action name cannot contain special characters';
+                    } else if (input === '') {
+                        return 'Your action name cannot be empty';
+                    } else if (input.charAt(0) === input.charAt(0).toUpperCase()) {
+                        return 'Your action name cannot start with an upper case letter';
+                    } else if (jhiCore.isReservedFieldName(input)) {
```
deepu105: @xetys does this matter? I guess reserved words are ok in a method name??

deepu105: Ok never mind seems some keywords cant be used anyway

generators/spring-controller/prompts.js
```diff
@@ -0,0 +1,99 @@
...
+        this.prompt(prompts).then((props) => {
+            if (props.actionAdd) {
+                const controllerAction = {
+                    actionName: props.actionName,
+                    actionMethod: props.actionMethod
+                };
+
+                this.controllerActions.push(controllerAction);
+
+                askForControllerAction.call(this, done);
```
deepu105: you cant do this for arrow methods just call `askForControllerAction(done)`

jdubois:
I'm rushing to do the `4.10.0` release, so sorry I'm going to modify this without much discussion:

- This generates a "FooController" class, but the entity sub-generator would generate a "FooResource" class. This is because that's supposed to be a REST "resource". Anyway, for consistency, I'm going to rename this with the "Resource" prefix.
- I'm also going to add the documentation to the website.

PR #6458, opened 2017-10-02
adding git init

### LINKED ISSUE #6453: [Feature] Create a "git init" sub-generator
This is the same idea as #6440

I would like to have a small sub-generator for Git:

- `jhipster git-init` would initialize a Git repository, add all files to it, and commit with message "Initial application generation by JHipster" (like https://start.jhipster.tech does today)
- This would run by default after each generation, but could be disabled with a flag

I'm also wondering if we could have the same after an entity is generated, but I'm less sure it's a good idea.

sendilkumarn:
closes #6453


generators/app/index.js
```diff
@@ -344,6 +351,16 @@ module.exports = class extends BaseGenerator {
                         }
                     }
                 }
+
+                if (!this.options['skip-git']) {
+                    this.gitExec('init', () => {
```
cbornet: Shouldn't `checkGit` be called to see if the user has git installed ?

generators/app/index.js
```diff
@@ -344,6 +351,26 @@ module.exports = class extends BaseGenerator {
...
                 }
+
+                if (!this.options['skip-git']) {
+                    this.isGitInstalled((code) => {
+                        if (code === 0) {
+                            this.gitExec('rev-parse --is-inside-work-tree', (err, gitDir) => {
+                                if (!gitDir) {
+                                    this.gitExec('init', () => {
+                                        this.log('git init successful');
+                                        this.gitExec(['add -A'], () => {
```
deepu105: no need of array for this syntax, remove array brackets here and in next command

sendilkumarn:
Is there some problem with `travis` this two tests are failing :(

jdubois:
Travis is OK, should we merge? LGTM

PR #6472, opened 2017-10-04
Set request path field to all Problem errors

cbornet:


DanielFran:
Can it be created directly in the project a specific page for each problem instead of redirecting to jhipster specific page?

cbornet:
Yes, that would be nice. Do you want to do it ?

DanielFran:
I can try, but I will need some tips to integrate in generator...

I found an example from zalando that we could use as template: http://zalando.github.io/problem/constraint-violation/

We need few pages, or only 1 that could have as content:
- Title => comming from Title message already translated
- Description => a specific text for each error that detail the error (like in zalando example) that will need to be translated
- The box with the error content (like in zalando exemple) comming from the problem

PR #6494, opened 2017-10-08
Add External contents to webpack

sudheerj:

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1099,7 +1099,7 @@ module.exports = class extends PrivateBase {
         const to = `content/${targetFolder}/`;
         const webpackDevPath = `${CLIENT_WEBPACK_DIR}/webpack.common.js`;
         let assetBlock = '';
-        if(from || to) {
+        if (sourceFolder && targetFolder) {
             assetBlock = `{ from: './${from}', to: '${to}' },`;
         }
 
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1089,6 +1089,36 @@ module.exports = class extends PrivateBase {
     }
 
     /**
+     * Copy third-party library resources path.
+     *
+     * @param {string} sourceFolder - third-party library resources source path
+     * @param {string} targetFolder - third-party library resources destination path
+     */
+    copyExternalAssetsInWebpack(sourceFolder, targetFolder) {
+        const from = `${CLIENT_MAIN_SRC_DIR}content/${sourceFolder}/`;
+        const to = `content/${targetFolder}/`;
+        const webpackDevPath = `${CLIENT_WEBPACK_DIR}/webpack.common.js`;
+        let assetBlock = '';
+        if(from || to) {
+            assetBlock = `{ from: './${from}', to: '${to}' },`;
+        }
+
+        try {
+            jhipsterUtils.rewriteFile({
+                file: webpackDevPath,
+                needle: 'jhipster-needle-add-assets-to-webpack',
+                splicable: [
+                    assetBlock
+                ]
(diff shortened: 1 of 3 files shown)


generators/generator-base.js
```diff
@@ -1089,6 +1089,36 @@ module.exports = class extends PrivateBase {
...
+     *
+     * @param {string} sourceFolder - third-party library resources source path
+     * @param {string} targetFolder - third-party library resources destination path
+     */
+    copyExternalAssetsInWebpack(sourceFolder, targetFolder) {
+        const from = `${CLIENT_MAIN_SRC_DIR}content/${sourceFolder}/`;
+        const to = `content/${targetFolder}/`;
+        const webpackDevPath = `${CLIENT_WEBPACK_DIR}/webpack.common.js`;
+        let assetBlock = '';
+        if(from || to) {
```
deepu105: shouldnt this be `from && to`

deepu105:
You have lint fail, make sure to run `npm run lint` on your fork
```
> generator-jhipster@4.9.0 lint /home/travis/build/jhipster/generator-jhipster
> eslint .
/home/travis/build/jhipster/generator-jhipster/generators/generator-base.js
  1102:9  error  Expected space(s) after "if"  keyword-spacing
✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

deepu105:
dont create new PR just push your change commits into the same branch

sudheerj:
Thanks Deepu. Done with the changes and no lint issues as well. By the way when we will have next Jhipster minor release. If it is not immediate, Do we have any nightly builds available. This is required for me to publish the module.

deepu105:
I cannot give you an ETA on this and unfortunately, we don't do nightly builds. I suggest you use the master branch in the meantime.
Clone the repo, checkout master
run npm link on the cloned `generator-jhipster `folder, now this version will be used instead of global
cc @jdubois

sudheerj:
Thanks. Currently I'am using local version of JHipster(with above changes) to work with PrimeNG module. But it cannot be published until JHipster released with above changes. If anybody tries to use the module then they are going to face below issues. i.e, Even users has to maintain the local version of Jhipster.

https://github.com/sudheerj/generator-jhipster-primeng/issues/4

I will also raise two more PRs to inject third party resources(one for index.html and other one for Vendor.SCSS).

pascalgrimaud:
Look at https://github.com/geraldhumphries/generator-jhipster-elasticsearch-reindexer/blob/master/generators/app/index.js#L204-L210
you can use a similar code, so your module will work with projects which was generated with a version older than 4.9.x

sudheerj:
Thanks Pascal. Is it the code snippet similar to above solves the issue of publishing module or the code snippets to work with multiple versions of JHipster.

pascalgrimaud:
Let me explain a little bit:
By adding this check on this function, your code will test if the function exists in the generator-jhipster dependency of the project (generated by the user):
- if **yes** (case 1),
    - the project has probably been generated with generator-jhipster 4.9.1+ (which is not yet released)
    - your module can use this function
    - it will add some code into `webpack.common.js`
- if **no** (case 2),
    - there will be no error when the user uses your module, because you catch the error
    - some instructions will be displayed to the user, so he can add manually the code needed into `webpack.common.js`

By using this code snippet, you can publish your module now. All users who use your module, will be on the case 2. As soon as there is a new release of generator-jhipster, it should be in the case 1.

Hope I'm clear.

sudheerj:
Thanks a lot Pascal for the detailed explanation. I will add this check to avoid this kind of errors.

PR #6507, opened 2017-10-11
Fix the --skip-server option

### LINKED ISSUE #6389: skip-server option and missing code in security module
Overview of the issue

When I am trying to use `--skip-server` option many issues happen in Security model when trying to build the system.

~~~
ERROR in [at-loader] ./src/main/webapp/app/blocks/interceptor/http.provider.ts:23:17
TS2304: Cannot find name 'AuthExpiredInterceptor'.

ERROR in [at-loader] ./src/main/webapp/app/shared/login/login.service.ts:19:18
TS2339: Property 'authServerProvider' does not exist on type 'LoginService'.

ERROR in [at-loader] ./src/main/webapp/app/shared/login/login.service.ts:38:21
TS2339: Property 'authServerProvider' does not exist on type 'LoginService'.

ERROR in [at-loader] ./src/main/webapp/app/shared/login/login.service.ts:42:14
TS2339: Property 'authServerProvider' does not exist on type 'LoginService'.**
~~~

After some debugging, I found that we are using some Services thas should not be in the code if we are 

using --skip-server option like below in 'login-service.ts':

`if (this.principal.isAuthenticated()) { this.authServerProvider.logout().subscribe(); }`

because we have condition that will remove authServerProvider from the class :

`<%_ if (!skipServer) { _%> private authServerProvider: AuthServerProvider <%_ } _%>`

P.S : not only this file there are other files which have the same issue.

##### **Reproduce the error**
`jhipster --skip-server --auth=jwt --db=sql `

##### **Related issues**
https://github.com/jhipster/generator-jhipster/issues/5429

##### **Suggest a Fix**
none

Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
koptan@0.0.0 C:\JS\jh-frontend
+-- generator-jhipster@4.8.2
`-- UNMET PEER DEPENDENCY popper.js@^1.11.0

```

<details>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.8.2",
    "baseName": "koptan",
    "useSass": false,
    "authenticationType": "jwt",
    "databaseType": "sql",
    "devDatabaseType": "sql",
    "prodDatabaseType": "sql",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "skipServer": true,
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "ar-ly"
    ]
  }
}
</pre>
</details>

<details>
<pre>
</pre>
</details>

java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

git version 2.14.1.windows.1

node: v6.10.2

npm: 3.10.10

bower: 1.8.0

yeoman: 2.0.0

Windows 10 , Chrome

clarenced:
Fix #6389
remove the `skipServer` test in http.provider.ts and login.service.ts
Also set to true the default condition for `generator.skipServer`

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -139,7 +139,8 @@ const files = {
                 // interceptors
                 'blocks/interceptor/_errorhandler.interceptor.ts',
                 'blocks/interceptor/_notification.interceptor.ts',
-                'blocks/interceptor/_http.provider.ts'
+                'blocks/interceptor/_http.provider.ts',
+                'blocks/interceptor/_auth-expired.interceptor.ts'
             ]
         },
         {
@@ -147,13 +148,6 @@ const files = {
             path: ANGULAR_DIR,
             templates: [
                 'blocks/interceptor/_auth.interceptor.ts'
-            ]
-        },
-        {
-            condition: generator => generator.skipServer,
-            path: ANGULAR_DIR,
-            templates: [
-                'blocks/interceptor/_auth-expired.interceptor.ts'
             ]
         }
     ],
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -150,7 +150,7 @@ const files = {
             ]
         },
         {
-            condition: generator => !generator.skipServer,
+            condition: generator => generator.skipServer,
             path: ANGULAR_DIR,
             templates: [
                 'blocks/interceptor/_auth-expired.interceptor.ts'
--- a/generators/client/templates/angular/src/main/webapp/app/_app.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_app.module.ts
@@ -29,7 +29,6 @@ import { <%=angularXAppName%>AdminModule } from './admin/admin.module';
 import { <%=angularXAppName%>AccountModule } from './account/account.module';
 <%_ } _%>
 import { <%=angularXAppName%>EntityModule } from './entities/entity.module';
-
 import { customHttpProvider } from './blocks/interceptor/http.provider';
 import { PaginationConfig } from './blocks/config/uib-pagination.config';
 
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_http.provider.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_http.provider.ts
@@ -33,9 +33,7 @@ import { LoginModalService } from '../../shared/login/login-modal.service';
     <%_ } _%>
 import { StateStorageService } from '../../shared/auth/state-storage.service';
 <%_ } _%>
-<%_ if (!skipServer) { _%>
 import { AuthExpiredInterceptor } from './auth-expired.interceptor';
-<%_ } _%>
 import { ErrorHandlerInterceptor } from './errorhandler.interceptor';
 import { NotificationInterceptor } from './notification.interceptor';
 
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts
@@ -23,9 +23,7 @@ import {
     <%=angularXAppName%>SharedLibsModule,
     <%=angularXAppName%>SharedCommonModule,
     CSRFService,
-    <%_ if (!skipServer) { _%>
     AuthServerProvider,
-    <%_ } _%>
     AccountService,
     <%_ if (!skipUserManagement || authenticationType === 'oauth2') { _%>
     UserService,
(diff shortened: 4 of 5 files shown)


generators/client/files-angular.js
```diff
@@ -150,7 +150,7 @@ const files = {
             ]
         },
         {
-            condition: generator => !generator.skipServer,
+            condition: generator => generator.skipServer,
```
ruddell: This will only generate the file when skipServer is true.  I believe you always want to copy the file, so it should be moved [up to here with the other interceptors](https://github.com/clarenced/generator-jhipster/blob/92c600803e67f45ed64c12d33379d55546da9abf/generators/client/files-angular.js#L142)

clarenced: OK

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_http.provider.ts
```diff
@@ -33,9 +33,7 @@ import { LoginModalService } from '../../shared/login/login-modal.service';
     <%_ } _%>
 import { StateStorageService } from '../../shared/auth/state-storage.service';
 <%_ } _%>
-<%_ if (!skipServer) { _%>
 import { AuthExpiredInterceptor } from './auth-expired.interceptor';
```
deepu105: Not very sure of this, Probabaly its better to add an abstract `auth-expired.interceptor` and `AuthServerProvider` or to skip the code calling these form the login service when skipServer is enabled. I would prefer the second option

clarenced: Hi @deepu105 , thx for your reply. I may be missing something here, but do we still have authentication when `skipServer` is enabled. ? Because the form login is always generated when the `authenticationType ` is different `oauth2` and the `authServerProvider` is always used to perform the authentication.

If it is mandatory, I would go with the 1st option you proposed.

generators/client/files-angular.js
clarenced: @ruddell done

rachidoulasri:
I think you are missing the provider for "`AuthServerProvider`" in _app.module.ts

> generator-jhipster/generators/client/templates/angular/src/main/webapp/app/_app.module.ts


maybe you should add the following lines
```
<%_ if (authenticationType !== 'jwt') { _%>
import { AuthServerProvider } from './shared/auth/auth-jwt.service'
<%_ } _%>
```

```
providers: [
        ProfileService,
        customHttpProvider(),
        PaginationConfig,
        UserRouteAccessService,
       <%_ if (authenticationType !== 'jwt') { _%>
       AuthServerProvider
       <%_ } _%>
    ],

```

clarenced:
@jdubois I will make the changes as requested

clarenced:
Hi @oulasri, I am missing something here, should the authenticationType `===` 'jwt' instead of `!==` ?

oulasri:
Hi @clarenced, Yes, you're right, sorry it's an error on my part !

generators/client/templates/angular/src/main/webapp/app/_app.module.ts
```diff
@@ -28,8 +28,10 @@ import { <%=angularXAppName%>AdminModule } from './admin/admin.module';
 <%_ if (authenticationType !== 'oauth2') { _%>
 import { <%=angularXAppName%>AccountModule } from './account/account.module';
 <%_ } _%>
+<%_ if (authenticationType !== 'jwt') { _%>
```
ruddell: AuthServerProvider is listed in [`shared/shared.module.ts`](https://github.com/clarenced/generator-jhipster/blob/e8324351d0fc6669338b4ff5d7c6c1323e9b35fa/generators/client/templates/angular/src/main/webapp/app/shared/_shared.module.ts#L26-L28) instead of `app.module.ts`.  This way you also don't have to check the authentication type either. So changes needed:

 - Remove changes to `app.module.ts`
 - Remove `<%_ if (!skipServer) { _%>` around `AuthServerProvider` in `shared/shared.module.ts`

clarenced: @ruddell OK I will do it

ruddell:
Just tested and it looks good to me.  This is a big improvement for people separating their client and server using the `skip-server`/`skip-client` flags

Previous behavior of apps using skip-server:
 - yarn fails to compile project on generation
 - can not authenticate even when specifying authentication type

New behavior apps using skip-server :
 - yarn compiles project correctly
 - can authenticate out-of-the-box

Alan759:
I still get the above errors with jHipster 4.10.2. when I generate the app using: 
jhipster --skip-server --with-entities --skip-user-management --auth=jwt --db=sql

Which release(and when) will this fix go into?

Issue #6061, opened 2017-07-08
[Enhancement] make a master JHipster pom instead of 'hardcoding' dependencies and plugin management


Tcharl:
##### **Overview of the issue**
The actual way for managing 'server' dependencies on pom.xml is to declare dependencies version and plugin behavior directly in the generated pom.xml

Instead of doing this, we could make that generated pom inherit from 'jhipster:jhipster-pom:${jhipster.version} instead of generating it.
##### **Motivation for or Use Case**

It will ease that pom maintenance (the _pom.xml template will be far easy, containing less code), as well as ease the pom upgrade (dependency or plugin management): client will just have to update their parent pom's ${jhipster.version} instead of backporting changes.

So that it will permit to tweak the pom more easily (like generating site, using more plugins, etc.
I've already made a good OSS pom on my previous [project](https://github.com/OsgiliathEnterprise/net.osgiliath.parent/tree/master/net.osgiliath.poms), may be I can make a PR?
It should be nice to make the same thing for Gradle project, as it would ease the maintenance of both frameworks, but I don't know anything about this build tool.

##### **Suggest a Fix**

Make a new jhipster project (jhipster-pom), adding dependencies and dependencymanagement on this pom. Then we should release these pom on maven central, then modifying the generator-jhipster/server/_pom.xml template to inherit from the newly created pom

##### **JHipster Version(s)**
4.6.x

- [ x] Checking this box is mandatory (this is just to show you read everything)

deepu105:
I think @erikkemperman was working on something like that?

erikkemperman:
Yes, I was. But a lot of other stuff got in the way :-)

If someone else wants to try: the gradle dependency management plugin, from the spring guys, should make it possible for the gradle build to use the same versions as the maven build (it reads poms!)

Tcharl:
So I'm on it ;-). Could a JHipster guy create a repo for the parent pom? For example jhipster-poms or jhipster-parent?

jdubois:
We could put it in jhipster/jhipster as everyone already depends on it.
Then, I'm not sure people will be happy to be tied too much to JHipster.  I believe they like their generated pom, and modify it as they want.

gzsombor:
Instead of creating a parent pom, it would be better if maven's import mechanism is used. So JHipster could be used in a multi-module maven project too.

Tcharl:
[...]

PR #6509, opened 2017-10-12
Use jhipster-dependencies BOM

erikkemperman:
See https://github.com/jhipster/generator-jhipster/issues/6061

So with this PR, almost all (*) dependencies are managed by the new `jhipster-dependencies` BOM, and unified across Maven and Gradle. There's one extra commit that can be discarded once 0.0.4 of the BOM is released with upgraded Cassandra driver artifacts.

(*) one exception are the inner deps within plugin configuration.

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -561,8 +561,8 @@ module.exports = class extends Generator {
                 this.warning('Java 8 is not found on your computer.');
             } else {
                 const javaVersion = stderr.match(/(?:java|openjdk) version "(.*)"/)[1];
-                if (!javaVersion.match(/1\.8/)) {
-                    this.warning(`Java 8 is not found on your computer. Your Java version is: ${chalk.yellow(javaVersion)}`);
+                if (!javaVersion.match(new RegExp(constants.JAVA_VERSION.replace('.', '\\.')))) {
+                    this.warning(`Java ${constants.JAVA_VERSION} is not found on your computer. Your Java version is: ${chalk.yellow(javaVersion)}`);
                 }
             }
             done();
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -48,6 +48,10 @@ const DOCKER_JENKINS = 'jenkins:latest';
 const DOCKER_SWAGGER_EDITOR = 'swaggerapi/swagger-editor:latest';
 const DOCKER_COMPOSE_FORMAT_VERSION = '2';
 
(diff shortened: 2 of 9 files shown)


jdubois:
I would like to merge this after #6435 and doing JHipster v4.10.0 - otherwise the release would be too big

erikkemperman:
Fair enough! Perhaps in the mean time we could discuss the merits of linking the releases of jhipster server library and the dependencies BOM, possibly putting them in a single repository so the maven-release-plugin can work as intended.

I think this might be a good idea because now there is an odd co-dependency: jhipster-dependencies manages the version of jhipster (for consumption by the generator and the registry) but the server lib itself also uses the BOM so all the versions will be automagically aligned.

Thoughts?

Oh and also I would very much appreciate if folks could try if this PR doesn't accidentally break their use cases (even though the tests all pass).

Thanks!

atomfrede:
Great work. I will try to test it during the weekend somehow.

erikkemperman:
@atomfrede thanks, that would be much appreciated!

erikkemperman:
FYI Rebased to resolve merge conflicts, and bumped jhipster-dependencies to 0.0.5.

erikkemperman:
Rebased to resolve conflicts with current master.

jdubois:
OK let's do this, I'm reviewing this and merging ASAP

erikkemperman:
@jdubois Great! FYI though, it needs to be rebased and due to recent merges to master branch it needs a new jhipster-dependencies in order to have no regression. See https://github.com/jhipster/jhipster-dependencies/pull/11. If you could look at that and do merge/release if you approve it, then I would be happy to resolve the merge conflicts on this PR!

jdubois:
@erikkemperman this is going to be OK - however I must admit I'm a bit disappointed, we only gain 20 lines in the `pom.xml`, which isn't a lot, so this looks like a lot of work without a big win in the end.

erikkemperman:
@jdubois Thanks for merging, but I was still resolving conflicts and bumping jhipster-dependencies to 0.0.7...

erikkemperman:
Also, the big win here isn't line count, in my humble opinion, but:

- Separate version management so upgrades won't (easily) conflict with user's generated apps (it will be just the line with jhipster-dependencies.version that changes)
- Automatically uniform versions across Maven and Gradle (because the gradle plugin just reads the new POM)
- No more parent POM, leaving users free to attach their own corporate parent.

PR #6514, opened 2017-10-13
Oidc support microservices

danielpetisme:
Suite of #6435


generators/server/index.js
```diff
@@ -202,7 +202,10 @@ module.exports = class extends BaseGenerator {
                 this.authenticationType = this.config.get('authenticationType');
                 // JWT authentication is mandatory with Eureka, so the JHipster Registry
                 // can control the applications
-                if (this.serviceDiscoveryType === 'eureka' && this.authenticationType !== 'uaa') {
+                if (this.serviceDiscoveryType === 'eureka' && (this.authenticationType !== 'uaa' && this.authenticationType !== 'oauth2')) {
+                    console.log("+++++++++++++++++++");
```
jdubois: You should remove the logs - that will probably cause the build to fail (I guess we check this in the lint configuration)

PR #6516, opened 2017-10-13
Fix OAuth compilation issues

mraible:
Fix EJS syntax in `AccountResource.java` and `UserService.java`.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/_UserService.java
+++ b/generators/server/templates/src/main/java/package/service/_UserService.java
@@ -257,6 +257,7 @@ public class UserService {
         log.debug("Created Information for User: {}", user);
         return user;
     }
+<%_ } _%>
 
     /**
      * Update basic information (first name, last name, email, language) for the current user.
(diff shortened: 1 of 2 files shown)


sendilkumarn:
@mraible I have tried to make sure that the user management is skipped completely when auth type is oauth2 and then we can overlay these changes on top of it. 

Probably to avoid confusion we can add in a new file that does this exclusively (may be we can reuse this for others (if we add them in future)).

mraible:
Sounds good. I'm traveling for the weekend with no laptop. I should be able to look at things Sunday morning (Denver time, -7).

jdubois:
I'm sorry @mraible but Travis fails because there is still some compilation issue. Besides, it looks like you have some formatting issue: lots of stuff is re-formatted, but I don't think it's related to what you are trying to solve.
I'm going to try to merge #6517 as it has less issues, but I guess this will cause some conflicts here. I'm having a look and do what I can.

jdubois:
Sorry again @mraible 

- I don't understand this PR, I think you have a formatting issue
- I have corrected several compilation issues, so I hope this is going to work now

-> I still have a huge problem, as Java unit tests don't pass. They need an OAuth2 server to work, and I guess you have one running so you don't see the issue. But for unit tests, this should be mocked or embedded.

PR #6523, opened 2017-10-15
Add User Synchronization to OAuth Monolith

mraible:
Fixes OAuth compiliation

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
+++ b/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
@@ -1,31 +1,41 @@
 <%#
- Copyright 2013-2017 the original author or authors from the JHipster project.
+Copyright 2013-2017 the original author or authors from the JHipster project.
 
- This file is part of the JHipster project, see http://www.jhipster.tech/
- for more information.
+This file is part of the JHipster project, see http://www.jhipster.tech/
+for more information.
 
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-      http://www.apache.org/licenses/LICENSE-2.0
+     http://www.apache.org/licenses/LICENSE-2.0
 
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- See the License for the specific language governing permissions and
- limitations under the License.
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
 -%>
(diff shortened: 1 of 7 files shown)


sendilkumarn:
I think we need to map the user to its login rather than id (as foreign keys everywhere) else this will cause an error. 

Or else we can have an id property in the user.java that just returns the login and change the argument from long to String everywhere we use User.

mraible:
That's why I liked the `User` syncing feature - because nothing had to change with User. If we don't want to add that, we should figure out how to map the login to the id.

sendilkumarn:
This is getting way interesting 😄

jdubois:
Oh I did the same "fix" at least twice - that's not an easy use-case... I guess we should re-work the UserDTO for an IODC microservice, so that it works.

jdubois:
Anyway, in this context a `User` isn't a JPA entity anymore, so does transforming it to a DTO makes a any sense? I would just suppress that feature

jdubois:
This is just not going to work: the issue is not only on the DTO, the JPA entities are also trying to do a relationship to a `User` JPA entity - and here the `User` domain class is not managed by JPA (it's just synced from OIDC).

-> for microservices, we should forbid to do relationships to the User, that's the only simple solution I can find. WDYT?

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -62,13 +85,131 @@ public String isAuthenticated(HttpServletRequest request) {
...
      * @return the current user
      * @throws RuntimeException 500 (Internal Server Error) if the user couldn't be returned
      */
     @GetMapping("/account")
     @Timed
     @SuppressWarnings("unchecked")
-    public ResponseEntity<User> getAccount(Principal principal) {
+    <%_ if (applicationType === 'monolith') { _%>
+    public UserDTO getAccount(Principal principal) {
+        if (principal != null) {
```
sendilkumarn: can we move this logic into service if possible ?

sendilkumarn:
LGTM 👍  🎆

jdubois:
OK I'm trusting you here - I'll merging and will test as soon as I can

jdubois:
I still got an error on the microservice, when doing a DTO and linking it to a `User`. That's a very specific use case, that doesn't worry me too much.

mraible:
@jdubois Can you please enter an issue to track this with steps to reproduce? We should also add a Travis test since this isn't currently caught.

jdubois:
Yes but first I'm testing with Docker

- with a monolith: everything starts OK, but I can't login with admin/admin in Keycloak
- with a gateway+micorservice: both do not start because I get

    `[org.springframework.security.oauth2.provider.token.store.JwtAccessTokenConverter]: Factory method 'jwtAccessTokenConverter' threw exception; nested exception is org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://localhost:9080/auth/realms/jhipster": Connection refused (Connection refused)`

-> so none work out of the box, which are much bigger issues

mraible:
For the monolith, are you running everything (including the app) in Docker? If so, you need an entry for "keycloak" in your hosts file.

jdubois:
@mraible yes I think I found the issue with Docker, let me do it

jdubois:
@mraible in fact I can correct my 2 issues above, and probably also remove your "keycloak in /etc/host" trick. I'll link the commit here.

jdubois:
OK I can't remove the "/etc/hosts" trick :-(

PR #6537, opened 2017-10-16
Allow multiple liquibase commands in Gradle, similar to Maven

harry-g:
Gradle and liquibase allows only diffChangeLog command.
With this change, you can specify others with task liquibase and -PliquibaseCommand, e.g. 
   gradlew liquibase -PliquibaseCommand=clearChecksums
This is similar to Maven liquibase:clearChecksums
The existing task liquibaseClearChecksums is retained for compatibility.

### DIFF
--- a/generators/server/templates/gradle/_liquibase.gradle
+++ b/generators/server/templates/gradle/_liquibase.gradle
@@ -77,7 +77,7 @@ task liquibase(dependsOn: compileJava, type: JavaExec) {
 }
 
 // for compatibility to former task
-task liquibaseClearChecksums() {
+task liquibaseDiffChangelog() {
 	liquibaseCommand = "diffChangeLog"
 	liquibase.execute()
 }
--- a/generators/server/templates/gradle/_liquibase.gradle
+++ b/generators/server/templates/gradle/_liquibase.gradle
@@ -73,12 +73,16 @@ def liquibaseCommand(command) {
     }
 }
 
-task liquibaseDiffChangeLog(dependsOn: initPaths) << {
-	liquibaseCommand("diffChangeLog")
+task liquibaseDiffChangeLog(dependsOn: initPaths) {
+    doLast {
+	    liquibaseCommand("diffChangeLog")
+    }
 }
 
-task liquibaseClearChecksums(dependsOn: initPaths) << {
-	liquibaseCommand("clearChecksums")
+task liquibaseClearChecksums(dependsOn: initPaths) {
+    doLast {
+        liquibaseCommand("clearChecksums")
+    }
 }
 
 def buildTimestamp() {
--- a/generators/server/templates/gradle/_liquibase.gradle
+++ b/generators/server/templates/gradle/_liquibase.gradle
@@ -18,6 +18,13 @@
 -%>
 import org.gradle.internal.os.OperatingSystem
 
+if (!hasProperty('liquibaseCommand')) {
+	ext.liquibaseCommand = 'diffChangeLog'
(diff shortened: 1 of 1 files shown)


atomfrede:
Hi @harry-g and thanks for your PR 👍 and especially congrats to your first contribution yo jhipster :)

I'm sure some might need this feature (at least I know one person), but nevertheless I;m a bit reluctant to merge this for mainly two reasons:

* We are currenty in progress of making gradle support fit more the gradle approach and not just try to mimic maven funtionality
* In particular this means (I know currently not the case) trying to avoid `-P` switches as they are not easy discoverable by the user (you can follow https://github.com/jhipster/generator-jhipster/issues/6252 for more information)
* Furthermore I would like to use external plugins where possible (currently we replaced custom docker tasks and custom gatling tasks with plugins). So for me the way to go is the https://github.com/liquibase/liquibase-gradle-plugin for better integration with liquibase

So if you like you can change your PR and add the plugin (need to make sure the correct version is used)

atomfrede:
@harry-g As it seems the `diffChangelog` task in the gradle plugin doesn't work well I suggest to stick with our custom taks. But I would prefer having not another `-P` switch in our build.

generators/server/templates/gradle/_liquibase.gradle
```diff
@@ -63,12 +70,18 @@ task liquibaseDiffChangelog(dependsOn: compileJava, type: JavaExec) {
...
     args "--driver=<% if (devDatabaseType === 'mysql') { %>com.mysql.jdbc.Driver<% } else if (devDatabaseType === 'mariadb') { %>org.mariadb.jdbc.Driver<% } else if (devDatabaseType === 'postgresql') { %>org.postgresql.Driver<% } else if (devDatabaseType === 'h2Disk') { %>org.h2.Driver<% } else if (devDatabaseType === 'oracle') { %>oracle.jdbc.OracleDriver<% } %>"
-    args "diffChangeLog"
+    args liquibaseCommand
     <%_ if (authenticationType === 'oauth2') { _%>
     args "--excludeObjects=oauth_access_token, oauth_approvals, oauth_client_details, oauth_client_token, oauth_code, oauth_refresh_token"
     <%_ } _%>
 }
 
+// for compatibility to former task
+task liquibaseClearChecksums() {
```
atomfrede: Shouldn't it be named `liquibaseDiffChangelog`?

harry-g: Ups, sure, sorry! Must have been some copy-paste error - thanks for reviewing!

harry-g:
Hi @atomfrede , thanks for your comments.
Sure -P is not nice, sorry that I was not aware of your valid ongoing discussion!
Maybe we can go for my new proposal until the liquibase plugin is useable?
It would have the clearChecksums task and further tasks could be added easily if needed.

atomfrede:
@harry-g Yes we will definitly go with our customs tasks as the liquibase plugin is currently not useable for us. Not sure if we can fix it upstream in a timely manner. So maybe we will put the liquibase stuff into an own jhipster-liquibase plugin, but not now.

harry-g:
@atomfrede , so maybe you want to merge this PR in the most recent variant with custom tasks and w/o -P? Would be a small improvement at least for Gradle users...

atomfrede:
Have you seen my latest comment about << ? But yes I will merge it

harry-g:
In this PR? Sorry, cannot see anything about << ...

atomfrede:
Sorry, somehow I missed to submit the review :( Now you should see what I mean

harry-g:
OMG, Gradle is always deprecating faster than we can hack *lol*
Ya, OK, it is one year already, I have missed it somehow... fixed now!

PR #6540, opened 2017-10-16
Fix sonar bugs so we pass the quality gate

jdubois:


generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -274,140 +282,149 @@ public User getAccount(Principal principal) {
...
     *
     * @return the current user
     * @throws RuntimeException 500 (Internal Server Error) if the user couldn't be returned
     */
-   @GetMapping("/account")
-   @Timed
-   public UserDTO getAccount() {
-       return Optional.ofNullable(userService.getUserWithAuthorities())
-           .map(UserDTO::new)
-           .orElseThrow(RuntimeException::new);
```
cbornet: `.orElseThrow(() -> new DefaultException("The current logged in user could not be found"));` would be better

jdubois: yes but the main issue comes from the `.orElseThrow()` : as the return parameter is not used, Sonar is complaining. This is what I want to remove.

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -200,17 +201,20 @@ public UserDTO getAccount(Principal principal) {
                 return userDTO;
             } else {
                 // Allow Spring Security Test to be used to mock users in the database
-                return Optional.ofNullable(userService.getUserWithAuthorities())
-                    .map(UserDTO::new)
-                    .orElseThrow(RuntimeException::new);
+                Optional<User> user = Optional.ofNullable(userService.getUserWithAuthorities());
```
cbornet: `.orElseThrow(() -> new DefaultException("User could not be found"));` would be better

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -231,8 +235,12 @@ public User getAccount(Principal principal) {
                             it.getAuthority()).collect(Collectors.toSet())
                     );
                 }
-            )
-            .orElseThrow(RuntimeException::new);
+            );
```
cbornet: `.orElseThrow(() -> new DefaultException("User could not be found"));` would be better

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
jdubois: Same as above, I want to remove the `.orElseThrow` method for Sonar

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -274,140 +282,149 @@ public User getAccount(Principal principal) {
...
     * @return the current open sessions
     * @throws RuntimeException 500 (Internal Server Error) if the current open sessions couldn't be retrieved
     */
-   @GetMapping("/account/sessions")
-   @Timed
-   public List<PersistentToken> getCurrentSessions() {
-       return persistentTokenRepository.findByUser(
-           userRepository.findOneByLogin(SecurityUtils.getCurrentUserLogin())
-               .orElseThrow(RuntimeException::new)
-       );
```
cbornet: `.orElseThrow(() -> new DefaultException("User could not be found"));` would be better

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -423,53 +440,58 @@ public void changePassword(@RequestBody String password) {
...
     */
-   @PostMapping(path = "/account/reset-password/init")
-   @Timed
-   public void requestPasswordReset(@RequestBody String mail) {
-       mailService.sendPasswordResetMail(
-           userService.requestPasswordReset(mail)
-               .orElseThrow(EmailNotFoundException::new)
-       );
-   }
+    @PostMapping(path = "/account/reset-password/init")
```
cbornet: Why ? This code is fine according to Sonar !

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -423,53 +440,58 @@ public void changePassword(@RequestBody String password) {
...
     * @throws RuntimeException 500 (Internal Server Error) if the password could not be reset
     */
-   @PostMapping(path = "/account/reset-password/finish")
-   @Timed
-   public void finishPasswordReset(@RequestBody KeyAndPasswordVM keyAndPassword) {
-       if (!checkPasswordLength(keyAndPassword.getNewPassword())) {
-           throw new InvalidPasswordException();
-       }
-       userService.completePasswordReset(keyAndPassword.getNewPassword(), keyAndPassword.getKey())
-           .orElseThrow(RuntimeException::new);
```
cbornet: `.orElseThrow(() -> new DefaultException("No user was found for this reset key."));` would be better

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: No the return param is used here and Sonar doesn't complain

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: No the return param is used here and Sonar doesn't complain

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
jdubois: On our Sonar Docker image yes, but not on SonarCloud, which uses the "enterprise" version (with more checks): https://sonarcloud.io/project/issues?id=io.github.jhipster.sample%3Ajhipster-sample-application&resolved=false&types=BUG

cbornet: No but here, even on SonarCloud !!

cbornet: There are only 3 orElseThrow to change, not all of them !

generators/server/templates/src/main/java/package/web/rest/errors/_DefaultException.java
```diff
@@ -0,0 +1,32 @@
...
+-%>
+package <%=packageName%>.web.rest.errors;
+
+import org.zalando.problem.AbstractThrowableProblem;
+import org.zalando.problem.Status;
+
+/**
+ * Generic exception with a message, that returns an Internal Server Error code.
+ */
+public class DefaultException extends AbstractThrowableProblem {
```
ctamisier: In zalando there is already a DefaultProblem class so i'm wondering if it could be confusing to have DefaultProblem(zalanda) and DefaultException(jhipster) implementing both Exception and Problem ?

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
jdubois: It's because the other are from other options, which are not selected on this SonarCloud config. All `.orElseThrow` have this issue, because they return a value we do not use. I know this rule is stupid, of course.

generators/server/templates/src/main/java/package/web/rest/errors/_DefaultException.java
jdubois: Oh yes, let's rename this... GenericException?

cbornet: Agreed ! The best would be to `throw ProblemBuilder.setTitle('whatever').build()`

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: I don't agree : we use the return valur of orElseThrow

cbornet: And Sonar doesn't complain

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
jdubois: oh yes you are correct

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
jdubois: yes I'm changing all those

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: `.orElseThrow(() -> ProblemBuiler.withTitle("User could not be found").build());` and DefaultException class can be removed (the method then throws DefaultProblem)

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: .orElseThrow(() -> ProblemBuilder.withTitle("No user was found for this reset key.").build());

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: .orElseThrow(() -> ProblemBuilder.withTitle("User could not be found").build());

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
```diff
@@ -200,17 +201,20 @@ public UserDTO getAccount(Principal principal) {
...
+                Optional<User> user = Optional.ofNullable(userService.getUserWithAuthorities());
+                if (user.isPresent()) {
+                    return new UserDTO(user.get());
+                } else {
+                    throw new DefaultException("User could not be found");
+                }
             }
         } else {
-            throw new RuntimeException();
+            throw new DefaultException("User could not be found");
```
cbornet: `throw ProblemBuilder.withTitle("User could not be found").build());`

jdubois: Unfortunately those doesn't work, as `withTitle` is not a static method

generators/server/templates/src/main/java/package/web/rest/errors/_DefaultException.java
jdubois: No I can't use `ProblemBuilder` as `withTitle` is not a static method.
Besides using a `GenericException` is more concise

generators/server/templates/src/main/java/package/web/rest/_AccountResource.java
cbornet: Oups ! It's `Problem.builder().withTitle...` !

generators/server/templates/src/main/java/package/web/rest/errors/_DefaultException.java
jdubois: Maybe `SimpleMessageException` is better. At least there is a meaning in the name

cbornet: Oups ! It's Problem.builder().withTitle... !

jdubois: Oh OK! Still I like `SimpleMessageException`, it's smaller - is it fine with you?

cbornet: Yes. And I just saw you also change the `type` so a custom exception is OK.
Actually I would vote for `InternalServerException` since that's the status used.

jdubois: OK let's go for a `InternalServerErrorException`

cbornet: :+1:

Issue #6269, opened 2017-08-21
UserServiceIntTest failure


cmyker:
##### **Overview of the issue**

Vanilla jHipster installation is failing integration test when you run whole integration tests package.
The reason is that another test class (AccountResourceIntTest) is running before and deleting the data that this test relies on.
##### **Reproduce the error**

I just used fresh jHipster 4.6.2 installation 
##### **Suggest a Fix**

https://github.com/jhipster/generator-jhipster/pull/6251
##### **JHipster Version(s)**

4.6.2
##### **JHipster configuration**

jhipster info
Executing jhipster:info
Options: 
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
admin@0.0.0 /home/igor/Projects/tracker/admin
└── generator-jhipster@4.6.2 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.sevensenders.tracker.admin"
    },
    "jhipsterVersion": "4.6.2",
    "baseName": "admin",
    "packageName": "com.sevensenders.tracker.admin",
    "packageFolder": "com/sevensenders/tracker/admin",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "no",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "mongodb",
    "devDatabaseType": "mongodb",
    "prodDatabaseType": "mongodb",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "rememberMeKey": "replaced-by-jhipster-info",
    "clientFramework": "angularX",
    "useSass": true,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
</pre>
</details>

##### **JDL for the Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

<details>
<summary>JDL entity definitions</summary>

<pre>

</pre>
</details>

##### **Environment and Tools**

openjdk version "1.8.0_144"
OpenJDK Runtime Environment (IcedTea 3.5.1) (suse-10.13.3-x86_64)
[...]

PR #6544, opened 2017-10-17
Make UserService test more robust

cbornet:

The UserService test is brittle since it relies on a predefined state of the User repository (eg. it assumes that an `admin` user [is present](https://github.com/jhipster/generator-jhipster/blob/v4.9.0/generators/server/templates/src/test/java/package/service/_UserServiceIntTest.java#L69)). So if a previous test does a cleaning of the User repository (eg. [UserResourceIntTest](https://github.com/jhipster/generator-jhipster/blob/v4.9.0/generators/server/templates/src/test/java/package/web/rest/_UserResourceIntTest.java#L196) if for some reason your tests are not executed in alphabetical path order), then the test will fail. See #6269 for a test failure case.

### DIFF
--- a/generators/server/templates/src/test/java/package/service/_UserServiceIntTest.java
+++ b/generators/server/templates/src/test/java/package/service/_UserServiceIntTest.java
@@ -16,45 +16,60 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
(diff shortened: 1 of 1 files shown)


jdubois:
Huge work!

PR #6553, opened 2017-10-18
Add UserResource to OAuth Monolith

mraible:
Adding `UserResource` allows many-to-one relations to the `User` object.

### DIFF
--- a/generators/server/templates/src/test/java/package/web/rest/_UserResourceIntTest.java
+++ b/generators/server/templates/src/test/java/package/web/rest/_UserResourceIntTest.java
@@ -31,7 +31,8 @@ import <%= packageName %>.repository.UserRepository;
 import <%= packageName %>.repository.search.UserSearchRepository;
 <%_ } _%>
 import <%= packageName %>.security.AuthoritiesConstants;
-import <%= packageName %>.service.MailService;
+<%_ if (authenticationType !== 'oauth2') { _%>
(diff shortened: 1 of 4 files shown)


generators/server/templates/src/test/java/package/web/rest/_UserResourceIntTest.java
```diff
@@ -63,10 +64,7 @@
 <%_ if (databaseType !== 'cassandra') { _%>
 import java.time.Instant;
 <%_ } _%>
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
mraible: `java.util.Arrays` and `java.util.List` is not used with OAuth; it seemed easier to import `java.util.*` than have if statements for each import.

cbornet:
Since this PR, the test context is restarted by AccountResourceIntTest for app-ng2-oauth :disappointed: : https://travis-ci.org/jhipster/generator-jhipster/jobs/289920669#L2381

jdubois:
I also have the `AccountResourceIntTest` which does not even compile on a OAuth2 microservice - there's definitely something broken here

jdubois:
Same thing for an OAuth2 gateway... I'm doing a specific ticket for all this, as this is getting really big. We don't test those in our build, for as far as I'm concerned this means the build is broken and we can't do a release now (and I wanted to do one for @xetys )

PR #6572, opened 2017-10-20
Fix to not call unnecessarily the /api/profile-info method multiple times

gzsombor:
Twice for opening the home page, one call is from the navbar.component.ts, the second is from the page-ribbon.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
@@ -39,7 +39,7 @@ export class PageRibbonComponent implements OnInit {
     constructor(private profileService: ProfileService) {}
 
     ngOnInit() {
-        this.profileService.getProfileInfo().subscribe((profileInfo) => {
+        this.profileService.getProfileInfo().then((profileInfo) => {
             this.profileInfo = profileInfo;
             this.ribbonEnv = profileInfo.ribbonEnv;
         });
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_profile.service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_profile.service.ts
@@ -18,7 +18,6 @@
 -%>
 import { Injectable } from '@angular/core';
 import { Http, Response } from '@angular/http';
-import { Observable } from 'rxjs/Rx';
 
 <%_ if (authenticationType !== 'uaa') { _%>
 import { SERVER_API_URL } from '../../app.constants';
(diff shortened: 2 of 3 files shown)


gzsombor:
Unfortunately, I was too fast. This patch doesn't fix the issue, as the second 'subscribe' triggers the call again, if there were more delay between the two call.

deepu105:
Not sure if changing to promise is the correct solution. If we really want to avoid the call then better way would be to cache the result in the service as an instance variable and use that when available instead of making api call

gzsombor:
This is what the fix tries to do, cache the promise in the service as an instance variable. Probably I misunderstood something :confused:

deepu105:
Ill take a closer look when i have time

generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_profile.service.ts
```diff
@@ -29,19 +28,23 @@ import { ProfileInfo } from './profile-info.model';
 export class ProfileService {
 
     private profileInfoUrl = <% if (authenticationType === 'uaa') { %>'<% } else { %>SERVER_API_URL + '<% } %>api/profile-info';
+    private profileInfo: Promise<ProfileInfo>;
```
deepu105: This can be done with Observable as well and I personally would prefer that to keep things consistent

generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_page-ribbon.component.ts
```diff
@@ -39,7 +39,7 @@ export class PageRibbonComponent implements OnInit {
     constructor(private profileService: ProfileService) {}
 
     ngOnInit() {
-        this.profileService.getProfileInfo().subscribe((profileInfo) => {
+        this.profileService.getProfileInfo().then((profileInfo) => {
```
deepu105: another option to avoid this call would be to pass the `profileInfo` to the component as a property from the navbar component

generators/client/templates/angular/src/main/webapp/app/layouts/profiles/_profile.service.ts
```diff
@@ -29,19 +28,23 @@ import { ProfileInfo } from './profile-info.model';
...
-                const data = res.json();
-                const pi = new ProfileInfo();
-                pi.activeProfiles = data.activeProfiles;
-                pi.ribbonEnv = data.ribbonEnv;
-                pi.inProduction = data.activeProfiles.indexOf('prod') !== -1;
-                pi.swaggerEnabled = data.activeProfiles.indexOf('swagger') !== -1;
-                return pi;
-            });
+    getProfileInfo(): Promise<ProfileInfo> {
+        if (!this.profileInfo) {
```
deepu105: I too don't understand why this wouldn't work unless the class is instantiated per call which would be strange. I dont have time to debug it and figure whats wrong but since this is not a pressing issue I guess we can wait to find a good solution

gzsombor: http.get returns a new Observable, every time when it's called, so that's why currently it sends multiple requests.
If the Observable is cached, it seems, that the 'subscribe' call triggers a second request, only if, the first request is already received.
 In other places of the app, we use Promise-s, for example for the authentication, this profile info, in my opinion, should work the same

deepu105: Yes we still use promise at some places but those are mostly legacy code and there was a ticket to migrate them to observable. Any way lets wait and tackle this later

jdubois:
LGTM let's merge this, thanks @gzsombor !

deepu105:
@jdubois I guess this was not working. @gzsombor did you manage to get it working?

gzsombor:
I don't understand, what was not working?

deepu105:
@gzsombor I was referring to this comment from you above

> Unfortunately, I was too fast. This patch doesn't fix the issue, as the second 'subscribe' triggers the call again, if there were more delay between the two call.

gzsombor:
Ahh ! It was about the first version of the commit, after I've force pushed a new one, where I've fixed the issue 😉

PR #6591, opened 2017-10-24
remove dirty file

ctamisier:
removing a probably undesired committed file ?


jdubois:
Yes, looks like an error, thanks!!

Issue #1810, opened 2015-08-03
MongoDb repos with ElasticSearch.


Selectorrr:
I think they are not compatible.
Example: https://github.com/Selectorrr/petsroom-jh
```
Caused by: org.springframework.data.mapping.PropertyReferenceException: No property insert found for type User!
	at org.springframework.data.mapping.PropertyPath.<init>(PropertyPath.java:75)
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:327)
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:307)
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:270)
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:241)
	at org.springframework.data.repository.query.parser.Part.<init>(Part.java:76)
	at org.springframework.data.repository.query.parser.PartTree$OrPart.<init>(PartTree.java:235)
	at org.springframework.data.repository.query.parser.PartTree$Predicate.buildTree(PartTree.java:373)
	at org.springframework.data.repository.query.parser.PartTree$Predicate.<init>(PartTree.java:353)
	at org.springframework.data.repository.query.parser.PartTree.<init>(PartTree.java:84)
	at org.springframework.data.elasticsearch.repository.query.ElasticsearchPartQuery.<init>(ElasticsearchPartQuery.java:40)
	at org.springframework.data.elasticsearch.repository.support.ElasticsearchRepositoryFactory$ElasticsearchQueryLookupStrategy.resolveQuery(ElasticsearchRepositoryFactory.java:117)
	at org.springframework.data.repository.core.support.RepositoryFactorySupport$QueryExecutorMethodInterceptor.<init>(RepositoryFactorySupport.java:369)
	at org.springframework.data.repository.core.support.RepositoryFactorySupport.getRepository(RepositoryFactorySupport.java:192)
	at org.springframework.data.repository.core.support.RepositoryFactoryBeanSupport.initAndReturn(RepositoryFactoryBeanSupport.java:239)
	at org.springframework.data.repository.core.support.RepositoryFactoryBeanSupport.afterPropertiesSet(RepositoryFactoryBeanSupport.java:225)
	at org.springframework.data.elasticsearch.repository.support.ElasticsearchRepositoryFactoryBean.afterPropertiesSet(ElasticsearchRepositoryFactoryBean.java:54)
[...]

PR #6595, opened 2017-10-25
Add elasticsearch option when mongo is selected

ramzimaalej:
This is a first attempt to fix this issue https://github.com/jhipster/generator-jhipster/issues/1810

@jdubois @deepu105 Let me know if you have any feedback on this. I also would like to add tests to this and solve this non-blocking issue:

`org.reflections.ReflectionsException: could not create Vfs.Dir from url, no matching UrlType was found [file:/Users/rmaalej/Documents/dev/workspace/expensy/expensy-analytics/build/classes/scala/main]
either use fromURL(final URL url, final List<UrlType> urlTypes) or use the static setDefaultURLTypes(final List<UrlType> urlTypes) or addDefaultURLTypes(UrlType urlType) with your specialized UrlType.
        at org.reflections.vfs.Vfs.fromURL(Vfs.java:109)
        at org.reflections.vfs.Vfs.fromURL(Vfs.java:91)`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ env:
     - JHIPSTER=app-ng2-infinispan
     - JHIPSTER=app-ng2-cassandra
     - JHIPSTER=app-ng2-mongodb
+    - JHIPSTER=app-ng2-mongodb-elasticsearch
     - JHIPSTER=app-ng2-oauth2 PROTRACTOR=1
     - JHIPSTER=app-react-default
     - JHIPSTER=app-mysql PROFILE=prod PROTRACTOR=1
--- a/generators/entity-server/templates/src/main/java/package/domain/_Entity.java
+++ b/generators/entity-server/templates/src/main/java/package/domain/_Entity.java
@@ -36,19 +36,22 @@ import org.hibernate.annotations.CacheConcurrencyStrategy;
 <%_ } if (databaseType === 'mongodb') { _%>
 import org.springframework.data.annotation.Id;
 import org.springframework.data.mongodb.core.mapping.Field;
+import org.springframework.data.mongodb.core.mapping.Document;
 <%_ } else if (databaseType === 'couchbase') { _%>
 import org.springframework.data.annotation.Id;
 import com.couchbase.client.java.repository.annotation.Field;
 import org.springframework.data.couchbase.core.mapping.Document;
 import org.springframework.data.couchbase.core.mapping.id.GeneratedValue;
 import org.springframework.data.couchbase.core.mapping.id.IdPrefix;
-<%_ } if (searchEngine === 'elasticsearch') { _%>
-import org.springframework.data.elasticsearch.annotations.Document;
 <%_ } if (databaseType === 'sql') { _%>
 
 import javax.persistence.*;
 <%_ } if (validation) { _%>
 import javax.validation.constraints.*;
+<%_ } _%>
+
+<%_ if (searchEngine === 'elasticsearch' && databaseType !== 'mongodb') { _%>
+import org.springframework.data.elasticsearch.annotations.Document;
 <%_ } _%>
 import java.io.Serializable;
 <%_ if (fieldsContainBigDecimal === true) { _%>
@@ -98,13 +101,15 @@ import static org.springframework.data.couchbase.core.mapping.id.GenerationStrat
 <%_         }
         }
 } if (databaseType === 'mongodb') { _%>
-@org.springframework.data.mongodb.core.mapping.Document(collection = "<%= entityTableName %>")
+@Document(collection = "<%= entityTableName %>")
 <%_ } if (databaseType === 'couchbase') { _%>
 @Document
 <%_ } if (databaseType === 'cassandra') { _%>
 @Table(name = "<%= entityInstance %>")
-<%_ } if (searchEngine === 'elasticsearch') { _%>
+<%_ } if (searchEngine === 'elasticsearch' && databaseType === 'mongodb') { _%>
 @org.springframework.data.elasticsearch.annotations.Document(indexName = "<%= entityInstance.toLowerCase() %>")
+<%_ } if (searchEngine === 'elasticsearch' && databaseType !== 'mongodb') { _%>
+@Document(indexName = "<%= entityInstance.toLowerCase() %>")
 <%_ } _%>
 public class <%= entityClass %> implements Serializable {
 
--- a/generators/entity-server/templates/src/main/java/package/domain/_Entity.java
+++ b/generators/entity-server/templates/src/main/java/package/domain/_Entity.java
@@ -35,7 +35,6 @@ import org.hibernate.annotations.Cache;
 import org.hibernate.annotations.CacheConcurrencyStrategy;
 <%_ } if (databaseType === 'mongodb') { _%>
 import org.springframework.data.annotation.Id;
-import org.springframework.data.mongodb.core.mapping.Document;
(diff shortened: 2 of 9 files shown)


pascalgrimaud:
Congrats @ramzimaalej !
It's a feature a lot of people wanted since 2 years, but no one contribute to it.

jdubois:
Yes, awesome!

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
```diff
@@ -88,11 +85,11 @@
 <%_         }
         }
 } if (databaseType === 'mongodb') { _%>
-@Document(collection = "<%= entityTableName %>")
+@org.springframework.data.mongodb.core.mapping.Document(collection = "<%= entityTableName %>")
```
deepu105: To make it less ugly when using mongo without ES I suggest to create an intance variable called mongoDocumentAnnotation and use it here the value of which can be qualified path when using ES and just document when using without ES. The same gies for ES annotation and the imports

ramzimaalej: sure, I will take care of it

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
```diff
@@ -91,7 +93,9 @@
 @Document(collection = "<%= entityTableName %>")
 <%_ } if (databaseType === 'cassandra') { _%>
 @Table(name = "<%= entityInstance %>")
-<%_ } if (searchEngine === 'elasticsearch') { _%>
+<%_ } if (searchEngine === 'elasticsearch' && databaseType === 'mongodb') { _%>
+@org.springframework.data.elasticsearch.annotations.Document(indexName = "<%= entityInstance.toLowerCase() %>")
```
deepu105: While this is correct my idea was to simplify this with a variable. Dont worry ill do it while merging. Ill merge this later

deepu105:
Please check the test failure for new setup

ramzimaalej:
@deepu105 I'm looking into the issue, it's related to mapstruct bean not being injected.

ramzimaalej:
Intellij was misleading me, the issue has nothing to do with Mapstruct. Actually, it's because of the ZonedDateTime converter.

`public static class DateToZonedDateTimeConverter implements Converter<Date, ZonedDateTime> {
        public static final JSR310DateConverters.DateToZonedDateTimeConverter INSTANCE = new JSR310DateConverters.DateToZonedDateTimeConverter();

        private DateToZonedDateTimeConverter() {
        }

        public ZonedDateTime convert(Date source) {
            return source == null ? null : ZonedDateTime.ofInstant(source.toInstant(), ZoneId.systemDefault());
        }
    }

    public static class ZonedDateTimeToDateConverter implements Converter<ZonedDateTime, Date> {
        public static final JSR310DateConverters.ZonedDateTimeToDateConverter INSTANCE = new JSR310DateConverters.ZonedDateTimeToDateConverter();

        private ZonedDateTimeToDateConverter() {
        }

        public Date convert(ZonedDateTime source) {
            return source == null ? null : Date.from(source.toInstant());
        }
    }`

while technically speaking the generated object represent the same timestamp, they are slightly different causing this line to fail `assertThat(fieldTestMapstructEntityEs).isEqualToComparingFieldByField(testFieldTestMapstructEntity);`

I have changed to use string instead of data and it works like a charm

`public static class StringToZonedDateTimeConverter implements Converter<String, ZonedDateTime> {
        public static final StringToZonedDateTimeConverter INSTANCE = new StringToZonedDateTimeConverter();

        private StringToZonedDateTimeConverter() {
        }

        public ZonedDateTime convert(String source) {
            return source == null ? null : ZonedDateTime.parse(source);
        }
    }

    public static class ZonedDateTimeToStringConverter implements Converter<ZonedDateTime, String> {
        public static final ZonedDateTimeToStringConverter INSTANCE = new ZonedDateTimeToStringConverter();

        private ZonedDateTimeToStringConverter() {
        }

        public String convert(ZonedDateTime source) {
            return source == null ? null : source.toString();
        }
    }`

I raised a PR on jhipster core: https://github.com/jhipster/jhipster/pull/31

generators/entity/templates/server/src/main/java/package/domain/_Entity.java
ramzimaalej: I think this way offers better readability of what's going in the template unless you want to put the logic here, but it's up to you. I don't have a strong opinion about it.

cbornet:
The issue I see is that ES stores `Date` objects as `Long` so using a String will be less preformant and compact.
The issue with `isEqualToComparingFieldByField()` is that it uses the equals() method which checks the timezone of ZonedDateTime but as we converted to Date before storage we lose this information. We have a matcher `TestUtil.sameInstant` for the other tests that checks ZonedDateTime ignoring the timezone.
So I think it would be better to use `isEqualToIgnoringGivenFields` , ignoring the ZonedDateTime fields and use individual `sameInstant` matchers for the remaining ZonedDateTime fields.

ramzimaalej:
If we use long we will also lose the timezone, because when you create a ZonedDateTime from an Instant you need to specify the Timezone. That said, there is no way to guess the original timezone. I would rather use a string representation that preserves the original timezone and avoids adding complexity to the generator to find out which fields that are ZonedDateTime to be treated differently in assert statement.

cbornet:
Most of our DBs (currently all of them except cassandra) don't preserve the timezone and the view layer will remove it anyway so it's not a big deal to lose it.

ramzimaalej:
done. let me know if there is anything else to address in this PR.

ramzimaalej:
@deepu105 @jdubois do you guys feel there is more stuff to do here?

jdubois:
I just need time to review but this looks quick to do

pascalgrimaud:
LGTM
Just give a quick try, and it works well. Good job @ramzimaalej 
- generated a project with MongoDB + ES
- generated some entities
- launched app
- played with entities + ES

So I don't know if this PR is needed to be merged https://github.com/jhipster/jhipster/pull/31

ramzimaalej:
@pascalgrimaud I think we can discard it for now (jhipster/jhipster#31).

ramzimaalej:
@pascalgrimaud can you please restart the build.

pascalgrimaud:
it's green, so it's ok ?

ramzimaalej:
I believe so. Once merged I will work on adding elasticsearch to couchbase as well.

ramzimaalej:
@deepu105 @jdubois any plans on when this will get merged?

jdubois:
@ramzimaalej not before the next release, where we modified our build system (with jhipster/jhipster-dependencies) and added Couchbase support. This is already a huge release, and we need some time to iron it out.
So that would be the release after.

ramzimaalej:
@deepu105 done

pascalgrimaud:
@ramzimaalej : just for your information, as soon as it's merged, I will probably move your Travis tests to https://github.com/hipster-labs/jhipster-travis-build because I'm not sure MongoDB + Elasticsearch will be used a lot. And I will add MongoDB + Elasticsearch + Gradle tests too.

I don't request change here as you already work a lot on this.
Thanks for your patience!

ramzimaalej:
no prob @pascalgrimaud Let me know if you need help.

jdubois:
LGTM and @deepu105 your changes were corrected, so let's merge this.
Congrats @ramzimaalej on a huge new feature!

PR #6606, opened 2017-10-26
Fix yarn failure when upgrading microservices

gmarziou:
When generating an app without client code, upgrade generator fails because yarn complains about no licence. Adding `"private": true` to `package.json` avoids this error and makes a lot of sense as there's no npm module to publish anyway.

See https://github.com/yarnpkg/yarn/issues/3821


jdubois:
We also need to have the UNLICENSED keyword like for an Angular app - I'll do this

gmarziou:
Thanks

PR #6608, opened 2017-10-28
use setters instead of constructor for UserDTO and ManagedUserVM

ctamisier:
- Define setters in UserDTO and ManagedUserVM
- Avoid to have a constructor with "too" many arguments..
- Knowing what fields are set, (vs comments in the constructor call)
- Avoid pass `null` for several fields in Tests (`createdBy`, `createdDate`, `lastModifiedBy`, `lastModifiedDate`)
- having UserDTO and ManagedUserVM looking the same that EntityDTO (getters + setters)

It is mostly about readability (big constructor vs setters) and people taste as well (?), so feel free to close it if the added value is not worth or if I missed something. thx.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
+++ b/generators/server/templates/src/main/java/package/service/dto/_UserDTO.java
@@ -86,36 +86,27 @@ public class UserDTO {
     }
 
     public UserDTO(User user) {
-        this(user.getId(), user.getLogin(), user.getFirstName(), user.getLastName(),
-            user.getEmail(), user.getActivated(),<% if (databaseType === 'sql' || databaseType === 'mongodb') { %> user.getImageUrl(), <% } %>user.getLangKey(),<% if (databaseType === 'sql' || databaseType === 'mongodb') { %>
-            user.getCreatedBy(), user.getCreatedDate(), user.getLastModifiedBy(), user.getLastModifiedDate(),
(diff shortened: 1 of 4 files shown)


ramzimaalej:
I like the idea of avoiding many arguments in the constructor, but I would rather use a builder than doing multiple assignments using setters.

ctamisier:
I started this PR by using fluent setters but because of the inheritance, once we use a fluent setter from the top class (UserDTO) we can not chain with the .password(pwd) because we have a UserDTO and not a ManagedUserVM. 
So I needed to introduce generics (for a Builder as well) to have it working and I concluded that it was maybe too much...
But yeah I'm totally willing to refactor it with a Builder.

ctamisier:
So I investigated a bit more about using Builder having entities with inheritance and I don't really find something simple and will generate a lot code.
Same thing by using fluent setters, the solution with generics implies some unchecked cast warnings. Finding some patterns that won't be too intrusive would be great..

deepu105:
Long constructors are not ideal but they are simple and easy to maintain
compared to generating builders etc. So I wouldn't spend too much time on
this as the value is not that much

deepu105:
Also these are mostly used in tests so im fine with both constructor or setter approach no need to complicate further

ramzimaalej:
sounds good to me.

Issue #6612, opened 2017-10-29
JHipster registry does not start with native cloud config profile in generated jhipster-registry.yml docker compose file


vberetti:
##### **Overview of the issue**

- Generate a jhipster microservice gateway with jhipster registry and eureka
- in src/main/docker/jhipster-registry.yml, change SPRING_PROFILES_ACTIVE from "dev" to "dev,native" in order to use native mode cloud config.
- docker-compose run -f jhipster-registry.yml will fail because "jhipster.registry.password" property is missing

##### **Motivation for or Use Case**

Can not use the native mode of cloud config server just by adding ",native" to the spring active profiles.

##### **Suggest a Fix**

This is due to the generated local file "src/main/docker/central-server-config/localhost-config/application.yml" that has the following definition :

eureka.client.service-url.defaultZone: http://admin:${jhipster.registry.password}@localhost:8761/eureka/

Docker compose for jhipster registry should contain a JHIPSTER_REGISTRY_PASSWORD environment variable.

PR : #6613

##### **JHipster Version(s)**

4.10.2

pascalgrimaud:
Closing this, as the PR is merged

PR #6613, opened 2017-10-29
#6612 add default jhipster registry password in server docker compose

vberetti:
See #6612 for details


pascalgrimaud:
Thanks @vberetti !

PierreBesson:
It's really bothering me that this was broken again ! We should really do integration testing on this.

PR #6474, opened 2017-10-05
Fix version of selenium chromedriver for AngularJS

pascalgrimaud:
Fix protractor builds for ng1 application:
```
[16:43:06] I/direct - Using ChromeDriver directly...
[16:43:06] E/direct - Error code: 135
[16:43:06] E/direct - Error message: Could not find chromedriver at /home/travis/app/node_modules/webdriver-manager/selenium/chromedriver_2.32. Run 'webdriver-manager update' to download binaries.
[16:43:06] E/direct - Error: Could not find chromedriver at /home/travis/app/node_modules/webdriver-manager/selenium/chromedriver_2.32. Run 'webdriver-manager update' to download binaries.
    at IError (/home/travis/app/node_modules/protractor/built/exitCodes.js:6:1)
    at ProtractorError (/home/travis/app/node_modules/protractor/built/exitCodes.js:11:9)
    at BrowserError (/home/travis/app/node_modules/protractor/built/exitCodes.js:52:9)
    at Direct.getNewDriver (/home/travis/app/node_modules/protractor/built/driverProviders/direct.js:68:27)
    at Runner.createBrowser (/home/travis/app/node_modules/protractor/built/runner.js:195:43)
    at q.then.then (/home/travis/app/node_modules/protractor/built/runner.js:339:29)
    at _fulfilled (/home/travis/app/node_modules/protractor/node_modules/q/q.js:834:54)
    at self.promiseDispatch.done (/home/travis/app/node_modules/protractor/node_modules/q/q.js:863:30)
    at Promise.promise.promiseDispatch (/home/travis/app/node_modules/protractor/node_modules/q/q.js:796:13)
    at /home/travis/app/node_modules/protractor/node_modules/q/q.js:556:49
[16:43:06] E/launcher - Process exited with error code 135
[16:43:06] Finished 'protractor' after 539 ms
[16:43:06] Starting 'itest'...
[16:43:06] Finished 'itest' after 12 μs
[16:43:06] E2E Tests failed
```

See https://travis-ci.org/hipster-labs/jhipster-travis-build/builds/283525249

___

- Please make sure the below checklist is followed for Pull Requests.

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [ ] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
[...]

PR #6615, opened 2017-10-29
Fix protractor tests for NPM

pascalgrimaud:
Related to @deepu105's https://github.com/jhipster/generator-jhipster/pull/6474#issuecomment-340284102

The reason why the build failed, is because you used NPM instead of Yarn. With that, we detected this issue. As the `node_modules` folder is managed differently between Yarn and NPM, the selenium folder is not at the same place.

It should fix the build in the [generator-jhipster-entity-audit](https://github.com/hipster-labs/generator-jhipster-entity-audit)

___


PierreBesson:
@deepu105 Well there are similar except that the goal of yarn is to prevent nested node_modules, so it makes sense.

deepu105:
ya i forgot that, wonder why it doesnt fail for Angular


Thanks & Regards,
Deepu

deepu105:
I guess we can make the protractor config for angularJS similar to angular
as well so that selenium jar need not be specified


Thanks & Regards,
Deepu

PR #6616, opened 2017-10-30
Remove duplicate security configuration

mraible:
When I was refactoring the OIDC / OAuth integration, I found I needed to have the security settings duplicated in the test directory's `application.yml`. This seemed strange, especially since the property values are the same as the ones in the main directory's `application.yml`. Now, it seems this is no longer necessary.

I tested this by making this modification in the generator-jhipster project and then running `npm link`. 

Then I generated three different projects with OAuth 2.0, Session Auth, and JWT. I ran `./mvnw test` in each to confirm everything worked. I did not test UAA, but I'm guessing the behavior should be the same.

### DIFF
--- a/generators/server/templates/src/test/resources/config/_application.yml
+++ b/generators/server/templates/src/test/resources/config/_application.yml
@@ -121,50 +121,6 @@ spring:
 liquibase:
     contexts: test
 
-<%_ if (authenticationType === 'oauth2') { _%>
-# ===================================================================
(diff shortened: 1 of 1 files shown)


mraible:
@danielpetisme Please review this PR. I want to make sure this isn't needed for microservices.

danielpetisme:
As former CI Administrator I'm tempted to trust the green badge 😉 . Count on me to some sanity checks tomorrow.

danielpetisme:
LGTM
Tests are green

Issue #6339, opened 2017-09-12
Dropdown Lists of OneToMany relation shows only the first 20 items


stieler-it:
##### **Overview of the issue**

Dropdown lists for OneToMany relations show only the first 20 options. (Probably this happens due to active pagination / endless scrolling).

##### **Motivation for or Use Case**

It is not possible to select every available option.

##### **Reproduce the error**

* Define two entities: Employee and Department. Employee has a OneToMany relationship to Department.
* Create 21 departments
* While creating an employee, open the department list

Expected behaviour: The list shows all 21 departments
Actual behaviour: The list shows only the first 20 departments

##### **Related issues**

n/a

##### **Suggest a Fix**

**Where the 20 comes from:**

`org.springframework.data.web.PageableHandlerMethodArgumentResolver` defines `DEFAULT_PAGE_REQUEST = new PageRequest(0, 20);` which is used when

1. The default is not overridden in the Resource
2. The REST client does not provide pagination options

**What we could do**:

1. Generate a `PageableDefault` Annotation into the `getAllDepartments()` method in the `Resource`
2. Provide paging options to `this.departmentService.query()` in `employee-dialog.component.ts`, e.g. `this.departmentService.query({size: 2000, sort: ['name,asc']})`

The page size seems to be limited to 2000 by `PageableHandlerMethodArgumentResolver`. However, I don't think showing so many entries in a dropdown anyway.

##### **JHipster Version(s)**

4.6.2

##### **JHipster configuration**

.yo-rc.json

```
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "XXX",
      "nativeLanguage": "de"
    },
    "jhipsterVersion": "4.6.2",
    "baseName": "babsapp",
    "packageName": "XXX",
    "packageFolder": "XXX",
    "serverPort": "8080",
    "authenticationType": "oauth2",
    "hibernateCache": "no",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "oracle",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "clientFramework": "angularX",
    "useSass": true,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [
      "protractor"
    ],
[...]

PR #6618, opened 2017-10-30
Auto-complete fields for the entities

gzsombor:
As discussed in #6339 - a possible solution is to create auto-complete fields for entity selection, instead of the current drop-down.

Currently it's not finished yet, the followings are missing:
- [ ] localization values
- [ ] it only works for entities, where filtering is enabled. It wouldn't be too hard to extend it to use ES if available, and fallback to load all the entities, and filter on the client side.
- [ ] currently it expects a string attribute named as 'name', it needs to be customized
- [ ] it's not added to any of the dialogs yet. This is where I've found the more complicated issue : these auto-complete fields are added into their respective 'entity' module, which needs to be imported in other entity modules, where these fields are needed, which will create a big spaghetti of entity inter-dependency, which is a bit scary. 
Probably we could separate the entity modules, for example the bank-account module, 
* _bank-account-base_ : bank-account.model.ts and bank-account.service.ts
* _bank-account-fields_ : the two new auto complete fields
* _bank-account-management_ : the dialogs, and listing components.

What do you think?

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/fields/base-autocomplete.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/fields/base-autocomplete.component.ts.ejs
@@ -1,0 +1,52 @@
+<%#
+ Copyright 2013-2018 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see http://www.jhipster.tech/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+
+export class BaseAutocompleteField<T> extends BaseFormField<T> {
+    searching = false;
+    searchFailed = false;
+    selected: T;
+    hideSearchingWhenUnsubscribed = new Observable(() => () => this.searching = false);
+
+    constructor(
+            config: NgbTypeaheadConfig
+        ) {
+            super();
+            config.showHint = true;
+        }
+
+    search = (text$: Observable<string>) =>
+        text$
+            .debounceTime(300)
+            .distinctUntilChanged()
+            .do(() => this.searching = true)
+            .switchMap((term) =>
+                this.<%= entityInstance %>Service.query({'<%= autocompleteField %>.contains': term})
+                    .do(() => this.searchFailed = false)
+                    .map((r: HttpResponse<I<%= entityAngularName %>[]>) => {
+                        this.searchFailed = (r.body.length === 0);
+                        return r.body;
+                    })
+                    .catch(() => {
+                        this.searchFailed = true;
+                        return Observable.of([]);
+                    }))
+                .do(() => this.searching = false)
+                .merge(this.hideSearchingWhenUnsubscribed)
+    
+    
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -405,7 +405,9 @@ const files = {
                 'shared/alert/alert.component.ts',
                 'shared/alert/alert-error.component.ts',
                 // dates
-                'shared/util/datepicker-adapter.ts'
+                'shared/util/datepicker-adapter.ts',
+                // fields
+                'shared/fields/base-form-field.component.ts'
             ]
         },
         {
--- a/generators/client/templates/angular/src/main/webapp/app/shared/fields/base-form-field.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/fields/base-form-field.component.ts.ejs
@@ -1,0 +1,62 @@
+<%#
+ Copyright 2013-2018 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see http://www.jhipster.tech/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+import { ControlValueAccessor } from '@angular/forms';
+
+export class BaseFormField<T> implements ControlValueAccessor {
+    private innerValue: T;
+    protected disabled: boolean;
+
+    private changed = new Array<(value: T) => void>();
+    private touched = new Array<() => void>();
+
+    public getValue(): T {
+        return this.innerValue;
+    }
+
+    protected setValue(value: T) {
+        if (this.innerValue !== value) {
+            this.innerValue = value;
+            this.changed.forEach((f) => f(value));
+        }
+    }
+
+    protected forceOnChange() {
+        this.changed.forEach((f) => f(this.innerValue));
+    }
+
+    touch() {
+        this.touched.forEach((f) => f());
+    }
+
+    writeValue(value: T) {
+        this.innerValue = value;
+    }
+
+    registerOnChange(fn: (value: T) => void) {
+        this.changed.push(fn);
+    }
+
+    registerOnTouched(fn: () => void) {
+        this.touched.push(fn);
+    }
+
+    setDisabledState(isDisabled: boolean) {
+        this.disabled = isDisabled;
+    }
+}
--- a/generators/client/templates/angular/src/main/webapp/app/shared/index.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/index.ts.ejs
@@ -20,6 +20,7 @@ export * from './constants/input.constants';
 export * from './alert/alert.component';
 export * from './alert/alert-error.component';
 export * from './auth/has-any-authority.directive';
+export * from './fields/base-form-field.component';
 <%_ if (enableTranslation) { _%>
 export * from './language/find-language-from-key.pipe';
 <%_ } _%>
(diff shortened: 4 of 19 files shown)


agoncal:
@gzsombor I've just pulled your PR and generated an application with it: I have a model with Address ---1-> Country (importing 250 countries from a CSV file) and would like to give it a try.

How can I use your selection component instead of the basic combobox ?

agoncal:
@gzsombor BTW I get the following warning:

`[1, 39]: The method "getValue" that you're trying to access does not exist in the class declaration.`

gzsombor:
Yes, there is a problem with exporting/importing the components, this needs to be fixed. That's why I've asked about opinions, but I haven't received anything, that was one of the reason, why it's not yet finished :) The other, is my lack of time :disappointed: 
The warning about getValue is known problem, it is because angular can't understand the concept of inheritance, so it hasn't see the method in the parent component. It would be nice, if we could fix it elegantly, but it's just ****script :wink:

agoncal:
@gzsombor My knowledge of Angular is limited so I feel I can't help much with code. But as I am very interested in this feature, I can help in reviewing/testing/giving you feedback. I would love to have this in JHipster. Ping my anytime if you want some feedback/testing

agoncal:
@gzsombor BTW in terms of naming, both components are called:

```
country-multi-selection.component
country-selection.component
```

What about renaming them with `selection` first so they all appear next to each other in alphabetical order ?

```
country-selection-multi.component
country-selection.component
```

agoncal:
@gzsombor I managed to make it work (importing/exporting the right module). But now, there are two things I can't figure out how it works:

#### How to affect the value

To get the selected value, I thought it was just a matter of overriding the `[(ngModel)]` like this:

`<spn-country-selection [(ngModel)]="organisation.address.country"></spn-country-selection>
`

But this doesn't work. Do you know how to get the value once selected ?

#### Doesn't display "Sorry, not found"

When in my CountrySelectionComponent I type "zzzz" I was expecting to see "Sorry, not found", but I don't see it.

agoncal:
@gzsombor when embedding different selectors in a same component, there are some id clashes. You should generate a different id for the component. Instead of `id="typeahead-http" ` you should add the name of the component `id="typeahead-country" `. This way you will end-up with different components:

```
<input id="typeahead-country" 
<input id="typeahead-language"
<input id="typeahead-organisation"
```

agoncal:
@gzsombor Would it be possible to have `required` as a parameter ? Because for the same selection, depending in which component it is used, the selection can be required or not. 

For the moment I just duplicated the code with 

```
<input id="typeahead-organisation" type="text" class="form-control"
       required
       #input
       [class.is-invalid]="searchFailed"
       ...
```

gzsombor:
Unfortunately, I can't claim that I have a huge Angular knowledge, I'm just a poor backend developer, who wants his UI autogenerated, because I don't like javascript. So that's why I write javascript. I feel the irony of the state.
This should work: `
<spn-country-selection [(ngModel)]="organisation.address.country"></spn-country-selection>
`

I haven't checked the required attribute, but that would be nice, as well :)

agoncal:
Gee, we are on the same boat then: two back-end guys wanting to have an auto-generated front-end ;o)

When I write `<spn-country-selection [(ngModel)]="organisation.address.country"></spn-country-selection>` I get the following error:

```
ERROR Error: If ngModel is used within a form tag, either the name attribute must be set or the form
control must be defined as 'standalone' in ngModelOptions.
```

So I add the `name`, but it doesn't seem to update my model either:

`<spn-country-selection name="country" [(ngModel)]="organisation.address.country"></spn-country-selection>`

gzsombor:
It's getting better, if you regenerate the app-ng2-default sample app, probably you will see it. :wink:

agoncal:
@gzsombor looks like you didn't take the _angularJSSuffix_ in account. In my application I have     "angularJSSuffix": "sponsor", this means that all components have a _sponsor_ suffix:

```
├── entities
│   ├── address-sponsor
│   ├── contact-sponsor
│   ├── tenant-sponsor
│   └── venue-sponsor
```
And in your module you import the component without the suffix

```
import { SponsorAddressModule } from '../address/address.module';
import { SponsorIndustrySectorModule } from '../industry-sector/industry-sector.module';
import { SponsorDocumentModule } from '../document/document.module';
```

it should be 

```
import { SponsorAddressSponsorModule } from '../address-sponsor/address-sponsor.module';
import { SponsorIndustrySectorSponsorModule } from '../industry-sector-sponsor/industry-sector-sponsor.module';
import { SponsorDocumentSponsorModule } from '../document-sponsor/document-sponsor.module';
```

agoncal:
@gzsombor about the "big spaghetti of entity inter-dependency", I've created an issue to talk about it. Tell me what you think: https://github.com/jhipster/generator-jhipster/issues/6693

gzsombor:
Interesting. Honestly, these suffixes looks insane :)  I haven't seen this feature yet. 
I will try to fix the other usability problems, before.

deepu105:
I still didnt find time to review and try this, probably will find some time next week

agoncal:
@gzsombor any progress on this ?

I've seen that test for `request-util` were introduced (https://github.com/jhipster/generator-jhipster/commit/f5c712e3f2f31fa40656258e7f4ce2c8e557ff04). Wonder if it affects the changes you did.

deepu105:
@gzsombor may be we can target this for v5 as well. I'll have a look when I have time

gzsombor:
I didn't want to progress further, until the module structure is sorted out. I try to catch up with the latest changes.

deepu105:
@gzsombor I could help with generalizing the component if you like

gzsombor:
Yes, the code sharing could be improved a bit for sure, the problems which I see:
* the html part is a bit different for the select and for the multi-select. A list of tags are rendered, which is a bit ugly, but for a first start it's ok, I think.
* I would like to provide customization options for the user - and it's simpler if most of the code is generated -
** Ability to add custom filters for the autocomplete, so they can write: <jhi-account-selection model="some.account" owner="currentUser.id" state="open"/>
** Ability to customize the rendering of the autocomplete, like : "Project Name<b>(Project Code)</b>"

And sure, there is no test coverage yet :( And even the old tests are braking.

deepu105:
Ok let me see how this can be generalized while meeting your requirements,
I think it's still possible as I have done something similar in React, so
it should definitely be possible in Angular as well. we can write tests
after that


Thanks & Regards,
Deepu

jdubois:
Lots of very interesting work here, but this has been stuck for more than 2 months - @gzsombor do you need help from my trainees? Or should we close it?

gzsombor:
Thanks for asking! I'm trying to keep up with the changes and it's not easy :) I think, the current version works, however, there is no javascript tests for it, and some of the old tests are broken because the 'select' -> 'autocomplete field' change. If someone could lend me a hand writing and fixing the unit tests ;-) that would be awesome.

jdubois:
Unfortunately I feel that this won't make it in the master branch - @gzsombor should we stop here? I'm sad about it, but as of today we have so many opened tickets and PRs, we need to remove work that is not going forward

gzsombor:
I hoped that someone could help in fixing the test cases, which I'm not familiar at all, and locally it seems that everything is green.

deepu105:
@gzsombor I really wanted to take a look at this, but this is quite big and i'm hardly finding enough time. I'll try my best coming weekened. In the meantime @agaspardcilia @chrisdns do any of you guys have time to look at this?

generators/client/templates/angular/src/main/webapp/app/shared/fields/base-autocomplete.component.ts.ejs
```diff
@@ -0,0 +1,52 @@
...
+            config.showHint = true;
+        }
+
+    search = (text$: Observable<string>) =>
+        text$
+            .debounceTime(300)
+            .distinctUntilChanged()
+            .do(() => this.searching = true)
+            .switchMap((term) =>
+                this.<%= entityInstance %>Service.query({'<%= autocompleteField %>.contains': term})
```
deepu105: can this service be generalized and moved to the lib, the only dependency is entitySevice, cant we pass it somehow?

deepu105:
Also as mentioned earlier, I don't think we should merge this without some generated tests, else this will bring down the overall coverage a lot and we really spent a lot of effort to bring it up and make sonar happy

deepu105:
I'm assigning this to me hoping i'll find some time soon. @jdubois hope you dont mind this being open for some more time

jdubois:
Of course I don't mind, but I'm just worried we go too far and take work that we cannot do or maintain by lack of time.
It's not against the idea, it's just that I'm afraid we don't have the bandwidth: we already have so many opened tickets, we are reaching a point were we can't do much more unless we grow the team (which is a though problem at the moment)

agaspardcilia:
We'll take a look and see what we can do.

agaspardcilia:
There's a problem with the completion, in the code it's supposed to do a simple search query to find entities that contain the input. It's calling `http://localhost:9000/api/projects?name.contains=d` but the server seems to always return the complete list of entities and to ignore the query. Is it not implemented yet or some Spring black magic that is not working ?

ruddell:
@agaspardcilia Did you enable filtering when generating your test entities? This would only work when that's enabled

agaspardcilia:
@ruddell Yep, that was the problem, thanks.

jdubois:
This really look complicated and we have tons of other things to do -> again, shouldn't we stop here? I understand there's been a lot of work and people want this, but at some point we need to accept not being able to do everything

deepu105:
@jdubois I agree. In its current form it only adds lot of code and maintenance complexity and in the end this might be used only by some people

gzsombor:
It also removes a lot's of code and complexity in the editor. The git line count change is a bit misleading, because I've created a couple of new files, and for this I had to add the license header 😉 

From the travis tests, it seems, that there is one problem with an import, in a custom prefix-suffix situation, and Protractor fails in a couple of places. Unfortunately, I can't reproduce it, when I locally try to run the script it doesn't fail. Maybe, protractor doesn't used - I'm a bit confused.

deepu105:
@gzsombor @jdubois I don't think I'll have time to work on this, I know I promised, but my plate is so full these days and this one would actually require a good amount of time. IMO, I still believe this adds too much code for end users and it's not complete anyway(no tests) and most of them might not want this. For most use cases a generic Typeahead solution available in the ecosystem might be enough and for others, they will always build their own solution. What I can suggest though is to move this into a Blueprint or Module to see the response before adding to the main generator

jdubois:
As this has been opened for 1 year, at some point we need to take a decision... same issue for me, we have so many opened tickets, it's just not possible for me to review.

I'm really sorry about this.

Grant-Lay:
@deepu105 -any chance you could share the react code for the typeahead somewhere?  I'm struggling trying to get this  [typeahead](http://ericgio.github.io/react-bootstrap-typeahead/) working within a jhipster update page.

shorstki:
Is there anything that solves the issue this PR was to fix? I'm checking for a fix for https://github.com/jhipster/generator-jhipster/issues/6339 but there is nothing, also here... edit pages after 3 years are still not usable if we've more than 20 items?

PR #6619, opened 2017-10-31
Angular-oidc support

danielpetisme:
The current implementation of JHipster OIDC delegates the authentication to the backend application. The aim of this PR is to let the client do the authentication and pass the Bearer to the token removing de facto the transitive dependency (client -> backend -> IdP). I used [angular-oauth2-oidc](https://github.com/manfredsteyer/angular-oauth2-oidc) for this PR

The support should work either on microservices and monoliths. 

Once of the motivation of this PR is to help to decouple the front and backend in order to be able to develop both components separately. 

- [x] [angular-oidc]() installation
- [x] Configuration for authentication
- [x] Map IdP user with Jhipster representation
- [x] Update `oauth2` `clientId` and `clientSecret` to use `internal` only in backend
- [x] Make Swagger-UI works
- [ ] Make the IdP URL configurable like `SERVER_API_URL`
- [ ] Provide the new Keycloak configuration
- [x] Test with mciroservices
- [ ] Test with monoliths  

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -35,6 +35,9 @@
     "@angular/platform-browser-dynamic": "4.3.2",
     "@angular/router": "4.3.2",
     "@ng-bootstrap/ng-bootstrap": "1.0.0-beta.5",
+    <%_ if (authenticationType === 'oauth2') { _%>
+    "angular-oauth2-oidc": "^2.1.8",
+    <%_ } _%>
     "bootstrap": "4.0.0-beta",
     "core-js": "2.4.1",
     "font-awesome": "4.7.0",
--- a/generators/client/templates/angular/src/main/webapp/app/_app-routing.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_app-routing.module.ts
@@ -27,7 +27,8 @@ const LAYOUT_ROUTES = [
 
 @NgModule({
     imports: [
-        RouterModule.forRoot(LAYOUT_ROUTES, { useHash: true })
+        RouterModule.forRoot(LAYOUT_ROUTES, { useHash: true <%_ if (authenticationType === 'oauth2') { _%>, initialNavigation: false<%_ } _%> })
+
     ],
     exports: [
         RouterModule
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/config/_auth.config.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/config/_auth.config.ts
@@ -1,0 +1,34 @@
+<%#
+Copyright 2013-2017 the original author or authors from the JHipster project.
+
+This file is part of the JHipster project, see http://www.jhipster.tech/
+for more information.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-%>
+import { AuthConfig } from 'angular-oauth2-oidc';
+
+export const authConfig: AuthConfig = {
+
+    // Url of the Identity Provider
+    issuer: 'http://localhost:9080/auth/realms/jhipster',
+
+    // URL of the SPA to redirect the user to after login
+    redirectUri: window.location.origin + '/index.html',
+
+    // The SPA's id. The SPA is registerd with this id at the auth-server
+    clientId: 'web_app',
+
+    // set the scope for the permissions the client should request
+    scope: 'openid profile email',
+};
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -150,6 +150,14 @@ const files = {
             templates: [
                 'blocks/interceptor/_auth.interceptor.ts'
             ]
+        },
+        {
+            condition: generator => generator.authenticationType === 'oauth2',
+            path: ANGULAR_DIR,
+            templates: [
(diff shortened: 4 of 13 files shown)


danielpetisme:
One of the benefits is to update the `clientId` and `clientSecret` of the gateway to use the `internal` client. That means the backend is not doing any authentication anymore it expects a token and validates it against the IdP.

Behind the since, the backend user details extraction relies on a `OAuth2AuthenticationDetails` whichi is still valid since it uses the token to get the information. That means the `/acccount` and `/authenticate` endpoints are not touched => User sync use as is.

generators/client/templates/angular/src/main/webapp/app/blocks/config/_auth.config.ts
```diff
@@ -0,0 +1,34 @@
...
+-%>
+import { AuthConfig } from 'angular-oauth2-oidc';
+
+export const authConfig: AuthConfig = {
+
+    // Url of the Identity Provider
+    issuer: 'http://localhost:9080/auth/realms/jhipster',
+
+    // URL of the SPA to redirect the user to after login
+    redirectUri: window.location.origin + '/index.html',
```
mraible: Can we make this dynamic? If the user hits a URL that requires authentication, they should be redirected back to it after logging in.

mraible:
@danielpetisme - thanks for taking the time to write this! 👍

I'm a big fan of angular-oauth2-oidc, it's a great library! As a warning, when I first started using it in March, it was unmaintained for months. Then the author came back to it and has been great at support ever since.

One thing I like about our current (server-side) implementation is you can easily switch IdP's by simply setting environment variables. If we switch to client-side, we'll have to change variables in a TypeScript file and recompile. It might not matter to the end user, but it's awesome for demos and deploying to the cloud.

Also, having Spring Security handle the redirect-to-IdP makes it easy for the various clients. Right now, we only have Angular, but @deepu105 is working on React and I hope to get an Ionic Module working soon. For Ionic + OAuth, I can use Cordova's InApp browser and send a request to `http://localhost:8080/login`. I'm pretty sure angular-oauth2-oidc doesn't work with Ionic because it can't launch a browser for the redirect. However, I could construct the URL manually. For example:

```ts
buildOAuthUrl(state, nonce): string {
  return this.oauthService.issuer + '/v1/authorize?' +
      'client_id=' + this.oauthService.clientId + '&' +
      'redirect_uri=' + this.oauthService.redirectUri + '&' +
      'response_type=id_token%20token&' +
      'scope=' + encodeURI(this.oauthService.scope) + '&' +
      'state=' + state + '&nonce=' + nonce;
}
```

For more information, see [this blog post](https://developer.okta.com/blog/2017/08/22/build-an-ionic-app-with-user-authentication). 

Yes, angular-oauth2-oidc is great, but I'm not sure we should switch to it. What are the advantages of using it over our current implementation? Does using it remove/add any features?

generators/client/templates/angular/_package.json
```diff
@@ -35,6 +35,9 @@
     "@angular/platform-browser-dynamic": "4.3.2",
     "@angular/router": "4.3.2",
     "@ng-bootstrap/ng-bootstrap": "1.0.0-beta.5",
+    <%_ if (authenticationType === 'oauth2') { _%>
+    "angular-oauth2-oidc": "^2.1.8",
```
deepu105: we use fixed versions

generators/client/templates/angular/src/main/webapp/app/_app-routing.module.ts
```diff
@@ -27,7 +27,8 @@ const LAYOUT_ROUTES = [
 
 @NgModule({
     imports: [
-        RouterModule.forRoot(LAYOUT_ROUTES, { useHash: true })
+        RouterModule.forRoot(LAYOUT_ROUTES, { useHash: true <%_ if (authenticationType === 'oauth2') { _%>, initialNavigation: false<%_ } _%> })
```
deepu105: the `<%_ _%>` templates are used for newlines for inline use `<% %>`

mraible:
@danielpetisme It looks like you'll need to fix the Protractor tests so they work with this new setup. I made them work with Keycloak/Okta when OAuth is selected. You'll probably need to modify Gatling tests as well since I made those work too.

danielpetisme:
TL;DR
If back and front live together => server side
If back and front are splited and live separately => client side

I tried to dig a little bit deeper the client  vs server side flow. First of all, the question is wrong, the right question is Implicit vs Authorization flow.
In a nutshell, it depends on how we do consider JHipster. If we consider it more as a server-side application (where the front is stronly coupled with the backend). Then the Autorization flow (ie. server side application) is the one to choose. 
Now if we consider JHipster as an a decoupled API + Js client application living their life seperately, then the Implicit flow is the one to choose. The API never sees the access_token and only trust request. It's also the one to choose for mobile applications for instance.

For technical details, the implicit flow (client side) looks easier (less network communication).

Further details:
https://stackoverflow.com/questions/7522831/what-is-the-purpose-of-the-implicit-grant-authorization-type-in-oauth-2
Have look to [this answer](https://stackoverflow.com/a/7688532) and [this one](https://stackoverflow.com/a/25644196)

This is the results of the survey
capture d ecran 2017-11-01 a 14 59 59 2

A personal thought, the survey represents the profile of JHipster users (more backend than frontend users).

So maybe for monolith application it doesn't make sense to have a client side flow since IMHO the developers manage the back+front stack as one.
For microservices, is more tricky, personally I split back and front because I'm API-first oriented (and that's the purpose of having an an api gateway also). The clients are multiple so client side makes sense for me.

Thoughts guys?

Note: Until we're clear on the future of this PR I won't spend time right now on it.

deepu105:
@danielpetisme while I agree that some of your points are valid lets not over complicate this at the moment. OIDC is still new and we need to see how people are using it and get feedback first then we can think of improvements. Frankly, we have so much on the pipeline that is more pressing so I suggest to drop this for now and reconsider once OIDC support is mature

danielpetisme:
:+1:

deepu105:
closing this as the server side solution is good enough for now, we can consider this in the future if there is enough need

PR #6622, opened 2017-10-31
AWS also supports mariadb.

java4africa:
I will update the docs in the next days - just to mention that mariadb is supported. The aws generator also needs some better error checks, but that's another issue.
I need to go back and try docker-compose - I think this would be simpler for my user-case.


jdubois:
Thanks & congrats for your first contribution!

PR #6642, opened 2017-11-03
Don't put two database drivers in a single package

erikkemperman:
When users select different databases for development and production (as when following the default choices) they will end up with two drivers in their package. This PR will make sure that they only get the appropriate one for the profile they're packaging.

It slightly increases the complexity of the generator(*) -- I wanted to keep the dependency in the general section in case dev and prod are the same database -- will have no effect on the complexity on the generated maven/gradle build, and will make the packaged artifact simpler.

Hope this makes sense!

(*) Alternatively, if you prefer, I could just put the driver in the profiles even when they're the same for dev and prod, which would not increase the generator complexity at all.

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -215,9 +215,7 @@
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
-            <%_ if (devDatabaseType !== 'h2Disk' && devDatabaseType !== 'h2Memory') { _%>
             <scope>test</scope>
-            <%_ } _%>
         </dependency>
         <%_ if (hibernateCache === 'hazelcast' || clusteredHttpSession === 'hazelcast' || applicationType === 'gateway') { _%>
         <dependency>
@@ -1192,6 +1190,12 @@
                     <artifactId>spring-boot-devtools</artifactId>
                     <optional>true</optional>
                 </dependency>
+                <%_ if (devDatabaseType === 'h2Disk' || devDatabaseType === 'h2Memory') { _%>
+                <dependency>
+                    <groupId>com.h2database</groupId>
+                    <artifactId>h2</artifactId>
+                </dependency>
+                <%_ } _%>
             </dependencies>
             <build>
                 <plugins>
--- a/generators/server/templates/_build.gradle
+++ b/generators/server/templates/_build.gradle
@@ -381,9 +381,6 @@ dependencies {
     <%_ if (devDatabaseType === 'mariadb' || prodDatabaseType === 'mariadb') { _%>
     compile "org.mariadb.jdbc:mariadb-java-client"
     <%_ } _%>
-    <%_ if (devDatabaseType === 'h2Disk' || devDatabaseType === 'h2Memory') { _%>
-    compile "com.h2database:h2"
-    <%_ } _%>
     <%_ if (devDatabaseType === 'mssql' || prodDatabaseType === 'mssql') { _%>
     compile "com.microsoft.sqlserver:mssql-jdbc"
     compile "com.github.sabomichal:liquibase-mssql"
--- a/generators/server/templates/gradle/_profile_dev.gradle
+++ b/generators/server/templates/gradle/_profile_dev.gradle
@@ -32,6 +32,9 @@ ext {
 
 dependencies {
     compile "org.springframework.boot:spring-boot-devtools"
+    <%_ if (devDatabaseType === 'h2Disk' || devDatabaseType === 'h2Memory') { _%>
+    compile "com.h2database:h2"
+    <%_ } _%>
 }
 
 def profiles = 'dev'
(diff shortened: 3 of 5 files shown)


generators/server/templates/_pom.xml
```diff
@@ -1357,6 +1357,44 @@
...
+                <%_ } _%>
+                <%_ if (devDatabaseType === 'oracle' && prodDatabaseType !== 'oracle') { _%>
+                <!-- more information at https://blogs.oracle.com/dev2dev/entry/how_to_get_oracle_jdbc -->
+                <dependency>
+                    <groupId>com.oracle.jdbc</groupId>
+                    <artifactId>ojdbc7</artifactId>
+                    <version>${oracle.version}</version>
+                </dependency>
+                <%_ } _%>
+                <%_ if (devDatabaseType === 'postgresql' && prodDatabaseType !== 'postgresql') { _%>
```
deepu105: h2 seems missing

erikkemperman: Here it is https://github.com/erikkemperman/generator-jhipster/blob/881f02b653bdd1124510c149801e2711bb2952e9/generators/server/templates/_pom.xml#L315

I didn't make the same split for H2, because it is used with `test` scope in case it isn't the dev database. But of course I could do something similar.

erikkemperman:
@deepu105 Thanks for taking a look! Yeah, it does make the templates slightly more complex (although I could reduce that by about 1/3 if you want).

I guess the question is a trade-off between some convenience for JH generator developers versus some bloat in JS users' artifacts.

PierreBesson:
Personally I like this way to do it and it optimizes the size of the jar. So +1 for me.

deepu105:
@erikkemperman seems others like the idea in that case ok from me as well

erikkemperman:
Cool! I was trying to also split H2 across the profiles, but ran into a bit of a problem -- `DatabaseConfiguration.java` is importing things from H2 and so it needs the driver on the classpath, even though it never actually uses it because the method `h2TCPServer` is annotated to run only in dev profile. Now trying to figure that out, hopefully I can amend this PR shortly.

gmarziou:
What's the gain in size?

erikkemperman:
@gmarziou That depends on the dev database that would now be excluded in production WAR. For H2 (which is not yet in this PR but I'm almost done with the workaround for the problem mentioned above) the saving is ~2MB.

jdubois:
I'm sorry but I don't understand this: with the CLI (or with JHipster Online), if you select one database in prod, you should use the same one in dev. For example, you can't have MySQL in prod and PostgreSQL in dev. This is because this doesn't make any sense.
The only other dev database you can select is H2, as it can be used in dev mode and in tests.

erikkemperman:
Ah! Just looking at the `<% if (...) %>` tags in the generator's `_pom.xml` it looked like any combination was possible. But you are of course right, CLI-produced configurations will only ever have H2 as a second database.

I guess the logic in the POM (and the gradle build) is conservative and wants to cover even cases when users edit their `.yo-rc.json`? If not it could be somewhat simpler.

Related question, is H2 still considered not ready for production? I have been using it in a small-medium application without any problems for a pretty long time now. It would be great if the CLI offered the option, in my opinion, perhaps with a warning?

jdubois:
Indeed, people modifying their `.yo-rc.json` configuration can mess up stuff. We should have a test to prevent that, when the generator starts it should validate the configuration.
I wouldn't use H2 in production, it's not meant to be a production database. But I know very big projects using it in development, so if you have a small non-critical server it could be good enough for you. Still, I would not propose that option.

erikkemperman:
All right, I've modified the PR -- it no longer assumes all combinations of dev/prod databases are possible -- and therefore it complicates the templates less than before.

In fact, it now just prevents H2 from ending up in the production packages. There was a slight issue because even in production profile there was code that required H2 on the classpath (see above). I have worked around that by instead loading these classes dynamically at runtime...

jdubois:
The reflection part is a bit ugly -> I'm pretty sure we could migrate this to https://github.com/jhipster/jhipster so it's cleaner, what do you think?

erikkemperman:
Yeah, agreed, kind of ugly. You mean like a general class loading / method invoking utility?

jdubois:
@erikkemperman no, those blocks of code could just be cut/pasted there - there is no need to generate those, it's only to help for dev

jdubois:
Anyway let's merge this, we can still refactor later

erikkemperman:
Ah, I thought you intended deduplication of similar code here. Also, one of these blocks is conditional `if (devDatabaseType === ...)` which the server lib doesn’t know about...

Anyway, thanks for review and merge and indeed it can always be refactored!

erikkemperman:
Actually thinking about this, I do think it ought to be a general method-invoking utility, both of these blocks are generator-specific conditioned. Would you consider something like that in the server lib?

deepu105:
Yes let's move the ugly part to server lib. Its not nice to generate these
in users code

erikkemperman:
Agreed :-)
https://github.com/jhipster/jhipster/pull/33

PR #6656, opened 2017-11-05
Improve logging of exceptions that get obfuscated

ramzimaalej:
I have noticed in some situation when an error occurs in the backend, the root exception got obfuscated making the debugging hard. While I was able to find the root cause by looking at newrelic, this is not an option for devs thus adding a statement to logs the exceptions.


cbornet:
See https://github.com/zalando/problem-spring-web/blob/master/src/main/java/org/zalando/problem/spring/web/advice/AdviceTrait.java#L234 . The log function swallows client 4xx errons. To change the behavior it would be better to override this function. Now I'm not sure it' s a good think to log client stack traces at the WARN level. Standard Spring Boot error management doesn't do it either.

ramzimaalej:
Looks like we can discard this PR. Thanks for the heads up!

Issue #6648, opened 2017-11-03
Use simple cache instead of Guava's?


erikkemperman:
I am packaging my app without springfox, in the production profile, and would now like to get rid of the large guava dependency.

However, it's being used by `PersistentTokenRememberMeServices.java` (for session authentication) and `OAuth2AuthenticationService.java` (for gateway/uaa authentication).

But it seems to me that in those places, it is actually a bit heavy-handed to use guava's Cache class -- which is specifically designed to be safe for use by multiple concurrent threads, but here we are bluntly synchronizing on a global lock around it!

https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/security/_PersistentTokenRememberMeServices.java#L118
https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/security/oauth2/_OAuth2AuthenticationService.java#L151

Would you consider a PR which replaces this use of guava's cache by a simple custom implementation? It's basically just a map with a maintenance thread. I have it mostly done already, and it works like a charm.

But I figured I'd check first, before I spend too much time polishing it and adding unit tests :-)

- [x] Checking this box is mandatory (this is just to show you read everything)

jdubois:
- yes you are totally correct, and that's a very good idea
- can we use Spring Cache's abstraction, to use some simple HashMap? I haven't looked too much, but I guess it's doable - one of the reasons is that I would like to be able to use a Memcache implementation in the future, so those caches could easily be distributed

erikkemperman:
Would it be all right to do it in two steps? First a simple map (have that working already) and then go to spring cache?

gmarziou:
Spring Cache abstraction uses a [ConcurrentHashMap by default](https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#cache-store-configuration-jdk)

jdubois:
So yes I would rather use Spring Cache

erikkemperman:
Well, the concurrent map doesn't have any benefits if we call it from within a globally locked section but sure, I'll see if I can make it work with spring cache!

ramzimaalej:
I have a question about the globally locked section. How would it solve the problem of receiving two requests when we have 2 instances of the app running. Do we really need the global lock?

atomfrede:
What about using [caffeine](https://github.com/ben-manes/caffeine) isntead of guava cache? It supports jcache and spring cache supports caffeine out of box.

[...]

PR #6666, opened 2017-11-08
Use simple cache and small TLD library (instead of Guava)

erikkemperman:
Ref. https://github.com/jhipster/generator-jhipster/issues/6648

Unfortunately I haven't found time to make this work with Spring caches, or Caffeine. But I'm not going to have a lot of time the next couple of days, so I wanted to put this out there for now.

If anyone wants to take a look at adding onto this, that would be great. Otherwise, this is also a fully functional (although pretty minimal) alternative to what was here before...

Note: I've added a dependency on a small public-suffix-list library to replace another little call we did to a Guava component. If this, or something like this, were adopted that would have to go into jhipster-dependencies.

### DIFF
--- a/generators/server/templates/src/main/java/package/security/_PersistentTokenRememberMeServices.java
+++ b/generators/server/templates/src/main/java/package/security/_PersistentTokenRememberMeServices.java
@@ -26,8 +26,6 @@ import <%=packageName%>.service.util.RandomUtil;
 <%_ if (databaseType === 'cassandra') { _%>
 import com.datastax.driver.core.exceptions.DriverException;
 <%_ } _%>
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 
 import io.github.jhipster.config.JHipsterProperties;
 
@@ -50,9 +48,11 @@ import java.time.LocalDate;
 <%_ if (databaseType === 'cassandra') { _%>
 import java.time.temporal.ChronoUnit;
 <%_ } _%>
-import java.util.concurrent.TimeUnit;
 import java.util.Arrays;
+<%_ if (databaseType === 'cassandra') { _%>
 import java.util.Date;
+<%_ } _%>
+
 
 /**
  * Custom implementation of Spring Security's RememberMeServices.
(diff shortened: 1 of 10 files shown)


erikkemperman:
I have unit tests for the new utility class, but they're on my machine at home... Will push them when I get there.

PS - That's cool, I got number #6666 -- do I get points for that?

erikkemperman:
Once again a randomly failing Travis test -- it [passed](https://travis-ci.org/erikkemperman/generator-jhipster/builds/299089362?utm_source=github_status&utm_medium=notification) before :-)

erikkemperman:
Just pushed those unit tests I mentioned, for what it's worth.

generators/server/templates/src/main/java/package/security/_TokenCache.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.LinkedHashMap;
+
+/**
+ * Simple time-limited cache for login tokens, necessary to avoid concurrent
+ * requests invalidating one another. It uses a {@link java.util.LinkedHashMap}
+ * to keep the tokens in order of expiration, and a dedicated low-priority purger
+ * thread which periodically removes expired tokens from the map.
+ */
+public class TokenCache<T> {
+<%_ /* TODO Should this class be in the server lib instead? */ _%>
```
deepu105: Yes I think this should go into server lib

generators/server/templates/src/main/java/package/security/_PersistentTokenRememberMeServices.java
```diff
@@ -92,11 +92,11 @@
 
     private static final int TOKEN_VALIDITY_SECONDS = 60 * 60 * 24 * TOKEN_VALIDITY_DAYS;
 
-    private static final int UPGRADED_TOKEN_VALIDITY_SECONDS = 5;
+    private static final long UPGRADED_TOKEN_VALIDITY_MILLIS = 5_000l;
```
deepu105: This number isn't big enough to add the underscore, it will just confuse novice users. I would rather just use `5000L`

erikkemperman: OK, sure, that makes sense.

generators/server/templates/src/main/java/package/security/_TokenCache.java
erikkemperman: I guess it depends on whether or not anyone is able to look into using spring caches instead of my simple custom class anytime soon -- because then this class will not be needed at all. But if not, and it is determined that something like this PR could be merged as is in the meantime, then yes it should definitely be in the server lib!

deepu105:
@erikkemperman the spring cache impl PR has been merged FYI

erikkemperman:
@deepu105 OK, thanks! I will look into it when I find some time... But thus far my feeling is a little custom implementation might still be warranted in this particular case, because the general implementations don't take advantage of some specific scenario constraints and therefore end up performing significantly worse.

However I would then prefer for the ugly bits to end up in the server lib, so I might wait for the mono-repo thing first, which would make this change a lot cleaner.

deepu105:
@erikkemperman closing this as we agreed that you would do this in the library

PR #6678, opened 2017-11-12
Add "unique" field validation

taurus227:


generators/entity/prompts.js
```diff
@@ -1130,6 +1146,9 @@ function logFieldsAndRelationships() {
                 if (field.fieldValidateRules.indexOf('required') !== -1) {
                     validationDetails = 'required ';
                 }
+                if (field.fieldValidateRules.indexOf('unique') !== -1) {
```
deepu105: this part changed slightly in master please update

generators/entity/templates/server/src/main/resources/config/liquibase/changelog/_added_entity.xml
```diff
@@ -55,6 +55,10 @@
             if (fields[idx].fieldValidate === true && fields[idx].fieldValidateRules.indexOf('required') !== -1) {
                 nullable = false;
             }
+            let unique = false
+            if (fields[idx].fieldValidate === true && fields[idx].fieldValidateRules.indexOf('unique') !== -1) {
```
deepu105: use `fields[idx].fieldValidateRules.includes('unique')` instaed of `fields[idx].fieldValidateRules.indexOf('unique') !== -1` I'm changing other usages

generators/entity/templates/server/src/main/resources/config/liquibase/changelog/_added_entity.xml
```diff
@@ -98,7 +102,7 @@
             }
             %>
             <column name="<%=columnName %>" type="<%=columnType %>">
-                <constraints nullable="<%= nullable %>" />
+                <constraints nullable="<%= nullable %>" unique="<%= unique %>" />
```
deepu105: add the unique constraint only when it is true else it will look busy

cbornet:
I think there is a problem in tests if an entity has 2 required relationships to an entity that has a unique field (already happened with User's login)

jdubois:
I'm sorry but I have no idea what this is about: there is no ticket, no description... There's also a review pending and some merge conflict -> sorry but I'm just totally lost here.

I'm closing this as it doesn't follow our process: please follow the guidelines and open up a ticket explaining what you are trying to solve.

deepu105:
@jdubois i think there was a ticket in Jhipster core or something.
@taurus227 please let us know if you still want to pursue this

taurus227:
Sorry, yes, I still want to pursue this, but I've been busy with some other stuff. I will update my pull request this week. The ticket is here: [Adding commands for unique constraints](https://github.com/jhipster/jhipster-core/issues/18)

PR #6679, opened 2017-11-13
6668 -  fix the logstash-appender stopping working issues

lmbai:
In the SpringCloudConfig, Spring will refresh its configuration even the app is already running.  Spring will call the `LoggerContext` to refresh all the logger config. which will `StopAndReset` the log config. it will stop all the logappender and remove all the `LoggerContextListener`.   After that, `LoggerContext `will load the new configuration based on the configuration file (logback-spring.xml) . then, Spring will call the configuration class ...... , but here, the `LoggingConfiguration` bean is a singleton and already been created. 
In this case, the above `LoggingConfiguration`  won't be created again (even the `logstash-appender` has been stopped).

We add the `@RefreshScope` here to force Spring to recreate this bean when there is configuration refresh

### DIFF
--- a/generators/server/templates/src/main/java/package/config/_LoggingConfiguration.java
+++ b/generators/server/templates/src/main/java/package/config/_LoggingConfiguration.java
@@ -53,11 +53,12 @@ import org.springframework.cloud.context.config.annotation.RefreshScope;
 @Configuration
 <%_ if (serviceDiscoveryType === "eureka") { _%>
 @ConditionalOnProperty("eureka.client.enabled")
+@RefreshScope
 <%_ } _%>
 <%_ if (serviceDiscoveryType === "consul") { _%>
 @ConditionalOnConsulEnabled
-<%_ } _%>
 @RefreshScope
+<%_ } _%>
 public class LoggingConfiguration {
 
     private static final String LOGSTASH_APPENDER_NAME = "LOGSTASH";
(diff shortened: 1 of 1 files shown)


gmarziou:
Travis build failed at compilation due to `@RefreshScope` annotation: 
https://travis-ci.org/jhipster/generator-jhipster/jobs/301245081

PierreBesson:
+1 It makes sense to have this `@RefreshScope` here so you can hot update the logstash appender config.

generators/server/templates/src/main/java/package/config/_LoggingConfiguration.java
```diff
@@ -48,13 +48,16 @@
 import org.springframework.cloud.consul.serviceregistry.ConsulRegistration;
 <%_ } _%>
 import org.springframework.context.annotation.Configuration;
+import org.springframework.cloud.context.config.annotation.RefreshScope;
```
PierreBesson: Please wrap this import statement with an "if serviceDiscovery enabled" template conditional so that it doesn't pollute the imports for application types who don't have service discovery.

lmbai: Hi @PierreBesson , thanks for the comment. i am not that familiar this the generator script.  Try my best to slove this. and you a welcom to edit it too.

PierreBesson:
Thanks @lmbai and congrats on your first contribution !

PR #6702, opened 2017-11-19
Entity refactoring

deepu105:


generators/entity-server/templates/src/main/java/package/domain/_Entity.java
```diff
@@ -0,0 +1,437 @@
...
+<%_ } if (databaseType === 'mongodb' || databaseType === 'couchbase') { %>
+    @Id<% if (databaseType === 'couchbase') { %>
+    @GeneratedValue(strategy = UNIQUE, delimiter = ID_DELIMITER)<% } %>
+    private String id;
+<%_ } if (databaseType === 'cassandra') { %>
+    @PartitionKey
+    private UUID id;
+<%_ } _%>
+
+<%_ for (idx in fields) {
```
cbornet: We should replace that by `for (field of fields)`

cbornet: I haven't tried forEach but if it works it's even better !

PR #6717, opened 2017-11-22
support login with both email and username

### LINKED ISSUE #6701: Feature Request: Login also with E-Mail address
A user should be able to also login with his email address rather than his login-name

- users tend to forget their username on webapplication, but often are able to remember their email address they used when registering. 
- more and more webapplication switch to only having the email-address as the main information for login
- there is a "nasty" Firefox autocompletion "feature", which autocompletes the users email address into the login form of a JHipster webapp. This is due to the fact, that Firefox remembers the textfield before a password field (which is the email on JHipster registration form).

On my JHipster based App I already had several users stumbling over this issue.

##### **Suggest a Fix**
I do have a working version locally, but am not sure I'm able to provide a proper fix on the generator sources. This might also influence other types of generators (like MongoDB) which I'm not capable of testing.

The changes are in a single class and quite easy.
Please look into DomainUserDetailsService. It's the central class of Spring to resolve a loginname into a UsersDetail object.
First we need to try to find a user also by his email.

```
        Optional<User> userFromDatabase = userRepository.findOneWithAuthoritiesByLogin(lowercaseLogin);
       // try also to retrieve by email
        if (!userFromDatabase.isPresent()) {
            userFromDatabase = userRepository.findOneWithAuthoritiesByEmail(login);
        }
```
As 'lowercaseLogin' might contain the email-address its important to not use it for creating 
org.springframework.security.core.userdetails.User.
Rather we use the login from the fetched userFormDatabase

```
            return new org.springframework.security.core.userdetails.User(
                // use the official login, rather than 'lowercaseLogin'
                user.getLogin(),
                user.getPassword(),
                grantedAuthorities);
```

##### **JHipster Version(s)**
the current version as of writing the feature request: 4.10.2

Currently I'm running in a monolith SQL based scenario.

thomastardy:
Fix #6701

### DIFF
--- a/generators/languages/templates/src/main/webapp/i18n/en/_global.json
+++ b/generators/languages/templates/src/main/webapp/i18n/en/_global.json
@@ -143,7 +143,8 @@
             "maxbytes": "This field cannot be more than {{ max }} bytes.",
             "pattern": "This field should follow pattern for {{ pattern }}.",
             "number": "This field should be a number.",
-            "datetimelocal": "This field should be a date and time."
+            "datetimelocal": "This field should be a date and time.",
+            "patternLettersAndDigits": "This field can only contain letters and digits."
         }
     },
     "error": {
--- a/generators/server/templates/src/main/java/package/repository/_UserRepository.java
+++ b/generators/server/templates/src/main/java/package/repository/_UserRepository.java
@@ -82,8 +82,8 @@ import static <%=packageName%>.config.Constants.ID_DELIMITER;
 public interface UserRepository extends <% if (databaseType === 'sql') { %>JpaRepository<User, Long><% } %><% if (databaseType === 'mongodb') { %>MongoRepository<User, String><% } %><% if (databaseType === 'couchbase') { %>N1qlCouchbaseRepository<User, String><% } %> {
     <%_ if (['ehcache', 'hazelcast', 'infinispan'].includes(cacheProvider) || clusteredHttpSession === 'hazelcast' || applicationType === 'gateway') { _%>
 
-    public static final String USERS_BY_LOGIN_CACHE = "usersByLogin";
-    public static final String USERS_BY_EMAIL_CACHE = "usersByEmain";
+    String USERS_BY_LOGIN_CACHE = "usersByLogin";
+    String USERS_BY_EMAIL_CACHE = "usersByEmain";
     <%_ } _%>
 <%_ if (authenticationType !== 'oauth2') { _%>
 
@@ -119,7 +119,7 @@ public interface UserRepository extends <% if (databaseType === 'sql') { %>JpaRe
     <%_ if (['ehcache', 'hazelcast', 'infinispan'].includes(cacheProvider) || clusteredHttpSession === 'hazelcast' || applicationType === 'gateway') { _%>
     @Cacheable(cacheNames = USERS_BY_EMAIL_CACHE)
     <%_ } _%>
-    Optional<User> findOneWithAuthoritiesByEmailIgnoreCase(String email);
+    Optional<User> findOneWithAuthoritiesByEmail(String email);
 <%_ } _%>
 
     Page<User> findAllByLoginNot(Pageable pageable, String login);
--- a/generators/server/templates/src/main/java/package/repository/_UserRepository.java
+++ b/generators/server/templates/src/main/java/package/repository/_UserRepository.java
@@ -80,6 +80,11 @@ import static <%=packageName%>.config.Constants.ID_DELIMITER;
 <%_ if (databaseType === 'sql' || databaseType === 'mongodb' || databaseType === 'couchbase') { _%>
 @Repository
 public interface UserRepository extends <% if (databaseType === 'sql') { %>JpaRepository<User, Long><% } %><% if (databaseType === 'mongodb') { %>MongoRepository<User, String><% } %><% if (databaseType === 'couchbase') { %>N1qlCouchbaseRepository<User, String><% } %> {
+    <%_ if (['ehcache', 'hazelcast', 'infinispan'].includes(cacheProvider) || clusteredHttpSession === 'hazelcast' || applicationType === 'gateway') { _%>
+
+    public static final String USERS_BY_LOGIN_CACHE = "usersByLogin";
+    public static final String USERS_BY_EMAIL_CACHE = "usersByEmain";
+    <%_ } _%>
 <%_ if (authenticationType !== 'oauth2') { _%>
 
     Optional<User> findOneByActivationKey(String activationKey);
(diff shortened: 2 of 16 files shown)


generators/server/templates/src/main/java/package/repository/_UserRepository.java
```diff
@@ -109,6 +109,9 @@
     @Cacheable(cacheNames = "users")
     <%_ } _%>
     Optional<User> findOneWithAuthoritiesByLogin(String login);
+
+    @EntityGraph(attributePaths = "authorities")
```
hdurix: Why this request is not `@Cacheable` as opposed to `findOneWithAuthoritiesByLogin`?
Especially if this new method is the first called, the cache could be necessary.

thomastardy: @hdurix could be done, but this cache needs to be maintained in the `UserService` as well.

hdurix: Yes you're right, for eviction

thomastardy: @hdurix added caching for `findOneWithAuthoritiesByLogin`

thomastardy:
Looking forward to getting a review. I hope this makes it into the next release as I want to use it in one of our projects. Feedback is very much appreciated.

thomastardy:
Thanks @ruddell for the review

jdubois:
I am very interested by this, as I have the same issue with start.jhipster.tech - I'll do a review this week

2 comments:
- Keeping a "user name" is important, in order to be able to do user profiles like https://github.com/jdubois without giving the user's email address (so that's important to keep it, in fact)
- one of the ideas I also have is to forbid the use of `@` in a user name, so you can't put an e-mail address as a user name, which would be better for security. Of course, that would break existing applications.

jdubois:
I'm sorry I still didn't have the time to review -> we're going to do a new release pretty soon, so this will be postponed a few days (it will be in the release after this one)

jdubois:
OK so I had a quick look at the code, this looks all good to me. The only issue I find is the following scenario:

- If we have "alice" who creates: user="alice@jhipster.tech" and email="bob@jhipster.tech"
- and "bob" who creates: user="bob" and email="alice@jhispter.tech"

-> then Bob can log in using "alice@jhipster.tech" but will get is own account: good for him
-> but Alice won't be able to use her email address anymore to log in

But: here Alice was pretty stupid, using an e-mail which isn't here real e-mail. And she can still log in using her real false e-mail address, so she's not totally locked out.

As a result:

- Does anybody has any other idea of how this system could be hacked?
- We should forbid using `@` in usernames from now on -> @thomastardy could you add a check when registering a new user? That will remove the problem in the future
- We also need to add a specific text for this, like "Please enter your e-mail address or username" on all login pages: this means in AngularJS/Angular/React and in all i18n translations. I can have a look at this afterwards, I'm used to this.

thomastardy:
@jdubois thanks for the feedback.
Regarding forbidding using `@` in usernames. If a user would use the email address as the username, wouldn't that be a valid use case? Then the check would be, that the username must be the same as the email address if you are using an `@`.

jdubois:
yes of course, but that would cause 2 issues:

- You can change your e-mail address (and also the login in theory, but we don't provide a method for that), so you need to add checks there too.
- The whole point of separating username/email address is that you can keep your email address private. At one point on https://start.jhipster.tech I will add user profiles, like https://github.com/jdubois -> so if you used your email address as a username, people will know it, which isn't great for privacy

thomastardy:
Very good point. I will add a check which prevents using `@` in usernames when creating a user and update the PR.

ruddell:
Our Social Auth implementation saves the user's email as the login for Facebook/Google in [SocialService.java#L149-L156](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/service/_SocialService.java#L149-L156).  Anyways, we can change that logic to get a username instead of an email from the social connections (email is not always present).

This is just something to keep in mind if we restrict the `@` from logins

thomastardy:
@ruddell the build is failing due to the SocialServieIntTest. You said, we can change the logic to use the username instead of the email. Do I understand that correctly that you would change `SocialService.getLoginDependingOnProviderId()` to always use `userProfile.getUsername().toLowerCase()` like we do for twitter?
But what about the comment in the javadoc
> Because provider like Google or Facebook didn't provide login or login like "12099388847393"
If there are providers which don't return you a login and only an email, that's a problem.
Appreciate your feedback.

ruddell:
I think it would still depend on the social provider. We already have an issue if Facebook doesn’t return an email. I won’t have time to look into it until tomorrow.

jdubois:
I'm going to cherry pick your code @thomastardy to have this on https://start.jhipster.tech -> that's a good way to test this "in real life", and with lots of users

jdubois:
It's done and in production at https://start.jhipster.tech -> @thomastardy please have a look at my code review comments as I saw quite a number of issues, but I'm confident it will work (well, I already use it in production, so that should work!! Thanks!)

thomastardy:
@jdubois probably a noob question, but where do I see your comments?

jdubois:
Sorry @thomastardy I forgot to submit my code review!! So you couldn't see them :-)

ruddell:
After comparing the three supported social sites, these are the fields that are available:

| Fields| Facebook | Twitter | Google |
| :---         |     :---      |          --- |          --- |
| **ID**   | null     | null    | null    |
| **Name**     | Jon Ruddell       | Jon Ruddell      | Jon Ruddell      |
| **First**     | Jon       | Jon      | Jon      |
| **Last**     | Ruddell       | Ruddell      | Ruddell      |
| **Email**     | optional       | null      | my-email@gmail.com      |
| **Username**     | null       | jonruddell      | 145932352034721158641      |

It looks like the only thing we can always depend on is the name.  Maybe change `return userProfile.getEmail();` to `return userProfile.getFirstName() + userProfile.getLastName();`?

thomastardy:
@ruddell thanks for the analysis of the social logins. I think generating the username based on first name and last name could work. It has the issue, that you can change these fields in Facebook, Twitter and Google, but your username in a JHipster application is not modifiable.

@jdubois thanks for the review comments. I will do the required changes. What's your opinion on how the handle the username for social sites?

pascalgrimaud:
One minor issue, but not related to this PR.
The user can edit his profile and use another email which doesn't exist in the current database. As there is no confirmation, the real owner won't be able to use his email to register.

generators/languages/templates/src/main/webapp/i18n/en/_global.json
```diff
@@ -143,7 +143,8 @@
             "maxbytes": "This field cannot be more than {{ max }} bytes.",
             "pattern": "This field should follow pattern for {{ pattern }}.",
             "number": "This field should be a number.",
-            "datetimelocal": "This field should be a date and time."
+            "datetimelocal": "This field should be a date and time.",
+            "patternLettersAndDigits": "This field can only contain letters and digits."
```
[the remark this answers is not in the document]
thomastardy: The texts on the registration component are different than the ones used in the user management. The registration is directed at the current user and the texts start with "Your username ...". The user management uses texts "This field ...".
Do you still think that we use for this new text the same value as during the registration?

thomastardy:
@jdubois I've updated the PR with your comments, but not the i18n one. Please have a look at my feedback. I also changed the `SocialService` according to @ruddell suggestion.

The build of `app-default-from-scratch` is failing on travis, but not locally. Do you have an idea why the build fails?

jdubois:
Sorry we merged #6796 and this impacts your code (you have conflicts now...). I spent a big part of the day debugging that other PR, and we need a few days before it's stable (it's a huge change, in fact). When that will be OK I'll be able to look at your PR, as I know both of them quite well now.

thomastardy:
@jdubois just go ahead. I will not have time to work on this anyways before the holidays.

jdubois:
Thanks @thomastardy - I'm going to try to merge this now

jdubois:
I'm going to finish this today - lots of code changes because it overlaps with #6796 !!
This is kind of tricky, I'm very happy with the end result but I hope we didn't introduce some bug!

jdubois:
And it's merged :-)
Lots of work here, this will need more testing

Issue #6736, opened 2017-11-25
issue from frontend-plugin cause npm&yarn tasks error


lmbai:
<!--
- Please follow the issue template below for bug reports and feature requests.
- If you have a support request rather than a bug, please use [Stack Overflow](http://stackoverflow.com/questions/tagged/jhipster) with the JHipster tag.
- For bug reports it is mandatory to run the command `jhipster info` in your project's root folder, and paste the result here.
- Tickets opened without any of these pieces of information will be **closed** without any explanation.
-->

##### **Overview of the issue**

when use maven to run `mvn -Pwebpack` profile, the build fail. 

##### **Motivation for or Use Case**

the default profile for maven  is `develop`, which won't do the frontend complile. the jhipster suggest to run `yarn start` to do the frontend compile. but i plan to run a dev environment in a test server.
so i would like to run `mvn -Pwebpeck -Pdev` to do the frontend and backed compile together. but it failed.

##### **Reproduce the error**

<!-- For bug reports, an unambiguous set of steps to reproduce the error -->

1. set the https-proxy  and http-proxy in your maven's `settings.xml` file. 
2. run `mvn -Pwebpeck` in your jhipster projects folder.
> it will thow a error say that `Unknown arguments: https-proxy, proxy`

##### **Related issues**

<!-- Has a similar issue been reported before? Please search both closed & open issues -->
[https://github.com/eirslett/frontend-maven-plugin/issues/660](https://github.com/eirslett/frontend-maven-plugin/issues/660)


##### **Suggest a Fix**

<!-- For bug reports, if you can't fix the bug yourself, perhaps you can point to what might be
  causing the problem (line of code or commit) -->

For `frontend-maven-plugin`, It will add the proxy(here it is the proxy for maven) as additional parameters to the command line for every command.
if you have a proxy setting in the maven config file, it will add the `--https-proxy=http://xxxxxx --proxy=http://xxxxxx` for each command. 
in the pom.xml, we have some frontend tasks:
<details>                                                                                                      
<summary>POM.xml file: prod frondend plugin</summary>                                                                            
                                                                                                         
```xml
<plugin>
	<groupId>com.github.eirslett</groupId>
	<artifactId>frontend-maven-plugin</artifactId>
	<version>${frontend-maven-plugin.version}</version>
	<executions>
		<execution>
			<id>install node and yarn</id>
			<goals>
				<goal>install-node-and-yarn</goal>
			</goals>
			<configuration>
[...]

PR #6737, opened 2017-11-25
disable the auto proxy for un-necessary jobs

lmbai:
disable the proxy setting for un-necessary jobs for the frontend plugin
which may cause error

### DIFF
--- a/generators/server/templates/_pom.xml
+++ b/generators/server/templates/_pom.xml
@@ -1176,6 +1176,11 @@
                                 <phase>generate-resources</phase>
                                 <configuration>
                                     <arguments>run webpack:build</arguments>
+                                    <%_ if (clientPackageManager === 'yarn') { _%>
+                                        <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
+                                    <%_ } else if (clientPackageManager === 'npm') { _%>
+                                        <npmInheritsProxyConfigFromMaven>false</npmInheritsProxyConfigFromMaven>
+                                    <%_ } _%>
                                 </configuration>
                             </execution>
                         </executions>
@@ -1393,6 +1398,11 @@
                                 <phase>test</phase>
                                 <configuration>
                                     <arguments>run webpack:test</arguments>
+                                    <%_ if (clientPackageManager === 'yarn') { _%>
+                                        <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
+                                    <%_ } else if (clientPackageManager === 'npm') { _%>
+                                        <npmInheritsProxyConfigFromMaven>false</npmInheritsProxyConfigFromMaven>
+                                    <%_ } _%>                                    
                                 </configuration>
                             </execution>
                             <execution>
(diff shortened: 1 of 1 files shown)


lmbai:
#6736

gmarziou:
What is the behavior with gradle?

deepu105:
I'm not sure if this affects everyone, this might be affecting only @lmbai so this needs to be tested by someone in a different non proxy and proxy environment to be sure

lmbai:
@deepu105 . This issue has been there since the begin, but in most of the circumstance， it didn't trigger.  I descripe it in the [https://github.com/eirslett/frontend-maven-plugin/issues/660](https://github.com/eirslett/frontend-maven-plugin/issues/660), and analaze  and give the scenario to reproduce it in the #6736 

1. it trigger only if we use `proxy` in the maven config. Because `frontedn-maven-plugin` will detect and use append it as "--https-proxy=http://xxxx" if the maven proxy is enabled.   i thinkd `frontedn-maven-plugin` implement it in this way to let `yarn install` use the proxy. but it is not nesseary for any other command like `yarn run webpack:build`, and it will trigger a fail in some condition.
2, for some npm task like `yarn run a && yarn run b` , the `--https-proxy=http://xxxx` will just work for the last task: `yarn run b`, so if we use `mvn -Pprod` to build the jhipster project.  the additional `--https-proxy=http://xxxx`  will apply in the `rimraf target//www/app/{src,target/}` , it won't trigger the error.
>but if we use `mvn -Pwebpack` to build the jhipster frontend, the additional  `--https-proxy=http://xxxx` will appy in the `node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js` eventually, which will trigger the error.

In the pull request, i didn't apply this in all the `frontend-maven-plugin` tasks. the `yarn insall` will still use this additional proxy setting. i just disable this proxy in the tasks which doesn't require the proxy at all.
i already suggest the `frontend-maven-plugin` to disable this by default in [https://github.com/eirslett/frontend-maven-plugin/pull/676](https://github.com/eirslett/frontend-maven-plugin/pull/676)

about the logic to add  this additional proxy setting, it is in the [eirslett/maven/plugins/frontend/lib/YarnRunner.java](https://github.com/eirslett/frontend-maven-plugin/blob/960bd8cba502ac30ecaf9e414f65b9969effb2f1/frontend-plugin-core/src/main/java/com/github/eirslett/maven/plugins/frontend/lib/YarnRunner.java#L41)

lmbai:
@gmarziou , Good question, since the `maven build`  is affected. how about the `gradle build` ?

i just dig some more about this with jhipster:

1. Jhipster choose  the `frontend-maven-plugin` for maven  to do  the frontend build. the way it run the yarn command cause this issue.
2. But Jhipster didn't use the `frontend-maven-plugin` for Gradle. it use `com.moowork.gradle:gradle-node-plugin:1.2.0` for the yarn tasks. so there is no issue in the gradle build. 


```
λ gradle buildEnvironment
Starting a Gradle Daemon, 1 stopped Daemon could not be reused, use --status for details

> Task :buildEnvironment

------------------------------------------------------------
Root project
------------------------------------------------------------

classpath
+--- org.springframework.boot:spring-boot-gradle-plugin:1.5.7.RELEASE
|    +--- org.springframework.boot:spring-boot-loader-tools:1.5.7.RELEASE
|    |    \--- org.springframework:spring-core:4.3.11.RELEASE
|    |         \--- commons-logging:commons-logging:1.2
|    \--- io.spring.gradle:dependency-management-plugin:1.0.3.RELEASE
+--- org.springframework.build.gradle:propdeps-plugin:0.0.7
+--- org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:2.5
|    \--- org.sonarsource.scanner.api:sonar-scanner-api:2.9.0.887
+--- net.ltgt.gradle:gradle-apt-plugin:0.11
+--- io.spring.gradle:dependency-management-plugin:1.0.3.RELEASE
\--- com.moowork.gradle:gradle-node-plugin:1.2.0
```

PierreBesson:
I can vouch for this solution. I was having problem at work today packaging the jhipster-registry behind a proxy and this works like a charm. Great work @lmbai !

PR #6739, opened 2017-11-25
Added maven site reporting

Tcharl:
Not totally finished:
* Should test release plugin (for example automating a travis site)
* Should update pipelines
* Should update documentation (for example, on GPG).

- [ x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x ] Tests are added where necessary
- [x ] Documentation is added/updated where necessary [PR](https://github.com/jhipster/jhipster.github.io/pull/518)
- [x ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,22 @@
 os:
-  - linux
+- linux
 services:
-  - docker
+- docker
 language: node_js
 node_js:
-  - "8.9.1"
+- 8.9.1
 jdk:
-  - oraclejdk8
+- oraclejdk8
 addons:
   apt:
     sources:
-      - google-chrome
+    - google-chrome
     packages:
-      - google-chrome-stable
+    - google-chrome-stable
 cache:
   directories:
-    - $HOME/.m2
-    - $HOME/.gradle
+  - "$HOME/.m2"
+  - "$HOME/.gradle"
 env:
   global:
     - PROFILE=dev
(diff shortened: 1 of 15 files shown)


Tcharl:
Hi,

Travis continuous delivery of site and code using maven release plugin (on central) works well now! Take a look at the configuration: https://github.com/Tcharl/test-travis-release

I'll try to make the code for Jenkins too, but these are mainly comments (on gpg encryption generation, ssh, Etc.), so IMO this one can be merged.

Once merged, I'll integrate it in jhipster server libraries, so that @jdubois won't spend many hours to release :-p

deepu105:
Im apprehensive about adding something specific to maven, so far we have kept features in sync between maven and gradle so that we dont favour one over the other. I don't think we should break that tradition now.
Can this be added for Gradle as well otherwise I would suggest to do this as a module

Tcharl:
Maintaining a module for the build files is really a pain: there are conflicts every two days in mean (I work on this PR from months, and try to fix conflicts everytime).

I think that after the @atomfrede  gradle plugin, users will have to choose between performance (Gradle), and Enterprise features like continuous release and site (Maven).

Gradle is not done to be as feature complete than Maven in term of industrialization (for example, there's no metadata declared in Gradle scripts).

Site can really help, for example in order to maintain jhipster bom: you've live dependency-convergence reports, and it's a real killer feature in companies trying to do things like Inner Source (free and standardized website for anyone).

Finally, Continuous Release (for example with Travis on Maven Central) is very cool: once you've tested it, you don't want to go back ^^ :-).

pascalgrimaud:
First of all, thanks for your hard work @Tcharl. I know you spend a lot of time on this.

I will give my opinion on this.

1) I agree with @deepu105, we must have same features for Gradle and Maven.

2) why did you have so much conflicts ? It's because you changed a lot the `pom.xml` which is regularly changed. I saw so much changes in this PR

3) about module: if you had choosen a module, you are totally free
    - you can add dependencies
    - you can replace some part, using regexp
    - you can add all files you want
    - and the hardest part, you can erase the current `pom.xml` by yours from your module
    - you can really do what you want, if you choose to support only Maven and not Gradle, it's your module, you're the owner so you decide the roadmap

4) I just tested your PR, and it generated some files, which contains hard link, which doesn't exist:
    - `<url>https://github.com/io.github.pascalgrimaud/jhipster</url>` with the comment `It may not be the right addresses`
    - I think some enterprises which don't want these as default -> a lot of generated projects are used internally, and are not published to GitHub

Nothing against you, but I don't understand why people prefers adding feature requests directly to main generator-jhipster instead of trying module.
In my opinion, for every big features:
- make 1st a module
- see if it's used a lot or not
- if the module is a great success, so yes, we can reconsider to add to the main generator


It's my personal opinion, not from the team :-)

deepu105:
@Tcharl its more of a team choice to keep gradle and maven support similar, yes indeed they both can do different things, but when we offer 2 technologies for the same thing we try to keep them similar as much as possible, same goes for our angular/react part. I know you have been pushing this for sometime now, but I wish you had first opened an issue and discuss this with team instead of doing PR to get everyones opinion. I do really appreciate your contributions but sometimes we also need to cutback on adding new stuff as it increases our maintanance burden. @jdubois WDYT?

erikkemperman:
To be fair, I think there was an issue but it may have gotten a bit convoluted with the effort to separate dependencies in the BOM...

Tcharl:
Sorry, @deepu105 , I thought https://github.com/jhipster/generator-jhipster/issues/6061 was the discussion, and then, with the help that PR, and its the reference sample, you (the team) would discuss on integrating it or not.

@erikkemperman , IMHO, the issue is not totally resolved: I think that not having a good release process is I real lack for the end user (also the team), but its my only my point of view.

I don't think there's much of a maintenance burden with this PR: build plugins are always the same between Gradle and Maven, so isofunctionality do not involve more things than before.

Thank you all for your feedback and your words, let's continue to exchange in order to add (or not) this contribution.

erikkemperman:
Well one thing has significantly in that direction, we can now attach our own corporate parent POM to the generated apps, which typically might have release sugar in it.

Tcharl:
@erikkemperman good point, the additions in here (the parent and reporting poms) can be added as parent 'enterprise' poms. However isn't JHipster made to replace swarc teams?

An alternative could be to offer these pom in a separate repo so that people without company poms can use them if wanted. WDYT?

deepu105:
sorry, @Tcharl that issue(#6061) had many different things and the BOM part took the attention there and at least I personally missed this. Anyway i'm more of a Gradle guy hence lets see what @jdubois has to say as he likes maven more

Tcharl:
I made a JHipster module for this. It is not fully finished, but it brings mosts of the functionalities.

Can this pr stay open a little more, just as a reminder in order to make the module as complete as this PR?

deepu105:
Ok


Thanks & Regards,
Deepu

PR #6745, opened 2017-11-26
Add hint for how to run jhipster after linking local generator

duderoot:
After linking the local generator, I have problems to see the changes when running jhipster in my local app. The problem was that I need to run `jhipster --with-entities` because the changes that I have done in the generator were in the entity template.
Such a hint will save times for newbies like me.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,6 +203,11 @@ Now, running the 'jhipster' command should use your specific JHipster version. Y
 ```shell
 jhipster
 ```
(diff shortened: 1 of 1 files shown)


gmarziou:
Alternatively you could have used also the upgrade generator `jhipster upgrade --force`

duderoot:
@gmarziou I use` jhipster upgrade --force` when I am updating my production apps and use tagged version of the generator. For working on generator I didn't think to use this option because I thought that it will be gone create also an upgrade of the app. Normally when working on generator man work, I guess, on the last version, so upgrade should work fine, but I have blocked that thought and not considered 'jhipster upgrade --force' as a viable solution.

Issue #6766, opened 2017-11-29
[Feature request] Improve OpenShift integration with Gradle and support for MSSQL


Spengreb:
##### **Overview of the issue**
The current implementation of OpenShift integration does not support any Gradle tasks and depends on the OpenShift CLI. It also seems to favor micro-service spring boot applications, creating an ocp/ the directory above the project root. Furthermore it also does not support MSSQL (as there is no stable Docker version).  

##### **Motivation for or Use Case**

- I wanted to deploy to OpenShift through GitLab CI and so I wanted to deploy via Gradle. 
- I use a Monolithic version and so I was not satisfied with what was generated (and was a bit confused at first). 
- I also use MSSQL with the paradigm of each branch I create having its own database in a single MSSQL instance. 

##### **Suggest a Fix**
I created a Gradle plugin with opinionated tasks for deploying to OpenShift. https://gitlab.com/hendosdo/gradle-openshift-tasks-plugin

This will start builds on Openshift and send deployments to it. This is done by modifying the deployment.yml a bit for updating application name and updating database location. 

I also have a local version of JHipster which will generate files properly in places the Plugin expects. I also retained existing functionality. It asks a few more questions as a result, like if mssql is selected it will ask for username, password, and URL. 

I will submit a PR if there is interest in this.

##### **JHipster configuration**
Running yo jhipster:openshift will create appropriate files, it checks the yo-rc.json file for existing configuration

- [x] Checking this box is mandatory (this is just to show you read everything)

jdubois:
- The MSSQL image we use is `'microsoft/mssql-server-linux:latest'` so I don't understand why you say we don't support it/we don't have a stable version? Anyway, in production, I wouldn't recommend running your database in Docker, but to use a managed service.
- I also don't understand your 3rd point - is it worrying you that we have one database per microservice? That's the whole point of doing microservices: if you have one single database, then you just have a distributed monolith. Note that if you want to do this, it isn't very hard, there is just one config key in `application-*.yml` to modify.
- I'm not sure what Gradle does here: our OpenShift support should work both with Maven and Gradle, and what I read at http://www.jhipster.tech/openshift/ looks good -> I'm not an expert here, @srinivasavasu is the boss on this, so I'm affecting the ticket to him, and hope he can answer us

deepu105:
@Spengreb 
[...]

PR #6768, opened 2017-11-29
Openshift

Spengreb:
#6766 

### DIFF
--- a/generators/openshift/files.js
+++ b/generators/openshift/files.js
@@ -10,9 +10,11 @@ function writeFiles() {
             for (let i = 0; i < this.appConfigs.length; i++) {
                 const appName = this.appConfigs[i].baseName.toLowerCase();
                 this.app = this.appConfigs[i];
-                this.template('_deployment.yml', `${this.directoryPath}/ocp/${appName}/${appName}-deployment.yml`);
+                this.template('_deployment.yml', `${this.directoryPath}/ocp/${appName}/${appName}-deployment.yml`); //microsservice
+                this.template('_deployment.yml', `src/main/docker/openshift/${appName}-deployment-template.yml`); //monolithic
+                this.template('gradle/_openshift.gradle', `gradle/openshift.gradle`);
 
-                if (this.app.prodDatabaseType !== 'no') {
+                if (this.app.prodDatabaseType !== 'no' && this.app.prodDatabaseType !== 'mssql') {
                     this.template(`db/_${this.app.prodDatabaseType}.yml`, `${this.directoryPath}/ocp/${appName}/${appName}-${this.app.prodDatabaseType}.yml`);
                 }
                 if (this.app.searchEngine === 'elasticsearch') {
--- a/generators/openshift/files.js
+++ b/generators/openshift/files.js
@@ -10,10 +10,14 @@ function writeFiles() {
             for (let i = 0; i < this.appConfigs.length; i++) {
                 const appName = this.appConfigs[i].baseName.toLowerCase();
                 this.app = this.appConfigs[i];
-                this.template('_deployment.yml', `${this.directoryPath}/ocp/${appName}/${appName}-deployment.yml`); //microsservice
-                this.template('_deployment.yml', `src/main/docker/openshift/${appName}-deployment-template.yml`); //monolithic
-                this.template('gradle/_openshift.gradle', `gradle/openshift.gradle`);
-
+                
+                if (this.app.applicationType === 'monolith') {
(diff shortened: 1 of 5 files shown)


generators/openshift/files.js
```diff
@@ -10,9 +10,11 @@ function writeFiles() {
             for (let i = 0; i < this.appConfigs.length; i++) {
                 const appName = this.appConfigs[i].baseName.toLowerCase();
                 this.app = this.appConfigs[i];
-                this.template('_deployment.yml', `${this.directoryPath}/ocp/${appName}/${appName}-deployment.yml`);
+                this.template('_deployment.yml', `${this.directoryPath}/ocp/${appName}/${appName}-deployment.yml`); //microsservice
```
srinivasa-vasu: There is no need to have a separate deployment file based on the nature of the app. May I know the reason behind doing this way?

generators/openshift/files.js
```diff
@@ -10,9 +10,11 @@ function writeFiles() {
...
+                this.template('_deployment.yml', `src/main/docker/openshift/${appName}-deployment-template.yml`); //monolithic
+                this.template('gradle/_openshift.gradle', `gradle/openshift.gradle`);
 
-                if (this.app.prodDatabaseType !== 'no') {
+                if (this.app.prodDatabaseType !== 'no' && this.app.prodDatabaseType !== 'mssql') {
```
srinivasa-vasu: This comes from main generator. MSSQL is already there as one of the prod db types. All you have to do is to add a file under /templates/db/_mssql.yml. These changes are not really required. Just follow how it is done for other DB types like MySQL, Postgres (these options come from the main generator)

generators/openshift/index.js
```diff
@@ -74,6 +74,11 @@ module.exports = class extends BaseGenerator {
                 this.dockerRepositoryName = this.config.get('dockerRepositoryName');
                 this.dockerPushCommand = this.config.get('dockerPushCommand');
                 this.openshiftNamespace = this.config.get('openshiftNamespace');
+                this.mssqlURL = this.config.get('mssqlURL');
+                this.mssqlUser = this.config.get('mssqlUser');
+                this.mssqlPass = this.config.get('mssqlPass');
+                this.mssqlDb = this.config.get('mssqlDb');
```
srinivasa-vasu: Just follow how it is done for other DB types like MySQL, Postgres etc. No need to do a separate track for MSSQL

generators/openshift/templates/_deployment.yml
```diff
@@ -63,6 +57,12 @@ parameters:
     description: Port of the application
     required: true
     displayName: Application Port
+  -
+    name: NAMESPACE
+    value: "<%= openshiftNamespace %>"
+    description: Openshift Namespace
+    required: true
+    displayName: OpenShift namespace
```
srinivasa-vasu: Templates are generated based on response to NAMESPACE prompt. I don't really see a value to parameterise this unless if you have a specific reason.

generators/openshift/files.js
Spengreb: Yeah, oops. Thats a mistake. I corrected that already and put an if to check if its a monolithic application or not. If its monolithic it can be deployed via the gradle deploy task.

generators/openshift/index.js
Spengreb: This and the comment above are to support MSSQL being an external database service. So it doesn't need to generate a new pod and instead needs the connection information.

generators/openshift/templates/_deployment.yml
Spengreb: Oh yeah, you're right.

generators/openshift/index.js
srinivasa-vasu: @Spengreb I get it :), but why MSSQL alone? Why not other db types? You can actually change connectionURL being specified in env property of the deployment.yml manually. If you want to make it generic for accessing external services, then it makes sense to do it through Open service broker API. This is in technical preview in OC 3.7.

Spengreb:
Fair enough. Thanks for the feedback. 

The Open Service Broker API seems very interesting and I'll have a look into it. Though we're running 3.6 currently.

Adding external DB support as a generic option is something i'd consider adding instead of the current implementation as it would be a nicer solution.  

Though i'm unwilling to write the deployment plugin again in MVN as I don't use it and i'm not at all familiar with it.

deepu105:
@Spengreb please dont be demotivated, its just that we cant do changes if they arent generic enough for everyone, you are still welcome to add a generic external DB option if you like

jdubois:
Yes Maven is 90% of our users, so we can't have something specific for Gradle - so sorry about this @Spengreb

PR #6769, opened 2017-11-29
Add methods for work with yaml file

antarus:
This code allows you to add one or more properties to a yaml file.
code update from hipster-labs/generator-jhipster-spring-cloud-stream

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -386,7 +386,7 @@ function getPropertyInArray(array, name, generator) {
 
 /**
  * Update or create a property by path in a array
- * 
+ *
  * @param {array} array - array in which we want to update
  * @param {string} name - property name to search "format myProperty.level2.level3"
  * @param {string} generator - The generator
@@ -398,7 +398,7 @@ function updatePropertyInArray(array, name, generator, value) {
 
 /**
  * Delete a property by path in a array
- * 
+ *
  * @param {array} array - array in which we want to delete
  * @param {string} name - property name to search "format myProperty.level2.level3"
  * @param {string} generator - The generator
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -23,6 +23,7 @@ const shelljs = require('shelljs');
 const ejs = require('ejs');
 const _ = require('lodash');
 const constants = require('./generator-constants');
+const jsyaml = require('js-yaml');
 
 const LANGUAGES_MAIN_SRC_DIR = `../../languages/templates/${constants.CLIENT_MAIN_SRC_DIR}`;
 
@@ -37,7 +38,32 @@ module.exports = {
     deepFind,
     getJavadoc,
     buildEnumInfo,
-    copyObjectProps
+    copyObjectProps,
+    // array
+    getPropertyInArray,
+    updatePropertyInArray,
+    deletePropertyInArray,
+    // yaml read
+    getYamlProperty,
+    // yaml properties
+    addYamlPropertiesAtBeginin,
+    addYamlPropertiesAtEnd,
+    addYamlPropertiesBeforeAnotherProperty,
+    addYamlPropertiesAfterAnotherProperty,
+    addYamlPropertiesAtLineIndex,
+    // yaml property
+    addYamlPropertyAtBeginin,
+    addYamlPropertyAtEnd,
+    addYamlPropertyBeforeAnotherProperty,
+    addYamlPropertyAfterAnotherProperty,
+    addYamlPropertyAtLineIndex,
+    // functions
+    getLastPropertyCommonHierarchy,
+    getPathAndValueOfAllProperty,
+    // main
+    updateYamlProperties,
+    updateYamlProperty
+
 };
 
 /**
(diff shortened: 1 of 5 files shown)


generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+
+/**
+ * get a property by path in a array
+ *
+ * @param {array} array - array in which we want to search
+ * @param {string} name - property name to search "format myProperty.level2.level3"
+ * @param {string} generator - The generator
+ * @return {string} the value of property or undefined
+ */
+function getPropertyInArray(array, name, generator) {
```
deepu105: Please use lodash directly. This method doesnt add any value so please remove it

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+
+/**
+ * Update or create a property by path in a array
+ *
+ * @param {array} array - array in which we want to update
+ * @param {string} name - property name to search "format myProperty.level2.level3"
+ * @param {string} generator - The generator
+ * @param {string} value - (optional) value of the property
+ */
+function updatePropertyInArray(array, name, generator, value) {
```
deepu105: Please use lodash directly. This method doesnt add any value so please remove it

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+}
+
+/**
+ * Delete a property by path in a array
+ *
+ * @param {array} array - array in which we want to delete
+ * @param {string} name - property name to search "format myProperty.level2.level3"
+ * @param {string} generator - The generator
+ */
+function deletePropertyInArray(array, name, generator) {
```
deepu105: Please use lodash directly. This method doesnt add any value so please remove it

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+/**
+ * Get a yaml property by path in a file
+ *
+ * @param {string} file - yaml file to search
+ * @param {string} name - property name to search "format myProperty.level2.level3"
+ * @param {string} generator - The generator
+ * @return {string} the value of property or undefined
+ */
+function getYamlProperty(file, name, generator) {
+    const treeData = jsyaml.safeLoad(generator.fs.read(file));
```
deepu105: This is not efficient if reading multiple props. Ideally, you should just load file once and read all required values from that. Its better to do this directly using the library. I dont see added value in having this method

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+}
+
+/**
+ * Add yaml properties at beginin
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ */
+function addYamlPropertiesAtBeginin(file, properties, generator) {
```
deepu105: method name should be `addYamlPropertiesAtBeginning`

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+ * Add yaml properties before another property
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ * @param {string} propertyBefore - The property before which we wish to insert new properties
+ * @param {string} addBeforeComment - (Optional) if this variable is defined, means that we will return the index before the previuous comment of the property.
+ */
+function addYamlPropertiesBeforeAnotherProperty(file, properties, generator, propertyBefore, addBeforeComment) {
+    const body = generator.fs.read(file);
```
deepu105: pass the body instead of reading it every time

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+    const newLines = jsyaml.safeDump(properties, { indent: 4 }).split('\n');
+    const applicationPropertyIndex = getIndexBeforeLineOfProperty(body, propertyBefore, generator, addBeforeComment);
+    if (applicationPropertyIndex === -1) {
+        throw new Error(`Property ${propertyBefore} not found`); // inside callback
+    }
+    const bodyLines = body.split('\n');
+    bodyLines.splice(applicationPropertyIndex, 0, newLines.join('\n'));
+    generator.fs.write(file, bodyLines.join('\n'));
```
deepu105: return the result back instead of writing it every time

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+
+/**
+ * Add yaml properties at beginin
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ */
+function addYamlPropertiesAtBeginin(file, properties, generator) {
+    const bodyLines = generator.fs.read(file).split('\n');
```
deepu105: pass the read value and dont read here

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+    const newLines = jsyaml.safeDump(properties, { indent: 4 }).split('\n');
+    bodyLines.splice(bodyLines.count, 0, newLines.join('\n'));
+    generator.fs.write(file, bodyLines.join('\n'));
```
deepu105: return result dont write here

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+
+/**
+ * Add yaml properties at end
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ */
+function addYamlPropertiesAtEnd(file, properties, generator) {
+    const bodyLines = generator.fs.read(file).split('\n');
```
deepu105: pass body in param, dont read here

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+    const newLines = jsyaml.safeDump(properties, { indent: 4 }).split('\n');
+    generator.fs.write(file, bodyLines.concat(newLines).join('\n'));
```
deepu105: return result, dont write here

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+/**
+ * Add yaml properties after another property
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ * @param {string} propertyAfter - The property after which we wish to insert new properties
+ */
+function addYamlPropertiesAfterAnotherProperty(file, properties, generator, propertyAfter) {
+    const body = generator.fs.read(file);
```
deepu105: pass body, dont read

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+    const newLines = jsyaml.safeDump(properties, { indent: 4 }).split('\n');
+    const applicationPropertyIndex = getIndexAfterLineOfProperty(file, propertyAfter, generator);
+    if (applicationPropertyIndex === -1) {
+        throw new Error(`Property ${propertyAfter} not found`); // inside callback
+    }
+    const bodyLines = body.split('\n');
+    bodyLines.splice(applicationPropertyIndex, 0, newLines.join('\n'));
+    generator.fs.write(file, bodyLines.join('\n'));
```
deepu105: return result dont write

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+ * Add yaml properties at a specific line
+ *
+ * @param {string} file - yaml file to update
+ * @param {array} properties - array of property to add
+ * @param {string} generator - The generator
+ * @param {string} indexLine - Index of line which we wish to insert new properties
+ * @param {string} numberSpace - number espace to start
+ */
+function addYamlPropertiesAtLineIndex(file, properties, generator, indexLine, numberSpace) {
+    const body = generator.fs.read(file);
```
deepu105: pass body, dont read

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+    const newLines = jsyaml.safeDump(properties, { indent: 4 }).split('\n');
+    const bodyLines = body.split('\n');
+    let spaceStr = '';
+    for (let i = 0; i < numberSpace; i++) {
+        spaceStr += ' ';
+    }
+    bodyLines.splice(indexLine, 0, newLines.map(line => spaceStr + line).join('\n'));
+    generator.fs.write(file, bodyLines.join('\n'));
```
deepu105: return result, dont write

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+}
+
+/**
+ * Add a yaml property at beginin
+ * TODO manage value of array type
```
deepu105: please don't add incomplete methods to public util file

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+ * @param {string} numberSpace - number espace to start
+ */
+function addYamlPropertyAtLineIndex(file, property, value, generator, indexLine, numberSpace) {
+    const properties = {};
+    updatePropertyInArray(properties, property, generator, value);
+    addYamlPropertiesAtLineIndex(file, properties, generator, indexLine, numberSpace);
+}
+
+/**
+ * Get the last property of a common hierarchical.
```
deepu105: write a better explanation comment

generators/utils.js
```diff
@@ -344,3 +370,457 @@ function buildEnumInfo(field, angularAppName, packageName) {
...
+            if (idxPropTmp !== -1) {
+                strPropTmp = strPropTmp.substring(0, idxPropTmp);
+            }
+        }
+    }
+    return lastValideParentProperty;
+}
+
+/**
+ * Retourne l'index de la ligne d'une propriété simple et unique
```
deepu105: please write comments in english

deepu105:
@antarus I don't think these should be added to the main generator. They are too specific to handle YML files. Maybe just publish them as a generic utility library

antarus:
Hello, the goal is to facilitate the integration of external module to the main generator. There is currently nothing in the generator to add, edit a spring's configuration file . For me, it's missing. There are function to add graddle or maven dependencies, for me it's the same notion. It took me a long time to manage this for the stream-cloud module.
I haven't found any library that allows you to add a yaml property without removing all the formatting, existing comments, etc. (I may not have searched well!)

I think the main generator lacks tools to make life easier for module developers and their integration.

After about the code deportation in an external library, I don't know, I'm not a node developer, I've never done it.

sendilkumarn:
I was thinking about something similar for JHipster Modules (while working on JHipster-kotlin)🤔 where we need to add few things extra to the base `build.gradle` or `pom.xml` something like this will be really useful. And again these should be a _separate library_.

something similar to angular-cli, how it is generating the application

deepu105:
@antarus let me think about this. May be I can create a utils project under jhipster and you can PR these there. Give me some time

pascalgrimaud:
As @antarus said, it's for adding properties into spring boot yaml configuration. For that, currently, we have differents solutions:
- add at the end of yaml file the full key/value, like: `hello.world.name.properties: value`. It's not consistent with the other properties but it works
- add properties into a new file: `application-yolo.yml` and say to the user to start the project by adding the profile `yolo`

- try to regexp / add / replace the yaml properties
    - every module can copy/paste these functions if needed -> it's in spring-cloud-stream module today
    - integrate these functions written by @antarus, into main generator
    - integrate these functions to an existing library -> it's strange that it doesn't exist today. Maybe pull request to one of existing library ?
    - made a new one

deepu105:
@pascalgrimaud yes I understand the need but I feel that if we start adding everything into the main generator it will become like a kitchen sink. Let's do a new library called `jhipster-utils` for these kinds of stuff. I'll create the repo and publish a skeleton and then @antarus can do the PR there

PierreBesson:
For me this is a valid use case. There should be an API to add properties from a module. However manipulating YAML files sucks, so I'm a bit reserved on this.

@deepu105, Do we really need to create yet another repo and npm package for which we will need to manage the release cycle. I have already lost track of how many we already have !

deepu105:
@PierreBesson multiple npm repo are better than making our core a kitchen sink. Imagine the amount of code that would need to be updated each time we rewrite something, say for example like the ES6 update I did. These are too specific functions for YML handling which ideally should be done by a lib like jsyaml, but if not either PR to that or worst case create a library to do that. This lib might never need much maintenance. Given the current size of Jhipster codebase, I'm reserved on adding stuff like these here. If is is one or two methods that's fine but this is too much code that i'm not willing to maintain in core. Remember every line of code we add makes our life tougher :wink:

deepu105:
Also the main problem is that these methods are never used by the generator itself its only used by one module now and if another module needs it in the future than an external lib is ideal for that

PierreBesson:
@deepu your 2nd point convinced me. Still wished we didn't have so many repos but I guess this can't be helped.

pascalgrimaud:
Agree with you @deepu105
That's why my prefered solution for this need is:
- add at the end of yaml file the full key/value, like: hello.world.name.properties: value. It's not consistent with the other properties but it works

deepu105:
@pascalgrimaud I dont understand one thing cant you load the YML using jsyaml and add new props and stuff to the parsed object and then dump the object back using jsyaml? Isnt that what we do in our Kubernetes generator and stuff? why is it required to do all these manual parsing? Sorry of its a stupid question

pascalgrimaud:
@antarus already tried this, it works fine but we lost all comments, which are so important

antarus:
Hello.

there are two problems with jsyaml, the first like the pascal says, we lose all the comments.

the second problem, when we load the application-dev file and save it, it adds 'null' to the 'application:' property

deepu105:
@pascalgrimaud @antarus ok I understand. Did we raise the issues to jsyaml guys? I'm pretty sure they might be able to fix them since its a valid usecase?

deepu105:
Since no other repo uses these at the moment if jsyaml guys can fix the issue I guess that would be the best solution

antarus:
jsyaml has not been updated since September 19th, the null is report since January 24th, and the comments not manage it seems to me that it's also reported ..

pascalgrimaud:
For now, I think we should not merge this into generator-jhipster as it's specific for modules and globally for Yaml.
So let keep these functions in [generator-jhipster-spring-cloud-stream](github.com/hipster-labs/generator-jhipster-spring-cloud-stream)

Depending if a lot of module need it or not, we can reconsider later:
- merge request to [js-yaml project](https://github.com/nodeca/js-yaml)
- fork the project and maintain it
- make a new project

So I'm closing this, I hope you agree, @antarus

antarus:
I agree that in the state this PR is too oriented yaml, but I remain convinced that it is necessary to facilitate the integration of modules in the main project.

deepu105:
I'm gonna follow YAGNIY principle here, so if another module ever needs it
we can reconsider


Thanks & Regards,
Deepu

PR #6783, opened 2017-12-01
Do not allow to update create date in audit entity

maivantan1992:
The create date in entity should not changeable. So do not allow to update create date when update entity.

### DIFF
--- a/generators/server/templates/src/main/java/package/domain/_AbstractAuditingEntity.java
+++ b/generators/server/templates/src/main/java/package/domain/_AbstractAuditingEntity.java
@@ -56,7 +56,7 @@ public abstract class AbstractAuditingEntity implements Serializable {
     private String createdBy;
 
     @CreatedDate<% if (databaseType === 'sql') { %>
-    @Column(name = "created_date", nullable = false)<% } %><% if (databaseType === 'mongodb' || databaseType === 'couchbase') { %>
+    @Column(name = "created_date", nullable = false, updatable = false)<% } %><% if (databaseType === 'mongodb' || databaseType === 'couchbase') { %>
     @Field("created_date")<% } %>
     @JsonIgnore
     private Instant createdDate = Instant.now();


gmarziou:
Thanks. 
How is it handled by client code?

maivantan1992:
@gmarziou This is the server side it self. It wont allow client side to update create date.

gmarziou:
I understood it but I would like to ask whether the field in form is in read only or disabled and how an error triggered by server side is correctly displayed. It seems to me that your PR might not cover user experience.

gmarziou:
Even server side, maybe there's something to do in DTO mapping and probably a unit test would be helpful.

maivantan1992:
@gmarziou Firstly, I current use generator for microservice, mean no UI for front end now. I don't know if some parts of generator will generate front end which have field to input create date, but I believe that no reason to audit entity can allow to edit create date.
Second, about DTO mapping: It has many way to fix the problem, but I think the most easy way is just mark it as updatable = false (just like the current config in `create by` field - may be some missing when coding this part).
Third, the unit test: I wish I have time to dounit test for it. But current I am not. Hopefully, this is just simple change with the same config of the `create by `field, that would not effect must to logic of another. Btw, the current CI seen not stable now, I see it fail to build many time.

gmarziou:
I agree that our Travis build is currently less stable though there is one test that fails on gateway and deal with creation date. I did not look at it and I was hoping you could: 

UserServiceIntTest.testFindNotActivatedUsersByCreationDateBefore:142 

https://travis-ci.org/jhipster/generator-jhipster/jobs/310442421

My point is that some changes in your project might be trivial but it does not mean that they don't have an impact on other configurations, this is why we have all these sample projects built by Travis.

maivantan1992:
Yup, I will check about the unitest fail if have time. Let see if the configuration makes unit test fail.

deepu105:
@maivantan1992 do you plan to work further on this? there has been no activity for sometime here

maivantan1992:
@deepu105 I am current busy and not have time to check the unit test issue. May need some help for that part.

jdubois:
I agree on the idea here, but @gmarziou is right there is a bit more work to do before we can merge this - and as I don't think any of us as the time for this, I'm closing this PR.
@maivantan1992 don't hesitate to reopen if you have time!
I personally have doubts whether we should keep the audits parts in JHipster : it's complex and I'm not sure many people use it or understand it. But that's another discussion!

PR #6788, opened 2017-12-02
Generate git information when building the package with the prod profile

raphaelbrugier:
This displays the git commit in the /management/info endpoint

This feature had been previously introduced in #3574 then reverted

see also https://docs.spring.io/spring-boot/docs/current/reference/html/howto-build.html#howto-git-info

Two things to note:

1.  I have tested this with Maven and Gradle locally.
I am not sure if that's possible to add a test for this? 
Do you have integration tests - something to check the endpoints when a generated application running?

2. For Maven - I have added the `spring-boot-starter-parent` artifact back as the parent for the pom.
This is supposed to be the parent with default configuration for any Spring boot application, I'm not sure why you had removed it? 
The reason seems to be for the new JHipster dependency - but it's not clear to me if that was intentional to remove Spring boot as the parent artifact?
See https://github.com/jhipster/generator-jhipster/commit/c678c15e6ac517b2ed4a6079915fe011756c170e#diff-4ac12e0abae2ba3bd72788629c00c4abL23

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management-detail.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/_user-management-detail.component.html
@@ -24,10 +24,16 @@
         <dt><span jhiTranslate="userManagement.login">Login</span></dt>
         <dd>
             <span>{{user.login}}</span>
-            <span class="badge badge-danger" *ngIf="!user.activated"
-                  jhiTranslate="userManagement.deactivated">Deactivated</span>
-            <span class="badge badge-success" *ngIf="user.activated"
-                  jhiTranslate="userManagement.activated">Activated</span>
+            <jhi-boolean
+                [value]="user.activated"
+            <%_ if (enableTranslation) { _%>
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
1. depending if the endpoint is secured or not, otherwise, you can look at these lines: https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/05-run.sh#L9-L12

2. related to https://github.com/jhipster/jhipster-dependencies/blob/master/pom.xml#L6-L11 we depend on Brussels-SR5 as parent -> @erikkemperman is our expert on this part

raphaelbrugier:
Well,

It seems adding `spring-boot-starter-parent` as the parent has broken the tests.
I'm a little confused by the new dependencies management used by JHipster.

I don't see the pom you linked in my classpath. 
If you don't have spring-boot-parent as the parent artifact, how do you inherit from the spring-boot default?

antarus:
Jhipster uses a dependency pom to handle the versions used. see https://github.com/jhipster/jhipster-dependencies/blob/master/pom.xml

PR #6802, opened 2017-12-04
skipping only parts of log config when eureka isn't there

### LINKED ISSUE #6799: LoggingConfiguration depends on Eureka

Currently the LoggingConfiguration has @ConditionalOnProperty("eureka.client.enabled") annotation set which makes the configuration disabled without eureka. It would be good to have some parts enabled even if the registry is not active.

There is a generated test that tests that the async appender is set. When running the tests with eureka disabled (which seems reasonable) this test will fail as the appender is not set.

##### **Reproduce the error**

Run the tests with eureka.client.enabled: false

##### **Suggest a Fix**

I'll provide a PR soon that uses Autowired(required=false) for the eureka instance id in this case instead but keeps the rest of the configuration enabled.

##### **JHipster Version(s)**

4.7.0

fhopf:
Fix #6799


PierreBesson:
Sorry it's too late now for a careful review but it seems like an improvement over my solution. Which was made do that tests would pass.
Thank you @fhopf !
Will review tomorrow.

jdubois:
LGTM - congrats @fhopf for your first contribution!

Issue #6791, opened 2017-12-03
EagerRelationships should be pageable


agoncal:
##### **Overview of the issue**

When there is a ManyToMany relationship, JHipster generates `findAllWithEagerRelationships` methods in the repositories. These methods should be pageable.

##### **Motivation for or Use Case**

Imagine that you have a ManyToMany relationship between a `SponsorAgreement` and a `Contact`. JHipster will generate the following code:

```
public interface SponsorAgreementRepository extends JpaRepository<SponsorAgreement, Long> {
    @Query("select distinct sponsor_agreement from SponsorAgreement sponsor_agreement left join fetch sponsor_agreement.contactsOthers")
    List<SponsorAgreement> findAllWithEagerRelationships();
}
```
The problem is that the method `findAllWithEagerRelationships` returns a `List` and not a `Page`. Instead, it should generated as follow:

```
public interface SponsorAgreementRepository extends JpaRepository<SponsorAgreement, Long>, JpaSpecificationExecutor<SponsorAgreement> {
    @Query(value = "select distinct sponsor_agreement from SponsorAgreement sponsor_agreement left join fetch sponsor_agreement.contactsOthers",
    countQuery = "select count(distinct sponsor_agreement) from SponsorAgreement sponsor_agreement")
    Page<SponsorAgreement> findAllWithEagerRelationships(Pageable pageable);
}
```

Notice the `countQuery`. Spring Data [needs this to be able to count the number of items](https://codingexplained.com/coding/java/spring-framework/fetch-query-not-working-spring-data-jpa-pageable) and calculate the number of pages.

BTW, the methods in the repository are not used (not even tested) when you use pagination.

##### **JHipster Version(s)**

```
featurepageable@0.0.0 /Users/agoncal/Documents/Code/Temp/jhipster/featurepageable
└── generator-jhipster@4.11.1 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "org.agoncal.featurepageable"
    },
    "jhipsterVersion": "4.11.1",
    "baseName": "featurepageable",
    "packageName": "org.agoncal.featurepageable",
    "packageFolder": "org/agoncal/featurepageable",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
[...]

PR #6805, opened 2017-12-04
Use Pageable in many-to-many relationship

ramzimaalej:

Fix for this: https://github.com/jhipster/generator-jhipster/issues/6791

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
@@ -43,11 +43,16 @@
         return <%= entityListToDtoListReference %>(<%= entityInstancePlural %>);<% } else { %>
         return <%= entityInstance %>Repository.<% if (fieldsContainOwnerManyToMany) { %>findAllWithEagerRelationships<% } else { %>findAll<% } %>();<% } %>
     <% } if (pagination !== 'no') { %>
-    public ResponseEntity<List<<%= instanceType %>>> getAll<%= entityClassPlural %>(Pageable pageable<% if (fieldsContainNoOwnerOneToOne) { %>, @RequestParam(required = false) String filter<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
-        log.debug("REST request to get a page of <%= entityClassPlural %>");<% if (viaService) { %>
-        Page<<%= instanceType %>> page = <%= entityInstance %>Service.findAll(pageable);<% } else { %>
-        Page<<%= entityClass %>> page = <%= entityInstance %>Repository.findAll(pageable);<% } %>
-        HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, "/api/<%= entityApiUrl %>");
+    public ResponseEntity<List<<%= instanceType %>>> getAll<%= entityClassPlural %>(Pageable pageable<% if (fieldsContainNoOwnerOneToOne) { %>, @RequestParam(required = false) String filter<% } %><% if (fieldsContainOwnerManyToMany === true) { %>, @RequestParam(required = false, defaultValue = false) boolean eagerload<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
+        log.debug("REST request to get a page of <%= entityClassPlural %>");<% if (viaService) { %><% if (fieldsContainOwnerManyToMany === true) { %>
+        Page<<%= instanceType %>> page = eagerload ? <%= entityInstance %>Service.findAllWithEagerRelationships(pageable) : <%= entityInstance %>Service.findAll(pageable);
+        HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, String.format("/api/<%= entityApiUrl %>?eagerload=%b", eagerload));<% } else { %>
+        Page<<%= entityClass %>> page = <%= entityInstance %>Service.findAll(pageable);
+        HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, "/api/<%= entityApiUrl %>");<% } %><% } else { %><% if (fieldsContainOwnerManyToMany === true) { %>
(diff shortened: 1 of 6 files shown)


generators/entity-server/templates/src/main/java/package/repository/_EntityRepository.java
```diff
@@ -76,9 +78,10 @@
     List<<%= entityClass %>> findBy<%= relationships[idx].relationshipNameCapitalized %>IsCurrentUser();
     <%_ } } _%>
     <%_ if (fieldsContainOwnerManyToMany === true) { _%>
-    @Query("select distinct <%= entityTableName %> from <%= entityClass %> <%= entityTableName %><% for (idx in relationships) {
-    if (relationships[idx].relationshipType === 'many-to-many' && relationships[idx].ownerSide === true) { %> left join fetch <%=entityTableName%>.<%=relationships[idx].relationshipFieldNamePlural%><%} }%>")
-    List<<%=entityClass%>> findAllWithEagerRelationships();
+    @Query(value = "select distinct <%= entityTableName %> from <%= entityClass %> <%= entityTableName %><% for (idx in relationships) {
+    if (relationships[idx].relationshipType === 'many-to-many' && relationships[idx].ownerSide === true) { %> left join fetch <%=entityTableName%>.<%=relationships[idx].relationshipFieldNamePlural%>",
+        countQuery = "select count(distinct <%= entityTableName %>) from <%= entityClass %> <%= entityTableName %><%} }%>")
+    Page<<%=entityClass%>> findAllWithEagerRelationships(Pageable pageable);
```
ctamisier: need to fix the call of this function as well in `_EntityServiceImpl.java` and `get_all_template.ejs`, and passing `null` (if I'm not wrong)

ramzimaalej: yes, I marked this PR as WIP for the time being.

deepu105:
@ramzimaalej please let us know when its complete

ramzimaalej:
sure @deepu105, sorry I have been quite busy lately.

ramzimaalej:
@deepu105 @jdubois Before I take this further, I would like to have your opinion regarding the web layer. 

Currently, I have overloaded findAllWithEagerRelationships to take Pageable. However, for the web, I don't think the default behavior should be loading eager relationships. So, I would leave it to the developer to decide to use this method or not. what do you think?

jdubois:
By default the user could load a Pageable with the default pagination size - then it's up to the developer to either increase the pagination size at the Spring Data level, or request for more entities in his REST request

ramzimaalej:
I have exposed the method at the web layer and left it up to the developer to decide to load eager relationship or not. I think this is done, please review and let me know if there is anything that should addressed.

ramzimaalej:
cc @deepu105

deepu105:
@ramzimaalej can you remove WIP from title if this is ready for review

deepu105:
@jhipster/developers can someone review and test this?

generators/entity-server/templates/src/main/java/package/service/impl/_EntityServiceImpl.java
```diff
@@ -99,6 +97,17 @@
...
+
+    /**
+     * Get all the <%= entityClass %> with eager load of many-to-many relationships.
+     *
+     * @return the list of entities
+     */
+    public Page<<%= instanceType %>> findAllWithEagerRelationships(Pageable pageable) {
+        return <%= entityInstance %>Repository.findAllWithEagerRelationships(pageable);
+    }
+    <% } %>
```
ruddell: If the entity uses a DTO, then it needs to be mapped.  Replace the semicolon on line 108 with:
```
<% if (dto !== 'mapstruct') { %>;<% } else { %>
            .map(<%= entityToDtoReference %>);<% } %>
```

ruddell:
It compiles after making the change mentioned above, but these methods still aren't used or tested anywhere.

deepu105:
@ramzimaalej can you add/generate some tests for the methods?

ramzimaalej:
@ruddell I will make the change, and generate tests for it.

ramzimaalej:
@ruddell I have addressed your feedback.

cc @deepu105 @jdubois This should be ready to be merged if it looks ok to u.

generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
```diff
@@ -36,19 +37,24 @@
...
-        log.debug("REST request to get a page of <%= entityClassPlural %>");<% if (viaService) { %>
-        Page<<%= instanceType %>> page = <%= entityInstance %>Service.findAll(pageable);<% } else { %>
-        Page<<%= entityClass %>> page = <%= entityInstance %>Repository.findAll(pageable);<% } %>
-        HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, "/api/<%= entityApiUrl %>");
-        return new ResponseEntity<>(<% if (!viaService && dto === 'mapstruct') { %><%= entityListToDtoListReference %>(page.getContent())<% } else { %>page.getContent()<% } %>, headers, HttpStatus.OK);
+    public ResponseEntity<List<<%= instanceType %>>> getAll<%= entityClassPlural %>(Pageable pageable<% if (fieldsContainNoOwnerOneToOne) { %>, @RequestParam(required = false) String filter<% } %><% if (fieldsContainOwnerManyToMany) { %>, @RequestParam(required = false, defaultValue = "false") boolean eagerload<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
```
deepu105: Can we use the <%_ syntax where possible to make this more readable? Plz note that it should be used only if the generated code from it will be in newline. For inline conditions <% is good

ruddell:
When the tests pass it's good for merge.  I went ahead and updated the templates in this PR to use `<%_`

jdubois:
Thanks @ramzimaalej !

Issue #6796, opened 2017-12-04
Improve spring cache implementation


antarus:
Hello,
 I would like to improve the cache support. For that I would like to add 'spring-boot-starter-cache' to dependencies, and give the possibility to the user to choose a cache system.
But before starting I have some questions.
Which default implementation do you want to use?
Spring cache offers the following services:
    - Generic
    - JCache (JSR-107) (EhCache 3, Hazelcast, Infinispan, etc.)
    - EhCache 2.x
    - Hazelcast
    - Infinispan
    - Couchbase
    - Redis
    - Caffeine
    - Guava (deprecated)
    - Simple
I have to add questions to configure the different implementations. if so which ones in priority?
It must be checked that the selected implementations do not collide with the selected hibernate cache.

I will take again the modifications made by erikkemperman https://github.com/jhipster/generator-jhipster/pull/6666

thanks

jdubois:
- In fact we already support it, have a look at http://www.jhipster.tech/using-cache/
- What is bad, is that for historic reasons this is using a "hibernateCache" variable, and this is linked to our "2nd level Hibernate cache" questions, so it only works with SQL databases

-> we should keep what we already have, but refactor it

antarus:
yes i know the cache is manage at some level in jhipster. The title of the issue should have been 'improve' and not add:)
I will try to look in more detail how refactor the 2nd level cache with spring-boot-starter-cache

antarus:
I look at the impact of refactoring the variable 'hibernateCache', it's a big job ..
I propose to create a variable 'cache'.
If the 'hibernatecache' variable is set, then we set its value to the 'cache' variable.
If there is no database, or if the value of 'hibernatecache' is 'no', then we add a question. Would you like to use spring cache? y / n
Then, if the cache variable is filled (not 'no'), we copy the class CacheConfiguration.java and configure it.
I also propose to use the 'cache' variable instead of 'hibernateCache' for the gateway case as hazelcast is mandatory

This solution has the merit of not modifying everything, we remain consistent on the cache configuration for hibernate and we have another variable for the cache spring.

What do you think ?

PierreBesson:
@antarus I think it's a very good Idea ! Very nice for no-db apps and clearer when your app actually does'nt have hibernate.

antarus:
@erikkemperman @PierreBesson   
I made the modification of the cache variable, add the dependency spring start cache and start the configuration of the cache.
[...]

PR #6811, opened 2017-12-05
Improve spring cache implementation

antarus:
init issue https://github.com/jhipster/generator-jhipster/issues/6796

simple rename hibernacheCache to cacheProvider
add an option for use spring cache and configure hibernate cache
If spring cache is selected, then we now have a question for configurate hibernate cache.
add variable enableHibernateCache

**TODO**
- [x] remove old variable hibernateCache in .yo-rc.json file. How?
- [x] replace `(hibernateCache !== 'no' && databaseType === 'sql')` by `enableHibernateCache`


generators/app/index.js
```diff
@@ -103,6 +103,13 @@ module.exports = class extends BaseGenerator {
             type: String
         });
 
+        // This adds support for a `--cache-impl` flag
```
deepu105: I dont see a need to provide this via flags, the flags are only for special case handlings, so please remove this or provide a reasonable explanation on why this is needed

generators/client/index.js
```diff
@@ -91,6 +91,13 @@ module.exports = class extends BaseGenerator {
             type: String
         });
 
+        // This adds support for a `--cache-impl` flag
+        this.option('cache-impl', {
```
deepu105: use `cache-provider` as the flag name

generators/client/index.js
```diff
@@ -253,6 +260,7 @@ module.exports = class extends BaseGenerator {
                     this.authenticationType && this.config.set('authenticationType', this.authenticationType);
                     this.uaaBaseName && this.config.set('uaaBaseName', this.uaaBaseName);
                     this.hibernateCache && this.config.set('hibernateCache', this.hibernateCache);
+                    this.cacheImplementation && this.config.set('cacheImplementation', this.cacheImplementation);
```
deepu105: use `cacheProvider` as the variable name everywhere please

generators/client/index.js
```diff
@@ -271,6 +279,9 @@ module.exports = class extends BaseGenerator {
                 if (this.configOptions.hibernateCache) {
                     this.hibernateCache = this.configOptions.hibernateCache;
                 }
+                if (this.configOptions.cacheImplementation) {
+                    this.cacheImplementation = this.configOptions.cacheImplementation;
```
deepu105: I guess its much easier to just replace `hibernateCache` variable everywhere with `cacheProvider` in one shot rather having a 2 variable solution. There will be more files changed but thats fine, i'll review it

generators/server/files.js
```diff
@@ -397,7 +397,7 @@ function writeFiles() {
 
             this.template(`${SERVER_MAIN_SRC_DIR}package/config/_package-info.java`, `${javaDir}config/package-info.java`);
             this.template(`${SERVER_MAIN_SRC_DIR}package/config/_AsyncConfiguration.java`, `${javaDir}config/AsyncConfiguration.java`);
-            if (this.hibernateCache === 'ehcache' || this.hibernateCache === 'hazelcast' || this.hibernateCache === 'infinispan' || this.clusteredHttpSession === 'hazelcast' || this.applicationType === 'gateway') {
+            if (this.cacheImplementation !== 'no' || this.clusteredHttpSession === 'hazelcast') {
```
deepu105: why are options removed from this?

generators/server/prompts.js
```diff
@@ -369,6 +369,31 @@ function askForServerSideOpts(meta) {
             ],
             default: (applicationType === 'microservice' || applicationType === 'uaa') ? 1 : 0
         },
+        {
+            when: response => ((response.databaseType === 'no' || response.hibernateCache === 'no') && applicationType !== 'gateway'),
+            type: 'list',
+            name: 'cacheImplementation',
+            message: 'Do you want to use spring cache?',
+            choices: [
```
deepu105: Lets not try to add new cache providers now, first iteration should be to support ehcache, hazelcast and infinispan and later more can be added

generators/server/templates/_build.gradle
```diff
@@ -212,8 +212,13 @@ repositories {
 
 dependencies {
     compile "io.github.jhipster:jhipster"
+    <%_ if (cacheImplementation !== 'no') { _%>
+    // TODO move to jhipster-dependencies
```
deepu105: we cant have TODO in generated code

antarus:
@deepu105 rename do :)

generators/server/prompts.js
```diff
@@ -156,6 +156,32 @@ function askForServerSideOpts(meta) {
...
+                {
+                    value: 'infinispan',
+                    name: '[BETA] Yes, with Infinispan (hybrid cache, for multiple nodes)'
+                },
+                {
+                    value: 'no',
+                    name: 'No'
+                }
+            ],
+            default: 'no'
```
deepu105: To retain current behaviour do the below 
`default: (applicationType === 'microservice' || applicationType === 'uaa'| ) ? 1 : 0`

generators/server/prompts.js
```diff
@@ -156,6 +156,32 @@ function askForServerSideOpts(meta) {
...
+                    value: 'hazelcast',
+                    name: 'Yes, with HazelCast implementation(distributed cache, for multiple nodes)'
+                },
+                {
+                    value: 'infinispan',
+                    name: '[BETA] Yes, with Infinispan (hybrid cache, for multiple nodes)'
+                },
+                {
+                    value: 'no',
+                    name: 'No'
```
deepu105: change to `No (Hibernate L2 cache won't be available as well)`

generators/cloudfoundry/index.js
```diff
@@ -35,7 +35,7 @@ module.exports = class extends BaseGenerator {
         this.buildTool = this.config.get('buildTool');
         this.packageName = this.config.get('packageName');
         this.packageFolder = this.config.get('packageFolder');
-        this.hibernateCache = this.config.get('hibernateCache');
+        this.cacheProvider = this.config.get('cacheProvider');
```
deepu105: This is still missing the || part for backward compatibility

generators/entity/index.js
```diff
@@ -113,7 +113,7 @@ module.exports = class extends BaseGenerator {
                 context.applicationType = this.config.get('applicationType');
                 context.packageFolder = this.config.get('packageFolder');
                 context.authenticationType = this.config.get('authenticationType');
-                context.hibernateCache = this.config.get('hibernateCache');
+                context.cacheProvider = this.config.get('cacheProvider');
```
deepu105: Missing backward compatibility or part

generators/generator-base.js
```diff
@@ -2312,7 +2312,8 @@ module.exports = class extends PrivateBase {
         generator.databaseType = generator.getDBTypeFromDBValue(context.options.db) || context.configOptions.databaseType || context.config.get('databaseType');
         generator.enableSocialSignIn = context.options.social || context.config.get('enableSocialSignIn');
         generator.searchEngine = context.options['search-engine'] || context.config.get('searchEngine');
-        generator.hibernateCache = context.options['hb-cache'] || context.config.get('hibernateCache');
+        generator.cacheProvider = context.options['cache-provider'] || context.config.get('cacheProvider');
```
deepu105: Missing backward compatibility or part

antarus:
I will look at the mistakes. I think it's related to the enableHibernateCache variable which is not backward compatible.
I noticed another problem, if we have the old variable 'hibernateCache' in the yo.rc file, and we regenerate the application, it still exists once the file has been saved. What is the best way to remove it?

deepu105:
Since we use the yeoman config, see if it has a remove method else its not
a huge problem to leave it as we will ignore it anyway if the other
variable is present

antarus:
Simply, `this.config.delete('hibernateCache')` :D

I made the changes, I PR tonight. It's boring to work from the office without internet connection: D

antarus:
@deepu105 travis ok ;)

generators/server/templates/src/main/java/package/domain/_Authority.java
```diff
@@ -18,7 +18,7 @@
 -%>
 package <%=packageName%>.domain;
 
-<% if (hibernateCache !== 'no' && databaseType === 'sql') { %>import org.hibernate.annotations.Cache;
+<% if (cacheProvider !== 'no' && databaseType === 'sql') { %>import org.hibernate.annotations.Cache;
```
deepu105: For all the hibernate related parts the check should be with the new enableHibernateCache variable and we can drop the cache and db check, this will apply to all domain objects/entities

antarus: I agree with you. I didn't want to change everything at the risk of making a mistake: D
In addition, I'm a little afraid of merge conflict given the number of code impacted currently.

I will not be able to try to do it until Monday

deepu105: thats ok its not urgent anyway

antarus:
add enableHibernateCache instead of cacheProvider !== 'no' && databaseType === 'sql'

PR #6812, opened 2017-12-05
Mono repo

erikkemperman:
As proposed on the mailing list, this merges the jhipster-dependencies and jhipster-server repositories into the generator as subdirectories. It includes all of the history intact, to preserve stats and praise/blame. The hashtag references in commit messages were expanded to explicitly point to the appropriate PRs, and the root commits of both repositories were given parents to form a linear overall history.

If this PR is adopted, I think the following two should also be merged, to make explicit the former repositories for jhipster-server and jhipster-dependencies are now discontinued:
https://github.com/jhipster/jhipster/pull/36
https://github.com/jhipster/jhipster-dependencies/pull/25


pascalgrimaud:
I'm not sure it's a good idea to delete files for the other repo.
We didn't do it for https://github.com/jhipster/jhipster-docker

erikkemperman:
@pascalgrimaud Well they stil exist in those repositories’ respective histories of course, but yes maybe that’s a little too much... I will amend the other two PRs to only change the README.

jdubois:
- This is of course incredibly awesome
- Tests needs to be done to be sure it all works well, as this is a bit "hacky"
- As we are currently closing a lot of bugs ->I think this will be merged after the next release, once everything is cleaned up, which should take a few days. This means there might be a few more commits on both jhipster/jhipster and jhipster/jhipster-dependencies but I don't expect much, they should be pretty stable now

erikkemperman:
@jdubois Thanks! Yes, that’s why I requested three reviews :-)

No problem to include more commits from those other repos, or rebase this PR on latest master here. The hard part was figuring this out but now it should be easy to repeat. Just let me know when you want it!

erikkemperman:
FYI I just rebased this PR onto the latest master of the generator, no changes were committed to jhipster-dependencies or the server library in the mean time.

.travis.yml
```diff
@@ -63,6 +63,8 @@ before_install:
   - yarn -v
   - yarn global add yo bower gulp-cli
 install:
+  - cd jhipster-dependencies/ && mvn clean install -Dgpg.skip=true && cd ..
```
deepu105: maybe we should use our maven wrapper here as well

deepu105:
@jdubois can we merge this?

jdubois:
Just a few more comments:

- I just released a new "jhipster-dependencies" version, so this would need to be updated (wait until tomorrow, I'm testing stuff at the moment - I found a strange bug!!)
- If we do that, shouldn't we also add https://github.com/jhipster/ng-jhipster and https://github.com/jhipster/react-jhipster ?
- To be totally honest I'm not 100% sure this is a good idea, but that would ease our development efforts, and be better for our GitHub stats

erikkemperman:
Well, personally I have a couple of things that would be a lot easier to do in a single repository -- moving things out of the generator and into the server-lib, would be much easier if it's not three steps and waiting for an otherwise pretty meaningless release in between... But if you guys feel this is a bad idea, let's think about it carefully. Curious how do the other devs feel?

deepu105:
Erik, my comment was for the npm modules Julien talked about for the server
lib and deps I guess the decision was already made and hence lets just go
with it


Thanks & Regards,
Deepu

erikkemperman:
@deepu105 Just want to be sure no-one is swallowing this with reservations...

jdubois:
@deepu105 @erikkemperman indeed, but that doesn't sound very logical - why is the server lib in the mono repo, and not the client libs?

pascalgrimaud:
What we can do for testing is: in the main generator-jhipster:

```
- git clone https://github.com/jhipster/jhipster-dependencies
- cd jhipster-dependencies/ && mvn clean install -Dgpg.skip=true && cd ..
- git clone https://github.com/jhipster/jhipster-server
- cd jhipster-server/ && mvn clean install -Dgpg.skip=true && cd ..
```

Putting everything in the main generator-jhipster would change the process for tag / release, because it will mix the tag for generator-jhipster, the tag for jhipster-dependencies and for jhipster-server

jdubois:
Hang on, that's a huge issue, thanks @pascalgrimaud I didn't thought of that! Yes we need the tags, and they should follow each project's version number. How come we didn't think of that before? I'm doing too many things at the same time

erikkemperman:
Well, we need at least of a "server" and a "dependencies" tag to keep track of changes that would require a dependencies or server release to go along with the next generator release. 

I'm not against adding client libraries to this repository, and it would be easy for me to add them now that I've figured things out.

My first concern here, however, was being able to test version upgrades in the BOM against the test suite in the generator, and to begin to resolve a mutual co-dependency issue between the BOM and the server lib. So that's why there are those two modules in this PR at the moment.

erikkemperman:
@pascalgrimaud  Ah, sorry, of course you mean "release tags", I was confused with labels on PRs. Never mind earlier remark :-)

So, in that case, it's either going all the way with linked releases of the several modules, like for instance spring-boot and many other large projects, or else we are back to separate repositories and need to figure out a way to test properly. 

The suggestion above of cloning the master-branches of jhipster-dependencies and jhipster server lib before testing the generator is a big improvement, to be sure. But it still doesn't help with reviewing PRs on those repositories, does it, until after they're merged?

jdubois:
The tag issue that @pascalgrimaud mentioned is a huge issue. So now I'm a bit reluctant to merge this. Besides this:

- In the end we are going to have all client-side libraries here (otherwise it wouldn't be consistent), and as @deepu105 says it's better to have those in separate projects. I also don't know how that will work with NPM releases -> usually it's one repo per NPM module, and even if that works technically, it will be surprising for people.
- For the Java part, indeed the server lib is linked to the dependencies project -> maybe only those 2 should be mixed together, but personally that doesn't give me a huge issue. Releasing each of them is easy, I really have no trouble here. Maybe with continuous integration in Travis?

PierreBesson:
Personally I don't really see the **huge** issue. I don't really see why we need specific git tags for jhipster-dependencies. Anyway people will be able to see the source code with maven source packages.
With the monorepo we can release new dependencies version only when releasing new version of the generator. The big advantage is that it keeps things in sync. It is very annoying to have to wait for a PR to jhispter-dependencies so that your generator PR can pass tests. In the current setup, I have noticed is that we always blindly trust PRs to dependencies and then release and test the generator PR.

TL;DR; The monorepo is good because it reduces the amount of work that the reviewer needs to do.

pascalgrimaud:
I'm not against this, but I'm not in favor either... For now, I don't know what is the best for us. There are pros and cons.

I wonder how we should manage the release version

**Solution 1**: different versions
- generator-jhipster: 4.13.0
- jhipster-dependencies: 0.1.4
- jhipster-server: 1.3.0

**Solution 2**: same version to align
- generator-jhipster: 4.13.0
- jhipster-dependencies: 4.13.0
- jhipster-server: 4.13.0

In all cases, if we decide to do a mono repo and to put here `jhipster-dependencies` and `jhipster-server`, to be consistent, maybe we should do the same for:
- ng-jhipster
- jhipster-core
- jhipster-registry
- jhipster-devbox

What is the reason to exclude the other projects ?

deepu105:
I agreed to mono repo half-heartedly for server lib and deps just because
of the development ease argument, but that argument doesn't apply to
ng-jhipster, react-jhipster, and jhipster-core, as developing and testing
them as separate or same repo is the same process and merging PRs on them
in same repo or separate is the same as well, I don't see any added value
in doing them in same repo and hence is a BIG NO for me for those.

I'm not sure if tags are important for stuff released to maven central but
they are important for NPM and I wouldn't want the lib versions being
upgraded unnecessarily during each release in any other lib.
I agree that having the dependency project and server lib together might
have value as they are very much dependent but having everything related to
JHipster in the same repo is a recipe for disaster, I could see many issues
like

   - managing tags will be difficult and confusing
   - releasing stuff to NPM will be difficult
   - it will be difficult for people who work on a particular project
   mostly, like Matheu for example, as now you have to work on this huge repo
   with everything, means more merge conflicts to solve, more code being
   indexed and so on
   - more issues in the generator, more PR to keep track of and IMO
   that's more effort for maintainers in general as you loose context easily,
   even with labels and stuff is not as intuitive as having separate trackers

Now for the problem of testing why can't we setup Travis tests so that each
PR on the server lib will clone the current generator master, install the
server lib locally and run the Travis test suite from the generator or a
subset, I believe its not very difficult to do with some bash scripts
(pascal plz confirm)
For the generator we can do the same clone server lib, o mvn install then
run Travis suite


I would suggest merging the server lib and dependency project first and see
if the above testing approach works first before merging everything into a
huge monolith


Thanks & Regards,
Deepu

gmarziou:
> I would suggest merging the server lib and dependency project first and see
if the above testing approach works first before merging everything into a
huge monolith

This sounds very reasonable to me.

pascalgrimaud:
I agree with ALL the points you mentionned Deepu.

And I confirm for the tests. In fact, it's already done with hipster-labs -> it depends on nothing, it can test what we want and it's easy to do the same for each repo

pascalgrimaud:
In all case, the final decision is for Julien :-)

erikkemperman:
All right, then how about I revisit the single repo PR I made a couple of weeks (months?) ago for the server lib and dependencies? I didn’t do it with history at the time, but I could try that this time if you like.

jdubois:
I'm sorry but the tag issue is really important for me. It's very useful to be able to tag each repo - not only for releases, but also for everyone (not just us) to understand what we changed.
If the idea is just to have the dependencies and server lib as sub-directories of the main repo, we can also do sub-modules - I think it would basically be the same result, and we can have the tags.

If we go with the "solution 2" from @pascalgrimaud then it is not satisfying for me:

- We would release lots of new versions of the server lib and the dependencies, when most of the time I actually don't have to do those releases. This means more work for me, and also for end-users (if there's a new release, they will at least have a look at it, for nothing)
- And indeed, it doesn't make sense to do it for the server lib and not for the client libs. This is also confusing for end users

Let me study/ask people to better understand how they do it at Google, as I know they have a single huge repo (but it doesn't really work the same way as Git, and there are probably tricks with their in-house IDE, that of course we don't have)

PierreBesson:
Personnaly I'm not too concerned about this "tag issue". Those dependencies project are tied to the templates so it makes sense to keep them in sync through git history in the same repo. So for me the most important thing is not too know what changes were introduced in jhipster-server on version X.Y but rather what PR introduced changes to the library's code.

IMO **jhipster-server and jhipster-dependencies should be treated as a part of JHipster's code and not simply external dependencies**, they are very tied to the project. Other projects like ng-jhipster, JHipster core can be kept in separate repo if it makes more sense for them.

@jdubois On the topic of managing releases of the maven libs @pgrimaud and I have some ideas on how to automate this, maybe with a `release` branch on which we set Travis to trigger auto deploys (and version increment) to maven central.

erikkemperman:
About those tags, @deepu105, do they carry a technical significance for NPM? For github (and how things like maven-release-plugin treat them) they’re basically just given names for particular points in history right?

So we could still have separate releases of these modules and just prefix the tags corresponding to jhipster-dependencies and hipster-server releases?

Or, like I said, I could also attempt to make a single repo with deps and lib separate from the generator. Whatever you guys prefer, let me know.

jdubois:
OK I understand most of you vote for the single repo:

- For the tag issue, let's just tag everything with the JHipster release. It's easier for everyone to understand. Please note that it can take up to 2 hours for Maven Central to synchronize, so that means the Maven dependencies must be deployed 2 hours before generator-jhipster - which means a patch release which doesn't impact the Maven dependencies will take a bit more time.
- There's also an issue if we have other server-side languages like .NET, but at the moment those are far in the future, let's not worry for something that doesn't exist yet
- I'm quite reluctant about this, but 1) I respect everybody's opinion 2) This is better for our stats and 3) let's just give it a try, we can still rollback later

-> So let's do it

@erikkemperman as you need to upgrade your PR (we probably did a few changes to both repos, my proposal is that we block a day for this. So we freeze the repos on that day, so you have time to sync your PR, and we can merge them quietly.
-> Which day would suit you best?

Tcharl:
@PierreBesson on the maven artifact release part, I implemented continuous release with Travis and maven-release-plugin with the [PR](https://github.com/jhipster/generator-jhipster/pull/6739).
The travis file of my [POC sample](https://github.com/Tcharl/test-travis-release/blob/master/.travis.yml) automatically tags the repo, increment versions and deploys to maven central staging when a commit is done on master, but it can easily changed for a specific branch (see l26).

erikkemperman:
@jdubois It looks like no commits were made to those two repositories since I updated this PR. So that should be alright. There have been a few commits to the generator but as this PR is almost exclusively in separate directories there are no merge conflicts. From my POV this is good to go!

I do think it would be good to add a prefix to the maven artifacts release tags, though, for bookkeeping purposes. If you use maven-release-plugin it’s as simple as adding a property I believe.

erikkemperman:
Oh sorry, I was looking at an old tab, there’s a commit I need to add after all. OK, I could do that this afternoon?

jdubois:
OK let's freeze the repos, then!
And I'll merge at the end of the day (if it's not too late, Christmas holidays are coming!)

erikkemperman:
@jdubois Since it was just a single commit, I cherry-picked that onto this PR branch, instead of replaying and rewriting all the histories again and grafting all over. Lot easier that way, less chance of mistakes. Hope that's all right? If so, all done!

erikkemperman:
And a quick reminder, I've made two PRs to those other repositories in case this is merged, making the README point out that work is no longer being done on those:
https://github.com/jhipster/jhipster-dependencies/pull/25
https://github.com/jhipster/jhipster/pull/36

There are a few open PRs on those repositories, some of mine which I will adapt/move/close ASAP. The remainder will have to be resubmitted to the new location though... But if you like I can try to run the same branch-filter magic on those to transplant them automatically?

deepu105:
Julien,

To be fair I don't think everyone agreed (I at least counted me, you,
Pascal, Gael against and only Erik and Pierre for), so let's take a vote
first and see how many are for and against


I vote against mono repo


Thanks & Regards,
Deepu

erikkemperman:
Hmm I’m not sure what happened, but I assumed there had been an actual vote? When Julien asked me to sync this PR? Didn’t mean to push anything through against anyone’s wishes!

jdubois:
Well, I just read the messages and I thought everybody agreed, there wasn't a vote, indeed! OK I'll do a vote on the core team mailing list, as we are the ones who will get the biggest impact.

erikkemperman:
Voted on the mailing list, most of you don’t like the idea so I’m closing this PR.

I still believe we have a problem with untested patches to BOM and server lib breaking the generator though, so I’m looking forward to your ideas on that!

pascalgrimaud:
> I still believe we have a problem with untested patches to BOM and server lib breaking the generator though, so I’m looking forward to your ideas on that!

I already have the solution in mind but didn't start the work yet, because we didn't totally decide what the way to go.

jdubois:
Sure, and I'd like to test the same idea with sub-modules.

Anyway, I'm not totally happy with this, but it's the holidays: merry Christmas to everyone, let's work on this again in 2018!

erikkemperman:
Git submodules, in my experience, are a pain in the derrière:-)

PR #6822, opened 2017-12-06
Upgrade libraries for angular 5

sendilkumarn:


generators/client/templates/angular/webpack/_webpack.dev.js
```diff
@@ -123,6 +121,11 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
         new webpack.WatchIgnorePlugin([
             utils.root('src/test'),
         ]),
+        new AngularCompilerPlugin({
```
deepu105: why are we doing aot in dev? I dont think AOT is suitable for incremental builds in yarn start

deepu105: also its slower than current setup which is bad in dev as it is already quite slow, we dont want to make it slower

deepu105: I just tried on a sample project from our Travis build and its initial build and yarn start is very slow(~65 sec even in my laptop, so it will be even slower on normal machines) compared to around 30 secs on non-AOT build. 

Also when I make changes to a project the incremental build is taking similar time, as I think AOT recompiles everything, which is very bad for dev, so lets not do this as I dont see any benefit of doing AOT in dev

AOT dev
image


Normal dev
image

sendilkumarn: Are they same ? 🤔 
screen shot 2017-12-06 at 19 41 37
screen shot 2017-12-06 at 19 41 41

For me it takes ~22s for dev all the time 🕐  and with loaded entities around 30-40s dev build and it will not give you an AOT output, if I am correct. It will produce a dev build only. 

(This is somewhat similar to what angular/cli does, for both prod and dev they use `@ngtools/webpack` ) 

This needs further investigation, I will revert this

sendilkumarn: > even in my laptop,

😏  (don't they have a sarcasm smiley here)

deepu105: He he, I'm just being technical, as @mraible would vouch for, being an octacore Xeon, Precision runs nodeJs stuff much faster than macs :smile: we could benchmark and see if you dont believe me :wink: 
Anyway I test with the default travis build and 15 entities.

deepu105: Btw in your PR when i tried I saw ngfactory stuff being built so not sure if it doesnt do that

deepu105: see
image

PR #6823, opened 2017-12-06
Open collective appeal update

deepu105:

@jdubois are you ok with this and the wordings?

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/home/_home.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/home/_home.component.html
@@ -55,7 +55,10 @@
         </ul>
 
         <p>
-            <span jhiTranslate="home.like">If you like JHipster, don't forget to give us a star on</span> <a href="https://github.com/jhipster/generator-jhipster" target="_blank" rel="noopener" jhiTranslate="home.github">GitHub</a>!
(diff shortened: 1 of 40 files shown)


jdubois:
I'm worried we push this a bit too far: people will then need to delete this from their app, and it's scattered in several files. I know we already do this, but it's mostly for pushing the documentation, so the goal is to help people.

deepu105:
I thought since we already add the github liking part this was ok as well, I also felt that additing i18n for this and the github liking part might be too much as people would have to delete it anyway. May be we should translate the the last paragraph on home page so that people just can delete it from home page. Wdyt?

deepu105:
I meant we shouldn't translate the last paragraph

jdubois:
Yes, something simple & easy to delete, as this is what people will eventually do.
And next to the documentation -> they will also delete the link to the documentation, so that means they have something very useful (the documentation), and in exchange they have those few lines which are easy to delete. That's a good deal I think.
Anyway, don't worry too much: for me, the voting system is what will make OpenCollective take off. We already have enough advertisement for it, people know about it, it's just a matter of giving them something good in exchange. Yeah, that's not really a "donation", then :-)

deepu105:
I guess you are right, lets not push it too far

PR #6827, opened 2017-12-07
Enable i18n cache busting

thmsltmnn:
ng-jhipster now supports the "buildTimestamp" config parameter to enable cache busting when loading translation files. It is not used yet in generator-jhipster, this pull request should fix this.

During development, my solution works.

HOWEVER, when compiling the app with AOT, the string replacement does not work (or happens to late), so buildTimestamp always remains at the predefined value of '0' set in app.constants.ts.

If anyone knows how to fix this, help would be very much appreciated.


generators/client/templates/angular/src/main/webapp/app/shared/_shared-libs.module.ts
```diff
@@ -23,6 +23,7 @@ import { NgbModule } from '@ng-bootstrap/ng-bootstrap';
 import { NgJhipsterModule } from 'ng-jhipster';
 import { InfiniteScrollModule } from 'ngx-infinite-scroll';
 import { CookieModule } from 'ngx-cookie';
+import {BUILD_TIMESTAMP} from '../app.constants';
```
deepu105: spacing inside braces

sendilkumarn:
I will look into it later today or early tomorrow 👍

sendilkumarn:
I think this is tricky, I confirm the issue. Can we try this with a simplified repo and check. I dont know whether the issue is with angular or ng-jhipster.

sendilkumarn:
@thmsltmnn can we just use normal versioning with plain string here.

deepu105:
It doesn't make any sense if the replacement for the other variables works how come this alone doesn't?

thmsltmnn:
@sendilkumarn setting the version number manually as "buildTimestamp" works, and this is the workaround I use right now since it became possible to use this feature in ng-jhipster. This is not optimal, because it is another file you have to modify when increasing the version of your app, in addition to pom.xml and package.json.

@deepu105 I think the replacement works, but none of the other replaced variables goes into an NgModule configuration parameter, which seems to be treated differently by the Angular compiler.

deepu105:
@thmsltmnn this issue should be solved by this https://github.com/jhipster/generator-jhipster/commit/a475422f52de85688b90bc7d63714852af1d8391
Can you try to add the buildTimestamp in the define section as I have done and try again

deepu105:
@thmsltmnn I have updated your PR please test and confirm if it works with AOT

thmsltmnn:
@deepu105 I tried it with a freshly generated JHipster 4.13.0 app: https://github.com/thmsltmnn/timely/commit/1eeee30477278846e668e0ed8d056589319cb48e

Sadly, it did not work. With AOT, the URL to request i18n files was http://localhost:8080/i18n/en.json?buildTimestamp=null

deepu105:
@thmsltmnn i'm lost may be this is related to https://github.com/angular/angular/issues/18170

deepu105:
@thmsltmnn btw hardcoding the value works but dynamic value doesn't so this is really an angular AOT issue similar to https://github.com/angular/angular/issues/18170
I hate angular AOT

This works
```
        NgJhipsterModule.forRoot({
            // set below to true to make alerts look like toast
            alertAsToast: false,
            i18nEnabled: true,
            buildTimestamp: '123',
            defaultI18nLang: 'en'
        }),
```

This doesn't
```
        NgJhipsterModule.forRoot({
            // set below to true to make alerts look like toast
            alertAsToast: false,
            i18nEnabled: true,
            buildTimestamp: new Date().getTime().toString(),
            defaultI18nLang: 'en'
        }),
```

deepu105:
@thmsltmnn can you open an issue in Angular for this?

deepu105:
@thmsltmnn a workaround would be to use `process.env.BUILD_TIMESTAMP` directly in the ng-jhipster lib instead of passing it via the module declaration, in theory it should work as its an env variable and webpack would replace it during compilation and should be AOT safe

thmsltmnn:
@deepu105 Using a hardcoded value that I change on every release is the solution I use now. Might be good enough.

Another solution would be to move the translation stuff back from ng-jhipster to generator-jhipster, but I don't know if that fits together with the rest of the structure.

sendilkumarn:
I don't think the second option is correct. Instead for the first, I would say yes. Till angular supports using symbol in here.

deepu105:
I use the env variable in ng-jhipster directly, I dont think it needs to be configurable anyway as its only for cache busting so its good enough for this

PR #6829, opened 2017-12-08
expose link and x-total-count headers when using CORS

ruddell:
If you try to access the user-management page or any entity page from an app using a separated client/server, the call to `this.links = this.parseLinks.parse(headers.get('link'));` fails because the `link` header is not exposed to the client.  Same situation for the `X-Total-Count` header.


jdubois:
Oh that's much better, great!

PR #6833, opened 2017-12-08
add SERVER_API_URL to UAA apps

ruddell:
Simplified the three API url variables into one - `this.apiUaaPath` which is only used on routes that the UAA controls (user auth/management).

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/services/auth/_account.service.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/services/auth/_account.service.js
@@ -26,7 +26,7 @@
     Account.$inject = ['$resource'];
 
     function Account ($resource) {
-        var service = $resource('<%- apiUaaUrlPrefix %>api/account', {}, {
+        var service = $resource('<%- apiUaaPath %>api/account', {}, {
             'get': { method: 'GET', params: {}, isArray: false,
                 interceptor: {
                     response: function(response) {
--- a/generators/client/templates/angularjs/src/main/webapp/app/services/auth/_sessions.service.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/services/auth/_sessions.service.js
@@ -26,7 +26,7 @@
     Sessions.$inject = ['$resource'];
 
     function Sessions ($resource) {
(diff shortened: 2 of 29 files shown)


generators/client/templates/angularjs/src/main/webapp/app/services/auth/_account.service.js
```diff
@@ -26,7 +26,7 @@
     Account.$inject = ['$resource'];
 
     function Account ($resource) {
-        var service = $resource('<%- apiUaaUrlPrefix %>api/account', {}, {
+        var service = $resource('<%- apiUaaPath %>api/account', {}, {
```
deepu105: @ruddell just for my understanding as my memory is quite weak, why is the server api url not needed for these?

generators/client/templates/angularjs/src/main/webapp/app/services/auth/_sessions.service.js
```diff
@@ -26,7 +26,7 @@
     Sessions.$inject = ['$resource'];
 
     function Sessions ($resource) {
-        return $resource('<%- apiUaaUrlPrefix %>api/account/sessions/:series', {}, {
+        return $resource('api/account/sessions/:series', {}, {
```
deepu105: Just for my understnading why is the uaa url or server api url not needed here?

generators/client/templates/angularjs/src/main/webapp/app/services/auth/_account.service.js
ruddell: There is no SERVER_API_URL variable in AngularJS

generators/client/templates/angularjs/src/main/webapp/app/services/auth/_sessions.service.js
ruddell: This file is not copied for UAA, only session-auth

generators/client/templates/angularjs/src/main/webapp/app/services/auth/_account.service.js
deepu105: Sorry didn't notice it was ng1 file

PR #6841, opened 2017-12-09
Upgrade Angular to 5.1.0 / TypeScript to 2.5.3

kevcodez:
Angular 5.1.0 officially supports TypeScript 2.5.x

### DIFF
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -54,7 +54,7 @@
     "zone.js": "0.8.16"
   },
   "devDependencies": {
-    "@angular/cli": "1.5.5",
+    "@angular/cli": "1.6.0",
     "@angular/compiler-cli": "5.1.0",
     "@ngtools/webpack": "1.8.5",
     "@types/jasmine": "2.5.53",
--- a/generators/client/templates/angular/_package.json
+++ b/generators/client/templates/angular/_package.json
@@ -26,14 +26,14 @@
     "node_modules"
   ],
   "dependencies": {
-    "@angular/common": "5.0.5",
-    "@angular/compiler": "5.0.5",
-    "@angular/core": "5.0.5",
-    "@angular/forms": "5.0.5",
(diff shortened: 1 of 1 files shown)


gmarziou:
Thanks, angular cli 1.6.0 has also [some dependencies on newer version of webpack](https://github.com/angular/angular-cli/blob/1.6.x/package.json), I wonder whether we should upgrade it too.

kevcodez:
@gmarziou Good point, webpack-dev-server and webpack-merge-tools seem to be fine. However, I upgraded webpack to 3.10.0 to align it with Angular 5 .1.0.

gmarziou:
OK, I see also that uglifyjs-webpack-plugin has also evolved and is no longer in beta. There may be other ones to check. Tomorrow I'll run some tests with your changes.

kevcodez:
@gmarziou  Aswell as rxjs, file-loader, postcss-loader, copy-webpack-plugin and extract-text-webpack-plugin.

I aligned the dependencies.

sendilkumarn:
@kevcodez I have updated the ng-bootstrap library too. The error in travis is because of some library issues I guess. Lets check this out.

gmarziou:
I have tested it on Windows but it failed in workbox-webpack-plugin only for a prod build: `mvnw -Pprod clean package` though I don't think it is related to this PR.

~~~
[ERROR]  95% emittingError: workbox-sw is needed by the service worker and could not be copied over to your new site. Error: EPERM: operation not permitted, open 'e:\projets\paratt_simpler\target\www\workbox-sw.prod.v2.1.2.js'                                                                                                                                                              
[ERROR]     at Promise.all.then.catch (e:\projets\paratt_simpler\node_modules\workbox-build\build\lib\utils\copy-workbox-sw.js:17:13)                                                           
[ERROR]     at <anonymous> 
~~~

sendilkumarn:
yeah I think it is because of my PR. I will have a look.

Issue #6836, opened 2017-12-09
Missing locale data and UI issues


ctamisier:
A freshly generated app causes some issues:

<img width="1270" alt="screen shot 2017-12-09 at 16 28 40" src="https://user-images.githubusercontent.com/766263/33796900-1a6e1f8e-dcfe-11e7-8bea-a934d5566de9.png">

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp",
      "nativeLanguage": "fr"
    },
    "jhipsterVersion": "4.12.0",
    "baseName": "jhi",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "postgresql",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "clientFramework": "angularX",
    "useSass": true,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "fr",
    "languages": [
      "fr",
      "en"
    ]
  }
}
</pre>
</details>

##### **Environment and Tools**

java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)

git version 2.15.1

node: v6.12.1

npm: 5.5.1

bower: 1.8.0

gulp:
[16:25:51] CLI version 3.9.1

yeoman: 2.0.0

yarn: 1.3.2

Docker version 17.09.0-ce, build afdb6d4

docker-compose version 1.16.1, build 6d1ac21

deepu105:
Did you run `mvn clean` while yarn start was running?
Ideally, you should run `yarn start` after `./mvnw`

ctamisier:
[...]

PR #6843, opened 2017-12-10
import locale for angular

ctamisier:
Fix for https://github.com/jhipster/generator-jhipster/issues/6836. Let's see about the coding rules (using a constructor ?)

check 'Pipes, i18n and breaking changes' in http://blog.ninja-squad.com/2017/11/02/what-is-new-angular-5/

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green **(waiting for travis)**
- [X] Tests are added where necessary **(no need I think...)**
- [X] Documentation is added/updated where necessary **(no need I think...)**

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
@@ -18,6 +18,9 @@
 -%>
 import { NgModule, LOCALE_ID } from '@angular/core';
 import { Title } from '@angular/platform-browser';
+import { registerLocaleData } from '@angular/common';
+<%_ const localeId = getLocaleId(nativeLanguage); _%>
+import locale from '@angular/common/locales/<%= localeId %>';
 
 <%_ if (websocket === 'spring-websocket') { _%>
 import { WindowRef } from './tracker/window.service';
@@ -57,7 +60,7 @@ import {
         <%_ if (skipLanguageForLocale(nativeLanguage)) { _%>
             useValue: 'en'
         <%_ } else { _%>
-            useValue: '<%= nativeLanguage %>'
+            useValue: '<%= localeId %>'
         <%_ } _%>
         },
     ],
@@ -70,4 +73,8 @@ import {
         <%=jhiPrefixCapitalized%>AlertErrorComponent
     ]
 })
-export class <%=angularXAppName%>SharedCommonModule {}
+export class <%=angularXAppName%>SharedCommonModule {
+    constructor() {
+        registerLocaleData(locale);
+    }
+}
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -484,6 +484,16 @@ module.exports = class extends PrivateBase {
     }
 
     /**
+     * return the localeId from the given language key (from constants.LANGUAGES)
+     * if no localeId is defined, return the language key (which is a localeId itself)
+     * @param {string} language - language key
+     */
+    getLocaleId(language) {
+        const langObj = this.getAllSupportedLanguageOptions().find(langObj => langObj.value === language);
+        return langObj.localeId || language;
+    }
+
+    /**
      * get all the languages options supported by JHipster
      */
     getAllSupportedLanguageOptions() {
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -102,8 +102,12 @@ const LANGUAGES = [
     },
     { name: 'Armenian', dispName: 'Հայերեն', value: 'hy' },
     { name: 'Catalan', dispName: 'Català', value: 'ca' },
-    { name: 'Chinese (Simplified)', dispName: '中文（简体）', value: 'zh-cn' },
-    { name: 'Chinese (Traditional)', dispName: '繁體中文', value: 'zh-tw' },
+    {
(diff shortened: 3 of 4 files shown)


pascalgrimaud:
Thanks @ctamisier
I will test it, specially for languages like ar_ly,  zh_cn, etc

ctamisier:
thanks for testing, good point because it's not working with zh_cn (at least).

ctamisier:
Or maybe we should just update the values of `LANGUAGES` in https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js to make them match with the angular localeID.

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -18,6 +18,16 @@
 -%>
 import { NgModule, LOCALE_ID } from '@angular/core';
 import { Title } from '@angular/platform-browser';
+import { registerLocaleData } from '@angular/common';
+<%_ const localeIds = {
+    'zh-cn': 'zh-Hans',
```
deepu105: I suggest to add a new attribute to the original language object in generator-constants so that this is not hidden, then fetch the value from that object lile how we do for rtl

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
```diff
@@ -70,4 +80,8 @@ import {
         <%=jhiPrefixCapitalized%>AlertErrorComponent
     ]
 })
-export class <%=angularXAppName%>SharedCommonModule {}
+export class <%=angularXAppName%>SharedCommonModule {
+    constructor() {
+        registerLocaleData(locale);
```
deepu105: Just ensure that this works fine with AOT as it has history of not working with normal js constructs sometimes

ctamisier:
yeah it's indeed a better idea.
I added `localeId` in `LANGUAGES` and created a function `getLocaleId(language)`.

There is the case of `ar-ly` which is `skipForLocale` and then fallback to 'en' as a `localeId` (but there is a file `@angular/common/locales/ar-ly`), so maybe we should remove this behavior ?

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
ctamisier: hum :), I don't know what this really means... can you guide me a bit on this ? thx

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
ctamisier: done

deepu105:
@ctamisier for `ar-ly` the default locale causes some issue in dates thats why the override, i'm not sure if the issue is fixed now, can you test ar-ly with some all date types in entities?

generators/client/templates/angular/src/main/webapp/app/shared/_shared-common.module.ts
deepu105: just run the app using `./mvnw -Pprod` and make sure everything works fine, as this will use AOT and if there is an issue here it should fail

ctamisier: So everything seems to work fine with this. Running `./mvnw -Pprod`compiles and runs with no problem. Then in the app everything is fine as well, I tried with different locales, so it's all good for me.

ctamisier:
If we use `ar-ly` as localeId (not what we currently have): 
`<td>{{user.createdDate | date}}</td>` shows `2017/12/11` which is OK.

Then if we switch to english language it shows `112017/12/` (actually `11&rlm;/12&rlm;/2017`) but because we switch back to LTR, it doesn't make sense (it's what I think...).

It wouldn't be a problem if we could change the localeId runtime which not doable today (https://github.com/angular/angular/issues/15039).

So I left the`ar-ly` and skipToLocale with no change.

marcelinobadin:
Just to aggregate some info here, I'm from Brazil and therefore I am always using pt-BR and the solution I have found was to change from pt-br to pt-BR everywhere because there is a case sensitive matter on the language code. Also using only 'pt' is more common use to only portuguese from Portugal which is quite different from brazilian portuguese and we will keep using pt-BR here to not use funny words that we don't use as Portugal people does heheheheeh.

ctamisier:
thanks @marcelinobadin for your input.
Actually in JHipster there are 2 things related to languages:
1) The JHipster language which is a code for JHipster to know what language the user is using. It is internal and not a standard. For Brazilian it's **pt-br**
2) The LOCALE_ID which is required for Angular and specially to load the locale file. For Brazilian it's this file https://github.com/angular/angular/blob/master/packages/common/locales/pt.ts (you can see indide the currency is Real brasileiro), the LOCALE_ID is **pt**.
There is no pt-BR in https://github.com/angular/angular/tree/master/packages/common/locales,
and then for Portugal the LOCALE_ID is pt-PT.

I guess Angular decided to design this way.. or I might  misunderstand something ?

ctamisier:
hmm, maybe using `pt-BR` as LOCALE_ID should work as well ?

PR #6862, opened 2017-12-13
Add git information to /info endpoint

PierreBesson:
Based on @raphaelbrugier #6788 
Sorry, I have closed the previous PR to do some small changes. @raphaelbrugier you will keep attribution for your commit.

Before merging let's make sure there is not problem with the gradle build.

### DIFF
--- a/generators/server/templates/gradle/_profile_prod.gradle
+++ b/generators/server/templates/gradle/_profile_prod.gradle
@@ -83,6 +83,12 @@ processResources {
     <%_ } _%>
 }
 
+generateGitProperties {
+    onlyIf {
+        !source.isEmpty()
+    }
+}
+
 gitProperties {
     keys = ['git.branch', 'git.commit.id.abbrev']
 }
--- a/generators/server/templates/gradle/_profile_prod.gradle
+++ b/generators/server/templates/gradle/_profile_prod.gradle
@@ -17,6 +17,7 @@
  limitations under the License.
 -%>
 apply plugin: 'org.springframework.boot'
+apply plugin: 'com.gorylenko.gradle-git-properties'
 <%_ if(!skipClient) { _%>
 apply plugin: 'com.moowork.node'
     <%_ if (clientFramework === 'angular1') { _%>
@@ -80,6 +81,10 @@ processResources {
         }
     }
     <%_ } _%>
+}
+
+gitProperties {
+    keys = ['git.branch', 'git.commit.id.abbrev']
 }
 
 <%_ if (!skipClient) { _%>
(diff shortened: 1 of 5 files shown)


ctamisier:
What about having all the tags on this commit as well (maximum one most of the time) to confirm that we just released the right tag for example.
I don't know if it's possible though

PierreBesson:
@ctamisier This is the point of `git describe` :  https://git-scm.com/docs/git-describe (however only available for the maven plugin).

ctamisier:
@PierreBesson ok I see thx.
But `git.commit.id.describe` seems included in the last commit of the project https://github.com/n0mer/gradle-git-properties/commit/629469305bc2ae1eef224d288c798595953e79d6...

raphaelbrugier:
No problem @PierreBesson, thanks for taking over this PR! I got sidetracked by some other work, unfortunately.
So the maven plugin is actually working without adding the `spring-boot-starter-parent` as a parent? This is still mysterious to me, but I think I have missed something :)

pascalgrimaud:
Now I remember the old issue about this.
With Gradle, if the machine doesn't have git, the build failed:
```

* What went wrong:
Execution failed for task ':generateGitProperties'.
> org.eclipse.jgit.errors.RepositoryNotFoundException: repository not found: /home/pgrimaud/tmp/46-git-gradle

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 1m 4s
5 actionable tasks: 4 executed, 1 up-to-date
```

PierreBesson:
@pascalgrimaud I thought the problem was fixed. Looking at the commit history indeed it was fixed 6 month ago (https://github.com/n0mer/gradle-git-properties/commits/master) but still unreleased.
I think we can apply the fix described in this issue : https://github.com/n0mer/gradle-git-properties/issues/35

pascalgrimaud:
I will test this tonight

generators/server/templates/gradle/_profile_prod.gradle
```diff
@@ -82,6 +83,10 @@ processResources {
     <%_ } _%>
 }
 
+gitProperties {
```
pascalgrimaud: Before this line, you need to add the following code. Otherwise, if there is not `.git/` folder, the build will fail.

```
generateGitProperties {
	onlyIf {
		!source.isEmpty()
	}
}
```

generators/server/templates/gradle/_profile_prod.gradle
```diff
@@ -82,6 +83,10 @@ processResources {
...
+    keys = ['git.branch', 'git.commit.id.abbrev']
```
pascalgrimaud: should we add `git.commit.id.describe` to be consistent with Maven ?

PierreBesson:
@pascalgrimaud everything should be working now.

n0mer:
@pascalgrimaud @PierreBesson 1.4.18 version released with `git.commit.id.describe`, please feel free to give it a try.

pascalgrimaud:
Thanks n0mer !
We should upgrade the version

PierreBesson:
@n0mer That's great ! Thanks.

PR #6873, opened 2017-12-15
Fix display after calling logout in react app

qmonmert:
When you logout, the page doesn't display the logout page even if the user is logout.


generators/client/templates/react/src/main/webapp/app/reducers/authentication.ts
```diff
@@ -123,8 +123,8 @@ export const login = (username, password, rememberMe = false) => async (dispatch
   dispatch(getSession());
 };
 
-export const logout = () => async dispatch => {
-  await dispatch({
+export const logout = () => dispatch => {
+  dispatch({
```
deepu105: Seems like for logout this api call is not required, I'm not sure why I added this originally

PR #6877, opened 2017-12-17
Format dates 'DD/MM/YY hh:mm' in user management screens (react)

qmonmert:
REACT : 

Format dates like 'DD/MM/YY hh:mm' in two screens (user-management)

### DIFF
--- a/generators/client/templates/react/_package.json
+++ b/generators/client/templates/react/_package.json
@@ -43,6 +43,7 @@ limitations under the License.
     "react-redux-loading-bar": "2.9.3",
     "react-router-dom": "4.2.2",
     "react-router-modal": "1.2.0",
+    "react-time-format": "0.0.5",
     "react-transition-group": "2.2.1",
     "reactstrap": "5.0.0-alpha.3",
     "redux": "3.7.2",
--- a/generators/client/templates/react/_package.json
+++ b/generators/client/templates/react/_package.json
@@ -43,7 +43,7 @@ limitations under the License.
     "react-redux-loading-bar": "2.9.3",
     "react-router-dom": "4.2.2",
     "react-router-modal": "1.2.0",
-    "react-time-format": "0.0.5",
+    "react-time": "4.3.0",
     "react-transition-group": "2.2.1",
     "reactstrap": "5.0.0-alpha.3",
     "redux": "3.7.2",
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx
@@ -5,10 +5,10 @@ import { Button } from 'reactstrap';
 import { Translate } from 'react-jhipster';
 import { FaArrowLeft } from 'react-icons/lib/fa';
 import Time from 'react-time-format';
-import { APP_DATE_FORMAT } from '../../../config/constants';
 
 import { ICrudGetAction } from '../../../shared/model/redux-action.type';
 import { getUser } from '../../../reducers/user-management';
+import { APP_DATE_FORMAT } from '../../../config/constants';
 
 export interface IUserManagementDetailProps {
   getUser: ICrudGetAction;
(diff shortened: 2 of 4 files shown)


deepu105:
@qmonmert thanks for the PRs, keep them coming. I'll test and merge them as soon as I find time. Thanks and good work

generators/client/templates/react/_package.json
```diff
@@ -43,6 +43,7 @@ limitations under the License.
     "react-redux-loading-bar": "2.9.3",
     "react-router-dom": "4.2.2",
     "react-router-modal": "1.2.0",
+    "react-time-format": "0.0.5",
```
deepu105: can we use https://github.com/andreypopp/react-time instead as it has more stars and hence more users. Another option is to use moment directly as we already have it on the path. But lest go with the lib as its more elegant

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx
```diff
@@ -51,11 +52,11 @@ export class UserManagementDetail extends React.Component<IUserManagementDetailP
           <dt><Translate contentKey="userManagement.createdBy">Created By</Translate></dt>
           <dd>{user.createdBy}</dd>
           <dt><Translate contentKey="userManagement.createdDate">Created Date</Translate></dt>
-          <dd>{user.createdDate}</dd>
+          <dd><Time value={user.createdDate} format="DD/MM/YY hh:mm" /></dd>
```
deepu105: add the format to `app/config/constants.ts`  as `export const APP_DATE_FORMAT = 'DD/MM/YY hh:mm';` and import it and use here

generators/client/templates/react/_package.json
qmonmert: I tried https://github.com/andreypopp/react-time but when the date is not good it writes "Invalid dates" (https://github.com/andreypopp/react-time/blob/master/src/__tests__/Time-test.js) so we should test the return before to display, so I think react-time-format is better

deepu105: what do you mean by date is not good? do you mean invalid date or invalid format? in either case if the date is invalid, I think its ok to show that, my main issue with react-time-format has only 4 stars and we dont want to depend on a lib that has too little user base. I would rather just use moment directly to format the date  something like 
`<dd>{moment(user.createdDate).format(DD/MM/YY hh:mm)}</dd>`

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -22,6 +22,8 @@ import { Link } from 'react-router-dom';
 import { Button } from 'reactstrap';
 import { Translate } from 'react-jhipster';
 import { FaPlus, FaEye, FaPencil, FaTrash } from 'react-icons/lib/fa';
+import Time from 'react-time-format';
+import { APP_DATE_FORMAT } from '../../../config/constants';
```
deepu105: this import should be grouped with the block below as its not from a lib

generators/client/templates/react/_package.json
qmonmert: The users created by default like 'admin' haven't the property 'lastModifiedDate' so it displays 'Invalid date' in all languages ...

deepu105: Oh that case, its a valid case and you should just do a truthy check
`<dd>{user.createdDate ? <Time value={user.createdDate} format={APP_DATE_FORMAT} /> :
 null}</dd>`

This is quite common in React, IMO printing empty for invalid date by a lib is not correct, coz this can lead to unintentional bugs which will go unnoticed when some data is missing and by doing the truthy check developer explicitly has control

deepu105: Aother option is to write our own wrapper which does the format using momnet and use that component. I guess it will be a very simple componenet. We can have it in the react-jhipster lib https://github.com/jhipster/react-jhipster/

PR #6883, opened 2017-12-18
Format numbers on metrics page (react)

qmonmert:
React : 

Format the numbers on the metrics page (same display that angular app).

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/config/constants.ts
+++ b/generators/client/templates/react/src/main/webapp/app/config/constants.ts
@@ -12,3 +12,5 @@ export const messages = {
 };
 
 export const APP_DATE_FORMAT = 'DD/MM/YY HH:mm';
+export const APP_WHOLE_NUMBER_FORMAT = '0,0';
+export const APP_TWO_DIGITS_AFTER_POINT_NUMBER_FORMAT = '0,0.[00]';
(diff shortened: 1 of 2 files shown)


deepu105:
I guess instead of adding libs here for each format, I'm thinking of making a wrapper in the react-jhipster project which does these and then uses them like angular pipe here. I'll make the required changes in the lib and let you know so that you can amend this PR to use those

qmonmert:
ok with that

deepu105:
@qmonmert please use `TextFormat` component from react-jhipster for this. Similar to https://github.com/jhipster/generator-jhipster/commit/f0689314220ffceb81079043bc0711cd7a6f60eb

I guess to have a format like `10,000` the below would be enough, to specify any other format use this syntax http://numeraljs.com/#format
```
<TextFormat value={number} type="number" />
```

qmonmert:
@deepu105 I use TextFormat now ;)

generators/client/templates/react/src/main/webapp/app/config/constants.ts
```diff
@@ -9,3 +9,8 @@ export const messages = {
   DATA_DELETE_SUCCESS_ALERT: 'Data deleted successfully!',
   DATA_UPDATE_ERROR_ALERT: 'Data deleted successfully!'
 };
+
+export const APP_DATE_FORMAT = 'DD/MM/YY HH:mm';
+
+export const APP_FORMAT_NUMBER_WHOLE = '0,0';
```
deepu105: could you please change this to `APP_NUMBER_WHOLE_FORMAT` & `APP_NUMBER_TWO_DIGITS_AFTER_POINT_FORMAT` for consistency

deepu105: never mind ill change it

qmonmert: ok I was ready to change ;)

deepu105:
@sendilkumarn I think you broke lint, can you fix it

PR #6897, opened 2017-12-20
Start dev for entity generation (react)

qmonmert:
PR to say that I begin the dev to generate entity for react.


deepu105:
I started working on them as well, but its fine if you would like to do it. Use this PR for it, i'll commit what I have here so you can use them. Follow the same pattern as the user management screen

qmonmert:
ok, user management was already my target ;)

deepu105:
I have commited what I had, you can continue with them :) thanks


Thanks & Regards,
Deepu

qmonmert:
Thanks @deepu105 I continue with that ;)

jdubois:
- FYI @qmonmert is working at Ippon next to me :-)
- For the entity sub-generator, I find what we have with Angular is too complex: we should generate "white applications" that are supposed to be easily modified by end-user to suit their business needs. Here we provide a beautiful, feature-rich UI, but that also makes it hard to modify. So for React, if we could have something simpler: I'm sure it will be better for end-users, and that should also help us to finish this sooner

deepu105:
Lets start with simple entity screens first without any fancy
datepicker/file upload stuff first

Sendil, I didn't get what you were talking about, so lets discuss that when
we meet next time

generators/client/templates/react/src/main/webapp/app/config/constants.ts
```diff
@@ -4,10 +4,14 @@ const config = {
...
+export const SERVER_API_URL = '';
+
 export const messages = {
   DATA_CREATE_SUCCESS_ALERT: 'Data created successfully!',
   DATA_DELETE_SUCCESS_ALERT: 'Data deleted successfully!',
   DATA_UPDATE_ERROR_ALERT: 'Data deleted successfully!'
 };
 
 export const APP_DATE_FORMAT = 'DD/MM/YY HH:mm';
+export const APP_FORMAT_LOCAL_DATE = 'DD/MM/YYYY';
```
deepu105: keep names consistent use `APP_LOCAL_DATE_FORMAT`, `APP_LOCAL_DATETIME_FORMAT`

generators/client/templates/react/src/main/webapp/app/config/translation.ts
```diff
@@ -1,20 +1,32 @@
 /* global: require*/
 import { TranslatorContext } from 'react-jhipster';
+import * as _ from 'lodash';
```
deepu105: import only what is needed like `import { merge } from 'lodash';`

generators/client/templates/react/src/main/webapp/app/config/translation.ts
```diff
@@ -1,20 +1,32 @@
...
 
 import { setLocale } from '../reducers/locale';
 import Storage from '../shared/util/storage-util';
 
-const mergeTranslations = requireContext => requireContext.keys().reduce(
-  (merged, key) => ({ ...merged, ...requireContext(key) }),
-  {}
-);
+// const mergeTranslations = requireContext => requireContext.keys().reduce(
```
deepu105: why are these changed? it has nothing to do with entities and I dont see why this change is needed as it will produce some weird code, please revert back to original code

generators/client/templates/react/src/main/webapp/app/entities/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { Route, Switch } from 'react-router-dom';
+import { ModalRoute } from 'react-router-modal';
+
+/* jhipster-needle-add-route-index-1 - JHipster will add routes here */
```
deepu105: give a proper name `jhipster-needle-add-route-import`

generators/client/templates/react/src/main/webapp/app/entities/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+const Routes = ({ match }) => (
+  <div>
+    <Switch>
+        /* jhipster-needle-add-route-index-2 - JHipster will routes here */
```
deepu105: give a proper name `jhipster-needle-add-route-path`

generators/client/templates/react/src/main/webapp/app/reducers/index.ts
```diff
@@ -8,6 +8,7 @@ import layout from './layout';
 import authentication from './authentication';
 import administration from './administration';
 import userManagement from './user-management';
+/* jhipster-needle-add-reducer-1 - JHipster will add reducer here */
```
deepu105: give a proper name `jhipster-needle-add-reducer-import`

generators/client/templates/react/src/main/webapp/app/reducers/index.ts
```diff
@@ -18,4 +19,5 @@ export default combineReducers({
   administration,
   userManagement,
   loadingBar
+  /* jhipster-needle-add-reducer-2 - JHipster will add reducer here */
```
deepu105: give a proper name `jhipster-needle-add-reducer-combine`

generators/client/templates/react/src/main/webapp/app/routes.tsx
```diff
@@ -15,6 +16,7 @@ const Routes = () => (
     <Route path="/logout" component={Logout} />
     <PrivateRoute path="/admin" component={Admin} />
     <PrivateRoute path="/account" component={Account} />
+    <Route path="**" component={Entities} />
```
deepu105: path should be `/entity`

generators/client/templates/react/src/main/webapp/app/shared/util/date-utils.ts
```diff
@@ -0,0 +1,11 @@
+import * as moment from 'moment';
+
+import { APP_FORMAT_DATETIME_LOCAL } from '../../config/constants';
+
+export const convertDateTimeFromServer = (date: any) => {
+    if (date) {
```
deepu105: change to 
```
export const convertDateTimeFromServer = date => 
  date ? moment(date).format(APP_FORMAT_DATETIME_LOCAL) : null;
```

type any is implicit so no need to specify

generators/entity-client/files.js
```diff
@@ -22,9 +22,11 @@ const constants = require('../generator-constants');
 /* Constants use throughout */
 const CLIENT_TEST_SRC_DIR = constants.CLIENT_TEST_SRC_DIR;
 const ANGULAR_DIR = constants.ANGULAR_DIR;
+const REACT_DIR = constants.ANGULAR_DIR;
 
 const CLIENT_NG1_TEMPLATES_DIR = 'angularjs';
 const CLIENT_NG2_TEMPLATES_DIR = 'angular';
+const CLIENT_NG3_TEMPLATES_DIR = 'react';
```
deepu105: CLIENT_REACT_TEMPLATES_DIR

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-detail.tsx
```diff
@@ -0,0 +1,137 @@
+<%#
```
deepu105: add licence header to all new template files added

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-detail.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { getEntity } from './<%= entityFileName %>.reducer';
+import { APP_DATE_FORMAT, APP_FORMAT_LOCAL_DATE } from '../../config/constants';
+
+export interface I<%= entityReactName %>DetailProps {
+  getEntity: ICrudGetAction;
+  <%= entityInstance %>: any;
+  match: any;
+}
+
+export class <%= entityReactName %>Detail extends React.Component<I<%= entityReactName %>DetailProps, undefined> {
```
deepu105: If there is no state then no need to define the generic just do `export class <%= entityReactName %>Detail extends React.Component<I<%= entityReactName %>DetailProps> {`

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-dialog.tsx
```diff
@@ -0,0 +1,350 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Modal, ModalHeader, ModalBody, ModalFooter, Button, Label } from 'reactstrap';
+import { AvForm, AvGroup, AvInput, AvFeedback } from 'availity-reactstrap-validation';
+import { Translate } from 'react-jhipster';
+import { FaBan, FaFloppyO } from 'react-icons/lib/fa';
+
```
deepu105: remove blank line here

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-dialog.tsx
```diff
@@ -0,0 +1,350 @@
...
+import * as moment from 'moment';
+
+import { ICrudGetAction, ICrudPutAction } from '../../shared/model/redux-action.type';
+import { getEntity, updateEntity, createEntity } from './<%= entityFileName %>.reducer';
+<%_ if (enableTranslation) { _%>
+import { locales } from '../../config/translation';
```
deepu105: locales is not required here

generators/client/templates/react/src/main/webapp/app/config/translation.ts
qmonmert: I changed this because when I generated several entities, only one file myEntityX.json (with the traduction) was used, the others were ignored :(

deepu105: it should fetch all files and merge them, as its doing the same currently and it works, don't worry about this just take care of entities and we can look into this later

qmonmert:
@deepu105 if you want test react, you can import https://github.com/jhipster/jdl-samples/blob/master/simple-online-shop.jh or https://github.com/jhipster/jdl-samples/blob/master/blog.jh
Maybe it will be good to merge quickly like that others developers can also contribute or we can wait last week (I will work for sure on this until next friday) ? cc @jdubois

deepu105:
Ok, ill have a final review and then will test and merge soon

deepu105:
@qmonmert I fixed some issues to make a basic entity work with current master. I didnt find any issue with entity translation, i'll check again later, but I guess this version might be good enough so once travis passes I'll merge this

generators/client/templates/react/src/main/webapp/app/config/translation.ts
```diff
@@ -1,10 +1,11 @@
 /* global: require*/
 import { TranslatorContext, Storage } from 'react-jhipster';
+import { merge } from 'lodash';
 
 import { setLocale } from '../reducers/locale';
 
 const mergeTranslations = requireContext => requireContext.keys().reduce(
-  (merged, key) => ({ ...merged, ...requireContext(key) }),
+  (merged, key) => merge({ ...merged }, { ...requireContext(key) }),
```
deepu105: @qmonmert this will solve the i18n issue you encountered

sendilkumarn: Cant we do this with `Object.assign` do we need `lodash` for this 🤔

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-delete-dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+-%>
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Modal, ModalHeader, ModalBody, ModalFooter, Button } from 'reactstrap';
+import { Translate, ICrudGetAction, ICrudDeleteAction } from 'react-jhipster';
+import { FaBan, FaTrash } from 'react-icons/lib/fa';
+
+import { getEntity, deleteEntity } from './<%= entityFileName %>.reducer';
+
+export interface I<%= entityReactName %>DeleteDialogProps {
```
sendilkumarn: The names are really long

generators/client/templates/react/src/main/webapp/app/config/translation.ts
deepu105: Object.assign is not recursive, its same as { ... } destructuring which is only first level, we need deep merge here

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity-delete-dialog.tsx
deepu105: ya not sure what to do as we cant strip the props, and how much ever we try there will be some case where name is too long, let me see what can be done

deepu105:
@sendilkumarn I overwrote the nit change you did as it was giving too many merge conflicts when I pulled it dies to rebase I have already done, those are minor things we can do later as we need to fix the formats anyway

sendilkumarn:
Well okay. Let me merge this 👍

PR #6899, opened 2017-12-20
only import rxjs modules that are used

ruddell:
For size-sensitive apps, rxjs can be imported differently as mentioned in the [rxjs readme](https://github.com/ReactiveX/rxjs).  We can reduce the size of the vendor bundle (default app, no entities) from 946 KB -> 818 KB (gzipped 209 KB -> 178 KB).  This is the [same method](https://github.com/angular/angular-cli/blob/a5edc12ee83264c6a691a688970d5c3deeada717/packages/%40angular/cli/tasks/eject.ts#L273-L279) that the angular-cli uses to optimize rxjs imports when you run `ng eject`.

Related PR in ng-jhipster: https://github.com/jhipster/ng-jhipster/pull/59

Webpack Visualizer RXJS size screenshots: [Before](https://user-images.githubusercontent.com/4294623/34221966-e56988ae-e587-11e7-86d3-fc3a458a611b.png) + [After](https://user-images.githubusercontent.com/4294623/34221967-e58a9b3e-e587-11e7-8892-a5927f27cce6.png)

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/_vendor.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/_vendor.ts
@@ -23,4 +23,6 @@ import '../content/scss/vendor.scss';
 <%_} else { _%>
 import '../content/css/vendor.css';
 <%_ } _%>
+import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/catch';
 // jhipster-needle-add-element-to-vendor - JHipster will add new menu items here
--- a/generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.service.ts
+++ b/generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.service.ts
@@ -18,7 +18,7 @@
 -%>
 import { Injectable } from '@angular/core';
 import { Http, Response, URLSearchParams } from '@angular/http';
-import { Observable } from 'rxjs/Rx';
+import { Observable } from 'rxjs/Observable';
 import { SERVER_API_URL } from '../../app.constants';
 
 @Injectable()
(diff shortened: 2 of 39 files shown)


generators/client/templates/angular/src/main/webapp/app/_vendor.ts
```diff
@@ -23,4 +23,6 @@ import '../content/scss/vendor.scss';
 <%_} else { _%>
 import '../content/css/vendor.css';
 <%_ } _%>
+import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/catch';
```
ruddell: You can either import these in one place, or in every file that needs them

deepu105:
I thought webpack tree shaking would get rid of unwanted code. Hmmm apparently not.

ruddell:
Importing from `rxjs/Rx` brings in the entire library and tree-shaking doesn't work specifically with RxJS because of how it is packaged.

>Unfortunately, Tree Shaking can’t help here as well, again due to the nature how RxJS is bundled: Tree Shaking relies on the symbol names of export and import definitions. As the RxJS operator modules don’t export anything, but change global state (“side effects”) instead, it’s not possible for the Tree Shaker to detect which operators are in use. As a result, all operators (and static methods of Observable) will be included in the build.
https://christianliebel.com/2017/07/import-rxjs-correctly/

This is also how all of the official Angular tutorials show to import from RxJS ([1](https://angular.io/tutorial/toh-pt4), [2](https://angular.io/tutorial/toh-pt6)).

jdubois:
This is awesome -> @deepu105 can you check and do a release of ng-jhipster for this?

deepu105:
Yes ill do it tomorrow

Issue #6379, opened 2017-09-19
Study using Prettier to format generated code


jdubois:
We spend a lot of time to generate well-formatted code.

[Prettier](https://github.com/prettier/prettier) is a source code formatter, available as a Node module: it should be very easy to send our generated code to Prettier before writing the file, so that the end-result is perfectly  formatted by Prettier.

In the end, we would win a lot of time: no need to worry anymore about formatting everything so that the end-result is good. We could focus on having some clean code inside the template, instead. Too often, our templates are unreadable because of the output we want to have.

This is also the only solution for things like import statements, as we don't know what the user will do (for example, what name will he use for his entities).

Unfortunately, Prettier doesn't support Java, so this would only work for TypeScript:

- We need a JavaScript tool, so that it works in our generation pipeline - we can't use Java-based code formatters
- TypeScript is already quite a lot of code, so that's a good starting point
- Maybe we could write the Java support for Prettier? I could try to find a trainee for that

deepu105:
Yes we should use it


Thanks & Regards,
Deepu

deepu105:
We could also explore what VSCode plugins uses for formatting as some of
the plugins support java, js and Ts


Thanks & Regards,
Deepu

PierreBesson:
Great idea to format code at generation time !

Note : code formatters for Java already exist: 
- https://github.com/google/google-java-format (doesn't suit me)
- Also you can use both the Eclipse and IntelliJ code formatters from the command line (requires an installation of the respective IDE).

gmarziou:
Why don't you like google-java-format ? 
It seems to be easy to use from both maven and gradle, so we could easily run it after project generation.

Anyway, the problem with these tools is that it's ok to run them on code we generate which is easy when you generate an app but how would you run them only on the files generated by the entity generator?

PierreBesson:
@gmarziou I'm not a big fan of google-java-format coding style (personal habits) and it can not be customized at all :disappointed:.

@gmarziou I think the idea was not to run this after generation but somehow during the generation itself (so the generated files are correctly formatted out of the box).

jdubois:
@gmarziou @PierreBesson we need an NPM-based tool, so we can integrate it into our generation pipeline: files should be "prettified" before being written. That's way faster, and a lot easier to integrate.

deepu105:
Found a NPM lib that can format java 
https://www.npmjs.com/package/auto-format

deepu105:
[...]

PR #6906, opened 2017-12-22
add prettier and tslint compatibility

arthurescriou:
Add prettier flag 'code-formatter' which format with prettier and tslint the code
Also add a yarn command : 'yarn code-format' which do the same
(tslint --fix take some time: ~1 minute) Tslint --fix is necessary because prettier break some tslint rules
There is no tests using this feature and it is not set by default then it should not change the behavior of jhipster 

#6379


PierreBesson:
@arthurescriou Good work !
However you should not have made the PR from your master branch, please follow [our guidelines](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md#submitting-a-pull-request) next time.

generators/app/index.js
```diff
@@ -116,6 +117,13 @@ module.exports = class extends BaseGenerator {
             defaults: false
         });
 
+        // This adds support for a `--code-formatter` flag
+        this.option('code-formatter', {
```
deepu105: This need not be an option, if we are adding it lets make it default

generators/app/index.js
```diff
@@ -363,9 +372,17 @@ module.exports = class extends BaseGenerator {
                         if (this.clientPackageManager === 'yarn') {
                             this.log(chalk.bold(`\nInstalling generator-jhipster@${this.jhipsterVersion} locally using yarn`));
                             this.yarnInstall();
+                            if (this.options['code-formatter']) {
```
deepu105: Use ` this.codeFormatter` instaed of options[]

generators/app/index.js
```diff
@@ -363,9 +372,17 @@ module.exports = class extends BaseGenerator {
...
+                                this.log(`${chalk.bold('\nFormatting code with Tslint')}\n please wait`);
+                                shelljs.exec('yarn lint --fix');
```
deepu105: This is an issue, we shoulnt be relying on this and prettier, what rules are prettier breaking? is there way to control them in prettier? if not we can consider changing the rules to align with prettier for the ones that are broken

generators/app/index.js
```diff
@@ -363,9 +372,17 @@ module.exports = class extends BaseGenerator {
...
+                            }
                         } else if (this.clientPackageManager === 'npm') {
                             this.log(chalk.bold(`\nInstalling generator-jhipster@${this.jhipsterVersion} locally using npm`));
                             this.npmInstall();
+                            if (this.options['code-formatter']) {
```
deepu105: same here

cli/codeFormat.js
```diff
@@ -0,0 +1,24 @@
+const prettier = require('prettier');
+const fs = require('fs');
+
+const walkSync = function (dir, filelist) {
```
deepu105: This is used by generated project so generate this file in the webpack/utils folder and use that

generators/generator-base.js
```diff
@@ -1517,6 +1517,40 @@ module.exports = class extends PrivateBase {
...
+    /**
+     * Copy Ts templates after stripping translation keys when translation is disabled.
+     *
+     * @param {string} source - path of the source file to copy from
+     * @param {string} dest - path of the destination file to copy to
+     * @param {object} generator - context that can be used as the generator instance or data to process template
+     * @param {object} opt - options that can be passed to template method
+     * @param {boolean} template - flag to use template method instead of copy
+     */
+    processTs(source, dest, generator, opt, template) {
```
deepu105: Is this method different from processJs? looks like you are relying on this to run prettier, I would rather just run prettier using the npm command from the generated project during the yarn install/build phase and avoid doing it here. If it needs to be done during generation then I would prefer to use "Transform output files through streams" section in http://yeoman.io/authoring/file-system.html

generators/generator-base.js
```diff
@@ -1517,6 +1517,40 @@ module.exports = class extends PrivateBase {
...
+    /**
+     * Copy config files templates directly on the disk to use it to configure while generating
+     *
+     * @param {string} source - path of the source file to copy from
+     * @param {string} dest - path of the destination file to copy to
+     * @param {object} generator - context that can be used as the generator instance or data to process template
+     * @param {object} opt - options that can be passed to template method
+     * @param {boolean} template - flag to use template method instead of copy
+     */
+    processConfig(source, dest, generator, opt, template) {
```
deepu105: why is this method needed? I dont understand the significance of using this for the prettier config file

generators/generator-base.js
```diff
@@ -1517,6 +1517,40 @@ module.exports = class extends PrivateBase {
...
+        this.log(chalk.magenta(dest));
```
deepu105: `this.debug` plz

package.json
```diff
@@ -62,10 +62,14 @@
     "mkdirp": "0.5.1",
     "opencollective": "1.0.3",
     "pluralize": "7.0.0",
+    "prettier": "1.9.2",
     "randexp": "0.4.6",
     "semver": "5.4.1",
     "shelljs": "0.7.8",
     "tabtab": "2.2.2",
+    "tslint": "5.5.0",
```
deepu105: why are these deps here?

deepu105:
Good work @arthurescriou I have some comments please take a look

generators/app/index.js
arthurescriou: Prettier don't use the shorthand {value} to {value: value} that tslint expect
I have fixed most of the other rules (quotes, parentheses) now
But i have not found any rules about this kind of format in prettier : 
https://prettier.io/docs/en/options.html

generators/generator-base.js
arthurescriou: To use prettier while generating file i need to give the function the path to the config file
But i did not found the path of this file before the copy so i force the generator to copy it before all of the other files i want to format

deepu105:
Lets go with this approach
1. Generate the prettier script under webpack folder of generated app
2. Generate prettier config file for the generated app
3. Add npm script task to run prettier
4. Invoke the task during yarn build phase after generation is complete (another option is to use the pipe feature but this can be future improvement)

This way you don't need to do anything during generation and can avoid most of the current changes. Lets do this first and then we can see how this can be improved

cli/codeFormat.js
```diff
@@ -0,0 +1,29 @@
+const prettier = require('prettier');
```
deepu105: move this file to `generators/client/templates/angular/webpack/_prettier.js` as a template and add an entry for this in the `generators/client/files-angular.js` file so that this file is generated during app creation.

generators/client/files-angular.js
```diff
@@ -32,6 +32,7 @@ const files = {
     common: [
         {
             templates: [
+                'prettierrc.json',
```
deepu105: prefix with `_` to be consistent

generators/client/templates/angular/_package.json
```diff
@@ -138,6 +138,7 @@
     "node": ">=6.9.0"
   },
   "scripts": {
+    "code-format":"node ../generator-jhipster/cli/format.js; yarn lint:fix",
```
deepu105: change command name to `prettier:format` or `prettify` so that its easy for people to relate to prettier

generators/client/index.js
```diff
@@ -401,6 +402,7 @@ module.exports = class extends BaseGenerator {
                     this.log(logMsg);
                 }
             );
+            codeFormat.formatter();
```
deepu105: don't do this instead do `this.spawnCommandSync(this.clientPackageManager, ['run', 'code-format'])` but call it before the build commands so may be at line 396/397 before invoking `webpack:build`

generators/client/templates/angular/_package.json
deepu105: also dont use yarn directly use `<%= clientPackageManager %>`

package.json
deepu105: remove all these new dependencies from here and revert the yarn lock file

cli/codeFormat.js
deepu105: add "prettier": "1.9.2" as dependency to the generated apps package.json

travis/samples/.jhipster/EntityWithPaginationDTO.json
```diff
@@ -8,7 +8,7 @@
         }
     ],
     "changelogDate": "20160806123743",
-    "entityTableName": "entity_with_pagination_and_dto",
+    "entityTableName": "entity_with_pagination_dto",
```
deepu105: What if users create entities with long name?

deepu105: Doesnt prettier has a setting to break imports into multiline when line limit exceeds? Else users will get error when they create something long

jdubois: In fact Prettier fails at this -> it puts all imports in one line

jdubois: Then if your import is more than 180 chars, you named it wrong

deepu105: ya but since the imports will have path as well its quite easy to reach the limit even with something like `CustomerCategoryServiceCode`, and if there are are more than one item imported then it easily will reach 180

deepu105: I thought prettier was smart enough to handle these, may be we should use `// prettier-ignore` comment for these as we already handle breaking stuff when its over 180 chars

arthurescriou: We can also put the file that <code>import</code> entities in <code>.prettierignore</code>  if it is the only one that break tslint rules.
Or there will be a new <a href='https://github.com/palantir/tslint/pull/3099'>tslint feature<a> in the next release that we can use to ignore <code> import</code>  lines

deepu105: most files will be importing entities, hence I dont how ignoring entire file will work

yarn.lock
```diff
@@ -194,6 +194,10 @@ bcrypt-pbkdf@^1.0.0:
   dependencies:
     tweetnacl "^0.14.3"
 
+big.js@^3.1.3:
```
deepu105: please revert this file

generators/generator-base.js
```diff
@@ -263,7 +263,8 @@ module.exports = class extends PrivateBase {
             let importStatement = `|import { ${appName}${entityAngularName}Module } from './${entityFolderName}/${entityFileName}.module';`;
             if (importStatement.length > constants.LINE_LENGTH) {
                 importStatement =
-                    `|import {
+                    `|//prettier-ignore
```
deepu105: add a space `// prettier-ignore`

generators/client/templates/angular/webpack/_prettier.js
```diff
@@ -0,0 +1,22 @@
+const prettier = require('prettier');
+const fs = require('fs');
+
+const walkSync = function (dir, filelist) {
```
deepu105: Isnt it possible to do the same using prettier CLI directly without this script?

generators/generator-base.js
deepu105: Since this is added please revert changes to the entity names in Travis as some of them were made longer on purpose as we do some specific thing when names are longer for some DB

deepu105:
@arthurescriou I have 2 more comments please take a look, once they are addressed i'll test and merge

deepu105:
@arthurescriou also please rebase and keep only relevant commits, there are too many commits due to changes

arthurescriou:
The <code>yarn run lint -- --fix</code> command took too much time so travis killed it because it look like something went wrong.
I think this is in a Jhipster project which have lots of entities and there is too many files to lint.

deepu105:
@arthurescriou normally linting works fast, but lint fix might take time. I'm still not satisfied with the lint --fix solution here. Give me some time, i'll play with this and see if there is some middle ground we can reach without using lint --fix, I'm ok to break existing style for that if required. hence i'm marking it as v5 for next major version.

deepu105:
@arthurescriou is lint --fix needed only for object shorthand rule? or are their other rules that are affected? also can you open an issue with Prettier for object shorthand rule?

deepu105:
@jdubois this change will create a lot of changes for existing project due to format changes and will be annoying for people who upgrade hence I propose to merge this for v5, WDYT?

jdubois:
Oh good point: that would be a breaking change, so let's put it for v5.
Besides, that leaves us time to test this better.

Le 13 janv. 2018 4:59 PM, "Deepu K Sasidharan" <notifications@github.com> a
écrit :

deepu105:
This is ready to merge but only once v4 is frozen and maintenance branch is created

jdubois:
-> @deepu105 then let's freeze it then :-)
I'll do a mail on the core dec team mailing list

deepu105:
@if we freeze now that would mean we would have to release maintenance from that branch. Maybe it is better to merge all possible non v4 PR first and make a final v4 release and then freeze it. Especially it would be nice to merge #6955 #6951(This is ready but would like someone other than me also to test it) #6805 #6783 and other minor stuff WDYT?

jdubois:
Oh yes you are correct @deepu105 (as always!)
But we should do this very soon, we are reaching the point where we must go forward on JHipster 5

deepu105:
@jdubois yes I agree, we need to start pushing like how we did for v4. Lets try to do the below if you agree

1. Test and merge as much non v5 PR as possible, especially (#6951 as its huge)
2. Create v4 maintenance branch
3. Merge spring boot 2 branch into master and make it v5 dev
4. Merge all v5 PR that are ready
5. Work aggressively to release v5 :wink:

jdubois:
And also remove AngularJS 1 :-)
Yes @deepu105 let's do it - but those are mainly tasks for you, how can I help? I'm still working on the Spring Boot 2 migration, I hope to finish in a couple of weeks but several builds are already OK

deepu105:
@jdubois I can take care of items 4, 5 and 6. I would need help on items 1 to 3 in above list

PR #6926, opened 2017-12-30
Fix month in audit component tests

pascalgrimaud:
- Please make sure the below checklist is followed for Pull Requests.

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [ ] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #6940, opened 2018-01-03
Refactorization of the 2 digits date part code in audit component tests

fgd99:
Fix #6926

just a little refactorization to remove the duplicate code that builds 2 digits date parts in the function getDate of the audits component tests


generators/client/templates/angular/src/test/javascript/spec/app/admin/audits/_audits.component.spec.ts
```diff
@@ -25,6 +25,10 @@ import { AuditsComponent } from '../../../../../../main/webapp/app/admin/audits/
 import { AuditsService } from '../../../../../../main/webapp/app/admin/audits/audits.service';
 import { ITEMS_PER_PAGE } from '../../../../../../main/webapp/app/shared';
 
+function build2DigitsDatePart(datePart: number) {
+    return ('0' + datePart).slice(-2);
```
MathieuAA: Please replace by
```javascript
return `0${datePart}`.slice(-2);
```

PR #6946, opened 2018-01-05
React : Sort and Pagination for users screen

qmonmert:
React : 

Sort and pagination for users screen

URL is not changed during the navigation (always http://localhost:9000/#/admin/user-management), is it mandatory to update the URL during the navigation ?


qmonmert:
Seen with @jdubois we must handle the update of the URLs

generators/client/templates/react/_package.json
```diff
@@ -29,12 +29,14 @@ limitations under the License.
     "availity-reactstrap-validation": "2.0.0-alpha.1",
     "axios": "0.17.0",
     "bootstrap": "4.0.0-beta.2",
+    "font-awesome": "4.7.0",
```
deepu105: we already have fontawesome from "react-icons" so use that and remove this please

generators/client/templates/react/_package.json
```diff
@@ -29,12 +29,14 @@ limitations under the License.
...
     "loaders.css": "0.1.2",
     "lodash": "4.17.4",
     "moment": "2.19.1",
     "react": "16.0.0",
     "react-addons-css-transition-group": "16.0.0-alpha.3",
     "react-addons-transition-group": "16.0.0-alpha.3",
+    "react-bootstrap": "0.31.5",
```
deepu105: we use reactstrap, why is this needed? please remove it

generators/client/templates/react/src/main/webapp/app/app.css
```diff
@@ -41,6 +42,53 @@ a:hover {
    font-weight: bold;
 }
 
+.pagination li a {
```
deepu105: why are these styles needed?

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,11 +20,13 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
 import { Button } from 'reactstrap';
+import { Pagination } from 'react-bootstrap';
```
deepu105: use https://reactstrap.github.io/components/pagination/ please

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -33,14 +35,91 @@ export interface IUserManagementProps {
   match: any;
 }
 
-export class UserManagement extends React.Component<IUserManagementProps, undefined> {
+export interface IUserManagementState {
+  itemsPerPage: number;
```
deepu105: these will be common for all entities, hence please make it a common interface and extend from that here

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -33,14 +35,91 @@ export interface IUserManagementProps {
...
+    }
+    this.state = {
+      itemsPerPage: ITEMS_PER_PAGE,
+      items: 0,
+      sort,
+      order,
+      activePage,
+      totalItems: 0
+    };
+    this.handlePagination = this.handlePagination.bind(this);
```
deepu105: use arrow methods instead of binding

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -33,14 +35,91 @@ export interface IUserManagementProps {
...
+  order: string;
+  activePage: number;
+  totalItems: number;
+}
+
+export class UserManagement extends React.Component<IUserManagementProps, IUserManagementState> {
 
   constructor(props) {
     super(props);
+    const pageParam = this.getUrlParameter('page', props.location.search);
```
deepu105: I guess this logic can be moved to util class

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -33,14 +35,91 @@ export interface IUserManagementProps {
...
+      order,
+      activePage,
+      totalItems: 0
+    };
+    this.handlePagination = this.handlePagination.bind(this);
   }
 
   componentDidMount() {
-    this.props.getUsers();
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, this.state.sort + ',' + this.state.order).then(action => {
```
deepu105: this is not how it should be handled in redux, you shouldnt handle sideeffect. Just update the store and use the value from store in component

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -57,18 +136,22 @@ export class UserManagement extends React.Component<IUserManagementProps, undefi
           <table className="table table-striped">
             <thead>
               <tr>
-                <th><Translate contentKey="global.field.id">ID</Translate></th>
-                <th><Translate contentKey="userManagement.login">Login</Translate></th>
-                <th><Translate contentKey="userManagement.email">Email</Translate></th>
+                <th onClick={this.sort.bind(this, 'id')}><Translate contentKey="global.field.id">ID</Translate> <span className="fa fa-sort"/></th>
```
deepu105: do not bind use arrow methods

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
qmonmert: ok I didn't see that for the pagination

generators/client/templates/react/src/main/webapp/app/app.css
qmonmert: it was useful to override the pagination with react-bootstrap

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
qmonmert: I wanted validate with you the mechanism (pagination) only on this screen to begin

deepu105: yes we should get it right here first then apply the same for entity :+1:

generators/client/templates/react/src/main/webapp/app/app.css
deepu105: react-bootstrap is still on bootstrap 3 that's why we went with react strap, it wont be nice to use both together especially one with different bootstrap impl, lets try to get it working with reactstrap, the features in react need not be pixel perfect in sync with one in angular, we can make changes where it makes sense

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
+      activePage
+    };
   }
 
   componentDidMount() {
-    this.props.getUsers();
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, this.state.sort + ',' + this.state.order);
+  }
+
+  sort = prop => e => {
```
deepu105: can be simplified to 
```
sort = prop => () => {
  this.setState({
    order: this.state.order === 'asc' ? 'desc' : 'asc',
    sort: prop
  }, () => this.sortUsers());
```

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
+    this.setState({
+       order: 'asc',
+       sort: prop
+    }, () => {
+      this.sortUsers();
+    });
+  }
+
+  sortUsers() {
+    location.href = location.href.split('?')[0] + '?page=' + this.state.activePage + '&sort=' + this.state.sort + ',' + this.state.order;
```
deepu105: I guess its not a good idea to mutate location directly, we should use the api provided by React router to update this so that states are in sync

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
+      itemsPerPage: ITEMS_PER_PAGE,
+      sort,
+      order,
+      activePage
+    };
   }
 
   componentDidMount() {
-    this.props.getUsers();
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, this.state.sort + ',' + this.state.order);
```
deepu105: can we use string literals instead of concatenation?

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
+       order: 'asc',
+       sort: prop
+    }, () => {
+      this.sortUsers();
+    });
+  }
+
+  sortUsers() {
+    location.href = location.href.split('?')[0] + '?page=' + this.state.activePage + '&sort=' + this.state.sort + ',' + this.state.order;
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, this.state.sort + ',' + this.state.order);
```
deepu105: can we use string literals instead of concatenation?

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
deepu105: can we use string literals instead of concatenation?

generators/client/templates/react/src/main/webapp/app/reducers/user-management.ts
```diff
@@ -100,10 +102,13 @@ export default (state = initialState, action) => {
 
 const apiUrl = '/api/users';
 // Actions
-export const getUsers: ICrudGetAction = (page, size, sort) => ({
-  type: ACTION_TYPES.FETCH_USERS,
-  payload: axios.get(`${apiUrl}?cacheBuster=${new Date().getTime()}`)
-});
+export const getUsers: ICrudGetAction = (page, size, sort) => {
+  const requestUrl = `${apiUrl}` + (sort ? `?page=${page}&size=${size}&sort=${sort}` : ``);
```
deepu105: you can also do `` `${apiUrl}${sort ? `?page=${page}&size=${size}&sort=${sort}` : ''}`;``

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
+export class UserManagement extends React.Component<IUserManagementProps, IPaginationState> {
 
   constructor(props) {
     super(props);
+    const pageParam = getUrlParameter('page', props.location.search);
+    const sortParam = getUrlParameter('sort', props.location.search);
+    let sort = 'id';
+    let order = 'asc';
+    let activePage = 1;
+    if (pageParam !== '' && !isNaN(Number(pageParam))) {
```
deepu105: parseInt is better than using wrapper Number Object

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -20,31 +20,82 @@ import * as React from 'react';
...
   match: any;
+  totalItems: 0;
 }
 
-export class UserManagement extends React.Component<IUserManagementProps, undefined> {
+export class UserManagement extends React.Component<IUserManagementProps, IPaginationState> {
 
   constructor(props) {
     super(props);
+    const pageParam = getUrlParameter('page', props.location.search);
```
deepu105: you can move this logic to a util method that returns `{ itemsPerPage, sort, order, activePage }` and pass `props.location` to the method as param
so this will become
```
this.state = {
  ...getSortState(props.location)
}
```

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
qmonmert: but it seems that with https://reactstrap.github.io/components/pagination/ we need to implement the logic of the pagination unlike react-bootstrap which provides a simple component

qmonmert: @deepu105 since today there is a message on this page https://react-bootstrap.github.io/ which prevents that the project is under active development, and APIs will change so what do you think ?

deepu105: That message was always there and I have been following related issue for sometime, seems like there is no much progress and they still dont have BS4 in roadmap so for me `react-bootstrap` is a strict no for us and should stick to `reactstrap` implementing our pagination using reactstrap doesnt seem that difficult. Ofcourse the componenet structure is different but its not a big deal. If you want you can write a wrapper on top of reactstrap to make a simple component

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -27,45 +27,32 @@ import { FaPlus, FaEye, FaPencil, FaSort, FaTrash } from 'react-icons/lib/fa';
...
     };
   }
 
   componentDidMount() {
-    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, this.state.sort + ',' + this.state.order);
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, `${this.state.sort},${this.state.order}`);
   }
 
-  sort = prop => e => {
+  sort = prop => () => {
```
deepu105: I suggested to simplify the method, plz look at my previous comment the branching is not needed

qmonmert:
@deepu105 the tests pass on my local machine

deepu105:
I'll take a look later

qmonmert:
ok

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -21,30 +21,59 @@ import { connect } from 'react-redux';
...
   constructor(props) {
     super(props);
+    this.state = {
+      ...getSortState(props.location)
+    };
   }
 
   componentDidMount() {
-    this.props.getUsers();
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, `${this.state.sort},${this.state.order}`);
```
deepu105: can we extract this line to a method as below and call the method here and in other places?
```
getUsers = () => {
  const { activePage, itemsPerPage, sort, order } = this.state;
  this.props.getUsers(activePage - 1, itemsPerPage, `${sort},${order}`);
}
```

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx
```diff
@@ -21,30 +21,59 @@ import { connect } from 'react-redux';
...
+      sort: prop
+    }, () => this.sortUsers());
+  }
+
+  sortUsers() {
+    this.props.getUsers(this.state.activePage - 1, this.state.itemsPerPage, `${this.state.sort},${this.state.order}`);
+    this.props.history.push(`${this.props.location.pathname}?page=${this.state.activePage}&sort=${this.state.sort},${this.state.order}`);
+  }
+
+  handlePagination = eventKey => {
```
deepu105: change to
```
handlePagination = activePage => this.setState({ activePage }, () => this.sortUsers());
```

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
```diff
@@ -0,0 +1,118 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
```
deepu105: unused import

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { Pagination, PaginationItem, PaginationLink } from 'reactstrap';
+
+export interface IPaginationComponentProps {
+  items: number;
+  activePage: number;
+  onSelect: any;
```
deepu105: `onSelect: (page: number) => void`

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  nextPage = () => {
+    this.setState({ currentPage: this.state.currentPage + 1 });
+    this.props.onSelect(this.state.currentPage + 1);
+  }
+
+  itemsToDisplay = activePage => {
+    const items = [];
+    let item: any = {};
```
deepu105: `: any` is not needed as its inferred from value

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
```diff
@@ -0,0 +1,118 @@
...
+              this.displayPaginationItem(i)
+              :
+              (
+                (paginationItem.display === 'disabled')
+                ?
+                <PaginationItem disabled key={i}>
+                  <PaginationLink href="#">...</PaginationLink>
+                </PaginationItem>
+                :
+                ''
```
deepu105: use `null` instead of `''`

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
```diff
@@ -0,0 +1,118 @@
...
+  activePage: number;
+  onSelect: any;
+  maxButtons: number;
+}
+
+export interface IPaginationComponentState {
+  currentPage: number;
+}
+
+export class PaginationComponent extends React.Component<IPaginationComponentProps, IPaginationComponentState> {
```
deepu105: :clap:

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-utils.ts
```diff
@@ -0,0 +1,38 @@
+import { ITEMS_PER_PAGE } from '../../shared/util/pagination.constants';
+
+export interface IPaginationState {
```
deepu105: 2 space indent please in this file

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-component.tsx
qmonmert: I think it's mandatory if not I have this error
`TS2339: Property 'display' does not exist on type '{}'.`

deepu105: oh, ok. Strange though

generators/client/templates/react/src/main/webapp/app/shared/util/pagination-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+  if (sortParam !== '') {
+    sort = sortParam.split(',')[0];
+    order = sortParam.split(',')[1];
+  }
+  return { itemsPerPage, sort, order, activePage };
+};
+
+export const getItemsNumber = (totalItems, itemsPerPage) => {
+  const division = Math.round(totalItems / itemsPerPage);
```
hdurix: I think it should be `Math.floor(totalItems / itemsPerPage)`.

For example with `totalItems = 35` and `itemsPerPage = 20`, division is 2 so the return is 2+1=3 (instead of 2).

deepu105:
@hdurix the pagination util is now in the react-jhipster lib and this PR is updated to use that is merged so you can rebase your branch to master and should be able to use it

PR #6949, opened 2018-01-05
Fix spring profile yml substitution for apps using service discovery

PierreBesson:
Fix profile substitutions for all configurations that were broken :
- microservice, gateway, uaa with maven
- monolith + service discovery, it was actually working partially as the profile was correctly set in application.yml. However this is subtly wrong as it prevent from retrieving the right profile specific configuration from the config server.


PierreBesson:
Note to self : I should add some travis tests to check that profile substitution works correctly for all configs and does not get broken again in the future.

jdubois:
Let me try to fix this, it's my fault

PR #6956, opened 2018-01-06
React - Field search -Elastic search option (no pagination and sort)

qmonmert:
React 

Add the field search when the option **elastic search** is taken.

Pagination and sort are not handle on this PR.

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
@@ -39,6 +39,9 @@ import { messages, SERVER_API_URL } from '../../config/constants';
 
 export const ACTION_TYPES = {
   FETCH_<%= entityActionNamePlural %>: '<%= entityInstance %>/FETCH_<%= entityActionNamePlural %>',
+  <%_ if (searchEngine === 'elasticsearch') { _%>
+  SEARCH_<%= entityActionNamePlural %>: '<%= entityInstance %>/SEARCH_<%= entityActionNamePlural %>',
+  <%_ } _%>
   <%_ for (idx in relationships) {
     const relationshipFieldNamePlural = relationships[idx].relationshipFieldNamePlural;
     const otherEntityNamePlural = relationships[idx].otherEntityNamePlural;
@@ -76,6 +79,9 @@ _%>
     case REQUEST(ACTION_TYPES.FETCH_<%= otherEntityNamePlural %>):
 <%_ } _%>
     case REQUEST(ACTION_TYPES.FETCH_<%= entityActionNamePlural %>):
+    <%_ if (searchEngine === 'elasticsearch') { _%>
+    case REQUEST(ACTION_TYPES.SEARCH_<%= entityActionNamePlural %>):
+    <%_ } _%>
     case REQUEST(ACTION_TYPES.FETCH_<%= entityActionName %>):
       return {
         ...state,
@@ -99,6 +105,9 @@ _%>
     case FAILURE(ACTION_TYPES.FETCH_<%= otherEntityNamePlural %>):
 <%_ } _%>
     case FAILURE(ACTION_TYPES.FETCH_<%= entityActionNamePlural %>):
+    <%_ if (searchEngine === 'elasticsearch') { _%>
+    case FAILURE(ACTION_TYPES.SEARCH_<%= entityActionNamePlural %>):
+    <%_ } _%>
     case FAILURE(ACTION_TYPES.FETCH_<%= entityActionName %>):
     case FAILURE(ACTION_TYPES.CREATE_<%= entityActionName %>):
     case FAILURE(ACTION_TYPES.UPDATE_<%= entityActionName %>):
@@ -127,6 +136,14 @@ _%>
         loading: false,
         entities: action.payload.data
       };
+    <%_ if (searchEngine === 'elasticsearch') { _%>
+    case SUCCESS(ACTION_TYPES.SEARCH_<%= entityActionNamePlural %>):
+      return {
+        ...state,
+        loading: false,
+        entities: action.payload.data
+    };
+    <%_ } _%>
     case SUCCESS(ACTION_TYPES.FETCH_<%= entityActionName %>):
       return {
         ...state,
@@ -154,6 +171,9 @@ _%>
 };
 
 const apiUrl = <% if (applicationType === 'gateway' && locals.microserviceName) { %>'/<%= microserviceName.toLowerCase() %>/<% } else if (authenticationType === 'uaa') { %>'<% } else { %>SERVER_API_URL + '<% } %>/api/<%= entityApiUrl %>';
+<%_ if (searchEngine === 'elasticsearch') { _%>
+const apiSearchUrl = <% if (applicationType === 'gateway' && locals.microserviceName) { %>'/<%= microserviceName.toLowerCase() %>/<% } else if (authenticationType === 'uaa') { %>'<% } else { %>SERVER_API_URL + '<% } %>/api/_search/<%= entityApiUrl %>';
+<%_ } _%>
 
 // Actions
 
@@ -171,6 +191,13 @@ export const getEntities: ICrudGetAction = (page, size, sort) => ({
   type: ACTION_TYPES.FETCH_<%= entityActionNamePlural %>,
   payload: axios.get(`${apiUrl}?cacheBuster=${new Date().getTime()}`)
 });
+<%_ if (searchEngine === 'elasticsearch') { _%>
+
+export const getSearchEntities: ICrudGetAction = query => ({
+  type: ACTION_TYPES.SEARCH_<%= entityActionNamePlural %>,
+  payload: axios.get(`${apiSearchUrl}?query=` + query)
+});
+<%_ } _%>
 
 export const getEntity: ICrudGetAction = id => {
   const requestUrl = `${apiUrl}/${id}`;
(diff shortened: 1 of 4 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
```diff
@@ -154,6 +171,9 @@ _%>
 };
 
 const apiUrl = <% if (applicationType === 'gateway' && locals.microserviceName) { %>'/<%= microserviceName.toLowerCase() %>/<% } else if (authenticationType === 'uaa') { %>'<% } else { %>SERVER_API_URL + '<% } %>/api/<%= entityApiUrl %>';
+<%_ if (searchEngine === 'elasticsearch') { _%>
+const apiSearchUrl = <% if (applicationType === 'gateway' && locals.microserviceName) { %>'/<%= microserviceName.toLowerCase() %>/<% } else if (authenticationType === 'uaa') { %>'<% } else { %>SERVER_API_URL + '<% } %>/api/_search/<%= entityApiUrl %>';
```
deepu105: we need to simplify this part following what is done in angular recently to use a single variable instead of all these checks

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -19,23 +19,34 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import { Button } from 'reactstrap';
+import { Button<%_ if (searchEngine === 'elasticsearch') { _%>, InputGroup<%_ } _%>
```
deepu105: use `<% %>` for inline templates as `<%_ _%>` slurps whitespace and newline

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -19,23 +19,34 @@
...
+ } from 'reactstrap';
```
deepu105: put this in same line above
`import { Button<% if (searchEngine === 'elasticsearch') { %>, InputGroup<% } %> } from 'reactstrap';`

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -19,23 +19,34 @@
...
+<%_ if (searchEngine === 'elasticsearch') { _%>
+import { AvForm, AvGroup, AvInput } from 'availity-reactstrap-validation';
+<%_ } _%>
 // TODO import TextFormat only when fieldContainsDate
 // tslint:disable-next-line:no-unused-variable
-import { Translate, ICrudGetAction, TextFormat } from 'react-jhipster';
-import { FaPlus, FaEye, FaPencil, FaTrash } from 'react-icons/lib/fa';
+import { Translate, translate, ICrudGetAction, TextFormat } from 'react-jhipster';
+import { FaPlus, FaEye, FaPencil, FaTrash<%_ if (searchEngine === 'elasticsearch') { _%>, FaSearch<%_ } _%>
```
deepu105: same comment as above here

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -45,11 +56,25 @@ export interface I<%= entityReactName %>Props {
...
-export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, undefined> {
+export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, I<%= entityReactName %>State> {
 
   constructor(props) {
     super(props);
+    <%_ if (searchEngine === 'elasticsearch') { _%>
+    this.state = {
+      search: ''
+    };
+    this.search = this.search.bind(this);
```
deepu105: no need to bind just use arrow methods instead

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -61,6 +86,25 @@ export class <%= entityReactName %> extends React.Component<I<%= entityReactName
     this.props.get<%= otherEntityNamePlural %>();
     <%_ } _%>
   }
+  <%_ if (searchEngine === 'elasticsearch') { _%>
+
+  search() {
```
deepu105: use arrow method
`search = () => {`

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -61,6 +86,25 @@ export class <%= entityReactName %> extends React.Component<I<%= entityReactName
...
+    if (this.state.search) {
+      this.props.getSearchEntities(this.state.search);
+    }
+  }
+
+  clear() {
```
deepu105: use arrow method

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -61,6 +86,25 @@ export class <%= entityReactName %> extends React.Component<I<%= entityReactName
...
+    this.props.getEntities();
+    this.setState({
+      search: ''
+    });
+  }
+
+  handleSearch(event) {
```
deepu105: use arrow method

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
qmonmert: there is a lint warning when I use arrow methods, so I use this ...

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
qmonmert: we have the same problem on _entity.service.ts (angular) ...

deepu105: oh ok, i think the improvement was done only on the non entity services like this `private resourceUrl = SERVER_API_URL + '<%- apiUaaPath %>api/users';` anyway then lets leave it for now

deepu105:
@qmonmert can you check the travis failure

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
qmonmert: ok

qmonmert:
@deepu105 normally I fix ;)

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -61,7 +61,7 @@ export interface I<%= entityReactName %>State {
 }
 <%_ } _%>
 
-export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, I<%= entityReactName %>State> {
+export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, <% if (searchEngine === 'elasticsearch') { %>I<%= entityReactName %>State<% } else { %>undefined<% } %>> {
```
deepu105: can you change this to
`<I<%= entityReactName %>Props<% if (searchEngine === 'elasticsearch') { %>, I<%= entityReactName %>State<% } %>> {`

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -45,11 +54,22 @@ export interface I<%= entityReactName %>Props {
...
+  search: string;
+}
+<%_ } _%>
 
-export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, undefined> {
+export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, <% if (searchEngine === 'elasticsearch') { %>I<%= entityReactName %>State<% } else { %>undefined<% } %>> {
 
   constructor(props) {
     super(props);
+    <%_ if (searchEngine === 'elasticsearch') { _%>
```
deepu105: can you move this check to wrap the constructor as its not needed when there is no state

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -45,11 +54,22 @@ export interface I<%= entityReactName %>Props {
   <%_ } _%>
   match: any;
 }
+<%_ if (searchEngine === 'elasticsearch') { _%>
+
+export interface I<%= entityReactName %>State {
```
deepu105: this could be added to a base interface as it will be needed for all entity, but we can do that while adding sorting etc

qmonmert: we will do this in the other PR to avoid conflicts ;)

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.tsx
```diff
@@ -45,13 +54,27 @@ export interface I<%= entityReactName %>Props {
...
+
+export class <%= entityReactName %> extends React.Component<I<%= entityReactName %>Props, <% if (searchEngine === 'elasticsearch') { %>I<%= entityReactName %>State> {
 
   constructor(props) {
     super(props);
+    this.state = {
+      search: ''
+    };
   }
+<% } else { %>undefined> {
```
deepu105: I guess you misunderstood me, for a stateless component constructor and the undefined in generic is not needed. I'll update this

qmonmert: ah yes sorry you're right, you fix this ?

deepu105:
@qmonmert there are lint errors can you please fix?

PR #6958, opened 2018-01-07
k8s and oc - elk update to use jhipster-import-dashboards

srinivasa-vasu:
elk update to use jhipster-import-dashboards

### DIFF
--- a/generators/kubernetes/files.js
+++ b/generators/kubernetes/files.js
@@ -63,6 +63,9 @@ function writeFiles() {
                 this.template('console/_jhipster-logstash.yml', 'console/jhipster-logstash.yml');
                 this.template('console/_jhipster-console.yml', 'console/jhipster-console.yml');
                 this.template('console/_jhipster-dashboard-console.yml', 'console/jhipster-dashboard-console.yml');
+                if (this.composeApplicationType === 'microservice') {
+                    this.template('console/_jhipster-zipkin.yml', 'console/jhipster-zipkin.yml');
+                }
             }
         },
 
--- a/generators/kubernetes/files.js
+++ b/generators/kubernetes/files.js
@@ -59,7 +59,6 @@ function writeFiles() {
 
         writeJhipsterConsole() {
             if (this.jhipsterConsole) {
-                // this.template('console/_logstash-config.yml', 'console/logstash-config.yml');
                 this.template('console/_jhipster-elasticsearch.yml', 'console/jhipster-elasticsearch.yml');
                 this.template('console/_jhipster-logstash.yml', 'console/jhipster-logstash.yml');
                 this.template('console/_jhipster-console.yml', 'console/jhipster-console.yml');
--- a/generators/kubernetes/files.js
+++ b/generators/kubernetes/files.js
@@ -59,10 +59,11 @@ function writeFiles() {
 
         writeJhipsterConsole() {
             if (this.jhipsterConsole) {
-                this.template('console/_logstash-config.yml', 'console/logstash-config.yml');
+                // this.template('console/_logstash-config.yml', 'console/logstash-config.yml');
                 this.template('console/_jhipster-elasticsearch.yml', 'console/jhipster-elasticsearch.yml');
                 this.template('console/_jhipster-logstash.yml', 'console/jhipster-logstash.yml');
                 this.template('console/_jhipster-console.yml', 'console/jhipster-console.yml');
+                this.template('console/_jhipster-dashboard-console.yml', 'console/jhipster-dashboard-console.yml');
             }
         },
 
--- a/generators/kubernetes/templates/console/_jhipster-elasticsearch.yml
+++ b/generators/kubernetes/templates/console/_jhipster-elasticsearch.yml
@@ -16,26 +16,460 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-apiVersion: extensions/v1beta1
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: es-config
+  namespace: <%= kubernetesNamespace %>
+data:
+    discovery.type: zen
+    minimum.master.nodes: "1"
+    unicast.host: jhipster-elasticsearch-discovery
+    network.host: 0.0.0.0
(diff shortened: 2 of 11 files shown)


generators/kubernetes/files.js
```diff
@@ -59,10 +59,11 @@ function writeFiles() {
 
         writeJhipsterConsole() {
             if (this.jhipsterConsole) {
-                this.template('console/_logstash-config.yml', 'console/logstash-config.yml');
+                // this.template('console/_logstash-config.yml', 'console/logstash-config.yml');
```
deepu105: can we remove it if not needed?

jdubois:
Great! @PierreBesson do you have time to take a look & merge? Otherwise I'll do it tonight

PierreBesson:
Thanks for doing this @srinivasa-vasu ! I had started to do this but yours is better as it includes Openshift. It still lacks the Zipkin server but otherwise is really good.
@jdubois I'm testing and merging this soon. I will also upgrade to the latest Console with ELK v6, currently there are a few issues with dashboards but I will fix it soon.

srinivasa-vasu:
@PierreBesson yeah, will add Zipkin.

generators/kubernetes/files.js
PierreBesson: Yes it should be removed.

srinivasa-vasu:
@PierreBesson Have pushed Zipkin changes as well. I think few prompts and ingress require some modification. Those can be taken up separately, will do it over the weekend

generators/kubernetes/templates/console/_jhipster-zipkin.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: jhipster-zipkin
+        image: <%= DOCKER_JHIPSTER_ZIPKIN %>
+        ports:
+        - containerPort: 9411
+          name: http
+          protocol: TCP
+        env:
+        - name: ES_HOSTS
+          value: http://jhipster-elasticsearch:9200
+        # TODO - better to update to an Ingress endpoint
```
sendilkumarn: WDYT, is it also better to add `JAVA_OPTS` in `jhipster-elasticsearch` ?

PierreBesson:
@srinivasa-vasu Just wait a few days for me to do some testing in minikube (I might add some little things too). In the meantime, @sendilkumarn has upgraded generator to v3, so can you please rebase this PR.

PierreBesson:
Just some thought here, actually the jhipster-elasticsearch is configured to be in elasticsearch dev mode. Ideally we should setup a "prod" (multi node) elasticsearch with k8s. We could try using something like the elasticsearch operator : https://github.com/upmc-enterprises/elasticsearch-operator.

sendilkumarn:
As discussed I agree to this completely. If not clusters atleast we have to add some memory to it. 

Deploying in k8 with console simply throws an error now.

a simple  XmX512M...

srinivasa-vasu:
@sendilkumarn JAVA_OPTS is better to add to all of the PODs till those get to run on Java9. What's the error that you are facing now?

srinivasa-vasu:
@PierreBesson Any reason why Prometheus operator manifest is not included in the generator? Since serviceaccount and prometheus resources are dependent on the custom resource provided by the operator, I think it would be better to have the operator as part of generator. Latest Operator has moved to CRD from TPR and the apiVersion has moved away from alpha. WDYT?

sendilkumarn:
@srinivasa-vasu By default, OpenJDK allocates `-Xms256m -Xmx256m` (a very low memory value) for the elastic search pod in kubernetes. And deploying them causes the pods to die.

srinivasa-vasu:
@sendilkumarn I doubt it. It is limited by the namespace's default (limitrange and quota), if you don't explicitly set it. Can you check those resources?

srinivasa-vasu:
@PierreBesson - I had a look at elasticsearch operator. It doesn't seem as generic as prometheus implementation. So I have resorted to statefulset cluster implementation and pushed the same.

generators/kubernetes/templates/console/_jhipster-elasticsearch.yml
```diff
@@ -16,41 +16,474 @@
...
-      - image: <%= DOCKER_JHIPSTER_ELASTICSEARCH %>
-        name: jhipster-elasticsearch
+      - name: jhipster-elasticsearch-master
+        image: <%= DOCKER_JHIPSTER_ELASTICSEARCH %>
+        env:
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: node.name
```
sendilkumarn: This will break applications that are using kubernetes version less than 1.8

srinivasa-vasu: Yes, supported in 1.8 and above

sendilkumarn: Interesting do we have this documented somewhere. I mean that we support only 1.8 and above? 

Since by default the kubernetes client (atleast in europe zone) is v1.7.11 so this throws an error.

srinivasa-vasu: yep, document needs to be updated. I thought about it. As it involves changing the config properties this is the way to go as es limits env prop by specifying the directory path instead of a specific file.

PierreBesson:
@srinivasa-vasu Great work on the Elasticsearch statefulset ! This is exactly what I had in mind yet I could would never have managed to do it myself.
No problem with requiring k8s 1.8, as it seems that k8s clusters are getting easier and easier to upgrade and keep up to date with new tooling.

@srinivasa-vasu Also you mention java 9 as being able to automatically set mem limits from docker config but actually this is already available in java > 8u131 with [`-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap`](https://blogs.oracle.com/java-platform-group/java-se-support-for-docker-cpu-and-memory-limits). I wanted to suggest moving to this but I've never tried it myself because of lack of time.

PR #7002, opened 2018-01-18
React : Alerts notifications

qmonmert:
PR in progress

React : 

Display alert notifications after CRUD operations


generators/client/templates/react/src/main/webapp/app/config/constants.ts
```diff
@@ -9,8 +9,11 @@ export const SERVER_API_URL = '';
 
 export const messages = {
   DATA_CREATE_SUCCESS_ALERT: 'Data created successfully!',
+  DATA_CREATE_ERROR_ALERT: 'Data not created!',
   DATA_DELETE_SUCCESS_ALERT: 'Data deleted successfully!',
-  DATA_UPDATE_ERROR_ALERT: 'Data deleted successfully!'
+  DATA_DELETE_ERROR_ALERT: 'Data not deleted!',
```
deepu105: use i18n values when i18n enabled using the translate methods

qmonmert:
@deepu105 notifications display well for entities and userManagement but not for account, an idea ?

ruddell:
@qmonmert I think this is intended behavior.  The entity and user management resources send alerts to the UI like so:
 - [UserResource](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_UserResource.java#L156)
 - [EntityResource](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-server/templates/src/main/java/package/web/rest/_EntityResource.java#L97)

AccountResource does not send alerts when changing passwords or other account actions.  Those responses/errors are handled in the UI
 - [AccountResource](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/_AccountResource.java#L350)

qmonmert:
@ruddell you're right

qmonmert:
There is a problem : in settings.json the property success is 
`<strong>Settings saved!</strong>`
so in the notification, the word strong appears

(the same for password)

generators/client/templates/react/src/main/webapp/app/config/constants.ts
```diff
@@ -8,9 +8,7 @@ export default config;
 export const SERVER_API_URL = '';
 
 export const messages = {
-  DATA_CREATE_SUCCESS_ALERT: 'Data created successfully!',
-  DATA_DELETE_SUCCESS_ALERT: 'Data deleted successfully!',
-  DATA_UPDATE_ERROR_ALERT: 'Data deleted successfully!'
+  DATA_ERROR_ALERT: 'Error'
```
deepu105: May be bit more expressive, how about `Internal Error`

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts
```diff
@@ -13,12 +14,16 @@ export default () => next => action => {
...
+        toast.success(action.meta.successMessage);
+      } else {
+        toast.success(action.meta.successMessage[0] + response.action.payload.data.id);
+      }
     }
     return Promise.resolve(response);
   }).catch(error => {
     if (action.meta && action.meta.errorMessage) {
-      // toast(action.meta.errorMessage, { type: toast.TYPE.ERROR });
+      toast.error(action.meta.errorMessage);
```
deepu105: may be we should show the error message in the error caught here by default so
`toast.error(action.meta.errorMessage || error.message);` similar to how we handle it in Angular notification interceptor. And then in the reducers return the `meta.errorMessage` filed only if we have a specific message to overriride

generators/client/templates/react/src/main/webapp/app/reducers/account.ts
```diff
@@ -57,17 +57,17 @@ const apiUrl = '/api/account';
 export const saveAccountSettings: ICrudPutAction = account => ({
   type: ACTION_TYPES.UPDATE_ACCOUNT,
   meta: {
-    successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-    errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+    successMessage: translate('settings.messages.success'),
+    errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/client/templates/react/src/main/webapp/app/reducers/account.ts
```diff
@@ -57,17 +57,17 @@ const apiUrl = '/api/account';
...
   payload: axios.post(apiUrl, account)
 });
 
 export const savePassword: ICrudPutAction = password => ({
   type: ACTION_TYPES.UPDATE_PASSWORD,
   meta: {
-    successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-    errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+    successMessage: translate('password.messages.success'),
+    errorMessage: translate('password.messages.error')
```
deepu105: Here we have a specific message so we can override by setting here, so this is fine

generators/client/templates/react/src/main/webapp/app/reducers/index.ts
```diff
@@ -5,12 +5,14 @@ import { loadingBarReducer as loadingBar } from 'react-redux-loading-bar';
 import locale from './locale';
 <%_ } _%>
 import layout from './layout';
+import account from './account';
```
deepu105: I think the settings pages are still using functions from `authentication.ts` file. we should change them to refer this

generators/client/templates/react/src/main/webapp/app/reducers/user-management.ts
```diff
@@ -122,8 +122,8 @@ export const createUser: ICrudPutAction = user => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.CREATE_USER,
     meta: {
-      successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('userManagement.created', { param: user.id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/client/templates/react/src/main/webapp/app/reducers/user-management.ts
```diff
@@ -135,8 +135,8 @@ export const updateUser: ICrudPutAction = user => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.UPDATE_USER,
     meta: {
-      successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('userManagement.updated', { param: user.id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/client/templates/react/src/main/webapp/app/reducers/user-management.ts
```diff
@@ -149,8 +149,8 @@ export const deleteUser: ICrudDeleteAction = id => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.DELETE_USER,
     meta: {
-      successMessage: messages.DATA_DELETE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('userManagement.deleted', { param: id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
```diff
@@ -211,8 +211,8 @@ export const createEntity: ICrudPutAction = entity => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.CREATE_<%= entityActionName %>,
     meta: {
-      successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('jhipsterApp.<%= entityInstance %>.created', { param: entity.id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
```diff
@@ -224,8 +224,8 @@ export const updateEntity: ICrudPutAction = entity => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.UPDATE_<%= entityActionName %>,
     meta: {
-      successMessage: messages.DATA_CREATE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('jhipsterApp.<%= entityInstance %>.updated', { param: entity.id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/entity-client/templates/react/src/main/webapp/app/entities/_entity.reducer.ts
```diff
@@ -238,8 +238,8 @@ export const deleteEntity: ICrudDeleteAction = id => async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.DELETE_<%= entityActionName %>,
     meta: {
-      successMessage: messages.DATA_DELETE_SUCCESS_ALERT,
-      errorMessage: messages.DATA_UPDATE_ERROR_ALERT
+      successMessage: translate('jhipsterApp.<%= entityInstance %>.deleted', { param: id }),
+      errorMessage: messages.DATA_ERROR_ALERT
```
deepu105: Lets use the error message from server as I said above, so no need to set this

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts
```diff
@@ -13,12 +14,16 @@ export default () => next => action => {
    */
   return next(action).then(response => {
     if (action.meta && action.meta.successMessage) {
-      // toast(action.meta.successMessage, { type: toast.TYPE.SUCCESS });
+      if (typeof action.meta.successMessage === 'string') {
```
deepu105: I dont understand this, what is the else part for?

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts
deepu105: The logic for this file should mimic `generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/_notification.interceptor.ts` and the error handling should also handle the logic here `/generators/client/templates/angular/src/main/webapp/app/shared/alert/_alert-error.component.ts`

deepu105:
@qmonmert btw thanks for putting up with my painful reviews :smile: hope its useful and not irritating for you

deepu105:
@qmonmert the issue about `<strong>Settings saved!</strong>` ideally the translate function should handle it as I wrote some html parsing for it. Lets check that once the feature is ready

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts
qmonmert: it's a hack because in creation mode we pass in meta : 
`successMessage: translate('jhipsterApp.projet.created', { param: entity.id })`
but entity.id is undefined because the entity isn't created

generators/client/templates/react/src/main/webapp/app/reducers/index.ts
qmonmert: you want merge account.ts and autentication.ts in one file authentication.ts ?

qmonmert:
@deepu105 starting next week, I will be less active on the project because I start a new mission 

cc @jdubois

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts
deepu105: We should get the values from response headers then

deepu105: That means ee can move the translation step into the interceptor

generators/client/templates/react/src/main/webapp/app/reducers/index.ts
deepu105: No account related should be in account and we should remove it from authentication

deepu105:
@qmonmert thats ok you have helped a lot already

Issue #6988, opened 2018-01-15
Not allowed to navigate top frame to data URL error when trying to open a blob object on Google Chrome


stefanobertini:
##### **Overview of the issue**

When opening a Blob Attachment, Chrome shows an empty page.
Looking at the console you can find this error:
**http://localhost:8080/app/components/util/data-util.service.js:66 Not allowed to navigate top frame to data URL**

##### **Motivation for or Use Case**

Seems related to this: https://groups.google.com/a/chromium.org/forum/#!searchin/blink-dev/top$20level$20data/blink-dev/GbVcuwg_QjM/JY6uJ6Q4BwAJ

##### **Reproduce the error**

Just create an entity with a blob field.
Run the app and add an entry to that entity, attaching a file.
Save the new entity and open it again.

Pressing "open" on the blob field causes the error.

##### **JHipster Version(s)**

4.13.3

##### **JHipster configuration**

Using JHipster version installed locally in current project's node_modules
Executing jhipster:info
Options:
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
gestione-cv@0.0.0 /Users/sbertini/Progetti/Plansoft/GestioneCV
└── generator-jhipster@4.13.3

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "it.plan.gestionecv",
      "nativeLanguage": "it"
    },
    "jhipsterVersion": "4.13.3",
    "baseName": "GestioneCV",
    "packageName": "it.plan.gestionecv",
    "packageFolder": "it/plan/gestionecv",
    "serverPort": "8080",
    "authenticationType": "session",
    "hibernateCache": "no",
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "searchEngine": "elasticsearch",
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "enableSwaggerCodegen": false,
    "rememberMeKey": "replaced-by-jhipster-info",
    "clientFramework": "angular1",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
[...]

PR #7006, opened 2018-01-18
added fix as its already used in angularx for opening blobs

NilsWild:
Fix for https://github.com/jhipster/generator-jhipster/issues/6988

### DIFF
--- a/generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
@@ -78,7 +78,7 @@
 
         function openFile (type, data) {
             var fileURL = 'data:' + contentType +';base64,' + data;
-            const win = window.open();
+            var win = window.open();
             win.document.write(
                 '<iframe src="' +
                 fileURL +
--- a/generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
+++ b/generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
@@ -77,7 +77,13 @@
         }
 
         function openFile (type, data) {
-            $window.open('data:' + type + ';base64,' + data, '_blank', 'height=300,width=400');
+            var fileURL = 'data:' + contentType +';base64,' + data;
+            const win = window.open();
+            win.document.write(
+                '<iframe src="' +
+                fileURL +
+                '" frameborder="0" style="border:0; top:0px; left:0px; bottom:0px; right:0px; width:100%; height:100%;" allowfullscreen></iframe>'
+            );
         }
 
         function toBase64 (file, cb) {


deepu105:
Next time say Fix #6988 so that issue autoclose when pr is merged

NilsWild:
Oh sorry

NilsWild:
Oh sorry

generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
```diff
@@ -77,7 +77,13 @@
         }
 
         function openFile (type, data) {
-            $window.open('data:' + type + ';base64,' + data, '_blank', 'height=300,width=400');
+            var fileURL = 'data:' + contentType +';base64,' + data;
```
sendilkumarn: `contentType` is wrong it should be `type`

generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
```diff
@@ -77,7 +77,13 @@
...
+            var win = window.open();
```
sendilkumarn: you should use `$window.open('','_blank')`

generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
deepu105: Oops, I did not check damn. You are right. @NilsWild can you please fix it

deepu105: how did this pass linting :thinking:

generators/client/templates/angularjs/src/main/webapp/app/components/util/_data-util.service.js
NilsWild: this should then be fixed here as well?  https://github.com/jhipster/ng-jhipster/blob/master/src/service/data-util.service.ts

stefanobertini: There is an error in the fix.

The name of the first parameter of the function is "type" but on line 80 it is used as "contentType".

Apart from this, the fix works !

Thanks a lot for your work!

NilsWild: @stefanobertini i know thats already corrected
anyway here  https://github.com/jhipster/ng-jhipster/blob/master/src/service/data-util.service.ts
@sendilkumarn s open comman is not used maybe update that although i am not sure if that is necessary

sendilkumarn: aah good catch. This should be updated

deepu105: @sendilkumarn can you first check the behavior without `_blank` I have a feeling it works the same since we use window.open

deepu105: check here https://developer.mozilla.org/en-US/docs/Web/API/Window/open

sendilkumarn: window.open it will work. But with _blank is preferred.

PR #7017, opened 2018-01-19
Rework on Audits components

### LINKED ISSUE #6974: Rework on Audits component

The audits view shows which user connected to the application and when. You can filter by date and you can sort by all fields (timestamp, principal, type, data.message).
However, only the events for the current page are sorted because the table is sorted in the Angular component: see https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.ts#L55.

This behavior is very disturbing for the app administration. The main use case when checking the audit event logs is for me to sort them by dates. When you sort the table, you can't imagine the data is sorted only for the current page, so for example you may think the last events were done many days ago (and not see the real last events).

##### **Reproduce the error**

Just create an app, login many times (> 20 times, or change the pagination constants), go to `http://localhost:8080/#/audits` and try to sort the table. Notice you have only the data for the current page.

##### **Suggest a Fix**

The sort should be done by the server (the request is already pageable so we just have to specify the sort param in the web service).
But I reckon that all this page should be modified to be similar to other CRUD list views (entity with pagination / user management) because all the mechanism is already done in those pages and the generated app would be more homogeneous (so less code, bugs and maintenance burden).

Concerning the different front-end frameworks:
- Angular: should be done first
- AngularJS: could stay as it is
- React: filters/sort part seems not implemented yet, and it would be simpler to do it with this solution

The only regression could be for the column "Extra data" as it would be not easy to sort the events by "data.message" (`data` is a map stored in an other table) but maybe it's not a big issue if the table is not sortable by this column.

What do you think?

I can do the Pull Request for that.
It could also be a "good first issue" if someone is interested. 

##### **JHipster Version(s)**

All.

hdurix:
- Use same mechanism as user management
- Audit table is not sortable by Extra data
- Cassandra is ignored as audits are not enabled for it

Fix #6974

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html
@@ -26,27 +26,28 @@
                 <div class="input-group-prepend">
                     <span class="input-group-text" jhiTranslate="audits.filter.from">from</span>
                 </div>
-                <input type="date" class="form-control" name="start" [(ngModel)]="fromDate" (ngModelChange)="onChangeDate($event)" required/>
+                <input type="date" class="form-control" name="start" [(ngModel)]="fromDate" (ngModelChange)="transition()" required/>
 
                 <div class="input-group-append">
-                        <span class="input-group-text" jhiTranslate="audits.filter.to">To</span>
-                    </div>
-                    <input type="date" class="form-control" name="end" [(ngModel)]="toDate" (ngModelChange)="onChangeDate($event)" required/>
+                    <span class="input-group-text" jhiTranslate="audits.filter.to">To</span>
+                </div>
+                <input type="date" class="form-control" name="end" [(ngModel)]="toDate" (ngModelChange)="transition()" required/>
             </div>
         </div>
     </div>
 
     <div class="table-responsive">
-        <table class="table table-sm table-striped table-bordered">
+        <table class="table table-sm table-striped">
             <thead>
-            <tr>
-                <th (click)="orderProp = 'timestamp'; reverse=!reverse"><span jhiTranslate="audits.table.header.date">Date</span></th>
-                <th (click)="orderProp = 'principal'; reverse=!reverse"><span jhiTranslate="audits.table.header.principal">User</span></th>
-                <th (click)="orderProp = 'type'; reverse=!reverse"><span jhiTranslate="audits.table.header.status">State</span></th>
-                <th (click)="orderProp = 'data.message'; reverse=!reverse"><span jhiTranslate="audits.table.header.data">Extra data</span></th>
+            <tr jhiSort [(predicate)]="predicate" [(ascending)]="reverse" [callback]="transition.bind(this)">
+                <th jhiSortBy="auditEventDate"><span jhiTranslate="audits.table.header.date">Date</span> <span class="fa fa-sort"></span></th>
+                <th jhiSortBy="principal"><span jhiTranslate="audits.table.header.principal">User</span> <span class="fa fa-sort"></span></th>
+                <th jhiSortBy="auditEventType"><span jhiTranslate="audits.table.header.status">State</span> <span class="fa fa-sort"></span></th>
+                <th><span jhiTranslate="audits.table.header.data">Extra data</span></th>
             </tr>
             </thead>
-            <tr *ngFor="let audit of getAudits()">
+            <tbody>
+            <tr *ngFor="let audit of audits">
                 <td><span>{{audit.timestamp| date:'medium'}}</span></td>
                 <td><small>{{audit.principal}}</small></td>
                 <td>{{audit.type}}</td>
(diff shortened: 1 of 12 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html
```diff
@@ -55,9 +56,10 @@ <h4 jhiTranslate="audits.filter.title">Filter by date</h4>
                     <span *ngIf="audit.data" ng-show="audit.data.remoteAddress"><span jhiTranslate="audits.table.data.remoteAddress">Remote Address</span> {{audit.data.remoteAddress}}</span>
                 </td>
             </tr>
+            </tbody>
         </table>
     </div>
-    <div *ngIf="audits">
+    <div>
```
sendilkumarn: why this condition is removed?

generators/client/templates/angular/src/test/javascript/spec/app/admin/audits/_audits.component.spec.ts
```diff
@@ -95,9 +95,9 @@ describe('Component Tests', () => {
                expect(comp.toDate).toBe(getDate());
                expect(comp.fromDate).toBe(getDate(false));
                expect(comp.itemsPerPage).toBe(ITEMS_PER_PAGE);
-               expect(comp.page).toBe(1);
+               expect(comp.page).toBe(10);
                expect(comp.reverse).toBeFalsy();
-               expect(comp.orderProp).toBe('timestamp');
+               expect(comp.predicate).toBe('id');
```
sendilkumarn: it has to be `timestamp` right ?

generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html
hdurix: This condition was useless because already present in the parent `div` (see https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/admin/audits/_audits.component.html#L19)

generators/client/templates/angular/src/test/javascript/spec/app/admin/audits/_audits.component.spec.ts
hdurix: No I think it has to be `id` because it is set by mock router (see https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/test/javascript/spec/helpers/_mock-route.service.ts#L33).

Nonetheless, I think you right, `timestamp` should be the default sort, see my last commit correcting this.

deepu105:
@hdurix by any chance will you be able to do tje same for the React part?

hdurix:
@deepu105 I could give it a try. I don't know react very much but I can take entity list as an example.

deepu105:
Yes please thank you

hdurix:
Even better, I will be able to use the new pagination util by @qmonmert from https://github.com/jhipster/generator-jhipster/pull/6946 !

hdurix:
I got the react part working in a project generated with #6946 code. I will add this to the generator once #6946 is merged because I reuse the same components.

React is nice to play with!

deepu105:
@hdurix i'll try to merge that so that you can rebase on to that and us it

hdurix:
@deepu105 I rebased and added react part

generators/client/templates/react/src/main/webapp/app/modules/administration/audits/audits.tsx
```diff
@@ -1,65 +1,146 @@
...
+      order: this.state.order === 'asc' ? 'desc' : 'asc',
+      sort: prop
+    }, () => this.transition());
   }
 
-  getAuditList = () => {
-    if (!this.props.isFetching) {
-      this.props.getAudits();
-    }
+  transition() {
```
deepu105: use const arrow method so that there is no bind issues.
`transition = () => {}`

generators/client/templates/react/src/main/webapp/app/shared/layout/header/header.tsx
```diff
@@ -9,7 +9,7 @@ import {
 import {
   FaHome, FaThList, FaUserPlus, FaUser, FaFlag, FaHeart,
   FaList, FaTasks, FaDashboard, FaBook, FaWrench, FaSignIn, FaSignOut,
-  FaClockO, FaHddO,
+  FaClockO, FaHddO, FaBell,
```
deepu105: add this after the // tslint:disable-next-line so that this doesn't fail when choosing Cassandra

deepu105:
@hdurix there are some lint errors on React part
```
ERROR: /home/travis/app/src/main/webapp/app/modules/administration/audits/audits.tsx[6, 3]: 'ICrudGetAction' is declared but its value is never read.
ERROR: /home/travis/app/src/main/webapp/app/modules/administration/audits/audits.tsx[92, 21]: 'onChangeFromDate' is declared but its value is never read.
ERROR: /home/travis/app/src/main/webapp/app/modules/administration/audits/audits.tsx[92, 39]: 'onChangeToDate' is declared but its value is never read.
```

hdurix:
@deepu105 see my last 2 commits.
Do you think the tests I added are useful enough? If yes, we can do the same for user management and entity management.

deepu105:
@hdurix yes any test is better than no tests :)

deepu105:
@hdurix will you be able to fix the failed build?

hdurix:
Sorry I didn't notice... It should be good now.

PR #7020, opened 2018-01-19
Generate i18n tooltip help from JDL comments

### LINKED ISSUE #6797: Generate i18n mouseover/tooltip help from JDL comments

When filling a JHipster form, we rely on a label and a entry field (textfield, combo...). But sometimes labels are not enough to give explanations. It would be good to generate some kind of i18n mouseover/tooltip help next to the label.

[JDL allows comments](http://www.jhipster.tech/jdl/#commentingjdl). So when we describe our model using JDL, this is what we can do:

```
/* This entity is complex and needs some documentation
*/
entity Complex {
    /** complex attribute */
    complexAttr1 String,
    /** another complex attribute */
    complexAttr2 Integer
}
```

These comments will be added to the Javadoc of the Entity, as well as the REST API

```
/**
 * This entity is complex and needs some documentation
 */
@ApiModel(description = "This entity is complex and needs some documentation")
@Entity
public class Complex implements Serializable {

    /**
     * complex attribute
     */
    @ApiModelProperty(value = "complex attribute")
    private String complexAttr1;

    /**
     * another complex attribute
     */
    @ApiModelProperty(value = "another complex attribute")
    private Integer complexAttr2;
```

For each comment, it would be great to generate an internationalized tooltip next to the label.

Dufgui:
Add tooltip based on javadoc field

Fix #6797

### DIFF
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-detail.component.html
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-detail.component.html
@@ -26,7 +26,7 @@
         const fieldName = fields[idx].fieldName;
         const fieldType = fields[idx].fieldType;
         const fieldTypeBlobContent = fields[idx].fieldTypeBlobContent; _%>
-        <dt><span jhiTranslate="<%= keyPrefix %>.<%= fieldName %>"><%= fields[idx].fieldNameHumanized %></span></dt>
+        <dt><span jhiTranslate="<%= keyPrefix %>.<%= fieldName %>"<% if (fields[idx].javadoc) { if (enableTranslation) { %> [ngbTooltip]="'<%= keyPrefix %>.help.<%= fieldName %>' | translate"<% } else { %> ngbTooltip="<%= fields[idx].javadoc %>"<% } } %>><%= fields[idx].fieldNameHumanized %></span></dt>
         <dd>
             <%_ if (fields[idx].fieldIsEnum) { _%>
             <span jhiTranslate="{{'<%= angularAppName %>.<%= fieldType %>.' + <%= entityInstance %>.<%= fieldName %>}}">{{<%= entityInstance %>.<%= fieldName %>}}</span>
--- a/generators/entity-client/templates/angularjs/src/main/webapp/app/entities/_entity-management-detail.html
+++ b/generators/entity-client/templates/angularjs/src/main/webapp/app/entities/_entity-management-detail.html
@@ -26,7 +26,7 @@
             const fieldName = fields[idx].fieldName;
             const fieldType = fields[idx].fieldType;
             const fieldTypeBlobContent = fields[idx].fieldTypeBlobContent; _%>
-        <dt><span data-translate="<%= keyPrefix %>.<%= fieldName %>"><%= fields[idx].fieldNameHumanized %></span></dt>
+        <dt><span data-translate="<%= keyPrefix %>.<%= fieldName %>"<% if (fields[idx].javadoc) { if (enableTranslation) { %> uib-tooltip="{{'<%= keyPrefix %>.help.<%= fieldName %>' | translate}}"<% } else { %> uib-tooltip="<%= fields[idx].javadoc %>"<% } } %>><%= fields[idx].fieldNameHumanized %></span></dt>
         <dd>
             <%_ if (fields[idx].fieldIsEnum) { _%>
             <span data-translate="{{'<%= angularAppName %>.<%= fieldType %>.' + vm.<%= entityInstance %>.<%= fieldName %>}}">{{vm.<%= entityInstance %>.<%= fieldName %>}}</span>
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-dialog.component.html
@@ -55,7 +55,7 @@ _%>
             }
         _%>
         <div class="form-group">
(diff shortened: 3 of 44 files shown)


generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-detail.component.html
```diff
@@ -26,7 +26,7 @@ <h2><span jhiTranslate="<%= keyPrefix %>.detail.title"><%= entityClassHumanized
         const fieldName = fields[idx].fieldName;
         const fieldType = fields[idx].fieldType;
         const fieldTypeBlobContent = fields[idx].fieldTypeBlobContent; _%>
-        <dt><span jhiTranslate="<%= keyPrefix %>.<%= fieldName %>"><%= fields[idx].fieldNameHumanized %></span></dt>
+        <dt><span jhiTranslate="<%= keyPrefix %>.<%= fieldName %>" <%_ if (fields[idx].javadoc) { _%> [ngbTooltip]="'<%= keyPrefix %>.help.<%= fieldName %>' | translate" <%_ } _%>><%= fields[idx].fieldNameHumanized %></span></dt>
```
deepu105: what about when i18n is disabled?

deepu105: use `<% %>` and not `<%_ _%>` as the first one is for inline and second one is for newline. Same for all other inline usage <%_ should never be used inline

generators/entity-client/templates/angularjs/src/main/webapp/app/entities/_entity-management-detail.html
```diff
@@ -26,7 +26,7 @@ <h2><span data-translate="<%= keyPrefix %>.detail.title"><%= entityClassHumanize
             const fieldName = fields[idx].fieldName;
             const fieldType = fields[idx].fieldType;
             const fieldTypeBlobContent = fields[idx].fieldTypeBlobContent; _%>
-        <dt><span data-translate="<%= keyPrefix %>.<%= fieldName %>"><%= fields[idx].fieldNameHumanized %></span></dt>
+        <dt><span data-translate="<%= keyPrefix %>.<%= fieldName %>" <%_ if (fields[idx].javadoc) { _%> uib-tooltip="{{'<%= keyPrefix %>.help.<%= fieldName %>' | translate}}" <%_ } _%>><%= fields[idx].fieldNameHumanized %></span></dt>
```
deepu105: I wouldnt bother with angularJS as we are planning to drop it

Dufgui: I know but it's quick to do it, so I do it and it will be integrate in the external generator in this way.

generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management-detail.component.html
Dufgui: ok i will do it.

Dufgui: Good question for i18n disabled. I have no idea how jhipster works in this case.
Is there any flag to detect it ? And in this case i put directly the javadoc value, ok ?

Dufgui:
Could i add a flag on context in generator base to know if current entity have a field with javadoc ? 
I want to do it to limite import in react which cause a problem in CI. And I don't think it 's a good idea to loop over all fields in react template to know if current entity have a javadoc field no ?

Dufgui:
Good question for i18n disabled. I have no idea how jhipster works in this case.
Is there any flag to detect it ? And in this case i put directly the javadoc value, ok ?

deepu105:
Yes plain javadoc is fine when noi18n, you can use the enableTranslation ctx variable. And yes you can add new context var for javadoc

Dufgui:
all tests pass in local so i create a double commit useless to relaunch travis. Because i don't understand the error in the travis log

Dufgui:
https://github.com/SitePen/remap-istanbul/issues/160

Dufgui:
CI test fail caused by the #7031

deepu105:
Our master builds are broken, not your fault. Can you squash your commits or edit messgaes to be more meaningful. Now all your commits look the same

Dufgui:
For the CI, i know I create the issue on it.
I squash them.  for the meaning, hope it's ok now.

Dufgui:
Please relaunch Travis without any Change

jdubois:
I'm doing a quick test - looks like a good idea, but I don't want our entities to become too complex, and I'm not sure this would suit everyone

jdubois:
OK this works well, but I'm afraid our entities become too complex if we keep adding new features - not sure this would be useful for most people...
Still I'm merging this as there is some good work and @agoncal seem to need it, and I'm not sure many people use the "javadoc" field anyway (I don't!), but next time please can we cast votes on adding new features like this, which will add code for everyone?

agoncal:
Just had some time to have a quick look at it. Looks great for me (didn't have time to test i18n but).

@jdubois when you say "our entities become too complex" are you talking about the templates? Because in terms of generated code, you go from this:

```
<dt><span>Title</span></dt>
```

To this 

```
<dt><span ngbTooltip="this is a comment for the title">Title</span></dt>
```

So if you don't use comments in JDL, you see no differences. Maybe people don't use Javadoc in JDL because they don't know much about it. When I discovered that my comments where inside the Swagger API comments, I loved it.

For us this is as important as i18n: users should have an app that is translated, as well as have some extra help.

jdubois:
Yes ok @agoncal - and if you have time to test, it should even work with Spring Boot 2 and React!

PR #7021, opened 2018-01-19
Remove --progress from production webpack build

jkutner:
The `--progress` option on webpack will fail for any non-interactive terminal. This breaks JHipster on CI servers, and Heroku (if deploying with `git`, which is not the default option today) with an error like:

```
[ERROR]  30% building modules 167/212 modules 45 active .sed: couldn't write 237439 items to stdout: Resource temporarily unavailable
```

For more information see https://github.com/webpack/webpack/issues/1553

If this change is not accepted, I recommend a new script entry `"webpack:prod:ci"` or similar. I will add this myself if desired.


deepu105:
I think it should be fine, ofcourse seeing progress would be nice but for prod its less important

jdubois:
Yes I just restarted the failing build - if that passes then let's merge this

pascalgrimaud:
it seems current master is broken, so nothing related to this PR

jdubois:
Yes my spring-boot-2 branch is also broken, and that doesn't look to my fault -> there are some Karma tests suddenly failing, I guess it's one more issue with a transitive NPM dependency (what a pain in the *** is NPM dependency management!)

PR #7026, opened 2018-01-20
add grafana with pre-configured dashboard view and few refactoring updates

srinivasa-vasu:
- Grafana added with pre-configured dashboard view
- Prometheus operator is included (as the app specific bits like service monitor and rbac objects are dependent on the operator crd)
- Aligned some of the prompts (jhipster console and prometheus) to be in sync with docker-compose
- Added a shell script pgm to deploy all of the k8s objects at one go
- Added init-container check to deployment object to wait for dependent services initialisation
- Output file structure is been refactored
- README and test spec updated accordingly

### DIFF
--- a/generators/docker-prompts.js
+++ b/generators/docker-prompts.js
@@ -253,7 +253,7 @@ function askForMonitoring() {
                 name: 'Yes, for metrics only with Prometheus (only compatible with JHipster >= v3.12)'
             }
         ],
-        default: 'no'
+        default: this.monitoring ? this.monitoring : 'no'
     }];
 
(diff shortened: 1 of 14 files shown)


deepu105:
@srinivasa-vasu @PierreBesson Since its a feature can we make it part of v5?

PierreBesson:
Wow huge work @srinivasa-vasu ! Yes let's have this for v5.

deepu105:
@PierreBesson can we merge this. The failing build is for React so should be ok. I restarted it anyway

PierreBesson:
Yes it can be merged without issue.

PR #7027, opened 2018-01-20
fix tslint semicolon config for Angular

ctamisier:
4d6bf48:
```
"semicolon": [
            "always"
]
```

should be

```
"semicolon": [
            true,
            "always"
]
```

5718bba:
fix tslint warning

### DIFF
--- a/generators/client/templates/angular/_tslint.json
+++ b/generators/client/templates/angular/_tslint.json
@@ -81,6 +81,7 @@
         ],
         "radix": true,
         "semicolon": [
+            true,
             "always"
         ],
         "triple-equals": [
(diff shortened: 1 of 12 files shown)


deepu105:
@ctamisier shount matter anymore as we will be using prettier soon. But thanks

pascalgrimaud:
With this change, there are a lot of new errors: https://travis-ci.org/jhipster/generator-jhipster/jobs/331473118#L3998
There are still some works to do before merging this.

ctamisier:
ouch indeed !
I am fixing it.
Can you confirm that this is something that we want a "always semicolon" ?
With prettier it will be automatically added, right ?
I think that prettier will be a value added for managing white-spaces.
For semicolon, as this is "textual", I think this is better to have it in the template as well.
(Might be confusing to see a semicolon in the generated code but note having it in the template ?!)

deepu105:
Yes, always semicolon should be true.

Issue #6773, opened 2017-11-30
[Feature Request] AWS Containerisation


ggotti:
##### **Overview of the issue**
The existing JHipster AWS Deployment option uses Elastic Beanstalk, which is no longer being encouraged as the way of deploying applications to AWS. It's also still in Beta, and there are some reliability issues with the sub-generator in its current state.

Based on the recent announcements at AWS re:Invent, we’d like to modify the generator to support the following deployment options (instead of Elastic Beanstalk):
- [AWS Fargate](https://aws.amazon.com/blogs/aws/aws-fargate/) : Initially for Monolithic applications only, with the option to run databases as either containers, or as Amazon RDS.
- [AWS EKS](https://aws.amazon.com/eks/)) (when it becomes available): Initially for Microservices only. We’ll potentially want to extend the existing Kubernetes deployment option to make it more AWS-centric.

Our current thinking is to re-use the existing plugin interface and structure, but modify the deployment to leverage CloudFormation more extensively (similar to how the [Serverless Framework](https://serverless.com/framework/docs/providers/aws/guide/resources/)  does it)

I believe supporting a Docker-first paradigm will help standardise the JHipster deployment options, and make it easier to deploy moving forward. [carvallegro](https://github.com/carvallegro)  and myself ([ggotti](https://github.com/ggotti)) and will be working on this throughout December. We’ll update this ticket with more details as we progress. Any assistance would be appreciated :)

##### **Motivation for or Use Case**
- AWS is fairly dominant in the Australia market, and the lack of modern AWS integration is a serious issue.
- Move the AWS plugin out of Beta. We've discovered some reliability issues with the current implementation.

deepu105:
You are welcome to do it :)

ggotti:
Just a status update for those that are interested. Our branch is located in [this repository](https://github.com/carvallegro/generator-jhipster), and our development branch is feature/aws-generator, if you'd like to run it. The generator itself is currently called "more-aws"; we'll rename is later on, probably to something like aws-containers as a way to maintain compatibility with the existing aws generator. We're currently only focusing on the monolith version (using Fargate).

**Status**
We've got the majority of prompts working, and talking back to AWS for valid value entry.
- SpringCloud integration with Amazon SSM
- We have the Cloud Formation scripts being uploaded into new s3 buckets, passwords being stored in SSM, and docker client login.
- Streaming of Cloud Formation log to terminal.
- At a high level, the Cloud Formation is able to create the following:
[...]

PR #7035, opened 2018-01-22
Feature/aws containers

ggotti:
This PR adds a new AWS specific sub-generator, called 'aws-containers', to deploy JHhipster applications to Amazon using Elastic Container Service (ECS). The version in this PR only works with monolithic applications. At a high level it does the following:
- 	Queries your AWS account for details around deployment topology: VPCs, subnets, database passwords.
- Introduces SpringCloud dependencies to allow your application to query Amazon's System Manager for credentials on startup.
- Uses a nested-stack CloudFormation arrangement to deploy your selected applications. 
- Creates an Elastic Container Registry (ECR) or private Docker repository to deploy applications to.

A detailed update of the various AWS technologies and services can be seen in this website PR. 

Thanks to @carvallegro for doing most of the work :)

Addresses monolithic component of #6773 


deepu105:
Lets do this for v5 as well since its a major feature

deepu105:
@jhipster/developers who can review this? I can do the JS part but not familiar with AWS containers

generators/aws-containers/README.md
```diff
@@ -0,0 +1,24 @@
+# AWS JHipster Generator - _A new hope_
+
+## AWS
+### ECR
+#### Authentication Token
+The token received by ECR to authenticate is structured as follow. We'll only take the first object.
```
MathieuAA: `follow` -> `followed`

generators/aws-containers/README.md
```diff
@@ -0,0 +1,24 @@
...
+     "proxyEndpoint":"https://foo.ecr.region.amazonaws.com"
+  }
+```
+
+## Development
+### Dependencies
+
+```
+# NPM
+npm i --save-dev aws-sdk@^2.167.0 progress@2.0.0 ora@1.3.0
```
MathieuAA: Can the versions be fixed for both yarn and npm?

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+    return new Promise((resolve, reject) => {
+        try {
+            AWS = require('aws-sdk') // eslint-disable-line
+            ProgressBar = require('progress'); // eslint-disable-line
+            ora = require('ora'); // eslint-disable-line
+        } catch (e) {
+            generator.log('Installing AWS dependencies');
+            let installCommand = 'yarn add aws-sdk@^2.167.0 progress@2.0.0 ora@1.3.0';
+            if (generator.config.get('clientPackageManager') === 'npm') {
+                installCommand = 'npm install aws-sdk@^2.167.0 progress@2.0.0 ora@1.3.0--save';
```
MathieuAA: If the version changes we'd have to update it here too... is there a better solution?

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+ * @returns {Promise<EC2.Region[]>}
+ */
+function listRegions() {
+    return spinner(ec2.describeRegions({})
+        .promise()
+        .then(data => data.Regions));
+}
+
+/**
+ * listVpcs(region) returns a Promise, which resolves to an array of AWS VPC objects
```
MathieuAA: The doc is not consistent with the code here

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+function listVpcs() {
+    return spinner(ec2.describeVpcs({})
+        .promise()
+        .then(data => data.Vpcs));
+}
+
+/**
+ * listSubnets(vpcId) returns a Promise, which resolves to an array of
+ * Subnets available within the supplied VPC
+ * @param region to look for subnets
```
MathieuAA: I guess you forgot to remove the param

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+/**
+ * Retrieve decoded information to authenticate to Docker with AWS credentials.
+ * @returns {Promise} Returns a promise that resolves when the informations are retrieved.
+ */
+function getDockerLogin() {
+    return spinner(new Promise((resolve, reject) => _getAuthorizationToken()
+        .then(authToken => sts.getCallerIdentity({}).promise()
+            .then((data) => {
+                const decoded = utils.decodeBase64(authToken.authorizationToken);
+                const splitResult = _.split(decoded, ':');
```
MathieuAA: lodash's `split` can be replaced by a regular `split`

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+    };
+    return spinner(new Promise((resolve, reject) => s3.headBucket({
+        Bucket: bucketName
+    }).promise()
+        .catch((error) => {
+            if (error.code !== 'NotFound') {
+                reject(new Error(`The S3 Bucket ${chalk.bold(bucketName)} in region ${chalk.bold(region)} already exists and you don't have access to it. Error code: ${chalk.bold(error.code)}`));
+            }
+        })
+        .then(() =>
```
MathieuAA: Can you call `then` before `catch` please?

jdubois:
Biggest issue is that you need AWS $$$$ for that

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+ * Upload the template in the S3Bucket
+ * @param bucketName S3 Bucket name to upload the template into
+ * @param filename Name to give to the file in the Bucket
+ * @param path Path to the file
+ * @returns {Promise}
+ */
+function uploadTemplate(bucketName, filename, path) {
+    return spinner(new Promise((resolve, reject) =>
+        fs.stat(path, (error, stats) => {
+            if (_.isNil(stats)) {
```
MathieuAA: As lodash's `isNil` just does `value == null`, I think you can just write `!stats` here

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+                    Body: fs.createReadStream(path)
+                },
+                {
+                    partSize: Math.max(stats.size, S3_MIN_PART_SIZE),
+                    queueSize: 1
+                }
+            );
+            let bar;
+            upload.on('httpUploadProgress', (evt) => {
+                if (bar === undefined && evt.total) {
```
MathieuAA: `!bar`

generators/aws-containers/aws-client.js
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/**
+ * Sanitize the bucketName following the rule found here:
+ * http://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html
+ * @param bucketName
+ * @returns {string}
+ */
+function sanitizeBucketName(bucketName) {
+    let labels = _.split(bucketName, '.');
```
MathieuAA: Two things:
  - Instead of using lodash, you can use the regular API
  - You can chain the calls

generators/aws-containers/lib/cf.js
```diff
@@ -0,0 +1,297 @@
...
+const _ = require('lodash');
+const chalk = require('chalk');
+
+const STACK_LISTENER_INTERVAL = 15000;
+const PHYSICAL_RESOURCE_SEPARATOR = ':';
+const STACK_EVENT_STATUS_DISPLAY_LENGTH = 35;
+
+let stdOut = message => console.error(message.trim()); // eslint-disable-line
+let stdErr = message => console.error(message.trim()); // eslint-disable-line
+module.exports = class CF {
```
MathieuAA: Is it okay to rename the class to `CloudFormation ` instead of using an abbreviation?

generators/aws-containers/lib/cf.js
```diff
@@ -0,0 +1,297 @@
...
+    constructor(region) {
+        this.cloudFormation = new AWS.CloudFormation({ region });
```
MathieuAA: If `CF` means `CloudFormation`, is the parameter an internal object?

generators/aws-containers/README.md
carvallegro: It's actually 'as follows' but thanks for the catch up :)

generators/aws-containers/aws-client.js
carvallegro: We'll definitely look into it but the issue with not specifying the version is that we don't have any control over the installed version. The code works with `2.167.0` so we're sure to avoid any breaking changes.

For example if `aws-sdk` has a 3.0.0 major release it might break the code. Hardcoding a version means that we only have to be careful at the npm version required by the `jhipster-gernerator` and update accordinlgy, which will happens less often than a `aws-sdk`update

generators/aws-containers/aws-client.js
carvallegro: It was originally like this but because of the way Promises behave we've had to put it first. If we put the `catch` `after` the `then` and the `s3.createBucket()` promise fails, it falls back to the `spinner()` which will change the expected behavior

generators/aws-containers/lib/cf.js
carvallegro: We've named it `cf` in order to be consistant with the other classes in `lib/` : `ecr` and `ssm`

generators/aws-containers/lib/cf.js
carvallegro: I'm not sure what you mean but here the constructor parameter, `region` is used to build `this.cloudFormation` which is an AWS component. It's a composition pattern.

generators/aws-containers/README.md
carvallegro: I'm not sure what you mean here. The installation instructions for both yarn and npm is the same

generators/aws-containers/README.md
MathieuAA: Right!! thanks

generators/aws-containers/README.md
deepu105: can we put the version numbers in a constant and use that in both places so that at least you only need to change it once

carvallegro: It is not possible to use variables in Markdown files but we could put constants in `aws-client.js` and maybe put a TODO comment to remind to update the doc

generators/aws-containers/aws-client.js
MathieuAA: That's weird.

generators/aws-containers/lib/cf.js
MathieuAA: I know. What I meant is I'd rather change the file's name (and the class' name too) and rename the `couldFormation` field.

generators/aws-containers/README.md
MathieuAA: By fixed I meant without `^` in the versions, like `npm i --save-dev aws-sdk@2.167.0 progress@2.0.0 ora@1.3.0`

deepu105: Oh sorry didn't notice that this was the readme of the sub gen itself

carvallegro:
@MathieuAA I've pushed some changes around your comments

generators/generator-base.js
```diff
@@ -1447,7 +1460,19 @@ module.exports = class extends PrivateBase {
...
+    /**
+     * A new dependency to build.gradle file in a specific folder.
+     *
+     * @param {string} scope - scope of the new dependency, e.g. compile
+     * @param {string} group - maven GroupId
+     * @param {string} name - maven ArtifactId
+     * @param {string} version - (optional) explicit dependency version number
+     */
+    addGradleDependencyInDirectory(directory, scope, group, name, version) {
+        const fullPath = `${directory}/build.gradle`;
```
MathieuAA: @deepu105 Maybe the use of the `path` module would be good here, don't you think?

generators/docker-utils.js
```diff
@@ -0,0 +1,96 @@
...
+ * @property forceBuild flag to force the image build.
+ * @property appConfig Configuration for the current application
+ * @returns {Promise.<TResult>|Promise}
+ */
+function checkAndBuildImages(opts = { cwd: './', forceBuild: false, appConfig: { buildTool: 'gradle' } }) {
+    if (this.abort) return null;
+    checkImageExist.call(this, opts);
+    const pwd = shelljs.pwd();
+    shelljs.cd(opts.cwd);
+    return new Promise((resolve, reject) => dockerCLI.command(`${opts.cwd}${this.dockerBuildCommand}`, (err) => {
```
MathieuAA: Can you format here?
```javascript
return new Promise((resolve, reject) =>
  dockerCLI.command(`${opts.cwd}${this.dockerBuildCommand}`, (err) => {
```

jdubois:
I will do a review of this code on Tuesday -> @ggotti could you also prepare some documentation for this?
Something like http://www.jhipster.tech/aws/ (the source is at https://github.com/jhipster/jhipster.github.io - and don't worry if this is not perfect graphically etc, what I need is documentation, I'll format and correct it myself)

jdubois:
Please have a look at my reviews, and update jhipster-dependencies.

Now I can deploy, but I have 2 major issues:

- My database connection fails, I got ` SQLException occurred while connecting to jhipster-test-ecs-tmp4-1ob0zht-jhipsterrdscluster-1lw3q3us4xyr7.cluster-co7tgvzcrcxe.us-east-1.rds.amazonaws.com:3306`- this is with Aurora/Postgresql. Please note I can see the database in Amazon RDS, so it's a configuration issue on our side.
- I'm worried that this created lots of services, and I have no idea how to clean up what was created. Can we have some kind of "tag" somewhere, to know what was deployed, how much does it costs, and delete everything easily?

jdubois:
You will also need to update the limitations in the documentation:

- this doesn't work with the JHipster Registry.
- if you use a cache, it will not scale (because no JHipster Registry), so it's easier a non-scalable ehcache cache, or no cache. That's a bit annoying.
- this only supports SQL databases with Aurora -> no MongoDB, Cassandra, etc.... so this should explained upfront

-> what would be even better is to check this when the sub-generator runs, and make it fail with a nice error message if there's something wrong

ggotti:
I'll look at actioning the PR feedback early next week, as I'm currently studying for an exam. In terms of comments in this thread. @jdubois 
-I'll look at adding tagging support in to the components where possible. However it should be noted that you should use the CloudFormation delete functionality where possible, as it makes it much easier to cleanup.
- There is an associated documentation PR [here](https://github.com/jhipster/jhipster.github.io/pull/532).
- I'll look at adding a pre-check in for the database configuration for ensure that it only creates for valid database configurations.

generators/aws-containers/templates/_base.template.yml
```diff
@@ -0,0 +1,35 @@
...
+        - true
+        - false
+    Default: false
+Resources:
+<% for (app of appConfigs) { %>  <%= app.baseName %>:
+    Type: 'AWS::CloudFormation::Stack'
+    Properties:
+      Parameters:
+        parentStackName: !Ref AWS::StackName
+        repositoryName: <%= app.baseName %>
```
carvallegro: @ggotti You can't have uppercase in the repository name. Needs to comply with the following regex:
```
 '(?:[a-z0-9]+(?:[._-][a-z0-9]+)*/)*[a-z0-9]+(?:[._-][a-z0-9]+)*'
```

ggotti:
Thanks to everyone for the feedback. We've made a bunch of changes based on the feedback we have received:

- 	Corrected issues with the Postgresql deployment option.
- 	Split out performance from scaling.
- 	Renamed templates to have a ejs suffix.
- 	Replaced Network Load Balancer with Application Load Balancer.
- 	Added a warning about distributed caching.
- 	Tightening of input control around stackname+application names.

The documentation [PR](https://github.com/jhipster/jhipster.github.io/pull/532)) has also been updated to reflect these changes.

jdubois:
@ggotti I tested it twice, I consistently got:

```
12:14:54 AM 	CREATE FAILED                       AWS::RDS::DBClusterParameterGroup	jhipster-test-ippon-tmp-ER0MYQ8VWR4X
12:14:55 AM 	CREATE FAILED                       AWS::CloudFormation::Stack	jhipster-test-ippon-tmp-ER0MYQ8VWR4X
✖ There was an error creating the stack: Creation of nested stack failed
```

Could you have a look?

ggotti:
Ok so the issue that @jdubois has raised is an interesting one, and was caused by insufficient timeout given to CloudFormation for the stack. The sub-generator was creating a "AWS::RDS::DBClusterParameterGroup", which according to the [CLI documentation](https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-cluster-parameter-group.html) (but interestingly enough not the CloudFormation documentation) requires at least 5 minutes before creating your new DB Cluster. We had set a 5 minute timeout on the nested-stack creation itself, and so it appears that we're now hitting that timeout limit; the sub-stack timeout is for individual items, not the stack itself. 

So, where to from here? The DBClusterParameterGroup is actually not giving us a lot of value, and speed wise is more of a headache than its worth. So what I've done is removed it from the CloudFormation nest-stack, and used the account default values (as per [this](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html)).  I've also removed the timeout function, which will account for situations where AWS is taking longer to provision than normal.

jdubois:
OK I tested it several times yesterday without any issue, let's merge this!

Issue #6987, opened 2018-01-15
Code Review: User/Account Management Bugs/Testcases/JWT concerns


RobinKamps:
##### **Overview of the issues**

Greetings JHipsters!

First of all thanks for your amazing work!
Just reviewed some of your code.
Here are some obvious bugs and concerns:

1.
Login of users can be changed via adminpanel.
However Login with invalid old Login is still possible, if cacheManager is enabled.
Solution: Both caches should be evicted on updateUser(UserDTO...) **also before** you update the user e.g.:

```
<%_ if (cacheManagerIsAvailable === true) { _%>
                cacheManager.getCache(UserRepository.USERS_BY_LOGIN_CACHE).evict(user.getLogin());
                cacheManager.getCache(UserRepository.USERS_BY_EMAIL_CACHE).evict(user.getEmail());
<%_` } _%>
```

2.
Email of users can be changed via adminpanel or via account by the users.
Login with old Email is still possible.
Solution: The email cache should evict on updateUser(String...) **also before** you update the user e.g.:

```
<%_ if (cacheManagerIsAvailable === true) { _%>
                cacheManager.getCache(UserRepository.USERS_BY_EMAIL_CACHE).evict(user.getEmail());
<%_` } _%>
```

Note: There are no Testcases wich cover the appropiate cache behavior - they should be added.

3.
Login via Email is messed up:
Emails can be stored with uppercases in db (on registration e.g.).
Then Login with email is not possible. Reason: the login is always converted tolowercase.
(DomainUserDetailService / String lowercaseLogin = login.toLowerCase(Locale.ENGLISH);)
Solution: emails should be stored in db tolowercase like the login - otherwise indexing and finding them etc. is just bad.

Testcases here do not cover tolower correct (no email in the testcases contains uppercase letters)

4. 
There is no need to check db everytime for email at login - its only needed if login is in email format.
Solution: check `(new EmailValidator()).isValid(lowercaseLogin,null)` first

Note:
Tests do not cover elasticsearch deletes/updates

...

Important Notes (I guess also pretty obvious for all that have jwt experience):

The overall design of how the jwt-tokens and UserDetailService/principal is currently implemented is not production ready.
[...]

PR #7057, opened 2018-01-29
Cache evict and email login

RobinKamps:
Greetings JHipsters!

Fixes for issue #6987

This PR closes the cache evict bugs, the "email can not signIn with email containing upperCase" and unnecessary email db lookup issue.

This PR does not store email toLowerCase always, as i suggested premature as a fix.

Reason:
Although it is quite common to identify the locale-part of email-addresses containing upper-case letters and the equivalent lowerCase as the same email-address locale-part, RFC5321 and RFC5322 are allowing email providers to distinguish strict between lowerCase and upperCase within locale-parts.
Due to this, converting/storing the email always toLowerCase could prevent users from receiving mails/the activation mail,  in case they have an email containing upperCase chars and their email-provider handles this strict.

Although JHipster stores emails not altered, JHipster does not support strict distinguishing (due to lookup via findOneByEmailIgnoreCase on signUp). 
There are several possible solutions to deal with the standards, but each comes with pros and cons, and the standards are widely considered to be broke at that part.
This PR however uses the approach JHipsters used before (cached findOneWithAuthoritiesByEmail on signIn, but ignoreCase on signUp) and just fixes the "not able to signIn with upperCase emails" for sql dbs resulting in the following behavior:
- users must type in their email-addresses on signIn using the exact matching case as stated on signUp/account/admin panel (for sql dbs).
- users from providers with strict email-address handling could be prevented from signUp, if they are not the first with an unique ignoreCase email.
- Note that this approach is relying at signUp on ignoreCase (and for nosql dbs also on signIn) and therefore relies on the used db-engine and locale/field type/utf-8 mb db settings configuration. This might or might not work as expected on email-addresses containing special chars as idn2008 domain part.

Possible alternative implementations:

1.
Similar to the current but refactor findOneWithAuthoritiesByEmail (cached) to findOneWithAuthoritiesByEmailIgnoreCase (must be non cached than)
- This would trade the cache benefits for a perhaps better user-experience (for ignoreCase on signIn - signUp/account/admin panel still depends on email-provider)
- Note that this approach would also have the cons as mentioned before and relies even more (now for signIn/sql dbs) on the current db-engine/settings

2.
If you want to implement the somewhat messed up standards and enable strict distinguishing within JHipster, just refactor findOneByEmailIgnoreCase to findOneByEmail.
Trade-offs will be:
- users with non strict providers can signUp multiple accounts with one/the same email-address just by stating a different upper-lower-case combination each time they signUp

Note:
Current email max length is set to 100 - however an email-address is valid/will be transported even when the email-address length is up to 254 bytes.

### DIFF
--- a/generators/server/templates/src/main/java/package/repository/_UserRepository.java
+++ b/generators/server/templates/src/main/java/package/repository/_UserRepository.java
@@ -94,16 +94,16 @@ public interface UserRepository extends <% if (databaseType === 'sql') { %>JpaRe
 
     String USERS_BY_EMAIL_CACHE = "usersByEmail";
     <%_ } _%>
-<%_ if (authenticationType !== 'oauth2') { _%>
+    <%_ if (authenticationType !== 'oauth2') { _%>
 
     Optional<User> findOneByActivationKey(String activationKey);
-<%_ } _%>
+    <%_ } _%>
 
     List<User> findAllByActivatedIsFalseAndCreatedDateBefore(Instant dateTime);
-<%_ if (authenticationType !== 'oauth2') { _%>
+    <%_ if (authenticationType !== 'oauth2') { _%>
 
     Optional<User> findOneByResetKey(String resetKey);
-<%_ } _%>
+    <%_ } _%>
 
     <%_ if (databaseType === 'couchbase' || databaseType === 'mongodb') { _%>
         <%_ if (cacheManagerIsAvailable === true) { _%>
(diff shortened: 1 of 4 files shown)


jdubois:
LGTM

3 comments before we merge:

- we should also force the email to be in lower-case in the Angular/React UI, so that would greatly reduce the issues
- we should increase the e-mail size to 254, as you explain
- for JHipster 5, we need to discuss whether we still keep the username, and not just the e-mail address

deepu105:
for JH5 let's remove username and make it simple. We even provide a tip on
how to change to use username instead of email for those who need that


Thanks & Regards,
Deepu

jdubois:
OK I'm merging this and then implementing the comments

jdubois:
So removing the email is really hard, as there are stuff like social login and OpenID which uses it, and I have no idea if it's a good idea to remove that for them... I'll do the easy comments first

PR #7069, opened 2018-02-01
Elasticsearch save bug

kazeki:
Repository.save will return the newest entity when creating an entity, but when updating, it will not.  So that when I use elasticsearch and entityAudit module, the test of entity update failed, because elasticsearch saved the old entity returned by Repository.save

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/save_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/save_template.ejs
@@ -27,10 +27,10 @@ if (!viaService) {
     if (dto === 'mapstruct') {
         resultEntity = entityInstance; %>
         <%= entityClass %> <%= entityInstance %> = <%= dtoToEntity %>(<%= instanceName %>);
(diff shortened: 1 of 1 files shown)


jdubois:
I don't get why you would need to flush the Hibernate session for that -> could you add a test case? That would be hugely useful, if our current code is broken

jdubois:
Oh sorry -> maybe because that's because you're not in our normal (simplistic) use case. As the Hibernate session is not flushed, if you have another request coming, you have some kind of dirty read as Elasticsearch does not participate in the transaction. Am I correct?
Then I'm not sure we need to flush the session for everybody, as there is a performance hit, when this is a "normal" behavior when you don't have distributed transactions. In that specific case, yes you need flush, but not in the other cases.

kazeki:
it isn't about transactions. when i use h2db + elasticsearch + entityAudit module，the rest API test of updating entity which just generated by jhipster will fail

PR #7072, opened 2018-02-01
Correct type handling of HTTP error messages

stieler-it:


generators/entity-client/templates/angular/src/main/webapp/app/entities/_entity-management.component.ts
```diff
@@ -88,26 +88,26 @@ export class <%= entityAngularName %>Component implements OnInit, OnDestroy {
...
+        this.totalItems = parseInt(headers.get('X-Total-Count'), 10);
         for (let i = 0; i < data.length; i++) {
             this.<%= entityInstancePlural %>.push(data[i]);
         }
     }
 
     <%_ }} _%>
-    private onError(error) {
-        this.jhiAlertService.error(error.message, null, null);
+    private onError(errorMessage: string) {
```
deepu105: So are we getting message directly without an error object?

stieler-it: Please look it up in `pagination-template.ejs` or `no-pagination-template.ejs` or `infinite-scroll-template.ejs`. In all files, `onError` is called with `this.onError(res.message)`, so the error object is already navigated to the message.

We can also change the other templates instead if you prefer to have the error object in the handler but my solution would conform to the success handler which is called like this `this.onSuccess(res.body, res.headers)`.

I have not checked if the current state actually works / shows errors. Maybe it is the global error handler jumping in and this one isn't needed at all?

deepu105: Its ok, most probably I wrote those but it has been a while. Just asking to confirm.

deepu105:
@stieler-it build seems to be failing. I restarted it once. If it still fails can you please check?

stieler-it:
Thanks, then there is probably something wrong. I will have a look as soon I'm done with our upgrade.

deepu105:
@stieler-it can you look at the conflicts plz

deepu105:
@stieler-it the builds are still failing, plz take a look

stieler-it:
Ok, I had to fix some types to the new interfaces. Let's see if it works now.

I noticed, however, that the generated component has compile errors. It appears that the `loadAll()` function builds the query using the `sort()` function, the `sort()` function is explicitely NOT generated if database type is Cassandra. This seems to be unrelated to my changes and I don't understand the reasons so I couldn't fix it on the way. Can you have a look or should I raise an issue for that?

stieler-it:
Apparently the linting rules are responsible for the failing build. Although my personal opinion is that generated code should be allowed to contain some unused imports, since you decided otherwise I will of course provide a compliant version.

deepu105:
It can have unused import, if it makes templating much easier but only in cases where templating it is difficult. In those spacial cases just add the tslint ignore comment. But if its easy to exclude using a simple ejs if condition then plz do that

jdubois:
For the Cassandra part: there is no "sort" function here as it's not possible with this database (this is the simple explanation of course, but let's keep it short as it's not the real problem here)

PR #7079, opened 2018-02-04
Entity folder structure

### LINKED ISSUE #6693: Brainstorming on a new Angular directory structure
#### **Overview of the issue**

Today's Angular directory structure is quite flat and it starts to bring a few challenges. We could use this issue to note the improvements that could be done.

```
├── entities
│   ├── country
│   │   ├── country-delete-dialog.component.html
│   │   ├── country-delete-dialog.component.ts
│   │   ├── country-detail.component.html
│   │   ├── country-detail.component.ts
│   │   ├── country-dialog.component.html
│   │   ├── country-dialog.component.ts
│   │   ├── country-popup.service.ts
│   │   ├── country.component.html
│   │   ├── country.component.ts
│   │   ├── country.model.ts
│   │   ├── country.module.ts
│   │   ├── country.route.ts
│   │   ├── country.service.ts
│   │   └── index.ts
```

#### **Motivation for or Use Case**

There are several problems we encounter with this flat structure.

##### **Same entity name in several MicroServices collide in the Gateway**

There is already [an issue about this](https://github.com/jhipster/generator-jhipster/issues/6480). Basically, we use several microservices and aggregate them into a single gateway. Some Entities are redundant (eg. Organisation, Contact, ...) and exist in several microservices (event though they don't have the same attributes). With this flat directory structure, you end up with twice the same entity. To avoid this, we need to add the _AngularJSSuffix_, but it's not really nice.

##### **Module inter-connection**

Today, [the client model in Angular is not typed](https://stackoverflow.com/questions/45843065/strong-typing-the-client-model-in-angular). We use the `BaseEntity` for relations (here, an `Organisation` has one `Country`):

```
export class Organisation implements BaseEntity {
    constructor(
        public id?: number,
        public name?: string,
        public country?: BaseEntity,
    ) {
    }
}
```

When we type this (and I hope JHipster will one day):

```
export class Organisation implements BaseEntity {
    constructor(
        public id?: number,
        public name?: string,
        public country?: Country,
    ) {
    }
}
```

We end-up with modules depending on each-other. Same for the attempt being done now about [auto-complete fields for the entities](https://github.com/jhipster/generator-jhipster/pull/6618). Basically, these components can be shared by any other, so modules depend on each other.

#### **Related issues**

* [Same entity name in several MicroServices collide in the Gateway](https://github.com/jhipster/generator-jhipster/issues/6480)
* [Auto-complete fields for the entities](https://github.com/jhipster/generator-jhipster/pull/6618)
* [Strong typing the client model in Angular](https://stackoverflow.com/questions/45843065/strong-typing-the-client-model-in-angular)

#### **Suggest a Fix**

If you take all these constraints into consideration, it looks like the directory structure could be improved:

* have a sub-directory per microservice
* have a shared directory per microservice (with model, service and other shared components such as the selection)

##### **Sub-directory per microservice and shared directory**

This could look like this:

```
├── billing     // microserivce
├── invoicing   // microserivce
├── scanning    // microserivce
│   ├── entities
│   │   ├── country  // notice that model and service are not there
│   │   │   ├── country-delete-dialog.component.html
│   │   │   ├── country-delete-dialog.component.ts
│   │   │   ├── country-detail.component.html
│   │   │   ├── country-detail.component.ts
│   │   │   ├── country-dialog.component.html
│   │   │   ├── country-dialog.component.ts
│   │   │   ├── country-popup.service.ts
│   │   │   ├── country.component.html
│   │   │   ├── country.component.ts
│   │   │   ├── country.module.ts
│   │   │   ├── country.route.ts
│   │   │   └── index.ts
│   │   ├── shared
│   │   │   ├── model   // every model 
│   │   │   │   ├── country.model.ts
│   │   │   │   └── organisation.model.ts
│   │   │   ├── service   // every service
│   │   │   │   ├── country.service.ts
│   │   │   │   └── organisation.service.ts
│   │   │   ├── selection  // every selection component
│   │   │   │   ├── country-selection
│   │   │   │   │   ├── country-selection.component.html
│   │   │   │   │   ├── country-selection.component.ts
│   │   │   │   │   ├── country-selection-multi.component.html
│   │   │   │   │   └── country-selection-multi.component.ts
│   │   │   │   └── organisation-selection
```

Some components can get quite complex, use helper classes or css. I quite like the "sub-directory per component", but that's another step further, not sure it's really needed

```
├── scanning
│   ├── entities
│   │   ├── country
│   │   │   ├── country-delete-dialog
│   │   │   │   ├── country-delete-dialog.component.html
│   │   │   │   ├── country-delete-dialog.component.ts
│   │   │   ├── country-detail
│   │   │   │   ├── country-detail.component.html
│   │   │   │   ├── country-detail.component.ts
│   │   │   ├── country-popup.service.ts
│   │   │   ├── country.module.ts
│   │   │   ├── country.route.ts
│   │   │   └── index.ts
```

##### **i18n**

And of course, i18n also comes to play. So instead of having a flat directory structure:

```
├── i18n
│   ├── en
│   │   ├── address.json
│   │   ├── contact.json
│   ├── fr
```
We would also seperate it per micr-service

```
├── i18n
│   ├── billing     // microserivce
│   ├── invoicing   // microserivce
│   ├── scanning    // microserivce
│   │   ├── en
│   │   │   ├── address.json
│   │   │   ├── contact.json
│   │   ├── fr
```

##### **Menu**

Today the UI shows an "Entity" menu, with all the entities. Again, it should be split into sub-menues (one per microservice) and then, have all the entities of this microservice list.

#### **JHipster Version(s)**

4.10.2

deepu105:
Fix #6693

This PR moves the entity model to the shared folder and also groups entities based on microservices. This also makes the grouping folder configurable via a flag `client-root-folder`

Microservices gateway
```
|_ entities
  |_ microserviceName
    |_ entityName
      |_ *.{component,module,service,route}.{ts,html}
|_ shared
  |_ *.model.ts
```
For monoliths
```
|_ entities
    |_ entityName
      |_ *.{component,module,service,route}.{ts,html}
|_ shared
    |_ *.model.ts
```

This also revamps the types to use interfaces rather than classes and removed the baseEntity

@agoncal @gmarziou I have kept it simple as I don't see any other actual issues that could happen from this structure and all of the use cases you mentioned should work

I don't see a need to move services out as I don't see an issue here(moving them out makes it more complicated due to the fact they need to be registered as a producer), circular dependency issue is only when you try to use barrel rolls, so when depending on services among entities use direct component paths instead of barrels. I don't see a need to import a service in another service as aggregation normally happens in the components. The same goes for proposed auto-complete fields, they are used in parent components and hence doesn't pose an issue.

For the arguments of splitting into multiple SPA/Modules, I don't see an issue there as well as you can add more modules and set up the base as you like easily. You could have separate mount points loading separate applications and control them using standard servlets.

Of course, we cant satisfy every possible use cases.

Let me know if you think I missed something.

**TODO:**

- [x] This doesn't group i18n files yet. I need to see how to do that later.
- [x] move model to shared for React entities
- [x] add a flag to disable grouping and have the original behavior
- [x] check if we can use path alias `resolve: { alias: { '@src': PATH_TO_SRC } }` in webpack and tsconfig to reduce some import clutter
- [x] fix double import issue highlighted by Antonio, add test if possible

**Known issues*

- If there are 2 microservices creating entity with the same name, you will have only the last one you generated in `.jhipster` folder. It should be fine if you are always generating it using JDL, but if you want to regenerate directly using `--with-entities` flag or something it will fail. There is no solution that I could think of as we have no way of knowing the microservice name before loading the json, so we cant group the json using service name. Ping @agoncal since you are the only one who I know having this usecase what do you think?

<hr>


deepu105:
@jhipster/developers @agoncal does anyone has any comments here? else I would like to merge them

jdubois:
Just reading from your description, this looks good to me. For gateways, it's good to group the entities by microservices, as it's exactly the same thing as what we have in the backend: this will remove a lot of issues, and will also be more logical.

gzsombor:
Seems good. However, it still not clear for me, where to put 'entity-components', like a 'jhi-project-link' or 'jhi-project-autocomplete', etc, which needs to be visible in everywhere - possible in every page. 
 Maybe, I'm missing something obvious. Or you don't want to solve that issue with this :)

deepu105:
@gzsombor I would put them in the same entity folder as they are part of the entity. There should not be a problem in visibility. If you are importing them into a non-entity page you could import them from the entity index and if you are importing it in another entity just import them using full path. Do you see any other issue?

deepu105:
@gzsombor sorry I made a mistake in my previous comment so edited it

agoncal:
@deepu105 When you have two different microservices, both with the same entity name (let's say `Customer`), there is a problem in the `entities/entity.module.ts` file: `GatewayCustomerModule` is declared twice : 

```
import { GatewayIsbnModule } from './microone/isbn/isbn.module';
import { GatewayCustomerModule } from './microone/customer/customer.module';
import { GatewayCustomerModule } from './microtwo/customer/customer.module';
/* jhipster-needle-add-entity-module-import - JHipster will add entity modules imports here */

@NgModule({
    imports: [
        GatewayIsbnModule,
        GatewayCustomerModule,
        /* jhipster-needle-add-entity-module - JHipster will add entity modules here */
    ],
    declarations: [],
    entryComponents: [],
    providers: [],
})
export class GatewayEntityModule {}
```

deepu105:
@agoncal thanks for testing ill fix that

agoncal:
@deepu105 We've already talked about this, but I'll try again ;o)  The advantage of having `microserviceName/entities` is that we can really isolated the code for each microservices, and even have seperate teams. The entire code of the microservice, goes into the microservice directory:

```
|_ microserviceName
  |_ entities
    |_ entityName
      |_ *.{component,module,service,route}.{ts,html}
  |_ mycomponent1
  |_ mycomponent2
  |_ mycomponent3
|_ shared
  |_ *.model.ts
```

That would also allow to have routes per microservice. Coming back to the redundant customer, we could have:

```
http://localhost:9000/microone/customer.html
http://localhost:9000/microtwo/customer.html
http://localhost:9000/auth
http://localhost:9000/admin
```

deepu105:
@agoncal let me give it a thought. I'll try to do it if its not complex to change

deepu105:
@agoncal I tried to implement your suggestion(spend half a day on it) but it makes the templates way too complex especially the below, and will require a lot of changes(way more than I imagined)
1. import paths will be much more complex and template becomes more complicated given different root folders that will exist
2. The common entity module definition we had earlier wont work for this structure and we need to find a way to group entities by module for microservices and simpler default for monoliths. This will require a lot of conditions on our already too complex templates
3. The test files currently have simpler import paths, with this it will become more complex

I understand your requirement and agree that it will be much more optimal organisation of code but considering the additional complexity, I'm reluctant to do it at this point. I'm sorry.

agoncal:
@deepu105 you are the master. If you are reluctant, leave it then. But does it mean you will at least implement your proposition:

```
|_ entities
  |_ microserviceName
    |_ entityName
```

deepu105:
@agoncal yes that is already done in this PR (which already made it more complex, still figuring some hidden bugs and trying to fix)

agoncal:
@deepu105 let me know we I can give it a try so I can give you some feedback

deepu105:
@agoncal you can try now

deepu105:
@agoncal look at my comment on the PR description under known issues. Let me know what you think

deepu105:
@jdubois @jhipster/developers does anyone know why the gateway builds are failing. The error looks strange

pascalgrimaud:
@deepu105 : all last PR failed for microservices, but it is green in Master. I don't know what happened, I need to check with a computer

deepu105:
ya just noticed, the same builds for gateway are failing in all PR

deepu105:
then i'm merging this since the issue needs to fixed in master

pascalgrimaud:
Yes no worry about the builds. It is not very stable because of Spring Boot 2.

agoncal:
@deepu105 a few comments.

It would be good to have submenues ("MicroOne", "MicroTwo") when doing microservices. Because we will get more entities (aggregating them from several microservices) and because some might have the same name.

_jhipstermenu2
BTW ATM because it's the same menu, when you select one, it selects the other.

_jhipstermenu1
Both Customer menu use the same route (http://localhost:9000/#/customer) so you can't go from one customer to another one (it always goes to the same).

BTW what about using the same sub-directory system in i18n ? At the moment we have this:

```
├── entities
│   ├── microone
│   │   └── customer
│   ├── microtwo
│   │   └── customer

i18n
├── en
│   ├── microoneCustomer.json
│   ├── microtwoCustomer.json
```

Wouldn't it be better to use the same directory structure?

```
i18n
├── en
│   ├── microone
│   │       └── customer.json
│   ├── microtwo
│   │       └── customer.json
```

Oh, and coming back to i18n, there is an issue with case because there is:

```
translation-not-found[microoneApp.microonecustomer.created]
```

And it should be `microoneCustomer`with uppercase `C`

deepu105:
@agoncal this issue took way more time than I initially imagined and it was way too complex to do and test. Though from the diff it might not look so, I went through so many iterations to reach this. Making this work with different directory layout based on microservice name is easy but templating and generating it is not as easy as it sounds. So frankly I'm not willing to do any further changes to this other than fixing bugs than what is done here as its simply too many combinations to test for and to make it work. Originally I even tried to make i18n grouped but gave up after failing so many times to generate it properly.
Now for your other use case of 2 entities with the same name in different services, to be honest, I even thought of not doing anything to support it at all as its kind of an edge case IMO, as the cost of complexity to do it by generation is far too high IMO. For the end user its just as simple as choosing a different name, or just prefixing the entity name with the service name while creating it, but still I implemented support for that as much as possible. I can try to fix the i18n issue but unfortunately, I think practically this is the closest I can offer for your use case

agoncal:
@deepu105 yes, I can imagine that generating such code is not trivial :o(  Well done for the new directory structure.

BTW will you be doing something similar with React ?

deepu105:
@agoncal for react yes, at least something similar

agoncal:
BTW @gzsombor what about adding your [auto-complete fields](https://github.com/jhipster/generator-jhipster/pull/6618) to see how it fits with the new structure

PR #7080, opened 2018-02-04
Remove unused HazelcastInstance

michael-wirth:


michael-wirth:
Do we really need a HazelcasrInstance reference in the WebConfigurer? It is not used at all.
Filter- and SevletRegistration can be replaced by Mockito.

jdubois:
This was to force Hazelcast to be initialized at this moment - it's a ugly
hack, I know.
I **think** this was only useful when doing http session clustering
(logically Hazelcast needed to be configured before the servlet filter),
and that's an option we removed a few months ago -> this needs to be
verified but indeed this can probably be removed

michael-wirth:
Javadoc changes reverted

PR #7097, opened 2018-02-10
support unused needle

axelhodler:
The needle was added to `_build.gradle` in this commit https://github.com/jhipster/generator-jhipster/commit/a1a89b3f3ba9cd201d7b82da5835b478140721ba but was apparently never implemented.

I would love to use it for a jhipster module im creating.

Maybe the name is not great `addGradlePluginToPluginsBlock` since `addGradlePlugin` is already taken by the needle adding to the buildscripts block

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1440,6 +1440,28 @@ module.exports = class extends PrivateBase {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jdubois:
LGTM!

jdubois:
Thanks @axelhodler !

PR #7110, opened 2018-02-11
Use a CoreModule to prevent from duplicate Service instances

wmarques:
I started working on Lazy Loading but saw that our SharedModule contains providers, and it shouldn't.
Indeed, each time the SharedModule is imported in a module, the providers are recreated, this might cause some issues as we think our services as singletons.

So I'm proposing to add a CoreModule which will be only imported in the AppModule and provide all the shared services. This is the recommended way from [Angular](https://angular.io/guide/singleton-services).


dancancro:
Here's an app I made with lazy loading that does this too https://github.com/dancancro/great-big-example-application/blob/master/src/main/webapp/app/core/core.module.ts

generators/client/files-angular.js
```diff
@@ -260,7 +260,7 @@ const files = {
                 { file: 'account/social/_social-register.component.html', method: 'processHtml' },
                 { file: 'shared/social/_social.component.ts', method: 'processJs' },
                 { file: 'shared/social/_social.component.html', method: 'processHtml' },
-                'shared/social/_social.service.ts'
+                'core/social/_social.service.ts'
```
deepu105: if a different folder is preferred then can you move the files to the `angularCore` section added

generators/client/templates/angular/src/main/webapp/app/account/activate/_activate.component.ts
```diff
@@ -21,7 +21,7 @@ import { NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 import { ActivatedRoute } from '@angular/router';
 
 import { ActivateService } from './activate.service';
-import { LoginModalService } from '../../shared';
+import { LoginModalService } from '../../core';
```
deepu105: please note that I merged folder restructure PR to master. Now you can do imports for this like `app/core` instead of providing relative paths. see if you can use that everywhere

generators/client/templates/angular/src/main/webapp/app/core/_core.module.ts
```diff
@@ -0,0 +1,83 @@
...
+    AuthServerProvider,
+    <%_ if (!skipUserManagement || authenticationType === 'oauth2') { _%>
+    UserService,
+    <%_ } _%>
+    UserRouteAccessService
+} from './';
+<%_ if (enableI18nRTL) { _%>
+import {
+    FindLanguageFromKeyPipe
+} from '../shared';<%_ } _%>
```
deepu105: if something from shared is needed here shouldnt it be moved here as well?

generators/client/templates/angular/src/main/webapp/app/core/language/_language.helper.ts
```diff
@@ -21,9 +21,9 @@ import { Title } from '@angular/platform-browser';
 import { Router, ActivatedRouteSnapshot } from '@angular/router';
 import { TranslateService, LangChangeEvent } from '@ngx-translate/core';
 
-import { LANGUAGES } from './language.constants';
+import { LANGUAGES } from '../../shared/language/language.constants';
```
deepu105: move the related files here as well, it doesnt make sense to have them in two places

deepu105:
@wmarques the conflicts are because of the import pattern change. I added alias for app folder so that we can do absolute path imports instead of relative ones

wmarques:
Thanks for the feedback, I'll update the PR tonight :)

generators/client/templates/angular/src/main/webapp/app/core/_core.module.ts
wmarques: I let this pipe in the shared because it's a pipe and can be used as a classic pipe in the HTML (so we need it in the declarations/exports arrays) or as a service (in the providers array), so it has to stay in shared...
The CoreModule is imported once in the AppModule and should only contains providers.

generators/client/files-angular.js
wmarques: I would only move the file to the `angularCore` from the `angularShared` section, the other sections group files by features (social, i18n...). Does it makes sense to you?

generators/client/templates/angular/src/main/webapp/app/core/language/_language.helper.ts
wmarques: I didn't move the other files because in some cases, the files were component files, so if we want to re-use some component in other modules of the app, they should stay in shared. That's why I only moved the services into the `CoreModule`. But I agree for this constant, I'm moving it

generators/client/files-angular.js
deepu105: It might be confusing to have core module files inside shared section. Here we try to group by folder mostly I guess

generators/client/templates/angular/src/main/webapp/app/core/_core.module.ts
deepu105: You are right. Ok

deepu105: app/shared plz

generators/client/templates/angular/src/main/webapp/app/core/language/_language.helper.ts
```diff
@@ -21,9 +21,9 @@ import { Title } from '@angular/platform-browser';
 import { Router, ActivatedRouteSnapshot } from '@angular/router';
 import { TranslateService, LangChangeEvent } from '@ngx-translate/core';
 
-import { LANGUAGES } from './language.constants';
+import { LANGUAGES } from 'app/core/language/language.constants';
 <%_ if (enableI18nRTL) { _%>
-import { FindLanguageFromKeyPipe } from './find-language-from-key.pipe';
+import { FindLanguageFromKeyPipe } from '../../shared/language/find-language-from-key.pipe';
```
deepu105: app/shared plz

generators/client/templates/angular/src/main/webapp/app/core/login/_login-modal.service.ts
```diff
@@ -19,7 +19,7 @@
 import { Injectable } from '@angular/core';
 import { NgbModal, NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
 
-import { <%=jhiPrefixCapitalized%>LoginModalComponent } from './login.component';
+import { <%=jhiPrefixCapitalized%>LoginModalComponent } from '../../shared/login/login.component';
```
deepu105: app/shared plz

deepu105:
Isnt any generated test files affected? If so thry should also move to same folder structure

wmarques:
Oh sorry didn't see those one.. I'll check tonight along with the test structure (my client doesn't allow github login...)
Thanks for the feedback

deepu105:
@wmarques plz merge once tests are green

PR #7127, opened 2018-02-15
For logout resource, return 204 OK response instead of 200 OK

vishal423:
With 200 OK response without body, Firefox gives warning on console as it expects
root xml element (treating response as text/xml in absence of content-type header)


gmarziou:
Shouldn't the logout method annotation include `produces = MediaType.APPLICATION_JSON_VALUE` ?

ramzimaalej:
@gmarziou no, since response does not have a body. As per the RFC, you should only return content-type header if the response has a body.

vishal423:
Irrespective of content-type, it doesn't make any sense to send 200 response as 204 is meant to handle this scenario.

jdubois:
Looks good to me! @vishal423 did you test this with the Angular front-end, just to be sure it doesn't break this?

vishal423:
@jdubois , yes I verified this change before raising merge request :)

jdubois:
Perfect! Thanks a lot!

PR #7131, opened 2018-02-16
Added support for Elasticsearch to the Heroku generator

jkutner:
This will automatically create a *free* Elasticsearch add-on attached to the Heroku app when the generator detects `"searchEngine": "elasticsearch"` in the config. For this to really work, I think the elasticsearch generator needs to add these to the `application.yml`

```
spring:
    data:
        elasticsearch:
            cluster-name:
            cluster-nodes:
            properties:
                path:
                    logs: target/elasticsearch/log
                    data: target/elasticsearch/data
```


jdubois:
This is so awesome!!

PR #7140, opened 2018-02-19
Fixing blueprint

sendilkumarn:
@deepu105 This fixes the error w.r.t blueprint when ran locally. 

@mraible Can you try this? this should fix the blueprint error that you posted in Google group. (long back ;) )

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -594,7 +594,7 @@ module.exports = class extends Generator {
         }
         const done = this.async();
         const localModule = path.join(process.cwd(), 'node_modules', blueprint);
-        if (!fs.existsSync(localModule)) {
+        if (fs.existsSync(localModule)) {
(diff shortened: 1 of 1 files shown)


generators/generator-base-private.js
```diff
@@ -594,7 +594,7 @@ module.exports = class extends Generator {
         }
         const done = this.async();
         const localModule = path.join(process.cwd(), 'node_modules', blueprint);
-        if (!fs.existsSync(localModule)) {
+        if (fs.existsSync(localModule)) {
```
deepu105: not sure about this. I need to test it. The idea was that if there is no module in your local node_modules folder then we check if there is a global one available. I guess the fix should be to call `done` in the else for this if

sendilkumarn: Yup agree that completely makes sense.

deepu105:
merged manually

PR #7155, opened 2018-02-22
sonar code smell fixes

ruddell:
This should reduce the "code smell" count by at around 60.  I also updated several ejs conditions to use `<%_` instead of `<%`


jdubois:
Thanks a lot!!!
I'm merging as the build fails because of me

generators/entity-server/templates/src/main/java/package/service/impl/_EntityServiceImpl.java
```diff
@@ -36,8 +36,10 @@
 import <%=packageName%>.service.mapper.<%= entityClass %>Mapper;<% } %>
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+<%_ if (pagination !== 'no') { _%>
```
jdubois: I need to revert this, as this makes the tests fail, when there is a relationship (hava a look at the `findAllWithEagerRelationships` method)

jdubois: oh I'll try with a  `|| fieldsContainOwnerManyToMany === true`

PR #6938, opened 2018-01-03
adding cipher suite SSL for prod 

raiden0610:
strengthen the SSL security with the correct cipher suite
Grade B in SSLlabs (https://www.ssllabs.com/ssltest/) goes to A- with this one

- Please make sure the below checklist is followed for Pull Requests.

- [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x] Tests are added where necessary
- [x] Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #7160, opened 2018-02-23
Adding undertow ordered cipher configuration

raiden0610:
It will force the order of the cipher define in the properties file.
The purpose of this is to achieve perfect forward secrecy for SSL [(More infos)](https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices#25-use-forward-secrecy) in combination of the cipher list in the comments SSL section in the spring boot properties file.

It's the missing part of my first [PR](https://github.com/jhipster/generator-jhipster/pull/6938)

The only thing left will be to add this JVM flag
`-Djdk.tls.ephemeralDHKeySize=2048`

 But I don't really know how we can automate this since it depends on the cloud / hosting providers

With all of these the grade in [SSLLabs](https://www.ssllabs.com/ssltest/) will be A+ instead of B

Tell me what you think about that 😃 

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -405,6 +405,7 @@ function writeFiles() {
             this.template(`${SERVER_MAIN_SRC_DIR}package/config/MetricsConfiguration.java.ejs`, `${javaDir}config/MetricsConfiguration.java`);
             this.template(`${SERVER_MAIN_SRC_DIR}package/config/ThymeleafConfiguration.java.ejs`, `${javaDir}config/ThymeleafConfiguration.java`);
             this.template(`${SERVER_MAIN_SRC_DIR}package/config/WebConfigurer.java.ejs`, `${javaDir}config/WebConfigurer.java`);
+            this.template(`${SERVER_MAIN_SRC_DIR}package/config/_UndertowConfiguration.java`, `${javaDir}config/UndertowConfiguration.java`);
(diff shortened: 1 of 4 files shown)


jdubois:
Very interesting, thanks a lot!
I'm wondering whether this shouldn't be in https://github.com/jhipster/jhipster instead - do you think people would modify this class? If not, it's easier for us, and also cleaner for end-users, to put that class in https://github.com/jhipster/jhipster instead of generating it. What do you think?

raiden0610:
Yes I understand.
Well I think it depends, if at a point people need to custom undertow it can be done in the same class no need to create anything.
But I don't have any use case in mind for that.
I think you are right, I'll try to integrate the change in the server side library

jdubois:
OK @raiden0610 I also think it's better, and also less work/maintenance for us.

We might also need:

- some specific properties in JHipsterProperties to configure it (for example to enable/disable it?)
- an @ConditionalOnClass to only run this when Undertow is in the classpath (in case some people run JHipster as a WAR in Tomcat - quite unlikely, I know!)

-> can you have a look, and do a proposal/PR?

raiden0610:
Ok for the requirements and ok, I'll do the proposal on jhipster server side 👍

jdubois:
Closing in favor of the new code in jhipster/jhipster

Issue #7163, opened 2018-02-24
Issues with ejs file extension


pascalgrimaud:
##### **Overview of the issue**

When generating a new project, choosing `no translation`, there are some issues in console.

##### **Motivation for or Use Case**

Full log: https://travis-ci.org/jhipster/generator-jhipster/jobs/345224839
An extract of Travis log:

```
Error in file: ../../languages/templates/src/main/webapp/i18n/en/_global.json
{ Error: ENOENT: no such file or directory, open '/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/languages/templates/src/main/webapp/i18n/en/_global.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
    at Object.readFileAsString (/home/travis/.config/yarn/global/node_modules/generator-jhipster/node_modules/html-wiring/index.js:281:13)
    at geti18nJson (/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/utils.js:261:25)
    at replacePlaceholders (/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/utils.js:235:26)
    at renderContent (/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/utils.js:166:24)
    at ejs.renderFile (/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/utils.js:191:13)
    at tryHandleCache (/home/travis/.config/yarn/global/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:228:10)
    at Object.exports.renderFile (/home/travis/.config/yarn/global/node_modules/generator-jhipster/node_modules/ejs/lib/ejs.js:437:10)
    at renderContent (/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/utils.js:189:9)
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/home/travis/.config/yarn/global/node_modules/generator-jhipster/generators/languages/templates/src/main/webapp/i18n/en/_global.json' }
```

##### **Reproduce the error**

Simply generate a new project, then choose N at this question:

```
Would you like to enable internationalization support? (Y/n) n
```

##### **Suggest a Fix**

[...]

PR #7169, opened 2018-02-25
Fix ejs noi18n and nodatabase

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/7163

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -250,9 +250,14 @@ function replacePlaceholders(body, generator) {
 function geti18nJson(key, generator) {
     const i18nDirectory = `${LANGUAGES_MAIN_SRC_DIR}i18n/en/`;
     const name = _.kebabCase(key.split('.')[0]);
-    let filename = `${i18nDirectory + name}.json`;
+    let filename;
     let render;
-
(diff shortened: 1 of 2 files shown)


jdubois:
Thanks @pascalgrimaud ! I'm having the "no database" issue right now!

jdubois:
And that did solve my problem!

generators/utils.js
```diff
@@ -250,9 +250,14 @@ function replacePlaceholders(body, generator) {
 function geti18nJson(key, generator) {
     const i18nDirectory = `${LANGUAGES_MAIN_SRC_DIR}i18n/en/`;
     const name = _.kebabCase(key.split('.')[0]);
-    let filename = `${i18nDirectory + name}.json`;
+    let filename;
     let render;
-
+    if (['activate', 'global', 'health', 'reset'].indexOf(name) !== -1) {
```
deepu105: You could use includes rather than index if as its more readable.  Also instead of checking for special names may be we should try with ejs ext when normal trial fails

pascalgrimaud: Ok, will do the change

PR #7174, opened 2018-02-26
[Don't Merge] fixing spring-service for blueprint

sendilkumarn:

@deepu105

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1737,7 +1737,7 @@ module.exports = class extends PrivateBase {
     composeExternalModule(npmPackageName, subGen, options) {
         let generatorTocall = path.join(process.cwd(), 'node_modules', npmPackageName, 'generators', subGen);
         try {
-            if (!fs.existsSync(generatorTocall)) {    
+            if (!fs.existsSync(generatorTocall)) {
                 this.debug('using global module as local version could not be found in node_modules');
                 generatorTocall = path.join(npmPackageName, 'generators', subGen);
             }
@@ -2395,7 +2395,14 @@ module.exports = class extends PrivateBase {
 
 
     setupSpringServiceOptions(generator, context = generator) {
-        generator.serviceName = context.options.name;
+        if (!context.options.name) {
+            if (context.options._) {
+                generator.name = context.options.name;
+            } else {
+                this.argument('name', { type: String, required: true });
+                generator.name = context.options.name;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


generators/spring-service/index.js
```diff
@@ -23,14 +23,21 @@ const constants = require('../generator-constants');
...
 
+let useBlueprint = false;
+
 module.exports = class extends BaseGenerator {
     constructor(args, opts) {
         super(args, opts);
-        this.argument('name', { type: String, required: true });
-        this.name = this.options.name;
+ 
+        this.name = this.options.name || this.options._;
```
sendilkumarn: This works for now, but I am not sure whether it is the correct way

generators/spring-service/index.js
```diff
@@ -23,14 +23,21 @@ const constants = require('../generator-constants');
 
 const SERVER_MAIN_SRC_DIR = constants.SERVER_MAIN_SRC_DIR;
 
+let useBlueprint = false;
+
 module.exports = class extends BaseGenerator {
     constructor(args, opts) {
         super(args, opts);
-        this.argument('name', { type: String, required: true });
```
deepu105: why is the argument removed? without it the help will not show it I guess

sendilkumarn: `this.argument` is not having name in it.

Since, there are two arguments here.

deepu105: I dont get it, service name is the only argument rite? can you explain the problem? this is working code right

sendilkumarn: w.r.t blueprint this is coming empty

deepu105: it should work, if not there might be something wrong. Needs to be checked. We cant remove the argument declaration as it would break validation and discovery in help provided by yeoman

sendilkumarn: Yeah, that is the main reason why I had put `don't merge` for the PR.

Let me describe the problem now,

* When you run `jhipster --blueprint generator-jhipster-kotlin spring-service foo`
* The arguments inside the blueprint's generator is empty and the blueprint generator is calling the main generator, hence there also the arguments are empty instead the argument is present in the `_` property of the object.

Is this the expected behavior or do we have to do something different to make sure that the arguments are sent across for `yeoman` generator?

generators/generator-base.js
```diff
@@ -2393,6 +2393,18 @@ module.exports = class extends PrivateBase {
         generator.experimental = context.configOptions.experimental || context.options.experimental;
     }
 
+
+    setupSpringServiceOptions(generator, context = generator) {
+        if (!context.options.name) {
+            if (context.options._) {
+                generator.name = context.options._;
+            } else {
+                this.argument('name', { type: String, required: true });
```
deepu105: The problem is without this line yeoman will not register the arg at all. may be there is a problem when composing the blueprint. Check how its done for entity gen, it also has name argument

deepu105:
closing this, I found the real issue, will push the fix after some cleanup later

PR #7177, opened 2018-02-26
Add needle to add maven repositories

axelhodler:

We're using jhipster for a bunch of microservices, project specific configs are then added by applying a jhipster-module on top of the default generator. Adding specific maven repositories is still done by search and replace. I guess using the needle functionality is way nicer and maybe helpful for others too.

Please feel free to ask for improvements of any kind

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1467,6 +1467,44 @@ module.exports = class extends PrivateBase {
     }
 
     /**
+     * Add a Maven Repository
+     *
+     * @param {string} url - url of the repository
+     * @param {string} username - (optional) username of the repository credentials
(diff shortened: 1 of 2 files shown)


jdubois:
Merging this, but I'm modifying the documentation and maybe the method signature, as this doesn't make much sense

cbornet:
We should have the same for Maven's pom !

sendilkumarn:
while needles are awesome, for POM I would suggest having some xml generator. there are many different options and many different ways. can we start to experiment using libraries like xml2js or maven specific xml parser.

cbornet:
BTW, grade adds jcenter to the repositories. Shouldn't we also add it to the pom for consistency ?

PR #7202, opened 2018-03-01
Fix React header

chrisdns:
hide react administration button when the user is not admin + fix react css admin tab + debug generated app when internationalization support is not selected (translations needed)


generators/client/templates/react/src/main/webapp/app/app.tsx.ejs
```diff
@@ -16,6 +16,7 @@ import AppRoutes from './routes';
 export interface IAppProps {
   location: any;
   isAuthenticated?: boolean;
+  isAdmin: boolean;
```
deepu105: I would like to do it differently so that role management is generic. So let me update this part

deepu105:
done via #7386

PR #7207, opened 2018-03-01
properly validate enum class names and values

### LINKED ISSUE #7205: entity enum type validation is permissive (class name and members name)

entity generator (or import-jdl generator) will accept java reserved keywords such as `true` and `false` with the fields of java enum type. Furthermore, the class name accepts values containing invalid characters such as the comma `,`.

The application will fail to start, it's tedious to recover if you didn't commit right before doing it and if you're using the entity generator, you must restart its creation.

##### **Reproduce the error**

Create an entity with reserved keyword such as `true` (through entity generator or JDL).

##### **Related issues**

Searched with the keywords `enum` and `validation` and didn't find any pertinent ones.

###### Entity generator :
1. Reference an existing method that validates java class names from the inquirer prompt that ask the enum name.
2. Implement a method that split around the character comma `,` and send resulting `String`s through an existing validation method that forbids java reserved keywords
3. Reference this new method from the `validate` field of the inquirer prompt that ask the question about the values of the enumeration.

###### JDL :
I don't know the JDL code as much as the entity generator but something similar could be done I guess. 
I know there is already something in place but if it can help, I implemented [something for a JHipster module](https://github.com/bastienmichaux/db-importer/blob/dev/lib/validation.js) that use a **single implementation** to validation inquirer prompts and a configuration file (could be the JDL if adapted).

##### **JHipster Version(s)**

4.14.0

Irrelevant

```json
{
    "fluentMethods": true,
    "relationships": [],
    "fields": [
        {
            "fieldName": "myenumfield",
            "fieldType": "True,false,hello",
            "fieldValues": "true,false,hello"
        }
    ],
    "changelogDate": "20180301134309",
    "dto": "no",
    "service": "no",
    "entityTableName": "boop",
    "jpaMetamodelFiltering": false,
    "pagination": "no"
}
```

Irrelevant

ruddell:
Fix #7205

### DIFF
--- a/generators/entity/prompts.js
+++ b/generators/entity/prompts.js
@@ -558,7 +558,8 @@ function askForField(done) {
                 if (input === '') {
                     return 'You must specify values for your enumeration';
                 }
-                if (!/^[A-Za-z0-9_,]*$/.test(input)) {
+                // We have whitespaces and commas allowed so that user can input a list of values split by commas.
+                if (!/^[A-Za-z0-9_,\s]+$/.test(input)) {
                     return 'Enum values cannot contain special characters (allowed characters: A-Z, a-z, 0-9 and _)';
                 }
                 const enums = input.replace(/\s/g, '').split(',');
--- a/generators/entity/prompts.js
+++ b/generators/entity/prompts.js
@@ -558,8 +558,8 @@ function askForField(done) {
                 if (input === '') {
                     return 'You must specify values for your enumeration';
                 }
-                // We have whitespaces and commas allowed so that user can input a list of values split by commas.
-                if (!/^[A-Za-z0-9_,\s]+$/.test(input)) {
+                // Commas allowed so that user can input a list of values split by commas.
+                if (!/^[A-Za-z0-9_,]+$/.test(input)) {
                     return 'Enum values cannot contain special characters (allowed characters: A-Z, a-z, 0-9 and _)';
                 }
                 const enums = input.replace(/\s/g, '').split(',');
@@ -579,9 +579,9 @@ function askForField(done) {
             },
             message: (answers) => {
                 if (!context.existingEnum) {
-                    return 'What are the values of your enumeration (separated by comma)?';
+                    return 'What are the values of your enumeration (separated by comma, no spaces)?';
                 }
-                return 'What are the new values of your enumeration (separated by comma)?\nThe new values will replace the old ones.\nNothing will be done if there are no new values.';
+                return 'What are the new values of your enumeration (separated by comma, no spaces)?\nThe new values will replace the old ones.\nNothing will be done if there are no new values.';
             }
         },
         {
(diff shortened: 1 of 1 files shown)


generators/entity/prompts.js
```diff
@@ -830,6 +837,7 @@ function askForField(done) {
         if (props.fieldAdd) {
             if (props.fieldIsEnum) {
                 props.fieldType = _.upperFirst(props.fieldType);
+                props.fieldValues = props.fieldValues.toUpperCase();
```
ruddell: This also fixes an issue when you run `export-jdl` but where the fieldValues contain lower-case letters.  JDL exects enum values in all caps.

generators/entity/prompts.js
```diff
@@ -534,6 +534,9 @@ function askForField(done) {
                 } else if (jhiCore.isReservedKeyword(input, 'JAVA')) {
                     return 'Your enum name cannot contain a Java reserved keyword';
                 }
+                if (!/^[A-Za-z0-9_\s]*$/.test(input)) {
```
MathieuAA: Why whitespaces?

ruddell: Copied it from below, thanks for catching that

MathieuAA:
@ruddell Your fix is good. However, I have two issues with the file:
  - There's a comma [here](https://github.com/jhipster/generator-jhipster/pull/7207/files#diff-59e22ebc06ff8316b80d94a3dcb8f476R561)
  - The code is duplicated.

Can I have a look before we merge it?

MathieuAA:
Oh I get it. It's an ugly trick but with the inquirer lib from yeoman I don't think there was any choice.
Here's the workflow:
  - The user gets to enter an enum's values like
    - ENGLISH
    - ENLIGH,FRENCH
    - ENGLISH, FRENCH

Then, the input (the whole string) is made exempt of whitespaces of any kind, then split by the comma character.
My bad, your PR was perfect. I'm gonna revert the change and add a comment.

ruddell:
I don't think we should allow spaces in the enum values unless we trim whitespace.  For example, `ENGLISH , FRENCH` breaks the Java build

deepu105:
@ruddell we could change the regex so that space after comma is allowed but not before comma

deepu105:
or we could just trim whitespace when splitting the values

deepu105:
I think we should just do the later to be safe

ruddell:
I edited the failing example, the actual issue is when the enum value itself contains spaces

deepu105:
Then we need to change the regex

MathieuAA:
@ruddell I don't think it comes to that, your example doesn't seem possible.

There's another solution: we change the prompt to make the user enter one enum value at a time, then clean each value.

deepu105:
If its not possible to do in current setup then lets not worry about that

ruddell:
@MathieuAA It is possible in the prompts since you added `\s` back.  I'm not familiar enough with regex to prevent spaces in words but allow them after commas.
screen shot 2018-03-02 at 3 23 54 pm

MathieuAA:
@ruddell You're right!! After I entered the enum values in the prompt, the generator kept it. That's crazy

PR #7208, opened 2018-03-02
Extract functions from entity model template

gmarziou:
Refactor Angular entity.model.ts.ejs template so that it gets simpler to maintain and test, and also to share common code with React.

### DIFF
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
@@ -16,96 +16,13 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<%_
-const variablesWithTypes = [];
-const typeImports = new Set();
-const defaultVariablesValues = {};
-let hasUserRelationship = false;
-let tsKeyType;
-if (pkType === 'String') {
-    tsKeyType = 'string';
-} else {
-    tsKeyType = 'number';
-}
-variablesWithTypes.push(`id?: ${tsKeyType}`);
-fields.forEach(field => {
-    const fieldType = field.fieldType;
-    const fieldName = field.fieldName;
-    let tsType;
-    if (field.fieldIsEnum) {
-        tsType = fieldType;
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
@gmarziou : FYI, there are some errors in Travis builds

gmarziou:
Thanks for notice, this is due to import path not handling the new entity root folder.

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
```diff
@@ -16,96 +16,13 @@
...
-import { IUser } from 'app/core/user/user.model';
-<%_ } _%>
-<%_ typeImports.forEach(typeImport => { _%>
-<%- typeImport %>;
+<%
+const variablesWithTypes = generateEntityClientFields(pkType, fields, relationships, dto);
+const typeImports = generateEntityClientImports(relationships, dto);
+const defaultVariablesValues = generateEntityClientFieldDefaultValues(fields);
+%>
+<%_ typeImports.forEach( (importedPath, importedType) => { _%>
```
deepu105: :clap: nice use of map

deepu105:
@gmarziou I merged it but then noticed that the React model file is not updated to use this. Could you plz update the same there as well?

gmarziou:
Yes I will do it, I thought you wanted to change the way imports are generated in React.

gmarziou:
BTW, React app does not compile when setting "`clientRootFolder": "test-root"` on entities.

gmarziou:
OK, I just looked at React code for entities and it's alien code to me 😄 

I thought the entity model was similar to Angular because I saw also an entity.model.ts.ejs file but it's not included in [generated files](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-client/files.js#L149-L181).
The model seems to be in entity.tsx.ejs but relationships are not typed.

@deepu105 What should I do?

deepu105:
Oh, yes that part is not finished yet. Ok nevermind, i'll do it when I make
cleanups next week. There are many items I have in pipeline to be done there


Thanks & Regards,
Deepu

PR #7213, opened 2018-03-02
Set version to 5.0.0-alpha.1 for v5 generator

gmarziou:
We discussed it on mailing list but did not reach a conclusion.
I predict issues in support if we don't set it as users are already using v5.

Feel free to reject it, not much work here :)

Also mentioned React in package.json keywords and removed AngularJS

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "generator-jhipster",
-  "version": "4.14.0",
-  "description": "Spring Boot + Angular in one handy generator",
+  "version": "5.0.0-alpha.1",
+  "description": "Spring Boot + Angular/React in one handy generator",
   "files": [
     "cli",
     "generators"
@@ -14,7 +14,7 @@
     "Spring Security",
     "JPA",
     "Hibernate",
-    "AngularJS",
+    "React",
     "Angular",
     "Twitter Bootstrap"
   ],


jdubois:
Yes let's do this, but let's release them on a beta channel on NPM.
We already discussed this we @deepu105 and that's the best solution.

Now, I need to understand how beta releases with NPM work!!!!

I've a few things to debug first, I'll do this just after.

gmarziou:
Travis fails with this error, I don't understand it:

~~~
error An unexpected error occurred: "Couldn't find any versions for \"generator-jhipster\" that matches \"5.0.0-alpha.1\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/app/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
~~~

gmarziou:
I don't have same issue with npm

gmarziou:
Seems like a known issue: https://github.com/yarnpkg/yarn/issues/2611
I'll dig further

jdubois:
It's hard to tell what's wrong here.... In related news, I had a look at doing a "beta" channel on NPM and that's very easy, so as soon as this solved, I'll do our first release.

deepu105:
I think the issue is that we are putting a version number that doesn't exist. If we just make a beta release without this PR it should work. As versions shouldn't be updated like this, it should be via a release

gmarziou:
OK why a difference of behavior between npm and yarn?

sendilkumarn:
I too feel updating the version number is not a very good idea.

gmarziou:
I can't help thinking that maven did a much better work solving all these issues than npm ;)

deepu105:
@gmarziou coz they are competing for market space :wink: 
I do agree that in order to be very flexible NPM made some rather bad choices in the beginning, like version range, which is more trouble than benefits, and many others. Some of it sounded like a good idea when the intention was only for NodeJS applications but once it spawned web and pretty much everything else all hell broke loose. But maybe its bit too late to change now I guess.

To be fair the problem we have here exists in the maven world as well If I'm not mistaken, i.e you cant download a version if it is not published right?

@jdubois can you make a beta release first, then we can clean up and merge the description change in this PR

deepu105:
You can do that with NPM as well using the `npm pack` command


Thanks & Regards,
Deepu

deepu105:
I do that all the time for ng-jhipster lib to test new unpublished versions
inside a project


Thanks & Regards,
Deepu

gmarziou:
OK, I'll try.

PR #7224, opened 2018-03-04
Fix MongoDB compile error

mraible:


generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
```diff
@@ -27,6 +27,7 @@ import liquibase.integration.spring.SpringLiquibase;
 <%_ if (databaseType === 'mongodb') { _%>
 import com.github.mongobee.Mongobee;
 import com.mongodb.MongoClient;
+import org.springframework.data.mongodb.repository.MongoRepository;
```
jdubois: I think this is only when you also use Elasticsearch - I'll have a look and correct if necessary

jdubois:
@mraible I'm moving the import - please have a look at the commit that will be linked just below

mraible:
Thanks @jdubois!

PR #7226, opened 2018-03-04
update gradle and plugins

atomfrede:
* Update gradle to 4.6
* Update apt plugins
* Adapt annotation processor config to new source set configuration


sendilkumarn:
wow, can we make use of cache benefits in the CI server?

atomfrede:
Maybe, but I think compiling java is not that crucial in caching. We should definitly make sure `bootRun` works with `--continuous` now as the boot plugin (and gradle) supports its.

PR #7231, opened 2018-03-05
Angular: Use ngtools/webpack instead of at-loader

wmarques:
It's better to lazy load modules to reduce the main bundle size and avoid giving the user data he will probably never need.

In order to make the lazy loading work with webpack we have two choices:
* Use the `angular-router-loader` for the dev config and `@ngtools/webpack` for prod (with AOT)
* Use `@ngtools/webpack` everywhere

For consistency and simplification I chose to use `@ngtools/webpack` everywhere except for the test configuration, I couldn't make it work... But I saw that we were using a fork of the "official" `istanbul-instrumenter-loader`so I put in our dependencies the official loader and not the fork which doesn't seems maintained anymore.

We can discuss this of course, I didn't see much performance issues using `@ngtools/webpack` in development mode.


wmarques:
If you have any hints on the `SyntaxError: Unexpected token import`I'm having with protractor... It doesn't seems related to the PR

generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -40,6 +41,10 @@ module.exports = (options) => ({
     },
     module: {
         rules: [
+            {
+                test: /(?:\.ngfactory\.js|\.ngstyle\.js|\.ts)$/,
+                use: [ '@ngtools/webpack' ]
```
sendilkumarn: brave move 😉 😜

generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -64,6 +69,11 @@ module.exports = (options) => ({
         ]
     },
     plugins: [
+        new AngularCompilerPlugin({
+            mainPath: utils.root('src/main/webapp/app/app.main.ts'),
```
sendilkumarn: <%= MAIN_SRC_DIR %>

generators/client/templates/angular/webpack/webpack.test.js.ejs
```diff
@@ -61,7 +61,7 @@ module.exports = (WATCH) => ({
                 test: /src[/|\\]main[/|\\]webapp[/|\\].+\.ts$/,
                 enforce: 'post',
                 exclude: /node_modules/,
-                loader: 'sourcemap-istanbul-instrumenter-loader?force-sourcemap=true'
+                loader: 'istanbul-instrumenter-loader?force-sourcemap=true&esModules=true'
```
sendilkumarn: can you change this to use syntax ?

generators/client/templates/angular/.angular-cli.json.ejs
```diff
@@ -56,11 +56,9 @@
         }
     },
     <%_ } _%>
-    "lint": [{
-            "project": "../../../tsconfig.json"
-        },
+    "lint": [
         {
```
sendilkumarn: can you try to lint this like the deleted line above ?

generators/client/templates/angular/webpack/webpack.common.js.ejs
wmarques: Haha yes ! Please try if you see any performance issues I'll rollback of course 😄

generators/client/templates/angular/package.json.ejs
```diff
@@ -78,6 +78,7 @@
     "html-loader": "0.5.0",
     "html-webpack-plugin": "3.0.0",
     "husky": "0.14.3",
+    "istanbul-instrumenter-loader": "3.0.0",
```
sendilkumarn: `3.0.0` doesn't support webpack 4 right ?

generators/client/templates/angular/webpack/webpack.dev.js.ejs
```diff
@@ -77,14 +77,6 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
             loaders: 'tslint-loader',
             exclude: ['node_modules', new RegExp('reflect-metadata\\' + path.sep + 'Reflect\\.ts')]
         },
-        {
-            test: /\.ts$/,
-            loaders: [
-                'angular2-template-loader',
```
sendilkumarn: is this dependency also removed from package.json ?

test/utils/expected-files.js
```diff
@@ -363,7 +363,6 @@ const expectedFiles = {
         `${CLIENT_TEST_SRC_DIR}spec/helpers/mock-state-storage.service.ts`,
         `${CLIENT_TEST_SRC_DIR}spec/helpers/spyobject.ts`,
         `${CLIENT_TEST_SRC_DIR}spec/test.module.ts`,
-        'tsconfig-aot.json',
         'tsconfig.json',
```
sendilkumarn: tsconfig.e2e ?

generators/client/templates/angular/webpack/webpack.common.js.ejs
sendilkumarn: Yeah sure 👍 
Did you find any good performance change? 
This also means that we will have to wait for angular-cli release cycle. `v6 - ivy` promises a lot let us see

generators/client/templates/angular/webpack/webpack.dev.js.ejs
wmarques: Nope because we need it in the webpack.test :(

generators/client/templates/angular/package.json.ejs
wmarques: I didn't have issue about this...

wmarques:
Seems that I have an issue with the interceptors who are not shared in the lazy loaded module. I think the cause is maybe the `ng-jhipster` library who exports the `HttpClientModule`, but this module should be imported only once in the app

sendilkumarn:
@wmarques WDYT about this?

Split this PR into 2, one to migrate to `ngtools-webpack` other one for `lazy loading`

The first one will fix the `yarn start` issue.

wmarques:
@sendilkumarn Yep LGTM :) I'm also investigating the HttpClient issues. So should we close this PRand i'll do two others ?

wmarques:
@sendilkumarn this PR will be about the ngtools usage, I'll create a new one for the lazy loading if that's ok for you

generators/client/templates/angular/webpack/webpack.common.js.ejs
wmarques: I'll do some tests to see the build time but it seemed pretty much the same...

sendilkumarn:
@wmarques yeah sure no issues at all. Just split it. I think `@ngtools` won't break now. So it can be merged and lazy loading we can check and merge

sendilkumarn:
pinging @deepu105 `at-loader` is not working here due to some issues. It is better time to switch to `@ngtools` WDYT ? I remember last time, you don't want to move to @ngtools

sendilkumarn:
@wmarques did you refactor the PR ?

deepu105:
@sendilkumarn @wmarques last time I didn't want to switch as it was slow in development, please test with some 10-15 entities. Each code change was taking so long to compile while running yarn start as everything was being recompiled instead of incremental changes. IF that issue no longer exists and the difference in performance while using Yarn start is very minimal then I'm ok to switch

Could you guys post some stats pre and post this change when using Yarn start

deepu105:
Just use one of our travis builds, it has lot of entities. Use the build
script under the travis folder and generate a sample app and copy the
config from that


Thanks & Regards,
Deepu

sendilkumarn:
can we try with ts-loader if we can ? It was also migrated to v4.

pascalgrimaud:
Just finished to align the registry with JHipster v5: https://github.com/jhipster/jhipster-registry/pull/251

@wmarques @sendilkumarn : do you think we should do the same for the registry too ?

deepu105:
Pascal, its not very important for Registry at the moment.We can do it once
we figure out all the details later


Thanks & Regards,
Deepu

wmarques:
@sendilkumarn You mean using ts-loader for the dev mode ? Or for dev and prod ?

deepu105:
@wmarques @sendilkumarn seems like migrating to webpack 4 didn't give us any performance improvements and our tests seems to be slower now, so I'm worried about this so can we first do a benchmark?

generators/client/templates/angular/webpack/webpack.common.js.ejs
deepu105: what matters is the time to compile when in hot reload mode for any changes to a file

deepu105:
@wmarques also take a look at https://github.com/s-panferov/awesome-typescript-loader/issues/497

deepu105:
Btw I can really see webpack performance improvement in React :)
Build time down to 15 secs with ts-loader, gonna try to make it even faster

deepu105:
So with ts-loader, cache-loader, thread-loader and fork-ts-checker-webpack-plugin I got build times of 12 secs for React with few entities. The yarn start compilation is very fast as well, ~10s fisrt run and ~3s for a change, also everything including yarn start works so we can try the same for Angular as well and compare to ngtools for performance.

deepu105:
@wmarques do you still want to do this? if so please benchmark this against the current setup with ts-loader first plz

PR #7255, opened 2018-03-08
React : Entity dialog generation, number case

agaspardcilia:
I added a case where the field is a number (Integer, Long, Float, Double) to make sure the generated dialog uses the type number for the corresponding AvInput.
It could be a good idea to display an Alert if the server returns a 400 at entity creation. 

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
@@ -256,6 +256,15 @@ _%>
                 </option>
               <%_ } _%>
               </AvInput>
+            <%_ } else if (fieldType === 'Integer' || fieldType === 'Long' || fieldType === 'Float' || fieldType === 'Double') { _%>
+              <Label id="<%= fields[idx].fieldName %>Label" for="<%= fields[idx].fieldName %>">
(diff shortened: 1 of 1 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
```diff
@@ -256,6 +256,15 @@ _%>
                 </option>
               <%_ } _%>
               </AvInput>
+            <%_ } else if (fieldType === 'Integer' || fieldType === 'Long' || fieldType === 'Float' || fieldType === 'Double') { _%>
```
deepu105: use includes it will be much nicer to read
`if (['Integer', 'Long', 'Float', 'Double'].includes(fieldType)) {`

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
```diff
@@ -256,6 +256,15 @@ _%>
...
+              <Label id="<%= fields[idx].fieldName %>Label" for="<%= fields[idx].fieldName %>">
```
deepu105: may be assign filedName as a const at line 210 to make code nicer

Issue #7252, opened 2018-03-07
React doesn't support parent-child relationships of the same entity


jdubois:
This is on the current master branch, using React.

Using the configuration below, the React front-end will not compile:
```
ERROR in [at-loader] ./src/main/webapp/app/entities/category/category.reducer.ts:13:3
    TS1117: An object literal cannot have multiple properties with the same name in strict mode.

ERROR in [at-loader] ./src/main/webapp/app/entities/category/category.reducer.ts:13:3
    TS2300: Duplicate identifier 'FETCH_CATEGORIES'.
```

I'm also worried that near to that line we have:
```
// TODO: this should be replaced by using the action and state from the related entity
```
I'm not sure if there is a link with that TODO, but it should also be solved.

Here's the output of `jhipster info` in order to reproduce:

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "io.jhipster.demo"
    },
    "jhipsterVersion": "4.14.0",
    "baseName": "tmp",
    "packageName": "io.jhipster.demo",
    "packageFolder": "io/jhipster/demo",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "clusteredHttpSession": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "clientFramework": "react",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [
      "gatling",
      "protractor"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": false,
    "aws": {
      "apps": [
        {
          "baseName": "tmp",
          "performance": "low",
          "fargate": {
            "CPU": "1024",
            "memory": "2GB",
            "taskCount": 1
          },
          "database": {
            "size": "db.t2.small",
            "supportedEngines": [
              "mariadb",
              "mysql"
            ],
            "instances": 1
          },
[...]

PR #7256, opened 2018-03-08
fix React entity camel case

chrisdns:
add camel case for entities get function


deepu105:
We should do this with lodash. Let me see if its already exposed else ill
expose lodash to templates. Then we can use the lodash method for this

deepu105:
I'm also fixing #7252 here

PR #7260, opened 2018-03-09
Add a simple error alert when an entity creation, update or deletion fails

agaspardcilia:

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
@@ -19,7 +19,7 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import { Button<% if (searchEngine === 'elasticsearch') { %>, InputGroup<% } %> } from 'reactstrap';
+import { Alert, Button<% if (searchEngine === 'elasticsearch') { %>, InputGroup<% } %> } from 'reactstrap';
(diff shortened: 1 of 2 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -232,8 +239,10 @@ export class <%= entityReactName %> extends React.Component<I<%= entityReactName
   }
 }
 
-const mapStateToProps = storeState => ({
-  <%=entityInstancePlural %>: storeState.<%= entityInstance %>.entities
+const mapStateToProps = ({ <%= entityInstance %> }) => ({
```
agaspardcilia: Looks better and don't seem to create any problem.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -91,6 +93,11 @@ export class <%= entityReactName %> extends React.Component<I<%= entityReactName
             <FaPlus /> <Translate contentKey="<%= keyPrefix %>home.createLabel" />
           </Link>
         </h2>
+        {(updateFailure) ? (
```
deepu105: I was gonna add toaster for this so that error and notification is global like in Angular. So let me update this

agaspardcilia: Alright

PR #7269, opened 2018-03-10
Angular: Remove angular2-moment

wmarques:
Angular2-moment makes the application bundle contains `moment` twice, which is quite huge... I think we should only use moment to do date manipulation.

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -34,7 +34,6 @@
     "@angular/platform-browser-dynamic": "5.2.0",
     "@angular/router": "5.2.0",
     "@ng-bootstrap/ng-bootstrap": "1.0.0",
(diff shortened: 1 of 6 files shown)


deepu105:
can you check ng-jhipster lib as well, I suggested to add the plugin to avoid writing our own pipes but if moment can directly do it than its fine. Just make sure the pipes are not used in HTML

wmarques:
Yep, I'll check that ! I agree that this lib is useful as it provides built in pipes but the cost is too high.. Moment is quite a huge lib,  Generally I use date-fns when I need to do date manipulation.

wmarques:
@deepu105  I didn't find `moment`in ng-jhipster and I didn't find pipes of the `angular2-moment`lib in our HTML templates

deepu105:
then its good, lets merge this


Thanks & Regards,
Deepu

PR #7282, opened 2018-03-12
Uses ngtools webpack plugin

sendilkumarn:
Dev build time is ~13-15s 😮 
yarn start ~6-8s 🙊 
reload time is 1s (astonishingly) on a plain project. 🎉 

This adds few minor changes on top of #7231 

Let us 👀 what travis says :)


deepu105:
@sendilkumarn @wmarques so i did some benchmarks with our ngx-default project setup (21 entities)

**Current master with ts-loader:**
yarn webpack:build
image

yarn test
image

**This PR with ngtools**

yarn webpack:build
image

yarn test:
image

I forgot to take screenshot of yarn start, but startup was ~7s and changes compiled in ~700ms with ts-loader, so IMO we should stick to ts-loader and use ngtools only for AOT, the only gain from switching would be reducing few dev dependencies, which again really doesnt matter as they are dev dependencies, let me know if you agree

jdubois:
@deepu105 can you also solve the ugly warnings about "System.import()"?

sendilkumarn:
Can you push this repo some where

deepu105:
@jdubois let me, see, I think I tried but couldn't get rid of it as its from a lib

deepu105:
@sendilkumarn which repo?

wmarques:
@deepu105 I agree with you, let's keep the ngtools only for prod build

sendilkumarn:
Can you share the huge repo with all entities in some github repo?

deepu105:
Use our travis build, go into the travis folder and run `./build-samples
generate ngx-default` it will create ngx-default-sample folder under
travis/samples


Thanks & Regards,
Deepu

sendilkumarn:
being lazy 😁

PR #7286, opened 2018-03-13
React : Fix relationship list of dialog

agaspardcilia:


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
```diff
@@ -515,7 +515,7 @@ _%>
 
 const mapStateToProps = storeState => ({
   <%_ otherEntityActions.forEach(val => { _%>
-  <%= val.instance %>: storeState.<%= val.reducer %>.<%= val.instance %>,
+  <%= val.instance %>: storeState.<%= val.reducer %>.entities,
```
deepu105: I think I used val.instance to handle user entity relations as well, can you check if this work with user entity as well

PR #7287, opened 2018-03-13
add React default translation placeholders

chrisdns:


generators/client/templates/react/src/main/webapp/app/modules/account/password-reset/init/password-reset-init.tsx.ejs
```diff
@@ -48,14 +48,18 @@ export class PasswordResetInit extends React.Component<IPasswordResetInitProps>
     if (resetPasswordFailure) {
       alertMessage = (
         <Alert color="danger">
-          <Translate contentKey="reset.request.messages.notfound" />
+          <Translate contentKey="reset.request.messages.notfound">
```
deepu105: Don't bother much about these, I'm planning to write a transform stream for this it will read msg from the i18n files and replace the component here with content when i18n is disabled. SO this PR is fine since you worked already but dont worry about updating this everywhere

PR #7293, opened 2018-03-14
React : configuration page fix

agaspardcilia:

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
@@ -66,8 +66,7 @@ export class ConfigurationPage extends React.Component<IConfigurationPageProps, 
 
   reverseIf = (condition, data) => condition ? data.reverse() : data;
 
-  sortPrefix = (configProps, rev) => this.reverseIf(rev, Object.values(configProps)
-                                                            .sort((a, b) => a['prefix'] && a['prefix'].localeCompare(b['prefix'])));
+  sortPrefix = (configProps, rev) => this.reverseIf(rev, Object.values(configProps).sort((a, b) => a['prefix'].localeCompare(b['prefix'])));
 
   sortProperties = (configProps, rev) => this.reverseIf(rev, Object.keys(configProps).sort());
 
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
@@ -18,9 +18,8 @@
 -%>
 import * as React from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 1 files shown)


generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
```diff
@@ -66,8 +66,7 @@ export class ConfigurationPage extends React.Component<IConfigurationPageProps,
 
   reverseIf = (condition, data) => condition ? data.reverse() : data;
 
-  sortPrefix = (configProps, rev) => this.reverseIf(rev, Object.values(configProps)
-                                                            .sort((a, b) => a['prefix'] && a['prefix'].localeCompare(b['prefix'])));
+  sortPrefix = (configProps, rev) => this.reverseIf(rev, Object.values(configProps).sort((a, b) => a['prefix'].localeCompare(b['prefix'])));
```
agaspardcilia: Kinda ugly way to do it, wouldn't mind an advise on how to make it a better way.

generators/client/templates/react/src/main/webapp/app/modules/administration/configuration/configuration.tsx.ejs
```diff
@@ -54,109 +53,109 @@ export class ConfigurationPage extends React.Component<IConfigurationPageProps,
...
-  sortProperties = (configProps, rev) => {
-    const { configuration } = this.props;
-    if (configuration && configuration.configProps) {
-      const listSorted = sortBy(configProps);
-      return rev ? reverse(listSorted) : listSorted;
-    }
-  }
+  sortPrefix = (configProps, rev) => this.reverseIf(rev, Object.values(configProps).sort((a, b) => a['prefix'].localeCompare(b['prefix'])));
+
+  sortProperties = (configProps, rev) => this.reverseIf(rev, Object.keys(configProps).sort());
```
deepu105: Yes, I dont think its a good idea to sort Object keys as it might cause a prefix to show wrong property. I would even suggest to get rid of sorting here and keep only filtering as that is what matters, I dont see much use of sorting these

deepu105: If it is required then best way would be to transform the object to array with both key and value first and then sort the array

agaspardcilia: Alright, I'll remove it then.

deepu105:
@jdubois I didnt test it yet, but I hope @agaspardcilia tested it and was working fine with SB2

agaspardcilia:
@deepu105 What is SB2 ?

deepu105:
Spring boot 2


Thanks & Regards,
Deepu

agaspardcilia:
Oh, how could I miss something that obvious. :see_no_evil:

I actually had to almost rewrite it completely because of SB2. This page has little to no dynamically generated stuff, it should work fine.

PR #7304, opened 2018-03-16
update node version

Tcharl:
Required for Sass and frontend-maven-plugin to work (at least on Macintosh).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - docker
 language: node_js
 node_js:
-  - "8.10.0"
+  - "9.7.1"
 jdk:
   - oraclejdk8
 addons:
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -57,7 +57,7 @@ const JAVA_VERSION = '1.8'; // Java version is forced to be 1.8. We keep the var
 const SCALA_VERSION = '2.12.1';
 
 // version of Node, Yarn, NPM
-const NODE_VERSION = '8.10.0';
+const NODE_VERSION = '9.7.1';
 const YARN_VERSION = '1.5.1';
 const NPM_VERSION = '5.7.1';
 


.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
   - docker
 language: node_js
 node_js:
-  - "8.10.0"
+  - "9.7.1"
```
deepu105: we only use LTS version, so no. Could you explain what problem you are seeing?

Tcharl:
The frontend maven plugin frontend compilation tells me that in order to be able to compile sass it need the latest node-sass (which is ok), but also node 9.7.1.

I'm on a mac, and downgrading node-sass does not work either

ruddell:
I don't see this issue on Mac.  How do you reproduce it?  I just tried a prod build with `./mvnw package -Pprod -DskipTests`

deepu105:
I think it happens because you have node 9.x installed locally. Can you use
NVM or downgrade your local node version

deepu105:
@Tcharl unfortunately we cant do this as our policy is to stick to LTS version, this might solve issue for you but might cause issue for people who are on 8.x as we might miss issues caused in that version, especially ES6 features used.

I suggest you do the below to fix your issue.

Make sure your local nodeJS version is exactly same as the version defined for the front end plugin, if not use NVM or downgrade your local version and then delete `node_modules` of the project and run `yarn install` again

The issue happens because when you generate your project and run `yarn install` the node-sass binding is compiled for the nodeJS version installed on your machine. Then when you run something via front end plugin, it tries to set up a local node version for the project and runs yarn install again, but this time it uses the version defined for the plugin and tries to compile node-sass, now since its already bound to a local version it fails.

Tcharl:
@deepu105 @jdubois thank you for the support and tips, and sorry for the inconvenience (I didn't know about the lts policy).
Let's find another convenient way to circumvent ;-)

PR #7329, opened 2018-03-20
Fix issue in docker-prompt when directory path doesn't have a trailing slash

PierreBesson:
Force adding a trailing slash in docker prompt.
This fixes the issue when the user input a path that doesn't end with a trailing slash.
Currently this broke the subgen and was very confusing for users.


pascalgrimaud:
I'm restarting Travis

PR #7330, opened 2018-03-20
React: Fix entity enum and relationship dialog

agaspardcilia:

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
@@ -139,15 +139,7 @@ export class <%= entityReactName %>Dialog extends React.Component<I<%= entityRea
     this.props.<%= val.action %>();
     <%_ }) _%>
   }
-  <%_ if (hasRelationshipQuery) { _%>
-  clearEmptyRelations = entity => {
-    Object.keys(entity).filter(k => entity[k] instanceof Object).forEach(k => {
-      if (entity[k]['id'] === '') {
-        delete entity[k];
-      }
-    });
-  }
-  <%_ } _%>
+
   saveEntity = (event, errors, values) => {
     <%_ for (idx in fields) {
         const fieldType = fields[idx].fieldType;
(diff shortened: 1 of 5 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
```diff
@@ -139,7 +139,15 @@ export class <%= entityReactName %>Dialog extends React.Component<I<%= entityRea
...
-
+  <%_ if (hasRelationshipQuery) { _%>
+  clearEmptyRelations = entity => {
+    Object.keys(entity).filter(k => entity[k] instanceof Object).forEach(k => {
+      if (entity[k]['id'] === '') {
+        delete entity[k];
+      }
+    });
+  }
+  <%_ } _%>
```
agaspardcilia: Not the best way to do it, but it's pretty quick and does not add bloat to the code.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-dialog.tsx.ejs
```diff
@@ -139,7 +139,15 @@ export class <%= entityReactName %>Dialog extends React.Component<I<%= entityRea
     this.props.<%= val.action %>();
     <%_ }) _%>
   }
-
+  <%_ if (hasRelationshipQuery) { _%>
+  clearEmptyRelations = entity => {
+    Object.keys(entity).filter(k => entity[k] instanceof Object).forEach(k => {
+      if (entity[k]['id'] === '') {
+        delete entity[k];
```
deepu105: React folks will kill us if they see use of delete, so no, lets not do this. Mutating stuff in React is heavily frowned upon

deepu105: could you explain the issue to me, maybe there is another way

agaspardcilia: When you don't put any value in a relationship field in a form, it creates a field in values that looks like this: `values.<relationship>.id: ""`. This by itself is no big deal but once you send that to the server it returns a 500 thinking that the id is invalid. I could try to create a new clean object for the server in the reducer right before calling axios. wdyt ?

deepu105: Can we check why it creates the empty id id field in the first place? May be we can avoidnit from happening unless a relationship is selected

agaspardcilia: I think it's caused by AvForm.

deepu105:
Ok lets do it in the reducer for now, make a copy of the object by copying
only valid props and use a generic util for it instead of creating a method
for all reducers. May be you could use a lodash method like 'pick' for this

PR #7338, opened 2018-03-21
Upgrade browser sync webpack plugin, remove tether and fix 7307

### LINKED ISSUE #7307: Admin Module Translations Missing for Angular
The translations are not loading for the administration pages in Angular webapps.

##### **Reproduce the error**
Generate an app (i18n enabled) and go to user management

##### **Related issues**

I think it's related to lazy-loading the admin module https://github.com/jhipster/generator-jhipster/pull/7235

##### **Suggest a Fix**
Not sure

##### **JHipster Version(s)**
Master 239acc14495d0069c64a35842065cb09ff6ea6ff

<details>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.14.0",
    "baseName": "newtest",
    "reactive": false,
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "useSass": false,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ]
  }
}
</pre>
</details>

<details>
<summary>Environment and Tools</summary>

<pre>
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

git version 2.14.3 (Apple Git-98)

node: v8.9.3

npm: 5.6.0

bower: 1.8.2

gulp:
[14:34:49] CLI version 3.9.0

yeoman: 2.0.0

yarn: 1.3.2

Docker version 18.03.0-ce-rc1, build c160c73

docker-compose version 1.19.0, build 9e633ef

</pre>
</details>

vishal423:
This change would ensure:
- Browser sync integration is compatible with webpack 4
- Remove tether.js dependency as it's not required.
- Fix #7307 - Load translation files for lazy loaded module

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -50,7 +50,7 @@
     "sockjs-client": "1.1.4",
     "webstomp-client": "1.2.0",
     <%_ } _%>
-    "tether": "1.4.3",
+    "popper.js": "1.14.1",
     "tslib": "1.9.0",
     "zone.js": "0.8.20"
   },
@@ -66,7 +66,7 @@
     <%_ } _%>
     "angular2-template-loader": "0.6.2",
     "browser-sync": "2.23.6",
-    "browser-sync-webpack-plugin": "2.0.1",
+    "browser-sync-webpack-plugin": "2.2.2",
     "cache-loader": "1.2.2",
     "codelyzer": "4.2.1",
     "copy-webpack-plugin": "4.5.1",
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -50,7 +50,6 @@
     "sockjs-client": "1.1.4",
     "webstomp-client": "1.2.0",
     <%_ } _%>
-    "popper.js": "1.14.1",
     "tslib": "1.9.0",
     "zone.js": "0.8.20"
   },
--- a/generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
@@ -18,6 +18,10 @@
 -%>
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { RouterModule } from '@angular/router';
+import { HttpClient } from '@angular/common/http';
+import { TranslateModule, TranslateLoader } from '@ngx-translate/core';
+import { TranslateHttpLoader } from '@ngx-translate/http-loader';
+import { JhiLanguageService } from 'ng-jhipster';
 
 import { <%=angularXAppName%>SharedModule } from 'app/shared';
 /* jhipster-needle-add-admin-module-import - JHipster will add admin modules imports here */
@@ -64,8 +68,22 @@ import {
     <%_ } _%>
 } from './';
 
+export function translatePartialLoader(http: HttpClient) {
+    // TODO: ideally this should load translations specific to lazy loaded module, however, few keys from lazy loaded modules are used in main bundle.
+    return new TranslateHttpLoader(http, 'i18n/', `.json?buildTimestamp=${process.env.BUILD_TIMESTAMP}`);
+}
+
 @NgModule({
     imports: [
+        TranslateModule.forChild({
+            loader: {
+                provide: TranslateLoader,
+                useFactory: translatePartialLoader,
+                deps: [HttpClient]
+            },
+            // TODO: reuse missing translation handler from ng-jhipster
+            isolate: true
+        }),
         <%=angularXAppName%>SharedModule,
         RouterModule.forChild(adminState),
         /* jhipster-needle-add-admin-module - JHipster will add admin modules here */
@@ -124,4 +142,6 @@ import {
     ],
     schemas: [CUSTOM_ELEMENTS_SCHEMA]
 })
-export class <%=angularXAppName%>AdminModule {}
+export class <%=angularXAppName%>AdminModule {
+    constructor(private languageService: JhiLanguageService) {}
+}
--- a/generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
@@ -19,11 +19,12 @@
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { RouterModule } from '@angular/router';
 import { HttpClient } from '@angular/common/http';
-import { TranslateModule, TranslateLoader } from '@ngx-translate/core';
+import { TranslateModule, TranslateLoader, MissingTranslationHandler } from '@ngx-translate/core';
 import { TranslateHttpLoader } from '@ngx-translate/http-loader';
-import { JhiLanguageService } from 'ng-jhipster';
+import { JhiLanguageService, JhiMissingTranslationHandler } from 'ng-jhipster';
+import { JhiConfigService } from 'ng-jhipster/src/config.service';
 
-import { <%=angularXAppName%>SharedModule } from 'app/shared';
+import { <%=angularXAppName%>SharedModule, i18nBundlePrefix, i18nBundleSuffix } from 'app/shared';
 /* jhipster-needle-add-admin-module-import - JHipster will add admin modules imports here */
 
 import {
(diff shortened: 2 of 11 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -50,7 +50,7 @@
     "sockjs-client": "1.1.4",
     "webstomp-client": "1.2.0",
     <%_ } _%>
-    "tether": "1.4.3",
+    "popper.js": "1.14.1",
```
deepu105: I think we should remove this from dependency and let the library which manages it to take care of it unless its required as a peer dependency

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
```diff
@@ -64,8 +68,22 @@ import {
     <%_ } _%>
 } from './';
 
+export function translatePartialLoader(http: HttpClient) {
```
deepu105: @pascalgrimaud @ruddell @wmarques @jdubois, @mraible  FYI, this is fixing the missing translation issue from lazy load but its not ideal. This is why I wasn't originally a fan of merging the jsons, may be we should split them by module

@vishal423 did you take a look at the merge json webpack plugin we use, it shouldn't be hard to split them my module. Else lets make this piece reusable as it would be needed in all lazy modules soon

generators/client/templates/angular/package.json.ejs
vishal423: @deepu105, it's specified as peer dependency in bootstrap 4. ng-bootstrap doesn't specify it as peer or/ dependency.

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
vishal423: @deepu105,
yes, I had look at merge plugin and also initially thought to split out. Few points related to TODO items:
- Few JSON keys specified in admin related bundles are shared between main and lazy loaded module. Before we split, we should move shared keys to common bundle so that they are available in main application and turn isolation off for lazy bundle to access main bundle keys (haven't tried, but, should work).
- To ensure reuse, can we move this as service to ng-jhipster project (which already have similar code for main)?

generators/client/templates/angular/package.json.ejs
deepu105: then lets remove it, we dont use JS from bootstrap

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
deepu105: Yes for both

generators/client/templates/angular/package.json.ejs
vishal423: I think we do load bootstrap JS as I see below module test to specify jquery loader for bootstrap JS:
  ```{ test: /bootstrap\/dist\/js\/umd\//, loader: 'imports-loader?jQuery=jquery' }```

In any case, I don't see any harm on explicitly loading peer dependency.

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
vishal423: I have incorporated changes for 2nd point. IMO, factory doesn't make sense as dependencies are static per module.

Also, noted another issue with changeLanguage method as that doesn't affect lazy loaded modules :(

generators/client/templates/angular/package.json.ejs
deepu105: I think its legacy code and we should remove it as well, we use ng bootstrap which doesn't use any bootstrap js

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
```diff
@@ -18,8 +18,14 @@
 -%>
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { RouterModule } from '@angular/router';
+import { HttpClient } from '@angular/common/http';
+import { TranslateModule, TranslateLoader, MissingTranslationHandler } from '@ngx-translate/core';
+import { TranslateHttpLoader } from '@ngx-translate/http-loader';
+import { JhiLanguageService, JhiMissingTranslationHandler } from 'ng-jhipster';
+import { JhiConfigService } from 'ng-jhipster/src/config.service';
```
deepu105: why qualified import? import { JhiConfigService } from 'ng-jhipster` should work

vishal423:
@deepu105, can you revert label to work in progress as I still need to split main json bundle to chunk based or should I take that as part of separate PR (not sure how much effort that would be)?

I also need to fix code when i18n is disabled, I will try that by today evening my time.

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
vishal423: I believe 'ng-jhipster' was not exporting that from index.js

generators/client/templates/angular/package.json.ejs
vishal423: If i remove popper.js dependency, then, during installation, I see below warning:
```warning " > bootstrap@4.0.0" has unmet peer dependency "popper.js@^1.12.9".```

**Comment by deepu105** [comment]
done

File: generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs (hunk 3 above)
Comment (deepu105) [reply generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs]: i'll look into it

File: generators/client/templates/angular/package.json.ejs (hunk 1 above)
Comment (deepu105) [reply generators/client/templates/angular/package.json.ejs]: thats better than bundling unused code right

File: generators/client/templates/angular/src/main/webapp/app/core/language/language.helper.ts.ejs [hunk 4]
```diff
@@ -29,6 +30,7 @@ import { FindLanguageFromKeyPipe } from 'app/shared/language/find-language-from-
 @Injectable()
 export class JhiLanguageHelper {
     renderer: Renderer2 = null;
+    private _language: BehaviorSubject<string>;
```
jdubois: @deepu105 is this good to use "_" for local variables?

deepu105: ok since there is a getter for this

deepu105:
@jdubois i'm working on this

jdubois:
Awesome !

deepu105:
@jdubois did you test user-management module in angular? it seems broken, when trying to create/edit uers

jdubois:
I thought it was broken because of i18n

deepu105:
No, its broken coz of lazy loading I guess

deepu105:
let me see if I can fix it

generators/client/templates/angular/package.json.ejs
vishal423: I had look at ng-bootstrap code and indeed they wrote custom js for popovers. I will remove popper.js dependency as that's independent.

generators/client/templates/angular/src/main/webapp/app/admin/admin.module.ts.ejs
vishal423: @deepu105 
At moment, I could only reduce i18n bundle size of admin module. In case of main chunk, it still load all bundles as there is 'title' key dependency on sub-module bundles.

I also noticed that we have each admin sub-module key under global.json that's being used for navbar menu items. If we can use same key for page title, then, it will become much simpler. WDYT?

deepu105:
@vishal423 i'm reverting your changes for splitting the json from this, lets do it separately after discussion, I might have some ideas to do it in a simpler way

deepu105:
@vishal423 please take a look at this commit https://github.com/jhipster/generator-jhipster/commit/afee53bbf146e57b69984aca65ed88b4201d3496

for splitting admin i18n files, we can do it in the ng-jhipster lib so that module just need to pass the suffix for the language file to be fetched and we can create those from webpack merge

deepu105:
For now since they are cached anyway its ok I guess

deepu105:
@jdubois I spent some time trying to figure out the failure in routing, seems to have something to do with the popup outlet, but I couldn't figure out exactly why. I'm tired so probably will take a look again tomorrow. In the meantime @wmarques  since you did the lazy loading, if you have an idea why this fails plz take a look or let me know what could be the reason so I can save some debugging time

deepu105:
@jdubois ok figured it out, pushing fix now

jdubois:
Great work @deepu105 ! It's a bit late for me too, I'll work on the v5 documentation and try to do the release tomorrow evening after work

deepu105:
No, its a partial fix, it works the first time you click edit but breaks second time :cry:

deepu105:
Probably this one https://stackoverflow.com/questions/47482230/angular-4-lazy-loading-with-named-router-outlet-not-working
Will try the solution suggested tomorrow eveing

deepu105:
I don't think we can do beta release with this part broken so lets wait to fix this

jdubois:
Yes agreed!

jdubois:
I had a look at https://github.com/angular/angular/issues/10981 and this bug is just crazy!!!

vishal423:
@deepu105, I had look at your change and it looks great. However, we should split bundles as single bundle defies the purpose of lazy loading.

deepu105:
Yes we need to but can be done sperately

Thanks & Regards,
Deepu

Issue #7350, opened 2018-03-23
[oAuth2] Unable to login when user has a specific role rather than default roles (ROLE_USER and ROLE_ADMIN)


maznag:
##### **Overview of the issue**

When I try to connect to JHipster application (Monolith with oAuth2 in my case) using a new user created with a specific role (ROLE_MANAGER for example), the server generate this exception:

<details>
<summary>Unable to find com.mycompany.myapp.domain.Authority with id ROLE_MANAGER'</summary>
<pre>
Exception in com.mycompany.myapp.service.UserService.getUserFromAuthentication() with cause = 'javax.persistence.EntityNotFoundException: Unable to find com.mycompany.myapp.domain.Authority with id ROLE_MANAGER' and exception = 'Unable to find com.mycompany.myapp.domain.Authority with id ROLE_MANAGER; nested exception is javax.persistence.EntityNotFoundException: Unable to find com.mycompany.myapp.domain.Authority with id ROLE_MANAGER'
</pre>
</details>

Tested with keycloak and OKTA

##### **Motivation for or Use Case**

Testing JHipster with oAuth2

##### **Reproduce the error**

1. Generate a simple application Monolith with oAuth2 and default options or using the following .yo-rc.json file.
mkdir test-monolith && cd test-monolith && jhipster
2. Launch keycloak server with docker-compose
docker-compose -f src/main/docker/keycloak.yml up -d
3. Start app
./mvnw
4. Connect to keycloak to add new Role and new User
- Add new role ROLE_MANAGER
http://keycloak:9080/auth/admin/master/console/#/realms/jhipster/roles
- Add new user 'manager' with password 'manager' and assign it ROLE_USER and ROLE_MANAGER roles
http://keycloak:9080/auth/admin/master/console/#/realms/jhipster/users

Note: add '127.0.0.1 keycloak' to your /etc/hosts if not exists

5. Try to login to the generated application

6. Oups ! login doesn't work

##### **Related issues**

Issue #7065

##### **Suggest a Fix**

To fix this bug, we must synchronise also the user authorities (extract from OAuth2Authentication) between IdP and JHipster's local database. To do this, I propose to modify the UserService.syncUserWithIdP as follow:
<details>
<summary>UserService.syncUserWithIdP</summary>
<pre>
private User syncUserWithIdP(Map<String, Object> details, User user) {
        // save authorities in to sync user roles/groups between IdP and JHipster's local database
        List<String> dbAuthorities = getAuthorities();
[...]

Issue #7281, opened 2018-03-12
OAuth2 configuration refinements


farrault:
##### **Overview of the issue**

I think that Jhipster generates some unused artifact with OAuth2 authentifications profiles. 
And that it could be simplified.

##### **Motivation for or Use Case**

In a Microservice or Gateway application, the application.yml file contains : 
```
security:
    oauth2:
        resource:
            user-info-uri: http://localhost:9080/auth/realms/jhipster/protocol/openid-connect/userinfo
            token-info-uri: http://localhost:9080/auth/realms/jhipster/protocol/openid-connect/token/introspect
            prefer-token-info: false
            jwt:
                key-uri: http://localhost:9080/auth/realms/jhipster
```

A vanilla Spring Boot application uses only one of these three uris. 
- it prefers the jwt.key-uri if present, to decode the access_token without calling the authorization server
- if the latter is not present, it uses user-info-uri OR token-info-uri (depending on prefer-token-info) to call the authz server and extract the user information
  but I'm not aware of a situation where it could use BOTH
(see Spring Boot's ResourceServerTokenServicesConfiguration) 
                    
That being said, in a JHipster microservice or gateway, there is a MicroserviceSecurityConfiguration that defined : 
- a custom ResourceServerTokenServices (based on UserInfoTokenServices that uses the user-info-uri )
- a custom TokenStore    
           
The presence of this UserInfoTokenServices disables the standard Spring Boot behaviour (offline jwt decoding) : 
```
   ResourceServerTokenServicesConfiguration.JwtTokenServicesConfiguration#jwtTokenServices:
      Did not match:
         - @ConditionalOnMissingBean (types: org.springframework.security.oauth2.provider.token.ResourceServerTokenServices; SearchStrategy: all) found bean 'userInfoTokenServices' (OnBeanCondition)
```
( And I didn't find the usage of the custom TokenStore ) 
      
So in the end, token-info-uri and jwt.key-uri doesn't seems to be used in either gateway or microservice application :
Jhipster calls the authz server to validate and decode the access_token
(which is probably the reason you use a cache on this call, on CachedUserInfoTokenServices (which is essentially useful on the micro-service)) 
      
      
Is that the intented behaviour ?
The presence of all these uris are miss-leading at the moment I think.
( And I can confirm that without the TokenStore/ and without token-info-uri/jwt.key-uri : a simple gateway/micro-service setup does work )

##### **Reproduce the error**

n/a

##### **Suggest a Fix**

[...]

Issue #7065, opened 2018-01-31
Improve User with OAuth Microservices


mraible:
JWT authentication seems to have the notion of a `User` when you create a gateway. This means that you can create relationships with `User` when generating code and importing JDL. However, the OAuth implementation of a gateway does not have a `User`. 

I also noticed that when importing JDL into a microservice app, it allows a relationship to `User` to go through, but then it fails on startup. 

I’d like to fix this and here are some thoughts on how we might do it:

1. If generating a gateway with OAuth, use user sync just like we do with a monolith. This means a user object/table exists and when you hit `/api/account`, the user is updated in the database with the values from the IdP. There is no write to the user table exposed to the UI or in the API.

2. Fix [the issue](https://github.com/jhipster/generator-jhipster/issues/7008) where a user doesn’t have an id and store the “sub” claim as the user’s id. This will require changing the “id” property of a user from a `Long` to a `String` for both microservices and monoliths with OAuth.

3. Fail fast or block relationships to `User` in microservice apps. Another alternative might be to allow relationships to `User`, then have a Feign client that talks back to the gateway for user information.

mraible:
I worked on storing the `sub` claim as the user's id tonight. I was unable to make it work because it seems like Hibernate won't update the id field, even if it set it and save. This [Stack Overflow Q & A] seems to confirm that. I'll try these steps:

```
1) clone the entity to a new entity.
2) delete the old entity. (be careful of your cascade children entities)
3) change the primary key of new entity (or set it null depend your generation strategy).
4) save the new entity.
```

mraible:
I tried the above steps by modifying `UserService#updateUser()` to have the following:

```java
public Optional<UserDTO> updateUser(UserDTO userDTO) {
    return Optional.of(userRepository
        .findOneByLogin(userDTO.getLogin()))
        .filter(Optional::isPresent)
        .map(Optional::get)
        .map(user -> {
            cacheManager.getCache(UserRepository.USERS_BY_LOGIN_CACHE).evict(user.getLogin());
            cacheManager.getCache(UserRepository.USERS_BY_EMAIL_CACHE).evict(user.getEmail());
            user.setId(userDTO.getId());
            user.setLogin(userDTO.getLogin());
            user.setFirstName(userDTO.getFirstName());
            user.setLastName(userDTO.getLastName());
            user.setEmail(userDTO.getEmail());
            user.setImageUrl(userDTO.getImageUrl());
[...]

PR #7351, opened 2018-03-23
save authorities in to sync user roles/groups between IdP and JHipste…

maznag:
save authorities in to sync user roles/groups between IdP and local database

this PR allow to fix a bug occurs when we try to login user with a specific role

Fix Issue #7350

### DIFF
--- a/generators/server/templates/src/main/java/package/service/UserService.java.ejs
+++ b/generators/server/templates/src/main/java/package/service/UserService.java.ejs
@@ -514,6 +514,17 @@ public class UserService {
     }
 
     private User syncUserWithIdP(Map<String, Object> details, User user) {
+        // save authorities in to sync user roles/groups between IdP and JHipster's local database
+        List<String> dbAuthorities = getAuthorities();
+        List<String> userAuthorities = user.getAuthorities().stream().map(Authority::getName).collect(Collectors.toList());
+        for (String authority :userAuthorities) {
+            if (!dbAuthorities.contains(authority)) {
+                log.debug("Saving authority '{}' in local database ...", authority);
+                Authority authoritytoSave = new Authority();
+                authoritytoSave.setName(authority);
+                authorityRepository.save(authoritytoSave);
+            }
+        }
         // save account in to sync users between IdP and JHipster's local database
         Optional<User> existingUser = userRepository.findOneByLogin(user.getLogin());
         if (existingUser.isPresent()) {


pascalgrimaud:
it would be nice to add tests too

maznag:
@pascalgrimaud There are no tests developed for this part of oAuth2 and synchronization the user infos between IdP and local database.

For this PR, I only modified the syncUserWithIdP method by adding the code that allows to save the new authorities in order to synchronize between IdP and local database.

Do you want me to add the tests for the whole syncUserWithIdP method or just for the part I added ? Thanks !

jdubois:
Did you test also test your code with MongoDB & Cassandra?
I'm worried this won't work with our "no database" option, but that's something more global, the "sync user" feature should not be there at all with this option. In fact, I'd rather have something like we have with JWT (with all user info in the token), and with OAuth2 that should be possible now.

pascalgrimaud:
Currently, the nodatabase option is broken, it doesn't compile, at least,
the last time I tried

Le lun. 26 mars 2018 à 07:31, Julien Dubois <notifications@github.com> a
écrit :

pascalgrimaud:
@maznag : to be consistent with this PR, if possible, just add a test which will cover your code. And for other methods, it can be done in another PR

maznag:
As we have now database is mandatory with oAuth2 option (for monolith and micro-service apps) we should we remove the option: 'No database'.

jdubois:
For the "no database" option: I totally understand this doesn't work with the current code. But with the new OAuth2 support, we should be able to have a token that holds both the user information and roles (like what we do in JWT), which would make those tables useless. Hopefully this will solve this issue in a much more elegant way.

maznag:
Yes I totally agree with you. But I don't understand why the feature 'sync user' between IdP and local database has been recently added for oAuth2 (work in progress for gateway + oAuth2 #7065). After searching, I found that this feature has been added to make relationship with the entity User. For me, itsn't the good reason to add this feature just to make relationship with User entity. We can add another entity (Employee or UserInfo for example) to store additionally informations for user and use the added entity to do relationship with other entities using email property as FK for example. Email can be used as primary key to identify authenticated user. So, these tables (authorities and user) will be useless.

jdubois:
Yes I totally agree with you @maznag - besides that's the design we have with JWT.
For me this comes from the time when we only had only an authorization token coming from OAuth2, but we're having the full user info in the token now (as far as I understand).
This was coded by @mraible and I wouldn't change this part without his agreement - Matt could you have a look?

farrault:
@jdubois :
Julien, as discussed quickly with you yesterday : 
The user information are indeed encoded in the access_token but : 
-  you currently do not decode the token : you call the OpenIdConnect User-info endpoint to retrieve these informations ( see #7281 ) 
- which, for me, is just fine : KeyCloak and Okta may produce JWT access_token but a compliant OIdC IdP doesn't have to do that :
  only the id_token MUST be a JWT (and only the new Spring Security 5 support for "OAuth 2.0 Login" ; that you do no use yet ; supports it natively)  
- (please also note that to be totally OpenIdConnect compliant, for security reason, you should not rely only with the information obtained through the access_token but to ALSO check the content of the id_token) 

However, this doesn't change anythings here :
=> instead of decoding a JWT, you call an endpoint to retrieve the same kind of information : in the end, Spring Security does succeed to create a User to populate the SecurityContext
I guess, you should be able to do the same things after in both case
(didn't have the time to check your JWT authentification type yet though ...)

jdubois:
@farrault with JWT we have both the user login and the authorities in the token. The main advantage of this is that there is no need to call "user info", so much less overhead. Also, there is no need for a local table where we would store that data.

For me, with OAuth2, we should have 2 designs:

- One for people using Okta/Keycloak, with a JWT token, and no need to call "user info"
- One for people using "authentication only" OAuth2 provider like Google, where you would need to store the authorities in a local table

deepu105:
@jdubois since matt has approved this can we merge it

jdubois:
@deepu105 I would like to be sure this works with MongoDB, Cassandra and Couchbase

jdubois:
No problem with Cassandra: according to the docs, you can't use OAuth2 with Cassandra (this would be an improvement in the future)

jdubois:
I'm merging this manually

jdubois:
Closing as I merged it manually

maznag:
@jdubois, thank you for your work. I'm sorry I wanted to do the tests but unfortunately I did not have the time last week.

jdubois:
Thanks @maznag !! This isn't probably the best part of JHipster - we need to find a way to make it better, maybe even put that code in https://github.com/jhipster/jhipster so people don't get this in their own projects. It's also very annoying to test all those databases, even with Docker.

maznag:
I agree with you. I started testing with mongodb under docker but I failed to run mongo with docker under MacOs.

Yes, as we have discussed before we should to found a best solution for oAuth2 as for JWT and do not need to synchronize the user infos in the local database.

PR #7357, opened 2018-03-26
Add jcenter to pom.xml and a needle to add a repository to the pom

cbornet:

Gradle build uses jcenter which is a superset of maven central.
I think we should do the same for the pom.xml at least for consistency.

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1077,6 +1077,32 @@ module.exports = class extends PrivateBase {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cbornet:
Well, jcenter is already used in the gradle build. So this is for consistency. We could remove it from both here and gradle. What  worries me is that ultimately there wouldn’t be any repository in the repositories tag that should be removed so how would we put the needle ?

cbornet:
I can remove jcenter from the pom. Do I also remove it from build.gradle ?

PR #7391, opened 2018-04-01
Fix classpath jar workaround for Windows

andreysaksonov:
Classpath is already too long to be handled on windows, so classpath should be **replaced** with classpath jar (btw, renamed task and jar to better reflect what it does). In current version of jhipster, classpath jar is **added** to already-too-long-classpath, so workaround doesn't work. Btw, this workaround is not needed on Windows 10, if "Long Paths" feature is enabled (see https://www.howtogeek.com/266621/how-to-make-windows-10-accept-file-paths-over-260-characters/), basically LongPathsEnabled should be "1" under HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem registry path.

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -95,9 +95,10 @@ springBoot {
 }
 
 if (OperatingSystem.current().isWindows()) {
-    task pathingJar(type: Jar) {
+    // https://stackoverflow.com/questions/40037487/the-filename-or-extension-is-too-long-error-using-gradle
+    task classpathJar(type: Jar) {
         dependsOn configurations.runtime
-        appendix = 'pathing'
+        appendix = 'classpath'
 
         doFirst {
             manifest {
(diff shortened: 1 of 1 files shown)


atomfrede:
Good Catch. Looks good to me. Thx for the pr.

jdubois:
I can't test but I totally understand your issue and agree on it. Should we do the same with Maven?

andreysaksonov:
I didn't find any traces of similar workaround in maven build, will take a deeper look

atomfrede:
Do we need that for maven? I think the boot run tasks work differently. At least we don't have something for maven in place as far as I can remember.

jdubois:
Oh I have no idea. But indeed I think we only had this issue with Gradle. So let's forget about Maven.

deepu105:
I don't know about Maven, I did the original workaround from my Windows days when I encountered it

PR #7406, opened 2018-04-03
Fix routing issues in admin module (user management, audit)

vishal423:
- Fix routing issues in user management, audit flows (navigation, sort etc)
- Use relative paths wherever applicable.
- Instead of aux route, use normal dialog for delete user operation.


deepu105:
@vishal423 again I'm unable to reproduce any issue in current master so unless someone else can confirm i'm putting this on hold
@pascalgrimaud did you try this with master?

deepu105:
@vishal423 ofcourse there are some improvement in this PR which we can merge once someone confirms there is an issue

vishal423:
@deepu105, @pascalgrimaud , try following flows:
- On Users, Audits list page, try pagination controls or sort by any column. You should see error in console
- When you try to delete user, you would notice a dialog, but, if you cancel or delete, url doesn't change. If you try delete again, you should notice console errors.

Also, instead of fixing current code, why do we require delete route in first place? One main reason to have route is to support bookmark and browser back and forth options. I don't think that would be applicable in this scenario. Shouldn't this be normal dialog without any route binding?

vishal423:
@deepu105, @pascalgrimaud , did you get chance to try above flows? If I am the only one facing above listed issues, then, I will close this PR.

vishal423:
FYI, above issues were reproducible with default gateway application without any changes: https://github.com/jhipster/jhipster-sample-app-gateway.git

If you are not facing above issues, then, your admin module is not lazily loaded.

ruddell:
I can reproduce using the steps mentioned by @vishal423.  You can't access the delete dialog by default or use any of the audit pagination.

I don't know much about lazy-loading but this PR does fix the functionality of the admin pages

deepu105:
@vishal423 @ruddell I tried with a fresh project again and I was able to reproduce error when trying to delete(create/view/edit works fine) so I'll try with this PR and see what happens. This is really weird as it was working last time I tried, may be something was wrong with my setup

vishal423:
If you haven't encountered, then you don't have lazy loaded admin module. I
also noted this lately with default test project.

deepu105:
It was definitely lazy loaded as I could see the partial assets being
loaded when calling an admin route

pascalgrimaud:
Just tested this.
Same as Jon, I could reproduce the issue and this PR solves it.

deepu105:
thanks @pascalgrimaud

PR #7410, opened 2018-04-04
Better Right-to-Left support according to recent changes of Bootstrap library

SidMorad:

Please let me know if something is wrong? or something else needed? thanks.


deepu105:
can you do it for React as well

SidMorad:
> can you do it for React as well

I did have a quick look at React generator and I did notice there is no concept of `rtl.*css.ejs` for React, (am I right?). If yes then more work need to be done for React in compare with these minor changes for Angular. 

I would like to help to push this forward but I almost have no experience with React. It may be helpful to point out current solution for Angular is done by pure CSS and watching html direction attribute(`<html dir=rtl>`). Therefore for supporting RTL in React, 1) `rtl.*css.ejs` need to be included as a global style and 2) html `dir` value change to `ltr` or `rtl` on language change event.

deepu105:
Doing it in React would be the same way as in Angular, you can copy the css over and include it in the app.css or app.scss file

PR #7415, opened 2018-04-05
User with only admin role should have access to account details

vishal423:


deepu105:
Ah sorry you are right

vishal423:
I did same for entity, but, then reverted as it may not be appropriate to grant access to admins by default

deepu105:
In our default setup, roles are incremental so if you are admin there is high chance you have user as well, if the users want finer control they have to change generated code anyway. Can you check whats default behaviour in Angular impl. We can follow the same here

PR #7416, opened 2018-04-05
change dbtype from timestamp to datetime for mysql

lmbai:
For the db type `ZonedDateTime` and `Instant` , we use convert it into `timestamp` for the db type. but there is limitation in mysql that the 'timestamp' only support a range of '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07'.

### DIFF
--- a/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
+++ b/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
@@ -82,17 +82,9 @@
             } else if (fieldType === 'LocalDate') {
                 columnType="date";
             } else if (fieldType === 'Instant') {
-                if (prodDatabaseType === 'mysql') {
-                    columnType="datetime";
-                } else {
-                    columnType="timestamp";
-                }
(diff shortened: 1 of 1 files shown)


lmbai:
@cbornet can u give a sample? I didn’t really get the question.

ruddell:
Yes this is compatible with `liquibase:diff`.  There are no changelogs generated after setting up the database and running `mvn liquibase:diff`

generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -82,9 +82,17 @@
             } else if (fieldType === 'LocalDate') {
                 columnType="date";
             } else if (fieldType === 'Instant') {
-                columnType="timestamp";
+                if (prodDatabaseType === 'mysql') {
```
ruddell: We can always use `datetime` in place of `timestamp` as Liquibase resolves it to the same `timestamp` where `datetime` is not supported  https://stackoverflow.com/a/28626825/3737815

PR #7424, opened 2018-04-05
Auto-add User for microservices and OAuth

### LINKED ISSUE #7065: Improve User with OAuth Microservices

mraible:
This adds a `UserEntityListener` (and config) to auto-add users in microservices. Hopefully, this will be the final piece of the puzzle to fix #7065.


mraible:
This is a work-in-progress because it works when you create a new entity, but not when you update an entity (the `UserEntityListener` is not called), or change from one user to another.

I'm not sure why the `@PreUpdate` is not called, but from reading Stack Overflow, it seems it might be because the Entity doesn't have a `@Version` attribute.

As far as changing from one user to another (when both are in the database), the error is:

```
Caused by: org.hibernate.HibernateException: Found shared references to a collection: com.okta.developer.blog.domain.User.authorities
        at org.hibernate.engine.internal.Collections.processReachableCollection(Collections.java:182)
        at org.hibernate.event.internal.FlushVisitor.processCollection(FlushVisitor.java:46)
        at org.hibernate.event.internal.AbstractVisitor.processValue(AbstractVisitor.java:104)
        at org.hibernate.event.internal.AbstractVisitor.processValue(AbstractVisitor.java:65)
        at org.hibernate.event.internal.AbstractVisitor.processEntityPropertyValues(AbstractVisitor.java:59)
        at org.hibernate.event.internal.DefaultFlushEntityEventListener.onFlushEntity(DefaultFlushEntityEventListener.java:155)
        at org.hibernate.event.internal.AbstractFlushingEventListener.flushEntities(AbstractFlushingEventListener.java:216)
        at org.hibernate.event.internal.AbstractFlushingEventListener.flushEverythingToExecutions(AbstractFlushingEventListener.java:85)
        at org.hibernate.event.internal.DefaultFlushEventListener.onFlush(DefaultFlushEventListener.java:38)
        at org.hibernate.internal.SessionImpl.doFlush(SessionImpl.java:1436)
        at org.hibernate.internal.SessionImpl.flush(SessionImpl.java:1422)
        at sun.reflect.GeneratedMethodAccessor370.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.springframework.orm.jpa.SharedEntityManagerCreator$SharedEntityManagerInvocationHandler.invoke(SharedEntityManagerCreator.java:304)
        at com.sun.proxy.$Proxy267.flush(Unknown Source)
        at org.springframework.data.jpa.repository.support.SimpleJpaRepository.flush(SimpleJpaRepository.java:534)
        at org.springframework.data.jpa.repository.support.SimpleJpaRepository.saveAndFlush(SimpleJpaRepository.java:505)
```

mraible:
FWIW, I used [blog.jh](https://github.com/jhipster/jdl-samples/blob/master/blog.jh) to test this and all Protractor tests pass on the gateway.

deepu105:
@mraible did you check how the update works in entity audit module, I dont remember adding any version fields but it was still working

mraible:
I looked into this for a couple hours this morning and still can't figure out how to solve it. A [Google search](https://www.google.com/search?q=spring+data+jpa+PreUpdate+not+called) seems to indicate that the entity is still in memory and JPA doesn't know it's updated, so that's why `@PreUpdate` isn't called.

To fix this, I tried adding `lastModifiedDate` and `createdDate` properties to my `Blog` entity:

```java
@LastModifiedDate
@Column(name = "last_modified_date")
@JsonIgnore
private Instant lastModifiedDate = Instant.now();

public Instant getLastModifiedDate() {
    return lastModifiedDate;
}

public void setLastModifiedDate(Instant lastModifiedDate) {
    this.lastModifiedDate = lastModifiedDate;
}

@CreatedDate
@Column(name = "created_date", nullable = false)
@JsonIgnore
private Instant createdDate = Instant.now();

public Instant getCreatedDate() {
    return createdDate;
}

public void setCreatedDate(Instant createdDate) {
    this.createdDate = createdDate;
}
```

Then in `BlogResource#updateBlog()`, I manually set the last modified date:

```java
@PutMapping("/blogs")
@Timed
public ResponseEntity<Blog> updateBlog(@Valid @RequestBody Blog blog) throws URISyntaxException {
    log.debug("REST request to update Blog : {}", blog);
    if (blog.getId() == null) {
    }
    blog.setLastModifiedDate(Instant.now());
    Blog result = blogRepository.save(blog);
    return ResponseEntity.ok()
        .headers(HeaderUtil.createEntityUpdateAlert(ENTITY_NAME, blog.getId().toString()))
        .body(result);
}
```

Unfortunately, the error still occurs:

```
2018-04-11 09:58:20.688 ERROR 79944 --- [  XNIO-4 task-4] c.o.d.blog.aop.logging.LoggingAspect     : 
Exception in com.okta.developer.blog.web.rest.BlogResource.updateBlog() with cause = 
'javax.persistence.EntityNotFoundException: Unable to find com.okta.developer.blog.domain.User with 
id c4af4e2f-b432-4c3b-8405-cca86cd5b97b' and exception = 'Unable to find 
com.okta.developer.blog.domain.User with id c4af4e2f-b432-4c3b-8405-cca86cd5b97b; nested 
exception is javax.persistence.EntityNotFoundException: Unable to find 
com.okta.developer.blog.domain.User with id c4af4e2f-b432-4c3b-8405-cca86cd5b97b'
```

farrault:
Hi @mraible, 
From [your github repo](https://github.com/mraible/jhipster5-ms-user), it appears that you are in a micro-service architecture.

From what I saw from the generated code : the user is created in the database ( from the data extracted from the OIdC Idp) (in com.okta.developer.blog.service.UserService.getUserFromAuthentication(OAuth2Authentication) ) only when the /account http endpoint is called by the frontend

The problem is probably the fact that this endpoint is actually handled directly by the gateway, the gateway doesn't forward the call to any micro-services : so the user is created only on the gateway database. 

The /account endpoint never being called on the microserviceS, the user is never created in their database neither.  This probably explains why your Blog microservice fails to find the user in its own database.

--- 

From an architecture point of view, in a microservice architecture, I think this is totally valid that the user doesn't exist in the microservices databases *by default* : if referenced, they are just external references, not entities and as such should be stored as plain userid only (or maybe plain login depending on the coupling)

@jdubois : when we talked about this, you said to me that in JWT case their was not any User entity available in the microserviceS neither right ?
=> whatever the target on this matter, it would probably nice to handle oauth2 and jwt in the same way no ?

mraible:
> On Apr 12, 2018, at 1:26 PM, Arrault Fabien <notifications@github.com> wrote:
> 
> Hi @mraible <https://github.com/mraible>,
> From your github repo <https://github.com/mraible/jhipster5-ms-user>, it appears that you are in a micro-service architecture.
> 
> From what I saw from the generated code : the user is created in the database ( from the data extracted from the OIdC Idp) (in com.okta.developer.blog.service.UserService.getUserFromAuthentication(OAuth2Authentication) ) only when the /account http endpoint is called by the frontend
> 

You are correct. Users aren’t created in the gateway until they log in. At Okta, we call this JIT (just-in-time) provisioning. The also won’t show up in any user dropdowns for entities that relate back to users.
> The problem is probably the fact that this endpoint is actually handled directly by the gateway, the gateway doesn't forward the call to any micro-services : so the user is created only on the gateway database.
> 
> The /account endpoint never being called on the microserviceS, the user is never created in their database neither. This probably explains why your Blog microservice fails to find the user in its own database.
> 
The code in this PR handles that. When a new entity is created, the `UserEntityListener` will add the User to the microservices database before creating the entity. In this case, the `@PrePersist` works. If you change the entity to reference a different user (that’s stored in the gateway, but not in the microservice), it fails. This is because the `@PreUpdate` does not seem to work. My guess is it could be caused by the fact that the `Blog` does not have any annotations that define its relationship to `User`, but I’m not certain.

> 
> From an architecture point of view, in a microservice architecture, I think this is totally valid that the user doesn't exist in the microservices databases by default : if referenced, they are just external references, not entities and as such should be stored as plain userid only (or maybe plain login depending on the coupling).
> 
I agree, this would be ideal. However, I’m not sure JPA allows you to just persist the id and not the whole object. If it does, we’d have to have different logic for monoliths vs microservices. 
> @jdubois <https://github.com/jdubois> : when we talked about this, you said to me that in JWT case their was not any User entity available in the microserviceS neither right ?
> => whatever the target on this matter, it would probably nice to handle oauth2 and jwt in the same way no ?
> 

In JHipster 4.x, it’s not possible to create relationships to User when you’re doing microservices. However, we don’t prevent this in `import-jdl`, so you won’t see any errors until you start your app. I’m trying to fix that because I believe it should be possible to have relationships to User in a microservices architecture.

jdubois:
@farrault yes for JWT we don't have the "User" table in the microservices.
The idea is that this is like a NoSQL database, like MongoDB for example: you don't have foreign keys. So you have a "User" table that has no link to your "BlogPost" table.

In other words, in the "BlogPost" table you can have an "author" field, which has the user login, but there is no guarantee that this login exists in the "User" table. That's a common pattern to me, when you do NoSQL.

Also, this design allows us to have "databaseless" microservices and gateways, which is important for some people (we have people having microservices talking to an SAP or a COBOL back-end).

Now, this design also has two issues:

- Using the "login" might not be a great reference key. If the user leaves the company, and his login is given to somebody else, for example.
- As @mraible says, it's not very good if you want to have some business code that uses the User object, like we often have with the JDL. In that case, synchronizing the User is in fact a good solution.

To me, there are pros and cons in synchronizing the User, so I'm mostly worried that we don't have the same behavior in JWT and in OAuth2, and that maybe that should be a specific option, so we could still have "databaseless" microservices.

farrault:
@mraible : ok , the fact that `com.okta.developer.blog.service.UserService.getUserFromAuthentication` is generated was misleading. And inserting the user at the blog insertion/update is quite tricky too.

So, at last, I succeed to run your sample  ( my Windows laptop and npm are not friend ... )

Your listener is correctly called on entity creation **and** on update : 
if you save a blog without changing the attached user, you can see it is called.

The fact is that this listener is only called at commit/flush time in update cases : 
<details>
<summary>Stack in listener on successfull update</summary>
<pre>
java.lang.Exception: null
	at com.okta.developer.blog.domain.UserEntityListener.beforeAnyOperation(UserEntityListener.java:34)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.hibernate.jpa.event.internal.jpa.ListenerCallback.performCallback(ListenerCallback.java:35)
	at org.hibernate.jpa.event.internal.jpa.CallbackRegistryImpl.callback(CallbackRegistryImpl.java:94)
	at org.hibernate.jpa.event.internal.jpa.CallbackRegistryImpl.preUpdate(CallbackRegistryImpl.java:68)
	at org.hibernate.jpa.event.internal.core.JpaFlushEntityEventListener.invokeInterceptor(JpaFlushEntityEventListener.java:51)
	at org.hibernate.event.internal.DefaultFlushEntityEventListener.handleInterception(DefaultFlushEntityEventListener.java:321)
	at org.hibernate.event.internal.DefaultFlushEntityEventListener.scheduleUpdate(DefaultFlushEntityEventListener.java:272)
	at org.hibernate.event.internal.DefaultFlushEntityEventListener.onFlushEntity(DefaultFlushEntityEventListener.java:143)
	at org.hibernate.event.internal.AbstractFlushingEventListener.flushEntities(AbstractFlushingEventListener.java:216)
	at org.hibernate.event.internal.AbstractFlushingEventListener.flushEverythingToExecutions(AbstractFlushingEventListener.java:85)
	at org.hibernate.event.internal.DefaultFlushEventListener.onFlush(DefaultFlushEventListener.java:38)
	at org.hibernate.internal.SessionImpl.doFlush(SessionImpl.java:1436)
	at org.hibernate.internal.SessionImpl.managedFlush(SessionImpl.java:493)
	at org.hibernate.internal.SessionImpl.flushBeforeTransactionCompletion(SessionImpl.java:3206)
	at org.hibernate.internal.SessionImpl.beforeTransactionCompletion(SessionImpl.java:2412)
	at org.hibernate.engine.jdbc.internal.JdbcCoordinatorImpl.beforeTransactionCompletion(JdbcCoordinatorImpl.java:473)
	at org.hibernate.resource.transaction.backend.jdbc.internal.JdbcResourceLocalTransactionCoordinatorImpl.beforeCompletionCallback(JdbcResourceLocalTransactionCoordinatorImpl.java:156)
	at org.hibernate.resource.transaction.backend.jdbc.internal.JdbcResourceLocalTransactionCoordinatorImpl.access$100(JdbcResourceLocalTransactionCoordinatorImpl.java:38)
	at org.hibernate.resource.transaction.backend.jdbc.internal.JdbcResourceLocalTransactionCoordinatorImpl$TransactionDriverControlImpl.commit(JdbcResourceLocalTransactionCoordinatorImpl.java:231)
	at org.hibernate.engine.transaction.internal.TransactionImpl.commit(TransactionImpl.java:68)
	at org.springframework.orm.jpa.JpaTransactionManager.doCommit(JpaTransactionManager.java:536)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:746)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:714)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:532)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:304)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:98)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:139)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
	at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.invoke(CrudMethodMetadataPostProcessor.java:135)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:92)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
	at org.springframework.data.repository.core.support.SurroundingTransactionDetectorMethodInterceptor.invoke(SurroundingTransactionDetectorMethodInterceptor.java:61)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
	at com.sun.proxy.$Proxy267.save(Unknown Source)
	at com.okta.developer.blog.web.rest.BlogResource.updateBlog(BlogResource.java:77)
</pre>
</details>

<p>

Your exception occurs before : when Hibernate merge your Blog entity in the JPA Persistence Context.
At this time Hibernate try to load the User entity in the JPA Persitence Context to correctly link the entities

<details>
<summary>Stack of the exception</summary>
<pre>
Caused by: javax.persistence.EntityNotFoundException: Unable to find com.okta.developer.blog.domain.User with id c4af4e2f-b432-4c3b-8405-cca86cd5b97b
	at org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl$JpaEntityNotFoundDelegate.handleEntityNotFound(EntityManagerFactoryBuilderImpl.java:159)
	at org.hibernate.event.internal.DefaultLoadEventListener.load(DefaultLoadEventListener.java:227)
	at org.hibernate.event.internal.DefaultLoadEventListener.proxyOrLoad(DefaultLoadEventListener.java:278)
	at org.hibernate.event.internal.DefaultLoadEventListener.doOnLoad(DefaultLoadEventListener.java:121)
	at org.hibernate.event.internal.DefaultLoadEventListener.onLoad(DefaultLoadEventListener.java:89)
	at org.hibernate.internal.SessionImpl.fireLoad(SessionImpl.java:1239)
	at org.hibernate.internal.SessionImpl.internalLoad(SessionImpl.java:1122)
	at org.hibernate.type.EntityType.resolveIdentifier(EntityType.java:672)
	at org.hibernate.type.EntityType.resolve(EntityType.java:457)
	at org.hibernate.type.EntityType.replace(EntityType.java:356)
	at org.hibernate.type.TypeHelper.replace(TypeHelper.java:163)
	at org.hibernate.event.internal.DefaultMergeEventListener.copyValues(DefaultMergeEventListener.java:394)
	at org.hibernate.event.internal.DefaultMergeEventListener.entityIsDetached(DefaultMergeEventListener.java:328)
	at org.hibernate.event.internal.DefaultMergeEventListener.onMerge(DefaultMergeEventListener.java:170)
	at org.hibernate.event.internal.DefaultMergeEventListener.onMerge(DefaultMergeEventListener.java:69)
	at org.hibernate.internal.SessionImpl.fireMerge(SessionImpl.java:883)
	at org.hibernate.internal.SessionImpl.merge(SessionImpl.java:869)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.springframework.orm.jpa.SharedEntityManagerCreator$SharedEntityManagerInvocationHandler.invoke(SharedEntityManagerCreator.java:304)
	at com.sun.proxy.$Proxy205.merge(Unknown Source)
	at org.springframework.data.jpa.repository.support.SimpleJpaRepository.save(SimpleJpaRepository.java:493)
</pre>
</details>

<p>

So it seems that a JPA Listener is not suitable for what you are trying to do.
The User has to exist before the call to the JPA merge with the current JPA mapping.

mraible:
@farrault I implemented an alternative version in https://github.com/jhipster/generator-jhipster/pull/7424/commits/4f8a753a04847099cc05c9180069369bf5ec2a64 that works. Thanks for taking a look and providing advice!

mraible:
Note: this implementation will likely work for JWT too (after conditionals are changed to include it).

mraible:
@deepu105 It looks like I need your help on this one. The Travis build is failing because `saveIdentitySnapshot` is not defined. I think I'm defining it in the right places in `generator-base.js`, but apparently not. I tried defining it in `entity/index.js` in my last commit, but that doesn't seem to help anything. Thanks in advance for any advice!

deepu105:
@mraible sure i'll take a look as soon as I find time, I'm holidaying in Paris now

generators/generator-base.js
```diff
@@ -1756,9 +1756,14 @@ module.exports = class extends PrivateBase {
...
             context.fieldNameChoices.push({ name: field.fieldName, value: field.fieldName });
         });
+        context.hasUserField = false;
         context.relationships.forEach((rel) => {
             context.relNameChoices.push({ name: `${rel.relationshipName}:${rel.relationshipType}`, value: `${rel.relationshipName}:${rel.relationshipType}` });
+            if (rel.relationshipName === 'user') {
+                context.hasUserField = true;
+            }
         });
+        context.saveIdentitySnapshot = context.applicationType === 'microservice' && context.authenticationType === 'oauth2' && context.hasUserField;
```
deepu105: I didnt take a  close look but may be this should be done in the loadInMemoryData section, as this method might not be called when creating entity from CLI. There is a forEach loop on relations there as well.

deepu105: @mraible could you let me know if that works?

mraible: It does work - thanks @deepu105! 🙏

mraible:
@jhipster/developers What do y'all think of this improvement? I can easily merge, but I'd rather have an approval first. ;)

jdubois:
@mraible so in your final code there is no `UserEntityListener `? You are manually saving the user, am I correct?
The main issue for me is, will we still be able to have "databaseless" microservices and gateways? I found them really cool.
Otherwise I'm fine with this, @farrault @deepu105 WDYT?

jdubois:
Oh no, if you are "databaseless" you can just use the Spring Controller sub-generator.

mraible:
In that case, this change should be fine since it only kicks in when you’re using the entity generator.

jdubois:
OK for me then (I didn't check your code, I'm late for my Devoxx demos)

generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
```diff
@@ -111,7 +114,12 @@ public class <%= entityClass %>Resource {
         log.debug("REST request to save <%= entityClass %> : {}", <%= instanceName %>);
         if (<%= instanceName %>.getId() != null) {
             throw new BadRequestAlertException("A new <%= entityInstance %> cannot already have an ID", ENTITY_NAME, "idexists");
-        }<%- include('../../common/save_template', {viaService: viaService, returnDirectly: false}); -%>
+        }<%_ if (saveUserSnapshot) { %>
+        if (<%= instanceName %>.getUser() != null) {
+            // Save user in case it's new and only exists in gateway
+            userRepository.save(<%= instanceName %>.getUser());
```
cbornet: Shouldn't we verify that the login and id are also not null ?

cbornet: Also it fails if the entity uses DTOs as DTOs don't have a `User` attribute but `userId` and `userLogin` instead.

mraible: Are you commenting on an old issue? You know that's not allowed, right? ;)

Sounds like a bug so probably best to open an issue. I don't think it's necessary w/o DTOs because the User is usually populated via dropdown.

cbornet: Yes, I'll open an issue later (no time right now...). It's just in case I forget...

dean0bambin0:
Hey, sorry if I shouldn't be posting on merged issue.  Just wondering about the final code here where user repository is injected into entity resource.  How will this work when user entity is updated in IDP, it will not sync with the microservice user.  Also DB's will be difficult to keep in sync across multiple microservices, no? I would have thought listener approach. Thanks and sorry if shouldn't be posting here.

dean0bambin0:
Awesome thanks for reply. I couldn't see this happening when looking through the code. Could only see the call in the gateway. I'll debug this eve. Cheer again

dean0bambin0:
> The user is synched in the microservice whenever the user logs in.

I just tested this and it is not working in my application, I also tested using the "Blog" project and it's not working.  I can see userservice is created in both the gateway and blog microservice with method called syncUserWithIdP.  This method is called on login in the gateway and user is synch'd to gateway DB but not happening in the microservice.  I can't see this method being used anywhere in the microservice for it to be called.

dean0bambin0:
Think should log a ticket??

mraible:
In the case of microservices, I believe it might be that the user isn't saved to the microservice database until a record is saved that links to that user.

dean0bambin0:
Yes, user is added when record is stored in microservice db. But the problem is that it is never re-synchd after inital creation so you easily end up with a user in your microservice db thats out of sync with user in gateway db/keycloak. For example if you create record in microservice and user is added. Then user updates all their details in keycloak/gateway (maybe numerous updates over time) this is never propogated to microservice to stay in sync.

mraible:
I’m not sure how to fix this, but you’re welcome to enter an issue to track it.

clement26695:
I think we discussed in #9949 about the same topic, approximately. To keep your databases in sync, you could use Kafka. We will try to enhance Kafka support in the future to ease this (ping @pascalgrimaud)

PR #7426, opened 2018-04-06
Mark created_date as non updatable

andreasfranz:
Mark created_date as non updatable, since it should never be changed after initial creation

### DIFF
--- a/generators/server/templates/src/main/java/package/domain/AbstractAuditingEntity.java.ejs
+++ b/generators/server/templates/src/main/java/package/domain/AbstractAuditingEntity.java.ejs
@@ -56,7 +56,7 @@ public abstract class AbstractAuditingEntity implements Serializable {
     private String createdBy;
 
     @CreatedDate<% if (databaseType === 'sql') { %>
-    @Column(name = "created_date", nullable = false)<% } %><% if (databaseType === 'mongodb' || databaseType === 'couchbase') { %>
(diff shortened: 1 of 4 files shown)


andreasfranz:
Could someone please advise me why the Travis checks failed? Following the steps in the documentation they run on my local machine.

jdubois:
I just looked at the first failed build, it looks like you have a JUnit error in `io.github.jhipster.sample.service.UserServiceIntTest` - do you know how to execute the Travis tests on your machine?

cbornet:
It’s because one of the tests needs to update the creation date

andreasfranz:
@cbornet Thanks for the info, but it sounds kind of strange to me that the only reason to not mark it as `updatable=false` is a test.

deepu105:
If you can find another way to test it then we can merge this. May some
reflection magic or something?

jdubois:
This is really annoying to close this because of a test!!! But indeed I can't find another solution, and the test is important

andreasfranz:
I thought about this again, I think I might have a solution and will try to implement it later.

jdubois:
I'm closing this as this has been stuck for 3 weeks and isn't very important - but that's definitely stupid to make it updatable only for tests!!

cbornet:
One way to do it could be to create a dedicated class to get the current time that is overridden during tests. Not too hard to do I think.

PR #7428, opened 2018-04-06
Minor polishing

erikkemperman:
Some minor polishing. Note, mutual dependency on a server-lib PR. Both PRs have a temporary commit at the tip of the branch which should _not_ be merged but are there to make Travis pass.

https://travis-ci.org/erikkemperman/generator-jhipster/builds/363105252?utm_source=github_status&utm_medium=notification

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,11 @@ env:
     - SPRING_JPA_SHOW_SQL=false
 
     # if JHIPSTER_LIB_BRANCH value is release, use the release from Maven
-    - JHIPSTER_LIB_REPO=https://github.com/jhipster/jhipster.git
-    - JHIPSTER_LIB_BRANCH=release
+    - JHIPSTER_LIB_REPO=https://github.com/erikkemperman/jhipster.git
+    - JHIPSTER_LIB_BRANCH=prepare-upgrade
     # if JHIPSTER_BRANCH value is release, use the release from NPM
-    - JHIPSTER_REPO=https://github.com/jhipster/generator-jhipster.git
-    - JHIPSTER_BRANCH=release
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -32,11 +32,11 @@ env:
     - SPRING_JPA_SHOW_SQL=false
 
     # if JHIPSTER_LIB_BRANCH value is release, use the release from Maven
-    - JHIPSTER_LIB_REPO=https://github.com/jhipster/jhipster.git
-    - JHIPSTER_LIB_BRANCH=release
+    - JHIPSTER_LIB_REPO=https://github.com/erikkemperman/jhipster.git
```
deepu105: I assume this was an accident :smiling_imp:

erikkemperman: Yes and no — it should not be merged, obviously, but I don’t know how else to make Travis pass on mutually dependent changes across server-lib and generator? See also my note above.

If approved these should both be merged (minus the “temp” commits) and pushed simultaneously, then it should work (this is what I did with my branches too).

If there is a better way to do this I’d love to learn about it!

erikkemperman: PS I could do the merge myself, if approved by you guys, I know this “temp” business is annoying!

deepu105: Shouldn't the change for server lib be done first? so that once we merge that we can merge it here, of course the generator tests on server lib might fail but if you have tested it locally I guess it should be good. But ya in general as long as you take care not to have unwanted changes ending up in master its fine.

erikkemperman: Indeed, I like the reassurance of both builds passing... Merging the server-lib PR first (despite red Travis) and then the generator later would also work, but requires intervention by Julien to publish the snapshot.

I understand the arguments against it, and this kind of situation is relatively infrequent I guess, but the mono-repo idea would have helped here ;-)

PR #7452, opened 2018-04-09
do not allow to create an entity inside @PutMapping

### LINKED ISSUE #7425: Entity is being created when using `@PutMapping`

_This is not a bug but rather an opinionated point._

For any entity generated by jhipster, `@PutMapping` code include to create the entity if the id is null.

1. By using HTTP PUT, we assume that the underlying resource will be updated. For creating a new resource, HTTP POST should be used. Currently in jhipster, PUT is also creating a new resource which also results in an another following issue.

1. Consider we do not want the user with role `ROLE_USER` to create a new resource. So we do `    @Secured(AuthoritiesConstants.ADMIN)` on `@PostMapping`, but this will not restrict a normal user from creating a new resource because he/she still can create using `@PutMapping`.

##### **Reproduce the error**

- start application
- using dev tool, change id of an entity to null before save method call.
- a new resource will be created

##### **Related issues**
-
##### **Suggest a Fix**

Remove method call from `@PutMapping` to `@PostMapping`.
And if client sends id as null, return some error message.

##### **JHipster Version(s)**

v4.x.x, v5.0.0-beta

-

-

-

pmverma:
Fixes #7425 , Once message for `idnull` confirmed I will update to rest of i18n files. Until that please do not merge.


generators/languages/templates/src/main/webapp/i18n/en/global.json.ejs
```diff
@@ -152,7 +152,8 @@
         "Size": "Field {{ fieldName }} does not meet min/max size requirements!",
         "userexists": "Login name already used!",
         "emailexists": "Email is already in use!",
-        "idexists": "A new {{ entityName }} cannot already have an ID"
+        "idexists": "A new {{ entityName }} cannot already have an ID",
+        "idnull": "Invalid ID"
```
deepu105: Please add the keybto all other languages as well. No need to translate just add the English version

pmverma: @deepu105 I have added to remaining langs, please review and merge.

PR #7456, opened 2018-04-09
Add a streaming/reactive endpoint for Mongo and Couchbase

cbornet:

This PR:
- creates a reactive repository if the reactive option is on for Mongo and Couchbase
- creates a streaming reactive WebFlux endpoint with mediatype application/stream+json
- adds a test using WebTestClient for this endpoint

This endpoint has the advantage to be fully reactive end-to-end. It can be used to fetch large collections out of the database without OOM (and without the drawbacks of pagination fetching such as add/delete of entities during the extract and repeated queries to the database).

Cassandra could also be done but this would require to completely move to spring-data-cassandra.
For JPA, there could be some interesting things to do with Scrollable ResultSets but I need to investigate more.

### DIFF
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -104,7 +104,7 @@ const serverFiles = {
             }]
         },
         {
-            condition: generator => (generator.reactive && ['mongo', 'couchbase'].includes(generator.databaseType)),
+            condition: generator => (generator.reactive && ['mongodb', 'couchbase'].includes(generator.databaseType)),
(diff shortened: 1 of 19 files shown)


jdubois:
This is an insane amount of work!! When do you have the time to code all that?
I'm merging this now as I'm about to test Cassandra and Couchbase, so I'd rather only do it once!

jdubois:
Yes we should move to Spring Data Cassandra, but that's a lot of work and only very few people use it...

PR #7476, opened 2018-04-12
Prettier on scss and css files generated + reformat css.ejs and scss.ejs

### LINKED ISSUE #7451: Use Prettier to css and scss files

Actually some css and scss files have style coding problems. This cause style inconsistence.

I've found some errors on css files and scss files in React and Angular part. 

##### **Related issues**

* pull #7448 from @agaspardcilia  or #7449 with their issue #7446 who show some css rules problems
* Pull: #7265 and its corresponding issue #7172 by @gmarziou who introduced the actual behaviour of Prettier
* issue: #7431 who debate about code indentation. 

##### **Suggest a Fix**
1. As said @deepu105, we could use Prettier to resolve easly this problems. Majors problems are the indentation, and some rules likes space around ":" or space around ">" or carriage-return after ",".  It's sometimes inconsistent. With Prettier, we could easy fix that, as @deepu105 says.
2. Make a PR to correct actual all `(s)css.ejs` problems (I could do that, it's easy). I work on a PR to fix that. We will have consistent (s)css.ejs if you accept that. 

##### **JHipster Version(s)**
JHipster 5.0.0-beta.0

JulioJu:

Fix #7451


deepu105:
@JulioJu great work.

generators/client/templates/angular/.prettierrc.ejs
```diff
@@ -1,7 +1,18 @@
...
-parser: typescript
-arrowParens: avoid
 tabWidth: 4
+useTabs: false
+overrides:
+- files: "src/**/{*.ts,*.tsx}"
+  options:
+    jsxBracketSameLine: false
+    arrowParens: avoid
+    parser: typescript
```
JulioJu: This line is not mandatory. Prettier detects the parser needed.
@gmarziou. To simplify prettierrc, we could delete all "parser" lines?  What do you think about that? I asked me before commit, but I didn't dare delete that… 
Maybe it's better to have a more accurate prettierc? We have a complete syntax… Maybe it's better ^^! I believe it's better to see what we do… As you wish ;-). 

Note: parser will be renamed "language" (see https://github.com/prettier/prettier/issues/2846).

generators/client/templates/angular/.prettierrc.ejs
```diff
@@ -1,7 +1,18 @@
...
+- files: "src/**/*.css"
```
JulioJu: @gmarziou : same as above: should we delete it ?

travis/scripts/00-install-jhipster.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
```
JulioJu: There is a strange error so I enable the bash option « trace for each command after it expands the command and before it executes  it » to see accurately where there is a bug. Advise at https://docs.travis-ci.com/user/installing-dependencies/#Installing-Projects-from-Source. 

**Note:** when I run `$ ./build-samples.sh build ngx-default` or `yarn prettier:format` or on my computer, all work fine. Strange!

jdubois:
This is totally awesome!!! Only downside is that @deepu105 you need to modify your Devoxx slides to include this :-D

generators/client/templates/angular/.prettierrc.ejs
deepu105: Lets leave it like that as its more explicit

JulioJu: @deepu105. I'm pretty sure, this syntax doesn't work. We could deduce that beccause we could see than 7fcd4b03a96bf10be6cd3ceb843b9c83f1f8a7cc works, but 7a854162f1d3729675bfabee2aa41666e3c3a3f0 fail. My assumption is that somewhere (but I don't know where) a command `$ yarn prettier --write --parser css "src/**/*.ts"` is launched. In my computer, when  in folder ngx-default-sample, when I run command `$ yarn prettier --write --parser css "src/**/*.ts"`, I have the same kind of error than in line 1526 of https://travis-ci.org/jhipster/generator-jhipster/jobs/365956929 (build of 7a854162f1d3729675bfabee2aa41666e3c3a3f0).

We could read at line 1526 of https://travis-ci.org/jhipster/generator-jhipster/jobs/365956929 than  
```
SyntaxError: (postcss) CssSyntaxError Unknown word (1:1)
> 1 | export * from './foo.service';
    | ^
  2 | export * from './foo-update.component';
  3 | export * from './foo-delete-dialog.component';
```
I don't understand this error beacause in folder "ngx-default-sample" there isn't file "foo.service", there isn't entity "foo". Have you an ideay why on remote Travis it tests folder "foo"?

As 7fcd4b03a96bf10be6cd3ceb843b9c83f1f8a7cc works, I could --hard reset to this commit ?

Or maybe I could try simply with a .prettierrc without explicit parser? Thanks 7fcd4b03a96bf10be6cd3ceb843b9c83f1f8a7cc, and with the result of the command « `$ yarn prettier --write --parser css "src/**/*.ts` » I believe it's better to let prettier choose the parser corresponding to the file extension. 
```typescript
printWidth: 140
singleQuote: true
tabWidth: 4
useTabs: false
overrides:
- files: "src/**/{*.ts,*.tsx}"
  options:
    jsxBracketSameLine: false
    arrowParens: avoid
```

What do you think about that?

JulioJu: @deepu105 I'm going to test the syntax above with a new commit…

JulioJu:
@deepu105 it works! 
I prefer 7fcd4b03a96bf10be6cd3ceb843b9c83f1f8a7cc than 3625f4778cfefefeabc78ff682524c9aaef13f70 because  7fcd4b03a96bf10be6cd3ceb843b9c83f1f8a7cc it's more clear and maybe more reliable and more human readable. 

There is a strange behaviour in the Travis build of 7a854162f1d3729675bfabee2aa41666e3c3a3f0. I've let an hypothesis, but it's strange. I believe it's better to let Prettier which rules and which parser works on a filetype. Be too much accurate in .prettierrc seems introduce confusion during Travis build. Only let a comment for humans is I believe better, and let Prettier works freely is I believe more reliable.

deepu105:
@JulioJu ok for your last comment, let prettier choose options, if that works fine. Can you cleanup your commits, may be squash the reverts

JulioJu:
@deepu105. Ok squashed ! Commit with `set -ex` deleted because this afternoon I will work on a separate PR for that (see issue #7479).

JulioJu:
@deepu105 
I havn't luck with Travis Builds ;-) 
Network connection problem again:
«
[FATAL] Non-resolvable parent POM for io.github.jhipster:jhipster-dependencies:[unknown-version]: Failure to find io.github.jhipster:jhipster-parent:pom:2.0.0 in http://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced @ io.github.jhipster:jhipster-dependencies:[unknown-version], /home/travis/.m2/repository/io/github/jhipster/jhipster-dependencies/2.0.0/jhipster-dependencies-2.0.0.pom, line 5, column 13
» (https://travis-ci.org/jhipster/generator-jhipster/jobs/366048640).
So I start again the Travis build

jdubois:
No the jhipster-parent failure is my fault!!

JulioJu:
@jdubois ok :-) ! 
So could you restart the travis build when all will be ok ? We could see at https://travis-ci.org/jhipster/generator-jhipster/builds/366077749?utm_source=github_status&utm_medium=notification that all Travis build won't success ;-)

What is « jhipster-parent » ?

jdubois:
"jhipster-parent" is a new library on which "jhipster-framework" and "jhipster-dependencies" depend - they should all be served from Maven Central, and this is what I failed this morning.
I just released version 2.0.1 and upgraded the generator, see https://github.com/jhipster/generator-jhipster/commit/f12c59c3326076d3346685c72cb1a470dfa767b1 - it should be good, but it's not fully tested yet. I'm monitoring this right now.

JulioJu:
@jdubois thanks for your explanation :-) ! All work fine now ;-). 
@deepu105 thanks for the review and the merge :-).

PR #7481, opened 2018-04-13
Reduce IDE warnings

erikkemperman:
I realize we're probably never going to get completely rid of IDE warnings for generated apps, but this will reduce the number considerably.

These changes are all quite minor -- but there are many of them...

Travis passes:
https://travis-ci.org/erikkemperman/generator-jhipster/builds/366090249

- Please make sure the below checklist is followed for Pull Requests.

- [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x] Tests are added where necessary
- [x]  Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #7483, opened 2018-04-13
Align versions

erikkemperman:
As requested [here](https://github.com/jhipster/generator-jhipster/pull/7481#issuecomment-381163037).


jdubois:
Hold on, Travis is still broken for jhipster/jhipster and I have no idea how to fix it....
It's crashing on `travis/scripts/03-replace-version-generated-project.sh` and I don't understand what the issue is

erikkemperman:
I am traveling now so can’t check properly. But this script still needs to be changed the properties from `jhipster-framework` to `jhipster-dependencies` after your change in the generator:
https://github.com/jhipster/jhipster/blob/master/travis/scripts/00-replace-version-generator.sh

I’ll check when I get home...

jdubois:
Thank a lot, so happy to have all this running fine, at least!

PR #7492, opened 2018-04-15
Feign improvements for JWT

PierreBesson:
While working on my Devoxx demo, I took the time to do these improvements which have been much discussed in the past and will align JWT with OAuth :

- Add a FeignInterceptor to forward the JWT token
- Enable Feign Clients by default for microservices (not for gateways currently, should I add the support ?).
- Set the correct hystrix/feign properties
- (refacto) Rename MicroserviceSecurityConfiguration -> SecurityConfiguration. It was an inconsistency from v3 which was my fault initially and it is very confusing.


mraible:
@PierreBesson It looks like you need to rename the class in `MicroserviceSecurityConfiguration.java` as well:

```
[ERROR] /home/travis/app/src/main/java/io/github/jhipster/travis/config/SecurityConfiguration.java:
[28,8] class MicroserviceSecurityConfiguration is public, should be declared in a file named 
MicroserviceSecurityConfiguration.java
```

jdubois:
@PierreBesson as soon as your build is green, don't hesitate to merge if you need this for Devoxx - I'm doing the same thing for my demo :-)

PierreBesson:
@jdubois It's OK I'm not going to be generating apps on stage.
The question is should I enable FeignClients for the gateway as well. As we already added the "openfeign" dependency it seems strange that we don't enable it and make it easy to use.

PR #7513, opened 2018-04-20
React: removed an useless regex

agaspardcilia:
@deepu105 I think that this regex is not finding anything and that the second one is doing all the work. Can you confirm this please ?

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -268,7 +268,6 @@ function replaceTranslation(body, generator) {
         }
     };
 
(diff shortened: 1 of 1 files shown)


generators/utils.js
```diff
@@ -268,7 +268,6 @@ function replaceTranslation(body, generator) {
         }
     };
 
-    replaceRegex(/(\{translate\('([a-zA-Z0-9.\-_]+)'(, ?null, ?'.*')?\)\})/g, '""');
```
deepu105: @agaspardcilia on closer look I think this might still be needed for cases where we use 
`<element prop={translate('foo')} />`
SO lets leave it in if its not causing any issues

Issue #7516, opened 2018-04-21
Feature request: Support font-awesome v.5


gomesp:
##### **Overview of the issue**

JHipster is currently supporting font-awesome 4.7.0. The project has now released version 5.0.10. Could we please add support to it?

Unfortunately it is not a simple change of the package version, as the icon names [have changed slightly](https://fontawesome.com/how-to-use/upgrading-from-4). 

This could even trigger another option for the generator, e.g. "What font-awesome icon set would you like to include in your project? 1. Solid, 2. Regular, 3. Light".

I am keen to have a go and open a pull-request at some point if the project is in agreement for the change.

##### **Motivation for or Use Case**

To keep up with the font-awesome project development.

- [x] Checking this box is mandatory (this is just to show you read everything)

gmarziou:
+1
Yes we should upgrade.

> This could even trigger another option for the generator, e.g. "What font-awesome icon set would you like to include in your project? 1. Solid, 2. Regular, 3. Light".

Currently we have 63 places where we use icons in our templates.
It's very easy for a user to do a search/replace on `class="fa` to select another variant.
So I'm not sure a new question is really needed.

pascalgrimaud:
+1 too

deepu105:
Will you consider doing a PR

gomesp:
@deepu105 if you give me 1 week I can do a PR. I imagine that the agreement is to simply replace the package with v.5 and test the generator again, right? 

I will keep the icon set with the solid family (fas or fa) to avoid a lot of regression testing. I may have change the css class for any icons that are now under the "font-awesome brands" (e.g. github, etc).

jdubois:
Yes it should not be very hard but that's quite time consuming to test... oh and please have a look at Angular and React - for this part it should be very similar. 
Anyway big big +1 if you do that PR!
(At work we are still with v4 and I'd love to have some of the new v5 icons)

gomesp:
I just opened the pull request https://github.com/jhipster/generator-jhipster/pull/7519.

I decided to open it sooner so that you guys could give feedback faster. I'm not sure if my approach for including the SVG/JS icon set is according to guidelines. If not I can change it and resubmit the PR. 

@jdubois I only worked in the angular codebase. But that was mostly because I could not find any instance of the FA icons in the react codebase! That may have to do with me not knowing the first thing about react. ;)

I can look into the react codebase again tomorrow to figure out where FA is being used and open another PR just for react.

[...]

PR #7519, opened 2018-04-22
Upgrading font-awesome from 4.7.0 to 5.0.10

gomesp:
The new font-awesome package uses the SVG with JS icon set. The previous
icon set has now all been renamed to the new icons present in FA5.

This change has been applied to the angular codebase only.


generators/client/templates/angular/src/main/webapp/app/layouts/index.ts.ejs
```diff
@@ -28,3 +28,41 @@ export * from './navbar/active-menu.directive';
 export * from './profiles/page-ribbon.component';
 export * from './profiles/profile.service';
 export * from './profiles/profile-info.model';
+// Imports all fontawesome core and solid icons
+
+import fontawesome from '@fortawesome/fontawesome';
```
gmarziou: I'm not sure layouts/index.ts is the best place for these global imports, personally I would have put them in vendor.ts as we do for CSS but I could be wrong.

gmarziou:
Have you tested a prod build? This is where most issues occur usually.

gmarziou:
Travis build fails on app using SASS, you must edit also vendor.scss.ejs

gomesp:
@gmarziou Thanks for pointing the sass issue. I have tested the generator with the default settings at first (so no sass). Can I run the travils build in my own fork before submitting the PR?

generators/client/templates/angular/src/main/webapp/app/layouts/index.ts.ejs
gomesp: That's exactly why I wanted to submit the PR soon, so I could get some input of the preferred place for the global imports. I'll move it vendor.ts. 👍

generators/client/templates/angular/package.json.ejs
```diff
@@ -36,7 +36,8 @@
     "@ng-bootstrap/ng-bootstrap": "1.0.1",
     "bootstrap": "4.0.0",
     "core-js": "2.5.3",
-    "font-awesome": "4.7.0",
+    "@fortawesome/fontawesome": "^1.1.5",
+    "@fortawesome/fontawesome-free-solid": "^5.0.10",
```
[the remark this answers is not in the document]
gomesp: Thanks, will do! 👍

pascalgrimaud:
Yes, just link your fork to your own Travis

gomesp:
Hi team, I fixed the issues with the saas version and applied the feedback given to me in this PR.

The travis tests are now passing in my for (thanks for pointing out how to run travis on it) and I have generated 2 apps with these changes locally, one with sass and the other without sass.

I hope everything is in good shape now to close the issue #7516.

gmarziou:
Looks good to me, maybe you could squash your 2 commits into a single one.

generators/client/templates/angular/src/main/webapp/app/vendor.ts.ejs
```diff
@@ -27,4 +27,45 @@ import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/catch';
 import 'rxjs/add/observable/of';
 import 'rxjs/add/observable/throw';
+
+// Imports all fontawesome core and solid icons
```
deepu105: Doesn't it support web fonts anymore?
Does the SVG + JS approach have a significant saving of bundle size?

gomesp: This is the recommended approach by the FA project. They still support web fonts as well. I guess this is a core project decision, which direction you want to take (if different than the recommended one). I don't have a comparison of the bundle size between the options, but I can figure out if you guys want.

PierreBesson: The "SVG font" approach seems to have a lot of advantages and it's been a hot topic of discussion, the consensus being that they are better than "Icon Fonts". Here are 3 links about it :

https://www.ianfeather.co.uk/ten-reasons-we-switched-from-an-icon-font-to-svg/
https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/
https://blog.github.com/2016-02-22-delivering-octicons-with-svg/

Especially now that it's supported by FA and easy to setup in our webpack build, I think it's safe to move to this approach.

deepu105: yup, I'm aware of it, I myself use SVG fonts in my projects, however, I was concerned about the JS part of if you look at some other SVG font libs they don't use JS. But then its not a huge concern

deepu105: so lets do it

deepu105: Also seems they will release a Angular version soon https://fontawesome.com/get-started/advanced-options

deepu105: we could migrate to the angular version when released

gomesp:
Woot, I got a PR merged! 🎉

pascalgrimaud:
thanks and congrats @gomesp :-)

jdubois:
Awesome, congrats!!

PR #7523, opened 2018-04-23
React : fix user management update page

agaspardcilia:
Using `model={isNew ? {} : user}` was sometimes displaying outdated values in form when updating an User. For instance, if you were editing `user` and then directly switching to `admin`, you had 50% chances that values in the form were still about `user` and not `admin` as you would expect.


deepu105:
do you see similar behavior in entity?

agaspardcilia:
No. Actually, I fixed it thanks to the entity implementation.

PR #7536, opened 2018-04-25
React : Upgrade font-awesome lib

agaspardcilia:
Related to #7516.
@gomesp Here's the PR.

@deepu105 The lib [documentation](https://github.com/FortAwesome/react-fontawesome/blob/master/README.md#build-a-library-to-reference-icons-throughout-your-app-more-conveniently) specifies that we could do something like this:
```javascript
import ReactDOM from 'react-dom';
import fontawesome from '@fortawesome/fontawesome'
import FontAwesomeIcon from '@fortawesome/react-fontawesome'
import brands from '@fortawesome/fontawesome-free-brands'
import faCheckSquare from '@fortawesome/fontawesome-free-solid/faCheckSquare'
import faCoffee from '@fortawesome/fontawesome-free-solid/faCoffee'

fontawesome.library.add(brands, faCheckSquare, faCoffee)
```
to add every icons we need once and then use a simmple string as key for the icon prop, `<FontAwesomeIcon icon="sync" />` for instance. I didn't do it because it would be a bit complicated to make sure we have the right icons loaded depending on the option the user chose for his application. Do you agree ?

### DIFF
--- a/generators/client/templates/react/package.json.ejs
+++ b/generators/client/templates/react/package.json.ejs
@@ -26,6 +26,9 @@ limitations under the License.
     "node_modules"
   ],
   "dependencies": {
+    "@fortawesome/fontawesome": "1.1.5",
+    "@fortawesome/fontawesome-free-solid": "5.0.10",
+    "@fortawesome/react-fontawesome": "0.0.18",
     "availity-reactstrap-validation": "2.0.2",
     "axios": "0.18.0",
     "bootstrap": "4.0.0",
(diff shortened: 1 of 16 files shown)


jdubois:
This looks all good to me, congrats @agaspardcilia !

deepu105:
@agaspardcilia sorry for late reply, was bit busy. I think its good to follow the suggestion and import all icons once. I think we can import all icons that will ve required if all options are enabled, ofcourse it means having few extra icons loaded, but that should be ok

PR #7555, opened 2018-05-02
Adds React icons loader + fix profile reducer

agaspardcilia:
For some reason var names are in camel case while others are in kebab case for `/management/info`. We might need to get more consistent with that.

### DIFF
--- a/generators/client/files-react.js
+++ b/generators/client/files-react.js
@@ -108,7 +108,8 @@ const files = {
                 'config/error-middleware.ts',
                 'config/logger-middleware.ts',
                 'config/notification-middleware.ts',
-                'config/store.ts'
+                'config/store.ts',
+                'config/icon-loader.ts'
             ]
         },
         {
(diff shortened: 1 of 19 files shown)


deepu105:
@agaspardcilia I guess I already discussed with @PierreBesson regarding the case of keys in the endpoint. @PierreBesson can you check what @agaspardcilia is referring to?

agaspardcilia:
@deepu105 I'm talking about `/management/info` that is answering me: 
```json
{
  "display-ribbon-on-profiles" : "dev",
  "activeProfiles" : [ "swagger", "dev" ]
}
```
As you can see, keys are written in two different styles. I don't know if I'm using an outdated version or not but it seems a bit strange.

PierreBesson:
We already discussed it in the MR. It's normal for the name to be activeProfiles, this is actually following our convention. What is not normal is the display-ribbon-on-profile field which is in kebab-case because it would look ugly in the yaml where it is defined.

PR #7566, opened 2018-05-03
Use getMongoClientDatabase() instead of getDatabase()

Falydoor:
This will prevent having a Mongobee exception when the database name is in the property **spring.data.mongodb.uri** (ex: mongodb://localhost:27017/jhipster) and the property **spring.data.mongodb.database** is null.

Looking at the source of [MongoProperties](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java#L169) shows that **getMongoClientDatabase()** will act as **getDatabase()** when the property **spring.data.mongodb.database** is not null.

I understand that this PR doesn't fix anything but it will save time for people that changed properties to use only **spring.data.mongodb.uri**.


jdubois:
Thanks @Falydoor !

PR #7578, opened 2018-05-06
React: typed redux store

bertwin:
These changes will make the redux store typed.

* In the reducers, the state type is infered from the `initialState`
* These are combined into a `RootState` type
* `mapStateToProps` uses the `RootState` to infer its return type `StateProps`
* This return type is combined with the 'typeof' dispatchToProps `DispatchProps` to create Component interfaces
* Where necessary `RouteComponentProps` were added.

This way the correct types have to be declared only once, in the initialState.

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/typings.d.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/typings.d.ts.ejs
@@ -16,6 +16,20 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+import { AxiosPromise } from 'axios';
+
+// TEMPORARY
+declare module 'react-jhipster' {
+  export interface IPayload<T> {
+    type: string;
+    payload: AxiosPromise<T>;
+    meta?: any;
+  }
+  export type ICrudGetAction<T> = (id: string | number) => IPayload<T> | ((dispatch: any) => IPayload<T>);
+  export type ICrudGetAllAction<T> = (page?: number, size?: number, sort?: string) => IPayload<T> | ((dispatch: any) => IPayload<T>);
+  export type ICrudSearchAction<T> = (search?: string) => IPayload<T> | ((dispatch: any) => IPayload<T>);
+}
+
 declare module '*.json' {
     const value: any;
     export default value;
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
@@ -18,7 +18,7 @@
 -%>
 import * as React from 'react';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
+import { Link, RouteComponentProps } from 'react-router-dom';
 import { Button, <% if (haveFieldWithJavadoc) { %>UncontrolledTooltip, <% } %>Row, Col } from 'reactstrap';
 // tslint:disable-next-line:no-unused-variable
 import {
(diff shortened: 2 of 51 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -60,29 +61,7 @@ import { APP_DATE_FORMAT, APP_LOCAL_DATE_FORMAT } from 'app/config/constants';
...
-  <%_ if (pagination !== 'no') { _%>
-  totalItems: 0;
-  location: any;
-  <%_ } _%>
-  <%_ if (pagination === 'pagination' || pagination === 'pager') { _%>
-  history: any;
-  <%_ } _%>
-  match: any;
-}
+export interface I<%= entityReactName %>Props extends StateProps, DispatchProps, RouteComponentProps<{url: string}> {}
```
deepu105: If it works well for autocompletion and refactoring in IDE/editors then this will be great

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
```diff
@@ -186,10 +183,13 @@ export class <%= entityReactName %>Detail extends React.Component<I<%= entityRea
...
 }
 
-const mapStateToProps = ({ <%= entityInstance %> }) => ({
+const mapStateToProps = ({ <%= entityInstance %> }: RootState) => ({
     <%= entityInstance %>: <%= entityInstance %>.entity
 });
 
 const mapDispatchToProps = { getEntity };
 
+type StateProps = ReturnType<typeof mapStateToProps>;
```
deepu105: I didnt know you could do it this way TBH :wink:

generators/client/templates/react/src/main/webapp/app/typings.d.ts.ejs
```diff
@@ -16,6 +16,15 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+
+// TODO: Fix in react-jhipster
+// Reason: getRoles from './user-management.reducer';
```
deepu105: may be we should use a custom type for getRoles?

bertwin: yes, that's probably better than making id optional for every ICrudGetAction

bertwin: it seems `getRoles` doen't do much , it fetches `${apiUrl}/authorities` but nothing is done with the result

jdubois:
As @deepu105 is on holidays, who could do the review @sendilkumarn @wmarques @agaspardcilia @chrisdns ? This looks like a cool feature to have, and as it's spread across the code base it can't stay for too long in a separate branch

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
```diff
@@ -63,8 +63,8 @@ export const ACTION_TYPES = {
 const initialState = {
   loading: false,
   errorMessage: null,
-  entities: [],
-  entity: {},
+  entities: [] as I<%= entityReactName %>[],
+  entity: new <%= entityReactName %>(),
```
bertwin: I'm not so sure about using classes in redux.

In `entity.model.ts.ejs` there currently are both an interface and a class. Is there any situation thinkable where a class would be used, except for setting the default value?

Maybe the class could be replaced with  
```js
// example from travis react-default
export const createDefault = (): IFieldTestEntity => ({
  booleanTom: false,
  booleanRequiredTom: false
});
```

this function could then be used in the initialState

This would also discourage creating methods on the entity model, since that doesn't fit well with react.

deepu105:
Can someone test this out. We need to ensure that IDE/editor (intellij and
vscode) works well with this especially for suggestions and refactoring

jdubois:
@chrisdns @agaspardcilia can you check @deepu105 comment and validate this? This looks awesome and I'd love to merge this

chrisdns:
There is no problem with IDEs but the current branch is still inducing some error type and some unused variables should be removed when using OAUth2 like noticed previously.

bertwin:
@chrisdns the problems with OAuth2 should be solved now.
@deepu105 I removed the type from getRoles

@jdubois I did a rebase to the current master, because the branch was getting far behind

deepu105:
@bertwin great I'm gonna test this tomorrow. There is a CI failure, I restarted it, lets see

deepu105:
No urgency here, it should be ok :)

Thanks & Regards,
Deepu

deepu105:
@bertwin this is really great, much much better than what I had originally. Thank you for this. Expecting a lot  more improvements from you :smile:

jdubois:
Awesome! Thanks @bertwin

PR #7579, opened 2018-05-06
remove duplicated index on email and login

ctamisier:
Having unique constraints on login and email,
`<constraints unique="true" nullable="false" uniqueConstraintName="ux_user_login"/>`
`<constraints unique="true" nullable="true" uniqueConstraintName="ux_user_email"/>`

It acts as an index:

`explain select login from jhi_user where login = 'admin';`
```
                                     QUERY PLAN
-------------------------------------------------------------------------------------
 Index Only Scan using ux_user_login on jhi_user  (cost=0.14..8.16 rows=1 width=118)
   Index Cond: (login = 'admin'::text)
(2 rows)
```

So we don't need those index creation:
```
<createIndex indexName="idx_user_login"
            tableName="<%= jhiTablePrefix %>_user"
            unique="true">
            <column name="login" type="varchar(50)"/>
</createIndex>
<createIndex indexName="idx_user_email"
            tableName="<%= jhiTablePrefix %>_user"
            unique="true">
            <column name="email" type="varchar(100)"/>
</createIndex>
```
It is the case for postgreSQL (and oracle as it is already ignored). I'm not sure about MySQL and MSSQL, it should be the same I guess but I don't have those databases installed yet. 
I'll let you know !

### DIFF
--- a/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
+++ b/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
@@ -101,19 +101,7 @@
             <column name="last_modified_by" type="varchar(50)"/>
             <column name="last_modified_date" type="timestamp"/>
         </createTable>
-    <%_ if (prodDatabaseType !== 'oracle') { _%>
-        <createIndex indexName="idx_user_login"
-            tableName="<%= jhiTablePrefix %>_user"
-            unique="true">
-            <column name="login" type="varchar(50)"/>
-        </createIndex>
 
-        <createIndex indexName="idx_user_email"
-            tableName="<%= jhiTablePrefix %>_user"
-            unique="true">
-            <column name="email" type="varchar(100)"/>
-        </createIndex>
-    <%_ } _%>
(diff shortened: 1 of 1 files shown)


jdubois:
Are you sure this is ignored in Oracle? I remember you need to put indexes on your foreign keys, so I thought you needed them also on constraints.

ctamisier:
Sorry when I say ignored, I meant that was already `if (prodDatabaseType !== 'oracle')` in the existing jhipster code.
Here (for postgresql at least), the unique constraint is an index.

ctamisier:
It is the same for mysql so we don't need the extra create createIndex.

`explain select * from jhi_user where login = 'admin';`
image
(use of ux_user_login)

Should we check about mariaDB and Mssql to remove the 2 createIndex xml entries.

ctamisier:
Same query plan for mariaDB (as mysql).

And as well for MSSQL: https://www.mssqltips.com/sqlservertip/4270/difference-between-sql-server-unique-indexes-and-unique-constraints/

jdubois:
Great work & analysis @ctamisier !!

PR #7584, opened 2018-05-07
Remove useless Moment locales from production build

wmarques:
Moment is quite a big library as it imports ALL the locales by default. We can use [this plugin](https://github.com/iamakulov/moment-locales-webpack-plugin) in order to keep only the locales the user wants to use and remove the rests, which result in a reduced bundle size 🎉 

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -101,6 +101,7 @@
     <%_ if (enableTranslation) { _%>
     "merge-jsons-webpack-plugin": "1.0.14",
     <%_ } _%>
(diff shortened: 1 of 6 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -101,6 +101,7 @@
     <%_ if (enableTranslation) { _%>
     "merge-jsons-webpack-plugin": "1.0.14",
     <%_ } _%>
+    "moment-locales-webpack-plugin": "^1.0.5",
```
deepu105: fix version plz

wmarques:
@deepu105 yep sure ! I do it in a few hours

PR #7595, opened 2018-05-10
 Pay attention to contextPath

asgh:
Maybe I'm misunderstanding this code, but it seems to me it will never work if there is a contextPath set, without these changes.


jdubois:
Just to be sure: are you using this inside an app server, and/or deploying it to something different than the root context?

asgh:
I have `server.contextPath ` set in `application.yml`. So all requests to the gateway (and all its microservices) are prefixed with that path.

PR #7614, opened 2018-05-12
Removes not null restriction from User password field

### LINKED ISSUE #7586: Generated 'User' entity's 'password' field's column is nullable

Generated `User.java` entity's `password` field looks like below:

    @JsonIgnore
    @NotNull
    @Size(min = 60, max = 60)
    @Column(name = "password_hash", length = 60)
    private String password;

Since, it is `@NotNull` at the field level, shouldn't it be `nullable = false` at the column level as well?

    @Column(name = "password_hash", length = 60, nullable = false)

thainan10:
Fix #7586


jdubois:
Oh sorry there is a misunderstanding here: I thought you wanted to have the database field non-nullable, not removing the @NotNull on the Java side (it's the other way round).
I'm closing this, let's discuss again on the ticket.

Issue #6929, opened 2017-12-31
Access token not renewed when using Keycloak


dkefer:
##### **Overview of the issue**

With a freshly generated microservice architecture (with just one microservice in my case) using Keycloak, the microservice responds only with 401 once the access token lifespan is reached (by default 5 minutes). 

No feedback in the jhipster application is provided. Fresh login is necessary to continue working with the entities. It is still possible with the gateway itself (the "Administration" menu if logged in as admin).

##### **Motivation for or Use Case**

I would expect the user to stay logged in as long as he's actively working with the application. 
Also, I think that a user should be given a feedback once the session expires, instead of displaying pages with no content. 

##### **Reproduce the error**

1. Create a microservice application containing some entities, Gateway and choose OIDC as the auth. mechanism for both.
1. Create docker images for both and a docker-compose using ```jhipster docker-compose```
1. Launch docker compose and log-in to the generated application.
1. After 5 minutes, the pages within the "Entities" menu will not contain any content when loaded. If you don't want to wait 5 minutes, set the "Access Token Lifespan" setting in the KeyCloak admin console (Realm settings > Tokens) to a lower value.

In the browser console, it can be seen that the requests to the microservice are responded with 401.
The docker-compose console says:
```
keycloak_1              | 13:59:57,066 WARN  [org.keycloak.events] (default task-15) type=USER_INFO_REQUEST_ERROR, realmId=jhipster, clientId=null, userId=null, ipAddress=172.18.0.9, error=invalid_token, auth_method=validate_access_token
requirement-app_1       | 2017-12-30 13:59:57.076  WARN 6 --- [ XNIO-2 task-13] o.s.b.a.s.o.r.UserInfoTokenServices      : Could not fetch user details: class org.springframework.security.oauth2.client.resource.OAuth2AccessDeniedException, Unable to obtain a new access token for resource 'null'. The provider manager is not configured to support it.
```

##### **Related issues**
https://github.com/jhipster/generator-jhipster/issues/6672

##### **Suggest a Fix**

I guess that a new access token should be fetched once the lifespan is over. 

##### **JHipster Version(s)**
4.13.1

##### **JHipster configuration**

**Microservice application**
<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "org.securityrat.requirementms"
    },
    "jhipsterVersion": "4.13.1",
    "baseName": "requirement",
[...]

PR #7616, opened 2018-05-13
WIP: Add feign configuration for GW

Tcharl:
I added Feign configuration for GW as stated in #7492. I've only one question about this PR (l551 on the files.js).

- [x ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x ] Tests are added where necessary
- [x ] Documentation is added/updated where necessary
- [x ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/generators/server/templates/src/main/resources/config/application.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application.yml.ejs
@@ -71,18 +71,8 @@ zuul: # those values must be configured depending on the application specific ne
     <%_ if (serviceDiscoveryType === 'consul') { _%>
     ignoredServices: 'consul,<%= baseName %>' # Prevent Consul and the "<%= baseName %>" app itself from having a Zuul route automatically created for them
     <%_ } _%>
-
-# See https://github.com/Netflix/Hystrix/wiki/Configuration
-hystrix:
-    command:
-        default:
-            execution:
-                isolation:
-                    thread:
-                        timeoutInMilliseconds: 10000
-
 <%_ } _%>
-<%_ if (applicationType === 'microservice') { _%>
+<%_ if (applicationType === 'microservice' || applicationType === 'gateway') { _%>
 feign:
     hystrix:
         enabled: true
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -531,14 +531,14 @@ const serverFiles = {
     ],
     serverMicroservice: [
         {
-            condition: generator => !(generator.applicationType !== 'microservice' && !(generator.applicationType === 'gateway' && (generator.authenticationType === 'uaa' || generator.authenticationType === 'oauth2'))),
+            condition: generator => (generator.applicationType === 'microservice' || (generator.applicationType === 'gateway' && (generator.authenticationType === 'uaa' || generator.authenticationType === 'oauth2'))),
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 { file: 'package/config/MicroserviceSecurityConfiguration.java', renameTo: generator => `${generator.javaDir}config/SecurityConfiguration.java` }
             ]
         },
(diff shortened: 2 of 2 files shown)


pascalgrimaud:
@PierreBesson : can you have a look on this? It's related to your question https://github.com/jhipster/generator-jhipster/pull/7492#issuecomment-381496359

@Tcharl : you have some lint error, with `yarn test`. Here the logs from Travis:

```
/home/travis/build/jhipster/generator-jhipster/generators/server/files.js
  552:130  error  Expected exception block, space or tab after '/*' in comment  spaced-comment
  552:130  error  Expected space or tab before '*/' in comment                  spaced-comment
  552:229  error  There should be no space before ','                           comma-spacing
```

Tcharl:
Hi @pascalgrimaud the lint error are from the question I asked in the code.
And I noticed that (at least) AuthorizedUserFeignClient does not work as expected from the gateway (401): I'll investigate on how to fix that, but I'm a little bit lost on the token negotiation from now ^^.

generators/server/templates/src/main/resources/config/application.yml.ejs
```diff
@@ -71,18 +71,8 @@ zuul: # those values must be configured depending on the application specific ne
     <%_ if (serviceDiscoveryType === 'consul') { _%>
     ignoredServices: 'consul,<%= baseName %>' # Prevent Consul and the "<%= baseName %>" app itself from having a Zuul route automatically created for them
     <%_ } _%>
-
-# See https://github.com/Netflix/Hystrix/wiki/Configuration
-hystrix:
```
jdubois: I also thought this wasn't working any more - can you confirm?

PierreBesson: Rereading the latest docs for Finchley (and yes they are hard to find as google always redirect to old versions): it seems that indeed there were some changes in recent versions:

```
If Zuul uses service discovery, you need to configure these timeouts with the ribbon.ReadTimeout and ribbon.SocketTimeout Ribbon properties.

If you have configured Zuul routes by specifying URLs, you need to use zuul.host.connect-timeout-millis and zuul.host.socket-timeout-millis.
```
https://cloud.spring.io/spring-cloud-static/Finchley.RC1/single/spring-cloud.html#_zuul_timeouts

So @Tcharl we need to set the `zuul.timeouts` properties for the gateway while keeping the `hystrix.timeouts` for microservices.

Tcharl:
I'm more and more skeptical on this one, that's why I'm holding it:

The more I dig into, the more I think that the UAA looks like a trick and is not 'Enterprise grade': spring sec 5 really encourages full OIDC in order to deal with oauth, and the UAA is not.

What do you think about:
* Either removing the UAA for JHipster, plebicit migration to OIDC (Keycloak) and capitalizing our effort on a fully fledged OIDC integration following the comment on #6929 made by @jgrandja, brainstorming on a good userdetails architecture and feign integration, Etc. For info, it took me two days to migrate a big JH app from UAA to OIDC.
* Either transforming the UAA into a full OIDC backend (and leaving the administration frontend to the gw): IMO it should be less difficult than to maintain variability.

I can finish this one, but I think that adding feature to the UAA may lead to developper frustration using it.

WDYT?

pascalgrimaud:
I suggested removing UAA too here: https://github.com/jhipster/jhipster-registry/pull/262#issuecomment-381345874

It required too much work, there is not enough value compared to OIDC, and so few people want to maintain it. That's why some tickets, some bugs for UAA are on hold.

So I agree to promote OIDC, and remove UAA.

PR #7633, opened 2018-05-15
fix: From Yarn 1.0 scripts don't require "--" for options to be forwarded.

aneri90:
Resolve YARN warn:
`From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.`


wmarques:
@aneri90 Thanks for your PR. I think we still use the `--` because we need to be still compliant with NPM even we encourage our users to use yarn... If I'm not wrong NPM still requires those ugly `--` 😞

aneri90:
Oh snap! You are right.. npm still needs it. 
Thanks for the reply! Closing PR

pascalgrimaud:
I think we can use something like that:
- use a variable at the beginning of the file:
    - for NPM: `-- --`
    - for Yarn: `--`
- then, use this variable instead of all these `-- --`, so it will work for both: NPM and Yarn

sendilkumarn:
@aneri90 go for it, it is kinda ugly to have that warning when running the application with `yarn`

aneri90:
I have changed and rebased my PR to avoid dirty commits. I cannot reopen this PR.. should I open another one?

jdubois:
I don't know why, but yes open up a new PR

PR #7634, opened 2018-05-15
Do not import the whole FontAwesome lib

wmarques:
We should avoid to import the whole fontawesome lib which is quite huge. It's better to use named imports so that webpack can treeshake the lib and include only the icons we use in our app. We gain about 300kb.

Maybe we can add a comment in `vendor.ts` to tell the user how to add icons, or a paragraph in our documentation, WDYT ?


sendilkumarn:
I don't think the failure is related. I restart travis

wmarques:
@sendilkumarn I rebased, let's wait for travis

PR #7636, opened 2018-05-15
[WIP] Use Jest instead of Karma

wmarques:
Jest is a widely used test library made by Facebook. There are many differences with Karma:
* Parallelization (faster)
* Use a virtual DOM instead of a real browser (no need of puppeteer or any browser driver, faster)
* Includes coverage reports (less dependencies)
* Less configuration IMO (just watch the PR..)

The syntax to write tests is the same so it's not a big refactoring and for me the differences are quite important, we gain around 20s on our tests and alot of time in the `yarn install` because we do not need to download Chromium...

Let me know what you think, it's a proposal and still WIP, if this is ok for you guys I'll continue my work to migrate also the React side.


mraible:
❤️

Related: I think we should move to using async/await in Protractor tests.

https://github.com/angular/protractor/blob/master/docs/async-await.md

gmarziou:
When I presented JHipster to Alpes JUG, a user asked me about when JHipster will switch to Jest because their team has experienced such a huge boost in test performance when using it. So perfect timing!

jdubois:
If this gets rid of Puppeteer then it's +1000 from me - @wmarques can you check the conflicts? Let's merge this asap!

jdubois:
I merged it manually - I don't get why the PR wasn't closed automatically ??
So it should all be OK on master!
And of course I tested with some entities and that looked awesome :-)

deepu105:
@jdubois wasnt it work in progress?

jdubois:
No it's fine, I had lunch with William, we work for the same client :-)

sendilkumarn:
@wmarques let me know if you need any help. I will work on this, if you are busy 👍

PR #7645, opened 2018-05-17
Add Memcached support

jdubois:
Memcached can be used as a Spring Cache implementation, but not as an Hibernate L2 cache.
This is great for people running in cloud environments like AWS, GCP, Heroku, as those platforms provide cheap Memcached support, allowing the applications to scale easily without hitting the database too hard.

Ping @jkutner we'll need to add support for the MemCachier add-on with Heroku!

### DIFF
--- a/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
@@ -48,6 +48,15 @@ eureka:
         service-url:
             defaultZone: http://admin:${jhipster.registry.password}@localhost:8761/eureka/
 <%_ } _%>
+<%_ if (cacheProvider === 'memcached') { _%>
+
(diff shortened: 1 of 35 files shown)


generators/server/templates/src/main/resources/config/application-dev.yml.ejs
```diff
@@ -48,6 +48,15 @@ eureka:
...
 <%_ } _%>
+<%_ if (cacheProvider === 'memcached') { _%>
+
+memcached:
+    cache:
+        servers: localhost:11211
+        mode: static
+        expiration: 86400
+        protocol: binary
+<%_ } _%>
```
ruddell: Does this section of the config do anything since memcached is disabled in dev?

jdubois:
Closing as I messed up my rebase :-(

PR #7647, opened 2018-05-17
Try to fix OOM errors with Jest by running tests serially

jdubois:
The OOM errors currently break the Travis builds


jdubois:
Yes @pascalgrimaud as mine "solution" wasn't working :-)

PR #7648, opened 2018-05-17
fix-oom

bertwin:
2 halve memory leaks !== 1 memory leak


pascalgrimaud:
the build is green ! let's try this

bertwin:
@pascalgrimaud maybe it would be better to make the jest workers limited to 2 only when running in a CI environment?

What would be the best way to do that? In the travis script?

jdubois:
@bertwin @pascalgrimaud yes, totally agree! We should only limit this for Travis

pascalgrimaud:
It would be an improvment. Maybe adding `test:ci` in `package.json` ? But we need to be carefull when packaging in prod, as it will use `test`
Need time to think about that :)

PR #7652, opened 2018-05-18
Add Memcached support

jdubois:
Memcached can be used as a Spring Cache implementation, but not as an Hibernate L2 cache.
This is great for people running in cloud environments like AWS, GCP, Heroku, as those platforms provide cheap Memcached support, allowing the applications to scale easily without hitting the database too hard.

Ping @jkutner we'll need to add support for the MemCachier add-on with Heroku!

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/home/home.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/home/home.component.html.ejs
@@ -36,7 +36,7 @@
             </div>
             <%_ if (authenticationType !== 'oauth2') { _%>
             <div class="alert alert-warning" *ngSwitchCase="false">
-                <span jhiTranslate="global.messages.info.register.noaccount">You don't have an account yet?</span>
+                <span jhiTranslate="global.messages.info.register.noaccount">You don't have an account yet?</span>&nbsp;
                 <a class="alert-link" routerLink="register" jhiTranslate="global.messages.info.register.link">Register a new account</a>
             </div>
             <%_ } _%>
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -29,6 +29,7 @@ const DOCKER_CASSANDRA = 'cassandra:3.9';
 const DOCKER_MSSQL = 'microsoft/mssql-server-linux:latest';
 const DOCKER_ORACLE = 'sath89/oracle-12c:latest';
 const DOCKER_HAZELCAST_MANAGEMENT_CENTER = 'hazelcast/management-center:3.9.1';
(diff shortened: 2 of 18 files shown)


jkutner:
@jdubois @jhipster/developers I'm sad to report that there are some serious issues with `com.amazonaws:elasticache-java-cluster-client`, which is used by `memcached-spring-boot-starter`. In short, `elasticache-java-cluster-client` does not work with usernames and passwords.

You can see why this happens in [MemcachedClient](https://github.com/awslabs/aws-elasticache-cluster-client-memcached-for-java/blob/ad77a2a992feaa58eb7bb12d701050d9e20b1f97/src/main/java/net/spy/memcached/MemcachedClient.java#L297). If `authDescriptor` is set, a NPE will always be thrown:

```
Caused by: java.lang.NullPointerException: null
 at net.spy.memcached.MemcachedClient.addObserver(MemcachedClient.java:2991)
 at net.spy.memcached.MemcachedClient.<init>(MemcachedClient.java:297)
 at net.spy.memcached.MemcachedClient.<init>(MemcachedClient.java:235)
 at io.sixhours.memcached.cache.MemcachedCacheManagerFactory.memcachedClient(MemcachedCacheManagerFactory.java:83)
```

In fact, I was only able to get this error by [patching memcached-spring-boot-starter](https://github.com/jkutner/memcached-spring-boot/commit/626bdca96513203e59757a14bbe6294de3bec3f5) to support username and password.

`elasticache-java-cluster-client` is a fork of the very old and very dead [spymemcached](https://code.google.com/archive/p/spymemcached/). But `elasticache-java-cluster-client` also seems very dead.

At this point, adding support on Heroku would require significant work to these libraries. I have to recommend that we switch to this somehow:

```xml
<dependency>
  <groupId>com.googlecode.xmemcached</groupId>
  <artifactId>xmemcached</artifactId>
  <version>2.4.3</version>
</dependency>
```

But I know that may not be possible.

I'l be very happy to add Redis support in the future though!

jkutner:
Here's my [WIP branch](https://github.com/jkutner/generator-jhipster/commit/b9a138003d57d452a17232af1987119d0f1a830c)

jdubois:
Oh @jkutner I should have asked you, you seem to know this far better than I do!!!! And I also realize that very few people seem to use Memcached with Java, let's change this as it's an awesome option!

- Yes let's switch to "xmemcached", the project looks much more popular. Can you do a PR? Do you want me to work on this?
- Once this is done, Redis support should be very easy, yes a PR is highly welcome

I'm creating 2 tickets for those and will assign them to you, if that's OK for you

PR #7658, opened 2018-05-19
fix metrics page in kubernetes

sendilkumarn:
In Kubernetes the metrics page is not working, because of the `jvm.garbage.PS-MarkSweep.count` name difference It is actually `jvm.garbage.MarkSweep.count`


generators/client/templates/react/src/main/webapp/app/modules/administration/metrics/metrics.tsx.ejs
```diff
@@ -94,6 +94,17 @@ export class MetricsPage extends React.Component<any, IMetricsPageState> {
...
 
+  gaugeRow = (metrics, label: String, key) => (
+    metrics.gauges[key] ?
+      <Row>
+        <Col md="9">{label}</Col>
+        <Col md="3" className="text-right">
+          {metrics.gauges[key].value}
+        </Col>
+      </Row>
+      : ''
```
deepu105: use null instead of empty string for empty node

Issue #7659, opened 2018-05-19
CamelCase bug in entity update react script


harry-g:
##### **Overview of the issue**
Using an entity with CamelCase name, many-to-many relations are generating wrong all-lowercase name at the last part of the entity updates script - I think only when using react.

##### **Motivation for or Use Case**
Running the generated app will fail because of wrong entity name.

##### **Reproduce the error**
1. run `jhipster` to Create a jHi app with default options, but chose `react` as client!
2. import e.g. the following model with `jhipster import-jdl src\model\Model.jdl`
```
entity MyEntityA {
	name String required
}

entity MyEntityB {
	name String required
}

relationship ManyToMany {
	MyEntityA{entityB(name)} to MyEntityB{name}
}
```
3. You will get an error:
```
ERROR in C:/Development/source/JHiTest/src/main/webapp/app/entities/my-entity-a/my-entity-a-update.tsx
(157,25): Cannot find name 'myentitybs'. Did you mean 'myEntityBS'?
```
check `src\main\webapp\app\entities\my-entity-a\my-entity-a-update.tsx`
Close to the end of the file (in my case line 157) you will find one map access with `myentitybs` (all lowercased) instead of `myEntityBS`.

Fix it manually and everything will work fine.

##### **Related issues**
Found none.

##### **Suggest a Fix**
Remove toLowerCase() in entity-update template
Will create a PR.

##### **JHipster Version(s)**

v5.0.0-beta-1

##### **JHipster configuration**
Using JHipster version installed globally
Executing jhipster:info
Options:
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
j-hi-test@0.0.0 C:\Development\source\JHiTest
`-- (empty)

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "de.harryg"
    },
    "jhipsterVersion": "5.0.0-beta.1",
    "baseName": "JHiTest",
    "packageName": "de.harryg",
    "packageFolder": "de/harryg",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
[...]

PR #7660, opened 2018-05-19
remove toLowerCase()

harry-g:
This will fail whenever the entity name contains uppercase chars after the first char.
Therefore removed toLowercase().
FIX https://github.com/jhipster/generator-jhipster/issues/7659

Hope this trivial change is OK w/o tests, docs and son on...


deepu105:
@chrisdns could you please check this?

chrisdns:
It's ok ! :+1:

PR #7687, opened 2018-05-24
Add Yarn to engines

jkutner:
This will ensure that Yarn is installed (with the correct version) on platforms like Heroku.

Normally, Heroku looks for a `yarn.lock` file, and installs and runs `yarn` accordingly. But in some cases (either with apps generated from https://start.jhipster.tech or `jhipster --skip-install`) no `yarn.lock` is present. But Heroku should still work.

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -138,8 +138,8 @@
     "write-file-webpack-plugin": "4.2.0"
   },
   "engines": {
-    "node": ">=8.9.0",
-    "yarn": ">=1.3.2"
+    "node": ">=8.9.0"<% if (clientPackageManager === 'yarn') { %>,
+    "yarn": ">=1.3.2"<% } %>
   },
   <%_ if(!skipCommitHook) { _%>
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -138,7 +138,8 @@
     "write-file-webpack-plugin": "4.2.0"
   },
   "engines": {
-    "node": ">=8.9.0"
+    "node": ">=8.9.0",
+    "yarn": ">=1.3.2"
```
pascalgrimaud: As the user can choose `Yarn` or `NPM` when generating the project, I think we should add this condition:

```
"node": ">=8.9.0"<% if (clientPackageManager === 'yarn') { %>,
"yarn": ">=1.3.2"<% } %>
```

Otherwise, `Yarn` would be mandatory ?

jkutner: oh i thought yarn was mandatory. yea, i'll fix this

jkutner:
@pascalgrimaud thanks

Issue #7337, opened 2018-03-21
Feature Request - Add options to generate Istio-friendly deployments MVP


saturnism:
##### **Overview of the issue**

Add a generator for Istio deployment

##### **Motivation for or Use Case**

With Istio deployment, we can potentially simplify the microservices and removing several components, and relying on Istio for Load Balancing and Circuit Breaking.

##### **Suggest a Fix**

- [x] Microservices/Gateway Generator
  - [x] Generate w/o Eureka, Hystrix, Zuul (#7695)
  - [x] Generate w/ `server.servlet.context-path` for easier Ingress mapping #7695)
- [x] Kubernetes Generator
  - [x] Generate Ingress mapping (#7695)
  - [x] Use Istio Ingress Class (#7697)
  - [x] Propagate Istio Headers (mostly done through Spring Cloud Sleuth 2) (#7697)
  - [x] Generate Istio Destination Rules for Circuit Breaker (#7697)
  - [x] If using Eureka/Consul, use host name rather than IP (#7697)
  - [x] Prefix service port names (e.g., `http`) for Istio (#7697)
- [x] Documentation Update (In Progress [Done - JHipster Doc 577](https://github.com/jhipster/jhipster.github.io/pull/577)]

After this initial MVP, see #7708

stephanj:
Would be awesome! More than happy to beta test. 🤓

jdubois:
@saturnism wouldn't this remove the need for large part of Spring Cloud? No more need for load balancing/circuit breakers, and maybe no gateway (not sure, as it's like with Traefik, our gateway does much more than just routing, so maybe you want the option to use both).
-> can you confirm this? BTW this is how it works with Heroku - the platform does load balancing/circuit breaking, making a big part of Spring Cloud useless. So this would be some kind of "common use case" for advanced platforms.

saturnism:
I guess it would remove some of Spring Cloud, but probably can't say all.  E.g.,
- Can keep Spring Cloud Sleuth for trace propagation. There are a lot of good integrations there. See [Istio By Example for a Span Extractor](https://github.com/saturnism/istio-by-example-java/tree/master/spring-boot-example/spring-istio-support/src/main/java/com/example/istio) that works w/ Spring Cloud.
- Still need bulk heading, i.e., ability to return default value when the service is ultimately not working

Probably don't need:
- Client Side Load Balancer, Ribbon
- Eureka

(But you are right - App Engine would also benefit from the common use case :)

[...]

PR #7695, opened 2018-05-28
WIP - #7337 - Generate microservices mapping for Kubernetes ingress

saturnism:
- Please make sure the below checklist is followed for Pull Requests.

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [ ] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #7697, opened 2018-05-28
Istio - intial commit

srinivasa-vasu:
#7337 
- tested with both eureka and consul
- included routerule and destinationpolicy rules
This is with minimal change to the existing code base. We can slowly introduce more features to it and over the period of time redundant components can be decoupled.

### DIFF
--- a/generators/kubernetes/prompts.js
+++ b/generators/kubernetes/prompts.js
@@ -22,7 +22,9 @@ const dockerPrompts = require('../docker-prompts');
 module.exports = _.extend({
     askForKubernetesNamespace,
     askForKubernetesServiceType,
-    askForIngressDomain
+    askForIngressDomain,
+    askForIstioSupport,
+    askForIstioRouteFiles
 }, dockerPrompts);
 
 function askForKubernetesNamespace() {
@@ -99,6 +101,67 @@ function askForIngressDomain() {
 
     this.prompt(prompts).then((props) => {
         this.ingressDomain = props.ingressDomain;
+        done();
+    });
+}
+
+function askForIstioSupport() {
+    if (this.composeApplicationType === 'monolith') {
+        this.istioSupportOpt = 'no';
+        return;
+    }
+    const done = this.async();
+
+    const prompts = [{
+        type: 'list',
+        name: 'istioSupportOpt',
+        message: 'Choose your preference for Istio Service Mesh',
+        choices: [
+            {
+                value: 'no',
+                name: 'Not required'
+            },
+            {
+                value: 'mi',
+                name: 'Manual sidecar injection (ensure istioctl in $PATH)'
+            },
+            {
+                value: 'ai',
+                name: 'Label namespace as automatic injection is configured already'
+            }
+        ],
+        default: this.istioSupportOpt ? this.istioSupportOpt : 'no'
+    }];
+
+    this.prompt(prompts).then((props) => {
+        this.istioSupportOpt = props.istioSupportOpt;
+        done();
+    });
+}
+
+function askForIstioRouteFiles() {
+    if (this.istioSupportOpt === 'no') return;
+    const done = this.async();
+
+    const prompts = [{
+        type: 'list',
+        name: 'istioRouteObjs',
+        message: 'Do you want to generate Istio route files?',
+        choices: [
+            {
+                value: 'no',
+                name: 'No'
+            },
+            {
+                value: 'yes',
+                name: 'Yes'
+            }
+        ],
+        default: this.istioRouteObjs ? this.istioRouteObjs : 'no'
+    }];
+
+    this.prompt(prompts).then((props) => {
+        this.istioRouteObjs = props.istioRouteObjs;
         done();
     });
 }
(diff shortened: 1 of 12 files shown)


generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -76,12 +81,26 @@ spec:
...
         - name: SPRING_CLOUD_CONSUL_PORT
           value: "8500"
+          <%_ if (app.istioSupportOpt !== 'no') { _%>
+        - name: SPRING_CLOUD_CONSUL_DISCOVERY_PREFER_IP_ADDRESS
+          value: "false"
+        - name: SPRING_CLOUD_CONSUL_DISCOVERY_HOSTNAME
+          value: <%= app.baseName.toLowerCase() %>
+        - name: SPRING_CLOUD_CONSUL_DISCOVERY_SERVICE_NAME
+          value: <%= app.baseName.toLowerCase() %>
+          <%_ } _%>
```
saturnism: can we add one more conditional env var to propagate additional Istio headers? e.g.:

```
- name: SPRING_SLEUTH_PROPAGATION_KEYS
   value: "x-request-id,x-ot-span-context"
```

srinivasa-vasu: Sure, I guess sleuth is not available as a dependency. Have to add it to the build file first.

PierreBesson: Sleuth is available when packaging the app with the zipkin maven profile.

generators/kubernetes/prompts.js
```diff
@@ -102,3 +104,64 @@ function askForIngressDomain() {
...
+    if (this.composeApplicationType === 'monolith') {
+        this.istioSupportOpt = 'no';
+        return;
+    }
+    const done = this.async();
+
+    const prompts = [{
+        type: 'list',
+        name: 'istioSupportOpt',
+        message: 'Choose your preference for Istio Service Mesh',
```
PierreBesson: I would use the following:
- name: `istio`
- message: `Do you want to setup Istio ?`
- value: 'manualInjection`or value: `automaticInjection`

generators/kubernetes/prompts.js
```diff
@@ -102,3 +104,64 @@ function askForIngressDomain() {
...
+
+function askForIstioRouteFiles() {
+    if (this.istioSupportOpt === 'no') return;
+    const done = this.async();
+
+    const prompts = [{
+        type: 'list',
+        name: 'istioRouteObjs',
+        message: 'Do you want to generate Istio route files?',
+        choices: [
```
PierreBesson: name: `istioRoute`
value: boolean

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -16,17 +16,22 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
PierreBesson: 👍

generators/kubernetes/templates/deployment.yml.ejs
srinivasa-vasu: @saturnism done!

generators/kubernetes/prompts.js
srinivasa-vasu: done

generators/kubernetes/templates/deployment.yml.ejs
saturnism: nice!

saturnism:
i think this is a good first step! i'll add additional checklist items that we can address later.

deepu105:
builds failed due to some network/yarn issue. I restarted them

saturnism:
@srinivasa-vasu one last thing, if istio support is enabled, then if when generating ingress, it should use istio ingress class w/ annotation:

```
  kubernetes.io/ingress.class: istio
```

saturnism:
#7695 is also ready for review, but will need all the changes made in this PR for it to work properly.

PierreBesson:
@srinivasa-vasu For the options you added as manual/auto sidecar injection, as this only impact the apply.sh script and it's potentially subject to change in the lifetime of the project so I'm thinking that maybe it could be a script parameter like this: `./apply.sh --sidecar-injection=manual/auto` and if none was given then output a message to the user.

saturnism:
can we merge the MVP first and add enhancements via https://github.com/jhipster/generator-jhipster/issues/7708 ?   Since not everyone will be using apply script either.

I feel we can track it separately, where it actually generates the YAML w/ `istio kube-inject` on the yaml file directly. The YAML with the injection is the final output.

PierreBesson:
I agree let's merge this quickly with a BETA tag. It's better to release early but I don't want to add useless questions especially because we have to provide configuration fallbacks when changing questions for people that regenerate from the same yo-rc.

saturnism:
Ah I see! Fair point because the answer is stored.

jdubois:
I just restarted the Travis build, I hope it will pass - yes let's merge this as soon as possible

saturnism:
exciting! :D

PR #7702, opened 2018-05-29
Bump dependencies

erikkemperman:

Cf., this depends on [this server lib PR](https://github.com/jhipster/jhipster/pull/70).

I'll cancel Travis: it will fail because of mutually dependent changes, see here for passing tests where both generator and server-lib PRs are applied:
https://travis-ci.org/erikkemperman/generator-jhipster/builds/385069631
https://travis-ci.org/erikkemperman/jhipster/builds/385070081

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -88,7 +88,7 @@
         <profile.swagger />
 
         <!-- Dependency versions -->
-        <jhipster-dependencies.version>2.0.3</jhipster-dependencies.version>
+        <jhipster-dependencies.version>2.0.5-SNAPSHOT</jhipster-dependencies.version>
         <!-- The spring-boot version should match the one managed by
         https://mvnrepository.com/artifact/io.github.jhipster/jhipster-dependencies/${jhipster-dependencies.version} -->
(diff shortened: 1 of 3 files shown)


generators/server/templates/pom.xml.ejs
```diff
@@ -88,7 +88,7 @@
         <profile.swagger />
 
         <!-- Dependency versions -->
-        <jhipster-dependencies.version>2.0.3</jhipster-dependencies.version>
+        <jhipster-dependencies.version>2.0.5-SNAPSHOT</jhipster-dependencies.version>
```
jdubois: I don't think those can work, our snapshot repo on Bintray isn't configured. I'll first do a 2.0.5 release of jhispter-framework etc, and then update this.

erikkemperman: Sorry, I wasn't sure if I remembered correctly -- but didn't we eventually figure out that we can publish snapshots via sonatype, previously?

jdubois: Yes but for this we need to have the bintray repository configured in the final pom.xml - and I don't think this is the case anymore
Not a big problem: I'm testing locally then pushing once it builds

PR #7713, opened 2018-05-31
Fixed application generation (partially)

MathieuAA:

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -130,6 +130,7 @@ module.exports = class extends BaseGenerator {
             defaults: false
         });
 
+        this.availableGeneratorConfig = this.options.availableGeneratorConfig;
         this.skipClient = this.configOptions.skipClient = this.options['skip-client'] || this.config.get('skipClient');
         this.skipServer = this.configOptions.skipServer = this.options['skip-server'] || this.config.get('skipServer');
         this.skipUserManagement = this.configOptions.skipUserManagement = this.options['skip-user-management'] || this.config.get('skipUserManagement');
@@ -189,6 +190,9 @@ module.exports = class extends BaseGenerator {
             },
 
             setupconsts() {
+                if (this.availableGeneratorConfig) {
+                    this.config = this.availableGeneratorConfig;
+                }
                 this.applicationType = this.config.get('applicationType');
                 if (!this.applicationType) {
                     this.applicationType = 'monolith';
@@ -273,7 +277,8 @@ module.exports = class extends BaseGenerator {
                     'client-hook': !this.skipClient,
                     configOptions: this.configOptions,
                     force: this.options.force,
-                    debug: this.isDebugEnabled
+                    debug: this.isDebugEnabled,
+                    availableGeneratorConfig: this.availableGeneratorConfig
                 });
             },
 
@@ -285,7 +290,8 @@ module.exports = class extends BaseGenerator {
                     'skip-commit-hook': this.options['skip-commit-hook'],
                     configOptions: this.configOptions,
                     force: this.options.force,
-                    debug: this.isDebugEnabled
+                    debug: this.isDebugEnabled,
+                    availableGeneratorConfig: this.availableGeneratorConfig
                 });
             },
 
@@ -357,7 +363,8 @@ module.exports = class extends BaseGenerator {
                             'skip-install': true,
                             force: this.options.force,
                             debug: this.isDebugEnabled,
-                            arguments: [entity.name]
+                            arguments: [entity.name],
+                            availableGeneratorConfig: this.config
                         });
                     });
                 }
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -130,7 +130,6 @@ module.exports = class extends BaseGenerator {
             defaults: false
         });
 
-        this.availableGeneratorConfig = this.options.availableGeneratorConfig;
         this.skipClient = this.configOptions.skipClient = this.options['skip-client'] || this.config.get('skipClient');
         this.skipServer = this.configOptions.skipServer = this.options['skip-server'] || this.config.get('skipServer');
         this.skipUserManagement = this.configOptions.skipUserManagement = this.options['skip-user-management'] || this.config.get('skipUserManagement');
@@ -190,9 +189,6 @@ module.exports = class extends BaseGenerator {
             },
 
             setupconsts() {
-                if (this.availableGeneratorConfig) {
-                    this.config = this.availableGeneratorConfig;
-                }
                 this.applicationType = this.config.get('applicationType');
                 if (!this.applicationType) {
                     this.applicationType = 'monolith';
@@ -277,8 +273,7 @@ module.exports = class extends BaseGenerator {
                     'client-hook': !this.skipClient,
                     configOptions: this.configOptions,
                     force: this.options.force,
-                    debug: this.isDebugEnabled,
-                    availableGeneratorConfig: this.availableGeneratorConfig
+                    debug: this.isDebugEnabled
                 });
             },
 
@@ -290,8 +285,7 @@ module.exports = class extends BaseGenerator {
                     'skip-commit-hook': this.options['skip-commit-hook'],
                     configOptions: this.configOptions,
                     force: this.options.force,
-                    debug: this.isDebugEnabled,
-                    availableGeneratorConfig: this.availableGeneratorConfig
+                    debug: this.isDebugEnabled
                 });
             },
 
@@ -363,8 +357,7 @@ module.exports = class extends BaseGenerator {
                             'skip-install': true,
                             force: this.options.force,
                             debug: this.isDebugEnabled,
-                            arguments: [entity.name],
-                            availableGeneratorConfig: this.config
+                            arguments: [entity.name]
                         });
                     });
                 }
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -109,6 +109,8 @@ module.exports = class extends BaseGenerator {
         });
 
         this.context = {};
+        this.availableGeneratorConfig = this.options.availableGeneratorConfig;
+
         this.setupEntityOptions(this, this, this.context);
         this.registerClientTransforms();
         const blueprint = this.config.get('blueprint');
@@ -128,6 +130,12 @@ module.exports = class extends BaseGenerator {
         if (useBlueprint) return;
         return {
             getConfig() {
+                if (this.availableGeneratorConfig) {
+                    this.config = this.availableGeneratorConfig;
+                    this.config.getAll = () => this.config;
+                    this.config.get = key => this.config[key];
+                    this.config.set = (key, value) => { this.config[key] = value; };
+                }
                 const context = this.context;
                 context.useConfigurationFile = false;
                 this.env.options.appPath = this.config.get('appPath') || constants.CLIENT_MAIN_SRC_DIR;
@@ -563,11 +571,7 @@ module.exports = class extends BaseGenerator {
                             field.fieldValidateRulesPattern.replace(/\\/g, '\\\\').replace(/"/g, '\\"') : field.fieldValidateRulesPattern;
                     }
 
-                    if (_.isArray(field.fieldValidateRules) && field.fieldValidateRules.length >= 1) {
-                        field.fieldValidate = true;
-                    } else {
-                        field.fieldValidate = false;
-                    }
+                    field.fieldValidate = _.isArray(field.fieldValidateRules) && field.fieldValidateRules.length >= 1;
 
                     if (fieldType === 'ZonedDateTime') {
                         context.fieldsContainZonedDateTime = true;
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -109,7 +109,6 @@ module.exports = class extends BaseGenerator {
         });
 
         this.context = {};
-        this.availableGeneratorConfig = this.options.availableGeneratorConfig;
 
         this.setupEntityOptions(this, this, this.context);
         this.registerClientTransforms();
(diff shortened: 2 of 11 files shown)


generators/import-jdl/index.js
```diff
@@ -100,50 +103,22 @@ module.exports = class extends BaseGenerator {
 
             parseJDL() {
                 this.log('The JDL is being parsed.');
+                const jdlImporter = new jhiCore.JDLImporter(this.jdlFiles, {
+                    databaseType: this.prodDatabaseType,
```
deepu105: for app generation these wont be present is that fine?

generators/import-jdl/index.js
```diff
@@ -154,93 +129,85 @@ module.exports = class extends BaseGenerator {
...
+                        generateApplicationFiles(
+                            this,
+                            this.importState.exportedApplications[0]
+                        );
+                    } catch (error) {
+                        this.error(`While generating applications from the parsed JDL\n${error}`);
+                    }
+                } else {
+                    // sub-folder generation, not yet handled
+                    this.applicationsLeftToGenerate = this.importState.exportedApplications
```
deepu105: what is this usecase?

MathieuAA: You could say it's temporary. Right now, the standard way of importing works. This is just to warn the user that generating entities/apps from one folder to another is not yet supported (I don't know if it's possible)

generators/import-jdl/index.js
MathieuAA: Why?

MathieuAA: You mean the variables?

MathieuAA: Oh I get it. It's perfectly fine. The importer will handle it just fine.

MathieuAA:
My chief concern here is the composability of the import-jdl subgen with the app & entity subgens.
If you try to import one application & _n_ entities, it will fail because:
  - I think I don't call the app subgen the right way
  - The app & entity subgens are concurrent when composed... and we actually want the app subgen to be the one executed (launched, processed, finished) before the other

generators/import-jdl/index.js
deepu105: lets not support that at all, thats why I wanted to remove the path option in new JDL as well, as this could be easily abused and adds a lot of edge cases and possible bugs, lets keep it simple. Also it would be an issue in JHipster online if we support generating apps in other folders, our users are already familiar with the restrictions of yeoman and hence lets just stick to that

deepu105:
@MathieuAA I might be able to help with composing, I have done my share of experiments with it for blueprints and stuff(and I added the original compose behaviours in the generators) so next week i'll spent some time to fix the blueprint issues, so i'll also try this out

generators/import-jdl/index.js
MathieuAA: Oh wait. I may have just understood the comment. I strongly disagree. Generating multiple apps is a must-have. Just so that we are on the same page here
  - there is no app path anymore
  - the chosen way to generate more than one app (instead of having a ton of JDL files) is to create folders so that any subfolder will be at `./<SUBFOLDER_i>` (the generator already handles not having paths as name)
  - a security would be not to generate anything if a baseName is an existing file (one can easily check this with the path module)

If we don't call the subgens in any of the subfolders, that's completely fine by me (that's why I wanted to inform the users of the remaining apps & entities to generate). But not supporting creating these subfolders with `.yo-rc.json` and `.jhipster` files in them would be a mistake in my opinion.

jdubois: @MathieuAA I understand, but for JHipster Online we need to re-think the whole system, and that's just not possible now

deepu105: ok, may be we were talking about different things here. Of course we have to support generating multiple apps form the same JDL, but then the only real use case for that is microservices and I dont see why we need to support that for a monolith, as IMO for monolith it would be a bad practice to have 2 apps on the same JDL.

Good to know that the path option is removed that means we have control of deciding where the path would be so I imagine the below use cases.

1. Entites in a JDL
  - same behavior as of today
2. single monolith app in a JDL with or without entities
  - `mkdir myApp && cd myApp`
  - Execute `jhipster import-jdl app.jh` and we create the below inside myApp folder
  - -- .yo-rc.json
    -- .jhipster
  - We call the app generator by specifying --with-entities flag
  - app is created in the `myApp` folder

3. Multiple microservice in a JDL with entities (say gateway, ms1, ms2)
  - `mkdir myMsApp && cd myMsApp`
  - Execute `jhipster import-jdl app.jh` and we create the below inside myApp folder
  - -- gateway
        -- .yo-rc.json
        -- .jhipster
    -- ms1
        -- .yo-rc.json
        -- .jhipster
    -- ms2
        -- .yo-rc.json
        -- .jhipster
  - We call the app generator by specifying --with-entities flag inside each sub folder, i.e `gateway, ms1 and ms2`
  - app is created in the sub folders accordingly

Now for case 3 the way yeoman works might be a challenge, but we have 2 options for that
1. Since we use our own CLI, we could cd into the folders we want and then execute the generators in the folders. This means import-jdl will not be a sub generator anymore and will be a CLI feature instead (to end user it doesnt matter as they will use it in the sameway, but a sideeffect would be that people wont be able to do `yo jhipster:import-jdl` anymore
2. Do some hack to cd into the folders before composing the app generators and cd out once done

MathieuAA: The first option looks good, but it implies having to work a bit to achieve that.
Let's not bother ourselves about how we generate these multiple apps. Anyone can temporarily `cd` inside the folders and do a `yo jhipster`. I'd rather have the app composition working.
I can, in the meantime, issue a warning that a JDL file contains multiple apps (to inform it's a bad practise).

generators/import-jdl/index.js
```diff
@@ -154,93 +129,85 @@ module.exports = class extends BaseGenerator {
...
-    });
-    return jdlApplications;
+function shouldGenerateApplications(generator) {
+    return !generator.options['ignore-application'] && generator.importState.exportedApplications.length !== 0;
 }
 
 function generateApplicationFiles(generator, application) {
-    const initialWorkingDir = shelljs.pwd();
-    shelljs.cd(path.join(shelljs.pwd().stdout, application.path || ''));
     callSubGenerator(generator, '..', 'app', {
```
MathieuAA: Logically, that's how a subgen should be called (if I'm not wrong). And after looking at the code for the app subgen, I can't understand why it doesn't pick up the .yo-rc.json file. The creation of the files takes place before the subgens are called (if I understand Yeoman precedence rules).

jdubois:
I restarted the build and it passes - should I merge this, so we got it in our upcoming beta release? @MathieuAA are you confident about it?

MathieuAA:
The entity generation works, so there's no regression. The app generation may fail, and I don't want that, even for a Beta. Do you want to release today?

jdubois:
I want to release "soon", but I'm not going to merge the PR if app generation "may fail" - we can still merge later, there is no hurry

MathieuAA:
If soon means before tomorrow, I may be able to make it.

jdubois:
OK, keep me updated, I might have time on Sunday evening for the release

MathieuAA:
Roger that!

MathieuAA:
I've finally found out why it didn't work! Turns out I have to reload the config, which isn't possible for some reasons with `this.config.getAll` (it may not do any I/O, which may be an issue, gotta check). I think I get it working before your beta release Julien.

jdubois:
OK this all look correct - @MathieuAA is it ready to merge? I'll have time today & tomorrow

MathieuAA:
@jdubois except the tests that fail. I'll be honest, I'm not happy with my solution. I'd be ready to merge, as it's a beta --improvable later on, but I'd rather clean up the tests if we keep this _ugly_ solution.

cc @deepu105 knowing you have a lot more experience with Yeoman that I do, do you have an input on this?

generators/import-jdl/index.js
```diff
@@ -154,42 +143,57 @@ module.exports = class extends BaseGenerator {
                     }
                     this.error(`Error while parsing applications and entities from the JDL ${error}`);
                 }
+            }
+        };
+    }
+
+    get install() {
```
deepu105: @MathieuAA not sure about doing the application generation in the install phase. I think it might be better to do the generation in the default phase. Any specific reason for this?
The install phase will queue up along with the install phase of all the sub gen that are composed from this as well and might cause race conditions

generators/import-jdl/index.js
```diff
@@ -201,63 +205,69 @@ module.exports = class extends BaseGenerator {
...
+    }
+    if (application['generator-jhipster'].skipUserManagement) {
+        args.push('--skip-user-management');
+    }
+    if (application['generator-jhipster'].jhiPrefix) {
+        args.push('--jhi-prefix');
+        args.push(application['generator-jhipster'].jhiPrefix);
+    }
+    const done = generator.async();
+    generator.spawnCommand('yo', args).on('close', () => {
```
deepu105: I dont like this part TBH, but its fine for now, i'll rework this when I find time, as its quite tricky to get composing working correctly here

generators/import-jdl/index.js
MathieuAA: According to the doc, being in the install scope is recommended when using the spawnCommand

generators/import-jdl/index.js
MathieuAA: I'm still trying to get it working when I have the time, but the main issue is passing the context from one generator to the other when composing. I thought it was automatic when composing generators, but I can't get it working.

MathieuAA:
@deepu105 I've managed to make it work without the hack.
I'll update the PR after I make the tests pass.

MathieuAA:
@jdubois Now you can merge it :)

jdubois:
Sorry @MathieuAA but the build badly fails :-(

MathieuAA:
And I can't reproduce it right now

generators/app/index.js
```diff
@@ -130,6 +130,7 @@ module.exports = class extends BaseGenerator {
             defaults: false
         });
 
+        this.availableGeneratorConfig = this.options.availableGeneratorConfig;
```
deepu105: @MathieuAA I think this might be duplication of the `configOptions` param that is passed around, can you check?
The purpose of `configOptions` was to pass context around when composing so please do see if you are able to use it without introducing a new param

generators/entity/index.js
```diff
@@ -128,6 +130,12 @@ module.exports = class extends BaseGenerator {
         if (useBlueprint) return;
         return {
             getConfig() {
+                if (this.availableGeneratorConfig) {
+                    this.config = this.availableGeneratorConfig;
+                    this.config.getAll = () => this.config;
```
deepu105: why are these needed?

generators/import-jdl/index.js
```diff
@@ -71,80 +77,69 @@ module.exports = class extends BaseGenerator {
...
-                this.skipClient = this.config.get('skipClient');
-                this.clientFramework = this.config.get('clientFramework');
+                if (!jhiCore.FileUtils.doesFileExist('.yo-rc.json') && !this.generated) {
+                    // we may have to parse the JDL first
+                    return;
+                }
+                let configuration = this.config.getAll();
+                if (!configuration.baseName && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
+                    configuration = JSON.parse(fs.readFileSync('.yo-rc.json', { encoding: 'utf-8' }))['generator-jhipster'];
+                    configuration.getAll = () => configuration;
```
deepu105: if such an update is needed, can we do it in a util method in generator-base and reuse it where needed instead of repeating this?

generators/app/index.js
MathieuAA: roger that, thanks for the tip

generators/entity/index.js
MathieuAA: to replace the seemingly absent config passing from one generator to another, but with your previous tip, I might get it to work without it

generators/import-jdl/index.js
MathieuAA: yes

jdubois:
@MathieuAA if you are near Ippon we can have lunch together today and go through all this

MathieuAA:
I'm 15mins from Ippon, I believe we can

pascalgrimaud:
@MathieuAA : just tested on my machine, I could reproduce the same issue on Travis.
I'm trying to fix it.

MathieuAA:
@pascalgrimaud You rock

generators/entity/index.js
pascalgrimaud: @MathieuAA : these 5 lines made `this.config` wrong. If I deleted these 5 lines, it seems to work

pascalgrimaud:
@MathieuAA : my use case to reproduce the issue:
- copy an existing `.yo-rc.json` into a new folder
- copy some entities inside `.jhipster/` folder, for example Operation, BankAccount and Label
- launch `jhipster --with-entities`

I wonder if the new variable you introduced `this.availableGeneratorConfig` is really usefull ?

MathieuAA:
@pascalgrimaud Not if we replace it by `this.configOptions`, which seems to be the convention for this use case.

pascalgrimaud:
Here the commit: https://github.com/pascalgrimaud/generator-jhipster/commit/89df4207e437b0ec754fd00eb21ca30388b42e05 -> can you check if it's ok for you ? If yes, you can take the commit, or I can do a PR against your branch

Here the current Travis builds: https://travis-ci.org/pascalgrimaud/generator-jhipster

But to be honest, I don't really understand what I'm doing :-D

MathieuAA:
If there's a `--with-entities`, that's one more use case, as I won't have to call the entity subgen (the app subgen will)

MathieuAA:
And the app generation + entity generation work?

pascalgrimaud:
I only tested the default use case in Travis.
I didn't test the new feature app+entity in same file.

MathieuAA:
I'm going to merge this and test later. Thanks a lot Pascal :)

MathieuAA:
I'm going to merge this and test later. Thanks a lot Pascal :)

jdubois:
Merged! Let's do this :-)

PR #7725, opened 2018-06-02
Fix #7723 - Change Angular resolver behavior

### LINKED ISSUE #7723: Change the behavior of Angular entity resolver

As of now, the resolver used in angular router cannot be strongly typed because it returns either a HTTPResponse or a new Entity.
Then, the update component and the detail component gets either the response.body or the element on their onInit() method.

I think we can improve it easily.

Instead of having this code:
```
@Injectable()
export class XDomainResolve implements Resolve<any> {
    constructor(private service: XService) {}

    resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
        const id = route.params['id'] ? route.params['id'] : null;
        if (id) {
            return this.service.find(id);
        }
        return new X();
    }
}
```

We could have this:
```
@Injectable()
export class XDomainResolve implements Resolve<X> {
    constructor(private service: XService) {}

    resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
        const id = route.params['id'] ? route.params['id'] : null;
        if (id) {
            return this.service.find(id).map((resp: HttpResponse<X>) => resp.body;
        }
        return new X();
    }
}
```

Strongly type resolver, avoid dealing with either the response or the entity.

Will do one in a minute!

Tcharl:

### DIFF
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management.route.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management.route.ts.ejs
@@ -23,6 +23,7 @@ import { Resolve, ActivatedRouteSnapshot, RouterStateSnapshot, Routes } from '@a
 import { JhiPaginationUtil, JhiResolvePagingParams } from 'ng-jhipster';
 <%_ } _%>
 import { UserRouteAccessService } from 'app/core';
+import { Observable } from 'rxjs';
 import { <%= entityAngularName %> } from 'app/shared/model/<%= entityModelFileName %>.model';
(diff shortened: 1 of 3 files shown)


generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management.route.ts.ejs
```diff
@@ -35,19 +35,21 @@ import {
     <%= entityAngularName %>DeletePopupComponent
 } from './<%= entityFileName %>-delete-dialog.component';
 <%_ } _%>
+import { Observable } from 'rxjs';
```
deepu105: please move lib import before local imports

PR #7728, opened 2018-06-04
Use openapi-generator for API-first option

cbornet:

[openapi-generator](https://openapi-generator.tech) is a community-driven fork of swagger-codegen. Using it has the following advantages:
* Supports both swagger v2.0 and OpenAPI v3.0 formats
* Will generate "delegate" interfaces that are called from the controller classes. So you don't need to add annotations such as `@RequestParam` in the implementation methods.
* Supports Webflux so it can be used with the future `reactive` option.

### DIFF
--- a/generators/server/templates/README.md.ejs
+++ b/generators/server/templates/README.md.ejs
@@ -206,9 +206,9 @@ will generate few files:
 <%_ } _%>
 
 <%_ if (enableSwaggerCodegen) { _%>
-### Doing API-First development using swagger-codegen
+### Doing API-First development using openapi-generator
 
-[Swagger-Codegen]() is configured for this application. You can generate API code from the `src/main/resources/swagger/api.yml` definition file by running:
+[OpenAPI-Generator]() is configured for this application. You can generate API code from the `src/main/resources/swagger/api.yml` definition file by running:
     <%_ if (buildTool === 'maven') { _%>
 ```bash
 ./mvnw generate-sources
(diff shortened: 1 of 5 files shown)


deepu105:
Since this is a breaking change, I suggest to merge this for v5

cbornet:
Any objection to merge this ? I need it for June 21st 😄

jdubois:
I have no idea what the difference is with swagger codegen, what happens?
Anyway I'm 100% trusting you -> if you need it, merge it :-)

cbornet:
Mostly I'm no longer member of swagger-codegen and switched to the openapi-generator fork 😄 
As for the delegate usage, it's because when using controller overrides, you need to repeat all the annotations on the method parameters which is not user-friendly and error-prone. With a delegated service, things are cleaner, no view setting bleeding in service logic. To work out-of-the-box, some enhancements were needed that I did only in openapi-generator 😄 .  And openapi-generator works with the latest OpenAPI version 3 specs.
I'll update the doc.

PierreBesson:
Openapi-generator is a fork and enhanced swagger-codegen and a community project, I'm trusting @cbornet 100% on this.

MartinDelille:
I must add that @wing328 (Swagger codegen lead developer) is at the origin of the community fork!

wing328:
We've documented the reasons of the fork in https://github.com/OpenAPITools/openapi-generator/blob/master/docs/qna.md. Please have a look and let us know if you've any question.

PR #7735, opened 2018-06-05
Caching Improvements

ruddell:
For Angular, CSS files are stored at the root of the of the app.  For React, CSS files and images referenced in CSS are stored at the root of the app.  This means those files [don't receive the caching headers](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/WebConfigurer.java.ejs#L181-L182).  I updated the Webpack config to put the CSS and referenced image files in the `content` folder like what was done in AngularJS

React also has a static folder with images such as the navbar logo, it isn't hashed by webpack or cached.  Not sure what we wanted to do with that so the react config may need tweaking.


deepu105:
Please test with prod profile as well

jdubois:
Thanks @ruddell - good catch!!

Issue #7731, opened 2018-06-05
Using jhipster server to create a project results in no applicationType in .yo-rc.json


mraible:
##### **Overview of the issue**

When creating a project with `jhipster server`, there is no `applicationType` key registered in `.yo-rc.json`. 

Projects like Ionic for JHipster rely on this key to determine if the app is a monolith or gateway. I could default to monolith if it's not found, but it seems like it should be there.

##### **Motivation for or Use Case**

https://github.com/oktadeveloper/generator-jhipster-ionic/issues/47

##### **Reproduce the error**

Create a project with `jhipster server`, then run the following command after it generates.

```
cat .yo-rc.json | grep applicationType
```

You will get no output.

##### **Suggest a Fix**

Add `applicationType: monolith` when `jhipster server` is used.

##### **JHipster Version(s)**

5.0.0-beta.1

- [x] Checking this box is mandatory (this is just to show you read everything)

manisi:
same error in JHipster version 4.14.4.

jdubois:
I don't think `jhipster server` works at all, it's only meant to be composed by the main generator, but indeed the documentation says this should work - @deepu105 I guess you refactored this part, wdyt?

pascalgrimaud:
It's just a missing option to add to `.yo-rc.json`
It shouldn't be hard to do

deepu105:
The documentation is correct and this is the expected behaviour. And yes if
used seperately this option will not be updated. May be its a good idea to
update the docs and promote the skip client/server flags instead of sub gens

PR #7737, opened 2018-06-06
Fix applicationType value in config when using jhipster server [ci skip]

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/7731

_____

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -360,6 +360,7 @@ module.exports = class extends BaseGenerator {
 
             saveConfig() {
                 this.config.set('jhipsterVersion', packagejs.version);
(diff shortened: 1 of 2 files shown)


generators/server/index.js
```diff
@@ -360,6 +360,7 @@ module.exports = class extends BaseGenerator {
 
             saveConfig() {
                 this.config.set('jhipsterVersion', packagejs.version);
+                this.config.set('applicationType', this.applicationType);
```
deepu105: if this is added here then you need to do it for client sub gen as well

pascalgrimaud: yes, you're right. I will do it later

PR #7752, opened 2018-06-10
fix entity conflicts with RouteComponentProps in React

ruddell:
Fix jhipster/jhipster-core#220

By extending RouteComponentProps in update/delete/detail React pages, we run into conflicts if an entity has the name `history`, `location`, etc.  This PR adds a suffix to the entity props object on those pages.  Another possible fix is to add the words to the reserved list, `Location` is in the example on JDL Studio so I went with this solution instead.
```
export interface RouteComponentProps<P, C extends StaticContext = StaticContext> {
  history: H.History;
  location: H.Location;
  match: match<P>;
  staticContext: C | undefined;
}
```

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-delete-dialog.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-delete-dialog.tsx.ejs
@@ -35,7 +35,7 @@ export class <%= entityReactName %>DeleteDialog extends React.Component<I<%= ent
   }
 
   confirmDelete = event => {
-    this.props.deleteEntity(this.props.<%= entityInstance %>.id);
+    this.props.deleteEntity(this.props.<%= entityInstance %>Entity.id);
     this.handleClose(event);
   }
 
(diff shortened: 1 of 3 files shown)


bertwin:
I agree that using reserved words is not needed, but may i suggest just using the names `entities` and `entity`? It would make the generator code simpler, and it could benefit code reuse in the front-end code.

jdubois:
basically you're adding a suffix - which isn't a bad idea at all in fact, even if that's not what we usually do.
About calling them "entity" instead: usually we use the name of the entity, as it's more descriptive. For the front-end, I expect that code will be changed and refactored quite a bit, so it's good to have the name of the entity, instead of something generic. This is meant to be boilerplate code that will be modified.

bertwin:
it's probably my lazyness :smile: 

screenshot from 2018-06-11 11-10-48
```js
const mapStateToProps = ({ entityWithPaginationAndDTO }: IRootState) => ({
  entities: entityWithPaginationAndDTO.entities,
  totalItems: entityWithPaginationAndDTO.totalItems
});
```

but if it is like this it would still be clear what the entity is. And replacing this button with `<MyCustomButton entity={entity} match={match}/>` in all entities would just be search and replace.

deepu105:
@bertwin I think this is how I had them originally, I don't remember why it was changed. I didnt check the history, may be I changed it or someone else did, but personally i'm ok with either style

PR #7772, opened 2018-06-13
CVE fixed at server side (Web Parameter Tampering)

robertwgil:
At gateway, attacker can intercept response of uaa/api/account and chance authorities that client will use for build screen, adding ROLE_ADMIN, you can see Administration menu, and list all users and routes, exposing privileged informations.

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/GatewayResource.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/GatewayResource.java.ejs
@@ -27,6 +27,8 @@ import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.cloud.netflix.zuul.filters.Route;
 import org.springframework.cloud.netflix.zuul.filters.RouteLocator;
 import org.springframework.http.*;
+import org.springframework.security.access.annotation.Secured;
+import <%=packageName%>.security.AuthoritiesConstants;
 import org.springframework.web.bind.annotation.*;
 
 import com.codahale.metrics.annotation.Timed;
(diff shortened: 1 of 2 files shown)


jdubois:
Not sure about this:

- For routes, indeed anyone can see them, but that's not really important as they are secured by each microservice, and by the `authorized-microservices-endpoints` property in the application.yml. And isn't this going to cause issues for normal users when they want to use those routes? So I'm not totally sure this is a big deal, and this needs to be tested for normal users.
- For users, by default we do have that REST endpoint opened for every logged in user. This often causes questions, and this is because our API is modeled after the GitHub API: this is available for business reasons, otherwise you wouldn't be able to do relationships to Users. So this is isn't specific to gateways, and it's done of purpose, but I agree this could be discussed again.

robertwgil:
@jdubois about routes, I understand that it is secured by each microservice, but this is an **Gain Information** - CVE, this doesn't pass by some penetration test for example, because expose information about infrastructure.

I did tests logging in with normal user and everything works, I'm not sure, but this resource is used only to show information to that screen.

screen shot 2018-06-13 at 11 03 39

jdubois:
If that does work for normal users (which I think it will), then we can indeed add this security check on routes. Please note that those routes will probably be used on the front-end, so anybody (even non-logged in users) will easily find them anyway.

gmarziou:
Also technically it's not a CVE as it is not registered in CVE database. Title should be reworded.

ruddell:
@jdubois 7232a62c2c0c0620edd02a20e7713bc6327d3b5f breaks the relationship dropdown when an entity relates to a user like you mentioned - it logs you out and gives a 401 error when requesting the users for the dropdown.  

While I agree `getRoutes` should definitely be secured, I'm not sure about  `getUsers`. It's confusing when we generate code that doesn't work out-of-the-box (someone in the Gitter pointed this out)

vishal423:
also, shouldn't these checks be added inside SecurityConfiguration.java instead at method level?

jdubois:
@vishal423 well, you have the choice... I usually try to do both, to be sure there is no issue

vishal423:
@jdubois, IMO, doing at both places seems code smell. Also, I believe, we should be consistent in our approach to ensure easy maintenance

PR #7781, opened 2018-06-14
replace entity name and suffix "Entity" on detail/update/delete pages

ruddell:
related to discussion in #7752 

@jdubois you mentioned https://github.com/jhipster/generator-jhipster/pull/7752#issuecomment-396172228 that you preferred to use entity names so feel free to close this if you don't like it.

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-delete-dialog.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-delete-dialog.tsx.ejs
@@ -35,7 +35,7 @@ export class <%= entityReactName %>DeleteDialog extends React.Component<I<%= ent
   }
 
   confirmDelete = event => {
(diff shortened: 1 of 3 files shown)


jdubois:
I'm not a big fan of this, as I'm sure you will end up having to refactor everything, and because I like to have my code with names that carry meaning: here you'll just have "entity" everywhere, which doesn't give any hint of what this is. If I had an entity "BlogEntry", then "blogEntryEntity" gives me more meaning that just "entity".

But the front-end part is the land of @deepu105 , so if Deepu you like this, I'll let you merge it!

deepu105:
As I said i'm really neutral on this one as I'm ok with both styles. @ruddell how bad do you want this? if its really not a big deal then I guess we could just leave it as it is. But if you see any benefit from this let me know

ruddell:
I'm fine without it

Issue #7775, opened 2018-06-14
Sometimes can't generate without prompt, even if there is file .yo-rc.json and other generation problems


JulioJu:
#### **Overview of the issue**

When there is already a folder node_modules in the project, we can't generate jhipster without prompt, even if there is the file .yo-rc.json, contrary to before.

It works randomly.

#### **Motivation for or Use Case**

Cause fails during local build for https://github.com/jhipster/generator-jhipster/pull/7484

More important, it cause problems if we want generate again the application for example for an upgrade.

#### **Reproduce the error**
```bash
#!/usr/bin/env bash
# -*- coding: UTF8 -*-
set -exu

PS4='\n\ntest@test:\w $ '

generatorDir="$HOME/generator-jhipster/"
export JHIPSTER="ngx-default"
export JHIPSTER_SAMPLES="$generatorDir""/travis/samples"
export APP_FOLDER="$JHIPSTER_SAMPLES""/$JHIPSTER""-sample"
bash "$generatorDir/travis/scripts/01-generate-entities.sh"
cd "$APP_FOLDER"
while true ; do
    yarn init -y
    yarn link "generator-jhipster"
    cp "$JHIPSTER_SAMPLES""/$JHIPSTER""/.yo-rc.json" .
    ls -la
    jhipster --force --no-insight --skip-checks --with-entities \
        --skip-git --skip-commit-hook
done
```

#### Explanations

An error occur started from 10071199b35c7940706d7173ccb12b22d8b28f4f from @MathieuAA (I've used `git bisect`).

At 10071199b35c7940706d7173ccb12b22d8b28f4f the output is:
```
Using JHipster version installed locally in current project's node_modules
Running default command
Executing jhipster:app
Options: force: true, insight: false, skipChecks: true, withEntities: true, skipGit: true, skipCommitHook: true, skip-checks: true, with-entities: true, skip-git: true, skip-commit-hook: true


        ██╗ ██╗   ██╗ ████████╗ ███████╗   ██████╗ ████████╗ ████████╗ ███████╗
        ██║ ██║   ██║ ╚══██╔══╝ ██╔═══██╗ ██╔════╝ ╚══██╔══╝ ██╔═════╝ ██╔═══██╗
        ██║ ████████║    ██║    ███████╔╝ ╚█████╗     ██║    ██████╗   ███████╔╝
[...]

PR #7800, opened 2018-06-17
Updated application/entity generation

MathieuAA:

Used provided ways to get the config file (`.yo-rc.json`) content to be read and used by the generators.

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -21,6 +21,7 @@ const path = require('path');
 const _ = require('lodash');
 const fs = require('fs');
 const Generator = require('yeoman-generator');
+const Storage = require('yeoman-generator/lib/util/storage');
 const chalk = require('chalk');
 const Insight = require('insight');
 const shelljs = require('shelljs');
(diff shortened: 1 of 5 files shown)


generators/generator-base-private.js
```diff
@@ -21,6 +21,7 @@ const path = require('path');
 const _ = require('lodash');
 const fs = require('fs');
 const Generator = require('yeoman-generator');
+const Storage = require('yeoman-generator/lib/util/storage');
```
deepu105: @MathieuAA is this public API or internal API?

generators/generator-base-private.js
```diff
@@ -54,6 +55,7 @@ module.exports = class extends Generator {
         this.env.options.appPath = this.config.get('appPath') || CLIENT_MAIN_SRC_DIR;
         // expose lodash to templates
         this._ = _;
+        this.createConfigFromNewConfFile();
```
deepu105: @MathieuAA why is this needed, could you please explain?

generators/generator-base-private.js
deepu105: ok

generators/generator-base-private.js
deepu105: you need this only for import-jdl right? i'm worried about the blueprints, as I dont know what side effect this would add to all generators

MathieuAA: I think I caught your meaning, I can try to make it only possible for the import-jdl subgen, now that I understand how the Storage works.

JulioJu:
@MathieuAA thanks, fix, perfectly https://github.com/jhipster/generator-jhipster/issues/7775 !

JulioJu:
Sorry, this commit introduce a new bug. See https://github.com/jhipster/generator-jhipster/issues/7775#issuecomment-397980893.

Issue #7238, opened 2018-03-06
Blueprint enhancements / fixes [ the future ]


sendilkumarn:
After playing a lot with blueprints, I feel few things are missing or good to have. Here is a set of things which would be super awesome to have it in the blueprint.

- [ ] `-b` shortcut for blueprint
- [x] create a blueprint generator (that helps to generate a blueprint)
- [x] the arguments passed to the jhipster commands is kind of flaky and it is not properly passed on
- [x] rework needed on the server files
- [ ] option to override the path for the files. This will be extremely helpful in the context of blueprint
- [x] Add blueprint support to the missing modules [only supported for client, server, entity, entity-client, entity-server, spring-controller and spring-service)

deepu105:
@sendilkumarn I moved these 2 out as we need to discuss them, as I don't agree with them. See reasons below
- [ ] remove `generator-jhipster` from the blueprint completely
Maybe we can remove jhipster from the name, but you still need the generator in the name otherwise yeoman will not be able to identify it during composing in legacy mode. I'll see what can be done here but cannot promise this
- [ ] Option to add a separate command for `blueprint` to work. Something like `khipster` (this removes the need of jhipster) in the client machine (however the blueprint will still need the generator as a dependency) ⚠️ (just an idea)
I dont agree with this, even if there is a way to do it, as this would hijack JHipster, it is impirtant that people know thet they are using JHipster. If a blueprint wants to become independent it should implement its own CLI and release it, in the CLI it could integrate with JHipster CLI

sendilkumarn:
> but you still need the generator in the name 

wasnt this removed in the new version. We can wrap or add them inside the base generator if needed. I feel it is redundant.

> as this would hijack JHipster,

I would argue not. (that is the reason why I added warning symbol). Technically, if we add blueprint support to all the generators it is completely doable. But all they will need the base generator, without that they cannot use it. It is like adding aliases. But it is pretty arguable on both the sides. Let us discuss this. But it is not mandatory.

deepu105:
1. Do you mean yeoman removed the need to have generator in the name? i'm not aware of that, the yeoman website still says its needed.

deepu105:
some work done in #7437

sendilkumarn:
@deepu105 after this PR, the blueprint is not working any more. now the blueprint is going in a loop

deepu105:
ok, i'll take a look

Thanks & Regards,
Deepu

[...]

PR #7802, opened 2018-06-17
Blueprint updates

deepu105:

related to #7238

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2434,4 +2434,22 @@ module.exports = class extends PrivateBase {
         dest.fieldNameChoices = [];
         dest.relNameChoices = [];
     }
+
+    /**
+     * Get all the generator configuration from the .yo-rc.json file
+     * @param {Generator} generator the generator instance to use
+     * @param {boolean} force force getting direct from file
+     */
+    getAllJhipsterConfig(generator = this, force) {
+        let configuration = generator.config.getAll();
+        if ((force || !configuration.baseName) && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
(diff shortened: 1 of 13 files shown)


generators/generator-base.js
```diff
@@ -2434,4 +2434,24 @@ module.exports = class extends PrivateBase {
...
+     * @param {Generator} generator the generator instance to use
+     * @param {boolean} force force getting direct from file
+     */
+    getAllJhipsterConfig(generator = this, force) {
+        let configuration = generator.config.getAll() || {};
+        if ((force || !configuration.baseName) && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
+            configuration = JSON.parse(fs.readFileSync('.yo-rc.json', { encoding: 'utf-8' }))['generator-jhipster'];
+        }
+        if (!configuration.get || typeof configuration.get !== 'function') {
+            Object.assign(configuration, {
```
MathieuAA: You're trying my previous solution, I suggest you try my new one.
If you need the configuration in other generators without having to pass it around, I suggest you use the `this.createConfigFromNewConfFile` (may have to rename it).
You're simulating a Storage class, and it may work but isn't valid in some cases.

MathieuAA: I wonder if you still need to do this by the way... with the method being plugged in the top class.

deepu105: @MathieuAA yes indeed my solution is a bit hacky and frankly, i'm not very proud of this. I saw your solution only after while resolving conflicts on this PR yesterday, but i'm quite tight on time, so if you have any spare time will you be kind enough to update this? else i'll do it later when I have time to study yours and use it. Also please note that the idea here is to always load the config ourself without relying on yeoman, so that we get the JHipster config all the time and not the blueprints config(so we might have to hardcode the `generator-jhipster` key in your solution as well

MathieuAA: Agreed!

PR #7804, opened 2018-06-18
Fix manifest icons to match Angular or React

mraible:
This is an attempt to convert the SVG JHipster icons to PNG and make them square. I think there's an error in the React SVG icon since it has a little white square next to the React tattoo. I used both online converters (which don't create a transparent background) and ImageMagick. The ImageMagick command I used is:

```
convert -background none logo-jhipster-react.svg logo-jhipster-react.png
```


mraible:
Added fix for icons not resolving with Angular. Here's the before screenshot:

angular-before

And the after:

angular-after

Issue #7842, opened 2018-06-23
TextBlob fields validations stopped working in V5


Tonterias:
##### **Overview of the issue**

Jhipster v5 does not like TextBlob fields validations:

~~~
entity Comment {
    creationDate Instant required
    commentText TextBlob minbytes(3) maxbytes(5000) required
    isOffensive Boolean
}


Hibernate JPA 2 Metamodel does not work with Bean Validation 2 for LOB fields, so LOB validation is disabled
WARNING! Cannot use validation in .jhipster/Comment.json for field {
"fieldName": "commentText",
"fieldType": "byte[]",
"fieldTypeBlobContent": "text",
"fieldValidateRules": [
"minbytes",
"maxbytes",
"required"
],
"fieldValidateRulesMinbytes": 3,
"fieldValidateRulesMaxbytes": 5000
}
~~~

This worked fine in version 4.13.

##### **Motivation for or Use Case**

As Gaël said, if it was working before, it should be a bug: https://stackoverflow.com/questions/50992210/how-to-create-a-field-validation-for-textblob-in-jhipster-v5-0-using-jdl

##### **Reproduce the error**

Try creating an Entity like this, please:

~~~
entity Comment {
    creationDate Instant required
    commentText TextBlob minbytes(3) maxbytes(5000) required
    isOffensive Boolean
}
~~~

##### **JHipster Version(s)**

v5 & v5.0.1 (it was working on v4.13)

##### **JHipster configuration**

D:\BasuraTemporal\Borrar\BOOTSRAP\JHispter\Packtpub\newstest501>jhipster info
Using JHipster version installed locally in current project's node_modules
Executing jhipster:info
Options:
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**
v5.0.1
```
newstest-501@0.0.0 D:\BasuraTemporal\Borrar\BOOTSRAP\JHispter\Packtpub\newstest501
`-- generator-jhipster@5.0.1

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.web.newstest501"
    },
    "jhipsterVersion": "5.0.1",
    "applicationType": "monolith",
    "baseName": "newstest501",
    "packageName": "com.web.newstest501",
    "packageFolder": "com/web/newstest501",
    "serverPort": "8080",
[...]

PR #7872, opened 2018-06-27
Fix react input for text blobs

agaspardcilia:
Fixed it as it was mentioned in #7842

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
@@ -273,7 +273,6 @@ _%>
 <%_ } _%>
 <%_ }) _%>
   render() {
-    const isInvalid = false;
     const { <%= entityInstance %>Entity,<%
     uniqueRealtionFields.forEach(rel => {
     %> <%= rel %>,<% }) %> loading, updating } = this.props;
@@ -427,6 +426,7 @@ _%>
                 </div>
               ) : null}
               <input id="file_<%= fieldName %>" type="file" onChange={this.onBlobChange(<%= isAnImage %>, '<%= fieldName %>')} <% if (isAnImage) { %>accept="image/*"<% } %> />
+              <AvInput type="hidden" name="<%= fieldName %>" value={<%= fieldName %>} <%- include react_validators %>/>
             </AvGroup>
 
             <%_ } else { _%>
@@ -645,7 +645,7 @@ _%>
               <span className="d-none d-md-inline" ><Translate contentKey="entity.action.back">Back</Translate></span>
             </Button>
             &nbsp;
-            <Button color="primary" id="save-entity" type="submit" disabled={isInvalid || updating}>
+            <Button color="primary" id="save-entity" type="submit" disabled={updating}>
               <FontAwesomeIcon icon="save" />&nbsp;
               <Translate contentKey="entity.action.save">Save</Translate>
             </Button>
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
@@ -435,7 +435,7 @@ _%>
                   <%= fieldNameHumanized %>
                 </Translate>
               </Label>
-              <AvField id="<%= entityFileName %>-<%= fieldName %>" type="text" name="<%= fieldName %>" <%- include react_validators %>/>
+              <AvInput id="<%= entityFileName %>-<%= fieldName %>" type="textarea" name="<%= fieldName %>" <%- include react_validators %>/>
               <%_ } _%>
             <%_ } else { _%>
             <Label id="<%= fieldName %>Label" for="<%= fieldName %>">
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
@@ -47,6 +47,13 @@ if (field.fieldValidate === true) {
     if (['Integer', 'Long', 'Float', 'Double', 'BigDecimal'].includes(fieldType)) {
         validators.push('number: { value: true, errorMessage: translate(\'entity.validation.number\') }');
     }
+    // Not supported anymore because the server can't check the size of the blob before downloading it completely.
+    // if (rules.includes('minbytes')) {
+    //     validators.push('minbytes: { value: ' + field.fieldValidateRulesMinbytes + ', errorMessage: translate(\'entity.validation.minbytes\', { min: ' + field.fieldValidateRulesMinbytes + ' }) }');
+    // }
+    // if (rules.includes('maxbytes')) {
+    //     validators.push('maxbytes: { value: ' + field.fieldValidateRulesMaxbytes + ', errorMessage: translate(\'entity.validation.maxbytes\', { min: ' + field.fieldValidateRulesMaxbytes + ' }) }');
+    // }
 
     result = validators.join(',\n                ');
     result = `validate={{\n                ${result}\n                }}`;
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
@@ -47,13 +47,6 @@ if (field.fieldValidate === true) {
     if (['Integer', 'Long', 'Float', 'Double', 'BigDecimal'].includes(fieldType)) {
         validators.push('number: { value: true, errorMessage: translate(\'entity.validation.number\') }');
     }
-    // Not supported anymore because the server can't check the size of the blob before downloading it completely.
-    // if (rules.includes('minbytes')) {
-    //     validators.push('minbytes: { value: ' + field.fieldValidateRulesMinbytes + ', errorMessage: translate(\'entity.validation.minbytes\', { min: ' + field.fieldValidateRulesMinbytes + ' }) }');
-    // }
-    // if (rules.includes('maxbytes')) {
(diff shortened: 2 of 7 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
```diff
@@ -435,7 +435,7 @@ _%>
                   <%= fieldNameHumanized %>
                 </Translate>
               </Label>
-              <AvField id="<%= entityFileName %>-<%= fieldName %>" type="text" name="<%= fieldName %>" <%- include react_validators %>/>
+              <AvInput id="<%= entityFileName %>-<%= fieldName %>" type="textarea" name="<%= fieldName %>" <%- include react_validators %>/>
```
deepu105: @agaspardcilia can you check if the validators applied still works for AvInput, some might be for AvField

agaspardcilia: Are you sure those are still supported ? I mean when I try to create an entity with `jhe` it doesn't even ask me about any validation rule. I don't even know if it's normal. And if it's not supported anymore, why is it still available in the JDL. 
Am I missing something ?

deepu105: There should be validation for text blob field as per below
```
{
            when: response => response.fieldAdd === true && response.fieldValidate === true,
            type: 'checkbox',
            name: 'fieldValidateRules',
            message: 'Which validation rules do you want to add?',
            choices: (response) => {
                // Default rules applicable for fieldType 'LocalDate', 'Instant',
                // 'ZonedDateTime', 'UUID', 'Boolean', 'ByteBuffer' and 'Enum'
                const opts = [
                    {
                        name: 'Required',
                        value: 'required'
                    }/* ,
                    {
                        name: 'Unique',
                        value: 'unique'
                    } */
                ];
                if (response.fieldType === 'String' || response.fieldTypeBlobContent === 'text') {
                    opts.push(
                        {
                            name: 'Minimum length',
                            value: 'minlength'
                        },
                        {
                            name: 'Maximum length',
                            value: 'maxlength'
                        },
                        {
                            name: 'Regular expression pattern',
                            value: 'pattern'
                        }
                    );
                } else if (['Integer', 'Long', 'Float', 'Double', 'BigDecimal'].includes(response.fieldType)) {
                    opts.push(
                        {
                            name: 'Minimum',
                            value: 'min'
                        },
                        {
                            name: 'Maximum',
                            value: 'max'
                        }
                    );
                } else if (response.fieldType === 'byte[]' && response.fieldTypeBlobContent !== 'text') {
                    opts.push(
                        {
                            name: 'Minimum byte size',
                            value: 'minbytes'
                        },
                        {
                            name: 'Maximum byte size',
                            value: 'maxbytes'
                        }
                    );
                }
                return opts;
            },
```

agaspardcilia: I guess I found a bug then.

deepu105: oh wait its disabled in the line above

```
        {
            when: response => response.fieldAdd === true && response.fieldType !== 'byte[]' && response.fieldType !== 'ByteBuffer',
            type: 'confirm',
            name: 'fieldValidate',
            message: 'Do you want to add validation rules to your field?',
            default: false
        },
```
I have no idea why its disabled @jhipster/developers does anyone remember why we removed validation for text fields

agaspardcilia: Julien told me that it was because of the size limit. The server was downloading the complete file before making any check. It might be a good idea to let the user mark the field as `requiered` tho.

deepu105: yes, at least required should be there

agaspardcilia: With validation enabled, everything works with text blobs but the `required`  validation is not working on regular blobs.

agaspardcilia:
I removed `required` test in backend for blobs. I think we can do without since it wasn't even supported before.

deepu105:
@jhipster/developers can someone test this. I don't have enough time

generators/entity-server/templates/src/main/java/package/common/field_validators.ejs
```diff
@@ -25,8 +25,12 @@ if (field.fieldValidate === true) {
     const rules = field.fieldValidateRules;
     const validators = [];
     const MAX_VALUE = 2147483647;
+    const isBlob = field.fieldType === 'byte[]';
 
-    if (rules.includes('required')) {
+    fieldTypeBlobContent
```
ruddell: Does this do anything?  I'm surprised it passed lint

deepu105: indeed this looks strange. @ruddell ejs linting is not as good as eslint. it only checks for syntax errors

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
```diff
@@ -645,7 +645,7 @@ _%>
               <span className="d-none d-md-inline" ><Translate contentKey="entity.action.back">Back</Translate></span>
             </Button>
             &nbsp;
-            <Button color="primary" id="save-entity" type="submit" disabled={isInvalid || updating}>
+            <Button color="primary" id="save-entity" type="submit" disabled={updating}>
```
deepu105: why is the invalid check removed? the idea was you cannot submit a form if it doesnt pass validation

generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
```diff
@@ -47,12 +47,12 @@ if (field.fieldValidate === true) {
     if (['Integer', 'Long', 'Float', 'Double', 'BigDecimal'].includes(fieldType)) {
         validators.push('number: { value: true, errorMessage: translate(\'entity.validation.number\') }');
     }
-    if (rules.includes('minbytes')) {
-        validators.push('minbytes: { value: ' + field.fieldValidateRulesMinbytes + ', errorMessage: translate(\'entity.validation.minbytes\', { min: ' + field.fieldValidateRulesMinbytes + ' }) }');
-    }
-    if (rules.includes('maxbytes')) {
-        validators.push('maxbytes: { value: ' + field.fieldValidateRulesMaxbytes + ', errorMessage: translate(\'entity.validation.maxbytes\', { min: ' + field.fieldValidateRulesMaxbytes + ' }) }');
-    }
+    // if (rules.includes('minbytes')) {
```
deepu105: Sorry, but I again forgot, why didi we disable this again? can you add a comment above this and other commented lines saying why we disable this so we dont forget in future

generators/entity-server/templates/src/main/java/package/common/field_validators.ejs
agaspardcilia: I was definitely tired when I did this. :sweat_smile:

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
agaspardcilia: Let me double check, but I think I removed it because it was unused. Like the variable was initialized but never modified or anything.

agaspardcilia: If you search for `isInvalid` refs in the file you only find it 2 times. The first is the declaration and the second is `disabled={isInvalid || updating}`. So yes, I think it's pretty much useless.

generators/entity-client/templates/react/src/main/webapp/app/entities/react_validators.ejs
agaspardcilia: Now that I think about it, we removed it because it's not supported well by the server. But `required` is also not supported at all by bean validation. I mean, using `@NotNull` on `byte[]` creates compilation errors on spring generated code. It might be a good idea to enable size constraints for the client only since it's already there and let the server side as it is. wdyt ?

deepu105: not sure about it, people might complain about a mismatch in validation @jdubois WDYT?

jdubois: doing it only on the client side is nice from an UI perspective, but don't protect the server at all, so that gives a false sense of security. Usually, we always take the lower common denominator when there is a mismatch with the front end and back end -> so I wouldn't do any validation here. I know it's bad, but we can't do it technically on the server side.

agaspardcilia: Alright, I'll remove the comments too then.

deepu105: I would still prefer to leave the comments with explanation, may be we can fix it in the future

generators/entity/prompts.js
```diff
@@ -735,18 +735,20 @@ function askForField(done) {
...
-                            name: 'Minimum byte size',
-                            value: 'minbytes'
-                        },
-                        {
-                            name: 'Maximum byte size',
-                            value: 'maxbytes'
-                        }
-                    );
                 }
+                // Not supported anymore
```
deepu105: can we add the reason in comment

agaspardcilia:
All done, a squash would be great here I think. :3

agaspardcilia:
@deepu105 Yup, that's right but `required` is a pretty important imo, do you think there's a way to make the validation work for the server ?

deepu105:
I'm not sure, I have to check, I was expecting @NotNull to work

Thanks & Regards,
Deepu

deepu105:
@agaspardcilia can you see if you can find a solution for @NotNull issue with some guys from Ippon, else lets remove the validation on front as well for this

PR #7874, opened 2018-06-27
Skip gway pages when skip service discovery

### LINKED ISSUE #7827: Redundant / broken gateway sub-module/component code without serviceDiscovery
If you generate a gateway without service discovery, then, on server side it doesn't generate the GatewayResource.java. However, no such checks are added on frontend layer and thus it generate gateway page/component code (along with associated service, routes, reducer etc) which are redundant as they will never work without backend API

##### **Suggest a Fix**
This change got introduced by below commit on server side. Either we should revert this or apply similar checks on frontend layer.
https://github.com/jhipster/generator-jhipster/commit/f3d100c6d624ac58157a6c72ee4c949da563e574

##### **JHipster Version(s)**
5.0.0-beta.3

{
  "generator-jhipster": {
    "promptValues": {
    },
    "jhipsterVersion": "5.0.0-beta.3",
    "applicationType": "gateway",
    "baseName": "rgtw5",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "jwtSecretKey": "4a95621ba2e00b566a6b5625a4304126fdc687e7",
    "clientFramework": "react",
    "useSass": false,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}

ruddell:
Fix #7827

- Skips gateway pages when no service discovery
- Don't copy gateway.json translation file into non-gateways or no service dsicovery
- Adds expected files tests for no service discovery


generators/client/index.js
```diff
@@ -184,11 +184,6 @@ module.exports = class extends BaseGenerator {
                     this.baseName = baseName;
                 }
 
-                this.serviceDiscoveryType = configuration.get('serviceDiscoveryType') === 'no' ? false : configuration.get('serviceDiscoveryType');
```
deepu105: why is this removed wouldn't this fail truthy checks as the default value will be `no` and would pass the check?

deepu105: I guess this was code handling legacy .yo-rc files when we had 'no' as default option. Can you check with an old .yo-rc file before we changed from 'no' to false

ruddell: Tested with `no` and fixed the language generator in that case.  Everything else looks good

PR #7882, opened 2018-06-28
Refactor Server HeaderUtil class

DanielFran:


jdubois:
BTW I'm thinking about refactoring this class and its sister PaginationUtil into the JHipster framework lib - there is little interest to generate this

deepu105:
@jdubois the reason for not moving this was due to the dynamic app name and dynamic i18n stuff on this class I guess

jdubois:
@deepu105 yes, but this is only for HeaderUtils, and that's indeed a good point - then I guess we could pass the name as a parameter, it wouldn't be as efficient but this is really over-optimizing

Issue #7896, opened 2018-06-29
Must remove the .yo-rc.json file if app is ever generated in the home directory


wilsonjl5:
##### **Overview of the issue**

When creating a test application for jHipster 5 I accidentally generated the app in my home directory. After noticing my mistake, I removed the files and created a new directory for my application. When trying to generate it I ran into the following issue: the previously generated project's .yo-rc.json file is being used as the config for any and all jHipster projects regardless of where they exist.

##### **Motivation for or Use Case**

First-time users are likely to encounter this problem. 

##### **Reproduce the error**

Generate a new app in your home directory. Then try generating another application anywhere else. You will see something similar to the screenshot below.
<img width="839" alt="screen shot 2018-06-29 at 12 05 55 pm" src="https://user-images.githubusercontent.com/7031213/42102759-eb083096-7b94-11e8-923c-800a811c17e9.png">

##### **Related issues**

None, that I could find.

##### **Suggest a Fix**

A work-around: if you want to generate an app elsewhere on your computer, outside of your home directory, was to remove this hidden file (.yo-rc.json) **from the home directory** via command line.
`
rm .yo-rc.json
`

##### **JHipster Version(s)**

```
myapplication@0.0.0 /Users/leighawilson/Desktop/Ippon/Jhipster/myapplication
└── generator-jhipster@5.0.1 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp"
    },
    "jhipsterVersion": "5.0.1",
    "applicationType": "monolith",
    "baseName": "myapplication",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "clientFramework": "angularX",
    "useSass": false,
    "clientPackageManager": "yarn",
    "testFrameworks": [],
[...]

PR #7898, opened 2018-06-30
Check home directory

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/7896

@leighadellinger : what do you think ?

@PierreBesson :
- the code is really simple, and it should work for all OS
- as discussed during  the OnBoarding, I reduce the timeout for git connection from 15sec to 5sec

warning_home_directory

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -789,7 +789,7 @@ module.exports = class extends Generator {
     checkGitConnection() {
         if (!this.gitInstalled) return;
         const done = this.async();
-        exec('git ls-remote git://github.com/jhipster/generator-jhipster.git HEAD', { timeout: 15000 }, (error) => {
+        exec('git ls-remote git://github.com/jhipster/generator-jhipster.git HEAD', { timeout: 5000 }, (error) => {
             if (error) {
                 this.warning(`Failed to connect to "git://github.com"
 1. Check your Internet connection.
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2060,7 +2060,7 @@ module.exports = class extends PrivateBase {
         this.log(chalk.white('Welcome to JHipster ') + chalk.yellow(`v${packagejs.version}`));
         this.log(chalk.white(`Application files will be generated in folder: ${chalk.yellow(process.cwd())}`));
         if (process.cwd() === this.getUserHome()) {
-            this.log(chalk.red.bold(`️⚠️  WARNING ⚠️  You are in your HOME folder!`))
+            this.log(chalk.red.bold('️⚠️  WARNING ⚠️  You are in your HOME folder!'));
         }
         this.log(chalk.green(' _______________________________________________________________________________________________________________\n'));
         this.log(chalk.white(`  Documentation for creating an application is at ${chalk.yellow('https://www.jhipster.tech/creating-an-app/')}`));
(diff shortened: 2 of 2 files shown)


jdubois:
Yes but maybe point to the documentation to explain the issue (here we see the issue but don't understand what happens)

pascalgrimaud:
@jdubois :
Is this documentation about 'Destination context' OK for you ? http://yeoman.io/authoring/file-system.html
If yes, I can add this link, something like:
> You are in your HOME folder! See the documentation at http://yeoman.io/authoring/file-system.html

Otherwise, I don't see any reference in our docs. Maybe we should add it ?

jdubois:
I'd rather have something in our own documentation - I guess most people don't even know we're based on Yeoman, so that page will confuse them.

We need a section in the "installation" page.

-> if you don't have the time I'll do it, it's pretty simple

pascalgrimaud:
Ok, I will do it.
I'll try to find time tonight, after the 2nd match :-)

Le sam. 30 juin 2018 à 18:46, Julien Dubois <notifications@github.com> a
écrit :

PierreBesson:
I think the error message should state "this can cause problems, you should always create a new directory and run the jhipster command from here."

pascalgrimaud:
@PierreBesson : I will add this

pascalgrimaud:
Last update:

warning_home_directory1

vishal423:
can we be more explicit on problems that it can cause so that end user can take correct decision? 
Also, is it possible to print path of .yo.rc.json during code generation if that's not in current directory and steps to fix?

PierreBesson:
I would write it like that :

Generation choices will be stored in a .yo-rc.json file which yeoman uses to remember those choices when regenerating an application. Be warned that when running `jhipster` yeoman will pick up any existing .yo-rc.json file present in a parent directory.
So it is strongly recommanded to not generate a JHipster project in your HOME directory as this will create a parent .yo-rc.

vishal423:
@pascalgrimaud, Per my understanding, it would happen if .yo-rc.json is present in any parent directory of current and not tied to only HOME directory. Following is the yeoman documentation explaining this:

> While running a generator, Yeoman will try to figure some things out based on the context of the folder it’s running from.
> 
> Most importantly, Yeoman searches the directory tree for a .yo-rc.json file. If found, it considers the location of the file as the root of the project. Behind the scenes, Yeoman will change the current directory to the .yo-rc.json file location and run the requested generator there.
> 
> The Storage module creates the .yo-rc.json file. Calling this.config.save() from a generator for the first time will create the file.
> 
> So, if your generator is not running in your current working directory, make sure you don’t have a .yo-rc.json somewhere up the directory tree.

pascalgrimaud:
@vishal423 : yes, agree with you! But it's completely related to Yeoman not directly to JHipster. We can of course added this to documentation (see https://github.com/jhipster/jhipster.github.io/pull/594 as it can be improved maybe), but not to the generator itself, as it already displays a lot of information

jdubois:
Yes, this is 100% due to Yeoman, that's supposed to be a feature, in fact.
The idea is that you could run that command in any directory of your application, and that would still work.
Of course, if you created your app in your home directory, then Yeoman thinks that everything inside your home directory is your app.
It's not a stupid feature, but it's surprising for many people. Then, I wouldn't recommend running an application inside your home directory, without reading the documentation or the warnings.

vishal423:
To be specific to HOME, can we abort installation if user tries to install in HOME directory with appropriate message?

deepu105:
Guys, I guess we do enough hand-holding already. So I would say in this case it's up to the user to at least read whats displayed on the console. We are talking about technically capable people here aren't we :wink: 

So let's leave this at it. We have more pressing stuff to spend time on, don't you agree?

jdubois:
Yes and if someone wants to generate this in his home directory, let him do what it wants... I even do it from time to time when I set up a test server quickly

PR #7903, opened 2018-06-30
set config at once to ensure updates are saved

ruddell:
Weird issue - when you change your answers on the second run of a generator, the config isn't saved. 
 Saving the config items at once instead of separately updates the `.yo-rc.json` with the new prompt answers, I went ahead and changed it everywhere to be consistent.  Not sure why this works and setting individually doesn't work

Steps to reproduce:
 - Generate a monolith (use `--skip-install` for faster generation)
 - Make docker-compose config directory
 - Run `jhipster docker-compose` to generate Docker Compose config, choosing any options
 - Re-run `jhipster docker-compose` and change any of the prompt answers
 - Check the `.yo-rc.json` and see the new choices aren't saved

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -326,23 +326,26 @@ module.exports = class extends BaseGenerator {
             },
 
             saveConfig() {
-                this.config.set('jhipsterVersion', packagejs.version);
-                this.config.set('applicationType', this.applicationType);
-                this.config.set('baseName', this.baseName);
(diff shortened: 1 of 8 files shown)


deepu105:
Regardless of the final solution, I do like this style of saving the config
rather than 1 by 1

jdubois:
Oh what an ugly bug, that's an awesome finding @ruddell !! And yes this style is much better.

PR #7916, opened 2018-07-02
avoid double suffixes in service/controller subgenerators

ruddell:
This PR ensures that when you use the service/controller subgenerators, it doesn't end in a double suffix.  For example, running `jhipster service FooService` generated `FooServiceService` instead of `FooService`

This would also make it easier to configure the suffixes with a flag in case a user wants to use "Controller" instead of "Resource" (in the future)

### DIFF
--- a/generators/spring-controller/index.js
+++ b/generators/spring-controller/index.js
@@ -106,8 +106,8 @@ module.exports = class extends BaseGenerator {
     _writing() {
         return {
             writing() {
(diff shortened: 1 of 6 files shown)


vishal423:
In one of my project, I actually required double suffixes, however, that could be treated as rare scenario.

deepu105:
@vishal423 I guess you could still do it like `jhipster service fooServiceService`

PR #7931, opened 2018-07-05
JHipster online statistics

agaspardcilia:
This PR aim to replace Insight by JHipster online statistics.


cli/commands.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
         argument: ['languages...'],
         desc: 'Select languages from a list of available languages. The i18n files will be copied to the /webapp/i18n folder'
     },
+    linker: {
```
deepu105: i'm not sure if its good name for this, `link` or `link-account` might be better

cli/commands.js
```diff
@@ -49,6 +49,9 @@ module.exports = {
...
+        desc: 'Link your generator to your JHipster online account'
```
deepu105: Link the installed JHipster CLI to your JHipster Online account

generators/app/index.js
```diff
@@ -346,7 +338,16 @@ module.exports = class extends BaseGenerator {
...
                 this.skipUserManagement && (config.skipUserManagement = true);
                 this.config.set(config);
+            },
+
+            insight() {
+                if (this.existingProject) return;
+
+                const yorc = Object.assign({}, this.configOptions);
+                yorc.applicationType = this.applicationType;
+                statistics.sendYoRc(yorc, this.jhipsterVersion);
```
deepu105: we shouldn't blindly send everything, we need to remove sensitive stuff like packageName, jwt token etc

generators/app/index.js
```diff
@@ -346,7 +338,16 @@ module.exports = class extends BaseGenerator {
                 this.skipServer && (config.skipServer = true);
                 this.skipUserManagement && (config.skipUserManagement = true);
                 this.config.set(config);
+            },
+
+            insight() {
+                if (this.existingProject) return;
```
deepu105: If it is existing project send the data with a flag to indicate this is regeneration

generators/client/index.js
```diff
@@ -249,16 +249,6 @@ module.exports = class extends BaseGenerator {
     // Public API method used by the getter and also by Blueprints
     _configuring() {
         return {
-            insight() {
```
deepu105: send `sendSubGenEvent` to track execution of this sub generator done seperately

generators/docker-compose/index.js
```diff
@@ -100,7 +101,6 @@ module.exports = class extends BaseGenerator {
                 this.monitoring = this.config.get('monitoring');
                 this.consoleOptions = this.config.get('consoleOptions');
                 this.useKafka = false;
-                this.useMemcached = false;
```
deepu105: this seems unrelated?

generators/docker-compose/index.js
```diff
@@ -243,15 +242,6 @@ module.exports = class extends BaseGenerator {
                     if (messageBroker === 'kafka') {
                         this.useKafka = true;
                     }
-                    // Add Memcached support
```
deepu105: ??

generators/generator-base-private.js
```diff
@@ -315,34 +314,6 @@ module.exports = class extends Generator {
...
 
-    /**
-     * insight
-     *
-     * @param trackingCode
-     * @param packageName
-     * @param packageVersion
-     * @returns {Insight}
-     */
-    insight(trackingCode = 'UA-46075199-2', packageName = packagejs.name, packageVersion = packagejs.version) {
```
deepu105: I wouldn't completely remove this since it will reduce stats a lot as the linking is really opt-in and multistep(due to the login part) and I don't think most people will do it.

@jdubois WDYT? IMO we should keep existing as well atleast for a while

generators/import-jdl/index.js
```diff
@@ -165,8 +166,42 @@ module.exports = class extends ImporterGenerator {
     get configuring() {
         return {
             insight() {
-                const insight = this.insight();
-                insight.trackWithEvent('generator', 'import-jdl');
+                statistics.sendSubGenEvent('generator', 'import-jdl');
+            },
+
+            parseJDL() {
```
deepu105: This seems unrelated here

generators/server/index.js
```diff
@@ -337,23 +337,6 @@ module.exports = class extends BaseGenerator {
     // Public API method used by the getter and also by Blueprints
     _configuring() {
         return {
-            insight() {
```
deepu105: send a subGen event

generators/statistics.js
```diff
@@ -0,0 +1,121 @@
+const uuid = require('uuid/v1');
```
deepu105: use `uuid/v4`

generators/statistics.js
```diff
@@ -0,0 +1,121 @@
...
+        this.config = new Config({
+            configName: 'jhipster-insight',
+            defaults: {
+                clientId: uuid(),
+                doNotAskCounter: 0
+            }
+        });
+        this.statisticsAPI = API_PATH;
+        this.clientId = this.config.get('clientId');
+        this.doNotAskCounter = this.config.get('doNotAskCounter');
```
deepu105: This shouldnt be per project, this should be saved in the a global config store. Is `new Config` creating a global store in users home or something? dow it work with winodws/linux/mac?

generators/import-jdl/index.js
```diff
@@ -220,6 +255,8 @@ module.exports = class extends ImporterGenerator {
                 } catch (error) {
                     this.error(`Error while generating entities from the parsed JDL\n${error}`);
                 }
+
+                statistics.sendCrashReport('source', 'stack', 'generatorVersion', 'yorc', 'jdl');
```
deepu105: I didnt find this method implementation?

cli/commands.js
deepu105: But in general I dont think we need a sub-gen for this

agaspardcilia:
`jhipster linker` just link a generator to a JHipster online user. Like insight, when a User launches the generator for the first time, he will be asked if he doesn't mind JHipster collecting stuffs. Data will be send even without linking an account. The linker subgen is just a way for the user to access his personal stats, nothing more.

And yes, there's a lot of required changes. This is not finished at all.

generators/generator-base-private.js
jdubois: normally (I didn't check the code), the user would send the data anonymously like today. Then, if he chooses to, he can also login (that step is optional), so he can have more features in JHipster Online - in that case, his data will be linked to his account, and he will also be able to delete it

cli/commands.js
agaspardcilia: What should I do ? Put it in the `app` subgen ?

generators/statistics.js
agaspardcilia: Julien wants to avoid collision as much as possible. Is it the right one for that ?

jdubois: I'm guessing the higher version, the better

generators/statistics.js
agaspardcilia: https://github.com/sindresorhus/conf

agaspardcilia: It's a global config that is used by yeoman. It's supposed to work pretty much on every common OS.

https://github.com/sindresorhus/conf

generators/import-jdl/index.js
agaspardcilia: I removed the implementation and I should have removed this call too. I got some issues with asking the user if he agrees to send the report as user inputs are just promise. It just gets swallowed by the flow of other terminal output, I just need the find out how to make the input a blocking instruction.

cli/commands.js
deepu105: Ok lets keep the sub-gen but rename it to `link-account` and we can have one more question after opt-in asking `Would you like to link your JHipster Online account, to get access to and manage your own stats` if the user says yes compose this sub-gen there and get credentials and do the link

generators/generator-base-private.js
deepu105: see my previous comment, I guess it would be better user experience to automate it more

generators/statistics.js
deepu105: yes v4 is better

generators/statistics.js
deepu105: ok

generators/import-jdl/index.js
deepu105: you need to use the done method for that. A phase will not continue unless the done is called

PierreBesson:
I don't really like the name `jhipster linker`, maybe it can be renamed to `jhipster login` same as `heroku login` or if this is not used just for login, maybe simply `jhipster online`.

generators/generator-base-private.js
jdubois: yes we should keep the existing Insight code - I need to check in the code but that's what @agaspardcilia told me he did

PierreBesson:
After reviewing this for a day and adapting JHipster online accordingly. I think this is ready to merge. However I will do a PR to disable the "login" feature for now as we have not fully defined what we are going to do with it.

PR #7936, opened 2018-07-06
React UAA paths and SERVER_API_URL

ruddell:
This PR does the following:
 - Fix React Gateway UAA paths (currently can't register or use user-management)
 - Implements `SERVER_API_URL` for all requests in the React client (through Axios Interceptor)
 - Fixes React app when running on a context-path other than `/`
 - Fixes error notifications on 401 when running on a context-path other than `/`
 - Removes unused/empty register.ts file

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/errorhandler.interceptor.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/errorhandler.interceptor.ts.ejs
@@ -29,7 +29,7 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
     intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
         return next.handle(request).pipe(tap((event: HttpEvent<any>) => {}, (err: any) => {
             if (err instanceof HttpErrorResponse) {
-                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.endsWith('/api/account'))))) {
+                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.includes('/api/account'))))) {
                     if (this.eventManager !== undefined) {
                         this.eventManager.broadcast({name: '<%=angularAppName%>.httpError', content: err});
                     }
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/errorhandler.interceptor.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/errorhandler.interceptor.ts.ejs
@@ -29,7 +29,7 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
     intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
         return next.handle(request).pipe(tap((event: HttpEvent<any>) => {}, (err: any) => {
             if (err instanceof HttpErrorResponse) {
-                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.indexOf('/api/account') === 0)))) {
+                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.endsWith('/api/account'))))) {
                     if (this.eventManager !== undefined) {
                         this.eventManager.broadcast({name: '<%=angularAppName%>.httpError', content: err});
                     }
--- a/generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
@@ -19,6 +19,8 @@
 import axios from 'axios';
 import { getBasePath, Storage } from 'react-jhipster';
 
+import { SERVER_API_URL } from 'app/config/constants';
+
 const TIMEOUT = 1000000; // 10000
 const setupAxiosInterceptors = onUnauthenticated => {
   const onRequestSuccess = config => {
@@ -27,7 +29,7 @@ const setupAxiosInterceptors = onUnauthenticated => {
       config.headers.Authorization = `Bearer ${token}`;
     }
     config.timeout = TIMEOUT;
-    config.url = `${getBasePath().replace(/\/$/, '')}${config.url}`;
+    config.url = `${SERVER_API_URL ? SERVER_API_URL : getBasePath()}${config.url}`;
     return config;
   };
   const onResponseSuccess = response => response;
--- a/generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
@@ -29,7 +29,7 @@ const setupAxiosInterceptors = onUnauthenticated => {
       config.headers.Authorization = `Bearer ${token}`;
     }
     config.timeout = TIMEOUT;
-    config.url = `${SERVER_API_URL ? SERVER_API_URL : getBasePath()}${config.url}`;
+    config.url = `${SERVER_API_URL}${config.url}`;
     return config;
   };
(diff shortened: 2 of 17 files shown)


generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
```diff
@@ -27,7 +29,7 @@ const setupAxiosInterceptors = onUnauthenticated => {
       config.headers.Authorization = `Bearer ${token}`;
     }
     config.timeout = TIMEOUT;
-    config.url = `${getBasePath().replace(/\/$/, '')}${config.url}`;
```
ruddell: This regex was changing the URL from `http://localhost:8080/#/` to `http://localhost:8080/#` but it works without the replace, can put back if needed for something

generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
```diff
@@ -17,13 +17,12 @@
...
 -%>
 const config = {
-  VERSION: process.env.VERSION,
-  SERVER_API_URL: process.env.SERVER_API_URL
+  VERSION: process.env.VERSION
 };
 
 export default config;
 
-export const SERVER_API_URL = '';
```
ruddell: This was declared twice (in `config` and below), moved to just one declaration.  Can put back in `config`, the import in other files would need changes.

generators/client/templates/react/src/main/webapp/index.html.ejs
```diff
@@ -19,7 +19,7 @@
 <!doctype html>
 <html class="no-js" lang="<%= nativeLanguage %>" dir="ltr">
 <head>
-    <base href="/" />
+    <base href="./" />
```
ruddell: This aligns the base href with Angular

generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/errorhandler.interceptor.ts.ejs
```diff
@@ -29,7 +29,7 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
     intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
         return next.handle(request).pipe(tap((event: HttpEvent<any>) => {}, (err: any) => {
             if (err instanceof HttpErrorResponse) {
-                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.indexOf('/api/account') === 0)))) {
+                if (!(err.status === 401 && (err.message === '' || (err.url && err.url.endsWith('/api/account'))))) {
```
ruddell: When running under a context-path such as `/test`, the path doesn't match as `/api/account` is not in the index 0 spot

deepu105: I think `err.url.includes('/api/account')` might be even better so that if the url ends up having any params it still works

generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
deepu105: I faced some issue coz of this in my own application from which I took these code for JH. I forgot the usecase. I guess it was some edge case. So I guess its better to just be safe here

generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
deepu105: its ok

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts.ejs
```diff
@@ -73,7 +73,7 @@ export default () => next => action => {
       } else if (error && error.response) {
         const response = error.response;
         const data = response.data;
-        if (!(response.status === 401 && (error.message === '' || (data && data.path && data.path.indexOf('/api/account') === 0)))) {
+        if (!(response.status === 401 && (error.message === '' || (data && data.path && data.path.endsWith('/api/account'))))) {
```
deepu105: same as before I think includes is better

generators/client/templates/react/src/main/webapp/app/modules/administration/administration.reducer.ts.ejs
```diff
@@ -181,28 +181,28 @@ export default (state: AdministrationState = initialState, action): Administrati
 <%_ if (applicationType === 'gateway' && serviceDiscoveryType) { _%>
 export const gatewayRoutes = () => ({
   type: ACTION_TYPES.FETCH_GATEWAY_ROUTE,
-  payload: axios.get('/api/gateway/routes')
+  payload: axios.get('api/gateway/routes')
```
deepu105: is the `/` automatically added? I didnt check earlier

generators/client/templates/react/src/main/webapp/app/shared/reducers/application-profile.ts.ejs
```diff
@@ -49,5 +49,5 @@ export default (state: ApplicationProfileState = initialState, action): Applicat
 
 export const getProfile = () => ({
   type: ACTION_TYPES.GET_PROFILE,
-  payload: axios.get('/management/info')
```
deepu105: whats with the `register.ts.ejs` file change above?

generators/client/templates/react/src/main/webapp/app/modules/administration/administration.reducer.ts.ejs
ruddell: Having the path start with `/` will make a request to the base domain/IP instead of the relative path.  

For example, app is running on `http://jhipster.com/sample-app/` 
- Requesting `/api/account` routes to `http://jhipster.com/api/account`
- Requesting `api/account` routes to `http://jhipster.com/sample-app/api/account`

generators/client/templates/react/src/main/webapp/app/shared/reducers/application-profile.ts.ejs
ruddell: It was an unused and empty file so I removed it

generators/client/templates/react/src/main/webapp/app/config/axios-interceptor.ts.ejs
ruddell: This part of the code seems like a hacky way to do relative links, [`getBasePath()`](https://github.com/jhipster/react-jhipster/blob/f911c0abee3f9f5147051ea09a7e8ee6d287db77/src/util/url-utils.ts#L4) returns the `window.location.href` to get the context path.  Using relative links  (without the leading `/`) achieves the same thing

So with the old code

 - Call to `/api/account`
 - `getBasePath().replace(/\/$/, '')` returns `jhispter.com/sample-app`
 - Request is made to `/api/account` so URL is built as `jhispter.com/sample-app/api/account`

Now:
- Call to `api/account`
- Request is made to `api/account` so URL is built as `jhispter.com/sample-app/api/account`

We use relative links everywhere [in Angular](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/layouts/profiles/profile.service.ts.ejs#L29) and the [docs for SERVER_API_URL](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/react/webpack/webpack.common.js.ejs#L114-L115) state it should end in `/` or it will use relative links.

deepu105: ok then

deepu105:
@ruddell is there any issue that is being closed by this?

ruddell:
I didn’t open an issue, the main bug fixed was React UAA gateway functionality

PR #7962, opened 2018-07-12
React: add icons

agaspardcilia:
This icons were imported but never actually added to the icon library.

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
@@ -76,6 +76,8 @@ export const loadIcons = () => {
     faFlag,
     faBell,
     faHome,
+    faRoad,
+    faTimesCircle,
     faRoad
   );
 };
--- a/generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
@@ -78,6 +78,6 @@ export const loadIcons = () => {
     faHome,
     faRoad,
     faTimesCircle,
-    faRoad
+    faSearch
   );
 };


generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
```diff
@@ -76,6 +76,8 @@ export const loadIcons = () => {
     faFlag,
     faBell,
     faHome,
+    faRoad,
```
vishal423: isn't this duplicate of last entry?

agaspardcilia: It is, I meant to add `faSearch`. :sweat_smile:

generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs
```diff
@@ -76,6 +76,8 @@ export const loadIcons = () => {
     faFlag,
     faBell,
     faHome,
-    faRoad
+    faRoad,
+    faTimesCircle,
+    faSearch
```
vishal423: shouldn't this be conditional as it's used only when elasticsearch is used

agaspardcilia: I don't know, I don't think it's really that important. I mean do we really need to start making that file complicated just to save a few kb. wdyt?

deepu105: I guess one icons doesn't harm that much, we can start adding conditions if there are many

deepu105:
@dcaravana where is it used? would you like to do a PR to add it

dcaravana:
Sorry, it's used as icon of Accounts/Sessions menu item.
I'd love to do a PR but I'm afraid I'll not have the time to do so, let me know if you need that anyway.

Issue #7925, opened 2018-07-04
Blueprint write phase, ENOENT: no such file or directory


pasqualevitale:
##### **Overview of the issue**

The bug is related to the usage of implemented blueprint, this case is about spring-controller sub-generator, but it seems the same for the others sub-generator.
When you simply implements: 
```
get writing() {
     // Here we are not overriding this phase and hence its being handled by JHipster
     return super._writing();
}
```
during the app generation with this blueprint, in `_writing` is thrown 
```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: ERROR! Copying template src/main/java/package/web/rest/Resource.java.ejs failed. [Error: ENOENT: no such file or directory, open 'D:\tests\b-helloworld\generators\spring-controller\templates\src\main\java\package\web\rest\Resource.java.ejs']
    at Environment.error (D:\tests\jhipster\generator-jhipster\node_modules\yeoman-environment\lib\environment.js:157:40)
    at module.exports.error (D:\tests\jhipster\generator-jhipster\generators\generator-base.js:1981:18)
    at ejs.renderFile (D:\tests\jhipster\generator-jhipster\generators\utils.js:197:23)
    at tryHandleCache (D:\tests\jhipster\generator-jhipster\node_modules\ejs\lib\ejs.js:254:14)
    at Object.exports.renderFile (D:\tests\jhipster\generator-jhipster\node_modules\ejs\lib\ejs.js:482:10)
    at Object.renderContent (D:\tests\jhipster\generator-jhipster\generators\utils.js:193:9)
    at module.exports.template (D:\tests\jhipster\generator-jhipster\generators\generator-base-private.js:636:23)
    at module.exports.writing (D:\tests\jhipster\generator-jhipster\generators\spring-controller\index.js:138:22)
    at Object.<anonymous> (D:\tests\jhipster\generator-jhipster\node_modules\yeoman-generator\lib\index.js:399:25)
    at D:\tests\jhipster\generator-jhipster\node_modules\run-async\index.js:25:25
```
The generator tries to open `D:\tests\b-helloworld\generators\spring-controller\templates\src\main\java\package\web\rest\Resource.java.ejs`
but it should tries to open `D:\tests\jhipster\generator-jhipster\spring-controller\templates\src\main\java\package\web\rest\Resource.java.ejs`

To workaround the problem I have to copy the templates into my blueprint folder templates.

[...]

PR #7971, opened 2018-07-15
blueprint: fix controller and service

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/7925

_____

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2454,7 +2454,7 @@ module.exports = class extends PrivateBase {
      */
     getAllJhipsterConfig(generator = this, force) {
         let configuration = generator.config.getAll() || {};
-        if ((force || !configuration.baseName) && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
+        if (!configuration.baseName && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
             configuration = JSON.parse(fs.readFileSync('.yo-rc.json', { encoding: 'utf-8' }))['generator-jhipster'];
         }
         if (!configuration.get || typeof configuration.get !== 'function') {
--- a/generators/spring-controller/index.js
+++ b/generators/spring-controller/index.js
@@ -136,15 +136,25 @@ module.exports = class extends BaseGenerator {
                     this.log(chalk.green(`adding ${action.actionMethod} action '${action.actionName}' for /api/${this.apiPrefix}/${action.actionPath}`));
                 });
 
-                this.template(
-                    `${SERVER_MAIN_SRC_DIR}package/web/rest/Resource.java.ejs`,
-                    `${SERVER_MAIN_SRC_DIR}${this.packageFolder}/web/rest/${this.controllerClass}.java`
-                );
-
-                this.template(
-                    `${SERVER_TEST_SRC_DIR}package/web/rest/ResourceIntTest.java.ejs`,
-                    `${SERVER_TEST_SRC_DIR}${this.packageFolder}/web/rest/${this.controllerClass}IntTest.java`
-                );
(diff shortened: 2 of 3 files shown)


generators/generator-base.js
```diff
@@ -2454,7 +2454,7 @@ module.exports = class extends PrivateBase {
      */
     getAllJhipsterConfig(generator = this, force) {
         let configuration = generator.config.getAll() || {};
-        if ((force || !configuration.baseName) && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
+        if (!configuration.baseName && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
```
pascalgrimaud: Without this change, some configurations are undefined when using blueprint, as it will use configuration from `generator-jhipster` instead of `generator-jhipster-helloworld`

Not sure if the `.yo-rc.json` is correct:

```yaml
{
  "generator-jhipster-helloworld": {
    "promptValues": {
    },
    "jhipsterVersion": "5.1.0",
    "baseName": "jhipster",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "jwtSecretKey": "60cdf1f2fc250f90434dc2865264445c4f6cef16",
    "useSass": false,
    "clientPackageManager": "yarn"
  },
  "generator-jhipster": {
    "promptValues": {
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "5.1.0",
    "baseName": "jhipster",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en"
    ],
    "blueprint": "generator-jhipster-helloworld"
  }
}
```

deepu105: actually, both values are needed,  so I think we can revert this change since I fixed this in a different way see https://github.com/jhipster/generator-jhipster/commit/9b6e29156885fa9e6e078e48e69d056ad390371d

generators/spring-controller/index.js
```diff
@@ -136,15 +136,25 @@ module.exports = class extends BaseGenerator {
...
+                if (useBlueprint) {
+                    this.template(
+                        `${this.fetchFromInstalledJHipster('spring-controller/templates')}/${SERVER_MAIN_SRC_DIR}package/web/rest/Resource.java.ejs`,
+                        `${SERVER_MAIN_SRC_DIR}${this.packageFolder}/web/rest/${this.controllerClass}.java`
+                    );
+                    this.template(
+                        `${this.fetchFromInstalledJHipster('spring-controller/templates')}/${SERVER_TEST_SRC_DIR}package/web/rest/ResourceIntTest.java.ejs`,
+                        `${SERVER_TEST_SRC_DIR}${this.packageFolder}/web/rest/${this.controllerClass}IntTest.java`
+                    );
+                } else {
```
deepu105: @pascalgrimaud actually I think it will work without the else part as well, since with or without blueprint the `this.fetchFromInstalledJHipster` will provide the correct path to templates

pascalgrimaud: ok, we should give a try without

pascalgrimaud: you're right @deepu105 !
No need to keep the condition if/else

deepu105: awesome

deepu105: so lets merge the PR then

Issue #7977, opened 2018-07-16
provide clear instructions for users behind corporate proxy


62mkv:
##### **Overview of the issue**

When trying to create JHipster application by running `jhipster`, the following WARNING occurs:

```WARNING! Failed to connect to "git://github.com"
1. Check your Internet connection.
2. If you are using an HTTP proxy, try this command: git config --global url."https://".insteadOf git://
```

I've already set this option, and the command `git ls-remote git://github.com/octocat/Spoon-Knife.git HEAD` is actually passing OK on my laptop, it just takes >75 seconds here and, probably, timeout value is less than that, thus failing the command. 

##### **Motivation for or Use Case**

I would expect to have more information here, like: how critical is this? will it somehow affect JHipster functionality later on ? what are the possible workarounds ? may be I could modify timeout value somewhere ? or set it as a option when running JHipster commands ?

##### **Reproduce the error**

1. Come into our office (just kidding; actually just setup some shitty corporate proxy with bad response time for https)
2. Set up necessary Git config (`http.proxy <proxyURL>; url."https://".insteadOf git://`)
3. Run `jhipster` in an empty folder 

##### **JHipster Version(s)**

5.1.0 

##### **JHipster configuration**

the default one, I guess

##### **Browsers and Operating System**

Windows 10 x64

- [x] Checking this box is mandatory (this is just to show you read everything)

jdubois:
The documentation is at https://www.jhipster.tech/configuring-a-corporate-proxy/

BTW I'm going to remove the Bower part in the documentation, that's obsolete.

If you are being a proxy, you'll probably have trouble with Maven and NPM, so a lot of stuff won't work, but that depends on what you doing in your company. Many companies have internal Maven/NPM mirrors, for example.

62mkv:
this documentation does not mention `git config url."https://"` part, by the way. also, proxy environment variables have to be specified as well, at least I could not find another way to get "webdriver-manager update" running. I understand that as JHipster is a mixed soup of all kind of tools, may be it's not exactly our concern here, but as users struggle with this when trying to use JHipster itself, I guess it would be great to give users more details and transparency. Link to the documentation about proxy configuration (displayed when aforementioned WARNING occurs) would also be great.

62mkv:
and I still don't understand - why and how is JHipster using "git://" connection to GitHub ? Is it for JHipster Marketplace ?

pascalgrimaud:
When using JHipster, we check the git connection:
[...]

PR #7985, opened 2018-07-16
Remove git connection check, used for Bower with AngularJS

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/7977

_____

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -787,22 +787,6 @@ module.exports = class extends Generator {
     }
 
     /**
-     * Check if git connection can be established
-     */
-    checkGitConnection() {
-        if (!this.gitInstalled) return;
-        const done = this.async();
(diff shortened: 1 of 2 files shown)


generators/generator-base-private.js
```diff
@@ -786,22 +786,6 @@ module.exports = class extends Generator {
         });
     }
 
-    /**
-     * Check if git connection can be established
-     */
-    checkGitConnection() {
```
sendilkumarn: Though, we might not use this directly. But this ensures that the user is having `git` and we can execute `git` based comments. I think it is better to have this check and not to relay on the internet to check

pascalgrimaud: the function checkGit already exists and is just above :p

sendilkumarn: 😬

PR #8011, opened 2018-07-22
Improved Jest service tests Angular and update form

Tcharl:

Improvements on Jest test using async await and mocks only verifying fields (not mocking angular http router because it's on the angular side ;-)).

- [X ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green


generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.ts.ejs
```diff
@@ -201,17 +210,4 @@ export class <%= entityAngularName %>UpdateComponent implements OnInit {
         return option;
     }
     <%_ } _%>
-    get <%= entityInstance %>() {
-        return this._<%= entityInstance %>;
-    }
-
-    set <%= entityInstance %>(<%= entityInstance %>: I<%= entityAngularName %>) {
```
deepu105: how is this related to e2e improvements, and if the moment stuff is not done in setter does it work for all cases?

generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs
```diff
@@ -86,61 +86,55 @@ describe('<%= entityClass %> e2e test', () => {
...
-            const relationshipFieldName = relationship.relationshipFieldName; _%>
-        <%_ if (relationshipType === 'many-to-one' || (relationshipType === 'one-to-one' && ownerSide === true)) { _%>
-        <%= entityInstance %>UpdatePage.<%=relationshipName %>SelectLastOption();
-        <%_ } else if ((relationshipType === 'many-to-many' && ownerSide === true)) { _%>
-        // <%= entityInstance %>UpdatePage.<%=relationshipName %>SelectLastOption();
-        <%_ } _%>
-        <%_ }); _%>
-        <%= entityInstance %>UpdatePage.save();
-        expect(<%= entityInstance %>UpdatePage.getSaveButton().isPresent()).toBeFalsy();
+        <%= entityInstance %>ComponentsPage.clickOnCreateButton().then (
```
deepu105: use async/await where possible

Tcharl: As protractor made its own promise wrappers, I choose to use them as I think they're made for something. But I could have been wrong ;-)

generators/entity-client/templates/angular/src/test/javascript/spec/app/entities/entity-management.service.spec.ts.ejs
```diff
@@ -42,16 +59,56 @@ describe('Service Tests', () => {
...
+                            body: returnedFromService
+                        })
+                    )
+                );
+                const rxPromise = service
+                    .find(<%- tsKeyId %>)
+                    .pipe(take(1))
+                    .toPromise();
+                expect.assertions(1);
+                return expect(rxPromise).resolves.toMatchObject({body: elemDefault});
```
Tcharl: here I followed the [jest doc on async](https://jestjs.io/docs/en/tutorial-async), and as we don't use async/await nowhere in the angular code I choose to use promise. Do you prefer I change the implementation?

generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs
deepu105: I have been using async/await with protractor for quite some time and didnt have any issues, so unless there is any recommendation from them to use promise instead of asyc, i would prefer to stick with async/await as they are neater and easier to read

generators/entity-client/templates/angular/src/test/javascript/spec/app/entities/entity-management.service.spec.ts.ejs
deepu105: yes, would prefer async/await. I did that for React but angular code was bit old and we were waiting for upgrade to node 8

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.ts.ejs
deepu105: ok

deepu105:
@Tcharl also please look at #8026

deepu105:
@Tcharl can you please fix conflicts and take care of the comments? also please switch to async/await where possible

Tcharl:
Sure, I'll work on it Sunday and next weekend

Tcharl:
Hi @deepu105 ,

I updated the PR rebasing, using async/await and fixing everything (I guess)

Tcharl:
Sorry, but I don't see what test to add to this PR: these tests are generated and executed on the travis suite ^^

deepu105:
@Tcharl that label is for others in the team to do a testing of the PR before merge

deepu105:
@jhipster/developers can some one test and merge this please @sendilkumarn @wmarques @agaspardcilia @chrisdns @ruddell

wmarques:
@Tcharl Why are you using `  spyOn(httpClient, 'get')` instead of using the `HttpTestingController` to test the HTTP calls ?

Tcharl:
@wmarques because here we're mocking the calls to the client, not the calls from the client to the protocol.
* Mocking httpClient here only tests the service methods implementations.
* Mocking using httpMock / HttpTestingController would test the service implementation plus angular httpclient internals.

wmarques:
@Tcharl I prefer using the httpMock personally. It's more readable IMO and it's provided by Angular for this use case. 
What do you mean by the `httpclient internals` ? The interceptors ?

Tcharl:
Sorry, but I'm a bit extremist on unit tests, so I won't change it:
Unit test should only test the method you want to test, and, because we have no real solution to tackle this, the transitive private methods calls by that former method. It's the only differentiator between unit vs integration tests.
In that specific case, if we followed your advice, we would test the method call plus the angular HttpClient#<verb> method call, which is located in another class, so it won't be a unit test anymore, but an integration test.

I really don't care about the misunderstanding on the distinction between unit and integration tests paradigm of the Angular community: Unit test should only test the method + transitive private method call and only these ones. Otherwise, I should make a new 'itests' package in the src/itest/javascript/ folder and put the angular services tests there, and maybe make the distinction on coverage reports to have the right Sonarqube informations.

wmarques:
@jhipster/developers WDYT ? I agree that the usage of `HttpClientTestingController` makes our tests not real unit tests as we also go through the HttpClient code, but I find it cleaner and it's the recommended way in the [Angular docs](https://angular.io/guide/http#testing-http-requests).

deepu105:
If I have to be honest, I don't really care if we are also testing some
internal code as long as the tests are cleaner and our test cases are
covered. And when it comes to Angular I really dont care being accurate and
calling them itests. I'll prefer the one with less code and cleaner to read
and understand

Thanks & Regards,
Deepu

Tcharl:
Done, but I felt hit in my craftmanship and principles.
Also, I don't find it either more readable nor natural but we should go the way community voted ;-).

Cheers team, Jhipster rock!

jdubois:
Can we merge this tomorrow? I wanted to do a release now, and that looks like a big change

Tcharl:
Ok for me.

PR #8016, opened 2018-07-24
handle missing translations for min and max validations

### LINKED ISSUE #7305: No translation for min and max validation errors
From https://github.com/mraible/21-points/issues/33

ruddell:
Fix #7305

 - Maps Min/Max/DecimalMin/DecimalMax to the Size i18n key on the client.  
 - Adds test for `alert-error.component.ts` 


vishal423:
@ruddell , can this conversion be done on client side as I believe the issue is due to client limitation to handle this out. API contract shouldn't be driven by client implementation, rather it's other way around. As you also said, elegant solution would be to add missing translations on client side to sync with API error contract.

ruddell:
@vishal423 How does the updated PR look?  I added the Min/Max i18n keys for all languages and made DecimalMin and DecimalMax use those keys instead of Size.  We could add translations for those two as well but it would be the same translation

vishal423:
@ruddell, sorry, but, looking at changes I don't see value add of adding new keys as conveyed message is similar to that of 'size' key. Instead, can this mapping be done on client itself to map to 'size' key instead of min/max/? That way server contract is respected and client side validation messages are reused.

deepu105:
@mraible @vishal423 would you be able to test this

ruddell:
To test, you have to remove the `jhiMin`/`jhiMax` Angular directives or `min`/`max` React validation on the entity update dialog (otherwise you just get client validation).  You can use the below JDL:

```
entity Preferences {
    goal Integer min(10) max(20)
    exactGoal Double min(10) max(20)
}
```

generators/client/templates/angular/src/test/javascript/spec/app/shared/alert/alert-error.component.spec.ts.ejs
```diff
@@ -0,0 +1,137 @@
...
+        let eventManager: JhiEventManager;
+
+        beforeEach(async(() => {
+            TestBed.configureTestingModule({
+                imports: [<%=angularXAppName%>TestModule<% if (enableTranslation) { %>, TranslateModule.forRoot()<% } %>],
+                declarations: [<%=jhiPrefixCapitalized%>AlertErrorComponent],
+                providers: [
+                    JhiEventManager,
+                    {
+                        provide: JhiAlertService,
```
vishal423: do we require prefix variable here?

vishal423:
Overall change looks good to me, however, wouldn't be able to test this out.

generators/client/templates/angular/src/test/javascript/spec/app/shared/alert/alert-error.component.spec.ts.ejs
ruddell: `JhiEventManager` and `JhiAlertService` come from `ng-jhipster` so it always uses the `jhi` prefix

generators/client/templates/react/src/main/webapp/app/config/notification-middleware.ts.ejs
```diff
@@ -101,6 +101,9 @@ export default () => next => action => {
                 const fieldErrors = data.fieldErrors;
                 for (i = 0; i < fieldErrors.length; i++) {
                   const fieldError = fieldErrors[i];
+                  if (['Min', 'Max', 'DecimalMin', 'DecimalMax'].includes(fieldError.message)) {
```
vishal423: would it be possible to add unit test for react as well?

ruddell: Done

ruddell:
Claimed the bug bounty here: https://opencollective.com/generator-jhipster/expenses/4421

PR #8026, opened 2018-07-26
Make Angular e2e tests use async/await

chrisdns:
As the Selenium promise manager will surely be deprecated in the future, we should use async/await to handle promises and it is more readable than .then() imo

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 <div *ngIf="audits">
-    <h2 class="audits-page-heading" jhiTranslate="audits.title">Audits</h2>
+    <h2 id="audits-page-heading" jhiTranslate="audits.title">Audits</h2>
 
     <div class="row">
         <div class="col-md-5">
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 <div *ngIf="audits">
-    <h2 jhiTranslate="audits.title">Audits</h2>
+    <h2 class="audits-page-heading" jhiTranslate="audits.title">Audits</h2>
 
     <div class="row">
         <div class="col-md-5">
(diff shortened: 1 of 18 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
```diff
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 <div *ngIf="audits">
-    <h2 jhiTranslate="audits.title">Audits</h2>
+    <h2 class="audits-page-heading" jhiTranslate="audits.title">Audits</h2>
```
deepu105: use id instead of class as these are unique names

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management.component.html.ejs
```diff
@@ -18,7 +18,7 @@
 -%>
 <div>
     <h2>
-        <span jhiTranslate="userManagement.home.title">Users</span>
+        <span class="userManagement-page-heading" jhiTranslate="userManagement.home.title">Users</span>
```
deepu105: use dash casing for ids and classes

generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
```diff
@@ -33,178 +34,159 @@ describe('account', () => {
     let settingsPage: SettingsPage;
     <%_ } _%>
 
-    beforeAll(() => {
-        browser.get('/');
-        browser.waitForAngular();
+    beforeAll(async () => {
+        await browser.get('/');
         navBarPage = new NavBarPage(true);
-        browser.waitForAngular();
```
deepu105: isnt this angular specific? protractor does some special handling for angular so is it ok to remove it?

chrisdns: the protractor documentation (https://www.protractortest.org/#/api?view=ProtractorBrowser.prototype.waitForAngular) says that waitForAngular() is already called before each action that's why I removed it, this does not seem to create any trouble

generators/client/templates/angular/tslint.json.ejs
```diff
@@ -117,7 +117,13 @@
...
         "no-consecutive-blank-lines": [true],
         "object-literal-shorthand": true,
-        "space-before-function-paren": [true, "never"],
+        "space-before-function-paren": [true, {
+            "asyncArrow": "always",
+            "anonymous": "never",
+            "constructor": "never",
+            "method": "never",
+            "named": "never"
+        }],
```
chrisdns: @deepu105 I have modified the tslint config in order to have

deepu105: yup its fine

Issue #8033, opened 2018-07-27
Feature: conditional dev webpack builds on maven/gradle run


deepu105:
##### **Overview of the feature request**

This is an idea I got while tinkering with some stuff. Currently, we do a webpack build as soon as app is generated so that when you run `mvnw` or `gradlew` you get a working app. Though this is the way we have been doing it for a while it has some disadvantages
1. If the initial webpack build fails for some reason, you get a blank page when starting the server, and this results in many invalid tickets. In angular 1, we used to have an error page in this case but we dont have it anymore with Angular/React
2. With Yeoman 3 upgrade the process is more async and it splits our final instruction step logs
3. Not very reliable when your workflow involves clean runs

I propose the below alternative

During maven/gradle dev profile builds we can always trigger a webpack dev build conditionally depending on the presence of the `target|build/www` folder(and maybe even the main assets inside it)
This is pretty easy to do in Gradle and I'm sure shouldn't be difficult in maven as well (@jdubois @pascalgrimaud correct me if I'm wrong)

So this way the webpack build will be run when you run `mvnw|gradlew` for the first time and whenever you do a clean run making sure the app doesn't give a blank page

No changes to prod builds as it always runs the webpack build

##### **Motivation for or Use Case**

Better development experience

##### **Related issues or PR**

- [ x] Checking this box is mandatory (this is just to show you read everything)

mraible:
+1 to anything that helps solve the blank page issue.

deepu105:
I can do the gradle part, but would appreciate help with Maven any volunteers @jhipster/developers ?

pascalgrimaud:
I thought about this some months ago, but never had the time to investigate.
So +1

atomfrede:
@deepu105 Great idea, because I had the exact same problem the last couple of days. Created a fresh project, yarn fails, need to do it manually. For first time users not great

vishal423:
do you also plan to install UI dependencies (with --force) during clean install?

jdubois:
With AngularJs 1 we used to have a nice error page that you did @deepu105 - it would be awesome to have something similar

pascalgrimaud:
@deepu105 : what is your exact idea when the `target|build/www` is missing ?
- have a error page like AngularJS
- or trigger the webpack profile: install yarn/npm, launch webpack:dev build

deepu105:
[...]

PR #8038, opened 2018-07-28
Improve webpack maven profile

pascalgrimaud:
fix https://github.com/jhipster/generator-jhipster/issues/8033

_____

### DIFF
--- a/generators/server/templates/gradle/profile_dev.gradle.ejs
+++ b/generators/server/templates/gradle/profile_dev.gradle.ejs
@@ -40,7 +40,8 @@ bootRun {
 }
 
 <%_ if (!skipClient) { _%>
-task webpackBuildDev(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task) {
+task webpackBuildDev(type: <%= _.upperFirst(clientPackageManager) %>Task, dependsOn: '<%= clientPackageManager %>_install') {
+    onlyIf { shouldWebpackRun() == true }
     args = ["run", "webpack:build"]
 }
 
@@ -71,5 +72,9 @@ processResources {
 
 <%_ if (!skipClient) { _%>
 processResources.dependsOn webpackBuildDev
-webpackBuildDev.onlyIf { project.hasProperty('webpack') }
+yarn_install.onlyIf { shouldWebpackRun() == true }
+
+def shouldWebpackRun() {
+    file('build/www/app/global.bundle.js').exists() == false || project.hasProperty('webpack')
+}
 <%_ } _%>
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1155,7 +1155,7 @@
         </profile>
         <%_ if (!skipClient) { _%>
         <profile>
-            <id>webpack-dev</id>
+            <id>webpack</id>
             <activation>
                 <file>
                     <missing>${basedir}/target/www/app/global.bundle.js</missing>
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -429,7 +429,7 @@ module.exports = class extends BaseGenerator {
                     this.log(`\n${chalk.bold.red('Running post run module hooks failed. No modification done to the generated app.')}`);
                     this.debug('Error:', err);
                 }
-                this.log(chalk.white(`If you find JHipster useful consider supporting our collective ${chalk.yellow('https://opencollective.com/generator-jhipster')}`));
+                this.log(chalk.green(`\nIf you find JHipster useful consider supporting our collective ${chalk.yellow('https://opencollective.com/generator-jhipster')}`));
             }
         };
     }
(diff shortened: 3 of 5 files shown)


PierreBesson:
As I explained in the issue discussion, I think we should rename the `webpack` profile to `dev-webpack` as this totally supersede the default dev profile.

generators/server/templates/pom.xml.ejs
```diff
@@ -1155,7 +1155,29 @@
         </profile>
         <%_ if (!skipClient) { _%>
         <profile>
-            <id>webpack</id>
+            <id>dev-webpack</id>
```
deepu105: to be consistent letc call it `webpack-dev`

deepu105: btw how is this triggered in default dev profile?

deepu105: originally you had to pass -Pdev,webpack to trigger this, ideally it should just work with -Pdev and with default `./mvnw`

pascalgrimaud: Renamed to webpack-dev

If the `target/www` does't exist, it will use the new `webpack-dev` maven profile, using `dev` spring boot profile (so dev properties).
If it exists, it will use the default `dev` maven profile, using `dev` spring boot profile.

Tested with:
- mvn clean, mvn, stop and start mvn
- mvn clean, mvn -Pdev, stop and start mvn -Pdev

generators/server/templates/gradle/profile_dev.gradle.ejs
```diff
@@ -40,7 +40,8 @@ bootRun {
 }
 
 <%_ if (!skipClient) { _%>
-task webpackBuildDev(type: <%= clientPackageManager.charAt(0).toUpperCase() + clientPackageManager.slice(1) %>Task) {
+task webpackBuildDev(type: <%= _.upperFirst(clientPackageManager) %>Task, dependsOn: '<%= clientPackageManager %>_install') {
+    onlyIf { shouldWebpackRun() == true }
```
deepu105: @atomfrede is this good or is there a better way to do this?

deepu105:
@pascalgrimaud so for maven does forcing webpack(current behavior) also work regardless of the presence of assets like `./mvnw -Pwebpack`
For gradle it works

generators/server/templates/gradle/profile_dev.gradle.ejs
atomfrede: Looks ok to me at first sight. Pretty sure one could do something with input and output files and let gradle do the magic to decide wether to run that task or not, but I think thats something for the jhipster plugin (shame on me I didn't had time)

pascalgrimaud:
@deepu105 : the `webpack` maven profile doesn't exist anymore. It has been renamed to `webpack-dev`. If you use `./mvnw -Pwebpack-dev`, it will work, regardless the presence of asserts.

generators/server/templates/gradle/profile_dev.gradle.ejs
deepu105: I thought of in/out files but figured we would have less control over the behaviour as it would run on change detection as well

deepu105:
@pascalgrimaud what was the reason to rename from `webpack` to `webpack-dev` again? I lost track of it. Is it possible to retain the name since its what we have documented

pascalgrimaud:
@deepu105 : because Pierre and you, suggested me this change. I can change back to webpack, so our documentation won't change.

deepu105:
read the commets again, I was suggesting using webpack-dev instead of dev-webpack I guess, But I still dont understand why @PierreBesson suggested the original change

pascalgrimaud:
Let's change back to keep the original maven profile: `webpack`. It can't confuse user as it's for advanced users.
I'm commiting the change

deepu105:
ok lets do this. @mraible can you check this out and provide a feedback

mraible:
@deepu105 Works great! Nice job.

BTW, is it possible to suppress output like the following? It really seems to clutter up the logs locally and in CI.

```
[INFO]     → [95%] 663 of 697 modules :: node_modules/moment/moment.js ~ external
[INFO]     → [95%] 664 of 697 modules :: node_modules/moment/moment.js ~ external
[INFO]     → [95%] 665 of 697 modules :: node_modules/moment/moment.js ~ external
[INFO]     → [96%] 666 of 697 modules :: node_modules/moment/moment.js ~ external
[INFO]     → [95%] 666 of 698 modules :: src/main/webapp/app/home/home.css ~ internal
[INFO]     → [96%] 667 of 698 modules :: node_modules/moment/moment.js ~ external
```

PierreBesson:
The reason I suggested renaming to *dev-webpack* is because it is nothing like the old *webpack* profile (which you needed to run with `mvn -Pdev,webpack`) as we now copied everything from the dev profile into it but it's not super important.

PierreBesson:
I just checked and `mvn -Pdev,webpack` still works however the behavior for `mvn -Pwebpack` has changed, in the past it would compile only the front-end and now it is compiling both.

deepu105:
@PierreBesson then I would say old behavior was not what I originally intended. The new behavior is correct. Its the same in Gradle as well.

PR #8056, opened 2018-08-01
fix issue with MAVEN_CONFIG

lmbai:
Fix the issue with `MAVEN_CONFIG` setting. 

- For `mvnw`:  we set the `MAVEN_CMD_LINE_ARGS` but we didn't use it
- For `mvnw`: we use the `MAVEN_CMD_LINE_ARGS`, but ignore the `MAVEN_CONFIG`

it will have issue when we depends on the `MAVEN_CONFIG`: eg: use pipline-maven-plugin


lmbai:
currently the fix here will solve the issue. 
another solution: should we use the (maven-wrapper)[https://github.com/takari/maven-wrapper] to regenerate the `mvnw` and `mvnw.cmd`. 
the new generated file also cater for the `MAVEN_CONFIG` env.

PierreBesson:
@lmbai I believe the maven wrapper was initially with the **takari/maven-wrapper**, you should regenerate it with the latest version and add this to your PR not modify the script by hand.

lmbai:
@PierreBesson @deepu105 updatd with the generated file with latest `maven-wrapper`

deepu105:
@lmbai doesn't the maven wrapper version and binary needs to be updated as well?

lmbai:
@PierreBesson @deepu105 updated the `.mvn` folder. 
it seems like there is a new file `MavenWrapperDownloader ` needed for the new `mvnw` script. 
now all the related files has been updated.

generators/server/templates/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,110 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
```
deepu105: you need to add this file to the server files section so that it is copied when generating anew project. but do we really need this file since the wrapper binary is included already? @PierreBesson WDYT

lmbai: I not sure this part.  It just provide a new way to download the jar file. 
Try wget , try curl , try this downloaded. 
Yeah, we have the jar file already, this may not needed. But I thinks it would be better to keep the original logic of the maven-wrapper. WDUT?

lmbai:
the above `Travis Ci` error was due to a timeout error. i am not sure how to re-try it.

PierreBesson:
I have restarted the build. I'm ok to add this new downloader.java but ideally I would prefer if it would be possible to not add the binary and to have it downloaded from the script (I believe the gradle wrapper works this way).

deepu105:
@PierreBesson I think we include the gradle binary as well, it only downloads if its not present I think

lmbai:
The Travis test failed due to the E2E timeout again.

PierreBesson:
@deepu105, Ok I'm going to have a look to see if we can remove the binary from our code... Will also with gradle.

pascalgrimaud:
I'm not sure to understand what this PR should do.
Why don't we keep things simple and just use the new `mvnw` and `mvnw.cmd` from https://github.com/takari/maven-wrapper ?

lmbai:
@PierreBesson thanks for the info. Actually i don't aware that we need to add it into the `files.js`...  
Just added it.

Btw: there is another pr at [https://github.com/jhipster/jhipster-console/pull/98](https://github.com/jhipster/jhipster-console/pull/98). pls help to take a look. 
we are upgrading our local jhipster consoles. the current `jhipster-alerter:v3.x` doesn't work with the `jhipster-elasticsearch:v3.x` due to the ES6 upgrade.

lmbai:
@pascalgrimaud , yeah, that what we thought at the begin, but the in the new `mvnw`, it will use the new file called `MavenWrapperDownloader` to download the jar. 
the logic is as follows:

1. the jar is there? end
1. if the jar is not there.
    1. use the `wget` to download the jar. 
    1. if fail, use `curl` to download the jar
    1. if fail, use the `MavenWrapperDownloader` to download the jar

so the result is if we use the new `mvnw`, we need the `MavenWrapperDownloader.java`.  the current unsure thing is should we include the `maven-wrapper.jar` or not. 

1. if we include it. then nothing will happen.
1. if we don't include it. the `mvnw` will try to download with the above logic.

pascalgrimaud:
Oh got it!
I thought you coded this class! -> https://github.com/takari/maven-wrapper/blob/master/.mvn/wrapper/MavenWrapperDownloader.java

PierreBesson:
Okay, so it seems that we could remove the `maven-wrapper.jar`. However for gradle there is no way to remove the jar, it is explicitly specified in the docs that it's mandatory: https://docs.gradle.org/current/userguide/gradle_wrapper.html

However for maven, I would like to remove the jar, if you all agree ?

lmbai:
I am open with both options.

- Keep the jar file, so that user don't need to download it.  it will benefit the people who are behind the proxy. but if they are behind the proxy, they need to solve the `apache-maven-3.5.4-bin.zip` issue anyway, because the `maven-wrapper` will download it anyway.
- Remove the jar file, the `mvnw` will download it automatically.

lmbai:
Anyway, i make this PR try to solve the `MAVEN_CONFIG` problems. not for the file structure. haha.

The issue i mentioned in the [https://github.com/jhipster/jhipster-console/pull/98](https://github.com/jhipster/jhipster-console/pull/98) is more critical.  Please take a look.

jdubois:
From a license perspective, I'm quite annoyed that we add a file from the ASF into our code. I understand it's meant to work this way, and this is caused by the ASF itself. Anyway, we have the same Apache 2 license, and we keep the copyright headers from the ASF, so that should be fine, but I was more confortable if the binary file.

PierreBesson:
Oh it is actually not copyrighted to the ASF, it should be copyrighted to the authors of https://github.com/takari/maven-wrapper. They are just too lazy to fill out their names in the license. But similarly to JHipster it is a generated file and under the same license so I'm sure there is no problem. And by the way, the binary is also subject to the license anyway.

It might be useful to add a NOTICE.txt file to jhipster to credit some of the upstream project we use: https://www.apache.org/dev/apply-license.html

jdubois:
If it's generated, the issue is that then the license is lost -> we do it on purpose for our code (it becomes the users' code), but I don't think the Maven wrapper authors intended that.
In that case, we should also generate their license header. That should solve the whole issue.

Concerning the NOTICE file we already have one for the project, and yes we could improve it (I mostly added this because "some" projects copy our templates and don't credit us). But for generated projects the issue is the same, as that file shouldn't be generated in the final projects (again, they belong the users).

So let's generate that file with the original header, that will solve the issue for me.

jdubois:
Thanks, I'm merging it!

PR #8069, opened 2018-08-03
Move to Terser plugin instead of UglifyJS

sendilkumarn:
It is fast, can see ~50% speed up in (angular) prod build 🤞

Lets see what travis says


sendilkumarn:
🎉 thanks @PierreBesson

PR #8073, opened 2018-08-04
add more test coverage for react

ruddell:
This gets all of the reducer specs to near 100%

@deepu105 The only component tests for React are for the Navbar, the other tests are service and reducer tests.  Is this something we want?


generators/client/templates/react/src/test/javascript/spec/app/shared/reducers/user-management.spec.ts.ejs
```diff
@@ -1,99 +0,0 @@
-<%#
```
ruddell: This file wasn't copied into the client, there is another user-management.spec.ts file [here](https://github.com/jhipster/generator-jhipster/blob/f1c05547f556c371adb3832fc2e7ed0bdf5b4919/generators/client/templates/react/src/test/javascript/spec/app/modules/administration/user-management/user-management.reducer.spec.ts.ejs)

vishal423: I may be wrong, but, think this was for OIDC flow.

ruddell: Can't believe I missed that

deepu105:
@ruddell ofcourse it would be nice to have more component tests

PR #8081, opened 2018-08-09
update gradle prod profile with h2 testCompile dependency

### LINKED ISSUE #7888: Gradle prod build: error: package org.h2.server.web does not exist

In JHI4 I used this command to build my gradle powered application using 

``` bash
./gradlew -P prod build buildDocker
``` 

which always did a great job, as it calls all required task, running tests and so on.

With JH5 this doesn't work anymore. It looks like h2 is compiled only for the default gradles dev profile, not when prod is enabled.

A workaround to this would be to 

```
./gradlew -x test -P prod build buildDocker
```

or using CI cache/artifacts to do it in several steps, like

```
./gradlew build
./gradlew test
./gradlew -P prod bootWar buildDocker
```

I think, that behavior should not be like this.

I want to do an entire build, test, dockerization in one command for the most simple CI/dev setup.

##### **Reproduce the error**

Generate a (microservice) application with gradle chosen as build-tool. Run the above command and you will see something like this:

```
> Task :compileTestJava FAILED
/home/david/jhipster/samples/bookstore/bookservice/src/test/java/com/mycompany/myapp/config/WebConfigurerTest.java:12: error: package org.h2.server.web does n
ot exist
import org.h2.server.web.WebServlet;
                        ^
/home/david/jhipster/samples/bookstore/bookservice/src/test/java/com/mycompany/myapp/config/WebConfigurerTest.java:78: error: cannot find symbol
        verify(servletContext, never()).addServlet(eq("H2Console"), any(WebServlet.class));
                                                                        ^
  symbol:   class WebServlet
  location: class WebConfigurerTest
/home/david/jhipster/samples/bookstore/bookservice/src/test/java/com/mycompany/myapp/config/WebConfigurerTest.java:90: error: cannot find symbol
        verify(servletContext).addServlet(eq("H2Console"), any(WebServlet.class));
                                                               ^
  symbol:   class WebServlet
  location: class WebConfigurerTest
3 

* What went wrong:
Execution failed for task ':compileTestJava'.
> Compilation failed; see the compiler error output for details.

````

##### **Suggest a Fix**

The most straightforward one would be include be making h2 compile in general, and not just for dev profile. But this would add h2 to all prod builds as well. 

A better way would be to figure out, how to **not** add h2 in prod, except we are running compileJavaTest

##### **JHipster Version(s)**

JHipster 5.0.1

### PR DESCRIPTION
**Pull request by joewhite101**
add h2 to prod profile for tests 

so that this command ./gradlew -P prod build buildDocker works

Fix #7888

I don't think this is testable other than adding the above build command to the CI scripts to ensure the prod profile tests always run correctly.  I also, ran the same process with maven `mvn clean package dockerfile:build` and it works correctly as the plugin inherits the test h2 dependency from the main pom section.

### DIFF
--- a/generators/server/templates/gradle/profile_prod.gradle.ejs
+++ b/generators/server/templates/gradle/profile_prod.gradle.ejs
@@ -23,7 +23,7 @@ apply plugin: 'com.moowork.node'
 <%_ } _%>
 
 dependencies {
-
+    testCompile "com.h2database:h2"
 }
 
 def profiles = 'prod'

### CODE REVIEW

**Comment by joewhite101** [comment]
Not sure what is going on with the travis build here. Looks like timeout fetching packages at the yarn step. Maybe just needs a re-run? Doesn't look connected to the change.

**Comment by PierreBesson** [comment]
Travis seems to be really flaky recently. I restarted your failed builds.

**Comment by joewhite101** [comment]
Pretty sure those build failures aren't connected to this change. Not sure if worth another re-run or not. If there is something to investigate here let me know.

**Comment by PierreBesson** [comment]
Yes this is an important fix. I'm merging.

File: generators/server/templates/gradle/profile_prod.gradle.ejs [hunk 1]
```diff
@@ -23,7 +23,7 @@ apply plugin: 'com.moowork.node'
 <%_ } _%>
 
 dependencies {
-
+    testCompile "com.h2database:h2"
```
pascalgrimaud: @PierreBesson @joewhite101 : I think it should be added only when h2 is selected for dev database. Need to be excluded for other database like MongoDB, Cassandra, etc.

joewhite101: @pascalgrimaud Yeah you are correct. May be a few days before I get a chance to look at it but I will create a new PR sometime soon.

pascalgrimaud: thanks @joewhite101 but it should be OK :)
PR already done this morning: https://github.com/jhipster/generator-jhipster/pull/8212

PR #8108, opened 2018-08-18
throw correct exception on bad password for UAA

### LINKED ISSUE #8014: Gateway Incorrectly Reports UAA's Error Status
When calling UAA, the translated exceptions from a login 'bad request' are converted to status 500, when routed through the gateway.

500_should_be_400

As you can see above, in postman, the request's status code is 500, and in the body of the response, there are two conflicting statuses (400 and 500).

I have not investigated other exceptions' conversion, but I suspect the following lines in `ExceptionTranslator`'s call to `AdviceTrait` (interface) to be at play:

`final HttpStatus status = HttpStatus.valueOf(Optional.ofNullable(problem.getStatus())
        .orElse(Status.INTERNAL_SERVER_ERROR)
        .getStatusCode());`

as the following error is logged in the gateway:

`2018-07-23 14:38:27.093 ERROR 56528 --- [ XNIO-2 task-64] o.z.p.spring.web.advice.AdviceTrait      : Internal Server Error`

##### **JHipster Version(s)**

Version 5.1.0

```
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.xxx.uaa"
    },
    "jhipsterVersion": "5.1.0",
    "baseName": "uaa",
    "packageName": "com.xxx.uaa",
    "packageFolder": "com/x/x/x/uaa",
    "serverPort": "9999",
    "authenticationType": "uaa",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "gradle",
    "jwtSecretKey": "xxxxxxx",
    "enableTranslation": false,
    "applicationType": "uaa",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "skipClient": true
  }
}
```

ruddell:
Fix #8014

### DIFF
--- a/generators/server/templates/src/main/java/package/security/oauth2/OAuth2AuthenticationService.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/oauth2/OAuth2AuthenticationService.java.ejs
@@ -18,11 +18,13 @@
 -%>
 package <%=packageName%>.security.oauth2;
 
+import <%=packageName%>.web.rest.errors.InvalidPasswordException;
 import io.github.jhipster.security.PersistentTokenCache;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.web.client.HttpClientErrorException;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
@@ -88,9 +90,9 @@ public class OAuth2AuthenticationService {
                 log.debug("successfully authenticated user {}", params.get("username"));
             }
             return ResponseEntity.ok(accessToken);
-        } catch (Exception ex) {
+        } catch (HttpClientErrorException ex) {
             log.error("failed to get OAuth2 tokens from UAA", ex);
-            throw ex;
+            throw new InvalidPasswordException();
         }
     }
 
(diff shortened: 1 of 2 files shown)


generators/server/templates/src/main/java/package/security/oauth2/OAuth2AuthenticationService.java.ejs
```diff
@@ -88,9 +90,9 @@ public class OAuth2AuthenticationService {
                 log.debug("successfully authenticated user {}", params.get("username"));
             }
             return ResponseEntity.ok(accessToken);
-        } catch (Exception ex) {
```
ruddell: We can throw the correct `InvalidPasswordException` by catching `HttpClientErrorException` instead of a generic `Exception`.  For other errors (no UAA up or connection refused), the behavior is the same as before - the exception is thrown

generators/server/templates/src/main/java/package/security/oauth2/OAuth2AuthenticationService.java.ejs
```diff
@@ -88,9 +90,9 @@ public class OAuth2AuthenticationService {
...
+        } catch (HttpClientErrorException ex) {
             log.error("failed to get OAuth2 tokens from UAA", ex);
-            throw ex;
+            throw new InvalidPasswordException();
```
deepu105: dont we need to pass the ex for stacktrace?

ruddell: The printed stack trace looks mostly the same, main difference is at the end where it returns the correct exception.  

[Logs from before change](https://gist.githubusercontent.com/ruddell/f5a8c406307bf92b231a6a5a93f1c04f/raw/dcd74aa83810534b8d8283060535a8191cc5a53a/gistfile1.txt) 
[Logs from after change](https://gist.githubusercontent.com/ruddell/b6c38a9b84958e5e351563cc72ede523/raw/b4309503930945911eb1f4dd2d071f7abf8d9db1/After%2520Changes%2520StackTrace%2520PR%25208108)

PR #8110, opened 2018-08-18
Create delete method in protractor test

### LINKED ISSUE #7844: Add delete tests to Protractor tests

The current Protractor tests verify that add works, but they don't delete the records. This means that when you run the tests, test data is left around.

Because I like to run e2e tests, but not have the test data lying around after.

https://github.com/jhipster/generator-jhipster/issues/7778

hdurix:
Test to delete last element of the entity list:
- click on last delete button
- check dialog title
- click on confirm button
- check an element has been removed

Fix #7844

### DIFF
--- a/generators/client/templates/react/src/test/javascript/e2e/util/utils.ts.ejs
+++ b/generators/client/templates/react/src/test/javascript/e2e/util/utils.ts.ejs
@@ -1,4 +1,4 @@
-import { ExpectedConditions, ElementFinder, browser, by, element } from 'protractor';
+import { ExpectedConditions, ElementArrayFinder, ElementFinder, browser, by, element } from 'protractor';
 
 const waitUntilDisplayedTimeout = 30000;
 
@@ -37,6 +37,18 @@ export const waitUntilHidden = (selector: ElementFinder, classname = '', timeout
     isHidden(selector),
     timeout,
     `Failed while waiting for "${selector.locator()}" of Page Object Class '${classname}' to be hidden.`
+  );
+};
+
+export const waitForCount = (elementArrayFinder: ElementArrayFinder, expectedCount: number) => () => {
+  return elementArrayFinder.count().then(actualCount => expectedCount === actualCount);
+};
+
+export const waitUntilCount = (elementArrayFinder: ElementArrayFinder, expectedCount: number, timeout = waitUntilDisplayedTimeout) => {
+  browser.wait(
+    waitForCount(elementArrayFinder, expectedCount),
+    timeout,
+    `Failed while waiting for "${elementArrayFinder.locator()}" to have ${expectedCount} elements.`
   );
 };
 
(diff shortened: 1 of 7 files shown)


hdurix:
The PR is in WIP until I add the same tests for react.

@mraible, to answer your [comment](https://github.com/jhipster/generator-jhipster/issues/7844#issuecomment-399669609) on the related issue, I'm not sure we should go further on deleting relationships between entities.

Indeed, we can't know in which order the entities should be deleted. I think that's why, the creation test is disabled when at least one relationship is required: see [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs#L34-L36) and [here](https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs#L89).

What do you think?

mraible:
Can you try out these changes on my [blog.jh](https://github.com/jhipster/jdl-samples/blob/master/blog.jh) JDL? It has relationships in it and the current Protractor tests do create data with relationships.

hdurix:
I created a [project](https://github.com/hdurix/jdl-blog) with your JDL file and protractor option (and my current generator branch).

Here what I understand from the protractor test. A test is done for testing the entity creation:
- for each relationship: if the relationship is many-to-many, [does nothing](https://github.com/hdurix/jdl-blog/blob/master/src/test/javascript/e2e/entities/entry/entry.spec.ts#L43): statement is commented
- for each relationship: if the relationship is NOT many-to-many, [selects the last option](https://github.com/hdurix/jdl-blog/blob/master/src/test/javascript/e2e/entities/entry/entry.spec.ts#L42) if any
- if at least one relationship is required (not the case on your sample JDL file), all the test is ignored: method is commented

All of the `delete` tests work properly with those cases, whether the target relationship contains entities (relationship with `User` or manually added with the app running) or not (test from scratch with no `Entry` for instance when testing `Blog` creation).

Now I need to understand why some Travis configurations are failing.

hdurix:
For random failures (especially in travis), I reckon I have to wait for the entity list to be displayed (→ retrieved from back-end) before trying to delete one (I need to find a way to do that with protractor).

hdurix:
Ok thanks! Next time I'll know...
It's a shame Travis does not let the committer to re run the build by himself.

pascalgrimaud:
probably to avoid an user to restart a build, and blocking the other builds, because there are some priorities.

Another solution is to use your own Travis instance, so you will be the boss there

hdurix:
It's okay for me!

jdubois:
Awesome, let's merge this! And don't forget the bug bounty :-)

pascalgrimaud:
impressive work @hdurix !!

hdurix:
Bug bounty claimed: https://opencollective.com/generator-jhipster/expenses/4936

PR #8112, opened 2018-08-19
Webpack polish

wmarques:
Just made some changes to make the Webpack configurations more consistent and replaced deprecated `loaders` by `use`.

I also removed jquery from the Angular dependencies, it doesn't seem used.

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -40,7 +40,6 @@
     "@ng-bootstrap/ng-bootstrap": "3.0.0",
     "bootstrap": "4.1.3",
     "core-js": "2.5.7",
-    "jquery": "3.3.1",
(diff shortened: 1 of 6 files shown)


jdubois:
Awesome! Thanks @wmarques

generators/client/templates/angular/package.json.ejs
```diff
@@ -40,7 +40,6 @@
     "@ng-bootstrap/ng-bootstrap": "3.0.0",
     "bootstrap": "4.1.3",
     "core-js": "2.5.7",
-    "jquery": "3.3.1",
```
deepu105: This means we dont use JQuery anymore, so may be we should remove the logo from website as well @wmarques

wmarques: I'll do that tonight 🙂

Issue #7810, opened 2018-06-19
App with Elasticsearch doesn't work out-of-the-box with Heroku


mraible:
##### **Overview of the issue**

I'm adding this issue because I believe @jkutner did work recently to make it so Elasticsearch was automatically configured when you deploy an app to Heroku. I see the following on startup, so I'm guessing provisioning Elasticseach is working.

```
2018-06-19T14:16:28.440721+00:00 app[api]: Attach SEARCHBOX (@ref:searchbox-adjacent-53104) by user matt@raibledesigns.com
```

However, when my app starts up, it fails to connect to this instance, and all CRUD functionality is broken.

```
2018-06-19T14:24:38.262001+00:00 app[web.1]: 2018-06-19 14:24:38.261 ERROR 4 --- [           main] .d.e.r.s.AbstractElasticsearchRepository : failed to load elasticsearch nodes : org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: [{#transport#-1}{hR944JNITnub8FEFr6wLEQ}{localhost}{127.0.0.1:9300}]
2018-06-19T14:24:38.561165+00:00 app[web.1]: 2018-06-19 14:24:38.560 ERROR 4 --- [           main] .d.e.r.s.AbstractElasticsearchRepository : failed to load elasticsearch nodes : org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: [{#transport#-1}{hR944JNITnub8FEFr6wLEQ}{localhost}{127.0.0.1:9300}]
2018-06-19T14:24:39.424443+00:00 app[web.1]: 2018-06-19 14:24:39.424 ERROR 4 --- [           main] .d.e.r.s.AbstractElasticsearchRepository : failed to load elasticsearch nodes : org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: [{#transport#-1}{hR944JNITnub8FEFr6wLEQ}{localhost}{127.0.0.1:9300}]
2018-06-19T14:24:39.730791+00:00 app[web.1]: 2018-06-19 14:24:39.730 ERROR 4 --- [           main] .d.e.r.s.AbstractElasticsearchRepository : failed to load elasticsearch nodes : org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: [{#transport#-1}{hR944JNITnub8FEFr6wLEQ}{localhost}{127.0.0.1:9300}]
[...]

PR #8120, opened 2018-08-20
Replace heroku elasticsearch addon with bonsai

jkutner:
This supports @mraible's changes in https://github.com/jhipster/generator-jhipster/issues/7810

### DIFF
--- a/generators/heroku/index.js
+++ b/generators/heroku/index.js
@@ -323,7 +323,7 @@ module.exports = class extends BaseGenerator {
 
                 this.log(chalk.bold('\nProvisioning addons'));
                 if (this.searchEngine === 'elasticsearch') {
-                    exec(`heroku addons:create searchbox:starter --as SEARCHBOX --app ${this.herokuAppName}`, addonCreateCallback.bind(this, 'Elasticsearch'));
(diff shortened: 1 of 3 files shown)


mraible:
@jkutner Should I do the migration to Jest in a separate PR?

jkutner:
@mraible i would said yes, and we can merge this first (since it doesn't break anything).

mraible:
FYI, I got an email this afternoon from Bonsai saying the following:

```
The usage for your Bonsai cluster, health-by-points-3 has exceeded the limits for its Sandbox ($0/mo) plan.

* Shards: Using 10 out of 6 shards (167% of limit).

What's next?

Your action is required in order to ensure continuity of service and avoid interruption to your cluster traffic. Please upgrade to the appropriate plan, or reduce your cluster's usage to bring it within its current plan's limits.

...

In order to avoid service interruption, please upgrade your plan or modify your usage before Thursday, August 30. If you are still over your limits past that date, your cluster will be placed into read-only mode. Searches will continue to work, and existing data can be deleted to make room, but any new data will not be indexed.
```

I only generated three entities in my project. It looks like it's $10/month to support 10 shards, which seems reasonable.

https://elements.heroku.com/addons/bonsai

jdubois:
@mraible at first I wanted to have sample applications working for free, but it is totally reasonable that if you select this kind of third-party tools (which cost a lot of $$$ to host), you need to pay for something.

PR #8121, opened 2018-08-20
WIP: Clean scss

bertwin:
### Proposal to change sass/css structure

In this PR i would like to test and share some ideas regarding css.

#### Remove division between sass / css
The prefered way to theme bootstrap is via setting bootstrap variables, right now this is not fully utilized.
On the other hand there are small local component styles that would probably never benefit from sass.
I would like to suggest to remove the sass option from the generator and always use scss for theming boostrap and plain css for the rest.

#### Replace node-sass with the js version of  dart-sass
So we don't need binaries when using sass.
http://sass-lang.com/dart-sass
This is not really needed, but i can imagine that node-sass is one of the reasons not to use the sass option when generating.

#### Share the themed bootstrap between  react / angular
`_bootstrap-variables.scss` and `bootstrap-jhipser.scss` could eventually be shared between frontends.

#### To try it:
These changes currently only work with a react monolith with sass support enabled.
It should look mostly identical, with a lot less ad-hoc styles.

What do you think, is it worth it to complete some or all of the idea's?

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/app.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/app.tsx.ejs
@@ -47,10 +47,9 @@ export class App extends React.Component<IAppProps> {
   }
 
   render() {
-    const paddingTop = '60px';
     return (
       <Router>
-        <div className="app-container" style={{ paddingTop }}>
+        <div className="app-container">
           <ToastContainer position={toast.POSITION.TOP_LEFT as ToastPosition} className="toastify-container" toastClassName="toastify-toast"/>
           <ErrorBoundary>
             <Header
@@ -65,8 +64,8 @@ export class App extends React.Component<IAppProps> {
               isSwaggerEnabled={this.props.isSwaggerEnabled}
             />
           </ErrorBoundary>
-          <div className="container-fluid view-container" id="app-view-container">
-            <Card className="jh-card">
+          <div className="container-fluid view-container mt-3" id="app-view-container">
+            <Card className="jh-card p-3">
               <ErrorBoundary>
                 <AppRoutes/>
               </ErrorBoundary>
--- a/generators/client/templates/react/src/main/webapp/app/_bootstrap-variables.scss.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/_bootstrap-variables.scss.ejs
@@ -43,4 +43,25 @@ $border-radius-sm:       .1rem;
 // Body:
 // Settings for the `<body>` element.
 
-$body-bg:                   #e4e5e6;
+$body-bg:                   #fafafa;
+
+
+$sizes: (
+  40: 40%,
+  60: 60%
+);
+
+
+$gray-jh:               #353d47;
+$blue-jh:               #3e8acc;
+$brown-jh:              #533f03;
+
+$blue:                  $blue-jh;
+$dark:                  $gray-jh;
+
+$component-active-bg:   $gray-jh;
+
+$link-color:            $brown-jh;
+$link-hover-decoration: none;
+
+$navbar-nav-link-padding-x: 1rem;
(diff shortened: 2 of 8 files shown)


deepu105:
@bertwin Thanks for taking the time to work on this. here are my thoughts

1. The reason for not using Sass by default was node-sass, so if dart-sass can be used instead without any binaries we can definitely switch on Sass by default. This means customizing bootstrap will be possible all the time.
2. I would prefer to have Sass everywhere in generated code since you can write plain CSS in SCSS files people who prefer that can always do that but generated code will be in Sass only. The advantage of this is you can use the variables defined by Bootstrap everywhere for consistency.

So yes +1 for this

@jhipster/developers does anyone has an objection to this?

generators/client/files-react.js
```diff
@@ -198,14 +199,6 @@ const files = {
             ]
         },
         {
-            condition: generator => generator.useSass,
```
deepu105: I would prefer to keep the scss versions and remove the css everywhere

generators/client/templates/react/src/main/webapp/app/app.tsx.ejs
```diff
@@ -65,15 +65,17 @@ export class App extends React.Component<IAppProps> {
               isSwaggerEnabled={this.props.isSwaggerEnabled}
             />
           </ErrorBoundary>
-          <div className="container-fluid view-container" id="app-view-container">
-            <Card className="jh-card">
+          <Container fluid tag="main" role="main"  className="view-container mt-3" id="app-view-container">
```
deepu105: I would prefer not to do any layout updates in this PR, if required this can be done separately. Can you do only the SCSS part in this PR

bertwin: :+1: I will keep only the minimum required changes in this PR. Shall I leave dart-sass in?

deepu105: yes please

gmarziou:
This sounds great, have you compared dart-sass with node-sass in terms of build speed?

By using sass by default, would it make any difference for integrating available themes (e.g. from themeforest.net)? Harder, simpler or same?

gmarziou:
Another benefit from using dart-sass is that maybe we would be able to use `--ignore-scripts` option during npm install and protect our users from command injection attacks. Last time I checked, there were less than 10 postinstall scripts out of 1000 node modules we install and the one we really needed was node-sass.

jdubois:
I didn't know about dart-sass : big +1 from me! That could indeed be a game changer and push the use of Sass with JHipster (I love the language, but I hate the current implementation)

wmarques:
Big +1 for dart-sass if the performances are good, the node-sass C++ binaries are just a mess to manage in big companies with proxies...
Also +1 to make Sass the default

deepu105:
@bertwin its merged, please update the PR to only have the common base file improvements and to remove generated CSS files.

@jdubois also we dont have to deprecate or remove CSS, with Sass, CSS also will just work fine(I do this in my own apps all the time), so if someone wants to use CSS that will be fine as well. For new apps everything should be fine, for existing apps when they upgrade there will be new Sass files and we could either skip them for existing apps with css option or just generate them and let the user decide which one to keep. WDYT?

deepu105:
@jdubois I have already made it default, will create another ticket for removal, we can do that in a later release

bertwin:
@deepu105 can I assume that all new apps will have sass enabled? Otherwise the css files are still needed.

deepu105:
Not untill we remove the CSS option, so for now just do the Sass part we
can later remove CSS

Thanks & Regards,
Deepu

deepu105:
@bertwin are you still workking on this?

deepu105:
@bertwin is this ready?

deepu105:
@bertwin I'm closing this as there is no progress, we will do this anyway for v6, and if you find time to continue this don't hesitate to open a new PR

PR #8122, opened 2018-08-20
Add Spring Data Jest to improve Elasticsearch

mraible:

- [x] Documentation added in https://github.com/jhipster/jhipster.github.io/pull/613

Implements changes in #7810.


mraible:
The Travis build for this PR keeps failing with errors like:

```
  JHipster generator
    Default configuration with
      AngularX
keytool error: java.io.FileNotFoundException: src/main/resources/config/tls/keystore.p12 (No such file or directory)
        1) "before each" hook
      React
keytool error: java.io.FileNotFoundException: src/main/resources/config/tls/keystore.p12 (No such file or directory)
        2) "before each" hook
      using npm flag
keytool error: java.io.FileNotFoundException: src/main/resources/config/tls/keystore.p12 (No such file or directory)
        3) "before each" hook
      Gradle
```

@jdubois Is this because of a change you recently made on the master branch?

generators/cleanup.js
```diff
@@ -156,4 +156,7 @@ function cleanupOldServerFiles(generator, javaDir, testDir, mainResourceDir, tes
         generator.removeFile(`${testDir}web/rest/ProfileInfoResourceIntTest.java`);
         generator.removeFile('gradle/mapstruct.gradle');
     }
+    if (generator.isJhipsterVersionLessThan('5.2.1')) {
```
mraible: @vishal423 Is this the correct version to use? `ElasticsearchConfiguration.java` is removed in this PR and no longer needed in 5.2.2+.

jdubois:
Sorry @mraible , it looks like I broke the build, I'm having a look right now

jdubois:
Master should be fixed now, sorry :-(

mraible:
FWIW, I tried to remove Log4J as a dependency using [this guide](https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/_using_another_logger.html), but failed.

jdubois:
@mraible oh I remember trying (and failing!) this also!

generators/cleanup.js
vishal423: looks good

jdubois:
This means people won't be able to use Spring Data Elasticsearch at all. As it's the official implementation, I'm not sure this is a good idea. I understand it won't work on cloud providers, and I agree Spring Data Jest looks like a better solution, but still that's an important change.
Besides, shouldn't you also remove Spring Data Elasticsearch from the Maven/Gradle dependencies?

generators/cleanup.js
jdubois: yes this API is really confusing!! I'm not totally sure, shouldn't that be 5.2.2?

generators/server/templates/src/main/resources/config/application-dev.yml.ejs
```diff
@@ -193,12 +193,9 @@ spring:
     <%_ } _%>
     <%_ if (searchEngine === 'elasticsearch') { _%>
         elasticsearch:
-            cluster-name:
-            cluster-nodes: localhost:9300
             properties:
                 path:
-                    logs: <%= BUILD_DIR %>elasticsearch/log
-                    data: <%= BUILD_DIR %>elasticsearch/data
+                    home: <%= BUILD_DIR %>elasticsearch
```
jdubois: as this is a Spring Data Elasticsearch property, is this being used by Jest also?

mraible: Yes. Spring Data Jest doesn’t reimplement anything in Spring Data Elasticsearch. It uses this property to create the embedded ES instance. If you don’t have spring.data.jest.uri defined (which we don’t in the dev profile), this property is required.

mraible:
Nothing changes from a code perspective. You can still use Spring Data Elasticsearch. Dependencies change and the port changes from 9300 (binary API) to 9200 (REST API), but that’s about it. I think it’s a good idea because: 1. It works and solves Elasticsearch on Heroku. 2. It allows for embedded Elasticsearch which is a much better development experience. 3. If people want to switch back, they only need to remove the dependency and change a couple properties.

generators/cleanup.js
mraible: That’s sounds more logical. I’ll fix it.

jdubois:
Thanks @mraible - you're right, let's do this! So much better than what is provided by default!

generators/cleanup.js
vishal423: yeah, looks like I mis-interpreted the api. I am bit lost in jhipster versioning :confused:

jdubois:
Matt, doesn't this need a new release of the JHipster dependencies?

DanielFran:
@jdubois For now it is not needed since @mraible added the version in the commits.

So for now it is not dependent of https://github.com/jhipster/jhipster/pull/87

jdubois:
Oh thanks @DanielFran - let me release a new version of JHipster dependencies, so those could be removed.

DanielFran:
@jdubois Do you want I split my PR in jhispter in 2, 1 that we could already aplly in this new version and another one with specific related with spring 2.1.x?

jdubois:
Thanks @DanielFran but it's already released :-)
However I have a couple of issues, see https://github.com/jhipster/generator-jhipster/issues/7810#issuecomment-415354308

cbornet:
Does it mean we can now use an embedded ES in the tests like we used to instead of the current mock-based implementation ?

jdubois:
@cbornet yes I guess we can do it again now.

PR #8139, opened 2018-08-21
Also ignore application.properties for Spring (close #8117)

### LINKED ISSUE #8117: jhipster upgrade deletes all ignored files

So I was migrating from a version to another. And my `.gitignore` is ignoring files like `/application.properties` which are not ignored by default by jhipster.

After migration, all these files were deleted.

The ignore files should not be touched.

##### **JHipster Version(s)**

Was going from 4.14.4 to 4.14.5

It probably doesn't matter

Doesn't matter

macOS 10.13.6

- [x ] Checking this box is mandatory (this is just to show you read everything)

henri-tremblay:

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -85,7 +85,7 @@ module.exports = class extends BaseGenerator {
     }
 
     _cleanUp() {
-        const filesToKeep = ['.yo-rc.json', '.jhipster', 'node_modules', '.git', '.idea', '.mvn'];
+        const filesToKeep = ['.yo-rc.json', '.jhipster', 'node_modules', '.git', '.idea', '.mvn', 'application.properties'];
         shelljs.ls('-A').forEach((file) => {
             if (!filesToKeep.includes(file)) {
                 this.info(`Removing ${file}`);


jdubois:
I'm not sure this should be ignored - you're storing secrets there, but maybe some other people would be happy to have that file commited

henri-tremblay:
My assumption is that it isn't the purpose of this file. If it's not a secret, why not putting it in `src/main/ressources`?

jdubois:
You could put it in both place, I don't think it's intended to have this one storing secrets. Maybe you could have a specific profile for that, and exclude it, for example having a `application-secret.yml`... But I would advise not putting it at all in the project, and for this there are 3 ways:

- Use the JHipster Registry. This is our "standard" way in JHipster.
- Use a file in the same folder as your production application (like I do for JHipster Online, because I don't want to pay for an extra server for JHipster Registry)
- Use environment variables: it's easy and widely supported, so a lot of people do this, but it's not very secure - it's usually quite easy to read them. I remember some time ago working for a big bank, and I accessed their app through JMX (yes, it was opened, for monitoring...), and from there you could see the environment variables... so I got the super secret private key :-)

henri-tremblay:
Hum. So far I think I still prefer my actual solution. I also think that this file is intended by Spring to never be committed. As I said, anything that can be committed should be in `src/main/resources`.

But if we manage to have the same result by parsing `.gitignore` or `.jhipsterignore` I will be happy to close this.

jdubois:
I will give it a try with .gitignore (as this will be a broader solution, if you use application.yml for example)

PR #8151, opened 2018-08-23
Add application name on path in application java

nguerrier:
With the new version of the generator, the message that the application started successfully is confusing. I added the name of the application in the end message after the local url and the external url. Indeed, the local and external url as it is presented returns to an error 404.

Example:
```
----------------------------------------------------------
        Application 'jhipster' is running! Access URLs:
        Local:          http://localhost:8081
        External:       http://127.0.1.1:8081
        Profile(s):     [dev, swagger]
----------------------------------------------------------
```

becomes:

```
----------------------------------------------------------
        Application 'jhipster' is running! Access URLs:
        Local:          http://localhost:8081/jhipster
        External:       http://127.0.1.1:8081/jhipster
        Profile(s):     [dev, swagger]
----------------------------------------------------------
```

### DIFF
--- a/generators/server/templates/src/main/java/package/Application.java.ejs
+++ b/generators/server/templates/src/main/java/package/Application.java.ejs
@@ -115,15 +115,17 @@ public class <%= mainClass %> {
         }
         log.info("\n----------------------------------------------------------\n\t" +
                 "Application '{}' is running! Access URLs:\n\t" +
-                "Local: \t\t{}://localhost:{}\n\t" +
-                "External: \t{}://{}:{}\n\t" +
+                "Local: \t\t{}://localhost:{}/{}\n\t" +
+                "External: \t{}://{}:{}/{}\n\t" +
                 "Profile(s): \t{}\n----------------------------------------------------------",
             env.getProperty("spring.application.name"),
             protocol,
             env.getProperty("server.port"),
+			env.getProperty("spring.application.name"),
             protocol,
             hostAddress,
             env.getProperty("server.port"),
+			env.getProperty("spring.application.name"),
             env.getActiveProfiles());
         <%_ if (serviceDiscoveryType && (applicationType === 'microservice' || applicationType === 'gateway' || applicationType === 'uaa')) { _%>
 


deepu105:
There is nothing wrong here, if you have added a context path, then you need to add it to log as well

This ticket has been closed as the **guidelines are not followed**.

Tickets must follow our [Guidelines](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md), as mentioned in:

1. our [Readme file on the front page of the project](https://github.com/jhipster/generator-jhipster/blob/master/README.md),
2. the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) and
3. our [Help page](http://jhipster.github.io/help/)

We have also created a template on the ["create a new ticket" page](https://github.com/jhipster/generator-jhipster/issues/new) to help you follow those guidelines.

**If this is indeed a JHipster bug**, please open a new issue with proper details or update this issue with all details and request to reopen.
Issues opened without proper details will be closed without explanation.

**If you have a question** please use Stack Overflow, and tag the question with [jhipster](http://stackoverflow.com/questions/tagged/jhipster). This helps the project to keep the issue tracker clean. Also, Stack Overflow will give your question a larger audience:

- This will increase your chances to get an answer
- Answers will be of higher quality, as there is a voting system
- This will also help other users who might have the same issue, as questions are tagged and easily searchable

jdubois:
Indeed I agree with @deepu105 the original URL is correct, there shouldn't be a context path, unless you deploy on Tomcat for example... If you are doing microservices, maybe you want to point to the gateway? That would explain your issue. But then again, I would keep the current message, as you don't know where the gateway is located (your code only works if you deploy the gateway and the microservice on the same host)

DanielFran:
@jdubois If I remember context-root os added by default for Heroku. So maybe it is needed changes for this case?

jdubois:
@DanielFran I don't remember that, and I'm sure Heroku also does deploy app on their root context, I do this all the time. Anyway this message is mostly for development, in production you could have any kind of front-end proxy that has another IP, or maybe add a context path, so indeed that only makes sense on the server that is running the application itself.

DanielFran:
@jdubois I was talking about https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/resources/config/application.yml.ejs#L221

jdubois:
Oh yes, that's a quite different thing: with Istio you don't need a service discovery server anymore, so in that case it would be normal to have microservices with no service discovery. But at the moment I doubt anybody uses that.

In that quite specific use case, indeed there is a context root: I guess we could add it to the end of the message as @nguerrier asks. I think we can get it reliably using the Servlet API, but that needs to be tested. Otherwise we could use the Spring Boot property that configures it: that's very specific and it wouldn't work with an app server for example (if you deploy your application as a WAR in Tomcat), but that would be enough for this use case.

deepu105:
My JH apps also have a context path and I show that in the url with below
```
log.info("\n----------------------------------------------------------\n\t" +
                "Application '{}' is running! Access URLs:\n\t" +
                "Local: \t\t{}://localhost:{}{}\n\t" +
                "External: \t{}://{}:{}{}\n\t" +
                "Profile(s): \t{}\n----------------------------------------------------------",
            env.getProperty("spring.application.name"),
            protocol,
            env.getProperty("server.port"),
            contextPath,
            protocol,
            InetAddress.getLocalHost().getHostAddress(),
            env.getProperty("server.port"),
            contextPath,
            env.getActiveProfiles());
```

DanielFran:
@deepu105 This is exactly what it is needed except that it should validate porperty "server.servlet.context-path".

I believe Spring renamed it with spring-boot 2.

PierreBesson:
Printing the context path would be OK as many people do use this and we set this by default for microservice with no discovery. The only thing that I would change from Deepu's code would be:
```
        if (contextPath == null || "".equals(contextPath)) {
            contextPath = "/";
        }
```
to :
```
        if (StringUtils.isBlank(contextPath)) {
            contextPath = "/";
        }
)
```

jdubois:
OK let's do this, @PierreBesson do you want to do the PR?

PierreBesson:
@jdubois I can, but maybe @nguerrier will want to do it.

nguerrier:
Hi. Thanks you all for your interest in this PR.
@PierreBesson I let you do this PR in my place. There is no problem 😉

PR #8155, opened 2018-08-23
Change secret length to conform with JJWT's (more secure) requirements

mraible:


jdubois:
Thanks a lot @mraible 
It still worries me a bit that now we need to write the secret key in base64, but that's not such a huge issue, and I documented it in https://github.com/jhipster/generator-jhipster/commit/03595a285a4e528dcdaa1efdd6778b13dd359b6a

generators/server/index.js
```diff
@@ -259,7 +259,7 @@ module.exports = class extends BaseGenerator {
 
                     // Generate JWT secret key if key does not already exist in config
                     if (this.authenticationType === 'jwt' && this.jwtSecretKey === undefined) {
-                        this.jwtSecretKey = Buffer.from(crypto.randomBytes(50).toString('hex')).toString('base64');
+                        this.jwtSecretKey = Buffer.from(crypto.randomBytes(100).toString('hex')).toString('base64');
```
les-okta: 64 bytes would be sufficient since that's 512 bytes - the max needed for any JWT HMAC SHA algorithm.

jdubois: Thanks!

Issue #8172, opened 2018-08-27
Can't Produce an Executable JAR


brunoborges:
##### **Overview of the issue**
Can't get a working executable JAR file (as in `.jar`) out of a Maven JHipster project.

##### **Motivation for or Use Case**
Due to requirements from other tools, including custom Maven plugins, we need the output of the Maven project to be an executable JAR file, not an executable WAR file.

##### **Reproduce the error**
Get an executable JAR by renaming the WAR file:
1. Create a monolithic dummy JHipster application
1. Build the application `mvn package`
1. Rename the `target/<app>.war` file to `target/<app>.jar`
1. Run it with `java -jar target/<app>.jar`
1. Application starts, but `curl -I http://localhost:8080` gives 404

Get an executable JAR by changing Maven packaging to `jar`:
1. Create a monolithic dummy JHipster application
1. Build the application `mvn package`
1. Change `<packaging>` inside `pom.xml` to `jar`
1. Run the JAR file with `java -jar target/<app>.jar`
1. Visit `http://localhost:8080` in your browser.
1. Application does not render properly (missing static content).

##### **Related issues**
See this StackOverflow thread: https://stackoverflow.com/questions/52046249/spring-boots-jhipster-war-artifact-renamed-to-jar-wont-run/52047100#52047100

See this Twitter thread: https://twitter.com/brunoborges/status/1034169585564495872

##### **JHipster Version(s)**

```
foobar@0.0.0 /Users/bruno/work/jhipster/jhipster-apps/foobar
└── generator-jhipster@5.2.1

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp"
    },
    "jhipsterVersion": "5.2.1",
    "applicationType": "monolith",
    "baseName": "foobar",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "no",
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSwaggerCodegen": false,
[...]

PR #8181, opened 2018-08-29
copy webapp into resources/static to generate executable jar

atomfrede:
when executing `bootJar` the compiled webapp is copied to `src/main/resources` such that it can be served from the jar file.

updates #8172

### DIFF
--- a/generators/server/templates/gradle/profile_dev.gradle.ejs
+++ b/generators/server/templates/gradle/profile_dev.gradle.ejs
@@ -48,6 +48,11 @@ task webpackBuildDev(type: <%= _.upperFirst(clientPackageManager) %>Task, depend
 war {
     webAppDirName = '<%= CLIENT_MAIN_SRC_DIR %>'
 }
+
+task copyIntoStatic (type: Copy) {
+    from '<%= CLIENT_DIST_DIR %>'
+    into 'build/resources/main/static'
+}
 <%_ } _%>
 
 processResources {
@@ -72,6 +77,9 @@ processResources {
 
 <%_ if (!skipClient) { _%>
 processResources.dependsOn webpackBuildDev
+copyIntoStatic.dependsOn processResources
+bootJar.dependsOn copyIntoStatic
+
 yarn_install.onlyIf { shouldWebpackRun() == true }
 
 def shouldWebpackRun() {
--- a/generators/server/templates/gradle/profile_prod.gradle.ejs
+++ b/generators/server/templates/gradle/profile_prod.gradle.ejs
@@ -51,6 +51,11 @@ task webpack(type: <%= _.upperFirst(clientPackageManager) %>Task, dependsOn: '<%
 war {
     webAppDirName = '<%= CLIENT_DIST_DIR %>'
 }
+
+task copyIntoStatic (type: Copy) {
+    from '<%= CLIENT_DIST_DIR %>'
+    into 'build/resources/main/static'
+}
 <%_ } _%>
 
 processResources {
@@ -86,4 +91,6 @@ gitProperties {
 <%_ if (!skipClient) { _%>
 test.dependsOn webpack_test
 processResources.dependsOn webpack
+copyIntoStatic.dependsOn processResources
+bootJar.dependsOn copyIntoStatic
 <%_ } _%>


deepu105:
@jhipster/developers so is anyone gonna attempt this for maven?

jdubois:
@atomfrede oh you have probably more control over what we have in Maven. Not sure this is easy to do in Maven... At least that allows us to try this with Gradle, and maybe switch to JAR files only in the future, as WAR are becoming more and more obsolete.

atomfrede:
`./gradlew clean bootJar -Pprod`:
screenshot from 2018-09-04 21-26-43

`./gradlew clean bootWar -Pprod`:
screenshot from 2018-09-04 21-30-36

So the jar contains the web app in `static` and the war has nothing in `static`. It looks fine to me.

erikkemperman:
Haven't followed this ticket closely, but it sound like it should be possible with Maven as well. Will have a look, probably this weekend.

jdubois:
Thanks a lot @erikkemperman

erikkemperman:
Hm, just had a quick trial -- you were right; this might be slightly tricky. Would be be all right to introduce a generator prompt asking whether the user wants to make a jar or a war? Because I don't see an elegant way to achieve something like this after the fact.

jdubois:
One more question for something this specific isn't good.... Maybe a specific goal? This would need documentation, but that would be enough for the moment. And then we'll migrate to executable JARs in the next major release, probably.

PierreBesson:
+1 for choosing between jar and war. However sadly Google AppEngine is only compatible with War packaging.

erikkemperman:
Yeah, I tried that -- the problem is we need to change the project.packaging property and we can't do dynamically, or override it from the commandline. So either we choose it during generation, or provide a separate POM file for jar packaging. Both aren't great options...

jdubois:
Oh, this sucks @erikkemperman ! And @PierreBesson is right, we need the WAR for Google App Engine, at least.

So we're stuck at the moment with Maven.

deepu105:
I think its still good. If someone really wants jar they can at least use Gradle :wink: 
And we can add some docs on how to change it for Maven. @erikkemperman did you check the Payara Micro Maven plugin mentioned by @brunoborges

brunoborges:
@jdubois how about this being an option when choosing between Maven and Gradle? What about a third option? As in:

[ ] Maven (WAR)
[ ] Maven (JAR)
[ ] Gradle

brunoborges:
Yeah, the Payara Micro Maven Plugin generates a classic WAR, and a bundle JAR. 

I'd guess this would require a fork and the creation/release of a "jhipster-maven-plugin" that does similar thing. 

Frankly, the ideal scenario would be for the spring-boot-maven-plugin to do that. :-)

https://github.com/payara/ecosystem-maven/tree/master/payara-micro-maven-plugin/

erikkemperman:
The spring-boot plugin supports executable jars, but it requires the `packaging` property that we can’t elegantly override post app generation, as far as I can tell.

jdubois:
I like the idea of @brunoborges with the 3 options, that's smart!

erikkemperman:
Yeah, me too. I'll try to get that working soon!

pascalgrimaud:
So, what I suggest:
- 3 options for question
    - Maven (WAR) <- by default
    - Maven (JAR)
    - Gradle
- In configuration, we could keep `buildTool` with value:
    - maven
    - gradle
- Add a new property `buildPackaging`:
    - empty or war: same result
    - jar

What do you think?

gmarziou:
Looks good, maybe the property name should reflect more that it's about server packaging

jkutner:
I'm -1 on the WAR/JAR option, but it still may be the best thing to do. I don't think the implications are clear to the user selecting between them. If that is implemented, be aware that there are several references to `*.war` that need to account for this (some are Gradle, so won't matter i guess):

```
$ grep -nr "\*\.war" generators
generators/heroku/index.js:560:                    let warFileWildcard = 'target/*.war';
generators/heroku/index.js:562:                        warFileWildcard = 'build/libs/*.war';
generators/heroku/templates/Procfile.ejs:19:web: java $JAVA_OPTS <% if (applicationType === 'gateway' || dynoSize === 'Free') { %>-Xmx256m<% } %> -jar <% if (buildTool === 'maven') { %>target<% } %><% if (buildTool === 'gradle') { %>build/libs<% } %>/*.war --spring.profiles.active=prod,heroku<% if (buildTool == 'maven' && herokuDeployType == 'git') { %>,no-liquibase<% } %> --server.port=$PORT <% if (prodDatabaseType === 'mongodb') { %>--spring.data.mongodb.database=$(echo "$MONGODB_URI" | sed "s/^.*:[0-9]*\///g")<% } %>
generators/heroku/templates/pom-profile.xml.ejs:36:                                                <exclude>*.war</exclude>
generators/heroku/templates/heroku.gradle.ejs:29:            delete fileTree(dir: "build/libs", exclude: "*.war")
generators/cloudfoundry/index.js:165:                    cloudfoundryDeployCommand += ` ${glob.sync(`${warFolder.trim()}*.war`)[0]}`;
generators/cloudfoundry/index.js:167:                    cloudfoundryDeployCommand += `${warFolder}*.war`;
generators/server/templates/gitattributes.ejs:85:*.war           binary
generators/server/templates/README.md.ejs:235:    java -jar target/*.war<% } %><% if (buildTool === 'gradle') { %>
generators/server/templates/README.md.ejs:236:    java -jar build/libs/*.war<% } %>
generators/server/templates/gitignore.ejs:87:*.war
generators/server/templates/src/main/docker/.dockerignore.ejs:32:!*.war
generators/server/templates/src/main/docker/Dockerfile.ejs:33:ADD *.war app.war
generators/ci-cd/templates/jenkins/Jenkinsfile.ejs:96:<%= indent %>        archiveArtifacts artifacts: '**/target/*.war', fingerprint: true
generators/ci-cd/templates/jenkins/Jenkinsfile.ejs:144:<%= indent %>        archiveArtifacts artifacts: '**/build/libs/*.war', fingerprint: true
generators/ci-cd/templates/jenkins/Jenkinsfile.ejs:174:<%= gitLabIndent %>        sh "cp target/*.war target/docker/"
generators/ci-cd/templates/jenkins/Jenkinsfile.ejs:178:<%= gitLabIndent %>        sh "cp build/libs/*.war build/docker/"
generators/ci-cd/templates/.gitlab-ci.yml.ejs:69:            - build/libs/*.war
generators/ci-cd/templates/.gitlab-ci.yml.ejs:103:            - target/*.war
```

Alternatively, cloud Maven generate both WAR and JAR by default?

deepu105:
@jkutner yes thats the best scenario and we implemented that for Gradle already but from what I understand its not easy to do that in Maven

erikkemperman:
As far as I understand it we’d need two poms for that, or intricate documentation for the user to switch by editing the pom. Maybe I’m missing something though?

erikkemperman:
Haven't had time to look into this over the weekend, after all. And the coming week is quite busy at work... If someone else wants to attempt this, please do. Otherwise I'll dive into as soon as I can.

erikkemperman:
Sorry guys, I'm not going to be able to tackle this any time soon... Lot of stuff on my plate at the moment.

vidhya03:
Hi ,

I am facing similar  [issue.](https://stackoverflow.com/questions/52933615/jhipster-spring-boot-static-resources-pdf-file-not-bundled-when-running-via-jenk ) https://stackoverflow.com/questions/52933615/jhipster-spring-boot-static-resources-pdf-file-not-bundled-when-running-via-jenk

when running via Jenkins the war size is 63MB, however when generate locally 93MB.

i don't know where to fix.

Issue #8164, opened 2018-08-26
Protractor test output is different between React and Angular


mraible:
The outputs from Protractor are different. With Angular, it shows dots, while React shows test names.

React:

```
[17:27:35] I/direct - Using ChromeDriver directly...


  Account
    ✓ should fail to login with bad password
    ✓ should login with admin account (2709ms)

  Administration
    ✓ should load metrics
    ✓ should load health
    ✓ should load configuration
    ✓ should load audits
    ✓ should load logs
    User management
      ✓ should load user management (2077ms)


  8 passing (13s)
```

Angular:

```
[17:35:43] I/direct - Using ChromeDriver directly...
Started
...........


11 specs, 0 failures
Finished in 31.203 seconds
```

I like React's output, especially the timings that show slow tests. Not sure if that's on purpose, or [zsh-command-time](https://github.com/popstas/zsh-command-time) that I have installed.

deepu105:
Thats coz of the framework used underneath. In Angular setup we use protractor with Jasmine and in React its Mocha + Chai.

Mocha has nice reporters out of the box. I once tried to get the same from Jasmine but it was no where near close to what mocha haf. May be things changed now and there is a better reporter for Jasmine now

deepu105:
Also the timing comes from Mocha as well

wmarques:
@deepu105 maybe we can use mocha + chai on Angular ?

deepu105:
We could and that way we can move protractor test templates outside react/angular folder and reuse them for both. E2e should be same regardless of technology anyway and what works for React should work for angular as well.

jdubois:
Oh that would be very impressive!

wmarques:
I can work on this on Friday as the typescript 3 work seems blocked by Angular...

wmarques:
@deepu105 I started to migrate from jasmine to mocha/chai on Angular. I'll provide a PR for this only and keep separate e2e tests. Then we can start "merging" the tests but we'll have some work to do to have same CSS selectors and same tests scenarios across the frameworks.

wmarques:
This is fixed by https://github.com/jhipster/generator-jhipster/pull/8197 right?

ruddell:
Yep, outputs now match.  Nice job

PR #8197, opened 2018-08-31
Angular: Use Mocha+Chai for end to end tests

wmarques:
This is the first step to merge the end to end tests for both Angular and React.
Once we've done that work, we can give the tests to the Vue guys so they can run tests on their Vue UI and see if it matches with other JHipster frameworks :) 

Related to #8164


pascalgrimaud:
You changed `beforeAll` to `before`, maybe you need to do the same for entities: https://github.com/wmarques/generator-jhipster/blob/ng-mocha-protractor/generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs#L52 ?

wmarques:
@pascalgrimaud Yep I need to update entities, I'll do that tonight

pascalgrimaud:
@wmarques : can you resolve merge conflict plz ?

wmarques:
@pascalgrimaud Done, hope the tests will pass this time !

wmarques:
I'll try to fix this asap, seems we have to tell Protractor to wait after entities test, he's probably clicking "too soon"

wmarques:
It works on my computer so I don't understand what is going on... Maybe I should add `browser.sleep(1000)`? But I'm not a fan of that kind of solution...

pascalgrimaud:
@wmarques : I'll have a look. I have some strange timeout in Travis with NPM, that I didn't have so much with Yarn

wmarques:
@pascalgrimaud The failing test is the one when you the user types "admin" and "admin" as user/password and then we expect to have the green message that tells a successful login.
Maybe the backend takes some time to respond and protractor is checking too early for the message...
I can also try to tell protractor "wait for the message" before the assertion.

pascalgrimaud:
I had a look on this:
- there are some error during `npm run lint`, with `No valid rules have been specified` in log
- so `npm test` failed

generators/entity-client/templates/angular/src/test/javascript/e2e/entities/entity.spec.ts.ejs
```diff
@@ -141,28 +143,27 @@ describe('<%= entityClass %> e2e test', () => {
         <%_ } _%>
         <%_ }); _%>
         await <%= entityInstance %>UpdatePage.save();
-        expect(await <%= entityInstance %>UpdatePage.getSaveButton().isPresent()).toBeFalsy();
+        expect(await <%= entityInstance %>UpdatePage.getSaveButton().isPresent()).to.be.false;
     });<%= closeBlockComment %>
 
-
+    after(async () => {
```
hdurix: Should not it be placed after the delete test?

wmarques:
Finally green ! 🎉🎉

generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
```diff
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import { browser, element, by, ExpectedConditions as ec } from 'protractor';
+import { browser, element, by, ExpectedConditions } from 'protractor';
```
mraible: Why make this change? I prefer to type `ec` over `ExpectedConditions`.

jdubois:
@wmarques this is totally awesome work -> can you fix the conflict? Then I think it's ready to merge

wmarques:
@jdubois  done !

pascalgrimaud:
@wmarques : you worked hard on this. Don't forget to claim the bounty :-)

ramden:
Just to give my 2 cents here (I am not involved at any level with them) -> I personally use https://www.cypress.io/ 

Works with every framework, works on docker, hast fixtures, gui app... Learning curve was very low... If you like have a look, maybe for jHipster 6 ;)

deepu105:
I use Cypress at work, I really like it but there are few things I dont like
1. It's very slow, it runs tests at human speed unlike Protractor
2. It cant run parallel suites

Thanks & Regards,
Deepu

ramden:
I agree, exept with 1. 
Protractor startup takes ages, cypress can execute 100 tests while protractor is loading :)
My experience was that protractor performance is horrible (and you dont get the power, tools and simplicity of cypress)

deepu105:
For me, startup time was never an issue and when you have a lot of e2e cypress really takes time. But I dont hold those gainst cypress its much more easier to use and debug than protractor. But then this is a lot of work and I dont think any of the core team members have the bandwidth for this

wmarques:
@hdurix Also told me about Cypress, I have to check this :) 
One of the advantage I see with Protractor is that it automatically "waits" for Promises and Timeouts so we don't have to manage the API loading etc... But it only works for Angular if I'm right so if we want to merge the end-to-end tests for all frontend frameworks, it won't be an advantage :(

deepu105:
Yes that is also an issue

Thanks & Regards,
Deepu

hdurix:
Indeed I also use Cypress with Vue.js at work and I also think it's nice!
It can be slow as human speed but it can be configured: for instance, the typing is done 1 letter every 10ms, but it's configurable.
And I didn't know about parallel testing, it's a big point.

IMHO, Protractor is a lot better than cypress for Angular because it hides a lot of complexities for developer. For the other frameworks (react and maybe Vue.js soon?), I think both have got pros and cons and are quite equivalents.

Below a table about how I see the options we have:

|                                                   | End user satisfaction | JHipster team Efforts |
|---------------------------------------------------|-----------------------|-----------------------|
| Protractor for all                                | +                     |  *                    |
| Cypress for all                                   | +                     | **                    |
| Protractor for Angular / Cypress for react & Vue.js | ++                    | **                    |
| 2 options (Protractor or Cypress) for all           | +++                   | ***                   |

- Switch from Protractor to Cypress will ask work for not a better end use satisfaction (at least for Angular apps).
- Use Cypress for react will represents a lot of work as well (for coding and maintaining it).
- The perfect solution would be to ask for the wanted tool during generation but the amount of work would be huge for creating and maintaining purposes.

As @deepu105 said, all is about time and priorities. We can imagine a Cypress [module](https://www.jhipster.tech/modules/creating-a-module/) if someone is interested to do it.

deepu105:
I'm definitely against supporting both as options, that's too much
maintenance work and no much value. I would say we stick to protractor for
now unless someone is ready to spend the effort and do the migration for
both React & Angular. I'm definitely not willing to do it. If anyone wants
to do it you are welcome.

Thanks & Regards,
Deepu

pascalgrimaud:
Protractor is an option here, that means doing cypress as a module should be doable !

ramden:
My comment was just an "option for the future", I think jumping on it at this moment maybe is not (yet) a good idea, at least until they have firefox support. 

Cypress can be used with any framework with any backend and because of that it would be an ideal candidate for jHipster.

So the end result could be : there is one codebase for e2e for all jHipster frameworks - cypress, 

Whatever framework jHipster is supporting today or will support in the feature (Angular, React, Vue, jQuery, frameworkXY...) it does not matter - as long as some html ids, classes or attributes and url routes stay the same in every framework generated project.

Also (I assume) 90% of people build their jHipster project in jenkins (and other CI tools) and probably love (as much as I do) the fancy video recording feature. Analyzing the problem (reconstructing) becomes so easy and simple when you see what actually happened with the complete app step by step and how you came to the problem. This is a part protractor cannot cope with, you have to analyze the logs, find the test code and understand what you wanted to do in this test. Then you recreate that propbably manually on the UI. Then you verify... yes it is not working.  

With cypress -> I watch the video, I see what the user has done to make it not work, then I go into code. I think this is a huge difference (maybe only my opinion but... isn't this how we imagine perfect e2e tests?).

I totally agree that cypress can be slow when you write 10000 tests with it but it can also be powerful and it depends how you write your tests.

mraible:
FWIW, Protractor can record screenshots, a video sounds pretty cool though.

https://blog.ng-book.com/taking-screenshots-with-protractor/

I found this plugin to record videos with Protractor, but haven’t tried it myself.

https://www.npmjs.com/package/protractor-video-reporter

deepu105:
For now, we will stick to Protractor we can reconsider again for JH6 or JH7
if Cypress adds Firefox support.

Thanks & Regards,
Deepu

PR #8200, opened 2018-08-31
React: update snapshot for CICD sub gen

pascalgrimaud:
Following this commit https://github.com/jhipster/generator-jhipster/commit/74f9525c0771d8ae4ec11a6ce38a05cd25c6f9f8#diff-826b1965e64cb5ef8452013ed545504e

I think it's better to add this directly to the package.json

_____

- [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green -> https://travis-ci.org/pascalgrimaud/generator-jhipster/builds/423074798

### DIFF
--- a/generators/ci-cd/index.js
+++ b/generators/ci-cd/index.js
@@ -96,6 +96,11 @@ module.exports = class extends BaseGenerator {
                 this.gitLabIndent = this.sendBuildToGitlab ? '    ' : '';
                 this.indent = this.insideDocker ? '    ' : '';
                 this.indent += this.gitLabIndent;
+                if (this.clientPackageManager === 'yarn') {
+                    this.frontTests = ' -u';
+                } else if (this.clientPackageManager === 'npm') {
+                    this.frontTests = ' -- -u';
+                }
             }
         };
     }
--- a/generators/ci-cd/templates/.gitlab-ci.yml.ejs
+++ b/generators/ci-cd/templates/.gitlab-ci.yml.ejs
@@ -89,7 +89,7 @@ maven-build:
         - ./mvnw compile -Dmaven.repo.local=$MAVEN_USER_HOME
         - ./mvnw test -Dmaven.repo.local=$MAVEN_USER_HOME
         <%_ if (!skipClient) { _%>
-        - ./mvnw com.github.eirslett:frontend-maven-plugin:<%= clientPackageManager %> -Dfrontend.<%= clientPackageManager %>.arguments=test -Dmaven.repo.local=$MAVEN_USER_HOME
+        - ./mvnw com.github.eirslett:frontend-maven-plugin:<%= clientPackageManager %> -Dfrontend.<%= clientPackageManager %>.arguments='test<%= frontTests%>' -Dmaven.repo.local=$MAVEN_USER_HOME
         <%_ } _%>
         <%_ if (cicdIntegrations.includes('sonar')) { _%>
         - ./mvnw org.jacoco:jacoco-maven-plugin:prepare-agent sonar:sonar <% if (sonarOrga) { %>-Dsonar.organization=<%= sonarOrga %> <% } %>-Dsonar.host.url=<%= sonarUrl %> -Dsonar.login=$SONAR_TOKEN -Dmaven.repo.local=$MAVEN_USER_HOME
--- a/generators/ci-cd/templates/jenkins/Jenkinsfile.ejs
+++ b/generators/ci-cd/templates/jenkins/Jenkinsfile.ejs
@@ -74,9 +74,9 @@ node {
 <%= indent %>    stage('frontend tests') {
 <%= indent %>        try {
         <%_ if (clientPackageManager === 'yarn') { _%>
-<%= indent %>            sh "./mvnw com.github.eirslett:frontend-maven-plugin:yarn -Dfrontend.yarn.arguments=test"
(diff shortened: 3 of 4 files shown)


generators/client/templates/react/package.json.ejs
```diff
@@ -195,7 +195,7 @@ limitations under the License.
     "postinstall": "webdriver-manager update --gecko false",
     <%_ } _%>
     "start": "<%= clientPackageManager %> run webpack:dev",
-    "test": "<%= clientPackageManager %> run lint && jest --coverage --logHeapUsage -w=2 --config src/test/javascript/jest.conf.js",
+    "test": "<%= clientPackageManager %> run lint && jest --coverage --updateSnapshot --logHeapUsage -w=2 --config src/test/javascript/jest.conf.js",
```
vishal423: Per my understanding, it's not recommended to update snapshots on each tests execution as it defies the purpose of snapshot testing. Instead, tests should fail in case of breaking change and snapshots should be updated/synced manually as part of fix

pascalgrimaud: so we have a different behaviour between local and cicd ? In that case, the cicd sub gen should fail for React as it doesn't contain `-u`: https://github.com/jhipster/generator-jhipster/blob/master/generators/ci-cd/templates/travis.yml.ejs#L79-L81

I'm working on switching back to NPM by default instead of Yarn, and `npm test -u` doesn't work.

vishal423: IMO, Local and CI/CD behavior should be in sync and both shouldn't contain `-u` flag to auto update snapshots. In case of breaking change, this would cause local snapshot tests to fail and would allow developer to check if this is correct failure. If yes, then, they will update snapshots by manually executing tests with -u flag. If not, they will fix UI component to preserve old behavior. 

From Jest docs:
>The test will fail if the two images do not match: either the change is unexpected, or the screenshot needs to be updated to the new version of the UI component.

pascalgrimaud: You're probably right, but I don't know how we can fix this:

```
Snapshot Summary
 › 2 snapshots failed from 1 test suite. Inspect your code changes or run `yarn test -u` to update them.
Test Suites: 1 failed, 52 passed, 53 total
Tests:       2 failed, 674 passed, 676 total
Snapshots:   2 failed, 2 total
Time:        28.924s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
``` 

A sample repo: https://github.com/pascalgrimaud/jhipster180901-react
Travis build: https://travis-ci.com/pascalgrimaud/jhipster180901-react

deepu105: For npm foes npm test -- -u work?

deepu105: Also the reason I added -u here is that unlike normal apps our travis test generates an app and does the tests so technically there is no snapshot to compare against and in ci env jest will just fail if the snapshot test is run for first time without the -u, that's why having -u in our travis test is ok, but we shouldn't do that directly in the package.json

travis/scripts/04-tests.sh
```diff
@@ -46,5 +46,5 @@ elif [ -f "gradlew" ]; then
         -Dlogging.level.io.github.jhipster.travis=OFF
 fi
 if [ -f "tsconfig.json" ]; then
-    yarn test -u
+    yarn test
```
deepu105: With npm `npm test -- -u` should work and with yarn it will produce an extra warning which should be ok

generators/client/templates/react/package.json.ejs
vishal423: @deepu105 , shouldn't snapshots required for auto generated tests be part of generator so that you don't require explicit `-u` flag for newly generated application tests to succeed?

travis/scripts/04-tests.sh
pascalgrimaud: ` npm test -- -u` works ! thanks guys

I will revert my commits and fix cicd sub gen accordingly.

jdubois:
Merging as @pascalgrimaud fixed the review issues

deepu105:
@pascalgrimaud I dont think we should be adding the -u param for the CICD config we generate, its ok to add that to our own travis test config for JHipster but not for the generated CICD tests. Was this causing any failures? I thought the issue was only with our own Travis tests?

pascalgrimaud:
@deepu105 :
Without this updateSnapshot flag, I got the issue when using react + cicd

A sample repo: https://github.com/pascalgrimaud/jhipster180901-react
Travis build: https://travis-ci.com/pascalgrimaud/jhipster180901-react

To reproduce :
- generate a react project
- use cicd sub gen, without this flag
- push to a repo

deepu105:
@pascalgrimaud Ok understood. I was thinking that IRL people would run tests at least once locally before CICD so that they will checkin the generated file, but in a very automated environment it could be possible so may be lets just comment out the snapshot tests in the generated files and document it so people can enable them and checkin the generated file while doing so. If we have the update flag in the CICD pipeline then snapshot testing is useless

deepu105:
@pascalgrimaud another option is to just document to say that they need to generate and check-in the snapshots first time for CICD to work and the failure they see is a valid case

pascalgrimaud:
I understand your point of view, but now we have another workflow with jhipster-online. You can generate the project and the cicd configuration. You don't need to do it locally. If we remove the updateSnapshot, in this case, it will fail.

deepu105:
Then lest comment out the snapshot tests by default with some nice comments
with instructions and update the React docs accordingly

Thanks & Regards,
Deepu

PR #8208, opened 2018-09-03
Prevent sending insight stats when running unit tests

PierreBesson:
I noticed strange patterns when looking at the graphs. They looked like they are produced by an automated system. Look how openshift and k8s stats follow the exact same trend !

screen shot 2018-09-03 at 17 56 48

It seems the --no-insight option is correctly set for Travis builds, however after adding some log in the analytics code, I noticed that it's our unit tests that are causing the issue.

screen shot 2018-08-30 at 18 15 45

The solution is to pass the --no-insight flag to npm test as well.

### DIFF
--- a/generators/statistics.js
+++ b/generators/statistics.js
@@ -128,6 +128,7 @@ class Statistics {
     }
 
     sendYoRc(yorc, isARegeneration, generatorVersion) {
+        if (this.noInsight) return;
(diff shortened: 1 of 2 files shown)


jdubois:
Oh indeed! Thanks a lot

jdubois:
yes indeed, much better!

Issue #8211, opened 2018-09-04
UAA using couchbase won't start [jh 5.3.0]


delkant:
##### **Overview of the issue**
Jhipster 5.3.0 : After generating a UAA service with couchbase, service wont start using ./gradlew

./gradlew

FAILURE: Build failed with an exception.

* Where:
Build file '/home/project/uaa-srv/build.gradle' line: 207

* What went wrong:
A problem occurred evaluating root project 'uaa'.
> Could not find method liquibaseRuntime() for arguments [com.h2database:h2] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

##### **Motivation for or Use Case**
UAA service with couchbase

##### **Reproduce the error**
generate a uaa service with couchbase and gradle and start service running ./gradlew

##### **JHipster Version(s)**

5.3.0
##### **Browsers and Operating System**

- [ ] Checking this box is mandatory (this is just to show you read everything)

ruddell:
Can you please post your `.yo-rc.json` config for the UAA?

pascalgrimaud:
This PR https://github.com/jhipster/generator-jhipster/pull/8212 should fix your issue.

For a quick fix in your project, you can search `h2database` and delete the dependencies, in files:
- build.gradle
- profile_prod.gradle

PR #8212, opened 2018-09-04
Fix H2 and liquibase dependencies for Gradle

pascalgrimaud:
fix https://github.com/jhipster/generator-jhipster/issues/8211

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -474,8 +474,10 @@ dependencies {
     testCompile "com.github.differentway:couchbase-testcontainer"
     <%_ } _%>
     testCompile "org.hamcrest:hamcrest-library"
-    <%_ if (devDatabaseType !== 'h2Disk' && devDatabaseType !== 'h2Memory') { _%>
+    <%_ if (databaseType === 'sql') { _%>
     testCompile "com.h2database:h2"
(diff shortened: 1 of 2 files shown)


generators/server/templates/build.gradle.ejs
```diff
@@ -474,8 +474,10 @@ dependencies {
     testCompile "com.github.differentway:couchbase-testcontainer"
     <%_ } _%>
     testCompile "org.hamcrest:hamcrest-library"
-    <%_ if (devDatabaseType !== 'h2Disk' && devDatabaseType !== 'h2Memory') { _%>
+    <%_ if (databaseType === 'sql') { _%>
```
PierreBesson: I don't think this is right, you have not taken into account sql database with no h2 and the same db in dev and prod.

pascalgrimaud: h2 is used during integration tests for sql database, isn't it ?

pascalgrimaud: In a SQL project, when you launch `mvn test`, it will use a h2 database for testing. That's why the dependency is here.

See the application.yml in test folder: https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/test/resources/config/application.yml.ejs#L61-L64

Issue #8225, opened 2018-09-05
Heroku generator doesn't persist deployType config


jkutner:
##### **Overview of the issue**

After prompting for deploy type (JAR or Git), the Heroku generator does not persist the value of `herokuDeployType` to the `.yo-rc.json`.

The `herokuAppName` value is persisted, and it seems that running `this.config.set(...)` always clobbers previous properties that are set.

##### **Reproduce the error**

Run the `jhipster heroku` command and select "Git". Then rerun `jhipster heroku`.

##### **Suggest a Fix**

I can work around this, but I'd like to figure out why only one config option persists.

##### **JHipster Version(s)**

5.3.0

PR #8226, opened 2018-09-05
Remove fallback behavior for herokuDeployType as a workaround for #8225

jkutner:
This is a workaround for #8225 until a proper fix can be found.

This change will cause the Heroku generator to ask for deployment type *every time*. But that's better than breaking the second time.


jdubois:
Too bad I just released v5.3.1 :-(
Does this need an urgent patch? I'm deploying with Git, so I never run `jhipster heroku` twice, in fact.

jkutner:
@jdubois no, i don't think it's urgent because the deploy will still work (it just might not be the Jar/Git type that people expect). And the manual deploy (i.e. `git push` or `heroku deploy:jar` is always still an option).

jkutner:
@jdubois don't merge this yet. I think i have a fix.

jkutner:
Ok, this should fix the problem.

PR #8229, opened 2018-09-06
[heroku] move Liquibase migrations to release phase

jkutner:
The change disabled liquibase migrations from running inline with the web process, and moves them to the [Heroku Release phase](https://devcenter.heroku.com/articles/release-phase). This will significantly improve startup time, and prevents common problems related to concurrency (i.e. migrations running on more than one server process at the same time).

In more detail, this will add special Liquibase configuration to the Maven profile, and a `release` command to the `Procfile`. Heroku treats the `release` process type specially, and runs it when appropriate (i.e. after a deploy, or when a new database is attached).

The change only affects Git deploy (which will eventually become the default deploy mechanism for this generator).

*NOTE:* This PR includes the changes from #8225 because they were require to get this working. We can choose to close that PR and merge this one, or I can rebase.

## Future work

* Apply this same improvement for Gradle
* Possibly pply a similar improvement for JAR deploys (this is a bit more complicated because the Maven plugin isn't present on the server)

### DIFF
--- a/generators/heroku/templates/pom-profile.xml.ejs
+++ b/generators/heroku/templates/pom-profile.xml.ejs
@@ -1,6 +1,5 @@
             <build>
                 <plugins>
-                    <%_ if (herokuDeployType == 'git' && (prodDatabaseType === 'postgresql' || prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb')) { _%>
                     <plugin>
                         <groupId>org.liquibase</groupId>
                         <artifactId>liquibase-maven-plugin</artifactId>
@@ -18,7 +17,6 @@
                             <promptOnNonLocalDatabase>false</promptOnNonLocalDatabase>
                         </configuration>
                     </plugin>
-                    <%_ } _%>
                     <plugin>
                         <artifactId>maven-clean-plugin</artifactId>
                         <version>${maven-clean-plugin.version}</version>
--- a/generators/heroku/templates/pom-profile.xml.ejs
+++ b/generators/heroku/templates/pom-profile.xml.ejs
@@ -1,5 +1,24 @@
             <build>
                 <plugins>
+                    <%_ if (herokuDeployType == 'git' && (prodDatabaseType === 'postgresql' || prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb')) { _%>
+                    <plugin>
+                        <groupId>org.liquibase</groupId>
+                        <artifactId>liquibase-maven-plugin</artifactId>
+                        <configuration combine.self="override">
+                            <changeLogFile>src/main/resources/config/liquibase/master.xml</changeLogFile>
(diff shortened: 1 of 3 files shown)


jdubois:
What an incredibly good idea @jkutner !!! That's truly awesome!

jkutner:
@atomfrede that sounds right. I'm not as familiar with the Gradle liquibase plugin though.

atomfrede:
@jkutner I hope I find some time during the weekend to implement it. The question is if we need a needle for that or just create a new gradle script for that. Will think about it.

jdubois:
This works incredibly well -> I used to have frequent timeouts, so boot up time was often over 90 seconds. I just did one test, and it booted in 21 secondes!!

jkutner:
@atomfrede i think this could go in the [heroku.gradle.ejs](https://github.com/jhipster/generator-jhipster/blob/master/generators/heroku/templates/heroku.gradle.ejs). would love the help, let me know if i can assist.

atomfrede:
@jkutner Some questions. Currently the heroku gradle plugin is added in version `0.2.0`. Somehow I can only use the new version when using the new `plugins` section not in the classpath section (but will try it later again). Furthermore is the plugin still required? `eploy Gradle based JVM applications directly to Heroku without pushing to a Git repository.` When we want to use git deploy only?

Regarding liquibase I think it can easily be added to `heroku.gradle`

jkutner:
@atomfrede i'm not sure why the heroku-gradle plugin is added. it might be for the CI/CD generator, but it's not used for the JAR and Git deployment. I would say ignore it, or remove it if possible.

generators/heroku/templates/pom-profile.xml.ejs
```diff
@@ -1,5 +1,22 @@
...
+                        <artifactId>liquibase-maven-plugin</artifactId>
+                        <configuration combine.self="override">
+                            <changeLogFile>src/main/resources/config/liquibase/master.xml</changeLogFile>
+                            <diffChangeLogFile>src/main/resources/config/liquibase/changelog/${maven.build.timestamp}_changelog.xml</diffChangeLogFile>
+                            <driver></driver>
+                            <url>${env.JDBC_DATABASE_URL}</url>
+                            <defaultSchemaName></defaultSchemaName>
+                            <username>${env.JDBC_DATABASE_USERNAME}</username>
+                            <password>${env.JDBC_DATABASE_PASSWORD}</password>
+                            <referenceUrl>hibernate:spring:com.mycompany.myapp.domain?hibernate.physical_naming_strategy=org.springframework.boot.orm.jpa.hibernate.SpringPhysicalNamingStrategy&amp;hibernate.implicit_naming_strategy=org.springframework.boot.orm.jpa.hibernate.SpringImplicitNamingStrategy</referenceUrl>
```
atomfrede: Not sure if required, but I think the package should be a variable: `<%=packageName%>`

PR #8240, opened 2018-09-06
Allow Spring boot Scheduled tasks to have multiple threads

ganeshkrishnan1:
The Async configuration currently only allows us to set the async pool size and not the scheduler pool size.
This results in spring scheduled tasks having only one thread to execute all it's tasks.

The proposed change allows us to have the number of scheduler tasks to be the size of the core pool.

Ideally we should have one more configuration value in application.yml which will dictate the size of the executor pool.

See: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/scheduling/annotation/EnableScheduling.html

"By default, will be searching for an associated scheduler definition: either a unique TaskScheduler bean in the context, or a TaskScheduler bean named "taskScheduler" otherwise; the same lookup will also be performed for a ScheduledExecutorService bean. If neither of the two is resolvable, a local single-threaded default scheduler will be created and used within the registrar. "

### DIFF
--- a/generators/server/templates/src/main/java/package/config/AsyncConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/AsyncConfiguration.java.ejs
@@ -29,13 +29,18 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.annotation.*;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.scheduling.annotation.SchedulingConfigurer;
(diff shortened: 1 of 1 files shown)


jdubois:
Indeed, our code isn't homogenous here, we should be able to configure both.
As you write, it would be better to have a new property, even if that's more complex. That means we need to create it in our JHipster Framework project, and it's a bit complex, so we need to discuss this first.
If anybody has comments, please add them here (note to @ganeshkrishnan1 please create a ticket first next time), and otherwise I'll do the JHipster framework part.

jdubois:
Awesome, I'm merging this! Thanks @ganeshkrishnan1

PR #8287, opened 2018-09-11
Reduce Webpack logs in Travis

wmarques:
This reduces the logs in our Travis builds using an environment variable `DISABLE_WEBPACK_LOGS`.
I'm not a fan of adding this into our webpack configurations because I prefer to keep it clean, but I don't know how to do that without doing this...

The `filter(Boolean)` is here to remove the `null` plugins in the configuration, see https://github.com/webpack/webpack/issues/5699


PierreBesson:
Oh and I think DISABLE_WEBPACK_LOG=true should be added to all pipeline configs in the CI-CD subgen. @pascalgrimaud, I'm sure you will want to do it.

pascalgrimaud:
yes, good idea

Issue #8275, opened 2018-09-11
Browsersync keeps refreshing


tibistibi:
##### **Overview of the issue**

When starting the application with yarn start  and editing some files often Browsersync keeps reloading.

##### **Reproduce the error**

run `yarn start` or `npm start` 
i think it gets worse after some changes.

##### **Suggest a Fix**
There seems some suggestions here:
https://stackoverflow.com/questions/37315302/gulp-browsersync-causing-multiple-reloads-with-each-filechange:
```    browserSync.stream({once: true})   ```
or 
```
    browserSync.init({
    server: dist.pages,
    files: dist.css + '*.css',
    watchOptions: {
        awaitWriteFinish : true
    }
});
```
But i'm unclear where i could find the browsersync configuration in jhipster. Maybe someone has some pointers. Or solved this problem already.

##### **Browsers and Operating System**

Ubuntu 18.04 Firfox and Chrome

##### **JHipster Version(s)**

```
stand-bystand@0.0.0 /home/tibi/git/standBystand
└── generator-jhipster@5.3.1 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "nl.tibi.sbys",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.14.4",
    "baseName": "standBystand",
    "packageName": "nl.tibi.sbys",
    "packageFolder": "nl/tibi/sbys",
    "serverPort": "8080",
    "authenticationType": "session",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSocialSignIn": true,
    "enableSwaggerCodegen": false,
    "rememberMeKey": "YourJWTSecretKeyWasReplacedByThisMeaninglessTextByTheJHipsterInfoCommandForObviousSecurityReasons",
    "clientFramework": "angularX",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "monolith",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
[...]

PR #8325, opened 2018-09-14
[Do not merge] Revert "Revert "Fixed: Browsersync keeps refreshing #8275""

jdubois:
Try to fix the current master branch where React tests fail

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -74,7 +74,7 @@
     <%_ } _%>
     "angular-router-loader": "0.8.5",
     "angular2-template-loader": "0.6.2",
-    "browser-sync": "2.24.6",
+    "browser-sync": "2.24.4",
     "browser-sync-webpack-plugin": "2.2.2",
     "cache-loader": "1.2.2",
     <%_ if (protractorTests) { _%>
@@ -144,7 +144,7 @@
     <%_ }); _%>
     "webpack": "4.8.0",
     "webpack-cli": "2.1.3",
-    "webpack-dev-server": "3.1.5",
+    "webpack-dev-server": "3.1.4",
     "webpack-merge": "4.1.2",
     "webpack-notifier": "1.6.0",
     "webpack-visualizer-plugin": "0.1.11",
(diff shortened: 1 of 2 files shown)


vishal423:
@jdubois , there was no change in dependencies and these were working before. This must not cause any issue.

vishal423:
looking at 2 failure build logs, it seems they are due to type mismatch.

> [INFO]  ERROR  Failed to compile with 63 errors21:55:06
> [INFO]
> [INFO]  error  in /home/travis/app/src/main/webapp/app/entities/entity-with-dto/entity-with-dto-delete-dialog.tsx
> [INFO]
> [INFO] ERROR in /home/travis/app/src/main/webapp/app/entities/entity-with-dto/entity-with-dto-delete-dialog.tsx(12,105):
> [INFO] TS2344: Type '{ id: number; }' does not satisfy the constraint '{ id?: string; }'.
> [INFO]   Types of property 'id' are incompatible.
> [INFO]     Type 'number' is not assignable to type 'string'.

jdubois:
Thanks @vishal423 !
Yes I know, I don't understand why this code would have caused any issue. But in Travis it looks like it broke with this commit, so I'm testing if reverting does anything (and I don't intend to merge this PR at all, of course).
I just deleted the Travis cache, in a desperate attempt to try to find the issue. I'm lost here.

vishal423:
did you try to restart those 2 failed builds to see if they go through?

jdubois:
yes, tons of time.... I'm totally lost here

jdubois:
After more tests: the issue seems to come from a transitive dependency! So I'm closing this

pascalgrimaud:
@jdubois : FYI, Travis cache concerns only `.m2` and `.gradle` because the backend libraries are stable enough and we don't need to download them each builds.
See https://github.com/jhipster/generator-jhipster/blob/master/.travis.yml#L16-L19

Intentionaly, there is no cache for `node_modules`. It could speed up a lot all our builds a lot !!! but we won't detect these transitive dependencies issues. That's why I removed the cache for `node_modules` long time ago.

Really sorry if I didn't communicate about it.

jdubois:
Thanks for the information! Too bad we can't have reproductible builds as easily with NPM as with Maven.

PR #8332, opened 2018-09-15
[React] Fix incorrect prefilling when creating new user

### LINKED ISSUE #8302: React: "Create a new user" dialog incorrectly prefills properties with data from a previously edited or detailed user

In the React frontend, under *User Management*, when clicking *Create a new user*, the dialog incorrectly prepopulates user property fields with data from the previously edited or previously in-detail-shown user - making creation of a new user impossible.

A workaround is to go back again (which causes `UserManagementUpdate` to dispatch a `RESET` action on `componentWillUnmount()`), then clicking  *Create a new user* again.

Creating a user should always work.

##### **Reproduce the error**

* Visit vanilla React-based JHipster frontend. Log in as `admin`. Click *Administration* -> *User Management*.
* Either *view* details of an existing user, then click *Back* ... or *edit* an existing user and click *Save*.
* Now being back in the User management interface, click *Create a new user*.
* The *Create or edit a user* dialog is shown and the route is `/admin/user-management/new`. But data is prefilled with data from the previous user, making creation of a new user impossible - as also the ID property is prepopulated - which isn't allowed for new users.

##### **Related issues**
No.

##### **Suggest a Fix**
The problem is that the user edit dialog and user creation dialog both use the `UserManagementUpdate` component. This component uses the Redux state `{user} : UserManagementState`. This state is also handled by  `UserManagementDetail`.

##### **JHipster Version(s)**
generator-jhipster@5.3.1. I don't know if this is a regression.

```
<details>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.example",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "5.3.1",
    "baseName": "tagadata",
    "packageName": "com.example",
    "packageFolder": "com/example",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "jwtSecretKey": "xxx",
    "clientFramework": "react",
    "useSass": true,
    "testFrameworks": [
      "cucumber",
      "protractor"
    ],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": [
      "en",
      "de"
    ]
  }
}
</pre>
</details>
```

Windows 7. Problem occurs at least in Chrome and Firefox, but problem is related to generated React codebase.

chrisdns:
This should fix #8302

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx.ejs
@@ -25,7 +25,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { APP_DATE_FORMAT } from 'app/config/constants';
 <% if (enableTranslation) { %>import { languages } from 'app/config/translation';<% } %>
-import { getUser } from './user-management.reducer';
+import { getUser, reset } from './user-management.reducer';
 import { IRootState } from 'app/shared/reducers';
 
 export interface IUserManagementDetailProps extends StateProps, DispatchProps, RouteComponentProps<{login: string}> {}
@@ -33,6 +33,10 @@ export interface IUserManagementDetailProps extends StateProps, DispatchProps, R
 export class UserManagementDetail extends React.Component<IUserManagementDetailProps> {
   componentDidMount() {
     this.props.getUser(this.props.match.params.login);
+  }
+
+  componentWillUnmount() {
+    this.props.reset();
   }
 
   render() {
@@ -103,7 +107,7 @@ const mapStateToProps = (storeState: IRootState) => ({
   user: storeState.userManagement.user
 });
 
-const mapDispatchToProps = { getUser };
+const mapDispatchToProps = { getUser, reset };
 
 type StateProps = ReturnType<typeof mapStateToProps>;
 type DispatchProps = typeof mapDispatchToProps;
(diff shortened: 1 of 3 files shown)


generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx.ejs
```diff
@@ -35,6 +35,10 @@ export class UserManagementDetail extends React.Component<IUserManagementDetailP
     this.props.getUser(this.props.match.params.login);
   }
 
+  componentWillUnmount() {
+    this.props.reset();
```
deepu105: Actually the fix should be the other way around. The value needs to be reset during the did mount step of create. Can you try that way

chrisdns: I tried it but this does not seem to work

Removing the ``user: action.payload.data`` in the reducer when ``CREATE`` or ``UPDATE`` actions are dispatched is working, or it is also possible to dispatch a reset just after the ``UPDATE`` without removing ``user: action.payload.data`` like the #8338 does, both work

deepu105: ok lets do it in the reducer, the object can be reset on action success

deepu105:
@chrisdns can you take a look at my comment

deepu105:
closing in favor of https://github.com/jhipster/generator-jhipster/pull/8529
@chrisdns see my PR, the issue was with defaultValue

Issue #7756, opened 2018-06-11
Feign Client Annotation Issues


ruddell:
Opening this on behalf of @brunnels who can't open an issue for some reason, I can do the PR later if no one sees any issues.

##### **Overview of the issue**

OAuth2InterceptedFeignConfiguration and OAuth2UserClientFeignConfiguration classes should not be annotated with `@Configuration`

UserFeignClientInterceptor class should not be annotated with `@Component`

##### **Motivation for or Use Case**

Component scan will pickup these configurations and add to the global feign configuration.  The configuration should only apply to clients annotated with `@AuthorizedFeignClient` or `@AuthorizedUserFeignClient`.

UserFeignClientInterceptor is instantiated via `@Bean` annotation in OAuth2UserClientFeignConfiguration so annotating with `@Component` is not necessary and doing so will cause feign autoconfig to add it as an interceptor to all feign clients.

##### **Reproduce the error**

Create a basic `@FeignClient` annotated interface and you will find that it has Authorized headers added by the UserFeignClientInterceptor 

##### **Suggest a Fix**

Remove the annotations.  I have tested and microservice to microservice communication is still working properly after they have been removed.

##### **JHipster Version(s)**

Tested and confirmed in 5.0.0-beta.1.

Annotations are also present in my 4.14.1 projects.

- [X] Checking this box is mandatory (this is just to show you read everything)

mraible:
Seems OK to me.

jdubois:
Impressive work! Let's do this

PR #8333, opened 2018-09-15
Configuration Annotated with configuration

Tcharl:
As this is a configuration, it should be annotated as if...

- [X ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green


vishal423:
It doesn't seems correct as per spring feign docs:

> 
FooConfiguration does not need to be annotated with @Configuration. However, if it is, then take care to exclude it from any @ComponentScan that would otherwise include this configuration as it will become the default source for feign.Decoder, feign.Encoder, feign.Contract, etc., when specified. This can be avoided by putting it in a separate, non-overlapping package from any @ComponentScan or @SpringBootApplication, or it can be explicitly excluded in @ComponentScan.

ruddell:
These were removed in https://github.com/jhipster/generator-jhipster/issues/7756

Tcharl:
I think this is correct: the AuthInterceptor shall be the default RequestInterceptor, no?

Tcharl:
An addition would be to import @Import(FeignClientsConfiguration.class) in FeignConfiguration as well

PR #8352, opened 2018-09-17
Initial jib integration

danielpetisme:

### DIFF
--- a/generators/server/templates/gradle/docker.gradle.ejs
+++ b/generators/server/templates/gradle/docker.gradle.ejs
@@ -18,10 +18,10 @@
 -%>
 buildscript {
     repositories {
-      gradlePluginPortal()
+        gradlePluginPortal()
     }
     dependencies {
-      classpath "gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:0.9.10"
+        classpath "gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:0.9.10"
     }
 }
 
--- a/generators/server/templates/gradle/docker.gradle.ejs
+++ b/generators/server/templates/gradle/docker.gradle.ejs
@@ -29,14 +29,14 @@ apply plugin: com.google.cloud.tools.jib.gradle.JibPlugin
 
 jib {
     from {
-        image = 'openjdk:8-jre-alpine'
+        image = '<%= DOCKER_JAVA_JRE %>'
     }
     to {
         image = '<%= baseName.toLowerCase() %>:latest'
     }
     container {
-        jvmFlags = ['-Djava.security.egd=file:/dev/./urandom']
-        ports = ['8080']
+        entrypoint = ['sh', '-c', 'chmod +x /entrypoint.sh && /entrypoint.sh']
+        ports = ['<%= serverPort %>'<% if (cacheProvider === 'hazelcast') { %>, '5701/udp' <% } %>]
         useCurrentTimestamp = true
     }
 }
--- a/generators/server/templates/gradle/docker.gradle.ejs
+++ b/generators/server/templates/gradle/docker.gradle.ejs
@@ -18,30 +18,27 @@
 -%>
 buildscript {
     repositories {
-        jcenter()
+      gradlePluginPortal()
     }
-
     dependencies {
-        classpath 'com.bmuschko:gradle-docker-plugin:3.5.0'
+      classpath "gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:0.9.10"
     }
 }
 
-apply plugin: com.bmuschko.gradle.docker.DockerRemoteApiPlugin
+apply plugin: com.google.cloud.tools.jib.gradle.JibPlugin
 
-import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
-
-task copyDockerFiles(type: Copy) {
-    description = "Copy Dockerfile and required data to build directory"
-    from 'src/main/docker'
-    from "${project.buildDir}/libs"
-    into { "${project.buildDir}/docker" }
-    include "*"
-    exclude "**/*.yml"
+jib {
+    from {
+        image = 'openjdk:8-jre-alpine'
+    }
+    to {
+        image = '<%= baseName.toLowerCase() %>:latest'
+    }
+    container {
+        jvmFlags = ['-Djava.security.egd=file:/dev/./urandom']
+        ports = ['8080']
+        useCurrentTimestamp = true
+    }
 }
 
-task buildDocker(type: DockerBuildImage, dependsOn: 'copyDockerFiles') {
-    description = "Package application as Docker image"
-    group = "Docker"
-    inputDir = project.file("${project.buildDir}/docker")
-    tags = ["<%= baseName.toLowerCase() %>:latest".toString(), "<%= baseName.toLowerCase() %>:${project.version}".toString()]
-}
+jibDockerBuild.dependsOn bootJar
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -99,7 +99,7 @@
         <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
         <maven-surefire-plugin.version>2.22.0</maven-surefire-plugin.version>
         <maven-war-plugin.version>3.2.2</maven-war-plugin.version>
-        <dockerfile-maven-plugin.version>1.4.4</dockerfile-maven-plugin.version>
+        <jib-maven-plugin.version>0.9.10</jib-maven-plugin.version>
         <%_ if (!skipClient) { _%>
         <frontend-maven-plugin.version>1.6</frontend-maven-plugin.version>
         <%_ } _%>
@@ -858,19 +858,16 @@
                     </execution>
                     <execution>
                         <id>docker-resources</id>
-                        <phase>validate</phase>
+                        <phase>verify</phase>
                         <goals>
                             <goal>copy-resources</goal>
                         </goals>
                         <configuration>
-                            <outputDirectory>target/</outputDirectory>
+                            <outputDirectory>target/classes/static/</outputDirectory>
                             <resources>
                                 <resource>
-                                    <directory><%= MAIN_DIR %>docker/</directory>
+                                    <directory>target/www</directory>
                                     <filtering>false</filtering>
-                                    <excludes>
-                                        <exclude>**/*.yml</exclude>
-                                    </excludes>
                                 </resource>
                             </resources>
                         </configuration>
@@ -1010,57 +1007,25 @@
                 </configuration>
             </plugin>
             <plugin>
-                <groupId>com.spotify</groupId>
-                <artifactId>dockerfile-maven-plugin</artifactId>
-                <version>${dockerfile-maven-plugin.version}</version>
-                <executions>
-                    <!--
-                     Uncomment the section below to build the docker image with mvn package and and push it with mvn deploy
-                    <execution>
-                        <id>default</id>
-                        <goals>
-                            <goal>build</goal>
-                            <goal>push</goal>
-                        </goals>
-                    </execution>
-                    -->
-                    <execution>
-                        <id>tag-latest</id>
-                        <phase>none</phase>
-                        <goals>
-                            <goal>build</goal>
-                            <goal>tag</goal>
-                            <goal>push</goal>
-                        </goals>
-                    </execution>
-                    <execution>
-                        <id>version</id>
-                        <phase>none</phase>
-                        <goals>
-                            <goal>build</goal>
-                            <goal>tag</goal>
-                            <goal>push</goal>
-                        </goals>
-                        <configuration>
-                            <tag>${project.version}</tag>
-                        </configuration>
-                    </execution>
-                    <execution>
-                        <id>commit</id>
-                        <phase>none</phase>
-                        <goals>
-                            <goal>build</goal>
-                            <goal>tag</goal>
-                            <goal>push</goal>
-                        </goals>
-                        <configuration>
-                            <tag>${git.commit.id.describe}</tag>
-                        </configuration>
-                    </execution>
-                </executions>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>jib-maven-plugin</artifactId>
+                <version>${jib-maven-plugin.version}</version>
                 <configuration>
-                    <repository><%= baseName.toLowerCase()%></repository>
-                    <contextDirectory>${project.build.directory}</contextDirectory>
+                    <from>
+                        <image>openjdk:8-jre-alpine</image>
+                    </from>
+                    <to>
+                        <image><%= baseName.toLowerCase() %>:latest</image>
+                    </to>
+                    <container>
+                        <jvmFlags>
+                            <jvmFlag>-Djava.security.egd=file:/dev/./urandom</jvmFlag>
+                        </jvmFlags>
+                        <ports>
+                            <port>8080</port>
+                        </ports>
+                        <useCurrentTimestamp>true</useCurrentTimestamp>
+                    </container>
                 </configuration>
             </plugin>
             <%_ if (enableSwaggerCodegen) { _%>
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1012,17 +1012,22 @@
                 <version>${jib-maven-plugin.version}</version>
                 <configuration>
                     <from>
-                        <image>openjdk:8-jre-alpine</image>
+                        <image><%= DOCKER_JAVA_JRE %></image>
                     </from>
                     <to>
                         <image><%= baseName.toLowerCase() %>:latest</image>
                     </to>
                     <container>
-                        <jvmFlags>
-                            <jvmFlag>-Djava.security.egd=file:/dev/./urandom</jvmFlag>
-                        </jvmFlags>
+                        <entrypoint>
+                            <shell>sh</shell>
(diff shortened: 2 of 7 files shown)


danielpetisme:
Here's a proposal to use jib ad Docker image producer.
Why use jib:
* The image building is smarter than just use a huge fat jar (a thinner app layer thus a faster build)
* because it's hype

You can test it via
`./mvnw clean verify jib:dockerBuild` and `./gradlew jibDockerBuild`

I don't know if we should remove the `Dockerfile`, devs are comfortable with it so removing the file right can be a bit too agressive.

jdubois:
@danielpetisme the current master is pretty broken, you'll need to rebase - we are fixing this at this very moment!

danielpetisme:
Ok let me know when I should rebase.

Le lun. 17 sept. 2018 à 22:31, Julien Dubois <notifications@github.com> a
écrit :

jdubois:
That should be good now... Still waiting for the build to succeed but I'm very confident

pascalgrimaud:
One good thing: we'll be consistent between Maven and Gradle.

If we decide to use Jib, does it mean the Dockerfile will completely disapear ?
Can we do the same thing than Dockerfile inside pom.xml and docker.gradle ?

generators/server/templates/gradle/docker.gradle.ejs
```diff
@@ -18,30 +18,27 @@
 -%>
 buildscript {
     repositories {
-        jcenter()
+      gradlePluginPortal()
```
pascalgrimaud: indent, 2 more spaces

generators/server/templates/gradle/docker.gradle.ejs
```diff
@@ -18,30 +18,27 @@
...
     }
-
     dependencies {
-        classpath 'com.bmuschko:gradle-docker-plugin:3.5.0'
+      classpath "gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:0.9.10"
```
pascalgrimaud: indent

generators/server/templates/gradle/docker.gradle.ejs
```diff
@@ -18,30 +18,27 @@
...
-task copyDockerFiles(type: Copy) {
-    description = "Copy Dockerfile and required data to build directory"
-    from 'src/main/docker'
-    from "${project.buildDir}/libs"
-    into { "${project.buildDir}/docker" }
-    include "*"
-    exclude "**/*.yml"
+jib {
+    from {
+        image = 'openjdk:8-jre-alpine'
```
pascalgrimaud: there is a constant for the base image

generators/server/templates/gradle/docker.gradle.ejs
```diff
@@ -18,30 +18,27 @@
...
+    }
+    to {
+        image = '<%= baseName.toLowerCase() %>:latest'
+    }
+    container {
+        jvmFlags = ['-Djava.security.egd=file:/dev/./urandom']
+        ports = ['8080']
```
pascalgrimaud: the port should be a variable, serverPort, from the config

generators/server/templates/pom.xml.ejs
```diff
@@ -1010,57 +1007,25 @@
...
-                    </execution>
-                </executions>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>jib-maven-plugin</artifactId>
+                <version>${jib-maven-plugin.version}</version>
                 <configuration>
-                    <repository><%= baseName.toLowerCase()%></repository>
-                    <contextDirectory>${project.build.directory}</contextDirectory>
+                    <from>
+                        <image>openjdk:8-jre-alpine</image>
```
pascalgrimaud: same comment: use the constant

generators/server/templates/pom.xml.ejs
```diff
@@ -1010,57 +1007,25 @@
...
+                    </from>
+                    <to>
+                        <image><%= baseName.toLowerCase() %>:latest</image>
+                    </to>
+                    <container>
+                        <jvmFlags>
+                            <jvmFlag>-Djava.security.egd=file:/dev/./urandom</jvmFlag>
+                        </jvmFlags>
+                        <ports>
+                            <port>8080</port>
```
pascalgrimaud: same, serverPort here

pascalgrimaud:
The entrypoint has been removed and it contains the famous hack: sleep Xxx.

That means the container will always fail the first time we launch it as the other services are not up when starting the docker compose file.

generators/server/templates/gradle/docker.gradle.ejs
danielpetisme: hum... I must say I'm not sure extracting the image and port values in `gradle.properties` worth it.
1- They are already consider as constant in the actual file and it does not looks like a limitation
2- It forces to work on 2 files to define a docker build
3- for consistency sake, it we do it for `from.image` and `ports` why not for the other constants like `jvmFlag` or even the target image name

generators/server/templates/pom.xml.ejs
danielpetisme: Makes more sense for Maven because the `pom.xml` is huge and it's easy to maintain only top level declarations.

generators/server/templates/pom.xml.ejs
danielpetisme: regarding the context of Spring and the server.port default value I think we could consider 8080 as a constant (just as its did in the actual Dockerfile)

danielpetisme:
@pascalgrimaud sorry for the indent...

Regarding the entrypoint, 2 options:
* build a oneLiner to start the app (`export JHIPSTER_SLEEP=0 && java -cp /app/resources:/app/classes:/app/libs/* MyMainClass`
* use jib [`extraDirectory`](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#adding-arbitrary-files-to-the-image) to provide a custom `entrypoint.sh` file

I would rather go for the `extraDirectory` option to composensate the lack of expresivness of the jib syntax (if could use the sh file to create a jhipster user, home etc.)

WDYT?

generators/server/templates/gradle/docker.gradle.ejs
PierreBesson: @danielpetisme You need to use the generator constants like docker base image, server port, etc. @pascalgrimaud does not mean extracting those values to a property file just using the right EJS templating.

sendilkumarn:
@danielpetisme awesome 🎉

PierreBesson:
Yes the Dockerfile need to be removed for this but it's not a huge problem.
For the entrypoint, I'm afraid we need it for JHIPSTER_SLEEP (which should actually be replaced by proper wait-for scripts in the future) but it should not be used to create the user.

Also when using jib, you want to use it's ability to create docker image without a docker daemon (yes just with maven !) so it means that we can't do typical dockerfile "RUN" commands as those need to be executed in a docker environment.

generators/server/templates/gradle/docker.gradle.ejs
danielpetisme: Ok make sense. Will do
Sorry for the misunderstanding.

pascalgrimaud: I was commenting with my phone and wasn't clear enough sorry
See https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/docker/Dockerfile.ejs#L19
We should use `image = '<%= DOCKER_JAVA_JRE %>'` instead

generators/server/templates/pom.xml.ejs
pascalgrimaud: here: `<%= DOCKER_JAVA_JRE %>`

generators/server/templates/pom.xml.ejs
pascalgrimaud: See https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/docker/Dockerfile.ejs#L37

And we should add another line for hazelcast too.
I think it should be something like that:

```
<ports>
    <port><%= serverPort %></port>
    <port>5701/udp</port> <--- need to check if the syntax is correct
</ports>
```

danielpetisme:
I'm working on an external `entrypoint.sh` I have 2 concerns:

- How can I force yeoman to keep the file attributes or force a chmod on the generated file. Otherwise I have to add a dirty chmod in the entrypoint 'sh -c chmod +x /entrypoint.sh && /entrypoint.sh'. 
See https://github.com/jhipster/generator-jhipster/pull/8352/files#diff-e17609d40a091dddab3de9f5b67c1771R38

- The current version of jib (ie. 0.9.10) has no support for declare env variables. It will be release with the next '0.9.11' version. So either we wait (I didn't contact the project lead) or as intermediate solution I add a poorman env var management in the `entrypoint.sh` script.
See https://github.com/jhipster/generator-jhipster/pull/8352/files#diff-775f634a8405a2e1b1fd139ce10bf828R4

I didn't do any cleaning yet. I prefer to finish the jib integration first before removing stuff.

PierreBesson:
@danielpetisme To chmod +x the generated template file, I used `fs.chmodSync()` after generation ([like here in the k8s subgen](https://github.com/jhipster/generator-jhipster/blob/master/generators/kubernetes/index.js#L243)). It seems to be the only way that work reliably for all OSes.

For env vars, let's keep it as you did and migrate when the new jib release to come out.

In the end I'm more confident with jib than other plugins considering that this gets a lot of resources from Google.

danielpetisme:
Jib does not take the source file permission into account. That means even if I chmod the file before the build it won't be executable in the container.
There is an open issue (still under discussion):
😢 https://github.com/GoogleContainerTools/jib/issues/794

The hack I proposed (include the chmod in the entrypoint) works even if it's ugly.

Do you want guys to go with the hack?

PierreBesson:
@danielpetisme I noticed a big issue on your PR. The frontend was actually not packaged and served by the docker container produced by jib. Hopefully I have found the fix and submitted a PR to your branch.

PierreBesson:
Also I guess once jib add support for env vars, we might not need the entrypoint script anymore and instead use the jib `<entrypoint>` property.

danielpetisme:
@PierreBesson I had a looked to your PR.
I'm confused because I've explicitly added the copy of the frontend.

For maven the copy is bound to the `verify` phase
https://github.com/danielpetisme/generator-jhipster/blob/jib/generators/server/templates/pom.xml.ejs#L860
which means you need to run `./mvnw verify jib:docker:build` to  have the frontend packaged

For gradle, the `jibDockerBuild` relies on `bootJar`. If you look to the `profile_dev/prod`.gradle you'll find the `copyIntoStatic` task (trigger by `bootJar`)

So normally both build tools are properly configured.

Now, your PR relies directly on webpack to generate the frontend files on the appropriate folders for jib.
I do prefer your solution. I'll integrate your PR and remove my hacks.

danielpetisme:
Regarding the entrypoint. Once the env vars supported by jib will be able to provide a one liner: 
`echo "The application will start in ${JHIPSTER_SLEEP}s..." && sleep ${JHIPSTER_SLEEP} && exec java ${JAVA_OPTS} -Djava.security.egd=file:/dev/./urandom -cp /app/resources/:/app/classes/:/app/libs/* "<%= packageName %>.<%= mainClass %>"  "$@"` 

as entrypoint.
The next steps would be to remove the `sleep` hack.

PierreBesson:
I see ! It did not work for me because I was using `mvn package` and `gradle bootWar` which are the recommended goals in our docs/readme.

danielpetisme:
@PierreBesson thanks for your PR it's way cleaner

If you want to test:
* `./mvnw -Pprod package jib:dockerBuild`
* `./gradlew -Pprod bootWar jibDockerBuild`

If everbody is ok with the implementation, time to go into details:
* Everybody is ok with a `:latest` tag or should I use the project version?  
* Any proposal on how to test the feature? (so far the `server` generator only assert file creation)

PierreBesson:
So far we were able to set only one docker tag so it was only "latest" but if possible I would like to set the "latest" "project version" and "git version" tags.

Yes the "server" tests only test generation of the files but the Travis builds will actually build and package the app. You have to look into Travis tests scripts to add a step to run the `jibDockerBuild` task.

danielpetisme:
**Docker Tagging**
Sor far we can't do multi-tagging, I propose to stay with `:latest` right now

**Testing**
I'll extend https://github.com/jhipster/generator-jhipster/blob/master/travis/scripts/05-run.sh to  `docker run` and curl (maybe later it will make sense to split tha `java` run and docker `run` in 2 separate files)

deepu105:
While this is good, I do get the feeling that its more restrictive and less
features than current setup, maybe we should wait. But if you guys feel
that the pros outweigh the co9n then its fine

Thanks & Regards,
Deepu

PierreBesson:
@deepu105, to sum up the pro of jib are:

- a huge optimisation and reuse in terms of docker layer, which means that when building and pushing a new docker image with changes to java code, it will only adds a kew kbs to exisiting layers.
- the ability to build and push docker images without a Docker daemon (great for CI)
- maintained by Google with new features frequently
- no need for hacks like .dockerignore

The cons:

- Slightly more complex (internally)
- Lose the flexibility/familliarity of the Dockerfile
- Not able to specify your own layers, run commands as part of the docker build.

danielpetisme:
@deepu105 I'm not sure to get your point. The images generated by the actual Docker plugin and jib are equivalent in terms of functions but they both run provide an `entrypoint` with run trigger the `java` process. 
In one hand the Gradle plugin run the fat jar in the other one jib build the classpath an run the application main class.

I would say jib cover 100% of the Gradle/Maven Docker plugins scope

I'm not an expert of Azure/AWS,K8s container prerequisites. Do you have any pointer I could use to verify the actual jib image compliance?

danielpetisme:
@PierreBesson as promised I've added some tests in the Travis build 😉 
After building the docker image with jib I start the app (via docker-compose) then I try to curl the app.
Sounds good?

travis/scripts/05-run.sh
```diff
@@ -124,3 +124,35 @@ if [ "$RUN_APP" == 1 ]; then
...
+            exit 0
+        fi
+        if [ $? -ne 0 ]; then
+            echo "Error when building Docker image"
+            exit 1
+        fi
+
+        cd "$APP_FOLDER"
+        docker-compose -f src/main/docker/app.yml up -d
+        sleep 40
```
PierreBesson: I'm afraid this will slow down our Travis build...

pascalgrimaud: I think for this kind of tests, it should go to daily builds at https://github.com/hipster-labs/jhipster-travis-build
See https://travis-ci.org/hipster-labs/jhipster-travis-build/branches

danielpetisme: @pascalgrimaud :+1: but I don't understand the branch organization (the docker branch is not synchronized with the master).
I'll remove the above `docker run` snippet to move to jhipster-travis-build

danielpetisme:
@pascalgrimaud I've create a gist from the `docker run` snippet
https://gist.github.com/danielpetisme/babf809ae0e7cb3aa8a500b6d8ef1c40

Let me know if I can help on the `travis-build`.

PierreBesson:
For some reason it seems this pr contains commits from other branches. It is normal ?

danielpetisme:
@PierreBesson because I suck at rebasing. It looks like I introduced duplicate commits. I'll clean it ASAP

DanielFran:
@danielpetisme I cancel 1 of the travis tests, since they were another one available.

danielpetisme:
@PierreBesson I fix my mess... Forgot to rebase somewhere and jhipster basecode is moving really fast 😉 
Should I squash in one commit?

PierreBesson:
No need to rebase more than what you already did.

PierreBesson:
Thanks a lot @danielpetisme !

pascalgrimaud:
Thanks! Now, we need to update:
- documentation: https://github.com/jhipster/jhipster.github.io
- zsh: https://github.com/jhipster/jhipster-oh-my-zsh-plugin/blob/master/jhipster.plugin.zsh#L65-L71
- fisherman: https://github.com/jhipster/jhipster-fisherman-plugin/blob/master/conf.d/jhipster.aliases.fish#L65-L71

pascalgrimaud:
@danielpetisme : can it work when my computer is offline without internet ? I saw this issue: https://github.com/GoogleContainerTools/jib/issues/718

briandealwis:
@danielpetisme you [wrote above](https://github.com/jhipster/generator-jhipster/pull/8352#issuecomment-422153788):
> What should be the default image tag?
> jib tagging is quite basic, I can only provide a target image name with a tag suffix (i.e myimage:latest or myimage:some-tag). Should I use :latest, :${project.version}, something else?

How could jib's tagging be improved?

danielpetisme:
@briancmills 
Previously jhipster used 2 tags `latest` and `${project.version}` maybe jib could propose something like
```
<to>
  <image>my-image</image>
  <tags>
    <tag>latest</tag>
    <tag>${project.version}</tag>
  </tags
</to>
```

briandealwis:
Jib now supports multiple tags with 0.9.11 (just released) 🎉

PierreBesson:
I would personally like to have the following tags: latest, project.version and git.commit.id.describe.

briandealwis:
You might want to use image labels to record that information too.

danielpetisme:
There is an open question: Dockerfile or not Dockerfile...

IMHO, jib is still a young project to force a big bang migration.
I would keep Dockerfile for some releases.

To encourage jib adoption I would apply the "don't feed the Monster" strategy and block Dockerfile enhancement PR.

deepu105:
I agree lets keep it for some time

PR #8402, opened 2018-09-25
Chunk Order Fix

tbickford:
Fixing the order that chuncks are loaded so that the global file is loaded after the main.  This will ensure CSS overrides of bootstrap in the global.css/global.sass work.  https://github.com/jantimon/html-webpack-plugin#options

### DIFF
--- a/generators/client/templates/angular/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.common.js.ejs
@@ -108,7 +108,7 @@ module.exports = (options) => ({
         <%_ } _%>
         new HtmlWebpackPlugin({
             template: './<%= MAIN_SRC_DIR %>index.html',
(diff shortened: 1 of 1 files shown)


deepu105:
I think originally the order was to overcome some weird issue, but that may
not be applicable anymore. Regardless this needs testing in both dev and
prod profile

wmarques:
@deepu105 Yep the issue was when we try to lazy load 2 or more modules the app was stuck in an infinite loop so I had to manually order the chunks.
Need to test that case before merging this but I don't think it should cause an issue since they are manually sorted.

deepu105:
@wmarques can you test this?

wmarques:
@deepu105 Done, I created another lazy loaded module and it works fine, thanks @Tcharl  !

Issue #6735, opened 2017-11-25
Bidrectional OneToOne relationship get two entries in the combobox


agoncal:
##### **Overview of the issue**

Let's say you have a required OneToOne relationship between a `SponsorAgreement` and a `Sponsor`. After creating a `SponsorAgreement`, when you edit it, you see the same `Sponsor` twice in the combobox.

##### **Motivation for or Use Case**

When you create a `SponsorAgreement` you only have one sponsor in the combox. When you edit it, you can see the same sponsor twice.

##### **Reproduce the error**

Generate a simple monolith with the following model : 

```
entity Sponsor {
    name String
}

entity SponsorAgreement {
    total Float
}

relationship OneToOne {
    SponsorAgreement{sponsor required} to Sponsor{sponsorAgreement}
}
```

Once the application generated, do the following steps:

* Create a new `Sponsor` (it gets the id 1)
* Create a new `SponsorAgreement`
  * In the sponsor combobox you only see the sponsor 1
* Edit the sponsor agreement
  * In the sponsor combobox you see the sponsor 1 **twice**

##### **Suggest a Fix**

On the `ngOnInit` method of the `SponsorAgreementDialogComponent`, the `subscribe` does a concatenation (`this.sponsors = [subRes].concat(res.json);`). I don't know why it does that, there's maybe a good reason, but it looks like it could be the problem: 

```
.subscribe((res: ResponseWrapper) => {
    if (!this.sponsorAgreement.sponsor || !this.sponsorAgreement.sponsor.id) {
        this.sponsors = res.json;
    } else {
        this.sponsorService
            .find(this.sponsorAgreement.sponsor.id)
            .subscribe((subRes: Sponsor) => {
                this.sponsors = [subRes].concat(res.json);
            }, (subRes: ResponseWrapper) => this.onError(subRes.json));
    }
```

##### **JHipster Version(s)**

```
bugdouble@0.0.0 /Users/agoncal/Documents/Code/Temp/jhipsterdouble
└── generator-jhipster@4.11.0 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "io.bugdouble"
    },
    "jhipsterVersion": "4.11.0",
    "baseName": "bugdouble",
    "packageName": "io.bugdouble",
    "packageFolder": "io/bugdouble",
[...]

PR #8411, opened 2018-09-27
Try to fix "Bidrectional OneToOne relationship get two entries"

jdubois:
See #6735


deepu105:
@jdubois does React also have the issue?

jdubois:
@deepu105 not exactly: there is a list with the normal items, with at the top an empty item. As the relationship is required, selecting the empty item gives a validation error `javax.validation.ConstraintViolationException: Validation failed`

ruddell:
Did you try with a snapshot of the `jhipster` server library?  I think this solution requires some of the latest PRs that were merged but not released yet

jdubois:
Thanks @ruddell - I didn't know that, so no I didn't test this. I'll try to do that today.

jdubois:
@ruddell sorry but no, it doesn't change anything

kaidohallik:
I tested it and seems that current `jhipster` server library still doesn't support `specified` on references.  
I tried solution in PR jhipster/jhipster#59 again and this worked again for me.

duderoot:
I add a new comment on my previous PR in [jhipster-framework/#59](https://github.com/jhipster/jhipster/pull/59) for @gzsombor in order to know if there is a possibility to build the query that we need with the latest version of QueryService that hi refactored a few days ago.

Issue #7722, opened 2018-06-02
Docker image generation issues


gclayburg:
##### **Overview of the issue**

I have been using jhipster for a while now and love the project.  There are certainly top-notch integration techniques with many technologies here.

However, I noticed a few issues with the way docker images are generated.  I have fixed these myself and I'm curious if this project might like to use some of these as a default.  Here are the issues I ran into:

- The app running in docker cannot respond to OS signals like TERM or QUIT.  The impact is that running a `docker stop` command will take a little longer to shutdown the app while it times out and sends a KILL signal.  This means that your app will not get a chance to run any Spring @Predestroy Java beans and your app may not shutdown correctly.
- The entire app war file is added as a single docker layer.  This can lead to slower build times and slower docker image pushes and pulls.  This is especially true when you need to push or pull from a remote repository.  If you change one line of code, the generated docker image layer will be dozens of MBs, even though most of it is unchanged from the previous build.
- The docker image does not allow command line options like this:
```
$ docker run synconsole:latest --spring.data.mongodb.uri=mongodb://syncuser:qwerty1234@yale:27017/synconsoledev
```
You can however pass environment variables with the docker --env flag.
- The docker image runs as the root user inside the container.  Applications are somewhat more secure when running as a normal unprivileged user.
-  The FROM line in the Dockerfile uses ':8-jre-alpine' as the version tag.  Currently, this tag is the same as ':8u151-jre-alpine' on Docker Hub, but chances are it will change as it has in the past when new versions are released.  This can lead to non-deterministic builds if your local docker installation already has an older ':8-jre-alpine' 

##### **Suggest a Fix**
See follow-up below

##### **JHipster Version(s)**
4.14.4 using gradle

##### **Browsers and Operating System**

- [x ] Checking this box is mandatory (this is just to show you read everything)

gclayburg:
I have fixed this issues in my project with this alternate  file `gradle/docker.gradle`

```groovy
buildscript {
    repositories {
        jcenter()
        maven {
            url "https://plugins.gradle.org/m2/"
        }
    }

    dependencies {
        classpath 'com.bmuschko:gradle-docker-plugin:3.2.0'
        classpath 'gradle.plugin.com.garyclayburg:dockerPreparePlugin:1.3.2'
    }
}

apply plugin: com.bmuschko.gradle.docker.DockerRemoteApiPlugin

import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage

[...]

PR #6744, opened 2017-11-26
Improve Docker image cache usage by reordering commands in app template Docker file

smasset:
- Please make sure the below checklist is followed for Pull Requests.

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [ ] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #8409, opened 2018-09-26
Move to jib 0.9.11

danielpetisme:
-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

This upgrade permits to declare environment variables directly in the build tool file (Maven/Gradle) rather than a shell script hack.

if you want a deeper look on the issue fixed with `jib 0.9.11` check the [Milestone report](https://github.com/GoogleContainerTools/jib/milestone/26)

PR #8413, opened 2018-09-28
Re introduced Docker image cache improvments from #6744

smasset:
These improvements were lost when merging #7722.

@gclayburg, @PierreBesson was this intentional. If not, please refer to #6744 for explanation.

### DIFF
--- a/generators/server/templates/src/main/docker/Dockerfile.ejs
+++ b/generators/server/templates/src/main/docker/Dockerfile.ejs
@@ -30,9 +30,9 @@ ADD entrypoint.sh entrypoint.sh
 RUN chmod 755 entrypoint.sh && chown jhipster:jhipster entrypoint.sh
 USER jhipster
 
-ADD *.war app.war
-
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Thanks @smasset 
I'm merging this, but not sure it is used now. Do you see this PR https://github.com/jhipster/generator-jhipster/pull/8409 ?

PierreBesson:
I don't understand. Now that we have migrated to jib, we should not have a Dockerfile anymore, the file should be removed.

pascalgrimaud:
The real PR is https://github.com/jhipster/generator-jhipster/pull/8352

@PierreBesson : see @danielpetisme's question, we didn't answer, so that's why the file is still here:
> Should I remove the existing Dockerfile?
Same as above, Devs are comfortable with it so removing the file right now can be a bit too agressive. Plus I didn't check yet if it's not used somewhere else.

smasset:
Wow, that was quick. Thanks.
@pascalgrimaud, I'll check the mentioned PR.

PR #8421, opened 2018-09-29
New env var for e2e in headless mode

pascalgrimaud:
When launching Protractor in Azure DevOps, I need to use Chrome in Headless mode.
So I added a new environment variable in Protractor config for that, beginning with `JH_` , so few chance to have same variable in their environment

I rename `DISABLE_WEBPACK_LOGS` to `JH_DISABLE_WEBPACK_LOGS` to be consistent (recently added by @wmarques)

```
export JH_E2E_HEADLESS=true
npm run e2e
```

Another solution would be to create a new goal in `package.json`, like `e2e-headless` but I don't know how to achieve this.

_____

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ env:
     # if JHIPSTER_BRANCH value is release, use the release from NPM
     - JHIPSTER_REPO=https://github.com/jhipster/generator-jhipster.git
     - JHIPSTER_BRANCH=release
-    - DISABLE_WEBPACK_LOGS=true
+    - JH_DISABLE_WEBPACK_LOGS=true
 
   matrix:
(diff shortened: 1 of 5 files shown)


pascalgrimaud:
I'm merging this, as Deepu approved it.

jdubois:
Don't we usually use "JHI_" for prefix? That's something we need to discuss more globally anyway, and one of my goals for JHipster 6. As we still have issues with reserved keywords a bit everywhere, and having a clean prefix like "JHI_" would probably solve it (at the expense of being "tied" to JHipster, but that shouldn't be a big problem).

pascalgrimaud:
Indeed `JHI_` is better.
I'll do the changes in Travis (and my Azure DevOps branch)

PR #8429, opened 2018-10-01
CI: rename JH_var to JHI_var

pascalgrimaud:


PierreBesson:
Seriously why not "JHIPSTER" in full ?

pascalgrimaud:
@PierreBesson : why not ? I'm not against. I fixed this to be consistent with the other part of JHipster, but as it only concerns env var in Travis (and Azure soon), we could use JHIPSTER_* here
But we need to be agree with this.

PierreBesson:
Well it's not only about Travis as we ship this in user generated code, so we should be extra explicit.

PR #8285, opened 2018-09-11
Set timeout to 1ms for Angular Protractor tests in Travis

hdurix:
- Please make sure the below checklist is followed for Pull Requests.

- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [x] Tests are added where necessary
- [x] Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #8436, opened 2018-10-02
Travis: fix speed up for noi18n + protractor

pascalgrimaud:
Following https://github.com/jhipster/generator-jhipster/pull/8285
Ping @hdurix 

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/shared-libs.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/shared-libs.module.ts.ejs
@@ -31,7 +31,7 @@ import { FontAwesomeModule } from '@fortawesome/angular-fontawesome';
         NgJhipsterModule.forRoot({
             // set below to true to make alerts look like toast
             alertAsToast: false,
-            alertTimeout: 5,
+            alertTimeout: 5000,
             <%_ if (enableTranslation) { _%>
             i18nEnabled: true,
             defaultI18nLang: '<%= nativeLanguage %>'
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/shared-libs.module.ts.ejs
```diff
@@ -31,6 +31,7 @@ import { FontAwesomeModule } from '@fortawesome/angular-fontawesome';
         NgJhipsterModule.forRoot({
             // set below to true to make alerts look like toast
             alertAsToast: false,
+            alertTimeout: 5,
```
deepu105: Want this in milliseconds? Sorry I'm on my fone. If it indeed is millis then 5 is too low

pascalgrimaud: oh indeed, I think it's millis ! So it should be 5000

hdurix:
Why is it needed? What is the exception with noi18n?

pascalgrimaud:
@hdurix : it's for noi18n project.

For the `shared-libs.module.ts.ejs` file, JHipster will generate:

```typescript
NgJhipsterModule.forRoot({
    // set below to true to make alerts look like toast
    alertAsToast: false,
}),
```

After the generation, the `lint` will change to:

```typescript
NgJhipsterModule.forRoot({
    // set below to true to make alerts look like toast
    alertAsToast: false
}),
```

So the replace won't work, because of missing char `,` :

```
sed -e 's/alertAsToast: false,/alertAsToast: false, alertTimeout: 1,/1;' src/main/webapp/app/shared/shared-libs.module.ts > src/main/webapp/app/shared/shared-libs.module.ts.sed
```

PR #8450, opened 2018-10-03
Update husky/lint-staged version

### LINKED ISSUE #8448: Update husky and use its new configuration format

Update husky to last stable release and use its new configuration format.

JHipster currently use husky@0.14.3. 
There's now a stable release (1.0) and even a 1.1, that uses a new configuration format, either in package.json or in a dedicated file: .huskyrc.

Falydoor:
Fix #8448

I noticed an issue during the generation during the automatic `npm install`

> husky > setting up git hooks
Can't find .git, skipping Git hooks installation.
Please check that you're in a cloned repository or run 'git init' to create an empty Git repository and reinstall husky.

Does it make sense to move the git init before `npm install`?


murdos:
What about moving hooks configuration in a separate file .huskyrc?
That would avoid duplication in package.json for angular and react.

generators/client/templates/angular/package.json.ejs
```diff
@@ -162,11 +162,13 @@
...
     ]
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
   <%_ } _%>
   "scripts": {
-    <%_ if(!skipCommitHook) { _%>
```
deepu105: Please retain the condition in the new husky block as well

Falydoor: It is the case, the husky blocked is merged with the lint-staged one (line 158).

generators/client/files-angular.js
```diff
@@ -47,6 +47,10 @@ const files = {
                 'postcss.config.js',
                 { file: 'webpack/logo-jhipster.png', method: 'copy' }
             ]
+        },
+        {
+            condition: generator => !generator.skipCommitHook,
+            templates: [{ file: '../.huskyrc', method: 'copy', noEjs: true, renameTo: () => '.huskyrc' }]
```
deepu105: this is weird, why do you have `../` in file name? and you dont have to rename it unless there is a change

deepu105: This should be fine `{ file: '.huskyrc', method: 'copy' }` and also where is the actual file? did you forget to commit it?

murdos:
> I noticed an issue during the generation during the automatic `npm install`
> 
> > husky > setting up git hooks
> > Can't find .git, skipping Git hooks installation.
> > Please check that you're in a cloned repository or run 'git init' to create an empty Git repository and reinstall husky.
 
I noticed this issue too with current version of JHipster and husky. 
I've created #8457 to be sure it's not forgotten.

generators/client/files-angular.js
Falydoor: The `../` in the filename is because I wanted to have only one file in the `templates` folder instead of one in `templates/react` and `templates/angular`.

So for angular, files are created with `this.writeFilesToDisk(files, this, false, this.fetchFromInstalledJHipster('client/templates/angular'));` so the `../` is to go back to `templates`. And then the renameTo is to avoid having the file in the parent directory of the project.

It's a little hacky but that's a way I found to avoid duplicating `.huskyrc` (or in `package.json`).

I see two other ways:
- Copy the file in the `write` method before the switch:

```javascript
// Public API method used by the getter and also by Blueprints
_writing() {
    return {
        write() {
            // COPY .HUSKYRC HERE
            switch (this.clientFramework) {
                case 'react':
                    return writeReactFiles.call(this, useBlueprint);
                default:
                    return writeAngularFiles.call(this, useBlueprint);
            }
        }
    };
}
```

- Call `this.writeFilesToDisk('.huskyrc', this, false, this.fetchFromInstalledJHipster('client/templates'));` in `file-angular.js` and `file-react.js`.

Which way you think is the best ?

deepu105: Please duplicate and put the files in respective folders for now, so that we don't have weird behaviors when using blueprints and stuff. We have some ideas for moving common code out but that's a bigger story for later

Falydoor:
@deepu105 I did the changes for the .huskyrc file.

PR #8461, opened 2018-10-04
[React] Fix webpack configuration module/rules/exclude

victorfu:
Follow the best practices mentioned in [https://webpack.js.org/configuration/](https://webpack.js.org/configuration/)
- Use RegExp only in test and for filename matching
- Use arrays of absolute paths in include and exclude

### DIFF
--- a/generators/client/templates/react/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/react/webpack/webpack.common.js.ejs
@@ -75,7 +75,7 @@ module.exports = options => ({
         test: /\.tsx?$/,
         use: getTsLoaderRule(options.env),
         include: [utils.root('./src/main/webapp/app')],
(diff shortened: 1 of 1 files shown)


generators/client/templates/react/webpack/webpack.common.js.ejs
```diff
@@ -75,7 +75,7 @@ module.exports = options => ({
         test: /\.tsx?$/,
         use: getTsLoaderRule(options.env),
         include: [utils.root('./src/main/webapp/app')],
-        exclude: /'node_modules'/
+        exclude: [path.resolve(__dirname, "node_modules")]
```
deepu105: I think utils.root does the same, can you check and use that method

victorfu: ok! let me check.

PR #8468, opened 2018-10-05
fix docker start with jib : added a sync before starting /entrypoint.sh

hcouplet:
When running pod ok k8s, I have an error : (on my local docker it's fine)
   sh: /entrypoint.sh : Text file busy

image

Looking in https://github.com/moby/moby/issues/9547,  it seems that the script can't start because the file is still busy with chmod.
I added a sync between the chmod and the script startup.
After that, all my microservices started correctly on kubernetes.

Changed in the pom.xml : 
`chmod +x /entrypoint.sh && /entrypoint.sh`
with
`chmod +x /entrypoint.sh && sync && /entrypoint.sh`

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1021,7 +1021,7 @@
                         <entrypoint>
                             <shell>sh</shell>
                             <option>-c</option>
-                            <arg>chmod +x /entrypoint.sh &amp;&amp; /entrypoint.sh</arg>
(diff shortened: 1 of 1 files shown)


jdubois:
This is really good, I love this as it is the kind of contribution that is nearly impossible to figure out without such an awesome community! Thanks @hcouplet !

deepu105:
@hcouplet can you also check the Gradle build?

Issue #8471, opened 2018-10-05
Gradle build doesnt work properly for client side build trigger


deepu105:
##### **Overview of the issue**

Look at the below logs, the front end builds are run only on the first run, it doesn't build it during a clean build or when there are changed files
```
$ ./gradlew bootWar -Pprod jibDockerBuild
Parallel execution is an incubating feature.
Download https://plugins.gradle.org/m2/gradle/plugin/com/google/cloud/tools/jib-gradle-plugin/0.9.11/jib-gradle-plugin-0.9.11.pom
Download https://plugins.gradle.org/m2/org/apache/commons/commons-compress/1.17/commons-compress-1.17.pom
Download https://plugins.gradle.org/m2/org/apache/commons/commons-parent/46/commons-parent-46.pom
Download https://plugins.gradle.org/m2/org/apache/httpcomponents/project/4.0/project-4.0.pom
Download https://plugins.gradle.org/m2/org/apache/commons/commons-compress/1.17/commons-compress-1.17.jar
Download https://plugins.gradle.org/m2/gradle/plugin/com/google/cloud/tools/jib-gradle-plugin/0.9.11/jib-gradle-plugin-0.9.11.jar

> Task :npm_install

> store@0.0.0 postinstall /home/deepu/workspace/temp/xldemo/store
> webdriver-manager update --gecko false

[17:24:21] I/update - chromedriver: file exists /home/deepu/workspace/temp/xldemo/store/node_modules/protractor/node_modules/webdriver-manager/selenium/chromedriver_2.42.zip
[17:24:21] I/update - chromedriver: unzipping chromedriver_2.42.zip
[17:24:21] I/update - chromedriver: setting permissions to 0755 for /home/deepu/workspace/temp/xldemo/store/node_modules/protractor/node_modules/webdriver-manager/selenium/chromedriver_2.42
[17:24:21] I/update - chromedriver: chromedriver_2.42 up to date
[17:24:21] I/update - selenium standalone: file exists /home/deepu/workspace/temp/xldemo/store/node_modules/protractor/node_modules/webdriver-manager/selenium/selenium-server-standalone-3.14.0.jar
[17:24:21] I/update - selenium standalone: selenium-server-standalone-3.14.0.jar up to date
npm WARN bootstrap@4.1.3 requires a peer of jquery@1.9.1 - 3 but none is installed. You must install peer dependencies yourself.
[...]

PR #8472, opened 2018-10-05
Fix gradle prod build issues #8471

deepu105:


atomfrede:
Ah there is also the typo. But it doesn't work or do you want to always execute the build?

deepu105:
It partially works when fixing typo(I mean runs when files are changed, but
doesnt run on clean build, which is starnge as it works in dev). Anyway I
think we should always run this in prod to avoid any kind of glitch in prod
packaging

Thanks & Regards,
Deepu

atomfrede:
👍

PR #8480, opened 2018-10-06
Improve Feign experience

Tcharl:
Use original Feign spring-boot feign configuration to benefit from loggers, encoders, ...
Import that Feign configuration to be used by all Feign Annotations: AuthorizedFeignClients...


generators/server/files.js
```diff
@@ -605,13 +605,8 @@ const serverFiles = {
     serverMicroservice: [
         {
             condition: generator =>
-                !(
-                    generator.applicationType !== 'microservice' &&
```
PierreBesson: I don't remember exactly why and it's such a headache, but there was a reason why this condition was so complicated, so really not sure about this simplification.

generators/server/templates/src/main/java/package/config/PageableSpringEncoder.java.ejs
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Provides support for encoding spring Pageable via composition.
```
PierreBesson: I'm not sure we can include this class from "Pascal Bütteriker" as part of the genrerated code. The license of this file is not clear for me. Ideally, we need to ask this person, if he agrees that his code (I know it's only one file but still...) be included in the JHipster lib, otherwise it will be "UNLICENSED" as part of the generated jhipster code and Mr Bütteriker might not agree to that.

Tcharl:
I tried to apply Morgan's laws on each sequence, may be I missed something: will triple check

generators/server/templates/src/main/java/package/config/PageableSpringEncoder.java.ejs
Tcharl: Sure, will try to find and ask him. The original code came from a PR to spring feign, so I guessed it was aclv2...

deepu105:
@Tcharl can you get the CIs to pass

PierreBesson:
@Tcharl Can you propose a PR to add the pageable class to the JHipster lib and then remove it from this PR. Then we will be able to merge I think.

Tcharl:
Sure, I just forgot to edit the label as WIP.

Expect it soon ;-).

Tcharl:
Should work once https://github.com/jhipster/jhipster/pull/104 will be merged

cbornet:
Just thinking : shouldn't this be fixed in Spring-Boot instead ?

Tcharl:
It has been tried...
https://github.com/spring-cloud/spring-cloud-netflix/issues/556
https://github.com/spring-cloud/spring-cloud-openfeign/issues/26

DanielFran:
@Tcharl Since we upgrade to Spring Cloud Greenwish SR1, will you review this PR to implement pageable?

Tcharl:
@DanielFran sure!

Tcharl:
Everythings' ready on my side :-)

PR #8496, opened 2018-10-08
Fix for Create User dialog pre-filling data from previously viewed/edited user 

sazeez:

Fix for issue #8302 React: "Create a new user" dialog incorrectly prefills properties with data from a previously edited or detailed user

Fixed the following issues:
- Issue of create user dialog pre-filling user data from previously visited user detail or user update pages. 
- Intermittent issue with activation checkbox not being set in user detail dialog, and deactivating user when data is saved.

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
@@ -85,7 +85,6 @@ export default (state: UserManagementState = initialState, action): UserManageme
     case SUCCESS(ACTION_TYPES.FETCH_ROLES):
       return {
         ...state,
-        loading: false,
         authorities: action.payload.data
       };
     case SUCCESS(ACTION_TYPES.FETCH_USERS):
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
@@ -107,7 +107,7 @@ export default (state: UserManagementState = initialState, action): UserManageme
         ...state,
         updating: false,
         updateSuccess: true,
-        user: action.payload.data
+        user: {}
       };
     case SUCCESS(ACTION_TYPES.DELETE_USER):
       return {
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management-detail.tsx.ejs
@@ -25,7 +25,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { APP_DATE_FORMAT } from 'app/config/constants';
 <% if (enableTranslation) { %>import { languages } from 'app/config/translation';<% } %>
-import { getUser } from './user-management.reducer';
+import { getUser, reset } from './user-management.reducer';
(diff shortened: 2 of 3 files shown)


deepu105:
@sazeez, unfortunately, there is already a PR open for this. See #8332

jdubois:
@deepu105 is right, but as the other PR has been stock for 3 weeks, if this one works we should use it

deepu105:
Both PR does the exact same thing so I would like to give the first one the
chance.

Thanks & Regards,
Deepu

sazeez:
Mine also fixes the intermitted issue with the activation checkbox, depending on the timing of the getRoles and getUser calls.

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
```diff
@@ -85,7 +85,6 @@ export default (state: UserManagementState = initialState, action): UserManageme
     case SUCCESS(ACTION_TYPES.FETCH_ROLES):
       return {
         ...state,
-        loading: false,
```
deepu105: If you dont reset this back it meand yiur loading icon will keep appearing if you try to rely on this variable. I don't get how this solves the issue, you might be suppressing the real issue by doing this, please check

deepu105:
@chrisdns since you did first PR what do you think?

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
sazeez: @deepu105, I am not a react expert, but as I understood the render() will kick in as soon as props are updated by the reducer, which means we really don't need the loading boolean, as soon as getUser and getRoles return props, they are supposed to dynamically be updated in the DOM. What I found in my testing is that, the AvInput checkbox refuses to update when DOM is re-rendered. I changed the checkbox to a text input, and it updates to true/false based on the value of the user.activated field on re-rendering. The text input updates fine on re-render when relying on the loading boolean or without it. 

If we were to rely on the loading boolean, then don't we need to wait for both the getRoles and getUser calls to complete and then render the page? But then we need to return two diff boolean props.

chrisdns:
@deepu105 I dont really mind, you can merge this PR when it is done

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.reducer.ts.ejs
```diff
@@ -107,7 +106,7 @@ export default (state: UserManagementState = initialState, action): UserManageme
         ...state,
         updating: false,
         updateSuccess: true,
-        user: action.payload.data
+        user: {}
       };
```
chrisdns: @sazeez maybe you can remove all the return statement here as it is the same just the case afte

deepu105:
The main idea of the loading param is to know that there is some data being
loaded and user can choose to show loading screen or whatever they prefer
based on that. If the AvForms are misbehaving then it's an issue with the
library as we should never depend on dom render sync for anything. So if we
make loading true we should also turn it off when action is complete.
Given that there are 2 PR and more than 2 people are claiming this to be
tricky, I think there some fundamental fault in design of this component (I
should be blamed mostly)
I think I'll take a look and refactor this in that case

deepu105:
closing in favor of https://github.com/jhipster/generator-jhipster/pull/8529
@sazeez see my PR, the issue was with defaultValue, thanks anyway for your PR and sorry we couldn't merge it

sazeez:
> closing in favor of #8529
> @sazeez see my PR, the issue was with defaultValue, thanks anyway for your PR and sorry we couldn't merge it

No worries, glad you took care of it.

PR #8513, opened 2018-10-09
Angular: Unify Principal and Account Service

wmarques:
We have two services who manages the same things IMO: the user account. I think we should merge them into the `AccountService` which will provide the necessary functions.
I also removed the async `hasAnyAuthority`, I don't know it was async as we don't need any network calls.


generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
```diff
@@ -18,19 +18,114 @@
...
+
+    fetch(): Observable<HttpResponse<Account>> {
         return this.http.get<Account>(SERVER_API_URL + '<%- apiUaaPath %>api/account', {observe : 'response'});
     }
 
     save(account: any): Observable<HttpResponse<any>> {
         return this.http.post(SERVER_API_URL + '<%- apiUaaPath %>api/account', account, {observe: 'response'});
     }
+
+    authenticate(identity) {
```
deepu105: could you add tests for this methods?

deepu105:
@jdubois seems like I do not have access to retry jobs on azure pipeline, could you give me access

deepu105:
ye but I can't even restart the entire PR build

Thanks & Regards,
Deepu

generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
wmarques: Sure, I'll write some during the weekend

wmarques:
@deepu105 it's done, I added some unit tests, tell me if that's ok for you :)

wmarques:
@DanielFran good catch

wmarques:
Finally green 🎉 I forgot to take care of your comment @DanielFran I'll do that asap and it should be good

wmarques:
@DanielFran it's done !

PR #8515, opened 2018-10-09
Angular: remove rxjs-compat and bump ngx-infinite-scroll

wmarques:
We should remove `rxjs-compat` as it is provided for migration between rxjs 5.5 and rxjs 6. Now all of our dependencies uses rxjs6 we can safely remove it.
It also allow to do tree-shaking on rxjs operators which save some kB (27 to be precise) 🎉 


jdubois:
Awesome work @wmarques !

PR #8525, opened 2018-10-10
change language after retrieve the account info instead of login

lmbai:
### Current Logic
 the current multiple language. there is 2 place we will change it

1. we will load it during the login process. if login success, we retrieve the account info, and change the laguage based on the user setting, (we keep it in a something like a globle variable)
```
            this.authServerProvider.login(credentials).subscribe(
                data => {
                    this.principal.identity(true).then(account => {
                        // After the login the language will be changed to
                        // the language selected by the user during his registration
                        if (account !== null) {
                            this.languageService.changeLanguage(account.langKey);
                        }
                        this.trackerService.sendActivity();
                        resolve(data);
```
1. when user choose the language temporarily, we change the language 
### Issue

- if the use set the `remember` me to true, we will save the JWT token in to the localStorage, if use close the brower and open again, the login process will be skipped,  we won't able to get the user preferred language, so we will display en as default.
- even user set the `remember` me to false, we still save the JWT token into sessionStorage, if user login succeed, it will display the user preferred language, but if in the same time, we open a new tap and load the gateway again, it will display the en as default (even the user set other langage as preferred).
### New Logic
change the `languageService.changeLanguage` when load the account info. so
- when first time login, it will display the user preferred lang
- when close and open the browser, it will display the user preferred lang
- when refresh the page, it will display the user preferred lang
- for the temporarily language user choosed, it will valid for the current window.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
@@ -17,6 +17,9 @@
  limitations under the License.
 -%>
 import { Injectable } from '@angular/core';
+<%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
+import { JhiLanguageService } from 'ng-jhipster';
+<%_ } _%>
 import { Observable, Subject } from 'rxjs';
 import { AccountService } from './account.service';
 <%_ if (websocket === 'spring-websocket') { _%>
@@ -30,6 +33,9 @@ export class Principal {
     private authenticationState = new Subject<any>();
 
     constructor(
+        <%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
+        private languageService: JhiLanguageService,
+        <%_ } _%>
         private account: AccountService<% if (websocket === 'spring-websocket') { %>,
         private trackerService: <%=jhiPrefixCapitalized%>TrackerService<% } %>
     ) {}
@@ -90,6 +96,13 @@ export class Principal {
                 <%_ if (websocket === 'spring-websocket') { _%>
                 this.trackerService.connect();
                 <%_ } _%>
+
+                <%_ if (enableTranslation) { _%>
+                // After retrieve the account info, the language will be changed to
+                // the user's preferred language configured in the account setting
+                this.languageService.changeLanguage(this.userIdentity.langKey);
+                <%_ } _%>
+
             } else {
                 this.userIdentity = null;
                 this.authenticated = false;
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 import { Injectable } from '@angular/core';
-<%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
+<%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
 import { JhiLanguageService } from 'ng-jhipster';
 <%_ } _%>
 import { Observable, Subject } from 'rxjs';
@@ -33,7 +33,7 @@ export class Principal {
     private authenticationState = new Subject<any>();
 
     constructor(
-        <%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
+        <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
         private languageService: JhiLanguageService,
         <%_ } _%>
         private account: AccountService<% if (websocket === 'spring-websocket') { %>,
@@ -97,7 +97,7 @@ export class Principal {
                 this.trackerService.connect();
                 <%_ } _%>
 
-                <%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
+                <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
                 // After retrieve the account info, the language will be changed to
                 // the user's preferred language configured in the account setting
                 this.languageService.changeLanguage(this.userIdentity.langKey);
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
@@ -19,6 +19,7 @@
 import { Injectable } from '@angular/core';
 <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
 import { JhiLanguageService } from 'ng-jhipster';
+import { SessionStorageService } from 'ngx-webstorage';
 <%_ } _%>
 import { Observable, Subject } from 'rxjs';
 import { AccountService } from './account.service';
@@ -35,6 +36,7 @@ export class Principal {
     constructor(
         <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
         private languageService: JhiLanguageService,
+        private sessionStorage: SessionStorageService,
         <%_ } _%>
         private account: AccountService<% if (websocket === 'spring-websocket') { %>,
         private trackerService: <%=jhiPrefixCapitalized%>TrackerService<% } %>
@@ -100,7 +102,8 @@ export class Principal {
                 <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
                 // After retrieve the account info, the language will be changed to
                 // the user's preferred language configured in the account setting
-                this.languageService.changeLanguage(this.userIdentity.langKey);
+                const langKey = this.sessionStorage.retrieve('local') || this.userIdentity.langKey;
+                this.languageService.changeLanguage(langKey);
                 <%_ } _%>
 
             } else {
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
@@ -97,7 +97,7 @@ export class Principal {
                 this.trackerService.connect();
                 <%_ } _%>
 
-                <%_ if (enableTranslation) { _%>
+                <%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
                 // After retrieve the account info, the language will be changed to
                 // the user's preferred language configured in the account setting
                 this.languageService.changeLanguage(this.userIdentity.langKey);
--- a/generators/client/templates/react/src/main/webapp/app/shared/layout/header/header.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/layout/header/header.tsx.ejs
@@ -19,7 +19,7 @@
 import './header.<%= styleSheetExt %>';
 
 import React from 'react';
-import { Translate<% if (enableI18nRTL) { %>, Storage<% } %> } from 'react-jhipster';
+import { Translate<% if (enableTranslation) { %>, Storage<% } %> } from 'react-jhipster';
 import {
   Navbar,
   Nav,
@@ -67,6 +67,7 @@ export default class Header extends React.Component<IHeaderProps, IHeaderState> 
 
   <%_ if (enableTranslation) { _%>
   handleLocaleChange = event => {
+    Storage.session.set('locale', event.target.value);
     this.props.onLocaleChange(event.target.value);
     <%_ if (enableI18nRTL) { _%>
     document.querySelector('html').setAttribute('dir', isRTL(event.target.value) ? 'rtl' : 'ltr');
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -147,7 +147,7 @@ export const getSession = () => async (dispatch, getState) => {
   
     const account = getState().authentication.account;
     if (account && account.langKey) {
-      await dispatch(setLocale(account.langKey));
+      dispatch(setLocale(account.langKey));
     }
     <%_ } else { _%>
     dispatch({
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -197,7 +197,7 @@ export const login = (username, password, rememberMe = false) => async (dispatch
     await dispatch(setLocale(account.langKey));
   }
   <%_ } else { _%>
-  dispatch(getSession());
+  await dispatch(getSession());
   <%_ } _%>
 };
 <%_ } _%>
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -147,7 +147,7 @@ export const getSession = () => async (dispatch, getState) => {
   
     const account = getState().authentication.account;
     if (account && account.langKey) {
-      dispatch(setLocale(account.langKey));
+      await dispatch(setLocale(account.langKey));
     }
     <%_ } else { _%>
     dispatch({
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -181,7 +181,7 @@ export const login = (username, password, rememberMe = false) => async (dispatch
       Storage.session.set(AUTH_TOKEN_KEY, jwt);
     }
   }
-  dispatch(getSession());
+  await dispatch(getSession());
 };
 <%_ } else if (authenticationType === 'uaa') { _%>
 export const login = (username, password, rememberMe = false) => async (dispatch, getState) => {
@@ -189,16 +189,7 @@ export const login = (username, password, rememberMe = false) => async (dispatch
     type: ACTION_TYPES.LOGIN,
     payload: axios.post('auth/login', { username, password })
   });
- <%_ if (enableTranslation) { _%>
   await dispatch(getSession());
-
-  const account = getState().authentication.account;
-  if (account && account.langKey) {
-    await dispatch(setLocale(account.langKey));
-  }
-  <%_ } else { _%>
-  await dispatch(getSession());
-  <%_ } _%>
 };
 <%_ } _%>
 
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -147,7 +147,9 @@ export const getSession = () => async (dispatch, getState) => {
   
     const account = getState().authentication.account;
     if (account && account.langKey) {
-      await dispatch(setLocale(account.langKey));
+      const langKey = Storage.session.get('locale', account.langKey);
+      Storage.session.set('locale', langKey);
+      await dispatch(setLocale(langKey));
     }
     <%_ } else { _%>
     dispatch({
--- a/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
@@ -181,7 +181,7 @@ export const login = (username, password, rememberMe = false) => async (dispatch
       Storage.session.set(AUTH_TOKEN_KEY, jwt);
     }
   }
-  await dispatch(getSession());
+  dispatch(getSession());
 };
 <%_ } else if (authenticationType === 'uaa') { _%>
 export const login = (username, password, rememberMe = false) => async (dispatch, getState) => {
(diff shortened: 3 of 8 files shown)


lmbai:
Finally pass the test. 
The react part really challenge for me. I never expected that my first time touch react is write Reacut Generator. 
Really impressed by the test cases for the React . It can be the guideline for React development work.
Please help review it and fell free to edit it if anything is not appropriate.

generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
```diff
@@ -17,6 +17,9 @@
  limitations under the License.
 -%>
 import { Injectable } from '@angular/core';
+<%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
```
deepu105: This part is being refactored in another PR by William

deepu105: The second condition can be cleaner, do !== Instead and remove the extra braces around it

generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
```diff
@@ -30,6 +33,9 @@ export class Principal {
     private authenticationState = new Subject<any>();
 
     constructor(
+        <%_ if (enableTranslation && !(authenticationType === 'oauth2')) { _%>
```
deepu105: Same everywhere for the second condition

generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
```diff
@@ -176,33 +181,15 @@ export const login = (username, password, rememberMe = false) => async (dispatch
       Storage.session.set(AUTH_TOKEN_KEY, jwt);
     }
   }
- <%_ if (enableTranslation) { _%>
   await dispatch(getSession());
```
deepu105: I think no need to await here now

generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
```diff
@@ -176,33 +181,15 @@ export const login = (username, password, rememberMe = false) => async (dispatch
...
-  <%_ } _%>
 };
 <%_ } else if (authenticationType === 'uaa') { _%>
 export const login = (username, password, rememberMe = false) => async (dispatch, getState) => {
   const result = await dispatch({
     type: ACTION_TYPES.LOGIN,
     payload: axios.post('auth/login', { username, password })
   });
- <%_ if (enableTranslation) { _%>
   await dispatch(getSession());
```
deepu105: Same no await

lmbai:
@deepu105  
in the `generator-jhipster\generators\client\templates\react\src\main\webapp\app\shared\reducers\authentication.ts.ejs`
```
export const getSession = () => async (dispatch, getState) => {
    await dispatch({
      type: ACTION_TYPES.GET_SESSION,
      payload: axios.get('<%= apiUaaPath %>api/account')
    });
  
    const account = getState().authentication.account;
    if (account && account.langKey) {
      dispatch(setLocale(account.langKey));  // here , !!!!!!!
    }
    <%_ } else { _%>
    dispatch({
      type: ACTION_TYPES.GET_SESSION,
      payload: axios.get('<%= apiUaaPath %>api/account')
    });
    <%_ } _%>
}
```
for the ` dispatch(setLocale(account.langKey));  ` no matter i use await or not. it always fail for the test case in `generator-jhipster\generators\client\templates\react\src\test\javascript\spec\app\shared\reducers\authentication.spec.ts.ejs`
```
    it('dispatches GET_SESSION_PENDING and GET_SESSION_FULFILLED actions', async () => {
      const expectedActions = [
        {
          type: REQUEST(ACTION_TYPES.GET_SESSION)
        },
        {
          type: SUCCESS(ACTION_TYPES.GET_SESSION),
          payload: resolvedObject
        }<% if (enableTranslation) { %>,
          {
            type: localeActionTypes.SET_LOCALE,
            locale: '<%= nativeLanguage %>'
          }<% } %>
      ];
      await store.dispatch(getSession()).then(() => expect(store.getActions()).toEqual(expectedActions));
    });
```
the error displayed in  #13710 failed and  #13706 failed
```
FAIL src/test/javascript/spec/app/shared/reducers/authentication.spec.ts (159 MB heap size)
  ● Authentication reducer tests › Actions › dispatches LOGIN, GET_SESSION and SET_LOCALE success and request actions
    expect(received).toEqual(expected)
    Expected value to equal:
      [{"type": "authentication/LOGIN_PENDING"}, {"payload": {"headers": {"authorization": "auth"}}, "type": "authentication/LOGIN_FULFILLED"}, {"type": "authentication/GET_SESSION_PENDING"}, {"payload": {"value": "whatever"}, "type": "authentication/GET_SESSION_FULFILLED"}, {"locale": "en", "type": "locale/SET_LOCALE"}]
    Received:
      [{"type": "authentication/LOGIN_PENDING"}, {"payload": {"headers": {"authorization": "auth"}}, "type": "authentication/LOGIN_FULFILLED"}, {"type": "authentication/GET_SESSION_PENDING"}, {"payload": {"value": "whatever"}, "type": "authentication/GET_SESSION_FULFILLED"}]
```

**i believe that in the getSessioin we `dispatch(setLocale(account.langKey)); ` , we should expected the `{"locale": "en", "type": "locale/SET_LOCALE"}` , why it always fail?**

in the test case for the login
```
**    <%_ if (authenticationType !== 'oauth2') { _%>
    it('dispatches LOGIN, GET_SESSION and SET_LOCALE success and request actions', async () => {
      <%_ if (authenticationType === 'jwt') { _%>
      const loginResponse = { headers: { authorization: 'auth' } };
      <%_ } else { _%>
      const loginResponse = { value: 'any' };
      <%_ } _%>
      axios.post = sinon.stub().returns(Promise.resolve(loginResponse));
      const expectedActions = [
        {
          type: REQUEST(ACTION_TYPES.LOGIN)
        },
        {
          type: SUCCESS(ACTION_TYPES.LOGIN),
          payload: loginResponse
        },
        {
          type: REQUEST(ACTION_TYPES.GET_SESSION)
        },
        {
          type: SUCCESS(ACTION_TYPES.GET_SESSION),
          payload: resolvedObject
        }<% if (enableTranslation) { %>,
        {
          type: localeActionTypes.SET_LOCALE,
          locale: '<%= nativeLanguage %>'
        }<% } %>
      ];
      await store.dispatch(login('test', 'test')).then(() => expect(store.getActions()).toEqual(expectedActions));**
``` 
**we also expected the `{"locale": "en", "type": "locale/SET_LOCALE"}`, but it does not fail.**

lmbai:
> adfsfsafdsafsadfdsafsaddsa

generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
lmbai: change the logic here, use `await` only when the `tranlation is enabled`

generators/client/templates/react/src/main/webapp/app/shared/reducers/authentication.ts.ejs
lmbai: changed the logic here, use `await` only when the `tranlation is enabled`

lmbai: use await only when the tranlation is enabled which match the test cases too

generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
```diff
@@ -90,6 +96,13 @@ export class Principal {
                 <%_ if (websocket === 'spring-websocket') { _%>
                 this.trackerService.connect();
                 <%_ } _%>
+
+                <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
+                // After retrieve the account info, the language will be changed to
+                // the user's preferred language configured in the account setting
+                this.languageService.changeLanguage(this.userIdentity.langKey);
```
deepu105: One problem I see is that account info might be fetched while not login in and you might already have changed language in UI, so we need to ensure this doesnt switch it back

lmbai: here, i don't get it. we only chang language when we fetche the account info. if user not login, it won't reach here, it will use the the native language.

lmbai:
@deepu105 i understand you consern. here is the story.
we have 3 languages here

1. `native language` which we config in the yo-rc.json
1. `user prefer language` which set in the user setting
1. `temporary language` which use use the selector

### Before Change
when initialize the page  we use the `native language` to render.  

- if user login: we fetch the account info , use the `user prefer language` to render again.
- if user change the user setting, we fetch the account info , use the `user prefer language` to render again.

so if we don't login:
1. refresh the page: **Angular will use back the `native language`. React will keep the current `language`**
1. remember me:
    1. restart the browser and load the page, both Angular and React will use back the `native language`
    1. open a new brower tap, both Angular and React will use back the `native language`

we persist the token in localstorage so that user don't need to login again, but we don't persist the language, so if use open the page with a valid token, he don't need to login. but we use the `native language` 
>here, for React, we actually persist the language in `sessionStorage` which will be valid for current window. that's why for the `refresh page`, react can keep the current language. but the current sessionstorage won;t be accessable for the new tap or restarted browser. 
### After Change
i change the logic a bit to cater the case when user enable the `remember me`.  use the `user prefer language` when we retrieve the account info. 
now the React and Angular have the same behavior. 
1. firt time open the page: use `native language`
1. login : use `user preferred language`
1. remember me
    1. open in a new window: use `user preferred language`
    1. open in a new tab: use `user preferred language`
1. refresh the current page use `user preferred language` 

### `temporary language`
After the above change, the Angular and React have the same behavior now. but it seems like the `temporary language` is not something we want.  
user choose the `temporary language` via the selector, it work for the whole interaction within this page.
it will be invalid when:
1. close and reopen it.
1. open a new tab.
1. refresh the page.

### refresh the page
**the point here is should we keep the `temporary language` when use refresh the page ?**
 @deepu105  i know that you would like to keep it since we are still in the same window from your piont of view.  Here, i am not sure, the age is a Single Page Application, if user refresh the page, every thing will be re renderred, it is also make sense to me.
>anyway, from the the bussness side, it is more common that a use choose a preferred language, so that they would it to be display in that language when they access this page.  i think it is more critical. for the  `temporary language` , it is good to have. but i doubt the frequency user use it.

### how to fix it if we really need it 
in the above i explain a bit the senario we use the  `temporary language`  and discuss a bit about the potential issue when refreshing.  now let's dicuss a bit how to fix if we really need it. 
1. **React**:  we keep the langage in the sessionStorage, the key is `local`
    1. we use `user preferred language` when we retrieve the acount info. as a consequnce, it overwrite the `temporary language` when user fresh the page. 
    1. so if we want to keep the `user preferred language`, we have to avoid to overwrite it.  in current design, is that possible? the answer is no.  the sessionstorage.local is been initialized with `native language` when the page is loading. there is no way for us to determince should we overwrite it or not. the only way here is to introduce a new key called `sessionstorage.selected_lang`, we set the value when user select the language via the selector, so when we retireve the account info, 
        1. if the `sessionstorage.selected_lang` is not null, we won't change the language (the refresh the page)
        1. if the `sessionstorage.selected_lang` is null, we we will use the `user preferred language`  (for the new tab or new window)
1. **Angular**:  we don't persist the language at all, we use the ng-jhipster to handle the language 
    1.  if we want to solve this, we also need to introduce a `sessionstorage.selected_lang`
```
// ng-jhipster\src\language\language.service.js
    JhiLanguageService.prototype.changeLanguage = function (languageKey) {
        this.currentLang = languageKey;
        this.configService.CONFIG_OPTIONS.defaultI18nLang = languageKey;
        this.translateService.use(this.currentLang);
    };
```

@deepu105  what do you think ?  
I personally don't think it is a issue.  while if you think we should cater for `temporary language` even for refreshing , can we do it in a seperate ticket with above thinking ?

deepu105:
@lmbai thanks for the detailed write-up. Its all about user experience. The ideal, what I would prefer is below for both Angular and React(I think React is already close to it)
1. Set native language when loading for the first time. Save this in SessionStorage
2. When logging in set user preference language and update it in SessionStorage (this can be done while loading account as you propose in this PR as its better :clap: )
3. If user selects another language from menu, update the SessionStorage with it
3. when refreshing the current tab, load the value from SessionStorage if it doesn't exist use the user preference else fallback to nativeLanguage

so ideally the logic to set language would be something like this, the SessionStorage value gets updated if user changes language and when loading preference

```
const language = SessionStorage.get("langKey") || account.language || nativeLanguage;
```

WDYT? your PR already improves the situation quite a lot for angular but its important to have similar experience in both Angular and React

lmbai:
@deepu105 , currently Angular and React will  have the same logic in this PR. 

For the refresh case, it is a bit more complicated, the sessionstorage.local have been intiaized before retrieve the account info.  So when retrieve the account info, a bit troublesome to differentiate it is refresh current page or open a new window.  That why I have the idea to add a new session key ‘selectd-lang’ .

While maybe we can try not to set the ‘local’ key for the native language.  But not sure the consequences if we don’t set it. 

For Angular it will be a bit more complicated .

Do u mind that I open another ticket to fix this ‘user selected language ‘ issue?   I may need more time to check for it.

deepu105:
@lmbai as long as we use sessionStorage it should be fine, if you open a new window it will be empty and you can set the user preference, I dont think we have to handle all cases now. And yes ok to have a new PR for that

generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
```diff
@@ -90,6 +98,14 @@ export class Principal {
                 <%_ if (websocket === 'spring-websocket') { _%>
                 this.trackerService.connect();
                 <%_ } _%>
+
+                <%_ if (enableTranslation && authenticationType !== 'oauth2') { _%>
+                // After retrieve the account info, the language will be changed to
+                // the user's preferred language configured in the account setting
+                const langKey = this.sessionStorage.retrieve('local') || this.userIdentity.langKey;
```
deepu105: the key name is very misleading, I think we should change it to something like `language` or `locale`

generators/client/templates/react/src/main/webapp/app/shared/layout/header/header.tsx.ejs
```diff
@@ -67,6 +67,7 @@ export default class Header extends React.Component<IHeaderProps, IHeaderState>
 
   <%_ if (enableTranslation) { _%>
   handleLocaleChange = event => {
+    Storage.session.set('locale', event.target.value);
```
deepu105: lets assign `event.target.value` to a const and use that everywhere in the method

lmbai:
@deepu105 , currently Angular and React will  have the same logic in this PR. 

For the refresh case, it is a bit more complicated, the sessionstorage.local have been intiaized before retrieve the account info.  So when retrieve the account info, a bit troublesome to differentiate it is refresh current page or open a new window.  That why I have the idea to add a new session key ‘selectd-lang’ .

While maybe we can try not to set the ‘local’ key for the native language.  But not sure the consequences if we don’t set it. 

For Angular it will be a bit more complicated .

Do u mind open another ticket to fix this ‘user selected language ‘ issue? Or u

generators/client/templates/react/src/main/webapp/app/shared/reducers/locale.ts.ejs
```diff
@@ -35,7 +35,6 @@ export default (state: LocaleState = initialState, action): LocaleState => {
     case ACTION_TYPES.SET_LOCALE:
       const currentLocale = action.locale;
       if (state.currentLocale !== currentLocale) {
-        Storage.session.set('locale', currentLocale);
```
deepu105: I think its better to do this here instead of in header so that this is responsibility of this reducer and I think it doesnt change any behavior

generators/client/templates/angular/src/main/webapp/app/core/auth/principal.service.ts.ejs
lmbai: yeah, spell error

generators/client/templates/react/src/main/webapp/app/shared/reducers/locale.ts.ejs
lmbai: first time open the page, here will set the `locale` to `native lang`. when it reach the `get account info`, we are not able to use user preferred language anymore since the `locale` is not null. 

In this PR. the React and Angular share the same logic: only persist the `locale` when user select the language.

lmbai:
@deepu105 
now the React and Angular share the same logic. 
only persist the language when user select the language. 
- first time open: use `native language`
- login: use `preferred language`
- refresh: use `preferred language`
- remember me, close the window and open again: use `preferred language`
- user select the language: use `selected language`, and persist it in sessionStorage
    - in current page, it always use  `selected language`
    - refresh the current page, when call get account, since the `locale` is not null, it still use `selected language`
    - open a window, the current `locale` is not accessable, it will use `preferred language`

lmbai:
@deepu105  there is another funny thing, we always talk about "first time open the page", "login" , "refresh", etc.

in this PR, since we will persit it when user select the language. so if user select the before login, it will not use the `preferred language` anymore. 

please advice that is it the righe behavior ?
or we want that: user login will always use the `preferred language`  no matter what they selected before the login ? 

BTW: i am not sure, what first test case fail , some kind of CSS error

deepu105:
@lmbai I think we will just assume that if the user selected a language s/he wanted to view in that language regardless of user preference, so I wouldn't worry much about those. Remember we are not gonna make everyone happy, we just provide opinionated defaults and end users have to adjust according to their preference if required anyway. For me the current solution in the PR is acceptable.

jsm174:
After this merge, I am now getting compile errors.

For jwt gateways: (jdl at bottom)

```
ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/modules/login/login.tsx
ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/modules/login/login.tsx(6,10):
TS2305: Module '"/Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/shared/reducers/authentication"' has no exported member 'login'.

ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/shared/reducers/authentication.ts
ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/shared/reducers/authentication.ts(105,29):
TS2339: Property 'session' does not exist on type '{ new (): Storage; prototype: Storage; }'.

ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/test/javascript/spec/app/shared/reducers/authentication.spec.ts
ERROR in /Users/jsm174/git/jhipster/react-test-9/gateway/src/test/javascript/spec/app/shared/reducers/authentication.spec.ts(8,52):
TS2305: Module '"/Users/jsm174/git/jhipster/react-test-9/gateway/src/main/webapp/app/shared/reducers/authentication"' has no exported member 'login'.
```

For oauth2 gateways:  (jdl at bottom)

```
ERROR in /Users/jsm174/git/jhipster/react-test-10/gateway/src/main/webapp/app/shared/reducers/authentication.ts
ERROR in /Users/jsm174/git/jhipster/react-test-10/gateway/src/main/webapp/app/shared/reducers/authentication.ts(81,29):
TS2339: Property 'session' does not exist on type '{ new (): Storage; prototype: Storage; }'.

ERROR in /Users/jsm174/git/jhipster/react-test-10/gateway/src/test/javascript/spec/app/shared/reducers/authentication.spec.ts
ERROR in /Users/jsm174/git/jhipster/react-test-10/gateway/src/test/javascript/spec/app/shared/reducers/authentication.spec.ts(118,17):
TS2304: Cannot find name 'localeActionTypes'.
```

jwt jdl:

```
application {
  config {
    baseName gateway,
    applicationType gateway,
    authenticationType JWT,
    packageName com.example.gateway,
    prodDatabaseType postgresql,
    clientFramework react,
    buildTool gradle
  }
  entities A, B, C, D
}

application {
  config {
    baseName microservicea,
    applicationType microservice,
    authenticationType JWT,
    packageName com.example.microservicea,
    prodDatabaseType postgresql,
    buildTool gradle
  }
  entities A, B
}

application {
  config {
    baseName microserviceb,
    applicationType microservice,
    authenticationType JWT,
    packageName com.example.microserviceb,
    prodDatabaseType postgresql,
    buildTool gradle
  }
  entities C, D
}

entity A {
    data String
}

entity B {
    data String
}

entity C {
    data String
}

entity D {
    data String
}

dto * with mapstruct
service * with serviceClass

microservice A, B with microservicea
microservice C, D with microserviceb
```

oauth2 jdl:

```
application {
  config {
    baseName gateway,
    applicationType gateway,
    authenticationType oauth2,
    packageName com.example.gateway,
    prodDatabaseType postgresql,
    clientFramework react,
    buildTool gradle
  }
  entities A, B, C, D
}

application {
  config {
    baseName microservicea,
    applicationType microservice,
    authenticationType oauth2,
    packageName com.example.microservicea,
    prodDatabaseType postgresql,
    buildTool gradle
  }
  entities A, B
}

application {
  config {
    baseName microserviceb,
    applicationType microservice,
    authenticationType oauth2,
    packageName com.example.microserviceb,
    prodDatabaseType postgresql,
    buildTool gradle
  }
  entities C, D
}

entity A {
    data String
}

entity B {
    data String
}

entity C {
    data String
}

entity D {
    data String
}

dto * with mapstruct
service * with serviceClass

microservice A, B with microservicea
microservice C, D with microserviceb
```

PR #8530, opened 2018-10-10
Remove extreme logging when running webpack from maven/gradle

deepu105:
fix #8528
-   Please make sure the below checklist is followed for Pull Requests.
-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #8533, opened 2018-10-11
Reduce webpack logs in dev

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/8528
Related to this PR: https://github.com/jhipster/generator-jhipster/pull/8530 -> as discussed @deepu105 , here another proposal. What do you think?

_____

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -183,7 +183,7 @@
     "test:watch": "<%= clientPackageManager %> test <%= optionsForwarder %>--watch --clearCache",
     "webpack:dev": "<%= clientPackageManager %> run webpack-dev-server <%= optionsForwarder %>--config webpack/webpack.dev.js --inline --hot --port=9060 --watch-content-base --env.stats=minimal",
     "webpack:dev-verbose": "<%= clientPackageManager %> run webpack-dev-server <%= optionsForwarder %>--config webpack/webpack.dev.js --inline --hot --port=9060 --watch-content-base --profile --progress --env.stats=normal",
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -183,7 +183,7 @@
     "test:watch": "<%= clientPackageManager %> test <%= optionsForwarder %>--watch --clearCache",
     "webpack:dev": "<%= clientPackageManager %> run webpack-dev-server <%= optionsForwarder %>--config webpack/webpack.dev.js --inline --hot --port=9060 --watch-content-base --env.stats=minimal",
     "webpack:dev-verbose": "<%= clientPackageManager %> run webpack-dev-server <%= optionsForwarder %>--config webpack/webpack.dev.js --inline --hot --port=9060 --watch-content-base --profile --progress --env.stats=normal",
-    "webpack:build:main": "<%= clientPackageManager %> run webpack <%= optionsForwarder %>--config webpack/webpack.dev.js --env.stats=normal",
+    "webpack:build:main": "<%= clientPackageManager %> run webpack <%= optionsForwarder %>--config webpack/webpack.dev.js --env.stats=minimal",
```
deepu105: but this means you won't have the progress when running these outside of maven/gradle as well right? My goal was not to have the progress at all when running from maven gradle and have progress and stats when running via npm

deepu105: ok just tried this and indeed it looks good on local console so I think I like this better, i'll close my PR

PierreBesson: So it seems like we agree there. So let's merge !

PR #8605, opened 2018-10-18
Update spring-boot version to 1.5.17.RELEASE (Spring platform Brussels-SR14)

DanielFran:
Depends on https://github.com/jhipster/jhipster-dependencies/pull/44

See: https://spring.io/blog/2018/10/18/spring-io-platform-brussels-sr14

### DIFF
--- a/generators/server/templates/_gradle.properties
+++ b/generators/server/templates/_gradle.properties
@@ -28,7 +28,7 @@ yarn_version=<%= YARN_VERSION %>
 jhipster_dependencies_version=0.1.14
 # The spring-boot version should match the one managed by
 # https://mvnrepository.com/artifact/io.github.jhipster/jhipster-dependencies/${jhipster_dependencies_version}
-spring_boot_version=1.5.16.RELEASE
+spring_boot_version=1.5.17.RELEASE
 # The hibernate version should match the one managed by
 # https://mvnrepository.com/artifact/io.github.jhipster/jhipster-dependencies/${jhipster_dependencies_version}
 hibernate_version=5.2.12.Final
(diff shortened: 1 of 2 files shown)


jdubois:
I'm sorry about this, but we have just so many bugs and features requests waiting that I just can't spend time updating the v4 branch.
Not sure what the solution would be here, maybe we should elect a maintainer for older versions, and give him some sponsor money for this (as I don't think anybody would do that for free)

DanielFran:
@jdubois It is OK, i just continue updating spring-boot because no decision was until now taken.
For me, if it is oficially declared that we do not make aymore update to v4, you can then close this PR.

deepu105:
What is the point of maintaining older branches, we are not a library, we
are a platform. I would expect people to use the latest available platform
always. There is no point using an old one. The only other valid scenario
is someone having an app created on v4 wants to update these deps. In that
case I think its quite reasonable to ask them to do it manually.

I say we set a hard policy of doing only security patches for older
versions that too for a limited period of 6 months.

DanielFran:
@deepu105 I agree, we just need to defined when we declare the new version is stable to stop updating older version.

deepu105:
I would suggest to add it as policy with a vote on the mailing list. Ill
propose it later when I'm on my laptop

pascalgrimaud:
I agree with @deepu105
And I'm pretty sure very few generated projects the last month use jhipster 4.x

When you want to start a new project with JHipster, you'll use the last version with Spring Boot 2, not an older branch. 

We didn't do this for v2 or v3, so I think we should stop to do this for v4.
We did it the last monthd because of release candidate, now spring-boot 2 is stable enough to stop maintaining our old branch.

It's just my opinion :)

Issue #7750, opened 2018-06-10
Feature Request - Google App Engine Generator


saturnism:
##### **Overview of the feature request**

Add a generator to prepare app for deployment into Google App Engine Standard Java 8.

##### **Motivation for or Use Case**

GAE is a PaaS that can deploy a single module or multiple modules (services) easily. It also has SSL support for both `*.appspot.com` domain and custom domain too. This feature can prepare existing JHipster apps for deployment into GAE.

##### MVP Support Proposal
- [x] Monolith
  - [x] Generate `appengine-web.xml` (Done #7765)
  - [x] Enable session store (Done #7765)
  - [x] Add App Engine plugin to `pom.xml` (Done #7765)
  - [x] Add App Engine plugin for Gradle
  - [x] Generate/update application properties to use Google Cloud SQL (Done #7765)
  - [x] Documentation (Done, https://github.com/jhipster/jhipster.github.io/pull/630)
- [ ] CI/CD
  - [ ] Generate Travis File for GAE
  - [ ] Documentation
- [ ] Micorservices
  - [x] Microservices - all of the monolith features (Done #7765)
  - [ ] Gateway
    - [ ] Generate dispatch mapping to let App Engine route microservices request.
    - [ ] Documentation

Not supported in MVP proposal:
* No registry/zuul support - App Engine supports request mapping already.
* Support only MySQL for now, and possibly PostgreSQL - CloudSQL.

- [x] Checking this box is mandatory (this is just to show you read everything)

saturnism:
@jdubois what's the recommended way to do this? e.g., for each app, we need to add:
* appengine maven plugin 
* WEB-INF/appengine-web.xml
* Update production DB URI

Would it make sense to add this option directly into the `app` generator? Or, is there a way to modify existing files via a new generator, like `jhipster appengine`?

saturnism:
i'll try w/ `jhipster appengine`, that tries to generate new profile/config

ruddell:
I would recommend modeling it off of the Heroku subgen (or one of the other deployment options)
[...]

PR #8616, opened 2018-10-20
App Engine - A series of Generator Fixes

saturnism:
- Re-introduce class loading priority; this is needed for specifically echcache, and adding the mail api just in case.
- Add initial Gradle support
- Store database name in the right variable for default value
- Configure UTF-8 encoding via web.xml for static files

For #7750


saturnism:
Alright I think that's the last set of updates for this PR :)

generators/gae/templates/appengine-web.xml.ejs
```diff
@@ -30,6 +30,13 @@
     <system-properties>
         <property name="java.util.logging.config.file" value="WEB-INF/logging.properties"/>
     </system-properties>
+    <class-loader-config>
+        <!-- temporary workaround to make sure the app uses it's memcache/mail jars.!-->
+        <priority-specifier filename="cache-api-1.0.0.jar"/>
+        <priority-specifier filename="cache-api-1.1.0.jar"/>
+        <priority-specifier filename="javax-mail-1.5.6.jar"/>
+        <priority-specifier filename="javax.mail-1.6.2.jar"/>
```
ludoch: LGTM..
Curious: 
1/ we were not doing that before?
2/ do we really need to have the 2 versions? If folks use the latest jHispter, they should only have the new cache and mail jars, so 1 of them would only be needed?

saturnism: we had these before, but i removed it while testing w/ hazelcast thought we didn't need these anymore. my mistake :)

i don't think we need the other versions; removing them from this list.

ludoch:
LGTM

deepu105:
@saturnism is this ready to be merged?

sendilkumarn:
@deepu105 I am testing it. Will merge it once done 👍

sendilkumarn:
Is it failing because of the flaky protractor?

saturnism:
looks like all green :D Thanks @sendilkumarn !

PR #8645, opened 2018-10-23
gitInitialized needs to be set if repo is already initialized

brunnels:

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -428,17 +428,14 @@ module.exports = class extends BaseGenerator {
                     this.isGitInstalled(code => {
                         if (code === 0) {
                             this.gitExec('rev-parse --is-inside-work-tree', { trace: false }, (err, gitDir) => {
-                                if (!gitDir) {
+                                // gitDir has a line break to remove (at least on windows)
+                                if(gitDir.trim() === 'true') {
+                                    this.gitInitialized = true;
+                                } else {
                                     this.gitExec('init', { trace: false }, () => {
                                         this.log(chalk.green.bold('Git repository initialized.'));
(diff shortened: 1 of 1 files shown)


generators/app/index.js
```diff
@@ -434,6 +434,12 @@ module.exports = class extends BaseGenerator {
                                         this.gitInitialized = true;
                                     });
                                 }
+
+                                // gitInitialized needs to be set if already initialized
+                                // gitDir has a line break to remove (at least on windows)
```
deepu105: why not change the if above to `if (!gitDir.trim())` and do this in else, that looks cleaner IMO

brunnels: Fair enough.  I was just trying to be clear what was happening.

generators/app/index.js
```diff
@@ -428,7 +428,10 @@ module.exports = class extends BaseGenerator {
                     this.isGitInstalled(code => {
                         if (code === 0) {
                             this.gitExec('rev-parse --is-inside-work-tree', { trace: false }, (err, gitDir) => {
-                                if (!gitDir) {
+                                // gitDir has a line break to remove (at least on windows)
+                                if(gitDir && gitDir.trim() === 'true') {
```
deepu105: is the value normally a text `'true'`?

brunnels: When not in a git repo
```
> git rev-parse --is-inside-work-tree
fatal: Not a git repository (or any of the parent directories): .git
```
when in a git repo
```
> git rev-parse --is-inside-work-tree
true
```

deepu105: please fix the prettier error, just run `npm run lint --fix`

PR #8646, opened 2018-10-23
fix(port): makes h2 DB to take a port within the range)

### LINKED ISSUE #8639: h2TCPServer config fails on server port >=10.000
Generated code (of the gateway backend) in DatabaseConfiguration.java was changed in Version 5.5.0 to:

    public Object h2TCPServer() throws SQLException {
        String port = "1" + env.getProperty("server.port");
        log.debug("H2 database is available on port {}", port);
        return H2ConfigurationHelper.createServer(port);
    }

Because there are only 2^16 ports available (on Windows), the program crashes when it tries to open a port >100.000

I use a Server Port >10.000

##### **Reproduce the error**

Use a Server Port >10.000 and run mvnw

##### **Suggest a Fix**

Don't add a leading 1 to the port if it already has 5 digits

##### **JHipster Version(s)**

5.5.0

<anything>
"serverPort": "10000",
<anything>

Windows 10

sendilkumarn:
closes #8639 

But first, have to validate whether this approach is fine. r? @jdubois 

### DIFF
--- a/generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
@@ -21,6 +21,7 @@ package <%=packageName%>.config;
 import io.github.jhipster.config.JHipsterConstants;
 <%_ if (devDatabaseType === 'h2Disk' || devDatabaseType === 'h2Memory') { _%>
 import io.github.jhipster.config.h2.H2ConfigurationHelper;
+import org.springframework.util.SocketUtils;
 <%_ } _%>
 <%_ if (databaseType === 'mongodb') { _%>
 import com.github.mongobee.Mongobee;
@@ -175,23 +176,9 @@ public class DatabaseConfiguration {
     @Bean(initMethod = "start", destroyMethod = "stop")
     @Profile(JHipsterConstants.SPRING_PROFILE_DEVELOPMENT)
     public Object h2TCPServer() throws SQLException {
-        String port = getValidPortForH2();
+        String port = String.valueOf(SocketUtils.findAvailableTcpPort());
         log.debug("H2 database is available on port {}", port);
         return H2ConfigurationHelper.createServer(port);
-    }
-
-    private String getValidPortForH2() throws NumberFormatException {
-        int port = Integer.parseInt(env.getProperty("server.port"));
-        if(port < 10000) {
-            port = 10000 + port;
-        } else {
-            if(port < 65536) {
-                port = port + 1;
-            } else {
-                port = port - 1;
-            }
-        }
-        return String.valueOf(port);
     }
     <%_ } _%>
 <%_ } _%>
--- a/generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
@@ -176,7 +176,7 @@ public class DatabaseConfiguration {
     @Profile(JHipsterConstants.SPRING_PROFILE_DEVELOPMENT)
     public Object h2TCPServer() throws SQLException {
         String port = String.valueOf(SocketUtils.findAvailableTcpPort());
-        log.debug("H2 database is available on port {}", port);
+        log.info("H2 database is available on port {}", port);
         return H2ConfigurationHelper.createServer(port);
     }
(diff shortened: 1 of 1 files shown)


jdubois:
I'm not fan of the "+1" part, as usually people will create several microservices which have ports numbers that follow each other.

I'd rather add a number that goes in a range where we have available ports. Not sure which number to use.

deepu105:
What about adding 1000 or something instead of 1

jdubois:
Yes something like that, but not 1000 as 8080 + 1000 = 9080 and that port is quite commonly used I think

sendilkumarn:
> I'm not fan of the "+1" part, 

@jdubois I can completely understand this concern.  I think we can go for 1000, since the logic kicks in only when the port number is greater than 10000 👍

PierreBesson:
Hey why not use [SocketUtils.html.findAvailableTcpPort](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/util/SocketUtils.html#findAvailableTcpPort--) ?

deepu105:
@PierreBesson in microservices there could be clash when two services start at same time, maybe we could do a combination of `SocketUtils.html.findAvailableTcpPort ` and something like what Sendil did

PierreBesson:
I think when using findTcpPort the risk of two ports being chosen at the same time is pretty low (and only concern the dev environment). Also I suspect the way findPort method works actually try to open random ports and retry until it finds an open one.

deepu105:
@sendilkumarn can you use `SocketUtils` as suggested by @PierreBesson

generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
```diff
@@ -174,7 +175,7 @@ public class DatabaseConfiguration {
     @Bean(initMethod = "start", destroyMethod = "stop")
     @Profile(JHipsterConstants.SPRING_PROFILE_DEVELOPMENT)
     public Object h2TCPServer() throws SQLException {
-        String port = "1" + env.getProperty("server.port");
+        String port = String.valueOf(SocketUtils.findAvailableTcpPort());
         log.debug("H2 database is available on port {}", port);
```
deepu105: maybe we should make this logging from debug to info to print the port

sendilkumarn:
@deepu105 We just proxy the request for the DB

generators/server/templates/src/main/java/package/config/DatabaseConfiguration.java.ejs
sendilkumarn: That makes sense. Well it is useless on anyway since we proxy the request.

sendilkumarn:
Good to go, I guess

jdubois:
No hold on

jdubois:
2 issues with this:

- The port will change all the time, and that's annoying in dev (some people also want to use Liquibase with H2 and wouldn't want to configure their pom.xml all the time, or use Intellij IDEA to look at their database schema, etc)
- The main issue with the random port: imagine you start the app, and take by random the Kafka port. Then you run Kafka... So your app can crash randomly, and that's not good

jdubois:
Could we just add "2000" to the app server port, so we have:

- Spring Boot at 8080 -> H2 at 10080
- Spring Boot at 8081 -> H2 at 10081

So no more issue.

sendilkumarn:
> The port will change all the time, and that's annoying in dev 

I agree that will be annoying. 

> Could we just add "2000" to the app server port

More in line with my previous solution. Let me redo things.

deepu105:
@jdubois this looks ok to me, are you ok to merge this?

pascalgrimaud:
sorry for late comment on this PR.
Pierre suggested to me yesterday, another solution could be to have this logic in property, when generating the project. 
So the final value for H2 port would be in application-dev.yml instead of hard coded.

This way, it will be easy to change in case of conflict. 
What do you think ?

jdubois:
I would keep the current logic (that's why I'm merging this), but indeed an improvement should be to make it configurable like @pascalgrimaud @PierreBesson want.

-> so I'm merging this, but then @pascalgrimaud @PierreBesson if you have time please do a PR with that change (it should be in the JHipster framework project, so it's a tiny bit annoying to do)

pascalgrimaud:
yes, it's better to merge this now and improve later

Issue #8647, opened 2018-10-24
Add an option to configure an entity/dto class name suffix


mselerin:
##### **Overview of the feature request**

Adding an option to configure an entity/dto class name suffix.

##### **Motivation for or Use Case**

Some naming conventions force you to have the entity's class suffixed by 'Entity' and the DTO's class without suffix.

For example, a 'foo' entity will produce : 
* `FooEntity` for the entity class name ;
* `Foo` for the DTO class name ;

Other classes name don't change (`FooService`, `FooRepository`, etc).

The default configuration would be an empty string for the entity suffix and 'DTO' for the dto suffix to reflect the current behaviour.

This configuration would be "application-wide" (no prompt on entity generation) and saved in the `.yo-rc.json`.

I have done almost all the job to see the feasibility of this feature (just took me about 1 hour for the simple case - without relationships). 💪 

I can make a PR as soon as all the work is done (unless a jHipster core member says it's not a good idea).

##### **Related issues or PR**

None.

-   [x] Checking this box is mandatory (this is just to show you read everything)

deepu105:
I'm ok if below conditions are satisfied
1. The code change for this will be minimum (I don't want to review a PR that touches too much)
2. Its passed while app generation as a flag and its persisted and read from the yo-rc.json
3. Support is added in jhipster-core to support JDL
4. should work for import-jdl as well

mselerin:
OK, just to give you a quick status :

- [x] Code change : only inside the "entity-server" templates
- [x] Passed with `--domain-suffix` and `--dto-suffix` flags with app generation
- [ ] Support added in jhipster-core
- [x] Work with import-jdl
- [ ] Unit tests added

I have added two functions inside the entity-server generator : `asDomain(string)` and `asDto(string)` to add the appropriate suffix. These functions are used inside the templates. Thanks to that, it will be super easy to change the behaviour for generating the class & variables names (add a prefix, get the suffix from another config key or something else).

For example, the entity class declaration looks like this now : 
```java
public class <%= asDomain(entityClass) %> implements Serializable { ... }
```

And a field declaration : 
```java
private <%= asDomain(entityClass) %> <%= asDomain(entityInstance) %>;
```

I'll do the PR by the end of the week.

Any feedback is welcome 😉

deepu105:
closing as PR merged. @mselerin next time please add the fix or close keyword on PR description to autoclose ticket

mselerin:
[...]

PR #8649, opened 2018-10-25
Feature: entity and dto suffix (#8647)

mselerin:

PR for #8647


generators/app/index.js
```diff
@@ -105,6 +105,20 @@ module.exports = class extends BaseGenerator {
             defaults: 'jhi'
         });
 
+        // This adds support for a `--domain-suffix` flag
+        this.option('domain-suffix', {
```
gmarziou: "domain suffix" does not really talk to me, if you google these terms you'll find DNS related content and this is what also what it means to me. Based on description I would rather call it "entity suffix". What do others think?

DanielFran: Yes, 'entity suffix' sounds better to me also.

mselerin: I called it 'domain' to match the package where the class stands. But I'm ok with 'Entity Suffix', it makes more sense.

I'll change this tomorrow.

generators/app/index.js
```diff
@@ -145,6 +159,13 @@ module.exports = class extends BaseGenerator {
         this.skipCheckLengthOfIdentifier = this.configOptions.skipCheckLengthOfIdentifier =
             this.options['skip-check-length-of-identifier'] || this.config.get('skipCheckLengthOfIdentifier');
         this.jhiPrefix = this.configOptions.jhiPrefix = _.camelCase(this.config.get('jhiPrefix') || this.options['jhi-prefix']);
+
+        this.entitySuffix = this.configOptions.entitySuffix =
+            this.config.get('entitySuffix') === false ? '' : this.config.get('entitySuffix') || this.options['entity-suffix'];
```
deepu105: You don't need the conditional, jist short circuiting is enough

generators/entity-server/index.js
```diff
@@ -43,6 +43,10 @@ module.exports = class extends BaseGenerator {
         } else {
             useBlueprint = false;
         }
+
+        // Expose 'asEntity' and 'asDto' functions for templating
+        this.asEntity = name => name + this.entitySuffix;
```
deepu105: Please move these methods to generator-base.js

generators/app/index.js
mselerin: I need the `this.config.get('entitySuffix') === false` because when the parameters come from the jdl (`entitySuffix false`), it's a real false value. So we need to test it and convert it to an empty string.

If the value stays to `false`, the entity name would be `Foofalse`.

I could have done the test in the `asEntity` function but the separation of concerns principle tells me to not. ;-)

Unless I misunderstood what you mean...

generators/entity-server/index.js
mselerin: Ok. I wasn't sure if the generator-base just for those functions was the right place but if you tell so, I'm 100% ok with that.

generators/app/index.js
deepu105: Ok understood

generators/entity-server/index.js
deepu105: Yes its the right place

mselerin: Just to let you know...

If I move `asEntity` and `asDto` to generator-base, I still need to define these inside a constructor (which does not yet exists in generator-base). Otherwise, the methods can't be used inside the templates.

I can do something like this : 

```javascript
module.exports = class extends PrivateBase {
  constructor(args, opts) {
        super(args, opts);
        
        // Expose asEntity and asDto to templates
        this['asEntity'] = this.asEntity;
        this['asDto'] = this.asDto;
    }

    /**
     * Construct the entity name by appending the entity suffix.
     * @param {String} name entity name
     */
    asEntity(name) {
        return name + this.entitySuffix;
    }

    /**
     * Construct the entity's dto name by appending the dto suffix.
     * @param {String} name entity name
     */
    asDto(name) {
        return name + this.dtoSuffix;
    }
};
```

But the constructor feel a little hacky... :-(

mselerin: OK I got it !
Everything is fine until a template call include(). For example :
```
<%- include('../../common/save_template', {viaService: viaService, returnDirectly: true}); -%>
```
When the function are defined inside the constructor (see above), I don't need to pass any additional values to the include function.

To fix this, I need to send more parameters to include, like this : 
```
<%- include('../../common/save_template', {asEntity; asEntity, asDto: asDto, viaService: viaService, returnDirectly: true}); -%>
```

Go with that solution ?

deepu105: hmmm, thats weird. I don't see how `this.asEntity = this['asEntity'];` is different from just declaring a class member method. Technically its the same. Anyway i'm ok with passing these to includes if needed

generators/app/index.js
```diff
@@ -145,6 +159,13 @@ module.exports = class extends BaseGenerator {
...
+
+        this.dtoSuffix = this.configOptions.dtoSuffix =
```
deepu105: @mselerin where are this being saved to config?

generators/entity-server/templates/src/main/java/package/service/impl/EntityServiceImpl.java.ejs
```diff
@@ -78,14 +78,14 @@ import static org.elasticsearch.index.query.QueryBuilders.*;
...
- * Service Implementation for managing <%= entityClass %>.
+ * Service Implementation for managing <%= asEntity(entityClass) %>.
  */
 @Service<% if (databaseType === 'sql') { %>
 @Transactional<% } %>
 public class <%= serviceClassName %><% if (service === 'serviceImpl') { %> implements <%= entityClass %>Service<% } %> {
 
     private final Logger log = LoggerFactory.getLogger(<%= serviceClassName %>.class);
-<%- include('../../common/inject_template', {viaService: viaService, constructorName: serviceClassName, queryService: false}); -%>
+<%- include('../../common/inject_template', {asEntity: asEntity, asDto: asDto, viaService: viaService, constructorName: serviceClassName, queryService: false}); -%>
```
deepu105: since we are on node 8+ can you check if the shortHand syntax works here?
```suggestion
<%- include('../../common/inject_template', {asEntity, asDto, viaService, constructorName: serviceClassName, queryService: false}); -%>
```

generators/app/index.js
mselerin: Same files, lines 380-381 in the `saveConfig()` function : 
```javascript
            saveConfig() {
                const config = {
                    jhipsterVersion: packagejs.version,
                    applicationType: this.applicationType,
                    baseName: this.baseName,
                    testFrameworks: this.testFrameworks,
                    jhiPrefix: this.jhiPrefix,
                    entitySuffix: this.entitySuffix,
                    dtoSuffix: this.dtoSuffix,
                    skipCheckLengthOfIdentifier: this.skipCheckLengthOfIdentifier,
                    otherModules: this.otherModules,
                    enableTranslation: this.enableTranslation,
                    clientPackageManager: this.clientPackageManager
                };
```

generators/entity-server/templates/src/main/java/package/service/impl/EntityServiceImpl.java.ejs
mselerin: OK this seems to work well. I have made the change everywhere.

generators/app/index.js
deepu105: Sorry missed it somehow

mselerin:
Documentation done !
https://github.com/jhipster/jhipster.github.io/pull/654

MathieuAA:
In the docs, it says that the suffixes can take two types for values:
  - a string (for an actual, non-empty suffix)
  - false (for no suffix)

Why not just an empty string?

mselerin:
> In the docs, it says that the suffixes can take two types for values:
> 
> * a string (for an actual, non-empty suffix)
> * false (for no suffix)
> 
> Why not just an empty string?

Because it seems not possible to specify an empty string inside the JDL. But if there's a way, I'll take it :-)

MathieuAA:
Okay, there may be a mismatch between what's generated in JCore (in the JSON entity files) and what's generated by the generator. I'll update JCore to only output empty strings.

DanielFran:
@mselerin do not forget to update PR in accordance with the changes from Mathieu.

MathieuAA:
@DanielFran There's nothing to change here (already done in JCore). That's why I asked the question: I was afraid of a potential value/type mismatch between the generated JSON files from JCore and here, but the issue only was in JCore, so nothing to do :)

mselerin:
BTW, the suffixes are not present in the JSON entity files. These parameters are application-wide, therefore only present inside the `.yo-rc.json` file.

And to be complete, an empty value inside `.yo-rc.json` is stored like this : 

```json
"dtoSuffix": ""
```

The `false` value is only needed inside a JDL file.

DanielFran:
@MathieuAA I was refering for the documentário 😎

MathieuAA:
Ahhhhh!Okay :)

deepu105:
@jhipster/developers I'm tight on time can someone test and merge this?

pascalgrimaud:
@deepu105 : we have our JHipster day at Ippon, this tuesday. I think our priority will be to test and merge all pending PR ; then try to reduce the number of opened tickets.

I'll try to help for this

generators/app/index.js
MathieuAA: Not anymore. Deepu's point still stands. The JDL doesn't export the suffixes as false (entity & dto)

generators/app/index.js
```diff
@@ -145,6 +159,13 @@ module.exports = class extends BaseGenerator {
...
+            this.config.get('dtoSuffix') === false ? '' : this.config.get('dtoSuffix') || this.options['dto-suffix'];
```
MathieuAA: Same as my previous comment

generators/entity/index.js
```diff
@@ -223,6 +223,18 @@ module.exports = class extends BaseGenerator {
                     this.log(chalk.green(`\nFound the ${context.filename} configuration file, entity can be automatically generated!\n`));
                     context.useConfigurationFile = true;
                 }
+
+                context.entitySuffix = configuration.get('entitySuffix');
+                if (context.entitySuffix === undefined || context.entitySuffix === null) context.entitySuffix = '';
```
MathieuAA:
```javascript
if (!context.entitySuffix) {
   context.entitySuffix = '';
}
```

generators/entity/index.js
```diff
@@ -223,6 +223,18 @@ module.exports = class extends BaseGenerator {
...
+                if (context.entitySuffix === false) context.entitySuffix = '';
+
+                context.dtoSuffix = configuration.get('dtoSuffix');
+                if (context.dtoSuffix === undefined || context.dtoSuffix === null) context.dtoSuffix = 'DTO';
+                if (context.dtoSuffix === false) context.dtoSuffix = '';
```
MathieuAA: This line can be removed

generators/app/index.js
mselerin: OK, just to be sure... Only the `false` check is not required anymore ?
The fix has been provided inside jhipster-core ?

MathieuAA:
Yep :)

generators/app/index.js
mselerin: Just to let you know...

I can't use this in Javascript : `this.config.get('dtoSuffix') || this.options['dto-suffix']` because if `this.config.get('dtoSuffix')` is an empty string, it's a falsy value, thus the `this.options['dto-suffix']` is applied.

I need to do this : 
```javascript
this.dtoSuffix = this.configOptions.dtoSuffix = _.isNil(this.config.get('dtoSuffix'))
            ? this.options['dto-suffix']
            : this.config.get('dtoSuffix');
```
(`_.isNil` is true if the value is null or undefined)

generators/entity/index.js
mselerin: Same as my previous comment, I need to do this : 
```javascript
if (_.isNil(context.dtoSuffix)) {
   context.dtoSuffix = 'DTO';
}
```

mselerin:
@MathieuAA all changes are done :-)

MathieuAA:
Awesome

deepu105:
shit seems like we merged the docs but not the PR

deepu105:
@jhipster/developers can someone test this please

deepu105:
@mselerin could you enable this in one of the travis test entity?

mselerin:
@deepu105 OK, I'll update the tests ASAP.

BTW, I've merge the master in the PR to keep it up-to-date ;-)

mselerin:
I have detected another error when generating a new app with entity suffix and an entity linked to the `User` entity.

The User entity is generated with the `server` sub-generator and doesn't respect the suffixes. So, when a new entity has a relation with User, it assumes that the class name would be `UserEntity` (if the entity suffix is 'Entity').

@deepu105, what do you think ? Should I modify the server generator as well ?

mselerin:
I have resolved the conflict with master (but messed on my first attempt, so had to rewrite history on my branch, :/)

ruddell:
Tested a few app combinations and the CI passed, so merging.  Congrats and nice work 👍

PR #8654, opened 2018-10-26
Add 'unique = true' to the @Column when needed

Akuka:
Add 'unique = true' to the @Column if the field validate rule is "unique"

Before:
`@Column(name = "just_a_test", nullable = false, unique = true)`

After:
`@Column(name = "just_a_test", nullable = false)`

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -165,7 +165,6 @@ public class <%= entityClass %> implements Serializable {
         if (fieldValidateRules.includes('unique')) {
             unique = true;
         }
-    }_%>
     <%- include ../common/field_validators -%>
     <%_ } _%>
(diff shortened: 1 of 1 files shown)


generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
```diff
@@ -158,9 +159,14 @@ public class <%= entityClass %> implements Serializable {
     const fieldNameUnderscored = fields[idx].fieldNameUnderscored;
     const fieldNameAsDatabaseColumn = fields[idx].fieldNameAsDatabaseColumn;
     if (fieldValidate === true) {
-        if (fieldValidate === true && fieldValidateRules.includes('required')) {
-            required = true;
-        } _%>
+        if (fieldValidate === true) {
```
murdos: Condition already tested at line 161 (yes, the original code was already testing twice).
Please change the start of the block to 
```
if (fieldValidate) {
    required = fieldValidateRules.includes('required')`;
    unique = fieldValidateRules.includes('unique')`;
```

PR #8660, opened 2018-10-26
Use builders to create response entities

Tcharl:
Some response entities were created using new, which is not as good as using builders

-   [x ] Tests are added where necessary
-   [x ] Documentation is added/updated where necessary
-   [x ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
@@ -31,7 +31,7 @@
     <%_ } else { _%>
         Page<<%= instanceType %>> page = <%= entityInstance %>QueryService.findByCriteria(criteria, pageable);
         HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, "/api/<%= entityApiUrl %>");
-        return new ResponseEntity<>(page.getContent(), headers, HttpStatus.OK);
+        return ResponseEntity.status(HttpStatus.OK).headers(headers).body(page.getContent());
     <%_ } _%>
     }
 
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
@Tcharl : FYI, I saw you push to upstream instead of your fork. It will trigger builds twice, 1 for the branch and 1 for PR. Then, all new branchs need to be deleted. 
I think it is better to push to your origin (fork) instead :)

deepu105:
Yes, please don't create PR branches on upstream unless its a long standing
feature that requires collaboration

generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
```diff
@@ -31,7 +31,7 @@
     <%_ } else { _%>
         Page<<%= instanceType %>> page = <%= entityInstance %>QueryService.findByCriteria(criteria, pageable);
         HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(page, "/api/<%= entityApiUrl %>");
-        return new ResponseEntity<>(page.getContent(), headers, HttpStatus.OK);
+        return ResponseEntity.status(HttpStatus.OK).headers(headers).body(page.getContent());
```
cbornet: You can use .ok() directly

Tcharl:
Damn, I'm really sorry for that: I was just crawling on the sources on the Github interface and fixing some of noticed code smells.
Will do that on my branch and PR in the future

PR #8669, opened 2018-10-29
Make the Spring services extendable

Akuka:
Make the Spring services extendable by changing the beans from `private` to `protected`, so the extending service will be able to use the extend service beans

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/inject_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/inject_template.ejs
@@ -39,7 +39,7 @@
     }
 _%>
 
-    <%= beans.map(bean => `private final ${bean.class} ${bean.instance};`).join('\n\n    ') %>
+    <%= beans.map(bean => `protected final ${bean.class} ${bean.instance};`).join('\n\n    ') %>
 
     public <%= constructorName %>(<%= beans.map(bean => `${bean.class} ${bean.instance}`).join(', ') %>) {
         <%= beans.map(bean => `this.${bean.instance} = ${bean.instance};`).join('\n        ') %>


jdubois:
I don't think you should modify a bean that was injected by Spring, this opens the door to an incredible number of problems... Could you give an example of what you are trying to do, and why?

Akuka:
@jdubois I dont want to modify the bean (and even if I wanted, I cant, as its `final`)
I am creating extended service, that extends the JHipster generated service, to keep the my custom logic separated from the JHipster generated code (its make the upgrade more simple)
```
@Service
public class MyEntityNameExtendedService extends MyEntityNameService {
```
In the methods of `MyEntityNameExtendedService` I want to use the `MyEntityNameRepository` bean, which is a member (private currently) of `MyEntityNameService`

Is this scenario sound reasonable?

Tcharl:
@Akuka I don't think that this is really helpful, as you must declare a constructor with the same parameter in you extension calling super. The trick is to also assign these injected services in your service:

```
public MyGeneratedService {
    private final myInjectedBean: BeanOfAnyKind;
    public MyGeneratedService(BeanOfAnyKind myInjectedBean) {
        this.myInjectedBean = myInjectedBean;
    }
}
public MyExtensionService extends  MyGeneratedService{
    private final myInjectedBean: BeanOfAnyKind;
    public MyExtensionService(BeanOfAnyKind myInjectedBean) {
        super(myInjectedBean);
        this.myInjectedBean = myInjectedBean;
    }
}
```

Akuka:
@Tcharl you right that this is not a blocker, and can be done with the way you just presented, but why complex things when we can keep it clean and simple? like this:
```
public MyGeneratedService {
    private final myInjectedBean: BeanOfAnyKind;
    public MyGeneratedService(BeanOfAnyKind myInjectedBean) {
        this.myInjectedBean = myInjectedBean;
    }
}
public MyExtensionService extends  MyGeneratedService{
    public MyExtensionService(BeanOfAnyKind myInjectedBean) {
        super(myInjectedBean);
    }
}
```

In this example its save few lines and in the more complex services its saving more than that, 
Changing it to `protected` would give the developer the flexibility to choose how to handle it, and flexibility is one of the most beautiful things in JHipster

Nymerea:
It's been a long time I wait this functionality.

I hope it will be merged soon :-)

MathieuAA:
Julien is absolutely right.---"don't expose your private parts"-- a former exhibitionist, strangelyIf you need them (the dependencies) to be shared, even between classes/subclasses, then there may be a better design you've not thought of yet. And your design isn't that good if you do share them.Even in this use case. I believe that if we require this to make our services available and usable by using inheritance, we fail.We should treat our generated code just like we treat our external dependencies. We don't use inheritance on them because of the coupling, we use composition.My 2 cents.

Akuka:
I got you,
I thought it would be nice but it's not that critical for me
Feel free to close this PR, I will not take it personally ;)

deepu105:
@Akuka thanks for understanding

jdubois:
Yes thanks, because honestly I spent 15 minutes on this before writing my answer - this wasn't obvious at all to me at the beginning. And that's always good to think and discuss that kind of thing you take for granted.

PR #8672, opened 2018-10-29
use the new favicon

atomfrede:
I added our new favicon (https://github.com/jhipster/jhipster-artwork/blob/master/icons/64x64favicon.ico) to the generator. It looks like this (not sure about the white background in the browser, but according to the guidelines that is correct). 

favicon

[ci-skip]


deepu105:
Lets wait to announce this first and also to add a random main image and so on.

atomfrede:
Will look into the random part today. But yes let's wait with merge until the official announcement

atomfrede:
I have added a very simple hipster choose based on the app name, doing a hash and then using modulo to select the final hipster image.  pretty sure we can do much smarter things, but for now it seems totally sufficient to me.

generators/client/index.js
```diff
@@ -76,6 +76,35 @@ module.exports = class extends BaseGenerator {
         }
     }
 
+    getHipster(baseName = this.baseName) {
```
deepu105: can we move this method to generator-base

generators/client/index.js
```diff
@@ -76,6 +76,35 @@ module.exports = class extends BaseGenerator {
...
+
+        var hash = 0, i, chr;
```
deepu105: use let, plz

generators/client/files-angular.js
```diff
@@ -85,6 +85,14 @@ const files = {
             templates: [
                 { file: 'content/images/hipster.png', method: 'copy' },
                 { file: 'content/images/hipster2x.png', method: 'copy' },
+                { file: 'content/images/hipster-man-1.png', method: 'copy' },
```
deepu105: can we use SVG instead of png? so that we dont have to copy 2 versions, it was done that way in react

generators/client/templates/react/src/main/webapp/app/modules/home/home.css.ejs
```diff
@@ -23,6 +23,6 @@ Main page styles
   display: inline-block;
   width: 100%;
   height: 497px;
-  background: url("../../../static/images/logo-jhipster-react.svg") no-repeat center top;
+  background: url("../../../static/images/<%=hipster%>.png") no-repeat center top;
```
deepu105: if we use svg it will scale and we dont have to copy many files

generators/client/index.js
atomfrede: Sure. Had that first but as it is only relevant for the client generator I moved to to the client generator.

deepu105: problem is yeoman will try to run it as a phase so base is better

atomfrede: :heavy_check_mark: Change it locally already. Going to use svgs too.

atomfrede:
I have changed following:

* `favicon.ico` is now the 64x64 bowtie
* logo is now the bowtie for every app
* hipster on homepage is selected randomly (sure it can be much smarter, but it suits the usecase) based on the app name
* the manifest uses the same hipster, but just the head
* did not add removal of old images for existing apps, not sure if needed, but we can add that later if desired

PR #8673, opened 2018-10-30
Azure: remove maximum execution time

pascalgrimaud:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,6 @@ jobs:
 - job: Test
   pool:
     vmImage: 'Ubuntu 16.04'
-  timeoutInMinutes: 40
   variables:
     JHI_PROFILE: dev
     JHI_RUN_APP: 1


azure-pipelines.yml
```diff
@@ -21,7 +21,6 @@ jobs:
 - job: Test
   pool:
     vmImage: 'Ubuntu 16.04'
-  timeoutInMinutes: 40
```
sendilkumarn: removing this won't possibly put us in an infinite loop. Why do we have to remove this now, do we have some issues?

pascalgrimaud: it's related to the last PR you merged yesterday. I don't know what happened, the job only ran for 8min and stopped because of maximum execution time.
So I prefer to remove it and let the default value (60min if I remember well)

sendilkumarn: You mean with `GAE` PR?

sendilkumarn: It is 60 by default. So no worries in removing it 👍 But I am curious which PR caused this problem.

pascalgrimaud: yes, this strange build: https://dev.azure.com/jhipster/generator-jhipster/_build/results?buildId=853&view=logs

sendilkumarn: This is indeed very strange and I suppose it is due to webpack build is stuck.  (🤞 it should be flaky)

PR #8685, opened 2018-10-31
Support JPA derived identifiers for one-to-one

### LINKED ISSUE #7060: adding MapsId to one-to-one mapping in jdl

With jdl it is possible to generate a one-to-one mapping. The code will look like this:

    public class Client  {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;
    @OneToOne
    @JoinColumn(unique = true)
    private User user;
     ....
    }

in line with the documentation here: http://www.jhipster.tech/tips/022_tip_registering_user_with_additional_information.html
it would be a performance improvement when the class would reuse the id of the user. like this:

    public class Client  {
    @Id
    private Long id;
    @OneToOne
    @JoinColumn(unique = true)
    @MapsId
    private User user;
     ....
    }

This improvement is for performance because there is only one index created and the database will have one less column. it should also be more efficient for hibernate. 

When changing this by hand and generate the corresponding changelog with liquibase take care that there is a bug there. a working changeset is:

		<modifyDataType columnName="id" newDataType="int" tableName="client" />
		<dropPrimaryKey tableName="client" />
		<dropColumn columnName="id" tableName="client" />
		<addPrimaryKey columnNames="user_id" constraintName="clientPK" tableName="client" />
		<addNotNullConstraint columnDataType="bigint" columnName="user_id" tableName="client" />

i think this is a minor improvment. so when it is to complicated don't do it. programmers can copy the above ;)

pmverma:
Fix #7060 

**I would like to have some discussions before implementation.** 

1. There will be support for both with `@MapsId` and without `@MapsId` for one-to-one
	1.1  In case of single one-to-one association, User can make choice to use `@MapsId` or not

	1.2 In case of multiple one-to-one associations, User can make choice to use:

		a. Regular `@OneToOne` without  `@MapsId` . We already have this in jhipster. For eg,

			class User{}
			class Profile{@OneToOne @JoinColumn(unique=true) private User user; } 
			class Preferences{@OneToOne @JoinColumn(unique=true) private User user; }
			class AppUser{@OneToOne @JoinColumn(unique=true) private User user;
							@OneToOne @JoinColumn(unique=true) private Profile profile; 
							@OneToOne @JoinColumn(unique=true) private Preferences pref; }

		b. @MapsId to one association and remaining without @MapsId. For eg,

			class User{}
			class Profile{@OneToOne @JoinColumn(unique=true) private User user; }
			class Preferences{@OneToOne @JoinColumn(unique=true) private User user; }
			class AppUser{@OneToOne @MapsId private User user;
							@OneToOne @JoinColumn(unique=true) private Profile profile; 
							@OneToOne @JoinColumn(unique=true) private Preferences pref; }

		c. @MapsId on all one-to-one associations as long as all related associations have same id(PK). For eg,

			class User{}
			class Profile{ @OneToOne @MapsId private User user; }
			class Preferences{ @OneToOne @MapsId private User user; }
			class AppUser{ @OneToOne @MapsId private User user;
							@OneToOne @MapsId private Profile profile; 
							@OneToOne @MapsId private Preferences pref; }

1. Entity being generated from cli will have option to ask in question - only for one-to-one
	Ask in prompts of relationship section of entity generation question.
	(Do you want to use JPA Derived Identifiers? y/N)

1. Entity being generated from jdl will have option similar to `required` effective to only one-to-one
1. JHipster Online?

@jdubois, what is your thought on this?


jdubois:
Hi @pmverma your PR has an issue: there is no changed files, and 4 very weird commits, could you have a look?

Then to support in JHipster Online, we would need to add the option to the JDL -> @MathieuAA would you be available to do this, could you have look once the PR is corrected?

MathieuAA:
@jdubois will do

pmverma:
@jdubois Those are commits for syncing my old fork with the upstream repo. Better had delete and fork again.

MathieuAA:
Once this is done for the generator, can you @pmverma create an issue in JCore issue tracker for this? thanks :)

generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
```diff
@@ -86,7 +99,8 @@ import static org.elasticsearch.index.query.QueryBuilders.*;<% } %>
  * REST controller for managing <%= entityClass %>.
  */
 @RestController
-@RequestMapping("/api")
+@RequestMapping("/api")<% if (isUsingMapsId === true && !viaService) { %>
+@Transactional <% } %>
```
DanielFran: Not sure we want @Transactional in Resource!

pmverma: Thanks for reviewing.

Two things:
1. This annotation will be added only in case where we use entity directly from Resource Controller.(option 1 during question)
2. Without this annotation, `detached entity passed to persist` will occur in `createXXX` as I have to load the parent entity to get its id.

Blackdread: In all cases, this is bad to have @Transactional anywhere in the controller

pmverma: Yes, you are correct. It is bad. But since we are serving the entity directly from controller, I can not see other option. One stupid thing can be done is annotate only the `createXXX` as only this method is needed `@Transactional`
For the entities managed via service class will not required to do this.

I am open to discuss and receive better suggestions.

Blackdread: I am not sure how behaves jHipster when no service and no DTO are selected in option -> is the entity directly returned in the controller (domain model)?
If not mistaken, DTOs require Service to be true

pmverma: Yes, entity is returned directly in that case.

pmverma:
@jhipster/developers The PR is ready. I would be happy if someone reviews it.

pmverma:
Bug bounty claimed https://opencollective.com/generator-jhipster/expenses/6352

jdubois:
As this is new we should have this question as "false" by default, but at some point I think this should become the new default. See https://vladmihalcea.com/the-best-way-to-map-a-onetoone-relationship-with-jpa-and-hibernate/ so at some point this should also be our recommended behavior.

jdubois:
As the PR was finished during the "bug bounty booster" program, that the bounty was claimed during that time, but that there is just a small bug left, @pmverma let's keep the $200 bounty on this even if this is merged after december, 31st.

pmverma:
@jdubois Thanks for keeping $200 bounty. 👏 

> As this is new we should have this question as "false" by default, but at some point I think this should become the new default. See https://vladmihalcea.com/the-best-way-to-map-a-onetoone-relationship-with-jpa-and-hibernate/ so at some point this should also be our recommended behavior.

Currently I have set "false" as default. Setting "true" as default requires a careful consideration - please see below which is related to your bug.
To follow the recommendations, my main reference is @vladmihalcea blog too. One of the best JPA/Hibernate blog 👍 

>When I create that entity all works well, but when I edit it to change the linked User, the change is not saved (I still have the older User that is linked).

Because we mapped the id of user to id of bar. Once the id(PK) of bar is set, we should not change the id(PK). In fact also not sure if this is allowed by JPA/Hibernate or not.

Thats why I think,

- If you want to update the association (ever), then use normal one-to-one mapping, without MapsId

- If your entity id is never going to change such as UserPreferences<->User, UserExtra<->User then use MapsId. Not limited to `User` entity but any entity applies the same.

I will add these in documentations.

generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
pmverma: Thanks @DanielFran! Good point.

DanielFran:
@pmverma Thanks for this awesome work, I will merge it now!
Can you please create a PR to document when we should use MapsId?
Thanks

generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
```diff
@@ -164,14 +179,19 @@ public class <%= entityClass %>Resource {
      * @param filter the filter of the request<% } %>
      * @return the ResponseEntity with status 200 (OK) and the list of <%= entityInstancePlural %> in body
      */
-    @GetMapping("/<%= entityApiUrl %>")<%- include('../../common/get_all_template', {asEntity, asDto, viaService}); -%>
+    @GetMapping("/<%= entityApiUrl %>")
+    <%_ if (databaseType === 'sql' && isUsingMapsId === true && !viaService) { _%>
+    @Transactional(readOnly = true) <%_ } _%>
```
pascalgrimaud: Sorry for old comment, but after this PR, there are some useless line, because it is not correctly formatted:

```java
/**
     * GET  /blablas : get all the blablas.
     *
     * @return the ResponseEntity with status 200 (OK) and the list of blablas in body
     */
    @GetMapping("/blablas")
    
    public List<Blabla> getAllBlablas() {
        log.debug("REST request to get all Blablas");
        return blablaService.findAll();
    }

    /**
     * GET  /blablas/:id : get the "id" blabla.
     *
     * @param id the id of the blabla to retrieve
     * @return the ResponseEntity with status 200 (OK) and with body the blabla, or with status 404 (Not Found)
     */
    @GetMapping("/blablas/{id}")

    public ResponseEntity<Blabla> getBlabla(@PathVariable Long id) {
        log.debug("REST request to get Blabla : {}", id);
        Optional<Blabla> blabla = blablaService.findOne(id);
        return ResponseUtil.wrapOrNotFound(blabla);
    }
```

pmverma: @pascalgrimaud  Thanks for pointing the formatting problem. I had same problem and fixed it. I will check what went wrong.

pmverma: @pascalgrimaud I have opened a PR at #9060 to fix this.

pascalgrimaud: thx a lot @pmverma !

PR #8691, opened 2018-10-31
removed Dockerfile/entrypoint.sh fromsrc/main/docker

### LINKED ISSUE #8690: Cleanups: Dockerfile and entrypoint.sh still created under 'src/main/docker'

After generating an application, `src/main/docker/Dockerfile` and `src/main/docker/entrypoint.sh` are still created in the source tree. Since `jib` is now used to generate the docker images, both files are no longer required.

Cleanups

##### **Suggest a Fix**

- remove `generators/server/templates/src/main/docker/Dockerfile.ejs`
- remove `generators/server/templates/src/main/docker/entrypoint.sh.ejs`

##### **JHipster Version(s)**

5.4.x, 5.5.0 and current master

rainerh:
updated expected files
added cleanup job

Fixes #8690


lmbai:
pls keep it,  i know the jib is good. But in some case, it is not applicable. we are still using the dockerfile:build even in Jhipster5

Tcharl:
On the opposite, I think that there's too much config files in this folder and pretty sure that many of them are useless/malformed...

jdubois:
I'm also pretty sure those files have issues in them, but I see many people do not want to use Jib yet (and I understand that, I just worked for a client where I couldn't use it due to some limitation).

I agree we should move 100% to Jib at some point, but it's too early and we're going to break people's applications.

Also, please note that we should have a huge work on Jib and Skaffold in the next few months, but that will be for JHispter 7.

I'm closing this as v6 is approaching and this is still too early.

PR #8697, opened 2018-11-01
Upgrade yarn version, fix 'test:watch' script and update angular prompt

vishal423:
- Upgrade yarn to `v1.12.1`
- Fix `test:watch` script in angular. If we pass `--clearCache` option to Jest, then, that would clear the cache, but, wouldn't execute the tests. I also checked equivalent script in react and don't see usage of `--clearCache`. If we have any specific reason to support this, then, we should support it as separate script. I didn't use `&&` to execute `test` script twice (one with `--clearCache` and other with `--watch`) as that would also execute `lint` script twice (again not sure why lint is executed along with `test` script)
- Change prompt message to reflect Angular 7 :smile:  


pascalgrimaud:
I removed the angular version in the prompt, otherwise, we'll need to change to Angular 8, Angular 9 etc
Can you resolve conflict plz ?

vishal423:
yes

vishal423:
looks like rebase didn't went well.. I will close this and create new

Issue #8682, opened 2018-10-30
Upgrade Node.js version to 10.13.0 (LTS)


DanielFran:
##### **Overview of the feature request**

Upgrade Node.js version to 10.13.0 (LTS)

##### **Motivation for or Use Case**
See https://nodesource.com/blog/what-s-new-to-lts-with-the-introduction-of-node-js-10-lts

##### **Related issues or PR**

-   [x] Checking this box is mandatory (this is just to show you read everything)

DanielFran:
@pascalgrimaud Is this something you will take care?

pascalgrimaud:
ok, self assigned

PR #8725, opened 2018-11-02
Upgrade to Node 10.13.0 LTS

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/8682
Another PR: https://github.com/jhipster/jhipster-devbox/pull/98

_____

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - docker
 language: node_js
 node_js:
-  - "8.12.0"
+  - "10.13.0"
 jdk:
   - oraclejdk8
 addons:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ services:
   - docker
 language: node_js
 node_js:
-  - "8.12.0"
+  - "10.13.0"
```
MathieuAA: Why not just use "10" as it will take the latest LTS version?

pascalgrimaud: I think we should keep exact version, like we do in package.json
But we're not consistent, because in Docker image, we download the latest LTS, using `https://deb.nodesource.com/setup_10.x` -> I should change this

PR #8746, opened 2018-11-04
Reduce logs

pascalgrimaud:
During backend tests, there are a lot of warnings logs, which are not necessary.
I apply this change in `src/test/resources/config/application.yml` config file.

_____


PierreBesson:
Yes if you set it to OFF, how can you fix your tests when they are failing ? I would leave the INFO level.

pascalgrimaud:
I see, let me try to convince you :)
Not sure you look the full stacktrace during the tests. I'm pretty sure you only check the name of failed tests, at the end.
Just look the 3 full logs, and tell me which one do you prefer ?

Current logs: https://gist.github.com/pascalgrimaud/6b4c7bd45ddf24f8a4e06fa6ec7b6c73
Logs INFO: https://gist.github.com/pascalgrimaud/caf43e43083dcdb4140ac73b5765198c
Logs OFF: https://gist.github.com/pascalgrimaud/d05ec8c5749b0b8df02ebc8b6b17f95a

deepu105:
What about setting to ERROR so that we get full logs on error and nothing
otherwise?

Thanks & Regards,
Deepu

pascalgrimaud:
Sorry, the 2nd was with ERROR, I will edit my previous post

PierreBesson:
I'm OK to set the level to ERROR when running with maven but definitely not OK to disable most logs for tests like you do in `test/application.yml`.
I know why you want less verbose logs in CI but logs are a powerful tool for testing.

pascalgrimaud:
I can set to ERROR, if we agree

PierreBesson:
To clarify, the reason I'm opposed is that this level is also going to apply when writing new tests.

PierreBesson:
One solution is to put those properties to `application-nologs.yml` and then running the tests with the nologs profile.

pascalgrimaud:
If you prefer keeping these full logs in dev for the users -> https://gist.github.com/pascalgrimaud/6b4c7bd45ddf24f8a4e06fa6ec7b6c73, no worry, I can close this PR.

Then, I will use ENV var to reduce logs in our CI, like I already did:
https://github.com/jhipster/generator-jhipster/blob/master/test-integration/scripts/24-tests-e2e.sh#L80-L88

PierreBesson:
@pascalgrimaud To have less logs when running maven or in CI can be nice to have. Actually Gradle already disable logging to the console when running tests.

deepu105:
Yes lets set to ERROR using env variables

pascalgrimaud:
Let's close this one, so

PR #8752, opened 2018-11-05
Update from Master

DanielFran:


deepu105:
@DanielFran instead of doing merges from master, I'll suggest doing a rebase from master when you want to update. this way history remains clean and you wont get any nasty merge conflicts when its time to merge the branch into master. Rebase is always better than merge

pascalgrimaud:
I agree with @deepu105 
And I think you're the only one to work on `spring-boot_2.1.0`, so I think you can push directly to this upstream branch. Just don't push to master :)

DanielFran:
@deepu105 @pascalgrimaud Sorry guys, I am still learning to use git/github, at work we are still using SVN :(

Next time I will try your suggestion thanks!

PR #8757, opened 2018-11-05
Add global logout for OIDC authentication

mraible:

Add global location for OIDC authentication. This PR covers monoliths, gateways, Angular, and React.


generators/client/templates/angular/src/main/webapp/app/core/login/login.service.ts.ejs
```diff
@@ -83,8 +83,20 @@ export class LoginService {
...
         }
         <%_ } else { _%>
-        this.authServerProvider.logout().subscribe();
-        this.principal.authenticate(null);
+        this.authServerProvider.logout().subscribe(response => {
+            this.principal.authenticate(null);
+            const data = response.body;
+            let logoutUrl = data.logoutUrl;
+            // keycloak
+            if (logoutUrl.indexOf('localhost') > -1) {
```
antarus: This test seems odd to me, Keycloak is not necessarily exposed on localhost, no?

mraible: You are correct, and Okta is not necessarily exposed on oktapreview.com. I'm open to suggestions for better logic.

mraible:
@deepu105 and @sendilkumarn: do you know how I might implement the redirect-after-logout logic for React?

deepu105:
@mraible I need to take a look, you can always use the location object to do that, did you check react router to see if there is a better way?

mraible:
Yeah, once I have the data it shouldn’t be hard. I’m trying to figure out how to get the logoutUrl and idToken from the POST to /api/logout.

mraible:
There’s nothing on the server that tells us if it’s Keycloak or Okta. Spring’s OAuth 2.0 support is agnostic. They’re just URLs to it.

ZiyaErcanErdem:
Enhancing existing AjaxLogoutSuccessHandler and logging out identity on gateway may be another option like below;

```
public class KeycloakLogoutSuccessHandler extends AjaxLogoutSuccessHandler {
	
    private final KeycloakIdentityManager keycloakIdentityManager;
    
    @Value("${security.oauth2.client.access-token-uri:#{null}}")
    private String accessTokenUri;
    
	private String logoutURI;
	
	public KeycloakLogoutSuccessHandler(
			@Value("${security.oauth2.client.access-token-uri:#{null}}") final String accessTokenUri,
			KeycloakIdentityManager keycloakIdentityManager) {
		
		this.keycloakIdentityManager = keycloakIdentityManager;
		
		if (StringUtils.isNotBlank(accessTokenUri)) {
			this.logoutURI = accessTokenUri.replace("token", "logout");
        }
		
	}
	
    @Override
    public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
        super.onLogoutSuccess(request, response, authentication);

        boolean logoutSuccess = false;
        String userName = this.extractAuthenticatedUserName(authentication);
        if(StringUtils.isNotBlank(userName)) {
        	logoutSuccess = this.keycloakIdentityManager.logoutUserByUsername(userName);
        }
        if (!logoutSuccess) {
            response.addHeader("SSO-Logout-URL", logoutURI);
        }
    }
    
    private String extractAuthenticatedUserName(Authentication authentication) {
    	if(null == authentication || !(authentication.getPrincipal() instanceof User)) {
    		return null;
    	}        	
    	User authenticatedUser = (User) authentication.getPrincipal();
    	String userName = authenticatedUser.getUsername();
    	return userName;
    }
}
```

```

@EnableOAuth2Sso
@Configuration
public class OAuth2SsoConfiguration extends WebSecurityConfigurerAdapter {

    
    private final CorsFilter corsFilter;
    private final KeycloakLogoutSuccessHandler keycloakLogoutSuccessHandler;

    public OAuth2SsoConfiguration(CorsFilter corsFilter, KeycloakLogoutSuccessHandler keycloakLogoutSuccessHandler) {
        this.corsFilter = corsFilter;
        this.keycloakLogoutSuccessHandler = keycloakLogoutSuccessHandler;
    }

 ...

   @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .csrf()
                .csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
            .and()
            .addFilterBefore(corsFilter, CsrfFilter.class)
            .headers()
            .frameOptions()
            .disable()
        .and()
            .logout()
            .logoutUrl("/api/logout")
            .logoutSuccessHandler(this.keycloakLogoutSuccessHandler)
        .and()
            .authorizeRequests()
            .antMatchers("/login").permitAll()
            .antMatchers("/api/profile-info").permitAll()
            .antMatchers("/api/**").authenticated()
            .antMatchers("/management/health").permitAll()
            .antMatchers("/management/**").hasAuthority(AuthoritiesConstants.ADMIN)
            .anyRequest().permitAll();
    }

...
}
```

mraible:
I like it! I don’t think it’ll work for Okta though. In my experience, Okta requires you logout using a browser redirect; it can’t be done server-side.

If we wanted to get fancy, we could prompt for Keycloak or Okta when creating an app. However, I like the current setup that allows easily switching between the two.

mraible:
The `ngx-mariadb-oauth2-sass-infinispan` and `ms-ngx-gateway-eureka-oauth2` example apps are failing because of Protractor tests. I've found that adding `browser.sleep(500)` in some of the `administration.spec.ts` tests solves the problem. I'll include this small change in this PR. @jhipster/developers Thoughts?

```diff
diff --git a/src/test/javascript/e2e/admin/administration.spec.ts b/src/test/javascript/e2e/admin/administration.spec.ts
index c39c6ea..ec5761a 100644
--- a/src/test/javascript/e2e/admin/administration.spec.ts
+++ b/src/test/javascript/e2e/admin/administration.spec.ts
@@ -35,6 +35,7 @@ describe('administration', () => {
 
     it('should load configuration', async () => {
         await navBarPage.clickOnAdmin('my-prefix-configuration');
+        await browser.sleep(500);
         const expect1 = 'Configuration';
         const value1 = await element(by.id('configuration-page-heading')).getText();
         expect(value1).to.eq(expect1);
@@ -42,6 +43,7 @@ describe('administration', () => {
 
     it('should load audits', async () => {
         await navBarPage.clickOnAdmin('audits');
+        await browser.sleep(500);
         const expect1 = 'Audits';
         const value1 = await element(by.id('audits-page-heading')).getText();
         expect(value1).to.eq(expect1);
@@ -49,6 +51,7 @@ describe('administration', () => {
 
     it('should load logs', async () => {
         await navBarPage.clickOnAdmin('logs');
+        await browser.sleep(500);
         const expect1 = 'Logs';
         const value1 = await element(by.id('logs-page-heading')).getText();
         expect(value1).to.eq(expect1);
```

mraible:
I'm trying to reproduce the issues with the `ms-ngx-gateway-eureka-oauth2` sample. Running it locally, the app is so slow it's unbearable. It seems this might have to do with the 20MB  `main.bundle.js` file. According to Protractor, it takes over 16 seconds to just log in!

```
  account
    ✓ should fail to login with bad password (1749ms)
    ✓ should login successfully with admin account (16558ms)
```

screen shot 2018-11-12 at 10 03 29 am

This particular example has a whole lot of entities. Shouldn't entities be lazy-loaded so the main bundle isn't so huge?

```
entity-with-dto
entity-with-pagination
entity-with-pagination-and-dto
entity-with-service-class
entity-with-service-class-and-dto
entity-with-service-class-and-pagination
entity-with-service-class-pagination-and-dto
entity-with-service-impl
entity-with-service-impl-and-dto
entity-with-service-impl-and-pagination
entity-with-service-impl-pagination-and-dto
entity.module.ts
field-test-entity
field-test-infinite-scroll-entity
field-test-mapstruct-entity
field-test-pager-entity
field-test-pagination-entity
field-test-service-class-entity
field-test-service-impl-entity
test-root
```

jdubois:
Yes I'm guessing this is because of the entities, because I did some tests recently with an empty project (no entities) and it was really good.

So the solution is indeed to lazy load them - @wmarques is our expert here, and I remember this is very buggy and complicated (don't believe the Angular evangelists, ever!!!)

Akuka:
🙄 
Will it be part of the next release?

mraible:
@Akuka This issue depends on https://github.com/jhipster/generator-jhipster/pull/8925. Without lazy-loaded entities, the Travis tests don't pass. 😕

ndywicki:
Hi guys,

For information,
At my work for production environment the OIDC Server is not Keycloak but another OIDC enterprise server implementation.

The logout (end-session) uri must contain the **idToken** like : 
https://oidc-server/oauth2/connect/endSession?realm=myrealm&id_token_hint={id_token}

So for this case I use generic logout handler with `oAuth2RestTemplate` and the logout URI set in the `application.yml`  :

`````
@Component
public class Oauth2SecurityContextLogoutHandler extends SecurityContextLogoutHandler {

    private final Logger log = LoggerFactory.getLogger(Oauth2SecurityContextLogoutHandler.class);

    private final UserInfoRestTemplateFactory userInfoRestTemplateFactory;

    private final ApplicationProperties applicationProperties;

    public Oauth2SecurityContextLogoutHandler(UserInfoRestTemplateFactory userInfoRestTemplateFactory, ApplicationProperties applicationProperties) {
        this.userInfoRestTemplateFactory = userInfoRestTemplateFactory;
        this.applicationProperties = applicationProperties;
    }

    @Override
    public void logout(HttpServletRequest request, HttpServletResponse response,
                       Authentication authentication) {

        OAuth2RestTemplate oAuth2RestTemplate = userInfoRestTemplateFactory.getUserInfoRestTemplate();
        if (oAuth2RestTemplate.getAccessToken() != null && oAuth2RestTemplate.getAccessToken().getAdditionalInformation() != null) {
            // Retrieved id_token
            String idToken = (String) oAuth2RestTemplate.getAccessToken().getAdditionalInformation().get("id_token");
            // Call end-session endpoint
            oAuth2RestTemplate.getForObject(applicationProperties.getOauth2().getClient().getLogoutUri(), String.class, idToken);
        }
        super.logout(request, response, authentication);
    }
}
`````


Regards,
Nicolas

mraible:
I believe this is ready to merge. All CI tests pass. The two samples that seemed to test this most thoroughly are:

* ngx-mariadb-oauth2-sass-infinispan
* ms-ngx-gateway-eureka-oauth2

They're tested by CI, but I'm not sure their equivalent React flavors are. I'll test them locally and confirm (or make changes until) all e2e tests pass.

PR #8791, opened 2018-11-10
Update .gitlab-ci.yml.ejs

saifali40:
Gitlab using Ruby as default so using the java image in generator

### DIFF
--- a/generators/ci-cd/templates/.gitlab-ci.yml.ejs
+++ b/generators/ci-cd/templates/.gitlab-ci.yml.ejs
@@ -32,6 +32,7 @@ cache:
     paths:
         - .maven/
 <%_ } _%>
+image: java:8
 stages:
     - build
     - test


pascalgrimaud:
Thanks for your 1st contribution to the project. 

But I'm not sure about it. 
When using the sub gen, the user has the choice between performing the build inside a container or not. Here, all GitLab builds will use container, which is wrong.
See the question here: https://github.com/jhipster/generator-jhipster/blob/master/generators/ci-cd/prompts.js#L115

atomfrede:
Thanks @saifali40 but @pascalgrimaud is right. We have a special image with all jhipster tools preinstalled we use, when the user chooses to execute the gitlab build inside docker:

```
image: jhipster/jhipster:v<%= jhipsterVersion %>

saifali40:
Thanks @pascalgrimaud , @atomfrede for the confirmation will check that.

PR #8792, opened 2018-11-10
Unit tests: use Mocha before() instead of beforeEach()

murdos:
Replace all occurrences of beforeEach() that could be replaced by before().
Indeed in our tests, the common use case is to run one or multiple generators with yeoman-test in a preliminary phase, then have multiple assertions with it(). 
Using beforeEach() for the preliminary phase results in multiple unneeded generations.


MathieuAA:
Awesome!

PR #8820, opened 2018-11-14
use primitive booleans for required boolean fields

### LINKED ISSUE #8786: Boolean type when required should be primitive type and getter should start with get when it's not

When attribute type is boolean and is required the java bean intrespector doesn't see the getter that starts with "is" exampe `isMyBoolean`. check [here](https://github.com/bpupadhyaya/openjdk-8/blob/master/jdk/src/share/classes/java/beans/Introspector.java#L513) for source code. this creates problem with validatores when they encouter a `@NotNull Boolean myVar` is should be `boolean` instead to avoid NullPointerExcpetions, it also avoids problems when persising data...
For Non required field a getter should be used for java to see it as linked to the bean `getMyBoolean` same code shows why. 

When using required boolean, the validator crashed with 500 (internal server error) instead of 400 "Bad Request".

##### **Reproduce the error**
generate service with a JDL with required boolean (and DTOs enabled just to make sure) and try to insert with null boolean.

##### **Suggest a Fix**
Suggested above

##### **JHipster Version(s)**
4.14.x but it pretty sure it's still here since I didn't find an issue/commit that fixes it.

```
entity Document {
	employeeUploaderId String
    type String required
    visibleToCustomer Boolean required
    description String required
    filename String required
    size Integer required
    createdAt ZonedDateTime required
    viewedAt ZonedDateTime
}
```

yelhouti:
fixes #8786

-   [X] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -153,7 +153,7 @@ public class <%= entityClass %> implements Serializable {
     const fieldValidate = fields[idx].fieldValidate;
     const fieldValidateRules = fields[idx].fieldValidateRules;
     const fieldValidateRulesMaxlength = fields[idx].fieldValidateRulesMaxlength;
-    const fieldType = fields[idx].fieldType;
+    const fieldType = getJavaFieldType(fields[idx]);
     const fieldTypeBlobContent = fields[idx].fieldTypeBlobContent;
     const fieldName = fields[idx].fieldName;
     const fieldNameUnderscored = fields[idx].fieldNameUnderscored;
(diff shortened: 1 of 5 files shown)


jdubois:
I'm not totally sure, but we need to check we have the correct accessors here. It's a bit more complex than just "is" and "get" (if you look at our code for getters you'll notice it!), and this needs to be perfectly well done or many tools won't work, so there's a small task to check if this is all good.

murdos:
> 
> 
> @murdos I advice you to make your booleans required, it will have an "is" and solves all your problems. Please, for your sanity, avoid tree-state booleans

My boolean fields are already required in JDL (and thus in database) ;-)

yelhouti:
@jdubois as you can check in the link mentioned above (java source code), `is` is used for primitive booleans only. With that, I'm pretty sure all libraries will work.

yelhouti:
Closing since not needed anymore in JHipster 5.x check for more info in the issue

PR #8844, opened 2018-11-16
switch angular entities classes from private to protected fields

### LINKED ISSUE #8839: improve angularX component generation for side-by-side

Like in #8509 the generated files for angular components have unnecessary private methods, which force a user doing side-by-side either reimplement those methods with a leading "_" in the name or change the actual code, what we don't want if doing SBS.

Like @agoncal, I try to work using the side-by-side approach in my projects. Changing the private methods at least to protected, will make extending components looking really good in angular.

 ##### **Suggest a solution**

Methods like onError, onSaveError, but also the injected dependencies should be at least protected, so

- the user can use injected services (or inject own extensions) down to the root components functionality. You simply don't use `private` or `protected` keywords inside constructor (but as parameters for super()).
- extended methods like `subscribeToSaveResponse` can use onError, without reimplementing in

Here is an example of how a side-by-side extended component looks like, if this feature request would be implented:

``` typescript
export class TrackUpdateExtendComponent extends TrackUpdateComponent implements OnInit {

    constructor(
        dataUtils: JhiDataUtils,
        jhiAlertService: JhiAlertService,
        trackService: TrackService,
        songService: SongService,
        activatedRoute: ActivatedRoute
    ) {
        super(dataUtils, jhiAlertService, trackService, songService, activatedRoute);
    }

    ngOnInit() {
        super.ngOnInit();
    }

    protected subscribeToSaveResponse(result: Observable<HttpResponse<ITrack>>) {
      super(result);
      if (something()) {
        this.onError('something went wrong');
      }
    }
}
```

Currently, I couldn't override `subscribeToSaveResponse`, so I would need to make a `_subscribeToSaveResponse`, override `save`, use '_' as prefix in all required services etc.

#8509 

xetys:
fix #8839


deepu105:
There are merge conflicts

jdubois:
Hold on I thought it was decided against this change?

jdubois:
Oh sorry, it was decided on the server side, not on the client side... different issue!
Then it's not consistent in our code.

deepu105:
On the client side its not a big deal since the modifiers are just compile
time and at run time its plain JS anyway

Thanks & Regards,
Deepu

jdubois:
@deepu105 you know this better than me, let's do it then.

Anyway, we need to discuss about improving the update process for our next major release.

xetys:
@deepu105 I need an OK from you to extend this PR to the client files also. It's not a big deal to use protected everywhere...it is actual only a TypeScript thing anyways, as you still can access private fields in JS, but then a user can extend any component without touching the code.

@jdubois We don't need such a change on the backend code, as Java gives a lot of good practices how to achieve the same. I had no issues with private methods or fields while doing side-by-side.

sendilkumarn:
My 2c, We can have it for entities for now. In future once we have decided the approach of upgrade process, we can make this common across library.

May be that time, JS will have these modifiers in it

deepu105:
I agree let's do for entries first

Thanks & Regards,
Deepu

xetys:
Then merge this one, I'll start another PR

xetys:
PS, didn't read whole discussion. Of course I'll wait how we decide the upgrade approach

PR #8862, opened 2018-11-21
Db identifiers length

rubensa:
This pull request contains two fixes:

- Fix constraint name limit length for oracle and mysql
As '_id' is appended limit is maxlength-3 (not maxlength-2)

- [PostgreSQL] Add constraint name length limit
PostgreSQL has 63 byte identifier length limit. (I think that "really"
is 64 -empirical check- but let's trust doc.)
See: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
"The system uses no more than NAMEDATALEN-1 bytes of an identifier;
longer names can be written in commands, but they will be truncated. By
default, NAMEDATALEN is 64 so the maximum identifier length is 63
bytes."

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2374,18 +2374,18 @@ module.exports = class extends PrivateBase {
             constraintName = `${constraintNamePrefix}${this.getTableName(entityName)}_${this.getTableName(columnOrRelationName)}`;
         }
         let limit = 0;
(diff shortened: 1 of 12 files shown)


rubensa:
My miss...  must roll back this:

- Fix constraint name limit length for oracle and mysql
As '_id' is appended limit is maxlength-3 (not maxlength-2)

PR #8863, opened 2018-11-21
[PostgreSQL] Add constraint name length limit

rubensa:
PostgreSQL has 63 byte identifier length limit.
See: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
"The system uses no more than NAMEDATALEN-1 bytes of an identifier;
longer names can be written in commands, but they will be truncated. By
default, NAMEDATALEN is 64 so the maximum identifier length is 63
bytes."

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2347,6 +2347,12 @@ module.exports = class extends PrivateBase {
             );
 
             limit = 64;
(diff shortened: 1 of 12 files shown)


jdubois:
Thanks a lot @rubensa !
There are some file conflicts, I'm going to try to merge them

jdubois:
Everything should be good, thanks again @rubensa !

rubensa:
Thanks to you @jdubois

PR #8871, opened 2018-11-22

henri-tremblay:


henri-tremblay:
The two commits works. But I prefer the latest version that get rids of `BeanClassLoaderAwareJCacheRegionFactory`.

generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
```diff
@@ -25,7 +25,10 @@ import org.ehcache.config.builders.*;
 import org.ehcache.jsr107.Eh107Configuration;
 
     <%_ if (enableHibernateCache) { _%>
-import io.github.jhipster.config.jcache.BeanClassLoaderAwareJCacheRegionFactory;
+import org.hibernate.cache.jcache.ConfigSettings;
+
+import javax.cache.Caching;
+import javax.cache.spi.CachingProvider
```
DanielFran: Missing ;

DanielFran:
```suggestion
import javax.cache.spi.CachingProvider
```
```suggestion
import javax.cache.spi.CachingProvider;
```

henri-tremblay: Thanks. My IDE went crazy at some point.

henri-tremblay:
Is it normalt that the tests are failing?

henri-tremblay:
Please don't merge now. I have a better solution.

henri-tremblay:
All good now. I like it!

henri-tremblay:
This is strange. The build is working perfectly in local but not on Travis. I found the CacheManager as a bean.

jdubois:
@henri-tremblay your tests are failing because in the test application.yml we have `spring.cache.type = simple`.

I think this is to speed up the tests, but for your use case that's a bad idea.

henri-tremblay:
Ah! Weird that it doesn't fail locally though. Let me see.

henri-tremblay:
Hopefully, it should work now.

henri-tremblay:
Awesome!

jdubois:
Thanks a lot @henri-tremblay

jdubois:
@henri-tremblay the `spring.cache.type = simple` trick also makes Infinispan tests fails, and I'm not even sure it's a good idea (sure, it must speed up tests, but then it's not behaving at all like a normal cache).

So I might remove it, and in that case maybe that would also clean up your code - no need for the conditionnal bean anymore (and that would be easier to understand for our end-users)

jdubois:
@henri-tremblay it was mostly to speed up the tests, but here it looks like we would have better quality tests when using the real implementation, so let's use it.

For cache cleanup that should already be OK as it works for "simple".

jdubois:
Hi @henri-tremblay I was some bad news: I'm testing with the current "spring-boot_2.1.0", which as just been upgraded to Spring Boot 2.1.1 (so yes, the branch name is wrong now), and the cache warning issue is back :-(

I'll investigate further, but there are high chances we need to re-open the ticket.

jdubois:
Sorry @henri-tremblay !! My bad, I messed up, this is working fine :-)

PR #8866, opened 2018-11-21
enable original war packaging 

atomfrede:
the original war is now also created when building an executable war via `bootWar` task.

closes #8811
closes #8860

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #8874, opened 2018-11-23
CI: use default build for Gradle and copy the non original war

pascalgrimaud:
Following https://github.com/jhipster/generator-jhipster/pull/8866

I'd prefer to keep `./gradlew bootWar -P"$JHI_PROFILE" -x test` so it is similar to what our end users will use. Are you ok with that @atomfrede ?

_____


atomfrede:
@pascalgrimaud Yes. In case the version is not `SNAPSHOT` it won't work, right?

pascalgrimaud:
yes but our CI will always generate a SNAPSHOT.war

PR #8906, opened 2018-11-30
Force user id to be considered as numeric value when loadData with postgresql and Liquibase

avdev4j:
1. Postgres Exception @ users loadData
In case of postgresql the id is considered as varying value
instead of numeric. The loadData failed. We have to force the type when load data.
It seems to be a regression but don't any information about it.

> Reason: liquibase.exception.DatabaseException: java.sql.BatchUpdateException: Batch entry 0 INSERT INTO public.jhi_user(id, login, password_hash, first_name, last_name, email, image_url, activated, lang_key, created_by, last_modified_by) VALUES('1', 'system', '$2a$10$mE.qmcV0mFU5NcKh73TZx.z4ueI/.bDWbj0T1BYyqP481kGGarKLG', 'System', 'System', 'system@localhost', '', 'TRUE', 'en', 'system', 'system') was aborted: ERROR: column "id" is of type bigint but expression is of type character varying

### DIFF
--- a/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
+++ b/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
@@ -147,6 +147,9 @@
                   identityInsertEnabled="true"
                   <%_ } _%>
                   tableName="<%= jhiTablePrefix %>_user">
+            <%_ if (prodDatabaseType === 'postgresql') { _%>
+            <column name="id" type="numeric"/>
+            <%_ } _%>
             <column name="activated" type="boolean"/>
             <column name="created_date" type="timestamp"/>
         </<% if (devDatabaseType === 'mssql' || prodDatabaseType === 'mssql') { %>ext:<% } %>loadData>


DanielFran:
@avdev4j I restarted the specific test to use last commited updates for spring-boot 2.1.1

avdev4j:
Unfortunately we can reproduce the problem : 

> An attempt was made to call the method org.elasticsearch.common.xcontent.XContentBuilder.string()Ljava/lang/String; but it does not exist. Its class, org.elasticsearch.common.xcontent.XContentBuilder, is available from the following locations:
    jar:file:/home/travis/app/app.war!/WEB-INF/lib/elasticsearch-x-content-6.4.3.jar!/org/elasticsearch/common/xcontent/XContentBuilder.class
It was loaded from the following location:
    jar:file:/home/travis/app/app.war!/WEB-INF/lib/elasticsearch-x-content-6.4.3.jar!/
Action:
Correct the classpath of your application so that it contains a single, compatible version of org.elasticsearch.common.xcontent.XContentBuilder

I wonder if it come from spring data jest ❓

DanielFran:
At least the liquibase issue is fixed!

avdev4j:
And the elasticsearch docker starting too !
If someone know what do mean "-E" parameters in "command" in docker-compose file...

DanielFran:
Maybe it is better to update the title just to confirm the issue is fixed for user type in postgre and open an issue in data-jest

avdev4j:
That's what I did first but because I commit a modification for elastic docker I changed the title.
I will revert the revet but should I keep my commit for the ES docker image ?

DanielFran:
Yes, revert it! I need to review elastic.yml

DanielFran:
I pushed a fix to see if it works....

DanielFran:
Thanks @avdev4j

avdev4j:
You're welcome ;)

Issue #7798, opened 2018-06-16
Angular production build failure with more than one lazy loading module


shai2017:
[dashboard.zip](https://github.com/jhipster/generator-jhipster/files/2108535/dashboard.zip)
##### **Overview of the issue**

Created another module named dashboard (very similar to "admin module" - structure wise). When importing this module at app.module level it works fine and production packaging working as expected. 
But when using lazy loading in app-routing.module (exactly same as admin module). Dev (./mvnw and yarn start) is working fine and I am able to see everything is loaded as expected. 
But production build keeps on failing to generate www package in target folder.
( ./mvnw -Pprod clean package -DskipTests=true) 

##### **Motivation for or Use Case**

It is stopping to package the web application for production which I am deploying to google cloud using CI build

##### **Reproduce the error**
1) Build jhipster microservice gateway without any entity 
2) create module dashboard under app folder (equal level as admin, core etc).  I have attached the folder with code.
3) build process is successful but not generating www folder. please see log message.

##### **Suggest a Fix**

it has something to do with webpack production build when using lazy loading ( used --progress flag in production build but not really helpful). When packaging is successful I am seeing it is emmitting and then displaying webpack assets and chunks. but in this scenario is it failing to do anything and going directly to $ rimraf target//www/app/{src,target/}

**See logs of webpack build for production.**  

<details>
<summary>Webpack output</summary>
<pre>
[INFO] --- frontend-maven-plugin:1.6:yarn (webpack build prod) @ gateway ---
[INFO] yarn not inheriting proxy config from Maven
[INFO] Running 'yarn run webpack:prod' in /Users/capalytix/Documents/DHUB/gateway
[INFO] yarn run v1.6.0
[INFO] $ yarn run cleanup && yarn run webpack:prod:main && yarn run clean-www
[INFO] $ rimraf target/{aot,www}
[INFO] $ yarn run webpack --config webpack/webpack.prod.js --progress --profile
[INFO] $ node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js --config webpack/webpack.prod.js --progress --profile
[ERROR]   0% compiling(node:39630) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead                                                                       21888ms building modules                                                                  
[ERROR] 74ms finish module graph                            
[ERROR] 8ms sealing                                
[ERROR] 0ms basic dependencies optimization 
[ERROR] 78ms dependencies optimization                      
[...]

PR #8925, opened 2018-12-04
angular: Lazy load entities

wmarques:

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -244,7 +244,7 @@ function writeFiles() {
                     this.entityAngularName,
                     this.entityFolderName,
                     this.entityFileName,
-                    this.enableTranslation,
+                    this.entityUrl,
                     this.clientFramework,
                     this.microserviceName
                 );
@@ -262,7 +262,7 @@ function writeFiles() {
                     this.entityAngularName,
                     this.entityFolderName,
                     this.entityFileName,
-                    this.enableTranslation,
+                    this.entityUrl,
                     this.clientFramework
                 );
             }
--- a/generators/client/templates/angular/src/main/webapp/app/entities/entity.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/entities/entity.module.ts.ejs
@@ -20,7 +20,6 @@ import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { RouterModule } from '@angular/router';
 
 @NgModule({
-    // prettier-ignore
     imports: [
         RouterModule.forChild([
             /* jhipster-needle-add-entity-route - JHipster will add entity modules routes here */
--- a/generators/client/templates/angular/src/main/webapp/app/entities/entity.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/entities/entity.module.ts.ejs
@@ -17,13 +17,14 @@
  limitations under the License.
 -%>
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
-
-/* jhipster-needle-add-entity-module-import - JHipster will add entity modules imports here */
+import { RouterModule } from '@angular/router';
 
 @NgModule({
     // prettier-ignore
     imports: [
-        /* jhipster-needle-add-entity-module - JHipster will add entity modules here */
+        RouterModule.forChild([
+            /* jhipster-needle-add-entity-route - JHipster will add entity modules routes here */
+        ])
     ],
     declarations: [],
     entryComponents: [],
--- a/generators/client/templates/angular/tslint.json.ejs
+++ b/generators/client/templates/angular/tslint.json.ejs
@@ -37,10 +37,6 @@
             "spaces"
         ],
         "label-position": true,
-        "max-line-length": [
-            true,
-            180
-        ],
         "member-access": false,
         "member-ordering": [
             true,
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
@@ -75,10 +75,10 @@ export class <%= entityAngularName %>DeletePopupComponent implements OnInit, OnD
                 this.ngbModalRef = this.modalService.open(<%= entityAngularName %>DeleteDialogComponent as Component, { size: 'lg', backdrop: 'static'});
                 this.ngbModalRef.componentInstance.<%= entityInstance %> = <%= entityInstance %>;
                 this.ngbModalRef.result.then((result) => {
-                    this.router.navigate([{ outlets: { popup: null }}], { replaceUrl: true, queryParamsHandling: 'merge' });
+                    this.router.navigate(['/<%= entityUrl %>', { outlets: { popup: null }}]);
                     this.ngbModalRef = null;
                 }, (reason) => {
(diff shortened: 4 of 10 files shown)


wmarques:
Protractor again...
I think the lazy loading part is making the e2e tests to fail because when the user click on an entity in navbar, it loads the entity JS bundle and protractor doesn't seem to "wait" for this loading.
I'll check this tonight, we really need to ship this asap to improve our client-side performances.

wmarques:
The end-to-end tests pass on my computer 😢 Wasn't the same entities...
I tried to restart the failed jobs to see.

wmarques:
@jhipster/developers does anyone knows why [this build](https://travis-ci.org/jhipster/generator-jhipster/jobs/470764347) fails ? I don't see any errors in the tests.

jdubois:
@wmarques I can't believe this didn't have the bug bounty tag!!!!! I just added, so this is included in our booster program for the holidays -> https://www.jhipster.tech/bug-bounties/

chrisdns:
Seems to be a lint error :

`ERROR: /home/travis/app/src/main/webapp/app/entities/entity.module.ts[55, 1]: Exceeds maximum line length of 180`

wmarques:
@chrisdns Yep, but I don't understand why we still have this rule if we use Prettier to format our files... Prettier should be handling this, and not tslint

jdubois:
Thanks @chrisdns -> this is because we generate a lot of entities, so at the end the line is too long. You could exclude it from the linter, but normally this should be automatically solved by Prettier, so there's a glitch somewhere here

wmarques:
@jdubois I think after we add a `needle` prettier isn't run again, so that's why we have those issues.
I'll check if I can easily tell the generator to run again prettier on the files after adding the needle, that should solve the issue.

jdubois:
Thanks @wmarques - that's bad, as Prettier is also configured as a Git pre-commit hook, so when the user will commit his file, it will be modified again.

wmarques:
Finally green ! If someone can review this :)

generators/entity-client/files.js
```diff
@@ -244,7 +244,7 @@ function writeFiles() {
                     this.entityAngularName,
                     this.entityFolderName,
                     this.entityFileName,
-                    this.enableTranslation,
```
mraible: Does removing this cause any issues?

wmarques: Was unused in the needle but I can double check it to be sure !

wmarques: Doesn't seem used, tested on i18n applications with entities so should be fine :)

wmarques:
@jhipster/developers Guys i'm having a last issue with the needle. I need to add a comma before the needle only if an entity has already been generated. Otherwise it will throw an error.
Do you know how can I do that ?
I can't add a trailing comma at the end to prevent this because prettier will remote it...

pascalgrimaud:
@wmarques : maybe you can look this code (if it can help you):
- https://github.com/jhipster/jhipster-vuejs/blob/master/generators/client/utils.js#L37-L39
- here the diff:

```diff
state: {
            logon: false,
            userIdentity: null,
            authenticated: false,
            currentLanguage: 'en',
            languages: {
                en: { name: 'English' },
                fr: { name: 'Français' },
                it: { name: 'Italiano' }
                // jhipster-needle-i18n-language-key-pipe - JHipster will add/remove languages in this object
            }
        },
```

wmarques:
@pascalgrimaud thanks ! But this seems to not add a comma on the last if I'm correct. So if you run this needle again, like when you want to add a new language, I think this code will fails too, as we don't add a comma at the beginning of the string. So in the code it will result in 
````
en: { name: 'English' },
                fr: { name: 'Français' },
                it: { name: 'Italiano' }
                es: { name: 'Spanish' }
````

And this will throw a compilation errror.

**Comment by pascalgrimaud** [comment]
You're right. The code I showed you is not correct. That means we have a bug in VueJS, but it's another subject.

Another sample, directly in the generator-jhipster:
https://github.com/jhipster/generator-jhipster/blob/17efde1a21618b0a04bc4fb28d77eb8d8be7cd83/generators/generator-base-private.js#L283-L290

In fact, it will recalculate all the content, then, do a replace.
Is it possible here to do the same?

**Comment by wmarques** [comment]
@pascalgrimaud I guess so, will try, thanks ! :)

**Comment by wmarques** [comment]
Ok I fixed this by simply checking the number of existing entities in the app. If we have more than one, I will add a comma at the beginning of the string.

**Comment by deepu105** [comment]
@wmarques cant you move the needle before an existing entry so that a comma is always needed? also prettier should remove unwanted trailing comma if you are not using prettier-ignore

**Comment by wmarques** [comment]
@deepu105 There are no existing entry on the first time we generate an entity :( 
But maybe you meant keeping the needle always on the top of the array?

**Comment by wmarques** [comment]
@deepu105 the current solution isn't good enough for you ?
I'm doing this to check if we already have an entity in the file: 
`const isEntityAlreadyGenerated = jhipsterUtils.checkStringInFile(entityModulePath, 'loadChildren', this);`

**Comment by deepu105** [comment]
Its ok not a big deal, I don't like unwanted complexity in general but its fine go ahead

**Comment by mraible** [comment]
All tests are passing. Is it OK to merge this one?

**Comment by wmarques** [comment]
@mraible For me it's ok, let's merge this, I'll keep an eye on issues if we find some bugs after merging this (let's hope not 🙏 )

**Comment by tibistibi** [comment]
it took me a bit to figure it out so here what i did to lazy load my own made entities:

First make sure this module is loaded last (in app.module.ts)

     StandBystandAppRoutingModule 


Than add this to a ngModule:

    imports: [
        RouterModule.forChild([
            {
                path: 'employee/declaration',
                loadChildren: './declaration/declaration.module#SbysDeclarationModule'
            }
        ])
    ],

Make sure all services are added to the provider part of your module.ts 
And for translation add the translation to provider and constructor something like this:
```
....
    providers: [
        { provide: JhiLanguageService, useClass: JhiLanguageService },
        DeclarationService,
        DeclarationPopupService,
        ClientInvoiceService,
        FreelancerInvoiceService
    ],

})
export class SbysDeclarationModule {
    constructor(private languageService: JhiLanguageService, private languageHelper: JhiLanguageHelper) {
        this.languageHelper.language.subscribe((languageKey: string) => {
            if (languageKey !== undefined) {
                this.languageService.changeLanguage(languageKey);
            }
        });
    }
}


```

tibistibi:
@wmarques i have all working local but on production and staging it is not working. it seems the modules are not there. i get the error:

    ERROR Error: "Uncaught (in promise): Error: Cannot find 'StandBystandNotificationModule' in './notification/notification.module'

some modules are loading fine others are not. i do not see any differences between them and they all are generated from jhipster. maybe it has something to do with this: #7798 issue where you did great work.

cr1st1p:
Not sure where to ask - related to this change: I do not understand why an entity handling module (even being it lazy loaded) has to handle the event of changing the language? Unless I'm missing something, why is the entity module doing the event passing between languageHelper and languageService? Every single one of them! Is this really its concern/task? Why is this not done by the navbar component? Or by the app module.

Also, I'm trying to reuse code generated by jhipster in other parts of the app - like the delete record popup + codebase of the entities listing. But this lazy loading feature broke it. Maybe you should add a warning + document how to reverse it.
Thank you.

tibistibi:
@cr1st1p true it was a quite difficult to find what to do. i think i have fixed all so if you have a question please ask because the changes are i have just done it.

tibistibi:
in any case popups are still possible. but the link is somewhat different. 
my links looks like this now:

      [routerLink]="[ { outlets: { popup: 'freelancer/'+ position.id} }]"

the base is now in the lazy loading and the last part of the url is in the router so that should also be done in the popup links.

PR #8934, opened 2018-12-08
Handle blueprint in upgrade subgenerator

murdos:
This branch is based on my other PR #8932

### DIFF
--- a/generators/upgrade/index.js
+++ b/generators/upgrade/index.js
@@ -48,6 +48,11 @@ module.exports = class extends BaseGenerator {
             desc: 'Upgrade to a specific version instead of the latest',
             type: String
         });
+        // This adds support for a `--target-blueprint-version` flag
+        this.option('target-blueprint-version', {
+            desc: 'Upgrade to a specific blueprint version instead of the latest',
+            type: String
+        });
 
         // This adds support for a `--skip-install` flag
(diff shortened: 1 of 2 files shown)


generators/upgrade/index.js
```diff
@@ -180,10 +190,48 @@ module.exports = class extends BaseGenerator {
                 });
             },
 
-            checkLatestVersion() {
-                if (this.targetVersion) {
-                    this.log(`Upgrading to the target version: ${this.targetVersion}`);
-                    this.latestVersion = this.targetVersion;
+            checkLatestBlueprintVersion() {
```
deepu105: I was wondering instead of doing specific methods for generator and blueprint cant we generalize the methods to check and upgrade an npm lib and then pass the variables to that instead of duplicating code

deepu105: So here this method can just call another method which does the check twice, first for package `generator-jhipster` and then for package `generator-${blueprint}`

murdos: I've done a bit of refactoring, please let me know what you think.

murdos:
@deepu105 Thanks for the review.
FYI I plan to do some manual tests on a real project. And if everything is OK, I'll merge this PR.

PR #8965, opened 2018-12-18
Remove finals that are now implicit

henri-tremblay:


murdos:
Hi. Could you just explain why these finals are now implicit?
I don't get it.

Blackdread:
No reason to remove.
I did not check all the code but it is not used in a lambdas so it is not implicit, anyone could eventually change the value later without the final

henri-tremblay:
I misread then. My real reason to remove it is that adding final to a local variable is not really useful and adds unnecessary noise to the code. Currently, JHipster seems to adopt a coding style without these unnecessary finals. They are present only in these two places. I recommend to remove them.

I will change the commit comment though.

jdubois:
Yes, our usual code style is that we remove that kind of "noise".

Issue #8954, opened 2018-12-14
Follow Dave Syer's advices to make Spring faster


jdubois:
This should be done on the `spring-boot_2.1.0` branch.

We should do a list of all the tips given by Dave Syer on https://spring.io/blog/2018/12/12/how-fast-is-spring and discuss them one-by-one, and then do small PRs to implement each of the ones we select.

jdubois:
I did the 4 most obvious tricks we could do. For the other points it's more complicated:

- several of them include adding arguments to the JVM, and we don't have much control on how the end-user sets up is IDE. Maybe we could do this only for Maven/Gradle.
- we could manually configure all beans, but then it would be a pain for our end-users when they want to add or configure more features, as they would lose all the auto-configuration and will need to do this manually like us.

PierreBesson:
@jdubois arguments to the JVM could be setup on the docker image as well. As for the exploded jar and running with `java -cp`, this is already what jib does 😃!

jdubois:
Thanks @PierreBesson good idea, I'm doing this!

jdubois:
BTW I won't use `-XX:TieredStopAtLevel=1` because this degrades performance later

jdubois:
Closing this as the list is over!

jdubois:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/6356

PR #8973, opened 2018-12-21
Add the spring-context-indexer annotation processor

jdubois:
Add the spring-context-indexer annotation processor.

This doesn't have a lot of impact, unless you have lots of components.

See #8954


srinivasa-vasu:
I don't think, this works with swagger integration. There is an open [swagger issue](https://github.com/springfox/springfox/issues/2451) which prevents from using this spring feature

jdubois:
Thanks a lot @srinivasa-vasu !! As explained in the blog post, this doesn't help a lot, so let's remove it. Also, we should not trade runtime startup time for development experience (annotation processors=bigger compilation time and tons of configuration issues), or we will lose the project's focus.

PR #8974, opened 2018-12-21
Remove JMX to improve start-up performance

jdubois:
Remove JMX setup by default.

If this works well, this should be set to false by default: https://github.com/jhipster/jhipster/blob/2.0.28/jhipster-framework/src/main/java/io/github/jhipster/config/JHipsterDefaults.java#L156

See #8954

### DIFF
--- a/generators/server/templates/src/main/java/package/Application.java.ejs
+++ b/generators/server/templates/src/main/java/package/Application.java.ejs
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.jmx.JmxAutoConfiguration;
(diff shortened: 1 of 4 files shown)


generators/server/templates/src/main/java/package/Application.java.ejs
```diff
@@ -58,7 +59,7 @@ import java.util.Collection;
     excludeFilters = @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE, classes = OAuth2InterceptedFeignConfiguration.class)
 )
 <%_ } _%>
-@SpringBootApplication
+@SpringBootApplication(exclude = JmxAutoConfiguration.class)
```
PierreBesson: Are you sure this needs to be excluded ? some people might want to activate it later.

jdubois: Yes I'll revert this: I'm just following the blog post here, I'm not finding any great improvements by removing JMX. Then, I see it also as a security risk, so let's remove it be default.

(I can't write here what I've done with JMX in the past, but trust me there were several very high security issues because JMX was left opened)

PR #8976, opened 2018-12-21
Lazy init Spring Beans at start-up.

jdubois:
This is gives a huge performance boost, but is considered experimental, so this is only configured in dev mode for the moment.

See #8954


sendilkumarn:
This is interesting stuff. 👍

jdubois:
This is still an experiment, but if it works fine this should go into the JHipster framework

PR #9001, opened 2019-01-01
add configuration property for main language used in server generator

### LINKED ISSUE #9000: RFE: Add configuration property for main language used in server generator
Happy new year to everyone!!!  :tada: :tada: :tada:

A way is needed to override the default settings of the `maven-compiler-plugin` in order to properly configure maven for use in Kotlin.

Need to configure Maven for use with Kotlin for https://github.com/jhipster/generator-jhipster.

More specifically we need to:
1. Remove the default `maven-compiler-plugin` configuration  and re-add it modified **after** the `kotlin-maven-plugin` configuration as explained in this comment https://github.com/jhipster/jhipster-kotlin/issues/45#issuecomment-450544754
2. Avoid having to create a duplicate `pom.xml` as it will create too high of a maintenance cost.

Since we do not want to maintain a secondary `pom.xml` as it would be too much work without any benefit and I could not find a better way to customize the `pom.xml` (I think that needles are mainly used for adding and not for removing), I believe the best way would be to add another configuration property (e.g. `serverLanguage='java'`) to the base server generator that would allow us to customize this behavior.

I will add a PR to show what I mean. If anyone knows of a better way to achieve this then please let us know.

None that I know of.

pvliss:
Fix #9000

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -133,6 +133,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 if (!this.applicationType) {
                     this.applicationType = 'monolith';
                 }
+                this.serverLanguage = 'java';
                 this.reactive = configuration.get('reactive') || this.configOptions.reactive;
                 this.reactiveRepository = this.reactive ? 'reactive/' : '';
                 this.packageName = configuration.get('packageName');
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -777,6 +777,7 @@
         </testResources>
         <%_ } _%>
         <plugins>
+<%_ if(serverLanguage === undefined || serverLanguage === 'java') { _%>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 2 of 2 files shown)


murdos:
I understand the need, but I don't think this is the better solution: you're introducing a concept, `serverLanguage`, that has no use in jhipster generator.

What I've done in another (non public) blueprint, is to manipulate the xml dom of pom.xml, in order to do the change required for the blueprint. Here the kotlin blueprint could just retrieve the maven-compiler-plugin subtree in pom.xml and delete it.
If need I could help do that in the kotline blueprint.

pvliss:
Hello @murdos,

Thanks for your input. Actually that is my current approach using xmldom. You can see it at https://github.com/jhipster/jhipster-kotlin/pull/74/commits/a9dd7db6da9ec544e41cb2698c1238f2cac0ef45#diff-d702dc7debe242a567b1c79be70e1621.

It did not always work and I could not figure out why. Maybe I will need to debug a bit more. You are welcome to help anytime

pvliss:
@murdos So I had another look at my code and found the problem. I was using the actual file system instead of the virtual file system provided by yeoman as mentioned in the [docs](https://yeoman.io/authoring/file-system.html). Thanks for putting me back to the right track again. :+1:

PR #9007, opened 2019-01-03
on user update, also update authorities

rabiori:
I don't actually know how to create a PR well, but I've noticed that `updateUser` does not update the user's authorities and this change fixes that.


Blackdread:
This endpoint is used by a "normal" user so I do not think this should be allowed, any user could add to his roles ADMIN...

deepu105:
@Blackdread is correct. Updating of a user role can only be done by an admin

deepu105:
@rabiori please create a ticket and state what issue you are trying to solve, else state clearly in the PR description what you are trying to do

PR #9019, opened 2019-01-07
add management endpoints security using a fixed API key, fix #8946

### LINKED ISSUE #8946: [FEATURE] Allow access to secured management endpoints using basic auth

Allow requests be made to `/management/XXX` using a fixed API key that would be defined in the properties (for example `jhipster.security.authentication.management-api-key`).
This api key could be passed in the header and let users access the management API without a regular user authentication. This would also lead to the creation of a new Spring "Authority" which would be call `ROLE_MANAGEMENT`.

@clement26695 is almost done with the Micrometer migration and we have decided to support prometheus by default. So in addition to the metrics screen, the app will expose prometheus metrics on `/management/prometheus`. However this poses a problem as this endpoint is not secured.

So I have looked at the [Prometheus security docs](https://prometheus.io/docs/operating/security/) on how they recommend to do this and the answers have been very lacking. It appalling that the Prometheus devs don't care too much about security and even point to a [security audit](https://prometheus.io/assets/downloads/2018-06-11--cure53_security_audit.pdf) which includes this section.

---
screenshot 2018-12-12 at 17 21 36

---
After reading the docs further, I found that it is possible to configure the Prometheus server to provide an API key to be passed in the `Authorization` header when querying the app prometheus endpoints. 

Implementing this fixed API key feature will also help for many other use cases for example to access actuator endpoints from scripts, to provide management rights for non admin users (now that we will have a management authority which can be attributed users other than admins). It can also be helpful for the registry to restrict its access as it currently has admin rights to all apps.

##### *Proposed implementation*

Add a new Spring Security filter and securityConfigurerAdapter to achieve fixed API key security for /management endpoints. The feature would only be enabled when the new `management-api-key` property is defined.

None that I know of

PierreBesson:


Tcharl:
No update on the Keycloak template?

PR #9021, opened 2019-01-08
fix comment of app.css.ejs

xkenmon:
css don't have comment like '//'
[Comment in css](https://developer.mozilla.org/en-US/docs/Web/CSS/Comments)


deepu105:
Thank you

PR #9022, opened 2019-01-08
rewrite Entity equals and hashCode to fix #8656

### LINKED ISSUE #8656: Accept subclasses as parameter to equals() method in entities

I ran into issues with the current `equals()` implementation returning false with a Hibernate proxy.

As IntelliJ says, it's better to accept subclasses in this case.

Avoid issues with Hibernate proxies of entities.

#5604

pblanchardie:
accept subclasses as parameter to equals to avoid problems with proxies, and provide consistent hashCode with database generated identifiers

fix #8656

this is the most reliable equals and hashCode implementation when using database generated identifiers with JPA and Hibernate (and still compatible with anything else).
As explained in https://vladmihalcea.com/the-best-way-to-implement-equals-hashcode-and-tostring-with-jpa-and-hibernate/, the constant hashCode provides consistency but can decrease performance when using huge sets and maps.

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -440,19 +440,15 @@ public class <%= asEntity(entityClass) %> implements Serializable {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof <%= asEntity(entityClass) %>)) {
             return false;
         }
-        <%= asEntity(entityClass) %> <%= entityInstance %> = (<%= asEntity(entityClass) %>) o;
-        if (<%= entityInstance %>.getId() == null || getId() == null) {
-            return false;
-        }
-        return Objects.equals(getId(), <%= entityInstance %>.getId());
+        return id != null && id.equals(((<%= asEntity(entityClass) %>) o).id);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(getId());
+        return 31;
     }
 
     @Override
--- a/generators/server/templates/src/main/java/package/domain/Authority.java.ejs
+++ b/generators/server/templates/src/main/java/package/domain/Authority.java.ejs
@@ -35,6 +35,7 @@ import javax.persistence.Column;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * An authority (a security role) used by Spring Security.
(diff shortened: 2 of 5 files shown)


Blackdread:
About hashcode as constant, I understand why but do you ever put an entity in a map/set when it is not managed yet (not persisted)?
Yes but not *very often*, which can be problematic in some cases but I did not have yet. 

Similar to this talk: https://www.youtube.com/watch?v=EZwpOLCfuq4  but he uses UUID which is a good way to do but get extra column in DB.

pblanchardie:
I agree it's not "very often" but I've done so in several cases (eg. when using CascadeType.PERSIST). Anyway, more often than manipulating huge sets of entities (in such cases, specific implementations would help performance tuning).
Debugging this kind of equals/hashCode problem can be hard, and that's my main concern. You can end up with silent loss of data, amongst others.
This article also explains the issue: https://vladmihalcea.com/the-best-way-to-implement-equals-hashcode-and-tostring-with-jpa-and-hibernate/

If UUID were used as `@Id` (and set at construction time rather than persist - ie. not `@GeneratedValue`) traditional hashCodes can be used, but it's not the JHipster standard.

jdubois:
- For me the best choice is to have a "natural id", but obviously we can't generate that with JHipster. My idea was to have more of template that people would then customize: in fact, I'm pretty sure nobody every does this. That's bad, but that's the way people work, so let's accept this.
- I'm very surprised, reading https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/ , that having "31" in the equals method is a good idea. I was OK with having an equals that changes once the entity is flushed, but indeed that wasn't such a great idea at all.

-> this is a very important, major change, so that should be in JHipster 6. I'm ready to merge this but this is so big that I'd be happy to have more feedback

cbornet:
I don't think we can use an UUID as `@Id` because joins would be too slow. But maybe use another field  with UUID type as "natural ID" would be a good idea ?

jdubois:
@cbornet for UUID that's what I thought, but some people pointed to me that it depends on the database - so I'm not too sure about this. So leaving speed aside:

- UUID are bad because they'll take more space and bandwidth (we're going to have them everywhere), so it's clearly not as simple or efficient as a Long
- One good thing is that they hide information: when you have `/invoice/1` you can get the ID of the next invoice... But it's also a false sense of security (someone can send you the link to `/invoice/generated-uuid-string`) and the URLs would be impossible to spell.

So in the end I'm not really convinced. Even if I do a lot of Cassandra, and here we have UUID everywhere.

pblanchardie:
@jdubois we all agree that having "31" in hashCode (and not equals) is not such a good idea because of bucket-related performance issues, but it's the only safe way we found to generate reliable generic implementations in our context. Let's wait for feedback, i agree it's a big change as it impacts Entity, but I hope we'll merge!

jdubois:
So no feedback here, this looks like a good solution, and anyway on JHipster we are basically following all the best practices from @vladmihalcea as he's the best on this subject.

Thanks a lot for the analysis @pblanchardie and congrats for your first contribution to the project!!

PR #9034, opened 2019-01-09
Change the packaging to jar and configure a war profile

PierreBesson:
This needs review and a lot of testing to make sure it works for every case.

In order to build the war, you can run:
`./mvnw -Pprod,war verify`

In my testing, running the produced war file in tomcat and deploying to GAE worked well.

### DIFF
--- a/generators/common/templates/README.md.ejs
+++ b/generators/common/templates/README.md.ejs
@@ -247,7 +247,7 @@ Refer to [Using JHipster in production][] for more details.
 To package your application as a war in order to deploy it to an application server, run:
 <% if (buildTool === 'maven') { %>
     ./mvnw -Pprod,war clean package<% } %><% if (buildTool === 'gradle') { %>
-    ./gradlew -Pprod -Pwar clean bootWar<% } %>
+    ./gradlew -Pprod clean bootWar<% } %>
 
 ## Testing
 
--- a/generators/common/templates/README.md.ejs
+++ b/generators/common/templates/README.md.ejs
@@ -223,22 +223,31 @@ Refer to [Doing API-First development][] for more details.
 
 ## Building for production
 
-To optimize the <%= baseName %> application for production, run:
+### Packaging as jar
+
+To build the final jar and optimize the <%= baseName %> application for production, run:
 <% if (buildTool === 'maven') { %>
     ./mvnw -Pprod clean package<% } %><% if (buildTool === 'gradle') { %>
-    ./gradlew -Pprod clean bootWar<% } %>
+    ./gradlew -Pprod clean bootJar<% } %>
 
 <%_ if (!skipClient) { _%>
 This will concatenate and minify the client CSS and JavaScript files. It will also modify `index.html` so it references these new files.
 <%_ } _%>
 To ensure everything worked, run:
 <% if (buildTool === 'maven') { %>
-    java -jar target/*.war<% } %><% if (buildTool === 'gradle') { %>
-    java -jar build/libs/*.war<% } %>
+    java -jar target/*.jar<% } %><% if (buildTool === 'gradle') { %>
+    java -jar build/libs/*.jar<% } %>
 
 <% if (!skipClient) { %>Then navigate to [http://localhost:<%= serverPort %>](http://localhost:<%= serverPort %>) in your browser.
 <% } %>
 Refer to [Using JHipster in production][] for more details.
+
+### Packaging as war
+
+To package your application as a war in order to deploy it to an application server, run:
+<% if (buildTool === 'maven') { %>
+    ./mvnw -Pprod,war clean package<% } %><% if (buildTool === 'gradle') { %>
+    ./gradlew -Pprod -Pwar clean bootWar<% } %>
 
 ## Testing
 
@@ -293,7 +302,7 @@ To stop it and remove the container, run:
 You can also fully dockerize your application and all the services that it depends on.
 To achieve this, first build a docker image of your app by running:
 
-    <% if (buildTool === 'maven') { %>./mvnw package -Pprod verify jib:dockerBuild<% } %><% if (buildTool === 'gradle') { %>./gradlew bootWar -Pprod jibDockerBuild<% } %>
+    <% if (buildTool === 'maven') { %>./mvnw package -Pprod verify jib:dockerBuild<% } %><% if (buildTool === 'gradle') { %>./gradlew bootJar -Pprod jibDockerBuild<% } %>
 
 Then run:
 
--- a/generators/docker-base.js
+++ b/generators/docker-base.js
@@ -49,7 +49,7 @@ function checkImages() {
             runCommand = './mvnw package -Pprod verify jib:dockerBuild';
         } else {
             imagePath = this.destinationPath(`${this.directoryPath + appsFolder}/build/jib-cache`);
-            runCommand = './gradlew bootWar -Pprod jibDockerBuild';
+            runCommand = './gradlew bootJar -Pprod jibDockerBuild';
         }
         if (shelljs.ls(imagePath).length === 0) {
             this.warning = true;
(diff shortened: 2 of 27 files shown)


PierreBesson:
For gradle I have had a look at it but I'm not sure how to do it : @atomfrede do you have an idea ?

atomfrede:
@PierreBesson Just to make sure, the title of this PR is wrong, right? You mean `Change the packaging to **jar** for maven and configure a war profile`?
For gradle would not use a `-P` switch. Need to check but I think we can just change the default packing to jar. As we still apply the `war` plugin there is `bootWar` task available. Everything that needs a war file must just call bootWar instead bootJar or assemble, but will have look to check if thats correct.

chanseokoh:
The title says "Change the packaging to war" and I was confused until I actually see the code. It should say "Change the packaging to jar".

atomfrede:
@loosebazooka We currently have `war` and `jar` packaging available, just a matter of calling `bootJar` directly. I would prefer not to add more hidden flags as they are not discoverable via `gradle tasks`, but maybe that needs a vote.

chanseokoh:
@atomfrede @PierreBesson is it possible to remove the WAR Plugin for Gradle?

```diff
 apply plugin: 'maven'
 apply plugin: 'org.springframework.boot'
-apply plugin: 'war'
 apply plugin: 'propdeps'
 apply plugin: 'com.moowork.node'
 apply plugin: 'io.spring.dependency-management'
```

chanseokoh:
Actually, more places with WAR:

```diff
diff --git a/build.gradle b/build.gradle
index 6e4c913..798935d 100644
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,6 @@ assert System.properties['java.specification.version'] == '1.8'
 
 apply plugin: 'maven'
 apply plugin: 'org.springframework.boot'
-apply plugin: 'war'
 apply plugin: 'propdeps'
 apply plugin: 'com.moowork.node'
 apply plugin: 'io.spring.dependency-management'
@@ -54,14 +53,6 @@ version = '0.0.1-SNAPSHOT'
 
 description = ''
 
-bootWar {
-   mainClassName = 'com.example.demoapp.JhipsterApp'
-}
-
-war {
-
-}
-
 springBoot {
     mainClassName = 'com.example.demoapp.JhipsterApp'
     buildInfo()
@@ -255,9 +246,6 @@ wrapper {
     gradleVersion = '4.10.2'
 }
 
-task stage(dependsOn: 'bootWar') {
-}
-
 if (project.hasProperty('nodeInstall')) {
     node {
         version = "${node_version}"
diff --git a/gradle/profile_dev.gradle b/gradle/profile_dev.gradle
index a1b1eb5..22bbce8 100644
--- a/gradle/profile_dev.gradle
+++ b/gradle/profile_dev.gradle
@@ -28,10 +28,6 @@ task webpackBuildDev(type: NpmTask, dependsOn: 'npm_install') {
     args = ["run", "webpack:build"]
 }
 
-war {
-    webAppDirName = 'build/www/'
-}
-
 task copyIntoStatic (type: Copy) {
     from 'build/www/'
     into 'build/resources/main/static'
diff --git a/gradle/profile_prod.gradle b/gradle/profile_prod.gradle
index 26a2978..1eec73c 100644
--- a/gradle/profile_prod.gradle
+++ b/gradle/profile_prod.gradle
@@ -27,10 +27,6 @@ task webpack(type: NpmTask, dependsOn: 'npm_install') {
     args = ["run", "webpack:prod"]
 }
 
-war {
-    webAppDirName = 'build/www/'
-}
-
 task copyIntoStatic (type: Copy) {
     from 'build/www/'
     into 'build/resources/main/static'
```

PierreBesson:
@chanseokoh will correct me if I'm wrong but I think the problem is as long as there is the war plugin enabled in build.gradle, jib (starting with v0.10) will assume that we want war packaging.

In my tests I have been able to work around that by removing all the war code highlighted above and added this to the build.gradle : 
```
if (project.hasProperty('war')) {
    apply plugin: 'war'
    war {
        enabled = true
    }
}
```
In this regard it will work similarly to maven. Should I go forward with this change ?

chanseokoh:
> as long as there is the war plugin enabled in build.gradle, jib (starting with v0.10) will assume that we want war packaging.

That is correct.

> Should I go forward with this change ?

From the perspective of the Jib developer, that (not applying the WAR Plugin) is the way we want.

Just from my personal curiosity, is `war.enabled = true` necessary?

atomfrede:
@PierreBesson I would like to remove most `-P` flags, but in that case it seems we need to have it.

PierreBesson:
@chanseokoh No it is actually not required. I'm not sure why it was there but it could very well be because it was needed for older Spring Boot versions. JHipster actually started with very early versions of Spring Boot.

atomfrede:
@chanseokoh No I think I didn't remove it. According to the boot documentation this is supposed to enable war creation in addition to jar creation, but I realized is has no effect (maybe because the war plugin is already applied). If jib can handle executable jars and we do not support container deploymend anymore we can definitl;y remove the war plugin as there are bootJar and bootWar available via the boot plugin.

PierreBesson:
@atomfrede That is not correct, if you don't add the `apply plugin: war` then the spring boot plugin won't show the bootWar task so I think the solution using -Pwar is required.

atomfrede:
+1 was not sure to be honest.

generators/common/templates/README.md.ejs
```diff
@@ -293,7 +302,7 @@ To stop it and remove the container, run:
 You can also fully dockerize your application and all the services that it depends on.
 To achieve this, first build a docker image of your app by running:
 
-    <% if (buildTool === 'maven') { %>./mvnw package -Pprod verify jib:dockerBuild<% } %><% if (buildTool === 'gradle') { %>./gradlew bootWar -Pprod jibDockerBuild<% } %>
+    <% if (buildTool === 'maven') { %>./mvnw package -Pprod verify jib:dockerBuild<% } %><% if (buildTool === 'gradle') { %>./gradlew bootJar -Pprod jibDockerBuild<% } %>
```
chanseokoh: FYI: `bootJar` may not be necessary. Jib doesn't look into a packaged JAR, so normally it's enough to do `./gradlew jibDockerBuild`, unless you wired up extra tasks with `bootJar`.

generators/docker-base.js
```diff
@@ -49,7 +49,7 @@ function checkImages() {
             runCommand = './mvnw package -Pprod verify jib:dockerBuild';
         } else {
             imagePath = this.destinationPath(`${this.directoryPath + appsFolder}/build/jib-cache`);
-            runCommand = './gradlew bootWar -Pprod jibDockerBuild';
+            runCommand = './gradlew bootJar -Pprod jibDockerBuild';
```
chanseokoh: ditto

generators/docker-base.js
```diff
@@ -49,7 +49,7 @@ function checkImages() {
             runCommand = './mvnw package -Pprod verify jib:dockerBuild';
```
chanseokoh: Related, `package` is no longer required for the Jib standard (non-WAR) containerization. Normally, `./mvnw compile jib:dockerBuild` should work (although if we were to do the WAR containerization, `package` is required.)

generators/common/templates/README.md.ejs
PierreBesson: I wished our build could be as fast as just compiling the sources but we have a complex builds with resource copying and filtering and in some case even a front-end nodejs build to run. However I wish we could create a special mode to speed up everything and have a fast dev workflow with jib + scaffold. Anyway thanks a lot for your help !

pascalgrimaud:
Good job @PierreBesson !
It is an important PR, but should I start to test everything now, or should I wait for the merge of Spring Boot 2.1 branch ?

PierreBesson:
This is not related to Spring boot 2.1 but maybe it might be better to wait for it to be merged as this will need extensive manual testing.
The things to test are :
- [ ] running the app as jar
- [ ] running the app as war in an app server
- [ ] running the app as docker image
- [ ] deploy to heroku, gae, etc..
- [ ] ci-cd in Jenkins, gitlab...

pascalgrimaud:
I'll start to review this, but as there are some conflicts, can you rebase your branch plz @PierreBesson ?
Tell me when it's done

PierreBesson:
The rebase is done.

pascalgrimaud:
Tested and it works well.
I'm going to merge this, waiting the CI to finish

pascalgrimaud:
good job @PierreBesson !

DanielFran:
@PierreBesson @pascalgrimaud @jdubois Did not found any update in documentation about those changes...

pascalgrimaud:
@DanielFran : maybe we forgot to do it... Can you open a new ticket plz ? So we won't forget to do it

DanielFran:
@pascalgrimaud Done!

PR #9036, opened 2019-01-10
exclude node_modules  dir in idea project

rover12421:
https://www.jhipster.tech/configuring-ide-idea/

can remove "Exclude directories" node

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -67,6 +67,12 @@ apply plugin: 'com.moowork.node'
 apply plugin: 'io.spring.dependency-management'
 apply plugin: 'idea'
 
+idea {
+  module {
+    excludeDirs += files('node_modules')
+  }
+}
+
 dependencyManagement {
   imports {
     mavenBom 'io.github.jhipster:jhipster-dependencies:' + jhipster_dependencies_version


jdubois:
Oh that's awesome, I'd love to have this in Maven (but I don't think it's as easy to do)

atomfrede:
@rover12421 Great, never thought of that. Thx.

atomfrede:
@rover12421 Great, never thought of that. Thx.

pvliss:
Just my 2 cents. I am using IntelliJ IDEA Ultimate 2018.3.2 and the `node_modules` directory is excluded by the IDE by default. I even tested selecting the dir, then clicking on "Mark as--> Not excluded" and then refreshed the project (Re-import for Maven and Refresh all Gradle projects) and the folder was automatically added to the excluded folders list by IntelliJ. So this step seems to be no longer necessary

pvliss:
Oh it might be because of the .gitignore file. Will check

atomfrede:
For me too, but it doesn't hurt to make it explicit because I remember older versions which didn't ignore `node_modules` (maybe it depends on installed js tools/plugins).

pvliss:
> For me too, but it doesn't hurt to make it explicit because I remember older versions which didn't ignore `node_modules` (maybe it depends on installed js tools/plugins).

I definitely agree as other maybe using older versions.

Out of interest though I re-checked this time generating projects skipping git ,
e.g: 
```bash
jhipster --force --no-insight --skip-checks --with-entities --skip-git --skip-commit-hook --from-cl
```
and then removing the .gitignore file manually. For both Maven ad Gradle and IDEA still seems to auto-exclude the node_modules dir although it also marks it as a 'library root' and continues to index it.

@rover12421's change is also properly excluding the dir and seems to be no longer indexed or marked as library root which makes the loading of the project a lot faster

pvliss:
It definitely makes project loading faster but it also seems to disable auto-completion, code navigation, etc. for the frontend development, which seems kind of expected as the folder is no longer indexed. I believe that many users will not be happy with that.

@atomfrede  can you please also check? Maybe I am missing something

PR #9038, opened 2019-01-10
exclude node_modules dir for idea(maven) project

rover12421:
@jdubois maven can exclude dir.

@pvliss I using `JHipster v5.7.2` create `Monolithic application`, idea don't auto  exclude `node_modules` dir. but idea can auto exclude directories containing `package.json`.
my test idea is 2018.3.2.

---
After the `node_modules` directory is added and excluded, the maven/gradle configuration is removed from the exclusion. idea cannot restore the exclusion status.  only click `Cancel Exclusion` on the right-click menu.😇

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -100,6 +100,7 @@
         <maven-compiler-plugin.version>3.8.0</maven-compiler-plugin.version>
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
         <maven-enforcer-plugin.version>3.0.0-M2</maven-enforcer-plugin.version>
+        <maven-idea-plugin.version>2.2.1</maven-idea-plugin.version>
         <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
         <maven-surefire-plugin.version>2.22.1</maven-surefire-plugin.version>
         <maven-war-plugin.version>3.2.2</maven-war-plugin.version>
@@ -769,6 +770,10 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-idea-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
             </plugin>
             <plugin>
@@ -1004,6 +1009,14 @@
                                 <version>[1.8,1.9)</version>
                             </requireJavaVersion>
                         </rules>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-idea-plugin</artifactId>
+                    <version>${maven-idea-plugin.version}</version>
+                    <configuration>
+                        <exclude>node_modules</exclude>
                     </configuration>
                 </plugin>
                 <plugin>


rover12421:
@pvliss  maven-idea-plugin, like maven-eclipse-plugin, has been [retired](http://maven.apache.org/plugins/index.html#Retired). But my test is still valid, and I think it's a better solution. Of course,
If you think that the directory that IDEA is automatically excluded is also ok(subdirectory containing `package.json`), then the node_modules directory is not excluded.

I see ‘[Configuring Intellij IDEA](https://www.jhipster.tech/configuring-ide-idea/)’ needs to exclude the `node_modules` directory. I think this step building tool can be done automatically, no need to set it manually.

ruddell:
I think the message about ignoring node_modules is left over from when we used AngularJS, which only used node_modules for building the app.

Like @pvliss said, doesn't this also break some things like autocompletion for libraries?
> It definitely makes project loading faster but it also seems to disable auto-completion, code navigation, etc. for the frontend development, which seems kind of expected as the folder is no longer indexed. I believe that many users will not be happy with that.

pvliss:
@rover12421 

> maven-idea-plugin, like maven-eclipse-plugin, has been [retired](http://maven.apache.org/plugins/index.html#Retired).

I did not know that maven-eclipse-plugin is also retired as I have never actually used it myself.

> But my test is still valid, and I think it's a better solution. Of course,
> If you think that the directory that IDEA is automatically excluded is also ok(subdirectory containing `package.json`), then the node_modules directory is not excluded.

Well your solution is better in terms of that it truly excludes the `node_modules` directory from the project to the point that the directory is also **not indexed** and that is why the project loads faster. This is definitely better when you also do **not** do frontend development with IDEA in the same project. e.g.
- You use another editor(atom, vs code, etc) to do the frontend dev and use IDEA only for the backend
- You do not generate a client app at all

> I see ‘[Configuring Intellij IDEA](https://www.jhipster.tech/configuring-ide-idea/)’ needs to exclude the `node_modules` directory. I think this step building tool can be done automatically, no need to set it manually.

I agree that doing it automatically is always better but I am just not sure if all end users will agree or how they actually develop their apps. At the very least the documentation needs to be updated accordingly to reflect the change in behavior. I guess core @jhipster/developers are more qualified to answer and make a decision.

Also, documentation might get stagnant at times, e.g the same ‘[Configuring Intellij IDEA](https://www.jhipster.tech/configuring-ide-idea/)’ also mentions how to setup Spring support but nowadays this is also done automatically by IDEA for me.

jdubois:
Indeed, I liked this idea a lot, but this is a bad idea: it makes the project load faster, but at the cost of not being able to develop properly the front-end...

I'm sorry but I think this was an error, and that we should also rollback the Gradle PR on this.

gmarziou:
@jdubois  I agree with you that not indexing frontend libs is not a good thing, however it's not the same for all generator-jhipster templates that completely pollute any text search with code that you also find in your app.

PierreBesson:
@gmarziou So you mean just exclude `node_modules/generator-jhipster` ?

jdubois:
Oh @PierreBesson you have an exciting idea here!

deepu105:
in Vscode excluding node_modules is fine its automatic since we define it bin .gitignore and it doesn't affect frontend dev, I have never used Idea for frontend so I don't know if the behavior is different there. Did someone try to exclude node_modules from indexing and see if autocomplete and stuff work? My belief(based on how vscode does stuff) was that indexing is not required for using the lib

jdubois:
After doing more tests: I have no idea how it works, but this indeed works correctly. It seems you need to do a `npm start` first, but that's something you'll do quickly anyway.

So let's keep this, and see the feedback we get with the next release.

PR #9064, opened 2019-01-15
Fix elastic docker for azure ci

pvliss:
I think that adding `discovery.type=single-node` within double quotes should fix the problem for Azure CI build and allow for upgrade to version 2.2 as per example at https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html


pvliss:
Nope! Did not work. :confused: 
My local test worked just fine both with and without quotes but I am on Ubuntu and I thought it would be some Windows to bash issue or something.

Anyways, closing and see if I can fix on my branch first

PR #9049, opened 2019-01-12
blueprint for languages sub generator

pascalgrimaud:
Related to https://github.com/jhipster/jhipster-vuejs/issues/132

This is needed in Vue.js blueprint.
cc @sahbi-ktifa 

Now, I understand the amount of work you did in blueprint / refactoring yeoman. I spent days on this!
That's why I assigned to you @murdos, if you can help in review.

_____

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #9082, opened 2019-01-16
Improve needle api

avdev4j:
WIP PR to starting an improvement on needle api.

This is an exemple on what we can do to simplify and improve the JHipster API

-   [X] Base needle Api
-   [X] (S)CSS Api
-   [x] ~~Sync react and Angular (S)CSS names~~ (TODO in another PR)
-   [x] Webpack needle Api
-   [X] Angular add Module and Admin module
-   [x] Add external resource to root (index.html)
-   [x] ~~Remove addAngularJsInterceptor() ?~~ (TODO in another PR)
-   [x] Angular add Entity to module
-   [x] React add Entity to module
-   [x] Angular add Element to menu
-   [x] React add Element to menu
-   [X] Maven needle Api
-   [x] Gradle needle Api
-   [x] liquibase needle Api
-   [x] Cache needle API
-   [X] RewriteFile in generator-base
-   [x] i18n needle Api

###


avdev4j:
Hi,
I probably found a bug when `addEntityToModule()` is called. The whole method is wrapped by a try catch that mask an error.

>      Uncaught Error: src/main/webapp/app/entities/entity.module.ts doesn't exist

On line `jhipsterUtils.checkStringInFile(entityModulePath, 'loadChildren', this);` 

By moving out the code from the try block the error occurred. I wonder if it's a intended behavior or we have here a bug?

To fix it I added a try catch block to my needle api method. Probably we must review this.

avdev4j:
I've just finished to move all needle code to the internal/api classes. 

There 2 things remaining to do in another PR : 
* Sync react and Angular (S)CSS names
* Remove addAngularJsInterceptor()

All feedbacks are appreciate. If this PR is validated I think we can do the same with JsonWriting file API (it's smaller).

the goal is the same, having a clearer internal API by separating code. What do you think about ?

pascalgrimaud:
@avdev4j : just tested your branch with vuejs, it failed because it didn't include https://github.com/jhipster/generator-jhipster/pull/9049

Once I rebase, it works fine.

Could you update your branch plz ?
So I could do more test with VueJS blueprint.

avdev4j:
Done ! 
Feel free to feedback.

DanielFran:
@pascalgrimaud Seems that the issue has been fixed.
It is OK to merge it?

pascalgrimaud:
@DanielFran : with the tests I did, it looks good! So I'm confident to merge it, as it will help @avdev4j, who is working on blueprint !

murdos:
I come a bit late, but I have a few remarks:
- by using a subfolder `internal` inside `generators` folder, we have the feeling there's a new Yeoman subgenerator, whereas as far as I can tell this is not true
- I would rather have each specific needle-api as close as possible to the templates files it handles. For example the needle-server-gradle should be inside `server` subgen, needle-client-react inside `client` subgen, ... So if one day we remove the react client, we won't forget to remove the related needle api that is hidden in another folder (and we may not notice it BTW, since there's not unit tests ^^).
- I don't see any unit tests for this whole new code. It's too bad. For maintenance and non regression. But also to document the way to use this new API.

murdos:
Nevertheless this is still an great idea that improves the existing code and make it easier to consume the API, both internally and for module/blueprint developers.
My remarks are more about the way to do it.

pascalgrimaud:
I agree with you @murdos but @avdev4j didn't write new functions, but mainly moved into new folders, so it's not his fault :)
I think it deserves a specific ticket to write all missing tests on these functions.

avdev4j:
@murdos 
It's never too late ! Now the code has been rework we can easily move it to another folder.
Dealing with an "internal" folder was not the best thing to do but I didn't find a better solution.
Putting all API classes in the generator they belong is, probably, a good choice. I wonder how to do with base and server/client base files.

Now all this things said, this is what I suggest : 
I will be happy to open a new PR to move files to each generators and adding tests if they are missing.

@pascalgrimaud What do you think about this  ? Should we open an issue for unit tests ? Should we do this in the same PR ? We also can have one for "moving files" and one for the tests when the first will be merged.

pascalgrimaud:
I know you're working on a private blueprint @avdev4j, so don't hesitate to refactor what you need to achieve it

deepu105:
I'm personally ok with having the internal folder as it is now, as its internal and only developers of core will be accessing it its ok. About moving it to specific sub gen is something we can consider but there are some needles which are shared between react/angular for example, so I don't think we can move it too close. Another issue is dispersing them into different folders makes it a bit more difficult to maintain IMO(I have been maintaining the base API for a while and it was easy since its all in one place, you may miss something when you are doing some generic updates etc). But I'm not gonna hold you guys back if you think it's better as I really want to give more power to the stream leads when it comes to these decisions.

murdos:
To be honest, I saw the changes to generator-base.js only after adding my comments, so I guess the unit tests and documentation on how to use the new API issues are partially resolved :)
Having a separate API would ease having dedicated unit tests, but we can live without as we did in the past.

However I would like to explore moving things to specific subgen folder, e.g. a needle-api in the root of each subgen folder, and a PR would be welcome :)

avdev4j:
I put all API files in the same folder to also to have one entrypoint to it. For method shared by react/angular (for example) we could keep the "base-client" file outside the generator's folders.

> Having a separate API would ease having dedicated unit tests

It's also a reason why I did this. It's easier to test, easier to maintain, easier to learn... ;)

I will open a PR. I will only move the server part and we can decide which solution is the best. Is it ok for you ?

deepu105:
This should have been merged for v6 as there are some minor method signature changes

avdev4j:
which method are you talking about @deepu105  ?

deepu105:
@avdev4j addVendorSCSSStyle, addMainSCSSStyle, addMainCSSStyle etc seems to have been changed. It would break modules which use them

avdev4j:
Oh yes, that's why I fix that with another PR #9151
This method could not work because they related to a non existing file but breaking the signature was not a good idea too.

My PR fix the signature method and the file and needle they related to.

deepu105:
ya but this is already released on v5.8.1 so no point rolling it back

deepu105:
anyway what is done is done, we just need to be a bit more careful when it comes to public API and stuff

PR #9098, opened 2019-01-19
enable html5-mode for angular and react

### LINKED ISSUE #8462: Implement HTML5 pushstate for Angular & React clients

HTML pushstate has been a long discussion on the JHipster issues thread.
It should be nice to implements it, because it has many advantages for the end users.
May be motivating contribution with a bounty? I can pay from my pocket if necessary.

- SEO and server side rendering

- Using anchors in pages

- Avoiding some tricky Identity provider hacks (tokens are usually returned on the form of anchors)

[list](https://github.com/jhipster/generator-jhipster/issues?utf8=%E2%9C%93&q=pushstate+)

-   [x ] Checking this box is mandatory (this is just to show you read everything)

ruddell:
Fix #8462

Removes the `#` from the URL. Also handles non-existing client pages for React, similar to how Angular handles it in https://github.com/jhipster/generator-jhipster/pull/8909

Tested manually:
-  [x] OAuth2
-  [x] JWT
-  [x] UAA
-  [x] React default
-  [x] Angular default
-  [x] Angular Microservices
-  [x] Email links
-  [x] Heroku https://html-5-mode.herokuapp.com/
-  [x] Traefik

### DIFF
--- a/generators/client/templates/angular/webpack/webpack.dev.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.dev.js.ejs
@@ -40,7 +40,6 @@ module.exports = (options) => webpackMerge(commonConfig({ env: ENV }), {
         contentBase: './<%= DIST_DIR %>',
         proxy: [{
             context: [
-                /* jhipster-needle-add-entity-to-webpack - JHipster will add entity api paths here */
                 '/'
             ],
             target: `http${options.tls ? 's' : ''}://127.0.0.1:<%= serverPort %>`,
--- a/generators/client/templates/angular/webpack/webpack.dev.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.dev.js.ejs
@@ -39,16 +39,9 @@ module.exports = (options) => webpackMerge(commonConfig({ env: ENV }), {
     devServer: {
         contentBase: './<%= DIST_DIR %>',
         proxy: [{
-            context: [<% if (authenticationType === 'uaa') { %>
-                '/<%= uaaBaseName.toLowerCase() %>',<% } %>
+            context: [
                 /* jhipster-needle-add-entity-to-webpack - JHipster will add entity api paths here */
-                '/api',
-                '/management',
-                '/swagger-resources',
-                '/v2/api-docs',
-                '/h2-console',<% if (authenticationType === 'oauth2') { %>
-                '/login',<% } %>
-                '/auth'
+                '/'
             ],
             target: `http${options.tls ? 's' : ''}://127.0.0.1:<%= serverPort %>`,
             secure: false,
--- a/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
@@ -1,0 +1,30 @@
+<%#
+Copyright 2013-2019 the original author or authors from the JHipster project.
+
+This file is part of the JHipster project, see https://www.jhipster.tech/
+for more information.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-%>
+package <%=packageName%>.web.rest;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+
+@Controller
+public class ClientForwardController {
+    @GetMapping(value = "/**/{path:[^\\.]*}")
+    public String redirect() {
+        return "forward:/";
+    }
+}
--- a/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
@@ -23,8 +23,11 @@ import org.springframework.web.bind.annotation.GetMapping;
 
 @Controller
 public class ClientForwardController {
+    /**
+     * Forwards any unmapped paths (except those containing a period) to the client index.html
+     */
     @GetMapping(value = "/**/{path:[^\\.]*}")
-    public String redirect() {
+    public String forward() {
         return "forward:/";
     }
 }
--- a/generators/client/files-react.js
+++ b/generators/client/files-react.js
@@ -310,6 +310,7 @@ const files = {
                 { file: 'shared/auth/private-route.tsx', method: 'processJsx' },
                 { file: 'shared/error/error-boundary.tsx', method: 'processJsx' },
                 { file: 'shared/error/error-boundary-route.tsx', method: 'processJsx' },
+                { file: 'shared/error/page-not-found.tsx', method: 'processJsx' },
                 // model
                 'shared/model/user.model.ts'
             ]
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -75,6 +75,7 @@
     "angular-router-loader": "0.8.5",
     "angular2-template-loader": "0.6.2",
     "autoprefixer": "9.4.7",
+    "base-href-webpack-plugin": "2.0.0",
     "browser-sync": "2.26.3",
     "browser-sync-webpack-plugin": "2.2.2",
     "cache-loader": "2.0.1",
--- a/generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
@@ -36,7 +36,7 @@ const LAYOUT_ROUTES = [
                 },
                 ...LAYOUT_ROUTES
             ],
-            { useHash: true , enableTracing: DEBUG_INFO_ENABLED }
+            { enableTracing: DEBUG_INFO_ENABLED }
         )
     ],
     exports: [
--- a/generators/client/templates/angular/src/main/webapp/index.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/index.html.ejs
@@ -19,7 +19,7 @@
 <!doctype html>
 <html class="no-js" lang="<% if (enableTranslation) { %><%= nativeLanguage %><% } else { %>en<% } %>" dir="ltr">
 <head>
-    <base href="./" />
+    <base href="/" />
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <title><%= baseName %></title>
--- a/generators/client/templates/angular/src/main/webapp/manifest.webapp.ejs
+++ b/generators/client/templates/angular/src/main/webapp/manifest.webapp.ejs
@@ -25,7 +25,7 @@
   ],
   "theme_color": "#000000",
   "background_color": "#e0e0e0",
-  "start_url": "/index.html",
+  "start_url": ".",
   "display": "standalone",
   "orientation": "portrait"
 }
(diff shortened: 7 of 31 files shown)


generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
```diff
@@ -0,0 +1,30 @@
...
+limitations under the License.
+-%>
+package <%=packageName%>.web.rest;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+
+@Controller
+public class ClientForwardController {
+    @GetMapping(value = "/**/{path:[^\\.]*}")
```
ruddell: > all paths that do not contain a period (and are not explicitly mapped already) are Angular routes, and should forward to the home page
https://spring.io/blog/2015/05/13/modularizing-the-client-angular-js-and-spring-security-part-vii

generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
```diff
@@ -0,0 +1,30 @@
...
+    public String redirect() {
```
gmarziou: Maybe it should be named `forward()` and the quote from spring blog could be added as a  javadoc comment.

jdubois:
I am very impressed @ruddell - I wanted to do this for at least 2 years!!!
Doesn't it cause any issue with Zuul routes on the gateway?

Maybe we should fix that all microservice calls go through the same URL prefix (like /services/** ) so they are easier to map. Like we already have /api/** for the API calls. It's easier to configure gateways (not just ours) like this.

ruddell:
@jdubois Thanks, I tried this many times and only finally got it to work.

The Zuul routes take precedence, so you are right that if you name a microservice and entity the same thing, then generate it in the gateway, there will be issues.  For example, I have a `order` microservice and generate an `order` entity, the paths will match causing issues (`/order` is mapped to the microservice, so refreshing in the client on the `/order` entity listing returns a 403).

Changing the microservice path has been discussed a few times and a major release is a good time to do that.

jdubois:
OK, then let's discuss this. It's so important that we could also do a vote on the mailing list, but let's start here.

For me we have 3 solutions:

- Stay as today: it's simpler and there's nothing to change, but it can cause issues like you mentioned, and also I wonder if it would work fine for other gateways. I'm currently having some customers with Traefik and some with a simple Kubernetes loadbalancer: I'm not sure how that would work for them. Maybe it's fine, maybe not.
- Create a `/services/**` prefix for all microservices. So calling the `invoice` REST endpoint on microservice `order` would be: `/services/order/api/invoice`. I find this a bit complicated, but that's my favorite choice at the moment.
- Put everything behind the `/api/` prefix, that we already have. So that's probably a little less work. In my previous example, you would be calling: `/api/order/api/invoice`. I don't like the double `/api`, but then it's annoying if we need to redefine those routes in all microservices. Also, we have the same issue as in the first point, you can't have an `order` entity on the gateway.

-> what do you think? Do you have any other idea?

ruddell:
Option 2 is can be accomplished by setting `zuul.prefix: '/services/'` in the gateway, then adding `/services` to the URL in the entity service.  It's nice that this is a simple change, and this is the same way the JHipster Registry is configured.  I just tested it and it resolves the above issue where the microservice path matches the client path.  I'm not sure how Traefik sets the routes, so this may require other fixes.

I think having `/api` twice is a little confusing and could cause issues with a gateway entity like you mentioned. I'd prefer Option 2 unless someone has another idea.

antarus:
Traefik uses consul, it intercept just the name of the service in / <service name>
I think that modifying in the file traefik.toml:
`frontEndRule = "PathPrefixStrip: /{{.ServiceName}}"` in
`frontEndRule = "PathPrefixStrip: /services/{{.ServiceName}}"`, this should work. To test

ruddell:
I tested with Traefik/Consul, and the change to the base href cause the same issue mentioned in the first post about context-paths.  The app assumes the assets should be served from the root `/`, but the gateway runs under `/basename/`.  Otherwise it works the same with and without `/services/` in the URL.

This context-path issue can be fixed by configuring the base href to the expected value. Using [`base-href-webpack-plugin`](https://www.npmjs.com/package/base-href-webpack-plugin) we can make it configurable for dev/prod. The Angular CLI lets you set the base href when building an app, so I think this would be good to include (or at least document for Traefik/Tomcat users).

Very simple config, in `webpack.prod.js`:
```
const { BaseHrefWebpackPlugin } = require('base-href-webpack-plugin');.
...
plugins: [
    new BaseHrefWebpackPlugin({ baseHref: '/services/gatewayconsul/' })
...
```

mraible:
I like putting everything behind the `/api` prefix. Would it be possible to use `/api/order/ms/invoice`, substituting `ms` for the 2nd `api`?

ruddell:
In that case you are modifying the request mappings from the microservice (that's where the second `/api/` comes from).  There's also endpoints such as `/management/info` which don't contain `api`.  Is there a way to configure the Zuul routes like that?    

It's simple to add a prefix (both `/api/` and `/services/` work), but I don't know how to do more complicated Zuul route changes that have been requested

gmarziou:
First, zuul supports a prefix that it strips in `SimpleRouteLocator`, you can set it application.yml

~~~yaml
zuul:
    prefix: /api
~~~

Then we can add a filter that /api to service call

~~~java
package com.jhipster.demo.store.gateway;

import com.netflix.zuul.ZuulFilter;
import com.netflix.zuul.context.RequestContext;
import org.springframework.cloud.netflix.zuul.filters.support.FilterConstants;
import org.springframework.stereotype.Component;

import static org.springframework.cloud.netflix.zuul.filters.support.FilterConstants.REQUEST_URI_KEY;

/**
 * http://gateway.example.com/api/order/invoice to http://order.example.com/api/invoice
 */
@Component
public class AddApiPrefixPathFilter extends ZuulFilter {

    @Override
    public Object run() {
        RequestContext context = RequestContext.getCurrentContext();
        String requestUri = (String) context.get(REQUEST_URI_KEY);
        context.put(REQUEST_URI_KEY, "/api" + requestUri);

        return null;
    }

    @Override
    public boolean shouldFilter() {
        return true;
    }

    @Override
    public String filterType() {
        return "pre";
    }

    @Override
    public int filterOrder() {
        return FilterConstants.RIBBON_ROUTING_FILTER_ORDER - 1;
    }
}
~~~

The filtering works well but then I get an error 401 due to bad JWT signature  reported by microservice.
I can't investigate further now.

ruddell:
Aside from the Zuul route issue, I think this PR is complete.  I'll add some docs about the context-path to Trafeik and Production pages.

The Zuul routes only cause an issue when you name a microservice entity the same as the microservice basename (then generate it on the gateway).  I'd prefer to keep the change simple, but if someone wants to write a custom Zuul path mapper and test it, I'm fine with that.

jdubois:
Awesome @ruddell - I also like how simple the end result looks like!

So that leaves us with 2 options, let me try to summarize. If if we have microservice "micro" with entity "foo", we could have:

- `/services/micro/api/foo`-> all microservices are simply behind a `/services` prefix. This is the easiest option to set up.
- `/api/micro/foo`-> this requires a bit more logic on the routing side, and adds a few constraints, but this is shorter and probably nicer to read.

-> WDYT? Any other arguments pros/cons each option? Let's first gather detailed options here, and I'm also creating a Twitter poll to get more feedback.

deepu105:
I would vote for option 1 as its simpler and clearly distinguishes microservice endpoints

deepu105:
@jdubois also lets merge it only after merging the spring boot branch after next minor release

jdubois:
@deepu105 yes totally agree for the merge

Then concerning both options:

- I think the core team mostly want `/services/micro/api/foo`, and I agree it's easier
- The Twitter poll is not over yet, but it looks like the vast majority of people want `/api/micro/foo`

ruddell:
Quote from [React docs](https://facebook.github.io/create-react-app/docs/deployment#serving-apps-with-client-side-routing) regarding the `start_url` in `manifest.webapp`:
>When users install your app to the homescreen of their device the default configuration will make a shortcut to /index.html. This may not work for client-side routers which expect the app to be served from /. Edit the web app manifest at public/manifest.json and change start_url to match the required URL scheme, for example:
>  "start_url": ".",

I ran a Lighthouse audit and it succeeded with the same results as an old app.

deepu105:
@sendilkumarn @wmarques @sahbi-ktifa can you guys test this for your streams

deepu105:
@ruddell there seems to be a minor conflict, can you check

sahbi-ktifa:
With Vue.js, this is just a matter of a property in the Router module. I'll wait for this to get merged before updating the blueprint!

DanielFran:
+1 to use it by default

jdubois:
-1 to have this "by default", this should be the only option, we should just get rid of the hashbang in v6

deepu105:
Yes this should be the only option

generators/client/templates/angular/webpack/webpack.dev.js.ejs
```diff
@@ -39,16 +39,8 @@ module.exports = (options) => webpackMerge(commonConfig({ env: ENV }), {
...
-                /* jhipster-needle-add-entity-to-webpack - JHipster will add entity api paths here */
-                '/api',
-                '/management',
-                '/swagger-resources',
-                '/v2/api-docs',
-                '/h2-console',<% if (authenticationType === 'oauth2') { %>
-                '/login',<% } %>
-                '/auth'
+            context: [
+                '/'
```
wmarques: Why are we removing this ? We don't need to provide those URLs with html5 mode ?

wmarques:
Couldn't test due to maven dependency issue, don't know if it's normal..
`Non-resolvable import POM: Could not find artifact io.github.jhipster:jhipster-dependencies:pom:2.2.0-SNAPSHOT in oss.sonatype.org-snapshot (http://oss.sonatype.org/content/repositories/snapshots) `

Maybe it's because you didn't rebased master ?

generators/client/templates/angular/webpack/webpack.dev.js.ejs
ruddell: In the past, we had to add every backend path here.  Now, every request goes to the backend and is forwarded to index.html if it doesn't match a path or resource.

ruddell:
I rebased again, I'm guessing that the `jhipster-dependencies` error was from bad timing on my part with the last rebase and SB 2.1 merge.

pascalgrimaud:
Just tested this PR and it works well ! Good job @ruddell 
But I have a question. I didn't see any changes related to:

> Then concerning both options:
>    * I think the core team mostly want `/services/micro/api/foo`, and I agree it's easier
>    * The Twitter poll is not over yet, but it looks like the vast majority of people want `/api/micro/foo`

I don't know if it should be included to this PR.

jdubois:
@pascalgrimaud for the microservices URL I don't think we decided anything in the end. @ruddell do you have any specific code for this?

-> let's make a decision, and make it fast, as this has been waiting for too long and I want JHipster 6 as soon as possible. So let's go for `/services/micro/api/foo`, as with most people in the team I believe it's the best and easiest option.

DanielFran:
@ruddell Since https://github.com/jhipster/generator-jhipster/pull/9173 has been merged, can you please rename ClientForwardControllerIntTest.java.ejs to ClientForwardControllerIT.java.ejs ?
Thanks

pascalgrimaud:
@ruddell : I did few tests with your PR and it seems good, excepting the big issue with Microservices, but it's not related to your PR I think
So +1 to merge it, so we can test manually the next days

jdubois:
Yes let's merge the important stuff now so we can do a beta release soon.

ruddell:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/6828

jdubois:
I just approved the bounty, thanks @ruddell !

avdev4j:
hi @ruddell 
does removing the `addEntityToWebpack` from the public API was really wanted?

I am currently working on a PR to add tests and I have added one for the `addEntityToWebpack` method, with this recent modification my tests are failed.

I can remove the test related to this method but I want to be sure it's really we want to do.

thanks

Issue #9081, opened 2019-01-16
gradle 5.x support


atomfrede:
Gradle 5 is already a few month old. We should support gradle 5 out of box for jhipster 6. There should be no big changes required, but I am pretty sure we will have some deprecation warnings regarding gradle 6. Need to investigate a little further, but pretty sure these are from some of the plugins we use, so most likely we just need to wait for new plugins releases.

Edit. Should consider removing the apt plugins and use default gradle features https://github.com/gradle/gradle/issues/2300#issuecomment-460065789

pvliss:
@atomfrede I assume you already have started work on this one but in any case I would be very interested in helping out on this one. BTW are you also considering switching to Kotlin gradle scripts or maybe even consider having it as another user option?

PierreBesson:
@pvliss, If you are interested to work on this, we can assign the issue to you. As for switching to a Kotlin based gradle build I don't see how it would be an improvement.

If you want to improve our gradle build, there was a great audit by a member of the Gradle team: https://github.com/jhipster/generator-jhipster/issues/6252 and a lot of his ideas were not actually implemented because we moved to other concerns. @atomfrede also started experimental work on a [jhipster gradle plugin](https://github.com/hipster-labs/jhipster-gradle-plugin).

atomfrede:
@pvliss I also see no value currently in switching to kotlin as the current script with groovy works just fine for us. What will most likely *not* happen is, that we support both groovy and kotlin. There are a lot of things we can optimize for sure, but nothing urgent imho. The greatest thing would be to make the jhpster gradle plugin reality, if you would like to know more I can tell you in detail more, but the linked issue describes quite good a few points.

Some of them are more or less resolve (e.g. repositories, plugin usage). We still have a lot of `-P` flags, but some of them won't go away I guess. The testing part could be done know as I remember we had a quite long conversation about it and decided not to split unit and integration tests as it would be a breaking change (but let me check). We have optimized the up-to-date checking for node/frontend builds already, not sure if the boot plugin has still some issues with that, need to check in detail.

pvliss:
@atomfrede + @PierreBesson Thanks for your input.

[...]

PR #9101, opened 2019-01-20
Speed up gradle build

pvliss:
- Define separate logic for the creation of SpringBoot `buildInfo` for the dev and pro profiles. For prod it remains the same but for dev we do not generate a build timestamp as it is not really needed and it triggers constant re-compilation, etc. Also removed clean of resources as it is not needed and always triggers a full re-build regardless if anything has changed or not
- Fix duplicate war creation. The `bootWar` task is actually extending the `war` task hence defining it again and with a separate name was causing the duplication. Removed tasks for renaming and deleting the war file as they are no longer necessary
- Fix the `webpackBuildDev` task to execute `npm install` only when needed. Moved the dependency within the task and now also checking for changes of the webpack config files excluding the prod

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -88,11 +88,15 @@ version = '0.0.1-SNAPSHOT'
 description = ''
 
 bootWar {
-   mainClassName = '<%= packageName %>.<%= mainClass %>'
+    mainClassName = '<%= packageName %>.<%= mainClass %>'
+}
+
+war {
     <%_ if (!skipClient) { _%>
     webAppDirName = '<%= CLIENT_DIST_DIR %>'
     <%_ } _%>
     enabled = true
+    extension = 'war.original'
 }
 
 springBoot {
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -89,19 +89,14 @@ description = ''
 
 bootWar {
    mainClassName = '<%= packageName %>.<%= mainClass %>'
-}
-
-war {
     <%_ if (!skipClient) { _%>
     webAppDirName = '<%= CLIENT_DIST_DIR %>'
     <%_ } _%>
     enabled = true
-    classifier = 'original'
 }
 
 springBoot {
     mainClassName = '<%= packageName %>.<%= mainClass %>'
-    buildInfo()
 }
 
 if (OperatingSystem.current().isWindows()) {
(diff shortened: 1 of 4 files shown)


pvliss:
Relates to #6252 and #9081

atomfrede:
Thx @pvliss Looks good, but the second war task was there on purpose as `bootWar` creates an executable `.war` (correct me if I am wrong), while we currently also support generating a normal `.war` file (name `.war.original`) which can deployed in a servlet container. But I think we need to rethink the whole packing part after https://github.com/jhipster/generator-jhipster/pull/9034 is merged.

generators/server/templates/build.gradle.ejs
```diff
@@ -89,19 +89,14 @@ description = ''
 
 bootWar {
    mainClassName = '<%= packageName %>.<%= mainClass %>'
-}
-
-war {
```
atomfrede: Currently we can generate the normal executable `.war` and a normal `.war.original` so maybe let's keep it for now until we have changed the default packaging to `.jar`.

generators/server/templates/build.gradle.ejs
```diff
@@ -515,21 +510,3 @@ if (project.hasProperty('nodeInstall')) {
...
-    destinationDir file("$buildDir/libs/")
-    rename "$project.name-$version-original.war", "$project.name-{$version}.war.original"
-}
-
-task deleteWarFile(type: Delete) {
-  delete "$buildDir/libs/$project.name-$version-original.war"
-}
-
-renameWarFile.dependsOn war
-bootWar.dependsOn war,renameWarFile,deleteWarFile
```
atomfrede: Good point to have `bootWar` not to depend on the normal `war` tasks as most people do not need the classical `.war` file

generators/server/templates/build.gradle.ejs
pvliss: @atomfrede Good catch :+1:  Sorry about that but I missed it completely. I have pushed another commit that restores the generation of the secondary non-executable war. The difference is that instead of using `classifier` and relying on copy and rename, etc. I have instead used `extension = war.original` which actually has the same effect.
**NOTE:** Starting with Gradle 5.1 `extension` has been deprecated and replaced by  `archiveExtension`

generators/server/templates/build.gradle.ejs
pvliss: Yes, they can be generated separately as per [docs](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/html/#packaging-executable-and-normal)

generators/server/templates/build.gradle.ejs
```diff
@@ -88,20 +88,19 @@ version = '0.0.1-SNAPSHOT'
...
+    mainClassName = '<%= packageName %>.<%= mainClass %>'
 }
 
 war {
     <%_ if (!skipClient) { _%>
     webAppDirName = '<%= CLIENT_DIST_DIR %>'
     <%_ } _%>
     enabled = true
-    classifier = 'original'
+    extension = 'war.original'
```
atomfrede: +1

pvliss:
Thanks @atomfrede

PR #9104, opened 2019-01-20
Use Faker.js to generate sample test data

### LINKED ISSUE #9103: Use Faker.js to generate test data
Please note I have a fully working prototype on the PR below.

[Faker.js](https://github.com/marak/Faker.js/) is a JavaScript library to generate "fake" data. For example, it can generate sample first names.

Like we already do with the "user" and "authority" tables, the idea here is to generate rows of fake data for each generated entity:

- For each entity, there would be 20 rows of "fake" data, so people can better test/understand the generated code
- This data is generated in .csv files, so it can easily be edited with Excel (for example), in order to have better data (maybe even true business data)
- The generator is smart enough to recognize some column names, like "first_name", and will generate appropriate fake data for those
- It is limited to what can be inserted into databases: if you put advanced validation logic in your entity's fields, it will not work totally well. That means the data will be inserted (as it will work at the database level), but you will have validation errors when trying to save it from the Web UI, going through bean validation. I do not think this is a big problem, everybody understands this is fake data, and not perfect business data

I think this would help many people when starting their application for the first time, and then would still be a great help later, as being able to have some initial data in all tables is often a requirement.

jdubois:
Fix #9103


test/utils/expected-files.js
```diff
@@ -209,9 +209,9 @@ const expectedFiles = {
     ],
 
     userManagementServer: [
-        `${SERVER_MAIN_RES_DIR}config/liquibase/authorities.csv`,
-        `${SERVER_MAIN_RES_DIR}config/liquibase/users.csv`,
-        `${SERVER_MAIN_RES_DIR}config/liquibase/users_authorities.csv`,
+        `${SERVER_MAIN_RES_DIR}config/liquibase/authority.csv`,
+        `${SERVER_MAIN_RES_DIR}config/liquibase/user.csv`,
+        `${SERVER_MAIN_RES_DIR}config/liquibase/user_authority.csv`,
```
fleboulch: Those 3 csv files are expected to be in the `config/liquibase/data/` directory no?

pascalgrimaud:
Travis / Azure failed for OAuth2 only. For the other config, it seems to work fine ! 
I'll check later why.

jdubois:
I think this PR is complete... There's just one test failing but this isn't due to the Faker.js migration (I hope I'm not mistaken here), so this should be OK to merge.

I'm getting weird Git error, in the form of `.git/logs/refs/heads/jdubois-faker.js: SyntaxError: Invalid number (1:1)`, so it seems I have corrupted something in my Git repo, which makes me unable to test this well.

I'll do more tests but I might just merge the PR, as those are not linked to this code at all.

pascalgrimaud:
@jdubois : random Azure builds, with Elasticsearch. Can I try to restart it ?

jdubois:
@pascalgrimaud yes of course

But the issue I get is a CSRF error when saving an entity, which I can reproduce locally. I don't understand why Faker.js would have an impact here.

jdubois:
Oh I got the Git error, and that's so stupid....

So I named my branch `jdubois-faker.js` as the project uses Faker.js. So when I want to commit, Prettier comes as a pre-commit hook and tries to reformat the `.git/logs/refs/heads/jdubois-faker.js`-> as the files ends with a `.js`... That's so stupid!!

jdubois:
After much testing I'm confident this is good, but still I will merge it tomorrow morning so I will have time to fix stuff if I ever break something!

jdubois:
Built failed again because of CSRF issue!! Sorry everyone, I'm going to push this, and if it fails I will fix it tonight. I really don't get the issue here.

DanielFran:
@jdubois This is failling master since you merged it.

See Test react-noi18n-es-ws-gradle-session of https://jhipster.visualstudio.com/generator-jhipster/_build/results?buildId=2089

jdubois:
This really sucks, I'll work on this today

PR #8941, opened 2018-12-11
Added api docs into dto

javasparx:
Close https://github.com/jhipster/generator-jhipster/issues/8924

-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #9126, opened 2019-01-26
Api docs into dto: tests and missing change from code review

murdos:
This is an addition to #8941

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
@@ -62,12 +62,10 @@ import javax.persistence.Lob;
 import <%=packageName%>.domain.enumeration.<%= fields[idx].fieldType %>;
 <%_ } } _%>
 
-<%_ if (typeof javadoc == 'undefined') { _%>
 /**
  * A DTO for the <%= entityClass %> entity.
  */
-<%_ } else { _%>
-<%- formatAsClassJavadoc(javadoc) %>
+<%_ if (typeof javadoc !== 'undefined') { _%>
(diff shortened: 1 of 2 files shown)


generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -62,12 +62,10 @@ import javax.persistence.Lob;
 import <%=packageName%>.domain.enumeration.<%= fields[idx].fieldType %>;
 <%_ } } _%>
 
-<%_ if (typeof javadoc == 'undefined') { _%>
 /**
  * A DTO for the <%= entityClass %> entity.
```
murdos: This description seems more appropriate to describe this class than the comment from the JDL for the entity.
Otherwise the comment from the JDL would be on both the entity domain and the DTO class.

test/entity.spec.js
```diff
@@ -9,583 +9,648 @@ const CLIENT_MAIN_SRC_DIR = constants.CLIENT_MAIN_SRC_DIR;
...
-                        fse.copySync(path.join(__dirname, '../test/templates/default-gateway'), dir);
-                    })
-                    .withArguments(['foo'])
-                    .withPrompts({
-                        useMicroserviceJson: true,
-                        microservicePath: '../'
-                    })
-                    .on('end', done);
-            });
+    context('regeneration from json file', () => {
```
murdos: Since I found no way to define an entity comment/javadoc using `entity` subgenerator by CLI, I added new tests that regenerate entities from `.jhipster/*.json` files.
All changes above are just introduction of a context `creation from CLI` and related tabulations changes.

DanielFran:
Thanks @murdos

PR #9078, opened 2019-01-16
Switch Main(S)CSS to Global(S)CSS

avdev4j:
The JHipster API provide functions to add style to css and scss files. 
But the files targeted are named main.css and main.scss.

As I can see 'main.css' and 'main.scss' is no longer existing and are replaces by 'global.css' and 'global.scss'.

I suggest to switch all 'main' references to 'global'. (needle, files references and needles).

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #9151, opened 2019-01-29
Fix react angular css api and needles

avdev4j:

After doing this PR #9082 . I saw that API for needles in CSS was totally broke.
It used the wrong filename and only for Angular.

I first open this PR #9078 but it legit to open a new due to huge modifications.

I rolled back to old api method name addMainSCSSStyle and addMainCSSStyle because it's neutral enough for every client framework we support. "Main" means the principal CSS file, it's not related to the css filename.

So we can have : 
- global.css for angular
- app.css for react

In every case we are using the "main" keyword in needles (angular and react): 
`jhipster-needle-scss-add-main
jhipster-needle-css-add-main
`
I think it's better, because it's not related to the filename. If the filename change the needle is still available.

To be more clearer : 
The API exposes two methods with generics name addMainCSSStyle and addMainSCSSStyle. Both of them are using the needle API (react or angular). Inside the needle api methods we are dealing with real css filenames that can be different (app.css for react / global.css for angular).
whatever framework you choose, needles in css files are the same between react and angular.

What do you think about @deepu105 ? Should we go further and rename react/angular css files to have the same filename ? 


avdev4j:
I've figure out that a bug was forget into my last PR #9082.

Needles in the API and in files don't match.

This PR fix that. If you think this one is ok we should merge it quickly otherwise the CSS will not work.
If @deepu105 think having css files with same names is better (as we said before), so I suggest to doing this in another PR.

avdev4j:
As discussed with @deepu105 (and @murdos) I change the targeted branch to v5 maintenance to revert the method signatures with their original values and fix needles values.

deepu105:
@avdev4j I know this part didn't have tests earlier and that's why we missed the issues, so maybe its an opportunity to add some tests, will you be interested in adding some tests for this part?

avdev4j:
I would add tests but I’m not sure how to do this. I mean add unit tests that create an application AND call the api. 
I know how to do this with a blueprint but do we can having tests that do both in the main test package? 

Envoyé de mon iPhone

deepu105:
There is no guidelines on that you can choose the approach that you think
is best. We do app generation and tests for the upgrade parts. You can also
create a blueprint on the fly and test if needed

avdev4j:
Blueprint on the fly seems to be the best way to do this. We can easily call the api and expected files to be modified as you want.

I added unit test for base needle and also for the css part.

I would add more tests for the whole needle API but certainly in a separate PR for the v6.

avdev4j:
For me I've finished unit tests for the CSS part and I opened a new PR to adding all test cases that we could merge for the v6.

Does it seems enough for you @deepu105 ? (Blueprints on the fly are awesome)

avdev4j:
finally also added react unit tests

Issue #8482, opened 2018-10-06
Angular - Switch to reactive forms?


Tcharl:
##### **Overview of the feature request**

Reactive forms is a new paradigm brought by Angular 6.
It could be nice to upgrade update forms to highlight it, as the approach looks better than the actual form validation implemented by JHipster.

##### **Motivation for or Use Case**

Easier validation. I've already tested it for some of our forms and it looks easier to understand than ngModel + validation bindings.

##### **Related issues or PR**

Even @wmarques is OK for it!
https://github.com/jhipster/generator-jhipster/issues/7575

-   [x] Checking this box is mandatory (this is just to show you read everything)

wmarques:
@Tcharl Indeed it's a good idea I also find it better for validation. That's quite a big work to do so maybe we can try to do that for JHipster v6 ?

Tcharl:
It could be a good target for V6, as it would be very visible to the Angular users.
I'll try to implement it on a WIP branch when I'll finish the current contributions

DanielFran:
@thainan10 Are you working in Reastive forms ? See https://github.com/jhipster/generator-jhipster/issues/7575

PR #9153, opened 2019-01-29
Use Angular reactive forms

Tcharl:
Fix https://github.com/jhipster/generator-jhipster/issues/8482


generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
```diff
@@ -43,50 +43,50 @@
...
+                <form *ngIf="!success" name="form" role="form" (ngSubmit)="finishReset()" [formGroup]="passwordForm">
                     <div class="form-group">
                         <label class="form-control-label" for="password" jhiTranslate="global.form.newpassword">New password</label>
-                        <input type="password" class="form-control" id="password" name="password" #passwordInput="ngModel"
+                        <input type="password" class="form-control" id="password" name="password"
                                placeholder="{{'global.form.newpassword.placeholder' | translate}}"
-                               [(ngModel)]="resetAccount.password" minlength=4 maxlength=50 required>
-                        <div *ngIf="passwordInput.dirty && passwordInput.invalid">
+                               formControlName="newPassword">
+                        <div *ngIf="passwordForm.get(['newPassword']).invalid && (passwordForm.get(['newPassword']).dirty || passwordForm.get(['newPassword']).touched)">
```
wmarques: I generally use a helper function like `isFieldInvalid(field: string)` to reduce the complexity within the HTML File.
WDYT about doing this for the error checks?

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
```diff
@@ -28,28 +30,32 @@ import { PasswordResetFinishService } from './password-reset-finish.service';
...
     error: string;
     keyMissing: boolean;
-    resetAccount: any;
     success: string;
     modalRef: NgbModalRef;
     key: string;
 
+    passwordForm = this.fb.group({
+        newPassword: ['', [Validators.required, Validators.minLength(4), Validators.maxLength(50)]],
+        confirmPassword: ['', [Validators.required, Validators.minLength(4), Validators.maxLength(50)]],
```
wmarques: Does this works? I always used `Validators.compose` to use multiple validators...

generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
```diff
@@ -31,51 +32,70 @@ import { Register } from './register.service';
...
     register() {
-        if (this.registerAccount.password !== this.confirmPassword) {
+        let registerAccount = {};
+        const login = this.registerForm.get(['login']).value;
+        const email = this.registerForm.get(['email']).value;
+        const password = this.registerForm.get(['password']).value;
+        if (password !== this.registerForm.get(['confirmPassword']).value) {
             this.doNotMatch = 'ERROR';
         } else {
+            registerAccount = Object.assign(registerAccount, {
```
wmarques: Maybe we can use spread operator here:
`registerAccount = {...registerAccount, login, email, password} `

generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
```diff
@@ -31,51 +32,70 @@ import { Register } from './register.service';
...
+            });
             this.doNotMatch = null;
             this.error = null;
             this.errorUserExists = null;
             this.errorEmailExists = null;
 <%_ if (enableTranslation) { _%>
             this.languageService.getCurrent().then((key) => {
-                this.registerAccount.langKey = key;
-                this.registerService.save(this.registerAccount).subscribe(() => {
+                registerAccount = Object.assign(registerAccount, {
```
wmarques: same, spread operator

generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
```diff
@@ -31,51 +32,70 @@ import { Register } from './register.service';
...
+                    langKey: key
+                });
+                this.registerService.save(registerAccount).subscribe(() => {
                     this.success = true;
                 }, (response) => this.processError(response));
             });
 <%_ } else { _%>
-            this.registerAccount.langKey = 'en';
-            this.registerService.save(this.registerAccount).subscribe(() => {
+            registerAccount = Object.assign(registerAccount, {
```
wmarques: spread operator

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
```diff
@@ -32,19 +33,31 @@ export class UserMgmtUpdateComponent implements OnInit {
...
-        private router: Router
+        private router: Router,
+        private fb: FormBuilder
     ) {}
 
     ngOnInit() {
         this.isSaving = false;
         this.route.data.subscribe(({user}) => {
             this.user = user.body ? user.body : user;
+            this.valorizeForm(this.user);
```
wmarques: I would prefer this method named `updateForm`, more clear for me, WDYT ?

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
```diff
@@ -57,12 +70,25 @@ export class UserMgmtUpdateComponent implements OnInit {
         <%_ } _%>
     }
 
+    private valorizeForm(user): void {
```
wmarques: Can we have some typing for the user  parameter ?

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.ts.ejs
```diff
@@ -70,62 +71,133 @@ export class <%= entityAngularName %>UpdateComponent implements OnInit {
...
+<%_ } _%>
+        protected activatedRoute: ActivatedRoute,
+        private fb: FormBuilder
     ) {
     }
 
     ngOnInit() {
         this.isSaving = false;
         this.activatedRoute.data.subscribe(({<%= entityInstance %>}) => {
+            this.valorizeForm(<%= entityInstance %>);
```
wmarques: rename to  `updateForm` if that's ok for you

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.ts.ejs
```diff
@@ -134,8 +206,25 @@ export class <%= entityAngularName %>UpdateComponent implements OnInit {
         return this.dataUtils.openFile(contentType, field);
     }
 
-    setFileData(event, entity, field, isImage) {
-        this.dataUtils.setFileData(event, entity, field, isImage);
+    setFileData(event, field: string, isImage) {
```
wmarques: Why this change?

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
vishal423: If used at more than one place, I generally create getter methods and access those in template and component class like below:

```
get password(): FormControl {
  return this.passwordForm.get('newPassword');
}
```

// use in template like:
`password.invalid `

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
```diff
@@ -62,10 +68,12 @@ export class PasswordResetFinishComponent implements OnInit, AfterViewInit {
     finishReset() {
         this.doNotMatch = null;
         this.error = null;
-        if (this.resetAccount.password !== this.confirmPassword) {
+        const password = this.passwordForm.get(['newPassword']).value
+        const confirmPassword = this.passwordForm.get(['confirmPassword']).value
+        if (password !== confirmPassword) {
```
vishal423: an alternative could be to write custom validator function

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
```diff
@@ -71,6 +97,16 @@ export class UserMgmtUpdateComponent implements OnInit {
         }
     }
 
+    private updateUser(user): void {
+        user.login = this.editForm.get(['login']).value;
```
vishal423: if form and model structure matches, then, you can get model object directly using `this.editForm.value`

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
```diff
@@ -28,28 +30,32 @@ import { PasswordResetFinishService } from './password-reset-finish.service';
...
-    confirmPassword: string;
     doNotMatch: string;
     error: string;
     keyMissing: boolean;
-    resetAccount: any;
     success: string;
     modalRef: NgbModalRef;
     key: string;
 
+    passwordForm = this.fb.group({
```
vishal423: an alternative could be to move form model outside in separate model class and do mapping between dto model and form model using services

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
Tcharl: Yes, it works, do you prefer compose?

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.ts.ejs
Tcharl: Because we don't have entity at this stage: I'll first try to make it work in a generated project, then contribute to ng-jhipster with a new method.

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
Tcharl: do you mean that I could use user = this.editForm.value in this case? if it is, this is AWESOME!

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
Tcharl: I wanted to stay coherent for form interaction between client and entity classes.
Generating a new model for form is a much time consuming and don't add much value (at least on my side, where forms are 3 to 5 attributes): may be doing this step by step if it is asked and voted (sorry in advance for those who will migrate)?

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
vishal423: yes

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
wmarques: No this way looks clearer for me so that's ok 😄

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
wmarques: Yes if the model and the form are the same structure it works well :) good catch @vishal423

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
wmarques: Not a bad idea but I would prefer a wrapper to check invalid, dirty and touched and then return a boolean

vishal423: wrapper can also leverage getter() functions. Point is to reduce complexity from templates (at moment, I see `passwordForm.get(['newPassword'])` code is repeated at more than 5+ places in this template and any future change in property name or object structure would require modification at all these places.

wmarques: You're right, but I would do a wrapper which takes in a parameter `fieldName` (like `newPassword`) and returns a boolean.
With getters we couldn't do that way, we would need to do a `isInvalid` method per field, if I understood you correctly

Tcharl: I'm not fan of wrapper: imagine a MOA want to show that some fields are mandatory (errored) without the end user to enter on the field.
If it's a sine qua non feature I'll do, but I would personally prefer without it

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
Tcharl: Yes, but I wouldn't change the original code so much

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
Tcharl: Tried but it only works if all the fields are identical: unfortunately, there are case where fields are undefined so that it fails...

Tcharl:
Everything's perfectly working on local: does someone knows why travis is failing?
May be because of Linux and date: does someone with a linux machine can test?

Thank you in advance

pascalgrimaud:
I'm closing and reopening to restart the build

wmarques:
@Tcharl looks like you have issues with e2e tests + compilation errors `ERROR in : Can't bind to 'translateValues' since it isn't a known property of 'h2'. ("<div> <div class="row justify-content-center"> <div class="col-md-8"> <h2 [ERROR ->][translateValues]="{username: settingsForm.get(['login']).value}" *ngIf="settingsForm.get(['login']).")`

Tcharl:
Everything's good to me: do you want something more in order to merge?

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
wmarques: Ok, let's keep it that way, it is indeed more flexible

generators/client/templates/angular/src/main/webapp/app/account/password-reset/init/password-reset-init.component.ts.ejs
```diff
@@ -50,7 +51,7 @@ export class PasswordResetInitComponent implements OnInit, AfterViewInit {
         this.error = null;
         this.errorEmailNotExists = null;
 
-        this.passwordResetInitService.save(this.resetAccount.email).subscribe(() => {
+        this.passwordResetInitService.save(this.resetRequestForm.get(['email']).value).subscribe(() => {
```
wmarques: Can't we just use `.get('email')`?

generators/client/templates/angular/src/main/webapp/app/account/password/password.component.html.ejs
```diff
@@ -32,63 +32,62 @@
...
-                    <input type="password" class="form-control" id="currentPassword" name="currentPassword" #currentPasswordInput="ngModel"
+                    <input type="password" class="form-control" id="currentPassword" name="currentPassword"
                            placeholder="{{'global.form.currentpassword.placeholder' | translate}}"
-                           [(ngModel)]="currentPassword" required>
-                    <div *ngIf="currentPasswordInput.dirty && currentPasswordInput.invalid">
+                           formControlName="currentPassword">
+                    <div *ngIf="passwordForm.get(['currentPassword']).invalid && (passwordForm.get(['currentPassword']).dirty || passwordForm.get(['currentPassword']).touched)" >
                         <small class="form-text text-danger"
-                               *ngIf="currentPasswordInput.errors.required" jhiTranslate="global.messages.validate.newpassword.required">
+                               *ngIf="passwordForm.get(['currentPassword']).errors.required" jhiTranslate="global.messages.validate.newpassword.required">
```
wmarques: use the string parameter instead of array one: `.get('currentPassword')`

wmarques:
@Tcharl looks good, only a comment about the usage of the `.get()` function, I think you can directly use a string parameter instead of an array one

generators/client/templates/angular/src/main/webapp/app/account/password/password.component.html.ejs
Tcharl: Done for each fields!

Tcharl:
@wmarques Done, there was many array to remove and it's now better from a end user point of view.
React tests are failing on ci, I don't think that it's my PR fault.

wmarques:
@Tcharl LGTM, thanks alot !

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
yelhouti: @Tcharl I know this is probably old, but could you please specify which field where problematic, since I already implemented all of this in a blueprint. but maybe I missed something.

PR #9154, opened 2019-01-29
Move client app to classpath:/public

### LINKED ISSUE #9118: Move client app to classpath:/public

By default Spring-boot serves anything that is into `classpath:/public`
So I propose to build the client into `target/classes/public` instead of `target/www`.
With this we can remove our custom addition of static assets for IDEs in WebConfigurer.
So far I've tested with IDEA, maven, and `java -jar`. Not yet by deploying in an app server, not yet with gradle.
We have a lot of ways to run the app and it can be tricky. Do you see use-cases I'm missing ?

* Policy 1, less custom code
* Our current implementation can have a bad effect if someone runs the prod war from a dir that has a `target/www` directory. In that case our detection will serve what is in this folder instead of what is in the war.
* This would be compatible with Webflux (our current is not : you can't do `setDocumentRoot` in Webflux)
* After that, we may be able to use Spring's built-in cache control instead of our custom cache filter.

cbornet:
Fix #9118 

### DIFF
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -115,6 +115,7 @@ const files = {
         {
             path: MAIN_SRC_DIR,
             templates: [
+                'WEB-INF/web.xml',
                 { file: 'favicon.ico', method: 'copy' },
                 'robots.txt',
                 '404.html',
--- a/generators/client/files-react.js
+++ b/generators/client/files-react.js
@@ -102,6 +102,7 @@ const files = {
         {
             path: MAIN_SRC_DIR,
             templates: [
+                'WEB-INF/web.xml',
                 { file: 'favicon.ico', method: 'copy' },
                 'robots.txt',
                 '404.html',
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -331,7 +331,8 @@ module.exports = class extends BaseBlueprintGenerator {
                 this.styleSheetExt = this.useSass ? 'scss' : 'css';
                 this.pkType = this.getPkType(this.databaseType);
                 this.apiUaaPath = `${this.authenticationType === 'uaa' ? `${this.uaaBaseName.toLowerCase()}/` : ''}`;
-                this.DIST_DIR = this.BUILD_DIR + constants.CLIENT_DIST_DIR;
+                this.DIST_DIR = this.getResourceBuildDirectoryForBuildTool(this.configOptions.buildTool) + constants.CLIENT_DIST_DIR;
+                this.AOT_DIR = `${this.getResourceBuildDirectoryForBuildTool(this.configOptions.buildTool)}aot`;
                 this.CLIENT_MAIN_SRC_DIR = constants.CLIENT_MAIN_SRC_DIR;
             },
 
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -167,8 +167,8 @@
     "lint": "tslint --project tsconfig.json -e 'node_modules/**'",
     "lint:fix": "<%= clientPackageManager %> run lint <%= optionsForwarder %>--fix",
     "ngc": "ngc -p tsconfig-aot.json",
-    "cleanup": "rimraf <%= BUILD_DIR %>{aot,www}",
-    "clean-www": "rimraf <%= BUILD_DIR %>/www/app/{src,<%= BUILD_DIR %>}",
+    "cleanup": "rimraf <%= DIST_DIR %> <%= AOT_DIR %>",
+    "clean-www": "rimraf <%= DIST_DIR %>app/{src,<%= BUILD_DIR %>}",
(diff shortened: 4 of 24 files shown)


cbornet:
Note : with this, there is no index.html error page when you start the IDE without first building the front-end (the one with instructions on how to troubleshoot).
So should we put index.html in `src/main/resources/public` ? Maybe also other resources that are not processed by webpack like favicon, loading.css, ... ?

deepu105:
@cbornet what about the issues with web servers that forced us to revert this last time? did you test that? cc @ruddell

deepu105:
Also is the move from `src/webapp` to `src/www` necessary? I personally like the `webapp` folder name as its clearer even to novice devs and also I would like to avoid an unnecessary breaking change if possible

cbornet:
The problem is that by default the maven-war-plugin copies everything that is in src/main/webapp to the root of the war

cbornet:
@deepu105 if it's important to keep the code in src/main/webapp it should be possible to prevent the copy by configuring maven-war-plugin with packageExcludes.

cbornet:
There is a discussion on this in #9118 . Let's discuss there !

cbornet:
Eventually I've left the client source code in `src/main/webapp` for the moment. We can move it to another dir later.
I tested the following which seem to work:
* maven/gradle
* angular/react
* executable war
* war.original in Tomcat
* docker/jib
* launch from IDE
* no devtool reload on /static change

deepu105:
So there is no encoding issue when running in Tomcat?

cbornet:
No everything looks fine for me. Would you double check ?

deepu105:
Unfortunately I'm on a tight schedule, may be @ruddel can help

ruddell:
This PR still has the same issue that the other one had.

Steps to reproduce:
 - Generate an app. I recommend a UAA gateway with no database and no service discovery, otherwise you have to set up a database reachable by Tomcat.
```
{ "generator-jhipster": { "promptValues": { "packageName": "com.mycompany.myapp", "nativeLanguage": "en" }, "jhipsterVersion": "5.8.1", "applicationType": "gateway", "baseName": "jhipster", "packageName": "com.mycompany.myapp", "packageFolder": "com/mycompany/myapp", "serverPort": "8080", "authenticationType": "uaa", "uaaBaseName": "uaa", "cacheProvider": "no", "enableHibernateCache": false, "websocket": false, "databaseType": "no", "devDatabaseType": "no", "prodDatabaseType": "no", "searchEngine": false, "messageBroker": false, "serviceDiscoveryType": false, "buildTool": "maven", "enableSwaggerCodegen": false, "clientFramework": "angularX", "useSass": true, "clientPackageManager": "npm", "testFrameworks": [], "jhiPrefix": "jhi", "entitySuffix": "", "dtoSuffix": "DTO", "otherModules": [], "enableTranslation": true, "nativeLanguage": "en", "languages": [ "en" ] } }
```
 - Add `<h2>&nbsp; &euro; &times;</h2>` to `home.component.html`
 - Package into war
 - Start Tomcat, I used below docker-compose:
```
version: '2'
services:
    tomcat:
        image: tomcat:8-jre8-alpine
        ports: 
            - 8080:8080
```
 - Copy `.war.original` to the Tomcat container 
```
docker cp target/jhipster-0.0.1-SNAPSHOT.war.original tomcat-container-id:/usr/local/tomcat/webapps/issue.war   
```
 - Open app once launched through Tomcat at http://localhost:8080/issue/

The string added above looks like `Â  â‚¬ Ã—` but it should look like &nbsp; &euro; &times;

cbornet:
Thanks @ruddell . I could reproduce. So now I can probably fix it :wink:

cbornet:
I think I found the cause : Tomcat has a charset mapper which [defaults to ISO-8859-1 for en and fr](https://github.com/apache/tomcat/blob/c4b006e3dd46ca8604b2c9418a4714b6261aab7c/java/org/apache/catalina/util/CharsetMapperDefault.properties).
If you change your browser language to something else then the issue disappears.
When using the embedded Tomcat, Spring overrides the default mapping in `resetDefaultLocaleMapping()`.

So we could:
1 / add a `web.xml` with `<locale-encoding-mapping-list>` with locales that we want to map to UTF-8 encoding.
2/ Or add a `web.xml` with `<mime-mapping>` to map `html` to `text/html;charset=utf-8`

I think 2/ is better since it will force UTF-8 whatever the locale and is closer to what we do for the embedded Tomcat in `WebConfigurer::setMimeMappings()`

Note : I haven't found a way to do it in Java with the ApplicationWebXml initializer.

cbornet:
Everything should be OK now. @ruddell @deepu105 can you test ?

ruddell:
The Tomcat encoding issue is fixed with your latest changes 👍

deepu105:
@cbornet @ruddell go ahead and merge if you guys have tested it

PR #9159, opened 2019-01-31
Fix bug on needle value for (s)css

avdev4j:
Hi,
We recently merge a modification on needle API. A bug has come with it because needles in files and in the method do not match.

To fix I opened a PR #9151 but it breaks the ne public API by modifying the method signature.
This is planned for the v6. To be fix quickly the bug I've just cherry pick the commit which fix the needle value bug.

@deepu105 
I think we can merge this one to first fix the bug and then, for the v6, decide if we revert public API method or not. Is it ok for you ? 


avdev4j:
I close this, since I opened two PRs with a more complete fix : 
- for master #9151
- for v5 maintenance #9214

PR #7923, opened 2018-07-03
add control for locale string format

vcupelloni:
add a control to distinguish different format of locale string en-US/en_US in reading profile infos in UserService class

Fix #7866

@mraible 
- [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
- [ ] Tests are added where necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

Issue #7866, opened 2018-06-26
java.lang.StringIndexOutOfBoundsException in UserService.java


vcupelloni:
##### **Overview of the issue**

Gateway generated with JHipster5 and configured to work with Okta have issues in UserService

`java.lang.StringIndexOutOfBoundsException: String index out of range: -1`
	at java.lang.String.substring(String.java:1967)
	at `com.mycompany.myapp.service.UserService.getUser(UserService.java:261)`
	at com.mycompany.myapp.service.UserService.getUserFromAuthentication(UserService.java:154)
	at com.mycompany.myapp.service.UserService$$FastClassBySpringCGLIB$$2169c3ce.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:746)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.aop.aspectj.MethodInvocationProceedingJoinPoint.proceed(MethodInvocationProceedingJoinPoint.java:88)
	at `com.mycompany.myapp.aop.logging.LoggingAspect.logAround(LoggingAspect.java:85)

##### **Motivation for or Use Case**

Cannot login to the gateway.

##### **Reproduce the error**

Create a fresh new gateway.
Configure to use Okta authentication.
Try to login.

##### **Suggest a Fix**

It seems that the user locale provided by Okta is like 'en_US' but at line 261 there is a substring on '-' that cause the exception.

##### **JHipster Version(s)**

5.0.1

##### **Browsers and Operating System**

- [ ] Checking this box is mandatory (this is just to show you read everything)

gmarziou:
Please paste output of `jhipster info`

vcupelloni:
✘ vcupelloni@ROMLAPA0693  ~/test/gateway> jhipster info
Using JHipster version installed locally in current project's node_modules
Executing jhipster:info
Options: 
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
gateway@0.0.0 /home/vcupelloni/test/gateway
└── generator-jhipster@5.0.1 

```


##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**


<details>
<summary>.yo-rc.json file</summary>
<pre>
{
[...]

PR #9161, opened 2019-01-31
Feature/enhancement translation for oauth

lmbai:
in the PR #8525 , we fix the language issue for the normal JWT cases. 
Actually for the oauth case, we should apply the same flow, the logic will be:
- when get the account info (when refresh the page or login), update the lang to the user preferred lang
- if use set the lang via the nav bar. this setting will persist into the sessionStorage, it will apply for current page only. 

 Extra things,  the Okta and KeyCloak will return the language with the key `locale`. but the value is different with our jhipster lang key.  
> in the #7866 ,  and [#7923](https://github.com/jhipster/generator-jhipster/pull/7923)  we fix the `en-US` -> `en` mapping for the Okta,  but it donesn't work for Keycloak.
Even for KeyCloak, they are have a lang key `zh-CN`, but in jhipster side, we are using `zh-cn`. 

there is no way for jhipster to cater all the lang keys.  here is the new logic for us:

- if the lang key start with `en-` or `en_`, we will set it to `en`, since in Jhipster, we only one en lang.
- otherwise, we will use lowercase of the lang-key from the oauth side. since in Jhipster, we only use lowercase for lang-key. 
- for the other case, the developer should mapping it accordingly. 

 
### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 import { Injectable } from '@angular/core';
-<%_ if (enableTranslation && authenticationType !== 'oauth2' && authenticationType !== 'uaa') { _%>
+<%_ if (enableTranslation && authenticationType !== 'uaa') { _%>
 import { JhiLanguageService } from 'ng-jhipster';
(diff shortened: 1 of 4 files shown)


lmbai:
@ruddell hi, i reopen the PR here. 
you a right, the UAA will have issue if i move the change lang logic in to the retrieve account info. 
now i revert the logic back. the UAA change lang logic will stay in the login flow.

jdubois:
LGTM, thanks @lmbai as this is very valuable work and quite hard to do!

PR #9179, opened 2019-02-02
Prevent routes in entity.module.ts to be duplicated upon reimporting JDL

### LINKED ISSUE #9178: Routes in entity.module.ts are duplicated upon reimporting jdl

If you make changes to a jdl file, and re-generate the model, the routes in entity.module.ts are duplicated (the generator fails to see they already exist).

Route shouldn't be duplicated

##### **Reproduce the error**

1. Create a JDL with an entity
1. Import it
1. Modify a field in the JDL
1. Re-import it
1. See entity.module.ts : the entity's route is present twice

##### **Suggest a Fix**

I have a fix in preparation

##### **JHipster Version(s)**

5.8.1

OS X 10.14

afillatre:
The generator fails to detect already existing routes because the template
was broken (missed a line separator after the comma), and the template
would not splitted into multiple lignes. This led to non-matching regex
in the utils.rewrite funtion.

closes #9178

### DIFF
--- a/generators/internal/needle-api/needle-client-angular.js
+++ b/generators/internal/needle-api/needle-client-angular.js
@@ -193,18 +193,19 @@ module.exports = class extends needleClientBase {
                 : `${appName}${entityAngularName}Module`;
 
             const splicable = isEntityAlreadyGenerated
-                ? `|,{
+                ? `|,
+                        |{
                         |                path: '${entityUrl}',
                         |                loadChildren: '${modulePath}#${moduleName}'
                         |            }`
                 : `|{
-                            |                path: '${entityUrl}',
-                            |                loadChildren: '${modulePath}#${moduleName}'
-                            |            }`;
+                        |                path: '${entityUrl}',
+                        |                loadChildren: '${modulePath}#${moduleName}'
+                        |            }`;
             const rewriteFileModel = this.generateFileModel(
                 entityModulePath,
                 'jhipster-needle-add-entity-route',
-                this.generator.stripMargin(splicable)
+                ...this.generator.stripMargin(splicable).split(/\n/)
             );
 
             this.addBlockContentToFile(rewriteFileModel, errorMessage);
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -90,6 +90,25 @@ function escapeRegExp(str) {
 }
 
 /**
+ * Escape leading comma in case the splicable is like a JSON array entry. This
+ * will allow it to match the first element of the list, in case that element is
+ * the same.
+ *
+ * @param {string} line the parsed line
+ * @param {number} idx the current line index
+ * @returns {string} a regex representing an optional comma if needed, or the
+ * common escaped regexp
+ */
+function escapeJsonArray(line, idx) {
+    if (idx === 0 && line === ',') {
+        // ignore the leading comma in order to allow the first element
+        // to match if the output is like a JSON array
+        return ',?';
+    }
+    return `\\s*${escapeRegExp(line)}`;
+}
+
+/**
  * Rewrite using the passed argument object.
  *
  * @param {object} args arguments object (containing splicable, haystack, needle properties) to be used
@@ -97,7 +116,7 @@ function escapeRegExp(str) {
  */
 function rewrite(args) {
     // check if splicable is already in the body text
-    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line)}`).join('\n'));
+    const re = new RegExp(args.splicable.map((line, idx) => escapeJsonArray(line, idx)).join('\n'));
 
     if (re.test(args.haystack)) {
         return args.haystack;


afillatre:
Since no tests were already present for this part, and my javascript skills are very low, I did not add a test on my own for this MR.

afillatre:
Hi @deepu105,

The main problem was not the prettier formatting (although the comma and the curly brace were on the same line, which would not reflect the generated file format), but the fact that the "splicable" were not splitted into multi-lines. Thus the `escapeRegExp()` function could not add the leading `\s*` in front of each line, and the regexp failed to match existing content (number of leading spaces are not the same in the template and the generated file, which seems pretty normal to me).

I'm not sure having the same routes multiple times in the `entity.module.ts` breaks anything, but it's cleaner to prevent it anyways.

avdev4j:
Hi guys,
I'm following this issue because I think this is related to the last changes on needle API.
But it is not, because I made some tests and I can reproduce it in 5.8.0 (where needle api was not merged).

In 5.7.2. it's working as expected.
After checking git histories I can't find where it come from.

Maybe checking 5.7.2/5.8.0 differences can help you.

avdev4j:
I have were it come.
In 5.8.0 we merged the Angular Lazy Loading (#8925) from @wmarques which change how we deal with adding import in Entity module.

The problem is that if a import is already here we need to add a comma with this splicable value `|,{
                        |                path: '${entityUrl}',
                        |                loadChildren: '${modulePath}#${moduleName}'
                        |            }`
But with prettier the comma is move to the previous import like that : 
`            {
                path: 'departement',
                loadChildren: './departement/departement.module#OhmymagDepartementModule'
            },
            {
                path: 'magasin',
                loadChildren: './magasin/magasin.module#OhmymagMagasinModule'
            }
`
When JHipster looking for something like `,{
                path: 'magasin',
                loadChildren: './magasin/magasin.module#OhmymagMagasinModule'
            }` it obviously can't find it.

Even if we can handle this correctly there is a second problem. We need to know if the import is at the first place or not (not only if there is already import existing in entity module).

I tried many solutions but I can't find which can satisfied all prerequisites.

afillatre:
@avdev4j Have you tried my patch ? I came to the same conclusions and fixed it. Please see the commit message for more details.

avdev4j:
To be honest, no I did not try but np I trust you ;).
I've just reading the commit message ;).

like @deepu105 I wonder what consequences will be when we modify a generic method called everywhere.

Probably that last modifications made on Entity module has rise up a disfunction in base methods. That's could explain why we did not see it before.

Otherwise I found a hack (less prettier as yours) but it seems to work (need more tests), I check if the import does not really exist in file with `path: '${entityUrl}'` as key.

```
            const appName = this.generator.getAngularXAppName();
            const hasEntityImported = jhipsterUtils.checkStringInFile(entityModulePath, 'loadChildren', this.generator);
            const isEntityAlreadyImported = jhipsterUtils.checkStringInFile(entityModulePath, `path: '${entityUrl}'`, this.generator)
            const modulePath = `./${entityFolderName}/${entityFileName}.module`;
            const moduleName = microServiceName
                ? `${this.generator.upperFirstCamelCase(microServiceName)}${entityAngularName}Module`
                : `${appName}${entityAngularName}Module`;
            let splicable = hasEntityImported ? '|,' : '|';
            splicable += `{
                        |                path: '${entityUrl}',
                        |                loadChildren: '${modulePath}#${moduleName}'
                        |            }`;

            if (!isEntityAlreadyImported) {
                const rewriteFileModel = this.generateFileModel(
                    entityModulePath,
                    'jhipster-needle-add-entity-route',
                    this.generator.stripMargin(splicable)
                );

                this.addBlockContentToFile(rewriteFileModel, errorMessage);
```


regards

afillatre:
Hi @wmarques,

This is not actually the same thing. With `const isEntityAlreadyGenerated = jhipsterUtils.checkStringInFile(entityModulePath, 'loadChildren', this);` you're looking for **any** existing `loadChildren` element in the file (thus whether to add a leading comma in the generated section), whereas I'm looking for the exact same element.

The code you're talking about is working as excepted, from what I saw.

@avdev4j yes you're right, I could have looked for both the keys and values in the file but : 
- I did not want to change the way it was designed in the first place
- I came to the conclusion there were actual bugs in both parts (indent and split problem in the needle-client, and json array problem in the util part), so I choose to fix them instead. I'll try to do more testing, but it'll be manual because JS is really not my strong suit and I won't be able to do something good enough in a reasonable amount of time for now.

avdev4j:
I'm currently working on adding unit tests on needle API but it's not easy than I expected.
For the moment manual tests are required and maybe reviews from people who master RegExp (yea I'm so bad at that...).

We also need a green build but the error sounds strange for me... 

> TypeError: Cannot read property 'minify' of undefined
> terser-webpack-plugin

Have you ever meet this error @wmarques ?

deepu105:
William, as I have previously suggested, lets use a simple approach. Add
the needle at the beginning, if there are other routes, so that a comma at
the end will always be present else add a dummy comment node after the
comma, like how I was doing earlier. No point in trying to make it perfect.
Also use prettier ignore if needed

deepu105:
I for one am not comfortable with changes to public API stuff a lot

afillatre:
So, here is the test I made (I'll give you the workflow, then go into details afterwards) : 
1. Create a vanilla JHipster project
1. Create a simple JDL model
1. Generate it using the `import-jdl`command from the master branch
1. Commit the changes
1. Delete the `.jhipster/*` (generated JSON entities) to allow a regeneration of the same model
1. Regenerate the same model from the master branch
1. Look at the modified files
1. Revert
1. Delete the `.jhipster/*` (generated JSON entities) to allow a regeneration of the same model
1. Regenerate the same model from _the fix branch_
1. Look at the modified files

After the first regeneration, the modified files are the following : 
- the JSON entities in the .jhpister folder => not relevant
- liquidbase files (master.xml, new scripts, etc.) => not relevant
- entity.module.ts => with the bug described in the ticket
- some code reformating in the *spec.ts files => don't know why, though

After the second regeneration (on the fix branch), only the `entity.module.ts` file is absent from the previous list.

This is far from a thorough testing, but the basic test case works fine.
Maybe some unit tests on the `utils.rewrite()` function would be relevant ?

wmarques:
@avdev4j For terser issue: https://github.com/webpack-contrib/terser-webpack-plugin/issues/66

wmarques:
Agree with Deepu's solution, let's keep the needle on top of the generated entities routes, so we always add a trailing comma, this should solve the problem

wmarques:
@deepu105 just checked, I don't think it's really easy to have the needle at the beginning of an expression, currently this is handled by the `rewriteFile` function which add the needle at the end if I understood it correctly and I don't want to break it... 
So I would prefer the @avdev4j solution, if that's ok for you.

deepu105:
@wmarques I don't get your point, the `rewriteFile` function just adds the provided content before the needle, so if you move the needle to the beginning it will just add there, I dont see any problem there unless i'm missing something.

for example 
```
const entities = [
  // needle-here
  FooEntity,
  BarEntity
]
```

should produce below when adding say `FooBarEntity`
```
const entities = [
  FooBarEntity,
  // needle-here
  FooEntity,
  BarEntity
]
```

The only problem is if we start with an empty array or something

wmarques:
@deepu105 I found a solution with @pascalgrimaud: We can add `// prettier-ignore` on top of the class, and then always add the entity route with a trailing comma. With this it should be ok.

Besides this, we also need to add a check to verify that the entity isn't already present in the file, maybe by checking the `loadChildren` value as @avdev4j suggested.

afillatre:
I'm not sure why you don't like the way I fixed it, but in all the solutions you propose you have a corner case you do not handle (for @deepu105 it's the empty array, for @wmarques it's the trailing comma).

Also, like I pointed out earlier, the real problem here is actually not the needle part (because it works for the most part), but the regexp that checks for the existing entry. With all the solutions you propose, you kind of miss the point. The regexp fails because of the comma (leading or trailing), and there is still the corner case where the expression we want to match is the first one (i.e. without any comma).

Here is a more detailed example, with the actual generator code :

Initial state
```
const entities = [
  // needle-here
  BarEntity
]
``` 
Add FooEntity (works ok, no matching)
```
const entities = [
  // needle-here
  FooEntity,
  BarEntity
]
``` 
Add another FooEntity (works ok, "FooEntity," matches the existing "FooEntity,")
```
const entities = [
  // needle-here
  FooEntity,
  BarEntity
]
``` 
Add another BarEntity (does not work, "BarEntity," does not match the existing "BarEntity", because the comma is missing)
```
const entities = [
  // needle-here
  BarEntity,
  FooEntity,
  BarEntity
]
``` 

I fixed that using the "?" char in regexp, to make the leading comma optional (and only if it's the leading comma). This will work if you append JSON content using the same method everywhere (and even so, it could be upgraded to make it work with more cases).

For the updates I did in the needle part, it's just to ensure the splicable part is split in lines for the regex to actually work.

I hope it's clearer.

deepu105:
I'm ok with making regex better. My only suggestion was not to overcomplicate stuff trying to perfect this

wmarques:
Agree with both of you guys.

@afillatre I just tested your branch, imported a JDL and re-imported it, didn't have any duplication so it seems to work good, i'm ok for merging this.

@deepu105 does that look correct for you ? For me it's quite a simple change and cover an edge case

deepu105:
Actually I'm not ok with this specific implementation as it modifies the
base rewrite method to address one specific case. If there is indeed an
issue it should be handled only on that specific parent method. I'm not
comfortable especially since we don't have unit tests for this part and we
never had issue here before so I would still suggest to find a fix on the
specific method in question rather than on the base rewrite method. Else
add test cases here to ensure this doesn't introduce any regression.

William, were you able to reproduce the actual issue mentioned? Coz I have
never seen the issue before

deepu105:
Ignore my comment about reproduction, I forgot that it was introduced recently. Honestly I'm ok with doing prettier-ignore with trailing comma. Its not a big deal. Else please add tests to ensure there is no regression here. We broke some public API already since we didn't pay close attention so I don't want the same to happen again and I'm really tight on time to personally test everything (I guess its the same for others as well) so its a good opportunity to add tests. May be someone on the thread can help to add tests

avdev4j:
@deepu105 @wmarques 
I've started unit tests for the whole needle API here #9254
and the angular part here #9151

The angular part only contains the css tests, I've plan to add more tests but I want to avoid doing this in many PR because of conflict issues.

The more I don't deal with conflict the better I am :D 

So I let you decide how we can do to add tests, waiting for other PR to be merged ? add tests and manage conflicts ? other solution ?

btw I've tried to add a unit test (not comited yet) which test if 'path: 'myEntity' appears twice when we call the `addEntityToModule` twice. But don't know how to deal with the cas "assert that text is present only one time". Probably with a regex like this

`                assert.fileContent(`${CLIENT_MAIN_SRC_DIR}app/entities/entity.module.ts`, /(path: 'entityUrl'){2}/);`

I'm not confortable with regexp so it does not work, ofc.... :')

wmarques:
@deepu105 I understand your concern. However, the fix that @afillatre provided isn't about the trailing comma, it's about the regexp check that verifies if the content we want to add isn't already present in the file. 

But yes, this is corrected by using prettier-ignore as there will always be a trailing comma so let's do this. I will provide a PR with those changes.

afillatre:
@wmarques what about the first entry ?
This entity.module.ts file starts empty, so you’ll have the problem the first time. And you can have the problem for any new json array-like file.

For this specific entity.module.ts file, duplicate entries do not seem to be a problem. But this may not always be the case.

I’d rather vote for the testing part instead. I’m not confortable writing the tests in js but I can provide the way to do them.

avdev4j:
I've start writing a test case like this : 
- creating a blueprint on the fly that call `addEntityToModule` twice

```
const mockBlueprintSubGen = class extends ClientGenerator {
    constructor(args, opts) {
        super(args, { fromBlueprint: true, ...opts }); // fromBlueprint variable is important
        const jhContext = (this.jhipsterContext = this.options.jhipsterContext);
        if (!jhContext) {
            this.error("This is a JHipster blueprint and should be used only like 'jhipster --blueprint myblueprint')}");
        }
        this.configOptions = jhContext.configOptions || {};
        // This sets up options for this sub generator and is being reused from JHipster
        jhContext.setupClientOptions(this, jhContext);
    }

    get initializing() {
        return super._initializing();
    }

    get prompting() {
        return super._prompting();
    }

    get configuring() {
        return super._configuring();
    }

    get default() {
        return super._default();
    }

    get writing() {
        const phaseFromJHipster = super._writing();
        const customPhaseSteps = {
            addEntityToModuleStep() {
                this.addEntityToModule(
                    'entityInstance',
                    'entityClass',
                    'entityName',
                    'entityFolderName',
                    'entityFileName',
                    'entityUrl',
                    'angularX',
                    'microServiceName'
                );
                this.addEntityToModule(
                    'entityInstance',
                    'entityClass',
                    'entityName',
                    'entityFolderName',
                    'entityFileName',
                    'entityUrl',
                    'angularX',
                    'microServiceName'
                );
            }
        };
        return { ...phaseFromJHipster, ...customPhaseSteps };
    }

    get install() {
        return super._install();
    }

    get end() {
        return super._end();
    }
};
```

- assert that the entity path is present only one time

```
            it('assert entity is correctly added in import entity module', () => {
                assert.fileContent(`${CLIENT_MAIN_SRC_DIR}app/entities/entity.module.ts`, /path: 'entityUrl',/);
            });

            it('assert entity is correctly added in import entity module', () => {
                assert.fileContent(`${CLIENT_MAIN_SRC_DIR}app/entities/entity.module.ts`, /(path: 'entityUrl'){2}/);
            });
```

But ` /(path: 'entityUrl'){2}/` does not working to test is the file contain the string twice. I need to find a way to test it properly.

afillatre:
@avdev4j If you can give me the url of your branch, I may take a look at it this week-end. Also, regex101 is a wonderful website to test and understand regexp, if that helps you.

avdev4j:
well I did not push it into a branch, I started angular client needle into this one : https://github.com/avdev4j/generator-jhipster/tree/sync_react_angular_css_api

And made other tests here : 
https://github.com/avdev4j/generator-jhipster/tree/unit_tests_needle_api

wmarques:
@afillatre I'm trying to understand what's happening:
* When I create a jhipster app using `import-jdl`, I reproduce your bug because my array is empty
* When I create a jhipster app and then generate an entity using `jhipster entity` it works, even with an empty array...

pascalgrimaud:
I'm testing this PR and I can still reproduce the issue.
@wmarques @avdev4j : can one of you have a look and try to test/complete this PR plz ?

pascalgrimaud:
closing this, in favor of https://github.com/jhipster/generator-jhipster/pull/9802

PR #9186, opened 2019-02-04
Fix call to the wrong stripMargin method

avdev4j:
In the angular needle Api we are calling a `this.stripMargin()` method that does not exist. 'stripMargin()' belongs to generator-base.js so I changed it to `this.generator.stripMargin()`.


murdos:
I think you need to create a PR against the 5.x maintenance branch for this fix, and you'll probably need to rebase your feature branch on it.

PR #9188, opened 2019-02-04
Fix call to the wrong stripMargin method

avdev4j:
In the angular needle Api we are calling a this.stripMargin() method that does not exist. 'stripMargin()' belongs to generator-base.js so I changed it to this.generator.stripMargin().

@murdos 
I don't use to handle maintenance branch with JHipster. I open a new PR and close the other one. I hope this is what you wanted me to do ;).

### DIFF
--- a/generators/internal/needle-api/needle-client-angular.js
+++ b/generators/internal/needle-api/needle-client-angular.js
@@ -111,7 +111,7 @@ module.exports = class extends needleClientBase {
     _generateRewriteFileModelWithImportStatement(appName, angularName, folderName, fileName, modulePath, needle) {
         const importStatement = this._generateImportStatement(appName, angularName, folderName, fileName);
 
-        return this.generateFileModel(modulePath, needle, this.stripMargin(importStatement));
+        return this.generateFileModel(modulePath, needle, this.generator.stripMargin(importStatement));
     }
 
     _generateImportStatement(appName, angularName, folderName, fileName) {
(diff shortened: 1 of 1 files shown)


murdos:
@avdev4j yes, that's what I was suggesting :)

avdev4j:
Nice !
There is any documentation about how you deal with JHipster maintenance branch ? 
We have just merged this PR into a maintenance but how people who are using JHipster 5X would benefit from it ?

I would probably do the same for #9151 but I'm not sure. Does every commit to v5 maintenance will be bring to 6.0.0 ?

murdos:
> 
> 
> Nice !
> There is any documentation about how you deal with JHipster maintenance branch ?

It's the first time I encounter this situation with JHipster (I was not there during v5 release) so I'm just using common sense and what I'm used to in my job.

>
> We have just merged this PR into a maintenance but how people who are using JHipster 5X would benefit from it ?

Fixes on the maintenance branch will be manually merged to master, regularly.
People using JHipster 5X will need to wait for a new patch release to benefit from it.

> 
> I would probably do the same for #9151 but I'm not sure. 

Yes, that's probably what we should do.

>
> Does every commit to v5 maintenance will be bring to 6.0.0 ?

If they apply and make sense (should be true most of the time), they will.

avdev4j:
Thanks you so much, it makes sens for me now.
I don't know why but I imagined the process in the other way. First merging in master and then merge to maintenance branch.

But anyway I'll open a new PR for css needles to v5 maintenance. 

👍

Issue #9194, opened 2019-02-05
PaginationUtil  loses the sort order and other query parameters such as criterias.


SergeyMitrofanov:
PaginationUtils insert in link header page, size and query(for ElasticSearch) parameters only. Othes query params loses.

Example query URI:
```
/uaa/api/users?login.contains=adm&sort=lastName%2Cdesc
```
Current response link header is:
```
</api/users?page=0&size=20>; rel="last",</api/users?page=0&size=20>; rel="first"
```

For useful expected:
```
</api/users?page=0&size=20&login.contains=adm&sort=lastName%2Cdesc>; rel="last",</api/users?page=0&size=20&login.contains=adm&sort=lastName%2Cdesc>; rel="first"
```

PR #9205, opened 2019-02-06
#9194 Fix paging link headers for use all query parameters.

SergeyMitrofanov:
Fix https://github.com/jhipster/generator-jhipster/issues/9194

1. Created universal single method PaginationUtil.generatePaginationHttpHeaders(HttpServletRequest request, Page<T> page). HttpServletRequest need for get URI and query params.
2. Tests adapted for new method calls. Assets not changed.
3. New test for sort and criteria params
4. Change templates for new call
5. Previous methods marks as deprecated.
  

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
@@ -23,14 +23,14 @@
     const entityListToDtoListReference = mapper + '.' + 'toDto';
     const entityToDtoReference = mapper + '::'+ 'toDto';
     if (jpaMetamodelFiltering) {  %>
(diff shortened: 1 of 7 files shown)


jdubois:
Could you explain why this would be better than the current method?

BTW I would like to migrate this class to jhipster/jhipster-framework, so if we do this we'll need to refactor this code. It shouldn't be difficult, just don't be surprised if this happens.

DanielFran:
Closing and reopening to launch tests

jdubois:
Thanks @SergeyMitrofanov I understand your issue now, but as commented above I have moved the PaginationUtil to our server-side library at https://github.com/jhipster/generator-jhipster/issues/9306

I'm having a look at whether I can refactor your code or not.

jdubois:
Merged! Thanks a lot @SergeyMitrofanov !!

cbornet:
Arf ! We don't have HttpServletRequest in Webflux so this is broken for the reactive option :cry:

jdubois:
Oh indeed @cbornet ! But does pagination makes sense in reactive? Wouldn't you just stream the data?

PR #9207, opened 2019-02-06
use underscore syntax to access command

cvk77:
Use underscore syntax to run `npm install` and `yarn install` as documented in https://github.com/srs/gradle-node-plugin/blob/master/docs/node.md.


pascalgrimaud:
Thanks and congrats for your 1st contribution to the project @cvk77 !

PR #9214, opened 2019-02-07
[v5 maintenance] Fix react angular css api and needles

avdev4j:
As discussed with @deepu105 (and @murdos) I'm opening a v5 maintenance PR to revert the method signatures with their original values and fix needles values.

same as #9151


deepu105:
@murdos can you test and merge this please

murdos:
I don't really know how to test that.

avdev4j:
You can create a blueprint client, with a « modify file » phase which call those methods.

Then add unit tests to check the css content with assert.fileContent

I could do this and share the project on github if you want.

ruddell:
I'm assuming this PR is related to the [breaking changes issue](https://groups.google.com/forum/#!topic/jhipster-dev/IImgpGcgwCU).  That was first mentioned in `jhipster-primeng` here: https://github.com/sudheerj/generator-jhipster-primeng/issues/67

Even with this PR, I still get the issue from there:
```
TypeError: Cannot read property 'clientWebpack' of undefined
    at constructor.copyExternalAssetsInWebpack (/Users/ruddell/Contrib/generator-jhipster/generators/generator-base.js:661:24)
    at constructor.writing (/private/tmp/jh/generator-jhipster-primeng/generators/app/index.js:224:18)
```
Which is referring to `this.copyExternalAssetsInWebpack('primeng', 'primeng');` . Since the Needle API addition, this method now looks like:
```
    copyExternalAssetsInWebpack(sourceFolder, targetFolder) {
        this.needleApi.clientWebpack.copyExternalAssets(sourceFolder, targetFolder);
    }
```

avdev4j:
hi @ruddell 
I'm surprized because it seems to be reproductible in 5.8.0 and the needle API was added in the 5.8.1.
Also I wonder how the 'needleApi' variable could be undefined when it is use in the whole classe.

This PR deal with css API not clientWebpack. So the bug should not be fix with this PR.

ruddell:
The issue I referenced is only reproducible on v5.8.1+.  v5.8.0 and below works fine.  I'm pretty sure the people in the issue who reported it with v5.8.0 were changing their global JHipster installation, not the version used by the module ([which is installed separately](https://github.com/sudheerj/generator-jhipster-primeng/blob/master/package.json#L30)).  If you need steps to reproduce let me know.

Anyway, I guess that's a separate issue but is a real example of breaking backwards compatability.

avdev4j:
I have just do some checks and indeed I think 5.8.0 issues should can from users installations.

I've add a response to the issue and I think it from the way we initialise the needleApi variable.

OFC I will bring some help to fix this issue. I need to switch in debug mode and add breakpoints into baseGenerator constructor to be sure primeNG call it.

avdev4j:
I figured out that we can't trust `this.clientFramework` to switch between angular and react before calling the needle API. 
In fact when testing primeNG bug I find that this can be undefined. To avoid bugs I removed this test and so the react part.

We don't have a needle api to add css in react css files just like before. I will open a issue to not forget to do this in the v6.

avdev4j:
@murdos 
I finally add unit tests like I would do with a blueprint. It could help you.

avdev4j:
I have added react unit tests

avdev4j:
I'm closing this one because the branch related to was orignally created from master.
For v5 I prefer to create branch from v5 maintenance and doing cherry-pick.

I'm opening a new one from v5 maintenance branch \o/

PR #9224, opened 2019-02-08
angular: change tabWidth for angular

deepu105:
Based on discussion https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jhipster-dev/s-BtFM-PSyg/AxaoR9pxFgAJ

### DIFF
--- a/generators/client/templates/react/.editorconfig
+++ b/generators/client/templates/react/.editorconfig
@@ -6,19 +6,18 @@ root = true
 
 [*]
 
-# Change these settings to your own preference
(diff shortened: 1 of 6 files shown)


murdos:
I'm wondering: shouldn't we have only one .editorconfig template? We have 3 ones: in `common`, in `client/react` and one in `entity-client/react`

deepu105:
@murdos some of them are not copied to the generated project, they are used to work on the templates itself

PR #9230, opened 2019-02-09
refactor: use strong instead of b tag

clakech:
Follow dev best practices, see sonar analysis message:

The strong/b and em/i tags have exactly the same effect in most web browsers, but there is a fundamental difference between them: strong and em have a semantic meaning whereas b and i only convey styling information like CSS.

While b can have simply no effect on a some devices with limited display or when a screen reader software is used by a blind person, strong will:

Display the text bold in normal browsers
Speak with lower tone when using a screen reader such as Jaws
Consequently:

in order to convey semantics, the <b> and <i> tags shall never be used,
in order to convey styling information, the <b> and <i> should be avoided and CSS should be used instead.
Noncompliant Code Example
```
<i>car</i>             <!-- Noncompliant -->
<b>train</b>         <!-- Noncompliant -->
```
Compliant Solution
```
<em>car</em>
<strong>train</strong>
```

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/account/password/password.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/password/password.component.html.ejs
@@ -19,7 +19,7 @@
 <div>
     <div class="row justify-content-center">
         <div class="col-md-8">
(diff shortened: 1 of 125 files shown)


clakech:
ouch !

clakech:
here we go @deepu105 ;)

PR #9233, opened 2019-02-10
[v5 maintenance] Fix needleApi injection, external resources public API and add unit tests

avdev4j:
According to https://github.com/sudheerj/generator-jhipster-primeng/issues/67 the needle API PR broke compatibility. I would like to handle needleApi construction in the generator-base constructor but it seems not working since module not always extend directly the generator-base.

To fix that I suggest to add a needleApi getter which initialise the variable if needed.

Also when during my tests I found a bug  in the addExternalResourcesToRoot and I fix that too.

@sudheerj
Could you please test this PR for me and tell me if you always have the problem ?
If it's ok for you I will remove the WIP tag.
thanks you

Regards,

PS: need to be merged to master to

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -48,11 +48,6 @@ const SERVER_MAIN_RES_DIR = constants.SERVER_MAIN_RES_DIR;
  * The method signatures in public API should not be changed without a major version change
  */
 module.exports = class extends PrivateBase {
-    constructor(args, opts) {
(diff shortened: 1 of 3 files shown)


avdev4j:
add client needle api unit test

jdubois:
From my understanding this is very important for @sudheerj -> once this is merged don't hesitate to ping me so I do a maintenance release of the 5.x branch

Issue #8250, opened 2018-09-07
Application fails to render start page for a fresh React project on IE11


hirro:
##### **Overview of the issue**

I had some problems with my application not rendering at all in IE11. 
After some troubleshooting I generated a blank React project in jHipster but I still get this problem.
The console error in IE reads:

> SCRIPT5009: 'Symbol' is undefined
  vendors.chunk.js (84301,60)

**vendors.chunk.js**
> 	84285: /***/ "./node_modules/react-jhipster/lib/src/language/translate.js":
> 	84286: /*!*******************************************************************!*\
> 	84287:   !*** ./node_modules/react-jhipster/lib/src/language/translate.js ***!
> 	84288:   \*******************************************************************/
> 	84289: /*! no static exports found */
> 	84290: /***/ (function(module, exports, __webpack_require__) {
> 	84291: 
> 	84292: "use strict";
> 	84293: 
> 	84294: var __extends = (this && this.__extends) || (function () {
> 	84295:     var extendStatics = Object.setPrototypeOf ||
> 	84296:         ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
> 	84297:         function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
> 	84298:     return function (d, b) {
> 	84299:         extendStatics(d, b);
> 	84300:         function __() { this.constructor = d; }
> 	84301:         d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
> 	84302:     };
> 	84303: })();
> 	84304: Object.defineProperty(exports, "__esModule", { value: true });
> 	84305: var React = __webpack_require__(/*! react */ "./node_modules/react/index.js");
> 	84306: var get = __webpack_require__(/*! lodash.get */ "./node_modules/lodash.get/index.js");
> 	84307: var sanitizeHtml = __webpack_require__(/*! sanitize-html */ "./node_modules/sanitize-html/dist/index.js");
> 	84308: var translator_context_1 = __webpack_require__(/*! ./translator-context */ "./node_modules/react-jhipster/lib/src/language/translator-context.js");
> 	84309: var REACT_ELEMENT = Symbol.for('react.element');

##### **Motivation for or Use Case**

According to the jHipster page IE11 is supported.

##### **Reproduce the error**

Generate project with settings in the provided .yo-rc file.
[...]

PR #9253, opened 2019-02-12
Remove IE polyfills

wmarques:
We should use the `nomodule` attribute to load our polyfills, so the evergreen browsers won't download this javascript file and it will be downloaded by the legacy browsers.

This reduces the payload downloaded by the evergreen browsers which doesn't need those polyfills.

Took the idea from the angular-cli 7.3 which took it from Vue apparently.

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -119,7 +119,6 @@
     "reflect-metadata": "0.1.13",
     "rimraf": "2.6.3",
     "simple-progress-webpack-plugin": "1.1.2",
-    "script-ext-html-webpack-plugin": "2.1.3",
     "style-loader": "0.23.1",
     "terser-webpack-plugin": "1.2.2",
     "thread-loader": "2.1.2",
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -48,6 +48,7 @@
     "ngx-infinite-scroll": "7.0.1",
     "ngx-webstorage": "2.0.1",
     "rxjs": "6.4.0",
+    "script-ext-html-webpack-plugin": "2.1.3",
     "swagger-ui": "2.2.10",
     <%_ if (websocket === 'spring-websocket') { _%>
     "sockjs-client": "1.3.0",
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -48,7 +48,6 @@
     "ngx-infinite-scroll": "7.0.1",
     "ngx-webstorage": "2.0.1",
     "rxjs": "6.4.0",
-    "script-ext-html-webpack-plugin": "2.1.3",
     "swagger-ui": "2.2.10",
     <%_ if (websocket === 'spring-websocket') { _%>
     "sockjs-client": "1.3.0",
@@ -120,6 +119,7 @@
     "reflect-metadata": "0.1.13",
     "rimraf": "2.6.3",
     "simple-progress-webpack-plugin": "1.1.2",
+    "script-ext-html-webpack-plugin": "2.1.3",
     "style-loader": "0.23.1",
     "terser-webpack-plugin": "1.2.2",
     "thread-loader": "2.1.2",
--- a/generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
@@ -16,10 +16,6 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import 'core-js/es6/reflect';
-import 'core-js/es7/reflect';
-import 'zone.js/dist/zone';
-require('../manifest.webapp');
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { ProdConfig } from './blocks/config/prod.config';
 import { <%=angularXAppName%>AppModule } from './app.module';
--- a/generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
@@ -16,6 +16,10 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+import 'core-js/es6/reflect';
+import 'core-js/es7/reflect';
+import 'zone.js/dist/zone';
+require('../manifest.webapp');
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { ProdConfig } from './blocks/config/prod.config';
 import { <%=angularXAppName%>AppModule } from './app.module';
--- a/generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
@@ -1,42 +1,3 @@
-<%#
- Copyright 2013-2019 the original author or authors from the JHipster project.
-
- This file is part of the JHipster project, see https://www.jhipster.tech/
- for more information.
-
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
-
-      http://www.apache.org/licenses/LICENSE-2.0
-
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- See the License for the specific language governing permissions and
- limitations under the License.
--%>
-/**
- * This file includes polyfills needed by Angular and is loaded before the app.
- * You can add your own extra polyfills to this file.
- *
- * This file is divided into 2 sections:
- *   1. Browser polyfills. These are applied before loading ZoneJS and are sorted by browsers.
- *   2. Application imports. Files imported after ZoneJS that should be loaded before your main
- *      file.
- *
- * The current setup is for so-called "evergreen" browsers; the last versions of browsers that
- * automatically update themselves. This includes Safari >= 10, Chrome >= 55 (including Opera),
- * Edge >= 13 on the desktop, and iOS 10 and Chrome on mobile.
- *
- * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
- */
-
-/***************************************************************************************************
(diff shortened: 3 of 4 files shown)


generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
```diff
@@ -1,42 +1,3 @@
-<%#
```
DanielFran: @wmarques Did you remove by mistake the copyright

generators/client/templates/angular/package.json.ejs
```diff
@@ -48,6 +48,7 @@
     "ngx-infinite-scroll": "7.0.1",
     "ngx-webstorage": "2.0.1",
     "rxjs": "6.4.0",
+    "script-ext-html-webpack-plugin": "2.1.3",
```
vishal423: shouldn't this be dev dependency ?

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
```diff
@@ -16,6 +16,10 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+import 'core-js/es6/reflect';
+import 'core-js/es7/reflect';
+import 'zone.js/dist/zone';
```
vishal423: shouldn't `zone` and below manifest import still be part of polyfills?

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
```diff
@@ -1,42 +1,3 @@
...
- *
- * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
- */
-
-/***************************************************************************************************
- * BROWSER POLYFILLS
- */
-
-/** IE9, IE10 and IE11 requires all of the following polyfills. **/
 import 'core-js/es6/symbol';
```
vishal423: recent release of `angular-cli` has removed all these pollyfills as these are now supported out of box by evergeeen browsers. Do we really require all these?

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
wmarques: No, they are required for Angular in JIT mode for the `reflect` imports, and Zone is always required.
Otherwise you will get a very strange `cannot find Parameter for ApplicationModule(?)` 😄

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
wmarques: That's the point of the PR, we load them only if necessary, so the evergreen browsers won't fetch this bundle, angular CLI does the same

vishal423: I thought we only support evergreen browsers and un-necessarily increases bundle size. This would also confuse developers on where to add new pollyfills that are also required in evergreen browsers.

As per my understanding, CLI doesn't add any support out of box and you need to manually add appropriate pollyfills to support old browsers. Do you have any code reference that points to this support by CLI?

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
vishal423: I am bit confused on mixing pollyfills with application code. If some polyfills are only required in JIT mode like `reflect`, then, can we separate those out and exclude from AOT builds or would AOT compiler be excluding those by default?

Thanks for reminding, I know angular is leveraging `zone` as base construct to support asynchronous events

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
wmarques: oh yes, thanks ! 😅

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
wmarques: I don't know what we support officially, so I just found a solution to keep supporting older browsers without impacting evergreen browsers... Maybe we should discuss about the browser support, it can be better to simply remove this...

Here is the link for the angular CLI: https://blog.ninja-squad.com/2019/01/31/angular-cli-7.3/

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
wmarques: Yes maybe we can separate this into a `required-polyfills` file ?
The AOT won't exclude those files automatically

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
vishal423: thanks for reference. Few points:
- I could see discussion on https://github.com/jhipster/generator-jhipster/issues/8250 related to browser support. May be v6 release is a good time to break officially.
- If we want to support, then, would prefer the angular cli approach to have specific polyfills containing only these and let main polyfills be containing others

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
vishal423: I like the angular CLI approach and naming convention to move those into specific `es2015-polyfills.ts` and let the main be `polyfills.ts`

`JIT` mode specific polyfills are moved to separate `es2015-jit-polyfills.ts` and included only in JIT compilation (to save some kbs)

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
wmarques: I guess it's better to break support, I prefer keeping generated code simple, if people want to add IE support, they can easily find out how to do it, we can even add a tip on the website to do this

wmarques:
I'll update this PR to simply remove the polyfills, this will simplify the generated code, I can add a tip on the jhipster website for the users who wants to support legacy browsers (mainly IE).
@jhipster/developers if you are against this please let me know :)

PierreBesson:
Apparently IE is not a browser anymore according to Microsoft so +1 https://www.zdnet.com/article/microsoft-security-chief-ie-is-not-a-browser-so-stop-using-it-as-your-default/

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
wmarques: That's good idea but I find it a little overkill just for to save two files which are not huge.. I prefer keeping it simple, even if we bundle them in AOT mode which is useless.

I think we will pretty soon use AOT even for dev so we will remove those polyfills once we've done that :)

wmarques:
It's done if someone can check this quickly so we can merge it, thanks ! 😄

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
```diff
@@ -16,73 +16,7 @@
...
-import 'core-js/es6/regexp';
-import 'core-js/es6/map';
-import 'core-js/es6/weak-map';
-import 'core-js/es6/set';
-
-/** IE10 and IE11 requires the following for NgClass support on SVG elements */
-// import 'classlist.js';  // Run `npm install --save classlist.js`.
-
-/** Evergreen browsers require these. **/
 import 'core-js/es6/reflect';
```
vishal423: I believe this can be removed as well

generators/client/templates/angular/src/main/webapp/app/app.main.ts.ejs
vishal423: agree, effort may not be worth to save few kbs

vishal423:
Separate module can be an alternative to hook these into applications.. I will check if can do in coming days

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
wmarques: No I think we still need the reflect files for the JIT mode

wmarques:
@vishal423 what do you mean by module ?

generators/client/templates/angular/src/main/webapp/app/polyfills.ts.ejs
vishal423: I think only `es7/reflect` is required in JIT mode. Can you please double check?

vishal423:
@wmarques, separate module to support legacy browsers using conditional polyfills

wmarques:
@jhipster/developers ready for review there :)

wmarques:
@vishal423 you were right about es6/reflect :)

fdelbrayelle:
@wmarques Hi, unfortunately we have to manage IE11 has a target. We tried to generate a brand new monolithic application with JHipster 5.8.2 and it works with the default polyfills.ts in IE11. We tried to generate as well in JHipster 6.0.2 with the same polyfills.ts.
We tried to check the diff between the two generated versions and cannot understand what is behind this. Could you help us please? You wrote about a tip on the site above.
Thank you by advance.

fdelbrayelle:
@wmarques Hi, unfortunately we have to manage IE11 has a target. We tried to generate a brand new monolithic application with JHipster 5.8.2 and it works with the default polyfills.ts in IE11. We tried to generate as well in JHipster 6.0.2 with the same polyfills.ts contents and it didn't work on IE11.

It doesn't work as well when trying to upgrading from 5.8.2 to 6.0.2 even when keeping the full polyfills.ts contents.

We tried to check the diff between the two generated versions and cannot understand what is behind this. Could you help us please? You wrote about a tip on the site above.
 Thank you by advance.

wmarques:
@fdelbrayelle hi, try to add following files in polyfills.ts:
```
import 'core-js/es6/symbol';
import 'core-js/es6/object';
import 'core-js/es6/function';
import 'core-js/es6/parse-int';
import 'core-js/es6/parse-float';
import 'core-js/es6/number';
import 'core-js/es6/math';
import 'core-js/es6/string';
import 'core-js/es6/date';
import 'core-js/es6/array';
import 'core-js/es6/regexp';
import 'core-js/es6/map';
import 'core-js/es6/weak-map';
import 'core-js/es6/set';
import 'core-js/es6/reflect';
```

This link can also help you: https://angular.io/guide/browser-support

fdelbrayelle:
Thank you very much! Finally, we stayed on 5.8.2 to keep the "target", for the moment.

Issue #9232, opened 2019-02-10
Blueprint issues


pvliss:
##### **Overview of the issue**

So I am trying to add mocha unit tests for the Kotlin blueprint and I have stumbled upon the following problems while working with the `entity` generator:
1. Options are not propagated to the blueprint sub-generator. So for example the test `with angular suffix` of the`entity-spec.js` is failing as the `angular-suffix` option is passed to the blueprint.
2. Defining blueprint as an option to the runner is not being picked up by `entity/index.js`. The workaround for this is to define a `.yo-rc.json` file that also includes the blueprint.

##### **Motivation for or Use Case**
Create unit tests for the Kotlin blueprint with minimal maintenance effort.

##### **Reproduce the error**
Use code from the following branch: https://github.com/pvliss/jhipster-kotlin/tree/mocha-tests
Run the following:
```bash
npm install
npm link generator-jhipster
npm test
```

##### **Suggest a Fix**

1. Pass the options to the blueprint, i.e:
```js
            useBlueprint = this.composeBlueprint(blueprint, 'entity', {
                ...this.options,
                'skip-install': this.options['skip-install'],
                'from-cli': this.options['from-cli'],
                force: this.options.force,
                arguments: [this.context.name]
            });

```
**NOTE:** Some options are obviously duplicated and could be removed
2. The same as with [client generator](https://github.com/jhipster/generator-jhipster/blob/5ff10fc5b7c156cbddaee253c74bd6b96a19a364/generators/client/index.js#L65) which already supports loading the blueprint from options as well. i.e change to:
```js
        const blueprint = this.options.blueprint || this.configOptions.blueprint || this.config.get('blueprint');
```

Final note, is that I think that these should be applied to all generators that support blueprints not just the entity and entity-server.

##### **JHipster Version(s)**
Current master

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
@pvliss : for your information:
- https://github.com/jhipster/generator-jhipster-blueprint/issues/10#issuecomment-431983686
- https://github.com/jhipster/generator-jhipster-blueprint/pull/13

In fact, it seems to be difficult to write `npm test` for entity sub generator.
Our current tests for entities are bad:
- https://github.com/jhipster/generator-jhipster/blob/master/test/entity.spec.js

pvliss:
Thanks for the extra info @pascalgrimaud!!!

[...]

PR #9266, opened 2019-02-14
Blueprint issues

pvliss:
This PR contains the following fixes:
1. Fix to not check for blueprint existence when `skipChecks` is true to allow for running mocha tests in blueprints as well, even if the npm module is no linked 
2. Fix to pass `options` and `configOptions` to blueprint generators and sub-generators
3. Add mocha tests for `spring-controller` and `service-blueprints` generators

I have tested(meaning `npm test`) these changes in the generator and also with the Kotlin and VueJS blueprints

Relates to issue 1 of #9232 

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -208,7 +208,7 @@ module.exports = class extends BaseGenerator {
             },
 
             validateBlueprint() {
-                if (this.blueprint) {
+                if (this.blueprint && !this.skipChecks) {
                     this.checkBlueprint(this.blueprint);
                 }
             },
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -346,35 +346,37 @@ module.exports = class extends BaseGenerator {
 
             composeServer() {
                 if (this.skipServer) return;
+                const options = this.options;
+                const configOptions = this.configOptions;
 
                 this.composeWith(require.resolve('../server'), {
+                    ...options,
+                    configOptions,
                     'client-hook': !this.skipClient,
-                    'from-cli': this.options['from-cli'],
-                    configOptions: this.configOptions,
-                    force: this.options.force,
                     debug: this.isDebugEnabled
                 });
             },
 
             composeClient() {
                 if (this.skipClient) return;
+                const options = this.options;
+                const configOptions = this.configOptions;
 
                 this.composeWith(require.resolve('../client'), {
-                    'skip-install': this.options['skip-install'],
-                    'skip-commit-hook': this.options['skip-commit-hook'],
-                    'from-cli': this.options['from-cli'],
-                    configOptions: this.configOptions,
-                    force: this.options.force,
+                    ...options,
+                    configOptions,
                     debug: this.isDebugEnabled
                 });
             },
 
             composeCommon() {
+                const options = this.options;
+                const configOptions = this.configOptions;
+
                 this.composeWith(require.resolve('../common'), {
+                    ...options,
                     'client-hook': !this.skipClient,
-                    'from-cli': this.options['from-cli'],
-                    configOptions: this.configOptions,
-                    force: this.options.force,
+                    configOptions,
                     debug: this.isDebugEnabled
                 });
             },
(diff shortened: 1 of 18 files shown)


pvliss:
Closing and re-opening to trigger a CI rebuild as I do not think that the e2e failure is related to this PR

murdos:
I agree with 2. and 3. and think this is a nice improvement, but I don't understand why 1. is needed.
Could you explain it? I don't remember having issues with mocha tests when not linking the npm module.

pvliss:
Hello @murdos and thanks for reviewing this.

So when making these changes other than testing them against the Kotlin blueprint I also wanted to test against another blueprint just to be safe. So I decided to use jhipster-vuejs which I know is currently really active :).  So I forked it and then tried to run the mocha tests using `npm test` without linking or having previously installed the module. The problem is that method `checkBlueprint` would try to find the jhipster-vuejs within `node_modules` but since it was not there it was failing even though the module was added as a dependency in the mocha tests. I could have solved this problem by just linking the jhipster-vuejs but at the time I had run too many tests too many times and that was my workaround.

So to sum it up it is really not that important and can be resolved by running `npm link` hence I can revert these changes if you wish.

murdos:
Alright. I'll take a look to vuejs blueprint to see how the tests are configured, and I'll let you know.

pvliss:
Thanks @murdos . Just to note that this is not specific to the VueJS blueprint. The Kotlin blueprint and most probably any blueprint has the same issue

pascalgrimaud:
thanks a lot @pvliss and @murdos
We plan to do a new release of Vue.js blueprint next week, probably monday with @sahbi-ktifa : let us know if you see any issue with blueprint / tests

murdos:
I checked what I did on my own blueprint for work, and the use of `skipChecks` makes sense and is a good idea! Good job :)
I was not convinced at beginning, because when you write unit test for a single subgenerator (that doesn't call other subgen), it's not useful (that's mainly what I did on my own blueprint). But as soon as you run unit tests on the `app` generator, this is very useful.
In fact I used a different technique to do that, but it's very loud and can't be redone on each blueprint: https://github.com/jhipster/generator-jhipster/blob/master/test/blueprint/app-blueprint.spec.js#L51 => could you check if this can be removed with your change (I guess so) and same thing for the entity-blueprint.spec.js ? Thanks

Last question: should we fix that on V5 or on V6 of the generator? If V6 is released in weeks or months, I'd opt for V5.

pvliss:
> I checked what I did on my own blueprint for work, and the use of `skipChecks` makes sense and is a good idea! Good job :)
> I was not convinced at beginning, because when you write unit test for a single subgenerator (that doesn't call other subgen), it's not useful (that's mainly what I did on my own blueprint). But as soon as you run unit tests on the `app` generator, this is very useful.

Thanks. Basically my issue was that the blueprint had to be installed or linked in order to run the unit tests which should not be the case as the `withGenerators` call of `yeoman-test` is what this is all about. Not sure if the optimal solution though.

> In fact I used a different technique to do that, but it's very loud and can't be redone on each blueprint: https://github.com/jhipster/generator-jhipster/blob/master/test/blueprint/app-blueprint.spec.js#L51 => could you check if this can be removed with your change (I guess so) and same thing for the entity-blueprint.spec.js ? Thanks

You mean remove all lines from 51-56? If we do that then the test will fail I think? From what I understand it is just mocking the presence of a blueprint even if it does nothing and simply asserts the existence of a blueprint. Am I missing something? Will test anyways

pvliss:
@murdos You were right. It works!!! But I think that this does not make the test on its own very useful, does it? Maybe I should add few more tests that actually assert something more useful, e.g. creating some files while skipping others, etc. WDYT?

murdos:
This test has one purpose: make sure the blueprint version is saved in .yo-rc.json and that the npm dependency is added to package.json, so in my opinion it is useful as it is.
Then each jhipster generator has its own blueprint test, to make sure it works as a blueprint.

Could you remove all lines from 51-56, do the same in `entity-blueprint.spec.js`, and also remove the `createBlueprintMockForSubgen` function in utils.js? Thanks :)

pvliss:
@murdos Done. Run locally `npm test` and everything passed. Let wait for the CI as well. Thanks again for the review.

Issue #9282, opened 2019-02-17
Failing CI Azure build for react


pvliss:
##### **Overview of the issue**
The `react-noi18n-es-ws-gradle-session` sample is failing at Azure Pipelines

So I spent a few hours today to find the cause of this problem as I feel uncomfortable submitting PRs and see builds failing :)

The problem seems to be caused by commit bda31032 (Migrate PaginationUtil and HeaderUtil to jhipster/jhipster-framework ). Reverting it seems to pass the CI again. The root cause I think is that the change does not take into account the disabled translation feature, hence its the only sample that fails.

You can see that the build is succesfull [here](https://dev.azure.com/pvlissidis/pvlissidis/_build/results?buildId=129) for the [fix-ci-branch](https://github.com/pvliss/generator-jhipster/tree/fix-ci-build)  at my fork where I have done the revert.

I do not want to submit a PR as this should be probably fixed instead of completely reverted.

##### **Motivation for or Use Case**
Fix the CI build

##### **Suggest a Fix**
Other than reverting the whole commit I think the following fixes should help:

1. The default application name of `HeaderUtils` was replaced by `<%= angularAppName %>` while the replacement `${spring.application.name:JHIPSTER}` is replaced by `<%= baseName %>`
2. The new [HeaderUtil](https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/web/util/HeaderUtil.java) class ignores case where translation is disabled.

##### **JHipster Version(s)**
Current master(6.x)

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
I'll have a look.
If your 2 fix are OK, I'll merge them. 
Thanks a lot !

pvliss:
Thanks @pascalgrimaud .

I believe it will work but I would like to note that I am not sure/convinced that this is the best way to go. The reason is that I have added an extra config property `jhipster.clientApp.name` in `application.yml` which might give the false impression to end users that changing it will make things work while this is not the case as the value will be hard-coded at the client code.

pascalgrimaud:
It works locally for me, so just merged the 2 PRs.
I saw an extra config property, agree with you, maybe it's not the best way to go, but at least, it will fix the CI, and not block other PRs.

pvliss:
Thanks @pascalgrimaud

PR #9283, opened 2019-02-17
Use config property "jhipster.clientApp.name" and `enableTranslation` flag when using HeaderUtil methods

pvliss:
This PR along with https://github.com/jhipster/jhipster/pull/188 are an attempt to fix issue #9282.

The JHipster PR will need to be merged first and then this one in order for everything to work properly

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
@@ -116,7 +116,7 @@ public class <%= entityClass %>Resource {
     } _%>
     private static final String ENTITY_NAME = "<%= entityName %>";
 
-    @Value("${spring.application.name:JHIPSTER}")
+    @Value("${jhipster.clientApp.name}")
     private String applicationName;
     <%_
     const instanceType = (dto === 'mapstruct') ? asDto(entityClass) : asEntity(entityClass);
@@ -147,7 +147,7 @@ public class <%= entityClass %>Resource {
         <%_ } _%>
 <%- include('../../common/save_template', {asEntity, asDto, viaService: viaService, returnDirectly: false, isUsingMapsId: isUsingMapsId, mapsIdAssoc: mapsIdAssoc, isController: true}); -%>
         return ResponseEntity.created(new URI("/api/<%= entityApiUrl %>/" + result.getId()))
-            .headers(HeaderUtil.createEntityCreationAlert(applicationName, ENTITY_NAME, result.getId().toString()))
+            .headers(HeaderUtil.createEntityCreationAlert(applicationName, <%= enableTranslation %>, ENTITY_NAME, result.getId().toString()))
             .body(result);
     }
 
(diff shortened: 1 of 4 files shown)


pvliss:
Also maybe we could add more tests to avoid this in the future. Will check tomorrow

cbornet:
This change makes my app crash:
```
2019-02-20 00:42:18.563  WARN 29523 --- [  restartedMain] onfigReactiveWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'cacheConfiguration' defined in file [/home/cbornet/workspace/jhipsterapps/reactive/target/classes/com/mycompany/myapp/config/CacheConfiguration.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.boot.context.properties.ConfigurationPropertiesBindException: Error creating bean with name 'io.github.jhipster.config.JHipsterProperties': Could not bind properties to 'JHipsterProperties' : prefix=jhipster, ignoreInvalidFields=false, ignoreUnknownFields=false; nested exception is org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'jhipster' to io.github.jhipster.config.JHipsterProperties
2019-02-20 00:42:18.580 ERROR 29523 --- [  restartedMain] o.s.b.d.LoggingFailureAnalysisReporter   : 

***************************
APPLICATION FAILED TO START
***************************

Description:

Binding to target [Bindable@37a4492 type = io.github.jhipster.config.JHipsterProperties, value = 'provided', annotations = array<Annotation>[@org.springframework.boot.context.properties.ConfigurationProperties(ignoreUnknownFields=false, ignoreInvalidFields=false, value=jhipster, prefix=jhipster)]] failed:

    Property: jhipster.clientapp.name
    Value: reactive
    Origin: class path resource [config/application.yml]:136:26
    Reason: The elements [jhipster.clientapp.name] were left unbound.

Action:

Update your application's configuration
```

cbornet:
`JHipsterProperties` is configured with `ignoreUnknownFields = false` so you can't add fields to `jhipster.*` without adding them also in the class.

pvliss:
Hello @cbornet. You are right that in some cases this problem occurs. See also https://travis-ci.org/jhipster/generator-jhipster/jobs/495668745. I reproduced locally using the `ngx-mongodb-kafka-cucumber` sample and the fix seems to be to also add the `jhipster.clientApp.name` property to the test `application.yml` as well.  I will do a new PR to fix this. Would you mind testing it as well locally. Thanks.

cbornet:
I don't think this is enough. You have to create the fields in `JHipsterProperties`

pascalgrimaud:
@cbornet : I think it's already the case, isn't it ?
https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/config/JHipsterProperties.java#L69

cbornet:
Mm... It seems I don't have the latest snapshot. I'll have a look.

cbornet:
The latest SNAPSHOT of jhipster-dependencies is from 02/15. Any reason it's not up-to-date ?

pascalgrimaud:
Not sure, but I think it's Julien who makes release / snapshot to Artifactory.
Locally, I need to clone `jhipster/jhipster` (or update my fork) and then, launch `./mvnw clean install -Dgpg.skip=true`. Otherwise, it won't work

cbornet:
Oh OK. I thought the SNAPSHOT was pushed automatically.

jdubois:
If you need a snapshot release just ping me directly - I'm doing one right now just in case. It has to be signed with PGP, not sure we can automate this.

cbornet:
We do automated releases of openapi-generator. It's not me who contributed it and I know it's a pain to setup so probably not worth it.

PR #9338, opened 2019-02-26
Increase precision value for decimal #9320

### LINKED ISSUE #9320: Increasing the precision value of BigDecimal

Increase the value of precision for decimal(SQL) & BigDecimal(Java) to handle a large but acceptable value.

https://github.com/jhipster/generator-jhipster/blob/d337f39c4794ea6d326dcda63acc325bc126fbfb/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs#L90-L91
The current precision and scale value of decimal(SQL) & BigDecimal(Java) covers only up to 10 million value. The scale should be fine in most case. However the current precision value is being a limitation here.

I want to add a max amount of 1 billion, 1000000000.00 which results in precision value of 12 and tests are failing with following error.
```
Value too long for column "STARTING_BALANCE DECIMAL(10, 2) COMMENT 'Starting balance for this customer.'": "-1000000000.00 (12)"; SQL statement:
```

JHipster should not limit its user to use only up to 10 million value for BigDecimal.
And as BigDecimal is often used in monetary, a higher precision value would definitely be useful.

I would like to suggest to add the precision of what a max `long` value would provide. 9,223,372,036,854,775,807 -> (19 + 2 for scale) = 21 for precision value.

I can also do a PR for it if this is approved.

https://github.com/jhipster/generator-jhipster/issues/2015
https://github.com/jhipster/generator-jhipster/issues/3552

pmverma:
Fix #9320 


jdubois:
I'm quite surprised it doesn't do anything else! LGTM, thanks @pmverma !

Issue #7886, opened 2018-06-29
Changes in global.scss loaded before bootstrap.scss


jdordoigne:
##### **Overview of the issue**

In jhispter 5, bootstrap rules have priority over styles in global.scss

##### **Motivation for or Use Case**

We have override bootstrap style in global.scss in our  jhipster 4 application. After migration in v5 all our style are broken.

##### **Reproduce the error**

For exemple add a button in home.component.html
<button class="btn btn-primary">Jhipster</button>

in global.scss add :

.btn {
    color:#28a745;
}

In v4 the text button is green, in v5 it's white. 

##### **JHipster Version(s)**

Seen in all jhipster 5 versions.

##### **JHipster configuration**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp"
    },
    "jhipsterVersion": "5.0.1",
    "applicationType": "gateway",
    "baseName": "bug",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "maven",
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "clientFramework": "angularX",
    "useSass": true,
    "clientPackageManager": "yarn",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
</pre>
</details>

##### **JDL for the Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

<details>
<summary>JDL entity definitions</summary>

<pre>

</pre>
</details>

##### **Environment and Tools**

java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

git version 2.15.2 (Apple Git-101.1)

node: v8.11.3

npm: 5.6.0

yeoman: 2.0.1

yarn: 1.7.0

[...]

PR #9339, opened 2019-02-26
Fix #9334 Remove javascript bundle global

### LINKED ISSUE #9334: Javascript bundle 'global' is always empty for angular production build

The JavaScript bundle 'global' does only contain CSS as its entrypoint is `global.(css|scss)`.
In the production build all CSS is extracted into a separate CSS file because of the MiniCssExtractPlugin.
Therefore the final `global.[hash].bundle.js` will only contain the standard webpack import stuff but no code from us and so it is useless.

##### **Reproduce the error**

- clone [jhipster-sample-app](https://github.com/jhipster/jhipster-sample-app)
- `npm install`
- `npm run webpack:prod`
- Look into `target/www/app/global.[hash].bundle.js`

##### **Suggest a Fix**

Remove the global entrypoint of the webpack production configuration and import the `global.(css|scss)` in`app.main.ts`.

I can do the PR.

##### **JHipster Version(s)**

```
jhipster-sample-application@0.0.0 C:\Home\Development\OpenSource\jhipster-sample-app
`-- generator-jhipster@5.8.1 

```

<details>
<pre>
{
  "generator-jhipster": {
    "baseName": "jhipsterSampleApplication",
    "packageName": "io.github.jhipster.sample",
    "packageFolder": "io/github/jhipster/sample",
    "authenticationType": "jwt",
    "hibernateCache": "ehcache",
    "websocket": false,
    "databaseType": "sql",
    "prodDatabaseType": "mysql",
    "useCompass": false,
    "buildTool": "maven",
    "searchEngine": false,
    "enableTranslation": true,
    "testFrameworks": [
      "gatling",
      "protractor"
    ],
    "languages": [
      "en"
    ],
    "serverPort": 8080,
    "jhipsterVersion": "5.8.1",
    "enableSocialSignIn": false,
    "useSass": false,
    "jhiPrefix": "jhi",
    "messageBroker": false,
    "jwtSecretKey": "bXktc2VjcmV0LXRva2VuLXRvLWNoYW5nZS1pbi1wcm9kdWN0aW9uLWFuZC10by1rZWVwLWluLWEtc2VjdXJlLXBsYWNl",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "reactive": false,
    "otherModules": [],
    "entitySuffix": "",
    "dtoSuffix": "DTO"
  }
}
</pre>
</details>

<details>

<pre>
entity BankAccount {
  name String required,
  balance BigDecimal required
}
entity Label {
  label String required minlength(3)
}
entity Operation {
  date DateTime required,
  description String,
  amount BigDecimal required
}
relationship OneToMany {
  BankAccount{operation} to Operation{bankAccount(name)}
}
relationship ManyToOne {
  BankAccount{user(login)} to User
}
relationship ManyToMany {
  Operation{label(label)} to Label{operation}
}

paginate Operation with infinite-scroll

</pre>
</details>

java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

git version 2.14.2.windows.1

node: v10.15.1

npm: 6.4.1

jsone-studios:
Fixes #9334 

### DIFF
--- a/generators/client/templates/angular/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.common.js.ejs
@@ -108,7 +108,7 @@ module.exports = (options) => ({
         <%_ } _%>
         new HtmlWebpackPlugin({
             template: './<%= MAIN_SRC_DIR %>index.html',
-            chunks: ['vendors', 'polyfills', 'main', 'global'],
+            chunks: ['vendors', 'polyfills', 'vendor', 'main'],
             chunksSortMode: 'manual',
             inject: 'body'
         })
--- a/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import './vendor.ts';
+import '../content/scss/global.scss';
 
 import { NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
--- a/generators/client/templates/angular/src/main/webapp/app/vendor.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/vendor.ts.ejs
@@ -97,4 +97,3 @@ library.add(faTrashAlt);
 library.add(faAsterisk);
 
 // jhipster-needle-add-element-to-vendor - JHipster will add new menu items here
-
--- a/generators/client/templates/angular/webpack/webpack.dev.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.dev.js.ejs
@@ -57,7 +57,7 @@ module.exports = (options) => webpackMerge(commonConfig({ env: ENV }), {
     },
     entry: {
         polyfills: './<%= MAIN_SRC_DIR %>app/polyfills',
-        global: './<%= MAIN_SRC_DIR %>content/scss/global.scss',
+        vendor: './<%= MAIN_SRC_DIR %>app/vendor',
         main: './<%= MAIN_SRC_DIR %>app/app.main'
     },
     output: {
--- a/generators/client/templates/angular/webpack/webpack.prod.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.prod.js.ejs
@@ -40,7 +40,7 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
     // devtool: 'source-map',
     entry: {
         polyfills: './<%= MAIN_SRC_DIR %>app/polyfills',
-        global: './<%= MAIN_SRC_DIR %>content/scss/global.scss',
+        vendor: './<%= MAIN_SRC_DIR %>app/vendor',
         main: './<%= MAIN_SRC_DIR %>app/app.main'
     },
     output: {


wmarques:
@jsone-studios thanks
Did you tested this with both configs? (CSS / SCSS)

jsone-studios:
> @jsone-studios thanks
> Did you tested this with both configs? (CSS / SCSS)

Yes, I have tested with both CSS and SCSS

ruddell:
This changes the order for `SCSS` so Bootstrap is loaded after the other styles.  Is this expected?  Will it cause issues customizing Bootstrap styles?  I'm not really familiar with how that works but I remember a related issue https://github.com/jhipster/generator-jhipster/issues/7886

**Master**:
master-scss

**This Pull Request**:
pr-scss

For CSS, the order is the same.

mraible:
I think Bootstrap styles should be first, as users (and JHipster) will want to override them.

jsone-studios:
@ruddell good catch, thanks! No this was not intended and it can cause issues with customizing bootstrap. Unfortunately I do not know why this happens only with `SCSS` and not with `CSS`. 

But I fixed it by moving the import of the `vendor.(css|scss)` into `global.(css|scss)`.

ruddell:
Another issue with this change is that for the `SCSS` option, the vendor css and user css are now together in one file instead of separate.  I think they should be separate for caching purposes (the vendor CSS won't change as often).  See the above screenshot.

I'm confused on how the global.js file gets generated.  This PR fixes it but I don't understand how or where it comes from.

jdubois:
Yes vendor and user css where separated before and that's good practice.

jsone-studios:
> Another issue with this change is that for the `SCSS` option, the vendor css and user css are now together in one file instead of separate. I think they should be separate for caching purposes (the vendor CSS won't change as often). See the above screenshot.

I will look into it and try to split the two again.

> I'm confused on how the global.js file gets generated. This PR fixes it but I don't understand how or where it comes from.

It comes from the global entrypoint in the webpack config which references a CSS or SCSS file. Webpack outputs a js file for every entrypoint. Because the global entrypoint does not contain any javascirpt, the global.js file only contains the webpack wrapper and no real code. According to [this comment](https://github.com/webpack/webpack/issues/6729#issuecomment-372359163) of a webpack member it is also not intended to use a CSS file as an entrypoint.

jsone-studios:
Bad news, there is no option to split the css into vendor and user css. So I think there are two options:
- Close this pull request and stick with the current setup and the empty global.js bundle
- Do the splitting by using a CDN for the bootstrap's css. Bootstrap should currently be the only vendor css jhipster use's, or am I wrong there?

What option do you like more?

jdubois:
Unfortunately @jsone-studios we can't use a CDN: we have many users who use JHipster internally, or in countries like China, so the CDN isn't going to work well for many of them.

jsone-studios:
I finally got it working. I have replaced the global entrypoint with a new entrypoint vendor, which will contain the `vendor.ts` and `vendor.(scss|css)`.

DanielFran:
@jsone-studios Can you please fix the conflits?

jsone-studios:
> @jsone-studios Can you please fix the conflits?

@DanielFran I am very busy the next days. Ich will fix the conflicts in a few days.

generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -108,7 +108,7 @@ module.exports = (options) => ({
         <%_ } _%>
         new HtmlWebpackPlugin({
             template: './<%= MAIN_SRC_DIR %>index.html',
-            chunks: ['vendors', 'polyfills', 'main', 'global'],
+            chunks: ['vendors', 'polyfills', 'vendor', 'main'],
```
vishal423: do we really require this new `vendor` chunk as third party packages are also contained inside `vendors` chunk? If yes, then, it would be good to provide better name to clarify content that it can contain.

jsone-studios: Yes we need the new `vendor` chunk as the vendor CSS is not extracted into the `vendors` chunk. As the `vendor` chunk contains only the import of fontawesome icons and the vendor CSS we could name it `vendor-styles`. What do you think?

wmarques: Can't we merge those two ?

ruddell:
We recently removed the CSS/SCSS option to only have SCSS, so testing this is now easier.

Comparing before and after this change, the only difference I can see is that the `global` js bundle is now renamed to a `vendor` js bundle.  Am I missing something?

**Before:**
js-before
**After:**
js-after

jsone-studios:
I could not find a solution to remove the empty `global` bundle but keep the separation of the vendor and user CSS files. As my current approach has no advantages I will close this PR.

ruddell:
Thank you for all your time and effort

PR #9340, opened 2019-02-26
Javadoc fixes

pvliss:
Yet some more minor javadoc fixes(less this time :) ) as I did another pass comparing with the Kotlin code to make sure they are aligned.

### DIFF
--- a/generators/server/templates/src/main/java/package/security/PersistentTokenRememberMeServices.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/PersistentTokenRememberMeServices.java.ejs
@@ -180,6 +180,10 @@ public class PersistentTokenRememberMeServices extends
      * <p>
      * The standard Spring Security implementations are too basic: they invalidate all tokens for the
      * current user, so when he logs out from one browser, all his other sessions are destroyed.
+     *
+     * @param request the request.
+     * @param response the response.
+     * @param authentication the authentication.
      */
     @Override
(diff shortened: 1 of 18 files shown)


generators/server/templates/src/main/java/package/security/PersistentTokenRememberMeServices.java.ejs
```diff
@@ -180,6 +180,10 @@ public class PersistentTokenRememberMeServices extends
      * <p>
      * The standard Spring Security implementations are too basic: they invalidate all tokens for the
      * current user, so when he logs out from one browser, all his other sessions are destroyed.
+     *
+     * @param request the request.
```
jdubois: I'm not a big fan of "response the response", I'd rather have no Javadoc here, as it doesn't help at all

jdubois:
Thanks a lot for the hard work @pvliss !
I made a small comment but I'll correct this myself

generators/server/templates/src/main/java/package/security/PersistentTokenRememberMeServices.java.ejs
jdubois: OK let's keep this for consistency with the other methods, as it's the only one in this case

pvliss: @jdubois Actually me neither. Its mainly to keep the javadoc tool happy :)

pvliss:
Thanks for merging @jdubois

PR #9333, opened 2019-02-24
Update angular dependencies

pvliss:
This PR updates most of the npm dependencies for Angular apps.

Relates to #9285 

-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

Issue #9285, opened 2019-02-18
Update npm packages


pvliss:
##### **Overview of the feature request**

I have lately been dealing with various issues mainly related to ones blocking dev of the Kotlin blueprint and I have noticed that many of the npm dependencies(prod+dev) of both the main generator-jhipster as well as the `package.json` for the generated apps(only checked React not sure about angular) are lacking behind.

Not sure if this on purpose(if so please let me know and close this) or simply no one has the time to deal with that task but I would like to help in bringing these up to date, at least the ones with minor and patch releases, as major upgrades will need more consideration/careful inspection.

Also while inspecting #9282 I noticed that for React there are some [dependencies](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/react/package.json.ejs)() used that have been moved/deprecated in favor of new packages(e.g [react-addons-css-transition-group](https://www.npmjs.com/package/react-addons-css-transition-group) has been replaced by [react-transition-group](https://github.com/reactjs/react-transition-group) ) and should also be fixed accordingly.

I think though that it would be best to update these before the 6.x release rather than later and the more they stay stale the more difficult they are to update.

###### **Update Tasks**

- [x] Check and update npm dependencies of the [main generator](https://github.com/jhipster/generator-jhipster/blob/master/package.json).
**NOTE:** I have done some initial work for that at https://github.com/pvliss/generator-jhipster/commits/update-npm-dependencies
- [x] Check and update npm dependencies of the [generated Angular apps](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/package.json.ejs)
- [ ] Check and update npm dependencies of the [generated React apps](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/react/package.json.ejs)

##### **Motivation for or Use Case**

Get latest features and fixes/patches of used libraries

-   [x] Checking this box is mandatory (this is just to show you read everything)

pvliss:
@jhipster-developers As this has not been closed I assume you want me to continue with this? If not then please let me know. Thanks

ruddell:
Please continue if you have time.  @DanielFran recently did some related work:
 - The Angular dependencies were upgraded (two weeks ago) here: https://github.com/jhipster/generator-jhipster/pull/9241
 - There was some work done to React but it required some code changes so it wasn't merged https://github.com/jhipster/generator-jhipster/pull/9243

pvliss:
[...]

PR #9354, opened 2019-03-01
Update node dependencies for Angular apps

pvliss:
Update to Angular 7.2.7 and other minor and patch releases

In continuation of #9333 

Relates to #9285 

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -38,7 +38,6 @@
     "@fortawesome/fontawesome-svg-core": "1.2.15",
     "@fortawesome/free-solid-svg-icons": "5.7.2",
     "@ng-bootstrap/ng-bootstrap": "4.0.4",
-    "@nguniversal/express-engine": "7.1.0",
     "@ngx-translate/core": "11.0.1",
     "@ngx-translate/http-loader": "4.0.0",
     "bootstrap": "4.3.1",
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -27,13 +27,13 @@
     "node_modules"
   ],
   "dependencies": {
-    "@angular/common": "7.2.6",
-    "@angular/compiler": "7.2.6",
-    "@angular/core": "7.2.6",
-    "@angular/forms": "7.2.6",
-    "@angular/platform-browser": "7.2.6",
-    "@angular/platform-browser-dynamic": "7.2.6",
-    "@angular/router": "7.2.6",
+    "@angular/common": "7.2.7",
+    "@angular/compiler": "7.2.7",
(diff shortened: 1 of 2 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -27,24 +27,25 @@
...
+    "@angular/core": "7.2.7",
+    "@angular/forms": "7.2.7",
+    "@angular/platform-browser": "7.2.7",
+    "@angular/platform-browser-dynamic": "7.2.7",
+    "@angular/router": "7.2.7",
     "@fortawesome/angular-fontawesome": "0.3.0",
     "@fortawesome/fontawesome-svg-core": "1.2.15",
     "@fortawesome/free-solid-svg-icons": "5.7.2",
     "@ng-bootstrap/ng-bootstrap": "4.0.4",
+    "@nguniversal/express-engine": "7.1.0",
```
wmarques: Why is this added?

pvliss: Oh sorry, forgot to comment about this one. It is a peer dependency of `ngx-cookie` and needs to be added manually. See also here https://github.com/salemdar/ngx-cookie/issues/78. This has changed in this commit: https://github.com/salemdar/ngx-cookie/commit/1956eff8940859147268f5b2b6cc3e7619a89247. Alternatively we can revert back to the previous version

wmarques: I don't understand why we should add this... The main purpose of this dependency is to do server-side rendering

pvliss: You are right. It should not be needed. Let me check again

pvliss: So you were right, it was not needed at all. I removed it. Lets wait for the CI again

For some reason it was failing locally and the error message was the same as the one described here https://github.com/salemdar/ngx-cookie/issues/78 and hastily I applied one of the workarounds without checking what the dep is actually for. Good thing we have review :)

pvliss: @wmarques So it appears that the problem is that webpack fails to build the bundle as can be seen at the CI logs after commit cf2bfad.Tried to silence webpack by using the [IgnorePlugin](https://webpack.js.org/plugins/ignore-plugin/) in commit 8115e8c but this is no good at runtime :(. So we either wait for a fix for ngx-cookie or revert back to 4.0.2, except if you have some better idea/solution to this problem. What do you think?

wmarques: Let's revert to 4.0.2 and wait for a fix. If it's not fixed in few weeks we can think about dropping this, I don't link being dependent on a non-maintained lib...

pvliss: :+1:

PR #9362, opened 2019-03-03
Make the prometheus endpoint public

PierreBesson:


Allsimon:
This endpoint gives tons of data to a potential attacker.

PierreBesson:
@allsimon if you have a good solution to this problem please discuss it on this issue : https://github.com/jhipster/generator-jhipster/issues/8946

Allsimon:
No idea, I was faced with the same issue at work and just decided to go with the public but denied to outsider via a proxy

I think we should put a disclaimer in the generated README.md

PierreBesson:
I'm merging this and doing a new PR for prometheus where I will disable prometheus by default on the prod profile.

PR #9385, opened 2019-03-07
Use Gradle 5 native BOM support

pvliss:
Gradle 5 now supports [Maven BOMs natively](https://docs.gradle.org/current/userguide/upgrading_version_4.html#rel5.0:bom_import)

The following changes have been made which **break** compatibility:
1. Switch to Gradle 5 native BOM support and remove `io.spring.dependency-management` plugin
2. Remove method `addGradleDependencyManagement() ` as the `dependencyManagement` block is no longer available. Code can use the [Gradle's native BOM support instead](https://docs.gradle.org/5.0/userguide/managing_transitive_dependencies.html#sec:bom_import)

Relates to  #9081 

Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #9389, opened 2019-03-07
Switch gradle node plugin and fix webpackBuildDev task to not execute if no client file has changed

pvliss:
As discussed in #9081 this PR switches the gradle plugin used for client tasks to https://github.com/node-gradle/gradle-node-plugin which is more active than the one currently used.

It also contains a fix to avoid constantly re-running webpack when nothing has changed.

**NOTE:** This PR is based on the branches of #9385 and #9388 to avoid conflicts, hence they should be merge first. Will do a rebase if needed.

### DIFF
--- a/generators/server/templates/gradle/profile_dev.gradle.ejs
+++ b/generators/server/templates/gradle/profile_dev.gradle.ejs
@@ -45,15 +45,16 @@ bootRun {
 
 <%_ if (!skipClient) { _%>
 task webpackBuildDev(type: <%= _.upperFirst(clientPackageManager) %>Task) {
-    inputs.files(fileTree("<%= CLIENT_MAIN_SRC_DIR %>"))
+    inputs.file("package-lock.json")
+    inputs.dir("<%= CLIENT_MAIN_SRC_DIR %>")
 
     def webpackDevFiles = fileTree("<%= CLIENT_WEBPACK_DIR %>/")
     webpackDevFiles.exclude("webpack.prod.js")
     inputs.files(webpackDevFiles)
 
-    outputs.files(fileTree("<%= CLIENT_DIST_DIR %>"))
+    outputs.dir("<%= CLIENT_DIST_DIR %>")
 
-    dependsOn <%= clientPackageManager %>_install
+    dependsOn <%= clientPackageManager %>Install
 
     args = ["run", "webpack:build"]
 }
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -44,7 +44,7 @@ plugins {
     id "com.google.cloud.tools.jib" version "0.9.11"
     id "com.gorylenko.gradle-git-properties" version "2.0.0"
     <%_ if (!skipClient) { _%>
-    id "com.moowork.node" version "1.2.0"
+    id "com.github.node-gradle.node" version "1.3.0"
     <%_ } _%>
     id "net.ltgt.apt-eclipse" version "0.21"
     id "net.ltgt.apt-idea" version "0.21"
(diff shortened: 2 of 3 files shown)


raychenon:
@pvliss , for this PR , the commit history shows that you forked from https://github.com/jhipster/generator-jhipster/pull/9388 .

Should #9388 be merged first to keep a consistent history ?

pvliss:
@raychenon Yes, it is forked from #9388 which is in turn is forked from #9385

atomfrede:
Somehow the changeset looks strange (even after merging the other branches). Can you check or rebase if needed?

pvliss:
It was indeed a bit meshed up ☹️ . I rebased and should be ok now. Lets wait for the CI again...

raychenon:
👍

generators/server/templates/gradle/profile_dev.gradle.ejs
```diff
@@ -45,15 +45,16 @@ bootRun {
 
 <%_ if (!skipClient) { _%>
 task webpackBuildDev(type: <%= _.upperFirst(clientPackageManager) %>Task) {
-    inputs.files(fileTree("<%= CLIENT_MAIN_SRC_DIR %>"))
+    inputs.file("package-lock.json")
```
atomfrede: Good catch to put `package-lock` here

generators/server/templates/gradle/profile_dev.gradle.ejs
```diff
@@ -45,15 +45,16 @@ bootRun {
...
 
     def webpackDevFiles = fileTree("<%= CLIENT_WEBPACK_DIR %>/")
     webpackDevFiles.exclude("webpack.prod.js")
     inputs.files(webpackDevFiles)
 
-    outputs.files(fileTree("<%= CLIENT_DIST_DIR %>"))
+    outputs.dir("<%= CLIENT_DIST_DIR %>")
 
-    dependsOn <%= clientPackageManager %>_install
+    dependsOn <%= clientPackageManager %>Install
```
atomfrede: Did the forked plugin change the convention task names? I think we might update the documentation as well. I think there is also something in the generated review.

generators/server/templates/gradle/profile_dev.gradle.ejs
pvliss: Thanks. Found it [here](https://guides.gradle.org/running-webpack-with-gradle/#step_3_declare_task_inputs_and_outputs) actually

generators/server/templates/gradle/profile_dev.gradle.ejs
pvliss: No they are the same. The same tasks exist in the original plugin as well. The difference is that the ones with underscore are generated dynamically. See also [here](https://github.com/node-gradle/gradle-node-plugin/blob/master/docs/node.md#executing-npm-tasks)

atomfrede: :heavy_check_mark: Does `yarnInstall` also exist? At least it is not explicitly mentioned in the documentation?

pvliss: Good question. I did not check but I assumed it does. Let me check in the code

pvliss: If I am not mistaken, it does. See https://github.com/node-gradle/gradle-node-plugin/blob/8ac7cbe248bae5c430cf4f9c6576db1e23b66501/src/main/groovy/com/moowork/gradle/node/NodePlugin.groovy#L57.

And a bit below that https://github.com/node-gradle/gradle-node-plugin/blob/8ac7cbe248bae5c430cf4f9c6576db1e23b66501/src/main/groovy/com/moowork/gradle/node/NodePlugin.groovy#L101  you can see the dynamic generation of the underscore variants, e.g: yarn_install

deepu105:
Sorry guys didn't have time to check. Hope you ensured everything works
fine when in dev/prod clean builds and incremental builds and so on

Issue #9400, opened 2019-03-08
Locale is not set with OAuth / OIDC and Keycloak


mraible:
##### **Overview of the issue**

When I log in with Keycloak, the user's locale is not set. I'm using the latest code from the master branch.

<img width="1345" alt="screen shot 2019-03-06 at 5 13 18 pm" src="https://user-images.githubusercontent.com/17892/54038269-1a111780-418e-11e9-87c8-66770b4f2a54.png">

If I try it using the latest release - 5.8.2 - this issue doesn’t exist.

Here's my JDL:

```
application {
  config {
    baseName gateway,
    packageName com.okta.developer.gateway,
    applicationType gateway,
    authenticationType oauth2,
    prodDatabaseType postgresql,
    searchEngine elasticsearch,
    serviceDiscoveryType eureka,
    testFrameworks [protractor]
  }
  entities Blog, Post, Tag, Product
}

application {
  config {
    baseName blog,
    packageName com.okta.developer.blog,
    applicationType microservice,
    authenticationType oauth2,
    prodDatabaseType postgresql,
    searchEngine elasticsearch,
    serverPort 8081,
    serviceDiscoveryType eureka
  }
  entities Blog, Post, Tag
}

application {
  config {
    baseName store,
    packageName com.okta.developer.store,
    applicationType microservice,
    authenticationType oauth2,
    databaseType mongodb,
    devDatabaseType mongodb,
    prodDatabaseType mongodb,
    enableHibernateCache false,
    searchEngine elasticsearch,
    serverPort 8082,
    serviceDiscoveryType eureka
  }
  entities Product
}

entity Blog {
  name String required minlength(3),
  handle String required minlength(2)
}

entity Post {
  title String required,
  content TextBlob required,
  date Instant required
}

entity Tag {
  name String required minlength(2)
}

entity Product {
  title String required,
  price BigDecimal required min(0),
  image ImageBlob
}

relationship ManyToOne {
  Blog{user(login)} to User,
  Post{blog(name)} to Blog
}

relationship ManyToMany {
  Post{tag(name)} to Tag{post}
}

paginate Post, Tag with infinite-scroll
paginate Product with pagination

microservice Product with store
microservice Blog, Post, Tag with blog
```

vishal423:
I think it would be good to add langKey attribute in the keycloak realm. I guess we already have server side code to map from that attribute

PR #9402, opened 2019-03-08
Set locale to 'en' if not specified by IdP

mraible:

Fixes https://github.com/jhipster/generator-jhipster/issues/9400.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/UserService.java.ejs
+++ b/generators/server/templates/src/main/java/package/service/UserService.java.ejs
@@ -851,7 +851,7 @@ public class UserService {
             // Since JHipter only has one "en" language, we will handle it here,
             // for other languages, please handle it accordingly.
             String locale = (String) details.get("locale");
-            if(locale.startsWith("en_") || locale.startsWith("en-")) {
+            if (locale.startsWith("en_") || locale.startsWith("en-")) {
                 locale = defaultLocale;
             }
             user.setLangKey(locale.toLowerCase());
--- a/generators/server/templates/src/main/java/package/service/UserService.java.ejs
+++ b/generators/server/templates/src/main/java/package/service/UserService.java.ejs
@@ -822,7 +822,6 @@ public class UserService {
 
     private static <%= asEntity('User') %> getUser(Map<String, Object> details) {
         <%= asEntity('User') %> user = new <%= asEntity('User') %>();
-        final String defaultLocale = "en";
         user.setId((String) details.get("sub"));
         if (details.get("preferred_username") != null) {
             user.setLogin(((String) details.get("preferred_username")).toLowerCase());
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -847,12 +848,16 @@ public class UserService {
...
             String locale = (String) details.get("locale");
-            if(locale.startsWith("en_") || locale.startsWith("en-")) {
-                locale = "en";
+            if (locale.startsWith("en_") || locale.startsWith("en-")) {
+                locale = defaultLocale;
             }
             user.setLangKey(locale.toLowerCase());
+        } else {
+            // set langKey to default if not specified by IdP
+            user.setLangKey(defaultLocale);
```
vishal423: If translations are enabled, then, can we use default language key `nativeLanguage` to set this value?

mraible: Good idea. Fixed!

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -847,12 +847,16 @@ public class UserService {
...
             // https://github.com/jhipster/generator-jhipster/issues/7866
             // for the en_US or en_** issue.
-            // Since JHipter only have one "en" language, we will handle it here, for other languages, please handle it accordingly.
+            // Since JHipter only has one "en" language, we will handle it here,
+            // for other languages, please handle it accordingly.
             String locale = (String) details.get("locale");
-            if(locale.startsWith("en_") || locale.startsWith("en-")) {
-                locale = "en";
+            if (locale.startsWith("en_") || locale.startsWith("en-")) {
+                locale = Constants.DEFAULT_LANGUAGE;
```
vishal423: Shouldn't this be `en` to comply with if statement?

mraible: Yep. Good catch. Fixed.

lmbai:
If we didn’t set the `local` key in the backend, the front end will use the default lang. 

right ? Correct me if I am wrong.

mraible:
If there's no langKey set on the user, there's an error in the JavaScript console after the user logs in. This is what this PR aims to fix.

lmbai:
I see, I didn’t test that yet, anyway this fix is good to have. But I would to fix the console  error too. The front end should not depend on the backend to give a local key. If no key or not value, it should the default key.

mraible:
This does fix the console error.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
lmbai: i think it will be better we use the following, what do you think ? @mraible 
```
            user.setLangKey("<%= nativeLanguage %>");
```

PR #9404, opened 2019-03-08
Fix Cucumber tests to use @DirtiesContext to have a clean context for each scenario

pvliss:
This fixes the failing integration tests for Cucumber in the `ngx-mongodb-kafka-cucumber` sample.

Example logs: also https://travis-ci.org/jhipster/generator-jhipster/jobs/503656883

### DIFF
--- a/generators/server/templates/src/test/java/package/cucumber/CucumberContextConfiguration.java.ejs
+++ b/generators/server/templates/src/test/java/package/cucumber/CucumberContextConfiguration.java.ejs
@@ -21,11 +21,13 @@ package <%=packageName%>.cucumber;
 import <%=packageName%>.<%= mainClass %>;
 import cucumber.api.java.Before;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.web.WebAppConfiguration;
 
 @SpringBootTest
 @WebAppConfiguration
+@DirtiesContext
 @ContextConfiguration(classes = <%= mainClass %>.class)
 public class CucumberContextConfiguration {
 
--- a/generators/server/templates/src/test/java/package/cucumber/CucumberIT.java.ejs
+++ b/generators/server/templates/src/test/java/package/cucumber/CucumberIT.java.ejs
@@ -20,8 +20,9 @@ package <%=packageName%>.cucumber;
 
 import org.junit.runner.RunWith;
 
-<% if (databaseType === 'cassandra') { %>
-import <%=packageName%>.AbstractCassandraTest;<% } %>
+<%_ if (databaseType === 'cassandra') { _%>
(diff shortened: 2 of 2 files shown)


cbornet:
Won't this make the tests very slow ?

pvliss:
Yes, possibly. As there are only a few tests it does not seem to take too long. I have not really used Cucumber before so there might be a better workaround. At least with this fix the tests do not fail.

pvliss:
Let me provide a few more details for the issue here.

So the problem is that for some reason the `CucumberIT` does not clean up the Spring context properly and this results in some of the cucumber tests/steps and all of the rest of the integration tests that run after it to fail. The key error message is:
` Unsatisfied dependency expressed through method 'cacheManager' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'jCacheCacheManager' defined in class path resource [org/springframework/boot/autoconfigure/cache/JCacheCacheConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [javax.cache.CacheManager]: Factory method 'jCacheCacheManager' threw exception; nested exception is javax.cache.CacheException: A Cache named [usersByLogin] already exists`

The only way I could find to mitigate the problem is using `@DirtiesContext` as explained [here](https://github.com/cucumber/cucumber-jvm/tree/master/spring#annotation-based-configuration). This will obviously be slower but it is my understanding that it only affects the Cucumber tests and not the other tests in terms of performance.

Regarding performance of the CI test run. I checked the last [5.8.2 Travis run ](https://travis-ci.org/jhipster/generator-jhipster/jobs/499900566) which has a total of 8 min 36 sec and for the `CucumberTest ` a total running time of 22.314 sec. In comparison the [Travis Run for this PR](https://travis-ci.org/jhipster/generator-jhipster/jobs/503711089) has a total of 10 min 15 sec and for `CucumberIT` a total of 32.061 s. The difference for the Cucumber test is bearable and the total is indeed bigger but I do not think iy this is related to this change as the total running time for the `21-tests-backend.sh` script is 172.67s for 5.8.2 and 187.95s for this PR. That is a difference of 15 seconds which is more or less close to the diff of the Cucumber tests

Finally, as I said I am not familiar at all with Cucumber so there might be a better solution. If anyone has a better idea of how to fix this please do so as we need to fix the CI builds to make our lives easier when checking new PRs as currently all of them will fail and might hide other errors too.

pvliss:
Since no one has commented with an alternative solution I am merging this to get CI back to green status.

pascalgrimaud:
@pvliss : since this PR, the daily build for cucumber failed.
See https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build/results?buildId=1721

Could you have a look, if you have time plz ? Otherwise, I should be able to try to fix it next week

pvliss:
@pascalgrimaud Yes, I will try and have a look today

PR #9414, opened 2019-03-10
Feature/fix null lang key under oauth2

lmbai:
In the PR #9402 , @mraible fix the null `locale` key issue from the backend.
Here, i would like to prevent it from the frontend. 

in the `react` side, it using the same logic:
```js
    const { account } = getState().authentication;
    if (account && account.langKey) {
      const langKey = Storage.session.get('locale', account.langKey);
      await dispatch(setLocale(langKey));
    }
```
```js
// angular: generators\client\templates\angular\src\main\webapp\app\core\auth\account.service.ts.ejs
	 import { LANGUAGES } from 'app/core/language/language.constants';
		if (langKey && LANGUAGES.includes(langKey)) {
			this.languageService.changeLanguage(langKey);
		} 
// react: generators\client\templates\react\src\main\webapp\app\shared\reducers\authentication.ts.ejs
 import { languages } from 'app/config/translation';
    if (account && account.langKey) {
      const langKey = Storage.session.get('locale', account.langKey);
      if(languages[langKey ]) {
         await dispatch(setLocale(langKey));
      }
    }
```
**But i think it may be also reasonable that let it throw error since it will fail to load the invalid lang-key.json, so that developer will know there is issue... WDYT ?**


pvliss:
Closing and re-opening to relaunch CI tests

DanielFran:
Closing/Reopenning

DanielFran:
Thanks @lmbai

Issue #9276, opened 2019-02-16
Upgrade Spring Security's OIDC Support


mraible:
### Overview of the feature request

Spring Security 5.1 and (native with Spring Boot 2.1) has a new way to configure OAuth 2.0 Login. Rather than using `@EnableOAuth2Ssso`, it's a regular part of the web security configuration. This issue is to track that upgrade progress.

- [x] Create monolith example using Spring Security 5.1's OIDC Support
- [x] Create microservices example using Spring Security 5.1's OIDC Support
- [x] Update Spring Security configuration for OAuth 2.0 monolith and microservices into generator
- [x] Ensure all integration tests pass (across all servers)

### Motivation for or Use Case

Using `@EnableOAuth2Sso` and `@EnableResourceServer` is no longer recommended in Spring Boot 2.1. The reasons for the change can be found in Josh Long's [Bootiful Podcast](https://itunes.apple.com/us/podcast/a-bootiful-podcast/id1438691771?mt=2), published on Jan 25, 2019. It's an interview with Madhura Bhave and the discussion starts at 21:30.

The pull request below shows the changes necessary to modify a v5.8.1 monolith with defaults + OAuth 2.0 for authentication. Please have a look and let me know if you see any issues.

https://github.com/mraible/jhipster-oidc-improved/pull/1

#### Known Issues

* The principal's authorities don't get updated by the user authorities mapper. The authorities are correct in the `OAuth2AuthenticationToken`, but not in its principal.

<img width="1047" alt="screen shot 2019-02-16 at 1 11 32 pm" src="https://user-images.githubusercontent.com/17892/52904964-4a4d4200-31f0-11e9-8eda-bc2deb398ef5.png">

yelhouti:
@mraible this implementation has security issues, please let's fix them before moving forward.

mraible:
@yelhouti Please provide a pull request to this PR’s branch with your suggested changes. I’ll be happy to review your fixes. Saying there’s issues without providing a fix doesn’t help me.

yelhouti:
Sorry @mraible I forgot to delete this comment, and in order to put everything in the long issue we already have.
The main issue here is that OIdC is used the some way you would with OAuth2 Authorization flow, on the other hand the [spec](https://openid.net/specs/openid-connect-core-1_0.html#TokenResponseValidation) clearly says that after validating the token the same way you would with OAuth2 (step 1) you MUST:
```
[...]

PR #9416, opened 2019-03-11
Upgrade to Spring Security 5.1's OIDC Support

mraible:
Contributes to fixing #9276.


pvliss:
Closing and re-opening to relaunch CI tests

mraible:
Closing and reopening to start new build.

generators/server/templates/src/main/java/package/security/oauth2/AuthRedirectController.java.ejs
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.GetMapping;
+
+/**
+ * This controller is protected by default, and only used to specify "/authorize" as a login-triggering link.
+ */
+@Controller
+public class AuthRedirectController {
+
+    @GetMapping("/authorize")
```
mraible: It'd be nice if this wasn't needed. It seems you need to request a protected resource to trigger OIDC login with Spring Security 5.1. Since it comes back to this endpoint after successful authentication, it redirects to the client app.

generators/server/templates/src/test/java/package/web/rest/LogoutResourceIT.java.ejs
```diff
@@ -58,36 +54,23 @@ public class LogoutResourceIT {
...
     @Before
     public void before() {
-        LogoutResource logoutResource = new LogoutResource(restTemplateFactory(), accessTokenUri);
+        LogoutResource logoutResource = new LogoutResource(registrations);
         this.restLogoutMockMvc = MockMvcBuilders.standaloneSetup(logoutResource)
             .setMessageConverters(jacksonMessageConverter).build();
     }
 
-    @Test
+    @Test @Ignore
```
mraible: This test needs improvement.

generators/server/templates/src/test/resources/config/application.yml.ejs
```diff
@@ -139,17 +139,21 @@ spring:
...
-            client-secret: web_app
-            scope: openid profile email
-        resource:
-            user-info-uri: http://DO_NOT_CALL:9080/auth/realms/jhipster/protocol/openid-connect/userinfo
+    security:
+        oauth2:
+            client:
+                provider:
+                    oidc:
+                        issuer-uri: http://localhost:9080/auth/realms/jhipster
```
mraible: I believe this means that Keycloak has to be running if you want integration tests to pass. This is because of OIDC discovery and how the endpoints are looked up from `.well-known/openid-configuration`. For example, https://dev-133320.okta.com/oauth2/default/.well-known/openid-configuration.

mraible:
@jhipster/developers Are you OK if I merge this? I'm leaving on vacation for 10 days tomorrow. It seems like a perfect time. ;)

pascalgrimaud:
We have a problem with GitHub and Travis. We can't see the Travis results in PR
The build is here: https://travis-ci.org/jhipster/generator-jhipster/builds/507021423

There is a lint error but I'm fixing it directly in your branch

After that, let's merge this !

mraible:
Thanks @pascalgrimaud! Since all tests pass, I'm merging.

pmverma:
@mraible I was generating an app from 5.8.2 to current master, then I got compile error for `OAuth2Configuration.java`.
I think this needs to be included in cleanup.js since this file has been removed with your new OIDC support.

mraible:
Can you please create an issue for this? I can do it when I get back from
vacation on March 26.

DanielFran:
@mraible dependency org.springframework.security.oauth.boot:spring-security-oauth2-autoconfigure was used by previous implementation only for OAUTH2

With your changes, it is now defined to be used by UAA. Is there any reason?

prior:
https://github.com/jhipster/generator-jhipster/blob/v5.x_maintenance/generators/server/templates/pom.xml.ejs#L585

after:
https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/pom.xml.ejs#L609

mraible:
@DanielFran I tried not to modify UAA in any way since I don't know much about it. If UAA doesn't need it, we should remove it.

jdubois:
I also have integration tests errors with this new code...
Here is my `.yo-rc.json`:
```

  "generator-jhipster": {
    "promptValues": {
    },
    "jhipsterVersion": "5.8.2",
    "baseName": "micro",
    "serverPort": "8081",
    "cacheProvider": "no",
    "enableHibernateCache": false,
    "websocket": false,
    "databaseType": "mongodb",
    "devDatabaseType": "mongodb",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "jwtSecretKey": "ZjAxMTZhN2MyNWMwYzAzOWVlMTY1ODY5M2Y1OGQzNGI5ZGYxYjQ0ZDgyYmExN2JjN2NiMjhmMTA5YWQ3ZmJiNzU2YjJmYzhjZmU2MWQ1MWQ2N2Q2N2JiMmZmNDNmODhhNWM2NDNkOWQ3Y2IwYzRhYzRhYTE0Njk0YjA5YTZiZDQ=",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "enableTranslation": false,
    "skipClient": true,
    "skipUserManagement": true
  }
}
```
Running `./mvnw package -Pprod verify jib:dockerBuild` fail during integration tests, with message:
```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.security.config.annotation.web.configuration.OAuth2ClientConfiguration$OAuth2ClientWebMvcSecurityConfiguration': Unsatisfied dependency expressed through method 'setClientRegistrationRepository' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'clientRegistrationRepository' defined in class path resource [org/springframework/boot/autoconfigure/security/oauth2/client/servlet/OAuth2ClientRegistrationRepositoryConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository]: Factory method 'clientRegistrationRepository' threw exception; nested exception is java.^Clang.IllegalArgumentException: Unable to resolve the OpenID Configuration with the provided Issuer of "http://localhost:9080/auth/realms/jhipster"

**Comment by mraible** [comment]
@jdubois I'll see if I can reproduce. Related: why do you have a `jwtSecretKey` in an app definition for oauth2?

**Comment by jdubois** [comment]
@mraible I don't know why it's there: it's probably always generated when using the CLI, even if it's not used.

**Comment by mraible** [comment]
@jdubois The more complete error you're seeing is:

```
Factory method 'clientRegistrationRepository' threw exception; nested exception is java.lang.IllegalArgumentException: Unable to resolve the OpenID Configuration with the provided Issuer of "http://localhost:9080/auth/realms/jhipster"
```

Integration tests now require Keycloak to be running because all the endpoints are looked up -on-the-fly - an OIDC feature. We could mock out the `http://localhost:9080/auth/realms/jhipster/.well-known/openid-configuration` endpoint with Wiremock or something, but these are integration tests and it seems like the integrated services should be running.

**Comment by jdubois** [comment]
Oh sorry @mraible !!! Indeed I remember seeing this... This is quite annoying, at least can you wrap the error in the test with a try/catch block, to provide a meaningful error message and explain people what they need to do?

**Comment by mraible** [comment]
I'm not sure how to try/catch a dependency injection issue in Spring.

@jgrandja Do you have any suggestions on how we might go about having better error messages (in tests) when the IdP's well-known endpoint is not available?

Here's what we see when running an integration tests and Keycloak is not running:

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate 
[org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository]: 
Factory method 'clientRegistrationRepository' threw exception; nested exception is 
java.lang.IllegalArgumentException: Unable to resolve the OpenID Configuration with the 
provided Issuer of "http://localhost:9080/auth/realms/jhipster"
```

**Comment by jgrandja** [comment]
@mraible I don't think it makes sense to catch an exception when the `ApplicationContext` does not initialize and startup properly - it should proceed to fail since it would be in an inconsistent state.

Could you not start Keycloak in `@BeforeClass` in the integration test class to ensure it's up before the tests run?

**Comment by mraible** [comment]
@jgrandja I figured out a solution thanks to some code and tests I found in Spring Security repositories. I created a new `@TestConfiguration` class that defines the `ClientRegistrationRepository` and `JwtDecoder` beans.

```java
import org.springframework.boot.test.context.TestConfiguration;
import org.springframework.context.annotation.Bean;
import org.springframework.security.oauth2.client.registration.ClientRegistration;
import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
import org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository;
import org.springframework.security.oauth2.core.AuthorizationGrantType;
import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
import org.springframework.security.oauth2.jwt.JwtDecoder;

import static org.mockito.Mockito.mock;

/**
 * This class allows you to run unit and integration tests without an IdP.
 */
@TestConfiguration
public class TestSecurityConfiguration {
    private ClientRegistrationRepository clientRegistrationRepository;

    public TestSecurityConfiguration() {
        ClientRegistration mockRegistration = clientRegistration().build();
        this.clientRegistrationRepository = new InMemoryClientRegistrationRepository(mockRegistration);
    }

    @Bean
    ClientRegistrationRepository clientRegistrationRepository() {
        return this.clientRegistrationRepository;
    }

    private ClientRegistration.Builder clientRegistration() {
        return ClientRegistration.withRegistrationId("oidc")
            .redirectUriTemplate("{baseUrl}/{action}/oauth2/code/{registrationId}")
            .clientAuthenticationMethod(ClientAuthenticationMethod.BASIC)
            .authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
            .scope("read:user")
            .authorizationUri("https://jhipster.org/login/oauth/authorize")
            .tokenUri("https://jhipster.org/login/oauth/access_token")
            .jwkSetUri("https://jhipster.org/oauth2/jwk")
            .userInfoUri("https://api.jhipster.org/user")
            .userNameAttributeName("id")
            .clientName("Client Name")
            .clientId("client-id")
            .clientSecret("client-secret");
    }

    @Bean
    JwtDecoder jwtDecoder() {
        return mock(JwtDecoder.class);
    }
}
```

Then I import this class in `@SpringBootTest`. 

```java
@SpringBootTest(classes = {MicroApp.class, TestSecurityConfiguration.class})
```

It'd be cool if it the `TestSecurityConfiguration` could be auto-detected, but I think this is "good enough".

jgrandja:
@mraible 

> It'd be cool if it the `TestSecurityConfiguration` could be auto-detected, but I think this is "good enough".

If you make the class `static`, it should auto-detect. Try this:

```
@TestConfiguration
static class TestSecurityConfiguration {
...
```

mraible:
@jgrandja Nope - `static` [isn't allowed in a top level class](https://javarevisited.blogspot.com/2017/04/can-we-declare-class-static-in-java.html). 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project micro: Compilation failure: Compilation failure: 
[ERROR] /Users/mraible/oauth2/src/test/java/com/mycompany/myapp/config/TestSecurityConfiguration.java:[18,8] modifier static not allowed here
```

FWIW, I just created a PR at https://github.com/jhipster/generator-jhipster/pull/9484.

jgrandja:
@mraible I totally missed that :(  What I meant to say is that a **static inner class** should be auto-detected. Not sure if making `TestSecurityConfiguration` as an static inner class in the test class makes sense? In general, static inner class `@Configuration` should get auto-detected.

mraible:
@jgrandja That would work for just one test, but there are around 8 integration tests in JHipster that need this OIDC mocked. Having one config class and imports might be the easiest to understand.

jgrandja:
@mraible Yeah, that does make sense. Forget anything I just said :)

avdev4j:
Oh @mraible 
This PR is very welcome.
I tried oauth2 with openAM on both 5.8.2 and Master.

Well this new way is highly better and easier to configure and use.

avdev4j:
Hi @mraible 
I have some trouble with oauth2 login on my k8s environment. 
Before upgrading to JHipster 6+ we used to use OAuth2Configuration enven in "production" env (mean in all env other than local).

like this:
```
/**
 * Development only configuration that is Browsersync-aware and redirects to the origin you clicked "login" from.
 * If you split your application into client and server on separate domains, you might want to enable this for prod
 * mode too.
 */
@Configuration
//@Profile("dev")
public class OAuth2Configuration {
```
This config class called another one named OAuth2AuthenticationSuccessHandler.
 
With the new Spring Security 5.1 and JHipster 6+ we don't have this class anymore because the spring-security-oauth2 dependency has been removed.
This dependency was mandatory for OAuth2Configuration to access to spring OAuth2ClientAuthenticationProcessingFilter.

I wonder if this modification is really expected or if we have another way to still use the filter.

thanks !

mraible:
@avdev4j Since this issue is closed, can you please ask your question on Stack Overflow with the "jhipster" tag? Please explain your use case and what you're trying to achieve. If I don't know the answer -- hopefully, the Spring Security folks can chime in.

mraible:
What are you trying to do? There might be a different way to do it with Spring Security 5.1.

avdev4j:
Hi @mraible 
I'm sorry to answer only now but I had a lack of time.
We have front and backend separate and deployed on different domain. In previous JHipster version (before Spring security 5.1) we could activate the `OAuth2Configuration` class for prod environement.

Now this class doesn't exist anymore I struggled a little bit to doing the same. Our problem was Spring sent redirect uri in http even if we first used https.

But the problem was not really related to the spring application because if we access the webapp by an https protocol our ingress (k8s cluster) redirect the request to the gateway in http. That's why we had this issue, the "https" had become "http".

To fix it (if people also meet this problem) :
Just add this in the spring configuration

```
server
    use-forward-headers: true
```

PR #9420, opened 2019-03-12
Add bootswatch themes question

wmarques:

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<nav class="navbar navbar-dark navbar-expand-md bg-dark">
+<nav class="navbar navbar-<% if (clientTheme === 'none' || clientThemeVariant !== 'light') { %>dark<% } else { %>light<% } %> navbar-expand-md <% if (clientTheme === 'none') { %>bg-dark<% } else { %>bg-<%= clientThemeVariant %><% } %>">
     <a class="navbar-brand logo" routerLink="/" (click)="collapseNavbar()">
         <span class="logo-img"></span>
         <span jhiTranslate="global.title" class="navbar-title"><%= capitalizedBaseName %></span> <span class="navbar-version">{{version}}</span>
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<nav class="navbar navbar-<% if (clientTheme === 'none' || clientThemeVariant !== 'light') { %>dark<% } else { %>light<% } %> navbar-expand-md <% if (clientTheme === 'none') { %>bg-dark<% } else { %>bg-<%= clientThemeVariant %><% } %>">
+<nav class="navbar navbar-<% if (clientTheme === 'none') { %>dark<% } else { %><%= clientThemeVariant %><% } %> navbar-expand-md <% if (clientTheme === 'none') { %>bg-dark<% } else { %>bg-<%= clientThemeVariant %><% } %>">
     <a class="navbar-brand logo" routerLink="/" (click)="collapseNavbar()">
         <span class="logo-img"></span>
         <span jhiTranslate="global.title" class="navbar-title"><%= capitalizedBaseName %></span> <span class="navbar-version">{{version}}</span>
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
@@ -39,11 +39,13 @@ Navbar
 
 .navbar {
     padding: 0.2rem 1rem;
+    <%_ if (clientTheme !== 'none') {_%>
     .dropdown-item.active,
     .dropdown-item.active:focus,
     .dropdown-item.active:hover {
         background-color: $dark;
     }
+    <%_ } _%>
 
     .dropdown-item:hover {
         color: $dropdown-link-hover-color;
@@ -59,11 +61,13 @@ Navbar
         font-weight: 400;
     }
 
+    <%_ if (clientTheme !== 'none') {_%>
     .navbar-toggler {
         &:hover {
             color: $navbar-dark-hover-color;
         }
     }
+    <%_ } _%>
 }
 
 /* ==========================================================================
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
@@ -18,6 +18,9 @@
 -%>
 
 @import '~bootstrap/scss/functions';
+<%_ if (clientTheme !== 'none') { _%>
+@import "~bootswatch/dist/<%= clientTheme %>/variables";
+<%_ } _%>
 @import '~bootstrap/scss/variables';
 
 /* ==========================================================================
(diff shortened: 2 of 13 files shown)


generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
```diff
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<nav class="navbar navbar-dark navbar-expand-md bg-dark">
+<nav class="navbar navbar-<% if (clientTheme === 'none' || clientThemeVariant !== 'light') { %>dark<% } else { %>light<% } %> navbar-expand-md <% if (clientTheme === 'none') { %>bg-dark<% } else { %>bg-<%= clientThemeVariant %><% } %>">
```
vishal423: do you need `clientThemeVariant !==` check in first if block? Can it be simplified like 2nd if block?

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
```diff
@@ -42,21 +45,25 @@ Navbar
         background-color: $dark;
```
vishal423: all color variations in this stylesheet are jhipster custom theme specific. On choosing external theme, it would be good to not apply these as default

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
```diff
@@ -42,21 +45,25 @@ Navbar
...
     }
 
+    .dropdown-item:hover {
+        color: $dropdown-link-hover-color;
```
vishal423: If I understood this correctly, then, you are trying to revert to default hover color. If yes, then we override that in style above this to align with custom Jhipster theme. If you don't override styles as per my last comment, then, this wouldn't be required.

generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
```diff
@@ -151,39 +151,11 @@ Custom alerts for notification
 entity tables helpers
```
vishal423: Styles like below may not be relevant with custom theme (like material). It would be good to make those conditional and not apply if user chooses some theme

> .ng-valid[required],
> .ng-valid.required {
>     border-left: 5px solid green;
> }
> 
> .ng-invalid:not(form) {
>     border-left: 5px solid red;
> }

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
```diff
@@ -23,9 +23,16 @@ put Sass variables here:
 eg $input-color: red;
 ****************************/
 // Override Boostrap variables
+<%_ if (clientTheme === 'none') { _%>
 @import "bootstrap-variables";
-// Import Bootstrap source files from node_modules
+<%_ } else {_%>
+@import "~bootswatch/dist/<%= clientTheme %>/variables";
```
vishal423: do you need this as you are also importing theme below that should include it as default?

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
```diff
@@ -23,9 +23,16 @@ put Sass variables here:
...
+<%_ } _%>// Import Bootstrap source files from node_modules
 @import '~bootstrap/scss/bootstrap';
```
vishal423: can you cross check if this is required when user chooses custom theme. Generally customized themes also import bootstrap and you don't have to do this yourself

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
```diff
@@ -23,9 +23,16 @@ put Sass variables here:
 eg $input-color: red;
 ****************************/
 // Override Boostrap variables
+<%_ if (clientTheme === 'none') { _%>
 @import "bootstrap-variables";
```
vishal423: we shouldn't make this conditional, as it would be useful to customize selected theme. Instead, we should make it's content conditional

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
wmarques: No because we need to also check if `clientThemeVariant` is not light

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
wmarques: Not it's because bootstrap apply the hover-color only on `a` elements and not on `dropdown-items`. We use sometimes dropdown-items without links and we need to have the same color on hover.

generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
wmarques: I would keep them as it's good for validation, user can change it afterwards if they do not fit the theme

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
wmarques: you're correct

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
wmarques: No we need bootstrap also: https://github.com/thomaspark/bootswatch

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
vishal423: I don't see this in navbar template. Are you talking about customization that you do post generation?

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
vishal423: ok

generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
vishal423: ok. I hope none of theme functionality depends on bootstrap javascript that we don't include

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
vishal423: IMO, if user has selected a theme, then, we should use that as is. Second if block is doing that by using theme variant to specify navbar background

generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
vishal423: I think the idea of pre-configured theme is to not apply any custom styles. If user require further customization, then, they should do manually.

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.scss.ejs
wmarques: No it is in the bootstrap code. They apply this color only for the `a` tags inside `dropdown-item`, however we want to apply this color on all `dropdown-item`, whether they contains a link or not.

vishal423: if this change is not scoped to `navbar`, then, shouldn't it go as global style? Also, this should be conditional and not applied if theme is selected to ensure that their default is respected

wmarques: I can move it into global yes.
For me it's a `bootstrap` strange behaviour so I would always apply this fix even with a theme. We use a variable so the color is not hard-coded.

mraible:
FWIW, this build is failing with the following error:

```
[INFO] Using 1 worker with 2048MB memory limit
[INFO]  ERROR  Failed to compile with 1 errors4:06:38 AM
[INFO] 
[INFO]  error  in ./src/main/webapp/content/scss/global.scss
[INFO] 
[INFO] Module build failed (from ./node_modules/sass-loader/lib/loader.js):
[INFO] 
[INFO]   color: $dropdown-link-hover-color;
[INFO]         ^
[INFO]       Undefined variable.
[INFO]    ╷
[INFO] 28 │   color: $dropdown-link-hover-color;
```

pascalgrimaud:
@wmarques : is it ready for beeing merged ?

wmarques:
@pascalgrimaud for me yes it's ready, I took all @vishal423 comments and the builds are green so that should be good

jdubois:
Awesome @wmarques !
You'll also need to update the "create application" documentation, and JHipster Online, if you have time

wmarques:
@jdubois done https://github.com/jhipster/jhipster.github.io/pull/720

PR #9429, opened 2019-03-14
Fix user management create/edit flows

vishal423:
Replace ngModel occurrence with corresponding reactive form control, update tests

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
@@ -34,6 +34,7 @@ export class UserMgmtUpdateComponent implements OnInit {
     isSaving: boolean;
 
     editForm = this.fb.group({
+        id: [''],
         login: ['', [Validators.required, Validators.minLength(1), Validators.maxLength(50), Validators.pattern('^[_.@A-Za-z0-9-]*')]],
         firstName: ['', [Validators.maxLength(50)]],
         lastName: ['', [Validators.maxLength(50)]],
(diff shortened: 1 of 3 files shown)


avdev4j:
Hi @vishal423 
Could you please provide some details about the goal of this PR ?

thanks

vishal423:
@avdev4j, done.

wmarques:
For me we shouldn't use the reactive form as the field isn't editable... That's why there was a readonly ngModel

vishal423:
IMO, It's bad design to mix `ngModel` and reactive forms in the same component. Irrespective of design choice, this code needs to be corrected to fix following console error:

> ERROR 
> Error: 
>       ngModel cannot be used to register form controls with a parent formGroup directive.  Try using
>       formGroup's partner directive "formControlName" instead.  Example:
>     <div [formGroup]="myGroup">
>       <input formControlName="firstName">
>     </div>
>     In your class:
>     this.myGroup = new FormGroup({
>        firstName: new FormControl()
>     });
>       Or, if you'd like to avoid registering this form control, indicate that it's standalone in ngModelOptions:
>       Example:
>     <div [formGroup]="myGroup">
>        <input formControlName="firstName">
>        <input [(ngModel)]="showMoreControls" [ngModelOptions]="{standalone: true}">
>     </div>
>

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
```diff
@@ -98,6 +100,7 @@ export class UserMgmtUpdateComponent implements OnInit {
     }
 
     private updateUser(user: User): void {
+        user.id = this.editForm.get(['id']).value;
```
wmarques: is it necessary as we won't update the user ID ?

wmarques:
@vishal423 oh yes you are correct about the error, let's do this this way then

generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-update.component.ts.ejs
vishal423: not really. Will update later today

PR #9461, opened 2019-03-21
update helm generator for istio changes

srinivasa-vasu:
- update helm generated services 
- remove kiali, as the latest istio release includes it out of the box
- helm services version upgrade


DanielFran:
Cosing/Reopenning

PR #9473, opened 2019-03-24
Fix gradle webpackBuildDev task to not fail when package-lock.json or `yarn.lock is missing

pvliss:
Fix Gradle `webpackBuildDev` task to not fail when either `package-lock.json` or `yarn.lock` is missing when using either npm or  yarn respectively.

When using `inputs.file(file)` and the referenced file is missing then gradle will throw an error and fail. Instead when using `inputs.files(file)`(mind the s) gradle considers it as a collection of paths and if any is missing it is simply ignored. We should not fail when the `package-lock.json` is missing but instead continue to execute `npm install` when `./gradlew` is executed.

This is another regression after merging #9389.

-   [] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green


pvliss:
Well it is usually not the case. But when playing around with dependencies and having linked modules(e.g. generator-jhipster + kotlin blueprint) I sometimes get errors running 'npm install'. Removing the lock file and running `npm install' makes everything work again :)

PR #9484, opened 2019-03-27
Add ability to run tests with no OIDC Provider running

mraible:

Fixes integration issues discussed in https://github.com/jhipster/generator-jhipster/pull/9416#issuecomment-476724974


generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
```diff
@@ -23,6 +23,7 @@ package <%=packageName%>.web.rest;
 import <%=packageName%>.AbstractCassandraTest;
 <%_ } _%>
 import <%=packageName%>.<%= mainClass %>;
+import <%=packageName%>.config.TestSecurityConfiguration;
```
DanielFran: Did you missed the condition if (authenticationType === 'oauth2')....?

mraible: This isn't needed in this case because there's an `<%_ if (authenticationType === 'oauth2') { _%>` on line 20.

mraible:
I'm seeing the following error fairly consistenly in the [JHI_APP=ngx-mariadb-oauth2-sass-infinispan JHI_PROTRACTOR=1 JHI_ENTITY=sql test](https://travis-ci.org/jhipster/generator-jhipster/jobs/512027047):

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

Any idea how to fix?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
DanielFran: But this condition ended in line 24, no?

DanielFran: Forget, I read it bad :(

mraible:
This may be caused by Travis having issues: https://www.traviscistatus.com/

Screen Shot 2019-03-27 at 1 12 19 PM

mraible:
C'mon Travis, you can do it!

DanielFran:
@mraible It seems that this didi not fixed all cases: see https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build/results?buildId=1995

mraible:
@DanielFran Thanks for the heads up. Looking now.

mraible:
These all seem to be failing because Keycloak is not running for the e2e tests. 

@pascalgrimaud Do you know where are Docker containers started in this jhipster-daily-build project's scripts?

pascalgrimaud:
@mraible : in [jhipster-daily-builds](https://github.com/hipster-labs/jhipster-daily-builds), it uses the same script in the main generator-jhipster, so it is in this folder:
- https://github.com/jhipster/generator-jhipster/blob/master/test-integration/scripts/20-docker-compose.sh

mraible:
Thanks for the fast response @pascalgrimaud. Keycloak is not started for these instances because no `keycloak.yml` is installed. This is because they're microservices and require a gateway for e2e tests. Why are we running e2e tests on microservices?

pascalgrimaud:
Oh I think I understand. I didn't notice we remove `keycloak.yml` for microservice.
For microservices, there is no e2e, just a check in `management/health` :

https://github.com/jhipster/generator-jhipster/blob/master/test-integration/scripts/24-tests-e2e.sh#L22-L24

mraible:
Hmmm, what's the best way to solve this then? 

1. Remove e2e tests for microservices
2. Figure out a way to add Keycloak and related files just for tests

We don't want to include Keycloak files for microservices because it just confuses the end user.

pascalgrimaud:
Thanks for pointing the problem, I'll take care of this.
I will add a specific Keycloak only for our CI: it is important to test if the microservice is launched correctly

PR #9491, opened 2019-03-28
Add default docker-compose files for Prometheus and Grafana

clement26695:
Create by default some Docker Compose files for using Prometheus and Grafana to monitor your applications, with an autoconfigured default dashboard in Grafana.

For the moment in this PR, Prometheus listen for instances on : 
- localhost:8080
- localhost:8081
- localhost:8082
- localhost:8083

I'm not sure if it's ok ?

-   [] Documentation is added/updated where necessary

### DIFF
--- a/generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
+++ b/generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
@@ -35,6 +35,7 @@ datasources:
     # <int> org id. will default to orgId 1 if not specified
     orgId: 1
     # <string> url
+    # Change localhost by host.docker.internal on MacOS
     url: http://localhost:9090
     # <string> database password, if used
     password:
@@ -47,7 +48,7 @@ datasources:
     # <string> basic auth username
     basicAuthUser: admin
     # <string> basic auth password
-    basicAuthPassword: foobar
+    basicAuthPassword: admin
     # <bool> enable/disable with credentials headers
     withCredentials:
     # <bool> mark as default datasource. Max one per org
--- a/generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
+++ b/generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
@@ -35,7 +35,7 @@ datasources:
     # <int> org id. will default to orgId 1 if not specified
     orgId: 1
     # <string> url
-    # Change localhost by host.docker.internal on MacOS
+    # On MacOS, replace localhost by host.docker.internal
     url: http://localhost:9090
     # <string> database password, if used
     password:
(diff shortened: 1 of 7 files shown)


generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
```diff
@@ -0,0 +1,67 @@
...
+    # <string> database user, if used
+    user:
+    # <string> database name, if used
+    database:
+    # <bool> enable/disable basic auth
+    basicAuth: false
+    # <string> basic auth username
+    basicAuthUser: admin
+    # <string> basic auth password
+    basicAuthPassword: foobar
```
jdubois: Could you use `admin` instead? As we usually have `admin/admin` everywhere in JHipster

jdubois:
For ports: you could use the one from the project itself, it's stored in the configuration (`.yo-rc.json`).

BTW the official list of ports is at https://www.jhipster.tech/common-ports/ and maybe you also need to adds the ports from Prometheus/Grafana?

generators/server/templates/src/main/docker/monitoring.yml.ejs
```diff
@@ -0,0 +1,40 @@
...
+        image: <%= DOCKER_GRAFANA %>
+        volumes:
+            - ./grafana/provisioning/:/etc/grafana/provisioning/
+        environment:
+            - GF_SECURITY_ADMIN_PASSWORD=jhipster
+            - GF_USERS_ALLOW_SIGN_UP=false
+            - GF_INSTALL_PLUGINS=grafana-piechart-panel
+        ports:
+            - 3000:3000
+        network_mode: "host" # to test locally running service
```
clement26695: It does not work on Mac OS...

PierreBesson: I remember we had found a solution for mac os, maybe we can add it as a comment.

clement26695: I manage to make it work by removing the network_mode "host" and using host.docker.internal as the ip address

PierreBesson: yes that's the solution I think.

generators/server/templates/src/main/docker/prometheus/prometheus.yml.ejs
```diff
@@ -0,0 +1,46 @@
...
+  - job_name: 'prometheus'
+
+    # Override the global default and scrape targets from this job every 5 seconds.
+    scrape_interval: 5s
+
+    # scheme defaults to 'http'.
+    metrics_path: /management/prometheus
+    static_configs:
+      - targets:
+          - localhost:8080
```
PierreBesson: I think you can access the serverPort variable from the app configuration in the template rather than guessing the port like that.

generators/server/templates/src/main/docker/grafana/provisioning/datasources/datasource.yml.ejs
```diff
@@ -35,6 +35,7 @@ datasources:
     # <int> org id. will default to orgId 1 if not specified
     orgId: 1
     # <string> url
+    # Change localhost by host.docker.internal on MacOS
```
PierreBesson: Please change this comment to be clearer: `On Mac OS, replace `localhost` by `host.docker.internal`

PR #9496, opened 2019-03-28
chore: avoid include test tables in liquibase

clakech:
When using `mvn liquibase:update` to update database schema with latest changes then the test table `jhi_date_time_wrapper` pop up in our database. By default we do not want this, but one could override this settings in command line if needed :)


jdubois:
Indeed! I'm assigning @avdev4j as he coded this part

jdubois:
LOL @avdev4j how come you're not part of the JHipster org yet??? I just invited you

avdev4j:
I have just tested this modification from @clakech and I think it's good.
I was not aware of this (and I'm surprised people don't raised it before)

But if we want to handle a new context we will have to add it to the plugin configuration. Should we document this ? (Indeed it's the normal behaviour).

jdubois:
Thanks @clakech @avdev4j - yes this is good, let's merge this

PR #9506, opened 2019-03-29
oauth2: remove redirect controller

vishal423:
- Remove intermediate oauth2 redirect controller by pointing UI to oidc login url


vishal423:
@mraible, do you foresee issue with this?

mraible:
@vishal423 The problem with doing this it prevents end users from using all of Spring Security's OIDC features. Currently, if users specify multiple IdPs in their `application.yml`, they'll be redirected to `/login` and Spring Security will display a list of providers they can click on to log in. This change basically hard-codes the "oidc" provider so users won't see multiple choices, even if they've configured them.

mraible:
We could remove the `/authorize` endpoint if I could get Spring Security's "redirect back to referrer" logic working. I did it on a previous app with the following bean:

```java
@Bean
public RequestCache refererRequestCache() {
    return new HttpSessionRequestCache() {
        @Override
        public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
            String referrer = request.getHeader("referer");
            if (referrer != null) {
                request.getSession().setAttribute("SPRING_SECURITY_SAVED_REQUEST", new SimpleSavedRequest(referrer));
            }
        }
    };
}
```

However, I wasn't able to get it working in JHipster. 

@jgrandja Is a `RequestCache` bean all that's needed to add redirect-to-referrer functionality in the latest Spring Security?

vishal423:
@mraible, I didn't check with multiple providers, however, `/login` endpoint is still available to use. Considering SPA, there are high chances that end users would like to create their own custom login page and use that instead of one provided by Spring security. We can still add a comment on the UI code to educate users to point to `/login` in case they configure multiple providers.

jgrandja:
@mraible The default `AuthenticationSuccessHandler` for `OAuth2LoginAuthenticationFilter` is `SavedRequestAwareAuthenticationSuccessHandler`, which has a `HttpSessionRequestCache`. If you want to provide your own `RequestCache` than you can `SavedRequestAwareAuthenticationSuccessHandler.setRequestCache(customRequestCache)` and than configure the instance via `http.oauth2Login().successHandler(customSuccessHandler)`.

vishal423:
@mraible, do you want to proceed with this? If not, I will close, otherwise, I will do similar change on react code as well

mraible:
@vishal423 That's fine, as long as it's documented.

generators/client/templates/angular/src/main/webapp/app/core/login/login.service.ts.ejs
```diff
@@ -49,7 +49,10 @@ export class LoginService {
         if (!contextPath.endsWith('/')) {
             contextPath = contextPath + '/';
         }
-        location.href = `//${location.hostname}${port}${contextPath}authorize`;
+
+        // If you have configured multiple OIDC providers, then, you can update this URL to /login.
+        // It will show Spring security generated login page with links to configured OIDC providers.
```
mraible:
```suggestion
        // It will show a Spring Security generated login page with links to configured OIDC providers.
```

generators/client/templates/react/src/main/webapp/app/shared/util/url-utils.ts.ejs
```diff
@@ -18,5 +18,8 @@
 -%>
 export const getLoginUrl = () => {
   const port = location.port ? `:${location.port}` : '';
-  return `//${location.hostname}${port}${location.pathname}authorize`;
+
+  // If you have configured multiple OIDC providers, then, you can update this URL to /login.
+  // It will show Spring security generated login page with links to configured OIDC providers.
```
mraible:
```suggestion
  // It will show a Spring Security generated login page with links to configured OIDC providers.
```

vishal423:
@mraible, I was looking into generated `SecurityConfiguration` and noted that we use `ProviderConfigurationRequest` to retrieve OIDC provider configurations and with current code, it will constrain the use of multiple providers. Do we have any specific reason to choose this approach?

    @Value("${spring.security.oauth2.client.provider.oidc.issuer-uri}")``

mraible:
@vishal423 If there's a better way to get the `issuer-uri` for the OIDC provider that's used to log in, I'm all for it!

vishal423:
Ok. I will spend sometime later this week to check on that part.

vishal423:
@mraible , Can you merge this? I think this change is independent of multiple OIDC providers support. I am bit occupied on my react sample and wouldn't be able to devote time on this before June

Issue #9145, opened 2019-01-29
Upgrade to JDK 11 while keeping JDK8 compatibility


PierreBesson:
##### **Overview of the feature request**

After merging the Spring-boot 2.1 branch, we will be able to fully support JDK 11. As discussed on the [mailing list java 11 thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jhipster-dev/O85Y96sp_Hk/a22j4FHiAgAJ), we are going to keep JDK 8 through 11 compatibility for now, however we can upgrade to JDK 11 as default for : 

- [x] Travis/Azure builds, mix between jdk8 and jdk11 builds
- [ ] Base docker image

For the docker image, we can use :
  - `adoptopenjdk/openjdk11:alpine-slim` (106MB not TCK verified for now)
  - `adoptopenjdk/openjdk11:slim` (238MB)
see https://github.com/AdoptOpenJDK/openjdk-docker

I'm recommending that we switch to adoptopenjdk as this is my understanding that AdoptOpenJDK jdk 11 docker images provide LTS while the official openjdk image does not (@mraible please correct me if I'm wrong). The official openjdk image also does not offer an alpine version.

Alternative: use the distroless java11 image (https://github.com/GoogleContainerTools/distroless/pull/285) once it becomes available: 

- [ ] The `jhipster/jhipster` docker image
- [x] Update the docs with instructions for installing JDK 11, either with SDKman or with AdoptOpenJDK (eg. `brew tap AdoptOpenJDK/openjdk && brew cask install adoptopenjdk`)
- [ ] Heroku
- [ ] GAE
- [ ] Cloudfoundry
- [ ] AWS
- [ ] CI-CD where applicable

##### **Motivation for or Use Case**

Use the latest JDK while adding integration tests for compatibility with JDK8.

##### **Related issues or PR**

#8689, #8573

-   [X] Checking this box is mandatory (this is just to show you read everything)

PierreBesson:
On the current master, we use openjdk:11-slim as the base docker image. I believe this will stop receiving updates when jdk12 will be released so I would recommend switching to either `adoptopenjdk/openjdk11:alpine-slim` (242MB) or `gcr.io/distroless/java:11` from google (189MB). @jhipster/developers What do you think ?

pascalgrimaud:
So no 'official' image ?
Which one has the most download ?
Which one is the most active ?

PierreBesson:
[...]

PR #9539, opened 2019-04-06
Switch to OpenJDK11

pascalgrimaud:
We discussed with @jdubois during our last JHipster day, and I think we should stick with a more classic installation of OpenJDK, instead of using SDKMAN.

@pvliss: do you agree with that ?
@PierreBesson : what do you think too ?

_____

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,10 +70,11 @@ before_install:
     - |
         if [[ $JHI_JDK = '11' ]]; then
             echo '*** Using OpenJDK 11'
-            curl -s get.sdkman.io | bash
-            echo sdkman_auto_answer=true > ~/.sdkman/etc/config
-            source "$HOME/.sdkman/bin/sdkman-init.sh"
-            sdk install java 11.0.2-open
+            sudo add-apt-repository ppa:openjdk-r/ppa
+            sudo apt-get update
+            sudo apt-get install -y openjdk-11-jdk
+            sudo update-java-alternatives -s java-1.11.0-openjdk-amd64
+            java -version
         else
             echo '*** Using OpenJDK 8 by default'
         fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,13 @@ RUN \
     g++ \
     libpng-dev \
     build-essential \
+    software-properties-common \
     sudo && \
+  # install OpenJDK 11
+  add-apt-repository ppa:openjdk-r/ppa && \
+  apt-get update && \
+  apt-get install -y openjdk-11-jdk && \
+  update-java-alternatives -s java-1.11.0-openjdk-amd64 && \
   # install node.js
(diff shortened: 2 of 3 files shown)


PierreBesson:
Why ? Do we have a bug with sdkman ? 

Sdkman allows us to have a large choice of current and beta releases of the jdk so I think it's valuable to use it. What about jdk13 ? Sdkman can allow us quickly launch a test with a devlopment release of jdk13.

Also using a ppa like you do is already an unofficial source. So is it really better than sdkman ?

pascalgrimaud:
@PierreBesson : no issue I think, it should be all resolved.

I proposed this PR as Julien seems to not like it. And I don't use sdkman either.
So is it the same way to install JDK11 in prod environment too ? I mean with sdkman ?

PierreBesson:
No sdkman is just for dev. It can install the various jdks in your home directory.

pvliss:
I personally use SDK man for dev and also believe that it makes it easy to
test other dists as well but don't really have a strong opinion about using
it in JHipster as long as the other dists work :). What I am not sure about
SDKMan is how long the dists are retained and if that's going to affect CI
and that's worries me a bit.

Στις Σάβ, 6 Απρ 2019, 11:56 μ.μ. ο χρήστης Pierre Besson <
notifications@github.com> έγραψε:

pvliss:
Close/Reopen to relaunch tests

pascalgrimaud:
As there is no issue with SDKMAN, let's keep this for now.
No need to merge this PR.

pascalgrimaud:
I'm reopening this, to trigger Azure build with JDK11, as current master failed with SDKMAN.

Related to https://github.com/jhipster/generator-jhipster/issues/9145

pvliss:
Sadly, it appears that SDKMan does not keep the versions which is obviously an issue for CI as noone wants to change the config each time a new release is out. Hence, I am merging this. If anyone has any objections please comment and I can revert.

PierreBesson:
This is super strange that sdkman jdk version cannot be fixed somehow...

pvliss:
@PierreBesson It appears that SDKMan does not keep minor versions once a new one is published. Also tested this locally. Removed locally installed version using 'sdk uninstall java 11.0.2-open' and then trying to re-install it using `sdk install java 11.0.2-open` I was getting the following error message:
```
Stop! java 11.0.2-open is not available. Possible causes:
 * 11.0.2-open is an invalid version
 * java binaries are incompatible with Linux64
 * java has not been released yet
```

Had to install 11.0.3-open.

An alternative could be to use the CI cache in order to keep it around(although not sure if entirely possible) but then again I don't think that would be a better solution.

pascalgrimaud:
Not possible with Azure for now. There is 'o cache, it is on their roadmap

PR #9552, opened 2019-04-11
Remove InternalServerErrorException

cbornet:

A generic exception will automatically be translated to an Internal Server Exception by problem-spring-web so we can remove InternalServerErrorException and use a simpler one that just extend RuntimeException.
We can't use RuntimeException directly because it would make Sonar unhappy.


DanielFran:
Thanks @cbornet

Issue #9286, opened 2019-02-18
Leverage Confluent Kafka docker images


framiere:
As seen face-to-face today with @jdubois , @Dabz and I propose to leverage our Confluent expertise to provide better Kafka support in JHipster.

### Outdated kafka images

JHipster is relying on old kaka version by `wurstmeister`
https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js#L35

### Leverage Confluent images

We propose to use the confluent images instead, so they are always up to date.
https://hub.docker.com/r/confluentinc/cp-kafka

### Target

We propose to update both the `docker-compose` and `helm` charts with these images.

### Longer term target

Being able to provide full [KSQL](https://github.com/confluentinc/ksql), [Kafka Connect](https://kafka.apache.org/documentation/#connect) and [Schema Registry](https://github.com/confluentinc/schema-registry) support.

Let us know what you think.

jdubois:
As we all agree, let's do this!

pascalgrimaud:
I suppose we could use confluentinc/cp-zookeeper
https://hub.docker.com/r/confluentinc/cp-zookeeper too ?

DanielFran:
Is this supposed to be done for v6?
Would love  to have this has soon...

pascalgrimaud:
@framiere @Dabz : I supposed you didn't have time to start the work.

So I had a quick look yesterday, and what I already did:
- migrate to official images for kafka and zookeeper
- update configuration for spring boot
- migrate from Spring Cloud Stream to native Apache Kafka client (related to https://github.com/jhipster/generator-jhipster/issues/9287)
- implement a simple rest controller, with 2 services (1 for Producer and 1 for Consumer)

Will you be able to do a review once I propose the PR ?

Dabz:
Sure, I will have a look to https://github.com/jhipster/generator-jhipster/pull/9567 today or tomorrow ;)

Issue #9287, opened 2019-02-18
Migrate from spring-cloud-stream to native Apache Kafka client


Dabz:
As seen face-to-face today with @jdubois , @framiere and I propose to leverage our Confluent expertise to provide better Kafka support in JHipster. 

##### **Overview of the feature request**

JHipster is currently generating the Apache Kafka source code with [Spring Cloud Stream](https://cloud.spring.io/spring-cloud-stream/), we propose to rely on the native Apache Kafka client instead. 

##### **Motivation for or Use Case**

Spring Cloud Stream is always [lagging a few version behind Apache Kafka](https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-stream-binder-kafka/2.1.1.RELEASE) and it makes the generated source code harder to read.

While we do see value in Spring Cloud Stream, from our experience, only advanced users would really benefit from this kind of framework and it is treacherous for beginners and intermediate users. While it helps, it hides too much semantic and blocks people from understanding the behaviour of their applications (e.g. automatic retry, error handling, idempotency, etc...).

jdubois:
As we all agree here, let's do this!

Tcharl:
I may be the only one that use rabbitmq. Not that important, Kafka is a pretty nice beast :-)

DanielFran:
@pascalgrimaud Can this issue be closed?

pascalgrimaud:
For now, we migrated from Spring Cloud Stream to Spring Kafka. If I understand well, we should use directly the native client.
So I don't know.

Maybe @ctamisier you have an advise on this ?
Otherwise, I think we can close and improve later if needed.

ctamisier:
I think there are some improvements in progress about the docker image properties.
https://github.com/Dabz/generator-jhipster/commit/b3fe482fad4b357f117f292ad0fb90e218e7d3ec
maybe this ? :)

I have some work in progress (80% done) using kafka native client (producer and consumer) and test containers, that can be a "drop-in" replacement of what already exists in JHipster, I'll submit it later as a PR.

Then I'm still wondering if a producer or consumer should be directly included in a JHipster web application rather than in an "external" component dealing directly between kafka and the underlying database.

If it does make sense we can even think about integrating kafka stream and let the developer taking advantage of advanced concepts (kstream, ktable, joins, windowing,...) or not.

pascalgrimaud:
[...]

PR #9567, opened 2019-04-16
Kafka: improve support

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/9286
Related to https://github.com/jhipster/generator-jhipster/issues/9287

I'm learning Kafka and following this blog https://www.confluent.io/blog/apache-kafka-spring-boot-application

If the docker-compose file is OK, I will update all configuration for Kubernetes, Openshift, etc.

It is a very first step.
Then, maybe we can push to specific topic for each entities ?

cc @jdubois @framiere @Dabz
_____ 


cbornet:
What would be great is to have something that shows up in the front end. For instance streaming the audit events either by websocket (if the option is used) or by http streaming.

cbornet:
Or the tracking activity already managed by websockets.

pascalgrimaud:
Yes it's a good idea @cbornet 
It would be cool to do something similar to the websocket page

cbornet:
With the advantage that it would work with a multi-instance/scaled app.

pascalgrimaud:
So for this first PR, let's focus only on backend. Improvements on Frontend can be done in another PR :D

I'll update the docker-compose files and Kubernetes config to use the new properties and this PR will be ready to be reviewed

cbornet:
The problem with the generated producer, consumer and endpoints is that although they are a good example of working code, they are most probably code that people will not need and that they will need to remove. So I'm not sure that's a great dev experience...

pascalgrimaud:
You're totally right, it should be better with entities but as I don't have any experience, I don't know how to handle it.

Here, it's a simple hello world of the use of producer / consumer. Nothing really exciting.

@cbornet : do you have experience with Kafka ?

jdubois:
For me this is just a simple template with a working example, and that's fine for people to start customizing.

cbornet:
@pascalgrimaud yes I have some experience with Kafka. We use it at work for a variety of use-cases. Actually, you can do a lot of different things with Kafka such as event bus, stream processing, event sourcing, ... That's why it's hard to provide a basic integration that would be interesting in all case.
I actually consider to migrate some of our Kafka Stream based micro-services to JHipster. This would add automatically some features that we require for such services : actuator endpoints (including prometheus), logger endpoint, dev/prod profiles, docker-compose setup, small admin UI, etc... But for that we also need to be able to generate apps without user management, without DB, etc... Hence the PRs I've pushed recently. I think JHipster would be valuable for purely stateless services/apps but we have some cleanup work to do first.
As for Kafka, the most interesting parts we can generate I think, is dependencies, Spring Boot autoconf, dev/prod profiles, and test conf (BTW there is probably some conf to do to configure an embedded kafka in test). The rest is not that useful because we don't know if users will do Kafka Streams, KSQL, KafkaConsumer/KafkaProducer or KafkaTemplate.

pascalgrimaud:
There are already a lot of changes.
I think it is enough for a 1st improvement for Kafka.

What could be done later:
- a UI / page for producing and seeing new message
- something related to websocket ?
- maybe some code related to entities ?

generators/server/templates/src/main/java/package/web/rest/KafkaResource.java.ejs
```diff
@@ -0,0 +1,43 @@
...
+package <%=packageName%>.web.rest;
+
+import <%=packageName%>.service.SampleKafkaProducer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping(value = "/api/kafka")
+public class KafkaResource {
```
cbornet: Maybe add an test for this class ?

generators/server/templates/src/test/java/package/CassandraKeyspaceIT.java.ejs
```diff
@@ -23,9 +23,15 @@ import com.datastax.driver.core.Session;
...
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+<%_ if (messageBroker === 'kafka') { _%>
+import org.springframework.kafka.test.context.EmbeddedKafka;
+<%_ } _%>
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+<%_ if (messageBroker === 'kafka') { _%>
+@EmbeddedKafka
```
cbornet: Is it needed to have an EmbeddedKafka in all tests ?

pascalgrimaud: Thanks a lot for starting to review my PR @cbornet 

I pushed quickly what I did before vacations, so not really sure about this part.
And I don't like this part, as I added `@EmbeddedKafka` everywhere. I started to use a dedicated `@Configuration` in `EmbeddedKafkaConfiguration` but I needed to customize the Embedded Kafka myself and didn't manage to make it works.

So ideas are welcome here

generators/server/templates/src/main/java/package/web/rest/KafkaResource.java.ejs
pascalgrimaud: sure, will do

pascalgrimaud:
Just pushed sample projects to play with:
https://github.com/pascalgrimaud/jhipster-kafka-samples

Tested with Docker-Compose and Kubernetes -> everything looks OK for me

pascalgrimaud:
@jdubois : related to your tweet https://twitter.com/juliendubois/status/1126192008283004930?s=19
Just keep in mind, this PR will remove spring-cloud-stream dependencies

vwiencek:
Hello, actually I see a use-case for Kafka with microservices / hipster. In a Microservice architecture, we often have :
- one main `User` entity holding credential information in a user management microservice
- for all other microservices, specific service related `User` entities (like `serviceAUser` or `serviceBUser` holding complementary information

Then each time we modify the main `User` entity (for example `email` or `firstname`, we need to keep other `serviceXUser` in sync with the main one.

The user management service might send an event to the kafka bus service that will be consumed exactly once by other microservices.

mraible:
Good idea. Or you could use oauth2 for the authentication type and let your identity provider do this for you.

pascalgrimaud:
Yes, there are a lot of use cases with Kafka, but I'd prefer to keep PR as little as possible.
This one is already big, so once it's ok and merged, we can improve some other use case.

pascalgrimaud:
Unless someone is against this, I think I will merge this PR.
For me, it's far better than what we currently have. So then, we can do some little PR to advance steps by steps.

If someone is against, let me know !

PR #9574, opened 2019-04-18
Enable blueprint to define custom subgenerator

sahbi-ktifa:
Blueprints can easily modify main generator behavior, however, quickly grows the need to define custom subgenerator for specific tasks.

To ensure that possibility, this PR enables to define a subgenerator in a blueprint using a simple file formatted as such:
*generator-jhipster-testblueprint/commands.js*
```
module.exports = {
    foo: {
        desc: 'bar', // A little description that will be displayed using --help
        blueprint: 'generator-jhipster-testblueprint' // This is required to determine the blueprint issuing the custom command
    }
};
```

When using `jhipster foo --blueprint testblueprint`, the foo subgenerator from the blueprint will be called directly.

I'd like some reviews on this one, as on the feature itself as on the code. This will clearly enable users to define specific stuff on their own enjoying all JHipster cli mechanism.

### DIFF
--- a/cli/commands.js
+++ b/cli/commands.js
@@ -18,10 +18,11 @@
  */
 
 let customCommands = {};
-if (process.argv.indexOf('--blueprint') > -1) {
+const indexOfBlueprintArgv = process.argv.indexOf('--blueprint');
+if (indexOfBlueprintArgv > -1) {
     /* eslint-disable import/no-dynamic-require */
     /* eslint-disable global-require */
-    customCommands = require(`generator-jhipster-${process.argv[process.argv.indexOf('--blueprint') + 1]}/cli/commands`);
+    customCommands = require(`generator-jhipster-${process.argv[indexOfBlueprintArgv + 1]}/cli/commands`);
 }
 
 const defaultCommands = {
--- a/cli/commands.js
+++ b/cli/commands.js
@@ -16,7 +16,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-module.exports = {
+let commands = {};
+if (process.argv.indexOf('--blueprint') > -1) {
+    commands = require(`generator-jhipster-${process.argv[process.argv.indexOf('--blueprint') + 1]}/commands`);
+}
+
+const defaultCommands = {
     app: {
         default: true,
         desc: 'Create a new JHipster application based on the selected options'
@@ -103,3 +108,4 @@ Example:
         desc: 'Upgrade the JHipster version, and upgrade the generated application'
     }
 };
+module.exports = Object.assign(defaultCommands, commands);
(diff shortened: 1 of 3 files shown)


pascalgrimaud:
Just to know, is it doable to add a test in this folder ?
https://github.com/jhipster/generator-jhipster/tree/master/test/blueprint

sahbi-ktifa:
I added a test to this !
I can only test that it will try to delegate its job to the blueprint directly :)

deepu105:
Let me take a look at this

cli/utils.js
```diff
@@ -188,7 +188,12 @@ const createYeomanEnv = () => {
     Object.keys(SUB_GENERATORS)
         .filter(command => !SUB_GENERATORS[command].cliOnly)
         .forEach(generator => {
-            env.register(require.resolve(`../generators/${generator}`), `${CLI_NAME}:${generator}`);
+            if (SUB_GENERATORS[generator].blueprint) {
+                /* eslint-disable prettier/prettier */
+                env.register(require.resolve(`../node_modules/${SUB_GENERATORS[generator].blueprint}/generators/${generator}`), `${CLI_NAME}:${generator}`);
```
deepu105: this expects the package to be installed on the node_modules of the generator? Shouldn't you rely on npms resolution mechanism here

cli/commands.js
```diff
@@ -103,3 +111,4 @@ Example:
         desc: 'Upgrade the JHipster version, and upgrade the generated application'
     }
 };
+module.exports = Object.assign(defaultCommands, commands);
```
deepu105: Use spread plz
```
module.exports = {
...defaultCommands,
...commands
}
```

deepu105:
I'm ok with idea

cli/utils.js
sahbi-ktifa: You are totally right !

cli/commands.js
sahbi-ktifa: I'm not used to use this syntax :) 
Thanks for the hint!

avdev4j:
Hi @sahbi-ktifa 
That's sound great since blueprints can't, always, match with the whole client need (call "the ultra specific part" :D).

I wonder if we should put the command file in a "cli" folder: "cli/commands.js" to match with the orignal JHipster file.

sahbi-ktifa:
Hello @avdev4j, thanks for the review!
What did you meanusing the cli? What was your intent ?

sahbi-ktifa:
Ok, I changed `commands.js` file under `cli` folder.
This is easier to understand and maintain.

cli/commands.js
```diff
@@ -16,7 +16,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-module.exports = {
+
+let customCommands = {};
+if (process.argv.indexOf('--blueprint') > -1) {
```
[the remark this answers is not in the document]
sahbi-ktifa: Fixed it :)

Issue #9569, opened 2019-04-17
Documentation: integration tests


pascalgrimaud:
Following https://github.com/jhipster/generator-jhipster/issues/8868 https://github.com/jhipster/generator-jhipster/pull/9174 https://github.com/jhipster/generator-jhipster/pull/9337 about Integration Tests

The documentation need to be updated:
- README in generated project
- jhipster.tech

pascalgrimaud:
Closing this, as the PR to jhipster.github.io is done.
It will be merged for v6

PR #9583, opened 2019-04-19
Update generated README, related to integration tests [ci skip]

pascalgrimaud:
Related to https://github.com/jhipster/generator-jhipster/issues/9569

### DIFF
--- a/generators/common/templates/README.md.ejs
+++ b/generators/common/templates/README.md.ejs
@@ -228,7 +228,7 @@ Refer to [Doing API-First development][] for more details.
 
 To build the final jar and optimize the <%= baseName %> application for production, run:
 <% if (buildTool === 'maven') { %>
-    ./mvnw -Pprod clean package<% } %><% if (buildTool === 'gradle') { %>
+    ./mvnw -Pprod clean verify<% } %><% if (buildTool === 'gradle') { %>
     ./gradlew -Pprod clean bootJar<% } %>
 
 <%_ if (!skipClient) { _%>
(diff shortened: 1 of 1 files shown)


generators/common/templates/README.md.ejs
```diff
@@ -247,14 +247,18 @@ Refer to [Using JHipster in production][] for more details.
...
+    ./mvnw -Pprod,war clean verify<% } %><% if (buildTool === 'gradle') { %>
     ./gradlew -Pprod clean bootWar<% } %>
 
 ## Testing
 
 To launch your application's tests, run:
 
-    <% if (buildTool === 'maven') { %>./mvnw clean test<% } else { %>./gradlew test<% } %>
+    <%_ if (buildTool === 'maven') { _%>
+    ./mvnw test integration-test
```
murdos: Unless I'm misunderstanding something, I would just use `verify` (it also includes `test`) rather than `integration-test`, according to https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html : 
>  If integration-test were to be called from the command line, no reports are generated. Worse is that the integration test container environment is left in a hanging state; the Tomcat webserver or Docker instance is left running, and Maven may not even terminate by itself.

pascalgrimaud: Ok thx, @murdos -> I'm doing another PR to change this

pascalgrimaud: The only minor issue I saw is `packaging` the application and `launching` tests is the same command now. So the information is here twice and is similar to Gradle
But it's really minor

PR #9585, opened 2019-04-19
Add possibility to also authenticate by Basic auth on JWT monoliths

cbornet:

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -278,6 +278,10 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter<% if (au
             .antMatchers("/management/prometheus").permitAll()
             .antMatchers("/management/**").hasAuthority(AuthoritiesConstants.ADMIN)<%_ if (authenticationType === 'session') { %>;<% } %>
         <%_ if (authenticationType === 'jwt') { _%>
+            <%_ if (applicationType === 'monolith') { _%>
+        .and()
+            .httpBasic()
+            <%_ } _%>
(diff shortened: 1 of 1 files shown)


cbornet:
@PierreBesson We could now protect the prometheus endpoint at least for monoliths with JWT :smiley:

PierreBesson:
Let me try it out... The problem is that it creates a "basic auth backdoor" to your app.

cbornet:
Not really a backdoor, just a little less secure than JWT since you transmit the no-expiry password on each request. You just have to know what you're doing when using it...

pascalgrimaud:
@PierreBesson : are you ok with this change ? Can we merge this ?

PierreBesson:
No, Currently the password should be admin/admin with no explicit property file to change. I think the property must be `spring.security.user.name/password` but we need to test. Also initially, I wanted to have this only on /management/** endpoints... This needs a lot of docs as it is a pretty big change ! Also it potentially has a performance impact has it will add an additional servlet filter for all requests.

cbornet:
This has no link with the property file. The login/pwd are those of the user DB.

PierreBesson:
Interesting ! What about my remark on the additional servlet filter this might introduce ?

cbornet:
That should be tested I guess. The worse would be an extra query to the DB...

cbornet:
So if you don't use basic auth the filter will only do:
```java
String header = request.getHeader("Authorization");

		if (header == null || !header.toLowerCase().startsWith("basic ")) {
			chain.doFilter(request, response);
			return;
		}
```
I guess the impact should be minimal.

cbornet:
Ok to merge ?

pascalgrimaud:
@cbornet : ok for me, let's merge this.

PR #9603, opened 2019-04-23
Reimporting jdl

### LINKED ISSUE #8880: Reimporting a JDL file is handled as a new project
When importing a JDL file, it doesn't ask for confirmation on conflict files where changes have been made and it simply overrides them. It also commits the changes with the message `Initial application generated by JHipster-5.7.0`

When importing a JDL file, it should as for confirmations prior overriding changed files and it shouldn't automatically commit those changes.

##### **Reproduce the error**
Create a project and import a JDL. It shouls, at least, ask for confirmation when overriding the narbar file.

##### **Related issues**
none found

##### **Suggest a Fix**
implement the confirmation before overriding the files changed by the user and don't auto-commit the changes.

##### **JHipster Version(s)**

```
factur-app@0.0.0 /Users/alejandro/IdeaProjects/test_comment
└── generator-jhipster@5.7.0

```

java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)

git version 2.16.2

node: v8.11.3

npm: 6.4.1

yeoman: 2.0.5

Docker version 18.09.0, build 4d60db4

docker-compose version 1.23.1, build b02f1306

INFO! Congratulations, JHipster execution is complete!

##### **JHipster Version(s)**
5.7.0

Shaolans:
Fix #8880 

### DIFF
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -268,12 +268,6 @@ class JDLProcessor {
     getConfig() {
         if (jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
             logger.info('Found .yo-rc.json on path. This is an existing app');
-            if (!Object.prototype.hasOwnProperty.call(this.options, 'interactive')) {
-                this.options.interactive = true;
-            }
-            if (!Object.prototype.hasOwnProperty.call(this.options, 'skip-git')) {
-                this.options['skip-git'] = true;
-            }
             const configuration = jhipsterUtils.getAllJhipsterConfig(null, true);
             this.applicationType = configuration.applicationType;
             this.baseName = configuration.baseName;
@@ -410,6 +404,9 @@ class JDLProcessor {
         if (this.options['json-only']) {
             logger.info('Entity JSON files created. Entity generation skipped.');
             return;
+        }
+        if (_.isUndefined(this.options.interactive)) {
+            this.options.interactive = true;
         }
         try {
             logger.info(
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -225,7 +225,7 @@ const generateEntityFiles = (generator, entity, inFolder, env, shouldTriggerInst
             command,
             {
                 ...options,
-                force: !options.interactive,
+                force: options.force || !options.interactive,
                 'skip-install': !shouldTriggerInstall
             },
             done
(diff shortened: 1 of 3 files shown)


pascalgrimaud:
It works fine and fix the issue.
But I discovered another strange behavior:
- `jhipster import-jdl toto.jdl` -> the app is created with all entities
- same `jhipster import-jdl toto.jdl` -> the app is created, in interactive mode, but the entities are not included
- if I used `jhipster import-jdl toto.jdl --with-entities`, it works well

cli/import-jdl.js
```diff
@@ -268,6 +268,12 @@ class JDLProcessor {
     getConfig() {
         if (jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
             logger.info('Found .yo-rc.json on path. This is an existing app');
+            if (!Object.prototype.hasOwnProperty.call(this.options, 'interactive')) {
```
MathieuAA: why not just have 
```javascript
this.options.interactive = true;
this.options['skip-git'] = true;
```

Shaolans:
It looks like I am getting the same behavior on the master.

cli/import-jdl.js
Shaolans: I thought that we want it as the default behavior but if the flags are already defined, that means the user want to behave to a certain way.

MathieuAA: Well, as there are only two possible values and we always set them to true...

Shaolans: I am not sure to understand

MathieuAA: Ah!I think I've misunderstood. Your code set the values to `true` if the values aren't set.
I didn't think of that.

However, why not did you write `Object.prototype.hasOwnProperty.call` instead of `MY_OBJECT.hasOwnProperty`?

Shaolans: Yes ! But I have discussed with @PierreBesson about whether we should enforce these flags and it seems that we maybe should enforce the flags since they were meant for the jhipster command and not import-jdl ?

> However, why did you write Object.prototype.hasOwnProperty.call instead of MY_OBJECT.hasOwnProperty?

Just to be sure to use the right function, since an object can define a field hasOwnProperty.
(and eslint does not like [it](https://eslint.org/docs/rules/no-prototype-builtins))

deepu105: I'm not okay with this, the default is non-interactive mode and if you want to run in interactive mode you specify the flag

Shaolans:
> so maybe we can run it interactive mode when there is only one application or when there is only entities. When there are multiple apps its better to run in parallel and user can disable that using the flag. Same goes for skip-git behavior

Is not this code doing it ?
If there is only one application interactive mode will be activated since the current directory will contain a .yo-rc.json and the same goes for a jdl containing only entities.
However if there are multiple apps, the directory won't contain .yo-rc.json but rather directories of these apps. Thus not activating the interactive mode.

amatosg:
actually the default was _interactive_, it was changed in 5.7.0 without any warning to the user (something like _warning: this and this is going to happen_ and that's why I opened the issue)

deepu105:
There was no different modes before 5.7, do the only surprising change was for people who are generating entities only and thats why I suggested restoring the interactive behaviour just for that. App generation using JDL was a new feature which was intended to be non interactive by default

deepu105:
I'm gonna work on this

deepu105:
This should be ok now, I'm not changing the `skip-git` behavior as it was the behavior in previous versions as well and you can disable that if you don't want with the flag

PR #9616, opened 2019-04-25
Ability to generate a deliverable documentation for your project

sahbi-ktifa:
As a real project, we lack of a deliverable documentation. Using Asciidoctor, we can now centralized documentation and linked as Javadoc as Typedoc with it.

This aims to be delivered with the project and will help developers to have a functional documentation that will evolved as the features are added to the project.

This is clearly a new feature and has to be discussed among us before being merged. Feel free to amend it with comments or even pros/cons arguments to the feature itself.


generators/client/templates/angular/package.json.ejs
```diff
@@ -180,6 +181,11 @@
     "build": "<%= clientPackageManager %> run webpack:prod",
     "test": "<%= clientPackageManager %> run lint && jest --coverage --logHeapUsage -w=2 --config src/test/javascript/jest.conf.js",
     "test:watch": "<%= clientPackageManager %> run test <%= optionsForwarder %>--watch",
+    <%_ if (buildTool === 'gradle') { _%>
+    "typedoc": "typedoc --out build/docs/html5/typedocs src/main/webapp/app",
+    <%_ } else { _%>
```
murdos: Here you assume maven is available, but this is not always true since you can generate a client only application, with the `--skip-server` option

murdos: Sorry, forget my last comment, I read it too quickly...

generators/common/files.js
```diff
@@ -54,6 +54,17 @@ const commonFiles = {
                 }
             ]
         }
+    ],
+    documentation: [
+        {
+            path: 'src/main/asciidoc/',
+            templates: ['index.adoc', 'auth.adoc', 'db.adoc']
```
murdos: 'db.adoc' should be added conditionally: a microservice can have no database.

generators/common/templates/src/main/asciidoc/db.adoc.ejs
```diff
@@ -0,0 +1,45 @@
...
+Production database is *<%_ if (databaseType === 'sql') { _%><%= prodDatabaseType %>%><%_ } else { _%><%= databaseType %><%_ } _%>*.
+
+.Databases configuration
+|===
+|Environment |URL |Login |Password
+
+| Local development
+<%_ if (databaseType === 'sql') { _%>
+<%_ if (devDatabaseType === 'mysql') { _%>
+| url: jdbc:mysql://localhost:3306/jhipster?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC
```
murdos: I wonder if we shouldn't add this to a variable in the subgenerator, and use the variable both here and in `application-dev.yml` to be sure the documentation doesn't diverge from the code.
Also this template could be moved to the `server` subgenerator.

generators/common/templates/src/main/asciidoc/front-end.adoc.ejs
```diff
@@ -0,0 +1,10 @@
+<%_ if (clientFramework) { %>
```
murdos: This template could probably be moved to `client` subgenerator, so each subgenerator contributes to the documentation.

generators/server/templates/pom.xml.ejs
```diff
@@ -128,6 +128,7 @@
         <%_ } _%>
         <properties-maven-plugin.version>1.0.0</properties-maven-plugin.version>
         <sonar-maven-plugin.version>3.6.0.1398</sonar-maven-plugin.version>
+        <asciidoctor-maven-plugin.version>1.5.6</asciidoctor-maven-plugin.version>
```
DanielFran: @sahbi-ktifa Can you please use version 1.6.0 so it is compatible with java 8+

generators/server/templates/build.gradle.ejs
```diff
@@ -53,6 +53,7 @@ plugins {
     id "org.liquibase.gradle" version "2.0.1"
     <%_ } _%>
     id "org.sonarqube" version "2.7"
+    id "org.asciidoctor.convert" version "1.5.9.2"
```
DanielFran: @sahbi-ktifa Can you please use version 1.6.1 so it is compatible with java 8+

generators/server/templates/pom.xml.ejs
DanielFran: Or we can target to version v2.0.0-RC.1?

generators/server/templates/build.gradle.ejs
DanielFran: Or we can target to version 2.1.0?

generators/common/files.js
sahbi-ktifa: Yes indeed

generators/server/templates/pom.xml.ejs
sahbi-ktifa: I followed asciidoctor maven/gradle plugin recommendations, but yes, we should definitely stick to the latest version!

PierreBesson:
But why Asciidoctor and not markdown ? Personally I think markdown is much simpler and better known and since v6, we already do automatic linting on it.

jdubois:
+1 for @PierreBesson I'm also used to Markdown, and that's what we already have everywhere, I would rather stay with it by default.

PierreBesson:
This is a bit strange but it seems that AsciiDoctor (the tool not the format) can support markdown as well : https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#markdown-compatibility
However it might be a bit confusing for the user.

sahbi-ktifa:
The main advantage with Asciidoctor is that it can import attributes from your build tools, therefore your documentation becomes interactive and self maintained (generated date, version, referencing source code directly, ...).
Using Markdown syntax is even possible so you would not have to learn a new syntax from your usual readme editing.

However, as @murdos pointed out, this feature might be a module instead of a part of the main generator, I'm a bit puzzled on this afterall. My primary goal is that, when producing an application within an enterprise environment, you're quickly asked to provide documentation. Producing a documentation process alongside your code would be I think an ease for many developers (because we all know that writing Word documents is not fun. At all)

DanielFran:
Actually I use it too in combination with swagger (API documentation to HTML/pdf) and spring rest docs (examples API request/response)

PierreBesson:
Honestly I don't know AsciiDoc myself but I know it is a very hip and powerful technology. Maybe @mraible could comment on this. I know that he is a heavy user of it.

sahbi-ktifa:
He is the one who introduced it to me :wink: 
It'd be great if we could have your thought on this one @mraible ?

mraible:
I'm not against this idea, but I've rarely seen docs that are part of the same project as the source code of a project. Also, if the docs are generated, it's likely they'll be out of date quickly. If we add this feature, I think it should be optional.

DanielFran:
I agree with @mraible, it should be optional, so we need a prompt question... Probably it should include the API doc with swagger2markup and test example with spring rest doc.

atomfrede:
> Honestly I don't know AsciiDoc myself but I know it is a very hip and powerful technology. Maybe @mraible could comment on this. I know that he is a heavy user of it.

Asciidoc is much much more powerful than markdown, so in case we will have something lets stick with asciidoctor (e.g. source code/content includes, using antora to aggregate, or having docs-as-code approach to publish the same documentation to different outputs, which is most likely html for everyone and word/confluence for managers or so). Yesterday asciidoctorj 2.0.0 was finally released, so there will be updates to the plugins soon I guess. In fact my old swagger2markup module (https://github.com/atomfrede/generator-jhipster-swagger2markup) already setup asciidoctor pipeline in maven and gradle but not for written documentation so maybe we can build upon that, but I guess it is easier to start a new module from scratch.

PierreBesson:
Yes it should include the API docs and also maybe the Javadoc and Typedoc as for Spring RestDocs, I am not a big fan as it's tedious to maintain for minimal gain over swagger.

atomfrede:
> Yes it should include the API docs and also maybe the Javadoc and Typedoc as for Spring RestDocs, I am not a big fan as it's tedious to maintain for minimal gain over swagger.

Yes. And currently swagger2markup does not support openapi 3 yet.

murdos:
I vote for Asciidoc rather than Markdown too since it's indeed more powerful.

deepu105:
I would suggest to add it as a sub-generator so that it can be run when required and hence can be updated frequently as well without having to regenerate the project. About format, I'm fine with Asciidoc but we should stick with Markdown for the content as much as possible for simplicity.

gmarziou:
Just a quick note about asciidoctor, I just converted my team's tech docs from Word and Markdown to AsciiDoc. Recently, I have tried several node-based tools to generate html/pdf from markdown, none worked as easy and as fast as Asciidoctor. The fact that it runs on JVM (using JRuby) and has great plugins for maven and gradle makes it a much better choice for java based projects (like Spring and Hibernate). GitHub and GitLab provide a good embedded preview.

As far as simplicity goes, I don't see significant difference bteween our current README.md and equivalent [README.adoc](https://github.com/gmarziou/jhipster-sample-app/blob/master/README.adoc)

For those who want to go beyond simple docs, https://github.com/docToolchain/docToolchain project is incredible.

pascalgrimaud:
Thanks for your proposal and the PR @sahbi-ktifa 
It has been opened for too long, and not sure everyone agreed on this.
So the best would be to go to module, as suggested
That's why I'm closing this.

Issue #9548, opened 2019-04-09
Generate feign clients from OpenAPI definition files


ecostanzi:
##### **Overview of the feature request**

Allow developers to generate feign clients from OpenAPI definition files. This could be done by integrating the code of the module https://github.com/cbornet/generator-jhipster-swagger-cli created by @cbornet.

##### **Motivation for or Use Case**
Improve communication and the integration between microservices. 

##### Example
I ran `yo jhipster-swagger-cli` on a microservice and then modified the code using the maven plugin instead of the npm module. [This](https://github.com/ecostanzi/jhipster-microservice-swagger-cli/commit/245e928af3a0b5a2950ce64ad802a828ad925a9c) is how it looks like, could be a starting point.

##### Possible Implementations

1. Modify the main generator with a new property in the `.yo-rc.json`
```json
...
"openApiClients": ["petstore", "anotherClientSpec"]
...
```

2. A separate generator `openapi-cli`
```bash
$ jhipster openapi-cli
Do you want to specify a path to an existing spec? (Yes/no) 
Yes -> Where is your Swagger/OpenAPI spec (URL or path) ? / Do you want keep a copy of the spec into the workspace?
No -> create a empty yml into the resources folder
What is the name of the client? (petstore)
```

##### **Open questions and consideration**

- Will the generator allow to generate more than one client?
- Even when passing a URL to the generator, I would suggest to keep a local copy of the spec file versioned into the workspace. This way the compilation keeps working out-of-the-box. 

There are certainly other aspects I didn't consider. According to [this](https://github.com/cbornet/generator-jhipster-swagger-cli/issues/13) @cbornet and @PierreBesson have already discussed about this feature so I'd be glad to know their opinion.

##### **Related issues or PR**

Communication between microservices: https://github.com/jhipster/generator-jhipster/issues/3649
Feign client annotations: https://github.com/jhipster/generator-jhipster/issues/7756
Improve service to service communication: https://github.com/jhipster/generator-jhipster/issues/6450
https://github.com/cbornet/generator-jhipster-swagger-cli/issues/13

-   [X] Checking this box is mandatory (this is just to show you read everything)

PierreBesson:
Yes this is a good idea and something we need to push. Note that the openapi-generator could also be integrated with the cli (installed from the openapi-generator npm dependency).

This openapi codegen feature could be used to generate clients on the front-end for angular httpClient and axios for react.

PierreBesson:
[...]

PR #9623, opened 2019-04-26
New openapi-cli subgenerator for OpenApi client generation

ecostanzi:
Create a sub-generator for generating clients from OpenAPI specs. Can be invoked using `jhipster openapi-client`. Open discussion in issue #9548.

TODO:
-  [X] import code from https://github.com/cbornet/generator-jhipster-swagger-cli and create a sub-module
-  [X] migrate from jar client to node openapi-generator-cli
-  [x] add generator name for each spec in .yo-rc.json file (to allow FE client generation)
-  [x] clean client code before regenerate

Jhipster checks:


.vscode/launch.json
```diff
@@ -260,6 +260,17 @@
             },
             "cwd": "${workspaceFolder}/test-integration/samples/app-sample-dev/",
             "console": "integratedTerminal"
-        }
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "jhipster openapi-cli",
```
cbornet: Maybe just `openapi` ? We are already in a CLI so that's a bit redundant

package.json
```diff
@@ -75,6 +75,9 @@
         "tabtab": "2.2.2",
         "through2": "3.0.1",
         "uuid": "3.3.2",
+        "sync-request": "6.0.0",
+        "underscore.string": "3.3.4",
+        "@openapitools/openapi-generator-cli": "0.0.9-3.3.4",
```
PierreBesson: Please use a fixed version here as per Policy 4 https://www.jhipster.tech/policies/

.vscode/launch.json
PierreBesson: Yes +1 for just `openapi`, openapi-cli seems weird.

PierreBesson: Maybe `openapi-client` or `openapi-rest-client` or `openapi-generator`.

cli/commands.js
```diff
@@ -99,6 +99,9 @@ Example:
         argument: ['name'],
         desc: 'Create a new Spring controller'
     },
+    'openapi-cli': {
+        desc: "Generates client code from an OpenAPI/Swagger definition"
```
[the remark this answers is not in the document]
PierreBesson: The description should be more specific: maybe use the term REST client. 
BTW @murdos this subge  can also generate js client code in the future.

package.json
ecostanzi: That's actually the latest stable version for the  of the `cli-3.3.4`, check the versions here https://www.npmjs.com/package/@openapitools/openapi-generator-cli/v/0.0.9-3.3.4. 

An alternative could be to use `cli-3.3.4` (as the README suggests) but I'm afraid that the `cli-` format might point to a new `0.0.10-3.3.4` in the future.

generators/openapi-cli/index.js
```diff
@@ -0,0 +1,285 @@
...
+    prompting() {
+        if (this.options.regen) {
+            return;
+        }
+
+        const hasExistingApis = Object.keys(this.apis).length !== 0;
+        let actionList;
+        const availableDocs = [];
+        try {
+            const swaggerResources = request('GET', 'http://localhost:8080/swagger-resources', {
```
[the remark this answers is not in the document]
ecostanzi: The URL was probably meant represent a gateway exposing all the microservices specs. However I think it's hard to modify it and make it flexible enough to be used in all the scenarios (monolith, microservices w/wo gateway etc...). The user will also have the endpoint available when regenerating the client. Maybe we should remove this part, any alternatives?

.vscode/launch.json
ecostanzi: `openapi-client` is good to me

package.json
PierreBesson: Indeed it's the only available version tag ! I didn't know that they didn't support Semver. I have opened an issue on their tracker: https://github.com/OpenAPITools/openapi-generator-cli/issues/14

generators/openapi-cli/index.js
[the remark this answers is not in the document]
PierreBesson: The /swagger-resources endpoint is the standard endpoint configured by Springfox to expose the list of swagger resources URL: (see [springfox-swagger-common/ApiResourceController.java](https://github.com/springfox/springfox/blob/master/springfox-swagger-common/src/main/java/springfox/documentation/swagger/web/ApiResourceController.java#L68)). But I agree that localhost:8080 should not be hardcoded but it should be replaced by a prompt in which the user can input an URL to a JHipster app in order to auto-discover the available swagger resources.

vwiencek:
Hello, just 1 comment/question.

Implementation of REST based clients to allow inter-microservices communication (based on `@FeignClient` is very useful. I see to options to share `DTOObjects` between services:
- option 1: have a shared library with all DTOObjects to make sure we keep them in sync ;
- option 2: reimplement (or regenerate with openapi-cli) and have duplicated objects in different services.

I guess the good pattern is to keep microservices as isolated as possible, which means that we should prefer option 1 over option 2, right ?

murdos:
> 
> 
> I guess the good pattern is to keep microservices as isolated as possible, which means that we should prefer option 1 over option 2, right ?

FYI on my current project we're using a 3rd option: each microservice publish its swagger contract in Nexus, and other microservices that need to communicate with this microservice generate DTO objects at compile time from the published swagger contract.

ecostanzi:
> FYI on my current project we're using a 3rd option: each microservice publish its swagger contract in Nexus 

@murdos that's interesting. How do you generate the spec file? I have read somewhere that it can be done with an integration test invoking the springfox endpoints.

murdos:
> 
> @murdos that's interesting. How do you generate the spec file? I have read somewhere that it can be done with an integration test invoking the springfox endpoints.

That's exactly how we generate it indeed. I tried with https://github.com/kongchen/swagger-maven-plugin at build time, but the generated swagger was not reliable.

generators/openapi-cli/index.js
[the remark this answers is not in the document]
ecostanzi: These are the current prompts:

```
? Where do you want to import your OpenAPI/Swagger specification from ? (Use arrow keys)
> From a Jhipster /swagger-resources live doc endpoint
  From the api.yml spec of an local Jhipster project
  From a custom specification file or URL
```

When choosing the first option then it prompts:
```
? Enter the URL of the running Jhipster instance http://javadev.intesys.it:8078
? Select the doc for which you want to create a client (Use arrow keys)
> openapi (/api/public/api.yml)
  auth (/v2/api-docs?group=auth)
  jhipster (/v2/api-docs?group=jhipster)
  management (/v2/api-docs?group=management)
```

When launching from command line I'm thinking at something like `jhipster openapi-client --spec=http://myjhipster:8080/v2/api-docs?group=jhipster --generator=spring --save`

What do you think?

PierreBesson:
@ecostanzi Do you need this merged and released before your talk at Jhipster conf ?

ecostanzi:
@PierreBesson if the code looks good to you it would be nice! Let's see if the test pass. If so I'll add some more tests and mark the PR as ready for review.

generators/client/templates/angular/package.json.ejs
```diff
@@ -143,6 +143,7 @@
     <%_ if (protractorTests) { _%>
     "webdriver-manager": "12.1.5",
     <%_ } _%>
+    "@openapitools/openapi-generator-cli": "0.0.9-3.3.4",
```
DanielFran: Better to put 0.0.9-4.0.2

See https://github.com/jhipster/generator-jhipster/pull/9925

generators/client/templates/react/package.json.ejs
```diff
@@ -166,6 +166,7 @@ limitations under the License.
     <%_ if (protractorTests) { _%>
     "webdriver-manager": "12.1.4",
     <%_ } _%>
+    "@openapitools/openapi-generator-cli": "0.0.9-3.3.4",
```
DanielFran: Better to put 0.0.9-4.0.2

See https://github.com/jhipster/generator-jhipster/pull/9925

generators/server/templates/package.json.ejs
```diff
@@ -29,7 +29,8 @@
     <%_ otherModules.forEach(module => { _%>
     "<%= module.name %>": "<%= module.version %>",
     <%_ }); _%>
-    "generator-jhipster": "<%= jhipsterVersion %>"
+    "generator-jhipster": "<%= jhipsterVersion %>",
+    "@openapitools/openapi-generator-cli": "0.0.9-3.3.4"
```
DanielFran: Better to put 0.0.9-4.0.2

See https://github.com/jhipster/generator-jhipster/pull/9925

package.json
```diff
@@ -91,7 +93,8 @@
         "mocha": "6.1.4",
         "sinon": "7.2.5",
         "yeoman-assert": "3.1.1",
-        "yeoman-test": "1.9.1"
+        "yeoman-test": "1.9.1",
+        "@openapitools/openapi-generator-cli": "0.0.9-3.3.4"
```
DanielFran: Better to put 0.0.9-4.0.2

See https://github.com/jhipster/generator-jhipster/pull/9925

package-lock.json
```diff
@@ -44,6 +44,12 @@
             "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz",
             "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
         },
+        "@openapitools/openapi-generator-cli": {
+            "version": "0.0.9-3.3.4",
```
DanielFran: Better to put 0.0.9-4.0.2

See https://github.com/jhipster/generator-jhipster/pull/9925

generators/client/templates/angular/package.json.ejs
ecostanzi: updated to 0.0.14-4.0.2

generators/client/templates/react/package.json.ejs
ecostanzi: updated to 0.0.14-4.0.2

package.json
ecostanzi: updated to 0.0.14-4.0.2

package-lock.json
ecostanzi: updated to 0.0.14-4.0.2

ecostanzi:
If it's ok for you members of the core team I'd like to do a demo preview of this feature during the talk at the jhipster conf. Otherwise if you prefer to take your time and review it before advertising it no problem, just tell me, i have something else to show.

PierreBesson:
Or you can just show it from your branch

ecostanzi:
@PierreBesson yes that's what I meant no need to have it in the main branch

DanielFran:
@ecostanzi Can you fix conflicts and confirm when this PR can be reviewed? Thanks

PierreBesson:
Hi everyone, let's move this forward, update the dependencies it introduces and merge this if this OK for you. For me it's important as a convergence of the OpenApiGenerator and JHipster projects.

ecostanzi:
Hi, I've just rebased the branch and removed a couple of dependencies. It should be ok for now.

DanielFran:
Restarted failing test

DanielFran:
Many thanks @ecostanzi for the awesome work

generators/client/templates/react/package.json.ejs
```diff
@@ -164,6 +164,7 @@ limitations under the License.
     <%_ if (protractorTests) { _%>
     "webdriver-manager": "12.1.5",
     <%_ } _%>
+    "@openapitools/openapi-generator-cli": "0.0.14-4.0.2",
```
vishal423: @ecostanzi,
It doesn't look appropriate to include generator sub-module dependency in the generated client application. Can't it be encapsulated in the module?

ecostanzi: Hi Vishal, what do you mean by saying "encapsulated in the module"? 

The documentations suggests:
> It is recommended to install the package as development dependency, because normally you only need this dependency during the development process

https://www.npmjs.com/package/@openapitools/openapi-generator-cli#package-mode

PierreBesson: Yes for me there is no problem to add this as dev dependency. Yes it does add significantly to what you have to download. But as it is useful, I don't see the problem of it being downloaded from npm rather than maven...

vishal423: IMO, it should be added as `generator-jhipster` dependency (not under dev) and not in the client applications dev dependency. Can you check if that works?

ecostanzi: Ok I'll let you know

ecostanzi: @vishal423 I opened issue https://github.com/jhipster/generator-jhipster/issues/10510 to discuss some changes. In that case I think that the dependency should remain in the package.json of the generated project.

vishal423: @ecostanzi, I will leave it to other members to decide. For me, I am planning to create a separate module that wouldn't contain it.

PR #9624, opened 2019-04-26
Remove CustomParameterizedException

cbornet:

You can achieve the same functionality as CustomParameterizedException with a Problem exception. Eg:

```java
throw Problem.builder()
                .withStatus(Status.BAD_REQUEST)
                .with("param0", "param0_value")
                .with("param1", "param1_value")
                .build();
```
so I propose to remove CustomParameterizedException as per Policy 2.


PierreBesson:
@cbornet You should add some info in the docs here by completing this page : https://www.jhipster.tech/managing-server-errors/

DanielFran:
@cbornet Maybe you can add a test with an example?

cbornet:
That would be testing the Problem lib which already has its own tests. Not sure it has a lot of value...

DanielFran:
ok, was just thinking about the migration, so people have an example... otherwise I am fine with the documentation update that should include an example.

jdubois:
Thanks @cbornet and anyway this kind of code should be in the framework, not generated

PR #9674, opened 2019-05-06
Fix session auth with skipUserManagement

cbornet:

The fix done in #9655 should also be done for skipUserManagement apps with DB.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -25,7 +25,7 @@ import <%= packageName %>.security.jwt.*;
 <%_ } _%>
 <%_ if (authenticationType === 'session') { _%>
 
-    <%_ if (databaseType !== 'no') { _%>
+    <%_ if (!skipUserManagement) { _%>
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
@cbornet : do you think I should add new configs + builds at daily-builds project, to test config with skipUserManagement ?

cbornet:
@pascalgrimaud that would be great ! At least some "no database builds"

DanielFran:
@cbornet The "no database" changes require changes in jhipster-core?

cbornet:
It should probably a added to the available options. Note however that it doesn't make sense for the entity generation.

pascalgrimaud:
@cbornet : do e2e work with no database option ? I don't think so, as I can still see some account / user management items when choosing no database.
I can try to fix them if you confirm, but what is the difference between no database and skipUserManagement ?

cbornet:
I'm working on this. We generate the front end account module for skipUserManagement although there's no account endpoint.
"No database" implies skipUserManagement but not the inverse.

cbornet:
And yes, e2e probably don't work atm.

pascalgrimaud:
Thanks for clarification :)
I will focus on adding the builds so

PR #9709, opened 2019-05-08
Add a no client option

### LINKED ISSUE #9708: No client option

Since we now have a question for the front-end framework we can also propose a `No client` choice that has the same effect as `--skipClient`

I think JHipster can be used for every type of applications, even monolithic ones without front-end

cbornet:
Fix #9708

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -439,6 +439,9 @@ module.exports = class extends BaseBlueprintGenerator {
                 if (this.configOptions.clientFramework) {
                     this.clientFramework = this.configOptions.clientFramework;
                 }
+                if (this.configOptions.skipClient) {
+                    this.skipClient = this.configOptions.skipClient;
+                }
                 if (this.configOptions.uaaBaseName !== undefined) {
                     this.uaaBaseName = this.configOptions.uaaBaseName;
                 }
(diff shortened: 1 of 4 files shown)


generators/server/index.js
```diff
@@ -439,6 +439,9 @@ module.exports = class extends BaseBlueprintGenerator {
                 if (this.configOptions.clientFramework) {
                     this.clientFramework = this.configOptions.clientFramework;
                 }
+                if (this.configOptions.skipClient) {
+                    this.skipClient = this.configOptions.skipClient;
```
murdos: Is there any use yet? Or is this just for future use, and to be sure the information is available in this subgenerator?

cbornet: Yes it's important as some files of the server are different if skipClient is set (pom.xml, README, ...)

murdos: Ok, but this assignation is already done in `this.setupServerOptions(this)`: https://github.com/jhipster/generator-jhipster/blob/master/generators/server/index.js#L61

cbornet: It needs to be done again at `configuring` phase after the `prompting` phase. You can try to remove it and you'll see the server will be generated with front-end related code.

murdos: Ok, I see. The `prompting` phase of the `client` subgen will indeed happen between the different phases of the `server` subgen.

PR #9719, opened 2019-05-09
Fixed password menu icon from 'clock' to 'lock' and completed 'uz-cyr' language support

javasparx:


javasparx:
It was always LOCK icon in password menu item. Reject if I am wrong.

pascalgrimaud:
you're totally right !
so funny, we didn't see it all this time !

Could you do the same for React too plz ?
And if possible, for Vue.js too -> if not, I'll do it at https://github.com/jhipster/jhipster-vuejs

javasparx:
I think I did the fix for React too, but please check it, because we do not use React at all.

clement26695:
I think you have to replace faClock by faLock in this file too :  generators/client/templates/react/src/main/webapp/app/config/icon-loader.ts.ejs

javasparx:
@clement26695 , I did as you suggested.

pascalgrimaud:
thanks @javasparx !

PR #9755, opened 2019-05-17
remove unused and confusing import

yelhouti:

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/shared.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/shared.module.ts.ejs
@@ -18,7 +18,6 @@
 -%>
 import { NgModule, CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import {
-    <%=angularXAppName%>SharedLibsModule,
     <%=angularXAppName%>SharedCommonModule,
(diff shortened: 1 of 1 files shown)


wmarques:
@yelhouti Thanks for your contribution.
Why this import were unusued for you?

pascalgrimaud:
@wmarques : we need to check and test this closely. It is probably here for a good reason.

yelhouti:
@pascalgrimaud the module is already imported in shared.module that imports this.
Also I feel that other imports are also a mess, since all global (application level) module imports are re-imoprted in each module...

wmarques:
@yelhouti Ok let me check this :) 
The purpose to have a SharedModule is to re-import it in each feature module to access to shared Components, Pipes or Directives so that's a normal thing.
But I need to check if we import other modules in each feature module.

wmarques:
@yelhouti LGTM. Can you also remove the `forRoot` static methods from `SharedModule` and `SharedLibs` ? It should work without them.

yelhouti:
@wmarques I removed only the very confusing import, I think the hwole import should reviewed, keeping only what's is strictly required in the module reimported for each entity.

wmarques:
Ok let's merge this then and I'll try to clean that.
I agree with you that we have too much imports, we should only have one simple `SharedModule`.

PR #9756, opened 2019-05-17
Allow all redirect URIs and CORS origin for keycloak

cbornet:

For instance, I use swagger-ui watcher to do hot reload of swagger-ui and which is served on localhost:8000.
This shouldn't be a security issue since the docker config shall only be used in dev (you would have other security issues in prod such as the secret being known)

### DIFF
--- a/generators/server/templates/src/main/docker/config/realm-config/jhipster-realm.json.ejs
+++ b/generators/server/templates/src/main/docker/config/realm-config/jhipster-realm.json.ejs
@@ -820,18 +820,12 @@
       "clientAuthenticatorType": "client-secret",
       "secret": "web_app",
       "redirectUris": [
-        "http://localhost:<%= serverPort %>/*",
-        "https://localhost:<%= serverPort %>/*",
-        "http://localhost:8100/*",
-        "http://127.0.0.1:8761/*",
-        "http://localhost:9000/*"
+        "http://localhost:*",
+        "https://localhost:*",
+        "http://127.0.0.1:*",
+        "https://127.0.0.1:*"
       ],
-      "webOrigins": [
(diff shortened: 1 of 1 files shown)


mraible:
This might make things easier, but the OAuth 2.0 spec discourages the use of wildcards in redirect URIs. https://tools.ietf.org/html/draft-ietf-oauth-security-topics-02#page-5

In fact, Okta doesn't support wildcards because of this.

If we make this change, we should document it in https://www.jhipster.tech/security/#-oauth2-and-openid-connect and be aware that Keycloak might become more spec compliant in a future release.

cbornet:
I know this is discouraged but the Keycloak docker-compose file shall only be used as is in dev only anyway

mraible:
True. What's your motivation for making this change? I like having the list so I know what I need to configure on other IdPs.

cbornet:
I serve the OpenApi from swagger-ui watcher on localhost:8000. With the security definitions defined in the api.yml, I'm able to do the OIDC login and attach the bearer token in the swagger-ui requests. But for it to work, localhost:8000 must be allowed.

mraible:
I’d prefer if we just added it as one additional URI. But I’m open to what others think.

cbornet:
We also generate a docker-compose for swagger-editor that binds to port 7742. So this one should also be allowed. We could of course put the full list of ports but I thought it was easier to just allow all. There's also the case where you use 127.0.0.1 instead of localhost.

cbornet:
Maybe we could allow "localhost:*" ?

cbornet:
@mraible I've changed to aceepting all `localhost:*` redirect URIs. It seems to work well.

mraible:
OK, fine. 😊

PR #9765, opened 2019-05-19
Only show health details for admin user

codelano:
Health details might contain some sensitive information so it should be showed to admin user only.

Example of health details response:

```
{
    "status": "UP",
    "details": {
        "diskSpace": {
            "status": "UP",
            "details": {
                "total": 254930841600,
                "free": 29542526976,
                "threshold": 10485760
            }
        },
        "mongo": {
            "status": "UP",
            "details": {
                "version": "4.0.6"
            }
        }
    }
}
```

### DIFF
--- a/generators/server/templates/src/main/resources/config/application.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application.yml.ejs
@@ -116,6 +116,7 @@ management:
     endpoint:
         health:
             show-details: when-authorized
+            roles: "ROLE_ADMIN"
         jhimetrics:
             enabled: true
     info:


pascalgrimaud:
You need to check if there is no impact with Microservices / Kubernetes.

jdubois:
@pascalgrimaud yes for me we didn't put security there on purpose, as many monitoring systems do not have the possibility to authenticate (either they don't support our authentication mechanism, or it's too complicated to give them a password...). There shouldn't be a security risk in telling people that the application is up, as if you can access it you already know it's up :-)

cbornet:
But the details such as mongo version or free disk space is more sensitive...

cbornet:
If I understand welll this PR, you'll still be able to call the health endpoint from an unauthenticated user but you won't get the details. LGTM.

clement26695:
I think we have the same issue with the prometheus endpoint, that is not protected. @PierreBesson tried to find a way to protect it, but he didn't find a satisfactory way to do it. You can see #8946 for more info

codelano:
> If I understand welll this PR, you'll still be able to call the health endpoint from an unauthenticated user but you won't get the details. So LGTM.

Correct. For other roles, but ROLE_ADMIN, you would get that:
```
{
    "status": "UP",
}
```

jdubois:
Oh sorry: I thought this was what we were actually doing! So totally LGTM

pascalgrimaud:
LGTM. Just tested and it works well
Thanks and congrats for your 1st contribution @codelano !

PierreBesson:
Thank you and congrats for your first contribution @codelano !

PR #9783, opened 2019-05-23
Refactor command suggestion logic

### LINKED ISSUE #9782: Improper command suggestion logic

As per now the command suggestion logic doesn't work as intended. 

It happens as `logger.error()` is being used to log the message which includes a `process.exit()` statement and as a result the control flow doesn't reach the command suggestion logic checks.

https://github.com/jhipster/generator-jhipster/blob/fc27a811fd16fb8ce44a7d66515b274e4fa049b6/cli/utils.js#L49

https://github.com/jhipster/generator-jhipster/blob/fc27a811fd16fb8ce44a7d66515b274e4fa049b6/cli/utils.js#L91

jamesgeorge007:
Closes #9782 

Refactored the command suggestion logic such that it suggests matching commands as intended. 

### DIFF
--- a/cli/utils.js
+++ b/cli/utils.js
@@ -46,7 +46,7 @@ const error = function(msg, trace) {
     if (trace) {
         console.log(trace);
     }
-    process.exit(1);
+    return; // eslint-disable-line
 };
 
 const init = function(program) {
@@ -88,7 +88,7 @@ const initHelp = (program, cliName) => {
     });
 
     program.on('command:*', name => {
-        logger.info(`${chalk.yellow(name)} is not a known command. See '${chalk.white(`${cliName} --help`)}'.`);
+        logger.error(`${chalk.yellow(name)} is not a known command. See '${chalk.white(`${cliName} --help`)}'.`);
 
         const cmd = Object.values(name).join('');
         const availableCommands = program.commands.map(cmd => {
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
@pascalgrimaud @atomfrede Thoughts :thinking:

pascalgrimaud:
LGTM but to be sure, we need to test if it impacts blueprint or not

jamesgeorge007:
@pascalgrimaud How are we gonna proceed with writing unit tests :thinking: 
I can find a `cli-utils-spec.js` within the `test `directory. Can you throw light on the intended procedure.

pascalgrimaud:
@jamesgeorge007 : yes it's this file `cli-utils-spec.js`. My comment was for the one who will review and merge (it could be me). We need to test if the [blueprint project](https://github.com/jhipster/generator-jhipster-blueprint works well with this change. So it's on our side, don't worry.

Oh, and I didn't notice it's your 1st contribution ! So big congrats !
Just give us some time to review/test it ;)

cli/utils.js
```diff
@@ -46,7 +46,7 @@ const error = function(msg, trace) {
     if (trace) {
         console.log(trace);
     }
-    process.exit(1);
+    return; // eslint-disable-line
```
deepu105: @jamesgeorge007  I don't get the point of this, how is the process killed on error now? this was to ensure we don't have any handing process/subprocess

jamesgeorge007: @deepu105 It was specific to the case here. I guess we could replace `logger.error()` with `logger.info()` :thinking:

jamesgeorge007: What would you suggest?

deepu105: Keep the process exit and do console log directly for the usage in question

jamesgeorge007:
@deepu105 Have a look at #9864

Issue #9095, opened 2019-01-18
Support java.time.Duration in entity creation


massimosiani:
##### **Overview of the feature request**

Add support for the java.time.Duration type as a field of an entity.

##### **Motivation for or Use Case**

Hibernate 5 natively supports this type, and little effort is necessary for supporting it in Mongo (a couple of converters are already defined in a companion JHipster project).

##### **Related issues or PR**

I did the homework but didn't find anything.

I am working on a PR for this. Help wanted for React! And feedback is welcome!

-   [x] Checking this box is mandatory (this is just to show you read everything)

massimosiani:
Hi,
let me give an update.
This PR is necessary for adding this feature to the generator:
https://github.com/jhipster/jhipster/pull/159

Moreover, if anyone would like to help me on a PR for the generator, here's the forked repository, there's some work on the backend and on the Angular part done:
https://github.com/massimosiani/generator-jhipster
branch: support-for-duration
Here's the diff with the master branch:
https://github.com/jhipster/generator-jhipster/compare/master...massimosiani:support-for-duration

Please upvote :)
Thank you!

DanielFran:
@massimosiani The PR https://github.com/jhipster/jhipster/pull/159 has been merged in the bom project.
Do you plan to make the PR in generator soon?

massimosiani:
Hi @DanielFran sorry for the delay.
I hadn't much time to complete the work on the generator, but I will make progress this week.
Thank you!

DanielFran:
@massimosiani Do you have news? It is just to understand if it can be included in v6 on time...

massimosiani:
Hi @DanielFran, I was facing an issue with a couple of tests.
A PR is on its way!

pascalgrimaud:
@massimosiani : for your information, I encountered some issue:
- Angular: updating an existing entity -> the field is empty
- React: can't create an entity with Duration

Did you add the support for React too ? I'll try to confirm these issue and open a ticket for that if needed

PR #9794, opened 2019-05-25
UUID field type for sql databases

murdos:
Requires https://github.com/jhipster/jhipster-core/pull/331

### DIFF
--- a/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
+++ b/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
@@ -165,8 +165,7 @@
                   file="config/liquibase/data/<%= entityTableName %>.csv"
                   separator=";"
                   tableName="<%= entityTableName %>"
-                  primaryKey="id"
-                  context="dev">
+                  primaryKey="id">
             <column name="id" type="numeric"/>
             <%_ for (idx in fields) {
                 let loadColumnType = 'string';
--- a/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
+++ b/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
@@ -172,8 +172,10 @@
                 let columnType = fields[idx].columnType;
                 if (columnType === 'integer' || columnType === 'bigint' || columnType === 'double' || columnType === 'decimal(21,2)' || columnType === '${floatType}') {
                     loadColumnType = 'numeric';
-                } else if (columnType === 'date' || columnType === 'datetime') {
+                } else if (columnType === 'date') {
                     loadColumnType = 'date';
+                } else if (columnType === 'datetime') {
+                    loadColumnType = 'datetime';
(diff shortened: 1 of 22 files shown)


generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -159,11 +161,11 @@
           Spring Boot configuration key.
     -->
     <changeSet id="<%= changelogDate %>-1-data" author="jhipster" context="faker">
-        <loadData
+        <loadUpdateData
```
murdos: This is required because of https://liquibase.jira.com/browse/CORE-3208 : `type="computed"` doesn't work currently with `loadData`

generators/server/templates/src/main/resources/config/liquibase/master.xml.ejs
```diff
@@ -20,13 +20,18 @@
...
+    <%_ if (prodDatabaseType === 'postgresql') { _%>
     <property name="clobType" value="longvarchar" dbms="h2"/>
     <%_ } else { _%>
     <property name="clobType" value="clob" dbms="h2"/>
     <%_ } _%>
-    
     <property name="clobType" value="clob" dbms="mysql, oracle, mssql, mariadb, postgresql"/>
+    <property name="uuidType" value="uuid" dbms="h2"/>
+    <%_ if (prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb') { _%>
+    <property name="uuidType" value="binary(16)" dbms="<%= prodDatabaseType %>"/>
```
murdos: `binary(16)` is the most efficient way to store UUID in MySQL, and also what Hibernate does by default.
Unfortunately this is not how Liquibase handle it by default: https://liquibase.jira.com/browse/CORE-3081

murdos:
> 
> 
> I don't get why you've added Duration fields in the JSON test files

Adding UUID is very similar to what has be done recently with Duration (https://github.com/jhipster/generator-jhipster/issues/9095), so I'm fixing things in the same time. Integration-tests data were incomplete.

murdos:
This MR is ready for review and integration. It only needs a release of jihpster-core and an update of the dependency in generator-jhipster.

MathieuAA:
@murdos roger that

pascalgrimaud:
Very good job @murdos !

murdos:
Thanks! It has been more difficult and burdensome than what I originally planned 😂

poornerd:
cool @murdos - I have been trying to use it with the JDL importer - does that work too?
```
entity MappingType {
	id UUID,
	typeCode String required unique,
    name String required
}
```
Because it is generating Long's for me :-(

MathieuAA:
JHipster having its own long id type, JCore ignores such fields when parsing a JDL content. This has nothing to do with this PR.

murdos:
@MathieuAA : good catch!

@poornerd : if you try with the following JDL, it will work (you can name the field as you want, but not `id`):
```
entity MappingType {
	uuid UUID,
	typeCode String required unique,
        name String required
}
```

poornerd:
@MathieuAA and @murdos thanks for the tips

PR #9818, opened 2019-05-29
react: Fix intermittent e2e failures

vishal423:
An attempt to fix intermittent failures seen on Azure for default react build. It's still not full proof as the presence of an overlay can cause an element to be not clickable.

### DIFF
--- a/generators/client/templates/react/src/test/javascript/e2e/modules/account/account.spec.ts.ejs
+++ b/generators/client/templates/react/src/test/javascript/e2e/modules/account/account.spec.ts.ejs
@@ -34,6 +34,7 @@ import {
   getModifiedDateSortButton,
 <%_ } _%>
   waitUntilDisplayed,
+  waitUntilClickable,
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
Thanks a lot @vishal423 !

Maybe we should try to do the same for Vue.js as we have the same problem since weeks: https://github.com/jhipster/jhipster-vuejs/issues/285
I tested a lot of thing but didn't manage to resolve the problem... 

cc @sahbi-ktifa

vishal423:
I have never used vue, however, will check if similar change can be applied there. As I said, we need to assert this change by monitoring further builds to see if they still fail on this as it's not full proof. 

There is an open issue https://github.com/angular/protractor/issues/2313 on the protractor (opened for 3+yrs) that should help to handle such scenarios gracefully.

pascalgrimaud:
your help would be really welcome on Vue.js :)

PR #9834, opened 2019-05-31
improve security headers

### LINKED ISSUE #9549: Improve Security Headers

We should improve our security headers to show we care about security.

jhipster.tech gets an F: https://securityheaders.com/?q=jhipster.tech&followRedirects=on
21-points.com gets a C: https://securityheaders.com/?q=21-points.com&followRedirects=on

In my experience with raibledesigns.com, a lot of the headers can be configured at the server level (rather than the app level), so it might just be a matter of documentation. 

raibledesigns.com gets an A: https://securityheaders.com/?q=raibledesigns.com&followRedirects=on

To show we care about security on the JHipster project.

Related: You can configure a [Content Security Policy with Spring Security](https://developer.okta.com/blog/2018/07/30/10-ways-to-secure-spring-boot#5-use-a-content-security-policy-to-prevent-xss-attacks).

atomfrede:
by setting `content-security`, `feature-policy`, `referrer-policy` header and deny embedding in an iframe.

Due to angular or an ng plugin we need to allow unsafe inline and eval (same for inline styles). Did not check if that is required for react or vue. Before the `X-Frame-Options` where disabled at all, now it is set to `deny` so embedding is not allowed by default.

See #9549 for the open question (thats why it is currently in draft mode). I tested also with local development via npm and gradle it works fine and I couldn't see any problems.

These are the current results for security headers https://securityheaders.com/?q=https%3A%2F%2Fgradleheroku.herokuapp.com%2F&followRedirects=on

And this is the deployed application https://gradleheroku.herokuapp.com/

closes #9549

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -88,6 +88,7 @@ import org.springframework.security.web.csrf.CsrfFilter;
 <%_ if (authenticationType === 'jwt' && applicationType !== 'microservice') { _%>
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 <%_ } _%>
+import org.springframework.security.web.header.writers.ReferrerPolicyHeaderWriter;
 <%_ if (applicationType !== 'microservice') { _%>
 import org.springframework.web.filter.CorsFilter;
 <%_ } _%>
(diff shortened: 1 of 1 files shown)


atomfrede:
@pascalgrimaud @jdubois @mraible If anyone of you has some time to have a look at this that would be great and would be nice to show on jhipster conf :)

mraible:
@atomfrede Looks good to me! I think you need to take it out of draft mode for CI to run. The only comment I have is the formatting. I've seen several Spring Security config examples that use the `and()` and outdent it after so things look better. For example: https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#oauth2resourceserver-sansboot

atomfrede:
@mraible Thx, now the format is the same as in the other parts fo the configuration.

pascalgrimaud:
let's wait the CI !

Issue #8920, opened 2018-12-02
React UI has issues in a microservices environment


mraible:
##### **Overview of the issue**

I created a microservices architecture using a JDL that I normally use for Angular apps. I tried it with React today and there's two main issues I found:

1. Adding a new blog results in an error.

```
blog-app_1                    | 23:11:07.406 [XNIO-2 task-13] ERROR org.hibernate.internal.ExceptionMapperStandardImpl - HHH000346: Error during managed flush [Validation failed for classes [com.okta.developer.blog.domain.User] during persist time for groups [javax.validation.groups.Default, ]
blog-app_1                    | List of constraint violations:[
blog-app_1                    | 	ConstraintViolationImpl{interpolatedMessage='must not be null', propertyPath=login, rootBeanClass=class com.okta.developer.blog.domain.User, messageTemplate='{javax.validation.constraints.NotNull.message}'}
blog-app_1                    | ]]
blog-app_1                    | 2018-12-02 23:11:07.410 ERROR 1 --- [ XNIO-2 task-13] o.z.problem.spring.common.AdviceTrait    : Internal Server Error
blog-app_1                    |
blog-app_1                    | org.springframework.transaction.TransactionSystemException: Could not commit JPA transaction; nested exception is javax.persistence.RollbackException: Error while committing the transaction
blog-app_1                    | 	at org.springframework.orm.jpa.JpaTransactionManager.doCommit(JpaTransactionManager.java:545)
blog-app_1                    | 	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:746)
blog-app_1                    | 	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:714)
blog-app_1                    | 	at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:532)
blog-app_1                    | 	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:304)
blog-app_1                    | 	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:98)
blog-app_1                    | 	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
[...]

PR #9845, opened 2019-06-03
fix relationship with user with oauth2 and microservices

clement26695:
Start fixing #8920

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
@@ -247,11 +247,21 @@ export const createEntity: ICrudPutAction<I<%= entityReactName %>> = entity => <
     return result;
 };
 
-export const updateEntity: ICrudPutAction<I<%= entityReactName %>> = entity => async dispatch => {
+export const updateEntity: ICrudPutAction<I<%= entityReactName %>> = entity => <% if (isEntityInMicroserviceSavingUserSnapshot) { %>async (dispatch, getState) => {
+  const entityWithFullUser = {
+    ...entity,
+    user: getState().userManagement.users.filter(user => user.id === entity.user.id)[0]
+  };
+
+  const result = await dispatch({
+    type: ACTION_TYPES.UPDATE_<%= entityActionName %>,
+    payload: axios.put(apiUrl, cleanEntity(entityWithFullUser))
+  });
+  <% } else { %>async dispatch => {
   const result = await dispatch({
     type: ACTION_TYPES.UPDATE_<%= entityActionName %>,
     payload: axios.put(apiUrl, cleanEntity(entity))
-  });
+  });<% } %>
   <%_ if (pagination !== 'infinite-scroll') { _%>
   dispatch(getEntities());
   <%_ } _%>
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
@@ -218,15 +218,33 @@ export const getEntity: ICrudGetAction<I<%= entityReactName %>> = id => {
   };
 };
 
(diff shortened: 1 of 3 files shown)


clement26695:
It would need an update in react-jhipster to update the ICrudPutAction<IEntitty>

pascalgrimaud:
@clement26695 : don't hesitate to ping me if you need a release of react-jhipster

clement26695:
I had to update the reducer test too. 

@deepu105, would it be possible to review this PR ? I'm not sure of how to use redux-thunk and its best practices...

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
```diff
@@ -218,22 +218,50 @@ export const getEntity: ICrudGetAction<I<%= entityReactName %>> = id => {
...
+    applicationType === 'gateway' &&
+    locals.microserviceName &&
+    authenticationType === 'oauth2' &&
+    hasUserField &&
+    dto === 'no';
+%>
+export const createEntity: ICrudPutAction<I<%= entityReactName %>> = entity => <% if (isEntityInMicroserviceSavingUserSnapshot) { %>async (dispatch, getState) => {
+    const entityWithFullUser = {
+      ...entity,
+      user: getState().userManagement.users.filter(user => user.id === entity.user.id)[0]
```
deepu105: This doesn't ensure the latest user detail as the state could be of before an update event from another instance. It might be better to do this on the server side, else trigger a call to fetch the user detail before this and then get the state

clement26695: As the generated entity is in a microservice, I think it's best to trigger a call to fetch the user detail (I don't think sure the microservice would know where to call for the user info).

But I was thinking : we create a table both in the gateway and the microservice with the user information : how de we keep these tables synchronized ? Is it a good practice to duplicate these tables ?

cbornet: I don't like it very much either. I think we shouldn't have relationships to users in microservices.

deepu105: Yes, I don't see the point in allowing a relationship to an entity which is in another service as well, this also creates coupling. It's the same as a microservice having a relationship to an entity in another microservice. I thought we didn't allow this earlier in the days, I might be wrong. I would vote to limit this. If someone needs such a relationship it needs to be handled manually IMO

clement26695: I agree. Let's close this PR if everyone is ok.

ping @mraible as he was the one raising the issue in #8920

mraible:
Looking at the issue you referenced, it seems strange that everything works in Angular, but not in React. It seems you’re closing this issue saying “there’s bugs in the React impl, but we don’t care”.

clement26695:
I just closed the PR, as my fix didn't seem good enouch, but I still can reopen it.

However, I still wonder if it does really works in the long term in Angular ? How do you keep your tables synchronized between the microservice and the gateway, as there is a coupling between the two ? Is there some kind of message broker to modify one database if the other one is modified ?

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.reducer.ts.ejs
vishal423: > As the generated entity is in a microservice, I think it's best to trigger a call to fetch the user detail (I don't think sure the microservice would know where to call for the user info).

micro-service should get logged-in user details from Oauth2 token claims. Since, user details are managed on the IDP, we should limit data stored in gateway / micro-services to avoid sync issues.

mraible:
The ability to create relationships to `User` currently only works when using microservices and OAuth 2.0. The `jhi_user` table is read-only when using OAuth 2.0, so keeping tables in sync isn't that big of a deal IMO. I don't believe JDL prevents you from creating relationships with `User` when using JWT or UAA, but IIRC, the code doesn't compile.

PR #9879, opened 2019-06-10
Fix for the default CSS not loading on index.html

ronswanson730:
Hi Guys,

I apologize if I get the below wrong as I am a bit new to contributing. I encountered a small issue when generating a jhipster react app in that the css file is not loaded because the folder is incorrect. "content" rather than "static".

For your review and perhaps approval.


ronswanson730:
I see.. I went with updating index.html as I saw the manifest.webapp was referencing content.

PR #9886, opened 2019-06-10
use 'a user' instead of 'an user'

ctamisier:


mraible:
The build seems to be failing because it can't make a connection to a process running on port 5000. https://travis-ci.org/jhipster/generator-jhipster/jobs/543879939#L3207

PR #9898, opened 2019-06-11
Use official adoptopenjdk docker image now that it is available from docker hub

PierreBesson:
Adoptopenjdk have been approved by the docker hub as official images now, see https://github.com/docker-library/official-images/pull/5710

However alpine is not offered as it is not recommended for now (see the PR). IMO if the people from the adoptopenjdk project don't recommend using those images, then we refrain from using them for now.


pascalgrimaud:
Just manually tested and it works well

PR #9916, opened 2019-06-14
Clean needle-api unit tests

murdos:
- Don't run each test twice with blueprint 'generator-jhipster-myblueprint' and 'myblueprint'
- Cleaned contexts (rename, regroup, delete empty ones, ...)

### DIFF
--- a/test/needle-api/needle-client-angular.spec.js
+++ b/test/needle-api/needle-client-angular.spec.js
@@ -76,154 +76,117 @@ const mockBlueprintSubGen = class extends ClientGenerator {
     }
 };
 
-describe('needle API SCSS: JHipster client generator with blueprint', () => {
-    const blueprintNames = ['generator-jhipster-myblueprint', 'myblueprint'];
-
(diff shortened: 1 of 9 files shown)


avdev4j:
hi @murdos 
I think `const blueprintNames = ['generator-jhipster-myblueprint', 'myblueprint'];` is used to test with both fullname and shortName, so we can resolve blueprint with 2 ways.

Doing this twice for each needle is not probably a good idea, but we need to be sure we keep testing this feature.

murdos:
> 
> hi @murdos
> I think `const blueprintNames = ['generator-jhipster-myblueprint', 'myblueprint'];` is used to test with both fullname and shortName, so we can resolve blueprint with 2 ways.

Yes, I can confirm it, since I'm the one who wrote the first blueprint unit tests that included this array to test both names. So this feature is already tested in [all blueprint tests](https://github.com/jhipster/generator-jhipster/tree/master/test/blueprint), no need to it again and again.

avdev4j:
Great !

Thank you ;)

PR #9918, opened 2019-06-15
Re-enable unit tests for upgrade subgenerator

murdos:
The target version is now forced, it should prevent issue when beta releases are done

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #9937, opened 2019-06-18
Support multiple blueprints

pvliss:
Add support for multiple blueprints as discussed in https://github.com/jhipster/generator-jhipster-blueprint/issues/31.

Here are the key points of the change:
- Added new option `--blueprints` which takes a comma separated list of blueprints, e.g. using cli:  `jhipster --force --no-insight --skip-checks --with-entities --from-cli --blueprints kotlin,vuejs`. The reason for not using the existing `blueprint` is that I have also removed the `blueprintVersion` from `yo-rc.json` which is mainly used by the upgrade sub-gen and converted the blueprint config property from a simple string to an array. The `blueprint` option and config property are still read in order to preserve backwards compatibility but in the case of the app sub-gen a warning is emitted to switch to using `blueprints` instead. e.g: So now instead of having:
```
        "blueprint": "generator-jhipster-kotlin",
        "blueprintVersion": "0.8.1",
```
we now have:
```
    "blueprints": [
      {
        "name": "generator-jhipster-kotlin",
        "version": "0.8.1"
      },
      {
        "name": "generator-jhipster-vuejs",
        "version": "1.0.0"
      }
    ]
```
- The `blueprint` option is still valid and will remain as is for the cli commands.
- The upgrade sub-gen has been updated to support multiple blueprints as well. The option `target-blueprint-version` has been replaced with `target-blueprint-versions` in which one needs to define both the blueprint name and version sperated by @, e.g in the form `kotlin@0.8.0`. Multiple distinct versions can be supplied here as well.
- Currently the blueprints are composed in their declaration order and they are all executed in sequence if they exists. No checking is done. Its also 

Few more notes:
- Tested with official Kotlin and VueJS blueprints and it seems to work well.(There is an issue with the Kotlin one but will shortly submit a PR to fix it).
- Most of the tests have been intentionally left with the old option `blueprint` to make sure nothing is broken but we can change them and add unit tests to check backwards compatibility if we agree on this change. We could also keep around both the `blueprint` and `blueprints` options and just keep the change in the config file.


deepu105:
Great I'm adding a bounty on this.

deepu105:
Let me know when ready and i'll review/test

pvliss:
@deepu105 Well it already works and all unit and CI tests pass. Added in draft mode because its a rather big change and wanted to trigger some discussion first. e.g. do you agree with the `blueprints` flag?

cli/utils.js
```diff
@@ -194,9 +194,11 @@ const createYeomanEnv = () => {
     Object.keys(SUB_GENERATORS)
         .filter(command => !SUB_GENERATORS[command].cliOnly)
         .forEach(generator => {
-            if (SUB_GENERATORS[generator].blueprint) {
-                /* eslint-disable prettier/prettier */
-                env.register(require.resolve(`${SUB_GENERATORS[generator].blueprint}/generators/${generator}`), `${CLI_NAME}:${generator}`);
+            if (SUB_GENERATORS[generator].blueprints && SUB_GENERATORS[generator].blueprints.length > 0) {
```
deepu105: Not sure if this should be changed to `blueprints` and iterated. AFAIK this flag is set on a custom command coming from any blueprint so the actual merge needs to be done in `commands.js` and not here. This would not work i think

generators/app/index.js
```diff
@@ -156,9 +156,14 @@ module.exports = class extends BaseGenerator {
...
-        // This adds support for a `--blueprint` flag which can be used to specify a blueprint to use for generation
+        // NOTE: Deprecated!!! Use --blueprints instead
         this.option('blueprint', {
-            desc: 'Specify a generator blueprint to use for the sub generators',
+            desc: 'Specify one or more generator blueprints to use for the sub generators',
+            type: String
+        });
+        // This adds support for a `--blueprints` flag which can be used to specify one or more blueprints to use for generation
+        this.option('blueprints', {
+            desc: 'Specify one or more generator blueprints to use for the sub generators',
```
deepu105: We should add an example here, also I think yeomen have a List type if I'm not wrong

generators/app/index.js
```diff
@@ -189,8 +194,44 @@ module.exports = class extends BaseGenerator {
 
         this.withEntities = this.options['with-entities'];
         this.skipChecks = this.options['skip-checks'];
-        const blueprint = jhipsterUtils.normalizeBlueprintName(this.options.blueprint || this.config.get('blueprint'));
-        this.blueprint = this.configOptions.blueprint = blueprint;
+
+        // check for old single blueprint declaration
+        let blueprint = this.options.blueprint;
```
deepu105: I dont think we need to do many checks. this can be simplified to
```suggestion
        let blueprints = this.options.blueprints || '';
        let blueprint = this.options.blueprint;
        if (blueprint) {
            this.warning('--blueprint option is deprecated. Please use --blueprints instead');
            if (!blueprints.split(',').includes(blueprint)) {
                blueprints = `${blueprint},${blueprints}
            }
        }
        if (!blueprints) {
            // load blueprints from config file
            const blueprintName = this.config.get('blueprint');
            if (blueprintName) {
                const blueprintVer = this.config.get('blueprintVersion') || 'latest';
                blueprint = {
                    name: blueprintName,
                    version: blueprintVer
                };
            }

            blueprints = this.config.get('blueprints') || [];
            if (blueprint) {
                blueprints.unshift(blueprint);
            }
        }
```
```

File: generators/app/index.js [hunk 4]
```diff
@@ -254,11 +295,16 @@ module.exports = class extends BaseGenerator {
...
+                if (this.blueprints && this.blueprints.length > 0) {
+                    this.blueprints.forEach(blueprint => {
+                        blueprint.version = this.findBlueprintVersion(blueprint.name) || 'latest';
+                    });
+
                     // Remove potential previous value to avoid duplicates
-                    this.otherModules = this.otherModules.filter(module => module.name !== this.blueprint);
-                    this.otherModules.push({ name: this.blueprint, version: this.blueprintVersion || 'latest' });
+                    this.otherModules = this.otherModules.filter(module => {
+                        return this.blueprints.findIndex(bp => bp.name === module.name) === -1;
```
deepu105: why is return needed? it can use shorter syntax as earlier

generators/generator-base.js
```diff
@@ -341,7 +341,7 @@ module.exports = class extends PrivateBase {
                     fullPath +
                     chalk.yellow('. Reference to ')}npm dependency (name: ${name}, version:${version})${chalk.yellow(' not added.\n')}`
             );
-            this.debug('Error:', e);
+            this.dnaebug('Error:', e);
```
fleboulch: Typo

deepu105: ??

deepu105:
Of course, I agree with the flag, and since you added deprecation warning
and backward compatibility its good to go. I reviewed most parts, except
the upgrade and util part will do that as well, feel free to change it to
ready

Thanks & Regards,
Deepu

cli/utils.js
pvliss: Thanks. Let me re-check

generators/generator-base.js
fleboulch: I checked on the codebase and I don't find any reference for the method `dnaebug`. I missed something?

generators/app/index.js
pvliss: Will add the example. There is indeed an `Array` type that will consume all remaining arguments. See https://yeoman.io/authoring/user-interactions.html. Should I change it?

generators/generator-base.js
pvliss: Fixed. Thanks!!!

cli/utils.js
pvliss: You are right. It was not needed and it has been removed.

generators/app/index.js
pvliss: Reworded and added example. Let me know if you wish me to change the type as well

generators/app/index.js
pvliss: Fixed

generators/app/index.js
pvliss: Fixed.

generators/generator-base.js
pvliss: My 4 year-old most probably is really excited with js as much as his dad :rofl:

generators/app/index.js
deepu105: I think using array type would be cleaner

generators/generator-base.js
deepu105: ha ha

cli/utils.js
deepu105: Thanks. You still need to update commands.js to work with the new flag as it still has some code for blueprint stuff

pvliss: I think this should not be changed as it does not make sense to call the same sub-command for multiple blueprints or am I missing something?

generators/app/index.js
pvliss: OK. Will get to it. Will also need to rebase after merge of #9918 tom ake sure that changes in the upgrade gen are also tested. Would it be ok to squash intermediate commits to keep the history cleaner or would that mess up the discussions, etc?

cli/utils.js
pvliss: Looking again at `commands.js` it simply registers commands instead of calling them, my bad. So you are right that it should be changed. What I fail to see though is where the `blueprint` flag is being set in order for it to be used at `utils.createYeomanEnv()`. Maybe @sahbi-ktifa  can help who I believe coded this part?

deepu105: I think the blueprint flag is set by the blueprints itself when defining its own command.js file which is merged here

generators/app/index.js
deepu105: you can squash if you like

cli/utils.js
pvliss: OK. Thanks. Fixed `commands.js` to register commands from multiple blueprints

generators/app/index.js
pvliss: Apparently I misread the docs :man_facepalming:  . The `Array` type is supported for arguments but not options, hence back to the comma separated list. In any case, I have made a few more fixes and code improvements(mainly remove code duplication).

generators/utils.js
```diff
@@ -472,9 +486,63 @@ function checkStringInFile(path, search, generator) {
...
+ * @returns {Array} an array that contains the info for each blueprint
+ */
+function loadBlueprintsFromConfiguration(generator) {
+    // load blueprints from config file
+    const blueprints = generator.config.get('blueprints') || [];
+
+    const oldBlueprintName = generator.config.get('blueprint');
+    if (oldBlueprintName) {
+        const version = generator.config.get('blueprintVersion') || 'latest';
+        blueprints.push(parseBlueprintInfo(`${oldBlueprintName}@${version}`));
```
murdos: Maybe do this only if `oldBlueprintName` is not also declared in `blueprints` sections of config file?

generators/generator-base-private.js
```diff
@@ -759,6 +759,42 @@ module.exports = class extends Generator {
         }
     }
 
+    /**
+     * Instantiates the blueprint generators, if any.
+     * @param {string} subGen - sub generator
+     * @param {any} extraOptions - extra options to pass to blueprint generator
+     * @return {true} useBlueprints - true if one or more blueprints generators have been constructed; false otherwise
+     */
+    instantiateBlueprints(subGen, extraOptions) {
```
murdos: This should probably belong to `generator-base-blueprint.js`, since only generators that can be overridden by blueprints need this.

generators/generator-base-private.js
```diff
@@ -759,6 +759,42 @@ module.exports = class extends Generator {
...
+            blueprints.forEach(blueprint => {
+                let bpOptions = {
+                    ...this.options,
+                    configOptions: this.configOptions
+                };
+                if (extraOptions) {
+                    bpOptions = { ...bpOptions, ...extraOptions };
+                }
+                const useBP = this.composeBlueprint(blueprint.name, subGen, bpOptions);
+                // use global variable since getters dont have access to instance property
```
murdos: From what I understand, this comment can be removed, it's outdated since there's no global variable anymor

generators/upgrade/index.js
```diff
@@ -48,9 +49,9 @@ module.exports = class extends BaseGenerator {
             desc: 'Upgrade to a specific version instead of the latest',
             type: String
         });
-        // This adds support for a `--target-blueprint-version` flag
-        this.option('target-blueprint-version', {
-            desc: 'Upgrade to a specific blueprint version instead of the latest',
+        // This adds support for a `--target-blueprint-versions` flag
+        this.option('target-blueprint-versions', {
+            desc: 'Upgrade to specific blueprint versions instead of the latest',
```
murdos: Could you provide an example in the description and/or add more detailed info in the description? 
Indeed this parameter changed a bit. I guess it should be used as `--target-blueprint-versions foo@0.0.1,bar@1.0.2`, right ?

generators/upgrade/index.js
```diff
@@ -211,34 +209,58 @@ module.exports = class extends BaseGenerator {
...
+                                            )} Application has already been generated with latest version for blueprint: ${blueprint.name}`
+                                        );
+                                    }
+                                    this.success(`Done checking for new version for blueprint ${blueprint.name}`);
+                                    resolve();
+                                });
+                            });
+                        })
+                ).then(() => {
+                    this.success('Done checking for new blueprint versions');
```
murdos: `new version of blueprints`? (beware I'm not a native English speaking person 😉)

cli/commands.js
```diff
@@ -19,20 +19,37 @@
...
+    if (blueprintNames.length > 0) {
+        blueprintNames
+            .filter((v, i, a) => a.indexOf(v) === i)
+            .map(v => jhipsterUtils.normalizeBlueprintName(v))
+            .forEach(blueprint => {
+                /* eslint-disable import/no-dynamic-require */
+                /* eslint-disable global-require */
+                try {
+                    const blueprintCommands = require(`${blueprint}/cli/commands`);
+                    result = { ...customCommands, ...blueprintCommands };
```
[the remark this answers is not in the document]
pvliss: Nice catch!!! Fixed

generators/app/index.js
```diff
@@ -156,9 +156,15 @@ module.exports = class extends BaseGenerator {
             type: String
         });
 
-        // This adds support for a `--blueprint` flag which can be used to specify a blueprint to use for generation
+        // NOTE: Deprecated!!! Use --blueprints instead
         this.option('blueprint', {
-            desc: 'Specify a generator blueprint to use for the sub generators',
+            desc: 'Specify one or more generator blueprints to use for the sub generators',
```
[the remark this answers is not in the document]
pvliss: It could actually be used as  `--blueprint A,B` similar to `--blueprints`. But you are right, I have added `DEPRECATED:` and restored the old description

generators/utils.js
pvliss: Fixed

generators/generator-base-private.js
pvliss: Fixed

generators/generator-base-private.js
pvliss: Fixed

generators/upgrade/index.js
pvliss: Fixed

generators/upgrade/index.js
pvliss: Not an native English speaker either :) Fixed

pvliss:
@murdos Thanks a lot for the review!!!

Once this is supported, we should also change jhipster-core to support it in the JDL as well and probably need to slightly update the docs too

DanielFran:
@MathieuAA Is this something you can add quickly in jhipster-core for latest version or we will need to wait the merge to implement it?

MathieuAA:
Frankly, I'd like to wait 'til after the merge. I'll review the work tonight, if this can be done quickly I may include it in the release.

clement26695:
It seems that removing the skip-checks flag make the generation crash. Is this normal ? 

I just tested with the commands `jhipster --blueprint vuejs` or `jhipster --blueprints kotlin,vuejs` and it doesn't work since the merge of this PR. It works with the skip-checks flag however.

The error I have is :
`TypeError: [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type object`

I'll try to take a look to see if I can fix this.

clement26695:
I think #10029 should solve this

MathieuAA:
I'm delaying the release of JCore (not because I've an issue with my machine) but I think this can be included.

PR #9942, opened 2019-06-19
Fix jhipsterContext for entity client, server and i18n blueprints and

pvliss:
Fix `jhipsterContext` for entity client, server and i18n sub-gens and allow creating blueprints without having to override the parent entity sub-gen.

The problem is actually described by @yelhouti at https://github.com/jhipster/generator-jhipster-blueprint/issues/38 and I used the same workaround for https://github.com/jhipster/jhipster-kotlin/pull/146.

This fixes the problem and will make the workaround obsolete and further simplify the code for the blueprints, for  example:

**Before:**
```
module.exports = class extends EntityServerGenerator {
    constructor(args, opts) {
        super(args, { fromBlueprint: true, ...opts }); // fromBlueprint variable is important

        const jhContext = (this.jhipsterContext = this.options.jhipsterContext || opts.jhipsterContext);

        if (!jhContext) {
            this.error(`This is a JHipster blueprint and should be used only like ${chalk.yellow('jhipster --blueprint kotlin')}`);
        }

        this.configOptions = jhContext.configOptions || {};
    }

    get writing() {
        return writeFiles();
    }
};
``` 
**After:**
```
module.exports = class extends EntityServerGenerator {
    constructor(args, opts) {
        super(args, { fromBlueprint: true, ...opts }); // fromBlueprint variable is important
        if (!this.jhipsterContext) {
            this.error(`This is a JHipster blueprint and should be used only like ${chalk.yellow('jhipster --blueprint kotlin')}`);
        }
    }

    get writing() {
        return writeFiles();
    }
};
``` 
#### Related issues:
- https://github.com/jhipster/generator-jhipster-blueprint/issues/38
- https://github.com/jhipster/generator-jhipster-blueprint/issues/10
- https://github.com/jhipster/jhipster-kotlin/pull/146

### DIFF
--- a/generators/entity-client/index.js
+++ b/generators/entity-client/index.js
@@ -27,8 +27,9 @@ let useBlueprint;
 module.exports = class extends BaseBlueprintGenerator {
     constructor(args, opts) {
         super(args, opts);
-        utils.copyObjectProps(this, this.options.context);
-        this.configOptions = this.options.configOptions || {};
+        utils.copyObjectProps(this, opts.context);
+        this.jhipsterContext = opts.jhipsterContext || opts.context;
+        this.configOptions = opts.configOptions || {};
         const blueprint = this.options.blueprint || this.configOptions.blueprint || this.config.get('blueprint');
         if (!opts.fromBlueprint) {
             // use global variable since getters dont have access to instance property
--- a/generators/entity-i18n/index.js
+++ b/generators/entity-i18n/index.js
@@ -27,8 +27,9 @@ let useBlueprint;
 module.exports = class extends BaseBlueprintGenerator {
     constructor(args, opts) {
         super(args, opts);
-        utils.copyObjectProps(this, this.options.context);
-        this.configOptions = this.options.configOptions || {};
+        utils.copyObjectProps(this, opts.context);
+        this.jhipsterContext = opts.jhipsterContext || opts.context;
+        this.configOptions = opts.configOptions || {};
         const blueprint = this.options.blueprint || this.configOptions.blueprint || this.config.get('blueprint');
         if (!opts.fromBlueprint) {
             // use global variable since getters dont have access to instance property
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -28,7 +28,8 @@ module.exports = class extends BaseBlueprintGenerator {
     constructor(args, opts) {
         super(args, opts);
         utils.copyObjectProps(this, opts.context);
-        this.configOptions = this.options.configOptions || {};
+        this.jhipsterContext = opts.jhipsterContext || opts.context;
+        this.configOptions = opts.configOptions || {};
         if (this.databaseType === 'cassandra') {
             this.pkType = 'UUID';
         }
--- a/test/blueprint/entity-client-blueprint.spec.js
+++ b/test/blueprint/entity-client-blueprint.spec.js
@@ -1,0 +1,74 @@
+const path = require('path');
+const fse = require('fs-extra');
+const assert = require('yeoman-assert');
+const helpers = require('yeoman-test');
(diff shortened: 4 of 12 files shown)


pvliss:
@jhipster/developers Should I merge this?

mraible:
This issue made me realize I'm doing things the hard way with Ionic for JHipster. Rather than extending the entity generator, the code is mostly copy/pasted from JHipster.

For example, see https://github.com/oktadeveloper/generator-jhipster-ionic/blob/master/generators/entity/index.js

deepu105:
You should convert it into a blueprint. You can reuse a lot for JHipster
then. take look at the Vue blueprint for example

Thanks & Regards,
Deepu

pvliss:
@mraible Indeed you have taken the hard way. I would say that most blueprints should not have to extend entity gen either but rather in your case just extend the entity-client.

mraible:
Can it still be a blueprint if it's not part of a JHipster app? Ionic4J generates a separate app.

pvliss:
@mraible Not really familiar with Ionic4J. Will check it out and let you know, except if someone else already knows the answer :smiley:  @deepu105 ?

pvliss:
@mraible So I just checked and I see that ionic4J is a module which executes separately from jhipster and just imports the generator to share common functionality and of course the config. My understanding is that blueprints, are mainly designed to extend or override jhipster but using the same work flow and within the same process. I suppose you could convert it to a blueprint and even have an option for a separate mobile app directory either within the same "backend" dir or externally. Would that be a problem? In that case you would most likely have to extend the entity-client to generate the extra files and also call super in order to generate the Angular or React frontend as well. The templates from the ionic app gen should also be moved to the entity client

Another approach could be to keep it as a module and see if you can extend the entity and/or entity-client instead of the `BaseGenerator` and remove as much duplication as possible. 

Did a quick diff mainly to `index.js` and `prompt.js` of the entity subgen between ionic4j and jhipster and there are several differences. I am not sure though if these are on purpose or the files have just got out of sync?

pvliss:
@mraible So I did a quick POC, by simply extending the entity sub-gen and keeping Ionic4J as a module as from what I understand that was your initial intention. You can see it here: https://github.com/pvliss/generator-jhipster-ionic/tree/test-extend-entity-subgen. The main change to look at is at the `generators/entity/index.js`(a caveat is that insights are not disabled) and the fact that `generators/entity/prompt.js` has been removed. The rest of the changes are formatting by prettier auto-applied using IDEA. Also did a quick test using the config from your latest [blog](https://developer.okta.com/blog/2019/06/24/ionic-4-angular-spring-boot-jhipster) and it seems to work :)

mraible:
Wow! Excellent work!! 

Do you know if it’s possible to extend JHipster and support “import-jdl” as well? I think that’s the main feature the module is missing.

pvliss:
Thanks. Although be warned that this was a quick POC and I am not sure if there was something in the entity gen you wish to keep, so please double-check

I believe the import-jdl is now in cli and not a yeoman sub-gen, but I have not really dived into the cli, so I am not sure. Will be having a look though at some time in order to also try and convert the upgrade sub-gen to cli(see [comment](https://github.com/jhipster/generator-jhipster/pull/9918#issuecomment-503513909)).

pvliss:
@mraible BTW, I believe the insights can also be skipped by passing `noInsight:true` on construction but have not tested it. See my latest [commit](https://github.com/pvliss/generator-jhipster-ionic/commit/0f0fea5074dd23a56481f7439fd445f2e2af0a5b)

mraible:
The only reason I had insights disabled is because it was causing an error. I commented it out because I didn’t have time to investigate.

pvliss:
Oh. Sorry, did not realize

Issue #9871, opened 2019-06-07
Add cache entry listener logger for ehcache


DanielFran:
##### **Overview of the feature request**

Implements a cache entry listener to allow to log what is happenning with cache.

##### **Motivation for or Use Case**

This might be usefull to log for investigation perpose.

##### **Possible implementation**

```java
import javax.cache.event.CacheEntryCreatedListener;
import javax.cache.event.CacheEntryEvent;
import javax.cache.event.CacheEntryExpiredListener;
import javax.cache.event.CacheEntryListenerException;
import javax.cache.event.CacheEntryRemovedListener;
import javax.cache.event.CacheEntryUpdatedListener;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

/**
 * A custom cache entry listener.
 */
public class CacheEntryListenerLogger implements CacheEntryCreatedListener<Object, Object>, CacheEntryUpdatedListener<Object, Object>, CacheEntryExpiredListener<Object, Object>, CacheEntryRemovedListener<Object, Object> {

    private static final Logger log = LoggerFactory.getLogger(CacheEntryListenerLogger.class);
    
    @Override
    public void onCreated(Iterable<CacheEntryEvent<? extends Object, ? extends Object>> events) throws CacheEntryListenerException {
        for (CacheEntryEvent<? extends Object, ? extends Object> entryEvent : events) {
            log.info("Cache event {} for key {} with value {}", entryEvent.getEventType(), entryEvent.getKey(), entryEvent.getValue());
        }
    }

    @Override
    public void onUpdated(Iterable<CacheEntryEvent<? extends Object, ? extends Object>> events) throws CacheEntryListenerException {
        for (CacheEntryEvent<? extends Object, ? extends Object> entryEvent : events) {
            log.info("Cache event {} for key {}, with old value {} and new value {}", entryEvent.getEventType(), entryEvent.getKey(), entryEvent.getOldValue(), entryEvent.getValue());
        }
    }

    @Override
    public void onExpired(Iterable<CacheEntryEvent<? extends Object, ? extends Object>> events) throws CacheEntryListenerException {
        for (CacheEntryEvent<? extends Object, ? extends Object> entryEvent : events) {
            log.info("Cache event {} for key {} with value {}", entryEvent.getEventType(), entryEvent.getKey(), entryEvent.getValue());
        }
    }

    @Override
[...]

PR #9951, opened 2019-06-20
Add cache entry listener logger for ehcache

DanielFran:
Depends on https://github.com/jhipster/jhipster/pull/300

Fix https://github.com/jhipster/generator-jhipster/issues/9871

### DIFF
--- a/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
@@ -28,6 +28,9 @@ import org.ehcache.jsr107.Eh107Configuration;
 import org.hibernate.cache.jcache.ConfigSettings;
     <%_ } _%>
 import io.github.jhipster.config.JHipsterProperties;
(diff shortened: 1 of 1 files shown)


DanielFran:
@henri-tremblay If you can review also, would be awesome. Thanks

henri-tremblay:
@DanielFran Will do today hopefully. Can I know why we need a listener? They are in general not recommended since they have a noticeable performance impact.

DanielFran:
This is just for logging and it is configured asynchronustly

henri-tremblay:
So, I would really not do that by default. As a way to debug the content of the cache.

1. When adding a listener, Ehcache goes in a different mode, a slower mode, to be able to call the listeners
2. A toString on a value can be quite slow. And could also trigger the cache when it's a Hibernate object. I'm afraid of sneaky bugs

PierreBesson:
Same for me, I don't like this solution. I'm sure there is a better way to log cache put and evict operations for debugging.

DanielFran:
OK, many thanks for the review.
Since there is no agreement to use it, no problem I close it.

PR #9963, opened 2019-06-21
replace hard-coded value of `test@localhost` with property value

TheJavaGuy:
In the `MailServiceIT.java.ejs` the value of sender is hard-coded. This value is, by default, the same one as given in the `application.yml` file under the key `jhipster.mail.from`. In the case we change the `test@localhost` value in the `application.yml` file, the tests will fail. This change will make tests pass in this case without having to do anything further.


pascalgrimaud:
Thanks and congrats for your 1st contribution @TheJavaGuy

Issue #9891, opened 2019-06-11
[React] Use hooks for components


qmonmert:
##### Use hooks in all React components

Hooks are available since the version 16.8 of React. The goal will be to integrate hooks in JHipster.

##### Todo

Here are all the components that seem be impacted :

-   [x] Header #9883 **MERGED**
-   [x] Password #9899 **MERGED**
-   [x] Register #9899 **MERGED**
-   [x] Configuration #9919 **MERGED**
-   [x] Health #9921 **MERGED**
-   [x] Logs #9921 **MERGED**
-   [x] Activate #9964 **MERGED**
-   [x] Logout #9968 **MERGED**
-   [x] Settings #9965 **MERGED**
-   [x] Metrics #9966 **MERGED**
-   [x] PasswordResetFinish #9924 **MERGED**
-   [x] UserManagementDetail #9929 **MERGED**
-   [x] UserManagementDeleteDialog #9938 **MERGED**
-   [x] UserManagementUpdate #9938 **MERGED**
-   [x] Home #9969 **MERGED**
-   [x] Login #9971 **MERGED**
-   [x] App #9972 **MERGED**
-   [x] Audits #9975 **MERGED**
-   [x] UserManagement #9981 **MERGED**

And the entities :

-   [ ] Entity #9983 ❌

deepu105:
@qmonmert I suggest doing small PRs so that its easy to review and test, may a PR per module(admin, layout, account, entity)

qmonmert:
@deepu105 yes I'm doing some small PRs

deepu105:
Once we implement the new bounty system, this issue should get 300$ bounty atleast

deepu105:
@jdubois I have changed the bounty amount as this is a lot of work and deserves more. Hope you are fine with it

jdubois:
Of course I am!!

pascalgrimaud:
Crazy work here @qmonmert !
Hope someone with more experience with React can review all your PRs

qmonmert:
thanks @pascalgrimaud :) I hope cc @deepu105 & @chrisdns ;)

qmonmert:
@deepu105 & @pascalgrimaud I am not able to make azure e2e tests pass for PRs #9968 & #9981 even after multiple attempts (it works locally). Are these PR still mergeable are should we wait for success ?

deepu105:
excellent work @qmonmert :clap:

pascalgrimaud:
[...]

PR #9968, opened 2019-06-21
[React] Hooks (logout component)

qmonmert:
WIP on #9891

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/modules/login/logout.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/login/logout.tsx.ejs
@@ -29,6 +29,7 @@ export interface ILogoutProps extends StateProps, DispatchProps {
 
 export const Logout = (props: ILogoutProps) => {
   useEffect(() => {
+    props.logout();
     const logoutUrl = props.logoutUrl;
     if (logoutUrl) {
       // if Keycloak, logoutUrl has protocol/openid-connect in it
@@ -37,7 +38,6 @@ export const Logout = (props: ILogoutProps) => {
           ? logoutUrl + '?redirect_uri=' + window.location.origin
           : logoutUrl + '?id_token_hint=' + props.idToken + '&post_logout_redirect_uri=' + window.location.origin;
     }
-    props.logout();
   });
 
   return (
--- a/generators/client/templates/react/src/main/webapp/app/modules/login/logout.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/login/logout.tsx.ejs
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import React from 'react';
+import React, { useEffect } from 'react';
 import { connect } from 'react-redux';
 
 import { IRootState } from 'app/shared/reducers';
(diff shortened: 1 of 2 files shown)


deepu105:
Could you check if the failures are actual errors? also aren't you able to
to restart onloy the failed build from Travis?

Thanks & Regards,
Deepu

qmonmert:
@deepu105 I don't understand the errors, the error is on Azure

deepu105:
@qmonmert any update on this one. Do you need any help?

qmonmert:
@deepu105 I need help because on my machine all tests pass :/

pascalgrimaud:
@qmonmert : did you test locally with prod profile ?

qmonmert:
@pascalgrimaud I just tried and it works locally :/

generators/client/templates/react/src/main/webapp/app/modules/login/logout.tsx.ejs
```diff
@@ -27,29 +27,26 @@ export interface ILogoutProps extends StateProps, DispatchProps {
...
-  componentDidMount() {
-    this.props.logout();
+  const logoutUrl = props.logoutUrl;
+  if (logoutUrl) {
+    // if Keycloak, logoutUrl has protocol/openid-connect in it
+    window.location.href =
+      logoutUrl.includes('/protocol')
+        ? logoutUrl + '?redirect_uri=' + window.location.origin
+        : logoutUrl + '?id_token_hint=' + props.idToken + '&post_logout_redirect_uri=' + window.location.origin;
   }
```
SoniaLahi: @qmonmert: wouldn't it be better to add these lines of code in the useEffect function, as the actions on a component lifecycle event are specified in this function ?

pascalgrimaud:
@qmonmert : first of all, sorry for taking time to review this.

I just gave a try locally and it failed, exactly with same error than in CI.
So the error comes from this PR, and not from Azure.
Then, it failed during entity e2e tests, so it's really strange...

I pushed a repo with the same config than Azure, so you can use it:

```
git clone https://github.com/pascalgrimaud/jh-react-pr-9968
cd jh-react-pr-9968
npm install
./mvnw clean verify -Pprod
docker-compose -f src/main/docker/mysql.yml up -d
java -jar target/*.jar --spring.profiles.active=prod,swagger
```

Then, run: `npm run e2e` in another terminal.

It looks like after a delay, the logout doesn't work anymore, so we have something like this:

```
FieldTestInfiniteScrollEntity e2e test
    1) "before all" hook: ret for "should load FieldTestInfiniteScrollEntities"
    2) "after all" hook: ret for "should delete last FieldTestInfiniteScrollEntity"

  FieldTestMapstructEntity e2e test
    3) "before all" hook: ret for "should load FieldTestMapstructEntities"
    4) "after all" hook: ret for "should delete last FieldTestMapstructEntity"

  FieldTestPagerEntity e2e test
    5) "before all" hook: ret for "should load FieldTestPagerEntities"
    6) "after all" hook: ret for "should delete last FieldTestPagerEntity"

  FieldTestPaginationEntity e2e test
    7) "before all" hook: ret for "should load FieldTestPaginationEntities"
    8) "after all" hook: ret for "should delete last FieldTestPaginationEntity"

```

pascalgrimaud:
@qmonmert : did you manage to reproduce the issue locally ?

Then, I think we need some React experts to help for achieving this PR.

@sendilkumarn @deepu105 ?

deepu105:
@pascalgrimaud @qmonmert I need to find some time to loom at it, also its been a while since I did any serious react, next two weeks i'll be busy with Devoxx Ukraine and Codemotion Berlin but after that I can see whats going on here

deepu105:
These kinds of issues are mostly due to some race conditions in our code which become apparent when protractor runs through things in superhuman speeds, it would be less evident when we use those pages as we are normally slow in those interactions. @qmonmert maybe try using some wait in between the actions in the protractor tests to see if this really is due to race conditions

pascalgrimaud:
@qmonmert : it looks ok, finally !!!

qmonmert:
@pascalgrimaud I found a solution but continuous-integration is always red 😕

pascalgrimaud:
@qmonmert : the failed build is related to relationships, there is a ticket for that. Nothing related to your PR. As it looks OK, I'm merging this.
Thanks for your patience, still 1 last PR !!

qmonmert:
@pascalgrimaud 🙂 I did not see the last failed 😇 so I have had a genius idea this night !

Issue #9836, opened 2019-06-01
Entity filtering: 'greaterOrEqualThan' should be 'greaterThanOrEqual'


yelhouti:
Hey :), I spent an houre trying to figure out why filters are not working, turns out that it's and english problem:
it's greaterThanOrEqual "gte" not greaterOrEqualThan expected by jhipster...
Any idea on how to fix this without introducing breaking changes, I would suggest sone json annotation at the same time be able to handle "gte", "gt"....

pascalgrimaud:
@yelhouti : please, provide more information, steps to reproduce your issue. It will be hard to guest your configuration here and what you tried to do.

jdubois:
I also have no idea what you talk about... And please don't use "guys", we try to be gender-neutral.

yelhouti:
@jdubois @pascalgrimaud my bad I could have been more specific, here are lines of code:
https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/service/filter/RangeFilter.java
```
   private FIELD_TYPE greaterThan;
    private FIELD_TYPE lessThan;
    private FIELD_TYPE greaterOrEqualThan;
    private FIELD_TYPE lessOrEqualThan;
```

pascalgrimaud:
cc @gzsombor as you look like the original author of the file. Sorry if I'm wrong :) 
What do you think ?

Blackdread:
could add setters with correct name (maybe not getters otherwise then need to tell Swagger to ignore one of the getter/setter for documentation of API)

gzsombor:
Sorry, I'm not a native English speaker - interesting, that no one complained before :wink: 
Having a new setter could be a good solution, unfortunately this problem is on the Java-Javascript layer, we could break the UI easily for an unsuspecting upgrader.

jdubois:
Yes this is breaking the API, but also this is correcting a bug... and having both until the next major release will not help much: I know we'll forget about it, and then this will confuse people to have the 2 APIs available... So in the end I would just refactor this and document it in the next release notes.

gzsombor:
I've opened a PR in the jhipster repo for this: https://github.com/jhipster/jhipster/pull/299

gzsombor:
I think, this ticket can be closed now - in the next major release, we can remove the compatibility, deprecated setters.

PR #9976, opened 2019-06-23
Fix to use greaterThanOrEquals in the tests

gzsombor:
and increase the code coverage to check more field types and more filters - following #9836 


generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
```diff
@@ -271,22 +272,27 @@ _%>
 
     private static final Integer <%=defaultValueName %> = <%= defaultValue %>;
     private static final Integer <%=updatedValueName %> = <%= updatedValue %>;
+    private static final Integer <%=smallerValueName %> = <%= defaultValue %> - 1;
```
murdos: If I'm not misunderstanding what this PR does, these new `SMALLER_` fields will only be used if the JPA filtering is enabled.

So it would be better to add them conditionally (`<%_ if (jpaMetamodelFiltering) { ... }`) otherwise they will be dead/unused code in the generated application.

PR #9984, opened 2019-06-25
use correct Spring profiles when building app with different Gradle profile

jack-kerouac:
Before, when first compiling the project with Gradle `dev` profile and then, without cleaning the
project, with Gradle `prod` profile, the list of active Spring profiles was not updated. Now it is.


atomfrede:
Thanks @jack-kerouac for your 1st contribution! I have also seen it a few days ago but didn't had time to investigate further, good work!

PR #9985, opened 2019-06-25
allow using Gradle project version in jib configuration

jack-kerouac:
We decided to not use `latest` as the tag for the Docker image built by jib, but the Gradle project version. While I think changing the default Docker image tag is not a good change for JHipster in general, moving the project version up allows it to be referenced by Jib. Not knowing too much about Gradle project configuration before, we stumbled over this and I want to save others this learning.

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -55,6 +55,11 @@ plugins {
     //jhipster-needle-gradle-plugins - JHipster will add additional gradle plugins here
 }
 
+group = "<%= packageName %>"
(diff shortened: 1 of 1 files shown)


atomfrede:
Interesting that it helps. Can you show your jib configuration maybe?

jack-kerouac:
image

Running `./gradlew jibDockerBuild` then says
> `Containerizing application to Docker daemon as <REPO>/<NAME>:0.0.1-SNAPSHOT...`

(I replaced our projects repo and name with `<REPO>` and `<NAME>` in both places above)

PR #9986, opened 2019-06-25
pass app version to NPM as env var instead of parsing it from build file

jack-kerouac:
Currently, the webpack `DefinePlugin` sets the `VERSION` by parsing it from either `build.gradle`
(in case of a Gradle build) or from `pom.xml` (in case of a Maven build). The problem with this is
that if the version is not statically set, but rather dynamically calcualted (e.g. using
https://github.com/palantir/gradle-git-version to determine it from the Git repo), the parsing
breaks.

Thus, this PR changes this by passing the version to show in the frontend from the server build
tool (Gradle/Maven) as an environment variable to the NPM webpack task. This is more robust and
does not rely on parsing the Gradle/maven file. The downside is that if only `npm start` is run,
instead of the version, `UNKNOWN` is shown.

I think the advantages (allowing dynamic version in Gradle, removing parsing code) outweigh this
downside.

I tested this by
1. generating Maven and Gradle projects
2. generating Angular and React clients
3. running it using `dev` vs. `prod` profiles
4. running the server build tool vs. `npm start`

(but not all combinations of the above :-).

This PR depends on https://github.com/jhipster/generator-jhipster/pull/9985, so that Gradle
project version can be accessed in the `gradle/profile_prod/dev.gradle` files.

### DIFF
--- a/generators/client/templates/angular/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.common.js.ejs
@@ -78,7 +78,13 @@ module.exports = (options) => ({
             'process.env': {
                 NODE_ENV: `'${options.env}'`,
                 BUILD_TIMESTAMP: `'${new Date().getTime()}'`,
-                VERSION: `'${utils.parseVersion()}'`,
+<%_ if (buildTool === 'gradle' || buildTool === 'maven') { _%>
+                // If webpack is run through Gradle 'webpack'/'webpackBuildDev' NpmTasks / Maven 'frontend-maven-plugin', then APP_VERSION is set.
+                VERSION: `'${process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'UNKNOWN'}'`,
+<%_ } else { _%>
+                // Server generation is skipped, so there is no way to determine the application version from server.
+                VERSION: `'0.0.1-SNAPSHOT'`,
+<%_ } _%>
                 DEBUG_INFO_ENABLED: options.env === 'development',
                 // The root URL for API calls, ending with a '/' - for example: `"https://www.jhipster.tech:8081/myservice/"`.
                 // If this URL is left empty (""), then it will be relative to the current context.
--- a/generators/client/templates/react/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/react/webpack/webpack.common.js.ejs
@@ -123,7 +123,13 @@ module.exports = options => ({
         <%_ if (enableTranslation) { _%>
         BUILD_TIMESTAMP: `'${new Date().getTime()}'`,
         <%_ } _%>
-        VERSION: `'${utils.parseVersion()}'`,
+<%_ if (buildTool === 'gradle' || buildTool === 'maven') { _%>
+        // If webpack is run through Gradle 'webpack'/'webpackBuildDev' NpmTasks / Maven 'frontend-maven-plugin', then APP_VERSION is set.
+        VERSION: `'${process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'UNKNOWN'}'`,
+<%_ } else { _%>
+        // Server generation is skipped, so there is no way to determine the application version from server.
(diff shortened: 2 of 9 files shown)


PierreBesson:
This is pretty cool. Thank you for your contribution @jack-kerouac !
Do you think you can add the git version as well. It should be `${git.commit.id.describe}`. At least it is for maven, I'm not sure about gradle.

atomfrede:
@PierreBesson Should work with gradle too

jack-kerouac:
I am not sure what exactly you mean with 'add the git version as well'. Append it to the `VERSION` property in the `DefinePlugin`?

I would prefer to not change functionality within this PR, to be honest. This IMHO is rather a refactoring. Fine?

PierreBesson:
Ok, never mind about my git version comment.

jack-kerouac:
Can you maybe help me find out why the Travis build failed? I do not know how to fine the output of mocha in Travis, to understand what exactly failed.

PierreBesson:
Looks like a random failure. I restarted the build.

jack-kerouac:
I just realize that we might now remove `xml2js` from `devDependencies`. What do you think?

jack-kerouac:
I amended the commit and removed this dependency.

generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -78,7 +78,16 @@ module.exports = (options) => ({
             'process.env': {
                 NODE_ENV: `'${options.env}'`,
                 BUILD_TIMESTAMP: `'${new Date().getTime()}'`,
-                VERSION: `'${utils.parseVersion()}'`,
+<%_ if (buildTool === 'maven') { _%>
+                // If webpack is invoked by Maven 'frontend-maven-plugin', then APP_VERSION is set.
+                VERSION: `'${process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'UNKNOWN'}'`,
+<%_ } else if (buildTool === 'gradle') { _%>
```
vishal423: can we combine `if` and `else if`?

generators/client/templates/react/webpack/webpack.common.js.ejs
```diff
@@ -123,7 +123,16 @@ module.exports = options => ({
         <%_ if (enableTranslation) { _%>
         BUILD_TIMESTAMP: `'${new Date().getTime()}'`,
         <%_ } _%>
-        VERSION: `'${utils.parseVersion()}'`,
+<%_ if (buildTool === 'maven') { _%>
+        // If webpack is invoked by Maven 'frontend-maven-plugin', then APP_VERSION is set.
+        VERSION: `'${process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'UNKNOWN'}'`,
+<%_ } else if (buildTool === 'gradle') { _%>
+        // If webpack is run through Gradle 'webpack'/'webpackBuildDev' NpmTasks, then APP_VERSION is set.
+        VERSION: `'${process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'UNKNOWN'}'`,
```
vishal423: same as above

generators/client/templates/angular/webpack/webpack.common.js.ejs
jack-kerouac: I am not sure if this makes it easier to read (since the comments above are different). What do you think?

atomfrede: Maybe we can make the comment `If webpack is invoked via Maven/Gradle then APP_VERSION is set` as the code is the same?

jack-kerouac: That seems to be a more pragmatic solution :). Thanks. Will change.

generators/client/templates/react/webpack/webpack.common.js.ejs
jack-kerouac: Solved.

vishal423:
thanks @jack-kerouac

PR #10000, opened 2019-06-27
Allow loading .yo-rc.json from an alternate root path other than cwd

pvliss:
Allow changing the parent root directory of the `.yo-rc.json`. The default remains the` cwd`.

The use case is to allow jhipster modules to take advantage of existing code and e.g. re-use the existing entity sub-gen while loading the configuration from a directory other than the `cwd`. See https://github.com/oktadeveloper/generator-jhipster-ionic/pull/109

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -414,11 +414,13 @@ function decodeBase64(string, encoding = 'utf-8') {
  * Get all the generator configuration from the .yo-rc.json file
  * @param {Generator} generator the generator instance to use
  * @param {boolean} force force getting direct from file
+ * @param {String} base path where the .yo-rc.json file is located. Default is cwd.
  */
-function getAllJhipsterConfig(generator, force) {
+function getAllJhipsterConfig(generator, force, basePath = '') {
     let configuration = generator && generator.config ? generator.config.getAll() || {} : {};
-    if ((force || !configuration.baseName) && jhiCore.FileUtils.doesFileExist('.yo-rc.json')) {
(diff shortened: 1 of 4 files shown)


generators/utils.js
```diff
@@ -414,11 +414,13 @@ function decodeBase64(string, encoding = 'utf-8') {
  * Get all the generator configuration from the .yo-rc.json file
  * @param {Generator} generator the generator instance to use
  * @param {boolean} force force getting direct from file
+ * @param {String} base path where the .yo-rc.json file is located. Default is cwd.
  */
-function getAllJhipsterConfig(generator, force) {
+function getAllJhipsterConfig(generator, force, basePath = '') {
```
deepu105: Can you check if this is exposed anywhere binnour public API if so we need to take care of backwards compatibility

pvliss: Isn't that being taken care of by the default parameter?

jdubois:
This is issue/PR #10000 !!!! Congrats! I can't believe we achieved so much!

generators/utils.js
pvliss: So I just added unit tests as well and an extra check within `utils.getAllJhipsterConfig()`. 

The method is not used directly within the generator but rather sub-gens use the one from `generator-base.js`. Nevertheless, since the method is exported in `utils.js` it could be used by 3rd party code(e.g. blueprints,modules,etc.)

deepu105: sorry didnt notice default param

PR #10001, opened 2019-06-27
Add a message when `yo` is not installed.

antarus:
Add a message when `yo` is not installed.

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -830,7 +830,15 @@ module.exports = class extends Generator {
             done();
             return;
         }
+
         shelljs.exec('yo --generators', { silent: true }, (err, stdout, stderr) => {
+            if (stderr.includes('yo:')) {
+                this.error(
+                    `The ${chalk.yellow('yo generator')} is not installed. Please install it using command ${chalk.yellow(
+                        `npm install -g yo`
+                    )}.`
+                );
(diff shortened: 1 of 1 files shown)


antarus:
Without yo, i can't test the blueprint vue.js

jdubois:
Yes, `yo` is needed to run the blueprints, so we can't remove it, at least not for advanced users

PierreBesson:
AFAIK, yo is only needed for modules not blueprints (as they should be run with `jhipster --blueprint blueprint`). I would rather we totally deprecate yo usage and find an alternative solution even if it means just wrapping yo in the jhipster cli (for example, implementing something like `jhipster --module ionic`).

jdubois:
Oh sorry, @PierreBesson you are probably right, it's for modules, not blueprints.

PierreBesson:
I have an alternate solution to this problem https://github.com/jhipster/generator-jhipster/pull/10004
Actually jhipster was depending on the globally installed yo command in order to check that the blueprint was correctly installed. In my PR I'm simply using yo from generator-jhipster's node_modules so this is hidden from the user.

PierreBesson:
Anyway @antarus thanks for doing this PR, which revealed the issue !

antarus:
ok, no problem ;)

PR #10007, opened 2019-06-28
Remove project name condition (underscore) in microservice generation

avdev4j:
When we creating a microservice application we don't allow an application name that contain an underscore even if it's authorized when we creating a monolith.

I understand it's for url definition but as we already do for kafka we can use c for urls definition and authorized underscore in applications name for every case.

We need to know where dasherizedBaseName is mandatory, so I start with a draft PR and feel free to tell me where I should replace baseName by dasherizedBaseName.

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -1685,9 +1685,6 @@ module.exports = class extends PrivateBase {
                     if (!/^([a-zA-Z0-9_]*)$/.test(input)) {
                         return 'Your base name cannot contain special characters or a blank space';
                     }
-                    if ((generator.applicationType === 'microservice' || generator.applicationType === 'uaa') && /_/.test(input)) {
-                        return 'Your base name cannot contain underscores as this does not meet the URI spec';
-                    }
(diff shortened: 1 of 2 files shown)


PierreBesson:
I'm not too sure that we should use the dasherizedName here but I would not touch this part of the code without testing end 2 end that it does not break anything.
Can't we agree not to use underscore in app names plz :cry:. This complicates our microservice support for no valid reason...

deepu105:
I'm ok to restrict `_` atleast in new projects, but we need handle existing
project regeneration

Thanks & Regards,
Deepu

PierreBesson:
It seems a tricky part to develop. I propose to close this MR @avdev4j. If you put an underscore in your app name then you brought trouble uppon youself.

avdev4j:
The problem is that we can have "_" in one case and can't in another case. But as @deepu105 said, we should handle this only for new projects. I think it's possible but TBH I did'nt have enough time to do that.

I suggest to keep this Issue closed and if I have time I will take a look.

Now I'm using camel case ;) (It's not really me but client specifications ofc).

PR #10069, opened 2019-07-09
Fix TranslateModule usage

wmarques:
Because of ng-jhipster packaging, we had one instance of TranslateModule per lazy loaded modules. This caused some issues when we introduced lazy-loading and we had to do some strange hacks. 
You can see the removed code in this PR to check the hacks.

We should do one `forRoot` call in our `AppModule` instead of doing it multiple times, so I think we should remove the `forRoot` in the ng-jhipster lib and do that in the generated app instead.

We need https://github.com/jhipster/ng-jhipster/pull/99 to be closed and a `ng-jhipster` release before merging this.

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -46,7 +46,7 @@
     <%_ } _%>
     "core-js": "3.2.1",
     "moment": "2.24.0",
-    "ng-jhipster": "0.11.0",
+    "ng-jhipster": "0.11.1",
     "ngx-cookie": "4.0.2",
     "ngx-infinite-scroll": "8.0.0",
     "ngx-webstorage": "4.0.1",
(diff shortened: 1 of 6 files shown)


vishal423:
@wmarques, did you test against lazy loaded modules and by changing language to see that they correctly update the translations? I think most of removed code was to make it work with lazy modules. If that's not required anymore, then, we can also remove `_language` behavior subject defined in the `JhiLanguageHelper` class. Also, don't we require `forChild` in lazy modules to correct import their translations?

wmarques:
@vishal423 Yep I just tested the admin part and it's working, I will double check to be sure.
Yes we can remove `_language`, will do that.

The `forChild` is there to lazy load translations also, for now we don't do that so we don't need `forChild` we always use the same `TranslationService` instance in each modules.

vishal423:
@wmarques , if user has disabled the i18n support, then, after this change do we still end up with `ngx-translate` in vendor bundle?

kaidohallik:
@wmarques `ng-jhipster` version 0.11.1 is released which can be used in this PR. Are you able to update `ng-jhipster` version in this PR and resolve conflicts?

vishal423:
@wmarques, I have pushed changes on your branch to fix conflicts.

wmarques:
@vishal423 looks good according to tests. Is that ok to merge this ?

kaidohallik:
Thanks @wmarques and @vishal423, this was a good improvement.

PR #10070, opened 2019-07-09
Lazy Load Moment Locales

wmarques:
We weren't using moment locales when switching language.

Now we lazy load the requested language and then set the correct moment locale.
I also removed the usage of useless Promises in `LanguageHelper` class


wmarques:
This can be quite complex for users and I'm afraid that some locales won't have the same identifier in moment... That can be hard to maintain for not such a big improvement IMO, so let's close this.

PR #10102, opened 2019-07-17
use default link colors of Bootstrap/Bootswatch theme

jack-kerouac:
With c55634ed99317aedebf5efad441a3a3e87f3f23b, it was decided to override the link color and font-weight of Bootstrap/the Bootswatch theme for passing the Lighthouse accessibility audit.

I think this is a bad idea since 1. the picked color might not match the color scheme of the Bootswatch theme and 2. it should be the responsibility of the theme to consider accessibility concerns and not JHipster building on top of this. This also again reduces the JHipster provided CSS (which I think is in general a good idea).

I thus suggest to remove this custom overriding of the link colors again.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
+++ b/generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
@@ -31,14 +31,15 @@ h4 {
     font-weight: 300;
 }
 
-a {
+<%_ if (clientTheme === 'none') { _%>
+/* Increase contrast of links to get 100% on Lighthouse Accessability Audit. Override this color if you want to change the link color, or use a Bootswatch theme */
+a, a:hover {
     color: #533f03;
     font-weight: bold;
 }
+<%_ } _%>
 
 a:hover {
-    color: #533f03;
-    font-weight: bold;
     /* make sure browsers use the pointer cursor for anchors, even with no href */
     cursor: pointer;
 }


generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
```diff
@@ -31,14 +31,7 @@ h4 {
     font-weight: 300;
 }
 
-a {
```
vishal423: Instead, can we make these conditional on `clientTheme === 'none'` ?

jack-kerouac: If there is no client theme, then the default Bootstrap styling is applied. It should again be the responsibility of this default styling to have good contrast and not of JHipster. If we are worried that this is a problem, I would rather create an issue in the Bootstrap project for this.

What do you think about this?

jack-kerouac: Related issues in Bootstrap:

https://github.com/twbs/bootstrap/issues/15304
https://github.com/twbs/bootstrap/issues/22549#issuecomment-422023478

And maybe this: https://getbootstrap.com/docs/4.1/getting-started/accessibility/#color-contrast

vishal423: as per linked docs:

> Most colors that currently make up Bootstrap’s default palette—used throughout the framework for things such as button variations, alert variations, form validation indicators—lead to insufficient color contrast (below the recommended WCAG 2.0 color contrast ratio of 4.5:1) when used against a light background. <b>Authors will need to manually modify/extend these default colors to ensure adequate color contrast ratios.</b>

Considering highlighted recommendation, I think it's appropriate to modify when no theme is chosen

jack-kerouac: Ok. Will do.

vishal423:
thanks @jack-kerouac

PR #10106, opened 2019-07-17
fix `google` meta tag attribute

TheJavaGuy:
For some reason, `google` meta tag had attribute `value` instead of `content`. According to the [documentation](https://support.google.com/webmasters/answer/79812?hl=en) this is wrong. Therefore, this commit introduces `content` attribute instead of `value`.


vishal423:
thanks @TheJavaGuy

TheJavaGuy:
@vishal423 You're welcome

PR #10119, opened 2019-07-19
make variable overrides in _bootstrap-variables.scss change components

jack-kerouac:
Currently, the `_bootstrap_variables.scss` file is not included into `vendor.scss`. Thus, the overriden variables in there are not affecting the bootstrap components. This commit fixes that. Now, the variables defined in this file influence the bootstrap component colors and other styles.

Since `_bootstrap-variables.scss` is also present (but being empty) when generating without a Bootswatch client theme, I think it still makes sense to also include it in this case. But in order to allow overriding variables already set in the `variables.scss` file of the Bootswatch theme, `_bootstrap-variables.scss` needs to be imported after this file.

The downside of this change is that now, the default layout has a white box on gray background:

image

I suggest to thus also change the value of `$body-bg` in `_bootstrap-variables.scss` to `white`. What do you think?

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
+++ b/generators/client/templates/angular/src/main/webapp/content/scss/vendor.scss.ejs
@@ -22,10 +22,12 @@
 put Sass variables here:
 eg $input-color: red;
 ****************************/
-// Override Boostrap variables
 <%_ if (clientTheme !== 'none') {_%>
 @import "~bootswatch/dist/<%= clientTheme %>/variables";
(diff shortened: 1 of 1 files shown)


atomfrede:
At least in our vue part it is normal and I think we had that kind of design for 5.x.

Screenshot_2019-07-24 javaprosample

jack-kerouac:
Ok. Then I suggest to merge this as it is, without changing the background color.

Anything I can do to move this forward?

atomfrede:
I will merge this, thanks for your contribution! In case we would like to change the style again it is easily possible.

PR #10154, opened 2019-07-30
Fix import angular service filename

mshima:
Entity's json service is created with name '${generator.entityServiceFileName}.service.ts' but is imported as '<%= entityFileName %>.service'.
Fix this bug.

https://github.com/jhipster/generator-jhipster/blob/a1023e100dce9689289383f57fbe7a064b1b84d8/generators/entity-client/files.js#L98

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -95,7 +95,7 @@ const angularFiles = {
                 },
                 {
                     file: 'entities/entity.service.ts',
-                    renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityServiceFileName}.service.ts`
+                    renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.service.ts`
                 }
             ]
         }


vishal423:
@mshima, As per my understanding both are same. Check this https://github.com/jhipster/generator-jhipster/blob/d25818cdca9a6337fc0e0374d437b8c70aa9a98e/generators/entity/index.js#L670

mshima:
Part of https://github.com/jhipster/generator-jhipster/issues/10160

mshima:
> @mshima, As per my understanding both are same. Check this
> 
> https://github.com/jhipster/generator-jhipster/blob/d25818cdca9a6337fc0e0374d437b8c70aa9a98e/generators/entity/index.js#L670

Correct, so
https://github.com/jhipster/generator-jhipster/blob/a1023e100dce9689289383f57fbe7a064b1b84d8/generators/entity-client/files.js#L98

should be changed to 

```
renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.service.ts\`
```

and entityServiceFileName should be removed.

In the process of changing entityFileName, the service file was not updated, so I found this bug.

mshima:
> @mshima, As per my understanding both are same. Check this
> 
> https://github.com/jhipster/generator-jhipster/blob/d25818cdca9a6337fc0e0374d437b8c70aa9a98e/generators/entity/index.js#L670

Should I create a separated issue?
Should I change the PR to the alternative fix?
Both seems ok to me.

vishal423:
It seems that the `entityServiceFileName` is used once in the complete codebase. I will prefer to reduce the change impact and will keep `entityFileName`. However, we should also not remove the `entityServiceFileName` variable and it's current assignment to be equal to the `entityFileName` as it might have been used in a module/blueprint. We can do the clean in the next major release.

mshima:
> It seems that the `entityServiceFileName` is used once in the complete codebase. I will prefer to reduce the change impact and will keep `entityFileName`. However, we should also not remove the `entityServiceFileName` variable and it's current assignment to be equal to the `entityFileName` as it might have been used in a module/blueprint. We can do the clean in the next major release.

Changed the patch to remove entityServiceFileName.

generators/entity/index.js
```diff
@@ -667,7 +667,6 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 context.entityModelFileName = context.entityFolderName;
                 context.entityParentPathAddition = this.getEntityParentPathAddition(context.clientRootFolder);
                 context.entityPluralFileName = entityNamePluralizedAndSpinalCased + context.entityAngularJSSuffix;
-                context.entityServiceFileName = context.entityFileName;
```
vishal423: I think you didn't read this part of comment 'However, we should also not remove the entityServiceFileName variable and it's current assignment to be equal to the entityFileName as it might have been used in a module/blueprint. We can do the clean in the next major release.'

Issue #8170, opened 2018-08-27
index.ts and Circular dependency


eosimosu:
##### **Overview of the feature request**

Barrels aka index.ts leads to maintenance hell and circular references that are sometimes extremely difficult to debug. 

##### **Motivation for or Use Case**

index.ts files were very useful prior to introduction of NgModules as they helped manage countless imports, but now they are obsolete and just causes a lot of headaches due to circular dependency that are sometimes extremely difficult to debug. 

More about this issue here: [https://github.com/angular/angular-cli/issues/7369](https://github.com/angular/angular-cli/issues/7369)

I suggest they should be removed in favor of direct import.

##### **Related issues or PR**

- [ x] Checking this box is mandatory (this is just to show you read everything)

deepu105:
You are right, I started adding barrels as per the Angular style guide recomendation at that time and soon realized that it was annoying. But I dont have the time to do the refactoring so if you is willing to do the refactoring you are most welcome.

eosimosu:
Great! I'd do the refactoring

jdubois:
Awesome! Thanks @eosimosu

vishal423:
I think barrels are removed from angular style guide in favor of NgModule, however, I still find them quite useful for cleaner imports. If you follow discipline, you will never run into circular dependency issue.

eosimosu:
@jdubois I started to work on this but haven't had time to finish it.
I'll re-open this issue when it's done.

jdubois:
Oh, awesome -> yes don't hesitate!!!! No pressure on you, when you have time to do it, that's perfect.

kaidohallik:
I opened PR #10185 which removes circular dependencies. I would like to extend this PR to completely remove barrels. Is this ok?

vishal423:
@kaidohallik, please go ahead. Should I re-open this ticket and assign to you?

kaidohallik:
@vishal423, for me it's not necessary to re-open this ticket. I added commit to PR which removes barrels, so hopefully work is done. Let's see if CI tests will pass.

deepu105:
Yes lets remove barrels. I added them since it was official recomendation
from the Angular style guide, but they have removed that recomendation
since I guess.

Thanks & Regards,
Deepu

PR #10185, opened 2019-08-06
remove Angular barrels

kaidohallik:

In issue #8170 was described problem that JHipster Angular modules have circular dependencies.  
This PR removes circular dependencies from JHipster Angular modules:
* added `webpack` plugin `circular-dependency-plugin` to detect circular dependencies
* modified JHipster Angular Typescript files to remove detected circular dependencies


kaidohallik:
CI passed. This PR is ready for review.  
One note: in cleanup, in line https://github.com/jhipster/generator-jhipster/pull/10185/files#diff-8076a6d0a7cff569bfbc9fdcdf900c38R80 I guessed that the version where this PR is landing is 6.3.0. If this PR will be in some other version then this line needs to be changed.

kaidohallik:
TSLint to ESLint migration caused conflicts. I am resolving these right now.

generators/client/templates/angular/package.json.ejs
```diff
@@ -96,6 +96,7 @@ Remove it once the issue is closed
     "chai-as-promised": "7.1.1",
     "chai-string": "1.5.0",
     <%_ } _%>
+    "circular-dependency-plugin": "5.0.2",
```
vishal423: I believe that the circular dependency issues should be resolved after removing barrel files. Do we still require this plugin? Did you find any circular dependency issues in code after removing barrels?  I will prefer to keep build slim and not include by default unless it provides a value add.

kaidohallik: No, after removing barrels there is no circular dependencies any more.  
This plugin detects circular dependencies. If someone makes change which introduces circular dependency then this plugin catches this. So this plugin ensures code quality.  
But I can remove this. Should I remove this?

vishal423: yes, lets remove from default configuration. We can add such recommendations on [tips](https://www.jhipster.tech/tips/) page

kaidohallik: Ok, I will remove this.

kaidohallik: circular-dependency-plugin removed now.

generators/cleanup.js
```diff
@@ -75,6 +77,29 @@ function cleanupOldFiles(generator) {
         generator.removeFile(`${ANGULAR_DIR}admin/metrics/metrics-modal.component.ts`);
         generator.removeFile(`${CLIENT_TEST_SRC_DIR}spec/app/admin/metrics/metrics-modal.component.spec.ts`);
     }
+    if (generator.isJhipsterVersionLessThan('6.3.0')) {
+        generator.removeFile(`${ANGULAR_DIR}account/index.ts`);
```
vishal423: shouldn't we add a check on client framework to restrict this cleanup to angular app?

generators/cleanup.js
```diff
@@ -75,6 +77,29 @@ function cleanupOldFiles(generator) {
         generator.removeFile(`${ANGULAR_DIR}admin/metrics/metrics-modal.component.ts`);
         generator.removeFile(`${CLIENT_TEST_SRC_DIR}spec/app/admin/metrics/metrics-modal.component.spec.ts`);
     }
+    if (generator.isJhipsterVersionLessThan('6.3.0')) {
```
vishal423: I am not sure if upgrade works correctly at moment. Yesterday, I was not able to get pass this check with `6.3.0` as version was already upgraded and this condition always evaluated to `false`. Were you able to test upgrade?

generators/cleanup.js
```diff
@@ -75,6 +77,29 @@ function cleanupOldFiles(generator) {
...
+        generator.removeFile(`${ANGULAR_DIR}account/index.ts`);
+        generator.removeFile(`${ANGULAR_DIR}admin/index.ts`);
+        generator.removeFile(`${ANGULAR_DIR}core/index.ts`);
+        generator.removeFile(`${ANGULAR_DIR}home/index.ts`);
+        generator.removeFile(`${ANGULAR_DIR}layouts/index.ts`);
+        generator.removeFile(`${ANGULAR_DIR}shared/index.ts`);
+        if (generator.withEntities) {
```
vishal423: I think entity related cleanup should be handled in `entity` sub-generator. Please check `entity/index.js`

generators/cleanup.js
kaidohallik: Yes, this is good idea! When doing PR I manually checked that React part have not such files ATM so that should work correctly. But will add this condition to be absolutely sure that no wrong files are deleted.

generators/cleanup.js
kaidohallik: Yes you're right! When doing PR I couldn't find suitable place in `entity` sub-generator, but now I found `cleanup()` method in `entity` sub-generator and will move entity cleaning into this function.

generators/cleanup.js
kaidohallik: I tested manual upgrade and that worked for me. Did you test manual upgrade or `upgrade` sub-generator?

kaidohallik: I am able to reproduce the problem. I updated `package.json`, set `version` to `6.3.0` and no files were deleted. Will investigate this.

vishal423: I think that's not an issue with your change. To test your change, you can use version as `6.4.0` in this condition and `6.3.0` in `generator-jhipster` package.json (please note this is temporary change to test and shouldn't commit)

generators/cleanup.js
kaidohallik: Change done.

generators/cleanup.js
kaidohallik: Change done.

generators/cleanup.js
kaidohallik: If I tested such way then files were deleted correctly. But I'm not sure if this PR can be merged while this problem is not solved.

generators/entity/index.js
```diff
@@ -1091,6 +1091,9 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 if (this.isJhipsterVersionLessThan('5.0.0')) {
                     this.removeFile(`${constants.ANGULAR_DIR}entities/${entityName}/${entityName}.model.ts`);
                 }
+                if (this.isJhipsterVersionLessThan('6.3.0')) {
+                    this.removeFile(`${constants.ANGULAR_DIR}entities/${context.entityFolderName}/index.ts`);
```
vishal423: can you please also add a client framework check here?

generators/cleanup.js
vishal423: I think that's separate issue and this PR can move ahead

generators/entity/index.js
vishal423: I am merging this as I need this before next upgrade

vishal423:
thanks @kaidohallik

generators/entity/index.js
kaidohallik: Change done. Thanks for pointing.

kaidohallik:
@vishal423 thanks for the review and thanks for merging

PR #10188, opened 2019-08-07
fix url inconsistencies between angular and react/vue

clement26695:


mraible:
Since Angular was implemented before Vue, does it make more sense to change Vue to match Angular?

clement26695:
There is also the inconsistency between React and Angular, so it was more convenient to make the change in one framework than in two :smile: 

But I can definitely change React and Vue.js templates instead, I don't know what makes more sense

vishal423:
I believe it's good practice to include module path and doing so makes it consistent with the `admin` module. However, we should do this for all route paths present under the `account` module and not limit to the current react/vue implementation.

clement26695:
So you would add the "account/" prefix to "register", "login", "logout" too ?

vishal423:
At the moment, I am more keen on the code structure-wise grouping. `login`, `logout` links though are displayed under the `account` menu, however, they aren't part of the account module. Further, I think that the account module routes can be lazy-loaded to further improve application startup performance.

clement26695:
Ok @vishal423, I'll try to add the 'account/' prefix to these urls too. 

I also noticed that there were also inconsistencies in the admin routes: it is 'jhi-metrics' in angular, and 'metrics' in react, without any prefix, for instance. I removed the jhi prefix everywhere, as I thought it was more logical, but I can also do it the other way around, if you think it is better

generators/client/templates/angular/src/main/webapp/app/account/password/password.route.ts.ejs
```diff
@@ -22,7 +22,7 @@ import { UserRouteAccessService } from 'app/core';
 import { PasswordComponent } from './password.component';
 
 export const passwordRoute: Route = {
-    path: 'password',
+    path: 'account/password',
```
vishal423: I think we can move `account` path prefix to `account.route.ts`.

generators/client/templates/angular/src/main/webapp/app/account/register/register.route.ts.ejs
```diff
@@ -21,7 +21,7 @@ import { Route } from '@angular/router';
 import { RegisterComponent } from './register.component';
 
 export const registerRoute: Route = {
-    path: 'register',
+    path: 'account/register',
```
vishal423: I think we can move `account` path prefix to `account.route.ts`.

generators/client/templates/angular/src/main/webapp/app/account/sessions/sessions.route.ts.ejs
```diff
@@ -22,7 +22,7 @@ import { UserRouteAccessService } from 'app/core';
 import { SessionsComponent } from './sessions.component';
 
 export const sessionsRoute: Route = {
-    path: 'sessions',
+    path: 'account/sessions',
```
vishal423: I think we can move `account` path prefix to `account.route.ts`.

generators/client/templates/angular/src/main/webapp/app/account/settings/settings.route.ts.ejs
```diff
@@ -22,7 +22,7 @@ import { UserRouteAccessService } from 'app/core';
 import { SettingsComponent } from './settings.component';
 
 export const settingsRoute: Route = {
-    path: 'settings',
+    path: 'account/settings',
```
vishal423: I think we can move `account` path prefix to `account.route.ts`.

generators/client/templates/react/src/main/webapp/app/modules/account/activate/activate.tsx.ejs
```diff
@@ -30,7 +30,7 @@ const successAlert = (
     <Translate contentKey="activate.messages.success">
       <strong>Your user account has been activated.</strong> Please
     </Translate>
-    <Link to="/login" className="alert-link">
+    <Link to="/account/login" className="alert-link">
```
vishal423: Lets not add the `account` path prefix to `login` route path.

generators/client/templates/react/src/main/webapp/app/shared/layout/menus/account.tsx.ejs
```diff
@@ -39,19 +39,19 @@ const accountMenuItemsAuthenticated = (
     <MenuItem icon="cloud" to="/account/sessions"><Translate contentKey="global.menu.account.sessions">Sessions</Translate></MenuItem>
     <%_ } _%>
   <%_ } _%>
-    <MenuItem icon="sign-out-alt" to="/logout"><Translate contentKey="global.menu.account.logout">Sign out</Translate></MenuItem>
+    <MenuItem icon="sign-out-alt" to="/account/logout"><Translate contentKey="global.menu.account.logout">Sign out</Translate></MenuItem>
```
vishal423: Lets not add the `account` path prefix to `logout` route path.

generators/client/templates/react/src/main/webapp/app/shared/util/url-utils.ts.ejs
```diff
@@ -19,7 +19,7 @@
 export const getLoginUrl = () => {
   const port = location.port ? `:${location.port}` : '';
 
-  // If you have configured multiple OIDC providers, then, you can update this URL to /login.
+  // If you have configured multiple OIDC providers, then, you can update this URL to /account/login.
```
vishal423: this url is pointing to server side path. please don't update here

clement26695:
I updated the PR accordingly to your comments, @vishal423

vishal423:
great work @clement26695

PR #10202, opened 2019-08-09
A pr to fix issue#10195

GitAntinus:
Hey guys, I cloned the io.github.jhipster package, and with pr#10194 , I recompiled the project so I get a full list of spring-configuration-metadata.json. Please add this to generator because it saves a lot of time for inetllij users.

### DIFF
--- a/generators/server/templates/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/generators/server/templates/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1,0 +1,579 @@
+{
+  "groups": [
+    {
+      "name": "jhipster",
+      "type": "io.github.jhipster.config.JHipsterProperties",
+      "sourceType": "io.github.jhipster.config.JHipsterProperties"
+    },
+    {
+      "name": "jhipster.async",
+      "type": "io.github.jhipster.config.JHipsterProperties$Async",
+      "sourceType": "io.github.jhipster.config.JHipsterProperties",
+      "sourceMethod": "getAsync()"
+    },
+    {
+      "name": "jhipster.cache",
+      "type": "io.github.jhipster.config.JHipsterProperties$Cache",
+      "sourceType": "io.github.jhipster.config.JHipsterProperties",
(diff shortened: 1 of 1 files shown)


GitAntinus:
> I don't agree with this MR. You should add the annotation processor in this project instead : https://github.com/jhipster/jhipster/blob/master/jhipster-framework/pom.xml

the problem is that the processor wont be invoked dude. You really need to invoke it before the user of jhipsiter compile their code. Because when the compile their code, even with the processor in dependency, the source code of this project wont be processed so there would be no metadata. Do you see the problem here?

GitAntinus:
> I don't agree with this MR. You should add the annotation processor in this project instead : https://github.com/jhipster/jhipster/blob/master/jhipster-framework/pom.xml

Besides, this metadata file is generated by spring-boot-c-processor with project jhipster-framework.

PierreBesson:
No I think the the metadata will be present generated by maven when publishing the lib, but let's not take my word for it and do some testing.

GitAntinus:
> No I think the the metadata will be present generated by maven when publishing the lib, but let's not take my word for it and do some testing.

there is no such file in any generated project.

PierreBesson:
So I just added the annotation processor to the jhipster-framework pom and ran the build. It generated a new file : https://gist.github.com/PierreBesson/c90ee89336003f1fa835cafc044750cd
Then I think the IDE will have no problem importing this file from the dependency jar.

PierreBesson:
Oh, do you have IDEA community or IDEA ultimate ?

GitAntinus:
> Oh, do you have IDEA community or IDEA ultimate ?

The ultimate version

GitAntinus:
> Oh, do you have IDEA community or IDEA ultimate ?

Anyway, I wish you are right because your workflow is better.

PierreBesson:
I wish I'm right as well but I'm afraid we might have to go with your solution. I will have a look later.

pascalgrimaud:
I didn't look in details but can't we do the same like Spring Boot ?

Issue #10105, opened 2019-07-17
Move from TSLint to ESLint


TheJavaGuy:
##### **Overview of the feature request**

TSLint will very soon move to maintenance mode and, starting from January 1st 2020 stop accepting anything except security fixes. This is a documented [issue](https://github.com/palantir/tslint/issues/4534), also explained in their [blog post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9). I believe that JHipster should move from TSLint to ESLint in the similar timeframe.

##### **Motivation for or Use Case**

For sure TypeScript will continue to be developed, and hopefully JHipster and ESLint too. I think that staying stuck with the tool which will be abandoned soon is not an optimal way ahead.

Tasks:
- [x] Migrate angular code to use ESLint - #10187 
- [x] Migrate reactjs code to use ESLint - #10213 

-   [x] Checking this box is mandatory (this is just to show you read everything)

vishal423:
I think we need to wait until https://github.com/mgechev/codelyzer/issues/763 is closed.

MathieuAA:
We needn't have to wait, as such a migration is kind of easy to do, and works (I've done this migration twice, this wasn't such a big deal).

vishal423:
@jhipster/developers, anyone willing to take. Otherwise, I will spend time over weekend

pascalgrimaud:
I think you can take it. Thanks a lot @vishal423 ! So happy to have you in the team :-)

wmarques:
Is that really urgent ? TSLint looks pretty well used today and there are not a clean migration path documented about this.
Also, some `codelyzer` rules aren't yet implemented in ESLint so we would loose some great lint features on the Angular side.

vishal423:
@wmarques, I believe we can use `eslint-plugin-tslint` to continue execute tslint specific rules. I have started work on this activity with an angular project and might be able to come up with something in next few days..

DanielFran:
@vishal423 typescript-eslint v2 hás been released and have breacking changes. Maybe it is better to use the nee version now? See https://github.com/typescript-eslint/typescript-eslint/commit/05ba26879dd5a5a0e1159951c8b24dc5e0e5cc4a

vishal423:
@DanielFran, thanks for update. Dependency update is a never ending task.

DanielFran:
And do not forget to claim the bug-bounty and close this issue :stuck_out_tongue_winking_eye:

vishal423:
[...]

PR #10213, opened 2019-08-12
react: Migrate from TSLint to ESLint

vishal423:
- Migrate react TSLint linting rules to their equivalent ESLint rules. Related to #10105 
- Removed `rxjs` and `import/export` lint rules.
- Upgrade/cleanup script seems to be broken. I will check on that in a separate PR
---


vishal423:
@jhipster/developers, this PR is also ready for review.

generators/client/files-react.js
```diff
@@ -51,6 +52,10 @@ const files = {
         {
             condition: generator => !generator.skipCommitHook,
             templates: ['.huskyrc']
+        },
+        {
+            condition: generator => generator.websocket === 'spring-websocket',
```
mraible: Why does spring-websocket require TSLint?

vishal423: because we use the following `rxjs` linting rules that are not yet available in `eslint`. If we agree to drop these rules, then, `tslint` dependency can be dropped from the react code.

>     "rxjs-collapse-imports": true,
>     "rxjs-pipeable-operators-only": true,
>     "rxjs-no-static-observable-methods": true,
>     "rxjs-proper-imports": true
>

deepu105: Lets drop them, they are not worth branching out

vishal423: sure. will update to remove `rxjs` linting rules. 

Also, what's your preference on import/export rules like below. These are not well supported by current code and I have to exclude at multiple places specifically in protractor and barrels. BTW, we don't have such rules on `angular` front. If we exclude these, then, one more less dependency to maintain :smile:

>     "import/no-internal-modules": [ "error", {
>         "allow": ["app", "test", "uuid/v4", "@fortawesome/**", "lodash/**", "react-toastify/**"]
>       }],
>     "import/no-extraneous-dependencies": [ "error", {
>         "devDependencies": true
>       }],
>     "import/no-unresolved": "off"

deepu105: Feel free to modify them as you see fit if it helps with maintainability.

vishal423: PR updated. @mraible, @deepu105 do you have further suggestions?

PR #10175, opened 2019-08-01
Upgrade Travis pipeline to OpenJDK11

SudharakaP:
I've removed the jdk option of travis; this I believe is useless since it's only used when used in conjunction with ```language: java```. 

Also I've changed the linux distro to bionic since the default distro used by travis seems to be trusty and that doesn't have openjdk11 preinstalled (bionic does). This I believe makes the travis pipeline cleaner so that we don't have to manually install stuff on it and also could get rid of the ```jdk_switcher use oraclejdk8``` part (which seems redundant anyways).

-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10216, opened 2019-08-13
Add support for Java11 in GAE

### LINKED ISSUE #10196: Add Java11 support to Google App Engine Generator

Currently the GAE generator can generate deployment files for java 8 (https://cloud.google.com/appengine/docs/standard/java/runtime). However starting recently GAE supports java 11 (https://cloud.google.com/appengine/docs/standard/java/runtime). I suggest we develop a branch of our GAE generator to support java 11. My idea is to initially give the user a choice of choosing between java  8 and 11 and generate the files based on the choice. 

On another note it is possible to run our jhipster project on java 11 on GAE standard as I've done on [>>this<<](https://github.com/python-sinhala-education-society/OpenLearnr) project. The only things we need is to remove the ```appengine-web.xml``` and add a ```app.yml```. Complete migration summary is given here; https://cloud.google.com/appengine/docs/standard/java11/java-differences

If this is something useful, I can help with doing the PR for this. Do we need to fix the current GAE generator before attempting this? 

jhipster supports java 11 and recently we migrated the Travis build process to java 11 as well (https://github.com/jhipster/generator-jhipster/pull/10175), it seems obvious that we need to add support for java 11 GAE generators at some point. 

#10175, #9605 

SudharakaP:
This migrates our existing GAE generator from Java 8 to Java 11. Note that this will remove all support for Java 8 GAE config file generation (in other words after this fix is merged users will only be able to generate Java 11 GAE configs). I thought this is the best since giving users the ability to choose between Java 8 and Java 11 seems messy since we have migrated other parts of jhipster-generator (such as the travis build to java 11) completely. Also the configs and the way GAE deployments work is very different from Java 8 to Java 11; and Java 11 uses JAR deployment which simplifies matters a lot.  

Closes #10196 


SudharakaP:
I should also add a note here; according to my experience if you try to deploy a jhipster application on a F1 instance in GAE (java11) the memory is somehow not adequate and thus it won't work. If anyone could reproduce this and verify we should probably put a warning in our generator highlighting this. And interestingly if you look at any example out there by the about Spring Boot application deployment in GAE (java 11) they use the F2 environment (https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/appengine-java11/springboot-helloworld/src/main/appengine/app.yaml). 😄

generators/ci-cd/templates/travis.yml.ejs
```diff
@@ -44,6 +43,8 @@ env:
     - JHI_DISABLE_WEBPACK_LOGS=true
     - NG_CLI_ANALYTICS="false"
 before_install:
+  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
+  - source install-jdk.sh --feature 11
```
PierreBesson: Please don't edit this file, this is @pascalgrimaud job as build master to decide on the process to install the jdk, please don't use some random script.

PierreBesson:
+1 for using only java 11 !

pascalgrimaud:
Yes, it has been discussed here: https://groups.google.com/forum/#!topic/jhipster-dev/KJ1Z0Gy0568
So let's remove Java8 support for CICD / Cloud Sub Generator
Only keep compatibility for the generated code

generators/ci-cd/templates/travis.yml.ejs
SudharakaP: @PierreBesson : Sorry about this. I should have mentioned this at https://github.com/jhipster/generator-jhipster/pull/10175 where we changed the travis pipeline to use openjdk 11. Later I've realized that perhaps it was better to use the install-jdk script rather than relaying on the default jdk of bionic (which probably can change over time).  

I'll revert this file here; and we'll discuss this in that pull request.

pascalgrimaud: @SudharakaP : I'd prefer to be consistent and do the same than our current build (for generator-jhipster). See here https://github.com/jhipster/generator-jhipster/blob/master/.travis.yml#L72-L81
Is it fine for you ?

SudharakaP: @pascalgrimaud : Yes, I am fine with that. Let me make that change to https://github.com/jhipster/generator-jhipster/pull/10175

PierreBesson:
I just tested it and it doesn't work for me. Also this is still using old versions of the app-engine plugins for gradle and maven (v1.3.3, it should be using v2.1.0).
It seems that the relevant docs is here:
- https://cloud.google.com/appengine/docs/standard/java/tools/migrate-maven
- https://cloud.google.com/appengine/docs/standard/java/tools/migrate-gradle

SudharakaP:
@PierreBesson : Sorry about that, I've missed the app-engine plugins. Although it's interesting since for me it worked. Let me do some more tests in the evening to fix this. 😈

SudharakaP:
@PierreBesson : Could you please give some more info on what didn't work for you exactly? I am doing some tests locally and I think I have do some more testing after deleting the app-engine plugins installed in my machine, so that all paths are covered. 😃

SudharakaP:
@PierreBesson : I've tested this several times, but couldn't figure out what doesn't work yet; for me it works it seems. Would you be able to provide some more info on what failed for you?

Issue #10223, opened 2019-08-15
blueprints not found in Windows


kaidohallik:
##### **Overview of the issue**

Since #10004 it's not possible to use blueprints in `Windows`.  
Error occurs in all tested shells: `Command Prompt`, `Git Bash` and `PowerShell`.
Not tested in Mac and Linux but as no problems have reported then I guess blueprints are working in these operating systems.

##### **Motivation for or Use Case**

Want to use blueprints also in Windows.

##### **Reproduce the error**

Install blueprint: `npm install -g generator-jhipster-vuejs`  
Run JHipster with blueprint: `jhipster --blueprint vuejs`  
This causes the following behavior:
* command line window is deactivated, some other window is activated
* if coming back to command line window then the following error is displayed:
```
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: The generator-jhipster-vuejs blueprint provided is not installed. Please install it using command npm i -g generator-jhipster-vuejs.
    at Environment.error (C:\projects\kaido-github\generator-jhipster\node_modules\yeoman-environment\lib\environment.js:176:40)
    at module.exports.error (C:\projects\kaido-github\generator-jhipster\generators\generator-base.js:1448:18)
    at shelljs.exec (C:\projects\kaido-github\generator-jhipster\generators\generator-base-private.js:845:22)
    at C:\projects\kaido-github\generator-jhipster\node_modules\shelljs\src\exec.js:121:9
    at ChildProcess.exithandler (child_process.js:285:7)
    at ChildProcess.emit (events.js:189:13)
    at maybeClose (internal/child_process.js:970:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
Emitted 'error' event at:
    at Environment.error (C:\projects\kaido-github\generator-jhipster\node_modules\yeoman-environment\lib\environment.js:178:15)
    at module.exports.error (C:\projects\kaido-github\generator-jhipster\generators\generator-base.js:1448:18)
    [... lines matching original stack trace ...]
    at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
```

##### **Related issues**

#10004

##### **Suggest a Fix**

In Windows I could fix this by changing line https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-base-private.js#L837 (added `node` at the beginning of the command) from
```
shelljs.exec(`${yoInternalCliPath} --generators`, { silent: true }, (err, stdout, stderr) => {
```
to
[...]

PR #10224, opened 2019-08-15
In entity cleanup check if the application is an Angular application

kaidohallik:
Follow up to #10185  

In Angular entity barrel cleanup Angular condition was missing before.  

In Angular client barrels cleanup Angular was checked from `Yeoman` config property `clientFramework` value but if using blueprint then actual client framework can differ.  
If I tested `vuejs` blueprint then in Windows after fixing #10223 I followed that route to get `Vue` app up and running:
* generate initial version with `jhipster --blueprint vuejs` - client framework question offered choices `Angular`, `React` and `No client`, choosing `Angular`, an Angular app was generated
* delete all files except `.yo-rc.json`
* run again `jhipster --blueprint vuejs` - now working `Vue` app is generated, but in `.yo-rc.json` is `clientFramework` value `angularX`

ping @vishal423

### DIFF
--- a/generators/cleanup.js
+++ b/generators/cleanup.js
@@ -75,7 +75,7 @@ function cleanupOldFiles(generator) {
         generator.removeFile(`${ANGULAR_DIR}admin/metrics/metrics-modal.component.ts`);
         generator.removeFile(`${CLIENT_TEST_SRC_DIR}spec/app/admin/metrics/metrics-modal.component.spec.ts`);
     }
-    if (generator.isJhipsterVersionLessThan('6.3.0') && generator.isAngularAppAccordingToPackageJson()) {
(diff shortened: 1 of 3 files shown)


vishal423:
> generate initial version with jhipster --blueprint vuejs - client framework question offered choices Angular, React and No client, choosing Angular, an Angular app was generated

Above should never happen and if that's the case, then, we should track that via separate issue. I think there are already issues related to that. As far as the cleanup script is concerned, let's keep that simple and rely on `clientFramework` value.

kaidohallik:
Ok, I removed check based on `package.json`. Now the only change from this PR is adding Angular check via `clientFramework` in entity cleanup.

PR #10241, opened 2019-08-18
Init test beans lazily

rabiori:

Something we did in our project, which cut down our test-time by ~1 min - I though it might be helpful to other devs.

I did not create an issue yet, I wanted to see if this seems like a welcome addition or not.

### DIFF
--- a/generators/server/templates/src/test/java/package/config/TestLazyBeanInitConfiguration.java.ejs
+++ b/generators/server/templates/src/test/java/package/config/TestLazyBeanInitConfiguration.java.ejs
@@ -1,0 +1,49 @@
+<%#
+ Copyright 2013-2019 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see https://jhipster.github.io/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+package <%=packageName%>.config;
+
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import java.util.Arrays;
+
+/**
+ * This class allows you to run your tests without initializing unneeded beans.
+ * <br/>
+ * In order to override this for any specific test, add to it: {@code @ActiveProfiles(TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)}
+ * <br/>
+ * When working with spring-boot 2.2.0 or later, should be replaced with property: <code>spring.main.lazy-initialization</code>.
+ * <br/>
+ * See <a href="https://spring.io/blog/2019/03/14/lazy-initialization-in-spring-boot-2-2">lazy-initialization-in-spring-boot-2-2</a>.
+ */
+@Component
+@Profile("!" + TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)
+public class TestLazyBeanInitConfiguration {
+    public static final String EAGER_BEAN_INIT = "eagerBeanInit";
+
+    @Override
+    public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+        Arrays.stream(beanFactory.getBeanDefinitionNames())
+            .map(beanFactory::getBeanDefinition)
+            .forEach(beanDefinition -> beanDefinition.setLazyInit(true));
+    }
+}
--- a/generators/server/templates/src/test/java/package/config/TestLazyBeanInitConfiguration.java.ejs
+++ b/generators/server/templates/src/test/java/package/config/TestLazyBeanInitConfiguration.java.ejs
@@ -37,8 +37,8 @@ import java.util.Arrays;
  */
 @Component
 @Profile("!" + TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)
-public class TestLazyBeanInitConfiguration {
-    public static final String EAGER_BEAN_INIT = "eagerBeanInit";
+public class TestLazyBeanInitConfiguration implements BeanFactoryPostProcessor {
+    public static final String EAGER_BEAN_INIT = "eager-bean-init";
 
     @Override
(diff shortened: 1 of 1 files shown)


rabiori:
@jkutner @jdigger @gunnarahlberg @pascalgrimaud - sorry to tag you guys, but I think this PR got lost somehow?

gmarziou:
Interesting, I'm sure JHipster team would love to cut down continuous integration time.
Few comments:

- documentation would be required (jhttps://www.jhipster.tech/profiles/#-profiles and project's README.md)
- where do you enable "eagerBeanInit" profile? In maven or gradle, in application properties?
- profile name (camel-cased) is quite different other JHipster profiles (kebab-cased). It should probably be renamed as "eager-bean-init"

generators/server/templates/src/test/java/package/config/TestLazyBeanInitConfiguration.java.ejs
```diff
@@ -0,0 +1,49 @@
...
+ * <br/>
+ * In order to override this for any specific test, add to it: {@code @ActiveProfiles(TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)}
+ * <br/>
+ * When working with spring-boot 2.2.0 or later, should be replaced with property: <code>spring.main.lazy-initialization</code>.
+ * <br/>
+ * See <a href="https://spring.io/blog/2019/03/14/lazy-initialization-in-spring-boot-2-2">lazy-initialization-in-spring-boot-2-2</a>.
+ */
+@Component
+@Profile("!" + TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)
+public class TestLazyBeanInitConfiguration {
```
vishal423: shouldn't you need to implement `BeanFactoryPostProcessor`?

rabiori: you are completely right, probably got lost on the way to the PR, I'll return it

rabiori:
> Interesting, I'm sure JHipster team would love to cut down continuous integration time.
> Few comments:
> 
> * documentation would be required (https://www.jhipster.tech/profiles/#-profiles and project's README.md)
> * where do you enable "eagerBeanInit" profile? In maven or gradle, in application properties?
> * profile name (camel-cased) is quite different from other JHipster profiles (kebab-cased). It should probably be renamed as "eager-bean-init"

In our project, in any test that we wanted to eagerly init the beans I added the following:
```
@ActiveProfiles(TestLazyBeanInitConfiguration.EAGER_BEAN_INIT)
```
You can pass it by gradle or maven, but the way I implemented this means that by default all beans will be lazy, and any specific test that requires otherwise can change it.

Regarding the profile name, I changed it as commented.

gmarziou:
Great but I did not understand how and where you would document it?

gmarziou:
I ran a little test with jhipster-sample-app-microservice app and 8 entities (https://github.com/jhipster/jdl-samples/blob/master/Oracle-Human-Resources-sample.jdl) `mwnw verify` ran in 43 s instead of 45 s.

rabiori:
> I ran a little test with jhipster-sample-app-microservice app and 8 entities (https://github.com/jhipster/jdl-samples/blob/master/Oracle-Human-Resources-sample.jdl) `mwnw verify` ran in 43 s instead of 45 s.

Perhaps in our code we have many more modules.
I don't see the harm in merging this, but if you think this is superfluous then just let me know.

gmarziou:
I don't think it's superfluous, I'm sure it helps when number of entities grows.

When JHipster upgrades to Spring Boot 2.2,  should we remove this class in favor of  setting `spring.main.lazy-initialization` property to true in `src/test/resources/application.yml` ?

Blackdread:
Yes for `spring.main.lazy-initialization`

rabiori:
> I don't think it's superfluous, I'm sure it helps when number of entities grows.
> 
> When JHipster upgrades to Spring Boot 2.2, should we remove this class in favor of setting `spring.main.lazy-initialization` property to true in `src/test/resources/application.yml` ?

indeed, and I added it to the javadoc of the class as well:
```
When working with spring-boot 2.2.0 or later, should be replaced with property: <code>spring.main.lazy-initialization</code>.
```

gmarziou:
> indeed, and I added it to the javadoc of the class as well
Yes this is where I read it. :)

So I'm wondering whether it makes sense to add this class if we are going to remove it as soon as we upgrade to SB 2.2. which should be soon I suppose. @pascalgrimaud @jdubois @deepu105 what do you think?

murdos:
> 
> So I'm wondering whether it makes sense to add this class if we are going to remove it as soon as we upgrade to SB 2.2. which should be soon I suppose. @pascalgrimaud @jdubois @deepu105 what do you think?

This is exactly what I was thinking. IMO we should just wait for SB 2.2 and don't include something that we know we'll remove soon.

DanielFran:
+1 to wait SB 2.2

vishal423:
I checked few builds(~14 entities) with and without this change to see actual performance gain, however, couldn't see any specific gain.
### Build including this change:
https://dev.azure.com/jhipster/generator-jhipster/_build/results?buildId=5116&view=logs
#### Backend tests took around `3m 21s`

### Build before this change:
https://dev.azure.com/jhipster/generator-jhipster/_build/results?buildId=5115
#### Backend tests took around `3m`

jdubois:
I would rather wait for Spring Boot 2.2 : I did the same trick earlier, but for running my app (not just tests!) and I had several annoying corner cases, which made me lose a lot of time.
So if we push this to people, I'm afraid they might have the same issues: using the "official" Spring Boot 2.2 option should make this easier to troubleshoot. Then, I haven't checked how the Spring team does it, and if they indeed take into consideration those corner cases (basically you need to eager-init some beans, so you need a specific config for that).

rabiori:
> I would rather wait for Spring Boot 2.2 : I did the same trick earlier, but for running my app (not just tests!) and I had several annoying corner cases, which made me lose a lot of time.
> So if we push this to people, I'm afraid they might have the same issues: using the "official" Spring Boot 2.2 option should make this easier to troubleshoot. Then, I haven't checked how the Spring team does it, and if they indeed take into consideration those corner cases (basically you need to eager-init some beans, so you need a specific config for that).

That's exactly the reason I did this in the tests only, and configured an "out" for when someone needed it.

If you intend to upgrade to 2.2 soon (you have a ticket for that?), I wouldn't merge this.
I just implemented this as a temporary measure, since I wasn't sure when you intended to upgrade.

atomfrede:
I think we could write a small [tip in our tips and tricks](https://www.jhipster.tech/tips/) section as end of month the first 2.2 release candidate should be released: https://twitter.com/snicoll/status/1173586725131739136
 So I would close this in favor of boot 2.2. @rabiori are you willing to provide a tip for all users that can't upgrade when 2.2 is released?

atomfrede:
I have created a tip, so let's close this PR and see how boot 2.2. works out.

See  https://github.com/jhipster/jhipster.github.io/pull/810

pascalgrimaud:
thanks for the proposal
I'm closing this as it goes to a tips rather than directly to the main generator-jhipster

Issue #9934, opened 2019-06-18
Migrate away from Jackson Afterburner


jdubois:
We've been using [Jackson Afterburner](https://github.com/FasterXML/jackson-modules-base/tree/master/afterburner) for quite a long time: it has a small performance cost at startup, but can make JHipster applications about 10% more performant overall than vanilla Spring Boot applications (One of the many tricks we use!!).

Unfortunately it gives an ugly warning message at startup since Java 9, see https://github.com/FasterXML/jackson-modules-base/issues/37

Given the way Afterburner works, I'm not sure this can be solved, so we need to find alternative solutions. It looks like https://github.com/stevenschlansker/jackson-blackbird (ping @stevenschlansker ) is promising, we could offer them our help, WDYT?

stevenschlansker:
Hi @jdubois , I'm very excited you've found my project!  It's a proof of concept right now, although we do run it in production on our own systems.  I believe the approach is flexible and should be much easier to extend looking forward (as the `LambdaMetafactory` hopefully gets more powerful e.g. generated field accessors).  Bytecode manipulation was always the most powerful approach but it's a PITA to be the maintainer!

I would be excited to collaborate on taking the blackbird project forward :)

jdubois:
Thanks @stevenschlansker  ! I didn't understand that Blackbird required JDK 11... Unfortunately we need to support JDK 8 for some time still, as we have many users who are locked on it, and that would be a stopper issue for us... Can you check the JDK version and only activate it when JDK 11 is found? Otherwise that's something we could on our side, and switch between Afterburner and Blackbird depending on the JDK.

jdubois:
I'm adding a bug bounty here, as this is an important added value for JHipster users.

SudharakaP:
@jdubois : I've tested jackson-blackbird on my local with Java 11 and it works as expected. I can do a pull request for this but would like to discuss my approach before doing it. 😄 

1. We can prompt the user to choose between Java 8 and 11 in the generator and based on that set the [JAVA_VERSION](https://github.com/jhipster/generator-jhipster/blob/29592ab0bd158ecd0829c406d594537a4f91b7d8/generators/server/index.js#L120) variable to the appropriate java version. This will enable us to choose between jackson-afterburner and jackson-blackbird when generating the initial pom. Does this sound good or any suggessions? 

[...]

PR #10244, opened 2019-08-19
upgrade to gradle 5.6 and using plugin management

atomfrede:
* regular upgrade to latest and final gradle 5.x version
* using the new [plugin management](https://docs.gradle.org/5.6/userguide/plugins.html#sec:plugin_version_management) to define plugin versions via properties
* removed the [propdeps plugin](https://github.com/spring-gradle-plugins/propdeps-plugin) as (at least) we don't use it directly and it seems quite outdated

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10245, opened 2019-08-19
Jackson BlackBird Support (for Java 11)

SudharakaP:
This adds ```jackson-blackbird``` support for java 11 as discussed in #9934. 

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -22,6 +22,7 @@ const statistics = require('../statistics');
 module.exports = {
     askForInsightOptIn,
     askForApplicationType,
+    askForJavaVersion,
     askForModuleName,
     askFori18n,
     askForTestOpts,
@@ -215,4 +216,51 @@ function askModulesToBeInstalled(done, generator) {
             done();
         }
     );
+}
+
+function askForJavaVersion() {
+    if (this.existingProject) {
+        return;
+    }
+
+    const choices = [{ name: 'Java 11', value: '11' }, { name: 'Java 8', value: '1.8' }];
+    const defaultChoice = [{ name: 'Java 11', value: '11' }];
+    const PROMPT = {
+        type: 'list',
+        name: 'javaVersion',
+        message: 'Which Java version would you like to use?',
+        choices,
+        default: defaultChoice
+    };
+
+    const done = this.async();
+
+    this.prompt(PROMPT).then(prompt => {
+        this.javaVersion = prompt.javaVersion;
+        if (prompt.javaVersion === '11') {
+            askJacksonBlackBirdToBeInstalled(done, this);
+        } else {
+            done();
+        }
+    });
+}
+
+function askJacksonBlackBirdToBeInstalled(done, generator) {
+    const PROMPT = {
+        type: 'input',
+        name: 'useJacksonBlackBird',
+        message: 'Add jackson-blackbird for increased performance (this is a experimental library) (Y/N) ?',
+        default: this.useJacksonBlackBird ? this.useJacksonBlackBird : 'N',
+        validate: input => {
+            if (input !== 'Y' && input !== 'N') {
+                return 'Input should be Y or N';
+            }
+            return true;
+        }
+    };
+
+    generator.prompt(PROMPT).then(prompt => {
+        generator.useJacksonBlackBird = prompt.useJacksonBlackBird;
+        done();
+    });
 }
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -64,7 +64,7 @@
     <properties>
         <!-- Build properties -->
         <maven.version>3.3.9</maven.version>
-        <java.version><%= JAVA_VERSION %></java.version>
+        <java.version><%= javaVersion %></java.version>
 <%_ if (!skipClient) { _%>
         <node.version>v<%= NODE_VERSION %></node.version>
     <%_ if (clientPackageManager === 'npm') { _%>
@@ -206,10 +206,19 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
+<%_ if (javaVersion === '1.8') { _%>
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-afterburner</artifactId>
         </dependency>
+<%_ } _%>
+<%_ if (javaVersion === '11' && useJacksonBlackBird === 'Y') { _%>
+        <dependency>
+            <groupId>org.jdbi.jackson-contrib</groupId>
+            <artifactId>jackson-module-blackbird</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+<%_ } _%>
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -357,6 +357,8 @@ module.exports = class extends BaseGenerator {
                     this.configOptions.websocket = this.options.websocket;
                 }
                 this.configOptions.clientPackageManager = this.clientPackageManager;
+                this.configOptions.javaVersion = this.javaVersion;
+                this.configOptions.useJacksonBlackBird = this.useJacksonBlackBird;
             },
 
             composeServer() {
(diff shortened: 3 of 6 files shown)


generators/server/templates/pom.xml.ejs
```diff
@@ -206,10 +206,19 @@
...
         <dependency>
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-afterburner</artifactId>
         </dependency>
+<%_ } _%>
+<%_ if (javaVersion === '11' && useJacksonBlackBird === 'Y') { _%>
+        <dependency>
+            <groupId>org.jdbi.jackson-contrib</groupId>
+            <artifactId>jackson-module-blackbird</artifactId>
+            <version>0.0.1</version>
```
DanielFran: Can you please create a property for the version: jackson-module-blackbird.version?

generators/app/prompts.js
```diff
@@ -216,3 +217,50 @@ function askModulesToBeInstalled(done, generator) {
...
+            askJacksonBlackBirdToBeInstalled(done, this);
+        } else {
+            done();
+        }
+    });
+}
+
+function askJacksonBlackBirdToBeInstalled(done, generator) {
+    const PROMPT = {
+        type: 'input',
```
cbornet: This should be a type "confirm" using a boolean

generators/app/prompts.js
```diff
@@ -216,3 +217,50 @@ function askModulesToBeInstalled(done, generator) {
...
+        choices,
+        default: defaultChoice
+    };
+
+    const done = this.async();
+
+    this.prompt(PROMPT).then(prompt => {
+        this.javaVersion = prompt.javaVersion;
+        if (prompt.javaVersion === '11') {
+            askJacksonBlackBirdToBeInstalled(done, this);
```
cbornet: Shouldn't this be a filter in the blackbird "validate" prompt ?

DanielFran:
@SudharakaP Gradle settings also require changes, see for example https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/build.gradle.ejs#L63

SudharakaP:
@DanielFran @cbornet : Thanks much for your comments and reviews. I'll fix all of these today evening or tomorrow morning. 😄

generators/server/templates/pom.xml.ejs
SudharakaP: @DanielFran : Thanks; done 😄

generators/app/prompts.js
SudharakaP: @cbornet : Thanks. Done 👍

SudharakaP:
> @SudharakaP Gradle settings also require changes, see for example:
> 
> * https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/build.gradle.ejs#L63
> * https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/build.gradle.ejs#L285
> 
> Note: I suggest to implement those changes on top of #10244 that should be merged soon... or wait for it.

@DanielFran : Thanks for pointing this out. Somehow I forgot about Gradle.

generators/app/prompts.js
SudharakaP: @cbornet : Sorry, I am not sure I understand you here. You mean the the prompt for java version should go after the prompt for blackbird prompt?

SudharakaP:
I've fixed some issues and also got all the tests pass at my end (I believe there was one test failing before). However two tests are failing in the azure pipelines and I don't think it's related to this. Could somebody please restart those two by any chance? 

@DanielFran : Added the Gradle stuff. Please let me know if you see anything else to be changed.

generators/app/prompts.js
cbornet: Sorry I should have said in the "when" field of blackbird prompt, test the Java version (so after the Java version prompt).

generators/server/templates/gradle.properties.ejs
```diff
@@ -35,6 +35,7 @@ hibernate_version=5.3.10.Final
 mapstruct_version=1.3.0.Final
 
 liquibase_hibernate5_version=3.7
+jackson_module_blackbird_version=0.0.1
```
DanielFran: Please add condition <%_ if (javaVersion === '11' && useJacksonBlackBird) { _%>

SudharakaP: Sorry, silly mistake; missed that somehow. Added it 😄

generators/app/prompts.js
SudharakaP: @cbornet : Thanks for the clarification. Now understand what you mean; I haven't done any yeoman generator development and just now realized that ```Inquirer.js``` provides the ```when``` attribute. 

Refactored the code; let me know if you see any further issues. 😄

generators/app/prompts.js
```diff
@@ -216,3 +218,43 @@ function askModulesToBeInstalled(done, generator) {
         }
     );
 }
+
+function askForJavaVersion() {
+    if (this.existingProject) {
+        return;
+    }
+
+    const choices = [{ name: 'Java 11', value: '11' }, { name: 'Java 8', value: '1.8' }];
```
murdos: We also support Java 12 (and may also support Java 13 soon), so we should probably either reword things a bit differently ("Java >= 11") or include other supported java versions.

PierreBesson:
I thought we agreed to not add a new question for java 8 or 11 ! If it's only to add blackbird then I'd rather we wait until JH v7 where we will drop java 8 support.
ping @jdubois

SudharakaP:
@PierreBesson @murdos : One thing off the top of my head; ask the blackbird prompt if ```JAVA_HOME``` is set to Java11+. We can also set the ```java.version``` tag in pom.xml based on ```JAVA_HOME```. What do you think about this approach? I think this way we don't have to wait until JH7.

murdos:
> 
> 
> @PierreBesson @murdos : One thing off the top of my head; ask the blackbird prompt if `JAVA_HOME` is set to Java11+. We can also set the `java.version` tag in pom.xml based on `JAVA_HOME`. What do you think about this approach? I think this way we don't have to wait until JH7.

This approach is not correct, because you can generate a project for a different target than what you have on the machine that is used for generating the project.

SudharakaP:
> > @PierreBesson @murdos : One thing off the top of my head; ask the blackbird prompt if `JAVA_HOME` is set to Java11+. We can also set the `java.version` tag in pom.xml based on `JAVA_HOME`. What do you think about this approach? I think this way we don't have to wait until JH7.
> 
> This approach does not work, because you can generate a project for a different target than what you have on the machine that is used for generating the project.

@murdos : Ah, got it; didn't think about that. Thanks 👍 .

murdos:
>  
> I thought we agreed to not add a new question for java 8 or 11 ! If it's only to add blackbird then I'd rather we wait until JH v7 where we will drop java 8 support.

I vaguely remember this decision, but I have not been able to retrieve the reason of this choice. 

Since we now have a new java version released every 6 months, I think having the question could have some benefits:
- be able to be more specific in source and target versions specified in Maven and Gradle. For example if someone currently want to work with Java 11 language feature, they should manually change their configuration after the project has been generated
- ditto for generated CI stuff
- potentially generate different code and/or dependency (like here) according to the target java version
- explicitly show java versions that jhipster supports
- ...

SudharakaP:
> I vaguely remember this decision, but I have not been able to retrieve the reason of this choice.
> 
> Since we now have a new java version released every 6 months, I think having the question could have some benefits:
> 
> * be able to be more specific in source and target versions specified in Maven and Gradle. For example if someone currently want to work with Java 11 language feature, they should manually change their configuration after the project has been generated
> * ditto for generated CI stuff
> * potentially generate different code and/or dependency (like here) according to the target java version
> * explicitly show java versions that jhipster supports
> * ...

Exactly, we have a [jhipster based project](https://github.com/python-sinhala-education-society/OpenLearnr) and I had to go an change the pom file manually to update it to Java 11. So I do see having a prompt has benefits. 

@murdos : Anyways, I've updated the prompt as you specified. 

Feel free to let me know if there's any other changes you guys could see.

PierreBesson:
Yes I guess the additional prompt for java version could have some use but as version are released more and more frequently, I'd much rather that we support only one version officially in the future as it makes the CI build matrix increase exponentially !  And our CI is already too slow with too many possibilities to test.

Supporting java 8 was necessary because of the huge amount of breaking changes with java 9 but I don't think an upgrade like this will happen in the near future.

So I'm still against the question. I'm appealing to @jdubois, @pascalgrimaud, @deepu105 for the final decision on this.

PR #10252, opened 2019-08-21
Define Typescript aliases in a single place

kaidohallik:
If adding new alias to Typescript then currently must add this alias both to Typescript and Webpack config. After this change it's enough to add alias to Typescript config. Webpack reads aliases directly from Typescript config.

If in current version adding new path only to `tsconfig.json` into section `compilerOptions.paths` then on building app the following error occur: `Module not found: Error: Can't resolve '...'`.

This improves developer experience:
* for these who doesn't know Webpack and Typescript in depth this can save time which it takes to understand and resolve errors if changing only `tsconfig.json`
* for these who know Webpack and Typescript well this avoids annoying double change, for example developers and testers of the ng-jhipster can route generated application to use locally built ng-jhipster with single change

### DIFF
--- a/generators/client/templates/angular/src/test/javascript/jest.conf.js.ejs
+++ b/generators/client/templates/angular/src/test/javascript/jest.conf.js.ejs
@@ -1,3 +1,5 @@
+const tsconfig = require('../../../tsconfig.json');
+
 module.exports = {
     preset: 'jest-preset-angular',
     setupFilesAfterEnv: ['<rootDir>/src/test/javascript/jest.ts'],
@@ -12,9 +14,7 @@ module.exports = {
     coveragePathIgnorePatterns: [
         '<rootDir>/src/test/javascript'
     ],
-    moduleNameMapper: {
-        'app/(.*)': '<rootDir>/<%= CLIENT_MAIN_SRC_DIR %>app/$1'
-    },
+    moduleNameMapper: addAliasesFromTypescriptConfig({}),
     reporters: [
         'default',
         [ 'jest-junit', { output: './<%= BUILD_DIR %>test-results/TESTS-results-jest.xml' } ]
@@ -25,3 +25,12 @@ module.exports = {
     rootDir: '../../../',
     testURL: 'http://localhost/'
 };
+
+function addAliasesFromTypescriptConfig(initialAliases) {
+  return Object.keys(tsconfig.compilerOptions.paths).reduce((aliases, aliasName) => {
+    const finalAliasName = aliasName.replace(/(.*)(\/\*)$/, '$1/(.*)');
+    const finalPath = tsconfig.compilerOptions.paths[aliasName][0].replace(/(.*)(\/\*)$/, '$1/\$ 1').replace(/\/\$ 1$/, '/$1');
+    aliases[finalAliasName] = `<rootDir>/${finalPath}`;
+    return aliases;
+  }, initialAliases);
+}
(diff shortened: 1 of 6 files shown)


generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -26,16 +26,20 @@ const MergeJsonWebpackPlugin = require("merge-jsons-webpack-plugin");
...
 
 module.exports = (options) => ({
     resolve: {
         extensions: ['.ts', '.js'],
         modules: ['node_modules'],
         mainFields: [ 'es2015', 'browser', 'module', 'main'],
-        alias: {
-            app: utils.root('<%= MAIN_SRC_DIR %>app/')
-        }
+        alias: Object.keys(tsconfig.compilerOptions.paths).reduce((aliases, aliasName) => {
```
vishal423: can you move this function to `utils`?

kaidohallik:
> Great initiative to simplify workflow. Can you also do similar change in `jest.conf.js` `moduleNameMapper` to map aliases?

Thanks for pointing this out! It's done now.  
If developing `ng-jhipster` locally and switching to this in `tsconfig.json` then before this change in tests was still used `ng-jhipster` from `node_modules`. Now in tests is also used locally built `ng-jhipster`.

generators/client/templates/angular/webpack/utils.js.ejs
```diff
@@ -28,3 +31,12 @@ function root(args) {
   args = Array.prototype.slice.call(arguments, 0);
   return path.join.apply(path, [_root].concat(args));
 }
+
+function addAliasesFromTypescriptConfig(initialAliases) {
```
vishal423: Can we instead write like below to separate map logic from reduction?
```
function mapTypescriptAliasToWebpackAlias(alias = {}) {
  const webpackAliases = { ...alias };
  Object.entries(tsconfig.compilerOptions.paths)
    .map(([key, value]) => {
      const regexToReplace = /\/\*$/;
      const aliasKey = key.replace(regexToReplace, '');
      const aliasValue = value.length ? value[0].replace(regexToReplace, '') : value.replace(regexToReplace, '');
      return [aliasKey, root(aliasValue)];
    })
    .reduce((aliases, [key, value]) => (aliases[key] = value), webpackAliases);
  return webpackAliases;
}
```

generators/client/templates/angular/src/test/javascript/jest.conf.js.ejs
```diff
@@ -25,3 +25,12 @@ module.exports = {
     rootDir: '../../../',
     testURL: 'http://localhost/'
 };
+
+function addAliasesFromTypescriptConfig(initialAliases) {
```
vishal423: can we write it like other comment? Also, can we simplify the usage of regex?

generators/client/templates/angular/webpack/utils.js.ejs
kaidohallik: I used ideas from this snippet and made code more readable. Thanks for the idea!

generators/client/templates/angular/src/test/javascript/jest.conf.js.ejs
kaidohallik: Function rewritten as in Webpack part and regex usage simplified.

vishal423:
thanks @kaidohallik

PR #10254, opened 2019-08-21
angular: Bumped library versions, remove acorn dependency

### LINKED ISSUE #10209: Update dependency on ng-bootstrap to 5.1 for Angular

Update the dependency on ng-bootstrap to 5.1 for Angular. ng-bootstrap 5.x is compatible with Angular 8.x so this should be possible.

Allow jhipster user to get the newest features and bugfixes in ng-bootstrap and stay up-to-date with this dependency.

At the moment, this is not possible out of the box, because https://github.com/jhipster/ng-jhipster includes code that depends on ng-bootstrap 4, and there were some small breaking changes in ng-bootstrap 5. However, I don't think this would be much effort and it would be highly appreciated.

I was not able to find related issues

vishal423:
- Closes #10209 
- Bumped `angular` and associated library versions.
- Removed direct dependency on `acorn` that was added to fix dynamic import issue. Related to https://github.com/FortAwesome/angular-fontawesome/blob/master/docs/upgrading/0.4.0-0.5.0.md
- Updated usage of font awesome library to fix warnings. Related to https://github.com/FortAwesome/angular-fontawesome/blob/master/docs/upgrading/0.4.0-0.5.0.md

---

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
@@ -16,15 +16,16 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import './vendor.ts';
-
+import { HTTP_INTERCEPTORS } from '@angular/common/http';
 import { NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
-import { HTTP_INTERCEPTORS } from '@angular/common/http';
+import { FaIconLibrary } from '@fortawesome/angular-fontawesome';
+import { fas } from '@fortawesome/free-solid-svg-icons';
 import { NgbDatepickerConfig } from '@ng-bootstrap/ng-bootstrap';
-import { NgxWebstorageModule } from 'ngx-webstorage';
 import { NgJhipsterModule } from 'ng-jhipster';
+import { NgxWebstorageModule } from 'ngx-webstorage';
 
+import { icons } from './vendor';
 <%_ if (authenticationType === 'jwt') { _%>
 import { AuthInterceptor } from './blocks/interceptor/auth.interceptor';
 <%_ } _%>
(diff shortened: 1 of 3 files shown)


vishal423:
@jhipster/developers, This PR is ready for review.

generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
```diff
@@ -16,15 +16,16 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import './vendor.ts';
```
kaidohallik: Currently README.md file states that you can add dependencies into vendor.ts: https://github.com/jhipster/generator-jhipster/blob/master/generators/common/templates/README.md.ejs#L171  
I this still working after removing this line?

vishal423: I have re-organized the imports. You can check this below below:

```import { icons } from './vendor';```

kaidohallik: I tested Leaflet and yes, this is working after this change. For me it was confusing that `import { icons } from './vendor';` adds all libraries support from `vendor.ts` behind the scenes.  
So in my testing all is working.

vishal423: Good to hear that everything is working fine. Regarding inclusion of libraries, `webpack` handles that for you. I think that's how module bundlers work. Imports defined in the vendor module are internal to that module unless they are exported to be imported into other modules.

kaidohallik: I tried to modify code in different ways and found out that if I define variable `const icons2 = ...` in module `app.module.ts` similar to variable `const icons = ...` in module `vendor.ts` and use variable `icons2` instead of `icons`: `iconLibrary.addIcons(...icons2);` then compiler is dropping whole `vendor.ts` file due to optimization and Bootstrap improvement and Leaflet support is gone.  
So this vendor import change seems a little confusing but without modifying code all is working well.

vishal423: As long as you are importing a symbol or default import and using that imported symbol in your current module, it should work fine. If you don't want to use imported symbol, then, you can always fallback to the global import. I don't think we can support both.

kaidohallik: I believe it's cleaner and better understandable to add new file for exporting FontAwesome icons  array constant, for example `layouts/font-awesome/font-awesome.constants.ts`, and restore global vendor import.

PR #10274, opened 2019-08-24
make service layer not depend on web layer

atomfrede:
As already proposed in [this sample](https://github.com/atomfrede/jh-9876) this PR does the following:

* Adding [archunit](https://www.archunit.org/) to check that the service layer does not depend on the web layer
* Add new exceptions in service layer package which are used by the userservice instead of the web excetions/problems
* Instead of renaming the web exceptions to something like `...WebException` or `...Problem` I just kept the name and use in the translator a fully qualified classname. Otherwise we would have had to rename/remove the all exceptions, change all imports elsewhere. With the current approach updating e.g. is easy and the diff will be pretty small.

I will take care of adding a paragraph about arch unit to our documentation after it is merged.

### DIFF
--- a/generators/server/templates/src/main/java/package/service/LoginAlreadyUsedException.java.ejs
+++ b/generators/server/templates/src/main/java/package/service/LoginAlreadyUsedException.java.ejs
@@ -1,0 +1,27 @@
+<%#
+ Copyright 2013-2019 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see https://www.jhipster.tech/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+package <%=packageName%>.service;
+
+public class LoginAlreadyUsedException extends RuntimeException {
+
+    public LoginAlreadyUsedException() {
+        super("Login name already used!");
+    }
+
+}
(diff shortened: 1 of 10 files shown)


generators/server/templates/src/main/java/package/service/LoginAlreadyUsedException.java.ejs
```diff
@@ -0,0 +1,27 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+package <%=packageName%>.service;
+
+public class LoginAlreadyUsedException extends RuntimeException {
```
mraible: UsernameAlreadyusedException seems like a better name because you don't have to guess what login means.

generators/server/templates/src/test/java/package/ArchTest.java.ejs
```diff
@@ -0,0 +1,45 @@
...
+
+        noClasses()
+            .that()
+                .resideInAnyPackage("..service..")
+            .or()
+                .resideInAnyPackage("..repository..")
+            .should().dependOnClassesThat()
+                .resideInAnyPackage("..<%=packageName%>.web..")
+        .because("Services and repositories should not depend on web layer")
+        .check(importedClasses);
```
mraible: Cool!

generators/server/templates/src/main/java/package/service/LoginAlreadyUsedException.java.ejs
atomfrede: Can do, just used the names we have already in `web/errors` package

atomfrede:
Seems I forgot the integration tests ti adapt :man_facepalming: Going to check that

atomfrede:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/9745

jdubois:
Thanks so much @atomfrede ! I didn't know about archtest, this looks like an awesome addition to JHipster! I'll have a look ASAP

PR #10281, opened 2019-08-26
Add support for scheduled audit event deletion

### LINKED ISSUE #10253: Add a schedule to delete audit logs after 30 days
The current implementation of audit log for authentication could result in DB becoming too large, so we should implement a simple @scheduled job to delete the content which are older than 30 days

SudharakaP:
Related to https://github.com/jhipster/jhipster/pull/379

Fixes #10253

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -1719,6 +1719,10 @@ const serverFiles = {
                     renameTo: generator => `${generator.testDir}service/MailServiceIT.java`
                 },
                 {
+                    file: 'package/service/AuditEventServiceIT.java',
+                    renameTo: generator => `${generator.testDir}service/AuditEventServiceIT.java`
+                },
+                {
                     file: 'package/service/UserServiceIT.java',
                     renameTo: generator => `${generator.testDir}service/UserServiceIT.java`
                 },
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -1534,6 +1534,10 @@ const serverFiles = {
                 {
                     file: 'package/web/rest/AuditResourceIT.java',
                     renameTo: generator => `${generator.testDir}web/rest/AuditResourceIT.java`
+                },
+                {
+                    file: 'package/service/AuditEventServiceIT.java',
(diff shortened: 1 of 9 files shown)


DanielFran:
@SudharakaP Can you add some tests, please?

generators/server/files.js
```diff
@@ -1718,6 +1718,10 @@ const serverFiles = {
                     file: 'package/service/MailServiceIT.java',
                     renameTo: generator => `${generator.testDir}service/MailServiceIT.java`
                 },
+                {
```
DanielFran: Move it to https://github.com/jhipster/generator-jhipster/blob/3d1cc2c2df5599be7ee73247f56f30cd3074fadf/generators/server/files.js#L1534, so this keep consistancy and is created only when needed.

generators/server/templates/src/main/resources/config/application-dev.yml.ejs
```diff
@@ -367,6 +367,8 @@ jhipster:
             host: localhost
             port: 5000
             queue-size: 512
+    audit-events:
```
DanielFran: Should be available only if generator.authenticationType === 'oauth2' && ['sql', 'mongodb'].includes(generator.databaseType)

generators/server/templates/src/main/resources/config/application-prod.yml.ejs
```diff
@@ -351,6 +351,8 @@ jhipster:
             host: localhost
             port: 5000
             queue-size: 512
+    audit-events:
```
DanielFran: Should be available only if generator.authenticationType === 'oauth2' && ['sql', 'mongodb'].includes(generator.databaseType)

DanielFran:
@SudharakaP I confirmed and we have 2 conditions to persist audit, so you need to change again files.js and add in line https://github.com/jhipster/generator-jhipster/blob/master/generators/server/files.js#L1697

See also https://github.com/jhipster/generator-jhipster/blob/master/generators/server/files.js#L1574 when to add properties in application-*.yml

PR #10283, opened 2019-08-26
Add Jackson Blackbird Support

### LINKED ISSUE #9934: Migrate away from Jackson Afterburner

SudharakaP:
This adds jackson-blackbird support for java 11 as discussed in #9934.

Fix #9934


SudharakaP:
Resolved merge conflict coming from master.

SudharakaP:
Travis build fails due to `apt-get` failure; close/opening.

SudharakaP:
Resolved minor conflict due to hibernate downgrade; https://github.com/jhipster/generator-jhipster/pull/10435

SudharakaP:
Just to follow up, is there anything that needs to be done on this further? If so feel free to let me know :smile:

pascalgrimaud:
Yes, the only need is time, sorry...
Just be patient :)

SudharakaP:
@pascalgrimaund : No problems. Please take your time. Just wanted to
confirm. :smile:

pascalgrimaud:
@SudharakaP : thanks for this PR. But adding a new option now is not a good idea. I think we should wait for v7 and include this feature directly, by replacing jackson afterburner
What do you think ?
Until this, I think we can let it opened

SudharakaP:
@pascalgrimaud : Sure, agreed. I'll change this PR to replace Afterburner with Blackbird. You can merge that when v7 is released. :+1:

SudharakaP:
@pascalgrimaud : Okay, I've changed the PR so that Blackbird is added in place of Afterburner. Merge it at your discretion. :smile:

jdubois:
I'm going to do some performance tests to check if this is good.

The repo is at https://github.com/jdubois/jackson-benchmarks

I will run this during the week end and post the results here.

SudharakaP:
@jdubois : Ah, wonderful. Thanks Julien. :smile:

jdubois:
Pinging @stevenschlansker who is the author of Blackbird! This is going to be interesting!

jdubois:
Not a lot of fun here :-(
I did 2 different sets of tests: a very simple one, and one with a lot of users doing a lot of JSON.... The first case didn't show any significant performance difference (Afterburner/Blackbird maybe have a 1% perf increase, nothing you can really measure). And in the second case, results where very close, but in fact a very tiny bit better for the "Vanilla" project.
I'm using AdoptOpenJDK (build 11.0.7+10), maybe there were some improvements which make those projects obsolete??? I don't understand what's happening.

SudharakaP:
@jdubois : Does Blackbird promise much greater performance? I think it's mainly for upgrading Jackson to Java 11. Isn't it? The [documentation mentions](https://github.com/stevenschlansker/jackson-blackbird#performance); 

> Blackbird performance seems to be almost exactly on par with Afterburner - up to 20% better than vanilla Jackson in some cases. At that point the method invocation overhead disappears behind the heat generated from the long parser and UTF8 decoder.

jdubois:
@SudharakaP yes this is supposed to give us better performance.
In fact, in the past, I've run some tests showing 10% performance increase with Afterburner, so I can confirm that those project's claims were real.
Then, I did more tests today, including testing with Java 8: it's very hard to spot any difference between "vanilla", "afterburner" and "blackbird". Also, Blackbird only works with Java 11, and that will cause an issue as many people are still stuck on Java 8.
So I'm extremely disappointed, and also I don't understand why this happens, but it seems that both Afterburner and BlackBird don't make any significant difference. As they add some significant risk and complexity, I would therefore remove them entirely.
Please comment and vote 👍(for removing) or 👎 (for keeping one of them)

SudharakaP:
@jdubois : Thanks for the information. I've done another PR to remove it. When we reach a consensus feel free to merge it. :smile:

stevenschlansker:
I agree, for most "out of the box" uses, neither Afterburner nor Blackbird will show much improvement with modern JVMs.  Reflection has gotten a *lot* faster over the years, and your database-across-a-network-segment is just as far away, probably even more so with all the fancy virtual cloud networking going on :)

It could well be that the most appropriate "out of the box" default is to include neither, with an option to "upgrade" if you happen to have an extraordinarily reflection-heavy workload.  Thanks for your interest!

SudharakaP:
@stevenschlansker : That's interesting. Thanks for your input. :smile: 

As Steven mentioned one thing we could do is to keep the Blackbird option as a optional prompt with defaulting to `no`. :smile:

atomfrede:
> @SudharakaP yes this is supposed to give us better performance.
> In fact, in the past, I've run some tests showing 10% performance increase with Afterburner, so I can confirm that those project's claims were real.
> Then, I did more tests today, including testing with Java 8: it's very hard to spot any difference between "vanilla", "afterburner" and "blackbird". Also, Blackbird only works with Java 11, and that will cause an issue as many people are still stuck on Java 8.
> So I'm extremely disappointed, and also I don't understand why this happens, but it seems that both Afterburner and BlackBird don't make any significant difference. As they add some significant risk and complexity, I would therefore remove them entirely.
> Please comment and vote +1(for removing) or -1 (for keeping one of them)

We have 5 votes for removal, none for keeping it. How should we proceed?

jdubois:
@atomfrede I think we all agree not to use Afterburner or Blackbird. I'm really sad about it, but we have done extensive tests and an agreement from everyone, so let's move on.

PR #10303, opened 2019-08-29
Add support for Caffeine Cache

### LINKED ISSUE #9351: Caffeine cache support

Caffeine provides a higher performance local cache than JHipster's [default provider](https://www.jhipster.tech/using-cache). This is in regards to both [concurrency](https://github.com/ben-manes/caffeine/wiki/Benchmarks) and [hit rates](https://github.com/ben-manes/caffeine/wiki/Efficiency).

The default provider may only be used as a local cache in JHipster's implementation ("... cannot work as a distributed cache"). Therefore the alternative provider should have a positive net impact.

Spring Cache includes [direct support](https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#cache-store-configuration-caffeine) against Caffeine's native api. This has a potential limitation where Spring's property setting only allows for a single configuration (e.g. all caches have the same maximum size), rather than configured per name. This was how Spring wrote their Guava provider and Caffeine's was a direct evolution of that code. There are [alternative configurators](https://github.com/stepio/coffee-boots) or the Java bean configuration to bypass this.

Spring Cache and Hibernate offer JCache integrations, which requires the additional caffeine-jcache dependency. This may be the simplest option to support both.

I should have some free cycles coming up, which means that I can try to do this work if given some guidance and if there is interest. However, I have not used Spring since the 1.x heydays so I lack familiarity in that respect.

SudharakaP:
This add Caffeine cache; related to the pull request; https://github.com/jhipster/jhipster/pull/383 in the jhipster framework side. 

Fixes #9351


DanielFran:
Closed/Reopened

SudharakaP:
Just for the record; documentation is updated as per https://github.com/jhipster/jhipster.github.io/pull/780

PierreBesson:
@pascalgrimaud Maybe we can set one of the integration tests builds (in `test-integration/samples/`) to use Caffeine so that it is active in end to end tests. I propose `ngx-gradle-fr`.

pascalgrimaud:
@PierreBesson : yes we should

SudharakaP:
> @pascalgrimaud Maybe we can set one of the integration tests builds (in `test-integration/samples/`) to use Caffeine so that it is active in end to end tests. I propose `ngx-gradle-fr`.

I've just committed this as well. :smile:

generators/server/templates/pom.xml.ejs
```diff
@@ -125,6 +125,11 @@
         <mapstruct.version>1.3.0.Final</mapstruct.version>
 <%_ if (enableSwaggerCodegen) { _%>
         <jackson-databind-nullable.version>0.2.0</jackson-databind-nullable.version>
+<%_ } _%>
+<%_ if (cacheProvider === 'caffeine') { _%>
+        <caffeine.version>2.8.0</caffeine.version>
```
avdev4j: Does this property is really used?

generators/server/templates/pom.xml.ejs
```diff
@@ -125,6 +125,11 @@
...
+        <jcache.version>2.8.0</jcache.version>
```
avdev4j: Same question for jCache property ;)

generators/server/templates/pom.xml.ejs
SudharakaP: Good catch. I've corrected it by defining caffeine.version and jcache.version explicitly in the dependencies section. Please take a look and let me know if you have any further issues.

generators/server/templates/pom.xml.ejs
SudharakaP: Good catch. I've corrected it by defining caffeine.version and jcache.version explicitly in the dependencies section. Please take a look and let me know if you have any further issues.

SudharakaP:
> Maybe we also should add tests in server spec file.
> 
> https://github.com/jhipster/generator-jhipster/blob/master/test/server.spec.js

I've added a test to generate the server with Caffeine as well. Since there's no new files generated for Caffeine it's basically identical to the one we have for the default ehcache test. Please let me know if you see any further issues :smile:

avdev4j:
Great ! Even if the test is the same, I think it's better to handle both case with minor dev cost :D.

Next step I will try it on my machine (I hope in next days). I think this PR is close to be merged, or maybe you have other things to do?

What ever this is a great job.

SudharakaP:
@avdev4j : I think all is done related to this PR and Caffeine. Over time we might need to improve Caffeine cache support even further but I think for now everything is good to go. :smile:

avdev4j:
Congratulations @SudharakaP don't forget the bug bounty ! ;)

pmverma:
Congratulations @SudharakaP <https://github.com/SudharakaP>.

SudharakaP:
@pmverma @avdev4j : Oh, thanks guys; love contributing to Jhipster. :smile:

SudharakaP:
And thanks for reminding about the bounty. I've claimed it; https://opencollective.com/generator-jhipster/expenses/9997

jdubois:
@SudharakaP awesome!! But why did you do a $90 bug bounty? It's $100, you should take all the money

pascalgrimaud:
@SudharakaP : why 90$ ? it's 100$ !

SudharakaP:
@jdubois @pascalgrimaud : Because I wanted to donate 10 dollars back. :smile:

jdubois:
Oh thanks, you really didn't need to 😍

PR #10316, opened 2019-08-30
remove duplicate test in UserServiceIT tests file

nonomoho:

### DIFF
--- a/generators/server/templates/src/test/java/package/service/UserServiceIT.java.ejs
+++ b/generators/server/templates/src/test/java/package/service/UserServiceIT.java.ejs
@@ -512,19 +512,6 @@ public class UserServiceIT <% if (databaseType === 'cassandra') { %>extends Abst
         assertThat(userDTO.getLangKey()).isEqualTo("en");
     }
 
-    @Test
-    <%_ if (databaseType === 'sql') { _%>
-    @Transactional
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/test/java/package/service/UserServiceIT.java.ejs
```diff
@@ -512,19 +512,6 @@ public class UserServiceIT <% if (databaseType === 'cassandra') { %>extends Abst
         assertThat(userDTO.getLangKey()).isEqualTo("en");
     }
 
-    @Test
-    <%_ if (databaseType === 'sql') { _%>
-    @Transactional
-    <%_ } _%>
-    public void testUserDetailsWithUSLocale2() {
-        userDetails.put("locale", "en-US");
```
vishal423: I think the other tests is asserting with locale as `en_US` and this one with `en-US`

nonomoho: Oh yes good catch ... My bad !
I close this PR
Do you think we should rename tests name to be more explicit?

vishal423: yes, that can be done to provide explicit context

PR #10323, opened 2019-08-30
Merge shared common module with shared module in Angular

kaidohallik:
Currently there are 3 shared modules in hierarchy under `shared` folder. It makes it too complex. This PR simplifies shared modules hierarchy, removes one level. If some component, directive or pipe seems common enough then this can be moved to `ng-jhipster` library.


kaidohallik:
@deepu105 Do you think this simplification (reducing shared modules hierarchy to 2 levels) is good or is it better to proceed with 3 level hierarchy of shared modules for some reason?

deepu105:
@kaidohallik originally this separation was due to some issues like cyclic dependencies and tree shaking, but if this doesn't cause any issues anymore, I think its fine

kaidohallik:
@deepu105 thanks for the feedback. While testing this PR I didn't encounter any problems and CI passed. So I believe that this PR is ready for merge.

vishal423:
thanks @kaidohallik

PR #10324, opened 2019-08-31
Use embeddableLaunchScript=true for apps generated by older JHipster versions

PierreBesson:
Linked to #10282 
Set embeddableLaunchScript to true for apps generated with old JHipster versions
We don't want to surprise users that were relying on this feature to be enable by default.

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -307,6 +307,12 @@ module.exports = class extends BaseGenerator {
                         this.clientPackageManager = 'yarn';
                     }
                 }
+                // Set embeddableLaunchScript to true if not defined, for backward compatibility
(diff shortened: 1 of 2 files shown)


generators/app/index.js
```diff
@@ -307,6 +307,12 @@ module.exports = class extends BaseGenerator {
                         this.clientPackageManager = 'yarn';
                     }
                 }
+                // Set embeddableLaunchScript to true if not defined, for backward compatibility
+                // See https://github.com/jhipster/generator-jhipster/issues/10255
```
saturnism: double checking - newly generated app will have this set to false by default right?

PierreBesson: Yes, false for new apps, true for old apps.

PR #10336, opened 2019-09-02
Enhancements to GAE generator

### LINKED ISSUE #10331: GAE Generator Enhancements and Corner Cases

This is a continuation of https://github.com/jhipster/generator-jhipster/issues/10196 which hopes to discuss and do some refactoring to the GAE App generator. It's my belief that this generator has some remaining issues outlined below. 

1) If the user hasn't installed Google Cloud SDK our app generator prompts the user to do so but it also give the unnecessary option of overwriting the pom file. This should be avoided. Typical output looks like;

> INFO! Using JHipster version installed globally
INFO! Executing jhipster:gae
INFO! Options: from-cli: true
Welcome to Google App Engine Generator (Beta)
WARNING! This sub-generator is still in development, please report bugs on Github
✖ You don't have the Cloud SDK (gcloud) installed. 
Download it from https://cloud.google.com/sdk/install
/bin/sh: 1: gcloud: not found
✖ Unable to determine the default Google Cloud Project ID
 conflict pom.xml
? Overwrite pom.xml? (ynaxdH) 

2) I would suggest we remove part of the generator where it tries to install the App Engine Java SDK using `gcloud components install app-engine-java --quiet`. This has the potential to fail since the Cloud SDK installation can be done in [several ways](https://cloud.google.com/sdk/install#installation_options) and it is operating system dependant. For example a typical case maybe in an Ubuntu or Debian where the user installs it using `apt-get`. Then he could encounter the following error; 

https://askubuntu.com/questions/1146879/how-do-i-fix-the-google-cloud-sdk-component-manager-is-disabled-for-this-instal

I think the best way to handle this is to check whether `app-engine-java` component is installed and if not prompt the user to do so; same thing we do for Google Cloud SDK. 

If this is in agreement let me know. I'll create the PR. 

More user friendly GAE generator with less confusion. 

#10196 and #10285

SudharakaP:
Fixes #10331


generators/gae/index.js
```diff
@@ -66,12 +65,12 @@ module.exports = class extends BaseGenerator {
...
-                                stdio: [process.stdin, process.stdout, process.stderr]
-                            });
-                            child.on('exit', code => {
-                                if (code !== 0) {
-                                    this.abort = true;
+                            exec(`gcloud components install ${component} --quiet`, err => {
+                                if (err) {
+                                    this.log.error(err);
+                                    done(
+                                        'App Engine Java SDK installation failed. \nPlease try to install manually; https://cloud.google.com/sdk/install'
```
saturnism: I think the error message here should prompt the user to run `gcloud components install` manually.

The `gcloud` sdk install was already checked previously.

generators/gae/index.js
```diff
@@ -736,14 +738,15 @@ module.exports = class extends BaseGenerator {
...
+                if (this.abort) return;
                 if (this.buildTool === 'maven') {
                     if (this.gaeCloudSQLInstanceNeeded === 'Y') {
                         this.render('pom-plugin.xml.ejs', rendered => {
                             this.addMavenPlugin('com.google.cloud.tools', 'appengine-maven-plugin', '2.1.0', rendered.trim());
                         });
                     }
                     this.render('pom-profile.xml.ejs', rendered => {
-                        this.addMavenProfile('prod-gae', `            ${rendered.trim()}`);
+                        this.addMavenProfile('prod-gae', `${rendered.trim()}`);
```
saturnism: i think the spaces were added originally to make sure the plugin is properly aligned/indented in the pom.xml. is that still the case if we remove the spacing here?

generators/gae/index.js
SudharakaP: Thanks. I've changed it. Let me know if you need any further changes.

generators/gae/index.js
SudharakaP: Yes, you are correct. I thought it was just an unnecessary space. 😄 I've added it back.

saturnism:
tested it the deployment myself, lgtm . /cc @PierreBesson if there are any other issues. otherwise, ready to merge 😄

Issue #10220, opened 2019-08-14
Polish file and class names in Angular part


kaidohallik:
##### **Overview of the issue**

Issues I discovered when refactoring imports in Angular part (barrel remove in #10185):

* inconsistent component and service class prefixes: components and services are mainly without prefix, but some of them have prefix `<%=jhiPrefixCapitalized%>`: `AlertErrorComponent`, `AlertComponent`, `ConfigurationComponent`, `ConfigurationService`, `DocsComponent`, `GatewayComponent`, `HealthModalComponent`, `HealthCheckComponent`, `HealthService`, `LoginModalComponent`, `MainComponent`, `MetricsMonitoringComponent`, `MetricsService`, `TrackerComponent`, `TrackerService`  
Why such difference? I think it would be good practice to use one style, so 2 possible solutions:
    * either remove prefix `<%=jhiPrefixCapitalized%>` from all components and services
    * or add prefix `<%=jhiPrefixCapitalized%>` to all components and services

* modules are prefixed differently, with prefix `<%=angularXAppName%>`, why this prefix is different, maybe use the same prefix `<%=jhiPrefixCapitalized%>` as used in components and services? If prefix will be removed from all components and services then this inconsistency will disappear, then only prefix `<%=angularXAppName%>` remains in use and this is not the case any more.

* as standard, service files have suffix `.service.ts`, but one file is different: `user-route-access-service.ts` (https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/core/auth/user-route-access-service.ts.ejs - dash instead of point before service), suggestion: rename this file to `user-route-access.service.ts`

* some class or file names doesn't follow general name convention:
    * register service class name is `Register` (https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/account/register/register.service.ts.ejs#L26), suggestion: rename this class to `RegisterService` (other service classes are ending with suffix Service)
    * in `health.component.ts` (https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/admin/health/health.component.ts.ejs#L29) class name is `HealthCheckComponent`, suggestion: rename this to `HealthComponent`
[...]

PR #10340, opened 2019-09-03
Improve shared libs, main and core modules in Angular

kaidohallik:
If currently using `CookieService` from `ngx-cookie` in some lazy loaded module which imports `SharedModule` (developer modifies code and this modification uses `CookieService`) then new instance of the `CookieService` is created. But the idea of the `ngx-cookie` library is that `CookieService` should be singleton. This PR makes `CookieService` to singleton.  
This also removes entirely `imports` part from shared libs module because shared libs module is not declaring any component, directive or pipe and therefore no imports needed.


generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
```diff
@@ -21,6 +21,7 @@ import { NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { NgbDatepickerConfig } from '@ng-bootstrap/ng-bootstrap';
 import { NgJhipsterModule } from 'ng-jhipster';
+import { CookieModule } from 'ngx-cookie';
```
vishal423: can you please add this to core module ?

kaidohallik: Good idea, it keeps main app module cleaner.  
But to keep similar style I think it's good to move also `NgxWebstorageModule.forRoot` and `NgJhipsterModule.forRoot` to core module. What do you think about that?

vishal423: yes, lets do one change at a time to keep change-set clean and easy to review :smile:

kaidohallik: When moving `CookieModule.forRoot`, `NgxWebstorageModule.forRoot` and `NgJhipsterModule.forRoot` to `core.module` I found that there is also some other configuration in the main module. I made full cleaning and moved all configuration from the main module to the core module. Now `core.module` is the single point for the configuration.  
I also changed PR title to better reflect changes done.

vishal423: @kaidohallik, We shouldn't introduce breaking changes unless they are required for the correct functioning of the generated application. Cleanup of the app module can be done before the next major release. Can you please revert changes and do only those that are relevant to the issue discussed above? Thanks

kaidohallik: @vishal423, Why do you think that this main module cleaning is breaking change? What is the definition of the breaking change for the `generator-jhipster`?

vishal423: @kaidohallik, I haven't looked in detail on changes made here, however, any change that has the potential to break an existing jhipster module and customization of an application post generation can be considered as a breaking change.

kaidohallik: @vishal423, I think that every single tiny change in `generator-jhipster` templates has potential to break an existing `JHipster module` and customization of an application post generation.  
I agree that changing file location increases this chance.  
You requested cleaning a little bit in a main module, I saw that more cleaning is needed and now it's hard to accept that only half of the job is allowed to do.  
Can you please confirm again that I move only these 3 `forRoot` constructs to core module and will revert all other cleaning?

kaidohallik: Moving `NgJhipsterModule.forRoot` from `app.module.ts` to `core/core.module.ts` caused CI to fail, because CI scripts changed `alertTimeout` in app config and this action failed now. I committed fix just in case. This is of course a good example to illustrate that changes in generator can break modules and old application customization in post generation.

vishal423: I had a quick look and seems that most changes mayn't have an impact on the customized code, however, it needs to be asserted on an application. Also, can you please check if we have any customization on the files to be removed as part of this exercise, then, those changes do end up in the git diff after application upgrade?

kaidohallik: I added functionality for showing warnings if in file renaming old and new content differ. Corresponding image from console output:

warning

Previously client side cleaning was runned from `app` generator context, now I moved this to `client` sub-generator context. It was needed to get right context for template rendering.  
I also rebased with current master and resolved conflicts and force pushed.  

If there are changes in renamed files then before commit these files are showed separately, new files as added and old files as deleted. But after commit Git was able to link files and shows added and removed lines count. Corresponding image from Git history:

git_history

vishal423: Currently, our approach is to remove a file and later add that file at a new location. Can you please check if we can leverage `move` command to do this job? It might help to preserve diff before commit.

On a side note, this discussion is going too far. In this commit, can we limit to external libraries and further cleanup can be done in separate PR as that requires more effort/discussion?

kaidohallik: Can I do new PR which limits to external libraries and after merging new PR we can close this and maybe use when new major release is prepared?

kaidohallik: Or is it better to add new commit here, so that possible solution to full cleaning remains in `upstream` Git history?

vishal423: I think let's have separate PR for external libraries, move app module refactoring as an item in issue #10220 and close this. Further, in that ticket, it would be good to discuss the solution before you proceed with actual implementation.

kaidohallik: I opened new simplified PR #10360

wmarques:
@kaidohallik if I'm correct, this PR will be merged before major release because it has breaking changes, right ?

kaidohallik:
@wmarques half of these changes are in the new PR #10360, half are linked to the issue #10220 for the major release with this comment: https://github.com/jhipster/generator-jhipster/issues/10220#issuecomment-528311220

PR #10138, opened 2019-07-23
Fix Istio URL issues 

deepu105:
Fix #10135 

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10342, opened 2019-09-03
Update /services prefix for the Kubernetes generator

saturnism:
For #10343 - issue based on https://medium.com/@simionrazvan/microservices-using-jhipster-kubernetes-istio-bf75e992612a

### DIFF
--- a/generators/kubernetes/templates/deployment.yml.ejs
+++ b/generators/kubernetes/templates/deployment.yml.ejs
@@ -194,13 +194,13 @@ spec:
           containerPort: <%= app.serverPort %>
         readinessProbe:
           httpGet:
-            path: <%= app.applicationType === 'microservice' && !app.serviceDiscoveryType ? "/" + app.baseName.toLowerCase() : "" %>/management/health
+            path: <%= app.applicationType === 'microservice' && !app.serviceDiscoveryType ? "/services/" + app.baseName.toLowerCase() : "" %>/management/health
             port: http
           initialDelaySeconds: 20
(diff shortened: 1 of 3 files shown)


PierreBesson:
What this is needed is completely beyond me. I don't know who added this prefix but it doesn't make any sense.

ruddell:
The discussion about the `/services/` prefix [starts here](https://github.com/jhipster/generator-jhipster/pull/9098#issuecomment-455891616), the purpose is to prevent client/server path conflicts.

>The Zuul routes take precedence, so you are right that if you name a microservice and entity the same thing, then generate it in the gateway, there will be issues.  For example, I have a `store` microservice and generate an `store` entity, the paths will match causing issues (`/store/**` is mapped to the microservice, so refreshing in the client on the `/store` entity listing returns a 403).

This PR does the same as https://github.com/jhipster/generator-jhipster/pull/10138

saturnism:
agreed it's kind of weird and also caught me by surprise when i use jhipster recently. /api/[service]/[entity] would've made more sense for me, but i think that was an option discussed that would've been much more complicated.

saturnism:
@PierreBesson if ok w/ you, i'd like to merge this, and then fix a couple of other issues :)

PierreBesson:
Yes of course it's the right solution. Thanks @ruddell for giving clarification. I'm still not convinced but we will have to tackle this issue later.

saturnism:
> ... I'm still not convinced but we will have to tackle this issue later.

+1 😄 thanks!

PR #10375, opened 2019-09-09
Fix terser version to 4.2.1 in Angular

### LINKED ISSUE #10371: Protractor: e2e tests with Angular are broken

Our current e2e tests with Angular are broken
The tests failed during create/save entities

See https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build

cc @vishal423 

All new PRs will fail, so it needed to be fixed as soon as possible
That's why I'm putting a bounty on this

kaidohallik:

fix #10371

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -132,6 +132,7 @@
     "sass-loader": "7.3.1",
     "simple-progress-webpack-plugin": "1.1.2",
     "style-loader": "1.0.0",
+    "terser": "4.2.1",
     "terser-webpack-plugin": "1.4.1",
     "thread-loader": "2.1.3",
     "to-string-loader": "1.1.5",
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/package.json.ejs
```diff
@@ -157,6 +158,11 @@
     "workbox-webpack-plugin": "4.3.1",
     "write-file-webpack-plugin": "4.5.1"
   },
+<%_ if (clientPackageManager === 'yarn') { _%>
```
vishal423: Do you require this as you have included that also as a dependency? Either case, I think we should also add a comment to the corresponding issue that we can track to update/remove this block later on.

pascalgrimaud: Yes, it is needed for Yarn, for a reason I don't remember
No, we can't add comment in JSON file :D

vishal423: we can always add a comment in generator that's used to include this code :smile:

pascalgrimaud: Oh indeed, nice tips !

kaidohallik:
Terser version 4.3.1 is released with fix for that: https://github.com/terser/terser/releases/tag/v4.3.1  
Currently testing, if successful then closing this PR, as transitive dependency then fixes problem.

kaidohallik:
In my testing `terser` version `4.3.1` fixes the problem, so this PR is not needed any more, that's why I'm closing this.

generators/client/templates/angular/package.json.ejs
kaidohallik: @vishal423, without this `resolve` part Yarn didn't resolve Terser to version 4.2.1, so different config for Npm and Yarn was needed to work correctly.

PR #10397, opened 2019-09-13
Fix faker and implement liquibase reproducibility

### LINKED ISSUE #10372: Reproducible build

When implementing a blueprint, we need to recreate a project a lot of times.
Every time it generates different sources.

There is 2 main source of problems:
- liquibase changelogs.
- faker data (dates, enums and RegExp).

- Easier to test generators and blueprints.
- Create automatized regeneration tests.

##### **Reproduce the error**

- Generate a project with default options.
- Regenerate it with `jhipster --with-entities`

Every time the project is regenerated, fake dates will be updated with recent values and fake enums will be randomized again.

- New projects will always create entities with changelog with recent date.
Automatized tests are not possible and comparing project creation will always differ.

mshima:
Fix faker reproducibility of dates and enums.
And implement reproducibility for liquibase files
Fixes #10372

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -183,6 +183,12 @@ module.exports = class extends BaseGenerator {
             defaults: false
         });
 
+        // This adds support for a `--creation-timestamp` flag which can be used create reproducible builds
+        this.option('creation-timestamp', {
+            desc: 'Project creation timestamp (used for reproducible builds)',
+            type: String
+        });
+
         this.skipClient = this.configOptions.skipClient = this.options['skip-client'] || this.config.get('skipClient');
         this.skipServer = this.configOptions.skipServer = this.options['skip-server'] || this.config.get('skipServer');
         this.skipUserManagement = this.configOptions.skipUserManagement =
@@ -436,8 +442,11 @@ module.exports = class extends BaseGenerator {
             },
 
             saveConfig() {
+                const creationTimestamp = this.parseCreationTimestamp() || this.config.get('creationTimestamp') || new Date().getTime();
+
                 const config = {
                     jhipsterVersion: packagejs.version,
+                    creationTimestamp,
                     applicationType: this.applicationType,
                     baseName: this.baseName,
                     testFrameworks: this.testFrameworks,
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -34,6 +34,38 @@ const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
 // In order to have consistent results with Faker, the seed is fixed.
 faker.seed(42);
 
+/*
+ * Current faker version is 4.1.0 and was release in 2017
+ * It is outdated
+ * https://github.com/Marak/faker.js/blob/10bfb9f467b0ac2b8912ffc15690b50ef3244f09/lib/date.js#L73-L96
+ * Needed for reproducible builds
+ */
+const getRecentDate = function(days, refDate) {
+    let date = new Date();
+    if (refDate !== undefined) {
+        date = new Date(Date.parse(refDate));
+    }
+
+    const range = {
+        min: 1000,
+        max: (days || 1) * 24 * 3600 * 1000
+    };
+
+    let future = date.getTime();
+    future -= faker.random.number(range); // some time from now to N days ago, in milliseconds
+    date.setTime(future);
+
+    return date;
+};
+
+const getRecentForLiquibase = function(days, changelogDate) {
+    let formatedDate;
+    if (changelogDate !== undefined) {
+        formatedDate = `${changelogDate.substring(0, 4)}-${changelogDate.substring(4, 6)}-${changelogDate.substring(6, 8)}`;
+    }
+    return getRecentDate(1, formatedDate);
+};
+
 /**
  * The default is to use a file path string. It implies use of the template method.
  * For any other config an object { file:.., method:.., template:.. } can be used
@@ -87,6 +119,7 @@ const serverFiles = {
                     options: {
                         interpolate: INTERPOLATE_REGEX,
                         context: {
+                            getRecentForLiquibase,
                             faker,
                             randexp
                         }
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -808,7 +808,21 @@ module.exports = class extends PrivateBase {
      * Generate a date to be used by Liquibase changelogs.
      */
     dateFormatForLiquibase() {
-        const now = new Date();
+        let now = new Date();
+        // Run reproducible timestamp when regenerating the project with with-entities option.
+        if (this.options.withEntities) {
+            if (this.configOptions.lastLiquibaseTimestamp) {
+                // Counter already started.
+                now = this.configOptions.lastLiquibaseTimestamp;
+            } else {
+                // Create a new counter
+                const creationTimestamp = this.parseCreationTimestamp() || this.config.get('creationTimestamp');
+                now = creationTimestamp ? new Date(creationTimestamp) : now;
+            }
+            now.setMinutes(now.getMinutes() + 1);
+            this.configOptions.lastLiquibaseTimestamp = now;
+        }
+
         const nowUTC = new Date(
             now.getUTCFullYear(),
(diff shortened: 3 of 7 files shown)


DanielFran:
@mshima Can you please fix conflicts?

mshima:
> @mshima Can you please fix conflicts?
Done.

generators/entity-server/files.js
```diff
@@ -34,6 +34,40 @@ const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
...
 faker.seed(42);
 
+/*
+ * Current faker version is 4.1.0 and was release in 2017
+ * It is outdated
+ * https://github.com/Marak/faker.js/blob/10bfb9f467b0ac2b8912ffc15690b50ef3244f09/lib/date.js#L73-L96
+ * Needed for reproducible builds
+ */
+/* eslint-disable-next-line */
+const recentDate = function (days, refDate) {
```
MathieuAA: if this is a function, please a "verbal" form (getRecentDate for instance)
What does this function do and why? I don't get it just by looking at the name

generators/entity-server/files.js
```diff
@@ -34,6 +34,40 @@ const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
...
+    };
+
+    var future = date.getTime();
+    future -= faker.random.number(range); // some time from now to N days ago, in milliseconds
+    date.setTime(future);
+
+    return date;
+};
+
+const recentForLiquibase = function (days, changelogDate) {
```
MathieuAA: same question about the name

generators/generator-base.js
```diff
@@ -808,7 +808,28 @@ module.exports = class extends PrivateBase {
      * Generate a date to be used by Liquibase changelogs.
      */
     dateFormatForLiquibase() {
-        const now = new Date();
+        let now = new Date();
+        // Run reproducible timestamp when regenerating the project with with-entities option.
+        if (this.options && this.configOptions && this.options.withEntities) {
```
MathieuAA: that's one lengthy if there, it would be better to refactor it

for instance, l. 816: what do we need to initialize the now variable? I don't think we need options, configOptions & withEntities, wouldn't configOptions be enough?

generators/entity-server/files.js
mshima: This is an exact update faker recent function with refDate parameter.
Faker lib seem unsupported right now, with no recent release.
Wanted to keep it the same.
https://github.com/Marak/faker.js/blob/9fd8d7d37b398842d0784a116a340f7aa6afb89b/lib/date.js#L73-L96

PR
https://github.com/Marak/faker.js/commit/816b658e26badba6a911b7065aa8ca413d848b9d#diff-9e4794cb4c334b7eaa3dfdff8725406e

generators/generator-base.js
mshima: Reproducibly build only makes sense with `jhipster --with-entities` command and it is not set to configOptions.

this.options and this.configOptions test probably com be removed.

generators/entity-server/files.js
MathieuAA: Ah okay, thanks for the explanation.
My point still applies though: if we introduce some code in our system, it has to abide to the same level of quality (or above) as the rest of the code.

generators/entity-server/files.js
```diff
@@ -87,6 +119,8 @@ const serverFiles = {
                     options: {
                         interpolate: INTERPOLATE_REGEX,
                         context: {
+                            getRecentDate,
```
MathieuAA: Are both functions are used somewhere else?

generators/generator-base.js
```diff
@@ -808,7 +808,28 @@ module.exports = class extends PrivateBase {
...
+        let now = new Date();
+        // Run reproducible timestamp when regenerating the project with with-entities option.
+        if (this.options.withEntities) {
+            if (this.configOptions.lastLiquibaseTimestamp) {
+                // Counter already started.
+                now = this.configOptions.lastLiquibaseTimestamp;
+            } else {
+                // Create a new counter
+                let creationTimestamp;
+                if (this.options.creationTimestamp) {
```
MathieuAA: I've seen this code somewhere...

generators/generator-base.js
```diff
@@ -808,7 +808,28 @@ module.exports = class extends PrivateBase {
...
+                    creationTimestamp = Date.parse(this.options.creationTimestamp);
```
MathieuAA: I wonder if [this](https://stackoverflow.com/questions/2587345/why-does-date-parse-give-incorrect-results) applies, what do you think?

generators/app/index.js
```diff
@@ -436,8 +442,20 @@ module.exports = class extends BaseGenerator {
             },
 
             saveConfig() {
+                let creationTimestamp;
+                if (this.options.creationTimestamp) {
+                    creationTimestamp = Date.parse(this.options.creationTimestamp);
```
MathieuAA: See my point below about Date.parse

generators/entity-server/files.js
MathieuAA: I only see the second being used elsewhere, so no need to export it

generators/generator-base.js
```diff
@@ -808,7 +808,28 @@ module.exports = class extends PrivateBase {
      * Generate a date to be used by Liquibase changelogs.
      */
     dateFormatForLiquibase() {
-        const now = new Date();
+        let now = new Date();
+        // Run reproducible timestamp when regenerating the project with with-entities option.
+        if (this.options.withEntities) {
```
MathieuAA: this is so complex that I think this deserves to be in its own function, your opinion?
It could be called getValueLastLiquibaseTimestamp

generators/generator-base.js
mshima: Seems that the following fixes this problem.
`const nowUTC = new Date(...`

generators/generator-base.js
mshima: Create a function parseTimestamp on generator-base-private.js?

mshima: or utils

MathieuAA: yeah maybe that'd be a good idea to put in the private thing

generators/generator-base.js
MathieuAA: yup

generators/generator-base.js
mshima: I think it became simple enough now.

mshima:
@MathieuAA what do you think about removing changelogDate generation on jhipster-core?
Than the --creation-timestamp will work for import-jdl too.
https://github.com/jhipster/jhipster-core/compare/master...mshima:master

MathieuAA:
For that to work, you'd need to update the jdl-import function to insert into the JSON entity files the changelogDate right?

mshima:
> For that to work, you'd need to update the jdl-import function to insert into the JSON entity files the changelogDate right?

No, if the entity doesn't find changelogDate, it will create.
Is jhipster-core used as dependency in any other package that can conflict with this?

MathieuAA:
Yeah maybe, that's the point of it being a lib. You know what, let's not do anything to JCore yet.
It's going to be merged later in the generator, so the need to refactor things will come later.

mshima:
@MathieuAA 
I will need to change https://github.com/jhipster/generator-jhipster/blob/de25969283ebea50039066763a8cae4805849edf/generators/generator-base.js#L813

to `if (this.options.withEntities || this.options.creationTimestamp) {`

Should I do this now?

MathieuAA:
Yeah, might be a good idea

mshima:
When this PR is merged I will create a PR on JCore to discuss this change there.

mshima:
@MathieuAA thank you for your help.

MathieuAA:
No, thank you :) your work is awesome

pascalgrimaud:
@mshima : thanks for this PR. Can you add some documentation too, before the next release plz ?

mshima:
@pascalgrimaud sure, where do you mean? is there another repos for docs?

PierreBesson:
@mshima yes https://github.com/jhipster/jhipster.github.io

PR #10414, opened 2019-09-14
enable use of enumerations with Cassandra

### LINKED ISSUE #10403: Enums are not correctly handled when using Cassandra

When using enums in JDL with cassandra, the fields of entities whose type is an enum are generated, but the enums themselves are not generated. The generated CQL doesn't include these fields either.

The generated project does not compile and doesn't work.
We should support enums in Cassandra since spring-data-cassandra [supports them](https://github.com/spring-projects/spring-data-cassandra/blob/master/src/main/asciidoc/reference/mapping.adoc#data-mapping-and-type-conversion) since https://github.com/spring-projects/spring-data-cassandra/pull/47.

##### **Reproduce the error**

```
application {
  config {
    baseName myapp
    applicationType microservice
    databaseType cassandra
    prodDatabaseType cassandra
    buildTool maven
  }
  entities *
}

enum MyEnum { A, B }

entity MyEntity {
    name String
    value MyEnum
}
```

Then `mvn verify`

##### **JHipster Version(s)**

6.2.0

See above JDL example

See above JDL example

clement26695:
PR candidate to fix #10403 

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -722,7 +722,7 @@ class EntityGenerator extends BaseBlueprintGenerator {
                         context.fieldsIsReactAvField = true;
                     }
 
-                    const nonEnumType = [
+                    field.fieldIsEnum = ![
                         'String',
                         'Integer',
                         'Long',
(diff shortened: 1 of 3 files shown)


generators/entity/index.js
```diff
@@ -738,7 +738,7 @@ class EntityGenerator extends BaseBlueprintGenerator {
                         'byte[]',
                         'ByteBuffer'
                     ].includes(fieldType);
-                    if (['sql', 'mongodb', 'couchbase', 'no'].includes(context.databaseType) && !nonEnumType) {
+                    if (['sql', 'mongodb', 'couchbase', 'no', 'cassandra'].includes(context.databaseType) && !nonEnumType) {
```
murdos: Since all databases types are not included, the `if` is not necessary anymore: 
```
field.fieldIsEnum = [
                        'String',
                        'Integer',
                        'Long',
                        'Float',
                        'Double',
                        'BigDecimal',
                        'LocalDate',
                        'Instant',
                        'ZonedDateTime',
                        'Duration',
                        'UUID',
                        'Boolean',
                        'byte[]',
                        'ByteBuffer'
                    ].includes(fieldType);
```

murdos:
Thanks @clement26695

PR #10436, opened 2019-09-17
Remove duplicated import in CacheConfiguration.java

avdev4j:
From:
```
<%_ if (enableHibernateCache) { _%>
import org.hibernate.cache.jcache.ConfigSettings;
<%_ } _%>
import io.github.jhipster.config.JHipsterProperties;

<%_ } _%>
<%_ if (cacheProvider === 'caffeine') { _%>
import com.github.benmanes.caffeine.jcache.configuration.CaffeineConfiguration;
import java.util.OptionalLong;
import java.util.concurrent.TimeUnit;

<%_ if (enableHibernateCache) { _%>
import org.hibernate.cache.jcache.ConfigSettings;
<%_ } _%>
import io.github.jhipster.config.JHipsterProperties;

<%_ } _%>
```
to:

```
<%_ if (cacheProvider === 'caffeine') { _%>
import com.github.benmanes.caffeine.jcache.configuration.CaffeineConfiguration;
import java.util.OptionalLong;
import java.util.concurrent.TimeUnit;

<%_ if (enableHibernateCache) { _%>
import org.hibernate.cache.jcache.ConfigSettings;
<%_ } _%>
import io.github.jhipster.config.JHipsterProperties;

<%_ } _%>
```

### DIFF
--- a/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
@@ -24,12 +24,6 @@ import java.time.Duration;
 import org.ehcache.config.builders.*;
 import org.ehcache.jsr107.Eh107Configuration;
 
-    <%_ if (enableHibernateCache) { _%>
-import org.hibernate.cache.jcache.ConfigSettings;
-    <%_ } _%>
-import io.github.jhipster.config.JHipsterProperties;
-
-<%_ } _%>
 <%_ if (cacheProvider === 'caffeine') { _%>
 import com.github.benmanes.caffeine.jcache.configuration.CaffeineConfiguration;
 import java.util.OptionalLong;


generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
```diff
@@ -24,12 +24,6 @@ import java.time.Duration;
 import org.ehcache.config.builders.*;
 import org.ehcache.jsr107.Eh107Configuration;
 
-    <%_ if (enableHibernateCache) { _%>
```
murdos: From what I understand this block is part of the conditional block `if (cacheProvider === 'ehcache')`, and the other is part of the conditional block `if (cacheProvider === 'caffeine')`, so the imports wont be duplicated and if you use ehcache as cache provider, these imports will be missing.

avdev4j: oh my bad, you're right.
So nothing to do here :D, I close

PR #10440, opened 2019-09-17
Add Information about Pull app

SudharakaP:
I thought it would be nice for users to know that there's an alternative method of reducing rebases and avoiding the hassle of forgetting to rebase from upstream.

Reference: https://github.com/wei/pull#readme

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -178,6 +178,8 @@ Sometimes your PR will have merge conflicts with the upstream repository's maste
     git push -f
     ```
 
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
I'd prefer to keep things simple, for our contributors: it's easier to add something like than:
git remote add upstream xxxxx
git fetch upstream
git rebase upstream/master

It's 3 commands instead of learning an external tools, which will hide how git works.

It's just my opinion. What do you think ?

atomfrede:
I like having just the plain git commands. They can be used everywhere else. So +1 to keep it simple.

SudharakaP:
@pascalgrimaud @atomfrede : I agree that it makes it a bit complicated in terms of the setup, and I am okay either-way; just thought that it made my life easier as I tend to forget pulling from upstream every so often; so I'll close this for now. Thanks for your input. 👍

PR #10443, opened 2019-09-18
Admin menu improvements

mshima:
Rename UserMgmt* to UserManagement*.
Migrate admin/* to sub modules and lazy load.


pascalgrimaud:
Next version will probably be 6.4.0, as it will contain Kafka native client and Redis

kaidohallik:
@mshima additional note about my `admin.module.ts` removal comment: if removing this intermediate level then need also change the following needle descriptions in the needle API:
* https://github.com/jhipster/generator-jhipster/blob/master/generators/client/needle-api/needle-client-angular.js#L57
* https://github.com/jhipster/generator-jhipster/blob/master/test/needle-api/needle-client-angular.spec.js#L177-L183

kaidohallik:
If we want to use one admin module and avoid module renaming from `admin.module.ts` to `admin-routing.module.ts` as this increases chance that this change breaks someones module or post generation customization (see conversation starting from https://github.com/jhipster/generator-jhipster/pull/10340#discussion_r320604037) then maybe use this approach:
* don't add `admin-routing.module.ts` in this PR, instead put this code into existing `admin.module.ts`
* add comment to #10220 for a new major version, that rename `admin.module.ts` to `admin-routing.module.ts`

mshima:
> If we want to use one admin module and avoid module renaming from `admin.module.ts` to `admin-routing.module.ts` as this increases chance that this change breaks someones module or post generation customization (see conversation starting from [#10340 (comment)](https://github.com/jhipster/generator-jhipster/pull/10340#discussion_r320604037)) then maybe use this approach:
> 
> * don't add `admin-routing.module.ts` in this PR, instead put this code into existing `admin.module.ts`
> * add comment to #10220 for a new major version, that rename `admin.module.ts` to `admin-routing.module.ts`

I want to create a new needle: `jhipster-needle-add-admin-route`.
I don't like the idea of creating this needle on an already deprecated file.

We should keep both or remove `admin.module`.

mshima:
> currently all generated modules are prefixed with `<%=angularXAppName%>`, I prefer that we keep unified style, maybe:
> 
> * remove this prefix from every module?
> * or add this prefix also to newly added modules?

In my first patch I had them removed, but reverted.
I prefer them without a prefix.
Or with a fixed prefix for everything provided by jhipster generator.

Probably is better to discuss it on another issue.

generators/cleanup.js
```diff
@@ -207,4 +207,7 @@ function cleanupOldServerFiles(generator, javaDir, testDir, mainResourceDir, tes
         generator.config.delete('blueprint');
         generator.config.delete('blueprintVersion');
     }
+    if (generator.isJhipsterVersionLessThan('6.4.0')) {
```
kaidohallik: As this is not server file then this should go to function `cleanupOldFiles` instead of `cleanupOldServerFiles`.

kaidohallik:
> I want to create a new needle: `jhipster-needle-add-admin-route`.
> I don't like the idea of creating this needle on an already deprecated file.
> 
> We should keep both or remove `admin.module`.

I vote for keeping 1 file.



> In my first patch I had them removed, but reverted.
> I prefer them without a prefix.
> Or with a fixed prefix for everything provided by jhipster generator.
> 
> Probably is better to discuss it on another issue.

#10220 is one possible place for discussion. For me it's ok to leave these new modules without prefix and then with follow up or as part of resolving #10220 remove prefix from all modules.

generators/cleanup.js
mshima: Sorry about that.
Didn't noticed.

generators/cleanup.js
```diff
@@ -88,6 +88,10 @@ function cleanupOldFiles(generator) {
     if (generator.isJhipsterVersionLessThan('6.3.0') && generator.configOptions && generator.configOptions.clientFramework === 'react') {
         generator.removeFile('tslint.json');
     }
+
+    if (generator.isJhipsterVersionLessThan('6.4.0')) {
```
kaidohallik: Latest cleanups are with additional condition `&& generator.configOptions && generator.configOptions.clientFramework === 'angularX'` to avoid deleting files if `clientFramework` is not `angularX` and file with the same name exists. Can you add this condition also here?

mshima: Done, thanks.

kaidohallik:
@mshima, I have 1 final note: `CanActivate` parameters and `UserRouteAccessService` import need to be removed from `user-management.route.ts`, because this is already applied to whole `admin` route in `app-routing.module.ts` and no need to repeat (other admin routes are not repeating this, there is this ok)

My test result:
* on manual upgrade old files are deleted correctly
* chunk loading is optimal both in `dev` and `prod` profile
* admin routes are accessible only for `ROLE_ADMIN` (this holds also after removing `CanActivate` from `user-management.route.ts`)
* no errors found in manual testing
* code looks clean to me, only `CanActivate` parameters and `UserRouteAccessService` import need to be removed from `user-management.route.ts`

@vishal423 would it be possible to take a look to this PR and if all is ok for you then merge?

vishal423:
Sorry, at moment, I am bit caught up in other issues and may not devote time to review this PR. If it's urgent, then, maybe someone else from core team can have a look at it.

pascalgrimaud:
@kaidohallik : we plan to do a new release today, probably v6.3.1 (instead of v6.4.0) which won't contain redis/kafka because I didn't have time to review and merge these PRs.
Do you need this PR to be merged before the release ?

kaidohallik:
It's not urgent for me. It would be good if this didn't stay on hold for long. @mshima do you need this in the next release?

pascalgrimaud:
The mini roadmap is:
- v6.3.1 today for fixing security issues
- v6.4.0 the next 2 weeks, with Redis and Kafka, as I need this for my current customer, so I'll do my best to find time to include these PR
So if this PR can wait the v6.4.0 because it includes clean files, it's ok !

kaidohallik:
For me it's ok to wait the v6.4.0.

mshima:
No problem for me.

kaidohallik:
Thanks for merging @pascalgrimaud !

PR #10483, opened 2019-09-24
Better exception/log for JWT token validation

marob:
There's no need to separate exception handling as the type of exception is self-explanatory. 
Also, `{}` should not be used for exception parameter if you want to log the stacktrace.

### DIFF
--- a/generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
@@ -119,18 +119,9 @@ public class TokenProvider implements InitializingBean {
         try {
             Jwts.parser().setSigningKey(key).parseClaimsJws(authToken);
             return true;
-        } catch (io.jsonwebtoken.security.SecurityException | MalformedJwtException e) {
-            log.info("Invalid JWT signature.");
-            log.trace("Invalid JWT signature trace: {}", e);
-        } catch (ExpiredJwtException e) {
-            log.info("Expired JWT token.");
-            log.trace("Expired JWT token trace: {}", e);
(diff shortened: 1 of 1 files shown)


cbornet:
:+1: LGTM.

generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
```diff
@@ -119,18 +119,9 @@ public class TokenProvider implements InitializingBean {
...
-            log.info("Expired JWT token.");
-            log.trace("Expired JWT token trace: {}", e);
-        } catch (UnsupportedJwtException e) {
-            log.info("Unsupported JWT token.");
-            log.trace("Unsupported JWT token trace: {}", e);
-        } catch (IllegalArgumentException e) {
-            log.info("JWT token compact of handler are invalid.");
-            log.trace("JWT token compact of handler are invalid trace: {}", e);
+        } catch (JwtException | IllegalArgumentException e) {
+            log.info("Invalid JWT token.");
```
cbornet: Minor: "token" is redundant here as it's already the T of JWT

marob: Indeed, but the term was already used in previous log messages.
But I agree it's redundant and the messages could be `"Invalid JWT."` and `"Invalid JWT trace."`.

Also, note that catching `JwtException` may intercept other exceptions that those in the previous code if other exceptions implement `JwtException` and can be thrown by the code in the `try` block (but I doubt it and catching them would probably be a good idea).

atomfrede:
Thanks @marob for your first contribution to jhipster, good work!

Issue #10482, opened 2019-09-24
Add all options to skip some prompts of app generator questions


amanganiello90:
**Overview of the feature request**
Hi all wonderful community,
I suggest to support option to skip the questions that we can't override.
For some questions jhipster already provides it (skip-server, skip-client and so on),
but for the **i18n** there is the parameter without cli settings [this.skipI18n](https://github.com/jhipster/generator-jhipster/search?q=this.skipI18n&unscoped_q=this.skipI18n).

So we have to provide skipping for:
* i18n
* additional [test framework](https://github.com/jhipster/generator-jhipster/blob/b1925bbe9283d4502d749d829fa1793afb1605e8/generators/app/prompts.js#L123) 
* Additional [modules to install](https://github.com/jhipster/generator-jhipster/blob/b1925bbe9283d4502d749d829fa1793afb1605e8/generators/app/prompts.js#L153)

After in every blueprint the developer can add for default that options.

**Motivation for or Use Case**
Refers to [this comment](https://github.com/jhipster/generator-jhipster/issues/10467#issuecomment-534034014) and #9692 issue.
As is difficult to override or extend the app generator class, is useful to not display the app generator questions that we don't want into the blueprint. Then it's easy to custom the other rest of prompts in every subgenerator to have the full control of blueprint.

**Related issues or PR**
The [nodejs blueprint issue](https://github.com/jhipster/generator-jhipster-nodejs/issues/44) requires it for the question about _additional test framework_ .

-   [x] Checking this box is mandatory (this is just to show you read everything)

Kerruba:
I think also supporting skipping questions by providing a default answer would be fantastic for the #9692 issue. As a blueprint developer for example I would like to make the assumption in the blueprint that the app is always a microservice

amanganiello90:
> 
> 
> I think also supporting skipping questions by providing a default answer would be fantastic for the #9692 issue. As a blueprint developer for example I would like to make the assumption in the blueprint that the app is always a microservice

Hi @Kerruba,
you already set a your value in the blueprint generator..
You can change any response from the server generator initialization phase (in your case for the application type). For example see this:  https://github.com/jhipster/generator-jhipster-nodejs/blob/9787a64234303f95b18385cc4f7c3a61e8ef6ca1/generators/server/index.js#L99

amanganiello90:
Hi @Kerruba,
For your happiness in my PR I have also add a **--skip-app-type** option. 
[...]

PR #10493, opened 2019-09-25
Skip App questions to reduce and customize general prompt

amanganiello90:

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -70,6 +70,13 @@ module.exports = class extends BaseGenerator {
             type: Boolean,
             defaults: false
         });
+        
+         // This adds support for a `--skip-app-type` flag
+        this.option('skip-app-type', {
+            desc: 'Skip the application type question',
+            type: Boolean,
+            defaults: false
+        });
 
         // This adds support for a `--skip-git` flag
         this.option('skip-git', {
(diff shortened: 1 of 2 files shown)


amanganiello90:
PR for  #10482

murdos:
FYI there's another solution proposed to fix the same issue: https://github.com/jhipster/generator-jhipster/pull/10586

amanganiello90:
Thanks @murdos, this is a good solution in order to provide a full customization...but my solution is only to skip some question prompts without customize any subgenerator, and it not exclude that you have referenced.

avdev4j:
hi,
If we can customize questions, that's mean we can skip which we don't want to display to the user. That's why I prefer to avoid adding many and many skip options for adding a full solution to customize prompts and get benefit for all other blueprints.

So I suggest to help @mshima to secure his PR and contribute to a more complete solution.

amanganiello90:
ok @avdev4j , what does miss in that PR?

avdev4j:
First we need to help for the review and try to how test and (if needed) improve the code.
I think we can converge for a more generic solution.

murdos:
I agree, I prefer a more generic solution too.

pascalgrimaud:
I think it can be closed as https://github.com/jhipster/generator-jhipster/pull/10752 should help to customize app question

PR #10498, opened 2019-09-26
Changing the frameOptions policy to sameOrigin - Enabling access of H2Console

### LINKED ISSUE #10481: H2 Console on microservice not working
Dear all,

When I create a microservice with the following model.jdl

```
application {
   config {
     baseName certification,
     applicationType microservice,
     packageName com.mla.reglab.certification,
     serviceDiscoveryType eureka,
     authenticationType oauth2,
     clientPackageManager npm,
     jhiPrefix certification,
     devDatabaseType h2Disk,
     prodDatabaseType postgresql,
     enableSwaggerCodegen true,
     enableHibernateCache true,
     cacheProvider hazelcast,
     nativeLanguage en,
     languages [en, fr, es],
     buildTool gradle,
     serverPort 8084,
     skipUserManagement true
   }
   entities EntityA
}

entity EntityA {
  name String required
}


dto all with mapstruct
service all with serviceClass
paginate all with pagination
```

And when I try to access http://localhost:8084/h2-console I get the H2 connection screen but when I try to connect, et get this error `No suitable driver found for 08001/0`
##### **JHipster Version(s)**

Version 6.3.1

SudharakaP:
Fixes #10481

This allows the use of accessing the H2 console for example.


generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
```diff
@@ -237,7 +237,7 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
             .featurePolicy("geolocation 'none'; midi 'none'; sync-xhr 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; fullscreen 'self'; payment 'none'")
         .and()
             .frameOptions()
-            .deny()
+            .sameOrigin()
```
atomfrede: Is it possible to change that depending on the profile? It makes sense to me to use it only in development  profile while on production (where no H2 console should be there) it would be good to disable it.

mraible: Will this affect CORS or the ability to be an OAuth 2.0 resource server?

atomfrede: If yes it won't work right now as currently the policy is stricter or do I get you wrong?

SudharakaP: @atomfrede : Yes, good idea; I've added some code to only enable it for the dev profile. Let me know if you see any further issues. :smile:

SudharakaP:
Close/Reopen for failing tests. 😢

vishal423:
@SudharakaP, were you able to reproduce issue? On my side with a monolithic oauth2 application, I don't see an issue connecting to H2 (disk).

SudharakaP:
> @SudharakaP, were you able to reproduce issue? On my side with a monolithic oauth2 application, I don't see an issue connecting to H2 (disk).

This is very interesting. :thinking: 

I did reproduce the exact error with a micro-service as the OP has indicated; it never occurred to me to try with a monolith since the OP never used a monolith. But looking at the code I do think it should give the same error for a monolith as well. Let me try it with a monolith and get back to you. :thinking:

SudharakaP:
Close/Reopen for tests

SudharakaP:
@vishal423 : Just to give a heads up; It would take me a couple of days to test this again since I don't have access to my home computer for a few days; but will do some more testing around Monday and hope I can find why it works for the Oauth2 monolith. 😄

vishal423:
@SudharakaP that's fine. I believe we don't display H2 console inside an iframe on the UI side and thus, it shouldn't be affected by frame policy. I also expect this behavior to be consistent across different authentication mechanisms supported in JHipster.

SudharakaP:
> @SudharakaP that's fine. I believe we don't display H2 console inside an iframe on the UI side and thus, it shouldn't be affected by frame policy. I also expect this behavior to be consistent across different authentication mechanisms supported in JHipster.

@vishal423 : I've done some tests with the oAuth2 monolith; I believe we handle spring web security configuration differently in monoliths and microservices. Notice that for oAuth2 monolith we make [web security exceptions](https://github.com/jhipster/generator-jhipster/blob/2e1e9884d63e8071e7f5ac193c8a5f0f0e509ae9/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L182) and also we seem to handle the [CSRF policies differently](https://github.com/jhipster/generator-jhipster/blob/2e1e9884d63e8071e7f5ac193c8a5f0f0e509ae9/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L198). If you start matching those two sections of a oAuth2 monolith with the oAuth2 micro-service counterpart you'll start seeing the same access problems due to the framesets used by the h2 consoles. 

Since the h2 console uses `framesets` which have [`frames`] within them, and since the [X-Frame-Options applies to both `frames` and `iframes`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) I believe we do see the same problems related to iframes.

Does this make sense? :smile: My fix fixes the micro-service oAuth2 problem which the OP observed without touching the oAuth2 monolith flow (which I think is safe but probably not the most elegant). As you indicated maybe there probably is a different approach where we could use the same type of solution that we've used for the oAuth2 monolith flow (off the top of my head maybe using a [web security exception](https://github.com/jhipster/generator-jhipster/blob/2e1e9884d63e8071e7f5ac193c8a5f0f0e509ae9/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L182) for h2-console. 

Do you guys think we are good with this or should we try to match it with the oAuth2 monolith flow to make it consistent? 

cc: @atomfrede @mraible

vishal423:
@SudharakaP, thanks for detailed analysis. I just checked `SecurityConfiguration` and we indeed bypass security for `h2-console`. Can we do same to be consistent?

>         <%_ if (devDatabaseType === 'h2Disk' || devDatabaseType === 'h2Memory') { _%>
>             .antMatchers("/h2-console/**")
>         <%_ } _%>

SudharakaP:
@vishal423 : I've changed the logic to have consistency. Let me know if you see any further issues. :smile:

SudharakaP:
Close/Open for tests.

SudharakaP:
Close/Open again 😄

DanielFran:
Since the failing tests are related with couchbase, will merge it now.

Issue #10491, opened 2019-09-25
Openapi-client generation with microservices


vwiencek:
Hello,

I've generated a basic microservice with JHipster 6.3.1, then I've tried to generate client connector on an other microservice and I get the following error.

Steps to reproduce
- Create a microservice `serviceA` with jhipster
- Create a microservice `serviceB` with jhipster
- generate openapi-client from `serviceB` with jhipster `opencpi-client` with options: user registry

Then I get the following error during generation
- `-attribute info.name is missing`

##### **Browsers and Operating System**

-   [X] Checking this box is mandatory (this is just to show you read everything)

ecostanzi:
Hi what are you using to generate the client code? A file or a live endpoint? Can you paste the openapi yml file and the console logs?

vwiencek:
Hello, 

in order to generate the client code I do:
- `jhipster openapi-client`
- then choosing `from a JHipster /swagger-resources live endpoint`
- entering `http://localhost:8081` from aa freshly generated 6.3.1 project (see .yo file below)
- selecting `default (/v2/api-docs)`
- use Eureka discovery -> yes

```

jhipster openapi-client
INFO! Using JHipster version installed locally in current project's node_modules
INFO! Executing jhipster:openapi-client
INFO! Options: from-cli: true
Welcome to the JHipster OpenApi client Sub-Generator
? Where do you want to import your OpenAPI/Swagger specification from ? From a Jhipster /swagger-resour
ces live doc endpoint
? Enter the URL of the running Jhipster instance http://localhost:8081/
? Select the doc for which you want to create a client default (/v2/api-docs)
? Do you want to use Eureka service discovery ? Yes
? Do you want to save this config for future reuse ? No


Generating java client code for client default (http://localhost:8081/v2/api-docs)

[...]

PR #10500, opened 2019-09-27
OpenAPI Client Bug Fix: Set proper default value for `cliName` variable

SudharakaP:
While working on #10491, I realized that the word `default` is used as the default value of the variable `cliName`. Since `default` is a keyword in Java this creates problems in the generated classes. This creates a minor fix for that. 

Related to  #10491

### DIFF
--- a/generators/openapi-client/prompts.js
+++ b/generators/openapi-client/prompts.js
@@ -191,7 +191,7 @@ function askExistingAvailableDocs() {
     this.prompt(prompts).then(props => {
         if (props.availableDoc !== undefined) {
             this.props.inputSpec = props.availableDoc.url;
-            this.props.cliName = props.availableDoc.name;
+            this.props.cliName = props.availableDoc.name === 'default' ? 'apidocs' : props.availableDoc.name;
         }
         done();
     });
--- a/generators/openapi-client/prompts.js
+++ b/generators/openapi-client/prompts.js
@@ -218,7 +218,7 @@ function askGenerationInfos() {
             when: response => this.props.action === 'new' && !response.useServiceDiscovery,
             type: 'input',
             name: 'cliName',
-            message: 'What is the unique name for your API client ?',
+            message: 'What is the unique name for your API client (please avoid using Java keywords) ?',
             default: this.props.cliName || 'petstore',
             validate: input => {
                 if (!/^([a-zA-Z0-9_]*)$/.test(input)) {
--- a/generators/openapi-client/prompts.js
+++ b/generators/openapi-client/prompts.js
@@ -191,7 +191,7 @@ function askExistingAvailableDocs() {
     this.prompt(prompts).then(props => {
         if (props.availableDoc !== undefined) {
             this.props.inputSpec = props.availableDoc.url;
-            this.props.cliName = props.availableDoc.name === 'default' ? 'apidocs' : props.availableDoc.name;
+            this.props.cliName = props.availableDoc.name === 'default' ? 'apidocs' : props.availableDoc.name; // "default" cannot be used as it's a keyword in java
         }
         done();
     });


generators/openapi-client/prompts.js
```diff
@@ -191,7 +191,7 @@ function askExistingAvailableDocs() {
     this.prompt(prompts).then(props => {
         if (props.availableDoc !== undefined) {
             this.props.inputSpec = props.availableDoc.url;
-            this.props.cliName = props.availableDoc.name;
+            this.props.cliName = props.availableDoc.name === 'default' ? 'apidocs' : props.availableDoc.name; // "default" cannot be used as it's a keyword in java
```
ecostanzi: if there are other forbidden keywords we could validate the cliName when "What is the unique name for your API client ?" is prompted to the user. WDYT?

SudharakaP: @ecostanzi : That question is shown conditionally based on `this.props.action === 'new' && !response.useServiceDiscovery`. So if we add it there, it might not be triggered for all flows of the generator. Does that make sense or were you implying something else? :smile: 

Also I was wondering (maybe I am not understanding the logic here correctly), is it strictly necessary to give the user the ability to choose the api client name? Sanitizing against all java keywords will be an interesting problem (maybe there's some library to do it); but wouldn't it be simpler to maybe just take any cliName and append it with say something like, `cliName + "_apidoc"`. This way it is guaranteed there's no unintentional keyword problems.

ecostanzi: > Does that make sense or were you implying something else? 
It doesn't, I'll check it

> Sanitizing against all java keywords will be an interesting problem
You're right, maybe avoiding to suggest java keywords is enough.

The client name is the unique identifier of the single generated client. It is necessary to fill the .yo-rc.json, to choose the package name etc. Also, when generating from static files or remote URIs the client name cannot be inferred unless by parsing the content of the yml file, so we let the user to choose it.

SudharakaP: > It doesn't, I'll check it

Sorry, maybe I should have been more clearer. What I meant was suppose when prompted "Do you want to use Eureka service discovery ?" the user chooses "true". Then the prompt "What is the unique name for your API client ?" will not appear. And thus what I mean is doing a check there for cliName is probably not suitable since the question might or might not show up. Does that makes things more clearer? :smile: 

> You're right, maybe avoiding to suggest java keywords is enough.

I've added a warning to the "What is the unique name for your API client ?" prompt make things clearer. Let me know what you think. :smile:

ecostanzi: > I've added a warning to the "What is the unique name for your API client ?" prompt make things clearer. Let me know what you think.

Ok, it's clearer now

> What I meant was suppose when prompted "Do you want to use Eureka service discovery ?" the user chooses "true". Then the prompt "What is the unique name for your API client ?" will not appear.

No worries your point was clear, my answer not so much. I created the submodule from this repository https://github.com/cbornet/generator-jhipster-swagger-cli/  . This configuration was [already there](https://github.com/cbornet/generator-jhipster-swagger-cli/blob/master/generators/app/index.js#L129) and to be honest I had never noticed that the cli name is not prompted when eureka service discovery is enabled. I'll try to understand if it makes sense. Thank you for pointing it out.

SudharakaP: @ecostanzi : No problem. I'll mark this resolved since the keyword issue (which is the intention of this PR) seems resolved now. But let me know if you have any other issues. 😄

DanielFran:
@ecostanzi Can we merge it?

ecostanzi:
yes!

DanielFran:
Thanks @ecostanzi and @SudharakaP

PR #10504, opened 2019-09-27
Support inline content for JDL

deepu105:
Makes it possible to invoke import-jdl with inline content

```
jhipster import-jdl --inline \
        "application {
            config {
                baseName jhapp,
                testFrameworks [protractor]
            }
        }"
```

or 

```
jhipster import-jdl --inline "application { config { baseName jhapp, testFrameworks [protractor] }}"
```

### DIFF
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -253,7 +253,9 @@ const isAllCompleted = items => Object.values(items).every(it => it);
 
 class JDLProcessor {
     constructor(jdlFiles, jdlContent, options) {
-        logger.debug(`JDLProcessor started with content: ${jdlFiles} and options: ${toString(options)}`);
+        logger.debug(
+            `JDLProcessor started with ${jdlContent ? `content: ${jdlContent}` : `files: ${jdlFiles}`} and options: ${toString(options)}`
+        );
         this.jdlFiles = jdlFiles;
         this.jdlContent = jdlContent;
         this.options = options;
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -425,19 +425,14 @@ class JDLProcessor {
     }
 }
 
-const validateContent = jdlFilesOrContent => {
-    if (jdlFilesOrContent) {
-        if (jdlFilesOrContent.length > 1) {
-            jdlFilesOrContent.forEach(key => {
-                if (!shelljs.test('-f', key)) {
-                    logger.error(chalk.red(`\nCould not find ${key}, make sure the path is correct.\n`));
-                }
-            });
-        } else if (!shelljs.test('-f', jdlFilesOrContent[0])) {
-            return jdlFilesOrContent[0];
-        }
+const validateFiles = jdlFiles => {
+    if (jdlFiles) {
+        jdlFiles.forEach(key => {
+            if (!shelljs.test('-f', key)) {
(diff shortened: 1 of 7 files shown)


cli/import-jdl.js
```diff
@@ -248,23 +252,14 @@ const shouldTriggerInstall = (generator, index) =>
 const isAllCompleted = items => Object.values(items).every(it => it);
 
 class JDLProcessor {
-    constructor(jdlFiles, options) {
-        logger.debug(`JDLProcessor started with jdlFiles: ${jdlFiles} and options: ${toString(options)}`);
+    constructor(jdlFiles, jdlContent, options) {
+        logger.debug(`JDLProcessor started with content: ${jdlFiles} and options: ${toString(options)}`);
```
MathieuAA: I don't think the log helps a lot in this case. Maybe telling that it started with files and/or content and options is a good idea. For instance: if I call the JDLProcessor with only files, the log tells me that the processor started with files and options (no content); or if only with a content that the processor started with only a content and options.

package.json
```diff
@@ -59,7 +59,7 @@
         "glob": "7.1.3",
         "gulp-filter": "5.1.0",
         "insight": "0.10.1",
-        "jhipster-core": "4.3.0",
+        "jhipster-core": "4.4.0",
```
MathieuAA: 5.0.0 might be better here (same content)

cli/import-jdl.js
deepu105: Sure

package.json
deepu105: What about the breaking changes? Don't we have to update anything in the generator?

MathieuAA: Nop! this is the good kind of breaking change, doesn't impact the generator.

deepu105: great, what about the custom annotation stuff, I saw a comment somewhere discussing aout changing the entity json files to have options inside an `options` key? was that done?

deepu105: Also for the `SEARCH_ENGINE` change doesn't it break existing JDL files? if it has `searchEngine` in it? given that we will update and release this in a minor JHipster version may be we need to add backward compatibility for JDL files that still uses `searchEngine`. WDYT?

MathieuAA: @deepu105 about the first question, I didn't see any comment so I can't relate, nor have I seen work about this `options` key.
About the second one: in the JDL, you still use the `search` keyword. Work has been done to harmonize what's written in a JDL file and the internal name of the option holding the value. The breaking change is about what's exposed by BinaryOptions.

MathieuAA: To be honest, the first point bothers me. @yelhouti do you have any input about this? I don't remember about such a key :(

murdos: @MathieuAA : the `options` key have been suggested by you in [this comment](https://github.com/jhipster/jhipster-core/issues/320#issuecomment-485381566), but this has not been done (yet?).

MathieuAA: @murdos ahah! yep. Now I remember. That was a suggestion for using custom annotations, but doesn't break existing options

deepu105: perfect

yelhouti: @MathieuAA @deepu105 , options field is only used for custom annotations for backward compatility reasons, once we switch to JDL centric approach, I could go through the code again to harmonize the logique.

MathieuAA: @yelhouti nice :)

PR #10508, opened 2019-09-28
Clean constructor parameters in Angular

kaidohallik:
This PR:
* removes unused injected services
* if injected service is used only in constructor then don't create class field (reference for good explanation: https://stackoverflow.com/a/48571668)

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/config/uib-pagination.config.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/config/uib-pagination.config.ts.ejs
@@ -22,7 +22,7 @@ import { ITEMS_PER_PAGE } from 'app/shared/constants/pagination.constants';
 
 @Injectable({providedIn: 'root'})
 export class PaginationConfig {
-    constructor(private config: NgbPaginationConfig) {
+    constructor(config: NgbPaginationConfig) {
         config.boundaryLinks = true;
         config.maxSize = 5;
         config.pageSize = ITEMS_PER_PAGE;
--- a/generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs
@@ -100,11 +100,11 @@ import { fontAwesomeIcons } from './icons/font-awesome-icons';
     ]
 })
 export class <%=angularXAppName%>CoreModule {
-    constructor(iconLibrary: FaIconLibrary, private dpConfig: NgbDatepickerConfig<% if (enableTranslation) { %>, private languageService: JhiLanguageService<% } %>) {
+    constructor(iconLibrary: FaIconLibrary, dpConfig: NgbDatepickerConfig<% if (enableTranslation) { %>, languageService: JhiLanguageService<% } %>) {
         registerLocaleData(locale);
         iconLibrary.addIconPacks(fas);
         iconLibrary.addIcons(...fontAwesomeIcons);
-        this.dpConfig.minDate = {year: moment().year() - 100, month: 1, day: 1};
+        dpConfig.minDate = {year: moment().year() - 100, month: 1, day: 1};
         <%_ if (enableTranslation) { _%>
         languageService.init();
         <%_ } _%>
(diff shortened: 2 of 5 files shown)


generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs
```diff
@@ -100,11 +100,11 @@ import { fontAwesomeIcons } from './icons/font-awesome-icons';
     ]
 })
 export class <%=angularXAppName%>CoreModule {
-    constructor(iconLibrary: FaIconLibrary, private dpConfig: NgbDatepickerConfig<% if (enableTranslation) { %>, private languageService: JhiLanguageService<% } %>) {
+    constructor(iconLibrary: FaIconLibrary, dpConfig: NgbDatepickerConfig<% if (enableTranslation) { %>, languageService: JhiLanguageService<% } %>) {
```
MathieuAA: why did you made it public? (and why the other are too...)

generators/client/templates/angular/src/main/webapp/app/blocks/config/uib-pagination.config.ts.ejs
```diff
@@ -22,7 +22,7 @@ import { ITEMS_PER_PAGE } from 'app/shared/constants/pagination.constants';
 
 @Injectable({providedIn: 'root'})
 export class PaginationConfig {
-    constructor(private config: NgbPaginationConfig) {
+    constructor(config: NgbPaginationConfig) {
```
MathieuAA: why did you made it public? (and why the other are too...)

generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs
kaidohallik: If modifier is missing then variable is accessible only in constructor. Without modifier it has as minimal accessibility as possible and is maximally optimized. There is link in PR description with additional information. Most important is: "Using private for a parameter property declares and initializes a private member; likewise, the same is done for public, protected, and readonly".

Before this change was automatically created class member, but we don't need class members here because these variables are used only in constructor.

vishal423: yes, it will localize scope to the method

generators/client/templates/angular/src/main/webapp/app/blocks/config/uib-pagination.config.ts.ejs
vishal423: it will localize scope to the method

vishal423:
thanks @kaidohallik

generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs
MathieuAA: Yeah, that means that if you don't put the private, then it will:
  - create a instance variable of the same name with the same accessibility level
    - if you don't put anything it will be public (it is the default)
    - if you put something, this _something_ will be the accessibility level for the instance variable

Yes, it will localize the scope. However, not declaring it outside the constructor params will create an instance variable because it is put in the constructor.

An example:
```typescript
class MyClass {
  constructor(public myField: string) {}
}
```
is equivalent to
```typescript
class MyClass {
  public myField: string;
  constructor(myField: string) {
    this.myField = myField;
  }
}
```
You can replace public by whatever accessibility level.

> Before this change was automatically created class member, but we don't need class members here because these variables are used only in constructor.

And afterwards? if there's no private level then anyone could update the internal state of the the class.

@wmarques you're the angular guy, do you have an opinion about it, and why is this code working by side-effect? Maybe I'm misunderstanding things because the angular framework requires it...

kaidohallik: constructor parameters work in this manner in Angular:
* if no modifier (like this is after this change) then variable is accessible only in constructor, no instance variable is created and this variable can't be accessed nowhere else but constructor
* if modifier is specified then instance variable is created and standard accessibility logic applies:
    * if this is private then this variable is accessible only inside class
    * if this is protected then this variable is additionally available in inherited classes
    * if this is public then this variable is accessible for everyone

PR #10531, opened 2019-10-01
Fix jacoco regression after migration to xml based format

### LINKED ISSUE #10529: Sonar code coverage regression after change from binary to xml format

Code coverage report from Jacoco are not taken into account anymore by Sonar.

In version 6.3.0, Sonar config was change to use XML report instead of .exec binary report file. 
However pom.xml config does not match the config in sonar-project.properties. Sonar config expect jacoco.xml files but the pom.xml is still configure with .exec files.

Also it seems that now if we try to skip tests, the builds fails because it is trying to get the jacoco report.

##### **Suggest a Fix**

Please align pom config with:
```
<jacoco.utReportFile>${jacoco.utReportFolder}/jacoco.xml</jacoco.utReportFile>
...
<jacoco.itReportFile>${jacoco.itReportFolder}/jacoco.xml</jacoco.itReportFile>
```
##### **JHipster Version(s)**
JHipster 6.3.0+

jfougere:
Fix regression on jacoco code coverage report in maven pom

Fix #10529

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -131,9 +131,9 @@
         <properties-maven-plugin.version>1.0.0</properties-maven-plugin.version>
         <sonar-maven-plugin.version>3.6.1.1688</sonar-maven-plugin.version>
         <jacoco.utReportFolder>${project.build.directory}/jacoco/test</jacoco.utReportFolder>
-        <jacoco.utReportFile>${jacoco.utReportFolder}/test.exec</jacoco.utReportFile>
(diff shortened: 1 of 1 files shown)


SudharakaP:
@jfougere : Thanks for the pull request; I kindly suggest changing the title to something that reflects the issue, as it will help us when searching. :smile:

jfougere:
@SudharakaP yes of course however, I realised that this PR is wrong and the previous code is actually correct.

Our problem with code coverage not taken into account is that with the new XML report we do need an extra JaCoCo Plugin installed in our SonarQube instance which was not needed before.

Starting from SonarQube 7.4+ however the plugin is installed by default.

PR #10535, opened 2019-10-01
Faker: don't generate unnecessary column declaration for loading data

murdos:
Columns are declared in liquibase changeset, but not generated in csv file, in the following cases:
 - one-to-many relationship
 - one-to-one with ownerSide = false or with useJPADerivedIdentifier = true

Fixes https://github.com/jhipster/generator-jhipster/issues/10577

### DIFF
--- a/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
+++ b/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
@@ -217,7 +217,11 @@
             <%_ } _%>
             <%_ for (idx in relationships) {
                     let loadColumnType = 'numeric';
-                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired) {
+                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired
+                        && (relationships[idx].relationshipType === "many-to-one"
+                            || (relationships[idx].relationshipType === "one-to-one" && relationships[idx].ownerSide === true
+                                && (relationships[idx].useJPADerivedIdentifier == null || relationships[idx].useJPADerivedIdentifier === false))
+                    )) {
                         let baseColumnName = getColumnName(relationships[idx].relationshipName) + '_id';
                         if (relationships[idx].otherEntityNameCapitalized === 'User' && authenticationType === 'oauth2') {
                             loadColumnType = 'varchar(100)';


generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -217,7 +217,11 @@
             <%_ } _%>
             <%_ for (idx in relationships) {
                     let loadColumnType = 'numeric';
-                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired) {
+                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired
+                        && (relationshipType === "many-to-one"
+                            || (relationshipType === "one-to-one" && relationships[idx].ownerSide === true
+                                && (relationships[idx].useJPADerivedIdentifier == null || relationships[idx].useJPADerivedIdentifier === false))
+                    )) {
```
murdos: This now matches https://github.com/jhipster/generator-jhipster/blob/706fc1f91d69de989ea7df3f49586cabdf1b6de8/generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs#L203-L207

generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -217,7 +217,11 @@
             <%_ } _%>
             <%_ for (idx in relationships) {
                     let loadColumnType = 'numeric';
-                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired) {
+                    if (relationships[idx].relationshipValidate === true && relationships[idx].relationshipRequired
+                        && (relationshipType === "many-to-one"
```
duderoot: @murdos you are using in the for loop always the same `relationshipType`(line221); you will need at least ` relationshipType = relationships[idx].relationshipType;`

murdos: Oh, right! 🙃 Thanks for catching this!
I'll fix it right now...

duderoot: In order to be able to test, I added in my repo and run the generator. With that fix is working.

murdos: I ran tests locally too, and confirmed the issue you reported is fixed.

PR #10538, opened 2019-10-02
adding support for dynamic env variables, fixes #9778

### LINKED ISSUE #9778: dynamic environment variables

I would like to be able to change front-end variables directly by edit a config file instead of having to rebuild the whole app 
Having the same war/docker image for all environments

Here is an example detailing how we could implement it:
https://www.jvandemo.com/how-to-use-environment-variables-to-configure-your-angular-application-without-a-rebuild/

yelhouti:
fixes #9778

Design choices are explained in the issue

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/app.constants.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.constants.ts.ejs
@@ -24,3 +24,4 @@ export const VERSION = process.env.VERSION;
 export const DEBUG_INFO_ENABLED = !!process.env.DEBUG_INFO_ENABLED;
 export const SERVER_API_URL = process.env.SERVER_API_URL;
 export const BUILD_TIMESTAMP = process.env.BUILD_TIMESTAMP;
+// export const MY_DYNAMIC_VARIABLE = window.__env.myDynamicVariable;
--- a/generators/client/templates/angular/src/main/webapp/env.js.ejs
+++ b/generators/client/templates/angular/src/main/webapp/env.js.ejs
@@ -1,0 +1,7 @@
+/*
+The code below creates a global __env variable if not already declared,
+it can be accessed in you angular file but we recommend exposing it through app.constants.ts .
+When using kubernetes, this file (with no comments) can be mounted as a configMap, this is why we keep such a simple key value syntax
+*/
+window.__env = window.__env || {};
+// window.__env.myDynamicVariable = 'http://127.0.0.1:8090';
--- a/generators/client/templates/angular/src/main/webapp/index.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/index.html.ejs
@@ -30,6 +30,7 @@
     <link rel="shortcut icon" href="favicon.ico" />
     <link rel="manifest" href="manifest.webapp" />
     <link rel="stylesheet" href="content/css/loading.css">
+    <script src="env.js"></script>
     <!-- jhipster-needle-add-resources-to-root - JHipster will add new resources here -->
 </head>
 <body>
--- a/generators/client/files-angular.js
+++ b/generators/client/files-angular.js
@@ -108,6 +108,7 @@ const files = {
                 '404.html',
                 'index.html',
                 'manifest.webapp',
+                'env.js',
                 'content/css/loading.css'
             ]
         }
--- a/generators/client/files-react.js
+++ b/generators/client/files-react.js
@@ -108,6 +108,7 @@ const files = {
                 '404.html',
                 'index.html',
                 'manifest.webapp',
+                'env.js',
                 'content/css/loading.css'
             ]
         }
--- a/generators/client/templates/angular/.eslintignore.ejs
+++ b/generators/client/templates/angular/.eslintignore.ejs
@@ -25,3 +25,4 @@ target/
 build/
 node/
 postcss.config.js
+src/main/webapp/env.js
--- a/generators/client/templates/angular/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/angular/webpack/webpack.common.js.ejs
@@ -99,6 +99,7 @@ module.exports = (options) => ({
             { from: './<%= MAIN_SRC_DIR %>content/', to: 'content' },
             { from: './<%= MAIN_SRC_DIR %>favicon.ico', to: 'favicon.ico' },
             { from: './<%= MAIN_SRC_DIR %>manifest.webapp', to: 'manifest.webapp' },
+            { from: './<%= MAIN_SRC_DIR %>env.js', to: 'env.js' },
             // jhipster-needle-add-assets-to-webpack - JHipster will add/remove third-party resources in this array
             { from: './<%= MAIN_SRC_DIR %>robots.txt', to: 'robots.txt' }
         ]),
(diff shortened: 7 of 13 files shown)


yelhouti:
If we all agree I can add some ejs comments explaining why this file should kept as simple as this, (without fancy js...), to avoid someone else modifying it in the future
I should also add some documentation on the website to explain how to use it

generators/client/templates/angular/src/main/webapp/env.js.ejs
```diff
@@ -0,0 +1,2 @@
+window.__env = window.__env || {};
+window.__env.myDynamicVariable = 'http://127.0.0.1:8090';
```
PierreBesson: Can you use a more meaningful name ? What is this variable used for. If it is not used and just an example then it should be commented out.

yelhouti: the `window.__env` varaible is just a way to group all variable in one place to avoid polluting the window. `myDynamicVariable` is just an example one how to use the `__env` object.

It was not commented for the reasons explained in my previous comment. but I can definitely add some comments, what do you think about this changes.

yelhouti:
Please someone double check the react code/comment.

generators/client/templates/angular/src/main/webapp/env.js.ejs
PierreBesson: Great.

wmarques:
For the Angular part we can provide a service to wrap the `window.__env` object, I think this will be useful.
Do you have time to do this @yelhouti  ? and WDYT about it ?

yelhouti:
@wmarques already added to app.constants.ts, don't you think it's enough?

wmarques:
@yelhouti we could have a `ConfigurationService` with a `getConfig(key: string)` method that access this variable.

yelhouti:
@wmarques I think it's more complicated than that, often system env variables contain sensitive info, example db password... so I wouldn't expose an endpoint that can read any value from there. Also your approach is solving a different problem than the one in the issue (and doesn't solve the on in the issue).
I Also think that the problem you are trying to solve is already solved by the existing ApplicationProperties read from application.yml and can already be customized from env variables. or am I missing something?

wmarques:
@yelhouti This is a good idea to have dynamic env variables. But for me this proposal add some confusion, we now have `app.constants` and `window.__env`, some users might be confused about this. Moreover, the replacement of the `env.js` file for each env might be difficult once you packaged your app in docker IMO.

However, I think your solution might be integrated as a tip on the JHipster website :)

As @PierreBesson suggested, I would go for a backend endpoint which exposes some Spring Properties.

@jhipster/developers WDYT about this ?

pascalgrimaud:
Indeed, I'd prefer this to go to a tips, if it's simpler

yelhouti:
> Moreover, the replacement of the env.js file for each env might be difficult once you packaged your app in docker IMO.

@wmarques axtually it is not, you just put it in the /static folder, (even if spring is compiled in jar, it knows it should reed there)
>  I would go for a backend endpoint which exposes some Spring Properties.

As you can see it  the original issue, the problem with that approach is that you must wait for angular to initialize, and the httpClient..., the http interceptor might need information before startup. This is the main use case for an env.js . But if you all agree that I should put it into tips, I will.

generators/client/templates/angular/src/main/webapp/index.html.ejs
```diff
@@ -30,6 +30,7 @@
     <link rel="shortcut icon" href="favicon.ico" />
     <link rel="manifest" href="manifest.webapp" />
     <link rel="stylesheet" href="content/css/loading.css">
+    <script src="env.js"></script>
```
vishal423: I think it will memoize these settings on the user browser. What's our recommendation to reflect update in the `env.js` values?

generators/client/templates/angular/src/main/webapp/app/app.constants.ts.ejs
```diff
@@ -24,3 +24,4 @@ export const VERSION = process.env.VERSION;
 export const DEBUG_INFO_ENABLED = !!process.env.DEBUG_INFO_ENABLED;
 export const SERVER_API_URL = process.env.SERVER_API_URL;
 export const BUILD_TIMESTAMP = process.env.BUILD_TIMESTAMP;
+// export const MY_DYNAMIC_VARIABLE = window.__env.myDynamicVariable;
```
vishal423: Instead of incorporating a change without any real integration, I think it would be good to specify external server url and debug settings in this `env.js` and remove those from the webpack build. If you generate a separate front end and backend, then, the capability to specify external url is very useful.  However, I think it as a breaking change, so, this PR should be deferred for `v7`.

generators/client/templates/angular/src/main/webapp/index.html.ejs
yelhouti: @vishal423 as you might think, this env.js file is mainly is pretty static, (unless you are debugging...) so In the use cases I though about this wasn't much of a problem. in case you have a use case where this file changes a lot (not only for debugging purposes) you can (I guess) configure webpack to add a hash in the filename...

generators/client/templates/angular/src/main/webapp/app/app.constants.ts.ejs
yelhouti: @vishal423 , Indeed it would be a breaking change. maybe you can add a PR with only that change once v7 is close

pascalgrimaud:
@vishal423 @wmarques : do you think we should merge this? 
If yes: @yelhouti you need to sign the CLA and resolve conflict
If no: let's close this PR and the ticket as it's opened for too long

vishal423:
In the present form, it's just a template and doesn't provide any real useful integration in the generated application. Unless anyone has a good reason to push, I am not in favor of going ahead with it.

pascalgrimaud:
Sorry, but following @vishal423's advice and other discussion, let's close this.

@yelhouti : anyway, as you put a lot of effort on that, I can propose you to write a tips at https://www.jhipster.tech/tips/ so it will probably help other people. In this case, you will be able to claim the bounty once the tips it's done. It would be a good deal I think. Tell me what do you think about that ?

yelhouti:
@pascalgrimaud that is a deal I cant refuse :) Thanks !

pascalgrimaud:
So it's good :)
Thanks a lot @yelhouti, don't forget to mention/ping me in your pull request for the tips plz

yelhouti:
@skmaingi would you like to do the copy paste and claim the 100$, also how can I contact you in private for a Job offer?

yelhouti:
@skmaingi you know what, I'm on it and i ll ask pascal to send the money to you.

yelhouti:
@pascalgrimaud please wire the money to @skmaingi

jdubois:
Thanks -> @skmaingi please do another OpenCollective invoice and I will approve it

skmaingi:
@yelhouti I lack words to describe my gratitude. I'm humbled and grateful. Thank you so much. 
About the job offer, my email: kyuste [at] gmail

skmaingi:
@yelhouti I have made the claim https://opencollective.com/generator-jhipster/expenses/21284
I am touched by your support. Thank you

yelhouti:
@skmaingi you are welcome, please check your e-mails :)

skmaingi:
@yelhouti I will. Thank you

PR #10541, opened 2019-10-02
Check compatibility with JHipster version used by blueprint(s)

### LINKED ISSUE #10415: The version of jhipster used by a blueprint should be checked against the version of jhispter used by user

Since a blueprint extends code/class of jhipster subgen(s), it has a dependency on a specific version of JHipster. E.g. let's the blueprint lightly overrides the server subgen (overriding Dockerfile) and depends on jhpster 6.0.0
However if the user has version 6.3.0 installed and launch the generator with the former blueprint, it will end with e.g. client code in v6.3.0 but server code in v6.0.0.

The generated code could be inconsistent since some new feature or fix involve multiple subgen.
Also only one version of jhipster is saved/stored in .yo-rc.json for later use by e.g. the upgrade command. Since this version is wrong the upgrade will only be partial.

##### **Reproduce the error**

See above.

##### **Suggest a Fix**

When we check that blueprint exists, we should probably check the jhipster version compatibility and stop if it doesn't match (if user still want to generates the app, it should launch the generation with the skipCheck option)

##### **JHipster Version(s)**

6.3.0

N/A

N/A

N/A

murdos:
Fixes #10415

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -875,7 +875,7 @@ module.exports = class extends Generator {
             this.error(
                 `The installed ${chalk.yellow(
                     blueprintPkgName
-                )} blueprint targets JHipster v${blueprintJhipsterVersion} and is not compatible with this JHipster version. Either update the blueprint or JHipster.`
+                )} blueprint targets JHipster v${blueprintJhipsterVersion} and is not compatible with this JHipster version. Either update the blueprint or JHipster. You can also disable this check using --skip-checks at your own risk`
             );
         }
     }
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -790,11 +790,11 @@ module.exports = class extends Generator {
     }
 
     /**
-     * Try to retrieve the version of the blueprint used.
+     * Try to retrieve the package.json of the blueprint used, as an object.
      * @param {string} blueprintPkgName - generator name
-     * @return {string} version - retrieved version or empty string if not found
+     * @return {object} packageJson - retrieved package.json as an object or undefined if not found
      */
-    findBlueprintVersion(blueprintPkgName) {
+    findBlueprintPackageJson(blueprintPkgName) {
         let packageJsonPath = path.join(process.cwd(), 'node_modules', blueprintPkgName, 'package.json');
         try {
             if (!fs.existsSync(packageJsonPath)) {
@@ -802,13 +802,26 @@ module.exports = class extends Generator {
                 packageJsonPath = path.join(blueprintPkgName, 'package.json');
             }
             // eslint-disable-next-line global-require,import/no-dynamic-require
-            const packagejs = require(packageJsonPath);
-            return packagejs.version;
+            return require(packageJsonPath);
         } catch (err) {
             this.debug('ERROR:', err);
+            this.warning(`Could not retrieve package.json of blueprint '${blueprintPkgName}'`);
+            return undefined;
+        }
+    }
+
+    /**
+     * Try to retrieve the version of the blueprint used.
+     * @param {string} blueprintPkgName - generator name
+     * @return {string} version - retrieved version or empty string if not found
+     */
+    findBlueprintVersion(blueprintPkgName) {
+        const blueprintPackageJson = this.findBlueprintPackageJson(blueprintPkgName);
+        if (!blueprintPackageJson || !blueprintPackageJson.version) {
             this.warning(`Could not retrieve version of blueprint '${blueprintPkgName}'`);
             return '';
         }
+        return blueprintPackageJson.version;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


generators/generator-base-private.js
```diff
@@ -846,6 +859,27 @@ module.exports = class extends Generator {
...
+    checkJHipsterBlueprintVersion(blueprintPkgName) {
+        const blueprintPackageJson = this.findBlueprintPackageJson(blueprintPkgName);
+        if (!blueprintPackageJson || !blueprintPackageJson.dependencies || !blueprintPackageJson.dependencies['generator-jhipster']) {
+            this.warning(`Could not retrieve version of JHipster declared by blueprint '${blueprintPkgName}'`);
+            return;
+        }
+        const mainGeneratorJhipsterVersion = packagejs.version;
+        const blueprintJhipsterVersion = blueprintPackageJson.dependencies['generator-jhipster'];
+        if (mainGeneratorJhipsterVersion !== blueprintJhipsterVersion) {
+            this.error(
```
pascalgrimaud: I'd prefer to put a warning, instead of error, because I think some blueprint can still work, although the version is not the same.
But I'm not sure :)

arabbani:
I prefer an error instead of a warning.

generators/generator-base-private.js
murdos: Since having a different version could potentially lead to some serious errors more or less silent, and also break the upgrade process (you'll never get the changes done between the version of jhipster of blueprint and the jhipster version you're using), I'd rather throw error.

However the check can be skipped with `--skip-checks` if you know what you're doing. I could add that to the error message.
What do you think?

pascalgrimaud: agree, let's keep the error here, with the message that we can skip this check, using skip-checks

generators/generator-base-private.js
```diff
@@ -846,6 +859,27 @@ module.exports = class extends Generator {
...
+     */
+    checkJHipsterBlueprintVersion(blueprintPkgName) {
+        const blueprintPackageJson = this.findBlueprintPackageJson(blueprintPkgName);
+        if (!blueprintPackageJson || !blueprintPackageJson.dependencies || !blueprintPackageJson.dependencies['generator-jhipster']) {
+            this.warning(`Could not retrieve version of JHipster declared by blueprint '${blueprintPkgName}'`);
+            return;
+        }
+        const mainGeneratorJhipsterVersion = packagejs.version;
+        const blueprintJhipsterVersion = blueprintPackageJson.dependencies['generator-jhipster'];
+        if (mainGeneratorJhipsterVersion !== blueprintJhipsterVersion) {
```
mshima: I'd prefer to put a semver comparison here.

I do the version check on my blueprint.

6.3.0 compatibility:
https://github.com/mshima/generator-jhipster-multitenancy2/blob/master/generators/client/patcher/files.v6.3.0.js
6.3.1 compatibility:
https://github.com/mshima/generator-jhipster-multitenancy2/blob/master/generators/client/patcher/files.v6.3.1.js

mshima: I don't know if that makes a difference, but I will try to move generator-jhipster to peer dependency.

I install a jhipster version, but my blueprint will always bring the last version, and that version is used.

murdos: > 
> 
> I'd prefer to put a semver comparison here.
> 

As explained above, it doesn't work. It should be the exact same version, since the blueprint generator extends the generator class from generator-jhipster. Otherwise your application is generated with a mix of different JHipster versions as explained in the issue #10415

I understand this might be annoying for your blueprint, but its behavior seems really different from usual blueprints (such as official ones and those generated by generator-jhipster-blueprint), and I think we should target those blueprints.
You still can bypass the check with the `--skip-check` flag

murdos: > 
> I install a jhipster version, but my blueprint will always bring the last version, and that version is used.

It could roughly work if your blueprint overrides all subgenerators. Roughly, because not all subgenerators can be overridden (e.g. `app` one, and all `ci-cd` and cloud deployments ones), so you still mix different JHipster version at runtime.

generators/generator-base-private.js
murdos: Done!

generators/generator-base-private.js
mshima: What about add kind of plugin support.

mshima: https://github.com/jhipster/generator-jhipster/pull/10550

mshima: Ok. Once this is merged I will change to plugin support to make a semver test against the peer dependency version.

pascalgrimaud:
@murdos : is it ready to be merged ?

murdos:
@pascalgrimaud : Yes, it's ready, let's merge it!

PR #10545, opened 2019-10-02
k8s: use kubeconfig instead of namespace in yml files fixes #9777

### LINKED ISSUE #9777: kubernetes namespace should not be present the yml files
Kubectl can't deploy to the right namespace since the valu ein the context is overridden by the yml file

Having different namespaces for test and prod

the namespace can be defined in kubeconf per context so that each one uses it's own namespace
##### **Reproduce the error**
generate kubernetes

##### **Suggest a Fix**
remove kubernetesNamespace from the generator (or have it optional is you want)

##### **JHipster Version(s)**

5.0.1 this not a regression

yelhouti:
fix https://github.com/jhipster/generator-jhipster/issues/9777

### DIFF
--- a/generators/kubernetes/templates/README-KUBERNETES.md.ejs
+++ b/generators/kubernetes/templates/README-KUBERNETES.md.ejs
@@ -14,12 +14,52 @@ $ <%= dockerPushCommand %> <%= appConfigs[i].targetImageName %>
 ```
 
 ## Deployment
-
+Add the ssh keys for the user
 You can deploy all your apps by running the below bash command:
 
 ```
+export KUBECONFIG=./kubeconfig
 ./kubectl-apply.sh
 ```
+
+example of kubeconfig file
+```
+kind: Config
+preferences: {}
+apiVersion: v1
+current-context: dev
+clusters:
+  - name: dev
+    cluster:
+      certificate-authority: ssl/dev/ca.crt
+      server: https://<dev-cluster-ip>:6443
+  - name: prod
+    cluster:
+      certificate-authority: ssl/prod/ca.crt
+      server: https://<dev-cluster-ip>:6443
+contexts:
+  - name: dev
+    context:
+      cluster: dev
+      user: admin-dev
+  - name: prod
+    context:
+      cluster: prod
+      user: admin-prod
+users:
+  - name: admin-dev
+    user:
+      client-certificate: ssl/dev/user.crt
+      client-key: ssl/dev/user.key
+  - name: admin-prod
+    user:
+      client-certificate: ssl/prod/user.crt
+      client-key: ssl/prod/user.key
+```
+
+to switch from dev to prod edit the kubeconfig file `currentContext` from dev to prod or using
+
+More info here: https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-context-and-configuration
 
 ## Exploring your services
 
--- a/generators/kubernetes/templates/deployment.yml.ejs
+++ b/generators/kubernetes/templates/deployment.yml.ejs
@@ -20,7 +20,6 @@ apiVersion: <%= KUBERNETES_DEPLOYMENT_API_VERSION %>
 kind: Deployment
 metadata:
   name: <%= app.baseName.toLowerCase() %>
-  namespace: <%= kubernetesNamespace %>
 spec:
   replicas: 1
   selector:
@@ -72,14 +71,14 @@ spec:
           value: prod
         <%_ if (app.serviceDiscoveryType === 'eureka') { _%>
         - name: SPRING_CLOUD_CONFIG_URI
-          value: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/config
+          value: http://admin:${jhipster.registry.password}@jhipster-registry:8761/config
         - name: JHIPSTER_REGISTRY_PASSWORD
           valueFrom:
             secretKeyRef:
               name: registry-secret
               key: registry-admin-password
         - name: EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE
-          value: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/
+          value: http://admin:${jhipster.registry.password}@jhipster-registry:8761/eureka/
           <%_ if (istio) { _%>
         - name: EUREKA_INSTANCE_PREFER_IP_ADDRESS
           value: "false"
@@ -89,7 +88,7 @@ spec:
         <%_ } _%>
         <%_ if (app.serviceDiscoveryType === 'consul') { _%>
         - name: SPRING_CLOUD_CONSUL_HOST
-          value: consul.<%= kubernetesNamespace %>.svc.cluster.local
+          value: consul
         - name: SPRING_CLOUD_CONSUL_PORT
           value: "8500"
           <%_ if (istio) { _%>
@@ -110,11 +109,11 @@ spec:
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'mysql') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql.<%= kubernetesNamespace %>.svc.cluster.local:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false
+          value: jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'mariadb') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb.<%= kubernetesNamespace %>.svc.cluster.local:3306/<%= app.baseName.toLowerCase() %>
+          value: jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb:3306/<%= app.baseName.toLowerCase() %>
         - name: SPRING_DATASOURCE_PASSWORD
           valueFrom:
             secretKeyRef:
@@ -123,7 +122,7 @@ spec:
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'postgresql') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql.<%= kubernetesNamespace %>.svc.cluster.local:5432/<%= app.baseName %>
+          value: jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql:5432/<%= app.baseName %>
         - name: SPRING_DATASOURCE_USERNAME
           value: <%= app.baseName %>
         - name: SPRING_DATASOURCE_PASSWORD
@@ -146,9 +145,9 @@ spec:
         <%_ } _%>
         <%_ if (app.searchEngine === 'elasticsearch') { _%>
         - name: SPRING_DATA_JEST_URI
-          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.<%= kubernetesNamespace %>.svc.cluster.local:9200
+          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch:9200
         - name: SPRING_ELASTICSEARCH_REST_URIS
-          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.<%= kubernetesNamespace %>.svc.cluster.local:9200
+          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch:9200
         <%_ } _%>
         <%_ if (app.messageBroker === 'kafka') { _%>
         - name: KAFKA_CONSUMER_KEY_DESERIALIZER
@@ -156,13 +155,13 @@ spec:
         - name: KAFKA_CONSUMER_VALUE_DESERIALIZER
           value: 'org.apache.kafka.common.serialization.StringDeserializer'
         - name: KAFKA_CONSUMER_BOOTSTRAP_SERVERS
-          value: 'jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'jhipster-kafka:9092'
         - name: KAFKA_CONSUMER_GROUP_ID
           value: '<%= app.baseName.toLowerCase() %>'
         - name: KAFKA_CONSUMER_AUTO_OFFSET_RESET
           value: 'earliest'
         - name: KAFKA_PRODUCER_BOOTSTRAP_SERVERS
-          value: 'jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'jhipster-kafka:9092'
         - name: KAFKA_PRODUCER_KEY_DESERIALIZER
           value: 'org.apache.kafka.common.serialization.StringDeserializer'
         - name: KAFKA_PRODUCER_VALUE_DESERIALIZER
--- a/generators/kubernetes/templates/istio/gateway.yml.ejs
+++ b/generators/kubernetes/templates/istio/gateway.yml.ejs
@@ -20,7 +20,6 @@ apiVersion: <%= KUBERNETES_ISTIO_NETWORKING_API_VERSION %>
 kind: Gateway
 metadata:
   name: <%= app.baseName.toLowerCase() %>-gateway
-  namespace: <%= kubernetesNamespace %>
   labels:
     gateway: <%= app.baseName.toLowerCase() %>-gateway
     istio: ingressgateway
@@ -45,7 +44,6 @@ apiVersion: <%= KUBERNETES_ISTIO_NETWORKING_API_VERSION %>
 kind: VirtualService
 metadata:
   name: <%= app.baseName.toLowerCase() %>-gw-virtualservice
-  namespace: <%= kubernetesNamespace %>
   labels:
     service: <%= app.baseName.toLowerCase() %>-gw-virtualservice
 spec:
--- a/generators/kubernetes/templates/messagebroker/kafka.yml.ejs
+++ b/generators/kubernetes/templates/messagebroker/kafka.yml.ejs
@@ -20,7 +20,6 @@ apiVersion: <%= KUBERNETES_DEPLOYMENT_API_VERSION %>
 kind: Deployment
 metadata:
   name: jhipster-kafka
-  namespace: <%= kubernetesNamespace %>
 spec:
   replicas: 1
   selector:
@@ -40,9 +39,9 @@ spec:
         image: <%= DOCKER_KAFKA %>
         env:
         - name: KAFKA_ADVERTISED_LISTENERS
-          value: 'PLAINTEXT://jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'PLAINTEXT://jhipster-kafka:9092'
         - name: KAFKA_ZOOKEEPER_CONNECT
-          value: 'jhipster-zookeeper.<%= kubernetesNamespace %>.svc.cluster.local:2181'
+          value: 'jhipster-zookeeper:2181'
         - name: KAFKA_BROKER_ID
           value: '2'
         - name: KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR
@@ -54,7 +53,6 @@ apiVersion: <%= KUBERNETES_CORE_API_VERSION  %>
 kind: Service
 metadata:
   name: jhipster-kafka
-  namespace: <%= kubernetesNamespace %>
 spec:
   selector:
     app: jhipster-kafka
@@ -65,7 +63,6 @@ apiVersion: <%= KUBERNETES_DEPLOYMENT_API_VERSION %>
 kind: Deployment
 metadata:
   name: jhipster-zookeeper
-  namespace: <%= kubernetesNamespace %>
 spec:
   replicas: 1
   selector:
@@ -97,7 +94,6 @@ apiVersion: <%= KUBERNETES_CORE_API_VERSION  %>
 kind: Service
 metadata:
   name: jhipster-zookeeper
-  namespace: <%= kubernetesNamespace %>
 spec:
   selector:
     app: jhipster-zookeeper
--- a/generators/kubernetes/templates/registry/application-configmap.yml.ejs
+++ b/generators/kubernetes/templates/registry/application-configmap.yml.ejs
@@ -20,7 +20,6 @@ apiVersion: <%= KUBERNETES_CORE_API_VERSION  %>
 kind: ConfigMap
 metadata:
   name: application-config
-  namespace: <%= kubernetesNamespace %>
 #common configuration shared between all applications
 data:
   application.yml: |-
@@ -51,5 +50,5 @@ data:
     eureka:
       client:
         service-url:
-          defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/
+          defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry:8761/eureka/
   <%_ } _%>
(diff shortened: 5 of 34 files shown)


generators/kubernetes/templates/registry/consul-config-loader.yml.ejs
```diff
@@ -39,7 +38,7 @@ spec:
         - name: INIT_SLEEP_SECONDS
           value: "5"
         - name: CONSUL_URL
-          value: consul.<%= kubernetesNamespace %>.svc.cluster.local
+          value: consul.
```
DanielFran: substitute "consul." by "consul"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -72,14 +71,14 @@ spec:
           value: prod
         <%_ if (app.serviceDiscoveryType === 'eureka') { _%>
         - name: SPRING_CLOUD_CONFIG_URI
-          value: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/config
+          value: http://admin:${jhipster.registry.password}@jhipster-registry.:8761/config
```
DanielFran: Substitute "http://admin:${jhipster.registry.password}@jhipster-registry.:8761/config" by "http://admin:${jhipster.registry.password}@jhipster-registry:8761/config"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -72,14 +71,14 @@ spec:
...
         - name: JHIPSTER_REGISTRY_PASSWORD
           valueFrom:
             secretKeyRef:
               name: registry-secret
               key: registry-admin-password
         - name: EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE
-          value: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/
+          value: http://admin:${jhipster.registry.password}@jhipster-registry.:8761/eureka/
```
DanielFran: Substitute "http://admin:${jhipster.registry.password}@jhipster-registry.:8761/eureka/" by "http://admin:${jhipster.registry.password}@jhipster-registry:8761/eureka/"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -89,7 +88,7 @@ spec:
         <%_ } _%>
         <%_ if (app.serviceDiscoveryType === 'consul') { _%>
         - name: SPRING_CLOUD_CONSUL_HOST
-          value: consul.<%= kubernetesNamespace %>.svc.cluster.local
+          value: consul.
```
DanielFran: Substitute "consul." by "consul"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -110,11 +109,11 @@ spec:
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'mysql') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql.<%= kubernetesNamespace %>.svc.cluster.local:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false
+          value: jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql.:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false
```
DanielFran: Substitute "jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql.:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false" by "jdbc:mysql://<%= app.baseName.toLowerCase() %>-mysql:3306/<%= app.baseName.toLowerCase() %>?useUnicode=true&characterEncoding=utf8&useSSL=false"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -110,11 +109,11 @@ spec:
...
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'mariadb') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb.<%= kubernetesNamespace %>.svc.cluster.local:3306/<%= app.baseName.toLowerCase() %>
+          value: jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb.:3306/<%= app.baseName.toLowerCase() %>
```
DanielFran: Substitute "jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb.:3306/<%= app.baseName.toLowerCase() %>" by "jdbc:mariadb://<%= app.baseName.toLowerCase() %>-mariadb:3306/<%= app.baseName.toLowerCase() %>"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -123,7 +122,7 @@ spec:
         <%_ } _%>
         <%_ if (app.prodDatabaseType === 'postgresql') { _%>
         - name: SPRING_DATASOURCE_URL
-          value: jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql.<%= kubernetesNamespace %>.svc.cluster.local:5432/<%= app.baseName %>
+          value: jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql.:5432/<%= app.baseName %>
```
DanielFran: Substitute "jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql.:5432/<%= app.baseName %>" by "jdbc:postgresql://<%= app.baseName.toLowerCase() %>-postgresql:5432/<%= app.baseName %>"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -146,23 +145,23 @@ spec:
         <%_ } _%>
         <%_ if (app.searchEngine === 'elasticsearch') { _%>
         - name: SPRING_DATA_JEST_URI
-          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.<%= kubernetesNamespace %>.svc.cluster.local:9200
+          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.:9200
```
DanielFran: Substitute "http://<%= app.baseName.toLowerCase() %>-elasticsearch.:9200" by "http://<%= app.baseName.toLowerCase() %>-elasticsearch:9200"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -146,23 +145,23 @@ spec:
...
         - name: SPRING_ELASTICSEARCH_REST_URIS
-          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.<%= kubernetesNamespace %>.svc.cluster.local:9200
+          value: http://<%= app.baseName.toLowerCase() %>-elasticsearch.:9200
```
DanielFran: Substitute "http://<%= app.baseName.toLowerCase() %>-elasticsearch.:9200" by "http://<%= app.baseName.toLowerCase() %>-elasticsearch:9200"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -146,23 +145,23 @@ spec:
...
         <%_ } _%>
         <%_ if (app.messageBroker === 'kafka') { _%>
         - name: KAFKA_CONSUMER_KEY_DESERIALIZER
           value: 'org.apache.kafka.common.serialization.StringDeserializer'
         - name: KAFKA_CONSUMER_VALUE_DESERIALIZER
           value: 'org.apache.kafka.common.serialization.StringDeserializer'
         - name: KAFKA_CONSUMER_BOOTSTRAP_SERVERS
-          value: 'jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'jhipster-kafka.:9092'
```
DanielFran: Substitute "'jhipster-kafka.:9092'" by "'jhipster-kafka:9092'"

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -146,23 +145,23 @@ spec:
...
         - name: KAFKA_CONSUMER_GROUP_ID
           value: '<%= app.baseName.toLowerCase() %>'
         - name: KAFKA_CONSUMER_AUTO_OFFSET_RESET
           value: 'earliest'
         - name: KAFKA_PRODUCER_BOOTSTRAP_SERVERS
-          value: 'jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'jhipster-kafka.:9092'
```
DanielFran: SUbstitute "'jhipster-kafka.:9092'" by "'jhipster-kafka:9092'"

generators/kubernetes/templates/messagebroker/kafka.yml.ejs
```diff
@@ -40,9 +39,9 @@ spec:
         image: <%= DOCKER_KAFKA %>
         env:
         - name: KAFKA_ADVERTISED_LISTENERS
-          value: 'PLAINTEXT://jhipster-kafka.<%= kubernetesNamespace %>.svc.cluster.local:9092'
+          value: 'PLAINTEXT://jhipster-kafka.:9092'
```
DanielFran: Substitute "'PLAINTEXT://jhipster-kafka.:9092'" by "'PLAINTEXT://jhipster-kafka:9092'"

generators/kubernetes/templates/messagebroker/kafka.yml.ejs
```diff
@@ -40,9 +39,9 @@ spec:
...
         - name: KAFKA_ZOOKEEPER_CONNECT
-          value: 'jhipster-zookeeper.<%= kubernetesNamespace %>.svc.cluster.local:2181'
+          value: 'jhipster-zookeeper.:2181'
```
DanielFran: Substitute "'jhipster-zookeeper.:2181'" by "'jhipster-zookeeper:2181'"

generators/kubernetes/templates/registry/application-configmap.yml.ejs
```diff
@@ -51,5 +50,5 @@ data:
     eureka:
       client:
         service-url:
-          defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/eureka/
+          defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry.:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry.:8761/eureka/
```
DanielFran: Substitute "defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry.:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry.:8761/eureka/" by "defaultZone: http://admin:${spring.security.user.password}@jhipster-registry-0.jhipster-registry:8761/eureka/,http://admin:${spring.security.user.password}@jhipster-registry-1.jhipster-registry:8761/eureka/"

yelhouti:
@DanielFran Oups, thank for catching that, one comment is enough :)

yelhouti:
@DanielFran Done :)

PierreBesson:
I don't understand why we generate a kubeconfig file. Normally this file depends on your Kubernetes cluster. I don't really understand what you are trying to do but I think it cannot fit all use cases.

yelhouti:
@PierreBesson I generate it for the same reason we generate all files, speed up the development process, and enforce best practices.

This file is written in a away to teach new users of kubernetes how to switch context..., you can still modify it with specific content..., also I hesitate to add a .gitignore file with ssl folder, to avoid users committing those..., but since they are not required (for other auth methods) I didn't

PierreBesson:
@yelhouti but if it doesn't work out the box then there is no need to generate this file. It will only confuse the users. It's better to write this information in the docs.

generators/kubernetes/templates/kubeconfig.ejs
```diff
@@ -0,0 +1,31 @@
+kind: Config
```
PierreBesson: Please remove this file.

yelhouti: sure, putting it in the README

PierreBesson: Thanks

yelhouti: Done

deepu105:
I also don't like this approach. I do not want to meddle with my kubeconfig as well, Also having the namespace on the YAML was a best practice AFAIK, what is the advantage of this over having the NS?

generators/kubernetes/templates/deployment.yml.ejs
```diff
@@ -72,14 +71,14 @@ spec:
           value: prod
         <%_ if (app.serviceDiscoveryType === 'eureka') { _%>
         - name: SPRING_CLOUD_CONFIG_URI
-          value: http://admin:${jhipster.registry.password}@jhipster-registry.<%= kubernetesNamespace %>.svc.cluster.local:8761/config
+          value: http://admin:${jhipster.registry.password}@jhipster-registry:8761/config
```
deepu105: would this still work if I specify a namespace during deployment?

deepu105:
@srinivasa-vasu what do you think?

generators/kubernetes/templates/deployment.yml.ejs
yelhouti: of course

yelhouti:
> Does everything go to default namespace now? 
Everything goes to what is specified in your kubeconfig (the file I was requested to remove, because everyone would know hoe to use one according to @PierreBesson ). I could put it back if you guys agree.
> Would this work if I set a default NS in config or pass it via `--namespace` flag during `kubectl apply`?
Yes it would

deepu105:
We shouldn't generate any kubeconfig file. What we could do is add the namespace in the commands of the generated `kubectl-apply.sh` file, so the namespace user chooses during generation, is used when applying using the file. Anyways I have mixed feelings about this change. Let's see what @srinivasa-vasu @saturnism etc have to say

deepu105:
@yelhouti did you test this with Istio, coz it also uses namepace in the generated files

yelhouti:
@deepu105 I disagree, if we add them there, they would take presedence over kubeconfig,

deepu105:
@yelhouti in that case you are kind of making the namespace question useless. IMO this change only makes it more confusing especially for novice users and I still don't see how this is more beneficial over having the namespace in the files

yelhouti:
@deepu105 I checked the code and it seemed to me that istio uses it own namespace (in it's files), I also looked at all places that had cluster.local and removed it

deepu105:
@yelhouti I meant if you look at Istio related files it uses the application namespaces in many places, especially in route URLs and stuff, for example, check `destination-rule.yml`, `gateway.yml` and so on, basically search for `<%= kubernetesNamespace %>` in the codebase

yelhouti:
@deepu105 it was removed from there two, my point is as follow:

everything that should be deployed to the namespace chosen by the user, should not have the namespace in it, the namespace should be defined in the kubeconfig.

 I didn't invent this rule it is suggested by kubernetes team:

https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-organizing-with-namespaces

quotes:
```
If you specify a namespace in the YAML declaration, the resource will always be created in that namespace. If you try to use the “namespace” flag to set another namespace, the command will fail.
```

yelhouti:
@deepu105 I use kubernets every week, and I'm pretty sure this is the right way to go...

deepu105:
@yelhouti I don't see the k8s team suggesting this as a best practice, they are just mentioning that declaring NS in file means you cant set it via a flag, do they anywhere say that one practice is better than other? What they say is

> There are two ways to explicitly tell Kubernetes in which Namespace you want to create your resources.

> One way is to set the “namespace” flag when creating the resource:
> You can also specify a Namespace in the YAML declaration.
> If you specify a namespace in the YAML declaration, the resource will always be created in that namespace. If you try to use the “namespace” flag to set another namespace, the command will fail.

Anyway let's see what others think

yelhouti:
@deepu105 sure, but if you think about it, the way it is said implies that, specifying it in the file makes it impossible to use the same config in two different namespace (ex: myProject-dev, myProject-prod). when not specifies, the kubeconfig allows to exactly that.

Check my comment in the k8s README: https://github.com/jhipster/generator-jhipster/pull/10545/files#diff-0

generators/kubernetes/templates/registry/application-configmap.yml.ejs
yelhouti: All resolved, please remove change request

srinivasa-vasu:
@yelhouti - I understand the reason behind removing the namespace tag. Both the approaches have their own pros and cons. The disadvantage of having namespace tag in the manifests is mitigated by the code generator tool. It so happens that if I want to apply the manifests to a specific namespace, then it needs to be passed as part of your kubectl command. But in my kubeconfig, if I set some other space as the default namespace and if I forget to pass the namespace in the command, then everything would get deployed to a namespace that I don't intend to do. That's the reason we have kept the namespace in all the files and it makes sense to do so while you are generating the code. This operation is idempotent, so if you need to generate it for a different space, just run the jhipster generator again. In that way, generated files control everything and I don't need to play around with the namespace tags. Hope this makes sense.

Really appreciate your thought process here, but personally I would prefer to keep the namespace tag as such.

yelhouti:
@srinivasa-vasu , I undertsand what you propose, but it clearly goes against the code duplication principal.

If I have 3 environment (dev, staging, prod) I would need 3 projects, and any change in one of them must be re applied to the others, so your disadvantage is not really solved.

About the kubeconfig approach, the risk of using the wrong namespace disappears if you use sane defaults. (ie: never override you globale kubeconfig) and use one kubeconfig par project (or even namespace)

If your environement are different from each other and you want a script for running each one, all you need is:
```
set -e
export ./kubeconfig
kubectl config use-context dev
kubectl apply -f ...
kubectl apply -f ...
kubectl apply -f ...
kubectl config use-context default
```

Or if you are paranoid, you can even have a kubeconfig file per environment...
 
By the way, I used to do exactly what do until a more senior person convinced me of the opposite, and please don't get me wrong, I don't mind being hard to convince, alsmost all of use (in this field) are pretty hard to convince :)

yelhouti:
@deepu105 Not necessarely, we could still use it to generate the kubeconfig. plus the place where it is still used are: 
- ingress, I think we could rename that variable,
- helm I think we should use the same approach
bur will still first need to agree on if ou guys want to switch. What do you think

srinivasa-vasu:
@yelhouti - Don't you keep the namespace (name) same between dev, stage and prod? If so, then there is no duplication. But the reality is, that's always not the case, right? You might have to maintain config info per environment if you decide to use K8s manifests. If that's the case, you anyway have to maintain separate config files per environment.

yelhouti:
@srinivasa-vasu in case you are a kubeconfig, then yes, it's one file per env, in case you hard code namespace into the file, you would need w whole different project.

yelhouti:
@deepu105 @DanielFran  sorry for the ping, could you guys validate the changes each of you requested, or direct me on how I should proceed? thanks

generators/kubernetes/templates/istio/gateway.yml.ejs
```diff
@@ -20,7 +20,6 @@ apiVersion: <%= KUBERNETES_ISTIO_NETWORKING_API_VERSION %>
 kind: Gateway
 metadata:
   name: <%= app.baseName.toLowerCase() %>-gateway
-  namespace: <%= kubernetesNamespace %>
```
deepu105: There is still `<%= app.baseName.toLowerCase() %>.<%= kubernetesNamespace %>.<%= ingressDomain %>` could you please check for all occurance of `kubernetesNamespace` in templates

yelhouti:
@deepu105 thank for the feedback, I will rename the variable when needed if we decide to go with it.
I think the main concern here it to make sure that nothing is deployed to the wrong namespace, this is why we keep the config that an env variable that is reset in each terminal.
I'm not just trying to get the PR passed, I am really looking for your constructive feedback so please indulge me and my questions.
1- I know that kubectl comes with a default kubeconfig poiting to localhost, that you should not modify, so if you don't like kubeconfig en variable you can use `--kubeconfig=config-file`, so unless I am missing something, I don't see how you can work without a kubeconfig file to point to your cluster with your credentials ... isn't a custom kubeconfig kind of required wven for beginners?
2- kubernetes states for beginners that they should use the default namespace so I don't feel too bas about that: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#when-to-use-multiple-namespaces
> Namespaces are intended for use in environments with many users spread across multiple teams, or projects. For clusters with a few to tens of users, you should not need to create or think about namespaces at all. Start using namespaces when you need the features they provide.

generators/kubernetes/templates/istio/gateway.yml.ejs
yelhouti: I will do as soon as we decide with @PierreBesson if we follow this route

generators/kubernetes/templates/README-KUBERNETES.md.ejs
```diff
@@ -14,13 +14,53 @@ $ <%= dockerPushCommand %> <%= appConfigs[i].targetImageName %>
 ```
 
 ## Deployment
-
+Add the ssh keys for the user
 You can deploy all your apps by running the below bash command:
 
 ```
+export KUBECONFIG=./kubeconfig
```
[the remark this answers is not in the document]
yelhouti: I don't see any workflow where you could avoid using a kubeconfig file, you could have many and export one of them to switch between clusters/namespaces... (or pass it as a command argument) so sorry I feel like I don't totally understand your concern

deepu105:
@yelhouti no worries, we are here to learn from each other.

Well, I use Kubernetes at my company a lot and our workflows are quite different. I(can say the same for my collegues) have honestly never used a kubeconfig file outside of the default(Execept for in automated tests) that is present in `~/.kube/config`. All my cluster information is stored in that file and I choose the `current-context` when I switch between them. Normally we never used same cluster to deploy different environment as our workloads were always on GKE, EKS or OpenShift, we have one cluster per environment and hence same namespace on all.


When I create a new cluster for example with GKE, I would execute `gcloud container clusters get-credentials <cluster-name>` or equivalent provided by EKS/AKS and so on so that the cluster details are added to my `~/.kube/config`. 

Also, I agree that NS is not a big deal unless you deploy multiple apps to the same cluster.

yelhouti:
@deepu105 Oh I see thanks, so with my approach the different would be that your `~/.kube/config` would need to be modified to add one line: 
```
- name: context1
  context:
    cluster: mycluster
    namespace: mynamespace
    user: admin
```
Or if you use the same cluster for two things, you would need to copy the context 
- name: context-ns1
  context:
    cluster: mycluster
    namespace: ns1
    user: admin
- name: context-ns2
  context:
    cluster: mycluster
    namespace: ns2
    user: admin
```
And It should all work as it used to, am I right?

deepu105:
Well, the thing is I don't want to add the namespace there, I rarely modify that file. But the real point is there is no universal workflow here, everyone follows a workflow that they are used to, we should just try to make what we generate as generic as possible without assuming any specific workflow and without the end-user having to fiddle around with any config just to get it working

srinivasa-vasu:
I know there is a separate issue opened to remove the namespace tag. Though I am inclined towards removing the `service` local references, but really not from the templates. As we are generating the code, to me it makes sense to be self-contained. However, we can provide kustomize template on top to externalize things. Have a look at [this](https://github.com/jhipster/generator-jhipster/pull/11038)
@yelhouti @deepu105

deepu105:
Looks good. I have not tried kustomize yet, will play around with it soon
and get back to you

saturnism:
eak i def missed this conversation for a while.

my preferences had always been - no ns in the yaml file. if needed, specify ns during deployment. use context to deploy to different clusters.

and +1 to use kustomize for to change configuration and/or override namespaces for different environments.

pascalgrimaud:
@deepu105 @PierreBesson @xetys @srinivasa-vasu @saturnism : what is the state of this PR ?
Do you all agree with this PR ? Should we merge this or not ?

As https://github.com/jhipster/generator-jhipster/issues/9777 is opened for too long, I'm trying to clean old tickets.

pascalgrimaud:
@yelhouti : sorry for taking so long, @PierreBesson said me he'll have a last look on this.
Can you resolve the merge conflict plz ?

yelhouti:
@pascalgrimaud when do you need it?

pascalgrimaud:
@yelhouti : I plan to do the release this monday, unless someone in the core team is against, so I'd like to have this, if @PierreBesson agrees :)

pascalgrimaud:
Sorry for this @yelhouti 
This PR is opened for too long (8 months), still conflicts, and no core team member want to merge this.
So that's why I'm closing this PR and the ticket too.

cc @jhipster/developers

Issue #10528, opened 2019-09-30
Improve jhipster client/server workflow.


mshima:
##### **Overview of the feature request**

Yeoman generators are designed to work in parallel (a priority must be run in every generator before moving to another).

We must be clear on what each priority should do:
- initializing: display logo, and load config needed by prompting.
- prompting: display prompts, prompts should save directly to this.config.
- configuring: configure itself
- default: do everything related to another generator (ex: server should load client config or setting minimum config if client is skipped)
- writing: write application files.
- end: prepare the application for the user.

Yeoman built in this.config is stored in a shared memory. So when we save in one generator, it will be available to read on another generator immediately.
- configs must be saved to this.config immediately instead of saving to the generator then saving to this.config on a latter step.

The variable configOptions name is confusing.
- we should rename to runtimeOptions.

##### **Motivation for or Use Case**

Make workflow simpler.

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/issues/10247
https://github.com/jhipster/generator-jhipster/issues/10205#issuecomment-520693655

-   [x] Checking this box is mandatory (this is just to show you read everything)

PR #10554, opened 2019-10-03
Parallel build client/common/server/languages

mshima:
Move client/common/server/languages to same execution cycle, so they are queued in parallel.
Making configOptions synchronized between priorities.

Part of https://github.com/jhipster/generator-jhipster/issues/10528
https://github.com/jhipster/generator-jhipster/pull/10527 follow up


generators/app/index.js
```diff
@@ -364,53 +364,9 @@ module.exports = class extends BaseGenerator {
...
-                });
-            },
-
-            askFori18n: prompts.askFori18n
-        };
-    }
-
-    get default() {
-        return {
+            askFori18n: prompts.askFori18n,
```
deepu105: Doesn't this depend on value from the client?

deepu105:
@mshima great work, these are the most complex parts of JHipster, I'm glad there are more people taking interest in these

generators/app/index.js
mshima: askFori18n shouldn't depend on value from client.

common/client/server modules depends on each other, but at default step, every module have configuration in place.

That the reason of https://github.com/mshima/generator-jhipster/commit/25c469bf7ee12a5538050ba8919a75a2c7823105

If you have doubt about yeoman workflow, take a look at the log from azure project step.
I've enabled yeoman debug.
And you can take a look at https://github.com/yeoman/generator/blob/2202fab7231707aeabbc2461d41e6bcafc743524/test/generators-compose-workflow.js

mshima:
> * So any refactoring here needs to be carefully tested. I agree that the current code is quite complex as it grew over time. I also agree that this needs to be refactored, but needs to manually tested as our tests still don't cover all sort of use cases

That's the reason of https://github.com/jhipster/generator-jhipster/pull/10527, it should be safe for 6.x.

This PR should be for 7.x

mraible:
Please fix conflicts in this pull request.

mshima:
I will fix this PR once jhipster 7 cycle starts.

PR #10563, opened 2019-10-05
Readonly entity: Server

### LINKED ISSUE #8800: Proposed new keyword: "readonly"

At the end of JDL file, users could add something like:

`readonly *
`

Or

`readonly A, B, C
`

Or

`readonly * except D, E
`

Entities defined as read-only should work almost exactly like others, with just a few differences:

* Frontend should only provide list and "View" pages for these entities. It won't provide "Create new", "Edit" or "Delete" options.

* Backend should only generate "findOne" and "findAll" methods for these entities. It won't generate "save" (create/update) and "delete" methods.

* Database won't need AUTO_INCREMENT in table id for these entities (thus no need for @GeneratedValue in entity id declaration).

* Sometimes an application must also reference some external data, such as additional tables imported periodically from a third-party system, or replicated online from another database, or simply managed by a different application. In this case, it's necessary to map these external tables because of relationships with other application tables referencing them, but the external tables must never be changed by the application.

* Sometimes applications need some kind of fixed "domain" information, stored in the database instead of hard-coded enums, that shouldn't be modified by the application. In this case, it can be useful to visualize this information inside the application (especially when editing entities referencing them), without ever modifying this information from the application itself.

It seems no similar features have been requested yet.

murdos:
Fixes #8800
⚠️ https://github.com/jhipster/generator-jhipster/pull/10562 should be merged before

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
@@ -119,6 +119,7 @@ import static org.elasticsearch.index.query.QueryBuilders.*;<% } %>
 public class <%= entityClass %>Resource {
 
     private final Logger log = LoggerFactory.getLogger(<%= entityClass %>Resource.class);
+<%_ if (!readOnly) { _%>
 
     <%_ let entityName = entityInstance;
     if (clientRootFolder && !skipUiGrouping) {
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
I think the service part should not be concerned by this, only the ressource end point.
Because I think he entity could be modified by batch or other services.

We should only forbid the client part and the endpoint (so only the ressource here)

What do you think ?

murdos:
> 
> 
> I think the service part should not be concerned by this, only the ressource end point.
> Because I think he entity could be modified by batch or other services.
> 
> We should only forbid the client part and the endpoint (so only the ressource here)
> 
> What do you think ?

I removed service code because it became dead code (and thus not covered by tests), and because in my own use case, data will be loaded with liquibase so I don't need it too. 
But I've no strong opinion here, if you think we should keep it, I can just revert the 2nd commit.

pascalgrimaud:
Yes please, I'd like to keep service to be able to create / update / delete the entity.
My real idea behind this is to use the read only mode with Kafka:
- no API
- create a consumer instead of resource, and this consumer will use the service to update the entity

PR #10567, opened 2019-10-06
update to swagger ui 3

### LINKED ISSUE #7966: Swagger UI latest version support (3.17.3)

Swagger UI very old version (2.2.10) is used which does not provide the support for Open API

Open API support

ecostanzi:
Fix #7966

Update swagger ui to version 3. 

-  still uses the iframe
- i added a swagger customizer for JWT to configure public and private endpoints. This is the only way I found to set the authentication token in the UI. 
- it works with oauth2 because the JSESSIONID and uaa auth tokens are passed automatically via cookie when performing requests

Related PRs:
https://github.com/jhipster/jhipster/pull/425
https://github.com/jhipster/generator-jhipster/pull/10570

If this is ok also the jhipste registry should be updated.


generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -92,9 +92,8 @@ module.exports = (options) => ({
             }
         }),
         new CopyWebpackPlugin([
-            { from: './node_modules/swagger-ui/dist/css', to: 'swagger-ui/dist/css' },
-            { from: './node_modules/swagger-ui/dist/lib', to: 'swagger-ui/dist/lib' },
-            { from: './node_modules/swagger-ui/dist/swagger-ui.min.js', to: 'swagger-ui/dist/swagger-ui.min.js' },
+            { from: './node_modules/swagger-ui-dist/*.{js,css,html,png}', to: 'swagger-ui', flatten: true, ignore: 'index.html'},
+            { from: './node_modules/axios/dist/axios.min.js', to: 'swagger-ui'},
```
ecostanzi: axios is present in the node_modules directory for both angular and react, but it is explicitly declared only in the react app. Should I add an explicit dependency in the package.json of the angular app?

ecostanzi:
@Hawkurane sorry I've just noticed you were also working on this. Let me know if your solution fits this one and if something can be done to improve for the UI.

generators/client/templates/react/src/main/webapp/swagger-ui/index.html.ejs
```diff
@@ -17,190 +17,65 @@
...
+                withCredentials: true,
+                presets: [
+                    SwaggerUIBundle.presets.apis,
+                    SwaggerUIStandalonePreset
+                ],
+                plugins: [
+                    SwaggerUIBundle.plugins.DownloadUrl
+                ],
+                <%_ if (authenticationType === 'jwt') { _%>
+                onComplete: function () {
```
ecostanzi: it seems to me that jwt is the only case where the token has to be passed via js to the UI. In case of oauth and uaa the authentication tokens and session info are all in the cookies. am I right?

Hawkurane:
Hey @ecostanzi This indeed goes along the lines of what I'm doing, except that instead of using swagger-ui npm packages, I'm using standalone js files :)
I have a working version for react and jwt but haven't done oauth2.0 (and uaa), and angular yet

generators/server/files.js
```diff
@@ -374,6 +374,10 @@ const serverFiles = {
                 {
                     file: 'package/security/jwt/JWTConfigurer.java',
                     renameTo: generator => `${generator.javaDir}security/jwt/JWTConfigurer.java`
+                },
+                {
+                    file: 'package/config/apidoc/SwaggerSecurityCustomizer.java',
+                    renameTo: generator => `${generator.javaDir}config/apidoc/SwaggerSecurityCustomizer.java`
```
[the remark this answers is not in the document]
ecostanzi: @Hawkurane thanks for the review. The reason why I didn't modify the jhipster-frameworks is because the security configuration depends on what the user chooses during generation. 

- a monolith has some private endpoints and some public endpoints
- a microservice has only private endpoints by default (all protected)
- the security definitions in the yml file when using jwt (Bearer) are different from the security definition we should define when using oauth or UAA

For what I know jhipster-framework should not be affected by what the user chooses during generation, and the only way to "dynamically" configure beans in the jhipster frameworks is to use spring conditional annotations. If we find how to use those annotations, we could think about moving SpringFox SecurityConfiguration, SecuritySchemes, and SecurityContext in the jhipster-framework, but we would miss the public/private endpoint part. 

Let me know what you think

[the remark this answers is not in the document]
Hawkurane: I think these are very valid points to consider. What about customizing the docket in 2 layers aka:

- First customization via jhipster/jhipster 
- Second one in a generated file (with SecurityConfiguration, SecuritySchemes and SecurityContext) for swagger ui v3, if we can keep this file I suggest we change the name for better clairty (e.g. SwaggerSecurityCustomizer, ...)

Another approach would migrate the swagger customizer completely from the jhipster frameworks to the generation as separating the two classes that essentially do the same thing might be confusing for the dev experience.

I am unsure as what could be considered as best practice so I am looking out for your response

zhx828:
@Hawkurane @ecostanzi thanks a lot for your work. Is there any updates on this PR? Github started to mark the swagger-ui 2.2.10 high severity. Nothing urgent, just wonder what the status is. Thanks again.

Hawkurane:
Hello, the work is mostly done. What is left is pretty much @ecostanzi choosing between what he thinks is best and tagging his PR as ready-to-merge :)

generators/server/files.js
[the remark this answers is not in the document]
ecostanzi: I'd keep it like it is now and see if some suggestions come out from the reviewers.

ecostanzi:
This is ready for merge I think. I resolved all the comments. The current configuration works and a code review can solve many doubts we had :)

zhx828:
@Hawkurane @ecostanzi thank you both!

@pascalgrimaud @cbornet @DanielFran since you've been involved in the issue, could some of you review the pr? Thanks in advance.

generators/server/templates/src/main/java/package/config/apidoc/SwaggerSecurityCustomizer.java.ejs
```diff
@@ -0,0 +1,97 @@
...
+
+import static com.google.common.collect.Lists.newArrayList;
+import static springfox.documentation.builders.PathSelectors.regex;
+
+@Component
+public class SwaggerSecurityCustomizer implements SwaggerCustomizer {
+<%_ if (applicationType === 'monolith' || applicationType === 'gateway') { _%>
+
+    private static final String[] PUBLIC_ENDPOINTS = {
+        "/api/account/change-password",
```
pascalgrimaud: I think it should be in application.yml instead of hard coded here

pascalgrimaud:
Just 1 comment from me.
Then, @ecostanzi : can you resolve the conflict plz ?

generators/server/templates/src/main/java/package/config/apidoc/SwaggerSecurityCustomizer.java.ejs
ecostanzi: I thought the same at the beginning. A the end I hard coded these paths because they are also hardcoded in the `SecurityConfiguration.java`. I was thinking "this isn't something that the user wants to configure".  BTW if you confirm I'll move them in the Swagger section of `JhipsterProperties.java`

pascalgrimaud: Ok then, so let's keep like this, and we'll improve it later if needed @ecostanzi !

jdubois:
I've added a bug bounty here, and I will try to help.
This is important as this issue a vulnerability warning on GitHub when creating a new project (with dependabot doing a PR to try to solve it)

pascalgrimaud:
@jdubois : there is already a bounty on the ticket, here it's the PR

jdubois:
Then I'm really tired!!! Thanks @pascalgrimaud !
Let me increase the bug bounty on the ticket to $300, then :-)

pascalgrimaud:
agree for increasing the bounty
I already tested JWT, it works well. 
I started session, but not finished
Then, I'll test:
- oauth2
- microservice jwt
- microservice oauth2

jdubois:
@pascalgrimaud shouldn't we remove "session" at some point? It adds a lot of maintenance, I don't think people use this anymore at all. We should work on the stats for this.

pascalgrimaud:
I don't the stat for session but if it's not used, let's remove it for v7

generators/server/templates/src/main/java/package/config/apidoc/SwaggerSecurityCustomizer.java.ejs
```diff
@@ -0,0 +1,96 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static springfox.documentation.builders.PathSelectors.regex;
+
+@Component
+public class SwaggerSecurityCustomizer implements SwaggerCustomizer {
+<%_ if (applicationType === 'monolith' || applicationType === 'gateway') { _%>
+
+    private static final String[] PUBLIC_ENDPOINTS = {
```
hdurix: It could be good to have the same list in SecurityConfiguration and here. I just tried to move this list in SecurityConfiguration and that's pretty nice:

Before:
```java
.antMatchers("/api/authenticate").permitAll()
.antMatchers("/api/register").permitAll()
.antMatchers("/api/activate").permitAll()
.antMatchers("/api/account/reset-password/init").permitAll()
.antMatchers("/api/account/reset-password/finish").permitAll()
```

After:
```java
.antMatchers(PUBLIC_ENDPOINTS).permitAll()
```

Then you could reuse this list here:
```java
.securityContexts(newArrayList(new SecurityContext(bearerAuth(), not(SecurityConfiguration.PUBLIC_ENDPOINTS))));
```

What do you think?

pascalgrimaud:
I try to resolve merge conflicts but something is broken, I don't understand what

pascalgrimaud:
@ecostanzi : after the upgrade of some libraries in current master, this PR needs to be updated, otherwise, it's broken when packaging.Sorry for that, but I need your help here to finish it and merge into master

ecostanzi:
No problem, I'll check it out

pascalgrimaud:
@ecostanzi : thanks a lot, I wanted to test and merge it this afternoon ! We need this PR for the next release !

generators/server/templates/src/main/java/package/config/apidoc/SwaggerSecurityCustomizer.java.ejs
ecostanzi: Looks good to me. I was also looking for a way to get these endpoints programmatically. I wanted to avoid modifying the security configuration. but I couldn't find a solution.

ecostanzi:
I'm sorry but I cannot look into it today :/ tomorrow probably

pascalgrimaud:
@ecostanzi : no worry, I'd like to do a release before Chrismast, so probably in 1 week !

generators/client/templates/angular/webpack/webpack.common.js.ejs
```diff
@@ -92,9 +92,8 @@ module.exports = (options) => ({
             }
         }),
         new CopyWebpackPlugin([
-            { from: './node_modules/swagger-ui/dist/css', to: 'swagger-ui/dist/css' },
-            { from: './node_modules/swagger-ui/dist/lib', to: 'swagger-ui/dist/lib' },
-            { from: './node_modules/swagger-ui/dist/swagger-ui.min.js', to: 'swagger-ui/dist/swagger-ui.min.js' },
+            { from: './node_modules/swagger-ui-dist/*.{js,css,html,png}', to: 'swagger-ui', flatten: true, ignore: 'index.html'},
```
pascalgrimaud: the ignore should be: `ignore: ['index.html']`

generators/client/templates/react/webpack/webpack.common.js.ejs
```diff
@@ -139,9 +139,8 @@ module.exports = options => ({
     }),
     new ForkTsCheckerWebpackPlugin({ eslint: true }),
     new CopyWebpackPlugin([
-      { from: './node_modules/swagger-ui/dist/css', to: 'swagger-ui/dist/css' },
-      { from: './node_modules/swagger-ui/dist/lib', to: 'swagger-ui/dist/lib' },
-      { from: './node_modules/swagger-ui/dist/swagger-ui.min.js', to: 'swagger-ui/dist/swagger-ui.min.js' },
+      { from: './node_modules/swagger-ui-dist/*.{js,css,html,png}', to: 'swagger-ui', flatten: true, ignore: 'index.html'},
```
pascalgrimaud: the ignore should be: `ignore: ['index.html']`

generators/client/templates/react/src/main/webapp/swagger-ui/index.html.ejs
```diff
@@ -17,194 +17,70 @@
...
+    window.onload = function () {
+        var urls = [];
+        axios.get("/swagger-resources").then(function (response) {
+            response.data.forEach(function (resource) {
+                urls.push({"name": resource.name, "url": resource.location});
+            });
 
-                function addApiKeyAuthorization() {
-                <%_ if (authenticationType === 'session' || authenticationType === 'uaa') { _%>
-                    var apiKeyAuth = new SwaggerClient.ApiKeyAuthorization("X-XSRF-TOKEN", getCSRF(), "header");
```
pascalgrimaud: how do you handle this part now ?
Just tested your PR and it doesn't work for OAuth2 when I want to POST an entity

ecostanzi: you are right, I forgot to restore that part sorry. I fixed it now.

pascalgrimaud: thanks, just tell me if it's ready, so I can do some tests to be sure, before merging this

ecostanzi: Yes it should work now you can test it

pascalgrimaud:
@ecostanzi, I just did some tests, here my results :
- monolith jwt : ok
- monolith session :
- monolith oauth2 : ok
- gateway : ok
- gateway -> micro : ok
- gateway -> micro with old version of swagger : KO
- registry (old version of swagger) -> gateway : KO
- registry (old version of swagger) -> micro : KO
- registry (old version of swagger) -> micro : ok

That means it seems not to be compatible between swagger 2 and swagger 3.
I thought this code `if(ui.specSelectors.isSwagger2()) {` could help but it's not the case.

The best solution would be:
- make this PR compatible with swagger2
- once it's done, upgrade the Registry with Swagger3 too

If it can't be achieve, I think we need to wait the new release v7 as it is considered as a breaking change for our users.

@jhipster/developers : what do you think ? Can somebody help here too plz ?

ecostanzi:
@pascalgrimaud I suspect that introducing the security definitions in the springfox yml caused most of the incompabilbility problems.

> gateway -> micro with old version of swagger : KO

Is it failing because the gateway does not send the token down to the microservice or for some other reason?

> registry (old version of swagger) -> gateway : KO
 registry (old version of swagger) -> micro : KO

Could you please tell what's causing the failure?

pascalgrimaud:
1) yes, you're right, the gateway doesn't use token so no security and we got a 401

2) about the new JHipster Registry v6.0.1, which is using Swagger 2, what I saw is, the Registry tried to get some resources (images or something else) in swagger-dist/, then get a 404 and the token is cleared.
Maybe it's a problem on the Registry side and it's not related to this PR. I simply wanted to share the results with you

cc @juliensadaoui for my last comment

ecostanzi:
> yes, you're right, the gateway doesn't use token so no security and we got a 401

The `SwaggerSecurityCustomizer` is causing this problem. I admit I didn't consider the compatibility.

For version 6, we can remove the customizer and send the token for every api call via requestInterceptor. This shouldn't be done for what we named "public endpoints" but that's how the current version works.

Then, in version 7, we can restore the `SwaggerSecurityCustomizer` and have the spec exposed properly with all the security schemes.

What do you think?

pascalgrimaud:
@ecostanzi : ok then, if it can solve the issue about compatibility !
And later, with v7, we can add back your customization as it looks better, so agree

ecostanzi:
> ok then, if it can solve the issue about compatibility 

Ok I pushed yesterday, now the JWT token is sent for every API call

DanielFran:
@ecostanzi can you please try using latest version of swagger-ui 3.24.x?

cbornet:
It would also be very nice to document security for session (httpbasic + CSRF) and OIDC in the customizer and somehow deactivate login/logout in the app's swagger-ui for these auth modes. That way you're auto-logged-in when using the app's swagger-ui and you also have correct definitions when using an external swagger-ui.
I'm also concerned by the future of Springfox and would like to give [Springdoc](https://github.com/springdoc/springdoc-openapi) a try.
All of this is less urgent than this PR of course.

generators/server/templates/src/main/java/package/config/apidoc/SwaggerSecurityCustomizer.java.ejs
ecostanzi: I had to remove the customizer to avoid introducing breaking changes. We will reconsider this for v7

pascalgrimaud:
@ecostanzi : thanks for updating the PR, I'll try to do some tests tomorrow

pascalgrimaud:
Finally, just spend my evening to test and it works well. So long to test !
I saw some issue but it comes from the Registry - cc @juliensadaoui 

So I'm merging this.
Thanks a lot @ecostanzi

jdubois:
Congrats, this is so awesome! I really wanted this one badly!

ecostanzi:
Good! Thank you for your time and patience @pascalgrimaud

juliensadaoui:
@pascalgrimaud I update swagger-ui to version 3 on the registry the next weekend

ecostanzi:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/12572

pascalgrimaud:
just approuved, thanks @ecostanzi !

micobarac:
How do I run Swagger v. 3 in jHipster v. 6.6.0 app? I can't find any info on this in the official documentation.

pascalgrimaud:
Swagger v3 replaced Swagger v2, you can access to it at `/admin/docs`, 
Nothing has changed, only the UI

micobarac:
I used `no client` option when generating an app. `/admin/docs` is unavailable.

cbornet:
Then this PR doesn't change anything for you. It's a front-end related one.

PR #10571, opened 2019-10-08
Rely on yeoman-environment to detect external modules (including blueprints)

murdos:
This is an attempt to solve issues with detecting/loading external Yeoman generators: https://github.com/jhipster/jhipster-kotlin/issues/179, https://github.com/jhipster/generator-jhipster/issues/10223, https://github.com/jhipster/generator-jhipster/issues/10223, ...

The idea is to rely on yeoman-environment `lookup()` feature that will detect available generators on your computer for us.
This also simplifies JHipster code and logic, since we don't need to make multiple attempts to find generator (lookup in local `node_modules`, try to do a `require()`, and in last resort use the `yo` command that internally uses yeoman-environment lookup())

### DIFF
--- a/cli/utils.js
+++ b/cli/utils.js
@@ -190,17 +190,8 @@ const done = errorMsg => {
 
 const createYeomanEnv = () => {
     const env = yeoman.createEnv();
-    /* Register yeoman generators */
-    Object.keys(SUB_GENERATORS)
-        .filter(command => !SUB_GENERATORS[command].cliOnly)
-        .forEach(generator => {
-            if (SUB_GENERATORS[generator].blueprint) {
-                /* eslint-disable prettier/prettier */
-                env.register(require.resolve(`${SUB_GENERATORS[generator].blueprint}/generators/${generator}`), `${CLI_NAME}:${generator}`);
-            } else {
-                env.register(require.resolve(`../generators/${generator}`), `${CLI_NAME}:${generator}`);
-            }
-        });
+    // Lookup for all available generators (blueprints, modules, ...)
+    env.lookup();
     return env;
 };
 
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -769,7 +769,7 @@ module.exports = class extends Generator {
         if (blueprint) {
             blueprint = jhipsterUtils.normalizeBlueprintName(blueprint);
             if (options.skipChecks === undefined || !options.skipChecks) {
-                this.checkBlueprint(blueprint, subGen);
+                this.checkBlueprint(blueprint);
             }
             try {
                 const finalOptions = {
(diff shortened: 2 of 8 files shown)


cli/utils.js
```diff
@@ -190,17 +190,8 @@ const done = errorMsg => {
...
-        .forEach(generator => {
-            if (SUB_GENERATORS[generator].blueprint) {
-                /* eslint-disable prettier/prettier */
-                env.register(require.resolve(`${SUB_GENERATORS[generator].blueprint}/generators/${generator}`), `${CLI_NAME}:${generator}`);
-            } else {
-                env.register(require.resolve(`../generators/${generator}`), `${CLI_NAME}:${generator}`);
-            }
-        });
+    // Lookup for all available generators (blueprints, modules, ...)
+    env.lookup();
```
deepu105: How does it know to register sub gens from blueprints? I wasn't aware of this API, is it new?

generators/generator-base-private.js
```diff
@@ -795,55 +795,47 @@ module.exports = class extends Generator {
...
     /**
      * Check if the generator specified as blueprint is installed.
      * @param {string} blueprint - generator name
      */
-    checkBlueprint(blueprint, subGen = '') {
+    checkBlueprint(blueprint) {
         if (blueprint === 'generator-jhipster') {
             this.error(`You cannot use ${chalk.yellow(blueprint)} as the blueprint.`);
         }
-        const done = this.async();
```
deepu105: This method is called inside a yeoman generator process, so its better to call done in order to avoid the process hanging, unless `this.env.getGeneratorNames()` doesn't do any async stuff

generators/generator-base-private.js
```diff
@@ -795,55 +795,47 @@ module.exports = class extends Generator {
      * @return {string} version - retrieved version or empty string if not found
      */
     findBlueprintVersion(blueprintPkgName) {
-        let packageJsonPath = path.join(process.cwd(), 'node_modules', blueprintPkgName, 'package.json');
-        try {
-            if (!fs.existsSync(packageJsonPath)) {
-                this.debug('using global module as local version could not be found in node_modules');
-                packageJsonPath = path.join(blueprintPkgName, 'package.json');
+        const blueprintGeneratorName = blueprintPkgName.replace('generator-', '');
```
deepu105: This specific replacement is done multiple times, may be lets have a util method as below to do it
```
const normalizeGeneratorName = name => name.replace('generator-', '');
```

generators/generator-base.js
```diff
@@ -1108,22 +1108,9 @@ module.exports = class extends PrivateBase {
...
-            this.debug('ERROR:', err);
-            const generatorName = npmPackageName.replace('generator-', '');
-            const generatorCallback = `${generatorName}:${subGen}`;
-            // Fallback for legacy modules
-            this.debug('Running yeoman legacy compose with options: ', generatorCallback, options);
-            this.composeWith(generatorCallback, options);
-        }
+        const generatorName = npmPackageName.replace('generator-', '');
+        const generatorCallback = `${generatorName}:${subGen}`;
+        this.composeWith(generatorCallback, options);
```
deepu105: So this now always uses the legacy wy of calling something in compose right. I remember the docs saying to use the `this.composeWith(require.resolve(generatorTocall), options)` way

mshima:
I've done a blueprint that tests situations like this.
(global|local) jhipster with (global|local) blueprints.

https://github.com/mshima/generator-jhipster-test

cli/utils.js
murdos: I don't think it's new, this is what yeoman uses with the yo command.
It scans local and global node_modules for generators, and register them and their subgenerators. If a generator is available both at local and global level, local one has priority.

generators/generator-base-private.js
murdos: `this.env.getGeneratorNames()` don't do any async stuff: https://github.com/yeoman/environment/blob/da6411e978b833d4cbc710cb2b3f3dda965f66aa/lib/environment.js#L305 

In fact the async stuff happens during `env.lookup();`, then everything is already in the (in memory) store.

generators/generator-base-private.js
murdos: Yeah, thanks. 
It belongs to the things I need to finish (with fixing upgrade unit test, merging app-blueprint.spec.js and scoped-blueprint.spec.js and fixing conflicts with https://github.com/jhipster/generator-jhipster/pull/10541), that's why it's still a draft. 

But I opened it in order to get some feedback, to know if I should continue 😉

generators/generator-base.js
murdos: I was not aware of that.
https://yeoman.io/authoring/composability.html indeed says `Even though it is not an encouraged practice, you can also pass a generator namespace to composeWith.`
But the jsdoc doesn't warn about that particular usage: https://github.com/yeoman/generator/blob/7b05f030622e05c9fa62b482f8ebe4a706c469ba/lib/index.js#L512

The goal is indeed to avoid `require.resolve()` that doesn't always work.

murdos:
> 
> 
> I've done a blueprint that tests situations like this.
> (global|local) jhipster with (global|local) blueprints.
> 
> https://github.com/mshima/generator-jhipster-test

Interesting. How do you use it for testing?

mshima:
> > I've done a blueprint that tests situations like this.
> > (global|local) jhipster with (global|local) blueprints.
> > https://github.com/mshima/generator-jhipster-test
> 
> Interesting. How do you use it for testing?

Clone it and add the desired target to azure pipeline (set your branch).

Example:
https://github.com/mshima/generator-jhipster-test/commit/40b14a7a218be6b5c8979526112d842404cecdf9

https://github.com/jhipster/generator-jhipster/pull/10555 is based on improvements from this "blueprint".

mshima:
Forgot that you have to add your repository to azure.
Or you can create PR to my repository to test.

mshima:
https://github.com/mshima/generator-jhipster-test/tree/detect-blueprint-installed-globally
https://dev.azure.com/marceloshima/generator-jhipster-test/_build/results?buildId=323

cli/utils.js
mshima: I think lookup should be called only if blueprints is passed.
Otherwise it's not used at all.

generators/generator-base.js
mshima: composeExternalModule is used by external modules, removing 'generator-' from modules packageName probably will break modules support.
Moving this new logic to composeBlueprint and deprecating modules support (callHooks, composeExternalModule) seems the way to go.

cli/utils.js
murdos: This is also necessary for modules hooks

generators/generator-base.js
murdos: I don't understand how it would break modules support. 
Removing 'generator-' is just the way to transform a npm package name to a Yeoman namespace. There's no differences between blueprints and modules in that perspective.

And I'm against deprecating modules support, they have their value, and always using blueprint is like using a sledgehammer to crack a nut.

cli/utils.js
mshima: Calling env.lookup() first time composeExternalModules/composeBlueprint will do the job?

generators/generator-base.js
mshima: Does modules without 'generator-' prefix works?

murdos: Modules are Yeoman generators, so they should use 'generator-' prefix. See Basic rules for a JHipster module in https://www.jhipster.tech/modules/creating-a-module/

cli/utils.js
murdos: No, this is required for checks and version retrieval.

mshima:
@murdos https://github.com/yeoman/environment/pull/124 will simplify this PR.

generators/generator-base-private.js
```diff
@@ -795,55 +795,47 @@ module.exports = class extends Generator {
...
+                }
             }
-            // eslint-disable-next-line global-require,import/no-dynamic-require
-            const packagejs = require(packageJsonPath);
-            return packagejs.version;
-        } catch (err) {
-            this.debug('ERROR:', err);
-            this.warning(`Could not retrieve version of blueprint '${blueprintPkgName}'`);
-            return '';
         }
```
mshima: With yeoman-environment v2.6.0:
const packageJsonPath = path.join(this.env. getPackagePath(blueprintGeneratorName), 'package.json');
 if (fs.existsSync(packageJsonPath)) {
    // eslint-disable-next-line global-require,import/no-dynamic-require
    const blueprintPackagejs = require(packageJsonPath);
    return blueprintPackagejs.version;
}

mshima:
Closing in favor of https://github.com/jhipster/generator-jhipster/pull/11313

PR #10595, opened 2019-10-10
fix metrics for services in react

ctamisier:
The following metric section isn't displayed for React:

image


pascalgrimaud:
need this for 6.4.0

PR #10606, opened 2019-10-12
feat (ci): Support Github CI in the ci-cd sub generator

### LINKED ISSUE #10599: Github Actions in CI sub generator
It would be cool to have Github Actions in the CI sub-generator, as it will be easy to integrate it for generated project in GitHub.

@vishal423 : do you want to took this ticket, as you already do an amazing work in our CI ?

vishal423:
* Closes #10599 
* Followed Azure CI template (without external integrations) with Java 8 and Java 11 build support.
* Tested with `yarn + gradle + angular` combination @ https://github.com/vishal423/sample-jhipster-ng-gradle-yarn-app/commit/64d1fd3759b33d53f255566421a44d1b31845603/checks?check_suite_id=262455522
* Tested with `npm + maven + react` combination @ https://github.com/vishal423/sample-jhipster-react-mvn-npm-app/commit/41dfd24c8ad079409008635594fd8dff6dd9f26e/checks?check_suite_id=262455872

---

### DIFF
--- a/generators/ci-cd/templates/github-ci.yml.ejs
+++ b/generators/ci-cd/templates/github-ci.yml.ejs
@@ -1,0 +1,69 @@
+<%#
+ Copyright 2013-2019 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see https://www.jhipster.tech/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+name: Application CI
+on: [push, pull_request]
+jobs:
+    applications:
+        name: <%= baseName %> test suite
+        runs-on: ${{ matrix.os }}
+        timeout-minutes: 40
+        strategy:
+            fail-fast: false
+            matrix:
+                node_version: [<%= NODE_VERSION %>]
+                os: [ubuntu-latest]
+                java_version: ['8.x', '11.x']
+        env:
+            NODE_VERSION: ${{ matrix.node_version }}
+            SPRING_OUTPUT_ANSI_ENABLED: NEVER
+            SPRING_JPA_SHOW_SQL: false
+            JHI_DISABLE_WEBPACK_LOGS: true
+            NG_CLI_ANALYTICS: false
+        steps:
+            - uses: actions/checkout@v1
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: ${{ matrix.node_version }}
+            - uses: actions/setup-java@v1
+              with:
+                  java-version: ${{ matrix.java_version }}
+            - name: Stop MySQL server
+              run: sudo /etc/init.d/mysql stop
+            - name: Install node.js packages
+              run: <%= clientPackageManager %> install
+            - name: Run backend test
+              <%_ if (buildTool === 'maven') { _%>
+              run: |
+                 chmod +x mvnw
+                 ./mvnw -ntp clean verify <% if (!skipClient) { %>-P-webpack<% } %>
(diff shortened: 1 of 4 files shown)


generators/ci-cd/templates/github-ci.yml.ejs
```diff
@@ -0,0 +1,69 @@
...
+    applications:
+        name: <%= baseName %> test suite
+        runs-on: ${{ matrix.os }}
+        timeout-minutes: 40
+        strategy:
+            fail-fast: false
+            matrix:
+                node_version: [<%= NODE_VERSION %>]
+                os: [ubuntu-latest]
+                java_version: ['8.x', '11.x']
```
pascalgrimaud: I'd prefer to keep only JDK 11, rather than support both version, as we'll probably remove JDK 8 soon

generators/ci-cd/templates/github-ci.yml.ejs
```diff
@@ -0,0 +1,69 @@
...
+        env:
+            NODE_VERSION: ${{ matrix.node_version }}
+            SPRING_OUTPUT_ANSI_ENABLED: NEVER
```
pascalgrimaud: I didn't have time to test Github action yet... so I don't know if we can some beautiful color here :)

generators/ci-cd/templates/github-ci.yml.ejs
```diff
@@ -0,0 +1,69 @@
...
+            SPRING_JPA_SHOW_SQL: false
+            JHI_DISABLE_WEBPACK_LOGS: true
+            NG_CLI_ANALYTICS: false
+        steps:
+            - uses: actions/checkout@v1
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: ${{ matrix.node_version }}
```
pascalgrimaud: no need, as there is only 1 fixed version of node

generators/ci-cd/templates/github-ci.yml.ejs
```diff
@@ -0,0 +1,69 @@
...
+            - uses: actions/setup-java@v1
+              with:
+                  java-version: ${{ matrix.java_version }}
```
pascalgrimaud: no need if we keep only java11

generators/ci-cd/templates/github-ci.yml.ejs
```diff
@@ -0,0 +1,69 @@
...
+                 chmod +x gradlew
+                 ./gradlew clean test integrationTest <% if (!skipClient) { %>-x webpack<% } %>
+              <%_ } _%>
+            <%_ if (!skipClient) { _%>
+            - name: Run frontend test
+              run: <%= clientPackageManager %> run <%= frontTestCommand %>
+            <%_ } _%>
+            - name: Package application
+              <%_ if (buildTool === 'maven') { _%>
+              run: ./mvnw -ntp package -Pprod -DskipTests -DskipIT
```
pascalgrimaud: are you sure about skipIT ?

pascalgrimaud:
Great job @vishal423 !
Very few minor comments from me

generators/ci-cd/templates/github-ci.yml.ejs
murdos: Indeed, it should be `-DskipITs`. 
Anyway, according to https://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html, `-DskipTests` should prevent all tests to be run and thus is sufficient

generators/ci-cd/templates/github-ci.yml.ejs
vishal423: I have no hard preference here, however, I think it would be good to support until `v7`.

generators/ci-cd/templates/github-ci.yml.ejs
vishal423: I think they do support, let me try this out

generators/ci-cd/templates/github-ci.yml.ejs
vishal423: yeah that needs correction. On further thought, I will remove it as `verify` goal comes after `package` and we anyway don't want to run tests here.

generators/ci-cd/templates/github-ci.yml.ejs
vishal423: I have tried both `DETECT` and `ALWAYS` and don't see any visual difference. For moment, I have set it to `DETECT`

Issue #10611, opened 2019-10-13
Add support to peer blueprints


mshima:
##### **Overview of the feature request**

Add support to create blueprints that have a peer dependency on generator-jhipster.

##### **Motivation for or Use Case**

Currently when we run generator-jhipster with blueprints, each blueprint will bring it's own generator-jhipster.
Those generator-jhipster version can be different and can cause unexpected results.

Current blueprints implementations will require generator-jhipster directly, that's why the blueprint needs to have a dependency on generator-jhipster.

This proposal is that instead of a hard require ex: 
```
const AppGenerator = require('generator-jhipter/generators/app)
```

We can use yeoman in our favor:
```
const Environment = require('yeoman-environment');
/* eslint-disable-next-line import/no-dynamic-require */
const AppGenerator = require(Environment.lookupGenerator('jhipster:app'));
```

Without a direct require of generator-jhipster modules, it's safe to move generator-jhispter to peerDependency.

On the blueprint side, we may have to change ci to install peer dependency. Ex for travis
```
 before_install:
     - npm install -g npm
     - npm install -g npm-install-peers
 install:
     - npm install
     - npm-install-peers
 script:
     - npm test
```

##### **Related issues or PR**

PR to yeoman-environment to add Environment.lookupGenerator:
https://github.com/yeoman/environment/pull/119

Verification of the running jhipster version with blueprint's one.
This PR make sure that the running generator-jhipster is the same that the blueprint brought.
https://github.com/jhipster/generator-jhipster/issues/10415

PR to use yeoman's namespaces and let yeoman do the hard work instead of loading blueprints ourselves.
https://github.com/jhipster/generator-jhipster/pull/10571

Problem on vuejs blueprint side:
https://github.com/jhipster/jhipster-vuejs/issues/469

-   [x] Checking this box is mandatory (this is just to show you read everything)

mshima:
https://github.com/yeoman/environment/pull/119 has been merged.
And yeoman-environment 2.5.0 has been released.
JHipster doesn’t need to depend on this version only the blueprint.

PR #10612, opened 2019-10-13
Add support to blueprint with peer dependency on generator-jhispter

mshima:
This PR adapts checkJHipsterBlueprintVersion to accept peerDependency on generator-jhipster.
https://github.com/jhipster/generator-jhipster/issues/10611

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -865,19 +865,32 @@ module.exports = class extends Generator {
      */
     checkJHipsterBlueprintVersion(blueprintPkgName) {
         const blueprintPackageJson = this.findBlueprintPackageJson(blueprintPkgName);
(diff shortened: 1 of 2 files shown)


generators/generator-base-private.js
```diff
@@ -871,6 +871,14 @@ module.exports = class extends Generator {
         }
         const mainGeneratorJhipsterVersion = packagejs.version;
         const blueprintJhipsterVersion = blueprintPackageJson.dependencies['generator-jhipster'];
+        if (!blueprintJhipsterVersion) {
+            // Check if generator-jhipster is on peerDependency
```
murdos: If generator-jhipster is on peerDependency, it most likely won't be declared as dependency, and thus this code will never be executed, since the function will return above (L870)

mshima: You are faster than me, I would have caught when implementing CI.
I got lost on blueprint, jhipster, yeoman tests.

Thanks

Issue #10631, opened 2019-10-16
Enable strict option for Typescript compiler in ng-jhipster


codecholeric:
##### **Overview of the feature request**

I wanted to ask, if you have ever considered to enable the `strict` compiler option for the Typescript compiler within the created Angular app.

##### **Motivation for or Use Case**

In my opinion one of the great values of Typescript is a stronger typing, providing documentation and more errors that can be caught at compile time instead of run time. The created Angular app has very unrestricted compiler options and uses many patterns closer to Javascript (many unspecified `any` types, unspecified return paths from if-statements, ...).
One major problem is, that it is easy to relax the compiler options for your projects, if you feel they are too restrictive. The other way around is quite hard. We want to enable strict compiler options, but there are a lot of errors within the generated JHipster Angular app then.
I would be willing to create a PR adjusting the templates to strict compiler options, but only if you are open to the change in the first place :wink: 
Otherwise I fear that we will get quite some merge conflicts in the future, when we'll try to upgrade JHipster.

-   [x] Checking this box is mandatory (this is just to show you read everything)

codecholeric:
Cool :smile: Can I support you in any way? I don't want to just make smart suggestions and then lean back, but I also don't want to slow you down, if you feel like you have to covered it anyway :wink: 
Is there any component where I could start to help you? Or do you want to learn from migrating the account module first?

kaidohallik:
I made PR for the account module: #10639.

Currently #10639 made the following in the account module:
* fixed all problems if `strict` is `true`
* fixed all problems after adding `TSLint` rule `"typedef": [true, "call-signature"]`

If this is OK (if we agree that this is the way to go) then interested people can do PR-s also in other modules/components.  
To avoid parallel work maybe it's good idea to leave comment to this ticket if someone started working with some new module/component cleaning.

#10639 turned off `ESLint` rule `@typescript-eslint/no-non-null-assertion`. If some other PR also needs this and #10639 is not merged yet then minor merge conflict can occur, but this is easily resolvable I think.

kaidohallik:
I made #10683 to resolve this for audits module.

pascalgrimaud:
ok @kaidohallik, forget my other comment. Found it :)

I'm putting a bounty, as it's a lot lot lot of work to achieve this

kaidohallik:
[...]

PR #10639, opened 2019-10-18
Use strict mode in account module in Angular

kaidohallik:
I switched `strict` to `true` in `tsconfig.json` and resolved all problems in account module.

Additionally I added `TSLint` rule `"typedef": [true, "call-signature"]` and resolved all problems in account module.

In the process of this cleaning if I found some unused code then I removed that and also renamed some variables to better names and made some minor fixes (like: error message was previously not shown in password reset initialization if network error occurred).

https://angular.io/guide/template-syntax#the-non-null-assertion-operator--- says:

> The non-null assertion operator, !, is optional with the exception that you must use it when you turn on strict null checks.

So if we turn on strict mode (which turns on also strict null checks) then we must turn off `ESLint` rule `@typescript-eslint/no-non-null-assertion`.  
In account module there are some places where I couldn't manage without this `!` operator and so I turned off `@typescript-eslint/no-non-null-assertion`.

If using `!` is not acceptable then seems that we can't set `strict` to `true` (or can someone offer solution to avoid `!` in account module?). Then we must find alternative solution like:
* turn on only some strict rules in `tsconfig.json`
* or instead of switching strict on in `tsconfig.json` look for equivalent `ESLint` rules and use `eslint disable` to disable rules in some code blocks

This PR resolves account module part for #10631


codecholeric:
Yes, I've noticed the need for `!` as well (since I've started to switch the TS config to strict in our project), in particular with the `FormBuilder` :frowning_face: 
If 3rd party components don't have strong enough types, there is not much one can do.
One other option is, to keep the check globally enabled and just suppress it in dedicated places, having a convention to comment, why this is okay. E.g.

```
// FormBuilder does not create a typesafe FormGroup, but we know that 'newPassword'
// is present from the way we constructed it
// eslint-disable @typescript-eslint/no-non-null-assertion
const newPassword = this.passwordForm.get(['newPassword'])!.value;
```

It's a little bit much bloat, but at least it's documented, why you can safely navigate this object.
Unfortunately I also don't know, how `FormBuilder` could make that type safe with the current Typescript language definition, since it would need something like

```
FormBuilder.group(controlOption: {k1: AbstractControlOption, k2: AbstractControlOption, ...}): 
    FormGroup & {k1: AbstractControl, k2: AbstractControl, ...}
```

(Does anyone know, if such a thing is possible with TS? Make the generic output of a function dependent on the structure of the generic input, but change the type of the properties?)

Could just create a wrapper for the `FormGroup`, that encapsulates the form navigation and unsafe operations.

kaidohallik:
In one intermediate version I wrote those `eslint-disable @typescript-eslint/no-non-null-assertion` comments but I didn't like it, too much bloat.

At some point I considered wrapper function but didn't implement that. Now if you are mentioning this too, I created wrapper function and restored `@typescript-eslint/no-non-null-assertion`. And `eslint-disable` in a single place only.

kaidohallik:
I forgot licence header from the new file. Now added that and force pushed.

generators/client/templates/angular/src/main/webapp/app/account/sessions/sessions.component.ts.ejs
```diff
@@ -21,41 +21,40 @@ import { Component, OnInit } from '@angular/core';
...
     selector: '<%= jhiPrefixDashed %>-sessions',
     templateUrl: './sessions.component.html'
 })
 export class SessionsComponent implements OnInit {
-
-    account: any;
-    error: string;
-    success: string;
-    sessions: Session[];
+    account: Account | undefined;
```
codecholeric: Could also write this shorter as `account?: Account`, couldn't you?

codecholeric:
I really like the additional types, feels like the changes make things cleaner :smiley: 
If you guys know, that this is the way to go, I would be happy to provide a PR for another module, e.g. admin module? Probably need `getFormControlUnchecked(..)` as well then, but I could just temporarily duplicate that and replace it, once this PR is merged...

kaidohallik:
I believe that it's better to make separate PR for every admin sub module. It simplifies review I believe.

While doing account module PR I actually started also with audits module. And now I'm finishing PR for audits module. So I would like to handle audits module cleaning.

@codecholeric For me it's OK if you take other admin modules for cleaning.

codecholeric:
Yes, makes sense! I'll just grab the next in the list then, i.e. `configuration`

kaidohallik:
> Yes, makes sense! I'll just grab the next in the list then, i.e. `configuration`

OK.

generators/client/templates/angular/src/main/webapp/app/account/sessions/sessions.component.ts.ejs
kaidohallik: @codecholeric Thanks for looking at that!  
`?:` seems to work here too, but as `?` means optional then it seems to me that `| undefined` fits better to context.

codecholeric: But what is the difference for you, between "it's optional", and "it can be `undefined`"? And where do you distinguish with `null` :wink: (that one always gave me a headache with JS)

kaidohallik: I changed `| undfined` to `| null = null` in all 3 places. I believe it's better now. @codecholeric thanks for pointing this out and helping to make code cleaner!

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
```diff
@@ -21,11 +21,11 @@
         <div class="col-md-4">
             <h1 jhiTranslate="reset.finish.title">Reset password</h1>
 
-            <div class="alert alert-danger" jhiTranslate="reset.finish.messages.keymissing" *ngIf="keyMissing">
+            <div class="alert alert-danger" jhiTranslate="reset.finish.messages.keymissing" *ngIf="keyMissing === true">
                 <strong>The password reset key is missing.</strong>
             </div>
 
-            <div class="alert alert-warning" *ngIf="!success && !keyMissing">
+            <div class="alert alert-warning" *ngIf="!success && keyMissing === false">
```
wmarques: Why this condition is changed ? Shouldn't we check that keyMissing is false or null ?

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
```diff
@@ -52,37 +50,38 @@ export class PasswordResetFinishComponent implements OnInit, AfterViewInit {
...
+            this.keyMissing = !this.key;
         });
-        this.keyMissing = !this.key;
     }
 
-    ngAfterViewInit() {
-        if (this.elementRef.nativeElement.querySelector('#password') != null) {
-          this.renderer.invokeElementMethod(this.elementRef.nativeElement.querySelector('#password'), 'focus', []);
+    ngAfterViewInit(): void {
+        if (this.elementRef.nativeElement.querySelector('#newPassword') != null) {
```
wmarques: I would prefer using a `ViewChild` with a TemplateRef instead of querying by id.. WDYT ? It's not in the scope of this PR so that's up to you if you want to do it :)

generators/client/templates/angular/src/main/webapp/app/account/settings/settings.component.ts.ejs
```diff
@@ -53,64 +52,53 @@ export class SettingsComponent implements OnInit {
...
-                if (settingsAccount.langKey !== current) {
-                    this.languageService.changeLanguage(settingsAccount.langKey);
+                if (this.account.langKey !== current) {
+                    this.languageService.changeLanguage(this.account.langKey);
                 }
             });
             <%_ } _%>
-        }, () => {
-            this.success = null;
-            this.error = 'ERROR';
```
wmarques: Shouldn't we also set success to false and error to true when an error occurs from the http call ?

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
```diff
@@ -44,7 +44,7 @@ const LAYOUT_ROUTES = [
                 <%_ if (!skipUserManagement) { _%>
                 {
                     path: 'account',
-                    loadChildren: () => import('./account/account.module').then(m => m.<%=angularXAppName%>AccountModule)
```
wmarques: Are you sure about this ? I think we added a prefix to handle some conflicts, maybe when generating entities... What if I generate an entity named `Account` ?

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
kaidohallik: Because this message can be seen only after route query parameters are resolved. So `keyMissing` had 3 values, `null` meant **not initialized**.  
As this code was confusing, I removed `keyMissing` and added `initialized` variable now.

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts.ejs
kaidohallik: Good idea, I made this change.

generators/client/templates/angular/src/main/webapp/app/account/settings/settings.component.ts.ejs
kaidohallik: Variable `error` was not used, so I removed this. Variable `success` is set to `false` at the beginning of this function. Errors are shown by `ErrorHandlerInterceptor`.  So this block is not needed.

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
kaidohallik: If trying to generate entity with name `Account` then entity sub generator throws error "ERROR! Error: The entity name cannot contain a Java or JHipster reserved keyword".

There is issue #10220 opened which contains also prefix removing and some prefixes are already removed. I put reference to this issue in previous commit but as this is in the second line then GitHub doesn't show this here (only first line of the commit message is shown).

kaidohallik:
Currently I feel that restoring `@typescript-eslint/no-non-null-assertion` by this commit: https://github.com/jhipster/generator-jhipster/pull/10639/commits/fe7a3a143e9f0bd74dd33709db389b3e4d290fe9 was not a right decision.

Angular documentation in https://angular.io/guide/template-syntax#the-non-null-assertion-operator--- says: **The non-null assertion operator, !, is optional with the exception that you must use it when you turn on strict null checks.**

I worked also enabling strict option in `audits` and `core` modules and I feel that if we go for strict option in Typescript then we must disable `@typescript-eslint/no-non-null-assertion` and use `!` if needed without providing additional code or we end with:
* we are improving code by resolving problems reported if strict option is turned on
* we make code worse by adding `eslint disable` lines (or if want to avoid that then add unnecessary if clauses or declare unnecessary variables or add unnecessary helper functions to avoid `eslint disable`)

@codecholeric @wmarques is this ok if I revert commit referenced above: https://github.com/jhipster/generator-jhipster/pull/10639/commits/fe7a3a143e9f0bd74dd33709db389b3e4d290fe9?

codecholeric:
@kaidohallik The mentioned part from the Angular docs, seems to refer specifically to templates to me. I.e. even with `no-non-null-assertion` it is possible to write things like

```
@Input()
iKnowItsThere!: string
```

Or within a template even `*ngIf="optionalProperty.iKnowItsThere.sub"`

At least for me that throws no compile error or eslint warning, even with the strict settings and eslint assertion on?

For things like the `FormGroup` I agree though, but in some way I also think that it is a limitation of Typescript (that it's impossible to define the correct type to access this in a type safe way and know it's not null).
Anyway, I don't have a super strong opinion about this, because I also see the downside of making code more bloated in places where you just know that this property is not optional, even though it's defined that way. However from a point of maintainability, this is of course a problem. Because tomorrow somebody might thing "thank God it's optional, I'm gonna put `null` there", and in another place someone put in `thing.optional!.sub`, because at the time this was never absent, instead of doing what should have been done, i.e. fix the type in the first place. If it is tedious, it might be an incentive to actually fix the types so it cannot be absent anymore. But I also know that sometimes you have no choice, e.g. if you use libraries. Then your only option really is to create a wrapper or similar to avoid this, like with `FormGroup`. And that is bloat of course.
Maybe you could share some examples, where you would want to use a non-`null` assertion?

kaidohallik:
I could simplify code in tracker service (PR #10786) so that currently no need to use this `eslint disable` in many places, so ATM I think that we should not revert https://github.com/jhipster/generator-jhipster/commit/fe7a3a143e9f0bd74dd33709db389b3e4d290fe9 - this `eslint` rule forced me to make tracker service code better, so this rule made work it should do

Currently I have 3 problematic places (in addition to `FormGroup`) after resolving `audits`, `core` and `tracker`:
* `datePipe` result can be null, in audits module:
    * https://github.com/jhipster/generator-jhipster/pull/10683/files#diff-56608689f1346dc198e6942fad4fbf71R76-R77
    * https://github.com/jhipster/generator-jhipster/pull/10683/files#diff-56608689f1346dc198e6942fad4fbf71R84-R85
* using class variable in callback function, typescript compiler thinks that class variable can be changed before callback function is executed, in `account.service` I currently added local function variable to suppress compiler error: https://github.com/jhipster/generator-jhipster/pull/10786/files#diff-c61338cd8464bf86b0a7f0faabf6214bR75-R76

wmarques:
@kaidohallik Sorry for late answer :( I agree that we should not have restored the ESLint rule. 
For me when strict mode is turned on, it's ok to use the non-null assertion operator when we are sure that the variable is non null... So i'm ok for reverting the suggested commit and turn this rule off.

kaidohallik:
I reverted @typescript-eslint/no-non-null-assertion restore commit.  
I believe it's better now. I didn't like that just for no null check rule I added unnecessary code, and even worse the new util file.  
I believe that people like seeing standard code as this is now
```
this.form!.get('field')
```
instead of JHipster util file specific
```
getFormControlUnchecked(this.form, 'field')
```

generators/client/templates/angular/src/main/webapp/app/account/sessions/sessions.component.ts.ejs
kaidohallik: @codecholeric, after resolving strict option in some other modules I think your suggestion to use `?:` makes sense, it bloats code as less as possible. So I changed my mind. I think that good approach is:
* if variable can initially be `undefined` and after assigning value to it, this variable is never set to `null`/`undefined` in the code then this variable can be defined as `myVariable?: MyType;`, for example:
```
account?: Account;
```
* if variable value can be "reset" in the code then use `myVariable: MyType | null = null;` syntax, for example:
```
account: Account | null = null;
```
and then in the code in the appropriate places assignment can be: `this.account = null;`

I added new commit to follow this path. @wmarques I hope it's OK for you.

Issue #10663, opened 2019-10-21
Duplicate blueprint configuration can be confusing and cause unexpected errors.


mshima:
##### **Overview of the issue**

Currently when setting a configuration from a blueprint generator, it will create a namespaced configuration duplicating all configurations.

https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/test/templates/ngx-blueprint/.yo-rc.json#L1-L6

- If a generator is not overridden by the blueprint, it will load the generator-jhipster namespace.

- If a generator is overridden by the blueprint, it will load the generator-jhipster-(blueprint) namespace.

To workaround this problem, there is getAllJhipsterConfig (https://github.com/jhipster/generator-jhipster/pull/7802/commits/ea50578c8db9af39bcc087502a8de805a62509c3) that will load every configuration and merge them, but the order is based on the blueprint load order.
https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/generators/utils.js#L423

Changing the .yo-rc.json in one place can not cause the expected result.

##### **Motivation for or Use Case**

Make the configuration cleaner and not confusing.

##### **Related issues**

PR https://github.com/jhipster/generator-https://github.com/jhipster/generator-jhipster/pull/7802

##### **Suggest a Fix**

Remove the duplicate configuration by:
1. Upgrade to yeoman-generator 4.1.0
2. Moving the blueprint configuration from generator.config to generator.blueprintConfig
3. Recreating generator.config with 'generator-jhipster' namespace.
4. Move the duplicate configurations from blueprintConfig to config.

##### **JHipster Version(s)**

master

##### **Browsers and Operating System**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

PR #10664, opened 2019-10-21
Unify blueprint configuration

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/10663

### DIFF
--- a/generators/generator-base-blueprint.js
+++ b/generators/generator-base-blueprint.js
@@ -25,6 +25,18 @@ const BaseGenerator = require('./generator-base');
  * The method signatures in public API should not be changed without a major version change
  */
 module.exports = class extends BaseGenerator {
+    constructor(args, opts) {
+        super(args, opts);
+
+        this.jhipsterConfig = this._getStorage('generator-jhipster');
+        this.fromBlueprint = this.rootGeneratorName() !== 'generator-jhipster';
+
+        if (this.fromBlueprint) {
+            this.blueprintConfig = this.config;
+            this.config = this.jhipsterConfig;
+        }
+    }
+
     // Public API method used by the getter and also by Blueprints
     _initializing() {
         return {};


generators/generator-base-blueprint.js
```diff
@@ -95,4 +106,14 @@ module.exports = class extends BaseGenerator {
         }
         return useBlueprints;
     }
+
+    removeConfigDuplicates() {
+        Object.keys(this.config.getAll()).forEach(key => {
```
sendilkumarn: For the blueprint generators the properties / configuration will remain inside the blueprint's key rather than inside the 'generator-jhipster` key. 

So iterating like this will not have an impact. Since there will not be any common property between them.

mshima: For client/server it works because app module doesn't have blueprint support.

I think that Kotlin is different.

I will try copying every config the first execution.

mshima: @sendilkumarn What about now?

generators/generator-base-blueprint.js
```diff
@@ -95,4 +106,17 @@ module.exports = class extends BaseGenerator {
         }
         return useBlueprints;
     }
+
+    migrateBlueprintConfig() {
```
sendilkumarn: I have mixed feelings about this. This will work but the configuration is heavily split now, any custom config in a blueprint will break things for the blueprint. I will strongly suggest to check this with various types of blueprints and validate that it works. 

Meanwhile, we will have #10737 to make sure we are not breaking anything. I feel #10737 is just an additional check but once this is validated we can remove things in the next major version.  WDYT?

sendilkumarn: And also what will happen when there are multiple blueprints?

mshima: This is a complicated issue, but the real problem of getAllJhipsterConfig is that
`jhipster --blueprints foo,bar`
can be different of
`jhipster --blueprints bar,foo`
The config of the last one will be used.

mshima: > And also what will happen when there are multiple blueprints?

The first run should settle the config.
If the blueprint want to keep the config separated, it should use this.blueprintConfig

mshima: Maybe unifying the config using blueprints parameter and config before creating the blueprint can help.

sendilkumarn: This is a tricky problem to solve. That is the reason I want to sort it out before we merge it. For now, AFAICT none of the blueprints fall into this weird case, but we should at least have a clear idea of what we should do.

generators/generator-base-blueprint.js
```diff
@@ -95,4 +106,17 @@ module.exports = class extends BaseGenerator {
...
+        if (!this.blueprintConfig.get('blueprintConfigOnly')) {
+            // Copy every config to jhipster config
+            this.config.set(this.blueprintConfig.getAll());
```
sendilkumarn: what will happen here, if there is a duplicate value?

mshima: Don't seems right, but It will be overridden.
Just like getAllJhipsterConfig =/

mshima: This is the main issue.
For this problem, we can revert last commit and you can use
`const configuration = {...this.getConfig.getAll(), ... this.blueprintConfig.getAll()}`
instead of getAllJhipsterConfig.
Remove the duplicates, but keep the specific ones.

sendilkumarn: In that case, I have touched upon only the things that are related to `Docker` and `k8s` which will need server values that will be present only in the blueprint configuration. But this changes globally for the blueprint. That is the main reason I want to have a proper solution here.

mshima: I didn't got your point here.
Do you have a blueprint and were it is used?

If it needs the configuration from blueprint, it should get by getAllJhipsterConfig or some way that the blueprint pass the config to the main generator.

sendilkumarn: Can you try testing with Kotlin and VueJS blueprint together?

mshima: Probably we should let the blueprint migrate their configs.
But create the means to create conflict free configs.

mshima: If we came up with a good migration strategy, then do it.

mshima: Removed the auto migration and if the blueprint wants a clean config pass opt.newConfig to the constructor.
WDYT?

sendilkumarn: @mshima this is so weird, what will happen when two server blueprints work together how does the configuration will result? How does blueprint config should look like? I think there should not be any configuration inside blueprint and all of them should live inside the main config. WDYT?

mshima: > what will happen when two server blueprints work together how does the configuration will result?

Depends on the order of the blueprint, but mainly the config will be duplicated (not guaranteed).

> How does blueprint config should look like?

Like this:
https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/test/templates/ngx-blueprint/.yo-rc.json

> I think there should not be any configuration inside blueprint and all of them should live inside the main config.

That last migration do that, but it only set a flag on blueprint config to don't migrate anymore.

With current version and the blueprint setting opts.blueprintConfig the new config will alway be generator-jhipster.
Unless the blueprint uses `this.blueprintConfig.set`, then it will be specific config.
But migration should be done by the blueprint.

At my blueprint I create some specific config, so I set to the blueprint config.

mshima: @sendilkumarn I suggest do the migration copying everything and if the blueprint is "config safe" it sets the opts.blueprintConfig=true.

This will be the inverse of current patch.

mshima: For example, at:
https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/test/templates/ngx-blueprint/.yo-rc.json

The config `promptValues` is a conflict.

https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/test/templates/ngx-blueprint/.yo-rc.json#L2-L5
will be used when:

- The generator is blueprinted
- getAllJhipsterConfig is used.


https://github.com/jhipster/generator-jhipster/blob/ae74d36c57f2aa90f680c0fa27fce67e1b11e81e/test/templates/ngx-blueprint/.yo-rc.json#L27-L30
will be used when:

- App generator (no blueprint support)
- The generator is not blueprinted and not using getAllJhipsterConfig

mshima:
@sendilkumarn I think it's ready.
This is the best we can get.
If configuration conflicts is found when migrating, a prompt will be shown with the alternatives.

generators/generator-base.js
```diff
@@ -2027,6 +2027,88 @@ module.exports = class extends PrivateBase {
         return jhipsterUtils.getAllJhipsterConfig(generator, force, configRootPath);
     }
 
+    migrateAllBlueprints() {
+        this.jhipsterConfig = this.jhipsterConfig || this._getStorage('generator-jhipster');
+        this.jhipsterOldVersion = this.jhipsterOldVersion || this.jhipsterConfig.get('jhipsterVersion');
```
mshima: With 6.6.0 and app blueprint, jhipsterVersion can be on blueprint config. Use getAllJhipsterConfig at this.

mshima: Quit if old jhipster is greater than 6.6.0

generators/generator-base-blueprint.js
```diff
@@ -26,6 +26,21 @@ const BaseGenerator = require('./generator-base');
...
 module.exports = class extends BaseGenerator {
+    constructor(args, opts) {
+        super(args, opts);
+
+        this.jhipsterConfig = this._getStorage('generator-jhipster');
+        this.fromBlueprint = this.rootGeneratorName() !== 'generator-jhipster';
+
+        if (this.fromBlueprint) {
+            this.blueprintConfig = this.config;
+            this.config = this.jhipsterConfig;
```
mshima: Just run migrate all

generators/generator-base.js
```diff
@@ -2027,6 +2027,88 @@ module.exports = class extends PrivateBase {
...
+            blueprintConfigs = [blueprintConfigs];
+        }
+        this.jhipsterConfig = this.jhipsterConfig || this._getStorage('generator-jhipster');
+        // Remove duplicated value
+        blueprintConfigs.forEach(bc => {
+            Object.keys(bc.getAll()).forEach(key => {
+                const blueprintValue = bc.get(key);
+                if (this.jhipsterConfig.get(key) === blueprintValue || blueprintValue === undefined) {
+                    this.blueprintConfig.delete(key);
+                }
```
mshima: Move not duplicate config

pascalgrimaud:
@murdos @sendilkumarn : can you have a last look plz ? Do you think it can be merged ?

mshima:
I kept only the core changes.
I will create another PR with the upgrade.

I really don't see any downside at this PR.
You cannot have a separation of blueprint config and core config with getAllJhipsterConfig method.
This is only the 1st step. Then we should remove getAllJhipsterConfig for jhipster 7.

Things got worse with app blueprint on 6.6.0 (see issue https://github.com/jhipster/generator-jhipster/issues/11045).
The blueprint and modules config field is saved on the blueprint namespace but we just look for blueprints and modules on the 'generator-jhipster' namespace (see the bug description).

https://github.com/jhipster/generator-jhipster/blob/bd2de0b9decc3ff9e571cd5097bb0a6639cecf54/generators/generator-base-blueprint.js#L72-L80
This runs only at the main generator (generator-jhipster namespace), the blueprint is never called if the blueprint name is not passed at the command prompt.

This PR fixes https://github.com/jhipster/generator-jhipster/issues/11045.

pascalgrimaud:
thanks @mshima !

PR #10667, opened 2019-10-22
Persistence improvements using hypersistence optimizer

### LINKED ISSUE #10650: JPA improvements using Hypersistence Optimizer

As the title says.

These improvements will make persistence more efficient.
Here is the discussion: https://groups.google.com/forum/#!topic/jhipster-dev/GL-yOZmtlws 

### LINKED ISSUE #10653: Update Mysql dialect to 8.0

This is not a feature request but rather a question regarding mysql dialect should be updated or not.
While working on #10650 with mysql database, I got a MAJOR warning message as following.

```
MAJOR - DialectVersionEvent - Your application is using the 
[org.hibernate.dialect.MySQL5InnoDBDialect] Hibernate-specific Dialect. Consider using the 
[org.hibernate.dialect.MySQL57Dialect] instead as it's closer to your current database server version 
[MySQL 8.0]. For more info about this event, check out this User Guide link - 
https://vladmihalcea.com/hypersistence-optimizer/docs/user-guide/#DialectVersionEvent
```
While I can update to `MySQL57Dialect `, it seems there is a new dialect available `org.hibernate.dialect.MySQL8Dialect` for our current hibernate version 5.3.x.

But before updating to either, I was wondering how many people are still on mysql db server 5.x.x.
I was not thinking to upgrade to `org.hibernate.dialect.MySQL8Dialect` but since we already have mysql docker image version 8.0.17, what do you think @jhipster/developers ?

Mysql 8.0 was out in Apr, 2018. https://www.mysql.com/news-and-events/

Have the latest dialect version.

pmverma:
Fixes #10650 
Fixes #10653 

Please note that all are verified using the `prod` profile only.

- [x] Postgresql
- [x] Mysql
- [x] Mariadb
- [x] Mssql
- [x] Oracle

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -296,7 +296,7 @@ public class <%= asEntity(entityClass) %> implements Serializable {
     <%_ } else if (relationshipType === 'many-to-one') {
             if (databaseType === 'sql') {
     _%>
-    @ManyToOne<% if (relationshipRequired) { %>(optional = false)<% } %>
+    @ManyToOne(fetch = FetchType.LAZY<% if (relationshipRequired) { %>, optional = false<% } %>)
         <%_ if (relationshipValidate) { _%>
     <%- include relationship_validators -%>
         <%_ }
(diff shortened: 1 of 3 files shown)


Blackdread:
For OneToOne, if MapsId is not used then on one side of the relationship (side without foreign key) will always be eager fetched, even if you put Lazy.

pmverma:
@Blackdread Yes you are correct. If it is bidirectional then there will be no effect. But for unidirectional, this will work.

pmverma:
Thanks @DanielFran!

cbornet:
There's a problem here : if we set the fetch strategy to be lazy at hibernate level, then it should be done at controller or service level. With the new behavior, if you call the REST endpoint for an entity, you don't see the related entity in the endpoint (it's `null`) and so it's not displayed in the front-end which is unexpected. It can even be very harmful : if you edit the entity then save without any modification, then the relationship is erased !

pmverma:
Thanks @cbornet for describing the problem. Actually yes, I generated and tested the PR by skipping the client during generation. I will check it.

pmverma:
For the moment, I will revert the lazy load. It is better to have a patch release for this.

Issue #10604, opened 2019-10-11
Some Daily Builds are Broken


SudharakaP:
##### **Overview of the issue**

While working on another issue, I noticed that Infinispan builds are broken on our [Daily Builds](https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build?definitionId=34); 

https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build?definitionId=31 

and

https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build?definitionId=30

##### **Motivation for or Use Case**

Just to make the daily builds perfect. 🌠 

##### **Related issues**

None that I know of. 

##### **Suggest a Fix**

Currently unknown.

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
cc @Hawkurane as he's the one who worked on JDL Daily Builds.

Hawkurane:
When I did the scripts there might have been steps that I did not do that could have resulted in the build failing (e.g. changing the hosts file for keycloak).
If there wasn't anything particular that needed to be done compared to other builds then that means the generator is broken and needs to be fixed 🛠

pascalgrimaud:
just cc the other ticket:

It seems that the Keycloak daily build has some problems with it's end-to-end tests phase.

https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build?definitionId=34&_a=summary

In every build the e2e phase seems to be broken.

SudharakaP:
@Hawkurane : Thanks for the information. I'll give it a go this evening. 😄

SudharakaP:
This seems to be failing on the `sendCacheInformation` call; I am trying to understand why; :thinking: 

`Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.infinispan.manager.DefaultCacheManager]: Factory method 'defaultCacheManager' threw exception; nested exception is org.springframework.beans.factory.BeanInitializationException: Cache (Infinispan protocol stack) configuration failed; nested exception is java.lang.NullPointerException
at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:185)
at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:622)
... 170 common frames omitted
[...]

PR #10673, opened 2019-10-22
Fix MariaDB Accessibility Issues on E2E Tests in Daily Builds

SudharakaP:
This fixes the [error seen in `docker-compose` logs](https://github.com/openzipkin/docker-zipkin/issues/181) due to MariaDB not being accessible. Although I did this as part of my investigations on #10604, I think this will make all MariaDB builds (running under docker) more stable; as you see currently all MariaDB builds in our Daily Builds are flaky. 

Fixes part of #10604 

Reference: https://github.com/openzipkin/docker-zipkin/issues/181

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -126,7 +126,7 @@ module.exports = class extends BaseDockerGenerator {
                         const databaseYaml = jsyaml.load(this.fs.read(`${path}/src/main/docker/${database}.yml`));
                         const databaseServiceName = `${lowercaseBaseName}-${database}`;
                         let databaseYamlConfig = databaseYaml.services[databaseServiceName];
-                        delete databaseYamlConfig.ports;
+                        if (database !== 'mariadb') delete databaseYamlConfig.ports;
 
                         if (database === 'cassandra') {
                             // node config


SudharakaP:
Close/Open for tests

pascalgrimaud:
Does it mean MariaDB is not accessible from other host, but only from localhost ?
Really strange because we don't have this for MySQL

SudharakaP:
@pascalgrimaud : Yes, correct. It seems to be the case. :thinking:

pascalgrimaud:
Reading the documentation at https://mariadb.com/kb/en/library/installing-and-using-mariadb-via-docker/ -> see `Connecting to MariaDB from Outside the Container`

We need to edit some config inside the container... no other option that mapping the port, so let's merge this

SudharakaP:
@pascalgrimaud : Thanks for linking that. I think we can have a more complex solution where we edit the file and then start the container; but I think exposing the port is just simple and no harm done as it's just a testing environment.

SudharakaP:
@pascalgrimaud : Apparently my fix didn't work. 🤔 I have to investigate more. But not to worry, I think I have an idea what is going on here; will look into it more today or tomorrow 😄

SudharakaP:
@pascalgrimaud : Found the missing piece. We need to give a decent sleep timeout to maria db. This seems to fix the issue completely. :smile:

pascalgrimaud:
I hope :)

SudharakaP:
> I hope :)

Yep, it does !!! Now we have only have JDLMs.Keycloak and the flacky Microservices.UAA to fix 😄

PR #10683, opened 2019-10-25
Use strict typescript option in audits module in Angular

kaidohallik:
This is next peace for #10631

Some comments about changes.

* Switching on typescript `strict` option forces to look at all kind of corner cases. So I made some changes in logic. I tested all these cases manually, but added also full unit tests.

* For date testing added `jest-date-mock` library.

* All internal methods are now private.

* Error messages were not shown before these changes.

* If start or end date was empty then NPE in server side before these changes.

* I removed `routeData.unsubscribe()`, because this in not necessary according to https://angular.io/guide/router#!#route-parameters (but it states: _Feel free to unsubscribe anyway. It is harmless and never a bad practice._) - currently in `audits` and `user-management` this `unsubscribe` is used, but in `entities` is not used. I believe it's better to not write unnecessary code as this is currently in entities.

* `link` header is not used, so removed variable for that

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
@@ -21,7 +21,7 @@
 
     <<%= jhiPrefixDashed %>-alert-error></<%= jhiPrefixDashed %>-alert-error>
 
-    <div class="row" *ngIf="initialized">
+    <div class="row">
         <div class="col-md-5">
             <h4 jhiTranslate="audits.filter.title">Filter by date</h4>
             <div class="input-group mb-3">
@@ -38,13 +38,13 @@
         </div>
     </div>
 
-    <div class="alert alert-warning" *ngIf="dataHaveBeenLoaded && audits.length === 0">
+    <div class="alert alert-warning" *ngIf="audits?.length === 0">
         <span jhiTranslate="audits.notFound">No audit found</span>
     </div>
 
-    <div class="table-responsive" *ngIf="audits.length > 0">
+    <div class="table-responsive" *ngIf="audits?.length > 0">
         <table class="table table-sm table-striped"  aria-describedby="audits-page-heading">
-            <thead [ngSwitch]="canLoad">
+            <thead [ngSwitch]="canLoad()">
             <tr jhiSort [(predicate)]="predicate" [(ascending)]="ascending" [callback]="transition.bind(this)" *ngSwitchCase="true">
                 <th scope="col" jhiSortBy="auditEventDate"><span jhiTranslate="audits.table.header.date">Date</span> <fa-icon [icon]="'sort'"></fa-icon></th>
                 <th scope="col" jhiSortBy="principal"><span jhiTranslate="audits.table.header.principal">User</span> <fa-icon [icon]="'sort'"></fa-icon></th>
@@ -72,12 +72,12 @@
         </table>
     </div>
 
-    <div *ngIf="audits.length > 0">
+    <div *ngIf="audits?.length > 0">
         <div class="row justify-content-center">
             <jhi-item-count [page]="page" [total]="totalItems" [itemsPerPage]="itemsPerPage"></jhi-item-count>
         </div>
         <div class="row justify-content-center">
-            <ngb-pagination [collectionSize]="totalItems" [(page)]="page" [pageSize]="itemsPerPage" [maxSize]="5" [rotate]="true" [boundaryLinks]="true" (pageChange)="loadPage(page)" [disabled]="!canLoad"></ngb-pagination>
+            <ngb-pagination [collectionSize]="totalItems" [(page)]="page" [pageSize]="itemsPerPage" [maxSize]="5" [rotate]="true" [boundaryLinks]="true" (pageChange)="loadPage(page)" [disabled]="!canLoad()"></ngb-pagination>
         </div>
     </div>
 </div>
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
@@ -16,10 +16,12 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<div *ngIf="audits">
+<div>
     <h2 id="audits-page-heading" jhiTranslate="audits.title">Audits</h2>
 
-    <div class="row">
+    <<%= jhiPrefixDashed %>-alert-error></<%= jhiPrefixDashed %>-alert-error>
+
+    <div class="row" *ngIf="initialized">
         <div class="col-md-5">
             <h4 jhiTranslate="audits.filter.title">Filter by date</h4>
             <div class="input-group mb-3">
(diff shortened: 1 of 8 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
```diff
@@ -30,113 +29,109 @@ import { AuditsService } from './audits.service';
...
-    links: any;
-    page: number;
-    routeData: any;
-    predicate: any;
-    previousPage: any;
-    reverse: boolean;
-    toDate: string;
-    totalItems: number;
+export class AuditsComponent implements OnInit {
+    private dateFormat = 'yyyy-MM-dd';
```
wmarques: I would put private fields at the end of public fields

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
```diff
@@ -30,113 +29,109 @@ import { AuditsService } from './audits.service';
...
         }
-
-        this.fromDate = this.datePipe.transform(fromDate, dateFormat);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        this.fromDate = this.datePipe.transform(date, this.dateFormat)!;
     }
 
-    today() {
-        const dateFormat = 'yyyy-MM-dd';
+    private today(): void {
```
wmarques: I would prefer this function returning today + 1 Date and then you can do this.toDate = this.today()
WDYT ?

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
```diff
@@ -30,113 +29,109 @@ import { AuditsService } from './audits.service';
...
-        this.alertService.error(error.error, error.message, null);
+    private onSuccess(audits: Audit[] | null, headers: HttpHeaders): void {
+        this.totalItems = Number(headers.get('X-Total-Count'));
+        if (audits === null) {
+            this.audits = [];
+        } else {
+            this.audits = audits;
+        }
+        this.dataHaveBeenLoaded = true;
+        this.initialized = true;
```
wmarques: What's the difference between dataHaveBeenLoaded and initialized ?
Btw it's dataHasBeenLoaded I think 😄

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.service.ts.ejs
```diff
@@ -28,10 +28,8 @@ import { Audit } from './audit.model';
 export class AuditsService  {
     constructor(private http: HttpClient) { }
 
-    query(req: any): Observable<HttpResponse<Audit[]>> {
+    query(req: { page?: number; size?: number; sort?: string[]; fromDate: string; toDate: string }): Observable<HttpResponse<Audit[]>> {
         const params: HttpParams = createRequestOption(req);
-        params.set('fromDate', req.fromDate);
-        params.set('toDate', req.toDate);
```
wmarques: Why are we removing thos params from the request?

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
kaidohallik: Done.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
kaidohallik: Good idea, done and made the same also for the `previousMonth()` function.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
kaidohallik: `initialized` is valued to `true` also then if error occurs on data loading. So if on first page load error occurs then anyway date filter is shown and user can try to change dates and retry data loading in this way.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.service.ts.ejs
kaidohallik: Those parameters are not removed from the request. These parameters are already assigned in the first line of this function and no need to assign values again.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
kaidohallik: I moved `initialized = true` to `ngOnInit` now.  
It aligns initial behavior with further behavior: running new query before previous query is not completed is now allowed also while first query is running.  
And code looks better after this change.  
I force pushed this change.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
```diff
@@ -30,113 +29,110 @@ import { AuditsService } from './audits.service';
...
+    itemsPerPage = ITEMS_PER_PAGE;
+    page!: number;
+    predicate!: string;
+    previousPage!: number;
+    ascending!: boolean;
+    toDate = '';
+    totalItems = 0;
+    canLoad = false;
+    initialized = false;
+    dataHaveBeenLoaded = false;
```
wmarques: I'm not a big fan of those 3 booleans... Can we instead use a `state` variable with 3 values `'initialized' | 'loading' | 'loaded'` or something like this ? WDYT ?

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
```diff
@@ -36,17 +38,24 @@
         </div>
     </div>
 
-    <div class="alert alert-warning" *ngIf="audits?.length === 0">
+    <div class="alert alert-warning" *ngIf="dataHaveBeenLoaded && audits.length === 0">
```
wmarques: Typo: `dataHasBeenLoaded`

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
kaidohallik: Thanks for pointing this out! I removed all 3 variables:
* canLoad --> now it's function, value is calculated
* initialized --> just removed, filter section survives if this is shown before `ngOnInit`
* dataHaveBeenLoaded --> `audits` array variable can now be also `null`, null means that data not loaded yet

I also removed `eslint disable` lines and set `@typescript-eslint/no-non-null-assertion` off as decided in #10639 - hopefully Git conflicts will not arise if merging different PR-s (exactly the same change in `.eslintrc`).

I amended previous review commit and force pushed.

generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.html.ejs
kaidohallik: variable now removed at all

kaidohallik:
As per https://github.com/jhipster/generator-jhipster/pull/10639#discussion_r348303798 I used `?:` in `audits` array declaration. I amended latest commit and force pushed.

wmarques:
Let's go ahead and see how it goes ! Again, thanks @kaidohallik 👏 
That's a huge work and it provides a much better Angular code 😄

kaidohallik:
@wmarques thank you for your reviews!  
These helped to make these PR-s better.

wmarques:
Well it took time but I managed to do it :)

PR #10686, opened 2019-10-26
refactor entity's integration test class

stonedMoose:
I use Jhipster for a few months now, and I was surprised to find some tests in EntityResourceIT which does not belong here in my opinion:

- The first one, is the `fromId`test which appear when mapstruct is involved. This test is not in anyway an integration test and it does not test the Resource but the Mapper. So I propose to move it to a class named EntityMapperTest

- The last two tests which do not belong to EntityResourceIT are the `equalsVerifier`. These tests ensure that the `equals`method on Entity or DTO works as intended, so they should be respectively in `EntityTest` and `EntityDTOTest`.

It's not a big improvment and it will not change the face of the earth, but it make sense to me, I hope it will make sense to you too. 
Plus it will lower the amount of integration test and increase the number of unit test in the generated apps.


generators/entity-server/files.js
```diff
@@ -239,6 +239,35 @@ const serverFiles = {
...
+            path: SERVER_TEST_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/service/dto/EntityDtoTest.java',
+                    renameTo: generator => `${generator.packageFolder}/service/dto/${generator.asDto(generator.entityClass)}Test.java`
+                }
+            ]
+        },
+        {
+            condition: generator => generator.dto === 'mapstruct' && (generator.databaseType === 'sql' || generator.databaseType === 'mongodb' || generator.databaseType === 'couchbase'),
```
stonedMoose: I'm not sure about this condition:

The `fromId`method of the mapper is generated only when DB are sql or mongodb or couchbase, but the test is only generated when the DB is sql...

generators/entity-server/files.js
```diff
@@ -239,6 +239,37 @@ const serverFiles = {
                     renameTo: generator => `gatling/user-files/simulations/${generator.entityClass}GatlingTest.scala`
                 }
             ]
+        },
+        {
+            path: SERVER_TEST_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/domain/EntityTest.java',
+                    renameTo: generator => `${generator.packageFolder}/domain/${generator.asEntity(generator.entityClass)}Test.java`
```
murdos:
```suggestion
                    renameTo: generator => `${generator.packageFolder}/domain/${generator.entityClass}Test.java`
```

stonedMoose: Thank you for the remark, I took it in account  in my last commit.

I fixed the missing import to UUID in MapperTest

murdos:
Thanks @stonedMoose and congrats for your first contribution!

pascalgrimaud:
Thanks a lot !
Congrats for your 1st contribution @stonedMoose !

PR #10688, opened 2019-10-27
Use simple modal in entity delete in Angular

kaidohallik:
I looked at entity delete component while dealing with #10631 and found heavy code for delete.

In user management module it's light and simple. I looked at Git history and found out that in user management this was simplified by https://github.com/jhipster/generator-jhipster/commit/3d180131fd6e0d47af6888f25695b1ee85f47a2e as part of #7406

I made the same simplification in entities and removed popup router outlet as this is not needed any more now.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.html.ejs
@@ -23,7 +23,6 @@
 <div class="container-fluid">
     <div class="card jh-card">
         <router-outlet></router-outlet>
-        <router-outlet name="popup"></router-outlet>
     </div>
     <<%= jhiPrefixDashed %>-footer></<%= jhiPrefixDashed %>-footer>
 </div>
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
@@ -19,17 +19,15 @@
 <%_
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
 _%>
-import { Component, OnInit, OnDestroy } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
+import { Component } from '@angular/core';
 
-import { NgbActiveModal, NgbModal, NgbModalRef } from '@ng-bootstrap/ng-bootstrap';
+import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap';
 import { JhiEventManager } from 'ng-jhipster';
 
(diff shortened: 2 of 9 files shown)


wmarques:
@kaidohallik The router-outlet was present to open modals based on the URL. What happens now when I have the delete modal opened and I refresh?

kaidohallik:
@wmarques If delete modal is opened and you refresh then delete modal is closed now. Do you think this is a problem?

I believe that it doesn't matter for end users, will this modal be still open or not after page refresh. And why should user refresh the page?

I believe that JHipster users will be happy if we remove this complex code and make entities code clearer and simpler.

vishal423:
I also don't like routing logic linked with delete flow. I generally view it as a way to bookmark a link and don't see a use case where an end-user would like to bookmark a delete entity view.

vishal423:
IMO, this would be a breaking change from current behavior. Should we wait for `v7`?

kaidohallik:
@vishal423 Do you feel that whole this change is breaking? Or if I remove this change: https://github.com/jhipster/generator-jhipster/pull/10688/files#diff-1c469cf25bdcf617d05b207f618d8da3L26, is this then non-breaking? And this popup router-outlet can be removed in v7?

vishal423:
After thinking a bit more about breaking changes, I believe to not worry much about this and instead deliver code improvements consistently (could be breaking).

Breaking change can be highlighted in the release notes and will not require team/contributors to wait for the next major release. What your opinion @jhipster/developers ?

pascalgrimaud:
@vishal423 : agree with you, but I'm not sure it's really a breaking change. It looks like more than a refactoring and an improvement of our code.
Breaking changes are:
- change folder name
- rename files
- delete files
- move to librairies

vishal423:
It could be breaking in the sense that we have changed our approach to delete an entity. If an end-user has modified code in deleted functions/files, then, they have to handle that manually during the upgrade.

pascalgrimaud:
@vishal423 : yes you're right, but it's not a big breaking change. As you said, mentioning it in the release note should be enough, it's just my opinion :)

vishal423:
yes, I think we can merge this after you create `v6.5.1` patch release.

deepu105:
Guys we are following the semver versioning which means we are making a
promise. I would not prefer to do breaking changes in minor/patch versions.
If there is strong reason for a breaking change then we should consider a
major release. Breaking changes for small improvements are unnecessary
burden on module/blueprint maintainers and for users upgrading apps. We
need to be more considerate.

deepu105:
Regarding what can be considered breaking, I agree with list from Pascal
along with
- public API changes.
- Changing code structure too much
- changing needles (they are part of public API)
-

wmarques:
For me it's not a big breaking change, I agree @pascalgrimaud. If that's ok for you I'm going to merge this

atomfrede:
I would also say the change is ok. Imho semver is a promise for blueprint/module developers,so users of our api. The generated code is not semver or at least I would not expect it

PR #10710, opened 2019-10-30
invalidate Account cache, if user has logged out

codecholeric:
With the upgrade to JHipster 6.4.0 we noticed some behavior surprising to us. I think it is related to https://github.com/jhipster/generator-jhipster/pull/10383/files#.

In particular, we noticed that if we explicitly log out, we will still receive the old `Account` after calling `AccountService.identity().subscribe(..)`. As far as I could see, this is due to the `accountCache` that does not get invalidated, even if the user logs out. In my opinion, it does not make sense to ever return the last cached `Account`, if `AccountService.authenticated` has been switched to `false` meanwhile, i.e. the user has been logged out. This PR changes the behavior to invalidate the cache within `identity()`, if `authenticated == false`. It would also possible to adjust `AccountService.authenticate(null)` to invalidate the cache, but since the cache is only used for `identity()` and the behavior is somewhat similar to `force`, I added it there.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/auth/account.service.ts.ejs
@@ -73,7 +73,7 @@ export class AccountService  {
     }
 
     identity(force?: boolean): Observable<Account> {
-        if (force) {
+        if (force || !this.authenticated) {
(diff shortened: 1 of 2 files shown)


kaidohallik:
Thanks and congrats for your 1st contribution @codecholeric !

PR #10727, opened 2019-11-05
Exit import-jdl with status from the forked process.

mshima:
Currently every import-jdl command exits with status 0, this patch will make import-jdl and probably other commands exit with the last non 0 status.

### DIFF
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -26,7 +26,7 @@ const pluralize = require('pluralize');
 const { fork } = require('child_process');
 
 const waitUntil = require('./wait-until');
-const { CLI_NAME, GENERATOR_NAME, logger, toString, getOptionsFromArgs, done, getOptionAsArgs } = require('./utils');
+const { CLI_NAME, GENERATOR_NAME, logger, toString, getOptionsFromArgs, done, getOptionAsArgs, setExitCode } = require('./utils');
 const jhipsterUtils = require('../generators/utils');
 
 const packagejs = require('../package.json');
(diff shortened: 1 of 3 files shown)


cli/import-jdl.js
```diff
@@ -204,6 +206,7 @@ const generateEntityFiles = (generator, entity, inFolder, env, shouldTriggerInst
 
             const childProc = forkProcess(runYeomanProcess, [command, ...getOptionAsArgs(options, false, !options.interactive)], { cwd });
             childProc.on('exit', code => {
+                if (code !== 0) setExitCode(code);
```
atomfrede: Is the guard required here? Would it be bad to set the code to 0 again? Would make it easier to read imho. And (just a personal matter of taste) in case the if statement is required, I would add `{` like

```json
if (code !== 0) {
 setExitCode(code);
}
```

mshima: It is, one success will overwrite a fail.
Fork is called at generateApplications, generateEntities and generateDeployments.
Each one, more then one time (ex: this.importState.exportedDeployments.forEach).

An alternative is to exit the script at first fail.

Moved it to the setExitCode.

MathieuAA: Hi @mshima! nice work :)

If we implement an exit-early strategy at first-fail, we risk making the user unhappy if, for instance, the user encounters more than one error.
Maybe we should go for an hybrid approach:
  - for "light" issues, we can aggregate them
  - for "unrecoverable" ones, I'd rather fail as early as possible

mshima: I think they are all unrecoverable.
But I suggest create another PR/issue for this.
This PR is kind of minor while fail as early as possible no.

MathieuAA: Agreed

MathieuAA:
can it be merged?

mshima:
> can it be merged?

Yes, I think so.

PR #10737, opened 2019-11-06
Fix/sub generators

sendilkumarn:

This enables sub-generators to use configurations from the blueprint. 

Fixes: #10738

### DIFF
--- a/generators/ci-cd/index.js
+++ b/generators/ci-cd/index.js
@@ -81,17 +81,18 @@ module.exports = class extends BaseGenerator {
             },
             getConfig() {
                 this.jhipsterVersion = packagejs.version;
-                this.baseName = this.config.get('baseName');
-                this.applicationType = this.config.get('applicationType');
-                this.skipClient = this.config.get('skipClient');
-                this.clientPackageManager = this.config.get('clientPackageManager');
-                this.buildTool = this.config.get('buildTool');
-                this.herokuAppName = this.config.get('herokuAppName');
+                const configuration = this.getAllJhipsterConfig(this, true);
+                this.baseName = configuration.get('baseName');
+                this.applicationType = configuration.get('applicationType');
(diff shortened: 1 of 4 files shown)


generators/ci-cd/index.js
```diff
@@ -81,17 +81,18 @@ module.exports = class extends BaseGenerator {
             },
             getConfig() {
                 this.jhipsterVersion = packagejs.version;
-                this.baseName = this.config.get('baseName');
-                this.applicationType = this.config.get('applicationType');
-                this.skipClient = this.config.get('skipClient');
-                this.clientPackageManager = this.config.get('clientPackageManager');
-                this.buildTool = this.config.get('buildTool');
-                this.herokuAppName = this.config.get('herokuAppName');
+                const configuration = this.getAllJhipsterConfig(this, true);
```
MathieuAA: I've seen this code somewhere, but I don't know why this work compared to what we already have!
What's the difference between these two approaches? thanks @sendilkumarn :)

mshima:
I think this is not needed with https://github.com/jhipster/generator-jhipster/pull/10664

generators/ci-cd/index.js
mshima: getAllJhipsterConfig is a workaround for https://github.com/jhipster/generator-jhipster/issues/10663.
It gets all configs from every namespace.

sendilkumarn:
@mshima I don't think so. For example in Kotlin all the server level keys will only be present in the sub-generator's configurations and not in main generator's configuration.

generators/ci-cd/index.js
MathieuAA: I think I get it, thanks :)

mshima:
I don't know how Kotlin works.
So https://github.com/jhipster/generator-jhipster/pull/10664 will break Kotlin?

PR #10750, opened 2019-11-08
update to gradle 6

atomfrede:
* We have some new deprecation warnings, but at least the no-http plugin still works as the announced deprecation was not done, so we are lucky
* Could not test the automatic shortening of jar commands via classpath jar as I don't have a windows at hand. So if someone wants to test if we can remove the custom classpath jar creation https://docs.gradle.org/6.0/release-notes.html#usability-improvements

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -32,7 +32,7 @@ buildscript {
 
 plugins {
     id "java"
-    id "maven"
+    id "maven-publish"
     id "idea"
     id "jacoco"
     id "org.springframework.boot"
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
@atomfrede : with this upgrade, we encounter an issue in packaging with openapi lib, probably related to https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/build.gradle.ejs#L42

See https://dev.azure.com/hipster-labs/jhipster-daily-builds/_build/results?buildId=7307

Do you have time to have a look plz ?

atomfrede:
I will have a look.

atomfrede:
Damn looks like it is fixed already upstream but not yet released. maybe I find a way to use an older gradle version when selecting open api until a new plugin release is done. If not we have to use 5.x until the plugin > 4.2.0 is released.

pascalgrimaud:
I'd prefer revert this one, instead of using conditional config for openapi, so we won't broke build for openapi config ?

atomfrede:
Give me a this evening, I think there is no much branching involved so not much code to remove afterwards.

pascalgrimaud:
Ok thanks @atomfrede

Issue #9692, opened 2019-05-07
Ability to extend the app sub-generator in blueprints


arjunnayak:
##### **Overview of the feature request**

While we can extend majority of the sub-generators in a blueprint, it would be great to have the option to extend the others, namely the app generator.

##### **Motivation for or Use Case**

In the interest of developer experience at our company, we're trying to simplify as much of the blueprint process as possible. In our case, this involves making modifications to questions that we can't control like type of application, internationalization, testing frameworks, installing other modules from jhipster marketplace, etc. Is this possible, or is there a technical limitation around this?

##### **Related issues or PR**

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
Not sure, I think it has been already asked but I didn't find the ticket or the comment.
Maybe do you remember @avdev4j or @murdos ?

avdev4j:
I don't think we have a ticket for that. I remember a private discussion about doing a blueprint for app generator.
One solution invoked was to do our own app sub generator. TBH I did not try.

I understand your problem @arjunnayak I have the same one for a company blueprint. 
But I wonder if it's really the goal of blueprints. What do you about that @deepu105?

murdos:
I'm a bit puzzled with this request (and I don't remember having already discussed it): by overriding the app subgenerator you can really change the full jhipster behavior, since this subgenerator is orchestrating the other subgenerators (`common`, `server`, `client`, ...).

But why not, we could try. I also wonder what @deepu105 thinks about it.

arjunnayak:
As mentioned in the description, our main concern is simplifying the prompts. At this time, we don’t have other needs to override the app generator. For example, we have our own front end generator based on create-react-app, so there’s no need to even prompt for application type since our blueprint is only used for generating micro services. 

Is it possible to move these prompts to a separate generator? That may lessen the potential for messing up crucial functionality happening in the app generator. Just a thought

avdev4j:
having pre and post prompt sub generators could be a way to do this.

Another solution for you could be having your own jhipster online (the original one is open source too) and remove the fields you don't need (application type for exemple) then it generates a .yo-rc.json and call your blueprint. Every prompt you want to skip will be fill up in the .yo-rc by the jhipster online. I assume it change the way you use your generators.

Kerruba:
Hello, 
We potentially have a similar issue in our company and I think at least being able to limit the type of application or other options in the app sub-generator could be useful.
[...]

PR #10752, opened 2019-11-09
Enable Blueprint for app generator 

sendilkumarn:
This enables app generator to be extedned by the blueprint.


pascalgrimaud:
@sendilkumarn : does it fix https://github.com/jhipster/generator-jhipster/issues/9692 ?

sendilkumarn:
I think yes it does fix it.

PR #10755, opened 2019-11-10
use older gradle version when using openapi plugin

atomfrede:
THis PR does two things:

1. It extrnalizes the gradle version into a constant such that updating should be easier
2. Until the open api plguin has been released in a new version (compatible with gradle 6) when using open api the latest 5.6.4 version is used. This involves minimal branching in two files so the effort to maintain and remove it is quite small

So with this change most of our users will get the latest and greatest gradle version without breaking the open api support.

For details see https://github.com/jhipster/generator-jhipster/pull/10750#issuecomment-552209980

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -25,6 +25,8 @@ const NODE_VERSION = '12.13.0';
 const YARN_VERSION = '1.19.0';
 const NPM_VERSION = '6.13.0';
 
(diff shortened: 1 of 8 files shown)


pascalgrimaud:
thanks @atomfrede 
I'm trusting you and merge this. Then, I can restart some daily builds

Issue #9027, opened 2019-01-08
UAA not working with eureka disabled


xetys:
##### **Overview of the issue**

consider the following JDL:

```
application {
	config {
  		baseName UserService
        	applicationType uaa
        	cacheProvider no
        	enableHibernateCache no
        	buildTool gradle
        	serviceDiscoveryType no
    }
}

application {
	config {
    		baseName GateWay
        	applicationType gateway
        	authenticationType uaa
        	uaaBaseName "UserService"
        	devDatabaseType no
        	prodDatabaseType no
        	databaseType no
        	buildTool gradle
        	serviceDiscoveryType no
    }
}

deployment {
  deploymentType docker-compose
  appsFolders [UserService, GateWay]
  dockerRepositoryName "xetys"
}

```

If we generate this at it is, the setup won't work as of following reasons:

1. we cannot reach any microservices using routes over the gateway as both zuul is missing and no proxy or what ever is present (see #8967)
2. if we enable zuul (what we can do without any additions to dependencies in gradle, as it is already there, while it lacks using maven), then we need to add static routes manually. But here, we need to respect the context-path @saturnism invented in #7697.
3. if we respect the context-path in the zuul routes, we still have problems, as the oauth2 endpoints are pointing to "userservice" instead of "userservice/UserService"
4. if we fix that, the `UaaTokenEndpointClient` and `UaaSignatureVerifierClient` will try to use the `loadBalancedRestTemplate`, which won't work if eureka is not active. We should use the `vanillaRestTemplate` instead
5. without pointing out the ports in the UAA/OAuth2 related endpoints, it will assume using port 80, which is usually not the correct one
6. if we even fix the ports, the gateway won't resolve the correct hostnames, as our docker-compose generate names them like `userservice-app` but not `userservice`

I described the behavior with docker-compose by example, but it will happen the same way using kubernetes.

On the other hand, if we re-activate eureka, everything works like intended.

This issue is very much like #8967, but this one is just an improvment request, while this is a critical issue and should be fixed ASAP.

##### **Suggest a Fix**

Out of my 6 points mentioned above, I see the following TODOs:

- [x] change docker-compose service name convention
- [x] remove the context-path extension invented by @saturnism. Regarding this, I would like to do this in general. I was thinking about what you wanted to do with ingress and stuff, but I suggest to remove that completely for microserivces, as
  - it doesn't work with UAA it all
  - we cannot use stuff like ratelimiter or user defined filter beans for normal microservices in general
[...]

PR #10763, opened 2019-11-11
enable UAA microservices w/o eureka

### LINKED ISSUE #8967: Improve docker-compose for non-eureka microservices

I talked with @saturnism during JHipster Conf in Paris about our microservice experience when working with kubernetes. In a nutshell, using K8s native service discovery, ingress etc, we can get rid of Zuul and Eureka. This is mainly interesting for K8s deployments but doesn't help so much for developer experience.

In other words, you can use Minikube for local development. But usually, this comes with certain limitations for the developer. Running like 10 java microservices inside a VM for Minikube will lack performance. When speaking for myself, my workflow is developing using docker-compose and use kubernetes in production. But our docker-compose isn't aware of that level of service discovery and ingress routing like k8s does, but there are approaches to achieve this.

So I am thinking of 2 changes for our docker-compose generator:

1. create a small project called "jhipster-proxy", which is an extension on nginx (or maybe traefik) in order to pass service names, so they are resolved similarly to what Zuul does automatically, or what our enhanced ingress generator does. In more details, if we have a "gateway", "fooservice" and "barservice", we should route the services over a reverse proxy, and all requests on '/' to the gateway and expose the proxy instead of the gateway to the host port (which usually is 8080). 

2. find a solution how to make the services being resolved with the same hostname like inside k8s or using service discovery. docker-compose already supports hostname resolution by the name of the services, but currently we generate names like "fooservice-app" instead of "fooservice". So we could just rename these names or use some option for docker-compose to make the default app names work.

I am currently working on a prototype at work to solve these 2 things and would love to PR them as soon as they start to work.

xetys:
this PR enhances the experience when using microservices (in particular
UAA) without eureka in docker-compose and kubernetes

closes #8967 and #9027

### DIFF
--- a/generators/server/templates/src/main/java/package/web/filter/RouteDetectorFilter.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/filter/RouteDetectorFilter.java.ejs
@@ -1,0 +1,111 @@
+<%#
+ Copyright 2013-2019 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see https://jhipster.github.io/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+package <%=packageName%>.web.filter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
+import org.springframework.cloud.netflix.zuul.web.ZuulHandlerMapping;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Detects if a call is to be forwarded to a microservice and adds it to zuul routes
+ */
+@Component
+@Order(1)
+public class RouteDetectorFilter implements Filter {
+    private static Logger log = LoggerFactory.getLogger(RouteDetectorFilter.class);
+    private final RestTemplate restTemplate;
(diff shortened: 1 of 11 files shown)


xetys:
Hey, 

here is my first approach how to make UAA based microservices work again when running without eureka. In a nutshell, I have the following issue. When we use eureka, we always can contact the services by their names, say `http://userservice/api/...`. With eureka as service discovery, we can fetch the IP and port implicitly. If we don't use eureka, we should:

1. pass the ports to all calls. For example, in all UAA related stuff like token endpoints (`http://userservice:9999/oauth/token_key`), but also for all microservice calls for entities and so on
2. setup Zuul routes in a different way then we do with eureka

To solve this, I see two solutions:

## Port 80 and services with `RouteDetector`
1. expect all *productive* ports to be 80!. This means:
  a. in docker-compose, all **microservices** will run with `SERVER_PORT=80`, so, inside the docker-compose network, they are resolvable without a port, like `http://serviceName`
  b. in kubernetes, all services will have port 80 for the same reason (while the targetPort can stay)
2. use a `Filter` called route detector in zuul, which detects microservice call and dynamically sets up a static route.

## Full Static approach

We carry the information about microservice names and ports to all required parts (mostly it's the gateway). This is tricky, as e.g. for UAA we have only `uaaBaseName` at the moment. For microservice entities, we had an approach using needles when we were adding webpack proxies on entity generation. Here we could do something similar to zuuls static routes. Then we should involve a duplicate check, which might be somehow hacky.


IMHO, the port-80 approach makes more sense to me. In production, we don't need different ports, as they aren't exposed anyways. Both in docker-compose and in kubernetes, we can configure all services to port 80 and solve a lot of tricky problems like hacking with needles. I'd love to hear what e.g. @PierreBesson @saturnism think about it.


If no one is critical about my first approach, I'll move it to a working and ready2merge state. I can change it later anyways

xetys:
Well, from the current state, this PR allows to run microservices (with UAA) without eureka in docker-compose and kubernetes (and consequently kubernetes-helm). In addition, I fixed the liveness probe bug introduced in #10548. While the PR has a good intension, we propably should open up the info uri for non-authenticated access or something similar.

At the current time, this PR closes #8967 #9027 and #10770

xetys:
P.S.: the one failing test is failing anyways...not related to this PR

pascalgrimaud:
closed / reopened to restart the build

pascalgrimaud:
@xetys : you're the expert on this part, so I'm trusting you. Let's merge this

generators/server/templates/src/main/java/package/web/filter/RouteDetectorFilter.java.ejs
```diff
@@ -0,0 +1,111 @@
...
+            String serviceName = matcher.group(1);
+
+            // test, if the service is missing in zuul routes
+            if (!zuulProperties.getRoutes().containsKey(serviceName)) {
+                try {
+                    URI uri = new URI(
+                        "http",
+                        null,
+                        serviceName,
+                        80,
```
jdubois: The port is hard-coded here? Isn't this going to cause trouble?

pascalgrimaud:
@jdubois : I think you're right about port 80, and we got an issue with docker compose

Then sorry @xetys, unless we can have a fix for this, I'll revert this PR for a new release

Issue #9559, opened 2019-01-24
Regex with escaped content is stripped


deepu105:
In the latest version, generating the below for a React app

```
entity Customer {
    firstName String required
    lastName String required
    gender Gender required
    email String required pattern(/^[^@\s]+@[^@\s]+\.[^@\s]+$/)
    phone String required
    addressLine1 String required
    addressLine2 String
    city String required
    country String required
}
```
this jdl with email field produces 
```
                  <AvField
                    id="customer-email"
                    type="text"
                    name="email"
                    validate={{
                      required: { value: true, errorMessage: translate('entity.validation.required') },
                      pattern: {
                        value: '^[^@s]+@[^@s]+.[^@s]+$',
                        errorMessage: translate('entity.validation.pattern', { pattern: '^[^@s]+@[^@s]+.[^@s]+$' })
                      }
                    }}
                  />
``` with React. The escape charecter for `\s` seems to be stripped.

deepu105:
This still seems like an issue

Shaolans:
Hello,
it seems that it is not the generator issue but more about prettier that delete the backslash.
If you use prettier on this following code:
``` const myregex = "^[^@\s]+@[^@\s]+\.[^@\s]+$";```
The output is the following ```const myregex = "^[^@s]+@[^@s]+.[^@s]+$";```.
This is normal since \s is not a [special character](http://w3schools.sinsixx.com/js/js_special_characters.asp.htm).

PR #10772, opened 2019-11-12
Fix #10520

### LINKED ISSUE #10520: Bug in regex escaping of entity field pattern validation
JHipster v6.3.1
Windows 10
openjdk version "12.0.2"
maven 3.6.1

I have added an entity to my JDL with a field called “Person” like so:

```
entity Person {
	email String unique required minlength(3) maxlength(255) pattern(/^\\S+@\\S+\\.\\S+$/),
}
dto * with mapstruct
```
 
The pattern validation on the Angular Component works fine:
```
export class PersonUpdateComponent implements OnInit {
  isSaving: boolean;


  editForm = this.fb.group({
    email: [null, [Validators.required, Validators.minLength(3), Validators.maxLength(255), Validators.pattern('^\\S+@\\S+\\.\\S+$')]]
  });
```

But as soon as the object hits the back-end code, the DTO Class cannot validate the email field. The problem seems to be an escaping of already escaped backslashes on the DTO side.
```

public class PersonDTO implements Serializable {

    private Long id;

    @NotNull
    @Size(min = 3, max = 255)
    @Pattern(regexp = "^\\\\S+@\\\\S+\\\\.\\\\S+$")
    private String email;

```
I think it would be best to leave the DTO Code as is and add escaping to the Angular Components.  This way developers don’t have to deal with escaping themselves.

Please add information about regex escaping in the JDL documentation at https://www.jhipster.tech/jdl/

Please also clarify in the documentation on when to use syntax pattern(/[A-Z]+/)  over the syntax pattern('[A-Z]+')

Shaolans:

Fix #10520
Looks like this is the similar problem encountered with react (https://github.com/jhipster/generator-jhipster/issues/9559)


atomfrede:
Thanks @Shaolans for the fix and pointing to the fixed react issue!

atomfrede:
Going to merge this as the failing travis build is unrelated.

pascalgrimaud:
@Shaolans : I edited the title, so it's better for the milestone

PR #10775, opened 2019-11-13
Load the parents eagerly

pmverma:
Partial revert the #10667 
Fixes the problem described https://github.com/jhipster/generator-jhipster/pull/10667#issuecomment-552960466


cbornet:
It should also be done for OneToOne

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
```diff
@@ -352,14 +352,14 @@ public class <%= asEntity(entityClass) %> implements Serializable {
...
-    @OneToOne(fetch = FetchType.LAZY<% if (relationshipRequired) { %>, optional = false<% } %>)
+    @OneToOne<% if (relationshipRequired) { %>(optional = false)<% } %>
             <%_ if (relationshipValidate) { _%>
     <%- include relationship_validators -%>
             <%_ }_%>
             <%_ if (isUsingMapsId === true) { %>
     @MapsId
     @JoinColumn(name = "id")
-            <%_ } else { %>
+            <%_ } else { _%>
```
pascalgrimaud: nice eagle eyes

amirensit:
Any idea what is the purposes behind removing lazy loading from manyToOne and oneToOne relations ?

cbornet:
That's because our front expects that the XXXToOne is eagerly loaded. If we don't do it at the hibernate level, it must be done somewhere else. Probably at the repository level.

PR #10765, opened 2019-11-12
CI: remove many-to-many with user entity

pascalgrimaud:
Fix https://github.com/jhipster/generator-jhipster/issues/10762

_____

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10777, opened 2019-11-14
Updated JCore to v6.0.3

MathieuAA:


MathieuAA:
@pascalgrimaud I finally got it!! This one was vicious: I had to debug JCore to find the needle in the haystack. I'm gonna improve some error messages in JCore and fix the wrong JSON test sample.

kaidohallik:
Some notes about my tests.

I made some additional relation removals and **invalid relationship** error was gone in Travis after these changes: https://github.com/kaidohallik/generator-jhipster/commit/9a193a157463192ae84f598810f0f89d3f9af99a

But e2e failed then in Travis after these changes: https://travis-ci.org/kaidohallik/generator-jhipster/jobs/612965788

I tried e2e test locally for `EntityWithServiceClassPaginationAndDTO` and locally this passed.

pascalgrimaud:
@kaidohallik :
- your proposal fix is similar to my PR : https://github.com/jhipster/generator-jhipster/pull/10765
- did you try to restart as it looks like a random failures ?

Then, do it mean we can't have many to many relationship with User anymore ?

kaidohallik:
I saw that @MathieuAA removed one relation, so I just in case searched for all places to give information to spare @MathieuAA time.

Now I see that this information was already available from #10765.

MathieuAA:
@pascalgrimaud 
> Then, do it mean we can't have many to many relationship with User anymore ?

You can, but this sample was wrong as it used the User entity directly and it broke the MtM contract.
The easy way has always been using another User entity through composition or inheritance.

pascalgrimaud:
Oh right thanks for the information.
So yes, if our samples are not correct, we need to fix them !

pascalgrimaud:
Sorry again, but I still don't understand something.
Here what I just tried:
- create default JHipster app (all conf by default), using master branch
- npm link generator-jhipster
- launch: jhipster entity Toto
- create a Many to Many with User -> it's well generated
- I can start the application, and play with it, without any problem: create / update -> works well
- when I use: `jhipster info` -> it failed
- when I use `jhipster export-jdl` -> same, it failed
- that means we're not consistent here as we can generate a MtM relationship but we can't export it ?

MathieuAA:
I'm closing this, this issue needs a clean slate to be fixed

MathieuAA:
This is what I've come up with:

```javascript
function checkOneToOneRelationship(jdlRelationship) {
  if (!jdlRelationship.injectedFieldInFrom) {
    throw new Error(
      `In the One-to-One relationship from ${jdlRelationship.from} to ${jdlRelationship.to}, ` +
        'the source entity must possess the source, or you must invert the direction of the relationship.'
    );
  }
}

function checkManyToOneRelationship(jdlRelationship) {
  if (
    jdlRelationship.injectedFieldInFrom &&
    jdlRelationship.injectedFieldInTo &&
    jdlRelationship.to.toLowerCase() !== 'user'
  ) {
    throw new Error(
      `In the Many-to-One relationship from ${jdlRelationship.from} to ${jdlRelationship.to}, ` +
        'only unidirectionality is supported, you should either create a bidirectional One-to-Many relationship or ' +
        'remove the injected field in the destination entity instead.'
    );
  }
}

function checkManyToManyRelationship(jdlRelationship) {
  if (
    (!jdlRelationship.injectedFieldInFrom || !jdlRelationship.injectedFieldInTo) &&
    jdlRelationship.to.toLowerCase() !== 'user'
  ) {
    const injectedFieldInSourceEntity = !jdlRelationship.injectedFieldInFrom
      ? 'not set'
      : `'${jdlRelationship.injectedFieldInFrom}'`;
    const injectedFieldInDestinationEntity = !jdlRelationship.injectedFieldInTo
      ? 'not set'
      : `'${jdlRelationship.injectedFieldInTo}'`;
    throw new Error(
      `In the Many-to-Many relationship from ${jdlRelationship.from} to ${jdlRelationship.to}, only ` +
        `bidirectionality is supported. The injected field in the source entity is ${injectedFieldInSourceEntity} ` +
        `and the injected field in the destination entity is ${injectedFieldInDestinationEntity}.`
    );
  }
}

```

PR #7129, opened 2018-02-16
Remove the unecessary api call

AnarSultanov:
Remove the unecessary call to get a new CSRF token when CSRF is disabled.

Fix #7090

Issue #7090, opened 2018-02-07
JHipster with Oauth2/OIDC option : csrf disabled but called in UI


warnonphilippe:
Hello,

I'm using Jhipster 4.13.3 with the Oauth2/OIDC option to generate a gateway connected keycloak. 

I think there could be an inconsistency in the generated code, or I missed something ;) In the java config (MicroserviceSecurityConfiguration.java and oAuth2Ss0Configuration.java), CSRF is disabled. 

`http.csrf().disable()`

But in the Angular UI (auth-session.service.ts), a resource is called to refresh the CSRF token.

```
logout(): Observable<any> {        
 // logout from the server         
return this.http.post(SERVER_API_URL + 'api/logout', {}).map((response: Response) => {            
 // to get a new csrf token call the api             
this.http.get(SERVER_API_URL + 'api/account')
```
So, is that call to refresh CSRF token really needed since the CSRF protection is disabled ? Or should CSRF be enabled ?

Thanks !

##### **JHipster Version(s)**
generator-jhipster@4.13.3

##### **JHipster configuration**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "be.phw.tmp"
    },
    "jhipsterVersion": "4.13.3",
    "baseName": "tmp",
    "packageName": "be.phw.tmp",
    "packageFolder": "be/phw/tmp",
    "serverPort": "8080",
    "authenticationType": "oauth2",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": false,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Memory",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "maven",
    "enableSocialSignIn": false,
    "enableSwaggerCodegen": false,
    "clientFramework": "angularX",
    "useSass": false,
    "clientPackageManager": "yarn",
    "applicationType": "gateway",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "enableTranslation": false
  }
}
</pre>
</details>

##### **Environment and Tools**

java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) Client VM (build 25.151-b12, mixed mode, sharing)

git version 2.15.0.windows.1

[...]

PR #10786, opened 2019-11-17
Use strict option in the majority of the core module and in the tracker module in Angular

kaidohallik:
This PR is next peace for the #10631

This PR resolves problems in the most places of **core module** and entirely in the **tracker module**.

I started only with `core/auth` folder, but as tracker module is related then I included also tracker module.

Unresolved stay after this PR:
* `core/login/login-modal.service.ts` - this can be done after in all places is `open()` used as `void`, account module PR #10639 is doing that already for account module
* `core/user/user.model.ts` - this can be done in conjunction with user-management module

Some notes about changes:

* there where memory leaks in tracker component and service (missing unsubscribes from observables), in my testing these are removed now

* following direction taken in #10220 I removed prefixes from tracker component and service

* UAA login returns from server object with tokens and other information, but as this object is not used then I didn't declare model object for that information and left return type to `any`

* user logout was not reflected in tracker component, because `trackerService.disconnect` was not called on user logout (`accountService.authenticate(null)` is used for logout), now this is fixed

* simplified `account.service` and removed repeated code from `account.service`

* `hasAnyAuthority` in `account.service` uses callback to check if user has authority and typescript compiler thinks that `this.userIdentity` can be changed before this callback function is called and gives null check error, currently I resolved this by defining local variable `account` and using this in callback function, so typescript compiler can be sure that no one changes this to `null` before callback function is called

* in `auth-jwt.service getToken()` for UAA was constantly returning `null` and the only usage was in tracker service where actions were performed on `if (getToken()) { .....` - so dead code and I removed that from both places (tracker is working fine with UUA in my testing, so this was really unnecessary code)

* in `auth-jwt.service` for JWT authentication the function `authenticateSuccess` was:
    * in old AngularJs style, this is now replaced with correct Typescript code
    * token was read from http header, now this is read from service response body

* in `auth-session.service` is comment "**to get a new csrf token call the api**" and under that was condition `<%_ if (applicationType !== 'gateway') { _%>`, this resulted the following behavior in 3 possible cases:
    * session authentication (only monolith is possible) is calling this and necessary work is done
    * oauth2 gateway doesn't call this and all is working fine
    * oauth2 monolith made this call but this is not necessary because in oauth authentication flow is user redirected to other page for authentication and after authentication redirected back to JHipster application, so I removed this redundant call also from oauth2 monolith and tested with both Keycloak and Okta and all is working fine

* FYI some links about this `gateway` condition in `auth-session.service`:
    * SO question: https://stackoverflow.com/questions/48660523/jhipster-with-oauth2-oidc-option-csrf-disabled-in-generated-java-config-but-ca?noredirect=1#comment84470664_48660523
    * Issue: #7090
    * PR adding gateway condition: #7129

* simplified tracker service and replaced promises with rxjs objects (mostly promises was removed by #10383, but this file was untouched by #10383)

* test "**should call /account only once**" in `accout.service.spec` was not testing what it claimed it tests, but as test "**should call /account again if user has logged out**" tested also the first one then I removed redundant test and changed correct test name to reflect all cases that it tests

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/tracker/tracker.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/tracker/tracker.component.ts.ejs
@@ -17,23 +17,25 @@
  limitations under the License.
 -%>
 import { Component, OnInit, OnDestroy } from '@angular/core';
+import { Subscription } from 'rxjs';
 
-import { <%=jhiPrefixCapitalized%>TrackerService } from 'app/core/tracker/tracker.service';
+import { TrackerService } from 'app/core/tracker/tracker.service';
+import { TrackerActivity } from 'app/core/tracker/tracker-activity.model';
 
 @Component({
     selector: '<%= jhiPrefixDashed %>-tracker',
     templateUrl: './tracker.component.html'
 })
-export class <%=jhiPrefixCapitalized%>TrackerComponent implements OnInit, OnDestroy {
-
-    activities: any[] = [];
+export class TrackerComponent implements OnInit, OnDestroy {
+    activities: TrackerActivity[] = [];
+    subscription: Subscription | null = null;
 
     constructor(
-        private trackerService: <%=jhiPrefixCapitalized%>TrackerService
+        private trackerService: TrackerService
     ) {
     }
 
-    showActivity(activity: any) {
+    showActivity(activity: TrackerActivity): void {
         let existingActivity = false;
         for (let index = 0; index < this.activities.length; index++) {
             if (this.activities[index].sessionId === activity.sessionId) {
@@ -50,14 +52,17 @@ export class <%=jhiPrefixCapitalized%>TrackerComponent implements OnInit, OnDest
         }
     }
 
-    ngOnInit() {
+    ngOnInit(): void {
         this.trackerService.subscribe();
-        this.trackerService.receive().subscribe((activity) => {
+        this.subscription = this.trackerService.receive().subscribe((activity: TrackerActivity) => {
             this.showActivity(activity);
         });
     }
 
-    ngOnDestroy() {
+    ngOnDestroy(): void {
         this.trackerService.unsubscribe();
+        if (this.subscription !== null) {
+            this.subscription.unsubscribe();
+        }
     }
 }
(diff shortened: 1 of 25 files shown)


generators/client/templates/angular/src/main/webapp/app/core/tracker/tracker.service.ts.ejs
```diff
@@ -112,28 +114,29 @@ export class <%=jhiPrefixCapitalized%>TrackerService {
...
-        this.connection.then(() => {
-            this.subscriber = this.stompClient.subscribe('/topic/tracker', (data) => {
-                this.listenerObserver.next(JSON.parse(data.body));
-            });
+    subscribe(): void {
+        if (this.connectionSubscription !== null) {
+            return;
+        }
+      
+        this.connectionSubscription = this.connectionSubject.subscribe(() => {
```
wmarques: Can we avoid nested `subscribe` by replacing the first `subscribe` with `tap` operator  ?

generators/client/templates/angular/src/main/webapp/app/admin/tracker/tracker.component.ts.ejs
```diff
@@ -50,14 +52,17 @@ export class <%=jhiPrefixCapitalized%>TrackerComponent implements OnInit, OnDest
...
-        this.trackerService.receive().subscribe((activity) => {
+        this.subscription = this.trackerService.receive().subscribe((activity: TrackerActivity) => {
             this.showActivity(activity);
         });
     }
 
-    ngOnDestroy() {
+    ngOnDestroy(): void {
         this.trackerService.unsubscribe();
+        if (this.subscription !== null) {
```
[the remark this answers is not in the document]
kaidohallik: Simplification done.

generators/client/templates/angular/src/main/webapp/app/core/tracker/tracker.service.ts.ejs
kaidohallik: These are different subscriptions, outer is `rxjs` one and inner is `Stomp` one, so can't avoid nesting.

kaidohallik:
I resolved conflicts with master and used `?:` according to https://github.com/jhipster/generator-jhipster/pull/10639#discussion_r348303798 and used correct unsubscribe order to avoid problems in possible edge cases and simplified code by using `!`.

kaidohallik:
I used too eagerly `!` in tracker service so rule `@typescript-eslint/no-unnecessary-type-assertion` was violated so I reverted using `!` in this PR, I amended last commit and removed using `!` (restored if checks) and force pushed changes.

wmarques:
Thanks alot @kaidohallik !

PR #10788, opened 2019-11-17
Correct typos

vivekmore:

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -448,7 +448,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 });
             },
 
-            askFori18n: prompts.askFori18n,
+            askForI18n: prompts.askForI18n,
         };
     }
 
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -448,7 +448,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 });
             },
 
-            askForI18n: prompts.askForI18n,
+            askFori18n: prompts.askForI18n,
         };
     }
 
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -189,7 +189,7 @@ module.exports = class extends BaseBlueprintGenerator {
         return {
             askForModuleName: prompts.askForModuleName,
             askForClient: prompts.askForClient,
-            askForI18n: prompts.askForI18n,
+            askFori18n: prompts.askForI18n,
             askForClientTheme: prompts.askForClientTheme,
             askForClientThemeVariant: prompts.askForClientThemeVariant,
 
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -189,7 +189,7 @@ module.exports = class extends BaseBlueprintGenerator {
         return {
             askForModuleName: prompts.askForModuleName,
             askForClient: prompts.askForClient,
-            askFori18n: prompts.askFori18n,
+            askForI18n: prompts.askForI18n,
             askForClientTheme: prompts.askForClientTheme,
             askForClientThemeVariant: prompts.askForClientThemeVariant,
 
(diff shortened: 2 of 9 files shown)


vivekmore:
Makes sense.
so what do you suggest? 
submit PR on separate 7.x branch?

murdos:
> 
> 
> Makes sense.
> so what do you suggest?
> submit PR on separate 7.x branch?

There's no 7.x branch, we only have maintenance branches.
I've added the 7.x label, I guess the PR will stay open until we start to work on 7.x

jdubois:
Let's first do the "easy" typos, that are not used by the sub-generators. For the others, it's going to be difficult.

vivekmore:
> Let's first do the "easy" typos, that are not used by the sub-generators. For the others, it's going to be difficult.

How about a gradual upgrade by keeping both variants like follows:

```
methodWithoutTypo() {
   ...
   // actual code
   ...
}

methodWITHtypo() {
   this.log.warn('Please use methodWithoutTypo() as this method will be deprecated in next major release');
   methodWithoutTypo();
}
```

MathieuAA:
@vivekmore that's a good idea! Can you add a comment to the "functions with typos" that says
```
@deprecated
To be removed for JHipster v7
```

That way, we could clean them up for the next major

murdos:
> Let's first do the "easy" typos, that are not used by the sub-generators. For the others, it's going to be difficult.

These will still be breaking changes, in particular for blueprints.

For example, if a blueprint is overriding a particular step (`setupClientconsts` in the example below), it won't be overridden anymore if we rename it in main generator to `setupClientConstants`:  
```
get writing() {
        const phaseFromJHipster = super._writing();
        const customPhaseSteps = {
            setupClientconsts() {
                // Do things that change generator-jhipster behavior...
            },
        };
        return { ...phaseFromJHipster, ...customPhaseSteps };
    }
```

Development on V7 will start soon, let's wait a bit more and don't break V6 for blueprints (and there're a more and more)

vivekmore:
@murdos - You're right
However, to get this small PR merged I've removed the breaking changes related to setup*() functions.

I'll make another PR as soon as v7 development starts that:
- removes deprecated methods from this PR
- corrects the setup*() methods (that I removed from my most recent update to this PR)

does that seem reasonable?

generators/app/index.js
```diff
@@ -448,7 +448,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 });
             },
 
-            askFori18n: prompts.askFori18n,
+            askForI18n: prompts.askForI18n,
```
murdos, line 451:
```suggestion
            askFori18n: prompts.askForI18n,
```

generators/client/index.js
```diff
@@ -189,7 +189,7 @@ module.exports = class extends BaseBlueprintGenerator {
         return {
             askForModuleName: prompts.askForModuleName,
             askForClient: prompts.askForClient,
-            askFori18n: prompts.askFori18n,
+            askForI18n: prompts.askForI18n,
```
murdos, line 192:
```suggestion
            askFori18n: prompts.askForI18n,
```

generators/server/index.js
```diff
@@ -348,7 +348,7 @@ module.exports = class extends BaseBlueprintGenerator {
             askForModuleName: prompts.askForModuleName,
             askForServerSideOpts: prompts.askForServerSideOpts,
             askForOptionalItems: prompts.askForOptionalItems,
-            askFori18n: prompts.askFori18n,
+            askForI18n: prompts.askForI18n,
```
murdos, line 351:
```suggestion
            askFori18n: prompts.askForI18n,
```

vivekmore:
> I highlighted the 3 remaining breaking changes that should be reverted. Then this PR could be merged ;)

Good catch @murdos ! Sorry for missing these, and thank you for being patient.
I realize I have a few more stuff to learn about things exported from Generators.

PR #10793, opened 2019-11-18
Modify Readme Page with Sponsor Logos Horizontally Aligned

SudharakaP:
The logos in our README.md file seems a bit space consuming with their vertical stacking. I re-formatted it to contain in a html table. 

To see how it looks --> https://github.com/SudharakaP/generator-jhipster/tree/modify-readme-page-with-sponsor-logo-alignment

I hope this looks better. :smile: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,47 +14,126 @@ Support this project by becoming a sponsor! [Become a sponsor](https://opencolle
 
 **Thank you to our sponsors!**
 
-**Gold sponsors**
+### Gold Sponsors
 
-[![Okta][okta-image]][okta-url]
+<table>
+  <tbody>
+    <tr>
+      <td align="center" valign="middle">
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
it looks better on computer but not on phone. Do you think it's possible to have both ?

SudharakaP:
> 
> 
> it looks better on computer but not on phone. Do you think it's possible to have both ?

@pascalgrimaud : Good catch. Yes, I think it should be possible; I've put absolute width measurements (pixels). Instead I have to convert them into relative measurements I think. Will do that in the evening. 😄

SudharakaP:
@pascalgrimaud : Okay, I've modified it; try it out now and let me know if it's better. I can't dynamically change the grid itself since I don't think the `README.md` file supports libraries like bootstrap. But I've fixed the grid to 5 columns which seems okay to me. :smile:

pascalgrimaud:
thanks, it's better now :)

PR #10794, opened 2019-11-18
Updated JCore to v6.0.4 & fixed the JDL export error logging 

MathieuAA:

### DIFF
--- a/generators/export-jdl/index.js
+++ b/generators/export-jdl/index.js
@@ -19,7 +19,6 @@
 const chalk = require('chalk');
 const jhiCore = require('jhipster-core');
 const BaseGenerator = require('../generator-base');
-const { logger } = require('../utils');
 const statistics = require('../statistics');
 
 module.exports = class extends BaseGenerator {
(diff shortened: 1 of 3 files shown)


pascalgrimaud:
The build still failed, because of this:
- the installation of JHipster in the CI is done without `npm link`, but with a real installation of the package
- once the application is generated, the `node_modules` contains the last release version (here 6.5.0), so it will use the JHipster Core 6.0.1, so it failed

That's why the CI didn't detect it until the release.
And that's why the CI will fail until a new release.
But as it concerns only the `jhipster info` commands, I think it's ok.

murdos:
Thanks for the explanation! I was wondering why we hadn't detected the issue earlier.
Is it possible to fix the CI and use `npm link`?

pascalgrimaud:
We can of course use `npm link`, but by doing this, we won't detect some potential other problems until the release. We already encounted these kind of things in the past.
- current installation -> similar as an npm install -g <package>
- npm link -> workflow for development

I think it's better to keep the current installation as it's the closest to the final release, and there is only 1 command which is done after the generation: `jhipster info`.

pascalgrimaud:
I'm confident, let's merge this
Then, I'll do a patch release, probably tomorrow, after waiting all our daily builds

PR #10796, opened 2019-11-18
fix docker task in GitLab CI-CD sub-gen

### LINKED ISSUE #10795: minor issues with Gitlab CI-CD

The out commented code block in the `.gitlab-ci.yml` for docker-push has two little issues I noticed:

1. when running `jib` without `-Pprod`, the build tool generates a dev frontend, which is not intended in production
2. we set `IMAGE_TAG` to `$CI_REGISTRY_IMAGE:$CI_COMMIT_REF_SLUG:$CI_COMMIT_SHA`, what creates an invalid container image tag

Using `docker-push` in Gitlab CICD

##### **Reproduce the error**

Use `jhipster ci-cd` to generate a `.gitlab-ci.yml` and uncomment the docker-push task.

xetys:
Includes `-Pprod` in both, maven and gradle `docker-push` tasks and fixes the tag name strategy to avoid invalid tag names

fix #10795

### DIFF
--- a/generators/ci-cd/templates/.gitlab-ci.yml.ejs
+++ b/generators/ci-cd/templates/.gitlab-ci.yml.ejs
@@ -134,11 +134,11 @@ gradle-package:
 #    stage: release
 #    variables:
 #        REGISTRY_URL: registry.gitlab.com
-#        IMAGE_TAG: $CI_REGISTRY_IMAGE:$CI_COMMIT_REF_SLUG:$CI_COMMIT_SHA
+#        IMAGE_TAG: $CI_REGISTRY_IMAGE:$CI_COMMIT_REF_SLUG-$CI_COMMIT_SHA
 #    dependencies:
 #        - gradle-package
 #    script:
-#        - ./gradlew jib -Djib.to.image=$IMAGE_TAG -Djib.to.auth.username="gitlab-ci-token"  -Djib.to.auth.password=$CI_BUILD_TOKEN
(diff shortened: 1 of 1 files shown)


generators/ci-cd/templates/.gitlab-ci.yml.ejs
```diff
@@ -134,11 +134,11 @@ gradle-package:
 #    stage: release
 #    variables:
 #        REGISTRY_URL: registry.gitlab.com
-#        IMAGE_TAG: $CI_REGISTRY_IMAGE:$CI_COMMIT_REF_SLUG:$CI_COMMIT_SHA
```
atomfrede: Fun fact I was about to do something similar. Can't we use ` IMAGE_TAG: $CI_REGISTRY_IMAGE/$CI_COMMIT_REF_SLUG:$CI_COMMIT_SHA`? With that the gitlab container scanning should work out of box. 

If you are curious I have setup sample apps on gitlab.com where I would like to integrate the complete gitlab ci feature set like dependency scanning etc here

https://gitlab.com/atomfrede/jhipster-sample-app 

https://gitlab.com/atomfrede/jhipster-sample-app-gradle

xetys: I think it is a question of favor. It would work, too. But then the actual image name would be 'master' or tag name etc. 

About the dep scanning, sure, but I'd like to have it kind of specialized for our applications...those templates are general and that leads to long CI times IMHO

PR #10808, opened 2019-11-20
introduce dynamic storage provisioning in k8s sub-gen

### LINKED ISSUE #10804: Allow using kubernetes PVCs for stateful resources

As already discussed in #5173 or #9622, our kubernetes sub-generator generates all "stateful resources" (such as MySQL, PostgreSQL, etc.) with `emptyDir` as volume definition. 

I propose the following add-on:

When `jhipster kubernetes` asks its prompts, it should ask

```
Do you want to use a storage class for dynamic storage provisioning?
- No
- Yes, the default storage class
- Yes, with a specified storage class
```

If the user chooses the last option, an additional question asks the `storageClassName`. 

I think the motivation is clear: we want to define non-empheral storage as we want to keep our data. In the past discussions, we were concerning about the cloud-specific storage configurations. AFAIK, kubernetes developed the concept of storage classes and persistent volume claim with the aim of decouple persistence with the persistent provider.

In usual clusters, with a recent kubernetes version, you just can create a PVC without specifying `storageClassName` or annotations. If the cluster has a default storage class, kubernetes assumes such an empty definition as a PVC for the default storage class implicitly. 

In scenarios, where cluster administrators enable several storage classes, we can ask the user for the storage class name. In earlier versions of k8s, there was no clear standard about how to set the storage class name, as you could either use the property, or `alpha`/`beta` annotations. However, as of today, it is well established.

I see the following tasks here:

- [ ] add the new prompts to sub-generator with corresponding cofigurations
- [ ] when using PVCs is enabled, propagate the correct PVC to all persistent resources, including app databases, JHipster console, etc
- [ ] add PVC related options to JDL in deployment->kubernetes

xetys:
with this feature, all stateful services (such as `Deployment`, `StatefulSet` for databases) will use persistent volume claims instead of `emptyDir`. There are two new options for the user, so it is possible to enable dynamic storage provisioning and specify a custom storage class.

Another update in context with this PR, is that `StatefulSet`s are no longer generated with `volumeClaimTemplate` by default, as it was before. Instead, the data dirs are either `emptyDir`, if dynamic storage provisioning is disabled, or we use volume claim templates with the same configuration as for the deployments.

close #10804

### DIFF
--- a/generators/kubernetes/prompts.js
+++ b/generators/kubernetes/prompts.js
@@ -274,7 +274,7 @@ function askForStorageClassName() {
     ];
 
     this.prompt(prompts).then(props => {
-        this.kubernetesStorageClassName = props.kubernetesStorageClassName;
+        this.kubernetesStorageClassName = props.kubernetesStorageClassName.trim();
         done();
     });
 }
--- a/generators/kubernetes/prompts.js
+++ b/generators/kubernetes/prompts.js
@@ -25,6 +25,8 @@ module.exports = {
     askForIngressType,
     askForIngressDomain,
     askForIstioSupport,
+    askForPersistentStorage,
+    askForStorageClassName,
     ...dockerPrompts
 };
 
@@ -216,6 +218,63 @@ function askForIstioSupport() {
 
     this.prompt(prompts).then(props => {
         this.istio = props.istio;
+        done();
+    });
+}
+
+function askForPersistentStorage() {
+    if (this.regenerate) return;
+    const done = this.async();
+    let usingDataBase = false;
+    this.appConfigs.forEach((appConfig, index) => {
+        if (appConfig.prodDatabaseType !== 'no') {
+            usingDataBase = true;
+        }
+    });
+
+    const prompts = [
+        {
+            when: () => usingDataBase,
+            type: 'list',
+            name: 'kubernetesUseDynamicStorage',
+            message: 'Do you want to use dynamic storage provisioning for your stateful services?',
+            choices: [
+                {
+                    value: false,
+                    name: 'No'
+                },
+                {
+                    value: true,
+                    name: 'Yes'
+                }
+            ],
+            default: this.kubernetesUseDynamicStorage
+        }
+    ];
+
+    this.prompt(prompts).then(props => {
+        this.kubernetesUseDynamicStorage = props.kubernetesUseDynamicStorage;
+        done();
+    });
+}
+
+function askForStorageClassName() {
+    if (this.regenerate) return;
+    const done = this.async();
+    const kubernetesUseDynamicStorage = this.kubernetesUseDynamicStorage;
+
+    const prompts = [
+        {
+            when: () => kubernetesUseDynamicStorage,
+            type: 'input',
+            name: 'kubernetesStorageClassName',
+            message: 'Do you want to use a specific storage class? (leave empty for using the clusters default storage class)',
+            default: this.kubernetesStorageClassName ? this.kubernetesStorageClassName : ''
+        }
+    ];
+
+    this.prompt(prompts).then(props => {
+        this.kubernetesStorageClassName = props.kubernetesStorageClassName;
         done();
     });
 }
--- a/generators/kubernetes/templates/console/jhipster-elasticsearch.yml.ejs
+++ b/generators/kubernetes/templates/console/jhipster-elasticsearch.yml.ejs
@@ -178,6 +178,10 @@ spec:
       - name: config
         configMap:
           name: es-config
+<%_ if (!kubernetesUseDynamicStorage) { _%>
+      - name: storage
+        emptyDir: {}
+<%_ } else { _%>
   volumeClaimTemplates:
   - metadata:
       name: storage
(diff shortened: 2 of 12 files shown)


generators/kubernetes/prompts.js
```diff
@@ -219,3 +221,60 @@ function askForIstioSupport() {
...
+            when: () => kubernetesUseDynamicStorage,
+            type: 'input',
+            name: 'kubernetesStorageClassName',
+            message: 'Do you want to use a specific storage class? (leave empty for using the clusters default storage class)',
+            default: this.kubernetesStorageClassName ? this.kubernetesStorageClassName : ''
+        }
+    ];
+
+    this.prompt(prompts).then(props => {
+        this.kubernetesStorageClassName = props.kubernetesStorageClassName;
```
srinivasa-vasu: Do you want to add .trim() in case if users press a space by mistake?

generators/kubernetes/templates/console/jhipster-elasticsearch.yml.ejs
```diff
@@ -186,6 +190,10 @@ spec:
       resources:
         requests:
           storage: 1Gi
+  <%_ if (kubernetesStorageClassName !== '') { _%>
```
srinivasa-vasu: nit: If you decide to add .trim(), can we change this to `if (kubernetesStorageClassName)`

generators/kubernetes/prompts.js
xetys: done!

generators/kubernetes/templates/console/jhipster-elasticsearch.yml.ejs
xetys: done here and in all similar parts

generators/kubernetes/templates/db/couchbase.yml.ejs
```diff
@@ -16,7 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: I think we should use KUBERNETES_CORE_API_VERSION

generators/kubernetes/templates/db/elasticsearch.yml.ejs
```diff
@@ -16,6 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: same with KUBERNETES_CORE_API_VERSION ?

generators/kubernetes/templates/db/mariadb.yml.ejs
```diff
@@ -16,6 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: same with KUBERNETES_CORE_API_VERSION ?

generators/kubernetes/templates/db/postgresql.yml.ejs
```diff
@@ -16,6 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: same with KUBERNETES_CORE_API_VERSION ?

generators/kubernetes/templates/db/mssql.yml.ejs
```diff
@@ -16,6 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: same with KUBERNETES_CORE_API_VERSION ?

generators/kubernetes/templates/db/mysql.yml.ejs
```diff
@@ -16,6 +16,23 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
+<%_ if (kubernetesUseDynamicStorage) { _%>
+apiVersion: v1
```
pascalgrimaud: same with KUBERNETES_CORE_API_VERSION ?

xetys: done

generators/kubernetes/templates/db/mssql.yml.ejs
xetys: done

generators/kubernetes/templates/db/postgresql.yml.ejs
xetys: done

generators/kubernetes/templates/db/mariadb.yml.ejs
xetys: done

generators/kubernetes/templates/db/elasticsearch.yml.ejs
xetys: done

generators/kubernetes/templates/db/couchbase.yml.ejs
xetys: done

xetys:
anything missing here?

pascalgrimaud:
let's merge this

PR #10809, opened 2019-11-20
Refactor Kafka generated code

cbornet:

The current Kafka generated code is a bit invasive : by default it starts a Kafka consumer and polls from a topic.
So I propose instead to generate generic endpoints to publish to and consume from any topic (with inspiration from the kafka-rest-proxy).
Also there is no reason to spawn the Spring context in the tests and the resource can be tested in full standalone mode.


cbornet:
Note: I tried to do an SSE streaming endpoint but it doesn't work well on undertow (it's better on Tomcat). Other possibilities could be to use a WebSocket if the option is selected or a reactor-kafka+Webflux endpoint for the reactive option.

jdubois:
@cbornet it's a much bigger issue, but we can also discuss migrating to Tomcat. I never see anybody using Undertow outside of us, and I'm not sure Red Hat will continue maintaining it as they do all of their efforts on Quarkus.

pascalgrimaud:
Thanks @cbornet, there is a lot of good idea here.

The reason why I created the Consumer and Producer in servive layout is because I'd like to add this feature :

1) entity as producer

- add a entity with service
- add an option to define this entity as producer 
- add a producer in service layer, in another class or in the same service as the entity
- each time there is a create, update or delete, publish to Kafka

2) entity as a consumer

- create an entity with read only (no endpoint) but with service
- add a consumer to this entity
- this entity can only be created/updated by consuming the topic from Kafka

It will be an approch to allow communication between microservices. 
What do you think @cbornet ? Is it possible with your current implementation ? I need your advice here :)

cbornet:
I see no incompatibility for this with the code here. This PR only brings some of the rest-proxy endpoints directly in the app. As for doing things for the entity, there are certainly some use cases where it would be valuable (as a kind of Change Data Capture). However I would warn that you should probably send business events in Kafka and not entity ones. Just like you should not use the CRUD endpoints between microservices.

pascalgrimaud:
Yes I agree @cbornet about business events, but as JHipster doesn't generate any business, I wanted to send entities in Kafka as sample, so the end-user have a real example and can recode this part.
Then, we could implement a producer for each entities, but should it be done directly in JHipster or in a specific module ?

cbornet:
This PR has code for producing and consuming Kafkak topics. So users can easily start from there. I think we shouldn't do more by default as there are a lot of different ways to use Kafka (producers/consumers, Kafka Streams, KSQL, Ksqldb, ...).
We could generate some code for the entities, maybe optionally, but that's another discussion I think. Meanwhile can we merge this ?

pascalgrimaud:
Yes @cbornet 
I'll use your PR today, give me some time to try, review and merge it plz

pascalgrimaud:
@cbornet : just give it a try

1) I created a microservice `Alpha` (as Producer)
2) I created a microservice `Beta` (as Consumer)
3) I created a `Gateway`
4) start the 3 app
5) go to swagger for `Alpha`, post a message to topic `toto`
6) go to swagger for `Beta`, create a Consumer `toto`
7) go to swagger for `Beta`, get records for topic `toto`

Here some comments:
- it doesn't' work the first time, I need to call several times the API to get the result
- sometimes, I got this `[Consumer clientId=consumer-1, groupId=sample] Attempt to heartbeat failed since group is rebalancing`, specially when I want to consum more than 1 topic at the same time

Maybe I'm doing something wrong, in this case, let me know.
Otherwise, I'll analyze this further

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -18,115 +18,146 @@
...
+    static void startServer() {
         if (!started) {
             startTestcontainer();
             started = true;
         }
     }
 
     private static void startTestcontainer() {
-        kafkaContainer = new KafkaContainer("<%= KAFKA_VERSION %>");
+        kafkaContainer = new KafkaContainer("5.3.1");
```
ctamisier: we can put <%= KAFKA_VERSION %> instead of 5.3.1 ?

cbornet: :+1:

cbornet:
> it doesn't' work the first time, I need to call several times the API to get the result

This is because at the first poll, the consumer starts the operations to join the group. The current default of 1s of polling is insufficient to join the group and get the messages. So it takes several poll attempts to do the whole process. Kafka also uses a heartbeat mechanism to detect dead consumers. The heartbeats are sent from a thread and if Kafka doesn't receive one in session.timeout.ms (10s by default), the consumer is considered dead. The sending of heartbeats only starts when the consumer has joined the group (https://chrzaszcz.dev/2019/06/kafka-heartbeat-thread/) so if there is more than 10s between 2 polls when the consumer is joining the group, then the process fails and has to restart from the beginning.

So the simplest solution I see is to increase the default poll timeout (eg. to 10s) to ensure the consumer has enough time to join the group.
WDYT?

pascalgrimaud:
Thanks for the analyze.
Yes, we should have something which works immediatly the first time. So if increasing to 10s works, let's do it

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
cbornet: done

cbornet:
@pascalgrimaud done. Can you test with the longer timeout ?

pascalgrimaud:
ok, I will

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -18,68 +18,44 @@
...
-@SpringBootTest(classes = <%= mainClass %>.class)
-<%_ } _%>
-@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)
-<%_ if (cacheProvider === 'redis') { _%>
-@ExtendWith(RedisTestContainerExtension.class)
-<%_ } _%>
-public class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
-
-    private MockMvc restMockMvc;
+class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
```
pascalgrimaud: - can we add `public class` ?
- then, is it a real IT test ? Otherwise, maybe Test instead

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -18,68 +18,44 @@
...
-    @Autowired
-    private <%= upperFirstCamelCase(baseName) %>KafkaConsumer consumer;
-
-    private static final int MAX_ATTEMPT = 5;
+    private <%= upperFirstCamelCase(baseName) %>KafkaResource kafkaResource;
+    private MockMvc restMockMvc;
 
     @BeforeAll
-    public static void startServer() {
+    static void startServer() {
```
pascalgrimaud: public ?

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -89,44 +65,99 @@ public class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
     private static void startTestcontainer() {
         kafkaContainer = new KafkaContainer("<%= KAFKA_VERSION %>");
         kafkaContainer.start();
-        System.setProperty("kafkaBootstrapServers", kafkaContainer.getBootstrapServers());
     }
 
     @BeforeEach
-    public void setup() {
-        <%= upperFirstCamelCase(baseName) %>KafkaResource kafkaResource = new <%= upperFirstCamelCase(baseName) %>KafkaResource(producer);
+    void setup() {
```
pascalgrimaud: public ?

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -89,44 +65,99 @@ public class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
...
+
+        kafkaResource = new <%= upperFirstCamelCase(baseName) %>KafkaResource(kafkaProperties);
 
-        this.restMockMvc = MockMvcBuilders.standaloneSetup(kafkaResource)
+        restMockMvc = MockMvcBuilders.standaloneSetup(kafkaResource)
             .build();
+    }
+
+    @Test
+    void producesMessages() throws Exception {
```
pascalgrimaud: public ?

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -89,44 +65,99 @@ public class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
...
-        consumer.start();
+        assertThat(records.count()).isEqualTo(1);
+        ConsumerRecord<String, String> record = records.iterator().next();
+        assertThat(record.value()).isEqualTo("value-produce");
     }
 
     @Test
-    public void producedMessageHasBeenConsumed() throws Exception {
-        restMockMvc.perform(post("/api/<%= dasherizedBaseName %>-kafka/publish?message=test"))
+    void createsConsumer() throws Exception {
```
pascalgrimaud: public ?

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -89,44 +65,99 @@ public class <%= upperFirstCamelCase(baseName) %>KafkaResourceIT {
...
+        Field groupField = ReflectionUtils.findField(KafkaConsumer.class, "groupId");
+        ReflectionUtils.makeAccessible(groupField);
+        String group = (String)groupField.get(consumer);
+        assertThat(group).isEqualTo("group-create");
+        Set<String> subscription = consumer.subscription();
+        assertThat(subscription).containsExactlyInAnyOrder("topic-create", "topic-create2");
+    }
+
+    @Test
+    void consumesMessages() throws Exception {
```
pascalgrimaud: public ?

pascalgrimaud:
Just tested your last change to 10000 and it works well
Good job @cbornet. Added some minor comments

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
cbornet: Why should it be public ? It was a proposition by IDEA to remove the public

pascalgrimaud: to be consistent with our other class, but it's fine for me here

cbornet: About whether it's a UT or IT, that's a good question. It doesn't spawn the Spring context but it spawns a real Kafka with testcontainer.

cbornet:
Since the polling of Kafka is not easy to grok, I gave another try to an SSE endpoint and finally got something working. This is easier to use and more in the philosophy of Kafka to have a continuous loop over consumer polls.
@pascalgrimaud can you review the new code ?

pascalgrimaud:
Just restarted the builds, I'll test it soon

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
hdurix: I would say to keep it package private as it does not need to be public (and we could change the others for consistency if they don't need to be public).
For me that's more an integration test than a unit one.

pascalgrimaud:
good job @cbornet !

PR #10810, opened 2019-11-20
Remove exception mapping of NoSuchElementException

cbornet:

NoSuchElementException is thrown when calling get() on an empty Optional. That's something that shouldn't happen on the server and should result in a HTTP error 500.
If we want to use an exception for 404, it should be a dedicated one. Although I haven't found a place in the code where this is needed.

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/errors/ErrorConstants.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/errors/ErrorConstants.java.ejs
@@ -29,9 +29,6 @@ public final class ErrorConstants {
     public static final String PROBLEM_BASE_URL = "https://www.jhipster.tech/problem";
     public static final URI DEFAULT_TYPE = URI.create(PROBLEM_BASE_URL + "/problem-with-message");
     public static final URI CONSTRAINT_VIOLATION_TYPE = URI.create(PROBLEM_BASE_URL + "/constraint-violation");
-    <%_ if (!reactive) { _%>
-    public static final URI ENTITY_NOT_FOUND_TYPE = URI.create(PROBLEM_BASE_URL + "/entity-not-found");
-    <%_ } _%>
     <%_ if (!skipUserManagement) { _%>
     public static final URI INVALID_PASSWORD_TYPE = URI.create(PROBLEM_BASE_URL + "/invalid-password");
     public static final URI EMAIL_ALREADY_USED_TYPE = URI.create(PROBLEM_BASE_URL + "/email-already-used");
--- a/generators/server/templates/src/main/java/package/web/rest/errors/ExceptionTranslator.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/errors/ExceptionTranslator.java.ejs
@@ -149,17 +149,6 @@ _%>
             .build();
         return create(ex, problem, request);
     }
-    <%_ if (!reactive) { _%>
-
-    @ExceptionHandler
-    public ResponseEntity<Problem> handleNoSuchElementException(NoSuchElementException ex, NativeWebRequest request) {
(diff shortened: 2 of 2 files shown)


xetys:
I don't get this one correctly. Maybe I'm using it wrong, but when I do some specific logic somewhere in the service layer and find that the user passed a missing entity, I just throw a `NoSuchElementException` as I know, the translator will make a proper HTTP response. Otherwise I should handle it manually in the REST controller...

If you remove that handling, it would be no longer possible to do so...right?

cbornet:
NoSuchElementException is a generic exception used in Optionals but also in a lot of other classes. It has no meaning at all that an entity is missing and that this error comes from validation of REST client input (which is what 404 means).
I just had the case where an Optional was empty because one of my service couldn't get information of a Kafka cluster. This is clearly a server error and shouldn't result in a 404.
So yes, you need to give meaning to the fact that the Optional is empty by translating it to another exception. If you want to translate to an HTTP error code globally, you should define some EntityMissingException with an ExceptionHandler mapping it to 404. Then you can do maybeEntity.orElse(() -> throw new EntityMissingException());
Note that we currenlty wrap entity missing by writing directly to the ResponseEntity. It would probably be better to throw this kind of Exception and let the ExceptionHandler translate to a 404 HTTP-Problem.

xetys:
I get your point...I just find it very useful, that we have this default handler for `NoSuchElementException` which a user can use out-of-the-box without adding new translators. 

I think the question is, how many people are using this the way I do?

atomfrede:
Maybe we can keep the mapping but instead of doing a 404 we map to a 500?

cbornet:
That's the default if you remove the mapping 🙂

xetys:
I had it yesterday when doing building a new microservice. Maybe I'm misusing the mapping, I dunno, but this is what I'm doing quite often:

Say, I have an entity `SomeEntity` and I want to add a user to its N:1 relation.

```java
    @Transactional
    public void addUserToSomeEntity(Long entityId, String userLogin) {
        SomeEntity entity = someEntityService.findOne(entityId) // returns Optional
            .orElseThrow(NoSuchElementException::new);
            
            // proceed with other custom logic
     }
```

as I know that we have a mapping for this, I don't bother about telling the rest-controller how to handle this and throw exceptions or whatever, I just can

```java
@PostMapping(...)
public ResponseEntity<Void> addUserToSomeEntity(@PathVariable Long id, @PathVariable String userLogin) {
    service.addUserToSomeEntity(id, userLogin);
    return ResponseEntity.noContent().build();
}
```

AFAIK, if you remove the mapping, my code would result in a 500 status, what is not what I want, as there is simply the entity missing, and no server error.


But again, if I'm doing bad practice, feel free to correct me :)

cbornet:
NoSuchElementException is generic and must be treated as a generic exception. Here you should throw EntityNotFoundException instead and have a mapping for it to 404 in ExceptionHandler.

atomfrede:
I would also say let's remove the mapping. I guess sonar will also complain when you throw a Nosuchelementexception in your code

cbornet:
We could create the EntityNotFoundException and it's mapping then use it in our endpoints instead of wrapOrNotFound. This way we would have better errors witrh Problem details instead of the current which is a simple 404 without body.

xetys:
Yeah, let's just create a new own exception and set the handler to it.

cbornet:
BTW you can already do 
```java
someEntityService.findOne(entityId)
    .orElseThrow(() -> new ResponseStatusException(HttpStatus.NOT_FOUND));
```
It will be correctly translated to 404.

cbornet:
Any objection to merge ?

atomfrede:
I have none.

Issue #10733, opened 2019-11-06
Better UI for not logged in and authentication expired users in Angular


kaidohallik:
##### **Overview of the feature request**

This is follow up to PR-s #10672 (corresponding issue is #10630) and #10675 (corresponding issue is #10551).  
PR for resolving this issue can be done after #10672 and #10675 are merged.

After above referenced 2 PR-s the logic for "not logged in" / "authentication expired" users in session and JWT authentication is:
* remember "requested" / "last seen" page
* route to home page
* open login modal
* on successful login route to "requested" / "last seen" page

In issues and PR-s referred above was proposed / discussed some enhancements. This issue is for discussion of these enhancements.

Description of possible enhancement:

* instead of routing to home page, route to https://github.com/jhipster/generator-jhipster/tree/master/generators/client/templates/angular/src/main/webapp/app/layouts/error where new special error sections with the following content will be added:
    * new error messages (translation keys are going to `error.json`)
        * `error.notLoggedIn`: **You are not logged in.**
        * `error.authenticationExpired`: **Authentication expired.**
    * information where user is routed after logging in in the form (translation can go to the `error.json` with the key `error.redirectUrl`): **After logging in you are routed to {{ redirectUrl }}**
    * button "Log in"

* don't open login modal by default, user can open login modal by clicking "Log in"

* if user navigates away from this error page without logging in then clear remembered "requested" / "last seen" page, because navigating away without logging in can be considered as cancellation of routing to "requested" / "last seen" page after logging in

##### **Motivation for or Use Case**

Improve UI for not logged in and authentication expired users.

##### **Related issues or PR**

#10672 #10630 #10675 #10551 #9800

-   [x] Checking this box is mandatory (this is just to show you read everything)

mraible:
There's a couple of PRs that've been merged related to this issue. Can we close it, or is there still work that needs to be done?

kaidohallik:
In the current main branch this is not resolved. But I think current open PR #11926 will resolve this issue. So this issue can be closed after #11926 is merged.

DanielFran:
Closed with #11926

PR #10835, opened 2019-11-27
Strict option in blocks folder in Angular

kaidohallik:
This is the next piece for the #10631 - blocks folder.

Some comments below.

https://angular.io/guide/http#getting-error-details says: _The HttpClient captures both kinds of errors in its HttpErrorResponse and you can inspect that response to figure out what really happened._  
So `HttpClient` error is always `HttpErrorResponse` and no need to check is this indeed `HttpErrorResponse`.  
`HttpErrorResponse` can contain HTTP error response or `ErrorEvent` in `HttpErrorResponse.error` (then `HttpErrorResponse.status` is `0`).

NotificationInterceptor: `HttpHeaders.get` is returning `string | null` - so no need to check `if (typeof alert === 'string')`, checking `if (alert)` is enough.

This PR has 2 commits:
* first one https://github.com/jhipster/generator-jhipster/commit/476b5e16314c0869f68791b4d0010553f2e9fc1d resolves strict option problems
* second one https://github.com/jhipster/generator-jhipster/commit/2fe069384e9a241e6589560b27083b72e470f174 improves UAA auth expired and not logged in users behavior to match how it's working with other authentication types, this also simplifies templates and if #10733 will be resolved then the same solution will be work also for UAA after these changes

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/auth-expired.interceptor.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/auth-expired.interceptor.ts.ejs
@@ -26,45 +26,29 @@ import { LoginService } from 'app/core/login/login.service';
 <%_ if (['session', 'jwt', 'uaa'].includes(authenticationType)) { _%>
 import { LoginModalService } from 'app/core/login/login-modal.service';
 <%_ } _%>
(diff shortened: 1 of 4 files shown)


kaidohallik:
As coming layouts PR is related to these changes then I'm merging this.

Issue #10840, opened 2019-11-27
Upgrade with pt_BR localization is broken


mshima:
##### **Overview of the issue**

Upgrading from jhipster 5.8.1 to 6.5.1 results in error with pt_BR localization.

PR https://github.com/jhipster/generator-jhipster/commit/492707876f4a6961b5ab2797818a5f7e5192c661 changed localization file name from 'xx_yyyy_zz to xx_yyyy_ZZ to match java locales'.

The following error is shown at git merge step:

```
error: The following untracked working tree files would be overwritten by checkout:
	src/main/resources/i18n/messages_pt_BR.properties
Please move or remove them before you switch branches.
```

##### **Reproduce the error**

- Create a project with server files and pt-br localization with jhipster < 6.1.0
- Upgrade it to last jhipster version

##### **Suggest a Fix**

Rename files on upgrade.

##### **Browsers and Operating System**

macOS

-   [x] Checking this box is mandatory (this is just to show you read everything)

murdos:
Seems to be very similar to #9908

pascalgrimaud:
yes it's similar to #9908 but I don't know if we can do something

PierreBesson:
Well for me it's not a bug. Yes it's kind of a breaking change  but this was the tradeoff to be able to support all locales. Especially Uzbek which can be written with 2 different alphabets. Actually it's fascinating that Uzbek has been using 3 different scripts for its language in the 20th century (arabic, cyrillic and latin). The cyrillic to latin migration is still in progress today: https://en.wikipedia.org/wiki/Uzbek_alphabet

PR #10842, opened 2019-11-28
Migrate problematic localization files on upgrade.

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/10840

This PR will rename the files and create a commit before the generation with the new jhipster version.

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -512,6 +512,7 @@ module.exports = class extends BaseBlueprintGenerator {
         return {
             cleanup() {
                 cleanup.cleanupOldFiles(this);
+                cleanup.upgradeFiles(this);
             },
 
             regenerateEntities() {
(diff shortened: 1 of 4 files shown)


PierreBesson:
Honestly, this looks like a big hack. I don't think we should add scripts like this. I have no ideas if it will effectively work in all possible use cases. Especially commiting on the user's befalf is strange (what if they are on their master branch.

mshima:
> Honestly, this looks like a big hack. I don't think we should add scripts like this.

I really don't see too much difference to what is done now.

> I have no ideas if it will effectively work in all possible use cases.

That is really hard to know, but the logic seems right.

> Especially commiting on the user's befalf is strange (what if they are on their master branch.

Upgrade does the following:

1. creates an empty branch called jhipster_upgrade and checkout
2. generate the project using the old jhipster version (commits)
3. do some merge to master (really don't know how this works).
4. remove every file that isn't control file (git).
5. generate the project using the new jhipster version (commits).
6. do some merge to master (don't know how this works too, but works).

This patch creates a commit between steps 3 and 4 that renames the problematic files.

This commit is needed?
Yes.

Why?
Step 4 removes every files and git looses reference to the problematic files.

PierreBesson:
@mshima you look confident enough so I trust you. Personally I don't have a "real life" jhipster app to help me test the upgrade subgen on it. I know how the upgrade subgen is supposed to work, it's just I'm not confident enough users are not going to screw themselves up with it.

In the end, I'm OK with this PR, please just add some big warning logs when it commits on the user's behalf.

PR #10857, opened 2019-12-02
Strict Typescript option in entities in Angular

kaidohallik:
Entities module for #10631 

Some notes about changes:

* replaced `any` with strict type in many places

* as I defined `IPagination` for entities query then I used this also in audits and user-management service

* if filtering is used for entity then ATM I left the type `any` in the service query parameter, but if someone is interested then it would be cool to have correct type declaration also for that case, maybe it's possible to use some generator, like `swagger-ts-generator` or something like that to generate filter definition instead of coding ourselves

* `onError` in entity update used `alertService.error`, but this doesn't make sense because `alertService.error` is not showing error messages, this just constructs `JhiAlert` and nothing more, now added new message `<%=angularAppName%>.error` for `JhiEventManager` and `AlertErrorComponent` subscribes to that and shows error messages
    * and as HTTP errors are handled by `ErrorHandlerInterceptor` then I removed `onError` from HTTP failure to avoid double error messages and broadcasting error only on file load error

* removed `pager` conditions from entities templates as pager was removed

* if in navigation to entity detail or update page entity is not found by id then server returns HTTP response 404 and `main.component.ts` is routing to 404 error page, but as `HttpClient` `get` method can also return `null` value by it's declaration then this case was needed to handle in strict option and I resolved this by routing also to this 404 error page (note: this should never happen in JHipster application as server is responsing HTTP 404 in that case, but if user changes that behavior in the back end then front end code is now prepared for that change)

* update form: `clearInputImage` third parameter was wrong, constant string `fileImage`, now passing correct field name and file input is now cleared

* removed some repeated code:
    * `trackById` is now declared only once in the update forms
    * elasticsearch: `clear()` was basically copy of `search('')` so removed `clear()`
    * defined `getTypescriptKeyType` in `generator-base-private.js` instead of copy-paste the same conditions into several places

* removed some unused code: in update form selects from related entities in **.html** were generated for `OneToOne` and `ManyToMany` only if `ownerSide=true`, but in **.ts** also for `ownerSide=false` - now this redundant code is not generated any more into **.ts** files

* update form, follow up to #10383: replaced `Promise` based `setFileData` content with `Observable` based `JhiDataUtils.loadFileToForm` from the new `ng-jhipster v0.11.5`

* problem with `trackId` return type in the main entity component: as in entity model is `id` declared as optional then strict option blocks returning just `number` or `string`, possible solutions (ATM I chose 2.b):
    1. add `| undefined` to return type, for example `number | undefined`, but this is strange, because in data loaded from database `id` is always defined
    2. add type assertion `!`, but this causes eslint rule `@typescript-eslint/no-unnecessary-type-assertion` failure, possible solutions:
        a. turn this rule off for whole app in `.eslintrc`
        b. turn this rule off only for the line in the `trackId` method where this rule causes errors
    3. use `any` as return type

* in `generateEntityQueries` was decided how to query related `OneToOne` entity service by `jpaMetamodelFiltering` of the owner entity, but this is wrong, query format depends on `jpaMetamodelFiltering` of the related entity (as this related entity is queried), this is now fixed


kaidohallik:
I made 2 additional commits.

* First one removed strict type from entity services `query` method as this can't be done before #10858 is resolved. If #10858 is resolved somehow then strict type can be added with separate PR.

* Second one:

    * added strict type for `<%=angularAppName%>.error` message in `AlertErrorComponent` and then it was logical resolve all strict option problems also in this `AlertErrorComponent`. I also removed prefix from AlertErrorComponent as per #10220 

    * polished imports order in entities

generators/client/templates/angular/src/main/webapp/app/shared/alert/alert-error.component.ts.ejs
```diff
@@ -106,21 +112,24 @@ export class <%=jhiPrefixCapitalized%>AlertErrorComponent implements OnDestroy {
         });
     }
 
-    setClasses(alert) {
-        return {
-            'jhi-toast': alert.toast,
-            [alert.position]: true
-        };
+    setClasses(alert: JhiAlert): { [key: string]: boolean } {
+        const classes = { 'jhi-toast': !!alert.toast };
```
wmarques: Boolean(alert.toast)

generators/client/templates/angular/src/main/webapp/app/shared/alert/alert-error.component.ts.ejs
```diff
@@ -106,21 +112,24 @@ export class <%=jhiPrefixCapitalized%>AlertErrorComponent implements OnDestroy {
...
+        if (alert.position) {
+            return { ...classes, [alert.position]: true };
+        }
+        return classes;
     }
-
-    ngOnDestroy() {
-        if (this.cleanHttpErrorListener !== undefined && this.cleanHttpErrorListener !== null) {
-            this.eventManager.destroy(this.cleanHttpErrorListener);
-            this.alerts = [];
```
wmarques: Why aren't we doing this anymore ?

kaidohallik: Assigning value to class member on destroy doesn't make sense. I believe that the idea was as I guessed in https://github.com/jhipster/generator-jhipster/pull/10901#issue-350746537 about `alert.component` `ngOnInit()`.

But `alert-error.component` actually is not tied to `JhiAlertService`, so this is really redundant, even `alertService.clear()` is not needed.

wmarques: Oh yes I didn't see the ngOnDestroy three lines up... Sorry 🤦‍♂

generators/client/templates/angular/src/main/webapp/app/shared/alert/alert-error.component.ts.ejs
kaidohallik: Thanks for pointing! Done.

PR #10861, opened 2019-12-02
Use stable dependency

jdubois:
Use a stable version of the Azure Spring Cloud dependency


atomfrede:
Is there already gradle support by the way?

jdubois:
Sorry @atomfrede but I would need your help here! It's only a matter of adding a JAR file, like for Maven, so very simple.
Then I'm going to do the same for Azure App Service, but we won't be able to use Gradle as there is no Gradle plugin yet (I've already pushed for it, of course)

PR #10865, opened 2019-12-03
Strict option in user management in Angular

kaidohallik:
User management module for #10631 

Some comments about changes:

* https://github.com/jhipster/generator-jhipster/pull/10383/files#diff-fbe805378490e83898dff42879a82b67L42-R43 changed `userService.find` return type from `Observable<HttpResponse<IUser>>` to `Observable<IUser>` so in detail and update component `ngOnInit` no need to check for `user.body` any more

* `activated` in update form: if reading code then can see that goal is to set `activated=true` in adding new user and disable changing that in update form (because back end forces `activated=true` if adding a new user) but in my testing either didn't work, so I adjusted code to make it work as intended

* http errors were not shown in user list component, added `<<%= jhiPrefixDashed %>-alert-error></<%= jhiPrefixDashed %>-alert-error>` and http errors are shown now

* `ActivatedRoute.data` observable doesn't need unsubscribe (https://angular.io/guide/router#observable-parammap-and-component-reuse). As this unsubscribe is not done in entities and was removed also from audits module then removed also from user management module, to follow the same style and to make code shorter

* in the `setActive` method in user list component I changed logic: if error occurs in saving (`activated` is not changed in the database) then restoring initial activated value (currently, if error occurs then UI shows as data was changed, shows wrong value)

* removed redundant `onError` method:
    * errors are shown by `AlertErrorComponent` now
    * in current master parameter to `onError` is `HttpErrorResponse.body` but as `HttpErrorResponse` misses property `body` then this throws constantly error: cant't read property error of undefined
    * in `onError` function is runned `this.alertService.error` - but this doesn't show error, this just constructs `JhiAlert` and that's all, so this wouldn't work if fixing parameter to `onError` and no need to duplicate `AlertErrorComponent`

* if adding or changing the user then the page title was `translation-not-found...` - this is fixed now

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-delete-dialog.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/user-management/user-management-delete-dialog.component.html.ejs
@@ -16,7 +16,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<form name="deleteForm" (ngSubmit)="confirmDelete(user.login)">
+<form *ngIf="user" name="deleteForm" (ngSubmit)="confirmDelete(user.login)">
     <div class="modal-header">
         <h4 class="modal-title" jhiTranslate="entity.delete.title">Confirm delete operation</h4>
         <button type="button" class="close" data-dismiss="modal" aria-hidden="true"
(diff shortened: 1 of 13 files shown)


kaidohallik:
I improved `setActive`: now `activated` is not changed before saving in the UI but only after successful saving.

kaidohallik:
I'm merging this as I need this for layouts PR.

PR #10873, opened 2019-12-04
Strict Typescript option in metrics module in Angular

kaidohallik:
Metrics module for #10631 

Some notes:
* removed prefix from component and service name and changed component name to standard according to #10220 
* FYI: metrics endpoint code is here: https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/config/metric/JHipsterMetricsEndpoint.java#L45-L65
* FYI: nice JSON to Typescript conversion tool: https://app.quicktype.io/

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.component.ts.ejs
@@ -16,21 +16,22 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import {Component, OnInit} from '@angular/core';
+import { Component, OnInit, ChangeDetectionStrategy, ChangeDetectorRef } from '@angular/core';
 import { flatMap } from 'rxjs/operators';
 
 import { MetricsService, Metrics, MetricsKey, ThreadDump, Thread } from './metrics.service';
 
 @Component({
     selector: '<%= jhiPrefixDashed %>-metrics',
-    templateUrl: './metrics.component.html'
+    templateUrl: './metrics.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class MetricsComponent implements OnInit {
     metrics?: Metrics;
     threads?: Thread[];
     updatingMetrics = true;
 
-    constructor(private metricsService: MetricsService) {}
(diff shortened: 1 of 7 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.service.ts.ejs
```diff
@@ -22,16 +22,21 @@ import { Observable } from 'rxjs';
 
 import { SERVER_API_URL } from 'app/app.constants';
 
+export type MetricsKey = 'jvm' | 'http.server.requests' | 'cache' | 'services' | 'databases' | 'garbageCollector' | 'processMetrics';
+export type Metrics = { [key in MetricsKey]: any };
+export type Thread = any;
```
wmarques: That's impossible to type ?

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.component.ts.ejs
```diff
@@ -17,46 +17,47 @@
  limitations under the License.
 -%>
 import {Component, OnInit} from '@angular/core';
+import { flatMap } from 'rxjs/operators';
 
-import { <%=jhiPrefixCapitalized%>MetricsService } from './metrics.service';
+import { MetricsService, Metrics, MetricsKey, ThreadDump, Thread } from './metrics.service';
 
 @Component({
```
wmarques: Can't we use the OnPush here ?
I see that you used functions inside `ngIf` so it would be much better in terms of performance to use the OnPush strategy here

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.service.ts.ejs
kaidohallik: It's possible, but as this is used inside `ng-jhipster` then I decided not to do that to keep code light, not to pollute code with unused information.

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics.component.ts.ejs
kaidohallik: Switched to OnPush now

PR #10883, opened 2019-12-05
[Feature Request] Enable autoprefixer for component specific styles

andreasfranz:
Treat component specific SASS styles the same way as vendor and global styles using the autoprefixer.


wmarques:
Good catch @andreasfranz thanks !

Issue #10887, opened 2019-12-05
New project (6.5.1) w/ postgresql, liquibase:diff won't find FixedPostgreSQL95Dialect and liquibase:diff changelog output fails to execute


nuba:
##### **Overview of the issue**

In a clean, new, project, using postgresql, liquibase won't find the custom jhipster postgresql driver dialect, and switching that driver results in a liquibase:diff output that fails to run. 

This is partially similar to #10885, created earlier today.

##### **Motivation for or Use Case**

I expect a newly created project to output no changes with liquibase:diff, and also expect the custom jhipster postgresql dialect driver to be found.

##### **Reproduce the error**

I created a repository each step in a commit: https://github.com/nuba/bug-report-jhipster-liquibase

Error messages are

```
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:3.6.3:diff (default-cli) on project jhipster: Error setting up or running Liquibase: liquibase.exception.UnexpectedLiquibaseException: liquibase.exception.DatabaseException: java.lang.ClassNotFoundException: io.github.jhipster.domain.util.FixedPostgreSQL95Dialect -> [Help 1]
```
And

```
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:3.6.3:update (default-cli) on project jhipster: Error setting up or running Liquibase: Migration failed for change set config/liquibase/changelog/20191205215152_changelog.xml::1575582718315-1::nuba (generated):
[ERROR]      Reason: liquibase.exception.DatabaseException: ERROR: constraint "uc_jhi_useremail_col" of relation "jhi_user" does not exist [Failed SQL: ALTER TABLE jhi_user DROP CONSTRAINT UC_JHI_USEREMAIL_COL]
[ERROR] -> [Help 1]
```

Here is the table definition in postgres

```
bugreport=# \d jhi_user;
                              Table "public.jhi_user"
       Column       |            Type             | Collation | Nullable | Default 
--------------------+-----------------------------+-----------+----------+---------
 id                 | bigint                      |           | not null | 
 login              | character varying(50)       |           | not null | 
 password_hash      | character varying(60)       |           | not null | 
 first_name         | character varying(50)       |           |          | 
 last_name          | character varying(50)       |           |          | 
 email              | character varying(191)      |           |          | 
 image_url          | character varying(256)      |           |          | 
 activated          | boolean                     |           | not null | 
 lang_key           | character varying(10)       |           |          | 
[...]

PR #10890, opened 2019-12-06
Added dependency to pom.xml templates so liquibase can find jhipster'…

nuba:
As requested by @pmverma on #10887 , this PR makes it possible for liquibase, in a maven-based project, to use jhipster's custom dialects.

I also created a new project with gradle and repeated the same steps mentioned in that issue. Liquibase with gradle had no problems finding the dialect, so I didn't change anything there.


pascalgrimaud:
thanks and congrats for your 1st contribution @nuba !

PR #10893, opened 2019-12-07
Strict Typescript option in layouts in Angular

kaidohallik:
This is layouts folder for #10631

Some comments.

* Code for redirecting to stored url and related tests were in 2 different places (for oauth2 in `main.component` and for others in `login.component`), now moved redirect code and related tests to `AccountService.identity`, this:
    * simplified generated code
    * simplified templates
    * removed duplicated code and duplicated tests
    * and to me seems that `account.service` is more logical place for that redirect logic

* Page title setting code was duplicated in 2 different places (for i18n in `language.helper` and for no i18n in `main.component`), now this code is only in `main.component` and I added also full unit tests to page title setting and these unit tests found errors in page title setter and fixed them and made some improvements:
    * if `pageTitle` was missing then in i18n enabled was searched translation for key `<%= angularAppName %>` but such key doesn't exist and page title was "translation-not-found...", changed that:
        * for i18n enabled: `global.title`
        * for i18n disabled: `<%= capitalizedBaseName %>`
    * if parent route had `pageTitle` in route data and child route had not then due to error in logic was used `<%= angularAppName %>`, now is used `pageTitle` from parent route data

* As changes touched `language.helper` then removed also `getAll()` from there - this returned `LANGUAGES` constant, now using this directly as all other constants are used (without getter around the constant).

* As `login-modal.service` from `core` and `login.component` from `shared` were related then cleaned also these.

* According to #10220 removed prefix from `MainComponent`, `LoginModalComponent` and `LanguageHelper`

* `errorMessage` in `error.component` was unused so removed that to avoid confusion in code readers

* in `profile.service` was declared that profile info coming from server is `ProfileInfo` but this is not true, so declared correct model `InfoResponse` for server response

* using `ViewChild` and removed `setTimeout` from login modal username focus - working fine

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/error/error.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/error/error.component.ts.ejs
@@ -57,7 +57,10 @@ export class ErrorComponent implements OnInit<% if (enableTranslation) { %>, OnD
     <%_ if (enableTranslation) { _%>
 
     private getErrorMessageTranslation(): void {
-        this.translateService.get(this.errorKey).subscribe(translatedErrorMessage => this.errorMessage = translatedErrorMessage);
+        this.errorMessage = '';
+        if (this.errorKey) {
+          this.translateService.get(this.errorKey).subscribe(translatedErrorMessage => (this.errorMessage = translatedErrorMessage));
+        }
     }
 
     ngOnDestroy(): void {
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/error/error.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/error/error.component.ts.ejs
@@ -16,34 +16,54 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit<% if (enableTranslation) { %>, OnDestroy<% } %> } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
+<%_ if (enableTranslation) { _%>
+import { Subscription } from 'rxjs';
+import { TranslateService } from '@ngx-translate/core';
+<%_ } _%>
 
 @Component({
     selector: '<%= jhiPrefixDashed %>-error',
     templateUrl: './error.component.html'
 })
-export class ErrorComponent implements OnInit {
-    errorMessage: string;
-    error403: boolean;
-    error404: boolean;
+export class ErrorComponent implements OnInit<% if (enableTranslation) { %>, OnDestroy<% } %> {
+    errorMessage?: string;
+    <%_ if (enableTranslation) { _%>
+    errorKey?: string;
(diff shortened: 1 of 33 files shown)


kaidohallik:
As this PR fixed extracting page title from parent route then we can now use this functionality. I added commit for user management where removed 4 identical `pageTitle` from child routes and defined this once in parent route.

I think we should use this also for entities. ATM I couldn't do that because it would cause conflicts with #10857

kaidohallik:
I made 2 additional changes:

* initially I removed general `errorMessage` as unused from `error.component` but I didn't like that and instead I removed booleans for 403 and 404 and implemented these using general `errorMessage`
    * while implementing this I discovered that in function `replaceTitle` in `utils.js` was used `global.json` if key starts with `error` but in fact error keys are both in `global.json` and `error.json`, so I implemented concatenation of these translations if error key starts with `error` and this also fixed error page title if i18n is disabled, now is correct title shown

* I reintroduced strange state in browser what was fixed by #9119 - now restored this functionality
    * to keep Git history clean I now rebased to 1 commit and force pushed changes

generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.ts.ejs
```diff
@@ -16,96 +16,58 @@
...
+        let title: string = (routeSnapshot.data && routeSnapshot.data['pageTitle']) ? routeSnapshot.data['pageTitle'] : '';
         if (routeSnapshot.firstChild) {
             title = this.getPageTitle(routeSnapshot.firstChild) || title;
         }
         return title;
     }
 
-    ngOnInit() {
-        this.router.events.subscribe((event) => {
+    private updateTitle(): void {
```
wmarques: Can we move the private functions under public ones ?

generators/client/templates/angular/src/main/webapp/app/layouts/error/error.component.ts.ejs
```diff
@@ -16,34 +16,54 @@
...
-            if (routeData.error403) {
-                this.error403 = routeData.error403;
-            }
-            if (routeData.error404) {
-                this.error404 = routeData.error404;
-            }
             if (routeData.errorMessage) {
+                <%_ if (enableTranslation) { _%>
+                this.errorKey = routeData.errorMessage;
+                this.setErrorMessage();
```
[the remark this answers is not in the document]
kaidohallik: Done

generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.ts.ejs
kaidohallik: Done

kaidohallik:
After checking this I found that I introduced strict error while refactored error component to general errorMessage. Latest commit fixes that.

kaidohallik:
While enabling strict option in general test logic I found that while enabling strict option in `mock-route.service` in this PR I forgot to add return types to functions. Latest commit fixes that.

PR #10897, opened 2019-12-09
Improve randexp usage in test data generation

kaidohallik:
This is follow up to #10892 and includes:

* using functionality provided by #10892 generates consistent result also for front end tests pattern fields

* improves #10892 for back end tests pattern fields:
    * after #10892 `sampleTextString` and `updatedTextString` were identical and constantly was shown warning in the console that sample and update strings are equal and adding B at the end of `updatedTextString`
    * now different consistent values are generated for `sampleTextString` and `updatedTextString`

* while testing this PR found that fake data was not generated for pattern fields if pattern field doesn't allow `"`, this is now fixed, fake data is now generated even if pattern doesn't allow `"`

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -270,7 +270,11 @@ function addEnumerationFiles(generator, templateDir, clientFolder) {
 function addSampleRegexTestingStrings(generator) {
     generator.fields.forEach(field => {
         if (field.fieldValidateRulesPattern !== undefined) {
-            field.fieldValidateSampleString = new Randexp(field.fieldValidateRulesPattern).gen();
+            const randExp = new Randexp(field.fieldValidateRulesPattern);
+            randExp.max = 5;
+            // In order to have consistent results with RandExp, the RNG is seeded.
+            randExp.randInt = generator.seededRandomNumberGenerator(10);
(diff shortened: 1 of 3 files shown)


kaidohallik:
First run of the react-maven-couchbase-caffeine failed, second was successful, but GitHub shows that checks are not successful...

mshima:
> Changes look good to me, thanks for fixing my PR. Did not test manually.

seededRandomNumberGenerator seems unnecessary for me.
Is there a reason to not use `faker.random.number({min: ??, max: ??})` as seed?

ruddell:
I didn't think to integrate Faker with RandExp, if it works then that seems simpler (less code in our codebase).  

The issue I see with Faker is that [the seed is set globally](https://github.com/Marak/faker.js/blob/f0b3aaf87c0ba2f855322a9084e4e1522295e7c5/build/src/docs.md#setting-a-randomness-seed).  This means if the entity generation order is changed, the regex-compliant values will change in the generated files.  Also if you set the seed based on row number, you'll have the same regex generated for all entities.

mshima:
Using faker instead of seededRandomNumberGenerator can be done on another PR.

DanielFran:
@mshima OK, so let's merge this.

PR #10901, opened 2019-12-09
Strict Typescript option in shared module in Angular

kaidohallik:
This is shared module for #10631

Some notes:

* `alert.component.ts` - in `ngOnDestroy()` was tried to clear messages by `this.alerts = [];` but this doesn't clear messages and the same success message is shown in another form if user quickly navigates to other form, substituted this with `this.alertService.clear();` and now are messages cleared and not shown in another form

* removed memory leak from `has-any-authority.directive.ts`, this happened because not unsubscribed from observable:
    * this occurred if using this directive in some component which can created multiple times while navigating in and out, for example if adding to home component: `<p *jhiHasAnyAuthority="'ROLE_ADMIN'">Only for admin</p>`
    * in generated code this is used only in navbar so if user didn't use this in any other place then no memory leak

* in` datepicker-adapter` added `!` to `null` return to suppress compiler errors, this can be removed after this is resolved: https://github.com/ng-bootstrap/ng-bootstrap/issues/1544

* according to #10220 removed prefix from `AlertComponent`

* login folder is cleaned by layouts PR and request-util is cleaned by entities PR

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
@@ -17,7 +17,7 @@
  limitations under the License.
 -%>
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { JhiAlertService } from 'ng-jhipster';
+import { JhiAlertService, JhiAlert } from 'ng-jhipster';
 
 @Component({
     selector: '<%= jhiPrefixDashed %>-alert',
@@ -30,24 +30,25 @@ import { JhiAlertService } from 'ng-jhipster';
             </div>
         </div>`
 })
-export class <%=jhiPrefixCapitalized%>AlertComponent implements OnInit, OnDestroy {
-    alerts: any[];
+export class AlertComponent implements OnInit, OnDestroy {
+    alerts: JhiAlert[] = [];
 
     constructor(private alertService: JhiAlertService) { }
 
-    ngOnInit() {
+    ngOnInit(): void {
         this.alerts = this.alertService.get();
     }
 
-    setClasses(alert) {
-        return {
-            'jhi-toast': alert.toast,
-            [alert.position]: true
-        };
-    }
-
-    ngOnDestroy() {
-        this.alerts = [];
+    setClasses(alert: JhiAlert): { [key: string]: boolean } {
+        const classes = { 'jhi-toast': !!alert.toast };
+        if (alert.position) {
+          return { ...classes, [alert.position]: true };
+        }
+        return classes;
+      }
+    
+    ngOnDestroy(): void {
+        this.alertService.clear();
     }
 
 }
--- a/generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
@@ -40,7 +40,7 @@ export class AlertComponent implements OnInit, OnDestroy {
     }
 
     setClasses(alert: JhiAlert): { [key: string]: boolean } {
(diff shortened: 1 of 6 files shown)


generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
```diff
@@ -30,24 +30,25 @@ import { JhiAlertService } from 'ng-jhipster';
...
-        return {
-            'jhi-toast': alert.toast,
-            [alert.position]: true
-        };
-    }
-
-    ngOnDestroy() {
-        this.alerts = [];
+    setClasses(alert: JhiAlert): { [key: string]: boolean } {
+        const classes = { 'jhi-toast': !!alert.toast };
```
wmarques: Instead of `!!alert.toast` I think you can write `Boolean(alert.toast)` , I find it more readable WDYT ?

generators/client/templates/angular/src/main/webapp/app/shared/language/find-language-from-key.pipe.ts.ejs
```diff
@@ -20,15 +20,15 @@ import { Pipe, PipeTransform } from '@angular/core';
...
+    private languages: { [key: string]: { name: string, rtl?: boolean } } = {
         // jhipster-needle-i18n-language-key-pipe - JHipster will add/remove languages in this object
     };
     transform(lang: string): string {
         return this.languages[lang].name;
     }
     <%_ if (enableI18nRTL) { _%>
     isRTL(lang: string): boolean {
-        return this.languages[lang].rtl;
+        return !!this.languages[lang].rtl;
```
wmarques: Boolean(this.languages[lang].rtl)

generators/client/templates/angular/src/main/webapp/app/shared/util/datepicker-adapter.ts.ejs
```diff
@@ -30,10 +30,12 @@ export class NgbDateMomentAdapter extends NgbDateAdapter<Moment> {
...
             return { year: date.year(), month: date.month() + 1, day: date.date() };
         }
-        return null;
+        // ! can be removed after https://github.com/ng-bootstrap/ng-bootstrap/issues/1544 is resolved
+        return null!;
     }
 
     toModel(date: NgbDateStruct): Moment {
-        return date ? moment(date.year + '-' + date.month + '-' + date.day, 'YYYY-MM-DD') : null;
+        // ! can be removed after https://github.com/ng-bootstrap/ng-bootstrap/issues/1544 is resolved
```
wmarques: I don't understand the comment, I don't see any `!`

generators/client/templates/angular/src/main/webapp/app/shared/alert/alert.component.ts.ejs
kaidohallik: I agree, made change, thanks for pointing out!

generators/client/templates/angular/src/main/webapp/app/shared/language/find-language-from-key.pipe.ts.ejs
kaidohallik: Done

generators/client/templates/angular/src/main/webapp/app/shared/util/datepicker-adapter.ts.ejs
kaidohallik: This is at the end after `null`, I modified comment to point out that this is after `null`

wmarques:
Thanks @kaidohallik ! :)

kaidohallik:
@wmarques thanks for those reviews and thanks for merging!

wmarques:
@kaidohallik Well it took time but I finally managed to have some time to review all those lines, hope we haven't missed anything 🙏

PR #10908, opened 2019-12-10
Azure App Service support

jdubois:
Adds support to deploy applications to Azure App Service

- Works like the Azure Spring Cloud sub-generator for consistency
- Allow to deploy locally or using GitHub Actions
- Adds support for Azure Application Insights for monitoring


generators/azure-app-service/templates/application-azure.yml.ejs
```diff
@@ -0,0 +1,35 @@
...
+# This configuration overrides the application.yml file.
+# ===================================================================
+
+# ===================================================================
+# Standard Spring Boot properties.
+# Full reference is available at:
+# http://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html
+# ===================================================================
+
+<%_ if (azureAppInsightsInstrumentationKey != '') { _%>
```
MathieuAA: !== (there should be a warning for this, amazing it never shows)

generators/generator-base.js
```diff
@@ -1838,6 +1838,33 @@ module.exports = class extends PrivateBase {
         return child;
     }
 
+    /**
+     * run a command using the configured Java build tool.
+     *
+     * @param {String} buildTool - maven | gradle
+     * @param {String} profile - dev | prod
+     * @param {Boolean} command - the command (goal/task) to run
```
MathieuAA: a boolean? not a string?

generators/generator-base.js
```diff
@@ -1838,6 +1838,33 @@ module.exports = class extends PrivateBase {
...
+     * @param {Function} cb - callback when build is complete
+     */
```
MathieuAA: there's a `@return` missing

generators/generator-base.js
```diff
@@ -1838,6 +1838,33 @@ module.exports = class extends PrivateBase {
...
+        if (os.platform() !== 'win32') {
+            buildCmd = `./${buildCmd}`;
+        }
+        buildCmd += ` -P${profile}`;
+        this.log(`Running command: '${chalk.bold(buildCmd)}'`);
+        const child = {};
+        child.stdout = exec(buildCmd, { maxBuffer: 1024 * 10000 }, cb).stdout;
+        child.buildCmd = buildCmd;
+
+        return child;
```
MathieuAA: simpler:
```js
return {
  stdout: exec(buildCmd, { maxBuffer: 1024 * 10000 }, cb).stdout,
  buildCmd
}
```

generators/azure-app-service/templates/application-azure.yml.ejs
jdubois: Thanks!

generators/generator-base.js
jdubois: Wrong copy/paste!! Thanks again

generators/generator-base.js
jdubois: Same issue with the method above, I corrected both

generators/generator-base.js
jdubois: Thanks for the tip! I'm also using this on the method above, as it was copy/pasted from there

generators/azure-app-service/templates/application-azure.yml.ejs
clementdessoude: FWIW, there is the eqeqeq rule in eslint which does that :)

jdubois: OK but it does not seem to work on this file, or maybe we didn't configure it correctly

MathieuAA:
Thanks @jdubois

PR #10909, opened 2019-12-11
Strict Typescript option in health module in Angular

kaidohallik:
This is a health module for the #10631 

Health checking API in Spring Boot changes in version 2.2, new API version V3 will be used, ATM in Spring Boot 2.1 we are using API version V2:
* Spring Boot 2.1: https://docs.spring.io/spring-boot/docs/2.1.10.RELEASE/actuator-api/html/#health
* Spring Boot 2.2: https://docs.spring.io/spring-boot/docs/2.2.1.RELEASE/actuator-api/html/#health

So if we switch to Spring Boot 2.2 then this health checking module needs to be aligned to the health checking API version V3. Or act as described in the referenced Spring Boot 2.2 API documentation: _If you need to return V2 JSON you should use an accept header or application/vnd.spring-boot.actuator.v2+json_.

In this PR I removed all code that processed result from actuator health endpoint and I see identical result in the UI. Some thoughts about removed code:
* this health component code is 3+ years old, maybe 3+ years ago this logic was needed (Spring Boot <= 1.4)
* in `health.component` test cases the input to tests is different from what it comes currently from API V2 - so maybe this was needed for API V1, then just lucky case that this API V1 code at all works for API V2 and something is shown
* tried to get API V1 result from health endpoint (to see how it looks like) with current master code (added `Accept` header `application/vnd.spring-boot.actuator.v1+json`) and got error: HTTP 406 - Could not find acceptable representation
* using `Accept` header `application/vnd.spring-boot.actuator.v2+json` returned correct result so seems that API V1 is so old that Spring Boot 2.1 doesn't support that any more
* health modal contained `h4` element for error, in current implementation this is not shown, if I made it visible for key error in the new version then it looked really ugly for mail service, long scrollbar appeared to modal, so I just removed that because this is not shown in the current master version

In Spring Boot 2.2 this version of `checkHealth()` in `health.service` should work with currently defined data model (we should get API V2 result) and we can migrate later to API V3 response:
```
  checkHealth(): Observable<Health> {
    let headers = new HttpHeaders();
    headers = headers.append('Accept', 'application/vnd.spring-boot.actuator.v2+json');
    return this.http.get<Health>(SERVER_API_URL + 'management/health', { headers });
  }
```

Additional note: removed prefix and aligned component name according to #10220

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/health/health.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/health/health.component.html.ejs
@@ -33,15 +33,15 @@
                 </tr>
             </thead>
             <tbody>
-                <tr *ngFor="let health of healthData">
-                    <td><% if (enableTranslation) { %>{{'health.indicator.' + baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ baseName(health.name) }}</span><% } %> {{subSystemName(health.name)}}</td>
+                <tr *ngFor="let healthKey of healthKeys()">
+                    <td><% if (enableTranslation) { %>{{'health.indicator.' + healthKey | translate}}<% } else { %><span class="text-capitalize">{{ healthKey }}</span><% } %></td>
                     <td class="text-center">
-                        <span class="badge" [ngClass]="getBadgeClass(health.status)" jhiTranslate="{{'health.status.' + health.status}}">
-                            {{health.status}}
+                        <span class="badge" [ngClass]="getBadgeClass(healthData.details[healthKey].status)" jhiTranslate="{{'health.status.' + healthData.details[healthKey].status}}">
+                            {{ healthData.details[healthKey].status }}
                         </span>
                     </td>
                     <td class="text-center">
-                        <a class="hand" (click)="showHealth(health)" *ngIf="health.details || health.error">
+                        <a class="hand" (click)="showHealth(healthKey, healthData.details[healthKey].details)" *ngIf="healthData.details[healthKey].details">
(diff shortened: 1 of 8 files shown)


DanielFran:
@pascalgrimaud Probably this is something to be changed also in React and VueJs?

kaidohallik:
@DanielFran I looked at React code and React code is using Spring Boot health endpoint data directly without processing. So after changes by this PR, Angular code is similar to React code.

generators/client/templates/angular/src/main/webapp/app/admin/health/health.component.html.ejs
```diff
@@ -33,15 +33,15 @@
                 </tr>
             </thead>
             <tbody>
-                <tr *ngFor="let health of healthData">
-                    <td><% if (enableTranslation) { %>{{'health.indicator.' + baseName(health.name) | translate}}<% } else { %><span class="text-capitalize">{{ baseName(health.name) }}</span><% } %> {{subSystemName(health.name)}}</td>
+                <tr *ngFor="let healthKey of healthKeys()">
```
wmarques: you can use this I think https://angular.io/api/common/KeyValuePipe

kaidohallik: Thanks for pointing out!

I tried `keyvalue` pipe, but this changed elements order. Then I used `keys` pipe from `ng-jhipster` (https://github.com/jhipster/ng-jhipster/blob/master/src/pipe/keys.pipe.ts), as this was before my changes, and code is simpler now.

PR #9593, opened 2019-04-21
fix websocket paths and exclude websockets from client redirect

ruddell:
Fix #9591

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10919, opened 2019-12-12
fix websocket fallback paths in ClientForwardController

ruddell:
Ensures that websocket fallbacks, such as `xhr_streaming` and `eventsource`, receive requests as expected rather than forwarding to the client.

The issue fixed here was that the regex for `websocket` was searching the end of the path rather than the beginning, due to the prefix of `/**/`.  When websockets don't work, they fallback to other options such as `xhr_streaming` and `eventsource` (URLs end with the method name).    Regular websockets work because the websocket URL ends in `websocket`

Before this PR, request for fallback methods would result in a `405 Method Not Allowed` because they are POST requests going to the client at `/` ([screenshot](https://user-images.githubusercontent.com/4294623/70653256-f725ff00-1c08-11ea-8a1e-8083f5bd7026.png))

Improved version of https://github.com/jhipster/generator-jhipster/pull/9593, based off [this StackOverflow](https://stackoverflow.com/questions/48565550/map-all-requests-instead-of-specific-pattern-in-spring-boot/48622614#48622614)

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
@@ -28,7 +28,11 @@ public class ClientForwardController {
      * Forwards any unmapped paths (except those containing a period) to the client {@code index.html}.
      * @return forward to client {@code index.html}.
      */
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
Trusting you on this

PR #10921, opened 2019-12-12
Improve reproducibility

mshima:
This PR addresses https://github.com/jhipster/generator-jhipster/pull/10897#issuecomment-564269736 and https://github.com/jhipster/generator-jhipster/pull/10897#issuecomment-564255169

- Use faker to provide seed for Randexp

- Use entity's changelogDate to initialize faker seed, so changing the entity's order will not change the fake data.

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -22,8 +22,6 @@ const faker = require('faker');
 const utils = require('../utils');
 const constants = require('../generator-constants');
 
-// In order to have consistent results with Faker, the seed is fixed.
-faker.seed(42);
 // In order to have consistent results with RandExp, the RNG is seeded.
 Randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
 
@@ -285,6 +283,13 @@ function addSampleRegexTestingStrings(generator) {
 
 function writeFiles() {
     return {
+        setupReproducibility() {
+            if (this.skipClient) return;
+
+            // In order to have consistent results with Faker, restart seed with current entity name hash.
+            faker.seed(utils.stringHashCode(this.name.toLowerCase()));
+        },
+
         writeClientFiles() {
             if (this.skipClient) return;
             if (this.protractorTests) {
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -18,8 +18,14 @@
  */
 const _ = require('lodash');
 const Randexp = require('randexp');
+const faker = require('faker');
 const utils = require('../utils');
 const constants = require('../generator-constants');
+
+// In order to have consistent results with Faker, the seed is fixed.
+faker.seed(42);
+// In order to have consistent results with RandExp, the RNG is seeded.
+Randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
 
 /* Constants use throughout */
 const CLIENT_TEST_SRC_DIR = constants.CLIENT_TEST_SRC_DIR;
@@ -272,8 +278,6 @@ function addSampleRegexTestingStrings(generator) {
         if (field.fieldValidateRulesPattern !== undefined) {
             const randExp = new Randexp(field.fieldValidateRulesPattern);
             randExp.max = 5;
-            // In order to have consistent results with RandExp, the RNG is seeded.
-            randExp.randInt = generator.seededRandomNumberGenerator(10);
             field.fieldValidateSampleString = randExp.gen();
         }
     });
@@ -283,7 +287,9 @@ function writeFiles() {
     return {
         writeClientFiles() {
             if (this.skipClient) return;
-            addSampleRegexTestingStrings(this);
+            if (this.protractorTests) {
+                addSampleRegexTestingStrings(this);
+            }
             if (this.clientFramework === 'angularX') {
                 // write client side files for angular 2.x +
                 this.writeFilesToDisk(
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -17,13 +17,12 @@
  * limitations under the License.
  */
 const _ = require('lodash');
-const Randexp = require('randexp');
 const faker = require('faker');
 const utils = require('../utils');
 const constants = require('../generator-constants');
 
-// In order to have consistent results with RandExp, the RNG is seeded.
-Randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
+/* Use customized randexp */
+const Randexp = utils.RandexpWithFaker;
 
 /* Constants use throughout */
 const CLIENT_TEST_SRC_DIR = constants.CLIENT_TEST_SRC_DIR;
@@ -275,7 +274,6 @@ function addSampleRegexTestingStrings(generator) {
     generator.fields.forEach(field => {
         if (field.fieldValidateRulesPattern !== undefined) {
             const randExp = new Randexp(field.fieldValidateRulesPattern);
-            randExp.max = 5;
             field.fieldValidateSampleString = randExp.gen();
         }
     });
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -17,12 +17,14 @@
  * limitations under the License.
  */
 const _ = require('lodash');
-const randexp = require('randexp');
 const chalk = require('chalk');
 const faker = require('faker');
 const fs = require('fs');
 const utils = require('../utils');
 const constants = require('../generator-constants');
+
+/* Use customized randexp */
+const randexp = utils.RandexpWithFaker;
 
 /* Constants use throughout */
 const INTERPOLATE_REGEX = constants.INTERPOLATE_REGEX;
@@ -30,9 +32,6 @@ const SERVER_MAIN_SRC_DIR = constants.SERVER_MAIN_SRC_DIR;
 const SERVER_MAIN_RES_DIR = constants.SERVER_MAIN_RES_DIR;
 const TEST_DIR = constants.TEST_DIR;
 const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
-
-// In order to have consistent results with RandExp, the RNG is seeded.
-randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
 
 /*
  * Current faker version is 4.1.0 and was release in 2017
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -31,8 +31,6 @@ const SERVER_MAIN_RES_DIR = constants.SERVER_MAIN_RES_DIR;
 const TEST_DIR = constants.TEST_DIR;
 const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
 
-// In order to have consistent results with Faker, the seed is fixed.
-faker.seed(42);
 // In order to have consistent results with RandExp, the RNG is seeded.
 randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
 
@@ -63,7 +61,10 @@ const getRecentDate = function(days, refDate) {
 const getRecentForLiquibase = function(days, changelogDate) {
     let formatedDate;
     if (changelogDate !== undefined) {
-        formatedDate = `${changelogDate.substring(0, 4)}-${changelogDate.substring(4, 6)}-${changelogDate.substring(6, 8)}`;
+        formatedDate = `${changelogDate.substring(0, 4)}-${changelogDate.substring(4, 6)}-${changelogDate.substring(
+            6,
+            8
+        )}T${changelogDate.substring(8, 10)}:${changelogDate.substring(10, 12)}:${changelogDate.substring(12, 14)}+00:00`;
     }
     return getRecentDate(1, formatedDate);
 };
@@ -336,6 +337,13 @@ function writeFiles() {
                 `${this.microservicePath}/${this.jhipsterConfigDirectory}/${this.entityNameCapitalized}.json`,
                 this.destinationPath(`${this.jhipsterConfigDirectory}/${this.entityNameCapitalized}.json`)
             );
+        },
+
+        setupReproducibility() {
+            if (this.skipServer) return;
+
+            // In order to have consistent results with Faker, restart seed with current entity name hash.
+            faker.seed(utils.stringHashCode(this.name.toLowerCase()));
         },
 
         writeServerFiles() {
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -33,6 +33,8 @@ const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
 
 // In order to have consistent results with Faker, the seed is fixed.
 faker.seed(42);
+// In order to have consistent results with RandExp, the RNG is seeded.
+randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
 
 /*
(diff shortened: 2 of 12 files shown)


mshima:
@kaidohallik can you test if this PR doesn't break https://github.com/jhipster/generator-jhipster/pull/10897?

mshima:
Integration test is passing https://github.com/jhipster/generator-jhipster/pull/10915 (default-jdl)

ruddell:
Does `this.changelogDate` always have a value?  I'm not familiar with that variable.  I was thinking of doing something similar (basing the faker seed on the entity name) but this looks good to me.

From some quick testing, works great.  It's nice to be able to export-jdl and import the same, then have no code changes. 👍

mshima:
> Does `this.changelogDate` always have a value? I'm not familiar with that variable. I was thinking of doing something similar (basing the faker seed on the entity name) but this looks good to me.

It's used for setting the liquibase changelog name.

Doesn't exist on some cases:
https://github.com/jhipster/generator-jhipster/blob/abb3fff743d010b33214c074f738eb6efe2145ec/generators/entity/index.js#L601-L606

mshima:
Changed the fallback to a calculated hash of entity's name instead of a fixed 42 value.

ruddell:
> Changed the fallback to a calculated hash of entity's name instead of a fixed 42 value.

We could remove `parseLiquibaseChangelogDate` by always using the hash of the entity's name.  That field is always guaranteed to have a value

mshima:
> We could remove `parseLiquibaseChangelogDate` by always using the hash of the entity's name. That field is always guaranteed to have a value

Is there any case we will want to rename an entity?

mshima:
I like the changelogDate approach because if the user wants to change the seed, it is possible.

mshima:
Just had another through. Hash is enough.
I will remove changelogDate.

kaidohallik:
Below is test result after I added fields as in just opened test data adding PR #10925

### Add field with pattern that doesn't allow simple alphabetical strings. ###

I added the following field to some entity:
```
        {
            "fieldName": "numberPatternTom",
            "fieldType": "String",
            "fieldValidateRules": [
                "pattern"
            ],
            "fieldValidateRulesPattern": "^[0-9]+$",
            "fieldValidateRulesPatternJava": "^[0-9]+$"
        }
```

In generation is shown warning:  
_Randomly generated first and second test values for entity "FieldTestEntity" field "numberPatternTom" with pattern "^[0-9]+$" in file "FieldTestEntityResourceIT" where equal, added symbol "B" to second value._

### Back end tests. ###

Current master generated:
```
    private static final String DEFAULT_NUMBER_PATTERN_TOM = "97658";
    private static final String UPDATED_NUMBER_PATTERN_TOM = "38";
```

This branch generated:
```
    private static final String DEFAULT_NUMBER_PATTERN_TOM = "";
    private static final String UPDATED_NUMBER_PATTERN_TOM = "B";
```

And back end tests failed.

### e2e tests. ###

Current master generated:
```
      fieldTestEntityUpdatePage.setNumberPatternTomInput('97658'),
```

This branch generated:
```
      fieldTestEntityUpdatePage.setNumberPatternTomInput('numberPatternTom'),
```

And e2e tests failed.

### Add mandatory field with pattern that doesn't allow simple alphabetical strings. ###

```
        {
            "fieldName": "numberPatternRequiredTom",
            "fieldType": "String",
            "fieldValidateRules": [
                "pattern",
                "required"
            ],
            "fieldValidateRulesPattern": "^[0-9]+$",
            "fieldValidateRulesPatternJava": "^[0-9]+$"
        }
```

Current master generates fake data for entity which contains that field.

This branch doesn't generate fake data for the entity containing such field (only header is present, 0 data lines is generated).

mshima:
@kaidohallik thank you for your detailed report.

mshima:
I think it is fixed.
Forgot to return the random number at randInt function :unamused:

Added the following patterns to IT tests:
https://github.com/jhipster/generator-jhipster/blob/38cd0dcb23e6d9ede77e7113ea09d277ac8fe31b/test-integration/samples/jdl-default/app.jdl#L89-L92

Results (https://dev.azure.com/jhipster/generator-jhipster/_build/results?buildId=7781 -> default-jdl):

- src/test/java/io/github/jhipster/sample/web/rest/EmployeeResourceIT.java

    private static final String DEFAULT_EMAIL = "axqa7B@o#(t?#.^wd";
    private static final String UPDATED_EMAIL = "]&g$\"@).$.R</H";

    private static final String DEFAULT_PATTER = "AAAAAAAAAA";
    private static final String UPDATED_PATTER = "BBBBBBBBBB";

    private static final String DEFAULT_NUMBER_PATTERN_TOM = "80346";
    private static final String UPDATED_NUMBER_PATTERN_TOM = "34201";

    private static final String DEFAULT_PATTERN_NAME = "Xnsxzl2";
    private static final String UPDATED_PATTERN_NAME = "Afcgbc3";


- src/main/resources/config/liquibase/fake-data/employee.csv

id;first_name;last_name;email;patter;number_pattern_tom;pattern_name;phone_number;hire_date;salary;commission_pct;sibag_id;gobag_id
1;Lola;Rodriguez;"Fd;\@,9d./^+";"zJf";"2104";"Ammm3";cross-platform;2019-12-13T14:45:36;7230;90426;1;1
2;Julio;Senger;".@lzU,.J--";"7lTO";"761216";"Gyj9";payment Expressway;2019-12-12T19:24:45;43151;12407;2;2
3;Edwin;Abshire;"8$'$@`p.I0k\}";"A";"77";"Loaqpc0";architecture Pants Profound;2019-12-13T08:46:29;18500;352;3;3
4;Magnolia;Block;"wZ@+fN-%K.W";"nN5";"32";"Gq0";Upgradable partnerships;2019-12-12T21:10:15;14242;70290;4;4

- src/test/javascript/e2e/entities/employee/employee.spec.ts:

           employeeUpdatePage.setEmailInput('g^[J@,hYFd.\C'),
            employeeUpdatePage.setPatterInput('patter'),
            employeeUpdatePage.setNumberPatternTomInput('74'),
            employeeUpdatePage.setPatternNameInput('Ecnij1'),

generators/entity-client/files.js
```diff
@@ -18,9 +18,13 @@
  */
 const _ = require('lodash');
 const Randexp = require('randexp');
+const faker = require('faker');
 const utils = require('../utils');
 const constants = require('../generator-constants');
 
+// In order to have consistent results with RandExp, the RNG is seeded.
+Randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
```
MathieuAA: It's not used elsewhere in this file, this should be removed.
Further more, I'm not fond of modifying the prototype of something like that. In this case, I see that we set a new function to Randexp:
  - if it's a new one, then it's kinda okay... until Randexp implements a `randIt` function.
  - if a `randInt` already exists and we're overriding it, then what if another developer comes and uses this function not knowing that it has been overridden. You can see the "oops" effect here :)

In my opinion, a better way of doing this would be to extract this function to another file (either new or already existing file) and not modify the prototype so that it may be used.

generators/entity-server/files.js
```diff
@@ -31,8 +31,8 @@ const SERVER_MAIN_RES_DIR = constants.SERVER_MAIN_RES_DIR;
 const TEST_DIR = constants.TEST_DIR;
 const SERVER_TEST_SRC_DIR = constants.SERVER_TEST_SRC_DIR;
 
-// In order to have consistent results with Faker, the seed is fixed.
-faker.seed(42);
+// In order to have consistent results with RandExp, the RNG is seeded.
+randexp.prototype.randInt = (min, max) => faker.random.number({ min, max });
```
MathieuAA: same comment as before, this should be removed

generators/utils.js
```diff
@@ -606,3 +607,25 @@ function normalizeBlueprintName(blueprint) {
     }
     return blueprint;
 }
+
+/**
+ * Calculate a hash code for a given string.
+ * @param {string} str - any string
+ * @returns {number} returns the calculated hash code.
+ */
+function stringHashCode(str) {
```
MathieuAA: this is hard to understand

MathieuAA: I see this has been taken from https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
Copy/pasting is actually bad in this case:
  - why do we need to declare all the variables outside the loop?
  - the naming has to be improved (chr => character, for instance)
  - there should be a comment on where this snippet has been taken from and adapted, or at least to say this has been adapted from what exists in Java
  - what about `Math.imul` ([resource](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/imul))

test/templates/reproducible/.yo-rc.json
```diff
@@ -0,0 +1,28 @@
...
+    "packageFolder": "com/mycompany/myapp",
+    "authenticationType": "session",
+    "cacheProvider": "ehcache",
+    "enableHibernateCache": true,
+    "websocket": "no",
+    "databaseType": "sql",
+    "devDatabaseType": "h2Disk",
+    "prodDatabaseType": "mysql",
+    "searchEngine": "no",
+        "buildTool": "maven",
```
MathieuAA: can you fix the indent here please? thanks

MathieuAA:
thanks @DanielFran for pinging me :)

generators/entity-client/files.js
mshima: Randexp.prototype.randInt is a function that returns a random value from min to max.

This is a recommended way to do this https://github.com/fent/randexp.js?#custom-prng

Otherwise we must set randInt with a custom value or function every time we call `new Randexp()` or  `const randExp = new this.randexp(fields[idx].fieldValidateRulesPattern)`

generators/utils.js
mshima: I took it from here:
https://github.com/jhipster/generator-jhipster/blob/e362242b0f750f04867fb2e087f1911af84e548b/generators/generator-base.js#L1673-L1686

mshima: Extract as an util function

test/templates/reproducible/.yo-rc.json
mshima: Thanks.
This was a copy/paste =/

https://github.com/jhipster/generator-jhipster/blob/e362242b0f750f04867fb2e087f1911af84e548b/test/templates/default/.yo-rc.json#L13-L15

test/templates/reproducible/.jhipster/Foo.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Foo",
+    "fields": [
+        {
+            "fieldName": "name",
+            "fieldType": "String"
+        },
+        {
+            "fieldName": "numberPatternRequiredTom",
```
kaidohallik: Maybe use just simply `numberPatternRequired` for better code readability? In #10925 I used suffix Tom because all fields are suffixed with Tom there.

test/entity.spec.js
```diff
@@ -816,5 +818,37 @@ describe('JHipster generator for entity', () => {
...
+                    assert.file(expectedFiles.server);
+                    assert.file(expectedFiles.clientNg2);
+                    assert.file(expectedFiles.gatling);
+                    assert.file(expectedFiles.fakeData);
+
+                    assert.fileContent(`${SERVER_MAIN_RES_DIR}config/liquibase/fake-data/foo.csv`, /1;New Leu Diverse Illinois;"0646"/);
+
+                    assert.fileContent(
+                        `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/web/rest/FooResourceIT.java`,
+                        /DEFAULT_NUMBER_PATTERN_REQUIRED_TOM = "387"/
```
kaidohallik: If field name will be changed to `numberPatternRequired` then `_TOM` suffix needs to be removed.

test/entity.spec.js
```diff
@@ -816,5 +818,37 @@ describe('JHipster generator for entity', () => {
...
+                    );
+                    assert.fileContent(
+                        `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/web/rest/FooResourceIT.java`,
+                        /UPDATED_NUMBER_PATTERN_REQUIRED_TOM = "468439"/
```
kaidohallik: If field name will be changed to `numberPatternRequired` then `_TOM` suffix needs to be removed.

generators/utils.js
MathieuAA: fair enough

generators/entity-client/files.js
MathieuAA: I can't believe this is actually in their doc. I'm generally okay with monkey-patching, but this is not a good use case.

> Otherwise we must set randInt with a custom value or function every time we call new Randexp() or const randExp = new this.randexp(fields[idx].fieldValidateRulesPattern)

There's an other way:
  - create a RandExp wrapper that will take care of "wrapping" this dependency
  - require this wrapper everywhere where you used to use randexp
That way, you really control how and where this dependency is used.

generators/utils.js
MathieuAA: can you at least improve the code?

generators/entity-client/files.js
mshima: > I can't believe this is actually in their doc. I'm generally okay with monkey-patching, but this is not a good use case.

https://www.npmjs.com/package/randexp#custom-prng

If we use Randexp, for sure we will want to make it reproducible.
In this case I see the "oops" effect when we forget to set randInt and the result is not reproducible.

generators/utils.js
mshima: done, I should reference stackoverflow on this?

MathieuAA: Maybe not

generators/entity-client/files.js
mshima: Take a look at:
https://github.com/fent/randexp.js/blob/2b13312e561953288a3cf9b3501e77519b400459/lib/randexp.js#L40-L41

`RandExp.prototype.max` is not set, but RandExp allows to set it and make it default.

Faker too, faker.seed is a global variable.
https://github.com/jhipster/generator-jhipster/blob/4ec3b1f3d216790fdf8e96ad3c0d540012986f7e/generators/entity-server/files.js#L34-L35

An implementation of a local RandExp would be:
```
const creator = new RandExp();
creator.randInt = function ...;
const expr = creator.create(/[0-9]*/).gen();
```
For a consistent result, creator should be passed around, and that doesn't work "out of the box".
That's the reason randomizers uses global implementation.
Creating a randomizer object and passing around seems a waste of time to me.

Faker for instance, there is no easy way to make it an object.

generators/utils.js
mshima: Is this resolved?

generators/entity-client/files.js
MathieuAA: Yeah, but there's a difference between:
  -  https://github.com/fent/randexp.js/blob/2b13312e561953288a3cf9b3501e77519b400459/lib/randexp.js#L40-L41 where the lib modifies an intern behavior (everything stays inside the lib here)
  - setting a "global thing", the way the setter works for faker

One way I see it working:
```javascript
const RandExp = require('randexp');

module.exports = { ... };

function randomInteger(min, max) {
  // whatever you want with RandExp
}
function anotherRandomThing(/* whatever args */) {
  // ...
}
```

Why wouldn't it work?

> Faker for instance, there is no easy way to make it an object.

I don't get this.

mshima: Done.

mshima:
@MathieuAA thanks.

PR #10926, opened 2019-12-13
update npm to latest version

atomfrede:
following the latest security advice
https://blog.npmjs.org/post/189618601100/binary-planting-with-the-npm-cli
updating npm and yarn for the generated clients

I didn't update/enforce npm version for the generator itself, should we do that? e.g. in `package.json`

image


pascalgrimaud:
should we do a release for that ?
anyway, I think we should do a release just before Christmas :)

atomfrede:
If users updates their global npm they are fine when not building with gradle/maven so a release might not be that urgent, but agree we should have a release hopefully with boot 2.2

PR #10932, opened 2019-12-13
Strict Typescript option in configuration module in Angular

kaidohallik:
This is a configuration module PR for the #10631

Some notes:

* As according to https://angular.io/guide/pipes#appendix-no-filterpipe-or-orderbypipe it's not recommended to perform filtering or ordering with pipe then now filtering and ordering is done in the component method when filter or order change event occurs.

* UI previously didn't show to user that it's possible to order by prefix. Now UI uses standard `jhiSort` + `jhiSortBy` directives to show that.

* According to #10220 removed prefix from configuration service and component.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.component.ts.ejs
@@ -18,14 +18,14 @@
 -%>
 import { Component, OnInit } from '@angular/core';
 
-import { ConfigurationService, Beans, Bean, PropertySource } from './configuration.service';
+import { ConfigurationService, Bean, PropertySource } from './configuration.service';
 
 @Component({
     selector: '<%= jhiPrefixDashed %>-configuration',
     templateUrl: './configuration.component.html'
 })
 export class ConfigurationComponent implements OnInit {
-    allBeans!: Beans;
+    allBeans!: Bean[];
     beans: Bean[] = [];
     beansFilter = '';
     beansAscending = true;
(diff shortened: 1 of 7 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.component.ts.ejs
```diff
@@ -18,46 +18,41 @@
...
-        this.configurationService.get().subscribe((configuration) => {
-            this.configuration = configuration;
+    filterAndSortBeans(): void {
+        this.beans = [];
 
-            for (const config of configuration) {
-                if (config.properties !== undefined) {
-                    this.configKeys.push(Object.keys(config.properties));
-                }
+        Object.keys(this.allBeans).forEach(beanKey => {
```
wmarques: Can we use `filter` here ? looks appropriate

generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.service.ts.ejs
```diff
@@ -17,70 +17,79 @@
...
-                        properties.push(propertiesObject[key]);
+    getBeans(): Observable<Beans> {
+        return this.http.get<ConfigProps>(SERVER_API_URL + 'management/configprops').pipe(
+            map(configProps => {
+                // This code is for Spring Boot 2
+                for (const key in configProps.contexts) {
+                    // If the key is not bootstrap, it will be the ApplicationContext Id
+                    // For default app, it is baseName
+                    // For microservice, it is baseName-1
+                    if (!key.startsWith('bootstrap')) {
```
wmarques: Why this was added ? Looks unrelated to the strict option no ?

kaidohallik: As I removed data conversion from one representation to another then code became so short that I moved functionality from `getConfigPropertiesObjects` to this function and removed `getConfigPropertiesObjects`. So this is not added by this PR but moved from one function to another.

But if I looked at React code now then I see that React is not dropping any context, React is showing all available information. So I'll do the same here, let's show all available information from `configprops` endpoint like this is done in React.

generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.component.ts.ejs
kaidohallik: Now I'm returning array from service and this allowed write it with filter and very shortly.

generators/client/templates/angular/src/main/webapp/app/admin/configuration/configuration.service.ts.ejs
kaidohallik: Now using all contexts beans as this is in React side.

PR #10938, opened 2019-12-14
Strict Typescript option in HTML templates in Angular

kaidohallik:
This fixes strict problems inside HTML templates. This PR is for #10631 

After this and #10932 are merged we can try to enable strict option in Angular:
* in `tsconfig.json` and in `tsconfig-aot.json` change
```
        "noImplicitAny": false,
```
to
```
        "strict": true,
```
* add new rule into `tslint.json`: `"typedef": [true, "call-signature"]`

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
@@ -49,21 +49,21 @@
                         <input type="password" class="form-control" id="newPassword" name="newPassword"
                                placeholder="{{'global.form.newpassword.placeholder' | translate}}"
                                formControlName="newPassword" #newPassword>
-                        <div *ngIf="passwordForm.get('newPassword').invalid && (passwordForm.get('newPassword').dirty || passwordForm.get('newPassword').touched)">
+                        <div *ngIf="passwordForm.get('newPassword')?.invalid && (passwordForm.get('newPassword')?.dirty || passwordForm.get('newPassword')?.touched)">
                             <small class="form-text text-danger"
-                               *ngIf="passwordForm.get('newPassword').errors.required" jhiTranslate="global.messages.validate.newpassword.required">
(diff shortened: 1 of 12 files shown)


generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
```diff
@@ -49,39 +49,39 @@
                         <input type="password" class="form-control" id="newPassword" name="newPassword"
                                placeholder="{{'global.form.newpassword.placeholder' | translate}}"
                                formControlName="newPassword" #newPassword>
-                        <div *ngIf="passwordForm.get('newPassword').invalid && (passwordForm.get('newPassword').dirty || passwordForm.get('newPassword').touched)">
+                        <div *ngIf="passwordForm.get('newPassword')?.invalid && (passwordForm.get('newPassword')?.dirty || passwordForm.get('newPassword')?.touched)">
```
wmarques: Are those fields really nullable ? I don't see the case where `passwordForm.get('newPassword')` could be undefined

kaidohallik:
Problems were in form.get('field').error.errorname - seems like compiler is not working brilliantly in html templates if both Angular `fullTemplateTypeCheck` and Typescript `strict` is turned on.

Compile error:
* `form.get('field')!.error.errorname`
* `form.get('field')!.error?.errorname`
* `form.get('field').error!.errorname`

Compile success:
* `form.get('field')?.error.errorname`
* `form.get('field')?.error?.errorname`
* `form.get('field')!.error!.errorname`

As in some places were used `error?.` then it was mandatory to use `?.error` and then I used `?.` in most places to keep similar style.  
Now I made change and using `!` as much as possible and use `?` only for error details check `form.get('field')?.error?.errorname`

I rebased against master and force pushed to pass CI.

generators/client/templates/angular/src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.html.ejs
kaidohallik: I accidentally answered in separate comment instead of this place.

wmarques:
That's weird indeed, LGTM if we are consistent, thanks @kaidohallik

PR #10940, opened 2019-12-14
GitHub Actions: split Angular and React builds

pascalgrimaud:
As the builds for React are too flacky, compared to Angular, I prefer to split them.
So it will be easier to restart only React builds.

_____

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -1,0 +1,160 @@
+#
+# Copyright 2013-2019 the original author or authors from the JHipster project.
+#
+# This file is part of the JHipster project, see https://www.jhipster.tech/
+# for more information.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


pascalgrimaud:
I'm merging this to help the other next PR

vishal423:
@pascalgrimaud, we can plan to move all `react` builds to azure devops as that supports the restart of an individual job on demand.

pascalgrimaud:
Yes, why not. It could be a temporary solution...

But the real solution is to fix the flacky protractor tests with React, similar to what we did with Vue.js but I know it will require a lot of work. 

I'd like to open a specific ticket for this, with all details and with a big bounty as it's a lot of work to solve all these flacky tests...

vishal423:
I think we can leverage most of code from `vue`, however, if we plan to move to `cypress`, we shouldn't spend more effort on this.

Issue #10934, opened 2019-12-13
--skip-server does not log error message on failure


ruddell:
##### **Overview of the issue**
Running `jhipster --skip-server` only prompts for the baseName of the application, then exits silently without any message.  Previously (v6.5.1 and before), it would exit with an error message:
```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: When using skip-server flag, you must pass a database option and authentication type using --db and --auth flags
```

##### **Motivation for or Use Case**
Error messages should display as expected

##### **Reproduce the error**
Run `jhipster --skip-server` and answer the baseName prompt, it will then exit without the expected error message

##### **Suggest a Fix**
I have not looked into it.  Did we change error handling since the last release?  Or maybe something related to yeoman?

Git bisect shows this issue first appeared after #10727
```
75c31ed5a5fc5b2df024a6508f415423fa670535 is the first bad commit
commit 75c31ed5a5fc5b2df024a6508f415423fa670535
Author: Marcelo Boveto Shima <marceloshima@gmail.com>
Date:   Mon Nov 4 16:32:33 2019 -0200

    Exit import-jdl with status from the forked process.

:040000 040000 6d924f61565ba9e31d779aed3956cd9093afc66e 8751ab748e47c88ff97d71d6dec91a4837d8e347 M	cli
```

##### **JHipster Version(s)**
Reprouced with Master
Error message shows correctly on v6.5.1

##### **Browsers and Operating System**
Mac, Node LTS

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
Adding a bounty on this, as it's a bug

SudharakaP:
So I analyzed this issue and I think the following lines, 

https://github.com/jhipster/generator-jhipster/blob/ba88f1424ddffd08589377a1dba9e432ceaa8a0b/cli/cli.js#L45-L47

have a problem. It tells node to end abruptly which is not necessary it seems (Refer: https://nodejs.org/api/process.html#process_process_exit_code). We can instead set the `process.exitCode` there. I've created a PR for this. Let me know if you guys agree with this. :smile: 

cc: @mshima

mshima:
Thank you @SudharakaP for finding this.
This makes getExitCode and setExitCode useless.
https://github.com/jhipster/generator-jhipster/pull/10946

I think blueprints has the same problem with:
[...]

PR #10948, opened 2019-12-15
Remove Explicit Calls to process.exit and Set the exitCode Instead

SudharakaP:
As discussed in #10934, remove explicit calls to node exit command (`process.exit()) and use `process.exitCode` instead. 

Reference: https://nodejs.org/api/process.html#process_process_exit_code

### DIFF
--- a/cli/run-yeoman-process.js
+++ b/cli/run-yeoman-process.js
@@ -30,7 +30,7 @@ logger.info(chalk.yellow(`Options: ${toString(options)}`));
 try {
     env.run(command, options, () => {
         done();
-        process.exit(0);
+        process.exitCode = 0;
     });
 } catch (e) {
     logger.error(e.message, e);
--- a/cli/utils.js
+++ b/cli/utils.js
@@ -47,7 +47,7 @@ const error = function(msg, trace) {
     if (trace) {
         console.log(trace);
     }
-    process.exit(1);
+    process.exitCode = 1;
 };
 
 const init = function(program) {
@@ -99,7 +99,7 @@ const initHelp = (program, cliName) => {
             logger.info(`Did you mean ${chalk.yellow(suggestion)}?`);
         }
 
-        process.exit(1);
+        process.exitCode = 1;
     });
 };
 


cli/utils.js
```diff
@@ -47,7 +47,7 @@ const error = function(msg, trace) {
     if (trace) {
         console.log(trace);
     }
```
mshima: By removing process.exit at this point, jhipster will not exit on log.error anymore.
We should find a cleaner way to exit the process.

mshima: Maybe throwing an exception.

SudharakaP: @mshima: Thanks for caching this. Yes, throwing an exception seems to be the way that the documentation suggests; https://nodejs.org/api/process.html#process_process_exit_code and https://nodejs.dev/error-handling-in-nodejs. Let me try to correct this in a separate PR. Or would you like to do the PR? :smile:

mshima: Feel free to do it. Don't have time right now.

cli/utils.js
```diff
@@ -99,7 +99,7 @@ const initHelp = (program, cliName) => {
             logger.info(`Did you mean ${chalk.yellow(suggestion)}?`);
         }
 
-        process.exit(1);
+        process.exitCode = 1;
```
mshima: Probably we should throw an exception here too.
This seems a 'command not found'.

cli/run-yeoman-process.js
```diff
@@ -30,7 +30,7 @@ logger.info(chalk.yellow(`Options: ${toString(options)}`));
 try {
     env.run(command, options, () => {
         done();
-        process.exit(0);
+        process.exitCode = 0;
```
mshima: Probably we should don't do anything:

```
env.run(command, options, done);
```

cli/utils.js
SudharakaP: Btw, I've just tested this function and when we set `process.exitCode=1` and for me, it seems to exit cleanly without any problems. Could you give me and example where you see this not exiting cleanly problem? :thinking:

cli/utils.js
SudharakaP: Again I've just tested this function and even when we set `process.exitCode=1` it seems to exit cleanly without any problems. :smile:

cli/run-yeoman-process.js
SudharakaP: Agreed. :smile:

cli/utils.js
mshima: I didn't say it's wrong, but is not safe for last minute merge just before release.

There is a change of behavior on logger.error (Exit on first error to set exit status and keep running)
I found 10 uses of logger.error, so we must check this new behavior.

https://github.com/jhipster/generator-jhipster/blob/08f36bbe77a92140516dc4a50401a202c9001f38/cli/jhipster.js#L28-L32

Currently, if node version is not supported, jhipster will log the error and quit.
With the new behavior, jhipster will keep running and maybe will success, but jhipster will exit with error code.

At this point, probably we should change to something like logger.fatal().

We must check done behavior too:
https://github.com/jhipster/generator-jhipster/blob/08f36bbe77a92140516dc4a50401a202c9001f38/cli/utils.js#L184-L190

cli/utils.js
mshima: Only place is called is at initHelp:
https://github.com/jhipster/generator-jhipster/blob/08f36bbe77a92140516dc4a50401a202c9001f38/cli/cli.js#L87-L101
Letting the process continue seems to make no harm.

cli/utils.js
SudharakaP: Ah, I understand now. We'll keep this as it is for now I think; since changing it might involve unnecessary problems. I've changed the other two and created #10956; feel free to let me know if you see any further issues. Thanks :smile:

PR #10976, opened 2019-12-17
Add Maven  --batch-mode option to improve CI readability in GitHub Actions

jdubois:
This should remove all the "downloading artefact" clutter in the console log, if this works this is going to be a lot better for reading the logs

### DIFF
--- a/test-integration/scripts/10-install-jhipster.sh
+++ b/test-integration/scripts/10-install-jhipster.sh
@@ -13,7 +13,7 @@ if [[ "$JHI_REPO" == *"/jhipster" ]]; then
     cd "$JHI_HOME"
     git --no-pager log -n 10 --graph --pretty='%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
 
-    ./mvnw -ntp clean install -DskipTests -Dmaven.javadoc.skip=true -Dgpg.skip=true
+    ./mvnw -ntp clean install -DskipTests -Dmaven.javadoc.skip=true -Dgpg.skip=true --batch-mode
 
 elif [[ "$JHI_LIB_BRANCH" == "release" ]]; then
     echo "*** jhipster: use release version"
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
I don't understand what happened here in the CI with these failures. It looks not related to your PR, we need to investigate

mshima:
> I don't understand what happened here in the CI with these failures. It looks not related to your PR, we need to investigate

Seems to be new chrome driver release.

kaidohallik:
Seems that Chrome version 79 is declared as stable since 17.12.2019 and this is incompatible with webdriver-manager 12.1.7. I tried to use webdriver-manager version 13.0.0 but the same error.

https://chromereleases.googleblog.com/2019/12/stable-channel-update-for-desktop_17.html

kaidohallik:
Hmm, error message is saying that chromedriver version is 79 - so this is latest, but Chrome version seems to be old in GitHub Actions.

In GitHub Actions page I can see that in https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners#ubuntu-1804-lts is information: **Google Chrome (Google Chrome 78.0.3904.108 )**

So one possible explanation is:
* webdriver-manager downloads chromedriver 79
* GitHub Actions uses fixed chrome version 78

And these are incompatible.

vishal423:
Looks like we don't control the version of driver to use. To fix broken builds, we can install latest chrome binary as part of build step.

kaidohallik:
I tried to add to GitHub Actions step:
```
sudo apt-get install google-chrome-stable
```

The I see in the CI logs: **google-chrome-stable is already the newest version (78.0.3904.108-1)** and nothing is installed additionally and error remains.

vishal423:
You need to update the package list before trying the install command. Anyway, I have done this change and will raise PR in next few mins

kaidohallik:
OK, thanks @vishal423 !

pascalgrimaud:
closing / reopening to trigger the builds

mshima:
@pascalgrimaud if think the patch needs to be rebased to fix the ci builds.

pascalgrimaud:
This is something I always wonder... Does it test the merge or only the branch ?
I think you're right, I need sleep :D

PR #10836, opened 2019-11-27
Strict option in home module in Angular

kaidohallik:
This is the next piece for the #10631 - home module.

One note about changes: `AccountService.identity()` must be called in `HomeComponent` only if authentication type is oauth2, for other authentication types this is already called in `LoginService.login()`: 
https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/core/login/login.service.ts.ejs#L60-L62

-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #10982, opened 2019-12-18
Fix auto login in Angular for JWT, session and UAA authentication

kaidohallik:
This fixes regression from #10836 - if user has valid authentication (valid JWT in session storage or valid cookie) then UI shows in home page like user is not logged in.  
That's because no more is executed `accountService.identity()` in `HomeComponent.ngOnInit()`.

This PR fixes that and makes logic better than this was before #10836 - now `accountService.identity()` is executed on app loading in `MainComponent.ngOnInint()` and this has the following advantage:

* this is a central place which is always executed on application loading and if user initially enters some other page than `HomeComponent` then auto login is always tried, even if initially entered component has not `accountService.identity()` in it's `ngOnInit()` method

* it's useful for pages which are not protected (for protected pages `UserRouteAccessService` executes `accountService.identity()` while checking rights) - when entering initially to such page then navbar is now correct, shows that user is logged in

It's useful also for `oauth2` authentication - if user changes redirect URI to some other unprotected page then now is correct navbar shown also in these cases.

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.ts.ejs
@@ -23,12 +23,15 @@ import { Router, ActivatedRouteSnapshot, NavigationEnd, NavigationError } from '
 import { TranslateService } from '@ngx-translate/core';
 <%_ } _%>
 
+import { AccountService } from 'app/core/auth/account.service';
+
 @Component({
     selector: '<%= jhiPrefixDashed %>-main',
     templateUrl: './main.component.html'
 })
 export class MainComponent implements OnInit {
     constructor(
+        private accountService: AccountService,
         <%_ if (enableTranslation) { _%>
         private translateService: TranslateService,
         <%_ } _%>
@@ -37,6 +40,9 @@ export class MainComponent implements OnInit {
     ) {}
 
     ngOnInit(): void {
+        // try to log in automatically
+        this.accountService.identity().subscribe();
+
         this.router.events.subscribe(event => {
             if (event instanceof NavigationEnd) {
                 this.updateTitle();
--- a/generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
+++ b/generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
@@ -186,6 +186,18 @@ describe('account', () => {
     });
 <%_ } _%>
 
+    it('should navigate to 404 not found error page on non existing route and show user own navbar if valid authentication exists', async () => {
+        // don't sign out and refresh page with non existing route
+        await browser.get('/this-is-link-to-non-existing-page');
+
+        // should navigate to 404 not found page
+        const url = await browser.getCurrentUrl();
+        expect(url).to.endWith('404');
+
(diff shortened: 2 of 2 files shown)


kaidohallik:
One problem this PR fixes is: if user has valid authentication and user tries to open application with non-existing URL then previously was shown wrong navbar (not authenticated user navbar), now is shown correct user navbar.

generators/client/templates/angular/src/main/webapp/app/layouts/main/main.component.ts.ejs
```diff
@@ -23,12 +23,15 @@ import { Router, ActivatedRouteSnapshot, NavigationEnd, NavigationError } from '
...
 
+import { AccountService } from 'app/core/auth/account.service';
+
 @Component({
     selector: '<%= jhiPrefixDashed %>-main',
     templateUrl: './main.component.html'
 })
 export class MainComponent implements OnInit {
     constructor(
+        private accountService: AccountService,
```
pascalgrimaud: maybe we need to inject it in tests too ?

kaidohallik: Good idea. I'll try to add Protractor test for that.

pascalgrimaud: I was talking about main.component.spec.ts in fact

kaidohallik: I don't know how to test this in unit tests. I wrote Protractor test, let's see if this is passing CI.

pascalgrimaud: Maybe I'm wrong as I'm not a front guy, but let me  try to explain.
I don't think a specific test is needed, but I think we need to inject the `accountService` in tests:
- in `main.component.ts`, we have `private translateService: TranslateService`
- so in `spec.ts` file, we have:
    - https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/test/javascript/spec/app/layouts/main/main.component.spec.ts.ejs#L39
    - https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/test/javascript/spec/app/layouts/main/main.component.spec.ts.ejs#L63-L65

That's why I though we need similar thing with this new `accountService`

kaidohallik: As `AccountService` is part of `TestModule`: https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/test/javascript/spec/test.module.ts.ejs#L70-L71 and `TestModule` is imported: https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/test/javascript/spec/app/layouts/main/main.component.spec.ts.ejs#L45 then `MockAccountService` is used, so should be OK.

By the way, CI with added Protractor test succeeded :)

Old code before this change didn't pass that added Protractor test.

pascalgrimaud:
good job @kaidohallik, let's merge this

kaidohallik:
@pascalgrimaud thanks for those reviews and merges in my Angular PR-s!

PR #10987, opened 2019-12-18
use lower case package instead of snake case one

avdev4j:
When using the openapi client generator and fill the project name in camel case it transform to snake case (with underscores) to named the java package.

We should use a lower case version to match with java naming convention. underscore are not forbidden but reserved to special case like when there is a special character, digit or reserved java keywords.

source: https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

 

### DIFF
--- a/generators/openapi-client/files.js
+++ b/generators/openapi-client/files.js
@@ -54,7 +54,7 @@ function writeFiles() {
                 let command;
                 if (generatorName === 'spring') {
                     this.log(chalk.green(`\n\nGenerating java client code for client ${cliName} (${inputSpec})`));
-                    const cliPackage = `${this.packageName}.client.${_.snakeCase(cliName)}`;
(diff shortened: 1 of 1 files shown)


mshima:
Cleanup of old files isn't necessary?

Seems to me that we should duplicate the cleanup for the old dir:
https://github.com/jhipster/generator-jhipster/blob/447b6f7c7b28b560c7410a6ae2ea56d6b2305f81/generators/openapi-client/files.js#L59-L62

avdev4j:
@mshima 
You're right, so this mean we need to manage both cases and delete package with the snake case name and the lower case one.

avdev4j:
I did it by removing the snake case folder everytime we generate the new one.
It works but I'm not fan to have code we will be able to explain the reason we did it in the futur

PR #10989, opened 2019-12-18
When using openapi-client with date field the java.time.LocalDate is not imported

avdev4j:
For this given api.yaml
```
definitions:
  student:
    type: 'object'
    properties:
      id:
        type: 'integer'
        format: 'int64'
      firstname:
        type: 'string'
      lastname:
        type: 'string'
      birthday:
        type: 'string'
        format: 'date'
      teacher:
        $ref: '#/definitions/teacher'
```

We have a Student model generated with a birthday date field. With our current configuration 
```
--type-mappings DateTime=OffsetDateTime,Date=LocalDate 
--import-mappings OffsetDateTime=java.time.OffsetDateTime,LocalDate=java.time.LocalDate
```
The generated field is a LocalDate one (as expected) but the import is not present and the file can't be compiled.

We have to change the DateLibrary value, use as parameter `-DdateLibrary=custom`, by removing it or use another value like 'java8'.
According to this issue https://github.com/swagger-api/swagger-codegen/issues/7795 we have some options to use but many of them seems to be legacy compliant.

I know @cbornet you also contribute to the swagger-codegen project and you probably have an idea to what is the best value for us. 

Thanks :)

### DIFF
--- a/generators/openapi-client/files.js
+++ b/generators/openapi-client/files.js
@@ -71,7 +71,7 @@ function writeFiles() {
                         ` --model-package ${cliPackage}.model` +
                         ' --type-mappings DateTime=OffsetDateTime,Date=LocalDate ' +
                         ' --import-mappings OffsetDateTime=java.time.OffsetDateTime,LocalDate=java.time.LocalDate' +
-                        ` -DdateLibrary=custom,basePackage=${this.packageName}.client,configPackage=${cliPackage},` +
+                        ` -DbasePackage=${this.packageName}.client,configPackage=${cliPackage},` +
                         `title=${_.camelCase(cliName)}`;
 
                     if (this.clientsToGenerate[cliName].useServiceDiscovery) {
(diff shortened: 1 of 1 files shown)


cbornet:
That's a very old workaround to config the generator (which btw is now openapi-generator, the fork of swagger-codegen). I'll check but it should be easier now to configure JSR310 dates.

avdev4j:
ok ! let me know if removing the "custom" value is enough or not. For sure this avoids compilation errors, but I want to be sure it's ok and support all cases. (I note for the openapi-generator name ;))

atomfrede:
We have java8 as date time config and it works fine with all date types (instant, local date and offsetdatetime) I can't remember any issues

avdev4j:
> We have java8 as date time config
Which config are you talking about ?

My case is easy to reproduce, define a `Date` field for a model in a openapi yaml file, use this yaml with the JHipster openapi subgenerator -> the model generate has a LocalDate field but not the import.

cbornet:
Actually Java8 is now the default so it should work OOTB. Can you also remove `-type-mappings` and `-import-mappings` as they are now the defaults.

PR #10991, opened 2019-12-18
Switch on strict Typescript option in Angular

### LINKED ISSUE #10631: Enable strict option for Typescript compiler in ng-jhipster

kaidohallik:
Closes #10631 

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/notification.interceptor.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/blocks/interceptor/notification.interceptor.ts.ejs
@@ -31,15 +31,15 @@ export class NotificationInterceptor implements HttpInterceptor {
         return next.handle(request).pipe(tap((event: HttpEvent<any>) => {
             if (event instanceof HttpResponse) {
                 const arr = event.headers.keys();
-                let alert = null;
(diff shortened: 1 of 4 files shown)


wmarques:
Finally! Thanks @kaidohallik ! That's a big improvement for our Angular code ! Hope this will not annoy our users alot :D

kaidohallik:
If someone doesn't like strict mode then it's easy to switch off. But if someone requires strict option then we offer this without modifying generated code now.

PR #10999, opened 2019-12-20
Optimize imports

jdubois:
Optimize imports of recently changed classes


pascalgrimaud:
@jdubois : LGTM, but we won't need this kind of thing soon, as prettier-java will do the job :)

PR #11003, opened 2019-12-20
Clean up the error.html page

jdubois:
Cleans up the page.
We should also validate if this is still useful or not, as it's not reachable anymore:

- If you are a human, you will get the error message from the front-end
- If you use the API, this is taken care of by Zalando Problems

So maybe this should be removed


pascalgrimaud:
It is still usefull for microservice application

PR #11004, opened 2019-12-20
Fix Sonar issue https://sonarcloud.io/project/issues?id=io.github.jhi…

jdubois:
…pster.sample%3Ajhipster-sample-application&issues=AWy0hYpvYwmYGWzysjfr&open=AWy0hYpvYwmYGWzysjfr

This is a false positive, it's just Sonar which doesn't understand Angular


vishal423:
btw, sonar rule also describe about deprecation of `summary` attribute

> adding a `summary` attribute to the `<table>` tag. However note that this attribute has been `deprecated` in `HTML5`.

jdubois:
Then my fix is bad, thanks @vishal423 !
I have no idea how to fix that stupid false positive, maybe I'll just fix the text so we don't use Angular for this. It's not very important, but it's so bad that visually impaired people will have a worse message, because of the "quality analysis".

jdubois:
I give up... The current code is really good, I'm not messing this up because of Sonar

Issue #11010, opened 2019-12-20
Models classes generated by openapi-generator have unknown imports


avdev4j:
##### **Overview of the issue**

Hi team,
when we use the openapi-generator (wrapped by our openapi-client subgenerator) the generated classes import the following class `import org.openapitools.jackson.nullable.JsonNullable;`.

The problem is that we don't have the `org.openapitools.jackson` dependency in our pom.

After some research it's appeard that this bug is known by the openapi community, but no fix has been done yet. https://github.com/OpenAPITools/openapi-generator/issues/2901

##### **Reproduce the error**
run the subgenerator openapi-client with an openapi spec api.yml

##### **Suggest a Fix**
I suggest to add the dependency when using our openapi-client. It's not really smart, but at least the generated code can be compiled and run. We can also add a comment in the code to link the openapi-generator issues and remember why we did the hack.

I know this a hack and I don't have a fully satisfying solution.

Otherwise the solutions suggested by the community are :
- use a workaround with a `templateDirectory` config in maven
- Post-process the generated files with https://github.com/google/google-java-format

##### **JHipster Version(s)**

```
consume-api-workshop@0.0.1-SNAPSHOT /Users/aviard/projects/perso/consumeApiWorkshop
└── (empty)

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "io.github.openapiworkshop",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "6.5.1",
    "applicationType": "monolith",
    "baseName": "consumeApiWorkshop",
    "packageName": "io.github.openapiworkshop",
    "packageFolder": "io/github/openapiworkshop",
    "serverPort": "8081",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSwaggerCodegen": false,
[...]

PR #11011, opened 2019-12-20
Add jackson-databind-nullable dependency with openapi-generator

avdev4j:
fix https://github.com/jhipster/generator-jhipster/issues/11010

### DIFF
--- a/generators/openapi-client/files.js
+++ b/generators/openapi-client/files.js
@@ -105,6 +105,8 @@ function writeFiles() {
                 return;
             }
 
+            const openApiToolsDependencyVersion = '0.2.1';
+
             if (this.buildTool === 'maven') {
                 if (!['microservice', 'gateway', 'uaa'].includes(this.applicationType)) {
                     let exclusions;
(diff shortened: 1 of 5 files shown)


generators/openapi-client/files.js
```diff
@@ -100,6 +100,8 @@ function writeFiles() {
                 return;
             }
 
+            const openApiToolsDependencyVersion = '0.2.1';
```
pascalgrimaud: I think we should put the version in this file: https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js

pascalgrimaud: then in pom.xml and gradle.properties, use this constant

generators/openapi-client/files.js
```diff
@@ -115,6 +117,9 @@ function writeFiles() {
                     this.addMavenDependency('org.springframework.cloud', 'spring-cloud-starter-openfeign', null, exclusions);
                 }
                 this.addMavenDependency('org.springframework.cloud', 'spring-cloud-starter-oauth2');
+                this.addMavenProperty('jackson-databind-nullable.version', openApiToolsDependencyVersion);
```
pascalgrimaud: should be called only if `!enableSwaggerCodegen`, otherwise, it will add twice

generators/openapi-client/files.js
```diff
@@ -126,6 +131,14 @@ function writeFiles() {
                     }
                 }
                 this.addGradleDependency('compile', 'org.springframework.cloud', 'spring-cloud-starter-oauth2');
+                this.addGradleProperty('jackson_databind_nullable_version', openApiToolsDependencyVersion);
```
pascalgrimaud: same comment than previously

avdev4j:
@pascalgrimaud All the changes have been done. I also move the code into a specific part to know what we do it and to easily remove that in the future.

pascalgrimaud:
Good job @avdev4j, I'll review it later

Issue #10630, opened 2019-10-16
Login dialog not displayed when trying to access protected page without login.


ashishjmeshram:
##### **Overview of the issue**

It seems that when an user is not logged in and if he tries to access the page which requires authentication, application should ideally ask him to login. But instead a blank page is shown.

Below is the stacktrace from the console.

```
webpack-internal:///./node_modules/@angular/core/fesm2015/core.js:5621 ERROR Error: Uncaught (in promise): HttpErrorResponse: {"headers":{"normalizedNames":{},"lazyUpdate":null},"status":401,"statusText":"Unauthorized","url":"http://localhost:8080/api/account","ok":false,"name":"HttpErrorResponse","message":"Http failure response for http://localhost:8080/api/account: 401 Unauthorized","error":{"type":"https://www.jhipster.tech/problem/problem-with-message","title":"Unauthorized","status":401,"detail":"Full authentication is required to access this resource","path":"/api/account","message":"error.http.401"}}
    at resolvePromise (webpack-internal:///./node_modules/zone.js/dist/zone.js:840)
    at resolvePromise (webpack-internal:///./node_modules/zone.js/dist/zone.js:799)
    at eval (webpack-internal:///./node_modules/zone.js/dist/zone.js:901)
    at ZoneDelegate.invokeTask (webpack-internal:///./node_modules/zone.js/dist/zone.js:435)
    at Object.onInvokeTask (webpack-internal:///./node_modules/@angular/core/fesm2015/core.js:35473)
    at ZoneDelegate.invokeTask (webpack-internal:///./node_modules/zone.js/dist/zone.js:434)
    at Zone.runTask (webpack-internal:///./node_modules/zone.js/dist/zone.js:202)
    at drainMicroTaskQueue (webpack-internal:///./node_modules/zone.js/dist/zone.js:615)
    at ZoneTask.invokeTask [as invoke] (webpack-internal:///./node_modules/zone.js/dist/zone.js:521)
    at invokeTask (webpack-internal:///./node_modules/zone.js/dist/zone.js:1675)
```

##### **Motivation for or Use Case**

Better user login experience.

##### **Reproduce the error**

Do not login to the jhipster application. Try to access pages like user management by directly typing the url in the browser window, a blank page is shown to user. 

http://localhost:8080/admin/user-management

##### **Related issues**

None

##### **Suggest a Fix**

Currently not known.

##### **JHipster Version(s)**

6.4.1

[...]

PR #11029, opened 2019-12-23
removed elastic keyword for Ids and some sonar fixes

Tcharl:
Elastic ids are creating some recurring issue due to index mismatch
Also, some sonarqube quality fixes has been made.


kaidohallik:
@tcharl looks like you are mixing unrelated different framework changes in one PR, looks like these are:
* some additional error checking in Angular `UserRouteAccessService`
* some Elasticsearch change
* upgrading Maven version
* formatting Javadoc and some other code cleaning in server side classes

It makes hard to review these changes. For example I would like to review and understand what is doing this `UserRouteAccessService` change in Angular side. I think it's better if you open different PR-s with explanation what problem one or another change resolves and how to reproduce current problem.

For this `UserRouteAccessService` change: is this duplicate of the #10630 which occurred in the v6.4.1 and was fixed in the v6.5.0 by the #10672 and this change is not needed any more for the code in the current master (since v6.5.0)?

Tcharl:
Hi @kaidohallik ,

Thank you for the advice and I totally agree with you that different PR shall be made for different purpose (like I would have done to split angular and elastic purposes).

However, making a PR to change some comment indentation, using curriyng instead of a lambda, adding comment or upgrading a patch version that we are sure that is retrocompatible is overkill IMO.

For this one, I should have done 4 if I stricly followed the rule of the law:
* One for maven version
* One for Elastic ID remove
* One for comment indentation and addition
* One for the lambda.

IMO, there's always shades of grey ;-)

Issue #10967, opened 2019-12-16
docker-compose subgenerator sets port to 80, but maps 8080


ruddell:
##### **Overview of the issue**
The monolith `docker-compose.yml` generated by `jhipster docker-compose` has conflicting configuration for ports.  It maps ports `8080:8080` but also sets `SERVER_PORT=80`

##### **Motivation for or Use Case**
The generated configuration should work out-of-the-box

##### **Reproduce the error**
- Generate monolithic app
- Run `jhipster docker-compose` and choose your monolith
- Start the stack with `docker-compose up -d`
- App is not accessible on port 80 or port 8080
- Open `docker-compose.yml` and you'll see:
```
      - JHIPSTER_SLEEP=30
      - SERVER_PORT=80
    ports:
      - '8080:8080'
```

##### **Related issues**
Introduced in https://github.com/jhipster/generator-jhipster/pull/10763

##### **Suggest a Fix**
@xetys will be more familiar with his goal for the new config, so hoping he'll know whether to remove `SERVER_PORT` or fix the port mapping.

https://github.com/jhipster/generator-jhipster/blob/b2232acecd0c7d8a754d6eca3447237ccd3972f5/generators/docker-compose/index.js#L120-L122

##### **JHipster Version(s)**
Only affects master, this change has not been released yet.

##### **JHipster configuration**
<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "6.5.1",
    "applicationType": "monolith",
    "baseName": "mono",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "maven",
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "bXktc2VjcmV0LXRva2VuLXRvLWNoYW5nZS1pbi1wcm9kdWN0aW9uLWFuZC10by1rZWVwLWluLWEtc2VjdXJlLXBsYWNl",
[...]

PR #11033, opened 2019-12-23
UAA Improvements / Docker Compose Fixes

### LINKED ISSUE #8967: Improve docker-compose for non-eureka microservices

### LINKED ISSUE #9027: UAA not working with eureka disabled

bradsheppard:
Re-introduces #10763 while also fixing #10967.

fix #8967
fix #9027

### DIFF
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -117,7 +117,11 @@ module.exports = class extends BaseDockerGenerator {
                         yamlConfig.environment.push(`JHIPSTER_REGISTRY_PASSWORD=${this.adminPassword}`);
                     }
 
-                    parentConfiguration[`${lowercaseBaseName}-app`] = yamlConfig;
+                    if (!this.serviceDiscoveryType && appConfig.applicationType !== 'gateway') {
+                        yamlConfig.environment.push('SERVER_PORT=80'); // to simplify service resolution in docker/k8s
+                    }
+
+                    parentConfiguration[`${lowercaseBaseName}`] = yamlConfig;
 
                     // Add database configuration
                     const database = appConfig.prodDatabaseType;
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -97,10 +97,15 @@ module.exports = class extends BaseDockerGenerator {
                         delete yamlConfig.ports; // Do not export the ports as Traefik is the gateway
                         this.keycloakRedirectUris += '"http://localhost/*", "https://localhost/*", ';
                     } else if (appConfig.applicationType === 'gateway' || appConfig.applicationType === 'monolith') {
-                        this.keycloakRedirectUris += `"http://localhost:${portIndex}/*", "https://localhost:${portIndex}/*", `;
-                        const ports = yamlConfig.ports[0].split(':');
-                        ports[0] = portIndex;
-                        yamlConfig.ports[0] = ports.join(':');
+                        if (!this.serviceDiscoveryType && appConfig.applicationType !== 'gateway') {
+                            const ports = [portIndex, 80];
+                            yamlConfig.ports[0] = ports.join(':');
+                        } else {
+                            this.keycloakRedirectUris += `"http://localhost:${portIndex}/*", "https://localhost:${portIndex}/*", `;
+                            const ports = yamlConfig.ports[0].split(':');
+                            ports[0] = portIndex;
+                            yamlConfig.ports[0] = ports.join(':');
+                        }
                         portIndex++;
                     }
 
--- a/generators/docker-compose/index.js
+++ b/generators/docker-compose/index.js
@@ -97,15 +97,10 @@ module.exports = class extends BaseDockerGenerator {
                         delete yamlConfig.ports; // Do not export the ports as Traefik is the gateway
                         this.keycloakRedirectUris += '"http://localhost/*", "https://localhost/*", ';
                     } else if (appConfig.applicationType === 'gateway' || appConfig.applicationType === 'monolith') {
-                        if (!this.serviceDiscoveryType && appConfig.applicationType !== 'gateway') {
-                            const ports = [portIndex, 80];
-                            yamlConfig.ports[0] = ports.join(':');
-                        } else {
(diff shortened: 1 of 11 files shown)


generators/docker-compose/index.js
```diff
@@ -117,7 +122,11 @@ module.exports = class extends BaseDockerGenerator {
                         yamlConfig.environment.push(`JHIPSTER_REGISTRY_PASSWORD=${this.adminPassword}`);
                     }
 
-                    parentConfiguration[`${lowercaseBaseName}-app`] = yamlConfig;
+                    if (!this.serviceDiscoveryType && appConfig.applicationType !== 'gateway') {
```
xetys: This is the place which is misconfigured. We should do it only for microservices + uaa. Considering our future roadmap, a better condition would be `appConfig.skipClient`

bradsheppard: Thanks for clarifying @xetys. I made your requested change.

generators/kubernetes/templates/service.yml.ejs
```diff
@@ -35,4 +35,9 @@ spec:
   <%_ } _%>
   ports:
   - name: http
+  <%_ if (app.applicationType === 'microservice' || app.applicationType === 'uaa') { _%>
```
pascalgrimaud: Unless there is a really good reason for this, I'd prefer using the condition : `!app.serviceDiscoveryType || app.applicationType === 'uaa'` -> so it will be specific to UAA and will not impact the other option as the other options works well with Kubernetes

bradsheppard: @pascalgrimaud Thanks, I made your requested change.

pascalgrimaud:
Thanks @bradsheppard for taking care of this

bradsheppard:
bounty claimed: https://opencollective.com/generator-jhipster/expenses/12722

pascalgrimaud:
just approuved, thanks @bradsheppard

PR #11035, opened 2019-12-25
Polish Angular templates

kaidohallik:
This is Angular templates polishing PR.

Makes code in templates more similar to the end code after Prettier reformation (these changes doesn't change code for end users).

Changes that modify code for end users.

1. Some code shortening and/or adding blank lines to make code better readable:
    * `audits.component` `onSuccess`
    * `sessions.service`
    * `user-management-detail.component` `ngOnInit`
    * `notification.interceptor`
    * `find-language-from-key.pipe`
    * `entity-management-detail.component` `ngOnInit`
    * `user.service.spec`
    * `entity-management.service.spec`

2. Removing unnecessary code:
    * `docs.component`

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
@@ -123,10 +123,6 @@ export class AuditsComponent implements OnInit {
 
     private onSuccess(audits: Audit[] | null, headers: HttpHeaders): void {
         this.totalItems = Number(headers.get('X-Total-Count'));
-        if (audits === null) {
-            this.audits = [];
-        } else {
-            this.audits = audits;
-        }
+        this.audits = audits ? audits : [];
     }
 }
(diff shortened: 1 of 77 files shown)


generators/client/templates/angular/src/main/webapp/app/admin/audits/audits.component.ts.ejs
```diff
@@ -123,10 +123,6 @@ export class AuditsComponent implements OnInit {
 
     private onSuccess(audits: Audit[] | null, headers: HttpHeaders): void {
         this.totalItems = Number(headers.get('X-Total-Count'));
-        if (audits === null) {
-            this.audits = [];
-        } else {
-            this.audits = audits;
-        }
+        this.audits = audits ? audits : [];
```
MathieuAA: Simpler: `this.audits = audits || []`

kaidohallik: Thanks for the hint! I simplified this and other places. While doing these changes made some more cleaning:
* in `user-management.route` removed unnecessary `null` assignment
* in `no-pagination-template` removed assigning empty string to `this.currentSearch` because this is already empty string (if not empty then this code is never reached)

Issue #10682, opened 2019-10-25
Use Spring Security test support in REST resource tests


codecholeric:
##### **Overview of the feature request**

I was wondering, why you don't use Spring Security testing support for your REST resource integration tests, but instead bypass the Spring proxy by manually constructing the resource.

E.g. in `UserResourceIT` the line

```
@BeforeEach
public void setup() {
    UserResource userResource = new UserResource(userService, userRepository, mailService);
    // ...
}
```

instead of just using

```
@Autowired
UserResource userResource;
```

and then something like

```
@Test
@Transactional
@WithMockUser(roles = {"ADMIN"})
public void createUser() throws Exception {
    // ...
}
```

##### **Motivation for or Use Case**

It would be closer to the real setup, it would be possible to test permissions as well, and it would need less boiler plate, since you wouldn't necessarily have to inject all dependencies and manually construct the resource.

I would be happy to contribute a PR changing this, but I wanted to ask first, if you had any reason to do it like this?

-   [x] Checking this box is mandatory (this is just to show you read everything)

atomfrede:
I can't remember any specific reason and looking at the code I would say using `@WithMockUser` seems like the better solution indeed

codecholeric:
So I guess I'll create a PR then changing this, if there are no arguments against it :smiley:

atomfrede:
Please go ahead, it will make our generated code easier to understand :+1:

qmonmert:
@pascalgrimaud I began to work on this feature #11047

codecholeric:
@qmonmert I know I'm late on this (didn't find the time up to Christmas), but we should definitely align on that! I've already changed many files, my problems are only, that there is such a combinatoric explosion with `reactive` vs `non-reactive` and all the options.
You can check out my current state here:
https://github.com/jhipster/generator-jhipster/pull/11050

codecholeric:
I would also want to gather some feedback though, if the infrastructure setup is easy to understand like that.

qmonmert:
@codecholeric ah ok no problem

codecholeric:
I also have one problem, that the `DomainUserDetailsServiceIT` fails if the 2nd level cache is selected, because the cache section would be closed, and I don't know how it is connected to my changes :thinking: 
Anyway, it will still take some time to test due to the vast possibilities of configuration... I only tested one simple case so far, that it really works (and considering the CI, there is much more testing necessary :wink:)
[...]

PR #11050, opened 2019-12-28
use more Spring Security testing support

codecholeric:
Many REST resource integration tests manually assemble the resource under test to bypass Spring Security proxies. If we consistently use Spring Security testing support, we can use one global MockMvc configuration to apply to all tests. This reduces boiler plate and makes the testing strategy more consistent.

Resolves: #10682


generators/server/templates/src/test/java/package/config/NoOpMailConfiguration.java.ejs
```diff
@@ -0,0 +1,40 @@
+<%#
+Copyright 2013-2019 the original author or authors from the JHipster project.
```
DanielFran: Update copyright years to 2013-2020

generators/server/templates/src/test/java/package/security/DomainUserDetailsServiceIT.java.ejs
```diff
@@ -69,7 +69,7 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 @ExtendWith(RedisTestContainerExtension.class)
 <%_ } _%>
 <%_ if (databaseType === 'sql') { _%>
-@Transactional
+@Transactio nal
```
DanielFran: Typo?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
```diff
@@ -55,89 +54,73 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import static com.tngtech.sstestsupport.web.rest.AccountResourceIT.TEST_USER_LOGIN;
```
DanielFran: Should this line be removed?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
```diff
@@ -55,89 +54,73 @@ import java.util.Collections;
...
             .andExpect(content().string(""));
     }
 
     @Test
     public void testAuthenticatedUser() throws Exception {
-        restUserMockMvc.perform(get("/api/authenticate")
+        restAccountMockMvc.perform(get("/api/authenticate")
             .with(request -> {
-                request.setRemoteUser("test");
+                request.setRemoteUser("someUser");
```
DanielFran: Maybe create also a constant?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
```diff
@@ -168,65 +151,53 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 <%_ } else { _%>
 import org.springframework.test.web.reactive.server.WebTestClient;
 <%_ } _%>
+import static com.tngtech.sstestsupport.web.rest.AccountResourceIT.TEST_USER_LOGIN;
```
DanielFran: Should this line be removed?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
```diff
@@ -304,9 +276,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 <%_ if (authenticationType === 'session' && !reactive) { _%>
 import static org.hamcrest.Matchers.hasItem;
 <%_ } _%>
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.when;
+import static com.tngtech.sstestsupport.web.rest.AccountResourceIT.TEST_USER_LOGIN;
```
DanielFran: Should this line be removed?

generators/server/templates/src/test/java/package/web/rest/UserResourceIT.java.ejs
```diff
@@ -67,6 +67,7 @@ import org.springframework.cache.support.NoOpCacheManager;
 import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
 <%_ } _%>
 import org.springframework.http.MediaType;
+import org.springframework.security.test.context.support.WithMockUser;
```
DanielFran: This should go to line 74

generators/server/templates/src/test/java/package/web/rest/WebTestConfiguration.java.ejs
```diff
@@ -0,0 +1,47 @@
+<%#
+Copyright 2013-2019 the original author or authors from the JHipster project.
```
DanielFran: Update copyright years to 2013-2020

generators/server/templates/src/test/java/package/web/rest/WithUnauthenticatedMockUser.java.ejs
```diff
@@ -0,0 +1,42 @@
+<%#
+Copyright 2013-2019 the original author or authors from the JHipster project.
```
DanielFran: Update copyright years to 2013-2020

generators/server/templates/src/test/java/package/web/rest/WebTestConfiguration.java.ejs
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+<%_ } _%>
+
+@Configuration
+public class WebTestConfiguration {
+    <%_ if (reactive) { _%>
+    @Bean
+    public WebTestClient webTestClient(ApplicationContext context) {
+        return WebTestClient.bindToApplicationContext(context).build();
```
murdos: I wonder if we should not try to use `@AutoConfigureWebTestClient` on each controller test instead (and `@AutoConfigureMockMvc` for non reactive application)

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "generator-jhipster",
-    "version": "6.6.0",
+    "version": "6.6.0-P",
```
murdos: Local modification I guess. Don't forget to revert it 😉

generators/server/templates/src/test/java/package/web/rest/WebTestConfiguration.java.ejs
codecholeric: I'll check out if that works!

generators/server/templates/src/test/java/package/config/NoOpMailConfiguration.java.ejs
codecholeric: Done :+1:

generators/server/templates/src/test/java/package/security/DomainUserDetailsServiceIT.java.ejs
codecholeric: Yeah, don't know how that happened, must have been in the last moment before I pushed (a little rushed because I wanted to share my progress)... :frowning_face:

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
codecholeric: If you want to use the constant within the annotation I think you need it, don't you?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
codecholeric: I usually don't create constants for things I only use in a very local context. Or do you mean a variable within the test method?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
codecholeric: See above, I don't think it compiles without it, if you want to use the constant within the annotation...

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
codecholeric: See above

generators/server/templates/src/test/java/package/web/rest/UserResourceIT.java.ejs
codecholeric: Done :+1:

generators/server/templates/src/test/java/package/web/rest/WebTestConfiguration.java.ejs
codecholeric: Done :+1:

generators/server/templates/src/test/java/package/web/rest/WithUnauthenticatedMockUser.java.ejs
codecholeric: Done :+1:

generators/server/templates/src/test/java/package/web/rest/WebTestConfiguration.java.ejs
codecholeric: Looks good, I think we should go this way, I've adjusted the PR!

package-lock.json
codecholeric: Yep, done :wink:

codecholeric:
One thing I wonder is about `NoOpMailConfiguration`. Principally I think it's a good idea to override it for the whole test scope, but the way it is now, it's only used for `AccountResourceIT` to mimic the old behavior. I've also noticed the test will also pass without it, because the mail sending happens `@Async` in the background, and will fail in an attempt to use `localhost` as a mail server.
It's cleaner to mock out the mail sending for tests, but I've noticed, that `UserResourceIT` seems to send mails anyway in the same way (causing an exception about the unreachable mail server `localhost`).
So I wonder if it would not be better to either consistently mock the `MailService` for all tests, or to live with the error for all tests. To adjust it for `AccountResourceIT` but not for `UserResourceIT` seems a little strange to me? Or am I wrong about the behavior?

Also I was surprised that there is no `@WithUnauthenticatedMockUser` out of the box? Does anyone know a better solution here? It's like all tests want `@WithMockUser` except for one. I know there is `@WithAnonymousUser`, but that is not quite what I want, because that simulates an authenticated user without authorities AFAIS?

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
DanielFran: Yes, I mean a variable

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
DanielFran: Ok

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
DanielFran: Ok

generators/server/templates/src/test/java/package/web/rest/AccountResourceIT.java.ejs
codecholeric: Done :+1:

codecholeric:
Hmm, it might take a little longer to get all those tests green :frowning_face: With some tests I'm also not sure (e.g. no heap space).

Off-topic: Not trying to be a smartass (always easy afterwards :wink:), but have you guys ever thought about moving this in the direction of a richer model for the templates? (I.e. somehow making it more about extension instead of modification, encapsulate the behavior belonging to an option together instead of sprinkled if-elses, etc.). Because I've spent about 3 full work days now and in particular templates like `AccountResourceIT` feel a lot like
Peter Griffin Window Blinds

Anyway, if someone feels like pitching in, I'd be really happy. Because it feels pretty far, but frankly I'm starting to get problems finding the time and there are so many combinations that need to be checked (and the turn-around is not super quick until you've reproduced the error locally). I think there is still some problem with the caching, but it behaves inconsistently (i.e. I can't reproduce the exact errors from the CI, but I get some errors about closed caches that I don't understand)

codecholeric:
At least I think I found the caching problem, in `CacheConfiguration.createCache(..)` the "native" `CacheManager` (e.g. `Eh107CacheManager`) is used to destroy an existing cache before creating it. However this seems to leave the Spring context in an inconsistent state, i.e. when cache loading is triggered via an aspect (e.g. `@Cacheable`), it is now referencing the old `Eh107Cache` which has already been closed...
Have to look into it, how to resolve this. Because on the other hand it looks like the Spring `CacheManager` is not available yet at the time when this operation is called.
A simple alternative might be to just add `@DirtiesContext` to the tests if 2nd level cache is active, but I don't think that's a great solution for performance (and custom tests probably run into the same problem). Another solution would be to deactivate the 2nd level cache for automatic tests, might not be bad, just loses a little bit of test coverage (but that hasn't been there before my changes either). Last but not least I could rewrite the `JCacheManagerCustomizer` to not destroy an existing cache but instead just reuse it. I'm not completely sure why an existing cache would have to be destroyed anyway? My intention would be that this case should never happen in production anyway, because it does not work reliably, does it? Any suggestions or preferences? :wink:

codecholeric:
I just tested and

```
private void createCache(javax.cache.CacheManager cm, String cacheName) {
    javax.cache.Cache<Object, Object> cache = cm.getCache(cacheName);
    if (cache == null) {
        cm.createCache(cacheName, jcacheConfiguration);
    }
}
```

would fix the test failures. I don't see any problem with this, so I'll go with it for now, if there are any objections, feel free :wink:

murdos:
@codecholeric: Could you resolve the conflicts? Also, do you need some help to finish this feature? It would be really great to have it 😉

codecholeric:
@murdos just did! Sorry for the delay. Now I just have the problem, that the failing tests seem mainly to be `OutOfMemoryErrors`, where I'm not sure if this really is a "failure", or if the application might just need a little more heap now because the testing context is bigger.
I could imagine that `@Autoconfigure...` initializes a little more than before...
Any opinions on that?
Also, is there an easy way to see what exactly the configuration was, when e.g. `Angular / ngx-gradle-psql-es-noi18n-mapsid (pull_request)` fails? Couldn't spontaneously find all the config parameters to reproduce this locally...
Or maybe as a related question, how do you guys test locally? I simply manually generate the application with my linked local version and "click" through the setup, then run `./gradlew integrationTest`. Is there a better way?
A little support how to track down those last problems would definitely be appreciated :smiley: (in particular because pretty much every bigger change of the tests will cause a painful merge conflict to resolve...)

codecholeric:
Oh, and how to checkout what failed with these Azure pipelines, where would those build logs be available that the jobs refer to? Because for those I really have no idea how to find out more information...

SudharakaP:
@codecholeric : For the Azure build logs click on the `errors/warnings` link as shown below.

image

For reproducing the configuration, you might find the JDL and `.yo-rc.json` printed out in the logs under `GENERATION: jhipster info` section of the builds (both Azure and Github Actions). 

Hope this answers your questions; let us know otherwise. 😄

codecholeric:
Thanks, I'll try to fix the rest of the problems then :+1: At least I guess it should be possible to find out of the heap was just a little too small or if there is some resource leak somehow in these test configurations.

DanielFran:
Closing/Reopenning to launch tests

DanielFran:
@codecholeric Do you need some help to finalise this PR?
I am for the same opinion of @murdos; this would be awesome to have it finally.

murdos:
@pascalgrimaud : is it possible to add a bounty here for @codecholeric ? He's doing a tedious work to make things work in all contexts :+1:

codecholeric:
@DanielFran sorry, this does now really involve some tedious debugging for `x^y` configurations to find out why some things fail for some config and not for some other :frowning: E.g. took me a while to find that now `SpaWebFilter` redirects `/test/` to `/index.html` for reactive apps, while for MVC there is an explicit ANT ignore for `/test/**` in the `SecurityConfiguration`.
But yes, I could definitely use some advice, because sometimes I'm not completely sure why tests behave differently / what the intention was. Can you for example tell me, why `/api/authenticate` should be accessible and return `200` without being authenticated for all configurations, but not for `oauth2` (see `AccountResourceIT.testNonAuthenticatedUser`)?
I'm surprised why the behavior should be different, shouldn't it be either `/api/authenticate` can in all configurations be accessed unauthenticated or just authenticated?

Also if somebody wants to grab a test / problem, feel free to write a comment about it and pick it. At the moment I try to get the behavior consistent / okay for the above use case in different configurations, but I need to understand why the configuration actually is different for `oauth2`...

There is still a problem with an `EntityResourceIT`, I have tried to migrate those tests, but then realized that in some places they create their own custom MockMVC within the test. So I'm not sure how to proceed there, it feels a little inconsistent if I only migrate a part. Also I could not reproduce this failure locally, even when running exactly the same config with the same entity generation?

codecholeric:
But I agree that it's worth it to push this through faster, because as soon as new tests are added, in particular if new reactive stuff pops up, this always causes big extra effort on this branch.

pascalgrimaud:
@murdos : unless I'm wrong, there is already a bounty on the ticket (not the PR) But I can increase the bounty if you think it's not enough

codecholeric:
Just that it really consumes quite some time to create all the failing configurations, then debug into the gory Spring details and find out why things behave differently. FYI, for the above case with the `SpaWebFilter`, I've excluded `/test/**` :wink: I hope that is okay, but I figured if it's okay for `SecurityConfiguration` it does not make any sense to be more strict for `SpaWebFilter`. But maybe we should revisit both places and think about solving this without hacks in the production code excluding `/test/**` or similar (at least in my eyes such a thing smells like a hack, because it opens `/test/**` on production configs, and I don't think that's a best practice)

mraible:
@codecholeric If you have issues with the reactive tests, you can ask @cbornet or myself to look at them. I can help with OAuth tests too.

codecholeric:
Thanks @mraible!! Maybe you can tell me, why in `SecurityConfiguration.java.ejs:262` there is this if statement for `oauth2`? Because it seems to me that this is the reason why an unauthenticated user can access `/api/authenticate` for non `oauth2` configs, but the tests fail with `401` if it is an `oauth2` config.
My problem is that I do not understand why the test should behave the same, if there is the different configuration of Spring Security? Or in other words, it seems fair to me, that the `oauth2` test fails trying to get a `200`, because in this case the ANT matcher does not exclude `/api/authenticate` from security checking.
Before my changes I think the security filter chain was simply not applied, so it did not make any difference, but maybe I'm wrong?

mraible:
> ... why in `SecurityConfiguration.java.ejs:262` there is this if statement for `oauth2`? Because it seems to me that this is the reason why an unauthenticated user can access `/api/authenticate` for non `oauth2` configs, but the tests fail with `401` if it is an `oauth2` config.

The reason `/api/authenticate` is not included in `SecurityConfiguration` is because the client never calls it. I didn't think it was useful to include it if it's never used. 

I tried adding it in a newly created OAuth 2.0 app and unit, integration, and e2e tests pass just fine.

codecholeric:
Thanks @mraible! But does it then make sense to even have this method at all when `oauth2` is selected? Before my application of Spring test support, the security advice was simply not added to `AccountResource`, so that's why with `oauth2` it behaved like the other configs in the test...

But if I understand you correctly, would it make sense to simply not generate the endpoint `/api/authenticate` for `oauth2` at all and thus also not the test method?

If I would just make the test pass, I could simply assert that for `oauth2` the result is `unauthorized` instead of `ok`, but I don't think that makes a lot of sense, if the endpoint is not used and the test unrealistic (it would just make people wonder again next time, why the endpoint behaves differently for `oauth2` and non-`oauth2` configs...)

codecholeric:
Is it possible that the `cassandra` fails that suddenly popped up after my last push are connected to https://github.com/jhipster/generator-jhipster/issues/11169? Because I can't reproduce them locally :frowning:

mraible:
> But if I understand you correctly, would it make sense to simply not generate the endpoint `/api/authenticate` for `oauth2` at all and thus also not the test method?

Yes, I believe so.

codecholeric:
Now I'm at a point where I don't know exactly how to go on:

* as mentioned [React / react-gradle-cassandra-session-redis](https://github.com/jhipster/generator-jhipster/pull/11050/checks?check_run_id=417774392) looks to me like https://github.com/jhipster/generator-jhipster/issues/11169
* [jhipster.generator-jhipster (Test ms-ngx-gateway-eureka-uaa)](https://github.com/jhipster/generator-jhipster/pull/11050/checks?check_run_id=417774372) I cannot reproduce no matter what I try :frowning: I've generated the same project in the same folder and I'm executing the commands from `21-tests-backend.sh`, but the build runs successfully on my machine. Do you have any tip how I could go on then? Or could someone else try to run that build and see if it fails?

SudharakaP:
> Is it possible that the `cassandra` fails that suddenly popped up after my last push are connected to #11169? Because I can't reproduce them locally 😦

@codecholeric : Yes, I believe the errors you see are connected to #11169. You can safely ignore those. We are currently trying to fix that problem.

SudharakaP:
> Now I'm at a point where I don't know exactly how to go on:
> 
> * as mentioned [React / react-gradle-cassandra-session-redis](https://github.com/jhipster/generator-jhipster/pull/11050/checks?check_run_id=417774392) looks to me like #11169
> * [jhipster.generator-jhipster (Test ms-ngx-gateway-eureka-uaa)](https://github.com/jhipster/generator-jhipster/pull/11050/checks?check_run_id=417774372) I cannot reproduce no matter what I try 😦 I've generated the same project in the same folder and I'm executing the commands from `21-tests-backend.sh`, but the build runs successfully on my machine. Do you have any tip how I could go on then? Or could someone else try to run that build and see if it fails?

@codecholeric : I see that you've changed `UserResourceIT`, `AuditResourceIT` and `AccountResourceIT` classes. I've did some similar changes and normally those errors (which are due to test failures) can be reproduced locally. Did you run `./mvnw clean verify` in your local machine?

codecholeric:
@SudharakaP yes. I've meanwhile reproduced a ton of these errors locally :wink: However with that specific one I'm out of luck :frowning: Do you think you could check out my branch, generate the project and see if it fails for you? Maybe I'm overlooking something there, don't know. I tried 4 times to reproduce it though (first time I switched it to Gradle because the integration tests seem easier to run then but after that I generated it with exactly the provided `.yo-rc.json` and ran the `./mvnw clean verify` command from `21-tests-backend.sh`) and couldn't. I always purged the whole project and regenerated it. So I don't know if I'm overlooking anything here, but for all the other configurations I managed to reproduce the behavior the way I did it...

DanielFran:
Closing/Reopening lo launch tests

codecholeric:
Ah, I've finally been able to reproduce `Test ms-ngx-gateway-eureka-uaa` locally :+1: (didn't realize that in this case it generates a second application to another folder...
Gonna look into it...

codecholeric:
Oh my God!! *breaksDownUncontrollablyCrying* :sob: I can't believe all the tests are green.
I would be happy if someone could take a look if all my changes make sense the way they are now :wink:

SudharakaP:
@codecholeric : Sorry Peter, I couldn't help you with the testing, was busy with some other stuff. Hope you figured things out now? 😄

codecholeric:
@SudharakaP at least the tests are all green for the first time :wink:

generators/server/templates/src/main/java/package/web/filter/SpaWebFilter.java.ejs
```diff
@@ -33,7 +33,7 @@ public class SpaWebFilter implements WebFilter {
     @Override
     public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
         String path = exchange.getRequest().getURI().getPath();
-        if (!path.startsWith("/api") && !path.startsWith("/management")<% if (authenticationType === 'oauth2') { %> && !path.startsWith("/login")<% } %>
+        if (!path.startsWith("/api") && !path.startsWith("/management") && !path.startsWith("/test")<% if (authenticationType === 'oauth2') { %> && !path.startsWith("/login")<% } %>
```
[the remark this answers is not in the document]
codecholeric: @murdos I completely agree that this is quite ugly :frowning_face: (but as I wrote, so is the exception in `SecurityConfiguration` :wink:) I thought I ran into problems when doing this (it was in fact my first attempt), because in certain configurations being beneath `/api` some other stuff kicked in that let the tests fail. However a quick test with a reactive and a non-reactive config that I did again right now seemed to work, so I'll see what the tests of all configs say about this change! If it's good, I'll fix up that old commit and remove this exception. (if not we could still change the path conditionally for reactive tests to get rid of this exception, but I thought that was somewhat ugly and complex, too)

test/utils/expected-files.js
```diff
@@ -234,8 +234,10 @@ const expectedFiles = {
         `${SERVER_MAIN_SRC_DIR}com/mycompany/myapp/web/rest/UserResource.java`,
         `${SERVER_MAIN_SRC_DIR}com/mycompany/myapp/web/rest/vm/KeyAndPasswordVM.java`,
         `${SERVER_MAIN_SRC_DIR}com/mycompany/myapp/web/rest/vm/ManagedUserVM.java`,
+        `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/config/NoOpMailConfiguration.java`,
         `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/web/rest/AuditResourceIT.java`,
         `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/web/rest/UserResourceIT.java`,
+        `${SERVER_TEST_SRC_DIR}com/mycompany/myapp/web/rest/WithUnauthenticatedMockUser.java`,
```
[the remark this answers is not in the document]
codecholeric: Done :+1:

generators/server/templates/src/main/java/package/web/filter/SpaWebFilter.java.ejs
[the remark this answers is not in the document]
codecholeric: I do remember now, it was because CSRF would kick in then for `session` and `oauth2`. But it seems to work by adding CSRF test support to `ExceptionTranslatorIT`. So I'll try to go that way because it makes the change of prod code unnecessary :smile: 
Also gonna unify CSRF conditions in tests a little... (`<%_ if (authenticationType == 'session' || authenticationType == 'oauth2' || ...) _%>.with(csrf())` is quite ugly) Would of course be even cooler if single tests wouldn't have to react in any way to this config, but it could be applied in a central point, but that is something for a separate PR :wink:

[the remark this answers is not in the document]
codecholeric: I do remember now, it was because CSRF would kick in then for `session` and `oauth2`. But it seems to work by adding CSRF test support to `ExceptionTranslatorIT`. So I'll try to go that way because it makes the change of prod code unnecessary :smile: 
Also gonna unify CSRF conditions in tests a little... (`<%_ if (authenticationType == 'session' || authenticationType == 'oauth2' || ...) _%>.with(csrf())` is quite ugly) Would of course be even cooler if single tests wouldn't have to react in any way to this config, but it could be applied in a central point, but that is something for a separate PR :wink:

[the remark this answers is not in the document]
codecholeric: @murdos I think I figured it out :smiley: Can you check it again? I fixed up the commit where I had added that ugly hack with `/test` and in my last commit I made the `NoOpMailConfiguration` I had introduced to mock the `MailService` for `AccountResourceIT` transparent, i.e. just put `@Configuration` on it and let it be picked up by Spring if it's in the test classpath. I think this makes sense because there was always some stacktrace from `UserResourceIT` because there was no mocking there and I think in the test scope `MailService` should generally be mocked for all tests (except `MailServiceIT`, but that one creates the `MailService` anyway and is thus not affected). Makes caching of contexts between tests more efficient as well.

codecholeric:
One thing I was wondering about: There is `SecurityBeanOverrideConfiguration` which claims in its Javadoc that `This configuration must be included in {@code @SpringBootTest} in order to take effect.`. I don't think that this is really accurate, because the class is annotated with `@Configuration` and is thus picked up automatically in `@SpringBootTest`.
Is this the intended behavior and the comment wrong? Or should it actually have been `@TestConfiguration` so that it does not automatically get picked up?
This is effectively no problem (the tests are green) and not related to this PR though :wink: (I have not changed that class) But I was still wondering about it...

murdos:
Thanks @codecholeric !

codecholeric:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/13770
(hope I followed the process correctly :wink:)

pascalgrimaud:
just approuved.
Thanks for your hard work here @codecholeric

PR #10521, opened 2019-09-30
Add rename file and move core functionality into core in Angular

kaidohallik:
This PR is related to closed PR #10340. Two PR-s, #10360 and #10492, already implemented some parts of initial #10340.  
I initially canceled finishing #10340 and added https://github.com/jhipster/generator-jhipster/issues/10220#issuecomment-528311220 because one requirement from @vishal423 was not met: Git commit window didn't show that files are renamed.  
After doing #10492 I looked at that again and found out that to fulfill this requirement I need to add basically one additional line of code to #10340: `git mv ${oldFile} ${newFile}`  
So I create this PR with remaining #10340 changes with addition of this `git mv` command. Description of this PR:
* adds functionality to rename files in upgrade process
* this added rename functionality is used to move Angular core classes under core folder

Two images to illustrate output after using this version.

Output from generator console:

regenerate-console

Output from Git commit window:

git-commit

-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #11063, opened 2019-12-31
Remove spring security SecurityExpressionHandler for oauth2

### LINKED ISSUE #10469: Jhipster version 6.x.x OAuth2MethodSecurityExpressionHandler never register automatic

Bug appear when generate microservice application
When I use jhipster version 6.x.x OAuth2MethodSecurityExpressionHandler never register automatic. So when i call to a controller with @PreAuthorize("#oauth2.hasScope('web-app')") annotation it always throw Exception "Attempted to call method hasScope(java.lang.String) on null context object".
**But it not has any problem with jhipster version 5.x.x with the same config**

Here is my config with error:

`
@Configuration
@EnableResourceServer
@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
public class SecurityConfiguration extends ResourceServerConfigurerAdapter {
    private final OAuth2Properties oAuth2Properties;

    public SecurityConfiguration(OAuth2Properties oAuth2Properties) {
        this.oAuth2Properties = oAuth2Properties;
    }

    @Override
    public void configure(HttpSecurity http) throws Exception {
        http
            .csrf()
            .disable()
            .headers()
            .frameOptions()
            .disable()
        .and()
            .sessionManagement()
            .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
        .and()
            .authorizeRequests()
            .antMatchers("/api/**").authenticated()
            .antMatchers("/management/health").permitAll()
            .antMatchers("/management/info").permitAll()
            .antMatchers("/management/prometheus").permitAll()
            .antMatchers("/management/**").hasAuthority(AuthoritiesConstants.ADMIN);
    }

    @Bean
    public TokenStore tokenStore(JwtAccessTokenConverter jwtAccessTokenConverter) {
        return new JwtTokenStore(jwtAccessTokenConverter);
    }

    @Bean
    public JwtAccessTokenConverter jwtAccessTokenConverter(OAuth2SignatureVerifierClient signatureVerifierClient) {
        return new OAuth2JwtAccessTokenConverter(oAuth2Properties, signatureVerifierClient);
    }

    @Bean
	@Qualifier("loadBalancedRestTemplate")
    public RestTemplate loadBalancedRestTemplate(RestTemplateCustomizer customizer) {
        RestTemplate restTemplate = new RestTemplate();
        customizer.customize(restTemplate);
        return restTemplate;
    }

    @Bean
    @Qualifier("vanillaRestTemplate")
    public RestTemplate vanillaRestTemplate() {
        return new RestTemplate();
    }
}
`

This is image when i debug, it load DefaultMethodSecurityExpressionHandler
image

##### **Reproduce the error**
java.lang.IllegalArgumentException: Failed to evaluate expression '#oauth2.hasScope('web-app')'
	at org.springframework.security.access.expression.ExpressionUtils.evaluateAsBoolean(ExpressionUtils.java:30)
	at org.springframework.security.access.expression.method.ExpressionBasedPreInvocationAdvice.before(ExpressionBasedPreInvocationAdvice.java:59)
	at org.springframework.security.access.prepost.PreInvocationAuthorizationAdviceVoter.vote(PreInvocationAuthorizationAdviceVoter.java:72)
	at org.springframework.security.access.prepost.PreInvocationAuthorizationAdviceVoter.vote(PreInvocationAuthorizationAdviceVoter.java:40)
	at org.springframework.security.access.vote.AffirmativeBased.decide(AffirmativeBased.java:63)
	at org.springframework.security.access.intercept.AbstractSecurityInterceptor.beforeInvocation(AbstractSecurityInterceptor.java:233)
	at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:65)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:688)

Caused by: org.springframework.expression.spel.SpelEvaluationException: EL1011E: Method call: Attempted to call method hasScope(java.lang.String) on null context object
	at org.springframework.expression.spel.ast.MethodReference.throwIfNotNullSafe(MethodReference.java:153)
	at org.springframework.expression.spel.ast.MethodReference.getValueRef(MethodReference.java:82)
	at org.springframework.expression.spel.ast.CompoundExpression.getValueRef(CompoundExpression.java:68)
	at org.springframework.expression.spel.ast.CompoundExpression.getValueInternal(CompoundExpression.java:89)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getTypedValue(SpelNodeImpl.java:114)
	at org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:300)
	at org.springframework.security.access.expression.ExpressionUtils.evaluateAsBoolean(ExpressionUtils.java:26)
	... 119 common frames omitted

##### **Suggest a Fix**

I fix it by two step:
1. Add a class 

`
@Configuration
@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
public class MethodSecurityConfig extends GlobalMethodSecurityConfiguration {

    @Override
    protected MethodSecurityExpressionHandler createExpressionHandler() {
        return methodSecurityExpressionHandler();
    }

    @Bean
    public MethodSecurityExpressionHandler methodSecurityExpressionHandler() {
        return new OAuth2MethodSecurityExpressionHandler();
    }
}
`

2. Remove @EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true) on 
SecurityConfiguration class

This is image after i change config, it load OAuth2MethodSecurityExpressionHandler so #oauth2.hasScope will execute without any error
image

##### **JHipster Version(s)**

Version 6.x.x

Tcharl:
Rollback proposal which should fix #10469


generators/server/templates/src/main/java/package/config/MethodSecurityConfiguration.java.ejs
```diff
@@ -19,16 +19,22 @@
 package <%= packageName %>.config;
 
 import org.springframework.context.annotation.Configuration;
+<%_ if (authenticationType !== 'oauth2') { _%>
```
vishal423: as this file template is skipped in the oauth2 flow, do you still require these conditionals?

generators/server/templates/src/main/java/package/config/MethodSecurityConfiguration.java.ejs
```diff
@@ -19,16 +19,22 @@
...
 import org.springframework.security.access.expression.method.MethodSecurityExpressionHandler;
+<%_ } _%>
 import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
 import org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration;
+<%_ if (authenticationType !== 'oauth2') { _%>
```
vishal423: same as above

generators/server/templates/src/main/java/package/config/MethodSecurityConfiguration.java.ejs
```diff
@@ -19,16 +19,22 @@
...
 import org.springframework.security.oauth2.provider.expression.OAuth2MethodSecurityExpressionHandler;
+<%_ } _%>
 
 @Configuration
 @EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
```
vishal423: Except UAA, you also need to revert to add this annotation back in the `SecurityConfiguration` class

generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
```diff
@@ -135,6 +135,11 @@ public final class SecurityUtils {
      * If the current user has a specific authority (security role).
      * <p>
      * The name of this method comes from the {@code isUserInRole()} method in the Servlet API.
+<%_ if (authenticationType === 'oauth2') { _%>
+     * /!\ Call to this method won't check Oauth2 SCOPEs but only groups and roles.
```
vishal423: I believe the method name itself clarifies this part and the comment can be omitted

generators/server/templates/src/main/java/package/config/MethodSecurityConfiguration.java.ejs
Tcharl: Sorry, a mistake on my side: what about keeping the class for oauth2 and keeping these guards?

vishal423: Unless we have a use case to override the default behavior, I don't see any value on having this additional empty configuration class.

Tcharl: It's easier for the ones that wants to do it, for example, I've my own on my project:

```
private final AuthorizedOnDTOEvaluator evaluator;

    // All beans here needs to be Lazy otherwise it'll break AOP (cache, transactions, etc...)
    public MethodSecurityConfiguration(@Lazy AuthorizedOnDTOEvaluator evaluator) {
        this.evaluator = evaluator;
    }

    @Override
    protected MethodSecurityExpressionHandler createExpressionHandler() {
        return new DTOAuthorizationSecurityExpressionHandler(evaluator);
    }
```

Also, the annotation itself deserve to be in another class than the SecurityConfiguration, which ATM is dedicated to authentication and endpoint security configuration, that is external interfaces

Tcharl:
@xetys good for you?

avdev4j:
I guess we can merge this PR ?

mraible:
FWIW, I removed the `MethodSecurityConfiguration` class in #11117.

mraible:
Is this PR still needed?

Tcharl:
Yes, as it removes some maven dependencies.
However, I've got to update it...

DanielFran:
@Tcharl can you fix conflicts?

Tcharl:
@DanielFran conflicts fixed, please merge whenever you want :-)

pascalgrimaud:
Excepting removing the dependencies, is the rename needed ? As it will introduce a minor breaking change by renaming files. In this case, there is no change in `cleanupOldServerFiles` ?

Tcharl:
As it is no more an extractor that renaming looks to make a sense.
Wasn't aware about this `cleanupOldServerFiles` capability, will take a look ASAP (this afternoon or tonight max)

kaidohallik:
When I tried to rename files then was requested to add Git rename support. I opened PR #10521 with rename functionality. But as those file renaming were rejected then this core rename functionality is missing ATM. If this PR will rename file, maybe it's then right time to add this core rename functionality?

mshima:
> When I tried to rename files then was requested to add Git rename support. I opened PR #10521 with rename functionality. But as those file renaming were rejected then this core rename functionality is missing ATM. If this PR will rename file, maybe it's then right time to add this core rename functionality?

Rename is already there.
But since it do git operation, I don’t thing we should use it outside upgrade generator.
We should create a structure that deletes on normal generator and rename at upgrade.
https://github.com/jhipster/generator-jhipster/blob/9c4276edb5906a743ee236996d3bc2f5808ba56f/generators/generator-base-private.js#L458

Added at:
https://github.com/jhipster/generator-jhipster/commit/9717f91aaaa2d2472697a0d29d53412b31e0ee5e

Tcharl:
Do you have something else to do? Otherwise someone can merge?

DanielFran:
@pascalgrimaud This PR is ready?

pascalgrimaud:
yes I think we can merge, although the renaming will introduce a minor breaking change but it is easy to fix it manually

generators/cleanup.js
```diff
@@ -221,6 +221,7 @@ function cleanupOldServerFiles(generator, javaDir, testDir, mainResourceDir, tes
         generator.removeFile(`${testDir}web/rest/ClientForwardControllerIT.java`);
     }
     if (generator.isJhipsterVersionLessThan('6.6.1')) {
+        generator.removeFile(`${javaDir}security/oauth2/JwtAuthorityExtractor.java`);
```
mshima: Should be after:
if (generator.isJhipsterVersionLessThan('6.8.0')) {

mshima:
@DanielFran I just found a minor problem, that must be fixed before or after merging.

generators/cleanup.js
DanielFran: @Tcharl Can you make this minor change?

The file should be removed only on version 6.8.0.

pascalgrimaud: I fixed it. So it can be merged if the CI pass

Issue #10963, opened 2019-12-16
Remove process.exit uses.


mshima:
##### **Overview of the feature request**

Remove process.exit uses and handle process shutdown cleanly.

##### **Motivation for or Use Case**

Node.js process.exit is not recommended because it has side effects https://nodejs.org/api/process.html#process_process_exit_code.

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/issues/10934
https://github.com/jhipster/generator-jhipster/pull/10956
https://github.com/jhipster/generator-jhipster/pull/10944
https://github.com/jhipster/generator-jhipster/pull/10961

-   [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
@pascalgrimaud I think we should go ahead and merge both PRs.

pascalgrimaud:
OK @mshima : I'm waiting the rebase for the green CI :heavy_check_mark:

pascalgrimaud:
Closing this as both PRs have been merged

PR #11077, opened 2020-01-03
Change cli logs from fatal to error.

mshima:

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -44,7 +44,7 @@ const runYoCommand = (cmd, args, options, opts) => {
     try {
         env.run(command, options, done);
     } catch (e) {
-        logger.fatal(e.message, e);
+        logger.error(e.message, e);
     }
 };
 
--- a/cli/import-jdl.js
+++ b/cli/import-jdl.js
@@ -86,7 +86,8 @@ function importJDL(jdlImporter) {
             const errorMessage = error.message || '';
             logger.log(chalk.red(`${errorName} ${errorMessage}`));
         }
-        logger.fatal(`Error while parsing applications and entities from the JDL ${error}`, error);
+        logger.error(`Error while parsing applications and entities from the JDL ${error}`, error);
+        throw error;
     }
     return importState;
 }
@@ -334,7 +335,8 @@ class JDLProcessor {
                 );
                 previousApp = getBaseName(application);
             } catch (error) {
-                logger.fatal(`Error while generating applications from the parsed JDL\n${error}`, error);
+                logger.error(`Error while generating applications from the parsed JDL\n${error}`, error);
+                throw error;
             }
         };
         this.importState.exportedApplications.forEach(application => {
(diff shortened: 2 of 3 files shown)


avdev4j:
hi @mshima 
I would like to merge this PR but I have one question, why sometimes we have to throw the error with an uncaught exception e.g. `throw error;` and why sometimes we do not ?

mshima:
> hi @mshima
> I would like to merge this PR but I have one question, why sometimes we have to throw the error with an uncaught exception e.g. `throw error;` and why sometimes we do not ?

If it is inside a loop, then throw to break the loop. Sometimes this isn’t needed, the process can terminate normally.

mshima:
We could catch outside the loop. But the error description must be right.
I am out this week.

avdev4j:
As I can see we do some unnecessary catch sometimes (not totally sure).
e.g. `jdlImporter.generateEntities(env, forkProcess);` is catch and the error is logged 
```
    try {
        const jdlImporter = new JDLProcessor(jdlFiles, options.inline, options);
        jdlImporter.getConfig();
        jdlImporter.importJDL();
        jdlImporter.sendInsight();
        jdlImporter.generateApplications(forkProcess);
        jdlImporter.generateEntities(env, forkProcess);
        jdlImporter.generateDeployments(forkProcess);
    } catch (e) {
        logger.error(`Error during import-jdl: ${e.message}`, e);
    }
```

but the method itself catch and log the error :
```
        } catch (error) {
            logger.error(`Error while generating entities from the parsed JDL\n${error}`, error);
            throw error;
        }
``` 

So the error is catch and log twice. Is it intended ?

mshima:
We could get some improvements in this area.
This PR is related to https://github.com/jhipster/generator-jhipster/issues/10963

avdev4j:
Ok, We should rework the way we manage the logging part, but not in this PR ;).

One thing I want to be sure before merging that:
- There is still a fatal logging in the  [import-jdl.js](https://github.com/jhipster/generator-jhipster/blob/aaea725f6cbc55e40c74d384a527ff2ca2bbaffe/cli/import-jdl.js#L452), is it intended ?

mshima:
Yes, but it can be changed to error and then a test
```
if (process.exitCode !=0){
  return;
}
```
after: 
https://github.com/jhipster/generator-jhipster/blob/aaea725f6cbc55e40c74d384a527ff2ca2bbaffe/cli/import-jdl.js#L471

avdev4j:
Ok, so as far as we should avoid `process.exit(1);` should we do remove this one ? 
I'm not expert of this part and how to manage node error code, that's why I'm asking you that. If you think it's ok, I'll merge it however let's remove the fatal logging and finally merge.

mshima:
I think we can keep fatal there. It’s cleaner. And the use case is very simple.
Otherwise the developer will need to know log.error sets exitCode as 1 to understand the conditional.

avdev4j:
Great, I see !

Thanks you very much @mshima

Issue #7608, opened 2018-05-11
TODO : Reactive/Webflux support


cbornet:
I have created a branch `reactive` to support reactive/webflux applications.

This issue is a TODO list for what remains:

- [x] Reactive DomainUserDetailsService (authentication) (@cbornet)
- [x] Account resource (@cbornet)
- [x] Security Audits (@cbornet #10405)
- [x] Serve front-end / static content (@cbornet)
- [x] Configure Locale resolver  (@cbornet)
- [x] Configure CORS  (@cbornet)
- [x] Exception handling (Problem) (https://github.com/zalando/problem-spring-web/pull/182)(@cbornet)
- [x] Http session / Cookie auth (#9326)  (@cbornet)
- [x] OAuth2 (@mraible  #11117)
- [x] SQL with R2DBC + reactor-pool (@cbornet #11439)
- [x] Cassandra (@cbornet in #9373)
- [x] Disable caching (@cbornet #10107)
- [x] Couchbase (@tchlyah in https://github.com/jhipster/generator-jhipster/pull/8997)
- [x] Skip user management on monolith and no database (#11014)(@cbornet)
- [x] API first (#8532) (@cbornet)
- [x] Server resources tests with WebTestClient
  - [x] User  (@cbornet)
  - [x] Account (@cbornet)
  - [x] Logs  (@cbornet)
- [x] Server resources
  - [x] User  (@cbornet)
  - [x] Logs  (@cbornet)
- [x] With service registry (Eureka)  (@cbornet)
- [x] Micro-service app  (@cbornet)
- [x] Metrics reporting with micrometer  (@cbornet)
- [x] Tune logback (#10462)(@cbornet)
- [x] Kafka option (#10935)(@cbornet)
- [x] Add BlockHound and Hooks.onOperatorDebug() during tests (#11009)(@cbornet)
- [x] Entity generator (#11416)(@murdos)

Blocked:
* Session Remember-me. (see https://github.com/spring-projects/spring-security/issues/5504)
* WebSockets option: no support for [security](https://github.com/spring-projects/spring-security/issues/5188) nor [STOMP](https://github.com/spring-projects/spring-framework/issues/19101)
* Springfox /swagger-ui (supported by future v3 of Springfox : https://github.com/springfox/springfox/issues/1773)
* Auditing : https://github.com/spring-projects/spring-security/issues/4961 and https://jira.spring.io/browse/DATACMNS-1231
[...]

PR #9326, opened 2019-02-22
Support for session auth in reactive apps

cbornet:
-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

Note : no support for remember-me (the checkbox is still present in the front but has no effect)

PR #11117, opened 2020-01-14
Add OAuth 2.0 / OIDC Auth for Reactive

mraible:
Adds OAuth2 support for reactive apps. Related to https://github.com/jhipster/generator-jhipster/issues/7608.

-   [x] Add continuous jobs for reactive monolith with e2e tests

Working example: https://github.com/mraible/jhipster-reactive-monolith-oauth2/pull/1

I'm guessing we need to add some tests for a reactive monolith with Protractor. Currently, if you generate a reactive monolith and run `yarn e2e`, tests fail.


test-integration/samples/webflux-mongodb/.yo-rc.json
```diff
@@ -17,7 +17,7 @@
         "enableTranslation": true,
         "nativeLanguage": "en",
         "languages": ["en", "fr"],
-        "testFrameworks": ["gatling"],
+        "testFrameworks": ["gatling","protractor"],
```
qmonmert: ["gatling","protractor"] => ["gatling", "protractor"] will be better :)

mraible:
@jhipster/developers Does anyone know where the `/page-does-not-exist` to client-side `/404` logic resides? The e2e test is failing for it when using webflux.

ruddell:
The backend logic that forwards 404s (that don't match an API endpoint) to the client is in [ClientForwardController.java](https://github.com/jhipster/generator-jhipster/blob/bbe475987898ef4a313318c2c71362f30b40f172/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs), it was removed from `reactive` apps in https://github.com/jhipster/generator-jhipster/pull/9326 . This also means that refreshing on a path like http://localhost:8080/account/settings will result in a 404.  Maybe there is a reactive way to do this?

mraible:
I was able to add a `SpaWebFilter` and now `/index.html` doesn't flash as it did before. 

```java
@Component
public class SpaWebFilter implements WebFilter {

    @Override
    public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
        String path = exchange.getRequest().getURI().getPath();
        if (!path.startsWith("/api") && path.matches("[^\\\\.]*")) {
            return chain.filter(
                exchange.mutate().request(exchange.getRequest().mutate().path("/index.html").build()
                ).build());
        }
        return chain.filter(exchange);
    }
}
```

Everything seems to work on the client, but I get a 401 for any routes that should be 404s.

mraible:
I was able to fix the security issue by adding the following to the `SecurityConfiguration`:

```java
.addFilterAt(new SpaWebFilter(), SecurityWebFiltersOrder.AUTHENTICATION)
```

mraible:
CI is failing after I merged from master:

```
[INFO] ERROR in /Users/mraible/webflux-mongodb/src/main/webapp/app/modules/login/login.tsx(26,11):
[INFO] TS2339: Property 'from' does not exist on type '{} | { from: { pathname: string; search: string; }; }'.
```

Line is:

```js
const { from } = location.state || { from: { pathname: '/', search: location.search } };
```

It doesn't look like this file has changed in a while. Maybe compiler settings changed or a React dependency?

mraible:
The `react-gradle-cassandra-session-redis` check is failing because of an `OperationTimedOutException`. This [happens on master too](https://github.com/jhipster/generator-jhipster/runs/389349931) and doesn't seem related to any changes I made. Maybe Cassandra takes too long to startup in Docker?

I'm not sure why JWT and session auth e2e tests are failing for WebFlux. The 404 test passes successfully, then the user can't log in after that. Any ideas @cbornet?

DanielFran:
@mraible It seems a dependency from react, actually all react tests are failing from Master

generators/server/templates/src/main/java/package/config/ReactiveSecurityConfiguration.java.ejs
```diff
@@ -201,14 +266,86 @@ public class SecurityConfiguration {
...
+            .pathMatchers("/api/auth-info").permitAll()
             .pathMatchers("/api/**").authenticated()
             .pathMatchers("/management/health").permitAll()
             .pathMatchers("/management/info").permitAll()
             .pathMatchers("/management/prometheus").permitAll()
             .pathMatchers("/management/**").hasAuthority(AuthoritiesConstants.ADMIN);
+        <%_ if (authenticationType === 'oauth2') { _%>
+
+        http<%_ if (['monolith', 'gateway'].includes(applicationType)) { _%>.oauth2Login()
+            .authenticationSuccessHandler(this::onAuthenticationSuccess)
```
cbornet: @mraible will you also do the authenticationErrorHandler ?

mraible: There's no reason to do one for OAuth 2.0 because the authentication will fail at the IdP and you won't be redirected back to the app.

mraible:
@DanielFran I'm not sure if it's worth it because the [fix was just adding `as any`](https://github.com/jhipster/generator-jhipster/pull/11117/commits/50973e295494d318eea8cbff36cb21f1871ff63e).

@cbornet All tests pass! Are you OK with me merging to master? If so, should we change Reactive from Alpha to Beta?

generators/server/templates/src/main/java/package/config/ReactiveSecurityConfiguration.java.ejs
cbornet: So we can't have the security audits for failed auth ?

cbornet:
LGTM.
We still need to do the SQL/R2DBC option (work started [here](https://github.com/cbornet/jhipster-r2dbc)) and the entity gen. Do you think that's mandatory for a beta ?

generators/server/templates/src/main/java/package/config/ReactiveSecurityConfiguration.java.ejs
mraible: No, I don't think so. At least not in JHipster. They should be available in the IdP though.

mraible:
@cbornet Good point. I'll hold off on changing to Beta. We could always cheat for the entity generator. If someone tries to use it with WebFlux, just print a message saying that you shouldn't build CRUD apps with WebFlux. ;)

Another thing that might be nice would be too prompt for reactive rather than having it as a separate option. It'd show up when the user chose monolith, microservice, or gateway.

PR #11125, opened 2020-01-15
Proposal: Remove explicit definition of database platform in Spring configuration

murdos:
See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes : 
>  Spring Boot now lets Hibernate chose the dialect to use rather than applying a default dialect based on the detected database. If you are interested by this feature, please make sure you don’t specify the database platform explicitly (i.e. via spring.jpa.database or spring.jpa.database-platform).

> If you had a dialect configured previously, you may want to remove your customization.

I only kept the dialects that have been overridden in jhipster framework.

### DIFF
--- a/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
@@ -143,29 +143,10 @@ spring:
             enabled: false
     <%_ } _%>
     jpa:
-        <%_ if (devDatabaseType === 'mysql') { _%>
-        database-platform: org.hibernate.dialect.MySQL8Dialect
-        <%_ } else if (devDatabaseType === 'mariadb') { _%>
-        database-platform: org.hibernate.dialect.MariaDB103Dialect
-        <%_ } else if (devDatabaseType === 'postgresql') { _%>
+        <%_ if (devDatabaseType === 'postgresql') { _%>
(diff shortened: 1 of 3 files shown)


generators/server/templates/src/main/resources/config/application-dev.yml.ejs
```diff
@@ -143,30 +143,11 @@ spring:
...
-        <%_ if (devDatabaseType === 'mysql' || devDatabaseType === 'mariadb') { _%>
-        database: MYSQL
-        <%_ } else if (devDatabaseType === 'postgresql') { _%>
-        database: POSTGRESQL
-        <%_ } else if (devDatabaseType === 'oracle') { _%>
-        database: ORACLE
-        <%_ } else if (devDatabaseType === 'mssql') { _%>
-        database: SQL_SERVER
-        <%_ } else { _%>
-        database: H2
```
hdurix: Could this be also removed from `test/application.yml`?

DanielFran: Not sure it is possible, because we might have a different database in dev and prod

murdos: @hdurix : good catch. I'll remove the `spring.jpa.database` entry too

DanielFran: Forget, a esse thinking about main/application.yml 🤪😅

PR #11134, opened 2020-01-17
Fix app loading other generators not from blueprint

cmyker:
So after this PR #10752 merge possibility to override app generator was introduced. But if one wants to override in addition to it, for example, server - it is not possible, since all calls similar to _this.composeWith(require.resolve('../server'_, are searching in directory relative to original generator index.js, not blueprint's. This PR solves that issue.

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -19,6 +19,7 @@
 /* eslint-disable consistent-return */
 const chalk = require('chalk');
 const _ = require('lodash');
+const path = require('path');
 const BaseBlueprintGenerator = require('../generator-base-blueprint');
 const cleanup = require('../cleanup');
 const prompts = require('./prompts');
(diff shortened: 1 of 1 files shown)


mshima:
I don’t know what problem this patch fixes.
Actually I think this patch will break blueprints that only override app blueprint and don’t override others generators.

cmyker:
Not true, in require.resolve two paths are passed, it will search in blueprint path and if not found in jhipster generator directory

mshima:
Blueprints are loaded at this point:
https://github.com/jhipster/generator-jhipster/blob/12c94a056b9cfac42c4de47b420d7ae08607a743/generators/server/index.js#L63

The bug needs to be confirmed.

cmyker:
How can I assist more?

mshima:
@cmyker the problem here is 'app blueprint fails to load others original generators' or 'app blueprint fails to load others blueprinted generators'?
How to reproduce this bug?
I suggest creating a bug report.

mshima:
Just confirmed this bug, but current PR will conflict with https://github.com/jhipster/generator-jhipster/issues/11142.
Closing in favor of https://github.com/jhipster/generator-jhipster/pull/11153

Thanks @cmyker

mshima:
Fill free to reopen if you think https://github.com/jhipster/generator-jhipster/pull/11153 doesn't fix your issue.

PR #11145, opened 2020-01-19
Prompt for reactive option after app type question

mraible:
As [discussed on the developer mailing list](https://groups.google.com/forum/#!topic/jhipster-dev/Cc20liWwPa8).

Note that this also changes `JHipster UAA server (for microservice OAuth2 authentication)` to `JHipster UAA server`. Are you OK with this change @xetys?


mraible:
@deepu105 If it's not nested, how would you suggest I implement it?

mraible:
OK. Thanks for the explanation @deepu105. I'll merge as-is for now. We can always refactor later.

PR #11164, opened 2020-01-23
use Mono.error instead of throwing in flatmap

atomfrede:
While working on the neo support and trying reactive option idea and sonar pointed out some issues. This fixes 
* Throwing instead of returning a mono in flatmap
*  (sonar) using primitive boolean if statement with boolean objects (for real there won't be a npe but aiming for a A grade in sonar)

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
@@ -194,15 +194,15 @@ public class UserResource {
         return userRepository.findOneByLogin(userDTO.getLogin().toLowerCase())
             .hasElement()
             .flatMap(loginExists -> {
-                if (loginExists) {
-                    throw new LoginAlreadyUsedException();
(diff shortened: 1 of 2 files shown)


generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -251,16 +251,16 @@ public class UserResource {
...
+                    return Mono.error(new EmailAlreadyUsedException());
                 }
                 return userRepository.findOneByLogin(userDTO.getLogin().toLowerCase());
             })
             .filter(user -> !user.getId().equals(userDTO.getId()))
             .hasElement()
             .flatMap(loginExists -> {
-                if (loginExists) {
-                    throw new LoginAlreadyUsedException();
+                if (Boolean.TRUE.loginExists) {
```
cbornet: .equals ?

atomfrede: :man_facepalming: Yes of course

PR #9981, opened 2019-06-24
[React] Hooks (user-management component)

qmonmert:
WIP on #9891

-   [x] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #11171, opened 2020-01-24
fix: sort on users

ctamisier:
Fix sorting users on React.

The current sort wasn't "sync" to the parameters in the URL.
When the url has 'asc', then 'desc' sorting was applied (and the opposite).

The sort is working for entities so I applied the same logic for users (just move a line).

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx.ejs
@@ -39,27 +39,17 @@ export interface IUserManagementProps extends StateProps, DispatchProps, RouteCo
 export const UserManagement = (props: IUserManagementProps) => {
   const [pagination, setPagination] = useState(getSortState(props.location, ITEMS_PER_PAGE));
 
-  const getAllUsers = () =>
-    props.getUsers(pagination.activePage - 1, pagination.itemsPerPage, `${pagination.sort},${pagination.order}`);
-
   useEffect(() => {
-    getAllUsers();
-  }, []);
+    props.getUsers(pagination.activePage - 1, pagination.itemsPerPage, `${pagination.sort},${pagination.order}`)
+    props.history.push(`${props.location.pathname}?page=${pagination.activePage}&sort=${pagination.sort},${pagination.order}`);
+  }, [pagination]);
 
-  const sortUsers = () => getAllUsers();
-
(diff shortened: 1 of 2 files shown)


MathieuAA:
There's no issue in Angular?

ctamisier:
No issue with Angular !

ctamisier:
I found out that this commit b0d7a85ffaee1d00af77f2299c74e52bf37b69f7 is the opposite of what I did in this PR. So I might have break something but I don't know what.

mshima:
Looks like it was to fix CI failing:
https://github.com/jhipster/generator-jhipster/pull/9981
https://github.com/jhipster/generator-jhipster/issues/9891

generators/client/templates/react/src/main/webapp/app/modules/administration/user-management/user-management.tsx.ejs
```diff
@@ -39,27 +39,17 @@ export interface IUserManagementProps extends StateProps, DispatchProps, RouteCo
...
-    getAllUsers();
-  }, []);
-
-  const sortUsers = () => getAllUsers();
-
   useEffect(() => {
-    sortUsers();
-  }, [pagination.activePage, pagination.order, pagination.sort]);
+    props.getUsers(pagination.activePage - 1, pagination.itemsPerPage, `${pagination.sort},${pagination.order}`)
+    props.history.replace(`${props.location.pathname}?page=${pagination.activePage}&sort=${pagination.sort},${pagination.order}`);
```
vishal423: any specific reason to `replace` history?

ctamisier: Thanks for spoting this! 
Indeed, to be coherent with Angular version it should be a **history.push**.
I will change this.

However with a history.push, when we _go back_ with the browser, the previous state sort is not applied (but the url is updated).
By using **history.replace** we will go back to the previous page no matter all the sorts that we could have performed on the page.

So I will try to see how we can fix this (in Angular, React and maybe Vue.js) in another PR.

vishal423:
@ctamisier, can you check react e2e failures. They seems to be failing after your last change (even after rerun). Let me know if they are not reproducible locally or are not related to your changes.

ctamisier:
@vishal423 , the issues were in the code, they are fixed now.

vishal423:
thanks @ctamisier

PR #11185, opened 2020-01-27
fix k8s mount-path config for psql, fix #11184

### LINKED ISSUE #11184: k8s: PostgreSQL mount path misconfigured

The mount path of the persistent storage in our postgresql manifests, generated by our kubernetes sub-generator, is misconfigured. Postgres does its own ephemeral mount on `/var/lib/postgresql/data`, so the database only persists until the pod is deleted 

##### **Reproduce the error**

Given you have a JHipster application using PostgreSQL:

1. generate k8s manifests using `jhipster kubernetes` (and choose dynamic storage provisioning)
2. deploy your application and generate some data
3. delete the PostgreSQL pod and wait until it is back up
4. examine the application for the data. All data should be lost

Reason:

PostgreSQL uses its own mount which overrides the existing, so if you `kubectl exec -it` into the running pod you can list the mounts to see the configuration:

``` bash
root@psql: mount
/dev/rbd2 on /var/lib/postgresql type ext4 (rw,relatime,stripe=1024,data=ordered)
/dev/sda1 on /var/lib/postgresql/data type ext4 (rw,relatime,errors=remount-ro,data=ordered)
```

(in my case I am using ceph-rbd as external storage)

##### **Suggest a Fix**

We should add a subMount path to the configuration, like this:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-postgresql
spec:
  replicas: 1
  selector:
    matchLabels:
      app: app-postgresql
  template:
    metadata:
      labels:
        app: app-postgresql
    spec:
      volumes:
        - name: data
          persistentVolumeClaim:
            claimName: app-postgresql-pvc
      containers:
        - name: postgres
          image: postgres:11.6
          env:
            - name: POSTGRES_USER
              value: App
            - name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: app-postgresql
                  key: postgresql-password
          ports:
            - containerPort: 5432
          volumeMounts:
            - name: data
              mountPath: /var/lib/postgresql/data
              subPath: postgres
          resources:
            requests:
              memory: '512Mi'
              cpu: '500m'
            limits:
              memory: '1Gi'
              cpu: '1'
```

Then, the mount table is showing only one mount, and the data persists after pod restarts...

##### **JHipster Version(s)**
6.6.0

xetys:
fix #11184

### DIFF
--- a/generators/kubernetes/templates/db/postgresql.yml.ejs
+++ b/generators/kubernetes/templates/db/postgresql.yml.ejs
@@ -86,7 +86,8 @@ spec:
         - containerPort: 5432
         volumeMounts:
         - name: data
-          mountPath: /var/lib/postgresql/
+          mountPath: /var/lib/postgresql/data
+          subMount: postgres
         resources:
           requests:
             memory: "512Mi"


generators/kubernetes/templates/db/postgresql.yml.ejs
```diff
@@ -86,7 +86,8 @@ spec:
         - containerPort: 5432
         volumeMounts:
         - name: data
-          mountPath: /var/lib/postgresql/
+          mountPath: /var/lib/postgresql/data
+          subMount: postgres
```
saturnism: should this be `subPath`?

xetys: oh yeah, you're right...that was a brain error xD

I'm fixing this asap

PR #11187, opened 2020-01-28
Update yeoman-generator to 4.5.0

mshima:

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
         "through2": "3.0.1",
         "uuid": "3.3.3",
         "yeoman-environment": "2.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
         "through2": "3.0.1",
         "uuid": "3.3.3",
         "yeoman-environment": "2.7.0",
-        "yeoman-generator": "4.4.0",
+        "yeoman-generator": "^4.5.0",
```
atomfrede: I think we should use fixed versions.

mshima: You are right.

PR #11205, opened 2020-01-29
Fix needles with multiple lines in Windows if Git autocrlf is used

kaidohallik:
If in Windows is used Git `autocrlf` and project is regenerated after committing to Git then those needles which content has more than 1 line are added multiple times to templates.

This PR fixes that.

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -118,7 +118,8 @@ function escapeRegExp(str) {
  */
 function rewrite(args) {
     // check if splicable is already in the body text
-    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line)}`).join('\n'));
+    // replacing \r\n with \n as this replacement is done also in haystack to resolve Windows Git autocrlf issue
+    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line.replace(/\r\n/g, '\n'))}`).join('\n'));
 
     // if in Windows is Git autocrlf used then need to replace \r\n with \n to avoid multiple additions
     if (re.test(args.haystack.replace(/\r\n/g, '\n'))) {
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -26,6 +26,7 @@ const fs = require('fs');
 const crypto = require('crypto');
 const randexp = require('randexp');
 const faker = require('faker');
+const os = require('os');
 
 const constants = require('./generator-constants');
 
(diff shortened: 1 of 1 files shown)


mshima:
If it's affecting only windows we should make this windows only.

generators/utils.js
```diff
@@ -118,9 +118,11 @@ function escapeRegExp(str) {
  */
 function rewrite(args) {
     // check if splicable is already in the body text
-    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line)}`).join('\n'));
+    // replacing \r\n with \n as this replacement is done also in haystack to resolve Windows Git autocrlf issue
+    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line.replace(/\r\n/g, '\n'))}`).join('\n'));
```
mshima:
```suggestion
    const isWin32 = os.platform() === 'win32';
    line = isWin32 ? line.replace(/\r\n/g, '\n') : line;
    const re = new RegExp(args.splicable.map(line => `\\s*${escapeRegExp(line)}`).join('\n'));
```

generators/utils.js
```diff
@@ -118,9 +118,11 @@ function escapeRegExp(str) {
...
 
-    if (re.test(args.haystack)) {
+    // if in Windows is Git autocrlf used then need to replace \r\n with \n to avoid multiple additions
+    if (re.test(args.haystack.replace(/\r\n/g, '\n'))) {
```
mshima:
```suggestion
    const haystack = isWin32 ? args.haystack.replace(/\r\n/g, '\n') : args.haystack;
    if (re.test(haystack)) {
```

kaidohallik:
@mshima thanks for the review! I'll make a new commit to change `CRLF` with `LF` only in Windows, I'll add a new function for that.

kaidohallik:
@mshima I made suggested changes.

DanielFran:
Trusting @mshima. Thanks @kaidohallik

mshima:
@kaidohallik have you tried .join('\r?\n') instead of normalize?
https://github.com/jhipster/generator-jhipster/issues/11231

kaidohallik:
I don't believe this affects #11231 in any way. #11231 is unrelated to this PR.

mshima:
I think jhipster is rewriting the file without crlf and as consequence, git cannot compare it.

kaidohallik:
This PR didn't change content which is written. This PR only changed comparison input.

But generally yes, JHipster writes content with LF and Git operations are converting this to CRLF. And sometimes I have seen that Git commit preview thinks file is changed if line endings are different, but ATM most of the time in my machine Git doesn't show those files as changed.

mshima:
@kaidohallik you are right, it’s pretty safe.
I've made the relation because it was just merged.

PR #9341, opened 2019-02-26
Add support for Couchbase relations

tchlyah:
Fix #9342 

This PR needs https://github.com/jhipster/jhipster-core/pull/309, and also [spring-data-couchbase:2.1.6](https://jira.spring.io/issues/?jql=project+%3D+DATACOUCH+AND+fixVersion+%3D+%223.1.6+%28Lovelace+SR6%29%22) release, since it fixes a bug with N1qlJoin (https://github.com/spring-projects/spring-data-couchbase/pull/177).

Depends on https://github.com/jhipster/generator-jhipster/pull/9526

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

Issue #7944, opened 2018-07-09
Support for rich document models for mongodb (embedding and referencing)


ivangsa:
Hi, 

I would like to add support for rich documents a relations for mongodb to jhipster

I think it would be easy to add support for both embedded/nested objects and collections as well as references between entities

I've been doing some work but I'm quite new to js development and I'm stuck with debugging some errors and finding my way with mocha tests..

https://github.com/ivangsa/generator-jhipster/issues/1

Let me know what you think about this...

Regards
Ivan

------

I would like to implement support for entity relations with mongodb, to generate microserices from jdl to cover two use cases:

1.- Support complex document models: embedded and nested objects and collections.
2.- Support documents with relations using @DBRef

I think the JDL would need extending to acommodote both embedding and referencing... Allow adding fields with an entity as fieldType, very much as the Enums are defined now, using jpa @Embeddable/@Embedded as field configuration for sql backend, benefiting sql backend as well as mongodb.
    entity Person {
        name String,
        address Address
    }
    entity Address { ... }
    
(There is a small problem when using @Embeddable/@Embedded is not posible mark the embedded Id as transient, but small tweaks on generated code can acomodate the mappings for exact needs..)

This is how I'm planning to interpret the JDL as it is right now plus the 'embedded field' extension:

#### Embedded (for embedding entities)
- Unidirectional: @Field (@Embeddable/@Embedded for sql)
    entity Person {
        name String,
        address Address
    }
    entity Address { ... }

#### One-to-One (for entities references)

- Bidirectional: @DBRef for both left and right side fields. @JsonBackReference on right side.
    relationship OneToOne {
      Car{driver} to Driver{car}
    }
- Unidirectional: @DBRef on left side
    relationship OneToOne {
      Citizen{passport} to Passport
    }
    

#### One-to-Many 
(Owner has none, one or more Cars) For embedding collections of entities with backreference.
- Bidirectional: @Field for left side, @DBRef for right side. @JsonBackReference on the non-collection (right) side.
    relationship OneToMany {
      Owner{car} to Car{owner}
    }
- Unidirectional: (not supported)

#### Many-to-One
- Bidirectional: For embedding collections of references: @DBRef on both sides. @JsonBackReference on the non-collection side.
    relationship ManyToOne {
      Car{owner} to Owner{car}
    }
- Unidirectional (where the car know its owner): @DBRef on the left side field.
    relationship ManyToOne {
      Car{owner} to Owner
    }

[...]

Issue #11122, opened 2020-01-14
DDD approach in JHipster


hdurix:
## DDD approach

For now, JHipster is only able to create CRUD entities and JHipster's goal is to help developers in building all the technical aspect of an application.
It could be interesting to explore how JHipster could also help them build the "business" of the application.
One approach for that is the methodology called [Domain Driven Design](https://en.wikipedia.org/wiki/Domain-driven_design).

### Why this ticket?

This ticket is to:
- record what is being done to explore a DDD approach in JHipster
- gather ideas from the community for turning a JHipster application in a more DDD-compliant one

### First step

I would like to work on a application generated by JHipster and to refactor it in order to bring DDD concepts (aggregates, entities, value objects, domain events, ...) and have example to show and to discuss about.

### More to come...

If some of you have experience with bringing DDD concepts to a JHipster application, feel free to participate, to give ideas and feedback.

pascalgrimaud:
For me, it is a must have. It can be helped by repositories as example, by documentation, by modifying the current generated code etc...
There is a lot of work I think, so I'm putting a big bounty on this. You'll be the lead on this part @hdurix, don't hesitate to ask help

pascalgrimaud:
cc @nonomoho as you probably have ideas on this

johnbanq:
I think the most important part of DDD lies in the strategic patterns(bounded contexts and context mapping) instead of the tatics patterns(repositories, aggregates etc). As long as you have a clear separation of contexts, a good idea of their relationships and a conceptual model within each context, you are using DDD even you are just CRUDing all the way through in each these contexts.

So perhaps we should start by telling people how to organize their code by contexts and apply “clean architecture“ within some key context which have a relatively complex domain model, instead of going for the tactics patterns right away?

huytmb:
I completely agree with @johnbanq advice. In early phase, starting with "clean architecture" for complex domain model.

gmarziou:
Do you have a specific flavor of "clean architecture" in mind? Hexagonal, onion, ...
Personally I like the modulith approach: package by feature/context and using spring events for inter module communication

mohamedzaki90:
@gmarziou spring events have the problem of lost events if an exception happens during event publication, listeners will never get notified

gmarziou:
Well, I never experienced this in my apps, do you have some reference to share? 

My understanding is that if an exception happens during event publication fails, the publisher will get it and its transaction will rollback. So I can't imagine how you could lose an event.

[...]

PR #10561, opened 2019-10-04
Readonly entity: Angular client

murdos:
See #8800

-   [ ] [Travis tests](https://travis-ci.org/jhipster/generator-jhipster/pull_requests) are green
-   [X] Tests are added where necessary
-   [X] Documentation is added/updated where necessary
-   [X] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #11239, opened 2020-02-05
Embedded entities for document databases (Couchbase, MongoDB)

tchlyah:
Hello guys,

Following jhipster/generator-jhipster#7944 and jhipster/generator-jhipster#9341 for supporting rich document models on document databases (MongoDB and Couchbase), I added a new embedded option that enables subdocuments as explained here : jhipster/generator-jhipster#7944 (comment)

This PR needs https://github.com/jhipster/jhipster-core/pull/429

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -170,6 +170,7 @@ const angularFiles = {
 const reactFiles = {
     client: [
         {
+            condition: generator => !generator.embedded,
             path: REACT_DIR,
             templates: [
                 {
@@ -187,10 +188,6 @@ const reactFiles = {
                     renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.reducer.ts`
                 },
                 {
-                    file: 'entities/entity.model.ts',
-                    renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
-                },
-                {
                     file: 'entities/index.tsx',
                     method: 'processJsx',
                     renameTo: generator => `entities/${generator.entityFolderName}/index.tsx`
@@ -198,7 +195,16 @@ const reactFiles = {
             ]
         },
         {
-            condition: generator => !generator.readOnly,
+            path: REACT_DIR,
+            templates: [
+                {
+                    file: 'entities/entity.model.ts',
+                    renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
+                }
+            ]
+        },
+        {
+            condition: generator => !generator.readOnly && !generator.embedded,
             path: REACT_DIR,
             templates: [
                 {
@@ -216,6 +222,7 @@ const reactFiles = {
     ],
     test: [
         {
+            condition: generator => !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
@@ -225,7 +232,7 @@ const reactFiles = {
             ]
         },
         {
-            condition: generator => generator.protractorTests,
+            condition: generator => generator.protractorTests && !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
@@ -239,7 +246,7 @@ const reactFiles = {
             ]
         },
         {
-            condition: generator => generator.protractorTests && !generator.readOnly,
+            condition: generator => generator.protractorTests && !generator.readOnly && !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -322,8 +322,10 @@ function writeFiles() {
                 files = reactFiles;
                 destDir = REACT_DIR;
                 templatesDir = CLIENT_REACT_TEMPLATES_DIR;
-            } else if (!this.embedded) {
-                this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
+            } else {
+                if (!this.embedded) {
+                    this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
+                }
                 return;
             }
 
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -44,7 +44,7 @@ const angularFiles = {
             templates: [
                 {
                     file: 'entities/entity.model.ts',
-                    // using entityModelFileName so that there is no conflict when genertaing microservice entities
+                    // using entityModelFileName so that there is no conflict when generating microservice entities
                     renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
                 }
             ]
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -43,6 +43,17 @@ const angularFiles = {
             path: ANGULAR_DIR,
             templates: [
                 {
+                    file: 'entities/entity.model.ts',
+                    // using entityModelFileName so that there is no conflict when genertaing microservice entities
+                    renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
+                }
+            ]
+        },
+        {
+            condition: generator => !generator.embedded,
+            path: ANGULAR_DIR,
+            templates: [
+                {
                     file: 'entities/entity-management.component.html',
                     method: 'processHtml',
                     template: true,
@@ -63,11 +74,6 @@ const angularFiles = {
                     renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.route.ts`
                 },
                 {
-                    file: 'entities/entity.model.ts',
-                    // using entityModelFileName so that there is no conflict when genertaing microservice entities
-                    renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
-                },
-                {
                     file: 'entities/entity-management.component.ts',
                     renameTo: generator => `entities/${generator.entityFolderName}/${generator.entityFileName}.component.ts`
                 },
@@ -82,7 +88,7 @@ const angularFiles = {
             ]
         },
         {
-            condition: generator => !generator.readOnly,
+            condition: generator => !generator.readOnly && !generator.embedded,
             path: ANGULAR_DIR,
             templates: [
                 {
@@ -110,6 +116,7 @@ const angularFiles = {
     ],
     test: [
         {
+            condition: generator => !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
@@ -128,7 +135,7 @@ const angularFiles = {
             ]
         },
         {
-            condition: generator => !generator.readOnly,
+            condition: generator => !generator.readOnly && !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
@@ -144,7 +151,7 @@ const angularFiles = {
             ]
         },
         {
-            condition: generator => generator.protractorTests,
+            condition: generator => generator.protractorTests && !generator.embedded,
             path: CLIENT_TEST_SRC_DIR,
             templates: [
                 {
@@ -308,7 +315,7 @@ function writeFiles() {
                 files = reactFiles;
                 destDir = REACT_DIR;
                 templatesDir = CLIENT_REACT_TEMPLATES_DIR;
-            } else {
+            } else if (!this.embedded) {
                 this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
                 return;
             }
@@ -317,17 +324,19 @@ function writeFiles() {
             this.writeFilesToDisk(files, this, false, this.fetchFromInstalledJHipster(entityTemplatesDir));
             addEnumerationFiles(this, templatesDir, destDir);
 
-            this.addEntityToModule(
-                this.entityInstance,
-                this.entityClass,
-                this.entityAngularName,
-                this.entityFolderName,
-                this.entityFileName,
-                this.entityUrl,
-                this.clientFramework,
-                microserviceName
-            );
-            this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
+            if (!this.embedded) {
+                this.addEntityToModule(
+                    this.entityInstance,
+                    this.entityClass,
+                    this.entityAngularName,
+                    this.entityFolderName,
+                    this.entityFileName,
+                    this.entityUrl,
+                    this.clientFramework,
+                    microserviceName
+                );
+                this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
+            }
         }
     };
 }
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
@@ -18,7 +18,7 @@
 -%>
 <%
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, embedded);
 const typeImports = generateEntityClientImports(relationships, dto);
 const defaultVariablesValues = generateEntityClientFieldDefaultValues(fields);
 const enumImports = generateEntityClientEnumImports(fields);
@@ -27,7 +27,7 @@ const enumImports = generateEntityClientEnumImports(fields);
 import { Moment } from 'moment';
 <%_ } _%>
 <%_ typeImports.forEach((importedPath, importedType) => { _%>
-    <%_ if (importedType !== `I${entityAngularName}`) { _%> 
+    <%_ if (importedType !== `I${entityAngularName}`) { _%>
 import { <%- importedType %> } from '<%- importedPath %>';
     <%_ } _%>
 <%_ }); _%>
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
@@ -18,7 +18,7 @@
 -%>
 <%
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, 'Moment', embedded);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, undefined, embedded);
 const typeImports = generateEntityClientImports(relationships, dto);
 const defaultVariablesValues = generateEntityClientFieldDefaultValues(fields);
 const enumImports = generateEntityClientEnumImports(fields);
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
@@ -18,7 +18,7 @@
 -%>
 <%
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, embedded);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, 'Moment', embedded);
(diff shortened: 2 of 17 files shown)


generators/entity-client/files.js
```diff
@@ -40,6 +40,17 @@ const CLIENT_REACT_TEMPLATES_DIR = 'react';
...
+            templates: [
+                {
+                    file: 'entities/entity.model.ts',
+                    // using entityModelFileName so that there is no conflict when genertaing microservice entities
+                    renameTo: generator => `shared/model/${generator.entityModelFileName}.model.ts`
+                }
+            ]
+        },
+        {
+            condition: generator => !generator.embedded,
```
MathieuAA: What exactly does this mean that a generator is embedded? I ask because I see the embedded attribute of the generator object and I don't understand.

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
```diff
@@ -283,15 +293,15 @@ public class <%= asEntity(entityClass) %> implements Serializable {
     @Cache(usage = CacheConcurrencyStrategy.NONSTRICT_READ_WRITE)
                 <%_ }
             }
-        } else if (databaseType === 'mongodb' || databaseType === 'couchbase') {
-            if (databaseType === 'mongodb') { _%>
+        } else if ((databaseType === 'mongodb' || databaseType === 'couchbase')) {
+            if (databaseType === 'mongodb' && !otherEntityEmbedded) { _%>
```
MathieuAA: Does `otherEntityEmbedded` mean that the other entity is embedded? In that case, I prefer `embeddedOtherEntity` or `otherEntityIsEmbedded` (but I prefer the first one for consistency)

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
```diff
@@ -283,15 +293,15 @@ public class <%= asEntity(entityClass) %> implements Serializable {
...
     @DBRef
         <%_ } _%>
-    @Field("<%= relationshipFieldName %>")
-        <%_ if (databaseType === 'couchbase') { _%>
+    @Field("<%= relationshipFieldNamePlural %>")
```
MathieuAA: Is it related to your PR?

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -54,7 +54,7 @@ import java.math.BigDecimal;
 <%_ if (fieldsContainBlob && databaseType === 'cassandra') { _%>
 import java.nio.ByteBuffer;
 <%_ } _%>
-<%_ if (fieldsContainOwnerManyToMany) { _%>
+<%_ if (fieldsContainOwnerManyToMany || fieldsContainEmbedded) { _%>
```
MathieuAA: What does `fieldsContainEmbedded` mean? That at least one of the fields is an embedded entity? Or all of them are?

MathieuAA:
Thank you for the PR @tchlyah!

generators/entity-client/files.js
tchlyah: I must admit that I don't understand it everything either, I followed readOnly PR (https://github.com/jhipster/generator-jhipster/pull/10561). It seems here that the generator contains current entity data (entityFolderName, entityFileName, readonly, ...).
So yes here embedded means that the current entity is embedded.

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
tchlyah: Yes, this is exactly what it means, I followed the same pattern for `otherEntity` variables (`otherEntityRelationshipName`, `otherEntityRelationshipNamePlural`, ...). I think it is explicit enough, but if you don't like it, I can rename it. In that case I prefer `otherEntityIsEmbedded`

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
tchlyah: Actually this is not, it is a bug that I discovered while I was testing the feature. If you like I can remove it for this PR and make an other for that.

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
tchlyah: It's the same meaning of `fieldsContainOwnerManyToMany`, it means that the relationship contains at least one `embedded` entity

generators/generator-base-private.js
```diff
@@ -1139,12 +1139,15 @@ module.exports = class extends Generator {
      * @param {Array|Object} fields - array of fields
      * @param {Array|Object} relationships - array of relationships
      * @param {string} dto - dto
+     * @param {boolean} embedded
      * @returns variablesWithTypes: Array
      */
-    generateEntityClientFields(pkType, fields, relationships, dto, customDateType = 'Moment') {
+    generateEntityClientFields(pkType, fields, relationships, dto, embedded, customDateType = 'Moment') {
```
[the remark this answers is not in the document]
tchlyah: Sorry but I don't understand what you mean by that. I've added the new argument in all usages, so it shouldn't cause any problem. Unless there is some usage I'm not aware of (in blueprint I guess ?).

I don't get your proposal either.

generators/generator-base-private.js
```diff
@@ -1139,12 +1139,15 @@ module.exports = class extends Generator {
      * @param {Array|Object} fields - array of fields
      * @param {Array|Object} relationships - array of relationships
      * @param {string} dto - dto
+     * @param {boolean} embedded
```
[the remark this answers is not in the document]
tchlyah: Sure, I'll add it with the other fixes

generators/entity-client/files.js
```diff
@@ -40,6 +40,17 @@ const CLIENT_REACT_TEMPLATES_DIR = 'react';
 const angularFiles = {
     client: [
         {
+            path: ANGULAR_DIR,
+            templates: [
+                {
+                    file: 'entities/entity.model.ts',
+                    // using entityModelFileName so that there is no conflict when genertaing microservice entities
```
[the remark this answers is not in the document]
tchlyah: I didn't write that, just moved the code, but I'll fix it 😉

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -241,23 +257,33 @@ public class <%= asDto(entityClass) %> implements Serializable {
             return false;
         }
 
+        <%_ if (!embedded) { _%>
         <%= asDto(entityClass) %> <%= asDto(entityInstance) %> = (<%= asDto(entityClass) %>) o;
         if (<%= asDto(entityInstance) %>.getId() == null || getId() == null) {
-            return false;
+        return false;
```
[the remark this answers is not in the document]
tchlyah: no, fixing it

generators/generator-base-private.js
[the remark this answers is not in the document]
mshima: It means that adding `embedded` before `customDateType` will make old blueprints use embedded instead of customDateType correct value.
This is a breaking change, embedded should be added after customDateType.

The suggestion is to create an object that will contain every parameter as attribute.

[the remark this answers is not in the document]
tchlyah: Ok, I understand. But creating an object holding parameters is also a breaking change, so it is not for this PR.

I will add `embedded` after `customDateType` with default value to `false`in order to not break actual blueprints!

generators/entity-client/files.js
mshima: This is correct. It's done this way currently.
But it should be renamed to isEmbedded too.
Actually you should use isEmbedded everywhere.

tchlyah: I don't really like using the keyword `is` in variables, it should be used only on methods. If I rename this to `isEmbedded`, `readOnly` should also be renamed to `isReadOnly`, otherwise it won't be consistent! same thing for `jpaMetamodelFiltering`, `validation` ...

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
mshima: This should be on a different PR.
It should be reviewed by a JPA developer.
And will be easier to get approved.

generators/entity-client/files.js
```diff
@@ -301,16 +315,18 @@ function writeFiles() {
...
-                    this.entityInstance,
-                    this.entityClass,
-                    this.entityAngularName,
-                    this.entityFolderName,
-                    this.entityFileName,
-                    this.entityUrl,
-                    this.clientFramework,
-                    this.microserviceName
-                );
+                if (!this.embedded) {
```
mshima: Became too confusing, must be refactored. 

```
let variables***;
let microserviceName = this.microserviceName;
if (this.clientFramework === 'angularX') {
    variables = somethingAngular
} else if (this.clientFramework === 'react') {
    variables = somethingReact;
    microserviceName = undefined;
} else {
    this.addEntityToMenu(this.entityStateName, this.enableTranslation, this.clientFramework, this.entityTranslationKeyMenu);
    return;
}
this.writeFilesToDisk( variables );
addEnumerationFiles(this, CLIENT_NG2_TEMPLATES_DIR, ANGULAR_DIR);
if (!this.isEmbeded) {
    this.addEntityToModule(variables, microserviceName);
    this.addEntityToMenu(variables);
}
```

***variables should include every specific variable like:
clientFiles(writeFilesToDisk), clientTemplatesDir(writeFilesToDisk), enumTemplatesDir(addEnumerationFiles), enumDestDir(addEnumerationFiles)

I think you should create a different PR for this refactoring, I can review it.

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity.model.ts.ejs
```diff
@@ -18,7 +18,7 @@
 -%>
 <%
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, embedded);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, 'Moment', embedded);
```
mshima: You should pass undefined here.
The default from the method should be applied.

generators/entity-client/templates/angular/src/test/javascript/spec/app/entities/entity-management.service.spec.ts.ejs
```diff
@@ -19,7 +19,7 @@
 <%_
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
 const tsKeyId = generateTestEntityId(primaryKeyType);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, embedded);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, 'Moment', embedded);
```
mshima: You should pass undefined here.
The default from the method should be applied.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.model.ts.ejs
```diff
@@ -18,7 +18,7 @@
 -%>
 <%
 const primaryKeyType = getPkTypeBasedOnDBAndAssociation(authenticationType, databaseType, relationships);
-const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, embedded);
+const variablesWithTypes = generateEntityClientFields(primaryKeyType, fields, relationships, dto, 'Moment', embedded);
```
mshima: You should pass undefined here.
The default from the method should be applied.

generators/entity-server/files.js
```diff
@@ -156,7 +156,13 @@ const serverFiles = {
                 {
                     file: 'package/domain/Entity.java',
                     renameTo: generator => `${generator.packageFolder}/domain/${generator.asEntity(generator.entityClass)}.java`
-                },
+                }
+            ]
+        },
+        {
+            condition: generator => !(['mongodb', 'couchbase'].includes(generator.databaseType) && generator.embedded),
```
mshima: `generator => ! generator.embedded` isn't enough?

generators/entity/index.js
```diff
@@ -877,8 +879,11 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 context.relationships.forEach(relationship => {
                     const otherEntityName = relationship.otherEntityName;
                     const otherEntityData = this.getEntityJson(otherEntityName);
-                    if (otherEntityData && otherEntityData.microserviceName && !otherEntityData.clientRootFolder) {
-                        otherEntityData.clientRootFolder = otherEntityData.microserviceName;
+                    if (otherEntityData) {
+                        if (otherEntityData.microserviceName && !otherEntityData.clientRootFolder) {
+                            otherEntityData.clientRootFolder = otherEntityData.microserviceName;
+                        }
+                        relationship.otherEntityIsEmbedded = otherEntityData.embedded;
```
mshima:
```
if (otherEntityData.embedded) {
    relationship.otherEntityIsEmbedded = true;
}
```
Otherwise otherEntityIsEmbedded will be written to entity.json for every relationship.

generators/generator-base-private.js
```diff
@@ -1173,10 +1176,11 @@ module.exports = class extends Generator {
             let fieldType;
             let fieldName;
             const relationshipType = relationship.relationshipType;
+            const otherEntityIsEmbedded = relationship.otherEntityIsEmbedded;
             if (relationshipType === 'one-to-many' || relationshipType === 'many-to-many') {
                 fieldType = `I${relationship.otherEntityAngularName}[]`;
                 fieldName = relationship.relationshipFieldNamePlural;
-            } else if (dto === 'no') {
+            } else if (dto === 'no' || otherEntityIsEmbedded) {
```
mshima: otherEntityIsEmbedded is used only once, use relationship.otherEntityIsEmbedded instead

generators/generator-base-private.js
```diff
@@ -1219,8 +1223,9 @@ module.exports = class extends Generator {
         const typeImports = new Map();
         relationships.forEach(relationship => {
             const relationshipType = relationship.relationshipType;
+            const otherEntityIsEmbedded = relationship.otherEntityIsEmbedded;
             let toBeImported = false;
-            if (relationshipType === 'one-to-many' || relationshipType === 'many-to-many') {
+            if (relationshipType === 'one-to-many' || relationshipType === 'many-to-many' || otherEntityIsEmbedded) {
```
mshima: otherEntityIsEmbedded is used only once, use relationship.otherEntityIsEmbedded instead

mshima:
Thanks @tchlyah!
Sorry for the late review.
This is big PR, so it can take some time to get everything properly reviewed.

@tchlyah you should create at least one integration test for each db.

@pascalgrimaud what do you think about this feature?

A review for java generated code is needed too.

pascalgrimaud:
the jhipster-core part has been merged so we need to go on :)

I agree with this feature if we add some new entities with this feature in our CI, so we are sure it works.

Then, we need to add some docs at jhipster.github.io too

murdos:
This seems to be similar to something that I was planning to do, but that I would have called `valueObject`, in a more [DDD approach](https://github.com/jhipster/generator-jhipster/issues/11122) and that I was planning for SQL databases (using JPA `@Embedded` and `@Embeddable`)

Are we sure that `embedded` is the word we want to retain and advertise?

tchlyah:
@mshima Thanks for your review, I tried to fix everything as much as I could.

@pascalgrimaud, thanks for your approval, I added integration tests for both Couchbase and MongoDB.

@murdos, in this PR, `embedded` is only used for Document databases, you can keep the same keyword for SQL databases.

generators/entity-client/files.js
MathieuAA: @mshima I don't agree. If it's a function, then yes it should be called `isXXX`. If it's a variable, the name prevails.

mshima: No problem

generators/entity-client/files.js
mshima: Just pushed the a PR with the refactoring https://github.com/jhipster/generator-jhipster/pull/11300.
Once merged, just rebase and add the necessary condition.

MathieuAA: The refactoring PR is merged, thanks @mshima!

generators/entity-client/files.js
```diff
@@ -308,7 +322,7 @@ function writeFiles() {
                 files = reactFiles;
                 destDir = REACT_DIR;
                 templatesDir = CLIENT_REACT_TEMPLATES_DIR;
-            } else {
+            } else if (!this.embedded) {
```
mshima: You should move this condition to inside else bracket.
Otherwise the method will continue on embedded.

tchlyah: Yes you're right! Fixing it right now

generators/entity-server/files.js
```diff
@@ -192,7 +198,7 @@ const serverFiles = {
             ]
         },
         {
-            condition: generator => generator.reactive && ['mongodb', 'cassandra', 'couchbase'].includes(generator.databaseType),
+            condition: generator => generator.reactive && !generator.embedded,
```
mshima: Removing `['mongodb', 'cassandra', 'couchbase'].includes(generator.databaseType)` was on purpose?

tchlyah: No, I didn't pay attention, it was a mistake after a review commit...

mshima:
Changes to the generators are ok now.
Last review https://github.com/jhipster/generator-jhipster/pull/11239#discussion_r378873813

tchlyah:
Do we need any other approval to merge this?

mshima:
I reviewed the generator changes.
If someone reviewed the java changes, then were are good to go.

tchlyah:
@pascalgrimaud can you please review the java part ?

pascalgrimaud:
I can try to review, but just to let you know, I'm not a MongoDB or Couchbase user at all.
Maybe @hdurix as he used a lot Couchbase ?

tchlyah:
It seems that he already approved the PR. @hdurix do you have any other remarks on that ?

hdurix:
Looks good :)

pascalgrimaud:
good job @tchlyah !

AlexGreg:
Hello

Is there any documentation on how to use embedded entities?

Thank you!

murdos:
@tchlyah: I just tried to test this feature using one of our sample project used for CI: [ngx-mongodb-kafka-cucumber](https://github.com/jhipster/generator-jhipster/tree/master/test-integration/samples/ngx-mongodb-kafka-cucumber) with the `DocumentBankAccount` and `EmbeddedOperation` entities (also from CI sample tests).

I don't understand how this is supposed to work: there's no way to create an embeddedOperation from the UI.
Also, on the server side, I don't understand why `EmbeddedOperation` has a (db) reference to `DocumentBankAccount` whereas the purpose of an embedded class is to be used in other entities. For e.g. if I have an other entity `CreditCard` that has operations, I don't want `EmbeddedOperation` to have both `documentBankAccount` and `creditCard` properties.
Finally, could you confirm that the JDL for the aforementioned entities would be something like that:
```
entity DocumentBankAccount

@embedded
entity EmbeddedOperation

relationship OneToMany {
  DocumentBankAccount to EmbeddedOperation
}

```

bro0k:
How to support Collection\<Type\>?
`
List<String>
`

pascalgrimaud:
ping @tchlyah : we're waiting your answer for @murdos's question

tchlyah:
> @tchlyah: I just tried to test this feature using one of our sample project used for CI: [ngx-mongodb-kafka-cucumber](https://github.com/jhipster/generator-jhipster/tree/master/test-integration/samples/ngx-mongodb-kafka-cucumber) with the `DocumentBankAccount` and `EmbeddedOperation` entities (also from CI sample tests).

Sorry for the delay, I just had a baby 👶so I don't really find time to do anything...

> I don't understand how this is supposed to work: there's no way to create an embeddedOperation from the UI.

Yes indeed, the MR is focused only on back side, I don't have much front experience, so I can't help on that part. I hope that someone can work this through.

> Also, on the server side, I don't understand why `EmbeddedOperation` has a (db) reference to `DocumentBankAccount` whereas the purpose of an embedded class is to be used in other entities. For e.g. if I have an other entity `CreditCard` that has operations, I don't want `EmbeddedOperation` to have both `documentBankAccount` and `creditCard` properties.

Yes, I'm aware of that, this is a limitation of jhipster, there is an issue about that: https://github.com/jhipster/generator-jhipster/issues/1569. You don't have this problem for other relationship types.

> Finally, could you confirm that the JDL for the aforementioned entities would be something like that:
> 
> ```
> entity DocumentBankAccount
> 
> @embedded
> entity EmbeddedOperation
> 
> relationship OneToMany {
>   DocumentBankAccount to EmbeddedOperation
> }
> ```

Yes, I confirm

tchlyah:
> How to support Collection\<Type\>?
> 
> `
> 
> List<String>
> 
> `

I would love to have that! But for now you can wrap your `String` into an entity

PR #11254, opened 2020-02-07
Prompts: return promises instead of using generator.async/done

mshima:
- When using this.async(), we should treat promises exceptions by ourselves.
   We are not doing that.
- generator.async is kept for legacy compatibility only.
- Promises are recommended.

Reference:
https://yeoman.io/authoring/running-context.html at 'Asynchronous tasks' topic

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -30,24 +30,20 @@ module.exports = {
     askForMoreModules,
 };
 
-function askForInsightOptIn() {
-    const done = this.async();
-
-    this.prompt({
+async function askForInsightOptIn() {
+    const answers = await this.prompt({
         when: () => statistics.shouldWeAskForOptIn(),
         type: 'confirm',
         name: 'insight',
         message: `May ${chalk.cyan('JHipster')} anonymously report usage statistics to improve the tool over time?`,
         default: true,
-    }).then(prompt => {
-        if (prompt.insight !== undefined) {
-            statistics.setOptOutStatus(!prompt.insight);
-        }
-        done();
     });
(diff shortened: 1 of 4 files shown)


atomfrede:
This looks nice, as it makes the whole code more easier to understand. I think we should bring this forward CC @jhipster/developers

DanielFran:
@mshima The scripts used in the generator are also used in the jhipster bom project... Can you take care of the impacts please?

mshima:
@DanielFran yep, I reverted most of the changes.
I am trying to understand the build failures, but looks like master is failing not specific to my branch.
Anyway the CI changes are to detect generated source changes between the PR and the branch destination of the merge (master in this case). Should be really useful for jhipster 7 refactoring that are coming. I will create a new PR for this.

mshima:
@deepu105 what’s the reason for https://github.com/jhipster/generator-jhipster/commit/7c331f5c902db5b0c2b122dfb3d907fe0943cf9a?

It’s not used anywhere inside generator-jhipster.
This PR will break it.

mshima:
Meta parameter can cause unexpected behaviour  as it is right now.
The function will be queued as a task, when the task is executed it passes the generator parameter as parameter to every task.
In this case meta will be the generator argument, so the prompt will be skipped if arguments is passed to the generator.
if meta is required by some other project, meta should be implemented with 2 functions, one for meta and another for the task.

deepu105:
You can break it. I started working on that with idea of exposing prompts to be used in jh online but I never pursued it further as it needed updates and some more breaking changes so you can remove those

PR #11256, opened 2020-02-07
refactor: remove cassandra unit

ctamisier:
Remove cassandra unit dependency.

Historically, `EmbeddedCassandraServerHelper.startEmbeddedCassandra()` feature was used and it was the way to have a running cassandra to perform tests.

Then; testcontainers has been used to start a cassandra instance (jhipster/generator-jhipster@9951387)

Today the only feature we use from cassandra unit is to load CQL files through `dataLoader.load(new ClassPathCQLDataSet(cqlDir + fileName, false, false, CASSANDRA_UNIT_KEYSPACE));`

This dependency is quite heavy, it pulls `org.apache.cassandra:cassandra-all` which is the whole cassandra server that we don't use .

### DIFF
--- a/generators/server/templates/src/test/java/package/CassandraKeyspaceIT.java.ejs
+++ b/generators/server/templates/src/test/java/package/CassandraKeyspaceIT.java.ejs
@@ -42,8 +42,8 @@ public class CassandraKeyspaceIT extends AbstractCassandraTest {
     private Session session;
 
     @Test
-    public void shouldListCassandraUnitKeyspace() throws Exception {
+    public void shouldListCassandraTestKeyspace() {
         Metadata metadata = session.getCluster().getMetadata();
-        assertThat(metadata.getKeyspace(CASSANDRA_UNIT_KEYSPACE)).isNotNull();
+        assertThat(metadata.getKeyspace(CASSANDRA_TEST_KEYSPACE)).isNotNull();
     }
 }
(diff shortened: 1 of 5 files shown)


generators/server/templates/src/test/java/package/CassandraKeyspaceIT.java.ejs
```diff
@@ -42,8 +42,8 @@ public class CassandraKeyspaceIT extends AbstractCassandraTest {
     private Session session;
 
     @Test
-    public void shouldListCassandraUnitKeyspace() throws Exception {
+    public void shouldListCassandraUnitKeyspace() {
```
DanielFran: Rename to shouldListCassandraTestKeyspace() ?

ctamisier: Oh yes !

DanielFran:
Thanks @ctamisier , I will remove the dependency in the bom

ctamisier:
Okay, thanks !!

SudharakaP:
> Thanks @ctamisier , I will remove the dependency in the bom

@DanielFran : Done. 😄

SudharakaP:
I should point out without further ado that starting from this change the cassandra builds started showing the same backend errors as before. I am not certain why; but maybe we should revert this and do some more testing before removing this dependency? :thinking: :thinking:

pascalgrimaud:
so strange...
You're right, I'd prefer stability !!

PR #11260, opened 2020-02-08
fix: use activated translation

ctamisier:
use `userManagement.activated` and `userManagement.deactivated` for the activation button in react.


pascalgrimaud:
thanks @ctamisier

Issue #11129, opened 2020-01-16
Redis Cluster support


pascalgrimaud:
##### **Overview of the feature request**

Our current implementation of Redis is for single server. In real production, it will never be the case, it will be with a Redis Cluster. 

##### **Motivation for or Use Case**

A single server of Redis will be never used in prod. 

See https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs#L729

See https://github.com/redisson/redisson/wiki/2.-Configuration#24-cluster-mode

The goal would be to have:
- a docker compose file to start a Redis cluster in dev
- the correct config in yaml
- the integration tests, which could still be with single server I think

Then, for all deployment (docker compose, k8s etc) use a Redis Cluster

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/pull/10057
https://github.com/jhipster/generator-jhipster/issues/9280

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
cc @Shaolans @juliensadaoui @avdev4j

pascalgrimaud:
I'm starting to work on the dev part.

pascalgrimaud:
About the Ops part, in fact, nothing is done today: Kubernetes, Openshift, AWS, etc...
It's only done with Docker-Compose.

So I'm reducing the bounty to 100$ to be fair -> if someone want to work on the Ops part, let me know and let's open a new ticket for that

pascalgrimaud:
I'm closing this as both PR have been merged.
I'll update the documentation about this

pascalgrimaud:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/13928

vrijmetse:
Does this feature works with AWS Elasticache Redis (with cluster mode enabled) like stated here https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Endpoints.html#Endpoints.Find.RedisCluster ?

pascalgrimaud:
Look at the specific code here:
https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs#L746-L756

For me, the last time I tried the cluster mode:
- it works locally
- it works for one of my customer project, with Redis deployed with Kubernetes

I don't see any reason why it won't work for AWS Elasticache Redis. I would say: try :)

PR #11264, opened 2020-02-08
Add property to use Redis Cluster

pascalgrimaud:
It's the dev part of https://github.com/jhipster/generator-jhipster/issues/11129
Depends on: https://github.com/jhipster/jhipster/pull/571

By default, it will use the single node. You need to update the application-dev.yml or application-prod.yml to use Redis Cluster. The related properties are commented.

Once it's enabled, you can start the cluster, using `docker-compose -f src/main/docker/redis-cluster.yml up -d` :

docker-compose-redis-cluster

Here when playing with the application, and check the different KEYS in my :hamburger: Redis Cluster : 

docker-compose-redis-keys

_____

### DIFF
--- a/generators/server/templates/src/main/docker/redis-cluster.yml.ejs
+++ b/generators/server/templates/src/main/docker/redis-cluster.yml.ejs
@@ -1,0 +1,86 @@
+<%#
+ Copyright 2013-2020 the original author or authors from the JHipster project.
+ This file is part of the JHipster project, see https://www.jhipster.tech/
+ for more information.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


generators/server/templates/src/main/docker/redis-cluster.yml.ejs
```diff
@@ -0,0 +1,86 @@
...
+        command:
+            - "redis-server"
+            - "--port 7003"
+            - "--cluster-enabled yes"
+            - "--cluster-config-file nodes.conf"
+            - "--cluster-node-timeout 5000"
+            - "--appendonly yes"
+        ports:
+            - "7003:7003"
+    <%= baseName.toLowerCase() %>-redis-4:
```
hdurix: maybe 3 nodes is enough?

pascalgrimaud: good question, honestly I don't know. I'm not a Redis expert at all, I simply saw that in several docker-compose files, that's why I did the same but without using network

pascalgrimaud:
Thanks for the review @Shaolans and @hdurix 
I'm merging this as I already merged the jhipster lib part.

PR #11276, opened 2020-02-11
fix invalid loadData end tag with namespace

csaban:
When ext: namespace is applied for open tag, the end tag gets generated as invalid - missing the namespace.

### DIFF
--- a/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
+++ b/generators/server/templates/src/main/resources/config/liquibase/changelog/initial_schema.xml.ejs
@@ -165,7 +165,7 @@
                   <%_ } _%>
                   tableName="<%= jhiTablePrefix %>_authority">
             <column name="name" type="string"/>
-        </loadData>
+        </<% if (devDatabaseType === 'mssql' || prodDatabaseType === 'mssql') { %>ext:<% } %>loadData>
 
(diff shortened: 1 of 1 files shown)


pmverma:
Thanks @csaban, I coded it and I really wonder the ci tests passed without a correct end tag. Honestly, I was unaware that mssql is using some namespace.

pmverma:
Oh and congrats for your first contribution. 🎉🎉🎉

pascalgrimaud:
@pmverma : that's because mssql is not covered by our CI
Maybe we can add some of them in daily builds project

pmverma:
Oh, I see @pascalgrimaud!
It would be really nice if all the options jhipster provides, are tested automatically using automation.
Also, daily builds with hundreds of options may become difficult to manage in a few years or maybe months.

pascalgrimaud:
It's nearly impossible to test all options in main generator-jhipster, and I don't think it's a good idea, as we have a lot of options which is not used a lot...

For example, Oracle is not tested, same for MsSQL.
I just look at the statistic : Oracle and MsSQL are more used than Cassandra or Couchbase, so maybe we should add them to the builds.

2 solutions for that:
- add new builds to jhipster-daily-builds : the project is here for this reason
- replace 1 build to use MsSQL instead of MySQL (as it is already tested several time) to main generator-jhipster

PR #11278, opened 2020-02-11
Prettier: fix some indent and add comment prettier-ignore to toString method

pascalgrimaud:
@clementdessoude : to avoid the toString method to be formatted by prettier, I'm adding this, because the generated toString is more readable. What do you think ? 

_____

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -554,6 +554,7 @@ public class <%= asEntity(entityClass) %> implements Serializable {
         return 31;
     }
 
+    // prettier-ignore
     @Override
     public String toString() {
         return "<%= asEntity(entityClass) %>{" +
--- a/generators/entity-server/templates/src/main/java/package/service/dto/EntityCriteria.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/service/dto/EntityCriteria.java.ejs
@@ -155,19 +155,20 @@ public class <%= entityClass %>Criteria implements Serializable, Criteria {
         return Objects.hash(
 <%_ filterVariables.forEach((filterVariable, index) => { _%>
     <%_ if (filterVariables.length === (index + 1)) { _%>
-        <%= filterVariable.name %>
+            <%= filterVariable.name %>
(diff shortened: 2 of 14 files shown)


hdurix:
I would prefer to find a better format in prettier if any, than adding all these rules here. Any idea of how and if this is possible?

pascalgrimaud:
@hdurix : agree. But I already discussed with @clementdessoude about that and it seems not so easy...

murdos:
I'm not fond of this formatter:off too. I was thinking that maybe building the toString() method in a different way could help  the formatting and (maybe the readability?), e.g. with apache-commons `ToStringBuilder`: 
```
public String toString() {
     return new ToStringBuilder(this).
       append("name", name).
       append("age", age).
       append("smoker", smoker).
       toString();
   }
```

clementdessoude:
Hi @hdurix!

It is difficult to "fix" this in prettier. Adding a custom option to not format toString method is not something we want to do, IMO (we want as less options as possible: see [here](https://prettier.io/docs/en/option-philosophy.html) for more infos), and there is no way we could know how to format this kind of concatenation of strings...

Concerning you PR @pascalgrimaud, I'll have to take a look at time once I'm home but LGTM. You also could use `// prettier-ignore` comments instead to have one comment instead of two for each toString

pascalgrimaud:
Ok, I'll replace formatter:off by prettier-ignore, it's better. 

Then, this PR is simply a first step to include prettier:java into JHipster. We need to agree together before merging this, so I see different solutions:
1) don't do anything, prettier:java will format the toString
2) change the current toString to be similar to the result with prettier:java
3) add prettier-ignore so it won't change if you generate / regenerate the toString with your IDE
4) from @murdos, change the toString to use ToStringBuilder

So WDYT ?

hdurix:
Yes no problem @clementdessoude, I understand this is not a good idea.

I like @murdos idea, especially as prettier:java is not in the generator yet.

pascalgrimaud:
I like @murdos idea too but I'm afraid our users won't be happy as it will be different from what the IDE will generate, unless you change manually the template.

pascalgrimaud:
Closing this, following https://github.com/jhipster/generator-jhipster/pull/11354#issuecomment-589168158

PR #11288, opened 2020-02-12
Fix OAuth user sync logic for Okta

mraible:
The `updated_at` claim is now a different type from Okta. It used to be an Integer, now it's an Instant. Not sure if this is a change in Spring Security or Okta. This code block doesn't seem to get hit for Keycloak.

Here's the stack trace that happens when using microservices:

```
gateway_1 | 2020-02-12 02:55:37.883 ERROR 1 --- [ XNIO-1 task-31] o.z.problem.spring.common.AdviceTraits   : Internal Server Error
gateway_1 |
gateway_1 | java.lang.ClassCastException: class java.time.Instant cannot be cast to class java.lang.Integer (java.time.Instant and java.lang.Integer are in module java.base of loader 'bootstrap')
gateway_1 | 	at com.okta.developer.gateway.service.UserService.syncUserWithIdP(UserService.java:171)
gateway_1 | 	at com.okta.developer.gateway.service.UserService.getUserFromAuthentication(UserService.java:216)
```

I did have this fix:

```java
Instant idpModifiedDate;
if (details.get("updated_at") instanceof Instant) { // Okta
    idpModifiedDate = (Instant) details.get("updated_at");
} else { // Keycloak
    idpModifiedDate = new Date(Long.valueOf((Integer) details.get("updated_at"))).toInstant();
}
```

But this seems like a more appropriate fix since it's Okta-only AFAICT:

```java
Instant idpModifiedDate = (Instant) details.get("updated_at");
```


generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -762,7 +762,12 @@ public class UserService {
             // if IdP sends last updated information, use it to determine if an update should happen
             if (details.get("updated_at") != null) {
                 Instant dbModifiedDate = existingUser.get().getLastModifiedDate();
-                Instant idpModifiedDate = new Date(Long.valueOf((Integer) details.get("updated_at"))).toInstant();
+                Instant idpModifiedDate;
```
sendilkumarn: nut: can we move this to a separate function?

mraible: @sendilkumarn I just simplified the logic since this seems to only affect Okta. Please have a look.

PR #11294, opened 2020-02-12
Redundant parameter eager

### LINKED ISSUE #11216: Redundant query parameter in many-to-many relationships in entity

When we generate an entity which contains many-to-many relationship with another entity, JHipster creates a query parameter for getAll endpoints called `eagerload`, but it's not used anywhere in EntityResource.java. 

There is already a logic to use either `findAllWithEagerRelationships` or `findAll` in the ejs template based on if the entity has a many-to-many relationship or not, so not sure why there is also an unused `eagerload` query parameter in the endpoints. 

If we want to give the option to load it eager or not during the runtime, shouldn't the generated Java code have:
```
if (eagerload) {
    return entityRepository.findAllWithEagerRelationships();
} else {
    return entityRepository.findAll();
}
```
?

To make it configurable to eager or lazy load many-to-many relationships during runtime

##### **Reproduce the error**

1- Generate a microservice
2- Add an entity that has many-to-many relationship with another entity
3- Check the generated EntityResource.java file, getAll method have `eagerload` parameter but it isn't used.

##### **Suggest a Fix**

We should either remove `eagerload` parameter or use it in the getAll method.

##### **JHipster Version(s)**

6.5.1

We used this sample JDL https://github.com/jhipster/jdl-samples/blob/master/simple-online-shop.jh

MacOS Catalina

seraphinandrieux:
PR to fix #11216 

I chose to remove the parameter eager load inside the generator. I based my solution on the logic already made with entities which have a Many to Many relation.
Inside the generator each time we have a many to many relation between two entities, the ejs chooses to use the method `findAllWithEager()`.
So I removed the query parameter `eagerload` which is useless if we follow the logic used someplace else.

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
@@ -51,7 +51,7 @@ _%>
     <%_ if (pagination === 'no') { _%>
     public List<<%= instanceType %>> getAll<%= entityClassPlural %>(<% if (fieldsContainNoOwnerOneToOne) { %>@RequestParam(required = false) String filter<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
         log.debug("REST request to get all <%= entityClassPlural %>");<% if (viaService) { %>
-        return <%= entityInstance %>Service.findAll();<% } else if (dto === 'mapstruct') { %>
+        return <%= entityInstance %>Service.<% if (fieldsContainOwnerManyToMany) { %>findAllWithEagerRelationships<% } else { %>findAll<% } %>();<% } else if (dto === 'mapstruct') { %>
         List<<%= asEntity(entityClass) %>> <%= entityInstancePlural %> = <%= entityInstance %>Repository.<% if (fieldsContainOwnerManyToMany) { %>findAllWithEagerRelationships<% } else { %>findAll<% } %>();
         return <%= entityListToDtoListReference %>(<%= entityInstancePlural %>);<% } else { %>
         return <%= entityInstance %>Repository.<% if (fieldsContainOwnerManyToMany) { %>findAllWithEagerRelationships<% } else { %>findAll<% } %>();<% } %>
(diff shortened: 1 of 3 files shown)


generators/entity-server/templates/src/main/java/package/common/get_all_template.ejs
```diff
@@ -49,23 +49,19 @@ _%>
...
         return <%= entityListToDtoListReference %>(<%= entityInstancePlural %>);<% } else { %>
         return <%= entityInstance %>Repository.<% if (fieldsContainOwnerManyToMany) { %>findAllWithEagerRelationships<% } else { %>findAll<% } %>();<% } %>
     <%_ } if (pagination !== 'no') { _%>
-    public ResponseEntity<List<<%= instanceType %>>> getAll<%= entityClassPlural %>(Pageable pageable<% if (reactive) { %>, ServerHttpRequest request<% } %><% if (fieldsContainNoOwnerOneToOne) { %>, @RequestParam(required = false) String filter<% } %><% if (fieldsContainOwnerManyToMany) { %>, @RequestParam(required = false, defaultValue = "false") boolean eagerload<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
+    public ResponseEntity<List<<%= instanceType %>>> getAll<%= entityClassPlural %>(Pageable pageable<% if (reactive) { %>, ServerHttpRequest request<% } %><% if (fieldsContainNoOwnerOneToOne) { %>, @RequestParam(required = false) String filter<% } %>) {<%- include('get_all_stream_template', {viaService: viaService}); -%>
```
MathieuAA: I'm not so sure about this. Previously, the default was `false`, right? So removing it but making fetching eager relationships by default kinda makes it "always true".
I feel like this is a breaking change, kinda.

seraphinandrieux: Indeed my bad, I'm updating the code to take account of this default value. 
Thanks for your review !

MathieuAA: Maybe we should make fetching eager relationships the default for the next major version (v7). Again, not so sure about this either :(

seraphinandrieux: Do you recommend me to put back the `eagerload` query argument and keep the logic used ?

Issue #11283, opened 2020-02-11
Create New Workflow for Label Cleanup


SudharakaP:
##### **Overview of the feature request**

As per the following discussion, I thought it would be nice to have a workflow that cleans up labels in our issues. So for example once a issue is closed labels like `triage`, `undefined`, `needs-reproduction` and `waiting for information` needs to be automatically removed. 

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/jhipster-dev/xfrODe_g14U

##### **Motivation for or Use Case**

Cleanup old labels of closed issues. 😄 

##### **Related issues or PR**

-   [x] Checking this box is mandatory (this is just to show you read everything)

SudharakaP:
I think we can add more labels for the cleanup (for example status labels maybe?). But given we don't adhere to a strict labeling policy I wanted to keep it simple. Please feel free to let me know if you want to include more labels or have any other suggestions. 😄

pascalgrimaud:
it's good idea ! I just need time to think more about it and propose something here too !

SudharakaP:
@pascalgrimaud : Thanks; [@mshima  suggested](https://github.com/jhipster/generator-jhipster/pull/11284#issuecomment-584813557) to keep the `area: triage` and `theme: undefined` and remove the `status: work in progress`.

Please feel free to let me know your thoughts as well; once everyone comes to an agreement I'll make some modifications. 😄

DanielFran:
I believe we should change, if possible, the workflow that add the 2 labels to add them only if no label of the same type is not present.

mshima:
> @pascalgrimaud : Thanks; [@mshima suggested](https://github.com/jhipster/generator-jhipster/pull/11284#issuecomment-584813557) to keep the `area: triage` and `theme: undefined` and remove the `status: work in progress`.

Close/reopen off the issue will remove those labels, it should be kept for reopen.

I think we should remove every `status: *`, and rename those that should be kept.
`status: bounty-requested` => `bounty: requested`
`status: bounty-approved` => `bounty: approved`
`status: should go to modules` => `resolution: should go to modules`
`status: wont-fix` => `resolution: wont-fix`

mshima:
> I believe we should change, if possible, the workflow that add the 2 labels to add them only if no label of the same type is not present.

I agree with you, but this will make the workflow much more complicated.
Instead of just writing, it will be read then write.

SudharakaP:
[...]

PR #11312, opened 2020-02-14
Create auto labeler workflow for PRs [skip ci]

mshima:
Following https://github.com/jhipster/generator-jhipster/issues/11283

This PR creates a workflow for auto labeling PRs based on the files changes.

It uses scheduled runs and tags PRs that aren't tagged yet.
Why don't use on pull_request? https://github.com/actions/labeler/issues/12
Once this bug is resolved, we should switch to on pull_request.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,61 @@
+"theme: cli":
+    - cli/*
+
+"theme: CI Builds":
+    - .github/workflows/*
+    - azure-pipelines.yml
+
+"theme: dependencies":
+    - package.json
+    - package-lock.json
+
+"theme: front":
+    - generators/client/index.js
+    - generators/client/prompts.js
+    - generators/needle-api/*
+    - generators/entity-client/*
+
+"theme: angular":
+    - generators/client/files-angular.js
+    - generators/client/templates/angular/**/*
+    - generators/entity-client/templates/angular/**/*
+
+"theme: react":
+    - generators/client/files-react.js
+    - generators/client/templates/react/**/*
+    - generators/entity-client/templates/react/**/*
+
+"theme: jhipster-internals":
+    - generators/generator-*.js
+    - generators/statistics.js
+    - generators/utils.js
+    # client/server common
+    - generators/app/**/*.js
+    - generators/common/**/*.js
+    - generators/needle-*.js
+    - generators/cleanup.js
+    - generators/entity/**/*
+
+"theme: i18n":
+    - generators/languages/**/*
+    - generators/entity-i18n/**/*
+
+"theme: java":
+    - generators/server/**/*
+    - generators/entity-server/**/*
+
+"theme: upgrade":
+    - generators/upgrade/**/*
+
+"theme: docker :whale:":
+    - generators/docker*.js
+    - generators/docker-compose/**/*
+
+"theme: aws":
+    - generators/aws/**/*
+    - generators/aws-containers/**/*
+
+"theme: kubernetes":
+    - generators/kubernetes/**/*
+    - generators/kubernetes-helm/**/*
+    - generators/kubernetes-knative/**/*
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,61 @@
+"theme: cli":
+    - cli/*
+
+"theme: CI Builds":
+    - .github/**/*
```
DanielFran: Should be: .github/workflows/*

DanielFran: Should be: .github/workflows/*

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+    schedule:
+        # * is a special character in YAML so you have to quote this string
+        - cron:  '*/5 * * * *'
+
+jobs:
+    triage:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: mshima/labeler@v2
+              if: github.repository == 'mshima/generator-jhipster' && github.ref == 'master'
```
DanielFran: This condition should reviewed

mshima: 🤦‍♂

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+    schedule:
+        # * is a special character in YAML so you have to quote this string
+        - cron:  '*/5 * * * *'
+
+jobs:
+    triage:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: mshima/labeler@v2
```
pascalgrimaud: why is it your own action ? did you fork the project ? in this case, we can host it under jhipster organization

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,13 @@
+name: "Pull Request Labeler"
+on:
+    schedule:
+        # * is a special character in YAML so you have to quote this string
+        - cron:  '*/5 * * * *'
```
pascalgrimaud: I'm afraid of this, as it will be launched all the time and we won't find in Actions list the other builds

SudharakaP:
I am a bit doubtful on this change. Maybe we need to think of a different approach :thinking:  

1) I think this action only adds labels and it doesn't remove them. Say I make a commit changing `package.json` and then in a later commit removes those changes. Then the label `theme: dependencies` might be staying; there's an issue for this in the parent repository; https://github.com/actions/labeler/issues/14

2) Since the action is forked from https://github.com/actions/labeler and there's [some customizations](https://github.com/mshima/labeler/commits/master) and we'll have to maintain it and probably merge upstream enhancements (such as the useful https://github.com/actions/labeler/issues/14) when they are available.

.github/workflows/labeler.yml
mshima: Sure.
The changes I’ve made is https://github.com/actions/labeler/pull/49.
I don’t have permission to create new repositories.
Can you clone my?
 If you clone upstream, then you have to copy the dependencies too.

.github/workflows/labeler.yml
mshima: We can reduce this to one an hour.
I’ve set to 5 min to test.

mshima:
> I am a bit doubtful on this change. Maybe we need to think of a different approach 🤔
> 
> 1. I think this action only adds labels and it doesn't remove them. Say I make a commit changing `package.json` and then in a later commit removes those changes. Then the label `theme: dependencies` might be staying; there's an issue for this in the parent repository; [actions/labeler#14](https://github.com/actions/labeler/issues/14)
> 2. Since the action is forked from https://github.com/actions/labeler and there's [some customizations](https://github.com/mshima/labeler/commits/master) and we'll have to maintain it and probably merge upstream enhancements (such as the useful [actions/labeler#14](https://github.com/actions/labeler/issues/14)) when they are available.
> 
> Just my two cents 😄

I think labeling removing is minor.
Another problem is that it don’t revised a modified PR to add new labels.
But it a start. We have a lot of unlabeled PRs.

Another approach is to create jhipster teams and use code owner.
It will ask the team to  review the PR based on code owner.

PR #11319, opened 2020-02-15
Move git methods to utils

pvliss:


generators/generator-base.js
```diff
@@ -1320,29 +1320,15 @@ module.exports = class extends PrivateBase {
...
      * @param {object} options[optional] - takes any of child process options
-     * @param {function} callback - a callback function to be called once process complete, The call back will receive code, stdout and stderr
+     * @param {function} callback[optional] - a callback function to be called once process complete, The call back will receive code, stdout and stderr
+     * @return {object} when in synchronous mode, this returns a ShellString. Otherwise, this returns the child process object.
      */
     gitExec(args, options, callback) {
-        callback = arguments[arguments.length - 1]; // eslint-disable-line prefer-rest-params
-        if (arguments.length < 3) {
-            options = {};
-        }
```
mshima: Are we loosing something removing this?

generators/generator-base-private.js
```diff
@@ -625,17 +625,16 @@ module.exports = class extends Generator {
...
+     * @return {boolean} true if installed; false otherwise..
      */
-    isGitInstalled(callback) {
-        this.gitExec('--version', { trace: false }, code => {
-            if (code !== 0) {
-                this.warning('git is not found on your computer.\n', ` Install git: ${chalk.yellow('https://git-scm.com/')}`);
-            }
-            if (callback) callback(code);
-        });
+    isGitInstalled() {
```
mshima: Do we want to keep callback compatibility?
It's really simple to add.

generators/generator-base.js
```diff
@@ -1320,29 +1320,15 @@ module.exports = class extends PrivateBase {
...
-
-        if (!Array.isArray(args)) {
-            args = [args];
-        }
-        const command = `git ${args.join(' ')}`;
-        if (options.trace) {
-            this.info(command);
-        }
-        shelljs.exec(command, options, callback);
+        return jhipsterUtils.gitExec(args, options, callback);
```
mshima: This one is in generator-base.
We should keep compatibility.
```
if (typeof options === 'function') {
  callback = options;
}
```

generators/generator-base.js
mshima: Got it

generators/generator-base.js
mshima: Add that to utils.gitExec.

generators/app/index.js
```diff
@@ -549,23 +549,20 @@ module.exports = class extends BaseBlueprintGenerator {
...
+                        const gitDir = this.gitExec('rev-parse --is-inside-work-tree', { trace: false }).stdout;
+                        // gitDir has a line break to remove (at least on windows)
+                        if (gitDir && gitDir.trim() === 'true') {
+                            this.gitInitialized = true;
                         } else {
-                            this.warning('Git repository could not be initialized, as Git is not installed on your system');
+                            const shellStr = this.gitExec('init', { trace: false });
+                            this.gitInitialized = shellStr.code === 0;
+                            if (this.gitInitialized) this.log(chalk.green.bold('Git repository initialized.'));
+                            else this.warning(`Failed to initialize Git repository.\n ${shellStr.stderr}`);
```
mshima: Isn't git required? Should be error here.

generators/app/index.js
```diff
@@ -549,23 +549,20 @@ module.exports = class extends BaseBlueprintGenerator {
...
                         }
-                    });
+                    } else {
+                        this.warning('Git repository could not be initialized, as Git is not installed on your system');
```
mshima: Isn't git required? Should be error here.

generators/generator-base-private.js
```diff
@@ -625,17 +625,16 @@ module.exports = class extends Generator {
...
+        const gitInstalled = jhipsterUtils.isGitInstalled();
```
mshima: Since we are setting this.gitInstalled.
This could be:
`const gitInstalled = this.gitInstalled || jhipsterUtils.isGitInstalled();`

generators/generator-base-private.js
pvliss: You are right. Will add it back

generators/generator-base.js
pvliss: > This one is in generator-base.
> We should keep compatibility.
> 
> ```
> if (typeof options === 'function') {
>   callback = options;
> }
> ```

Sorry, but I can't find that in generator-base. Are you sure its there?

mshima: generator-base.js should be api stable.
https://github.com/jhipster/generator-jhipster/blob/448d48ec9df570ee4c68f46050e23c0c936b49b7/generators/generator-base.js#L43-L49

https://github.com/jhipster/generator-jhipster/blob/448d48ec9df570ee4c68f46050e23c0c936b49b7/generators/generator-base.js#L1330-L1333
Was removed. It allowed `gitExec('...', callback)`.
So we should add 
```
if (typeof options === 'function') {
  callback = options;
}
```
To fix api compatibility.
This could be added to gitExec at generator-base, but it will be better to add to gitExec at utils.

generators/app/index.js
pvliss: Fixed

generators/app/index.js
pvliss: Fixed

generators/generator-base.js
pvliss: You are right. Did not realize although you actually pointed it out in your first comment as well. Thought I have moved the code verbatim. Anyways, fixed it as per your suggestion

generators/generator-base-private.js
pvliss: Added back

generators/generator-base-private.js
pvliss: Since I added back the callback though we should execute it anyways(with code 0) if we want to keep compatibility, right?

mshima: Sure

generators/app/index.js
```diff
@@ -549,23 +549,20 @@ module.exports = class extends BaseBlueprintGenerator {
...
                         } else {
-                            this.warning('Git repository could not be initialized, as Git is not installed on your system');
+                            const shellStr = this.gitExec('init', { trace: false });
+                            this.gitInitialized = shellStr.code === 0;
+                            if (this.gitInitialized) this.log(chalk.green.bold('Git repository initialized.'));
+                            else this.error(`Failed to initialize Git repository.\n ${shellStr.stderr}`);
                         }
-                    });
+                    } else {
+                        this.error('Git repository could not be initialized, as Git is not installed on your system');
```
mshima: I was wrong it should be warning.
Above too.
It's required only for upgrade.
Sorry about that.

pvliss: No problem. Changed it back to warning

mshima:
Great work.
Thank you @pvliss.

Issue #11321, opened 2020-02-15
react: Multiple issues on entity list page


vishal423:
##### **Overview of the issue**

Few scenarios:

- [x] ~~Delete a record. Post successful operation, you should see two backend calls being made to fetch list page records.~~  Done in #11322
  - http://localhost:8080/api/operations?page=0&size=20&sort=id,asc&cacheBuster=1581786413254
  - http://localhost:8080/api/operations?page=0&size=20&sort=id,asc&cacheBuster=1581786413260
- [x] ~~Update a record. Post successful operation, you should see two backend calls being made to fetch list page records~~  Done in #11322
  - http://localhost:8080/api/operations?page=0&size=20&sort=id,asc&cacheBuster=1581786577562
  - http://localhost:8080/api/operations?page=0&size=20&sort=id,asc&cacheBuster=1581786577564
- [ ] Search (elasticsearch) flow seems to be broken on UI.
  - If you enter any search criteria and click the search icon, no backend API call is being made. Instead, I always see entity equivalent `No records found` message. 
  - If you try to clear the search criteria, you can see entity equivalent `No records message`.

There could be more flows with different combination and I think it would be appropriate to do proper manual testing of generated applications. Ideally, you should have only one useEffect per api call with variable parameters, however, that doesn't seems to be scenario in current code and causing these issues.

##### **Related issues**
#11222 

##### **JHipster Version(s)**
`master`

##### **JHipster configuration**
Refer #11222 for configurations and entity details

##### **Browsers and Operating System**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

qmonmert:
@vishal423 I try to fix the first two bullets points with #11322 
It's not easy :(
If you have an idea for elasticsearch bug ...

qmonmert:
@vishal423 I don't see the bug with elastisearch
On my PR #11322 there is call when I click search icon 
![image](https://user-images.githubusercontent.com/9989211/74609735-1f404b80-50ed-11ea-9d33-414ffe0af8f9.png)
I see the message "No records found" but it's the same on Angular

vishal423:
[...]

PR #11322, opened 2020-02-15
Fix two backend calls when delete and update react entities

qmonmert:
Fix two points (backend calls when delete and update react entities) on 3 for #11321 

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
@@ -101,9 +101,7 @@ export const <%= entityReactName %> = (props: I<%= entityReactName %>Props) => {
   <%_ if (pagination === 'no' || pagination === 'infinite-scroll') { _%>
   useEffect(() => {
     <%_ if (pagination !== 'no') { _%>
-      <%_ if (pagination === 'infinite-scroll') { _%>
     resetAll();
-      <%_ } _%>
     <%_ } else { _%>
     props.getEntities();
     <%_ } _%>
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
@@ -110,14 +110,6 @@ export const <%= entityReactName %> = (props: I<%= entityReactName %>Props) => {
   }, [<%_ if (searchEngine === 'elasticsearch' && pagination === 'no') { _%>search<%_ } _%>]);
   <%_ } _%>
 
-  <%_ if (pagination === 'infinite-scroll') { _%>
-  useEffect(() => {
-    if (props.updateSuccess) {
-      getAllEntities();
-    }
-  });
-  <%_ } _%>
-
   <%_ if (searchEngine === 'elasticsearch') { _%>
   const startSearching = () => {
     if (search) {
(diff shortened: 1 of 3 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -129,6 +121,7 @@ export const <%= entityReactName %> = (props: I<%= entityReactName %>Props) => {
         ...paginationState,
         activePage: 1
       });
+      props.getSearchEntities(search, paginationState.activePage - 1, paginationState.itemsPerPage, `${paginationState.sort},${paginationState.order}`);
```
SudharakaP: Tiny thing; missing a space between `${paginationState.sort}` and `${paginationState.order}`

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -107,15 +107,7 @@ export const <%= entityReactName %> = (props: I<%= entityReactName %>Props) => {
...
-  }, [<%_ if (searchEngine === 'elasticsearch' && pagination === 'no') { _%>search<%_ } _%>]);
-  <%_ } _%>
-
-  <%_ if (pagination === 'infinite-scroll') { _%>
-  useEffect(() => {
-    if (props.updateSuccess) {
-      getAllEntities();
-    }
-  });
+  }, []);
```
SudharakaP: @qmonmert : While you are at this I think you can remove the `<%_ if (pagination === 'infinite-scroll') { _%>` statement as I think it's unnecessary. After all if `<%_ if (pagination !== 'no') { _%>` it implies that `<%_ if (pagination === 'infinite-scroll') { _%>` :smile:

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
qmonmert: there is no space but a comma between ${paginationState.sort} and ${paginationState.order} 🧐

SudharakaP: Yes I am thinking we need to add a space after the comma. 🙃

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
qmonmert: indeed I think you are right, I do that 👍

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
qmonmert: I tried but we can't because this string is used in the request
image

SudharakaP: Ops you are right. My mistake. 🙃

qmonmert: 😉

pascalgrimaud:
@SudharakaP : is the review done and ok ? can it be merged ?

SudharakaP:
@pascalgrimaud : Yes I think this is fine and can be merged. :smile: 

There definitely are other cases such as the one I mentioned at; https://github.com/jhipster/generator-jhipster/issues/11321#issuecomment-593119615. But I think we can handle them separately. I am a bit sick these days and once I get well will try to do investigate further on that. :smile:

PR #11327, opened 2020-02-16
jaxb dependencies for liquibase

Tcharl:
Jaxb B dependency is mandatory to execute liquibase.
Looks like it's not in the main dependencies anymore

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1189,6 +1189,11 @@
                             <artifactId>spring-beans</artifactId>
                             <version>${spring.version}</version>
                         </dependency>
+                        <dependency>
+                            <groupId>jakarta.xml.bind</groupId>
+                            <artifactId>jakarta.xml.bind-api</artifactId>
+                            <version>${jaxb-runtime.version}</version>
+                        </dependency>
                     </dependencies>
                 </plugin>
 <%_ } _%>


pascalgrimaud:
Can you describe better the errors you encounter plz ?
I remember some issue about Jaxb when upgrading to JDK11. Hope we won't introduce some new issue with this

generators/server/templates/pom.xml.ejs
```diff
@@ -1189,6 +1189,11 @@
                             <artifactId>spring-beans</artifactId>
                             <version>${spring.version}</version>
                         </dependency>
+                        <dependency>
```
pascalgrimaud: it should not be present for NoSQL database as there is no liquibase

clementdessoude:
> Can you describe better the errors you encounter plz ?
> I remember some issue about Jaxb when upgrading to JDK11. Hope we won't introduce some new issue with this

It is just that jaxb was included by default with Java 8 but it is not the case anymore in Java 11.

mraible:
I've been using JHipster with Java 11 for months and have yet to experience this issue. Is it a specific Maven or Gradle command that causes it? @Tcharl Can you please provide the steps to produce the problem?

Tcharl:
Hi @mraible , hi @pascalgrimaud , thank you for the explanation @clementdessoude :-)

My configuration is now on Java 13, but I think it's the same on java 11: 
There's a stacktrace when using ./mvnw liquibase:diff, at least with h2disk databases, adding jaxb runtime do get rid of that one :-)

clementdessoude:
Concerning your remark @mraible: do you use the mvn liquibase:diff command? There is indeed a difference between the liquibase dependency and the liquibase plug-in that is not used when running the app (it is only used for running liquibase diff commands)

generators/server/templates/pom.xml.ejs
pascalgrimaud: @Tcharl : we don't have Liquibase for MongoDB, Couchbase, Cassandra, so I still think adding this for these database is useless

mraible:
@Tcharl You are correct. I can reproduce it if I use the `mvn liquibase:diff` command.

```
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:3.6.3:diff (default-cli) on project liquibase: 
Execution default-cli of goal org.liquibase:liquibase-maven-plugin:3.6.3:diff failed: 
A required class was missing while executing org.liquibase:liquibase-maven-plugin:3.6.3:diff: 
javax/xml/bind/JAXBException
```

pascalgrimaud:
There are still 2 things to do before merging this:
- resolve merge conflicts
- add conditional on this dependencies: `<%_ if (databaseType === 'sql') { _%>`

Can you do it plz @Tcharl ?

Tcharl:
Sorry, totally forgot this one: will do it today :-)

pascalgrimaud:
Thanks @Tcharl

Tcharl:
Restested, it seems that another change introduced fixed the behavior: diff is working as expected with h2 without any change on the current codebase.

PR #11335, opened 2020-02-18
Correct Mistake in Adding Support for [skip ci] in GitHub Actions

SudharakaP:
This corrects a mistake I've made in https://github.com/jhipster/generator-jhipster/pull/11304

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -37,7 +37,7 @@ jobs:
     applications:
         name: ${{ matrix.app-type }}
         runs-on: ${{ matrix.os }}
-        if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.pull_request.title, '[skip ci]') && !contains(github.pull_request.title, '[ci skip]')"
+        if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.event.pull_request.title, '[skip ci]') && !contains(github.event.pull_request.title, '[ci skip]')"
         timeout-minutes: 40
         strategy:
             fail-fast: false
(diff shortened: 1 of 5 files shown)


SudharakaP:
Sorry, I've corrected a small mistake I've made in my last PR related to this; this will take care of the pull requests for GitHub Actions, but for Azure Pipelines it seems that PRs still trigger the builds; this is due to two reasons; 

1) The environment variable `$(Build.SourceVersionMessage)` has a weird value of the form `Merge 25dfa26c51310ceef36281375dc9394fb4abdcf1 into 87741f64131e66d80f832789ac20b9f7cbcce421
` in the case of PRs. This is pointed out at; https://github.com/Microsoft/azure-pipelines-agent/issues/858#issuecomment-557869509 as well. 

2) I managed to create a workaround using a different environment variable and getting the commit message using `git log --format=%B -n 1 ${SYSTEM_PULLREQUEST_SOURCECOMMITID}` but as [I've pointed out earlier](https://github.com/jhipster/generator-jhipster/pull/11304#issuecomment-585535791) it seems we have to add the condition to every single step. The workaround @pascalgrimaud suggested by adding `exit 0` only works for the current step and doesn't prevent from running all the other steps. Here's an example repo I've created to test this out; https://github.com/SudharakaP/test-ci-ski/blob/master/azure-pipelines.yml and a PR run for that workflow is at; https://dev.azure.com/sudharakap/ci-testing/_build/results?buildId=1316&view=logs&j=5264e576-3c6f-51f6-f055-fab409685f20.

pascalgrimaud:
thanks @SudharakaP
this rules should be included directly to Azure Pipelines instead of here...

SudharakaP:
@pascalgrimaud : I agree as well. Let us revert the previous change I made to Azure pipelines for now. It's now so bad after all, as for a PR Azure Pipelines will run the tests once; it won't run when merging it (as long as [skip ci] is there). :smile:

PR #11346, opened 2020-02-19
Ingress with nginx: regular expressions in the path

lsorba:
By default regular expressions are not supported by the nginx controller (commonly used), it  transforms the `/.*` from the gateway path (or any service `/services/uaa/.*`) to 

```
location /.* {
  ...
}
```
which is wrong.

It should be transformed to 
```
location ~* "^/.*" {
  ...
}
```

> This can be enabled by setting the nginx.ingress.kubernetes.io/use-regex annotation to true (the default is false).

See: https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/

We either have to add the annotation but then it adds a **dependency to nginx which is bad**.
```
  annotations:
    nginx.ingress.kubernetes.io/use-regex: "true"
```

or remove the wildcard match in the path, suggested in my pull request.

Tested successfully with
```
NGINX Ingress controller
  Release:       0.26.1
  Build:         git-2de5a893a
  Repository:    https://github.com/kubernetes/ingress-nginx
  nginx version: openresty/1.15.8.2
```


xetys:
Oh, thank goodness, I faced that one so many times, but I forgot to open an issue all the time.

This PR LGTM

Issue #10641, opened 2019-10-18
Format Java code with prettier-java


yotamshapira:
##### **Overview of the feature request**

Now that [prettier-java](https://github.com/jhipster/prettier-java) is available, it should be possible to integrate it the generator-jhipster project.

##### **Motivation for or Use Case**

As @jdubois said in a similar issue (see #6379):

> In the end, we would win a lot of time: no need to worry anymore about formatting everything so that the end-result is good. We could focus on having some clean code inside the template, instead. Too often, our templates are unreadable because of the output we want to have.

Supporting Java would benefit a lot of users, make it easier to merge PRs and standardize code.

##### **Related issues or PR**

Continuing the discussion from #6379.

-   [X] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
I didn't follow closely the work done by the crazy prettier team but it will be done when it's ready

I'll let @clement26695 doing an official answer

What could easily done without big impact:
- add prettier-java as dependency
- add new task in package.json file, so the user can run it to format all Java files
- no webhook for now -> it can be added later

What do you think @clement26695 ?

murdos:
I would really like to use prettier-java in JHipster (and on other projects) too, and that's the reason I reported a few issues on prettier-java resulting from tests on real code coming from JHipster, and also had a nice discussion with @clement26695 on the subject.

I think I would use a different strategy: 
1. Enable formatting of java templates on the fly during generation, like any other code supported by prettier (typescript, json, markdown, ...)
2. Wait a bit for feedback, and then enable auto formatting on pre-commit hook in the generated project (this will require moving lintstaged and husky from `client` to `common` subgenerator)

Sure, step 1 might have some small impacts (we're only talking about formatting here), but this is how we'll move forward I think. 
If we agree, I can work on a PR.

murdos:
For the record, I just did new tests with prettier-java 0.4.0, and I've identified at least 3 issues blocking (to my opinion) issues, related to:
- https://github.com/jhipster/prettier-java/issues/262
- https://github.com/jhipster/prettier-java/issues/287 (for SecurityConfiguration.java)
- and the resulting formatting of the toString() of entities that is weird

clement26695:
Sorry, I was not available this weekend to answer to this thread. 

[...]

PR #11371, opened 2020-02-22
Use prettier-java for preformatting before writing disk

pascalgrimaud:
1st step for https://github.com/jhipster/generator-jhipster/issues/10641

@murdos : thanks for pointing me in my other PRs. After our lunch yesterday, I gave a try and you're right: it was easier than expected !

I don't think it's a breaking change here as I didn't introduce the hook with Java, only for our generated code. So we won't change the custom code of our users.

What do you think ?
cc @hdurix @murdos @clementdessoude @Shaolans @Hawkurane @jdubois 
cc @jhipster/developers too

_____

### DIFF
--- a/generators/common/templates/.prettierrc.ejs
+++ b/generators/common/templates/.prettierrc.ejs
@@ -10,3 +10,9 @@ arrowParens: avoid
 
 # jsx and tsx rules:
 jsxBracketSameLine: false
+
+# java rules:
+overrides:
+  - files: "*.java"
+    options:
+      tabWidth: 4
--- a/generators/common/templates/.prettierrc.ejs
+++ b/generators/common/templates/.prettierrc.ejs
@@ -10,9 +10,11 @@ arrowParens: avoid
 
 # jsx and tsx rules:
 jsxBracketSameLine: false
+<%_ if (!skipServer) { _%>
 
 # java rules:
 overrides:
   - files: "*.java"
     options:
       tabWidth: 4
+<%_ } _%>
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
         "parse-gitignore": "1.0.1",
         "pluralize": "7.0.0",
         "prettier": "1.19.1",
-        "prettier-plugin-java": "0.7.0",
+        "prettier-plugin-java": "0.7.1",
         "progress": "2.0.3",
         "randexp": "0.5.3",
         "semver": "5.6.0",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
         "url": "git://github.com/jhipster/generator-jhipster.git"
     },
     "scripts": {
-        "prettier:format": "prettier --write \"{,**/}*.{js,json,md,yml}\"",
+        "prettier:format": "prettier --write \"{,**/}*.{js,json,md,yml,java}\"",
         "completion": "tabtab install --name jhipster --auto",
         "lint": "npm run eslint && npm run ejs-lint",
         "lint-fix": "npm run prettier:format && npm run eslint -- --fix",
@@ -71,6 +71,7 @@
         "parse-gitignore": "1.0.1",
         "pluralize": "7.0.0",
         "prettier": "1.19.1",
+        "prettier-plugin-java": "0.7.0",
         "progress": "2.0.3",
         "randexp": "0.5.3",
         "semver": "5.6.0",
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -192,6 +192,13 @@ module.exports = class extends BaseBlueprintGenerator {
             type: String
         });
 
+        // This adds support for a `--skip-prettier-java` flag
+        this.option('skip-prettier-java', {
+            desc: 'Skip the prettier-java pre-formatting at generation',
+            type: Boolean,
+            defaults: false
+        });
+
         this.skipClient = this.configOptions.skipClient = this.options['skip-client'] || this.config.get('skipClient');
         this.skipServer = this.configOptions.skipServer = this.options['skip-server'] || this.config.get('skipServer');
         this.skipUserManagement = this.configOptions.skipUserManagement =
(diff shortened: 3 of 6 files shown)


clementdessoude:
LGTM

murdos:
Before merging this I think we should solve the issue of the formatting of toString() methods, related to the discussion around https://github.com/jhipster/generator-jhipster/pull/11278

pascalgrimaud:
@murdos : yes, that's why I put 'on hold' label

PierreBesson:
It's not a breaking change but I think it will introduce a huge number of minor formatting changes for people who upgrade so that's why I think it's better if we keep this for v7. However this can be the first merged PR of the v7 branch !

murdos:
> @murdos : yes, that's why I put 'on hold' label

Oh, sorry, I had not seen the label.

mshima:
I am in favor of merging this with a cli flag.

pascalgrimaud:
@mshima : do you suggest a new flag `--prettier` ? 
When using `jhipster --prettier`, it will format all generated java file. Otherwise, it will generate java like it is in template

jdubois:
The flag is a great idea, but I would do the opposite: have Prettier by default, and a flag to disable it in case of trouble.

deepu105:
LGTM. But I'm also in favour of doing it in v7 rather than adding more flags

jdubois:
I don't want to wait... Could we do a test on the sample project, in a specific branch, to see what changes? If that's only formatting it shouldn't too annoying, or so I hope!

pascalgrimaud:
@jdubois : for my current customer, we use prettier-java (with pre commit hook) for 2 months now and it works very well. I encounter only 2 minors issues:
- the import is different than IntelliJ but it's so minor and now, we don't care about it
- the toString is not really readeable, that's why I suggested to add `//prettier-ignore` to this method but we didn't manage to agree - see https://github.com/jhipster/generator-jhipster/pull/11278

I was not alone, @hdurix and @murdos did the same.

@deepu105 : the v7 won't be released before at least 6 months, maybe more... as our roadmap is not clear enough (see https://github.com/jhipster/generator-jhipster/issues/10958). We have too much items here and no work in progress.

About this PR, we didn't introduce any break change, that's why I'm suggesting this.
The break change would be to introduce the pre commit hook which will format all existing / custom code of our users, and I didn't do it here. We'll add this pre commit hook only for v7.

It's similar to **manually** format this kind of things:
- https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs#L848
- https://github.com/jhipster/generator-jhipster/blob/master/generators/entity-server/templates/src/main/java/package/service/mapper/BaseEntityMapper.java.ejs#L36-L38

mshima:
> @mshima : do you suggest a new flag `--prettier` ?

I was thinking about `—prettier-java`. Only prettier can be confusing.
But any flag would work for me. `—experimental` too.

> When using `jhipster --prettier`, it will format all generated java file. Otherwise, it will generate java like it is in template

Exactly.

pascalgrimaud:
I added `--skip-prettier-java` instead, so we'll have it by default

deepu105:
Ok lets do it

pascalgrimaud:
Before merging this, we need to discuss about the `toString` formatting.
We need to make a choice:

1) use the new formatting provided by `prettier-java`

```java
    @Override
    public String toString() {
        return (
            "User{" +
            "login='" +
            login +
            '\'' +
            ", firstName='" +
            firstName +
            '\'' +
            ", lastName='" +
            lastName +
            '\'' +
            ", email='" +
            email +
            '\'' +
            ", imageUrl='" +
            imageUrl +
            '\'' +
            ", activated='" +
            activated +
            '\'' +
            ", langKey='" +
            langKey +
            '\'' +
            ", activationKey='" +
            activationKey +
            '\'' +
            "}"
        );
    }
```

2) add `// prettier-ignore` just before the toString method, so it will be similar to what IntelliJ will generate.

```java
    // prettier-ignore
    @Override
    public String toString() {
        return "User{" +
            "login='" + login + '\'' +
            ", firstName='" + firstName + '\'' +
            ", lastName='" + lastName + '\'' +
            ", email='" + email + '\'' +
            ", imageUrl='" + imageUrl + '\'' +
            ", activated='" + activated + '\'' +
            ", langKey='" + langKey + '\'' +
            ", activationKey='" + activationKey + '\'' +
            "}";
    }
```

3) use apache-commons `ToStringBuilder`

```java
    @Override
    public String toString() {
        return new ToStringBuilder(this)
            .append("firstName", firstName)
            .append("lastName", lastName)
            .append("email", email)
            .toString();
    }
```

generators/common/templates/.prettierrc.ejs
```diff
@@ -10,3 +10,9 @@ arrowParens: avoid
 
 # jsx and tsx rules:
 jsxBracketSameLine: false
+
+# java rules:
```
murdos: This should be added conditionally: it's not useful if skipServer is true

package.json
```diff
@@ -71,6 +71,7 @@
         "parse-gitignore": "1.0.1",
         "pluralize": "7.0.0",
         "prettier": "1.19.1",
+        "prettier-plugin-java": "0.7.0",
```
DanielFran: Update version to 0.7.1?

pascalgrimaud: thanks, done

pascalgrimaud:
@jdubois : I'll change my PR to have `—prettier-java` instead. We won't have prettier-java by default as some of us are afraid of breaking change.

Then, after my change, this PR will be ready to be merged.

jdubois:
I agree, but if it's not by default, people won't use it... Let's put it that way for a couple of months, and we need to raise awareness on this.
Some blog posts, documentation, conference talks should be good -> we need to team up on that

pascalgrimaud:
@murdos : is this last version ok for you ?

pascalgrimaud:
Thanks ! I'll update the documentation with this new flag `--prettier-java`

PR #11410, opened 2020-02-29
Use SUPPORTED_CLIENT_FRAMEWORKS in templates

vivekmore:

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -86,6 +86,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 // Make constants available in templates
                 this.MAIN_SRC_DIR = constants.CLIENT_MAIN_SRC_DIR;
                 this.TEST_SRC_DIR = constants.CLIENT_TEST_SRC_DIR;
+                this.ANGULAR = constants.SUPPORTED_CLIENT_FRAMEWORKS.ANGULAR;
                 const configuration = this.getAllJhipsterConfig(this, true);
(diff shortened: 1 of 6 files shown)


generators/client/index.js
```diff
@@ -82,10 +82,11 @@ module.exports = class extends BaseBlueprintGenerator {
                 }
             },
 
-            setupClientconsts() {
+            setupClientConstants() {
```
murdos: This one is a breaking change. You should add it to https://github.com/jhipster/generator-jhipster/pull/10788 instead ;-)

generators/common/index.js
```diff
@@ -49,13 +49,14 @@ module.exports = class extends BaseBlueprintGenerator {
                 this.checkInvocationFromCLI();
             },
 
-            setupConsts() {
+            setupConstants() {
```
murdos: Breaking change too

generators/client/index.js
vivekmore: Ok I'll gladly revert and move these changes to the suggested PR.

These methods didn't appear to be used anywhere, hence I didn't think it'd be an issue.

Am I correct in understanding that you're saying it'll break blueprints and modules extending these core sub-generators?

murdos: Thanks! Yes, you're correct: these methods could be used externally and thus should be considered as public API.

vivekmore: Cool! 
I've moved the breaking changes to https://github.com/jhipster/generator-jhipster/pull/10788

Issue #11403, opened 2020-02-28
Reactive doesn't work with all NoSQL databases


mraible:
##### **Overview of the issue**

After experiencing [some failing tests](https://github.com/jhipster/generator-jhipster/pull/11223#issuecomment-589893481) when integrating Spring Cloud Gateway, I discovered it was caused by using Couchbase for the DB. If I switch to Mongo, it works.

Here's the `.yo-rc.json` that causes failing tests with `./mvnw verify -P '!webpack,dev'`. Note that it fails for both `monolith` and `gateway` application types.

```
{
  "generator-jhipster": {
    "applicationType": "monolith",
    "reactive": true,
    "baseName": "reactiveGateway",
    "packageName": "io.github.jhipster.gateway",
    "packageFolder": "io/github/jhipster/gateway",
    "authenticationType": "jwt",
    "cacheProvider": "no",
    "enableHibernateCache": false,
    "websocket": false,
    "databaseType": "couchbase",
    "devDatabaseType": "couchbase",
    "prodDatabaseType": "couchbase",
    "searchEngine": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "maven",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": ["en", "es"],
    "testFrameworks": ["protractor"],
    "serverPort": "8080",
    "jhiPrefix": "jhi",
    "jhipsterVersion": "6.7.1",
    "messageBroker": false,
    "jwtSecretKey": "NWE0MjUyY2Y3ZjExNDBlMTZhNDQ2NjQxYjkxYWZmNDA2OWRlZDBiZWY2M2Q3M2IxYjYzNmZiMmNkN2QxMzUwMTg0NmQ0ZTQwOTFiZjc3YTdjNjEyODM0YmEzZjFmODgwMzlkMjUwYWFlODg4MGE0NzQ5NDY5ZTIzOTU3NGIxZDE=",
    "embeddableLaunchScript": false,
    "useSass": true,
    "clientPackageManager": "npm",
    "clientFramework": "angularX",
    "clientTheme": "none",
    "creationTimestamp": 1582856430345,
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "blueprints": []
  }
}
```

Failing tests are:

```
[ERROR] Failures: 
[...]

PR #11223, opened 2020-02-01
Replace Zuul with Spring Cloud Gateway for Reactive Microservices

mraible:
-   Please make sure the below checklist is followed for Pull Requests.

-   [x] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #11417, opened 2020-03-02
Fix Couchbase Reactive

tchlyah:
Fixes https://github.com/jhipster/generator-jhipster/issues/11403


azure-pipelines.yml
```diff
@@ -85,6 +85,18 @@ jobs:
...
+                  JHI_ENTITY: none
+                  JHI_PROTRACTOR: 1
+              webflux-couchbase-session:
+                  JHI_APP: webflux-couchbase-session
+                  JHI_ENTITY: none
+                  JHI_PROTRACTOR: 1
+              webflux-couchbase-oauth2:
+                  JHI_APP: webflux-couchbase-oauth2
+                  JHI_ENTITY: none
+                  JHI_PROTRACTOR: 1
```
pascalgrimaud: We are above the limit of 10 here, so it will increase the build time.

The best would be to mix with the other webflux builds, and add new builds at https://github.com/hipster-labs/jhipster-daily-builds so these config will be tested each day. So we'll stay with 10 builds.

mraible: @pascalgrimaud I added a bunch for reactive tests and [two more for reactive gateway](https://github.com/jhipster/generator-jhipster/pull/11223). I could move them, but I prefer reactive to be a first-class citizen.

pascalgrimaud: I was suggesting this because Couchbase is not used a lot. So launch Couchbase + Reactive builds for each PR is not really usefull.

We need to remember that the more builds we have, the more random failures... So restarting pull request can become a pain.

We have more and more options: testing all options for each PR is impossible. We need to make a choice. That's why the Daily Builds project is here ! :-)

Anyway, what you did is a good idea : testing these configs during the PR, when it's in Work In Progress, then once the PR is merged, we can move it to Daily Builds.

tchlyah: This is exactly why I added the tests here, as soon as you validate the PR, I'll keep just one, or if you prefer I'll remove the 3 tests.

I'll do the PR for daily builds ASAP

pascalgrimaud: ok then. Can you just resolve the problem with lint plz ?

tchlyah: Sorry, I missed it! Just fixed it

PR #11447, opened 2020-03-07
Enable prettier for .jhipster/* and use this.fs for read and verify.

mshima:

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -19,7 +19,6 @@
 /* eslint-disable consistent-return */
 const chalk = require('chalk');
 const _ = require('lodash');
-const shelljs = require('shelljs');
 const pluralize = require('pluralize');
 const jhiCore = require('jhipster-core');
 const prompts = require('./prompts');
@@ -214,8 +213,9 @@ class EntityGenerator extends BaseBlueprintGenerator {
                         context.clientRootFolder = context.microserviceName;
                     }
                 }
+                // TODO 7.0 context.filename = this.destinationPath(context.filename);
                 context.filename = `${context.jhipsterConfigDirectory}/${context.entityNameCapitalized}.json`;
-                if (shelljs.test('-f', context.filename)) {
+                if (this.fs.exists(context.filename)) {
                     this.log(chalk.green(`\nFound the ${context.filename} configuration file, entity can be automatically generated!\n`));
                     context.useConfigurationFile = true;
                 }
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -1548,12 +1548,11 @@ module.exports = class extends Generator {
      */
     registerPrettierTransform(generator = this) {
         // Prettier is clever, it uses correct rules and correct parser according to file extension.
-        let prettierFilter;
+        let filterPatternForPrettier = '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml}';
         if (this.prettierJava) {
-            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
-        } else {
-            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
+            filterPatternForPrettier = '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}';
         }
+        const prettierFilter = filter(['.yo-rc.json', filterPatternForPrettier], { restore: true });
         // this pipe will pass through (restore) anything that doesn't match typescriptFilter
         generator.registerTransformStream([prettierFilter, prettierTransform(prettierOptions), prettierFilter.restore]);
     }
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -1550,9 +1550,9 @@ module.exports = class extends Generator {
         // Prettier is clever, it uses correct rules and correct parser according to file extension.
         let prettierFilter;
         if (this.prettierJava) {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
         } else {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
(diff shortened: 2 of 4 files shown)


mshima:
@MathieuAA, about this test case that is failing:
```
getExistingEntities
     when entities change on-disk
           returns an up-to-date state
```
Is this test really needed?
I think mixing mem-fs read/write with disk read/write is an error.

MathieuAA:
I'll check tomorrow

generators/generator-base-private.js
```diff
@@ -1550,9 +1550,9 @@ module.exports = class extends Generator {
         // Prettier is clever, it uses correct rules and correct parser according to file extension.
         let prettierFilter;
         if (this.prettierJava) {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '.jhipster/*.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
```
murdos: Does this expression includes `.jhipster/modules/jhi-hooks.json` ? I think it should be formatted too.
The following expression might be a bit more inclusive:
```suggestion
            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
```

generators/generator-base.js
```diff
@@ -1301,22 +1303,28 @@ module.exports = class extends PrivateBase {
...
 
+        // TODO 7.0 this.destinationPath(JHIPSTER_CONFIG_DIR);
         return shelljs
             .ls(path.join(JHIPSTER_CONFIG_DIR, '*.json'))
             .reduce((acc, file) => {
                 try {
-                    const definition = jhiCore.readEntityJSON(file);
+                    const definition = this.fs.readJSON(file);
+                    if (recreate) {
+                        // Trigger an write to keep the file on mem-fs to trigger prettier.
```
mshima:
```suggestion
                        // Execute a write operation to set the file as modified on mem-fs to trigger prettier.
```

generators/generator-base-private.js
```diff
@@ -1550,9 +1550,9 @@ module.exports = class extends Generator {
         // Prettier is clever, it uses correct rules and correct parser according to file extension.
         let prettierFilter;
         if (this.prettierJava) {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
         } else {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '.jhipster/*.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
```
mshima:
```suggestion
            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
```

pascalgrimaud:
I'd prefer to merge this after the v6.8.0 if you agree

generators/app/index.js
```diff
@@ -542,7 +542,7 @@ module.exports = class extends BaseBlueprintGenerator {
                     this.configOptions.skipComposeEntity = true;
                     const options = this.options;
                     const configOptions = this.configOptions;
-                    this.getExistingEntities().forEach(entity => {
+                    this.getExistingEntities(true).forEach(entity => {
```
MathieuAA: This makes little sense to have a Boolean parameter for such a function... Plus, it's a code-smell. Does it deserve its own function?

generators/generator-base.js
```diff
@@ -1301,22 +1303,28 @@ module.exports = class extends PrivateBase {
     /**
      * get sorted list of entities according to changelog date (i.e. the order in which they were added)
      */
-    getExistingEntities() {
+    getExistingEntities(recreate = false) {
```
MathieuAA: Recreate? What's its use?

generators/entity/index.js
```diff
@@ -214,8 +213,9 @@ class EntityGenerator extends BaseBlueprintGenerator {
                         context.clientRootFolder = context.microserviceName;
                     }
                 }
+                // TODO 7.0 context.filename = this.destinationPath(context.filename);
```
MathieuAA: Why not setDestinationPath ?

MathieuAA:
@Mshima, you're right to remove the test. I don't think it belongs to this project

generators/generator-base.js
mshima: core writes, directly to disk.
To make prettier work it needs to load the file and modify it on mem-fs.
this.fs.{exists|read} loads it, this.append sets it as modified.

maybe the var name has to be changed.

generators/entity/index.js
mshima: destinationPath is not a setter, it’s a path calculation.
Every yeoman-generator built in method uses destinationRoot as working directory.
destinationPath adds root to the path if the path isn’t absolute.

generators/app/index.js
mshima: See bellow response.
It’s possible to remove this parameter and always set the file as modified on mem-fs.
I was removing overhead, there isn’t much.

mshima:
@MathieuAA can you take another look at this one?

generators/generator-base-private.js
```diff
@@ -1550,9 +1550,9 @@ module.exports = class extends Generator {
         // Prettier is clever, it uses correct rules and correct parser according to file extension.
         let prettierFilter;
         if (this.prettierJava) {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}'], { restore: true });
         } else {
-            prettierFilter = filter(['.yo-rc.json', '{,**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
+            prettierFilter = filter(['.yo-rc.json', '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml}'], { restore: true });
```
MathieuAA:
```javascript
        let extensionFilterForPrettier = '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml}';
        if (this.prettierJava) {
            extensionFilterForPrettier =  '{,**/,.jhipster/**/}*.{md,json,ts,tsx,scss,css,yml,java}';
        }
        const prettierFilter = filter(['.yo-rc.json', extensionFilterForPrettier], { restore: true });
```

MathieuAA: I tend to avoid if-else conditionals in favor of simple ifs, makes the code easier to read and doesn't created two paths in the code

Issue #11454, opened 2020-03-10
Error deploying to Azure Spring Cloud locally


yucwan:
##### **Overview of the issue**

Deploying application...
× Deployment failed!
 Error: Command failed: az spring-cloud app deploy --resource-group yuchensp --service yuchensp --name yucjhapp --jar-path target/*.jar
Command group 'spring-cloud' is in preview. It may be changed/removed in a future release.
WARNING: This command usually takes minutes to run. Add '--verbose' parameter if needed.
WARNING: [1/3] Requesting for upload URL
WARNING: [2/3] Uploading package to blob
ERROR: The command failed with an unexpected error. Here is the traceback:

ERROR: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'target/*.jar'
Traceback (most recent call last):
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\knack\knack\cli.py", line 206, in invoke
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\commands\__init__.py", line 608, in execute
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\commands\__init__.py", line 666, in _run_jobs_serially
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\commands\__init__.py", line 659, in _run_job
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\six\six.py", line 703, in reraise
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\commands\__init__.py", line 636, in _run_job
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\commands\__init__.py", line 306, in __call__
  File "C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-3pnsb50r\azure-cli-core\azure\cli\core\__init__.py", line 493, in default_command_handler
  File "C:\Users\yucwan\.azure\cliextensions\spring-cloud\azext_spring_cloud\custom.py", line 334, in app_deploy
    True)
[...]

PR #11459, opened 2020-03-10
Improve Gradle support for Azure Spring Cloud

jdubois:
See #11454


generators/azure-spring-cloud/index.js
```diff
@@ -417,19 +417,22 @@ for more detailed information.`
 
                 const done = this.async();
                 this.log(chalk.bold('\nDeploying application...'));
-
+                let buildDir = 'target';
```
pascalgrimaud: there should be some method or constant to improve this part, but should be ok for now

Issue #11274, opened 2020-02-10
jhipster ci-cd does not generate a ci-cd pipelines whith --skip-server option


maznag:
##### **Overview of the issue**

No ci-cd pipelines generated with "jhipster ci-cd" when we generate only a frontend app with --skip-server option

##### **Motivation for or Use Case**

Generate a ci-cd pipeline only for frontend applications.

##### **Reproduce the error**

* generate a new app with "jhipster --skip-server --db=sql --auth=jwt"
* generate a GitLab ci-cd pipeline for example with "jhipster ci-cd"
* check the generated file .gitlab-ci.yml

Only these lines are generated:

stages:
- check
- build
- test
- analyze
- package
- release
- deploy

##### **JHipster Version(s)**
V6.7.0

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "nativeLanguage": "fr"
    },
    "jhipsterVersion": "6.7.0",
    "applicationType": "monolith",
    "baseName": "TestCICD",
    "useSass": true,
    "clientPackageManager": "npm",
    "clientFramework": "angularX",
    "clientTheme": "none",
    "clientThemeVariant": "",
    "authenticationType": "jwt",
    "cacheProvider": "no",
    "databaseType": "sql",
    "devDatabaseType": "sql",
    "prodDatabaseType": "sql",
    "creationTimestamp": 1581324311414,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "enableTranslation": true,
    "nativeLanguage": "fr",
    "languages": ["fr"],
    "blueprints": [],
    "skipServer": true
  }
}

</pre>
</details>

##### **Environment and Tools**

java version "12.0.2" 2019-07-16
Java(TM) SE Runtime Environment (build 12.0.2+10)
Java HotSpot(TM) 64-Bit Server VM (build 12.0.2+10, mixed mode, sharing)

git version 2.13.1

node: v11.14.0

npm: 6.13.7

yeoman: 3.1.1

Docker version 19.03.1, build 74b1e89

docker-compose version 1.24.1, build 4667896b

vishal423:
I believe we support this flow in the GitHub Actions. @maznag, can you try that?

maznag:
[...]

PR #11461, opened 2020-03-10
GitLab-ci: add sonar-analyze stage for skipServer option

maznag:
Related issue #11274 
Related PR #11438

### DIFF
--- a/generators/ci-cd/templates/.gitlab-ci.yml.ejs
+++ b/generators/ci-cd/templates/.gitlab-ci.yml.ejs
@@ -258,6 +258,9 @@ cache:
 stages:
     - build
     - test
+    <%_ if (cicdIntegrations.includes('sonar')) { _%>
+    - analyze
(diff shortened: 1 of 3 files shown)


avdev4j:
Add sonar-scanner when skipping server generation is a good idea imho.
We should support sonar analysis for other CI than Gitlab.

generators/ci-cd/templates/.gitlab-ci.yml.ejs
```diff
@@ -279,6 +282,18 @@ webpack-test:
         paths:
             - build/test-results
         expire_in: 1 day
+<%_ if (cicdIntegrations.includes('sonar')) { _%>
```
avdev4j: we should handle the case !skipServer isnt'it ?
We already have a sonar plugin in both maven and gradle config files. We need to check if server is not skipped so we have to call wether maven or gradle sonar stage.

PR #11478, opened 2020-03-14
Entity subgen: Refactoring of determination of entity key type

murdos:
Content of this PR:
- Factored the definition of Java key type of the entity (string or numeric, depending on datatabase type, auth and use of mapsId) 
- Factored the definition of TypeScript key type of the entity (string or numeric, depending on datatabase type, auth, ...) 
- Factored, for each relationship, the computing of the other entity primary key type

### DIFF
--- a/generators/entity-client/index.js
+++ b/generators/entity-client/index.js
@@ -37,6 +37,20 @@ module.exports = class extends BaseBlueprintGenerator {
     }
 
     // Public API method used by the getter and also by Blueprints
+    _configuring() {
+        return {
+            setup() {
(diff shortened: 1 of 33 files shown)


pascalgrimaud:
@murdos : can you tell me if there is some impact for blueprint plz ?

mshima:
I think changes to entity/index.js and generator-base-private.js aren’t safe.
- hasUserField isn’t exposed anymore.
- pkType isn’t exposed anymore.
- generateEntityClientFields is in generator-base-private. But since it’s used on a template. It think it should be moved to generator-base.

murdos:
I've re-exposed `hasUserField` and ` pkType` for backward compatibility with a deprecated comment, and reverted signature change of `generateEntityClientFields()`, it should be safe now!

PR #11498, opened 2020-03-23
add bulgarian language

stoyants:
add translations for bulgarian language support


atomfrede:
Thanks @stoyants and congrats to your first contribution!

generators/languages/templates/src/main/webapp/i18n/bg/error.json
```diff
@@ -0,0 +1,14 @@
+{
+    "error": {
+        "title": "Страница за грешка!",
+        "http": {
+            "400": "Неправилна заявка.",
+            "403": "Не сте упълномощени да получите достъп до тази страница.",
+            "404": "Тази страница не съществува.",
+            "405": "Използваният от вас HTTP глагол не се поддържа от този URL.",
```
martin-g: "`за` този URL" sounds better to me

generators/languages/templates/src/main/webapp/i18n/bg/error.json
```diff
@@ -0,0 +1,14 @@
...
+            "500": "Вътрешна грешка в сървъра."
+        },
+        "concurrencyFailure": "Друг потребител е променил данните едновременно с вас. Вашите промени са отхвърлени.",
+        "validation": "Грешка при валидиране на сървъра."
```
martin-g: "Грешка при валидиране на данните на сървъра." ?!

generators/languages/templates/src/main/webapp/i18n/bg/global.json.ejs
```diff
@@ -0,0 +1,163 @@
...
+                },
+                "confirmpassword": {
+                    "required": "Полето за потвърждаване на вашата парола е задължително.",
+                    "minlength": "Полето за потвърждаване на вашата парола трябва да съдържа поне 4 символа.",
+                    "maxlength": "Полето за потвърждаване на вашата парола не може да бъде по-дълго от 50 символа."
+                },
+                "email": {
+                    "required": "Вашият имейл е задължителен.",
+                    "invalid": "Вашият имейл е невалиден.",
+                    "minlength": "Вашият имейл трябва да съдържа поне 5 символа.",
```
martin-g: Not related to the BG translation: I think it would be better if `5` and `50` are parameters, e.g. `{{min}}` and `{{max}}`

generators/languages/templates/src/main/webapp/i18n/bg/metrics.json
```diff
@@ -0,0 +1,102 @@
+{
+    "metrics": {
+        "title": "Метрики на приложението",
+        "refresh.button": "Обновяване",
```
martin-g: Обнови ?!

generators/languages/templates/src/main/webapp/i18n/bg/global.json.ejs
martin-g: Hm. Maybe this won't be that easy. I guess the DB fields have the same length too.

generators/languages/templates/src/main/webapp/i18n/bg/error.json
stoyants: Accepted, will change it to "за ..."

generators/languages/templates/src/main/webapp/i18n/bg/global.json.ejs
stoyants: I guess someone has to provide "min" and "max parameters as well ;-)

generators/languages/templates/src/main/webapp/i18n/bg/error.json
stoyants: OK, I was also thinking about more verbose explanation, in fact changed it multiple times back and forth. At the end decided to keep it close to the original message in English.
But in Bulgarian maybe should be more specific and not to put users in doubt.
Accepted, will add "на данните".

generators/languages/templates/src/main/webapp/i18n/bg/metrics.json
stoyants: OK, this is a complicated one. I'm not a philologist, so please excuse my humble explanation.
As a source of translation I used English. In English there is (nearly) no difference between "Second Person Singular" and "Second Person Plural". So it is difficult to decide should application provide "Polite" or "Familiar" form of communication.
In Bulgarian language, like in German for example, there is a clear distinction between the two forms (Geman - "du" and "Sie", Bulgarian - Вие" and "ти"). I decided for the "Polite" form. 
And I did my best to keep it consistent all over the messages. This is the way application "talks" to the user.
And here we come to the translation of the buttons. Because more or less these are the activities, that application offers to the user. Or the way user "talks" to the application.
English again has the advantage  - the same word is a Noun and a Verb, even a Verb in its imperative form. But I don't find it nice, if application "talks" in a polite form to put imperative Verbs as button labels. We don't give orders to someone, who is polite to us, do we? 
For this reason I decided to use "Verbal nouns" ("отглаголни съществителни") - Open = "Отваряне", Save = "Запазване", Edit = "Редактиране", and so on.
For this reason, to keep some consistency Refresh = "Обновяване".
I suggest to keep it as it is. Or ... change all buttons to imperative verbs. (Отвори, Изтрий и т.н.)

PR #11500, opened 2020-03-23
Fix Jwts usage to get rid of the deprecation warning

dschulz:
jjwt project has deprecated the `Jwts.parser()` method in favor of `Jwts.parserBuilder()` some months ago. This patch corrects the relevant code to use the updated versions and avoids the deprecation warning.

More info: jwtk/jjwt#499

### DIFF
--- a/generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
@@ -100,8 +100,9 @@ public class TokenProvider {
     }
 
     public Authentication getAuthentication(String token) {
-        Claims claims = Jwts.parser()
+        Claims claims = Jwts.parserBuilder()
             .setSigningKey(key)
+            .build()
             .parseClaimsJws(token)
             .getBody();
 
@@ -117,7 +118,7 @@ public class TokenProvider {
 
     public boolean validateToken(String authToken) {
         try {
-            Jwts.parser().setSigningKey(key).parseClaimsJws(authToken);
+            Jwts.parserBuilder().setSigningKey(key).build().parseClaimsJws(authToken);
             return true;
         } catch (JwtException | IllegalArgumentException e) {
             log.info("Invalid JWT token.");


dschulz:
The changes are trivial but I'm afraid I'll need some guidance to comply with the requirements.

generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
```diff
@@ -100,8 +100,9 @@ public class TokenProvider {
     }
 
     public Authentication getAuthentication(String token) {
-        Claims claims = Jwts.parser()
+        Claims claims = Jwts.parserBuilder()
```
bdemers: Ideally, you want to reuse the parser `parserBuilder()...build()` when possible. (and only incur the cost of construction once)
Maybe stick it in the PostConstruct above?

bdemers:
Note, reusing the `parser` could be done before this change too, but you could mutate the Parser when not using the `parserBuilder()` (the mutable methods will be removed in future versions of JJWT)

generators/server/templates/src/main/java/package/security/jwt/TokenProvider.java.ejs
dschulz: You're right @bdemers, I guess I was too focused in the deprecation warning and absolutely not in the performance.

jdubois:
The performance point is very interesting! Indeed `JwtParser` is supposed to be immutable, so we could reuse it, which would be a very good improvement

jdubois:
@dschulz is that fine to make this PR "ready for review"? It's currently in draft mode.

dschulz:
> @dschulz is that fine to make this PR "ready for review"? It's currently in draft mode.

@jdubois, absolutely. I'll make it ready for review and try to come up with another, better patch after that. I need to take some time examining the jjwt code to be more confident.

pascalgrimaud:
@DanielFran : not for this build, but for this release, you mean ? In this case, yes, I'm short on time to test everything but I'll trust on the daily builds this night

DanielFran:
Yes, for the release of course 😴😂

DanielFran:
@jdubois Did you test this changes?

PR #11540, opened 2020-04-02
Allow to use TLS in prod mode

jdubois:

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1773,7 +1773,7 @@
             </build>
             <properties>
                 <!-- default Spring profiles -->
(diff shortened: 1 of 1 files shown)


jdubois:
@atomfrede it's not the default, it's just that we can now do "mvn -Pprod,tls", and before that wasn't possible.
(and you could do "mvn -Pdev,tls", so that wasn't even consistent)

Can you have a look for Gradle? I'm sorry, I really suck so much at Gradle

atomfrede:
Sure. I already looked at the source, I think we need to add it there too.

jdubois:
Merging as the checks are green and that shouldn't cause any issue.
@atomfrede can you have a look at the Gradle equivalent?

atomfrede:
> Merging as the checks are green and that shouldn't cause any issue.
> @atomfrede can you have a look at the Gradle equivalent?

Yes. No big thing to add.

PR #11545, opened 2020-04-04
Replace jjwt deprecated parser by parserBuilder

caodangtinh:
Generated TokenProvider.java class is using deprecated method from [jjwt](https://github.com/jwtk/jjwt/pull/499). It should replace by their new suggestion method.


murdos:
FYI there's already an open PR with similar changes and an interesting point that has been raised: https://github.com/jhipster/generator-jhipster/pull/11500

caodangtinh:
> FYI there's already an open PR with similar changes and an interesting point that has been raised: #11500

Ok, I will close my PR.

Issue #11533, opened 2020-03-31
In a blueprint, `--prettier-java` can't find the java plugin


danielpetisme:
##### **Overview of the issue**
Use the `--prettier-java` in a JHipster blueprint.
https://github.com/jhipster/jhipster-quarkus/pull/37#issuecomment-606827625

##### **Reproduce the error**

Quite tricky to reproduce since it's very specific to blueprints planning to use the `--prettier-java` option, so I think Jhipster quarkus may be the only one?

##### **Suggest a Fix**

According to https://prettier.io/docs/en/plugins.html Prettier looks for plugin in the same `node_modules` but in a blueprint Prettier and the java plugin are not in the same directory

Within a blueprint, the dependencies are a bit transformed (i'm not a node_module so don't ask me why).
Anyway, in the blueprint node_modules  I can found the `generator-jhipster` dependency which has also couple of dependencies
![Screenshot 2020-03-31 at 10 57 58 PM](https://user-images.githubusercontent.com/593564/78074583-486a2200-73a3-11ea-8ae4-c4b7924d7f0f.png)
You may notice prettier is present as dependency but not the java plugin. 

I can find the java plugin in the blueprint dependencies (I remove the declaration or prettier and java plugin in the blueprint `package.json`).
![Screenshot 2020-03-31 at 10 58 15 PM](https://user-images.githubusercontent.com/593564/78074711-7bacb100-73a3-11ea-8898-06d2ae308481.png)

So  we end up with this structure in a blueprint:
```
node_modules/
  + generator-jhipster/
  |-+node_modules/
     |--+prettier/ <-- the one invoke by `--prettier-java`
  +prettier/
  +prettier-plugin-java
```

I had a deeper look to:
https://github.com/jhipster/generator-jhipster/blob/d3cc5f6b8e021422bbab6d8151172ececb0eaf2d/generators/generator-transforms.js#L44
I tested by updating `generator-transform.js
```
onst prettierTransform = function(defaultOptions) {
    const transform = (file, encoding, callback) => {
        /* resolve from the projects config */
        prettier.resolveConfig(file.relative).then(options => {
            if (file.state !== 'deleted') {
                const str = file.contents.toString('utf8');
                if (!options || Object.keys(options).length === 0) {
                    options = defaultOptions;
                }
                // for better errors
                options.filepath = file.relative;
                this.
                options.plugins = ["/Users/daniel/workspace/jhipster-quarkus/node_modules/prettier-plugin-java"]
[...]

PR #11567, opened 2020-04-09
Register prettier-java plugin.

mshima:
Prettier lookup for plugins at same level node_modules repository.
Workaround prettier-java failing with npm changing it's location.

Fixes https://github.com/jhipster/generator-jhipster/issues/11533

### DIFF
--- a/generators/generator-transforms.js
+++ b/generators/generator-transforms.js
@@ -18,6 +18,7 @@
  */
 const through = require('through2');
 const prettier = require('prettier');
+const prettierJava = require('prettier-plugin-java');
 
 const prettierOptions = {
     printWidth: 140,
@@ -27,7 +28,8 @@ const prettierOptions = {
     // js and ts rules:
     arrowParens: 'avoid',
     // jsx and tsx rules:
-    jsxBracketSameLine: false
+    jsxBracketSameLine: false,
+    plugins: [prettierJava]
 };
 
 const prettierTransform = function(defaultOptions) {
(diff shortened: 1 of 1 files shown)


danielpetisme:
Tested and as is it does not work.
You have updated the `prettierOptions` but they are only used in `prettierFormat` which is... never used: https://github.com/jhipster/generator-jhipster/search?q=prettierFormat&unscoped_q=prettierFormat

But I used what you have made to update `prettierTransform`
```
const prettierTransform = function(defaultOptions) {
    const transform = (file, encoding, callback) => {
        /* resolve from the projects config */
        prettier.resolveConfig(file.relative).then(options => {
            if (file.state !== 'deleted') {
                const str = file.contents.toString('utf8');
                if (!options || Object.keys(options).length === 0) {
                    options = defaultOptions;
                }
                // for better errors
                options.filepath = file.relative;
                options.plugins = [prettierJava]
                const data = prettier.format(str, options);
                file.contents = Buffer.from(data);
            }
            callback(null, file);
        });
    };
    return through.obj(transform);
};
```

The interesting line is `options.plugins = [prettierJava]` which does the trick 🎉 .

I propose to use both of them, the one you already defined in `prettierOptions` and this one to be sure to cover all possible prettier transformation.

WDYT?

mshima:
Updated the PR with a different approach.
I don't know the consequences, but if there is any, I think it's a bug.

danielpetisme:
I tested your latest changes, it works for me!

Issue #11542, opened 2020-04-03
[Reactive] Support R2DBC in entity generator


cbornet:
We now have support for r2dbc in the main generator and entity generation for NoSQL databases.
So remains r2dbc entity generation.

Some things that make it not so easy:
* There's currently no support for query derivation although that should arrive soon : https://github.com/spring-projects/spring-data-r2dbc/issues/282
* r2dbc is not an ORM so the relational code has to be handled explicitly. Maybe the first version could be done without support for relationships.

mraible:
I've added a bug bounty as this seems like a large task that will be difficult to do.

gzsombor:
I would like to work on it - I've already found a couple of issue, which prevented running the reactive app, and login to it.

mraible:
Go for it @gzsombor!

brahimk:
Hello @gzsombor 

Do you want help on this issue ?

thanks.

gzsombor:
@brahimk thanks for asking, there are some progress implementing this - and there is an open PR about
https://github.com/jhipster/generator-jhipster/pull/11742 . 
Currently I'm blocked by three things:
* Spring-boot 2.3 support is not yet merged, so building and running the CI is not possible out of the box. I hope, this can be solved with using the jitpack.io repository.
* Spring-data-r2dbc bug - https://github.com/spring-projects/spring-data-r2dbc/pull/390 - it's not yet merged, and released, unfortunately this error is triggered by our CI suite.
* IRL problems - like lack of time/travel/holiday

mraible:
@gzsombor Are you still blocked? Can we split this up into smaller issue and give you more $$ to make it happen?

murdos:
@mraible : as far as I can tell, this ticket should be closed as resolved, since https://github.com/jhipster/generator-jhipster/pull/11742 has been merged.

mraible:
Fixed by https://github.com/jhipster/generator-jhipster/pull/11742. I'm not sure how to do the bug bounty payout. Maybe @pascalgrimaud can help?

gzsombor:
Yes, sorry, the code has been merged, but there are a couple of open issues, in decreasing importance:
* filtering is not implemented
[...]

PR #11584, opened 2020-04-14
Testcontainers for SQL databases

### LINKED ISSUE #11544: Include Testcontainers in the main generator

I just released a new version of this module: https://github.com/intesys/generator-jhipster-testcontainers. It allows to run integration tests on sql databases with testcontainers. It's very simple (it basically adds a properties file in test resources) and it doesn't impact the existing code. A user can keep running integration tests on H2 or switching to testcontainers simply add the `testcontainers` profile when running tests.

What about including this in the main generator?

We've been using testcontainers for a while and I think it is really helpful. Jhipster already uses this library to run test with other databases. Adding this also for sql database would be a plus for developers. 

https://github.com/jhipster/generator-jhipster/issues/10958 

ecostanzi:
Fixes #11544

Allows to run tests with by enabling testcontainers with the `testcontainers` profile:

```
./mvnw verify -Dspring.profiles.active=testcontainers
```
```
./gradlew integrationTest -Ptestcontainers
```
- [x] testcontainers for sql db
- [x] HibernateTimeZoneIT tests for Mysql and MariaDB
- [ ] r2dbc support
- [x] ci-cd templates


ecostanzi:
I'm looking for an advice. I'm not sure on how to deal with tests.

I added a JHI_TESTCONTAINERS flag to run backend test with testcontainers, but doing so when JHI_TESTCONTAINERS=1, H2 integration tests are not executed by `21-test-backend.sh`. 

An alternative: when JHI_TESTCONTAINERS=1, integration tests run both with h2 and then with testcontainers.

generators/server/templates/src/test/resources/config/application-testcontainers.yml.ejs
```diff
@@ -0,0 +1,93 @@
...
+                cachePrepStmts: true
+                prepStmtCacheSize: 250
+                prepStmtCacheSqlLimit: 2048
+                useServerPrepStmts: true
+        <%_ } _%>
+    jpa:
+        <%_ if (prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb') { _%>
+        database-platform: org.hibernate.dialect.MySQL5InnoDBDialect
+        <%_ } else if (prodDatabaseType === 'postgresql') { _%>
+        database-platform: io.github.jhipster.domain.util.FixedPostgreSQL82Dialect
```
pascalgrimaud, line 68: FixedPostgreSQL10Dialect instead

generators/server/templates/src/test/resources/config/application-testcontainers.yml.ejs
```diff
@@ -0,0 +1,93 @@
...
+        <%_ if (prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb' || prodDatabaseType === 'mssql') { _%>
+            data-source-properties:
+                cachePrepStmts: true
+                prepStmtCacheSize: 250
+                prepStmtCacheSqlLimit: 2048
+                useServerPrepStmts: true
+        <%_ } _%>
+    jpa:
+        <%_ if (prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb') { _%>
+        database-platform: org.hibernate.dialect.MySQL5InnoDBDialect
```
pascalgrimaud, line 66: are you sure ? as we don't have this anymore -> see https://github.com/jhipster/generator-jhipster/pull/11125

generators/server/templates/src/test/resources/config/application-testcontainers.yml.ejs
```diff
@@ -0,0 +1,93 @@
...
+        <%_ } else if (prodDatabaseType === 'postgresql') { _%>
+        database-platform: io.github.jhipster.domain.util.FixedPostgreSQL82Dialect
+        <%_ } else if (prodDatabaseType === 'oracle') { _%>
+        database-platform: org.hibernate.dialect.Oracle12cDialect
```
pascalgrimaud, line 70: maybe same for other database-platform here, except for PSQL

generators/server/templates/pom.xml.ejs
```diff
@@ -409,6 +409,48 @@
...
         </dependency>
 <%_ } _%>
+<%_ if (databaseType === 'sql') { _%>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>test</scope>
+        </dependency>
+<%_ } _%>
+<%_ if (prodDatabaseType === 'mysql') { _%>
```
pascalgrimaud, line 419: you can add this condition to ̀artifactId` only, I think. Something like:

```
       <dependency>
            <groupId>org.testcontainers</groupId>
<%_ if (prodDatabaseType === 'mysql') { _%>
            <artifactId>mysql</artifactId>
<%_ } _%>
<%_ if (prodDatabaseType === 'mariadb') { _%>
            <artifactId>mariadb</artifactId>
<%_ } _%>
            <scope>test</scope>
        </dependency>
```

generators/server/templates/pom.xml.ejs
```diff
@@ -409,6 +409,48 @@
             <scope>test</scope>
         </dependency>
 <%_ } _%>
+<%_ if (databaseType === 'sql') { _%>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
```
pascalgrimaud, line 415: do we need this ? Any way, we need to check with other dependencies as there can be a duplicated here: testcontainers is used for other options

pascalgrimaud:
@ecostanzi :

> An alternative: when JHI_TESTCONTAINERS=1, integration tests run both with h2 and then with testcontainers.

No, don't run both test. Let's just mixt : some builds will run with H2, some other with testcontainers. We can't cover all options :)

For example:
- ngx-default VS react-default
- both have exactly same options for backend
- ngx-default should run with testcontainers and react-default should run with H2
- or, ngx-default should run with H2 and react-default should run with testcontainers

generators/server/templates/gradle/testcontainers.gradle.ejs
```diff
@@ -0,0 +1,53 @@
...
+    <%_ } _%>
+    <%_ if (prodDatabaseType === 'mssql') { _%>
+    testImplementation "org.testcontainers:mssqlserver"
+    <%_ } _%>
+    <%_ if (prodDatabaseType === 'oracle') { _%>
+    testImplementation "org.testcontainers:oracle-xe"
+    <%_ } _%>
+}
+
+if (project.hasProperty('testcontainers')) {
```
atomfrede, line 51: I think you forgot to apply this gradle skript (see https://github.com/intesys/generator-jhipster-testcontainers/blob/master/generators/app/index.js#L95). But maybe having this in a dedicated gradle script is a bit too much?
If we want to keep it separate I would suggest to check for the `testcontainers` property in the main file and apply this script only when the property is present

ecostanzi:
There are some issues with mysql. Apparently without the dialect tests are failing with this error: java.sql.SQLSyntaxErrorException: Unknown table 'SEQUENCES' in information_schema.

If I restore the dialect, HibernateTimeZoneIT tests fail.

I'll try to solve them by using TC_MY_CNF

generators/server/templates/pom.xml.ejs
ecostanzi: I'll try to remove it as soon as all test pass :)

generators/server/templates/gradle/testcontainers.gradle.ejs
ecostanzi: Makes sense. In the module wanted something separate. I'll try to bring everything into the build.gradle.

generators/server/templates/build.gradle.ejs
```diff
@@ -145,6 +145,10 @@ task integrationTest(type: Test) {
     testLogging {
         events 'FAILED', 'SKIPPED'
     }
+    if (project.hasProperty('testcontainers')) {
```
atomfrede, line 148: I think it doesn't make any difference, but I would say let's keep all properties together at the top of the build file.

atomfrede:
@ecostanzi Nice work, thanks. We should also adapt the CI configs (at least for gitlab using docker) we should generate in a way it works out of box on gitlab.com using DinD.

generators/server/templates/build.gradle.ejs
ecostanzi: This works:
```
task integrationTest(type: Test) {
...
}
if (project.hasProperty('testcontainers')) {
    integrationTest {
        environment 'spring.profiles.active', 'testcontainers'
    }
}
```
But this one doesn't:
```
if (project.hasProperty('testcontainers')) {
    integrationTest {
        environment 'spring.profiles.active', 'testcontainers'
    }
}
task integrationTest(type: Test) {
...}
```

```

* Where:
Build file 'C:\Users\ecostanzi\dev\tmp\daigradle\build.gradle' line: 102

* What went wrong:
A problem occurred evaluating root project 'daigradle'.
> Could not find method integrationTest() for arguments [build_danqkclw8b0bg52d6bfrx5o9e$_run_closure6@69a2634e] on root project 'daigradle' of type org.gradle.api.Project.
```

I don't use gradle very often, so I'm open to suggestions :)

generators/server/templates/src/test/resources/config/application-testcontainers.yml.ejs
ecostanzi: @pascalgrimaud I had to restore the dialects to ovverride the one specified in the test `application.yml`. When testcontainers profile is active it has to override the `database-platform: io.github.jhipster.domain.util.FixedH2Dialect` in the default test properties.

WDYT?

generators/server/templates/build.gradle.ejs
atomfrede: Ah okay I see, was not sure about. Let's keep it as is.

ecostanzi:
Quick recap:
- JDBC support. It works, But I spent most of the time on HibernateTimezoneIT. With mysql and mariadb tests pass only if the JVM timezone is set to GMT (`-Duser.timezone=GMT`). When running tests on my laptop (Timezone Europe/Rome) this si what happens: https://github.com/ecostanzi/generator-jhipster/runs/599329024?check_suite_focus=true. 
- R2DBC support. Not working. It is supported in the latest version of testcontainers. But I didn't manage to make it work. Changing the URL https://www.testcontainers.org/modules/databases/r2dbc/ is not enough apparently.
- CI/CD: I added few commented lines on the gitlab-ci.yml templates, trying to understand if something has to be done for travis and circleCI. 

If you have any suggestions on one of the above points that could help me to speed up the work :)

generators/ci-cd/templates/.gitlab-ci.yml.ejs
```diff
@@ -80,6 +80,15 @@ gradle-test:
 
 gradle-integration-test:
     stage: test
+<%_ if (databaseType === 'sql') { _%>
```
atomfrede, line 83: Maybe we can make the out of box experience with gitlab.com better (like we did just now for other dbs which already require testcontainers).
E.g.  check if we run inside docker, start the dind service and use the correct spring profile (so basically not just a comment) what do you think?

atomfrede:
> Quick recap:
> 
>     * JDBC support. It works, But I spent most of the time on HibernateTimezoneIT. With mysql and mariadb tests pass only if the JVM timezone is set to GMT (`-Duser.timezone=GMT`). When running tests on my laptop (Timezone Europe/Rome) this si what happens: https://github.com/ecostanzi/generator-jhipster/runs/599329024?check_suite_focus=true.
> 
>     * R2DBC support. Not working. It is supported in the latest version of testcontainers. But I didn't manage to make it work just with properties file. Changing the URL https://www.testcontainers.org/modules/databases/r2dbc/ is not enough apparently.
Usually the tc maintainers have very good tests. How did you define the database url?
> 
>     * CI/CD: I added few commented lines on the gitlab-ci.yml templates, trying to understand if >something has to be done for travis and circleCI.

I think we can leave it out for now

generators/ci-cd/templates/.gitlab-ci.yml.ejs
SudharakaP: I think this is already added as per https://github.com/jhipster/generator-jhipster/pull/11603. You probably just need to add the extra condition. :smile:

SudharakaP: This is already added as per #11603, you probably just need to add the extra condition. smile

ecostanzi: Since the work on the gitlab-ci.yml has already been done in #11603 I think I can fix it accordingly. 

@SudharakaP Only one thing. I see that DinD has been activated in the `gradle-test` job, but it should probably be moved in the `gradle-integration-test` job, right?

SudharakaP: @ecostanzi : You are right on that. I am doing some further testing for the Gradle side in Gitlab. I'll do a PR today with those changes. :smile:

SudharakaP: @ecostanzi : Done at https://github.com/jhipster/generator-jhipster/pull/11614. Thanks for pointing it out.

generators/ci-cd/templates/.gitlab-ci.yml.ejs
```diff
@@ -80,7 +80,7 @@ gradle-test:
 
 gradle-integration-test:
 <%_ if (insideDocker) { _%>
-    <%_ if (cacheProvider === 'redis' || ['cassandra', 'couchbase', 'neo4j'].includes(prodDatabaseType)) { _%>
+    <%_ if (cacheProvider === 'redis' || databaseType === "sql" || ['cassandra', 'couchbase', 'neo4j'].includes(prodDatabaseType)) { _%>
```
atomfrede, line 83: Very minor thing, but we can simplify it to 

`<%_ if (cacheProvider === 'redis' || ['sql', 'cassandra', 'couchbase', 'neo4j'].includes(databaseType)) { _%>`

I think

ecostanzi:
@atomfrede thank you for the review. 

> Usually the tc maintainers have very good tests. How did you define the database url?

I tried to repace the URL in the `spring.r2dbc.url` using the prefix r2dbc:tc prefix but it doesn't work out of the box. I didn't find the reactive equivalent of jdbc ContainerDatabaseDriver. I should try again, but since there are no sql tests in the webflux workflow of github actions I'll leave this part out of this PR. 

If you agree with the above, the only think that seems to be left is to fix the HibernateTimezoneIT integration test for mysql and mariadb. It works in the pipelines, but only because the container timezone is GMT.

atomfrede:
Totally fine for me. Let's move in small steps!

atomfrede:
> If you agree with the above, the only think that seems to be left is to fix the HibernateTimezoneIT integration test for mysql and mariadb. It works in the pipelines, but only because the container timezone is GMT.

This seems to be fixed, is that right? So let's get it merged I would say :tada:

murdos:
> I should try again, but since there are no sql tests in the webflux workflow of github actions I'll leave this part out of this PR.

FYI there's one now (`webflux-psql`), and I think we'll have more in the future when entity subgen handle R2DBC (#11542) since SQL is the most used database type with JHipster.

murdos:
> Allows to run tests with by enabling testcontainers with the `testcontainers` profile:
> 
> ```
> ./mvnw verify -Dspring.profiles.active=testcontainers
> ```
> 
> ```
> ./gradlew integrationTest -Ptestcontainers
> ```

WDYT of adding the following commented configuration in `application.yml` of test resources, in order to better advertise this feature?
```
spring:
  profiles:
    # Uncomment the following line to enable tests against production database type rather than H2, using Testcontainers
    #active: testcontainers
```

generators/server/templates/src/test/resources/config/application-testcontainers.yml.ejs
```diff
@@ -0,0 +1,87 @@
...
+<%_ } _%>
+<%_ if (prodDatabaseType === 'oracle') { _%>
+#
+# You have to specify an Oracle image name in a classpath file named testcontainers.properties.
+# Follow the instructions at https://www.testcontainers.org/modules/databases/oraclexe/
+<%_ } _%>
+# ===================================================================
+
+spring:
+    datasource:
```
murdos, line 40: Could you add this block conditionally on `!reactive`, and add the `reactive` part you've done, event if it's not working yet?
It would help to understand the error and to fix it in a later PR.

murdos: Also, it seems that `org.testcontainers:r2dbc` maven dependency need to be added in addition to the database module. Have you tried that?

ecostanzi: > Could you add this block conditionally on !reactive

I disabled the file generation when the application is reactive. 

> Also, it seems that org.testcontainers:r2dbc maven dependency need to be added in addition to the database module. Have you tried that?

Yes I did. I lost the changes I stashed when I tried to add reactive support, but I put the generated code here: https://github.com/ecostanzi/jhipster-r2dbc-testcontainers. It's a reactive+mysql example.

ecostanzi:
> This seems to be fixed, is that right? So let's get it merged I would say 🎉

@atomfrede it's not :( HibernateTimezoneIT seems to be too tied up to the h2 database, in particular to the fact that H2 timezone is the JVM timezone [This line](https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/test/java/package/config/timezone/HibernateTimeZoneIT.java.ejs#L206) basically looses the timezone and make all the tests fail.


I created another branch to make this test fail: https://github.com/ecostanzi/generator-jhipster/runs/621423170?check_suite_focus=true

Given that it should always work out of the box, the options are:
- try to modify the test to get a version compatible with all databases (i tried but without success)
- add a comment that says "if you want to run this test with testcontainers set -Duser.timezone=GMT"
- forcing the timezone to GMT for all integration tests (via maven/gradle)
- removing this integration test 

cc @avdev4j maybe you can help us on this

ecostanzi:
I read the conversation here https://github.com/jhipster/generator-jhipster/issues/8284. These two comments were useful: https://github.com/jhipster/generator-jhipster/issues/8284#issuecomment-443332067, https://github.com/jhipster/generator-jhipster/issues/8284#issuecomment-443452568. I was able to fix the issue mysql timezone by changing the servertimezone parameter in the connection string of testcontainers properties. 

I think this can be merged now!

atomfrede:
Can you check the prettier errors?

ecostanzi:
I launched tests again, they passed. Don't know what happened with eslint.

DanielFran:
Relaunched failing tests

atomfrede:
I think the react tests are flaky longer already if I remember correctly.

atomfrede:
@ecostanzi Can you rebase against master? The failing tests are related to some changes with couchbase which are fixed on master already.

ecostanzi:
@atomfrede should be ok now

atomfrede:
Great work @ecostanzi. Let's merge this and we should update both the generated and our documentation for the next release.

PR #11588, opened 2020-04-14
Fix enums

### LINKED ISSUE #11587: Java code with enums with custom values don't compile
##### **Overview of the issue** & **reproduction**

When importing such a JDL content:
```
entity MyEntity {
  myField MyEnum
}

enum MyEnum {
  AAA(aaa),
  BBB(bbb)
}
```

The Java code won't accept the custom values because they are generated (didn't seem so before, but anyway...).

Generated code should compile.

##### **Reproduce the error**

Generate an app with
```
application
  config {
    baseName testWithEnums
  }}

entity MyEntity {
  myField MyEnum
}

enum MyEnum {
  AAA(aaa),
  BBB(bbb)
}
```

Check the generated enum, you should see it doesn't compile.

##### **Related issues**

Don't think so. Related to https://github.com/jhipster/jhipster-core/issues/459.

##### **Suggest a Fix**

#11588

##### **JHipster Version(s)**

master

```
application
  config {
    baseName testWithEnums
  }}

entity MyEntity {
  myField MyEnum
}

enum MyEnum {
  AAA(aaa),
  BBB(bbb)
}
```

Just generate an app with the provided JDL, it should do the trick.

Irrelevant.

MathieuAA:
Fixes #11587

Also added tests and refactored the code


mshima:
At least one integration test should be added:
https://github.com/jhipster/generator-jhipster/search?q=fieldValues&unscoped_q=fieldValues

generators/entity-server/files.js
```diff
@@ -386,22 +386,36 @@ function writeFiles() {
...
+                    );
                 }
             });
         }
     };
 }
+
+function getTemplateFileForEnums(jhipsterInstallationPath, enumInfo) {
+    const pathToEnumFolder = `${jhipsterInstallationPath}/${SERVER_MAIN_SRC_DIR}package/domain/enumeration`;
+    if (enumInfo.withoutCustomValues) {
```
mshima, line 414: I don’t like the idea of creating 2 templates.
Usually the template is the same:
```
If (withCustomValues ) {
With custom values  template
} else {
Without
}
```

generators/utils.js
```diff
@@ -454,39 +454,56 @@ function getJavadoc(text, indentSize) {
 /**
  * Build an enum object
  * @param {any} field : entity field
- * @param {string} angularAppName
- * @param {string} packageName
  * @param {string} clientRootFolder
  */
-function buildEnumInfo(field, angularAppName, packageName, clientRootFolder) {
+function buildEnumInfo(field, clientRootFolder) {
```
mshima, line 459: I think we should keep api stability for utils.

MathieuAA: This isn't used at all... The generator is full of this

MathieuAA: I'll create a second one, cleaned up and deprecate the other, so as to be removed for v7.

mshima:
> At least one integration test should be added:
> https://github.com/jhipster/generator-jhipster/search?q=fieldValues&unscoped_q=fieldValues

There are places that fieldValues is used directly. 
At least should use a replace with regexp like :
https://github.com/jhipster/generator-jhipster/blob/594fc34269150c57a1450afba8077c6c5e08adc9/generators/entity-client/templates/angular/src/test/javascript/spec/app/entities/entity-management.service.spec.ts.ejs#L71

generators/entity-server/files.js
MathieuAA: The issue is that the differences are quite clear between the templates and merging them is gonna lead to a messy result (yes, I find most templates messy but still... :))

MathieuAA: I'm gonna push the one-file solution

MathieuAA:
> There are places that fieldValues is used directly.
At least should use a replace with regexp like :

I don't get what you mean but I've added tests for enum generation.

mshima:
You need to add integration test.
Like changing:
https://github.com/jhipster/generator-jhipster/blob/0567113f9bf99471853695d1565f3c8f3ad35bae/test/templates/export-jdl/.jhipster/JobHistory.json#L41

to:
`"fieldValues": "FRENCH (French),ENGLISH (English),SPANISH (Spanish)"`

So this way all tests are executed against custom enum (cli, backend, e2e).

@murdos you did a recent refactoring on the IT entities, is necessary to add a new entity to it test custom enum? Or changing one of current ones is enough?

mshima:
Integration test probably will raise the following problems:
Faker is broken (I think it makes e2e fail):
https://github.com/jhipster/generator-jhipster/blob/20164cdcd5b16569deb9575991f290e3ab99fb5e/generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs#L141

Should be (not tested):
```
fields[idx].fieldValues.split(',').replace(/(.+?)\s?\(.*\)/, (match, firstGroup) => firstGroup)
```

Same for gatling:
https://github.com/jhipster/generator-jhipster/blob/e22ba2f3cbd275570f08c60149917c7d9c73f363/generators/entity-server/templates/src/test/gatling/user-files/simulations/EntityGatlingTest.scala.ejs#L191

Cli test is ok (maybe is missing a `\s?` in the middle and `\)` in de end of the regex) :
https://github.com/jhipster/generator-jhipster/blob/594fc34269150c57a1450afba8077c6c5e08adc9/generators/entity-client/templates/angular/src/test/javascript/spec/app/entities/entity-management.service.spec.ts.ejs#L71

Cli (e2e test)
https://github.com/jhipster/generator-jhipster/blob/83a9c2c3c273ea7c5c4032164a58be4f39fcf00d/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-update.component.html.ejs#L58

Java (backend test)
https://github.com/jhipster/generator-jhipster/blob/6092a0693d3ad1f2228bd2269af445e81660379c/generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs#L236

Probably is better to write getEnumsKeys on generator-base.

MathieuAA:
CLI test is okay, it's a greedy operator. I'm gonna see about other cases

MathieuAA:
@mshima I've improved the code based on your review, can you check again when you have some time? thanks :)

mshima:
I took a closer look, I think there is a confusion between custom values vs custom labels.
I think we should use the custom labels approach.

Example
```
enum status {
  HALT(Halt),
  IN_PROGRESS(In progress)
}
```

The `HALT/IN_PROGRESS` should be the db value.
`Halt/In progress` should be shown at the client side (select).

Take a look at https://github.com/jhipster/jhipster-core/issues/459 example.

So I will review with this approach.

MathieuAA:
I hesitated to implement it for this PR, but this is mixing things: fixes and features. That's why I didn't take this approach. I'll read over what you've reviewed later

mshima:
> I hesitated to implement it for this PR, but this is mixing things: fixes and features. That's why I didn't take this approach. I'll read over what you've reviewed later

I don't consider it contains fixes, because in my opinion it wasn't implemented at all (at least at current version, I don't know if it worked at a earlier version).
But once we implement values only, it will be a breaking change to implement labels.

MathieuAA:
> > I hesitated to implement it for this PR, but this is mixing things: fixes and features. That's why I didn't take this approach. I'll read over what you've reviewed later
> 
> I don't consider it contains fixes, because in my opinion it wasn't implemented at all (at least at current version, I don't know if it worked at a earlier version).
> But once we implement values only, it will be a breaking change to implement labels.

It was too buggy to actually work as intended. Let's take the label path

test/templates/enums/.jhipster/A.json
```diff
@@ -0,0 +1,32 @@
...
+    "fields": [
+        {
+            "fieldName": "myFieldA",
+            "fieldType": "MyEnumA",
+            "fieldValues": "AAA,BBB"
+        },
+        {
+            "fieldName": "myFieldB",
+            "fieldType": "MyEnumB",
+            "fieldValues": "AAA (aaa_aaa),BBB"
```
mshima, line 12: I think we should change to 
```suggestion
            "fieldValues": "AAA (aaa aaa),BBB"
```
White space here should be almost done.

generators/generator-base.js
```diff
@@ -2177,4 +2177,19 @@ module.exports = class extends PrivateBase {
...
+     * From an enum's values (with or without custom values), returns the enum's values without custom values.
+     * @param {String} enumValues - an enum's values.
+     * @return {Array<String>} the formatted enum's values.
+     */
+    getEnumValuesWithoutCustomValues(enumValues) {
+        if (!enumValues || enumValues === '') {
+            throw new Error('Enumeration values must be passed to get the formatted values.');
+        }
+        return enumValues
+            .replace(/\s/g, '')
```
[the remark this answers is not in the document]
mshima: This change is to allow white space like `IN_PROGRESS (In progress)`

test/templates/enums/.jhipster/A.json
```diff
@@ -0,0 +1,32 @@
+{
```
mshima, line 1: Copy this file to https://github.com/jhipster/generator-jhipster/tree/master/test-integration/samples/.jhipster
With name `FieldTestEnumWithValue.json` or something like this.
Add at:
https://github.com/jhipster/generator-jhipster/blob/30b21eae1e8df878dbd6c21e60e691d54f380fc3/test-integration/scripts/11-generate-entities.sh#L102

And:
https://github.com/jhipster/generator-jhipster/blob/30b21eae1e8df878dbd6c21e60e691d54f380fc3/test-integration/scripts/11-generate-entities.sh#L142

`sql` and `sqlfull` sections.

generators/generator-base.js
```diff
@@ -2177,4 +2177,19 @@ module.exports = class extends PrivateBase {
...
+            .split(',')
+            .map(enumValue => enumValue.replace(/(.+?)\(.*/, (match, firstGroup) => firstGroup));
```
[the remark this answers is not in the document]
mshima: The `\s?` is to allow white space like IN_PROGRESS (In progress)

test/templates/enums/.jhipster/A.json
MathieuAA: JCore and the generator don't produce this

mshima: The generator will with the 2 changes I am proposing to getEnumValuesWithCustomValues.
And 1 of the 2 must be the first to support, preferably the generator.

mshima: I will not block the PR due to this change.
Once the integration are added and i18n is fixed, the PR is done.
Integration test is more important than mocha test here.
But I will create another PR in sequence to add white space support.

MathieuAA: So you know, the feature was intended as JDL-only. Are you sure you want to introduce it in the generator? I don't mind but I'd rather do less work than more :)

mshima: Why JDL only? Nothing should be JDL only in my opinion.
Once it’s supported on generator-jhipster, it can by added by blueprints and is really easy to add a prompt.

MathieuAA: I disagree. Some things are just too ill-suited to do in CLIs. If I recollect this correctly this was the reason and I still agree with this. As the JDL is a "visible" thing, it's easy to do such a thing as having a custom enum label. In the CLI, you have to know how to do it prior to writing your values (that are unformatted). Or at least you have to inform the user of the possibility. Which can bloat the output

mshima: As I said your PR, your call.
But jhipster is a platform, not jdl dependent (which is lagging behind without blueprint support).
More then that, we are a community, if someone needs some feature that will not interfere too much on the maintenance, I see no reason to decline.
Anyway, you should add enums to integration test and fix i18n.
Currently the main IT tests are not jdl only.
So we can merge this PR.

mshima:
@MathieuAA looks like there are deserialization problems with the generated java enum with some configurations (ex: infinispan).
I don't know how do you want to proceed.
But I suggest to move the java generated enum change to another PR. It's not required for custom labels support.

mshima:
I identified the infinispan problem with the stacktrace, but I couldn't identify the others problems (can be as simple as increasing the timeout).

MathieuAA:
I'll check as soon as I get back to my computer, thanks for the investigation!!

MathieuAA:
Indeed, there seems to be an issue with infinispan, but it's not related to this PR...
See [this one](https://github.com/jhipster/generator-jhipster/pull/11606/checks?check_run_id=602335602) for instance.
It may be related to [this](https://stackoverflow.com/questions/58815818/infinispan-10-0-1-final-no-marshaller-registered-for-java-type-java-util-uuid). 

However, I'm at loss about the other issues =/

mshima:
@MathieuAA I have the filling that all errors are cache related.
You can try adding something like:
```
If (containsSomeCustomValues)
   addEntryToCache(enum.value, enum.packageName, this.cacheProvider);
```

at (change all chain to add enums like relationships):
https://github.com/jhipster/generator-jhipster/blob/d3cc5f6b8e021422bbab6d8151172ececb0eaf2d/generators/server/needle-api/needle-server-cache.js#L26

Or at:
https://github.com/jhipster/generator-jhipster/blob/b3c83b139e058827ac6d9913a100673cc15b9173/generators/entity-server/files.js#L387

mshima:
@MathieuAA I have the filling that all errors are cache related.
You can try adding something like:
```
If (containsSomeCustomValues)
   addEntryToCache(enum.value, enum.packageName, this.cacheProvider);
```

at (change all chain to add enums like relationships):
https://github.com/jhipster/generator-jhipster/blob/d3cc5f6b8e021422bbab6d8151172ececb0eaf2d/generators/server/needle-api/needle-server-cache.js#L26

Or at:
https://github.com/jhipster/generator-jhipster/blob/b3c83b139e058827ac6d9913a100673cc15b9173/generators/entity-server/files.js#L387

MathieuAA:
What would it do? What's your feeling about this?

mshima:
@MathieuAA you have changed the enum from java enum type to java class.
So deserialization and caching should be changed according.

Probably this should be added to the enum type.
https://github.com/jhipster/generator-jhipster/blob/9e1f366e439bb7c17a11732fd70e2671cc657d8d/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs#L152-L159

Probably should be added to the field that references the enum
https://github.com/jhipster/generator-jhipster/blob/9e1f366e439bb7c17a11732fd70e2671cc657d8d/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs#L306-L312

Maybe should add others specific configurations:
https://github.com/jhipster/generator-jhipster/blob/9e1f366e439bb7c17a11732fd70e2671cc657d8d/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs#L336
with `<%= packageName %>.domain.enumerations.*`

I don't think it's worth.

MathieuAA:
Not really, serialization & deserialization are handled [this way](https://docs.oracle.com/javase/8/docs/platform/serialization/spec/serial-arch.html#a6469) way. An enum is a class, it's just hidden from you (I've been told not to criticize Java's syntax, I don't want to get "accidentally" killed):
  - without specific values, there's a hidden private constructor that's used to instantiate one
  - when you declare them, you have to specify (at least) one yourself

That said, cache providers are supposed to handle that by not handling the specific values, right? they're just gonna use the default behavior and don't handle specific values. I'm no expert at caching, but I don't see why caching is the issue here.

generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs
```diff
@@ -138,8 +138,8 @@ for (lineNb = 1; lineNb <= numberOfRows; lineNb++) {
         }
         if (fields[idx].fieldIsEnum) {
             if (fields[idx].fieldValues.length !== 0) {
-                const enums = fields[idx].fieldValues.split(',');
-                data = enums[this.faker.random.number(enums.length-1)];
+                const enumValues = getEnumValuesWithCustomValues(fields[idx].fieldValues);
```
pascalgrimaud, line 141: Not 100% sure but these 2 lines causes some troubles with fake data.

This CSV can't be imported into the application.

```
id;my_field_a;my_field_b;my_field_c
1;AAA;BBB;aaa_aaa
2;BBB;aaa_aaa;aaa_aaa
3;BBB;aaa_aaa;aaa_aaa
4;BBB;BBB;b and b
5;AAA;BBB;b and b
6;AAA;aaa_aaa;aaa_aaa
7;BBB;BBB;b and b
8;BBB;aaa_aaa;aaa_aaa
9;BBB;aaa_aaa;aaa_aaa
10;BBB;BBB;b and b
```

This one can:

```
id;my_field_a;my_field_b;my_field_c
1;AAA;BBB;AAA
2;BBB;AAA;AAA
3;BBB;AAA;AAA
4;BBB;BBB;BBB
5;AAA;BBB;BBB
6;AAA;AAA;AAA
7;BBB;BBB;BBB
8;BBB;AAA;AAA
9;BBB;AAA;AAA
10;BBB;BBB;BBB
```

So maybe try to use the "key" instead of the "value" here?

pascalgrimaud: You can play with:
- KO -> https://github.com/pascalgrimaud/jh-issue-11588
- OK -> https://github.com/pascalgrimaud/jh-issue-11588/tree/fix-enums

Just launch ./mvnw and go to http://localhost:8080/field-test-enum-with-value

MathieuAA: I love you man

generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs
```diff
@@ -138,8 +138,8 @@ for (lineNb = 1; lineNb <= numberOfRows; lineNb++) {
...
+                data = enumValues[this.faker.random.number(enumValues.length-1)].value;
```
mshima, line 142:
```suggestion
                data = enumValues[this.faker.random.number(enumValues.length-1)].name;
```

generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs
mshima: Should be fixed in next line.

generators/entity-server/templates/src/main/resources/config/liquibase/fake-data/table.csv.ejs
MathieuAA: Thanks Marcelo!

generators/entity-server/templates/src/main/java/package/domain/enumeration/Enum.java.ejs
```diff
@@ -19,8 +19,32 @@
 package <%= packageName %>.domain.enumeration;
 
 /**
- * The <%= enumName %> enumeration.
- */
+* The <%= enumName %> enumeration.
+*/
```
mshima, line 23:
```suggestion
 */
```

generators/entity-server/templates/src/main/java/package/domain/enumeration/Enum.java.ejs
```diff
@@ -19,8 +19,32 @@
 package <%= packageName %>.domain.enumeration;
 
 /**
- * The <%= enumName %> enumeration.
- */
+* The <%= enumName %> enumeration.
```
mshima, line 22:
```suggestion
 * The <%= enumName %> enumeration.
```

MathieuAA:
@mshima I want to split the bounty with you, you've really helped me a lot here and for the whole feature.

MathieuAA:
I'm merging this, thanks @pascalgrimaud for the fix and @mshima for the review and the constant code improvement ideas on this!

pascalgrimaud:
@MathieuAA thanks in advance, for the beers !

MathieuAA:
Bounty claimed: https://opencollective.com/generator-jhipster/expenses/16642
@mshima I've claimed my half, you deserve the other

mshima:
Thank yo @MathieuAA

pascalgrimaud:
FYI, this PR will impact the Vue.js blueprint too
Will you have time to have a look on this @MathieuAA ?

MathieuAA:
I'll try to find some today

pascalgrimaud:
@MathieuAA : in fact, I already start the work here https://github.com/jhipster/jhipster-vuejs/pull/604
I'll ask you if I need help. Thanks :)

MathieuAA:
Alright then.

mshima:
Half bounty claimed: https://opencollective.com/generator-jhipster/expenses/19713

pascalgrimaud:
@mshima : approved

mshima:
@pascalgrimaud thanks. I have ~4 pending bounties, so you can take a look together if you want.

pascalgrimaud:
@mshima : yes, just mention me when you claim it on GitHub, so I got the email for notification, otherwise, I won't see it immediatly unless I participated to the ticket :)

PR #11598, opened 2020-04-19
New constants for SpringBoot and JHipster lib versions + snapshots repository for jhipster lib

murdos:
This way they don't need to be updated in Maven and Gradle build configuration for each update

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -82,10 +82,10 @@
         <profile.tls />
 
         <!-- Dependency versions -->
-        <jhipster-dependencies.version>3.7.0-SNAPSHOT</jhipster-dependencies.version>
+        <jhipster-dependencies.version><%= JHIPSTER_DEPENDENCIES_VERSION %></jhipster-dependencies.version>
         <!-- The spring-boot version should match the one managed by
         https://mvnrepository.com/artifact/io.github.jhipster/jhipster-dependencies/${jhipster-dependencies.version} -->
-        <spring-boot.version>2.2.6.RELEASE</spring-boot.version>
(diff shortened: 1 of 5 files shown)


generators/server/templates/pom.xml.ejs
```diff
@@ -82,10 +82,10 @@
         <profile.tls />
 
         <!-- Dependency versions -->
-        <jhipster-dependencies.version>3.7.0-SNAPSHOT</jhipster-dependencies.version>
+        <jhipster-dependencies.version><%= JHIPSTER_DEPENDENCIES_VERSION %></jhipster-dependencies.version>
         <!-- The spring-boot version should match the one managed by
```
Tcharl, line 86: Just an idea while I see that: would it be possible to add a conditional on this property?
if (JHIPSTER_DEPENDENCIES_VERSION .endwith("-SNAPSHOT") {
// declare Snapshot repository in pom or gradle build file
}

murdos: This is what I want to setup for SpringBoot in the `spring-boot-2.3` branch, by conditionally defining Spring milestones repository. 
However I don't see how it will works for JHipster dependencies, since the snapshots are not regularly published and thus are not really useful (that might change if we use GitHub Packages in the future).

Tcharl: Automated publication of jhipster-dependencies snapshots is somewhat that is easily doable :-)

murdos: Yep. I'm working on it :)

murdos: @Tcharl : Done! Julien was in fact already publishing automatically snapshots, so I've added conditionally the snapshot repository ;)

murdos:
Can this be merged? I need this for cleaning the Spring Boot 2.3 branch I'm working on

PR #11608, opened 2020-04-20
fix deprecated code from JJWT

jdubois:
See https://github.com/jwtk/jjwt/blob/master/api/src/main/java/io/jsonwebtoken/Jwts.java#L102


mshima:
See https://github.com/jhipster/generator-jhipster/pull/11500. 
It’s stale but there is a review to reuse the parser.

jdubois:
Thanks @mshima !
Indeed it's the same code - I'm definitely tired tonight!! And the performance part is very interesting

Issue #11615, opened 2020-04-21
Missing Testcontainers Dependency: Redis with Gradle 


SudharakaP:
##### **Overview of the issue**

If we generate a basic project with Redis and Gradle there seem to be a compilation error in `RedisTestContainerExtension.java` class due to missing Testcontainers dependency on Gradle.

##### **Motivation for or Use Case**

Obviously we don't want compilation errors. :smile: 

##### **Reproduce the error**

Just use the below configuration to generate a project and try to compile it. 

##### **Suggest a Fix**

PR will follow soon. :smile: 

##### **JHipster Version(s)**

v6.8.0

##### **JHipster configuration**

```json
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp"
    },
    "jhipsterVersion": "6.8.0",
    "applicationType": "monolith",
    "baseName": "jhipster",
    "packageName": "com.mycompany.myapp",
    "packageFolder": "com/mycompany/myapp",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "redis",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": false,
    "buildTool": "gradle",
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "MDI4ZjQ2OTQ2ZmEyYmVlNGM4YjE4MjgzZWYwYjAyZTdlZjg5ZjA4NzMyNWM2ODI4ZjZlMmViNzNmZjA4M2ZlZjAzNDc1YzA5NGE5NDA5ZGM1MmUxZDg4NDNjNjgzOTQyNjczYjI5ODQyMzM4NDcwMjE3ZjFiZWQwZjI3OTk0NTA=",
    "embeddableLaunchScript": false,
    "useSass": true,
    "clientPackageManager": "npm",
    "clientFramework": "angularX",
    "clientTheme": "none",
    "clientThemeVariant": "",
    "creationTimestamp": 1587512202352,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "enableTranslation": false,
    "blueprints": []
  }
}
```

-   [x] Checking this box is mandatory (this is just to show you read everything)

PR #11616, opened 2020-04-22
Adds Testcontainers as a Test Dependency for Gradle + Redis Configuration

SudharakaP:
This adds Testcontainers as a test dependency for Gradle + Redis configuration. I've added the `database-common` artifact just as we do in `Cassandra`. 

@bsideup: Hey, if you don't mind me asking a quick question; I am wondering what is the difference between; `org.testcontainers:database-commons` vs `org.testcontainers:testcontainers` dependencies?  

Fixes https://github.com/jhipster/generator-jhipster/issues/11615

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -596,6 +596,9 @@ dependencies {
     <%_ if (messageBroker === 'kafka') { _%>
     testImplementation "org.testcontainers:kafka"
     <%_ } _%>
+    <%_ if (cacheProvider === 'redis') { %>
+    testImplementation "org.testcontainers:testcontainers"
+    <%_ } _%>
     //jhipster-needle-gradle-dependency - JHipster will add additional dependencies here
 }
 


generators/server/templates/build.gradle.ejs
```diff
@@ -596,6 +596,9 @@ dependencies {
     <%_ if (messageBroker === 'kafka') { _%>
     testImplementation "org.testcontainers:kafka"
     <%_ } _%>
+    <%_ if (cacheProvider === 'redis') { %>
```
atomfrede, line 599: For maven we just use `org.testcontainers:testcontainers`. `database-commons` depends on `testcontainers` so not problem with that, just curious why explicitly using `database-commons` here

SudharakaP: @atomfrede : I wasn't sure to use `org.testcontainers.database-commons` or `org.testcontainers.testcontainers`. I see that in [Cassandra we explicitly use `database-commons`](https://github.com/jhipster/generator-jhipster/blob/3b008bcbfcbdd190dd951b36743285bbe6a680b6/generators/server/templates/pom.xml.ejs#L485-L495). This is why I asked bsidesup above the difference between the two (do you know? :thinking:). So you mean using `org.testcontainers.testcontainers` is more appropriate. If so do you know why we use `org.testcontainers.database-commons` in Cassandra when using a similar `GenericContainer`? :smile:

bsideup:
@SudharakaP `org.testcontainers:testcontainers` is the main, core dependency, while `database-commons` is an utility module that contains things like SQL script parsing and should only be used by concrete modules like mysql, cassandra and others.

generators/server/templates/build.gradle.ejs
bsideup: Cassandra depends on `database-commons`, no need to use it explicitly.

SudharakaP: @bsideup: You mean [Cassandra module[(https://www.testcontainers.org/modules/databases/cassandra/) depends on `database-commons` right? We don't use that; we use a `GenericContainer` for Cassandra instead and still use `database-common`. My thinking is this is unnecessary probably. :thinking:

SudharakaP:
> @SudharakaP `org.testcontainers:testcontainers` is the main, core dependency, while `database-commons` is an utility module that contains things like SQL script parsing and should only be used by concrete modules like mysql, cassandra and others.

Thanks much. That clears things up :smile:

generators/server/templates/build.gradle.ejs
SudharakaP: > For maven we just use `org.testcontainers:testcontainers`. `database-commons` depends on `testcontainers` so not problem with that, just curious why explicitly using `database-commons` here

@atomfrede : So I've replaced it with `org.testcontainers:testcontainers`. Thanks for pointing it out. :smile:

atomfrede:
Thanks @bsideup for the quick response!

PR #11626, opened 2020-04-22
add missing property annotations

atomfrede:
This pr adds missing property annotations. Currently the Map can't be converted without a custom converter, going to check how to best do that.


generators/server/templates/src/main/java/package/domain/PersistentAuditEvent.java.ejs
```diff
@@ -143,6 +148,9 @@ public class PersistentAuditEvent implements Serializable {
     @Transient
         <%_ } _%>
     <%_ } _%>
+    <%_ if (databaseType === 'neo4j') { _%>
+    @Transient
```
atomfrede, line 152: Right now mapping of maps is not part of SDN/RX (on purpose). So let's make it transient for now. As we plan to remove the audit feature (and neo is still in beta) I would say thats okay.

atomfrede: This fixes also the daily build as before the test was not generated

PR #11645, opened 2020-04-26
Finish integration of prettier java

murdos:
- Install prettier-plugin-java in generated application so it can be used for later development, not only during initial generation
- Enable commit hooks for server code, and move commit hooks configuration from the client subgen to the common subgen
- Use constants to globally set version of husky, lint-staged and prettier, and automatically align version of prettier between generator-jhipster and generated applicaition
-  Save prettierJava option .yo-rc.json. This is necessary to keep consistent formatting when upgrading or when generating entities.


mshima:
I didn’t run the code, but seems everything is correct.

PR #11646, opened 2020-04-26
Fix OpenApiConfiguration to not use google.guava.predicates - which i…

gzsombor:
…s not referred by directly anymore

### DIFF
--- a/generators/server/templates/src/main/java/package/config/OpenApiConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/OpenApiConfiguration.java.ejs
@@ -18,7 +18,6 @@ limitations under the License.
 -%>
 package <%= packageName %>.config;
 
-import com.google.common.base.Predicates;
 import io.github.jhipster.config.JHipsterConstants;
 import io.github.jhipster.config.JHipsterProperties;
 import io.github.jhipster.config.apidoc.customizer.SwaggerCustomizer;
@@ -47,7 +46,7 @@ public class OpenApiConfiguration {
     @Bean
     public SwaggerCustomizer noApiFirstCustomizer() {
         return docket -> docket.select()
-            .apis(Predicates.not(RequestHandlerSelectors.basePackage("<%= packageName %>.web.api")));
+            .apis(RequestHandlerSelectors.basePackage("<%= packageName %>.web.api").negate());
     }
 
     @Bean


gzsombor:
Thx!

pascalgrimaud:
With this change, all JDL daily builds failed.
See https://github.com/hipster-labs/jhipster-daily-builds/actions?query=workflow%3A%22JDLMS+Ehcache+MySQL%22

Here an extract of stacktrace:
```
> Task :compileJava
/home/runner/app/store/src/main/java/com/jhipster/demo/store/config/OpenApiConfiguration.java:31: error: incompatible types: java.util.function.Predicate<RequestHandler> cannot be converted to com.google.common.base.Predicate<RequestHandler>
            .apis(RequestHandlerSelectors.basePackage("com.jhipster.demo.store.web.api").negate());
                                                                                               ^
Note: /home/runner/app/store/src/main/java/com/jhipster/demo/store/security/jwt/TokenProvider.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
1 error
```

@gzsombor @atomfrede : one of you can have a look plz ?

gzsombor:
Interesting. Does it use a different version of springfox-core instead of 3.0.0-SNAPSHOT? I think, in 3.0, springfox doesn't depend on guava anymore  - but uses code from Java 8 finally.

pascalgrimaud:
I didn't have a look into details.
But what I can, is the use of master branch:
- generator-jhipster
- jhipster (lib)
- https://github.com/hipster-labs/jhipster-daily-builds/blob/master/.github/workflows/jdl-ms-ehcache-mysql.yaml#L28-L31

atomfrede:
@gzsombor [We are still on 2.9.2](https://github.com/jhipster/jhipster/blob/master/jhipster-dependencies/pom.xml#L90) which has a strange mixture of guava predicates and java predicates (looked at their source). I would say let's revert that change.

gzsombor:
This is very strange, I've generated an app with sql+reactive configuration, and I've got springfox-3.0.0-snapshot

atomfrede:
:thinking: Indeed very strange. Did you use maven? The jdl sample was gradle, both use the jhipster bom so should have the same dependencies

atomfrede:
Ah found it. In maven we have 

```xml
    <repository>
            <id>springfox-snapshots</id>
            <url>https://oss.jfrog.org/artifactory/oss-snapshot-local/</url>
            <snapshots>
                <enabled>true</enabled>
            </snapshots>
        </repository>

PR #11654, opened 2020-04-27
use git hash or build information to prefix cache keys

### LINKED ISSUE #9822: Use the Git hash in Spring cache keys
Following [my recent article on using Redis cache](https://dev.to/azure/configuring-azure-redis-cache-to-boost-spring-boot-performance-7-7-52dl), I had a very interesting discussion with @tkvangorder as there is an interesting edge case: when you do a new release, you still inherit from the old cache, and maybe you will have some wrong data there, because you changed something in your code.

I think of it like a database: we use Liquibase to update the database, but what about the cache?

At first I thought we should have some specific task in our release process to clear out the cache, and I know many people who do that (including several of my own former clients!), but that's not a perfect solution, as you can't have a perfect system where the cache is cleared exactly at the same moment as the application is upgraded.

So @tkvangorder has a great solution: he's using the application version in the cache keys. This solves everything nicely!

With JHipster we could easily have something similar, as we have (at least with Maven) the Git commit id which is available as a Spring Boot property (`${git.commit.id.describe}`). So I wonder if we could have some specific Spring Cache implementation in JHipster Framework, where the default key would be automatically prefixed by that commit hash.

atomfrede:
Finally I found some time to integrate the custom prefixed cash key implementation. For details how keys for spring cache abstractions are generated look at the implementation in the jhipster lib.

https://github.com/jhipster/jhipster/pull/618

closes #9822

### DIFF
--- a/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/CacheConfiguration.java.ejs
@@ -51,9 +51,7 @@ import com.hazelcast.core.Hazelcast;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-    <%_ if (serviceDiscoveryType === 'eureka' || serviceDiscoveryType === 'consul') { _%>
-import org.springframework.beans.factory.annotation.Autowired;
-    <%_ } _%>
+
 <%_ } _%>
 <%_ if (cacheProvider === 'ehcache' || cacheProvider === 'caffeine') { _%>
 import org.springframework.boot.autoconfigure.cache.JCacheManagerCustomizer;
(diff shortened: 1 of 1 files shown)


atomfrede:
Going to have look at the failed gateway builds most likely tonight

pascalgrimaud:
@atomfrede : as the jhipster lib part has been merged, do you feel confident to merge this too, for including it to the next release ?

atomfrede:
I tested it with multiple cache implementations and it worked well. We are more or less just using spring cache customizations, but would be great if someone else would try it. I guess it is not urgent to have that in the next release.

pascalgrimaud:
Then let's go and let's see the next day. I'm trusting the daily builds :)

atomfrede:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/17723

PR #11656, opened 2020-04-27
Update NodeJS Template Version to 12

### LINKED ISSUE #11650: The generator project - Deprecate node v8 versions

Node v8 is not longer supported, and node v10 is in maintenance mode (v12 being the LTS version).

First motivation: we can use the newer features of node (`mkdir('p', ...)`), bump (I'm talking semver-major updates) some dependencies.
Second motivation: JCore is gonna do the same.
Third motivation: remove the `mkdirp` dependency, because it won't be needed anymore.

Didn't find any.

-   [X] Yes, I do solemnly swear that I've read everything.

SudharakaP:
This updates NodeJS in templates to version 12.

Related to #11650


MathieuAA:
I think this can be merged unless a build fails

generators/server/templates/package.json.ejs
```diff
@@ -33,6 +33,6 @@
     "@openapitools/openapi-generator-cli": "1.0.12-4.3.0"
   },
   "engines": {
-    "node": ">=8.9.0"
+    "node": "12.16.1"
```
mshima, line 36: Do you mean to force a specific version?

mshima:
Is there a reason to don’t support node 10 too?

generators/server/templates/package.json.ejs
SudharakaP: @machima: Not really, that was a mistake. Thanks. :+1:

generators/client/templates/react/package.json.ejs
```diff
@@ -186,7 +186,7 @@ limitations under the License.
   },
 <%_ } _%>
   "engines": {
-    "node": ">=8.9.0"<% if (clientPackageManager === 'yarn') { %>,
+    "node": ">=12.6.1"<% if (clientPackageManager === 'yarn') { %>,
```
mshima, line 189:
```suggestion
    "node": ">=12.16.1"<% if (clientPackageManager === 'yarn') { %>,
```

SudharakaP:
> Is there a reason to don’t support node 10 too?

I think you are right. We should do Node 10 since it's also Active; https://nodejs.org/en/about/releases/

generators/client/templates/react/package.json.ejs
SudharakaP: I've updated Node to 10 instead. :smile:

mshima:
https://github.com/jhipster/generator-jhipster/pull/11657 was merged so:
There is no reason to make the generator's node >=12.6.1 and keep the templates generated with node >=12.6.1 requiring node >=10.20.1

You should extract a constant like @murdos did at https://github.com/jhipster/generator-jhipster/pull/11645/files see: https://github.com/jhipster/generator-jhipster/blob/c3b06470e66b5272d9b44f5428bb56b3b33f0532/generators/generator-constants.js#L50

SudharakaP:
@mshima: I was under the impression that we should give users the flexibility of using the actively maintained Node versions; whereas for `generator-jhispter` we can use whatever most recent Node version we prefer.
 
I think with the approach you are describing both (templates + generator) will have the same node version right? :smile:

mshima:
> I think with the approach you are describing both (templates + generator) will have the same node version right? 😄

Yep, same version makes more sense to me. I don't see a reason for them to differ.
But I think we should allow node >=10.20.1, and only move to 12 when 10 is deprecated.

pascalgrimaud:
I'm in favor of merging this.

It will probably work for Node 10 but we don't have any builds for that. And I don't want to add another support.

What we use in our build is Node 12, so we should promote this version.

murdos:
Just regarding prettier version: the case is a bit different since we need to use the same version in the generator and in the generated application, since prettier is used to format code on the fly during generation. Otherwise, if versions are different, running prettier again inside generated application, and just after generation might generates changes and that would be awkward for our users.

mshima:
> Just regarding prettier version: the case is a bit different since we need to use the same version in the generator and in the generated application, since prettier is used to format code on the fly during generation. Otherwise, if versions are different, running prettier again inside generated application, and just after generation might generates changes and that would be awkward for our users.

I agree, prettier should use the exact same version.

A bit but not entirely different, what I am saying is that if we are supporting generating the application on node `>=12.16.1`, we should support running the generated application on `>=12.16.1`.

But since engines is only advisory, there will be only a warning if generator-jhipster is installed on node 10.

SudharakaP:
> A bit but not entirely different, what I am saying is that if we are supporting generating the application on node `>=12.16.1`, we should support running the generated application on `>=12.16.1`.
> 
> But since engines is only advisory, there will be only a warning if generator-jhipster is installed on node 10.

I am okay with both ways, so I've went ahead and updated it back to 12. Also I've used the `NODE_VERSION` variable so that we have the same version everywhere. Let me know if you see any further issues. :smile:

vishal423:
It seems a bit rigid to force users to always have the latest nodejs version to be able to use jhipster cli. I just tried creating a new application from the master, and noticed an error:

```
You are running Node version 12.13.1
JHipster requires Node version >=12.16.1
Please update your version of Node.
```

murdos:
> It seems a bit rigid to force users to always have the latest nodejs version to be able to use jhipster cli. I just tried creating a new application from the master, and noticed an error:
> 
 
I agree. I got the same error. I vote to downgrade to 10 as minimum.

SudharakaP:
> > It seems a bit rigid to force users to always have the latest nodejs version to be able to use jhipster cli. I just tried creating a new application from the master, and noticed an error:
> 
> I agree. I got the same error. I vote to downgrade to 10 as minimum.

I agree. So I've went ahead and created a PR to set the generator and template node version to 10.20.1. :smile:

PR #11665, opened 2020-04-29
Update prettier to version 2.0.5 and reformat code

murdos:


mshima:
I don’t think we should merge this.
This probably will break every single PR due to the array trailing coma.

murdos:
@jhipster/developers : what should we do in your opinion?

A: update to prettier 2.0 but set `trailingComma` to `none` (the default value before prettier 2.0) and maybe switch back `trailingComma` to default prettier value for v7

B: wait for v7 to switch to prettier 2.0

vishal423:
I prefer trailing comma as that provides cleaner diffs and readability.

pascalgrimaud:
Thanks a lot for this PR @murdos 
Just my opinion, I'm in favor of merging this.

The more we try to improve our code, the more we are stuck because of possible breaking change.
It's pretty easy to update the existing code.

Concerning me, I already did my own personal fork of generator-jhipster for my customer with some items which will be in v7 and everything works well. I'll include this PR in my fork, so again, thanks for your work @murdos !

deepu105:
I think we can merge this. I was also doing the exact same in my React branch along with some other upgrades. I'll do a separate PR with other upgrades then

PR #11666, opened 2020-04-29
Factorize the construction and definition of JDBC urls

murdos:
We have many templates that uses a JDBC url depending on the database type.
This PR creates a function dedicated to that responsibility, used by the templates.

Because of this duplication, I think, there was subtle differences in options used for MySQL and MariaDB URL, I chose to uniformize them, but your feedback is welcome.


generators/server/templates/src/main/resources/config/application-dev.yml.ejs
```diff
@@ -105,31 +105,31 @@ spring:
...
         username: root
         password:
         <%_ } else if (devDatabaseType === 'postgresql') { _%>
-        url: jdbc:postgresql://localhost:5432/<%= baseName %>
+        url: <%- getJDBCUrl(devDatabaseType, { databaseName: baseName, hostname: 'localhost' }) %>
         username: <%= baseName %>
         password:
         <%_ } else if (devDatabaseType === 'oracle') { _%>
-        url: jdbc:oracle:thin:@localhost:1521:xe
+        url: <%- getJDBCUrl(devDatabaseType, { databaseName: baseName, hostname: 'xe' }) %>
```
mshima, line 120:
```suggestion
        url: <%- getJDBCUrl(devDatabaseType, { databaseName: 'xe', hostname: 'localhost' }) %>
```

murdos: Thanks @mshima for spotting this error!

MathieuAA:
Can you add some tests for the function?

murdos:
@MathieuAA : done 😉

PR #11674, opened 2020-04-30
Revert "Fix OpenApiConfiguration to not use google.guava.predicates - which i…"

atomfrede:
Reverts jhipster/generator-jhipster#11646

SpringFox 2.9.x has a mixture of java predicates and guava predicates.

### DIFF
--- a/generators/server/templates/src/main/java/package/config/OpenApiConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/OpenApiConfiguration.java.ejs
@@ -18,6 +18,7 @@ limitations under the License.
 -%>
 package <%= packageName %>.config;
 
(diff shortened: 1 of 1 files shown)


atomfrede:
This is the [current code in 2.9.2](https://github.com/springfox/springfox/blob/2.9.2/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/ApiSelectorBuilder.java)

And the migration to `java.util.Predicate` is [here](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/plugins/ApiSelectorBuilder.java). It was done in 2018(!!!) already and it seems there is no new non-snaphsot release

PR #11683, opened 2020-05-01
[SpringBoot 2.3] Upgrade to Spring Data Elasticsearch 4.0 and Elasticsearch 7.6

murdos:
- replace Jest with RestHighLevelClient. Jest still [does not support Elasticsearch 7](https://github.com/searchbox-io/Jest/issues/670) and is not supported anymore by SpringBoot. This also perfectly align with Policy #1
- remove configuration related to embedded node server, since it's not supported by Elasticsearch and Spring Boot since 2.0


mraible:
We should make sure this works with Heroku. One of the reasons I switched to Jest is because it did not. See https://github.com/jhipster/generator-jhipster/issues/7810#issuecomment-399144535 for more info.

murdos:
@mraible : thanks for the comment, it helped me realize that I forgot to clean up our deployments subgenerators (kubernetes, openshift, heroku, ...).

I did some tests, and in fact I was not able to have a working application with heroku and master of generator-jhipster. I guess it's related to jest not supporting ES 7.
I also did some tests with this branch, and it's not working out of the box because Spring Boot doesn't accept and parse user:password information in BONSAI_URL, so I've open an issue for that: https://github.com/spring-projects/spring-boot/issues/21291

mraible:
@murdos Do you think this is a breaking change? If so, maybe we should start a v7 branch?

pascalgrimaud:
@murdos @mraible : do you want me to release v6.9.0 ? then, we'll start the v7 branch from v6.9.0 and we can continue to release v6 branch when needed ?

atomfrede:
As a workaround can't we parse the url ourselfs and set username/password manually? I once did something similar using dokku, which does not populate a jdbc url to parse the database url and create a valid jdbc url from it.

murdos:
> @murdos Do you think this is a breaking change? If so, maybe we should start a v7 branch?

Which change? 
The only constraint I can think of is the requirement of version >= 7.6 of Elasticsearch since the Java High Level REST Client [is only guaranteed to be able to communicate with any Elasticsearch node running on the same major version and greater or equal minor version](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/master/java-rest-high-compatibility.html). 
But I'm not sure this requires switching to a v7 branch.

> 
> @murdos @mraible : do you want me to release v6.9.0 ? then, we'll start the v7 branch from v6.9.0 and we can continue to release v6 branch when needed ?

As you can see in `cleanup.js` I was expecting that a 6.9.0 version will indeed be released before the version including Spring Boot 2.3, and I was targeting a 6.10.0 version for SB2.3. But depending on discussion SB2.3 might end on v7 😉
Anyway, there's no emergency for a 6.9.0 for me, but since there a bit of new features in master, that might be interesting.

mraible:
The breaking change for me is my app with ES will no longer work on Heroku. If we can fix that, I'm 😎 with merging to 6.x.

IMO, we should keep upgrading Spring Boot (even to 2.3) on 6.x. The only reason I can see waiting for 2.3 in v7 is if we're closed to finished with v7.

murdos:
FYI I'm going to merge this PR (in the `spring-boot-2.3` branch).
There's no reason to keep using Jest in JHipster: https://github.com/VanRoy/spring-data-jest/issues/116#issuecomment-637779531

pascalgrimaud:
agree @murdos, go ahead

PR #11686, opened 2020-05-02
Generate reactive client with openapi generator

avdev4j:

This a first version of what openapi client generator could be done to support both reactive and spring openfeign clients.

In the prompt phase, we need to detect if the JHipster application is reactive or not. Then in the writing phase, call the openapi generator with the Java Webclient generator parameter.

TODO:
- [x] Clean code between both Spring and Webclient functions
- [x] Be sure that Webclient si still the best option
- [x] if yes, clean all unwanted files
- [x] Test on Gradle applications
- [x] Fix Localdate compilation errors (import org.threeten.bp.LocalDate is missing)
- [x] Fix JUnit failing import (wrong version ?)


avdev4j:
hum... should we add the following dependencies when using webclient generator ?

```        
        <dependency>
            <groupId>org.threeten</groupId>
            <artifactId>threetenbp</artifactId>
            <version>1.4.4</version>
        </dependency>
        <dependency>
            <groupId>com.github.joschi.jackson</groupId>
            <artifactId>jackson-datatype-threetenbp</artifactId>
            <version>2.10.0</version>
        </dependency>
```

avdev4j:
It seems that generated tests use the JUnit4 framework instead of JUnit5.

package-lock.json
```diff
@@ -2327,7 +2327,7 @@
                 },
                 "load-json-file": {
                     "version": "2.0.0",
-                    "resolved": "https://registry.npmjs.org/load-json-file/-/load-json-file-2.0.0.tgz",
+                    "resolved": "http://registry.npmjs.org/load-json-file/-/load-json-file-2.0.0.tgz",
```
atomfrede, line 2330: Thats a bit strange, why did it change from `https` to `http`?

avdev4j: you're right, I don't know why. I can revert this change anyway.

atomfrede:
> hum... should we add the following dependencies when using webclient generator ?
> 
> ```
>         <dependency>
>             <groupId>org.threeten</groupId>
>             <artifactId>threetenbp</artifactId>
>             <version>1.4.4</version>
>         </dependency>
>         <dependency>
>             <groupId>com.github.joschi.jackson</groupId>
>             <artifactId>jackson-datatype-threetenbp</artifactId>
>             <version>2.10.0</version>
>         </dependency>
> ```

I think we should try to rely on standard java dataypes if possible. Durations are available in `java.time`. Not sure if `threetenbp` offers more features, but I don't think so.

MathieuAA:
From the lib's page:
> An adaptation of the JSR-310 backport for Android.
Attention: Development on this library is winding down. Please consider switching to Android Gradle plugin 4.0, java.time.*, and its core library desugaring feature in the coming months.

Maybe we should remove it?

avdev4j:
I think it will be easy to hack this, according to the doc we can change the date Library by using the "java8" value for the dateLibrary option.

https://openapi-generator.tech/docs/generators/java/

avdev4j:
As I thought using the java8 dateLibrary option fix the date issue !

avdev4j:
Ok, so I struggle a little bit with the generator I have to choice.
As I can see, we are currently using the spring generator, which is a server generator, with the spring-cloud library to generate a client.

I tried to do the same and adding the `reactive` flag but I only can do this with the spring-boot library. I turn the library to spring-boot and it generate a Spring boot server, not a client (indeed Spring generator is a server one... logic results).

Maybe @cbornet or @ecostanzi  you can help me and explain me why we are using a server generator to generate a client api (according to https://openapi-generator.tech/docs/generators).
I feel like spring/spring-cloud generator is not a server one and should be under java client generators.

Anyway, we cannot use the spring/spring-boot + reactive because it's generate a spring boot implementation of our api definition, not a client.

Java/Webclient seems to remain the best option, but it generate a loooot of files I don't want to add to my application (e. g. gradle wrapper, gradle config, Androi manifest etc.).

thanks teammates !

cbornet:
Yes spring cloud generator is a library of spring generator because it's only a change in the templates. I agree it can seem a bit weird and I apologize since I was the one who coded it.

cbornet:
Spring-boot +reactive does indeed generate server stubs and that's what we use for the "API-first" option. For a reactive client I believe Java+webclient is the good choice.

cbornet:
As for the files you don't want to generate, you should be able to choose which files you want with the supportingFiles option. That's what we do for the non-reactive client.

avdev4j:
> Yes spring cloud generator is a library of spring generator because it's only a change in the templates. I agree it can seem a bit weird and I apologize since I was the one who coded it.

To be honest I needed times to understand the links between Spring libraries and server/client roles. It seems it's clearer now. It was easier to include spring-cloud as client generator into the server one ;).

> For a reactive client I believe Java+webclient is the good choice.
I did this choice since we can't use the reactive flag, good !

> you should be able to choose which files you want with the supportingFiles option
WDYT about using a specific webclient .openapi-generator-ignore file ? I'm a little bit confused about how the supportingFiles works because it seems if the parameter is present or not there is no consequences. the ApiKeyRequestInterceptor.java is always generated.

avdev4j:
I've tested with both Maven and Gradle. I also tested on "real" projects and it seems to be fine. Ready for review

avdev4j:
Could we merge this one @juliensadaoui ?

PR #11687, opened 2020-05-02
Downgrade node version to 10.20.1 in both templates and generator

SudharakaP:
This sets the template and generator node engine version to >=10.20.1

Related to https://github.com/jhipster/generator-jhipster/pull/11656

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -23,7 +23,7 @@ const packagejs = require('../package.json');
 const JAVA_VERSION = '1.8'; // Java version is forced to be 1.8. We keep the variable as it might be useful in the future.
 
 // Version of Node, Yarn, NPM
-const NODE_VERSION = '12.16.1';
+const NODE_VERSION = '10.20.1';
 const YARN_VERSION = '1.22.4';
 const NPM_VERSION = '6.14.4';
 
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -23,7 +23,7 @@ const packagejs = require('../package.json');
 const JAVA_VERSION = '1.8'; // Java version is forced to be 1.8. We keep the variable as it might be useful in the future.
 
 // Version of Node, Yarn, NPM
-const NODE_VERSION = '10.20.1';
+const NODE_VERSION = '12.16.1';
 const YARN_VERSION = '1.22.4';
 const NPM_VERSION = '6.14.4';
 
--- a/package.json
+++ b/package.json
@@ -105,8 +105,8 @@
         "inquirer": "6.3.1"
     },
     "engines": {
-        "node": ">=12.16.1",
-        "npm": ">=6.13.4"
+        "node": ">=10.20.1",
+        "npm": ">=6.14.4"
     },
     "license": "Apache-2.0",
     "collective": {


generators/generator-constants.js
```diff
@@ -23,7 +23,7 @@ const packagejs = require('../package.json');
 const JAVA_VERSION = '1.8'; // Java version is forced to be 1.8. We keep the variable as it might be useful in the future.
 
 // Version of Node, Yarn, NPM
-const NODE_VERSION = '12.16.1';
```
DanielFran, line 26: In the generator we define always the LTS version, not sure we should downgrade here?

SudharakaP: No sure; I downgraded so as to make sure we have the same version both in the generator and the templates as pointed out at https://github.com/jhipster/generator-jhipster/pull/11656#issuecomment-620812371. I'll let you guys to decide the final course of action here. :thinking: 

cc: @mshima @murdos

mshima: In my opinion we should use always use LTS in maintenance mode (v10 right now) (it should be more stable https://nodejs.org/en/about/releases/).

But downgrading here will downgrade generated ci versions too.
My opinion is that we should create a new field that reads from jhipster package.json to write every generated package.json like I pointed in the other PR. And that should be >=10.19.0 (10.20.x doesn't have any security fixes).

MathieuAA: If there are some issues with v12 (pointed out in another issue I believe), then switching to v10 makes sense. They're both LTS so it should be okay for now. We should then migrate to v12 next year (I believe v10 won't be "active" April next year).

SudharakaP: > In my opinion we should use always use LTS in maintenance mode (v10 right now) (it should be more stable https://nodejs.org/en/about/releases/).
> 
> But downgrading here will downgrade generated ci versions too.
> My opinion is that we should create a new field that reads from jhipster package.json to write every generated package.json like I pointed in the other PR. And that should be >=10.19.0 (10.20.x doesn't have any security fixes).

@mshima : Isn't the generated CI versions should be downgraded as well? I mean if we are using v10 for the generator; isn't the CI that is generated through it downgraded to v10 as well? :thinking: 

For the security fixes; I am confused. How come, 10.20.x which succeeds 10.19.0 doesn't have the same security fixes? :thinking:

MathieuAA: If there's a newer version, and a feature on this newer version also somehow concerns v10 (not a breaking change), I guess the "natural" thing to do is to backport the feature in v10. That, or I don't know... It's a good question.

mshima: > @mshima : Isn't the generated CI versions should be downgraded as well? I mean if we are using v10 for the generator; isn't the CI that is generated through it downgraded to v10 as well? 🤔

Probably @DanielFran is pointing that:
6.8.0 is already generating ci with 12.x, so it will be a downgrade.
6.8.0 is generating package.json with 8.x, so 6.9.0 will upgrade to 10.x.

> For the security fixes; I am confused. How come, 10.20.x which succeeds 10.19.0 doesn't have the same security fixes? 🤔

I don't think we should make the user to always have the latest version.
10.20.0 had a major change to compiler stack on macOS.
Making node >=10.12.0 (first v10 LTS version, I think) is fine to me too.

SudharakaP: @mshima : So your suggestion is to have >=10.12.0 for both templates (excluding CI templates) and the generator but have >=12.16.1 for CI templates. Am I correct?

mshima: Yes having package.json >=10.12.0 will remove the warning.
And the CI will actually install the specific version, so keep it 12.16.1.

pascalgrimaud: This value should stay at the latest LTS: https://nodejs.org/en/

package.json
```diff
@@ -105,8 +105,8 @@
         "inquirer": "6.3.1"
     },
     "engines": {
-        "node": ">=12.16.1",
-        "npm": ">=6.13.4"
+        "node": ">=10.20.1",
```
pascalgrimaud, line 108: this change is the good one. This PR should only change these 2 values to be less restricted

generators/generator-constants.js
mshima: @SudharakaP this constant is used for creating CI. Should be kept 12.16.1
Add:
`const PACKAGE_JSON_NODE_VERSION = packagejs.engines.node;`
And use it on the package.json templates

package.json
SudharakaP: @pascalgrimaud : I've reverted the generator node version back to 12.16.1. Could you please let me know if everything looks okay now? I haven't changed the template or CI node versions [as described in the above comment](https://github.com/jhipster/generator-jhipster/pull/11687#discussion_r420035667) there seem to be differing opinions on that. If that is also needed let me know.

pascalgrimaud: good to go, thanks :)

PR #11720, opened 2020-05-06
Try to load config from mem-fs so it can load current working version.

mshima:
Fixes config synchronization with getAllJhipsterConfig.
When saving config with `this.config.set(‘foo’, ‘bar’);`
getAllJhipsterConfig will not read the updated config until next jhipster execution.


mshima:
@pascalgrimaud I think we should merge this for 6.9.0.
Otherwise should be closed, since we should remove getAllJhipsterConfig for 7.

mraible:
What's the alternative to `getAllJhipsterConfig()`? I'm [using it in Ionic for JHipster](https://github.com/oktadeveloper/generator-jhipster-ionic/blob/master/generators/app/index.js#L64).

mshima:
> What's the alternative to `getAllJhipsterConfig()`? I'm [using it in Ionic for JHipster](https://github.com/oktadeveloper/generator-jhipster-ionic/blob/master/generators/app/index.js#L64).

Normal `this.config`.

PR #11724, opened 2020-05-07
Set PostgreSQL as default database

agaspardcilia:
Not my biggest PR but it was mentioned in #10958.

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -142,16 +142,16 @@ const R2DBC_DB_OPTIONS = [
 
 const SQL_DB_OPTIONS = [
     {
+        value: 'postgresql',
+        name: 'PostgreSQL',
+    },
+    {
         value: 'mysql',
         name: 'MySQL',
     },
     {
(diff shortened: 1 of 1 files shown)


mraible:
@jhipster/developers Does this need to wait for v7?

DanielFran:
I added label v7 but for me this can be merged

pascalgrimaud:
In my opinion, it's mainly for people who want to try or do demo, so you can press "enter enter enter" with the cli. There is no breaking change.

Then, if we decide to switch to PostgreSQL by default, there are more work than simply this change:
- change all default config in CI to use and test more PostgreSQL
- do the same thing in jhipster-daily-builds
- change the behavior in jhipster-online
- do the change in all other blueprint backend (kotlin, nodejs, quarkus, etc.)

Because, if it becomes our default database, it must be tested more.

agaspardcilia:
Alright, I thought things mentioned in the road map were agreed upon, looks like there not... I'll close the PR for now and open a ticket to discuss that then.

PR #11725, opened 2020-05-07
Skip fake data as default behavior

agaspardcilia:
As mentioned in #10958, I didn't had a lot of information but I guess this was what `Disable fake data by default` meant.

Also, if nobody is having any problem with the argument I've modified, the doc will need to be updated.

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -81,9 +81,9 @@ module.exports = class extends BaseBlueprintGenerator {
             defaults: false,
         });
 
-        // This adds support for a `--skip-fake-data` flag
-        this.option('skip-fake-data', {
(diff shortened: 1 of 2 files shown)


murdos:
I was not aware this item was in the v7 roadmap. I don't why it was added, and I don't see a discussion about it in  #10958

I personally disagree with disabling this feature by default. 
While generated fake data are not perfect, and used to have some issues, I think they are useful to populate dev environment with some data.

agaspardcilia:
Alright, I'll put that on draft then.

agaspardcilia:
I thought it was something confirmed, my bad, I'll close that one.

PR #11729, opened 2020-05-07
update login regex to be more consistent with current released one

### LINKED ISSUE #11726: Login RegEx is a breaking change

[One recent PR](https://github.com/jhipster/generator-jhipster/pull/11450) changed the validation patterns logins. This regex has a problem.

* There is a potential breaking change. E.g. with the old regex `foo@bar@com` was a valid login. With the regex in master not. So data which is currently in the database and a valid login will be invalid now

It should not happen.

##### **Related issues**

None.

##### **Suggest a Fix**

Update the regex such that

* It behaves more like the old regex, such that breaking changes do not happen

##### **JHipster Version(s)**

Current master only. The was **NO** release with the breaking change, so no user is in danger.
 
atomfrede:
This makes the login regex more consistent with the current released version. In particular for example:

* Released version: `foo@bar@com` -> VALID :heavy_check_mark: 
* Current master: `foo@bar@com` -> INVALID :red_circle: 
* New regex: `foo@bar@com` -> VALID :heavy_check_mark: 

closes #11726

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -84,7 +84,7 @@ module.exports = class extends BaseBlueprintGenerator {
 
             setupClientconsts() {
                 // Make constants available in templates
-                this.LOGIN_REGEX = constants.LOGIN_REGEX;
+                this.LOGIN_REGEX = constants.LOGIN_REGEX_JS;
                 this.ANGULAR = constants.SUPPORTED_CLIENT_FRAMEWORKS.ANGULAR;
                 this.HUSKY_VERSION = constants.HUSKY_VERSION;
                 this.LINT_STAGED_VERSION = constants.LINT_STAGED_VERSION;
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -84,6 +84,7 @@ module.exports = class extends BaseBlueprintGenerator {
 
             setupClientconsts() {
                 // Make constants available in templates
+                this.LOGIN_REGEX = constants.LOGIN_REGEX;
                 this.ANGULAR = constants.SUPPORTED_CLIENT_FRAMEWORKS.ANGULAR;
                 this.HUSKY_VERSION = constants.HUSKY_VERSION;
                 this.LINT_STAGED_VERSION = constants.LINT_STAGED_VERSION;
--- a/generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
@@ -42,7 +42,7 @@ export class RegisterComponent implements AfterViewInit {
     success = false;
 
     registerForm = this.fb.group({
-        login: ['', [Validators.required, Validators.minLength(1), Validators.maxLength(50), Validators.pattern('<%- LOGIN_REGEX %>')]],
+        login: ['', [Validators.required, Validators.minLength(1), Validators.maxLength(50), Validators.pattern(/<%- LOGIN_REGEX %>/)]],
         email: ['', [Validators.required, Validators.minLength(5), Validators.maxLength(254), Validators.email]],
         password: ['', [Validators.required, Validators.minLength(4), Validators.maxLength(50)]],
         confirmPassword: ['', [Validators.required, Validators.minLength(4), Validators.maxLength(50)]]
--- a/generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/register/register.component.ts.ejs
@@ -42,7 +42,7 @@ export class RegisterComponent implements AfterViewInit {
     success = false;
 
     registerForm = this.fb.group({
-        login: ['', [Validators.required, Validators.minLength(1), Validators.maxLength(50), Validators.pattern('<%= LOGIN_REGEX %>')]],
+        login: ['', [Validators.required, Validators.minLength(1), Validators.maxLength(50), Validators.pattern('<%- LOGIN_REGEX %>')]],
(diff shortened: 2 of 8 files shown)


DanielFran:
@atomfrede Maybe it is better to define a constant and use it everywhere?

atomfrede:
> @atomfrede Maybe it is better to define a constant and use it everywhere?
> Probably on Vuejs you can reuse the same constant, this will avoid to make the changes on other blueprints?

Done.

atomfrede:
There is a problem with the javascript regex when it is generated. Taking a look now.

DanielFran:
@atomfrede It seems that it was affecting only react... I restarted them 1hour ago

atomfrede:
@DanielFran Thanks, but the generated regex contained invalid characters, adapted the escaping, should work now correct what I can tell from my local tests.

ruddell:
Angular does not have E2E tests for the registration page but React does, which explains the earlier failing tests. 

-------
The two regexes return different results for JS vs Java.  For example, the below value is valid for `login` in the JS (no validation message displayed), but Java returns an error for not matching the pattern:

    user!

For that failure, a generic `Registration failed! Please try again later` error is displayed to the register page, and a `translation-not-found[error.Pattern]` is displayed on the admin user management edit page.

-------
- Create a user `test#@localhost` 
- Click the edit button in user management for `test#@localhost`
    - It fails to find the user `test` and sends you to a 404 page
- Create a user `test`
- Click the edit button in user management for `test#@localhost`
    - It takes you to the `test` edit page instead of the `test#@localhost` edit page

 I'm guessing it's because the user login is used in the URL (tested with Angular).

atomfrede:
Thanks @ruddell for checking. Sadly js does not support atomic groups, so the workaround with the lookahead seems not to perfect.

atomfrede:
Okay, having the correct regex in js should work. But, we need to have `\1` in the script. Which seem to be forbidden in strict mode (octal numbers are not allowed). Has anyone an idea how to make that possible without disabling strict mode?

This should be the resulting regex:
`^(?=([_.@A-Za-z0-9-]+))\1|(?=([a-zA-Z0-9!$&*+=?^_`{|}~.-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*))\1$`. 
But it won't compile because of `\1` seen as octal number

atomfrede:
So just to say the regex engine of javascript is *****

pascalgrimaud:
really strange behavior with javascript... indeed !

pascalgrimaud:
Unless someone wants to review it, I'll merge it tonight

ruddell:
Just tested again with the latest changes (both Angular and React). The first issue in my last comment is still happening. `user!` is never marked as invalid on the client-side. I can't seem to find any invalid value for the client-side pattern, so server-side validation is always returning an error.

atomfrede:
Strange. Let me check whats the difference to my manual adapted application,

atomfrede:
@ruddell If you have spare time can you try again?

ruddell:
There are still some minor differences between the Java and JS regexes, for example `test#a` and `test$a`  are valid to the JS regex (so no client-side validation) but invalid to the Java regex. It works as expected for other cases I tested.

I still think we shouldn't overcomplicate the solution, especially if we aren't following the email spec exactly anymore, but you've done all the work so I'm fine either way (and you're pretty close to a good solution).

atomfrede:
The problem with js is, that the test method is stateful. Calling test on the same regex with the same string may yield different results. It is documented, but very strange compared to java or other languages imho,

ruddell:
The regex `test` method should only be stateful if you add `/g` or /`y` to the regex ([docs for global flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#Using_test_on_a_regex_with_the_global_flag) mention this). We don't pass a global flag in the regex, so it shouldn't affect us.

`user#test` always returns as valid.  I can't reproduce what you have in your GIF without passing in a regex with `/g` at the end (instead of string).  When I test the latest generated code, it works correctly except for `<valid-char><invalid-char><valid-char>`. Here's a [screen capture](https://www.loom.com/share/77150b148084435bb4922d03cba4f4cd), it shows the difference between passing the pattern as String, Regex (same as String), and Regex with global flag (matches your GIF).

I'm not a regex expert though so I may be missing something.

atomfrede:
Yes, you are right, we don't pass a global flag anymore, my bad.

ruddell:
Just saw your edit. You're not stupid, regex is just a pain to get right lol.

If there is a match, it returns true.  If you use `^[a-z]+$` instead of `[a-z]+`, then it returns false like you expected.  The `^` and `$` make it match the whole line of text rather than just part (`user1` would have to match instead of just the partial `user`).  

The issue is the `|` operator, we need `^` and `$` surrounding both statements. Currently we have it around the entire statement.  [Current regex with tests](https://regexr.com/54f68), and [modified regex with tests](https://regexr.com/54f65) to have `^` and `$` surrounding both statements.

With that change it works as expected.
```
const str = 'user1';

const regex1 = RegExp('[a-z]+');
const regex2 = RegExp('^[a-z]+$');
const regex3 = RegExp('^[a-z]+|[a-z]+$');
const regex4 = RegExp('^[a-z]+$|^[a-z]+$');

console.log(regex1.test(str)); //true
console.log(regex2.test(str)); //false
console.log(regex3.test(str)); //true
console.log(regex4.test(str)); //false
```

atomfrede:
@ruddell :facepalm: Of course! So I would say, at least I am a bit stupid! Will change the frontend part! Thanks so much!

atomfrede:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/18385

PR #11742, opened 2020-05-08
[webflux] Reactive sql entity generator

gzsombor:

### DIFF
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -103,7 +103,7 @@ const serverFiles = {
         },
         {
             condition: generator =>
-                (!generator.reactive || !['mongodb', 'cassandra', 'couchbase', 'neo4j', 'sql'].includes(generator.databaseType)) &&
+                !generator.reactive &&
                 !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
@@ -116,7 +116,6 @@ const serverFiles = {
         {
             condition: generator =>
                 generator.reactive &&
-                ['mongodb', 'cassandra', 'couchbase', 'neo4j', 'sql'].includes(generator.databaseType) &&
                 !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -102,9 +102,7 @@ const serverFiles = {
             ],
         },
         {
-            condition: generator =>
-                !generator.reactive &&
-                !generator.embedded,
+            condition: generator => !generator.reactive && !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 {
@@ -114,9 +112,7 @@ const serverFiles = {
             ],
         },
         {
-            condition: generator =>
-                generator.reactive &&
-                !generator.embedded,
+            condition: generator => generator.reactive && !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 {
@@ -126,19 +122,18 @@ const serverFiles = {
             ],
         },
         {
-            condition: generator =>
-                generator.reactive && ['sql'].includes(generator.databaseType) && !generator.embedded,
+            condition: generator => generator.reactive && ['sql'].includes(generator.databaseType) && !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 {
                     file: 'package/repository/EntityReactiveRepositoryInternalImpl.java',
-                    renameTo: generator => `${generator.packageFolder}/repository/${generator.entityClass}RepositoryInternalImpl.java`
+                    renameTo: generator => `${generator.packageFolder}/repository/${generator.entityClass}RepositoryInternalImpl.java`,
                 },
                 {
                     file: 'package/repository/rowmapper/EntityRowMapper.java',
-                    renameTo: generator => `${generator.packageFolder}/repository/rowmapper/${generator.entityClass}RowMapper.java`
-                }
-            ]
+                    renameTo: generator => `${generator.packageFolder}/repository/rowmapper/${generator.entityClass}RowMapper.java`,
+                },
+            ],
         },
         {
             condition: generator => generator.service === 'serviceImpl' && !generator.embedded,
--- a/generators/entity-server/files.js
+++ b/generators/entity-server/files.js
@@ -122,7 +122,7 @@ const serverFiles = {
             ],
         },
         {
-            condition: generator => generator.reactive && ['sql'].includes(generator.databaseType) && !generator.embedded,
+            condition: generator => generator.reactive && generator.databaseType === 'sql' && !generator.embedded,
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 {
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -116,7 +116,7 @@ module.exports = class extends BaseBlueprintGenerator {
             if (rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true)) {
                 const colName = this._getJoinColumnName(rel);
                 joins.push(
-                    ` LEFT JOIN ${rel.otherEntityTableName} ${rel.relationshipName} ON entity.${colName} = ${rel.relationshipName}.id`
+                    ` LEFT JOIN ${rel.otherEntityTableName} \\"${rel.relationshipName}\\" ON entity.${colName} = \\"${rel.relationshipName}\\".id`
                 );
             }
         });
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -124,7 +124,7 @@ module.exports = class extends BaseBlueprintGenerator {
     }
 
     _getUseMapsIdRelation(relationships) {
-        const jpaDerivedRelation = Object.values(relationships).find(rel => rel.useJPADerivedIdentifier);
+        const jpaDerivedRelation = Object.values(relationships).find(rel => rel.useJPADerivedIdentifier === true);
         if (jpaDerivedRelation) {
             return { isUsingMapsId: true, mapsIdAssoc: jpaDerivedRelation };
         }
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -81,34 +81,37 @@ module.exports = class extends BaseBlueprintGenerator {
                 this.importSet = false;
                 this.uniqueEnums = {};
 
-                this.importApiModelProperty = Object.values(this.relationships).filter(v => typeof v.javadoc != 'undefined').length > 0;
+                this.importApiModelProperty = this.relationships.filter(v => typeof v.javadoc != 'undefined').length > 0;
                 if (!this.importApiModelProperty) {
-                    this.importApiModelProperty = Object.values(this.fields).filter(v => typeof v.javadoc != 'undefined').length > 0;
+                    this.importApiModelProperty = this.fields.filter(v => typeof v.javadoc != 'undefined').length > 0;
                 }
 
-                Object.values(this.relationships).forEach(v => {
-                    if (v.ownerSide === false && ['one-to-many', 'one-to-one', 'many-to-many'].includes(v.relationshipType)) {
+                this.relationships.forEach(relationship => {
+                    if (
+                        relationship.ownerSide === false &&
+                        ['one-to-many', 'one-to-one', 'many-to-many'].includes(relationship.relationshipType)
+                    ) {
                         this.importJsonIgnore = true;
-                    } else if (v.relationshipType === 'many-to-one') {
+                    } else if (relationship.relationshipType === 'many-to-one') {
                         this.importJsonIgnoreProperties = true;
                     }
-                    if (v.relationshipType === 'one-to-many' || v.relationshipType === 'many-to-many') {
+                    if (relationship.relationshipType === 'one-to-many' || relationship.relationshipType === 'many-to-many') {
                         this.importSet = true;
                     }
                 });
 
-                Object.values(this.fields).forEach(v => {
+                this.fields.forEach(field => {
                     if (
-                        v.fieldIsEnum &&
-                        (!this.uniqueEnums[v.fieldType] || (this.uniqueEnums[v.fieldType] && v.fieldValues.length !== 0))
+                        field.fieldIsEnum &&
+                        (!this.uniqueEnums[field.fieldType] || (this.uniqueEnums[field.fieldType] && field.fieldValues.length !== 0))
                     ) {
-                        this.uniqueEnums[v.fieldType] = v.fieldType;
+                        this.uniqueEnums[field.fieldType] = field.fieldType;
                     }
                 });
             },
 
             useMapsIdRelation() {
-                const jpaDerivedRelation = Object.values(this.relationships).find(rel => rel.useJPADerivedIdentifier === true);
+                const jpaDerivedRelation = this.relationships.find(rel => rel.useJPADerivedIdentifier === true);
                 if (jpaDerivedRelation) {
                     this.isUsingMapsId = true;
                     this.mapsIdAssoc = jpaDerivedRelation;
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -32,6 +32,7 @@ module.exports = class extends BaseBlueprintGenerator {
         this.jhipsterContext = opts.jhipsterContext || opts.context;
 
         this.testsNeedCsrf = ['uaa', 'oauth2', 'session'].includes(this.jhipsterContext.authenticationType);
+        this.officialDatabaseType = constants.OFFICIAL_DATABASE_TYPE_NAMES[this.jhipsterContext.databaseType];
 
         useBlueprints = !this.fromBlueprint && this.instantiateBlueprints('entity-server', { context: opts.context });
     }
@@ -101,4 +102,27 @@ module.exports = class extends BaseBlueprintGenerator {
         });
         return { importApiModelProperty, importJsonIgnore, importJsonIgnoreProperties, importSet, uniqueEnums };
     }
+
+    _generateTableJoins(relationships) {
+        let joins = [];
+        for (idx in relationships) {
+            let rel = relationships[idx];
+            if (rel.relationshipType === 'many-to-one' || rel.relationshipType === 'one-to-one') {
+                joins.push(" LEFT JOIN " + rel.otherEntityTableName + " " + rel.relationshipName + " ON entity." + this.getColumnName(rel.relationshipName) + "_id = " + rel.relationshipName + ".id");
+            }
+        }
+        return joins;
+    }
+
+    _generateEagerRelationsAndEntityTypes(entityClass, relationships) {
+        let eagerRelations = relationships.filter(function(rel) {
+            return rel.relationshipType === 'many-to-one' || rel.relationshipType === 'one-to-one';
+        });
+        let uniqueEntityTypes = new Set(eagerRelations.map(function(rel) {
+            return rel.otherEntityNameCapitalized;
+        }));
+        uniqueEntityTypes.add(entityClass);
+        return { eagerRelations, uniqueEntityTypes };
+    }
+
 };
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -127,16 +127,10 @@ module.exports = class extends BaseBlueprintGenerator {
     }
 
     _generateEagerRelationsAndEntityTypes(entityClass, relationships) {
-        const eagerRelations = relationships.filter(function (rel) {
-            return rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true);
-        });
-        const regularEagerRelations = eagerRelations.filter(function (rel) {
-            return rel.useJPADerivedIdentifier !== true;
-        });
+        const eagerRelations = relationships.filter(rel => rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true));
+        const regularEagerRelations = eagerRelations.filter(rel => rel.useJPADerivedIdentifier !== true);
         const uniqueEntityTypes = new Set(
-            eagerRelations.map(function (rel) {
-                return rel.otherEntityNameCapitalized;
-            })
+            eagerRelations.map(rel => rel.otherEntityNameCapitalized)
         );
         uniqueEntityTypes.add(entityClass);
         return { eagerRelations, uniqueEntityTypes, regularEagerRelations };
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -21,6 +21,7 @@ const constants = require('../generator-constants');
 const writeFiles = require('./files').writeFiles;
 const utils = require('../utils');
 const BaseBlueprintGenerator = require('../generator-base-blueprint');
+const { isReservedTableName } = require('../../jdl/jhipster/reserved-keywords');
 
 /* constants used throughout */
 let useBlueprints;
@@ -108,6 +109,13 @@ module.exports = class extends BaseBlueprintGenerator {
             return 'id';
         }
         return `${this.getColumnName(relationship.relationshipName)}_id`;
+    }
+
+    _generateSqlSafeName(name) {
+        if (isReservedTableName(name, 'sql')) {
+            return `e_${name}`;
+        }
+        return name;
     }
 
     _getUseMapsIdRelation(relationships) {
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -106,9 +106,8 @@ module.exports = class extends BaseBlueprintGenerator {
     _getJoinColumnName(relationship) {
         if (relationship.useJPADerivedIdentifier === true) {
             return 'id';
-        } else {
-            return this.getColumnName(relationship.relationshipName) + '_id';
         }
+        return `${this.getColumnName(relationship.relationshipName)}_id`;
     }
 
     _generateTableJoins(relationships) {
@@ -125,10 +124,9 @@ module.exports = class extends BaseBlueprintGenerator {
     }
 
     _getUseMapsIdRelation(relationships) {
-        for (idx in relationships) {
-            if (relationships[idx].useJPADerivedIdentifier === true) {
-                return { isUsingMapsId: true, mapsIdAssoc: relationships[idx] };
-            }
+        const jpaDerivedRelation = Object.values(relationships).find(rel => rel.useJPADerivedIdentifier);
+        if (jpaDerivedRelation) {
+            return { isUsingMapsId: true, mapsIdAssoc: jpaDerivedRelation };
         }
         return { isUsingMapsId: false, mapsIdAssoc: null };
     }
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -124,6 +124,15 @@ module.exports = class extends BaseBlueprintGenerator {
         return joins;
     }
 
+    _getUseMapsIdRelation(relationships) {
+        for (idx in relationships) {
+            if (relationships[idx].useJPADerivedIdentifier === true) {
+                return { isUsingMapsId: true, mapsIdAssoc: relationships[idx] };
+            }
+        }
+        return { isUsingMapsId: false, mapsIdAssoc: null };
+    }
+
     _generateEagerRelationsAndEntityTypes(entityClass, relationships) {
         const eagerRelations = relationships.filter(function (rel) {
             return rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true);
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -103,13 +103,21 @@ module.exports = class extends BaseBlueprintGenerator {
         return { importApiModelProperty, importJsonIgnore, importJsonIgnoreProperties, importSet, uniqueEnums };
     }
 
+    _getJoinColumnName(relationship) {
+        if (relationship.useJPADerivedIdentifier === true) {
+            return 'id';
+        } else {
+            return this.getColumnName(relationship.relationshipName) + '_id';
+        }
+    }
+
     _generateTableJoins(relationships) {
         const joins = [];
         Object.values(relationships).forEach(rel => {
             if (rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true)) {
-                const colName = this.getColumnName(rel.relationshipName);
+                const colName = this._getJoinColumnName(rel);
                 joins.push(
-                    ` LEFT JOIN ${rel.otherEntityTableName} ${rel.relationshipName} ON entity.${colName}_id = ${rel.relationshipName}.id`
+                    ` LEFT JOIN ${rel.otherEntityTableName} ${rel.relationshipName} ON entity.${colName} = ${rel.relationshipName}.id`
                 );
             }
         });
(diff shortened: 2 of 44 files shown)


mraible:
Excellent work @gzsombor! Can you please add a CI test for this so we know it's working?

mraible:
@pascalgrimaud Any chance we can hold release 6.9 for this PR? I'd be happy to help get it merge-worthy!

pascalgrimaud:
ok @mraible
ping me once it's ready

generators/entity/index.js
```diff
@@ -738,6 +738,9 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 if (!context.relationships) {
                     context.relationships = [];
                 }
+                var humanDatabaseNames = {"mongodb": 'MongoDB', 'cassandra': 'Cassandra', 'sql': 'SQL', 'couchbase': 'Couchbase', 'neo4j': 'Neo4j'};
```
murdos, line 741: This should be factored with `server/prompts.js` and probably moved to `generator-constants.js`, but it should be a different PR. 
Meanwhile, please move it to `entity-server` since there's no use for `entity-client` so it doesn't need to be in `entity` subgen. And I suggest `officialDatabaseNames`, because `humanDatabaseNames` is a bit weird. Or `humanReadableDatabaseNames` but that's longer.

generators/generator-base.js
```diff
@@ -1451,6 +1451,17 @@ module.exports = class extends PrivateBase {
         return joinTableName;
     }
 
+    generateTableJoins(relationships) {
```
murdos, line 1454: Could you move it in `entity-server` subgen, prefixed with `_` so it can be used in templates?

mraible:
@gzsombor There's a conflict in `Entity.java.ejs`. Can you please take a look?

gzsombor:
There are a couple of things, which is still an open question for me, and as here, we try to use r2dbc which is much lower level than Hibernate, there are multiple things, which is pain.
Basically, it seems, that multiple things behave a little bit differently between the async drivers itself, and the async drivers vs sync drivers. As I mostly tried to fix the integration tests, currently I can report mostly the H2 issues:
* blob/clob fields returned as a custom io.r2dbc.spi.Blob/Clob field, which is a Reactive producer - we can consume it, and get the result byte[]/char[] but it's a bit ugly, and probably not the thing what the user wants. Currently, I've modified the integration tests to use longvarchar for H2 instead of clob, to simplify my life :wink: 
* The reactive Spring Data doesn't support tables, without an autoincrement field. The repository.save(..) method just checks, if the id field contains a value, if not, issues an insert, if contains, issues an update. So even if we just get a value from a sequence, and put that into the entity - Spring Data will just call the update :disappointed: So, we have to generate custom inserts - or wait, until Spring Data will support it. And we need the code to get values from a sequence, reactively. Currently, I've just removed the possibility to use PostgreSQL/MSSQL/Oracle :cry: 
* the conversions between the various date and time types is a bit ad-hoc. For H2 I've added a couple of new conversions, but I'm a bit afraid, that for MariaDB/MySQL/etc we will need to add more. 
* Currently, the code generates left joins to load efficiently the many-to-one and one-to-one relations, but not the one-to-many or many-to-many relations, which match with the Hibernate behavior, however the misleadingly named findOneWithEagerRelationships is not yet implemented. It shouldn't be too difficult, I think everything is there to do it. 
* Similarly, the optimization, when two table shares their primary key - when @ MapsId is used for Hibernate - is not yet handled.
* Entity filtering is also skipped.

murdos:
> we try to use r2dbc which is much lower level than Hibernate, there are multiple things, which is pain.

This is exactly why I was waiting for Spring Boot 2.3 and last Spring Data R2DBC release before trying to handle this task, while still knowing that a lot of things would be hard.

cbornet:
For the issue with the save() method, I had the same issue with the User entity. My workaround was to use a distinct create() method when I wanted to force an INSERT statement. See https://github.com/jhipster/generator-jhipster/blob/a9549c781d593995b877fc0315f8572066324b47/generators/server/templates/src/main/java/package/repository/UserRepository.java.ejs#L326

cbornet:
As for the datetime converters, I think LocalDateTime is the only type supported by all drivers. So you'll probably need converters for OffsetDateTime for other DBs.

mp911de:
@gzsombor here a few thoughts about the items you mentioned:

* True blob/clob columns typically map to `Blob`/`Clob` types to avoid heap exhaustion. It makes sense to think about these columns as if they would hold a 2-gigabyte value and design the model accordingly.
* Mixed-flavor types (`TEXT`, `VARBINARY`) column types can be consumed/provided either as `String`/`Clob` respective `ByteBuffer`/`Blob`. Per spec, those should return a scalar value when called through `Row.get(identifier)`. If that's not the case, please raise a bug report against the affected driver.
* Spring Data R2DBC provides three interfaces when it comes to database interaction: Pure SQL via `DatabaseClient`, `insert`/`update` methods, and the Repository API. The Repository interface tries to discover from a given object whether it is new or whether it should exist in the database. This happens by default by inspecting the `@Id` property. You can also implement the `Persistable` interface to supply the information from your application. Please note that Postgres allows auto-generated columns through the `SERIAL` column type.
See the [reference documentation](https://docs.spring.io/spring-data/r2dbc/docs/1.1.0.RC1/reference/html/#r2dbc.entity-persistence.state-detection-strategies) for further reference. On a related note, Spring Data JDBC implements the same strategy.

mraible:
The `webflux-psql` job is failing with the following error:

```
[ERROR] Tests run: 11, Failures: 1, Errors: 5, Skipped: 0, Time elapsed: 0.361 s <<< FAILURE! - in io.github.jhipster.sample.web.rest.BankAccountResourceIT
[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.updateBankAccount  Time elapsed: 0.021 s  <<< ERROR!
org.springframework.dao.DataIntegrityViolationException: 
executeMany; SQL [INSERT INTO bank_account (name, guid, bank_number, agency_number, last_operation_duration, mean_operation_duration, mean_queue_duration, balance, opening_day, last_operation_date, active, account_type, attachment, attachment_content_type, description) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)]; NULL not allowed for column "ID"; SQL statement:
INSERT INTO bank_account (name, guid, bank_number, agency_number, last_operation_duration, mean_operation_duration, mean_queue_duration, balance, opening_day, last_operation_date, active, account_type, attachment, attachment_content_type, description) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15) [23502-200]; nested exception is io.r2dbc.spi.R2dbcDataIntegrityViolationException: [23502] [23502] NULL not allowed for column "ID"; SQL statement:
```

I was able to fix this by changing `*_added_entity_BankAccount.xml` to add autoIncrement:

```
<column name="id" type="bigint" autoIncrement="${autoIncrement}">
```

However, now there seems to be an issue with `BigDecimal` and formatting. I'm not sure how to solve it.

```
[ERROR] Tests run: 11, Failures: 5, Errors: 0, Skipped: 0, Time elapsed: 77.291 s <<< FAILURE! - in io.github.jhipster.sample.web.rest.BankAccountResourceIT
[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.updateBankAccount  Time elapsed: 0.736 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <2.00>
to be equal to:
 <2>
but was not.
        at io.github.jhipster.sample.web.rest.BankAccountResourceIT.updateBankAccount(BankAccountResourceIT.java:400)

[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.getBankAccount  Time elapsed: 0.612 s  <<< FAILURE!
java.lang.AssertionError: 
JSON path "$.balance"
Expected: is <1>
     but: was <1.0>
        at io.github.jhipster.sample.web.rest.BankAccountResourceIT.getBankAccount(BankAccountResourceIT.java:338)

[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.getAllBankAccounts  Time elapsed: 0.219 s  <<< FAILURE!
java.lang.AssertionError: 
JSON path "$.[*].balance"
Expected: a collection containing <1>
     but: mismatches were: [was <1.0>]
        at io.github.jhipster.sample.web.rest.BankAccountResourceIT.getAllBankAccounts(BankAccountResourceIT.java:308)

[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.createBankAccount  Time elapsed: 0.152 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <1.00>
to be equal to:
 <1>
but was not.
        at io.github.jhipster.sample.web.rest.BankAccountResourceIT.createBankAccount(BankAccountResourceIT.java:180)

[ERROR] io.github.jhipster.sample.web.rest.BankAccountResourceIT.getAllBankAccountsAsStream  Time elapsed: 0.244 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <1.0>
to be equal to:
 <1>
but was not.
        at io.github.jhipster.sample.web.rest.BankAccountResourceIT.getAllBankAccountsAsStream(BankAccountResourceIT.java:278)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   BankAccountResourceIT.createBankAccount:180 
Expecting:
 <1.00>
to be equal to:
 <1>
but was not.
[ERROR]   BankAccountResourceIT.getAllBankAccounts:308 JSON path "$.[*].balance"
Expected: a collection containing <1>
     but: mismatches were: [was <1.0>]
[ERROR]   BankAccountResourceIT.getAllBankAccountsAsStream:278 
Expecting:
 <1.0>
to be equal to:
 <1>
but was not.
[ERROR]   BankAccountResourceIT.getBankAccount:338 JSON path "$.balance"
Expected: is <1>
     but: was <1.0>
[ERROR]   BankAccountResourceIT.updateBankAccount:400 
Expecting:
 <2.00>
to be equal to:
 <2>
but was not.
[INFO] 
[ERROR] Tests run: 11, Failures: 5, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

mraible:
NOTE: There is an error when using the `sql` entity set:

```
Found the .jhipster/MapsIdUserProfileWithDTO.json configuration file, entity can be automatically generated!

The entity MapsIdUserProfileWithDTO is being updated.

Found the .jhipster/FieldTestEnumWithValue.json configuration file, entity can be automatically generated!

The entity FieldTestEnumWithValue is being updated.

WARNING! This entity has the DTO option, and it has a relationship with entity "operation" that doesn't have the DTO option. This will result in an error.

KeyStore 'src/main/resources/config/tls/keystore.p12' generated successfully.
```

mraible:
I tested the `webflux-mysql` test too and it fails because of the description mapping in `BankAccount`:

```
./mvnw test -Dtest=BankAccountResourceIT -P'!webpack,dev'
```

Results in:

```
org.springframework.data.mapping.MappingException: Could not read property 
@org.springframework.data.relational.core.mapping.Column(value="description", keyColumn="")
private java.lang.String io.github.jhipster.sample.domain.BankAccount.description from result set!
        at org.springframework.data.r2dbc.convert.MappingR2dbcConverter.readFrom(MappingR2dbcConverter.java:170)
        Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.FluxMapFuseable] :
```

This column is a `clob` according to Liquibase.

```xml
<column name="description" type="${clobType}">
    <constraints nullable="true" />
</column>
```

NOTE: For both our webflux-postgresql and webflux-mysql jobs, the dev database is H2, so we're not really testing PostgreSQL or MySQL. On the upside, we are testing and finding issues with H2!

gzsombor:
Yes, clob doesn't work with H2, because the driver doesn't convert it to
string, I tried a temporary solution, with setting clobType to longvarchar
for H2 - it works, but I didn't like it. (The driver - I think - would
prefer to have io.r2dbc.spi.Clob as type)
The BigDecimal problem - I think - comes from that we use decimal(21,2) and
the H2 driver sets the scale accordingly. But currently, the integration
tests assumes that returned values are integers - there are a couple of
.intValue() calls, here and there...
I thought about setting the test values to floating point everywhere (To
1.5 and 2.25 instead of the current 1 and 2), but it would affect all the
other configurations/databases - and indeed, it braked everything, so I
postponed that battle.
There is a new spring data r2dbc release (1.1.0-RELEASE), I've played with
that upgrade a little bit, the only issue is that we need to override a
couple of versions.

mraible:
If I try running the `BankAccountResourceIT#createBankAccount()` against MySQL using the `prod` profile, it fails too:

```
java.lang.IllegalArgumentException: Cannot encode value of type 'class java.util.UUID'
	at dev.miku.r2dbc.mysql.codec.DefaultCodecs.encode(DefaultCodecs.java:182)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	|_ checkpoint ⇢ SQL "INSERT INTO bank_account (name, guid, bank_number, agency_number, last_operation_duration, mean_operation_duration, mean_queue_duration, balance, opening_day, last_operation_date, active, account_type, attachment, attachment_content_type, description) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)" [DatabaseClient]
	|_ checkpoint ⇢ Handler io.github.jhipster.sample.web.rest.BankAccountResource#createBankAccount(BankAccountDTO) [DispatcherHandler]
Stack trace:
		at dev.miku.r2dbc.mysql.codec.DefaultCodecs.encode(DefaultCodecs.java:182)
```

generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
```diff
@@ -1097,27 +1123,92 @@ public class <%= entityClass %>ResourceIT <% if (databaseType === 'cassandra') {
...
+        // Check, that the count call also returns 0
+        webTestClient.get().uri("/api/<%= entityApiUrl %>/count?sort=id,desc&" + filter)
+            .accept(MediaType.APPLICATION_JSON)
+            .exchange()
+            .expectStatus().isOk()
+            .expectHeader().contentType(MediaType.APPLICATION_JSON)
+            .expectBody().json("0");
+    }
+<%_ } else { _%>
+    private void default<%= entityClass %>ShouldNotBeFound(String filter) throws Exception {
```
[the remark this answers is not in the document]
gzsombor: I really dont like to repeat that if, everywhere, because it adds too much noise. This file is already too complex

gzsombor:
I've made some progress on the entities, it looks, if we always call:
```        
try {
            // try, directly the driver
            return row.get(columnName, target);
} catch (Exception e) {
            Object obj = row.get(columnName);
            return useSpringConverters(obj, target);
}
```

It works a bit better - for example, the H2 driver already implements Clob -> String conversion. (Yes, I know, this is not the best solution, if your Clob is really a Large Object, but this is what we already doing with Hibernate/etc )
To be effective everywhere, I had to override the findById and findAll methods too on ReactiveCrudRepository too, which also has the benefit of getting all the x-to-one, eager relations loaded too.
The list of mostly constant helper classes are growning, after this PR will be merged, I think, they should be added to jhipster-framework.

gzsombor:
Because Spring Data is not able to handle entities with only null values - I agree, this is a very pathological case - but our TestServiceImpl (among others) integration tests were failing due to this reason - I had to generate a bit more low level code. But the progress is unstoppable!

generators/entity-server/files.js
```diff
@@ -196,6 +191,20 @@ const serverFiles = {
                 },
             ],
         },
+        {
+            condition: generator => generator.reactive && ['sql'].includes(generator.databaseType) && !generator.embedded,
+            path: SERVER_MAIN_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/repository/EntityReactiveRepositoryInternalImpl.java',
+                    renameTo: generator => `${generator.packageFolder}/repository/${generator.entityClass}RepositoryInternalImpl.java`,
```
murdos, line 200: The purpose of an internal repository is to be internal and thus hidden. So this class should not be public and should be in `${generator.entityClass}Repository.java` IMO.
Note that it could still be in a separate template, but it should be included in the other template so there's only one java class at end.

gzsombor:
Just to summarize the current state - despite the tests are failing, the state is not that bad - and far from terrible :wink: 
JHipster related issues:
* filtering is not yet implemented
* many-to-many relations and one-to-many relations are not yet handled (interestingly, this functionality it not yet covered by integration tests)
* Lot of code is generated, due to spring-data-* limitations, but most of it is static, so after merge, these could go to a jhipster-library. Including the field conversion support functions, which is also used in Couchbase, as I saw. 
* Currently, I've rebased on spring-boot-2.3 branch, so it is using the latest from Spring.
* It seems, that for Mariadb, there is a separate r2dbc Mariadb connector, for which, the version number is not inherited from any external pom. So, either we could go with using the Mysql connector, or we need to explicitly declare the supported version.
* handling databases, which doesn't have autoincrement field is not yet implemented, probably we can live without them. 
Bugs:
* there are a various spring-data-r2dbc issues, which I need to circumvent, unfortunately, there is one issue, which needs spring-data-r2dbc patch ( https://github.com/spring-projects/spring-data-r2dbc/pull/390 ), because there are entities - even in our test suite - which don't have any field in it - (just many-to-many relations), but spring-data-r2dbc throws an error when there is no value specified for an insert.

gzsombor:
Ok, this looks green now! And I think, it is not because I've disabled all the tests.
PostgreSql, MySQL and MariaDB is definitely works

mraible:
Are you OK if I merge this @gzsombor?

generators/entity-server/files.js
mraible: This seems to be the only unresolved comment on this PR. Can you please have a look @gzsombor?

generators/entity-server/index.js
```diff
@@ -101,4 +103,42 @@ module.exports = class extends BaseBlueprintGenerator {
...
+        if (jpaDerivedRelation) {
+            return { isUsingMapsId: true, mapsIdAssoc: jpaDerivedRelation };
+        }
+        return { isUsingMapsId: false, mapsIdAssoc: null };
+    }
+
+    _generateEagerRelationsAndEntityTypes(entityClass, relationships) {
+        const eagerRelations = relationships.filter(function (rel) {
+            return rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true);
+        });
```
murdos, line 132:
```suggestion
        const eagerRelations = relationships.filter(rel => rel.relationshipType === 'many-to-one' || (rel.relationshipType === 'one-to-one' && rel.ownerSide === true);
```

generators/entity-server/index.js
```diff
@@ -101,4 +103,42 @@ module.exports = class extends BaseBlueprintGenerator {
...
+        const regularEagerRelations = eagerRelations.filter(function (rel) {
+            return rel.useJPADerivedIdentifier !== true;
+        });
```
murdos, line 135:
```suggestion
        const regularEagerRelations = eagerRelations.filter(rel => rel.useJPADerivedIdentifier !== true);
```

generators/entity-server/index.js
```diff
@@ -101,4 +103,42 @@ module.exports = class extends BaseBlueprintGenerator {
...
+        const uniqueEntityTypes = new Set(
+            eagerRelations.map(function (rel) {
+                return rel.otherEntityNameCapitalized;
+            })
```
murdos, line 139:
```suggestion
        const uniqueEntityTypes = new Set(
            eagerRelations.map(rel => rel.otherEntityNameCapitalized)
```

generators/entity-server/templates/src/main/java/package/common/get_all_stream_template.ejs
```diff
@@ -19,16 +19,31 @@
 <%
 const mapper = entityInstance  + 'Mapper';
 const entityToDtoReference = mapper + '::' + 'toDto';
+const reactiveSql = (databaseType === 'sql' && reactive);
+let toResponseEntityPrefix = '';
+let toResponseEntityPostfix = '';
+if (pagination !== 'no') {
+    if (reactiveSql) {
+        toResponseEntityPostfix = '.map(ResponseEntity::ok)';
```
murdos, line 27: I've the feeling that it make the template more difficult to understand.
Sometimes a little duplication is not problematic if it makes things more maintainable.

generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
```diff
@@ -813,7 +872,19 @@ class <%= entityClass %>ResourceIT <% if (databaseType === 'cassandra') { %>exte
...
+                        if ((fields[idx].fieldType === 'byte[]' || fields[idx].fieldType === 'ByteBuffer') && fields[idx].fieldTypeBlobContent !== 'text') { %><% 
+                            if (databaseType === 'cassandra') { %>.array()<% } %>)<% } else 
+                                if (fields[idx].fieldType === 'Integer') { %><% } else 
+                                    if (fields[idx].fieldType === 'Long') { %>.intValue()<% } else 
+                                        if (fields[idx].fieldType === 'Float' || fields[idx].fieldType === 'Double') { %>.doubleValue()<% } else 
+                                            if (fields[idx].fieldType === 'BigDecimal') { %>)<% } else
+                                                if (fields[idx].fieldType === 'Boolean') { %>.booleanValue()<% } else 
+                                                    if (fields[idx].fieldType === 'ZonedDateTime') { %>)<% } else 
+                                                        if (fields[idx].fieldType !== 'String') { %>.toString()<% } %>))<%= !reactive ? ')' : '' %><%_ } _%>;
```
murdos, line 887: Very nice asciii art ! 😉

generators/entity-server/index.js
```diff
@@ -101,4 +103,42 @@ module.exports = class extends BaseBlueprintGenerator {
...
+        );
+        uniqueEntityTypes.add(entityClass);
+        return { eagerRelations, uniqueEntityTypes, regularEagerRelations };
+    }
```
mshima, line 143: I prefer to do preparation before templates, instead of calling functions there:
```
    _preparing() {
        return {
            setup() {
                this.joinColumName = this. _getJoinColumnName(this.relationships);
                this.derivatedRelationship = this.relationships.find(rel => rel.useJPADerivedIdentifier === true);
                this.isUsingMapsId = !!this.derivatedRelationship;
            },
        };
    }

    get preparing() {
        if (useBlueprints) return;
        return this._preparing();
    }
```

generators/server/templates/build.gradle.ejs
```diff
@@ -532,7 +532,11 @@ dependencies {
     <%_ } _%>
     <%_ if (prodDatabaseType === 'mariadb') { _%>
     implementation "org.mariadb.jdbc:mariadb-java-client"
+        <%_ if (!reactive) { _%>
     liquibaseRuntime "org.mariadb.jdbc:mariadb-java-client"
+        <%_ } else { _%>
+        implementation "org.mariadb:r2dbc-mariadb:${r2dbc_mariadb_version}"
```
murdos, line 538: R2DBC driver versions are managed by Spring Boot dependencies / R2DBC BOM
```suggestion
        implementation "org.mariadb:r2dbc-mariadb"
```

generators/server/templates/gradle.properties.ejs
```diff
@@ -51,6 +51,7 @@ liquibaseTaskPrefix=liquibase
 <%_ if (reactive) { _%>
 blockhound_junit_platform_version=1.0.4.RELEASE
 junit_platform_launcher_version=1.6.2
+r2dbc_mariadb_version=0.8.2-alpha2
```
murdos, line 54:
```suggestion
```

generators/server/templates/pom.xml.ejs
```diff
@@ -114,6 +114,9 @@
 <%_ if (reactive) { _%>
         <blockhound-junit-platform.version>1.0.4.RELEASE</blockhound-junit-platform.version>
 <%_ } _%>
+<%_ if (reactive && databaseType === 'sql') { _%>
+        <r2dbc-mariadb.version>0.8.2-alpha2</r2dbc-mariadb.version>
+<%_ } _%>
```
murdos, line 119:
```suggestion
```

generators/server/templates/pom.xml.ejs
```diff
@@ -426,19 +429,6 @@
             <scope>test</scope>
         </dependency>
 <%_ } _%>
-<%_ if (databaseType === 'sql' && reactive) { _%>
```
murdos, line 429: 👍

generators/server/templates/pom.xml.ejs
```diff
@@ -462,6 +452,13 @@
             <groupId>org.mariadb.jdbc</groupId>
             <artifactId>mariadb-java-client</artifactId>
         </dependency>
+        <%_ if (reactive) { _%>
+        <dependency>
+            <groupId>org.mariadb</groupId>
+            <artifactId>r2dbc-mariadb</artifactId>
+            <version>${r2dbc-mariadb.version}</version>
```
murdos, line 459:
```suggestion
```

generators/server/templates/src/test/java/package/ReactiveSqlTestContainerExtension.java.ejs
```diff
@@ -51,7 +51,6 @@ public class ReactiveSqlTestContainerExtension implements BeforeAllCallback {
             .withTmpFs(Collections.singletonMap("/testtmpfs", "rw"));
     <%_ } else if (prodDatabaseType === 'mssql') { _%>
     private static MSSQLServerContainer<?> container =  new MSSQLServerContainer<>("<%= DOCKER_MSSQL %>")
-            .withDatabaseName("<%= baseName %>")
```
murdos, line 54: Is there a specific reason for this change?

utils/entity.js
```diff
@@ -137,8 +137,7 @@ function prepareEntityForTemplates(entityWithConfig, generator) {
             : entityWithConfig.entityStateName
     );
 
-    entityWithConfig.reactiveRepositories =
-        entityWithConfig.reactive && ['mongodb', 'cassandra', 'couchbase', 'neo4j'].includes(entityWithConfig.databaseType);
+    entityWithConfig.reactiveRepositories = entityWithConfig.reactive;
```
murdos, line 140: I think (but probably in another PR since this once is quite big) that we should kill this `reactiveRepositories` config parameter, since it's now equivalent to  `reactive`

gzsombor: Yes, definitely!

generators/entity-server/index.js
gzsombor: Sorry, I was not aware of this complex way of doing template setup. Is the 'preparing()' and setup()' function name is somehow privileged?

generators/server/templates/src/test/java/package/ReactiveSqlTestContainerExtension.java.ejs
gzsombor: Yes, in the MSSQLServerContainer the withDatabaseName is not overriden ( https://github.com/testcontainers/testcontainers-java/blob/1.14.3/modules/mssqlserver/src/main/java/org/testcontainers/containers/MSSQLServerContainer.java ), so the super class's method is called: https://github.com/testcontainers/testcontainers-java/blob/48534069037e17147208e65c3c7d3d03e3c83a47/modules/jdbc/src/main/java/org/testcontainers/containers/JdbcDatabaseContainer.java#L89 - which throws an UnsupportedOperationException().

So it is never used before. Unfortunately, even after my change, there are other issues with the MSSQL support - but at least the tests are not failing in the startup.

generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
gzsombor: Thanks, I did my best :wink:

generators/entity-server/index.js
mshima: `preparing` is a custom priority. See https://github.com/jhipster/generator-jhipster/issues/12338.
https://github.com/jhipster/generator-jhipster/blob/9aee4a29381a1ebb52b83f13c8e7083620b5ef9b/generators/entity/index.js#L465-L481

Setup is the task name can be anything.

generators/server/templates/build.gradle.ejs
gzsombor: Unfortunately that dependency is only added recently, so it does not contain yet: https://github.com/r2dbc/r2dbc-bom/commit/3ee8980068c69bfdd6ad14d7d4bdc55b8e6b4ccc

generators/entity-server/templates/src/main/java/package/common/get_all_stream_template.ejs
gzsombor: Well, I think, adding more repetition doesn't make it more maintainable, or more readable. 
The real problem is that in these templates it's really hard to create and use bigger abstractions - without painful escaping.
I'm thinking about, something like this:
```

function wrapResponse(response) {
   if (pagination !== 'no') {
       if (reactiveSql) {
          return response + ".map(ResponseEntity::ok)";
       } else {
          return "new ResponseEntity<>(" + response + ".map(ResponseEntity::ok)";
       }
   }
   return response;
}

```

and in template, it would be nice, if we could wrote something like:
` <%= wrapResponse( %>....<% ); %>
`
or 
`<%wrapResponse%>.....</%wrapResponse%> `

But AFAIK, this is not possible :disappointed:

generators/entity-server/index.js
gzsombor: Ok, I see now, thanks!

generators/entity-server/index.js
```diff
@@ -71,34 +73,81 @@ module.exports = class extends BaseBlueprintGenerator {
...
+                this.importJsonIgnoreProperties = false;
+                this.importSet = false;
+                this.uniqueEnums = {};
+
+                this.importApiModelProperty = Object.values(this.relationships).filter(v => typeof v.javadoc != 'undefined').length > 0;
+                if (!this.importApiModelProperty) {
+                    this.importApiModelProperty = Object.values(this.fields).filter(v => typeof v.javadoc != 'undefined').length > 0;
+                }
+
+                Object.values(this.relationships).forEach(v => {
```
mshima, line 89: Great.
Rename `v` to `relationship`
```suggestion
                this.relationships.forEach(v => {
```

generators/entity-server/index.js
```diff
@@ -71,34 +73,81 @@ module.exports = class extends BaseBlueprintGenerator {
...
+                        this.importJsonIgnore = true;
+                    } else if (v.relationshipType === 'many-to-one') {
+                        this.importJsonIgnoreProperties = true;
+                    }
+                    if (v.relationshipType === 'one-to-many' || v.relationshipType === 'many-to-many') {
+                        this.importSet = true;
+                    }
+                });
+
+                Object.values(this.fields).forEach(v => {
```
mshima, line 100: Rename `v` to `relationship`.
```suggestion
               this.fields.forEach(v => {
```

generators/entity-server/index.js
gzsombor: Done

mshima (approved):

mshima:
@murdos I am doing some changes related to relationships that probably will conflict to this.
if you agree, I want to merge it before doing so.

murdos:
> 
> 
> @murdos I am doing some changes related to relationships that probably will conflict to this.
> if you agree, I want to merge it before doing so.

Sure! 👍

gzsombor:
Thanks!

Issue #11670, opened 2020-04-29
Core: Upgrade commander to v5


deepu105:
I upgraded commander from v2 to v4 in https://github.com/jhipster/generator-jhipster/pull/11669

V5 has lot of breaking changes and hence probably is considerable effort to upgrade to https://github.com/tj/commander.js/blob/master/CHANGELOG.md

PR #11769, opened 2020-05-12
Upgrade to commander 5

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/11670

-   Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

PR #11770, opened 2020-05-12
Create command for creating sample application.

mshima:
Closes https://github.com/jhipster/generator-jhipster/issues/11670


cli/cli.js
```diff
@@ -52,6 +54,8 @@ const sharedOptions = loadSharedOptions(blueprintsPackagePath) || {};
 // Env will forward sharedOptions to every generator
 Object.assign(env.sharedOptions, sharedOptions);
 
+program.storeOptionsAsProperties(false).passCommandToAction(false).version(version).usage('[command] [options]').allowUnknownOption();
```
deepu105, line 57: plz break lines for readability

cli/cli.js
```diff
@@ -72,38 +76,124 @@ const runYoCommand = (cmd, args, options, opts) => {
...
+            command.option(cmdString, value.description, value.default);
+        });
+    }
+
     command
         .allowUnknownOption()
         .description(opts.desc)
-        .action(args => {
-            const options = getCommandOptions(packageJson, process.argv.slice(2));
+        .action((first, second, third) => {
```
deepu105, line 115: are thses changes to support new commander version? looks a bit complicated

cli/utils.js
```diff
@@ -184,19 +176,18 @@ const getCommand = (cmd, args, opts) => {
...
-const getCommandOptions = (pkg, argv) => {
+const addKebabCase = (options = {}) => {
+    const kebabCase = Object.keys(options).reduce((acc, key) => {
+        acc[_.kebabCase(key)] = options[key];
+        return acc;
+    }, {});
+    return { ...kebabCase, ...options };
+};
+
+const getCommandOptions = (pkg, argv = []) => {
```
deepu105, line 187: add or update tests plz

cli/sample.js
```diff
@@ -0,0 +1,39 @@
...
+const https = require('https');
+const path = require('path');
+
+module.exports = (args, options, env) => {
+    const filename = args[0];
+    if (!fs.existsSync(filename)) {
+        fs.mkdirSync(filename);
+    }
+    const file = fs.createWriteStream(`${filename}/${filename}.jdl`);
+    https.get(`https://raw.githubusercontent.com/jhipster/jdl-samples/master/${filename}.jdl`, response => {
```
deepu105, line 29: add or update tests plz

cli/cli.js
```diff
@@ -72,38 +76,124 @@ const runYoCommand = (cmd, args, options, opts) => {
...
+            let args;
+            let cmdOptions;
+            let unknownArgs;
+            logger.debug(`first command arg: ${toString(first)}`);
+            logger.debug(`second command arg: ${toString(second)}`);
+            logger.debug(`third command arg: ${toString(third)}`);
+
+            if (opts.argument) {
```
deepu105, line 123: add or update tests plz

cli/cli.js
mshima: eslint complains about that

cli/cli.js
mshima: More or less. We were using commander, but parsing options manually, so:
* changed to use options parsed by the commander.
* registered options based on the generator options.
* help will work for the commands.
* commander will parse known options.

mshima:
@deepu105 sample command needs more error handling, i prefer to merge https://github.com/jhipster/generator-jhipster/pull/11769 before.

cli/utils.js
mshima: Added addKebabCase test, getCommandOptions is already updated.

cli/cli.js
mshima: done

cli/sample.js
mshima: done

cli/cli.js
mshima: Commander 5 changed the behaviour of the default command, so instead of relying on `program.on('command:*'` we must bail if an argument is found for app or forward to import-jdl.

deepu105:
Even better why don't we just add an alias called `jdl` for `import-jdl` and implement the sample stuff directly in existing import jdl sub gen, so you can run `jhipster jdl default`, IMO it makes more sense. WDYT @pascalgrimaud @jdubois @mraible @jhipster/developers ?

mraible:
> Even better why don't we just add an alias called `jdl` for `import-jdl` and implement the sample stuff directly in existing import jdl sub gen, so you can run `jhipster jdl default`, IMO it makes more sense. 

Sounds good to me!

mshima:
@deepu105 what should I do?
Add jdl alias and fetch samples if the file doesn’t exists?
Or do you have another plan for implementing this?

deepu105:
Lets go with jdl alias, so that we don't have to add a new command.

So the idea would be to first look for the specified file in local disk, if
not found look in samples repo. Also nice to look uo directly by given URL
as well to support custom repos

mraible:
I really like the idea of a jdl alias and lookup from samples. A full URL will allow for versioning if files too. 

I think `jhipster jdl default` would be cool because you could (in theory) change it by putting a default.jdl in samples. 

Another thing I thought of: what if a ? in a property's value caused a prompt? That would allow us to create JDLs that just prompt for buildTool, for example. Not sure how much work it'd be, just throwing the idea out there. 🙂

deepu105:
Ya that could be something we can build but would be a little bit of work due to how we handle yo-rc files now

cli/cli.js
```diff
@@ -72,38 +76,124 @@ const runYoCommand = (cmd, args, options, opts) => {
...
             } else {
                 const namespace = opts.blueprint ? `${packageNameToNamespace(opts.blueprint)}:${key}` : `${JHIPSTER_NS}:${key}`;
-                runYoCommand(namespace, program.args, options, opts);
+                runYoCommand(namespace, args, options, opts);
             }
         })
         .on('--help', () => {
             if (opts.help) {
-                logger.info(opts.help);
-            } else {
```
deepu105, line 104: The else was for getting help from the yo command, do you handle it somewhere else?

cli/jdl.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+    const promises = args.map(toJdlFile).map(filename => {
+        if (!fs.existsSync(filename)) {
+            let url;
+            try {
+                const urlObject = new URL(filename);
+                url = filename;
+                filename = path.basename(urlObject.pathname);
+            } catch (_error) {
+                url = new URL(filename, 'https://raw.githubusercontent.com/jhipster/jdl-samples/master/').toString();
```
deepu105, line 86: can we do a small extra check here to see if the filename ends with `.jdl` or `.jh` if not we just append `.jdl` to it so we can do `jhipster jdl default` and `jhipster jdl default.jdl`

test/cli/cli.spec.js
```diff
@@ -28,7 +31,7 @@ describe('jhipster cli test', () => {
     });
 
     it('should return error on unknown command', function (done) {
-        this.timeout(4000);
+        this.timeout(10000);
```
deepu105, line 34: was it intentional?

deepu105:
Great work @mshima :clap:

cli/cli.js
mshima: It was not working.
I am build the help with commander.

cli/jdl.js
mshima: Doing at line 33.

test/cli/cli.spec.js
mshima: Yep the test was failing.

mshima:
@deepu105 any news about this PR?
I waiting to start working on https://github.com/jhipster/generator-jhipster/issues/11815.
It's cli related too.

cli/commands.js
```diff
@@ -57,22 +62,61 @@ const defaultCommands = {
...
+            },
+            {
+                option: '--skip-db-changelog',
+                desc: 'Disable generation of database changelogs',
+                default: false,
+            },
+            {
+                option: '--inline <value>',
+                desc: 'Pass JDL content inline. Argument can be skipped when passing this',
+            },
```
MathieuAA, line 115: Instead of passing an inline JDL that can potentially override the sample entirely, this should be used to customize the pulled sample.
@mshima @deepu105 

And there shouldn't be any need for `--db` or `--blueprint/s`

mshima: This is a kickstart feature, the user don't need to know the jdl structure.

deepu105:
I'll test it this weekend

deepu105:
@mshima do you know why 2 tests are failing?

pascalgrimaud:
@deepu105 : flaky tests with React. We have this since months, I already mentionned it several times, but it comes back again and again.
It's because our React front is not maintained at same level than Angular or Vue.js

See https://github.com/jhipster/generator-jhipster/pull/11814 by @MathieuAA

cli/cli.js
```diff
@@ -72,38 +76,124 @@ const runYoCommand = (cmd, args, options, opts) => {
...
+            } else {
+                // Option that doesn't contain arguments
+                // first=cmdOptions second=unknownArgs
+                args = [];
+                cmdOptions = first;
+                unknownArgs = second || [];
+            }
+
+            const firstUnknownArg = Array.isArray(unknownArgs) && unknownArgs.length > 0 ? unknownArgs[0] : undefined;
+            if (key === 'app' && firstUnknownArg !== undefined && !firstUnknownArg.startsWith('-')) {
```
deepu105, line 158: why is unknown command handling done for `app` command, we don't have second level commands also I didn't see the didYouMean part working here, can you double check. The didYouMean hint should be for global unknown command like `jhipster entittty`

cli/jdl.js
```diff
@@ -0,0 +1,94 @@
...
+            .get(url, response => {
+                if (response.statusCode !== 200) {
+                    return reject(new Error(`Error downloading ${url}: ${response.statusCode} - ${response.statusMessage}`));
+                }
+
+                logger.debug(`Creating file: ${path.join(filename)}`);
+                const fileStream = fs.createWriteStream(`${filename}`);
+                fileStream.on('finish', () => {
+                    fileStream.close(() => {
+                        resolve(filename);
```
deepu105, line 51: This doesn't see right. Filestream.close doesn't take any args. My VSCode also complains
image

deepu105:
@mshima once the issues I mentioned above is fixed and tests pass this can be merged

cli/jdl.js
mshima: It exists but I could not find a doc for WriteStream.close().
https://github.com/nodejs/node/blob/49f549e064e92c1db0392994d8a4636726d38a75/lib/internal/fs/streams.js#L461-L468

mshima: But 'close' event is cleaner 👍

cli/cli.js
mshima: `If you use .command('*') to add a default command, you may be be able to use isDefault:true with a normal command instead.`
Commander 5 changed the behaviour and the `*` event is only emitted if there is no default command.
So `jhipster entittty` is treated like `jhipster app entittty`, since `app` doesn't takes any argument, we are good for didYouMean.
It makes sense to treat `jhipster foo.jdl` as `jhipster app foo.jdl` since app is the default command.

For jhipster 7 we could remove `.allowUnknownOption()`.
This will require every command and option to be registered, otherwise it will throw an exception.

mshima:
The only thing, I think we should keep `jhipster jdl` and `jhipster import-jdl` as separated commands, `import-jdl` for local files only.
For example when running CI, I don't think it should look for jdl at sample repository if the file is not found.

cli/cli.js
mshima: Can I close this review @deepu105 ?

deepu105: if you are getting did you mean suggestions then yes, it wasn't working for me. May be remove `allowUnknownOption` as its not a breaking change that would affect anyone

mshima: I didn't realised it wasn't working. I thought it was a design preference.
Fixed now.

mshima: About allowUnknownOption, we must be sure to cover blueprint's custom commands support.

mshima:
@deepu105 didYouMean is fixed.
Should we make import-jdl local only or keep as alias for jdl?
@jhipster/developers

mraible:
@mshima I think `import-jdl` as local-only is a good start.

mshima:
Heroku test suite doesn't pass with a previous load of child_process to node cache...

sendilkumarn:
> @deepu105 : flaky tests with React. We have this since months, I already mentionned it several times, but it comes back again and again.
> It's because our React front is not maintained at same level than Angular or Vue.js
> 
> See #11814 by @MathieuAA

@pascalgrimaud  Let me take a look at this.

MathieuAA:
I would love the help @sendilkumarn!

deepu105:
@mshima i would prefer to keep it as alias

mshima:
@deepu105 I really think a local only alternative is required mainly for CI.
Looks like @DanielFran @mraible and @MathieuAA reacted with let's do it or why not for keeping import-jdl as local only.

Alternatives would be
1 - Add option like --local-only-jdl (local-only will conflict with cli)
2 - Disable sample with CI env var and force to use absolute url.
3 - Add another command for local jdl.

Anyway let's ask @jhipster/developers 
❤️ for reverting last commit and make import-jdl a jdl alias
🚀 for keeping import-jdl as local-only.

mshima:
@deepu105 reverted the change.
I will keep the implementation in the log for reference.
We can handle CI in another PR.

MathieuAA:
@mshima the options aren't mutually exclusive, we can always choose the first and move to the second one later on

mshima:
You are saying to keep as an alias but use different behavior for each one?

deepu105:
Sorry but I really don't see a reason to have two commands doing the same thing with slightly different inputs. Its not elegant and will confuse end users. We can handle the CI in many different ways(very easy ways), its an internal issue and shouldn't affect user experience. Those in favour of having two commands please justify why coz I don't see any strong enough reason to do so and is a hard no from me unless there is a strong justification

MathieuAA:
There's also the maintenance cost, true. Alright. I had two concerns. The first was about not confusing the user with another command, which has almost the same name as one users were used to. We can remove the import-jdl, as long as there's a warning saying that this command is deprecated.
The expected behaviour could be something like that:
  - jhipster import-jdl toto.jdl
  - warning: the import-jdl command is deprecated in favor of <insert shiny new command here>
  - we automatically use the new sub-gen behind the curtains

My second concern is more about compatibility. The new command must be able to behave the same as the previous one (with a few exceptions though, let's not prevent us from cleaning up things if we want to).
If the two points can be addressed, I'm all in.

pascalgrimaud:
I'd prefer to keep import-jdl for v6, don't change command until v7.

mshima:
The only behavior change between import-jdl and jdl is that jdl looks for files at sample repository if the file is not found locally.

For commander an alias will have the same behavior.
To keep import-jdl as a alias to jdl I must add something like this:
```
if (process.argv[3] === 'import-jdl') {
  options['localOnlyJdl'] = true;
}
```
Then ignore sample repository when localOnlyJdl option is true at jdl command.
So instead of maintaining a command, we will be maintaining a workaround.

I really think both implementations are valid.

test/cli/jdl.js
```diff
@@ -143,10 +146,78 @@ describe('jdl command test', () => {
...
+        describe('when passing skipSampleRepository=true', () => {
+            describe('with local file argument', () => {
+                beforeEach(() => {
+                    sinon.stub(https, 'get');
+                });
+                afterEach(() => {
+                    https.get.restore();
+                });
+                it('should return file not found', () => {
+                    proxyquire('../../cli/jdl', {})(
```
mshima, line 158:
```suggestion
                    return proxyquire('../../cli/jdl', {})(
```

cli/cli.js
```diff
@@ -72,38 +76,126 @@ const runYoCommand = (cmd, args, options, opts) => {
...
+                const suggestion = didYouMean(cmd, availableCommands);
+                if (suggestion) {
+                    logger.info(`Did you mean ${chalk.yellow(suggestion)}?`);
+                }
+
+                logger.fatal(`${chalk.yellow(firstUnknownArg)} is not a known command. See '${chalk.white(`${CLI_NAME} --help`)}'.`);
+                return;
+            }
+
+            const customOptions = {};
```
deepu105, line 172: @mshima I have merged this for now, but honestly, I would like this part to be handled in a simpler way, by just adding a new flag for the command `--skip-sample-repo` and pass that flag in tests instead of doing this here so that both commands behave exactly the same, like an alias should be, can you do a new PR for that?

mshima: This should be jhipster 7 only https://github.com/jhipster/generator-jhipster/pull/11770#issuecomment-634892631.
But I think we should remove `import-jdl` instead.

deepu105: That comment is about the import-jdl command being removed nit about handling tests, tests are internal and we can handle it however we like

mshima: I understood @pascalgrimaud asked to keep import-jdl behavior unchanged for jhipster 6.

pascalgrimaud: Yes, I'd like to have the same current behavior for import-jdl.
I don't care if it can take a url instead of a file, as it would be an improvement. 
What I want is to keep:
jhipster import-jdl file.jdl : it imports all entities

Plz, don't change it now. No breaking change until v7.

deepu105: That behavior doesn't change and AFAIK there is no breaking change.

pascalgrimaud: thank you :)

mshima:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/19714

pascalgrimaud:
@mshima : approved

PR #11778, opened 2020-05-14
fix: jhipster version for marketplace

fdelbrayelle:

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -18,6 +18,7 @@
  */
 const chalk = require('chalk');
 const statistics = require('../statistics');
+const packagejs = require('../../package.json');
 
 module.exports = {
     askForInsightOptIn,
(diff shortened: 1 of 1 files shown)


generators/app/prompts.js
```diff
@@ -163,7 +163,7 @@ function askForMoreModules() {
 
 function askModulesToBeInstalled(done, generator) {
     generator.httpsGet(
-        'https://api.npms.io/v2/search?q=keywords:jhipster-module+jhipster-5&from=0&size=50',
+        'https://api.npms.io/v2/search?q=keywords:jhipster-module+jhipster-6&from=0&size=50',
```
murdos, line 166: I think we should compute this dynamically, using the major number from package.json version

fdelbrayelle: You're right, this is fixed!

PR #11779, opened 2020-05-14
Set VM to UTC

Blackdread:
See issue: https://github.com/jhipster/generator-jhipster/issues/11780


DanielFran:
@Blackdread Maybe it is better to remove some configuration on application property files, so the user only needs to take care about what is configured on the main?

Blackdread:
Sorry, I do not understand what you mean.

I did not touch the property file.
I know about the ```serverTimezone=UTC```, it solved the date problem at the time it was added (with hibernate UTC settings as well). Just recently found out that it is not enough.

DanielFran:
I mean the property hibernate.jdbc.time_zone: UTC added in https://github.com/jhipster/generator-jhipster/pull/8501 that it is not necessary now if we configure directly the server

Blackdread:
yeah could I guess but I would not be so sure.
Would need test to see when DB is on another server, the connection might need to be set as UTC.

generators/server/templates/src/main/java/package/Application.java.ejs
```diff
@@ -102,6 +104,8 @@ public class <%= mainClass %> {
      * @param args the command line arguments.
      */
     public static void main(String[] args) {
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
gmarziou, line 107: Maybe we should log the default timezone before changing it?

        ZoneId defaultZoneId = ZoneId.systemDefault();
        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
        log.info("Timezone changed from {} to UTC", defaultZoneId);

Blackdread: Added

avdev4j:
If we finally do this, should we move the timezone configuration part to a dedicated configuration class? It could be easier to understand?

Also if we remove the `hibernate.jdbc.time_zone: UTC` and having the db in a different machine, it will not have consequence because it's related to Hibernate, not the DB. So we could think that's it's a good idea to remove the property but it's also mean that we trust totally the JVM to set (along the project life) the right timezone.

For example, I have the JVM set to UTC -> Hibernate will store the value on the same timezone.
I change the JVM timezone -> Hibernate will take the JVM timezone and no longer UTC, I can't know on wich timezone the  date as been stored... I have different values on different timezones :cry:

Conclusion: whatever I do with the default JVM timezone, I should keep the hibernate timezone property to UTC :)

mraible:
Is this still an issue? If so, can you please fix conflicts.

Blackdread:
original issue was closed so not sure should merge or not. I am still in favor of setting VM to UTC. Many developers do not know how to handle timezone (for DB, etc), so this is a small code that is easy to remove after generated if the person knows what he is doing with the timezone.

avdev4j:
I think @mraible forget to close the PR when he closed the issue.

My main concern here is we will change the timezone for the whole JVM. Logging (for example) will be affected by that. I'm not saying this is a bad thing, but it's a side effect we need to be sure we understand the consequences.
I think we need to do some checks before this kind of modification.

By the way, this PR is good and I would merge it if we are sure this modification makes sense.

Blackdread:
@avdev4j then a vote within jhipster team? :P

generators/server/templates/src/main/java/package/Application.java.ejs
```diff
@@ -117,7 +123,8 @@ public class <%= mainClass %> {
                 "Application '{}' is running! Access URLs:\n\t" +
                 "Local: \t\t{}://localhost:{}{}\n\t" +
                 "External: \t{}://{}:{}{}\n\t" +
-                "Profile(s): \t{}\n----------------------------------------------------------",
+                "Profile(s): \t{}\n\t" +
+                "VM timezone: \t{}\n----------------------------------------------------------",
```
mraible, line 127: It might be better to just say timezone since you won't always be in a VM when running a JHipster app.

Blackdread: fixed

Blackdread:
fixed

Blackdread:
we merge?

avdev4j:
I don't think we made a decision about changing the JVM timezone. This is a discussion open for a long time now.

I think we should be sure what we are doing here.

PR #11795, opened 2020-05-15
Save dateFormatForLiquibase.

mshima:
Reproducible builds can set changelogDate to future date, so save it to keep changelogs linear.

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -891,21 +891,46 @@ module.exports = class extends PrivateBase {
 
     /**
      * Generate a date to be used by Liquibase changelogs.
+     *
+     * @param {Boolean} [reproducible=true] - Set true if the changelog date can be reproducible.
+     *                                 Set false to create a changelog date incrementing the last one.
+     * @return {String} Changelog date.
      */
-    dateFormatForLiquibase() {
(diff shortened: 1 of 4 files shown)


mshima:
@MathieuAA can you check if I missed any test at this PR?

generators/generator-base.js
```diff
@@ -882,21 +882,46 @@ module.exports = class extends PrivateBase {
 
     /**
      * Generate a date to be used by Liquibase changelogs.
+     *
+     * @param {Boolean} reproducible - Set true if the changelog date can be reproducible.
```
mshima, line 886:
```suggestion
     * @param {Boolean} [reproducible=true] - Set true if the changelog date can be reproducible.
```

generators/generator-base.js
```diff
@@ -882,21 +882,46 @@ module.exports = class extends PrivateBase {
 
     /**
      * Generate a date to be used by Liquibase changelogs.
+     *
+     * @param {Boolean} [reproducible=true] - Set true if the changelog date can be reproducible.
+     *                                 Otherwise a changelog date will be created incrementing the last one.
```
mshima, line 887:
```suggestion
     *                                 Set false to create a changelog date incrementing the last one.
```

PR #11813, opened 2020-05-18
fix: Angular locale module name

### LINKED ISSUE #11745: bug ; local angular genrator
when using the angular generator and make select ar-ly as a language the angular client bug.
the code generated on core.module.ts contain the import of the local wich refer to an non existing file

angular client generated code should import modules with case sensitive names.
##### **Reproduce the error**
try to generate an application with an angular client and choosing Arabic (Libya) as language then use docker containers or linux to see the results

Linux is case sensitive and the generated import contains ar-ly or under node_modules there is ar-LY.js (i think the iso format is xx-AA)

##### **Suggest a Fix**
i suggest to fix value key in generators-constants file and verify if there are other impacts

##### **JHipster Version(s)**
v 6.8 
i think all versions

Shaolans:

Some of the languages differs between the values and the localeId used in Angular resulting this kind of issue #11745
Close #11745

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -173,6 +173,7 @@ const LANGUAGES = [
         dispName: 'Shqip',
         value: 'al',
         momentLocaleId: 'sq',
+        localeId: 'sq',
     },
     {
         name: 'Arabic (Libya)',
@@ -180,6 +181,7 @@ const LANGUAGES = [
         value: 'ar-ly',
         rtl: true,
         skipForLocale: true,
+        localeId: 'ar-LY',
     },
     {
         name: 'Armenian',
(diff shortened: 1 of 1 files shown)


MathieuAA:
This isn't happening in react?

Shaolans:
As far as a I know, this is not happening in React.

kaidohallik:
`localeId` is used only in Angular part.  
FYI: I tried to understand why is `ar-ly` used partially (https://github.com/jhipster/generator-jhipster/blob/master/generators/generator-constants.js#L182 - the only place where `skipForLocale: true,`) and found out that because of #6101  
If in https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/core/core.module.ts.ejs logic for `ar_ly`:

* in `LOCALE_ID` use `en`
* in `registerLocaleData` use `ar_LY`

still holds then this LGTM.

MathieuAA:
Alright then. It already works for real so let's merge it.

MathieuAA:
Thanks @Shaolans for the fix, and thanks @kaidohallik for the explanation!

Issue #11805, opened 2020-05-16
Fix frequently failing react builds (e2e)


MathieuAA:
I think we're all getting tired to see failing react builds.
This is the issue to fix'em.

Even though I'm assigning myself to this issue, anyone is free to contribute and fix them! The goal here is to improve our QoL as contributors.

---

Found issues:
  - StaleElementReferenceError: stale element reference: element is not attached to the page document:  https://github.com/jhipster/generator-jhipster/runs/682093799?check_suite_focus=true
  - `should delete last item` expected 0 got 1 or (10 got 11)
  - Account tests: https://github.com/hipster-labs/jhipster-daily-builds/runs/659933931?check_suite_focus=true#step:18:364

pascalgrimaud:
adding a bounty for this
React builds are flaky since at least 6 months...

MathieuAA:
I'm gonna try the recent changes made by @skmaingi against my private fork with duplicated builds to see if it works

pascalgrimaud:
I'm reopening this, because I need to test if it totally fixed, with the related PRs.

skmaingi:
@MathieuAA @pascalgrimaud Confirm if pull request #11916 has fixed this.

mshima:
If this issue is fixed I think the bounty should be increased an shared between @MathieuAA @skmaingi and @sendilkumarn

MathieuAA:
I wish to give my part to the issue for cypress.

sendilkumarn:
👍 of course for Cypress :)

jdubois:
I'm going to increase the bounty to $300 so it can be shared evenly.
-> Mathieu, I will have a look at how to give $100 to Cypress

And @MathieuAA  what is blocking from closing this issue?

MathieuAA:
Nothing :) I guess I'm so used to have the classic "fixes #XXX" that I've completely forgot about this issue!

jdubois:
Ok then don’t forget to claim your bounties!
@MathieuAA for the $100 you want to give to Cypress I know there’s a way to transfer money directly between 2 collectives and not pay any fee, I need to have a look

pascalgrimaud:
@jdubois : I think @MathieuAA means the issue in JHipster related to support Cypress and replace Protractor :)

jdubois:
Oh then that would be much easier! @MathieuAA can you clarify?

MathieuAA:
[...]

PR #11814, opened 2020-05-18
Fix failing react tests

MathieuAA:

The goals of this PR:
  - fix the tests
  - refactor them after

---

For `Entity.spec.ts.ejs`:
Noticed issue: `beforeRecordsCount` was a global variable used in different async
functions, so not atomic. The tests weren't actually independent because of that.
Another: the `ComponentsPage` was initialized in one test and used in the others.

---

Related to #11805


vishal423:
Last time when I looked into the react e2e tests, there were visible issues with code (multiple calls, broken functionality) and that was causing e2e intermittent failures. You may also want to check on that aspect.

MathieuAA:
@vishal423 thanks! If you want to try fixing them with me that'd be great! If you don't have time though, just pointing out where the issues would help me tremendously!

vishal423:
@MathieuAA , I have taken a short break from any work/code on e2e test cases after burnout of fixing those on my schematic. For your reference, looking at recent failures, I could see the following failure reasons in deleteXXX scenario:
- StaleElementReferenceError: stale element reference: element is not attached to the page document
> From the log, it's not obvious the statement that caused it. It requires more debugging. We should fix configuration on our side to log complete stack traces
- AssertionError: expected 1 to equal 0
> Ideally it shouldn't happen. The most probable cause is the delete API failure. I suspect it could be an issue in the generated code and not with test cases. Your point on the atomic test case shouldn't affect here as we run tests in the sequential order.

MathieuAA:
>  AssertionError: expected 1 to equal 0

this one is hopefully fixed after the commits but I haven't noticed the first one, more investigation is indeed needed

I'm currently looking at the Account tests

vishal423:
Link to first failure  https://github.com/jhipster/generator-jhipster/runs/682093799?check_suite_focus=true

MathieuAA:
Ah thanks @vishal423! I'll investigate it

MathieuAA:
> Ideally it shouldn't happen. The most probable cause is the delete API failure. I suspect it could be an issue in the generated code and not with test cases. Your point on the atomic test case shouldn't affect here as we run tests in the sequential order.

That's weird, tests pass for angular cases

vishal423:
I suspect the assertion issue is due to old data being displayed on the UI. A quick look into code reveals that we don't clear old data while making a fetch call and also don't consider the in progress fetch call in render.

MathieuAA:
Interesting. That may be a reason why angular tests don't fail, because protractror is hooked on angular and its internal state. 
I'll do more tests tomorrow.
Great point @vishal423!!

MathieuAA:
@vishal423 Just thought about something and I wanted your opinion on this. If the issue is that we never wait for the state to change, i.e. for requests to complete so that the internal state changes and the view to change in consequence, what about:
  - intercepting & waiting for the requests' completion
  - waiting a bit, just for the state to change
  - checking if the view has changed

It's not the prettiest solution, but I don't see any other simpler way to do it.

vishal423:
@MathieuAA, we do try to wait for requests to complete before doing assertions, however, we do that at a high level. For example, to ensure that the GET request to fetch and render the list page contents is complete, we do wait for the outcome of that; either a message showing no records found or a table to display the content. Do you see shortcomings in the existing test suite code that don't follow this? 

As I pointed before in my previous comment, fixing the react code can be one solution to ensure the correct functioning of protractor test suite.

MathieuAA:
@vishal423 That's the thing: the fact that there's a tight coupling between what we do and what we expect make the tests (i.e. we can see the "one-to-one" relationship between the code and the test), well, not robust as they should be. At least, they're not as robust as the angular tests.
That's why I asked about a "lower level" of check, to reduce the coupling. Ultimately, I don't think it will solve anything... The issue might reside in how we design our tests.

Open the e2e tests for the account and you'll see that each test depend on the previous one. What happens when we want to write an additional test? Will it fail because the one after expects us to do something? I can't answer that myself TBH.

Now, if there's an issue with the code dealing with account management, I don't think I can rely on the tests to tell me where the issue is. The same goes for the e2e tests for entities.

> As I pointed before in my previous comment, fixing the react code can be one solution to ensure the correct functioning of protractor test suite.

AFAIK the react code works. If it's hard to test then changing it will help us but I don't think the issue resides here.

vishal423:
I don't fully agree with the high code coupling point of view, instead, I consider those coupled with functional expectations. I think you are mixing unit tests with the functional tests. In real application usage, there is a dependency on user actions typically aligned to the business flow and you want those flows to be tested fully in sequence. Failure in a step should typically have a cascading effect.

You can't rely on e2e tests to point out the implementation issues, but, they can give you a glimpse of something wrong somewhere with intermittent failures. As an example, you can look at the last few fixes on the react code that were causing intermittent failures, however, due to inadequate unit tests, they were not caught.

On broad terms, I think we need to look beyond protractor and in the long run put more effort towards improving overall generated source code and test code quality.

MathieuAA:
I want us to look into this little failure:
```
   Account
    ✓ should fail to login with bad password
    ✓ should login with admin account
    ✓ should be able to sign up
    ✓ should load user management
    1) should activate the new registered user
    2) should not be able to sign up if login already taken
    3) should not be able to sign up if email already taken
    4) should be able to log in with new registered account
    5) should login with admin account
    ✓ should fail to update password when using incorrect current password
    ✓ should be able to update password
    ✓ should be able to log in with new password
    6) should login with user_test account
    7) should be able to change user_test settings
    ✓ should login with admin account
    8) should not be able to change admin settings if email already exists
    ✓ should delete previously created fake user
```

I expect my tests to fail in cascade too, especially where there's a flow involved, like here. What I don't expect, however, is that some parts of my flow fail intermittently. I think I can safely assume that none of like Christmas tree decorations in tests.

I expect my tests to yell at me when I do something stupid. Intermittent yelling isn't proper yelling, it's just noise, but that's just beside the point. The whole point I'm trying to address is the **why**. Why are there intermittent failures?

The code doesn't seem to have intermittent behavior though.

> On broad terms, I think we need to look beyond protractor and in the long run put more effort towards improving overall generated source code and test code quality.

Yup.

pascalgrimaud:
I have an easy solution: disable Protractor tests for React.
As it's flaky since months and after so much effort from everybody, it's not fixed, then let's just remove these tests from the CI.

vishal423:
Can you point me to the build that contains the above failures? I will try to spend some time tomorrow on it.

MathieuAA:
@vishal423 there's [this one](https://github.com/jhipster/generator-jhipster/runs/681476590)

MathieuAA:
@pascalgrimaud Okay to remove them if:
  - they are replaced
  - we add more tests to the react code with, for instance, [testing library](https://testing-library.com/docs/react-testing-library/intro)

vishal423:
Account failure seems related to the incomplete rendering of the table contents. If you manually access this page, then, you may notice that initially only the table header without any content is rendered and later refreshed to include the content of the API response. This rendering is intermittent and you may not see it often (clean reload of this page may highlight this)

To improve reliability, we can include a check to wait for the table contents to be visible on the page before the failing statement.

MathieuAA:
Yes, do you think the last commit helps in that regard?

vishal423:
Instead of waiting on the sort icon header field, I suggest waiting for table rendering to complete. In real usage, users would definitely wait for content to be visible before taking another action. It would also ensure that we are not dealing with an intermediate state that could lead to stale element exceptions.

MathieuAA:
Good point but how can we wait for that? I've seen that protractor doesn't provide us with too many options regarding this particular matter. I've seen plugins though. That can be considered.

MathieuAA:
I've also forked the generator to have an all-react build system [here](https://github.com/MathieuAA/private-generator-jhipster). 
The builds include the ones from this repo and others from the "daily builds" repo.
Pascal has suggested me to duplicate builds to, I'm pushing them. This should give some insight, relatively speaking though. If we're looking at async errors, this is gonna take time...

MathieuAA:
Two failures (out of 59:
  - https://github.com/MathieuAA/private-generator-jhipster/runs/702120747#step:17:343
  - https://github.com/MathieuAA/private-generator-jhipster/runs/702120722#step:17:224

sendilkumarn:
Oops sorry for this. I will take a look into this.

sendilkumarn:
@MathieuAA Can we have a quick sync over the weekend?

MathieuAA:
Of course @sendilkumarn. Would a video call be okay for you?

sendilkumarn:
Yeah sure. I will forward you a hangout or zoom invite later 👍

MathieuAA:
A better way to achieve what the last commit says is to rework the notification aspect of the generated apps so that notifications only appear **after** an action has been performed. This isn't the case right now.

MathieuAA:
The error seem related to memcached (both here and in my duplicated builds)

MathieuAA:
@jhipster/developers I propose you the following:
  - create an issue to make e2e tests with memcached-based builds work again
  - disable memcached-based builds
  - review this PR
  - hopefully merge it if it's good enough

pascalgrimaud:
@MathieuAA : for your information, I reproduce the failures easily with memcached, in my local environment, at my 1st try

generators/entity-client/templates/react/src/test/javascript/e2e/entities/entity.spec.ts.ejs
```diff
@@ -194,11 +194,7 @@ describe('<%= entityClass %> e2e test', () => {
 
         expect(await isVisible(<%= entityInstance %>DeleteDialog.deleteModal)).to.be.false;
 
-        await waitUntilAnyDisplayed([<%= entityInstance %>ComponentsPage.noRecords,
-        <%= entityInstance %>ComponentsPage.table]);
-    
-        const afterCount = await isVisible(<%= entityInstance %>ComponentsPage.noRecords) ? 0 : await getRecordsCount(<%= entityInstance %>ComponentsPage.table);
-        expect(afterCount).to.eq(beforeRecordsCount);
+        await waitUntilDisplayed(<%= entityInstance %>ComponentsPage.noRecords);
```
pascalgrimaud, line 197: The React build for memcached failed because of this line.

Why only for memcached ? It's because:
- it's the only build with **dev** profile
- see https://github.com/jhipster/generator-jhipster/blob/master/.github/workflows/react.yml#L78
- see https://github.com/jhipster/generator-jhipster/blob/master/.github/workflows/react.yml#L102
- with dev profile, there are data, with fake-data

When I replaced with old code, it works well:
```
await waitUntilAnyDisplayed([<%= entityInstance %>ComponentsPage.noRecords, <%= entityInstance %>ComponentsPage.table]);
```

MathieuAA: Well spotted. This won't fix the issue, at least this shows there's room for improvement regarding the error messages, among other things.

sendilkumarn: That is awesome @pascalgrimaud 

Let us move the `memcached` to `prod` build. Also we still are not fixing the root cause. The test cases should ideally be of the following format:

```
// setup data before each test case.
// do the action.
// validate the results.
```

Also, after discussing with @MathieuAA I strongly feel that we should move away from protractor to Cypress. This will solve a lot of problems. WDYT?

pascalgrimaud: About Cypress, it's another subject. FYI, Vue.js blueprint already started the work -> https://github.com/jhipster/jhipster-vuejs/pull/399

As Vue.js is not in main generator-jhipster, we could easily choose to remove Protractor or keep it, as we don't need to be totally consistent with main generator-jhipster. That's the power of blueprint.

MathieuAA: > > Let us move the memcached to prod build
> 
> It will be probably fixed the build. But don't forget a lot of us (in the core team) do demo, and start application with dev profile, and run e2e tests in dev profile. That's the reason I think e2e must work in dev profile too.
> 

Pascal, of course they must work no matter the env. The principal thing here is to be able to insert/remove data at will. That helps with the design issues we have with our tests. A solution could to have a third env, where data can be inserted/removed/whatever without messing with anything.

sendilkumarn: > That's the reason I think e2e must work in dev profile too.

I do agree, it should work irrespective of the env. But this should be a temporary work around until we fix the test-case designs

> About Cypress, it's another subject. FYI, Vue.js blueprint already started the work

That is the beauty of Cypress, it will not tie you to a single framework. So, if we have built for vueJS then we can use it here.

sendilkumarn:
Can we also merge this with #11814 ?

sendilkumarn:
Closing this, this is merged into the master via this commit e77ebe851394af18e289aaf4b1481c3a42f8f540

PR #11857, opened 2020-05-27
Correct typos (breaking changes for v7)

vivekmore:
Continuation of #10788 (for next major release)

### DIFF
--- a/generators/client/index.js
+++ b/generators/client/index.js
@@ -91,7 +91,7 @@ module.exports = class extends BaseBlueprintGenerator {
                 }
             },
 
-            setupClientconsts() {
+            setupClientConstants() {
                 // Make constants available in templates
                 this.LOGIN_REGEX = constants.LOGIN_REGEX_JS;
                 this.ANGULAR = ANGULAR;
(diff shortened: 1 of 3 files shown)


avdev4j:
hi @vivekmore 
I wonder if we should go further and rename the generator base function `aski18n` while we introduce breaking changes.

WDYT

vivekmore:
> I wonder if we should go further and rename the generator base function `aski18n` while we introduce breaking changes.


I like it - I don't remember if there was a reason for leaving it out of my PR.
I've updated my PR with your suggested change, hopefully the PR build doesn't break

avdev4j:
I think we will have break issues with modules and blueprints that will still use the old methods.

vivekmore:
> I think we will have break issues with modules and blueprints that will still use the old methods.

yes I thought that was clear from the title (these are breaking changes)

avdev4j:
hi @vivekmore 
I just wanted to confirm you that there is a risk to have breaking issues for some blueprints. unfortunately I don't know which blueprint can be affected, maybe none.

PR #11879, opened 2020-06-01
Generate the code of search option only when searchEngine option is present

maznag:
Generate the code of search option only when searchEngine option is present.

### DIFF
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -162,7 +162,7 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 context.prodDatabaseType = configuration.get('prodDatabaseType') || this.options.db;
                 context.devDatabaseType = configuration.get('devDatabaseType') || this.options.db;
                 context.skipFakeData = configuration.get('skipFakeData');
-                context.searchEngine = configuration.get('searchEngine');
+                context.searchEngine = _.isUndefined(configuration.get('searchEngine')) ? false : configuration.get('searchEngine');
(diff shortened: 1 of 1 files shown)


murdos:
Could your provide more information about the issue you're fixing?

maznag:
@murdos  The related code of search option is generated even if the searchEngine option isn't present. due to this test condition: `searchEngine !== false` whitch present in several files (for example in entity-management.component.ts.ejs). In fact, when searchEngine option isn't present this test condition always evaluate to `true` because searchEngine hase '' as value. 

To reproduce the issue:
* Generate a new app with the default options without specifying searchEngine option
* Generate a sample entity
* Check the generated code and you will see that the related code of search option will be generated.

generators/entity/index.js
```diff
@@ -162,7 +162,7 @@ class EntityGenerator extends BaseBlueprintGenerator {
                 context.prodDatabaseType = configuration.get('prodDatabaseType') || this.options.db;
                 context.devDatabaseType = configuration.get('devDatabaseType') || this.options.db;
                 context.skipFakeData = configuration.get('skipFakeData');
-                context.searchEngine = configuration.get('searchEngine');
+                context.searchEngine = _.isUndefined(configuration.get('searchEngine')) ? false : configuration.get('searchEngine');
```
mshima, line 165: I prefer this notation:
```suggestion
                context.searchEngine = _configuration.get('searchEngine')) || false;
```

maznag: done !

DanielFran:
Thanks @maznag

Issue #11824, opened 2020-05-19
Angular Client fails to compile With more than one entity


JasonTypesCodes:
##### **Overview of the issue**

Generating an Angular client on 6.9.0 with entities results in:
```
ERROR in src/main/webapp/app/entities/entity.module.ts:6:27 - error TS2345: Argument of type '({ path: string; loadChildren: () => Promise<any>; } | undefined)[]' is not assignable to parameter of type 'Routes'.
  Type '{ path: string; loadChildren: () => Promise<any>; } | undefined' is not assignable to type 'Route'.
    Type 'undefined' is not assignable to type 'Route'.

  6     RouterModule.forChild([
                              ~
  7       {
    ~~~~~~~
... 
 16       /* jhipster-needle-add-entity-route - JHipster will add entity modules routes here */
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 17     ]),
    ~~~~~
```

##### **Reproduce the error**

1. Generate a new Monolith app
2. Add an entity
3. Add another entity

##### **Related issues**

None that I could locate

##### **Suggest a Fix**

There is a superflous comma in the generated entity.module.ts.  It should be removed.
```
{
        path: 'something',
        loadChildren: () => import('./something/something.module').then(m => m.JhipsterSomethingModule),
      },
      , <------ This one
      {
        path: 'something-else',
        loadChildren: () => import('./something-else/something-else.module').then(m => m.JhipsterSomethingElseModule),
      },
```

##### **JHipster Version(s)**

6.9.0

##### **JHipster configuration**

```
INFO! Using JHipster version installed locally in current project's node_modules
INFO! Executing jhipster:info
INFO! Options: from-cli: true
Welcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
jhipster@0.0.1-SNAPSHOT /home/schindlerj/src/temp/jhipster/upstream/angular-multi-entity
└── generator-jhipster@6.9.0 

```


##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**


<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.mycompany.myapp",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "6.9.0",
[...]

PR #11890, opened 2020-06-03
Fix addIcon needle function generating double comma

kaidohallik:
#11665 switched `Prettier` to `v2` which by default adds trailing commas to many places.  
`addIcon` needle function assumed that there is no trailing comma.  
After this change `addIcon` works both with and without trailing comma in icons array.

### DIFF
--- a/generators/client/needle-api/needle-client-angular.js
+++ b/generators/client/needle-api/needle-client-angular.js
@@ -125,16 +125,11 @@ module.exports = class extends needleClientBase {
         const iconImport = `fa${this.generator.upperFirstCamelCase(iconName)}`;
         if (!jhipsterUtils.checkRegexInFile(iconsPath, new RegExp(`\\b${iconImport}\\b`), this.generator)) {
             try {
-                let newIconContent = `\n  ${iconImport},\n  // jhipster-needle-add-icon-import`;
-                const iconNeedle = '\\r?\\n\\s*\\/\\/ jhipster-needle-add-icon-import';
-                if (!jhipsterUtils.checkRegexInFile(iconsPath, new RegExp(`,${iconNeedle}`), this.generator)) {
-                    newIconContent = `,${newIconContent}`;
-                }
                 jhipsterUtils.replaceContent(
                     {
                         file: iconsPath,
-                        pattern: new RegExp(iconNeedle, 'g'),
-                        content: newIconContent,
+                        pattern: /(\r?\n)(\s*)\/\/ jhipster-needle-add-icon-import/g,
+                        content: `\n  ${iconImport},\n  // jhipster-needle-add-icon-import`,
                     },
                     this.generator
                 );
--- a/generators/client/needle-api/needle-client-angular.js
+++ b/generators/client/needle-api/needle-client-angular.js
@@ -125,11 +125,16 @@ module.exports = class extends needleClientBase {
         const iconImport = `fa${this.generator.upperFirstCamelCase(iconName)}`;
         if (!jhipsterUtils.checkRegexInFile(iconsPath, new RegExp(`\\b${iconImport}\\b`), this.generator)) {
             try {
+                let newIconContent = `\n  ${iconImport},\n  // jhipster-needle-add-icon-import`;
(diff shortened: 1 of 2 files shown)


mshima:
Probably we should create a bug and verify every client needle.
Similar issue related to prettier: https://github.com/jhipster/generator-jhipster/issues/11824

generators/client/needle-api/needle-client-angular.js
```diff
@@ -125,11 +125,16 @@ module.exports = class extends needleClientBase {
         const iconImport = `fa${this.generator.upperFirstCamelCase(iconName)}`;
         if (!jhipsterUtils.checkRegexInFile(iconsPath, new RegExp(`\\b${iconImport}\\b`), this.generator)) {
             try {
+                let newIconContent = `\n  ${iconImport},\n  // jhipster-needle-add-icon-import`;
+                const iconNeedle = '\\r?\\n\\s*\\/\\/ jhipster-needle-add-icon-import';
+                if (!jhipsterUtils.checkRegexInFile(iconsPath, new RegExp(`,${iconNeedle}`), this.generator)) {
```
mshima, line 130: I don’t think we should add this test.
We should add the comma to the template instead.

The only use case I can think the comma in the template doesn’t cover is a module with updated jhipster been applied to a old jhipster source.
Or I am missing something?

kaidohallik: Two possible use cases where trailing comma is missing:
1. update JHipster version, don't regenerate, run module which uses `addIcon`
2. user changes `.prettierrc` and sets `trailingComma: none`

First one is unlikely, second can happen, question is: should we support that or not.

If we support then code in `generator-jhipster` is more complex, if we don't support then some users can get compilation errors but these errors are very easily fixable.

So, should I remove this check to make generator simpler but risk with some minor errors in edge cases?

mshima: If we support, we should support in a more generic way.
I suggest to add both and remove in the transform step before prettier.

mshima: @kaidohallik what about `// prettier-ignore`?

kaidohallik: I think we should not put `// prettier-ignore` in the icons file. End user should get nicely formatted code.

I will remove edge cases handling from this particular place. So if someone wants to implement general trailing comma edge cases protection then this place doesn't contain redundant code.

I retain first commit so if someone wants then can investigate one possible solution.

mshima:
@kaidohallik we need to add the trailing coma to the template.
https://github.com/jhipster/generator-jhipster/blob/4589cb6aa6c298c4155ffebf5671a26167694c35/generators/client/templates/angular/src/main/webapp/app/core/icons/font-awesome-icons.ts.ejs#L35
https://github.com/jhipster/generator-jhipster/blob/4589cb6aa6c298c4155ffebf5671a26167694c35/generators/client/templates/angular/src/main/webapp/app/core/icons/font-awesome-icons.ts.ejs#L73

kaidohallik:
Trailing commas added to template.

generators/client/templates/angular/src/main/webapp/app/core/icons/font-awesome-icons.ts.ejs
```diff
@@ -70,6 +70,6 @@ export const fontAwesomeIcons = [
   faCalendarAlt,
   faSearch,
   faTrashAlt,
-  faAsterisk
+  faAsterisk,
   // jhipster-needle-add-icon-import
```
mshima, line 74:
```suggestion
  // jhipster-needle-add-icon-import
  faAsterisk,
```

generators/client/templates/angular/src/main/webapp/app/core/icons/font-awesome-icons.ts.ejs
```diff
@@ -32,7 +32,7 @@ import {
   faTrashAlt,
   faAsterisk,
   faTasks,
-  faHome
+  faHome,
   // jhipster-needle-add-icon-import
```
mshima, line 36:
```suggestion
  // jhipster-needle-add-icon-import
  faHome,
```

PR #11938, opened 2020-06-15
Improve MSSQL performance

pascalgrimaud:
Following advices from http://www.jochenhebbrecht.be/site/2014-05-01/java/fixing-slow-queries-running-sql-server-using-jpa-hibernate-and-jtds

It improves a lot of performance of my MSSQL application.
What do you think @ruddell ?

_____

### DIFF
--- a/generators/generator-base-private.js
+++ b/generators/generator-base-private.js
@@ -1508,7 +1508,7 @@ module.exports = class extends Generator {
         } else if (databaseType === 'oracle') {
             jdbcUrl = `jdbc:oracle:thin:@${options.hostname}:1521:${options.databaseName}`;
         } else if (databaseType === 'mssql') {
-            jdbcUrl = `jdbc:sqlserver://${options.hostname}:1433;database=${options.databaseName}`;
+            jdbcUrl = `jdbc:sqlserver://${options.hostname}:1433;database=${options.databaseName};sendStringParametersAsUnicode=false`;
         } else if (databaseType === 'h2Disk') {
             if (!options.localDirectory) {
                 throw new Error(`'localDirectory' option should be provided for ${databaseType} databaseType`);
--- a/test/generator-base-private.spec.js
+++ b/test/generator-base-private.spec.js
@@ -223,9 +223,9 @@ export * from './entityFolderName/entityFileName.state';`;
             });
         });
         describe('when called for mssql', () => {
-            it('return jdbc:sqlserver://localhost:1433;database=test', () => {
+            it('return jdbc:sqlserver://localhost:1433;database=test;sendStringParametersAsUnicode=false', () => {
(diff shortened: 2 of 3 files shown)


murdos:
Can this change causes some encoding issues with the application?

pascalgrimaud:
@murdos : from my current use for my customer, the encoding is the same with or without this property. The only difference is performance, with some requests:
- without this property: 4sec~
- with this property: 100ms

That's why I'm proposing this, but I'm not a MSSQL expert at all, that's why I was asking to @ruddell  who proposed MSSQL support in the past :)

avdev4j:
@pascalgrimaud @murdos 
According to the documentation it depends if we use non-national character methods. In this case we could lost some datas.
But, the official documentation suggest, also, the set this params to false to improve performances when using such JDBC data types as VAR, VARCHAR or LONGVARCHAR.

Maybe we should merge it and add some documentation about this feature on the official website. (in tips and tricks section?).

PS: the official documentation -> [https://docs.microsoft.com/fr-fr/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver15](https://docs.microsoft.com/fr-fr/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver15)

pascalgrimaud:
I spent 1 full day to try to understand how our generated application with JHipster+MSSQL has so poor performance.

Now I got the solution, I don't care if it's merged or not, as I won't use MSSQL a lot.

Maybe it's better to not merge, as it is the default configuration.
I'll let you the final decision :)

avdev4j:
@pascalgrimaud don't worry I think we should merge it, especially if the offical doc suggest that. The only thing I struggle with is the notion of "non-national character methods" if someone could explain us, maybe we would be able to easily validate this change or not :D.

cc @jhipster/developers

ruddell:
I tested emojis and international characters (google translated 'foo' to Chinese `富`). With our default config, neither of those work since we use `varchar` instead of `nvarchar` for MSSQL. 

When changing liquibase config to use `nvarchar`, both emojis and the international character saved and returned successfully. With the changes from this PR, the database saves and returns question marks (`???`) instead of the emoji or chinese character, like it would when using `varchar`.

pascalgrimaud:
Thanks for feedback @ruddell 
For our project, we have some fields using nvarchar, but we added `org.hibernate.annotations.Nationalized` annotation

So, without my PR (change in jdbc url), does nvarchar work well ?

ruddell:
`nvarchar` works fine without this PR, and is broken with this PR

`nvarchar` works fine with this PR with `@Nationalized` annotation

Tested with `♥️😘😂富` as the data entered in a String field.

pascalgrimaud:
ok then, I think it's better to close this.
At least, for people who have performance issue, they'll know how to fix it :)

avdev4j:
Anyway @pascalgrimaud it deserve a thread in the tips website section 😉

pascalgrimaud:
yes, will try to find a place to our documentation

PR #11943, opened 2020-06-16
Made the Navbar fixed to Top.  

bhushanon:

-   [x]  All continuous integration tests (https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Not applicable, Tests are added where necessary

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
+++ b/generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
@@ -35,6 +35,10 @@ h4 {
     font-weight: 300;
 }
 
+.container-fluid {
+  padding-top: 45px;
+}
+
 <%_ if (clientTheme === 'none') { _%>
 /* Increase contrast of links to get 100% on Lighthouse Accessability Audit. Override this color if you want to change the link color, or use a Bootswatch theme */
 a {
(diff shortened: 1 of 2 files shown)


avdev4j:
hi @bhushanon 
I'm started with a little tip, you could put a `x` into the braces to check the box, instead of write "YES". You also could check the box after the message has been posted, `- []` is convert to a checkbox html element. 

Second tip, I consider that merging from your master branch is not a good practice. You should create a different branch, even on your own repo. Why? Because it will be easier for you to rebase from the upstream master if needed and If you have. It will not be possible to work on different contributions if you work directly on master. This is an advice, let the master clean and open PR from your branches. (Of course it change nothing about your contribution, for this one it's ok ;) ).

What is your motivation for changing this? Could you please explain us more?

Whatever, we have failing  e2e tests. Could you, please, fix them?

thanks you for your contribution!

regards,

generators/client/templates/angular/src/main/webapp/content/scss/global.scss.ejs
```diff
@@ -35,6 +35,10 @@ h4 {
     font-weight: 300;
 }
 
+.container-fluid {
+  padding-top: 45px;
```
avdev4j, line 39: I'm not a CSS specialist, but is not a good idea to use px to set a padding value?

bhushanon:
I will adhere to all suggestions & directions that have been given, and create new PR and close this one.

Issue #10345, opened 2019-09-03
Add Support for Microservices in GAE


SudharakaP:
##### **Overview of the feature request**

I was thinking we could add support for Microservices in our GAE Environment(s). It seems that [GAE supports microservices through services](https://cloud.google.com/appengine/docs/standard/java11/an-overview-of-app-engine).

If this is something useful; I could work on this one. Let me know your thoughts. 😄  

##### **Motivation for or Use Case**

Add Microservice support for GAE will make our GAE generator more complete with support for all types of applications we generate. 

##### **Related issues or PR**

#10284 

-   [x] Checking this box is mandatory (this is just to show you read everything)

saturnism:
that would be great! i wanted to have that initially but there were some difficulties. I think most of the problems are now resolved through work done on Kubernetes.

Some thoughts on how to approach this:
* Use a [dispatch file](https://cloud.google.com/appengine/docs/standard/java/reference/dispatch-yaml) to map sub-urls to the backend services.
  * Do this in the Gateway application
  * Disable Zuul and everything similar to Kubernetes Ingress / Istio
  * Similar to Kubernetes Ingress, we can map by prefixes, e.g., /services/inventory/* -> inventory service

SudharakaP:
@saturnism : Wonderful; thanks for the suggestions; Although I have experience deploying monoliths on GAE, I haven't done any microservice deployments. So it will take me some time to figure this one out. :smile:

pascalgrimaud:
Adding a bounty for this

SudharakaP:
Just to give a heads up; I'll be working on this after finalizing the [AWS bug fixes](https://github.com/jhipster/generator-jhipster/issues/8366) which I think is more important (correct me if I am wrong). :smile:

SudharakaP:
Do you guys like to add this to the existing generator, create a new generator (i.e: gae-microservices) or add this as a module? Let me know what you think. Just asking this since in some threads I've seen the reluctance to add new generators to due to maintainability. 😄

SudharakaP:
Okay, so I am gonna try the separate module approach. 😄

saturnism:
i think we should use the same generator :D

SudharakaP:
@saturnism : Thanks for the suggestion; yes we could do that. But curious why do you think this should be included in the same generator? I was thinking this might make it a bit more complicated with monolith and microservice logic in the same generator?

saturnism:
[...]

PR #11969, opened 2020-06-21
Add Microservice Support for GAE sub-generator

SudharakaP:
This adds micro-service support for GAE sub-generator

Resolves https://github.com/jhipster/generator-jhipster/issues/10345

### DIFF
--- a/generators/gae/index.js
+++ b/generators/gae/index.js
@@ -18,10 +18,12 @@
  */
 const os = require('os');
 const shelljs = require('shelljs');
+const fs = require('fs');
 const chalk = require('chalk');
 const _ = require('lodash');
 const BaseGenerator = require('../generator-base');
 const statistics = require('../statistics');
+const dockerPrompts = require('../docker-prompts');
 
 const constants = require('../generator-constants');
 
@@ -116,33 +118,64 @@ module.exports = class extends BaseGenerator {
         };
     }
 
-    defaultProjectId() {
-        if (this.abort) return null;
-        if (this.gcpProjectId) {
-            return this.gcpProjectId;
-        }
-        try {
-            const projectId = shelljs.exec('gcloud config get-value core/project --quiet', { silent: true }).stdout;
-            return projectId.trim();
-        } catch (ex) {
-            this.log.error('Unable to determine the default Google Cloud Project ID');
-            return undefined;
-        }
-    }
-
-    defaultServiceNameChoices(defaultServiceExists) {
-        if (this.applicationType === 'monolith') {
-            return defaultServiceExists ? ['default', _.kebabCase(this.baseName)] : ['default'];
-        }
-        if (this.applicationType === 'gateway') {
-            return ['default'];
-        }
-
-        return [_.kebabCase(this.baseName)];
-    }
-
     get prompting() {
         return {
+            askForPath() {
+                if (this.abort) return undefined;
+                if (this.applicationType !== 'gateway') return undefined;
+                const messageAskForPath = 'Enter the root directory where the microservices are located';
+                const prompts = [
+                    {
+                        type: 'input',
+                        name: 'directoryPath',
+                        message: messageAskForPath,
+                        default: this.directoryPath || '../',
+                        validate: input => {
+                            const path = this.destinationPath(input);
+                            if (shelljs.test('-d', path)) {
+                                const appsFolders = this._getMicroserviceFolders(input);
+
+                                if (appsFolders.length === 0) {
+                                    return `No microservices are found in ${path}`;
+                                }
+                                return true;
+                            }
+                            return `${path} is not a directory or doesn't exist`;
+                        },
+                    },
+                ];
+
+                return this.prompt(prompts).then(props => {
+                    this.directoryPath = props.directoryPath;
+                    // Patch the path if there is no trailing "/"
+                    if (!this.directoryPath.endsWith('/')) {
+                        this.log(chalk.yellow(`The path "${this.directoryPath}" does not end with a trailing "/", adding it anyway.`));
+                        this.directoryPath += '/';
+                    }
+                    this.appsFolders = this._getMicroserviceFolders(this.directoryPath);
+                    this.log(chalk.green(`${this.appsFolders.length} applications found at ${this.destinationPath(this.directoryPath)}\n`));
+                });
+            },
+            askForApps() {
+                if (this.regenerate) return undefined;
+                if (this.applicationType !== 'gateway') return undefined;
+                const messageAskForApps = 'Which microservice applications do you want to include in your configuration?';
+                const prompts = [
+                    {
+                        type: 'checkbox',
+                        name: 'chosenApps',
+                        message: messageAskForApps,
+                        choices: this.appsFolders,
+                        default: this.defaultAppsFolders,
+                        validate: input => (input.length === 0 ? 'Please choose at least one application' : true),
+                    },
+                ];
+
+                return this.prompt(prompts).then(props => {
+                    this.appsFolders = props.chosenApps;
+                    dockerPrompts.loadConfigs.call(this);
+                });
+            },
             askForProjectId() {
                 if (this.abort) return;
                 const done = this.async();
@@ -151,7 +184,7 @@ module.exports = class extends BaseGenerator {
                         type: 'input',
                         name: 'gcpProjectId',
                         message: 'Google Cloud Project ID',
-                        default: this.defaultProjectId(),
+                        default: this._defaultProjectId(),
                         validate: input => {
                             if (input.trim().length === 0) {
                                 return 'Project ID cannot empty';
@@ -236,7 +269,7 @@ module.exports = class extends BaseGenerator {
                         type: 'list',
                         name: 'gaeServiceName',
                         message: 'Google App Engine Service Name',
-                        choices: this.defaultServiceNameChoices(this.defaultServiceExists),
+                        choices: this._defaultServiceNameChoices(this.defaultServiceExists),
                         default: this.gaeServiceName ? this.gaeServiceName : 0,
                     },
                 ];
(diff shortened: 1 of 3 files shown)


SudharakaP:
@mshima : I don't know if you could help me on this one but there's a small problem which I couldn't figure out in this PR. If I didn't add the `if` condition at [line 146 of index.js](https://github.com/jhipster/generator-jhipster/pull/11969/files#diff-db6b3ff2520083c61a939019cc621219R146), it seems to give the following error; 

<pre><font color="#CC0000">ERROR! The &quot;path&quot; argument must be of type string. Received undefined</font>
TypeError [ERR_INVALID_ARG_TYPE]: The &quot;path&quot; argument must be of type string. Received undefined
<font color="#555753">    at validateString (internal/validators.js:117:11)</font>
<font color="#555753">    at Object.join (path.js:1039:7)</font>
    at module.exports.destinationPath (/home/sudharaka/IdeaProjects/generator-jhipster/generators/generator-base-private.js:71:22)
    at module.exports.getMicroserviceFolders (/home/sudharaka/IdeaProjects/generator-jhipster/generators/gae/index.js:147:38)
    at Object.&lt;anonymous&gt; (/home/sudharaka/IdeaProjects/generator-jhipster/node_modules/<u style="text-decoration-style:single">yeoman-generator</u>/lib/index.js:893:25)
    at /home/sudharaka/IdeaProjects/generator-jhipster/node_modules/<u style="text-decoration-style:single">run-async</u>/index.js:25:25
    at new Promise (&lt;anonymous&gt;)
    at /home/sudharaka/IdeaProjects/generator-jhipster/node_modules/<u style="text-decoration-style:single">run-async</u>/index.js:24:19
    at runLoop.add.once.once (/home/sudharaka/IdeaProjects/generator-jhipster/node_modules/<u style="text-decoration-style:single">yeoman-generator</u>/lib/index.js:894:11)
<font color="#555753">    at processImmediate (internal/timers.js:456:21)</font> {
  code: <font color="#4E9A06">&apos;ERR_INVALID_ARG_TYPE&apos;</font>
}
</pre>

I am very curious as I couldn't figure out where this error is coming from and how that function is called; perhaps due to my lack of knowledge about Yeoman. Do you have any idea? :thinking:

mshima:
> I am very curious as I couldn't figure out where this error is coming from and how that function is called; perhaps due to my lack of knowledge about Yeoman. Do you have any idea? Thanks a bunch. 😄

That should be the argument `jhipster gae input`, using this way is kind of confusing.
But there is no argument registered `this.agument(...)` at the constructor. 😕 
 
This is a bug, if there is an argument, it should be registered.
In this case input probably is `[]`. 
I will take a look later this week if the empty array is supposed to be there.

generators/gae/index.js
```diff
@@ -141,8 +142,93 @@ module.exports = class extends BaseGenerator {
...
+                                    return `No microservices are found in ${path}`;
+                                }
+                                return true;
+                            }
+                            return `${path} is not a directory or doesn't exist`;
+                        },
+                    },
+                ];
+
+                this.prompt(prompts).then(props => {
```
mshima, line 198:
```suggestion
                return this.prompt(prompts).then(props => {
```
And remove 
```
const done = this.async();
```
and 
```
done();
```

Returning the prompt promise is cleaner.

generators/gae/index.js
```diff
@@ -141,8 +142,93 @@ module.exports = class extends BaseGenerator {
...
 
+    getMicroserviceFolders(input) {
+        if (!input) return null;
+        const destinationPath = this.destinationPath(input);
+        const files = shelljs.ls('-l', destinationPath);
+        const appsFolders = [];
+
+        files.forEach(file => {
+            if (file.isDirectory()) {
+                if (shelljs.test('-f', `${destinationPath}/${file.name}/.yo-rc.json`)) {
```
mshima, line 153: I prefer ```fs.existsSync(`${destinationPath}/${file.name}/.yo-rc.json`)``` here.

Using shelljs should be avoided.

mshima:
> @mshima : There's a small problem which I encountered which I was very curious about. If I didn't add the `if` condition at [line 146 of index.js](https://github.com/jhipster/generator-jhipster/pull/11969/files#diff-db6b3ff2520083c61a939019cc621219R146), it seems to give the following error;
> 
> ERROR! The "path" argument must be of type string. Received undefined
> TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
>     at validateString (internal/validators.js:117:11)
>     at Object.join (path.js:1039:7)
>     at module.exports.destinationPath (/home/sudharaka/IdeaProjects/generator-jhipster/generators/generator-base-private.js:71:22)
>     at module.exports.getMicroserviceFolders (/home/sudharaka/IdeaProjects/generator-jhipster/generators/gae/index.js:147:38)
>     at Object.<anonymous> (/home/sudharaka/IdeaProjects/generator-jhipster/node_modules/yeoman-generator/lib/index.js:893:25)
>     at /home/sudharaka/IdeaProjects/generator-jhipster/node_modules/run-async/index.js:25:25
>     at new Promise (<anonymous>)
>     at /home/sudharaka/IdeaProjects/generator-jhipster/node_modules/run-async/index.js:24:19
>     at runLoop.add.once.once (/home/sudharaka/IdeaProjects/generator-jhipster/node_modules/yeoman-generator/lib/index.js:894:11)
>     at processImmediate (internal/timers.js:456:21) {
>   code: 'ERR_INVALID_ARG_TYPE'
> }

How to reproduce this error?

generators/gae/index.js
SudharakaP: @mashima : Not sure how this can be done it gives the lint error; `ESLint: Method 'askForPath' expected no return value.(consistent-return)` :thinking:

generators/gae/index.js
SudharakaP: Thanks. Done. :+1:

mshima:
@SudharakaP I just realized you are adding the argument input.

You should register at the generator:
https://yeoman.github.io/generator/Generator.html#argument

generators/gae/index.js
mshima: Need to change the others returns too:
`if (this.abort) return undefined;`

Or add eslint disable rule.

SudharakaP:
@mshima : 

> @SudharakaP how to reproduce this error?

For this we need to run the subgenerator on top of a gateway application. I use the following JDL file to create the gateways and micro-services. 

https://github.com/jhipster/jdl-samples/blob/master/microservice-ecommerce-store-k8s.jdl

Then run `jhipster gae` on the application named `store` which is the gateway. Although you need to login to a Google Cloud account in order for the generator to work correctly. Let me know if you have any further questions.

SudharakaP:
> @SudharakaP I just realized you are adding the argument input.
> 
> You should register at the generator:
> https://yeoman.github.io/generator/Generator.html#argument
> 
> Register at jhipster commands (can be integrated):
> 
> https://github.com/jhipster/generator-jhipster/blob/990eeb70ce5d5c666ca01bf8e1bb6ded4f7ec6d5/cli/commands.js#L59-L61

@mshima : But the command `jhipster gae` shouldn't take any arguments. The intention of the argument `input` to the function is only to pass it on to `destinationPath` function. It's not a cli argument. Even in the function `defaultServiceNameChoices` we use an argument so I am quite confused why this function seems to be called automatically by yeoman. :thinking:

generators/gae/index.js
SudharakaP: Ah, thanks. Done. :smile:

mshima:
> > @SudharakaP I just realized you are adding the argument input.
> > You should register at the generator:
> > https://yeoman.github.io/generator/Generator.html#argument
> > Register at jhipster commands (can be integrated):
> > https://github.com/jhipster/generator-jhipster/blob/990eeb70ce5d5c666ca01bf8e1bb6ded4f7ec6d5/cli/commands.js#L59-L61
> 
> @mshima : But the command `jhipster gae` shouldn't take any arguments. The intention of the argument `input` to the function is only to pass it on to `destinationPath` function. It's not a cli argument. Even in the function `defaultServiceNameChoices` we use an argument so I am quite confused why this function seems to be called automatically by yeoman. 🤔

Every method from the generator is queued to the corresponding priority or to default priority if the method name does not match a priority.

To workaround this you should prefix with `_` or add to a super class.

generators/gae/index.js
```diff
@@ -141,8 +143,91 @@ module.exports = class extends BaseGenerator {
...
+                        type: 'checkbox',
+                        name: 'chosenApps',
+                        message: messageAskForApps,
+                        choices: this.appsFolders,
+                        default: this.defaultAppsFolders,
+                        validate: input => (input.length === 0 ? 'Please choose at least one application' : true),
+                    },
+                ];
+
+                this.prompt(prompts).then(props => {
```
mshima, line 225: Return prompt here too.

SudharakaP: Thanks much for the explanation. Done. :+1:

SudharakaP:
> > > @SudharakaP I just realized you are adding the argument input.
> > > You should register at the generator:
> > > https://yeoman.github.io/generator/Generator.html#argument
> > > Register at jhipster commands (can be integrated):
> > > https://github.com/jhipster/generator-jhipster/blob/990eeb70ce5d5c666ca01bf8e1bb6ded4f7ec6d5/cli/commands.js#L59-L61
> > 
> > 
> > @mshima : But the command `jhipster gae` shouldn't take any arguments. The intention of the argument `input` to the function is only to pass it on to `destinationPath` function. It's not a cli argument. Even in the function `defaultServiceNameChoices` we use an argument so I am quite confused why this function seems to be called automatically by yeoman. thinking
> 
> Every method from the generator is queued to the corresponding priority or to default priority if the method name does not match a priority.
> 
> To workaround this you should prefix with `_` or add to a super class.

@mshima : Ah, now I get it. That fixed the issue. :+1:

mshima:
> @mshima : Ah, now I get it. That fixed the issue; thanks a lot. 👍

Looks good to me now.

Just a small suggestion is to move the non task function to after the priorities for organization and add `_` to others non task functions.

SudharakaP:
> > @mshima : Ah, now I get it. That fixed the issue; thanks a lot. +1
> 
> Looks good to me now.
> 
> Just a small suggestion is to move the non task function to after the priorities for organization and add `_` to others non task functions.

I've done that at de06aca. Does that look good? :thinking:

mshima:
> @mshima : I've done that at [de06aca](https://github.com/jhipster/generator-jhipster/commit/de06aca12e283c487e023eeca34837444a16069c). Does that look good? 🤔

Yep. I think it cleaner.

SudharakaP:
@mshima : Wonderful. Thanks much for the review. Really appreciate it. :+1: :love_you_gesture:

mshima:
@SudharakaP I’m glad I could help.

SudharakaP:
Bug bounty submitted at https://opencollective.com/generator-jhipster/expenses/21088

PR #11975, opened 2020-06-22
Set default private license

SandraAhlgrimm:
A default value for the license should be set as it is required by multiple other applications. So it would ease my daily duties as a developer.
On the other hand there is automatically a license depending on the country where the new application gets generated. In order to avoid confusion here, a default value can help.

Another possibility could be "unlicensed". I am open for discussion.


jdubois:
+1
- This is what we do in the `package.json` (the keyword used is "UNLICENSED" which means this is a private project) -> we need the same thing here to be consistent
- This license name is required by some third-parties

SandraAhlgrimm:
got duplicated here https://github.com/jhipster/generator-jhipster/pull/11977

PR #11988, opened 2020-06-25
Dependabot adjustments

mshima:
- PR branchs considers target branch.
- On conditions don’t show at actions list.
- dependabot builds are duplicating at dependabot/* branch and PR

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -18,7 +18,13 @@
 #
 
 name: Angular
-on: [push, pull_request]
+on:
+    push:
+        branches-ignore:
+            - 'dependabot/**'
+    pull_request:
+        branches:
+            - '*'
 env:
     JHI_RUN_APP: 1
     JHI_JDK: 11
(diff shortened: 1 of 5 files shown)


SudharakaP:
Also my suggestion above will skip the branches when merged to master too; but do we really need to have the CI running on master for something we don't check before merging. I would suggest that's not quite optimal right. After all we should probably check what we check before merging not after merging. :smile:

mshima:
> And remove the if conditions. wdyt? 😄

```
on:
    pull_request:
        branches-ignore:
            - 'dependabot/**'
```
Doesn’t work branches-ignore matches target branch only, should be master in this case.

SudharakaP:
> > And remove the if conditions. wdyt? smile
> 
> ```
> on:
>     pull_request:
>         branches-ignore:
>             - 'dependabot/**'
> ```
> 
> Doesn’t work branches-ignore matches target branch only, should be master in this case.

@mshima : Ah, got it. This is the same case with Azure which I got to know later. :smile:

PR #11997, opened 2020-06-26
Add prettier check workflow

mshima:
This PR add a prettier:check script and a github workflow to check.
`skip-ci` is not be applied to this workflow since README.md and others text files should be formatted too.

An alternative is to run prettier:check at pretest script.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
     - package-ecosystem: npm
-      directory: "/"
+      directory: '/'
       schedule:
           interval: weekly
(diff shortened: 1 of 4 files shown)


murdos:
I'd rather add this check as an additional step of the `generator` workflow

pascalgrimaud:
I agree with @murdos : no need to add another parallel job, include it in generator workflow with npm test would be good

mshima:
@murdos @pascalgrimaud changed to use generator workflow.
Moved the `skip-ci` condition to `npm test` step, since we need to check yml and md files changes for every PR.

PR #12021, opened 2020-07-01
Remove java8 support

avdev4j:
According to v7 roadmap. Remove Java8 support. I suggest to set Java 11 as default version and allow from 1.9 to 14. 
---

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -20,7 +20,7 @@ const semver = require('semver');
 const packagejs = require('../package.json');
 
 // Version of Java
-const JAVA_VERSION = '1.8'; // Java version is forced to be 1.8. We keep the variable as it might be useful in the future.
+const JAVA_VERSION = '11'; // Java version is forced to be 11. We keep the variable as it might be useful in the future.
 
 // Version of Node, Yarn, NPM
 const NODE_VERSION = '12.16.1';
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -60,7 +60,7 @@ description = ""
 
 sourceCompatibility=<%= JAVA_VERSION %>
 targetCompatibility=<%= JAVA_VERSION %>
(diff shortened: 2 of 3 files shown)


cbornet:
In that case, Java 11 should be the min version. `var` FTW !

avdev4j:
To be honest, I'm doubtful. But I'm ok to set min version at 11 ;)

jdubois:
Yes in that case we should go to 11 at the minimum.
But I’m not convinced we should drop Java 8, it’s still the default for many people.

avdev4j:
I have just change the min to 11.
@jdubois I did it because of the v7 roadmap issue. However we still be able to run with java8 if we change the java version in the pom.xml but it could be complicated if we introduce "var" in generated code, for example.

agilob:
I'm not sure if this is mostly about compatibility with `var` keyword, it's certainly a big one, but I'm sure we can think about other features that aren't supported on 8, eg.  allowing autogenerated config for modern GCs, better container-awareness, introduced after 9.

avdev4j:
for sure, yes.

pascalgrimaud:
+1 for removing Java8 support
If some companies still want to use Java8, the users can easily do the change manually.

For your information:
- all our CI builds are running with Java11 since months, no more tests with Java8 anymore
- the CI-CD sub-generator should already generated files for Java11 -> can you check this part quickly plz @avdev4j ?

cbornet:
IMO, Java 11 should be the default everywhere (pom, docker, ...). But it's probably good to keep code compatibility with Java 8 (ie no var) for the moment.

jdubois:
Yes, +1 for Java 11 by default while keeping Java 8 compatibility (I haven't used "var" yet, by the way)

avdev4j:
I think it's a good compromise. We just need to check for new codes and avoid features that could be not supported by Java 8.

@cbornet remind me that I need to check Docker files before merge this PR ;)

avdev4j:
I have set version to 11 but allow versions between 1.8 and 14.

PR #12029, opened 2020-07-03
Add id fluent setter to Entity.

mshima:
Currently we are not generating fluent id setter.

---


murdos:
I think the reason was that we're not supposed to manually set the id of an entity.

mshima:
> I think the reason was that we're not supposed to manually set the id of an entity.

I think so, but for specific use cases this can be useful.
Eg: `foo.bar(new bar().id(barId));`.

For completeness I think it is better to include.

PR #12038, opened 2020-07-05
Refactor jhipster configuration.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/10528
- Refactor app/client/server/common/languages generators to use jhipsterConfig.
- App generator compose with client/server/common generators in the same task, so they will run in parallel (run a priority on every generator before moving to another priority).
- Implements `jhipster --defaults`
- Simplifies configuration by applying rules in fewer points.

Breaking changes:
The workflow was changed, configs are available at the generator instance only at default priority now, this prevents synchronization problems.
- Runtime only configs is available at configOptions and stored configs are available at jhipsterConfig.
- setup*Options will not set defaults anymore, they are executed at generator constructor, creating synchronization problems with prompts which are executed later.

Migration:
- Use the configuration at the source configOptions or jhipsterConfig.
- Move your customization to default priority.
- If you still want the configuration to be available at generator instance at an earlier priority, load it.

---


pascalgrimaud:
I think it will break all existing blueprints, right ? If yes, maybe some docs or examples to help the authors of blueprint to migrate to v7 : it can be done later, no worry for that

mshima:
The only breakage I can see is blueprints that customises the workflow by overriding the priorities functions.

Eg:
https://github.com/jhipster/jhipster-kotlin/blob/fdd8ca9b019f13b6adf4ccdf0cd9d1208c6f8d81/generators/app/index.js#L30-L35

It would break if I changed the name of https://github.com/jhipster/generator-jhipster/blob/8a939555af2933247174b3c627d7409ac7327ee6/generators/app/index.js#L259

This is a too advanced use case to be documented in my opinion.

mshima:
There is only a suggesting but I will create an issue later:

`setup*Options` is kind of deprecated.
https://github.com/jhipster/jhipster-kotlin/blob/fdd8ca9b019f13b6adf4ccdf0cd9d1208c6f8d81/generators/server/index.js#L48

The old implementation was doing too much at the wrong time.
Now it can be replaced by `Object.assign(this, this.jhipsterConfig);` at default priority.

mshima:
@pascalgrimaud, @jhipster/developers is there anyone good at drawing flow charts, would be nice to have one showing jhipster 7 workflow.

pascalgrimaud:
For doing something like that ? https://www.jhipster.tech/microservices-architecture/
I don't remember the tools we use here...

mshima:
@pascalgrimaud something like this.
Before this PR:
jhipster6

After this PR:
jhipster7

Flow diagrams with blueprints and sbs-blueprints would be useful too.

MathieuAA:
@mshima You rock. Just had to say it.

generators/generator-base.js
```diff
@@ -2141,8 +2111,170 @@ module.exports = class extends PrivateBase {
...
+            this.jhipsterConfig.searchEngine = options.searchEngine;
+        }
+        if (options.build) {
+            this.jhipsterConfig.buildTool = options.build;
+        }
+        if (options.websocket) {
+            this.jhipsterConfig.websocket = options.websocket;
+        }
+        if (options.jhiPrefix) {
+            this.jhipsterConfig.jhiPrefix = options.jhiPrefix;
```
pascalgrimaud, line 2171: @mshima : some tests on my side, it seems this.jhipsterConfig.jhiPrefix is always empty and is not initialized with `jhi` value.

My quick fix would be:
```
        if (options.jhiPrefix) {
            this.jhipsterConfig.jhiPrefix = options.jhiPrefix;
        } else {
            this.jhipsterConfig.jhiPrefix = 'jhi'
        }
```

But I know it's not the correct fix, because it should read the default value `jhi` from `generator-defaults.js`
 Can you help here plz ?

mshima: @pascalgrimaud https://github.com/jhipster/generator-jhipster/pull/12063

PR #12039, opened 2020-07-05
Remove jcore dep

MathieuAA:
This PR removes the uses of the JCore dep in favor of the files now in the generator.

---

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -22,7 +22,6 @@ const path = require('path');
 const shelljs = require('shelljs');
 const ejs = require('ejs');
 const _ = require('lodash');
-const jhiCore = require('jhipster-core');
 const fs = require('fs');
 const crypto = require('crypto');
 const randexp = require('randexp');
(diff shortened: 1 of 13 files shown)


generators/utils.js
```diff
@@ -577,7 +577,7 @@ function decodeBase64(string, encoding = 'utf-8') {
 }
 
 function loadYoRc(filePath = '.yo-rc.json') {
-    if (!jhiCore.FileUtils.doesFileExist(filePath)) {
+    if (!FileUtils.doesFileExist(filePath)) {
```
mshima, line 580: Use `fs.existsSync` instead.

MathieuAA: There's a notable difference between the FileUtils' version and the Node FS' one: the latter only checks the path (https://nodejs.org/api/fs.html#fs_fs_existssync_path) and doesn't core if it's a folder or a regular file.

mshima: I don't think we need that, I would prefer fs.existsSync for simplicity.
This kind of test is not done at yeoman-generator for `.yo-rc.json`, it will just throw if it cannot parse.

Anyway your call.

mshima: In fact I think loadYoRc should throw if `.yo-rc.json` is a folder.
This is not a supported scenario at all, it's a bad problem and should not be ignored by returning undefined.

MathieuAA:
Thanks Marcelo!

mshima:
Not sure why, but `npm ci` is printing an error at macOS now, github CI isn't affected.
See https://github.com/jhipster/generator-jhipster/pull/12041

Issue #12042, opened 2020-07-05
Side affect mock DateTimeProvider


Romancha:
##### **Overview of the issue**

Using mock for dateTimeProvider may side affect users test that use dateTimeProvider. To avoid side affect we need to use @MockBean

##### **Motivation for or Use Case**

I have my own integration tests, which use the createdDate, lastModifiedDate in the verification of business logic. I was a bit confused when my tests fell and saw that the createdDate, lastModifiedDate in my objects equals Instant.now() - minus 4 days. The reason for this is using @Mock DateTimeProvider in UserServiceIT.

##### **Reproduce the error**

1. Create a test in which an object with the createdDate, lastModifiedDate will be created.
2. Compare createdDate, lastModifiedDate with current date.

**Expected result:** createdDate, lastModifiedDate have Instant.now() value
**Actual result:** createdDate, lastModifiedDate have Instant.now() - 4 days value

##### **Related issues**

Also fixed in jhipster-kotlin - https://github.com/jhipster/jhipster-kotlin/issues/246

##### **Suggest a Fix**

https://github.com/jhipster/generator-jhipster/pull/12043

##### **JHipster Version(s)**

generator-jhipster@6.8.0

##### **JHipster configuration**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "org.romancha.test"
    },
    "jhipsterVersion": "6.8.0",
    "applicationType": "monolith",
    "baseName": "test",
    "packageName": "org.romancha.test",
    "packageFolder": "org/romancha/test",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "postgresql",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "gradle",
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "-",
    "embeddableLaunchScript": false,
    "useSass": true,
    "clientPackageManager": "npm",
    "clientFramework": "react",
    "clientTheme": "none",
    "clientThemeVariant": "",
    "creationTimestamp": 1593984745881,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "entitySuffix": "",
[...]

PR #12043, opened 2020-07-05
Remove side effect of mock dateTimeProvider

Romancha:
Using mock for dateTimeProvider may affect users test that use dateTimeProvider. To avoid side affect we need to use @MockBean

[issue](https://github.com/jhipster/generator-jhipster/issues/12042)
[pr jhipster-kotlin](https://github.com/jhipster/jhipster-kotlin/pull/247)


atomfrede:
Thanks @Romancha and congrats to your first contribution! @jhipster/developers How do we handle this? Just merge it to master or also in v6 maintenance branch? If also going to v6 is there some automated way?

pascalgrimaud:
@atomfrede : it depends on the issue. Is it an important issue which needs a new patch release ? Or is it related to more custom code ? As all our CI passed...

atomfrede:
Guess it is not critical and only happens under special circumstances.

Issue #10575, opened 2019-10-08
Integrate VueJS blueprint as a mainline generator (like Angular and React)


cmoine:
##### **Overview of the feature request**

The actual flow for VueJS is awkward (CLI and https://start.jhipster.tech/#/generate-application as well):
- Choose React/Angular generation
- THEN choose VueJS blueprint

Moreover the web site claims:
Spring Boot + Angular / React / Vue Web applications
I think it is very frustrating/confusing that we cannot choose VueJS instead of React/Angular. I am pretty sure some users just give up after that and do not dig deeper.

##### **Motivation for or Use Case**

VueJS is really gaining popularity, it really makes sense to do it. Some recent articles are now only comparing React and VueJS, Angular is no more part of the comparison :-(

Moreover, I think the VueJS generator would gain popularity and stability since it would be part of the release train. At the moment, a new JHipster version does not imply the blueprints have been tested against it (from JHipster perspective) (tell me if I'm wrong).

##### **Related issues or PR**

None, but I would be glad to contribute for this feature in a PR.

pascalgrimaud:
It's an important discussion
cc @jhipster/developers 
Plz, vote the original ticket with +1 or -1

avdev4j:
If we want to do that we need to:
- handle blueprint specific questions
- handle all other official blueprints (maybe in other PRs) -> so thinking of a modular code to dynamically adding fields from blueprint questions (and ofc the backend part related to)

wmarques:
I'm ok for merging Vue, the blueprint is quite stable and very well made :) 

Just one thought about that quote:

> VueJS is really gaining popularity, it really makes sense to do it. Some recent articles are now only comparing React and VueJS, Angular is no more part of the comparison :-(

Both 3 frameworks are made to build applications, if you do shitty things with Angular, ofc the app will suck, same goes for Vue and React.
Both have pros/cons, I always see the 3 frameworks on "comparison articles".
Angular is very widely used especially in the Java ecosystem.

I agree that Vue is great etc, I even use it myself for small personal projects, but don't discard other technologies because they do have less Github stars or less representation in articles, that doesn't mean anything.

I would have said the same for React of course.

mraible:
[...]

PR #12066, opened 2020-07-13
Rename page to vue-page

pascalgrimaud:
Related to #10575
Following this https://github.com/jhipster/generator-jhipster/pull/12064#pullrequestreview-447060090

---


mshima:
I don’t agree with the rename.
Feature parity is good, but is really hard for one developer to implement all 3 clients, especially for non-essential feature.
There are needles that are exclusive to angular and nobody cares enough to add to react.

I think we should:
- prepare page command for others clients.
- throw if the client isn’t supported
- integrate into app generator for reproducibility.

pascalgrimaud:
The `throw if the client isn't supported` is already implemented here: https://github.com/jhipster/generator-jhipster/blob/master/generators/page/index.js#L77-L79
But maybe not correctly implemented :-)

I agree with you @mshima : maybe we should keep as `page` and if there are contributors who want to code the Angular and React part, they are welcome. Otherwise, it will throw an error like today

pascalgrimaud:
following the refactoring from @mshima, no need anymore this PR

PR #12067, opened 2020-07-13
Improvements to blueprints composing

mshima:
- Fixes a problem that if the blueprint constructor throws an error, jhipster will treat like the blueprint doesn't exists.
- Move blueprint related functions and logic to generator-base-blueprint.
- Create a common merging blueprint file and option configs, for cli and generator.

---


generators/utils.js
```diff
@@ -667,12 +669,40 @@ function loadBlueprintsFromConfiguration(config) {
...
+    if (typeof blueprints === 'string') {
         return blueprints
             .split(',')
             .filter(el => el != null && el.length > 0)
             .map(blueprint => parseBlueprintInfo(blueprint));
     }
+    return [];
+}
+
+function mergeBlueprints(...blueprintsToMerge) {
```
MathieuAA, line 684: Judging by the tests, this function does two things:
  1. flattens the internal structure of the array `([ [{}], [{}] ] => [ {}, {} ])`
  2. it resolves versions based on the input while removing duplicates

It's not clear what this function does by just looking at the code (the tests helped me)

You can do all that by:
  - flattening the array (Array#flat)
  - creating a map for (2), filling the map by using your algorithm and then getting the map's values

generators/utils.js
```diff
@@ -684,15 +714,18 @@ function parseBluePrints(blueprints) {
...
         bpName = bpName.slice(0, idx);
+        return {
+            name: bpName,
+            version,
+        };
     }
     return {
         name: bpName,
-        version,
     };
```
MathieuAA, line 729: for consistency blueprintName

test/blueprint/constructor-error-blueprint.spec.js
```diff
@@ -0,0 +1,33 @@
...
+const EnvironmentBuilder = require('../../cli/environment-builder');
+
+describe('JHipster application generator with blueprint with constructor error', () => {
+    describe('generate monolith application with scoped blueprint', () => {
+        it('rejects the environment', () => {
+            return helpers
+                .create('jhipster:app', {}, { createEnv: EnvironmentBuilder.createEnv })
+                .inTmpDir(dir => {
+                    // Fake the presence of the blueprint in node_modules
+                    const fakeBlueprintModuleDir = path.join(dir, 'node_modules/generator-jhipster-myblueprint');
```
MathieuAA, line 14: 'node_modules/generator-jhipster-myblueprint' => 'node_modules', 'generator-jhipster-myblueprint'

test/blueprint/constructor-error-blueprint.spec.js
```diff
@@ -0,0 +1,33 @@
...
+                    fse.ensureDirSync(fakeBlueprintModuleDir);
+                    fse.copySync(path.join(__dirname, '../../test/templates/throwing-blueprint'), fakeBlueprintModuleDir);
```
MathieuAA, line 16: '../../test/templates/throwing-blueprint' => '..', '..', 'test', 'templates', 'throwing-blueprint'

test/templates/throwing-blueprint/generators/server/index.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    get install() {
+        return super._install();
+    }
+
+    get end() {
+        return super._end();
+    }
+};
```
MathieuAA, line 34: this will be used in a failing case for the constructor, right ? why have all the methods below?

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
-
-                if (!this.options.skipChecks) {
-                    const namespaces = blueprints.map(bp => jhipsterUtils.packageNameToNamespace(bp.name));
-                    // Verify if the blueprints has been registered.
-                    const missing = namespaces.filter(namespace => !this.env.isPackageRegistered(namespace));
-                    if (missing && missing.length > 0) {
-                        this.error(`Some blueprints were not found ${missing}, you should install them manually`);
-                    }
+        // Load and verify blueprints, should be executed only once
+        if (!this.configOptions.blueprintConfigured) {
```
MathieuAA, line 94: maybe this block's content should be in another function then

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
 
-                this.configOptions.blueprintsVerified = true;
+            let otherModules = this.jhipsterConfig.otherModules || [];
+            if (blueprints && blueprints.length > 0) {
+                blueprints.forEach(blueprint => {
+                    blueprint.version = this.findBlueprintVersion(blueprint.name) || blueprint.version;
+                });
+
+                // Remove potential previous value to avoid duplicates
+                otherModules = otherModules.filter(module => this.blueprints.findIndex(bp => bp.name === module.name) === -1);
```
MathieuAA, line 122: if you need write comments, then it should be moved to another function with an explicit name :)

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
+                this.warning('--blueprint option is deprecated. Please use --blueprints instead');
+                if (!argvBlueprints.split(',').includes(blueprint)) {
+                    argvBlueprints = `${blueprint},${argvBlueprints}`;
                 }
+            }
+            const blueprints = jhipsterUtils.mergeBlueprints(
+                jhipsterUtils.parseBluePrints(argvBlueprints),
+                jhipsterUtils.loadBlueprintsFromConfiguration(this.config)
+            );
+            // Run a lookup to find blueprints.
```
MathieuAA, line 109: if you need write comments, then it should be moved to another function with an explicit name :)

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
-                    // Verify if the blueprints has been registered.
-                    const missing = namespaces.filter(namespace => !this.env.isPackageRegistered(namespace));
-                    if (missing && missing.length > 0) {
-                        this.error(`Some blueprints were not found ${missing}, you should install them manually`);
-                    }
+        // Load and verify blueprints, should be executed only once
+        if (!this.configOptions.blueprintConfigured) {
+            this.configOptions.blueprintConfigured = true;
+            let argvBlueprints = this.options.blueprints || '';
+            // check for old single blueprint declaration
```
MathieuAA, line 97: if you need write comments, then it should be moved to another function with an explicit name :)

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
+            ...this.options,
+            configOptions: this.configOptions,
+            ...extraOptions,
+            jhipsterContext: this,
+        };
+
+        const blueprintGenerator = this.composeWith(generatorNamespace, finalOptions, true);
+        if (blueprintGenerator instanceof Error) {
+            throw blueprintGenerator;
+        }
```
MathieuAA, line 188: I don't get it. A generator is a generator, right? why would something called `composeWith` **return** an Error?

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
+        if (!this.env.isPackageRegistered(generatorName)) {
+            this.env.lookup({ filterPaths: true, packagePatterns: blueprint });
+        }
+        if (!this.env.get(generatorNamespace)) {
+            this.debug(
+                `No blueprint found for blueprint ${chalk.yellow(blueprint)} and ${chalk.yellow(
+                    subGen
+                )} subgenerator: falling back to default generator`
+            );
+            return false;
```
MathieuAA, line 175: There's an issue with what's returned: I see a boolean here (perhaps it should be `undefined`), but also an object l.190.

generators/generator-base-blueprint.js
```diff
@@ -84,51 +90,187 @@ module.exports = class extends BaseGenerator {
...
         }
         return useBlueprints;
     }
+
+    /**
+     * Compose external blueprint module
+     * @param {string} blueprint - name of the blueprint
+     * @param {string} subGen - sub generator
+     * @param {any} options - options to pass to blueprint generator
+     */
```
MathieuAA, line 157: having a @return annotation here would help

generators/generator-base-blueprint.js
```diff
@@ -16,6 +16,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+const chalk = require('chalk');
```
MathieuAA, line 19: (I don't have the option to put it below so I put this here): the generator has a lot of public methods, are they all tested? Or should some of them be private (or moved to other files because they don't have the same responsibility)?

mshima: They are all private, there is no reason to use outside.
We only guarantee stability for generator-base and generator-constants.
https://www.jhipster.tech/modules/creating-a-blueprint/

generators/generator-base-blueprint.js
mshima: This is a legacy behaviour of yeoman-generator.
Nothing I can do for yeoman-generator 4.x.
If you want to create an issue at yeoman-generator repository for 5.x, I can handle it.

generators/generator-base-blueprint.js
mshima: Changed to private.

generators/generator-base-blueprint.js
mshima: I don't agree.
Too many methods makes less linear and more complicated.
The alternative is to let the developer guess or look at the api documentation.

generators/generator-base-blueprint.js
mshima: Same as above.

generators/generator-base-blueprint.js
mshima: Same as above.
And in many cases like this one, the option is deprecated.

generators/utils.js
mshima: Array.flat requires node 12, so I did the bump.
Maps doesn't have order guarantee.

MathieuAA: Maps have order guarantee :)
From the MDN doc:
The Map object holds key-value pairs and remembers the original insertion order of the keys.

generators/generator-base-blueprint.js
MathieuAA: >  Too many methods makes less linear and more complicated.

Yup, and that's when one can see if the private code really belongs in its current place or if it's to be moved elsewhere.
My point is more about clean code than something else here. A function/method should be:
  - well named
  - short so as to understand what it does according to its name, around 8-10 lines
  - etc.

A comment inside this kind of block sometimes is an indication of something that can be improved. For instance, this kind of comments should be kept for really complex code (that almost never happens, unless you're working on a new spacecraft reactor or something that has a visible algorithmic complexity).
If you deem the comment code is clear enough, then the comment isn't necessary. Otherwise, moving it should be an option to consider.
A better explanation than mine: https://refactoring.guru/smells/comments.

mshima: `otherModules` is used by templates, I can’t see a reason for extracting into a function.
We probably should start merging blueprints and modules at load instead of before save.
It duplicating and is implemented wrong, otherModules version will take precedence, this is wrong, I fix that later.

Your comments to extract to methods were:
- One field that should be improved. But this PR is big enough.
- A external api that could be improved.
- One deprecated option.

generators/utils.js
mshima: Yep, only for english js, not for portuguese js... 🤕 
Order is not mentioned at portuguese translation.

pascalgrimaud:
Thanks for the review @MathieuAA 
So let's merge it

pascalgrimaud:
Good job @mshima as usual !

Issue #12085, opened 2020-07-16
GCP/GAE region mismatch between gcloud app and gcloud sql commands


leozin:
##### **Overview of the issue**

When jhipster gae tries to generate a sql instance, it uses the region set by gcloud app, which is loaded by running 
```
gcloud app describe --format="value(locationId) 
```

However, sometimes the region described in gcloud app doesn't exist in gcloud sql, and the error below occurs:

```
Creating New Cloud SQL Instance

Running: gcloud sql instances create "my-app" --region='europe-west' --project=my-app-12345
✖ ERROR: (gcloud.sql.instances.create) HTTPError 400: Invalid request: Invalid value for region: europe-west.
```

In the case above, there's no "europe-west" region, but "europe-west1" for database regions.

The list of official regions can be found in the links below:
https://cloud.google.com/sql/docs/mysql/locations
https://cloud.google.com/sql/docs/postgres/locations

##### **Motivation for or Use Case**

Trying to deploy for the first time on gcloud gae

##### **Reproduce the error**

No entities were used here, other than the default: User

1- Have a project in Europe-West (or us-central)
2- Run jhipster gae
3- Try to create a SQL instance

##### **Suggest a Fix**

I already forked the project but I was wondering which would be the best way to fix it. For example:
- Should another input option be displayed so the user can choose the region for his/her database server?
or
- Should we append "1" if the region doesn't end with a number? eg.: europe-west or us-central (so far they are the only exceptions I found, but just had a quick look)

##### **JHipster Version(s)**

```
monu-admin@0.0.0 /Users/lrafaeli/dev/projects/monu-admin
└─┬ generator-jhipster-vuejs@1.8.2
  └── generator-jhipster@6.9.1 

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "one.monu.admin",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "6.9.1",
    "applicationType": "monolith",
    "baseName": "MonuAdmin",
    "packageName": "one.monu.admin",
    "packageFolder": "one/monu/admin",
    "serverPort": "8080",
    "authenticationType": "session",
    "cacheProvider": "no",
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
[...]

PR #12089, opened 2020-07-17
Update Google App Engine Locations

SudharakaP:
This updates app engine locations as per; https://cloud.google.com/appengine/pricing

Fixes https://github.com/jhipster/generator-jhipster/issues/12085

---


pascalgrimaud:
I wonder if there is an API to get this list, instead of hard coding it ?

SudharakaP:
@pascalgrimaud : We could use [`gcloud app regions list`](https://cloud.google.com/sdk/gcloud/reference/app/regions/list) command, but the only problem I see is that it will give the region tag (such as `asia-east2, us-east1` and so on). Currently we display to the user more friendlier region names (such as `us-central1 - Iowa, europe-west1 - Belgium` etc). I don't know an api method that could get us this actual country/city name which I why I didn't go that route. :smile:

Issue #11896, opened 2020-06-03
Cli changes for jhipster 7


mshima:
##### **Overview of the feature request**

I propose the following changes to cli for jhipster7
- [x] ~~Remove import-jdl command~~ import-jdl: do not force `--skip-sample-repository`
  `import-jdl` is now an alias for jdl command forcing `--skip-sample-repository`.
  We should promote `jdl --skip-sample-repository`. https://github.com/jhipster/generator-jhipster/pull/11770#discussion_r434523752, https://github.com/jhipster/generator-jhipster/pull/11770#issuecomment-634892631, https://github.com/jhipster/generator-jhipster/pull/11770#issuecomment-634880598
- [x] Remove kebab case options.
  Having to deal with `this.options['skip-install']` and `this.options.skipInstall` should be removed.
- [ ] Disable allowUnknownOption
  Makes easier to deal with cli options. https://github.com/jhipster/generator-jhipster/pull/11770#discussion_r430536146
  We must make sure every options is registered and load options from blueprints before disabling allowUnknownOption.
  

##### **Related issues or PR**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

deepu105:
Lets not remove the alias for 7, it doesn't cost anything to have the alias. We can consider removing it for v8

mraible:
@mshima It seems you re-opened this issue yesterday. Is there something else you need to fix?

mshima:
@mraible the I reverted the PR for jhipster code due to a bug and to it don’t interfere at the work there.
The issue is https://github.com/tj/commander.js/issues/1343.
I have a branch ready based on commander 7, so I am waiting it to be released.

mshima:
@pascalgrimaud commander 7 is planned for January, so we have 2 options for jhipster 7:
- bundle a commander 7 snapshot with jhipster (just 1 file).
- postpone `allowUnknownOption` to jhipster 8.

pascalgrimaud:
Honestly, I don't know. Maybe we can give a try to snapshot.
What's your advice on this ?

Le lun. 12 oct. 2020 à 22:57, Marcelo Shima <notifications@github.com> a
écrit :

mshima:
> Honestly, I don't know. Maybe we can give a try to snapshot.

I think we should do it.

pascalgrimaud:
I'm trusting you, so let's follow your advice

mraible:
Added bug bounty for your help @mshima.

[...]

PR #12094, opened 2020-07-18
Migrate options to camel case only.

mshima:
Removing kebab case support because:
- We are using camel for some options, kebab for others.
- Probably we are using different case for the same option in different generators.
- commander only generates camel case.
- Its confusing having both cases.
- Camel case is easier to use.

Related to https://github.com/jhipster/generator-jhipster/issues/11896

---


mshima:
@pascalgrimaud no, commander converts `--skip-install` into `skipInstall:true`, only changing the handling.
Cli options keeps intact:
https://github.com/jhipster/generator-jhipster/blob/1dd4804f585c46c104a106be4e1fad713e43554b/test/cli/cli.spec.js#L144-L162

pascalgrimaud:
it's totally perfect so !
thanks @mshima

mshima:
This is required due to complexity involved in having 2 keys for the same config.
@pascalgrimaud the problem is that is a breaking changes for blueprints when they use `this.options['skip-install']` instead of `this.options.skipInstall`.
Maybe we can consider to apply kebab case when composing with blueprints for jhipster 7, to reduce breaking changes, and remove for jhipster 8.

pascalgrimaud:
> This is required due to complexity involved in having 2 keys for the same config.

I totally agree with you. I just asked for documentation, to not forget it, as we'll have a lot of new stuff :)
About breaking change, we can, as it's a major version. We simply need to document it, then it will be ok.

Issue #11736, opened 2020-05-08
Change default database to Postgres


agaspardcilia:
##### **Overview of the feature request**

As mentioned in #10958 we could use Postgres as our default database.

##### **Motivation for or Use Case**

The point must have been discussed without my knowledge to be in the v7 roadmap, so I'm making this ticket to have a place to talk about it.

##### **Related issues or PR**

I've already made some shallow stuff on #11724.

-   [x] Checking this box is mandatory (this is just to show you read everything)

agaspardcilia:
As @pascalgrimaud pointed out, changing this would require to: 
+ change all default config in CI to use and test more PostgreSQL
+ do the same thing in jhipster-daily-builds
+ change the behavior in jhipster-online
+ do the change in all other blueprint backend (kotlin, nodejs, quarkus, etc.)

MathieuAA:
And update what the JDL parsing system outputs to take that into account.

mshima:
- Migrate to auto increment.
https://github.com/liquibase/liquibase/releases/tag/v4.0.0-beta1
https://github.com/liquibase/liquibase/releases/tag/v3.10.1

DanielFran:
@mshima You mean "postgresql >= 10: use identity columns for autoincrement"?

mshima:
Maybe we should use autoIncrement feature for postgresql too.
https://github.com/jhipster/generator-jhipster/blob/902160664a79cce8130e8077867f52ef4c394eac/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs#L27-L33

DanielFran:
@mshima Yes I agree, since I migrate to liquibase 3.10.1 (and PR is ready for 4.0.0), we can now switch to autoincrement.

Will you make the PR or do you want I take care of it?

mshima:
@DanielFran posted to keep note, pretty busy in the generator refactoring.
Feel free to work on it.

murdos:
Why would you want to use autoincrement for Postgres?
As far as I understand, the point in this article still stands: https://vladmihalcea.com/postgresql-serial-column-hibernate-identity/
The new identity feature in Postgres is just a more convenient and cleaner way to manage the underlying sequence used by Postgres, but this is still problematic with JPA/Hibernate.

DanielFran:
@murdos The only changes expected is to use autoincrement in liquibase side that define the "GENERATED AS IDENTITY" that still uses a sequence

[...]

Issue #9322, opened 2019-02-22
[Feature Request] PostgreSQL: Unique Sequence for each entity


amatosg:
##### **Overview of the feature request**

I see no reason for having a single sequence for **all** the entities. It would make more sense (to me, at least) if each entity has its own sequence.

##### **Motivation for or Use Case**

Entities' id would have a logic sequence. For me is *weird* to see one entity having id 1050, 1055 and then 1065, while others could have 1056, 1057.

##### **Related issues or PR**

I've searched both open and closed issues and haven't found anything.

-   [x] Checking this box is mandatory (this is just to show you read everything)

jdubois:
I believe that what we do is the default in Hibernate, so that would be against [policy 1](https://www.jhipster.tech/policies/), could you check?

On a personal side, I'd rather have just one sequence, I find this simpler.

cbornet:
We use the sequence generator with Hilo algorithm which allows performant batch. See https://vladmihalcea.com/hibernate-identity-sequence-and-table-sequence-generator/

jdubois:
Yes I think we have the recommended approach from Vlad, so I'm closing this (of course you can still prove me wrong and I'll re-open)

amatosg:
What about using `id BIGINT GENERATED ALWAYS AS IDENTITY`? That way each table would have its own sequence generator (just like in MySQL/MariaDB). With that, things get simplify because:

1. There is no need to create sequences
2. The IDs for each entity will be correlative
3. No additional queries are needed as there are no sequences 

My 3rd point would make it even more performant than the Hilo algorithm as there are no queries involved for generating the id.

The only drawback I see is that that feature was introduced in PostgreSQL 10 in October 2017, which it might impact on users having 9.x.

What do you guys think?

gmarziou:
> 3\. No additional queries are needed as there are no sequences

Have you checked this with datasource-proxy or p6spy?
I don't understand how Hibernate would know the id without re-loading the entity.

Using sequences in Hibernate is better because it can provision a batch of new ids in one statement and then use them for many inserts without having to ask sequence fir each insert.

That said, I'd prefer too to have a sequence per table.

murdos:
> 
> That said, I'd prefer too to have a sequence per table.

I prefer that too.
Would it be ok if it was available as an option?

amatosg:
[...]

PR #12098, opened 2020-07-18
Use autoincrement property in liquibase for PostgreSQL

DanielFran:
Related with https://github.com/jhipster/generator-jhipster/issues/11736:
- Use liquibase property autoincrement for PostgreSQL that define 'GENERATED BY DEFAULT AS IDENTITY' (see https://github.com/liquibase/liquibase/pull/874 and https://github.com/liquibase/liquibase/pull/941)
- Define 'POSTGRES_10' by default for GAE
- Define 'aurora-postgresql10.4' by default for AWS

---

### DIFF
--- a/generators/gae/index.js
+++ b/generators/gae/index.js
@@ -649,7 +649,7 @@ module.exports = class extends BaseGenerator {
                 const name = this.gcpCloudSqlInstanceName;
                 // for mysql keep default options, set specific option for pg
                 const dbVersionFlag =
-                    this.prodDatabaseType === 'postgresql' ? ' --database-version="POSTGRES_9_6" --tier="db-g1-small"' : '';
+                    this.prodDatabaseType === 'postgresql' ? ' --database-version="POSTGRES_10" --tier="db-g1-small"' : '';
                 let gaeCloudSqlLocation = this.gaeLocation;
                 if (gaeCloudSqlLocation === 'us-central') {
                     gaeCloudSqlLocation = 'us-central1';
--- a/generators/aws-containers/index.js
+++ b/generators/aws-containers/index.js
@@ -361,9 +361,9 @@ module.exports = class extends BaseGenerator {
                     app.dbType = appConfig.prodDatabaseType;
 
                     app.auroraEngine = appConfig.prodDatabaseType === postgresqlType ? 'aurora-postgresql' : 'aurora-mysql';
-                    app.auroraFamily = appConfig.prodDatabaseType === postgresqlType ? 'aurora-postgresql9.6' : 'aurora-mysql5.7';
+                    app.auroraFamily = appConfig.prodDatabaseType === postgresqlType ? 'aurora-postgresql10.4' : 'aurora-mysql5.7';
                     app.auroraClusterParameterGroupName =
-                        appConfig.prodDatabaseType === postgresqlType ? 'default.aurora-postgresql9.6' : 'default.aurora-mysql5.7';
+                        appConfig.prodDatabaseType === postgresqlType ? 'default.aurora-postgresql10.4' : 'default.aurora-mysql5.7';
                     app.auroraDbParam =
                         appConfig.prodDatabaseType === postgresqlType ? 'check_function_bodies: 0' : 'sql_mode: IGNORE_SPACE';
                 });
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -175,7 +175,7 @@ public class <%= asEntity(entityClass) %> implements Serializable {
 <%_ if (!embedded) { _%>
     @Id
     <%_ if (databaseType === 'sql' && isUsingMapsId === false) { _%>
-        <%_ if (prodDatabaseType === 'mysql' || prodDatabaseType === 'mariadb') { _%>
+        <%_ if (['mysql', 'mariadb', 'postgresql'].includes(prodDatabaseType)) { _%>
     @GeneratedValue(strategy = GenerationType.IDENTITY)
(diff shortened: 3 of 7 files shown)


DanielFran:
@murdos Maybe I am wrong but my understanding is that in Postgresql 10 on hibernate side it uses the sequence: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/identity/PostgreSQL10IdentityColumnSupport.java

mshima:
@DanielFran we need to change this to identity:
https://github.com/jhipster/generator-jhipster/blob/902160664a79cce8130e8077867f52ef4c394eac/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs#L178-L184

@murdos this problem applies to serial and big serial.
IFAIK with this postgres 10 feature the type isn't serial.

Anyway marking as `on-hold` until we make sure this.

DanielFran:
@mshima You are right, I forget to change it.

mshima:
> @murdos Maybe I am wrong but my understanding is that in Postgresql 10 on hibernate side it uses the sequence: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/identity/PostgreSQL10IdentityColumnSupport.java

@DanielFran that's what I was looking for.

I think it's safe.

generators/gae/index.js
```diff
@@ -649,7 +649,7 @@ module.exports = class extends BaseGenerator {
                 const name = this.gcpCloudSqlInstanceName;
                 // for mysql keep default options, set specific option for pg
                 const dbVersionFlag =
-                    this.prodDatabaseType === 'postgresql' ? ' --database-version="POSTGRES_9_6" --tier="db-g1-small"' : '';
+                    this.prodDatabaseType === 'postgresql' ? ' --database-version="POSTGRES_10" --tier="db-g1-small"' : '';
```
SudharakaP, line 652: Are we adding support for PostgreSQL 10 specifically? Not anything higher than that? Just wondering since otherwise if we do support the current version 12, we can drop this `--database-version` flag completely.

DanielFran: @SudharakaP I was just maintaining the default value, the version needs now to be >= 10

murdos:
> @murdos Maybe I am wrong but my understanding is that in Postgresql 10 on hibernate side it uses the sequence: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/identity/PostgreSQL10IdentityColumnSupport.java

Yes, and it was already doing it in Postgresql < 10, nothing changed: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/identity/PostgreSQL81IdentityColumnSupport.java

You seems to be confused by the fact that PosgreSQL internally use its own sequence.

Please take the time to read https://vladmihalcea.com/postgresql-serial-column-hibernate-identity/ : the issue is that Hibernate ideally needs to know the entity identifier before hand, and relying on the internal sequence prevents Hibernate from doing batch INSERT statements, which has an impact on performance.

> 
> @murdos this problem applies to serial and big serial.
> IFAIK with this postgres 10 feature the type isn't serial.

I don't see any fundamental changes related to the use of an internal sequence and our issue with identity column when I read https://www.2ndquadrant.com/en/blog/postgresql-10-identity-columns/
@vladmihalcea's article was written when Postgresql was only supporting serial, so of course he doesn't mention the case of identity column.

murdos:
Another interesting article: https://vladmihalcea.com/how-to-replace-the-table-identifier-generator-with-either-sequence-or-identity-in-a-portable-way/

> Database sequences are the best identifier generator choice when using JPA and Hibernate. This is because, unlike IDENTITY generator, you can take advantage of JDBC batch updates. Therefore, you should always use a database sequence if the underlying database supports this feature.

That confirms that we should keep the SEQUENCE strategy for Postgresql, and the we should also use it for MariaDB since [sequence are now supported since version 10.3](https://mariadb.com/resources/blog/sequences-support-in-mariadb-server-10-3/)

mshima:
In my opinion:
- https://vladmihalcea.com/postgresql-serial-column-hibernate-identity/
Does not apply for Postgres 10.
- https://vladmihalcea.com/how-to-replace-the-table-identifier-generator-with-either-sequence-or-identity-in-a-portable-way/
Does apply for batch update, but I don't see this as a blocking issue, since it won't make difference for the major for the use cases. (A recent change removed a lib that caused a performance drop of ~3% https://github.com/jhipster/generator-jhipster/pull/10283#issuecomment-616491178 and the developer should take care of improving by choosing another lib).
Doing one call to sequence generator for each insertion, in jhipster applications (web/microservice) will have a negligible performance drop IMO.
- I would like to have one sequence per table, but that required creating a sequence at liquibase changelog for each table.
Now it's just migrate to identity.

A previous discussion about this theme: https://github.com/jhipster/generator-jhipster/issues/9322

mshima:
Missing here too:
https://github.com/jhipster/generator-jhipster/blob/86a0ecdeca071ba4098d3a115f548841a821ab9e/generators/server/templates/src/test/java/package/repository/timezone/DateTimeWrapper.java.ejs#L33-L35

vladmihalcea:
@mshima The way Hibernate handles IDENTITY is the same no matter what PostgreSQL version you are using.

mshima:
Ok then, we should close this or change mariadb to sequence as @murdos said.
@vladmihalcea is there any way to benefit from identity with hibernate?

vladmihalcea:
If you are not concerned about batch inserts, then it's all fine. You could do the batch inserts using some other framework.

DanielFran:
Thanks @vladmihalcea for the feedback.

Issue #12109, opened 2020-07-19
Prettier Java by default


pascalgrimaud:
##### **Overview of the feature request**

In v6.10.1, we have the flag `--prettier-java`
For v7, we should put `prettier-java` by default, in all generated projects.

We should discuss about the flag `--no-prettier-java` but I'm not in favor for keeping this, as we don't have it for our frontend.

cc @clementdessoude @Shaolans @murdos 

##### **Motivation for or Use Case**

I use prettier-java for my customers, since months, and it works so well.
It should absolutely be by default in all generated projects.

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/issues?q=prettier

-   [x] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
@mshima : as it's related to `prettier`, I can mention it here. But I'm not sure it's related to prettier-java, maybe related to yeoman phase. I'm a little lost for this minor issue.
When I generate a Vue application, at the end of generation, there are some files which are not commited by git:

```
Application successfully committed to Git from /home/pgrimaud/tmp/12-vue.

If you find JHipster useful consider sponsoring the project https://www.jhipster.tech/sponsors/

Server application generated successfully.

Run your Spring Boot application:
./mvnw

Client application generated successfully.

Start your Webpack development server with:
 npm start


> jhipster@0.0.0 cleanup /home/pgrimaud/tmp/12-vue
> rimraf target/classes/static/

INFO! Congratulations, JHipster execution is complete!

17:46:28 in ~/tmp/12-vue on  master [!+] took 1m 30s 
➜ git status   
Sur la branche master
Modifications qui seront validées :
  (utilisez "git restore --staged <fichier>..." pour désindexer)
	modifié :         .lintstagedrc.js
	modifié :         .postcssrc.js
	modifié :         src/test/javascript/jest.conf.js
	modifié :         webpack/loader.conf.js
	modifié :         webpack/utils.js
	modifié :         webpack/vue.utils.js
	modifié :         webpack/webpack.common.js
	modifié :         webpack/webpack.dev.js
	modifié :         webpack/webpack.prod.js

Modifications qui ne seront pas validées :
  (utilisez "git add <fichier>..." pour mettre à jour ce qui sera validé)
  (utilisez "git restore <fichier>..." pour annuler les modifications dans le répertoire de travail)
	modifié :         .lintstagedrc.js
	modifié :         .postcssrc.js
[...]

PR #12124, opened 2020-07-20
Remove Prettier java option in favor of having it baseline

avdev4j:
fix https://github.com/jhipster/generator-jhipster/issues/12109

---

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -197,12 +197,6 @@ module.exports = class extends BaseBlueprintGenerator {
             type: String,
         });
 
-        // This adds support for a `--prettier-java` flag
-        this.option('prettier-java', {
-            desc: `Launch prettier-java pre-formatting at generation (default: ${appDefaultConfig.prettierJava})`,
-            type: Boolean,
-        });
-
         // Just constructing help, stop here
         if (this.options.help) {
             return;
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -127,7 +127,7 @@
     "optimize-css-assets-webpack-plugin": "5.0.3",
     "postcss-loader": "3.0.0",
     "prettier": "<%= PRETTIER_VERSION %>",
-    <%_ if (!skipServer && prettierJava) { _%>
+    <%_ if (!skipServer) { _%>
     "prettier-plugin-java": "<%= PRETTIER_JAVA_VERSION %>",
     <%_ } _%>
     <%_ if (protractorTests) { _%>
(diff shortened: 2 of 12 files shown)


avdev4j:
I did not notice that @mshima already did the job. However I found a bug in search templates that can be fix with my last commit. 

I know we did not take the same way but I have no problem to close my PR in favour or yours @mshima  (and open a specific open for the bug).

Also what do you think about print the error object instead of error.message in prettierTransform ? In my case the error was undefined and I struggled a bit to find the reason.

mshima:
> I did not notice that @mshima already did the job. However I found a bug in search templates that can be fix with my last commit.

No problem.

> I know we did not take the same way but I have no problem to close my PR in favor or yours @mshima (and open a specific open for the bug).
> 
> Also what do you think about print the error object instead of error.message in prettierTransform ? In my case the error was undefined and I struggled a bit to find the reason.

Sure.

@avdev4j the problem is that I think prettier options is not been honoured.
We don't have useful tests for prettier java, the only prettier java test does not do the job anymore.
Can you add a blueprint test that overrides `.prettierrc` with custom java options and add a not prettified java file to make sure it working?

I had to disable prettier java for some needles because the prettified code does not match the expected needles.
I prefer to merge my branch before removing prettierJava switch completely, I don't want to deal with that test.

avdev4j:
hi @mshima 
I've just added a new tests which override App, Common and Server generators through a blueprint and test if .prettierrc is present and if it's applied as expected. WDYT ?

This is what we can do:
- You finish you work on your branch
- I open a new PR with the bug fix and the tests
- Close this one and postpone removing --prettier-java totally

I suppose I don't need to wait for your branch to be merge to open and merge mine since I add prettier-java option in my test (and you would remove it after)

mshima:
> hi @mshima
> I've just added a new tests which override App, Common and Server generators through a blueprint and test if .prettierrc is present and if it's applied as expected. WDYT ?

I found the bug I was trying to find with the test I asked. It's not easily fixed =\\.
```
echo > .prettierrc
jhipster --defaults --prettier-java
```
The source will be generated and there will be diff (git pre-hook commit applied prettier correctly).

At the normal test I did a different approach, I created a side-by-side blueprint that overrides `.prettierrc` with `tabWidth: 10` and tested the generated source with that width.

> This is what we can do:
> 
> * You finish you work on your branch
> * I open a new PR with the bug fix and the tests
> * Close this one and postpone removing --prettier-java totally

Looks good, the test can be dropped.

> I suppose I don't need to wait for your branch to be merge to open and merge mine since I add prettier-java option in my test (and you would remove it after).
> 
> Indeed, we should mark `--prettier-java` option as dreprecated as soon as possible and remove it 7+ releases (when we will be trustful).

I think we can drop prettier java flag.

> Also what do you think about print the error object instead of error.message in prettierTransform ? In my case the error was undefined and I struggled a bit to find the reason.

Just remembered that the correct approach is to call the callback with the error instead of throwing.
Since it's running at a queued function, the process will die otherwise.

avdev4j:
So, we should merge https://github.com/jhipster/generator-jhipster/pull/12114 and I'll do a rebase and remove the blueprint test.

avdev4j:
Since you PR has been merged @mshima I trust you to review mine if all go green :D

PR #12126, opened 2020-07-20
Add JHCC docker compose files

avdev4j:
---

### DIFF
--- a/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
+++ b/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
@@ -42,6 +42,11 @@ version: '<%= DOCKER_COMPOSE_FORMAT_VERSION %>'
 services:
   jhipster-control-center:
     image: '<%= DOCKER_JHIPSTER_CONTROL_CENTER %>'
+    command:
+      - /bin/sh
+      - -c
+      # Patch /etc/hosts to support resolving host.docker.internal to the internal IP address used by the host in all OSes
+      - echo "`ip route | grep default | cut -d ' ' -f3` host.docker.internal" | tee -a /etc/hosts > /dev/null && java -jar /jhipster-control-center.jar
     environment:
       - _JAVA_OPTIONS=-Xmx512m -Xms256m
       # TODO change 'swagger' to 'api-docs' when JHCC will be migrated to 7.x JHipster version. https://github.com/jhipster/jhipster/pull/764
--- a/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
+++ b/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
@@ -1,0 +1,38 @@
+<%#
+Copyright 2013-2020 the original author or authors from the JHipster project.
+
+This file is part of the JHipster project, see https://www.jhipster.tech/
+for more information.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-%>
+version: '<%= DOCKER_COMPOSE_FORMAT_VERSION %>'
+services:
+  jhipster-control-center:
+    image: '<%= DOCKER_JHIPSTER_CONTROL_CENTER %>'
+    environment:
+      - _JAVA_OPTIONS=-Xmx512m -Xms256m
+      - SPRING_PROFILES_ACTIVE=prod,swagger<% if (authenticationType === 'oauth2') { %>,oauth2<% } %>
+      <%_ if (authenticationType === 'jwt') { _%>
+      - SPRING_SECURITY_USER_PASSWORD=admin
+      <%_ } _%>
+      - MANAGEMENT_METRICS_EXPORT_PROMETHEUS_ENABLED=true
+      - JHIPSTER_SLEEP=30 # gives time for other services to boot before the application
+      <%_ if (authenticationType === 'oauth2') { _%>
+      # For keycloak to work, you need to add '127.0.0.1 keycloak' to your hosts file
+      - SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_OIDC_ISSUER_URI=http://keycloak:9080/auth/realms/jhipster
+      - SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_OIDC_CLIENT_ID=jhipster-control-center
+      - SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_OIDC_CLIENT_SECRET=jhipster-control-center
+      <%_ } _%>
+    ports:
+      - 7419:7419
--- a/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
+++ b/generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
@@ -17,6 +17,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -%>
 ## How to use JHCC docker compose
+# To allow JHCC to reach JHipster application from a docker container note that we set the host as host.docker.internal
+# To reach the application from a browser, you need to add '127.0.0.1 host.docker.internal' to your hosts file.
 ### Discovery mode
 # JHCC support 3 kinds of discovery mode: Consul, Eureka and static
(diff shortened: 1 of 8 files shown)


generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+## How to use JHCC docker compose
+### Discovery mode
+# JHCC support 3 kinds of discovery mode: Consul, Eureka and static
+# In order to use one, please set SPRING_PROFILES_ACTIVE to one (and only one) of this values: consul,eureka,static
+### Discovery properties
+# According to the discovery mode choose as Spring profile, you have to set the right properties
+# please note that current properties are set to run JHCC with default values, personalize them if needed
+# and remove those from other modes. You can only have one mode active.
+#### Eureka
+#       - EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=http://admin:admin@host.docker.internal:8761/eureka/
```
pascalgrimaud, line 28: we can add this by default, when discoverytype is eureka, then, we can suppose the registry is started with docker, so the host should be jhipster-registry
we can anyway keep this line as commented, in case the registry is started direct from host, instead of docker

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+#### Consul
```
pascalgrimaud, line 29: same comment here for consul

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+#       - SPRING_CLOUD_CONSUL_HOST=host.docker.internal
+#       - SPRING_CLOUD_CONSUL_PORT=8500
+#### Static
+# Add instances to "MyApp"
+#       - SPRING_CLOUD_DISCOVERY_CLIENT_SIMPLE_INSTANCES_MYAPP_0_URI=http://host.docker.internal:8081
+#       - SPRING_CLOUD_DISCOVERY_CLIENT_SIMPLE_INSTANCES_MYAPP_1_URI=http://host.docker.internal:8082
+# Or add a new application named MyNewApp
+#       - SPRING_CLOUD_DISCOVERY_CLIENT_SIMPLE_INSTANCES_MYNEWAPP_0_URI=http://host.docker.internal:8080
```
pascalgrimaud, line 37: we can add one, by default for monolith, with the server port from config. what do you think ?

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+version: '<%= DOCKER_COMPOSE_FORMAT_VERSION %>'
+services:
+  jhipster-control-center:
+    image: '<%= DOCKER_JHIPSTER_CONTROL_CENTER %>'
+    environment:
+      - _JAVA_OPTIONS=-Xmx512m -Xms256m
+      - SPRING_PROFILES_ACTIVE=prod,swagger,static<% if (authenticationType === 'oauth2') { %>,oauth2<% } %>
```
pascalgrimaud, line 44: in v7 and in current master, swagger is renamed
I know JHCC was generated and updated to last v6, but we need to be careful about this

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+      <%_ if (authenticationType === 'jwt') { _%>
+      - SPRING_SECURITY_USER_PASSWORD=admin
+      <%_ } _%>
+      - MANAGEMENT_METRICS_EXPORT_PROMETHEUS_ENABLED=true
```
pascalgrimaud, line 48: is it needed ?

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+      - JHIPSTER_SLEEP=30 # gives time for other services to boot before the application
+      <%_ if (authenticationType === 'oauth2') { _%>
+      # For keycloak to work, you need to add '127.0.0.1 keycloak' to your hosts file
+      - SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_OIDC_ISSUER_URI=http://keycloak:9080/auth/realms/jhipster
+      - SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_OIDC_CLIENT_ID=jhipster-control-center
```
pascalgrimaud, line 53: need to add these info in keycloak realm

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,58 @@
...
+#       - SPRING_CLOUD_DISCOVERY_CLIENT_SIMPLE_INSTANCES_MYNEWAPP_0_URI=http://host.docker.internal:8080
+version: '<%= DOCKER_COMPOSE_FORMAT_VERSION %>'
+services:
+  jhipster-control-center:
+    image: '<%= DOCKER_JHIPSTER_CONTROL_CENTER %>'
+    environment:
+      - _JAVA_OPTIONS=-Xmx512m -Xms256m
+      - SPRING_PROFILES_ACTIVE=prod,swagger,static<% if (authenticationType === 'oauth2') { %>,oauth2<% } %>
+      <%_ if (authenticationType === 'jwt') { _%>
+      - SPRING_SECURITY_USER_PASSWORD=admin
```
pascalgrimaud, line 46: do we need an additional common jhipster security secret ?

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
avdev4j: no

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
avdev4j: you're right, for the moment we should keep swagger and change it to the new profile when we will upgrade JHCC to V7

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
avdev4j: done

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
avdev4j: I wonder if we should keep host.docker.internal as it should work for every cases (to be confirm)

PierreBesson: I think it's best to move this comment to the docs and generate only the relevant option (static/eureka/consul). This means we need to write a page on the JHipster Control Center !

avdev4j: For sure, we need a dedicated page for JHCC. Now the yaml is adjusted according to the project configuration we can move out this or add only what we need.

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
```diff
@@ -0,0 +1,67 @@
...
+      <%_ if (discoveryProfile === 'eureka') { _%>
+      - EUREKA_CLIENT_SERVICE_URL_DEFAULTZONE=http://admin:admin@host.docker.internal:8761/eureka/
+      <%_ } else if (discoveryProfile === 'consul') { _%>
+      - SPRING_CLOUD_CONSUL_HOST=host.docker.internal
+      - SPRING_CLOUD_CONSUL_PORT=8500
+      <%_ } else { _%>
+      - SPRING_CLOUD_DISCOVERY_CLIENT_SIMPLE_INSTANCES_<%= baseName.toUpperCase() %>_0_URI=http://host.docker.internal:<%= serverPort %>
+      <%_ } _%>
+    ports:
+      - 7419:7419
```
PierreBesson, line 67: @avdev4j To make the `host.docker.internal` host resolution work on Linux, here is the magic line you need to add to the docker-compose.

```
    command:
      - /bin/sh
      - -c
      # Patch /etc/hosts to support resolving host.docker.internal to the internal IP address used by the host in all OSes
      - echo "`ip route | grep default | cut -d ' ' -f3` host.docker.internal" | tee -a /etc/hosts > /dev/null && java -jar /jhipster-control-center.jar
```

I have installed the iproute command in the docker image in this jhcc PR which is already merged and deployed to docker hub: https://github.com/jhipster/jhipster-control-center/pull/85

generators/server/templates/src/main/docker/jhipster-control-center.yml.ejs
avdev4j: https://github.com/jhipster/generator-jhipster/issues/12163

PR #12128, opened 2020-07-20
prevent leakage of technical details trough problem message

### LINKED ISSUE #12051: Exception Handling: Leakage of Implementation Details

Some exceptions trigger a leakage of implementation details to the end user. The example shown here relates to Jackson's exception details, but maybe there are other examples that I am not aware of.

We recently underwent security testing -- one of the issues that was caught is that certain exceptions leak inside information. An example is shown below.

##### **Reproduce the error**

Use a malformed JSON request body.

Example Request

```
GET /auth/login

35 {"username":"asd","password":"asd","rememberMe":true} 1
Z
Q
```

Example Response

```json
{
    "type": "https://www.jhipster.tech/problem/problem-with-message",
    "title": "Bad Request",
    "status": 400,
    "detail": "JSON parse error: Cannot deserialize instance of 
       `java.util.LinkedHashMap<java.lang.Object,java.lang.Object>` out of VALUE_NUMBER_INT token; nested exception is com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.util.LinkedHashMap<java.lang.Object,java.lang.Object>` 
       out of VALUE_NUMBER_INT token\n at [Source: (PushbackInputStream); line: 1, column: 1]", 
    "path": "/auth/login",
    "message": "error.http.400"
}
```

As can be seen above, the error message discloses internal details that could help an attacker understand the system's vulnerabilities better.

In this case, "The input is not valid JSON: expecting a JSON object but found a number instead." is a better message because it better communicates the error without disclosing any implementation details. However, if it is too complicated to implement such detailed error messages, we should simply say that the JSON is malformed (this is what I have done, as I am not familiar enough with Jackson to go any further).

##### **Suggest a Fix**

It is easy to fix this particular exception (I have done so on my end, using if statements in `ExceptionTranslator#process(...)`).

However, there may be other examples that I am not aware. If you have any examples in mind, please comment below or create a new issue!

##### **JHipster Version(s)**

Reproduced on both 6.7.1 Gateway and 6.9.1 micro-service. I haven't seen a fix in recent versions' change logs; correct me if I am wrong.

  <details>
  <summary>.yo-rc</summary>
  <pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.***.gateway",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "6.7.1",
    "applicationType": "gateway",
    "baseName": "gateway",
    "packageName": "com.nxtsens.backend.gateway",
    "packageFolder": "com/nxtsens/backend/gateway",
    "serverPort": "8080",
    "authenticationType": "uaa",
    "uaaBaseName": "uaa",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": true,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": "kafka",
    "buildTool": "gradle",
    "clientFramework": "react",
    "useSass": true,
    "testFrameworks": ["cucumber"],
    "jhiPrefix": "jhi",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "languages": ["en", "fr"],
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "clientTheme": "none",
    "blueprints": [],
    "embeddableLaunchScript": false,
    "creationTimestamp": 1579029215188
  }
}
  </pre>
  </details>

atomfrede:
This prevents leaking of technical details via the problem message which is by default filled with the message of the exception. Most often the message contain/reveal technical details which should not be leak via an api call to the caller.

closes #12051

**This targets v6.x branch on purpose. I think this should be backported as it impacts security.**

---

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/errors/ExceptionTranslator.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/errors/ExceptionTranslator.java.ejs
@@ -18,13 +18,17 @@
 -%>
 package <%= packageName %>.web.rest.errors;
 
+import io.github.jhipster.config.JHipsterConstants;
 import io.github.jhipster.web.util.HeaderUtil;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Value;
 <%_ if (databaseType !== 'no' && databaseType !== 'cassandra') { _%>
 import org.springframework.dao.ConcurrencyFailureException;
 <%_ } _%>
+import org.springframework.dao.DataAccessException;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.HttpMessageConversionException;
 <%_ if (reactive && databaseType === 'sql') { _%>
 import org.springframework.stereotype.Component;
 <%_ } _%>
@@ -41,10 +45,12 @@ import org.springframework.web.server.ServerWebExchange;
 <%_ if (!reactive) { _%>
 import org.springframework.web.context.request.NativeWebRequest;
 <%_ } _%>
+import org.springframework.core.env.Environment;
 import org.zalando.problem.DefaultProblem;
 import org.zalando.problem.Problem;
 import org.zalando.problem.ProblemBuilder;
 import org.zalando.problem.Status;
+import org.zalando.problem.StatusType;
 import org.zalando.problem.spring.web<% if (reactive) { %>flux<% } %>.advice.ProblemHandling;
 import org.zalando.problem.spring.web<% if (reactive) { %>flux<% } %>.advice.security.SecurityAdviceTrait;
 import org.zalando.problem.violations.ConstraintViolationProblem;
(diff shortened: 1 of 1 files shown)


AlexandreCassagne:
This is something I have manually added to my stack. I can confirm it does the job!

atomfrede:
@pascalgrimaud Any idea why some builds are stuck forever? I have seen a similar effect here https://github.com/jhipster/generator-jhipster-micronaut/pull/125

pascalgrimaud:
@avdev4j reported it to me, the last days too.
I think the problem comes from GitHub: we have too much builds in our pipelines, so we must reduce the number of builds. The only way to solve it is to restart the build.

avdev4j:
I confirm, the only solution I found is to restart workflows by workflows without trying to restart them all at the same time.

atomfrede:
Good for this one, the gradle wrapper validation will be fine :)

atomfrede:
Here is the PR for the documentation update https://github.com/jhipster/jhipster.github.io/pull/990

PR #12164, opened 2020-07-27
Update spring-boot to 2.2.9 and other dependencies

DanielFran:
Please make sure the below checklist is followed for Pull Requests.

-   [x] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [x] Tests are added where necessary
-   [x] Documentation is added/updated where necessary
-   [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #12183, opened 2020-07-31
Filter builds according to modified generators

avdev4j:
We have some issues with Github action workflows at the moment, many of them are stuck in a queue. This is a Github bug, but I suggest to reduce the number of ran builds according to `paths` or `paths-ignore` properties to improve our CI.

WDYT @jhipster/developers ?

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -22,9 +22,17 @@ on:
     push:
         branches-ignore:
             - 'dependabot/**'
+        paths:
+            - '!generators/client/**react.js'
+            - '!generators/client/**vue.js'
+            - 'generators/client/templates/angular/**'
     pull_request:
         branches:
             - '*'
+        paths:
+            - '!generators/client/**react.js'
+            - '!generators/client/**vue.js'
+            - 'generators/client/templates/angular/**'
 env:
     JHI_RUN_APP: 1
     JHI_JDK: 11
--- a/.github/workflows/default-app.yml
+++ b/.github/workflows/default-app.yml
@@ -22,9 +22,13 @@ on:
     push:
         branches-ignore:
             - 'dependabot/**'
+        paths-ignore:
+            - 'generators/client/**'
     pull_request:
         branches:
             - '*'
+        paths-ignore:
+            - 'generators/client/**'
 env:
     JHI_RUN_APP: 1
     JHI_JDK: 11
--- a/.github/workflows/default-app.yml
+++ b/.github/workflows/default-app.yml
@@ -1,0 +1,144 @@
+#
+# Copyright 2013-2020 the original author or authors from the JHipster project.
+#
+# This file is part of the JHipster project, see https://www.jhipster.tech/
+# for more information.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


pascalgrimaud:
I like a lot, but I see 1 improvement here:
- add `default-app.yml` which will be triggered all the time -> maybe 1 angular, 1 react, 1 vue

avdev4j:
I've added the default workflow. Tell me wdyt ;)

avdev4j:
I've added a paths-ignore for default-app if we modify a client file. We need to discuss about that because it can be an issue:

Without paths-ignore:
- default-angular/react/vue will be launched twice if we do a modification in a specific client (e.g Angular).

With paths-ignore:
- default-angular/react will not be launched at all if we do a modification in a specific client (e.g Vue).

mshima:
I don't like this change.
With this change, a PR like https://github.com/jhipster/generator-jhipster/pull/12164 won't trigger any build, but IMO it should trigger every build, webflux included which is not been triggered currently.

mshima:
- Each e2e test tests a different combination of backend features.
- Push events should not be filtered IMO, there are not enough PR merges to worth it. A missed broken PR can be difficult to track.

avdev4j:
> I don't like this change.
> With this change, a PR like #12164 won't trigger any build, but IMO it should trigger every build, webflux included which is not been triggered currently.

This is not really true. The default-app and the generator will be triggered, also all azure-pipelines ones, but I'm agree with that in certain cases we probably need to run all scenario. We should going further and find a solution.

Webflux workflow is now an option and it's not related to this PR.

.github/workflows/angular.yml
```diff
@@ -22,9 +22,17 @@ on:
     push:
         branches-ignore:
             - 'dependabot/**'
+        paths:
+            - '!generators/client/**react.js'
+            - '!generators/client/**vue.js'
+            - 'generators/client/templates/angular/**'
```
mshima, line 28: Drop filtering from push event.
```suggestion
```

.github/workflows/angular.yml
```diff
@@ -22,9 +22,17 @@ on:
...
     pull_request:
         branches:
             - '*'
+        paths:
+            - '!generators/client/**react.js'
+            - '!generators/client/**vue.js'
+            - 'generators/client/templates/angular/**'
```
mshima, line 35:
```suggestion
        paths:
            - 'utils/**'
            - 'test-integration/**'
            - 'generators/*.js'
            - '!generators/docker-*.js'
            - '!generators/*-docker.js'
            - '!generators/kubernetes-*.js'
            - 'generators/app/*.js'
            - 'generators/client/**/*.js'
            - '!generators/client/**/*-react.js'
            - '!generators/client/**/*-vue.js'
            - 'generators/client/templates/angular/**'
            - 'generators/server/**/*.js'
            - 'generators/server/templates/**'
            - 'generators/languages/**/*.js'
            - 'generators/languages/templates/**'
            - 'generators/common/**/*.js'
            - 'generators/common/templates/**'
            - 'generators/entity/**/*.js'
            - 'generators/entity-client/*.js'
            - 'generators/entity-client/templates/angular/**'
            - 'generators/entity-server/*.js'
            - 'generators/entity-server/templates/**
            - 'generators/entity-i18n/*.js'
            - 'generators/entity-i18n/templates/**
```

avdev4j:
I think we can close this PR @mshima, can't we?

mshima:
> I think we can close this PR @mshima, can't we?

Yep, If we became constrained in runners again we can create a new one, but with our current structure it's really complicated to filter by paths without loosing coverage.

avdev4j:
great!

PR #12189, opened 2020-08-03
[vue] Entity page - default order should be asc

qmonmert:
Fix issue https://github.com/jhipster/jhipster-vuejs/issues/596

---

### DIFF
--- a/generators/entity-client/templates/vue/src/main/webapp/app/entities/entity.component.ts.ejs
+++ b/generators/entity-client/templates/vue/src/main/webapp/app/entities/entity.component.ts.ejs
@@ -188,7 +188,7 @@ export default class <%= entityAngularName %> extends <% if (fieldsContainBlob |
   <%_ } _%>
   <%_ if (pagination !== 'no') { _%>
   public sort() : Array<any> {
-    const result = [this.propOrder + ',' + (this.reverse ? 'asc' : 'desc')];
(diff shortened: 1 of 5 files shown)


generators/entity-client/templates/vue/src/main/webapp/app/entities/entity.component.ts.ejs
```diff
@@ -24,7 +24,7 @@ export default class <%= entityAngularName %> extends <% if (fieldsContainBlob |
   public page = 1;
   public previousPage = 1;
   public propOrder = 'id';
-  public reverse = <%_ if (pagination === 'infinite-scroll') { _%>true<%_ } else { _%>false<%_ } _%>;
+  public reverse = false;
```
avdev4j, line 27: why does this condition is removed?

qmonmert: because by default entities must be sorted by id ASC and not reverted

avdev4j: I'm still wonder why do we set it to true when choosing the infinite-scroll

qmonmert: it was a mistake because with true, entities are bad sorted

Issue #12178, opened 2020-07-30
Implement incremental liquibase.


mshima:
##### **Overview of the feature request**

Implement linear liquibase changelog.
- Every time some entity is added or removed, create a new changelog.
- Don't touch old changelogs.
- Don't recreate master.xml.
- Append new entries to master.xml.
- New master.xml's entries must be linear.
- Constraints depends on the table existence.

Tasks:
- [x] 1. stop passing fork function to import-jdl. (https://github.com/jhipster/generator-jhipster/pull/12137)
- [x] 2. add a option to jdl to stop saving config files. (https://github.com/jhipster/jhipster-core/pull/442, https://github.com/jhipster/jhipster-core/pull/450)
- [x] 3. change jdl to return a complete application structure. (https://github.com/jhipster/generator-jhipster/pull/12191)
- [x] 4. change import-jdl to save the config to keep fork compatibility. (https://github.com/jhipster/generator-jhipster/pull/12193)
- [ ] 5. stop calling environment using fork for 1 application and pass the application structure as option instead of saving.
- [x] 6. make entities generator more friendly to composing. (https://github.com/jhipster/generator-jhipster/pull/12176)
- [ ] 7. implement incremental-changelog and inclemental-changelog-liquibase.
- [x] 8. Allow app/entities configs to passed at options instead of saving. (https://github.com/jhipster/generator-jhipster/pull/12197)
- [x] 9. Don't touch master.xml and initial.xml.

Support for multiples applications (not planned):
- [ ] ~stop calling environment using fork for multiples applications.~ (Postponed due to fork/cwd requirement).

##### **Motivation for or Use Case**

Make easier to developers to deal with liquibase.

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/issues/11398
https://github.com/mshima/generator-jhipster-liquibase/

-   [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
@pascalgrimaud bounty claimed https://opencollective.com/generator-jhipster/expenses/36634

pascalgrimaud:
@mshima : approved

PR #12238, opened 2020-08-10
Implement incremental liquibase.

mshima:
Closes https://github.com/jhipster/generator-jhipster/issues/12178.

---


pascalgrimaud:
Great work here ! 

@mshima : if you feel confident, plz go ahead and merge this one.
At the same time, is there some documentation to try this, or not yet ?

mshima:
> Great work here !
> 
> @mshima : if you feel confident, plz go ahead and merge this one.
> At the same time, is there some documentation to try this, or not yet ?

@pascalgrimaud just enable it with `jhipster --incremental-changelog --with-entities`.

mshima:
@pascalgrimaud I am in doubt how to handle field changes like type and validation.
Current approach is to let the user handle it, I could drop and recreate the column, but if you are ok, I prefer to handle it after merging this.

DanielFran:
Better to keep user manage the changes, this can involve scripts to manage data conversion, etc...

mshima:
> Better to keep user manage the changes, this can involve scripts to manage data conversion, etc...

Yep, that's the reason.

avdev4j:
Could you confirm @mshima that we have to choose between incremental-log or not at the beginning of the project. What happen if I want to have an incremental log for a given entity and recreate log for another? Should I edit the .yo-rc.json each time I manipulate the entity?

mshima:
> Could you confirm @mshima that we have to choose between incremental-log or not at the beginning of the project.

@avdev4j 
No. Can be done at an already existing project, just need to make sure the incremental needle is in place.
`jhipster --incremental-changelog ...` enables it.

> What happen if I want to have an incremental log for a given entity and recreate log for another? Should I edit the .yo-rc.json each time I manipulate the entity?

No .yo-rc.json/.jhipster manipulation is required, nothing is stored there, only `incrementalChangelog: (true/false)`.
If you manipulate `.jhipster/*` manually, you can lose old state (old state is .jhipster/*, new state is prompt answers/jdl generated entities).

It's incremental only. Old changelogs aren't touched. You can manipulate manually every liquibase changelog.
If you want to recreate the changelog `--recreate-initial-changelog`, an initial changelog will be created for every entity, but none will be deleted.

In my blueprint I store every change at .yo-rc.json for reproducibility. I left this out for simplicity.
If think reproducible changelog is useful, fell free to open a feature request. I don't think it's useful anymore.

I know 2 issues:
- It can not work correctly for jdls with more than 1 applications.
Is required to force it with `jhipster --no-fork microservice.jdl`
The reason is that we jhipster is not prepared to run in a cwd defined by variable, we need to fork it with a custom cwd.
Eg: `npm install` is executed in current cwd, without fork we need to pass the cwd variable to spawn.
- It will not work correctly `jhipster --ignore-application` or with no application.
Reason: `jhipster entity` is executed for each changed entity, so added relationship will not find the other side and will break the changelog. Reference will be created before the referenced file.
To fix this we will need to create a generator called entities to make the entities to be regenerated in sync.

62mkv:
@mshima 

> "To fix this we will need to create a generator called entities to make the entities to be regenerated in sync." 

is it possible to ask if could you define an issue for this? I am afraid `--ignore-application` is my primary mode of running `jhipster jdl` on existing application (otherwise it creates too much mess). Maybe in v7 it's gonna be different, idk. But if you feel it's worth a shot, then.. an issue would be nice to have, I think.

Thanks

mshima:
@62mkv it should be fixed, entities generator was created.
If it’s not working you can create a new bug report.

62mkv:
ok.. seems to work now ! ) so, once this command above has been run, I can add new field definition to some entity in my JDL file, then run `jhipster jdl app.jdl --ignore-application` and voila - it will generate separate Liquibase files for adding column and loadData! thanks @mshima

mshima:
@62mkv raise a new issue.
Even better if you provide a PR removing the javadoc/remarks from addColumn 😄 .

@gmarziou yep, some tips would be great.
And perhaps we should make it default soon.

Issue #12251, opened 2020-08-16
Create jhipster-lib snapshots and use them at CI.


mshima:
##### **Overview of the feature request**

Our current test approach is to clone jhipster-lib, build and use it.
We should consider to create snapshots for each push and deploy to github packages.
Those snapshots could be used at generator-jhipster tests.

##### **Motivation for or Use Case**

Benefits:
- ~~Reduce CI builds by 30~40 seconds each.~~
- Allows to simplify big dependencies update like spring boot major version bump. It could be done gradually.
Eg: create snapshots for jhipster-lib `-SB23` and add an option to use it.
- Sonatype snapshots are daily, so not suitable for our goal.

Limitations:
- Github packages requires authentication to read the packages
  - Not sure it's worth to migrate azure builds.
  - We should keep using daily snapshots at non github actions generated projects.

##### **Related issues or PR**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
> Our current test approach is to clone jhipster-lib, build and use it.

I like this approach, because we can reproduce the issue, with the exact commit of jhipster lib and generator-jhipster. 

By using snapshot version, we won't be able to reproduce the exact build, if there is a merge in jhipster lib : the previous snapshot will be erase.

The main problem would be the different approach with Azure as the github token is not available there. 

So for now, I'd prefer to keep our current system. 
But let's discuss ! :-) 
Let's discuss

mshima:
> > Our current test approach is to clone jhipster-lib, build and use it.
> 
> I like this approach, because we can reproduce the issue, with the exact commit of jhipster lib and generator-jhipster.

Confused about this.
Even if a snapshot was deleted it would be reproducible.

> By using snapshot version, we won't be able to reproduce the exact build, if there is a merge in jhipster lib : the previous snapshot will be erase.

A public package version is never deleted from github packages.
https://docs.github.com/en/packages/publishing-and-managing-packages/deleting-a-package
I don’t know about public package snapshots.

> The main problem would be the different approach with Azure as the github token is not available there.

We could create a token with read package permission only and use it.
Anyway not allowing unauthenticated read is my main concern.

pascalgrimaud:
Some packages have been published:
https://github.com/jhipster/jhipster/packages

Let's see if it works well.

DanielFran:
We need to update the generator to use this repository

pascalgrimaud:
[...]

PR #12252, opened 2020-08-16
Use jhipster-lib snapshots from github packages.

mshima:
- Use jhipster-lib snapshots.
- ~Add spring-boot 2.3 option example.~

Closes https://github.com/jhipster/generator-jhipster/issues/12251.
Depends on https://github.com/jhipster/jhipster/pull/785.

---

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -262,7 +262,15 @@ repositories {
     jcenter()
     <%_ if (reactive && databaseType === 'sql' || JHIPSTER_DEPENDENCIES_VERSION.endsWith('-SNAPSHOT')) { _%>
     maven {
+        <%_ if (process.env.JHI_GITHUB_CI === 'true') { _%>
+        credentials {
+            username = System.getenv("GITHUB_ACTOR")
+            password = System.getenv("GITHUB_TOKEN")
+        }
+        url "https://maven.pkg.github.com/jhipster/jhipster/"
+        <%_ } else { _%>
         url "https://oss.sonatype.org/content/repositories/snapshots/"
+        <%_ } _%>
         mavenContent {
             snapshotsOnly()
         }
--- a/test/server.spec.js
+++ b/test/server.spec.js
@@ -105,4 +105,80 @@ describe('JHipster server generator', () => {
             );
         });
     });
+
+    describe('java repositories', () => {
+        describe('when JHI_GITHUB_CI env variable is set', () => {
+            let githubCiBackup;
+            before(() => {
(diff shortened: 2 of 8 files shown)


generators/app/index.js
```diff
@@ -197,6 +197,11 @@ module.exports = class extends BaseBlueprintGenerator {
             type: Boolean,
         });
 
+        this.option('spring-boot23', {
+            desc: 'Use spring boot 2.3 (beta)',
```
pascalgrimaud, line 201: we should not rely on specific branch here... otherwise, next versions, we'll add spring boot 2.4, 2.5, 2.6 etc etc

generators/server/templates/build.gradle.ejs
```diff
@@ -262,7 +262,15 @@ repositories {
     jcenter()
     <%_ if (reactive && databaseType === 'sql' || JHIPSTER_DEPENDENCIES_VERSION.endsWith('-SNAPSHOT')) { _%>
     maven {
+        <%_ if (process.env.JHI_GITHUB_CI === 'true') { _%>
+        credentials {
```
pascalgrimaud, line 266: one another reason I'm not convinced about this approach is to generate specific code for our CI. 

It means it's different from local / CI

mshima:
> I don't like having specific code for a specific version of Spring Boot, in our template

The second commit it’s just an example of how to create a gradual upgrade of spring-boot.
JHipster features became too wide and difficult to upgrade at once.
The spring 2.3 branch has diverted a lot, most of the features were ready.
So I tried to create a solution to allow setting sb2.3 as optional temporary.

generators/server/templates/build.gradle.ejs
mshima: Maven could be configured with a custom settings.xml.
Don’t know about Gradle.

test/server.spec.js
```diff
@@ -105,4 +105,80 @@ describe('JHipster server generator', () => {
...
+            });
+            after(() => {
+                process.env.JHI_GITHUB_CI = githubCiBackup;
+            });
+
+            it('generates pom.xml without github repository', () => {
+                assert.noFileContent('pom.xml', '<id>github</id>');
+            });
+        });
+    });
```
mshima, line 183:
```suggestion
```

generators/app/index.js
mshima: Will be removed before merging the PR, and added if a partial spring-boot 2.3 support is added.

DanielFran:
Close/Reopen to validate if it uses the github package repo

generators/app/index.js
mshima: Removed

PR #12298, opened 2020-08-24
[Cypress] Upgrade to Cypress 5.0.0

nassimerrahoui:
---

### DIFF
--- a/generators/client/templates/angular/package.json.ejs
+++ b/generators/client/templates/angular/package.json.ejs
@@ -145,7 +145,7 @@
     "ts-node": "8.10.2",
     <%_ } _%>
     <%_ if (cypressTests) { _%>
-    "cypress": "4.12.1",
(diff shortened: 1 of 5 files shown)


avdev4j:
Do you check that we upgrade tests according to the changelog (if needed)?  https://docs.cypress.io/guides/references/changelog.html#5-0-0

nassimerrahoui:
> Do you check that we need to upgrade tests according to the changelog (if needed)? https://docs.cypress.io/guides/references/changelog.html#5-0-0

There is a change with Cypress.blob type

nassimerrahoui:
> Do you check that we need to upgrade tests according to the changelog (if needed)? https://docs.cypress.io/guides/references/changelog.html#5-0-0

It seems that there is an error with tsconfig :

`TS5053: Option 'sourceMap' cannot be specified with option 'inlineSourceMap'.`

PR #12307, opened 2020-08-25
Cypress

avdev4j:
---


.github/workflows/react.yml
```diff
@@ -183,7 +183,13 @@ jobs:
               run: npm run e2e:update-webdriver --versions.chrome ${{ steps.chromedriver-version.outputs.version }}
               working-directory: /home/runner/app
             - name: 'TESTS: End-to-End'
+              id: e2e
               run: $JHI_SCRIPTS/24-tests-e2e.sh
+            - uses: actions/upload-artifact@v2
+              if: ${{ always() && steps.e2e.outcome == 'failure'}}
```
qmonmert, line 189: ${{ always() && steps.e2e.outcome == 'failure'}} => ${{ always() && steps.e2e.outcome == 'failure' }} 😇

avdev4j: done ;)

qmonmert:
LGTM 👏

qmonmert (approved):

generators/client/files-common.js
```diff
@@ -0,0 +1,39 @@
...
+const commonFiles = {
+    common: [
+        {
+            templates: ['.eslintignore'],
+        },
+    ],
+};
+
+function writeFiles() {
+    mkdirp(MAIN_SRC_DIR);
```
mshima, line 37:
```suggestion
```

Remove this, not sure why it was added.

azure-pipelines.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       variables:
           JHI_PROFILE: dev,webpack
           JHI_RUN_APP: 1
-          JHI_PROTRACTOR: 0
+          JHI_CYPRESS: 0
```
mshima, line 31:
```suggestion
          JHI_E2E: 0
```
Should change this?

generators/client/templates/common/.eslintignore.ejs
```diff
@@ -18,7 +18,12 @@
 -%>
 node_modules/
 src/main/docker/
+<%_ if (cypressTests) { _%>
+src/test/javascript/cypress/
```
mshima, line 22: Should the entire cypress for be ignored?

azure-pipelines.yml
avdev4j: yes, for sure

generators/client/templates/common/.eslintignore.ejs
avdev4j: > I though it was merged to master 😆.

Not yet, but soon :p

I guess it can done when I'll review all your last suggestions.

avdev4j:
> I though it was merged to master 😆.

Not yet, but soon :p

I guess it can done when I'll review all your last suggestions.

generators/client/templates/common/.eslintignore.ejs
avdev4j: I think your are right, we should not ignore cypress at all (well I see no reason) but we if do that we need to add cypress folder in includes array in the tsconfig and fix all lint errors. I'll do that

generators/cypress/templates/src/test/javascript/cypress/tsconfig.json.ejs
```diff
@@ -0,0 +1,7 @@
+{
```
qmonmert, line 1: you can replace all by: 
```
{
  "references": [
    {
      "path": "./../../../../tsconfig.json"
    }
  ]
}
```

generators/client/templates/common/.eslintignore.ejs
avdev4j: https://github.com/jhipster/generator-jhipster/pull/12309

generators/cypress/templates/src/test/javascript/cypress/tsconfig.json.ejs
nassimerrahoui: [jhipster#12310](https://github.com/jhipster/generator-jhipster/pull/12310)

generators/client/templates/common/.eslintignore.ejs
avdev4j: hum... it seems we did that because adding Cypress in the project (tsconfig by using include) fire errors and conflicts between libraries. The solution was to exclude Cypress to the tsconfig file and also add it to the eslint ignore to avoid errors like this:
The file does not match your project config [...]
he file must be included in at least one of the projects provided at [...]

mshima: Agreed, this is not simple to fix.
A reference:
https://github.com/cypress-io/cypress-and-jest-typescript-example

nassimerrahoui (approved):

mshima:
IMO we disable cypress for consul+oauth2 and merge.
https://github.com/jhipster/generator-jhipster/blob/1b52b14315842df746142896b87117f972473e5c/test-integration/samples/ms-react-gateway-consul-oauth2/.yo-rc.json#L20

mraible:
@mshima I'm curious: does Cypress have issues with Consul and OAuth 2.0, but Protractor does not?

mshima:
> @mshima I'm curious: does Cypress have issues with Consul and OAuth 2.0, but Protractor does not?

I think cypress tests are much more complete than protractor was.
- Protractor was using fixed values for fields (AAA, BBB, etc), now it is using faker (same logic we use for liquibase fake data).
- It's probably enabled in more use cases (including more fields).
- It's only disabled for required relationships or if it fails to create a compatible value.

That specific error maybe related only to react+gateway+consul+oauth2 (hard to debug and there is no stacktrace).
I think that this fail is not related to this PR, it only discovered it, so it should not block the entire PR.
We could open a new issue to address that specific use case.

We could the test to github actions, we will have the screenshot available, it could be helpful.

avdev4j:
Ok @mshima in this case I've just switch back to Protractor for this build, I think it's better than nothing at all.

mshima:
The failure didn't happened in the merge event, so let's merge.

test-integration/scripts/24-tests-e2e.sh
```diff
@@ -43,7 +43,7 @@ launchCurlOrProtractor() {
         return 1
     fi
 
-    if [ "$JHI_PROTRACTOR" != 1 ]; then
+    if [ "$JHI_E2E" != 1 ]; then
```
mshima, line 46: This will disable daily builds tests and other repositories that uses this script.
I think we should test both JHI_PROTRACTOR and JHI_E2E.

avdev4j: Exact, we should migrate daily build to use JHI_E2E too.

DanielFran: Do not forget that there are used also by jhipster/jhipster project

avdev4j: Thanks Daniel, I will migrate all tomorrow and open PRs.

PR #12309, opened 2020-08-25
[Cypress] Enable eslint for Cypress

avdev4j:
---

### DIFF
--- a/generators/entity-client/templates/common/src/test/javascript/cypress/integration/entity/entity.spec.ts.ejs
+++ b/generators/entity-client/templates/common/src/test/javascript/cypress/integration/entity/entity.spec.ts.ejs
@@ -166,10 +166,7 @@ describe('<%= entityClass %> e2e test', () => {
 
     <%_ } else if (fieldType === 'Boolean') { _%>
     cy.get(`[data-cy="<%= fieldName %>"]`).should('not.be.checked');
-    if ('<%= fieldValue %>' === 'true') {
-        cy.get(`[data-cy="<%= fieldName %>"]`).click().should('be.checked');
-    }
-
+    cy.get(`[data-cy="<%= fieldName %>"]`).click().should('be.checked');
     <%_ } else if (['byte[]', 'ByteBuffer'].includes(fieldType) && fieldTypeBlobContent === 'text') { _%>
     cy.get(`[data-cy="<%= fieldName %>"]`).type('<%= fieldValue %>', { force: true }).invoke('val').should('match', new RegExp('<%= fieldValue %>'));
 
--- a/generators/cypress/templates/src/test/javascript/cypress/support/entity.ts.ejs
+++ b/generators/cypress/templates/src/test/javascript/cypress/support/entity.ts.ejs
@@ -43,10 +43,10 @@ Cypress.Commands.add('setFieldImageAsBytesOfEntity', (fieldName: string, fileNam
   cy.fixture(fileName).as('image')
     .get(`[data-cy="${fieldName}"]`).then(function(el) {
       const blob = Cypress.Blob.base64StringToBlob(this.image, mimeType);
-      let file = new File([blob], fileName, { type: mimeType });
-      let list = new DataTransfer();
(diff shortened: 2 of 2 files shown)


nassimerrahoui (approved):

generators/entity-client/templates/common/src/test/javascript/cypress/integration/entity/entity.spec.ts.ejs
```diff
@@ -166,10 +166,7 @@ describe('<%= entityClass %> e2e test', () => {
 
     <%_ } else if (fieldType === 'Boolean') { _%>
     cy.get(`[data-cy="<%= fieldName %>"]`).should('not.be.checked');
-    if ('<%= fieldValue %>' === 'true') {
-        cy.get(`[data-cy="<%= fieldName %>"]`).click().should('be.checked');
-    }
-
+    cy.get(`[data-cy="<%= fieldName %>"]`).click().should('be.checked');
```
mshima, line 169: This will loose randomization
```suggestion
    <%_ if (fieldValue) {  _%>
    cy.get(`[data-cy="<%= fieldName %>"]`).click().should('be.checked');
    <%_ } _%>
```

nassimerrahoui: It is necessary because some entites need to have their boolean to be true

mshima: Why?

nassimerrahoui: When we try to create entity's instance, sometimes tests failed because it can't validate creation because of a boolean field which is not checked

nassimerrahoui: So, we think that is better to always check boolean's field to avoid problem with required booleans

mshima: It’s a bug at the client side so.
The model should be created with false value if the field is required.
Or we should switch to a radio instead of a checkbox.
Anyway I think you can open an issue to discuss it, WYT?
Updated the suggestion

mshima: IMO requiring the user to click the checkbox, falsely implies the checkbox must be true.

mshima: This is cypress discovering bugs 😄

nassimerrahoui: There is so many cases right here with entities :sweat_smile:

nassimerrahoui: > It’s a bug at the client side so.
> The model should be created with false value if the field is required.
> Or we should switch to a radio instead of a checkbox.
> Anyway I think you can open an issue to discuss it, WYT?
> Updated the suggestion

For the moment, I can't reproduce the error. So, let's not open an issue for that

nassimerrahoui: My bad, I think it is probably link to this issue [jhipster#12296](https://github.com/jhipster/generator-jhipster/pull/12296).

mshima: @nassimerrahoui can we use a randomized 1 click then?

nassimerrahoui: IMO, as we are not sure that the problem will not occur next times, I think it is more safe to always keep the click

PR #12321, opened 2020-08-27
Avoid password and resetKey column to be sorted by the /users API

### LINKED ISSUE #12312: Security issues in the user management

On the user management webpage: `http://localhost:8080/admin/user-management`, it is possible to sort columns and the following URL can be observed when sorting by Login:
`http://localhost:8080/admin/user-management?page=1&sort=login,desc`
The data displayed on the screen is from the `generators/server/templates/src/main/java/package/domain/User.java.ejs` which contains data from the `jhi_user` database table. The issue is that some properties of the model contain security related data and that an administrator can sort them out. For example:
`http://localhost:8080/admin/user-management?page=1&sort=password,desc`
In this case, the `password` property is the `password_hash` column of the database.

Even if this feature is supposed to be only for administrative purpose, thus requiring a high privilege account, the `password` and `resetKey` properties should not be exposed.
Indeed, the `resetKey` of any user could be retrieved with the following methodology:
* create a user on the application
* reset the password and retrieve the key, then sort the users and compare the `resetKey` with a user we want to retrieve its `resetKey`
* re-reset the password so the first character change again and compare it again. By dichotomy, this technique might help recover each character of the `resetKey` of a certain user

This technique would require a lot of requests to generate `resetKey` that would allow retrieving the resetKey of another user because of its 20 character lengths. It is pretty unlikely that a working attack could be made.

This technique can not be applied to the password as the bcrypt hashing algorithm is used and contains a random salt that is unkown to an attacker.

Moreover, this technique could be applied to any other data exposed by the model.

##### **Reproduce the error**

Explains on last paragraph

##### **Related issues**

N/A

##### **Suggest a Fix**

Security related elements specific to a user should not be present in a model used for arbitrary sorting to a user/administrator.

##### **JHipster Version(s)**

```
blog@0.0.1-SNAPSHOT /home/laxa/Documents/blog
└── generator-jhipster@6.10.1
```

<details>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "org.jhipster.blog"
    },
    "jhipsterVersion": "6.10.1",
    "baseName": "blog",
    "packageName": "org.jhipster.blog",
    "packageFolder": "org/jhipster/blog",
    "serverPort": "8080",
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "buildTool": "maven",
    "jwtSecretKey": "YourJWTSecretKeyWasReplacedByThisMeaninglessTextByTheJHipsterInfoCommandForObviousSecurityReasons",
    "embeddableLaunchScript": false,
    "useSass": true,
    "clientFramework": "react",
    "clientTheme": "none",
    "clientThemeVariant": "",
    "creationTimestamp": 1598011041769,
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "otherModules": [],
    "enableTranslation": false,
    "blueprints": []
  }
}

</pre>
</details>

N/A

Not applicable here

-   [ X] Checking this box is mandatory (this is just to show you read everything)

avdev4j:
fix #12312

---

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
@@ -31,6 +31,8 @@ import <%= packageName %>.repository.search.UserSearchRepository;
 import <%= packageName %>.security.AuthoritiesConstants;
 <%_ if (authenticationType !== 'oauth2') { _%>
 import <%= packageName %>.service.MailService;
+import org.springframework.data.domain.Sort;
+import java.util.Collections;
 <%_ } _%>
 import <%= packageName %>.service.UserService;
 import <%= packageName %>.service.dto.<%= asDto('User') %>;
@@ -126,6 +128,9 @@ import static org.elasticsearch.index.query.QueryBuilders.*;
 @RestController
 @RequestMapping("/api")
 public class UserResource {
+    <%_ if (authenticationType !== 'oauth2') { _%>
+    private static final List<String> FORBIDDEN_ORDERED_PROPERTIES = Collections.unmodifiableList(Arrays.asList("resetKey", "password"));
+    <%_ } _%>
 
     private final Logger log = LoggerFactory.getLogger(UserResource.class);
 
@@ -287,17 +292,36 @@ public class UserResource {
     <%_ if (databaseType === 'sql' || databaseType === 'mongodb' || databaseType === 'neo4j' || databaseType === 'couchbase') { _%>
         <%_ if (!reactive) { _%>
     public ResponseEntity<List<<%= asDto('User') %>>> getAllUsers(Pageable pageable) {
+        <%_ if (authenticationType !== 'oauth2') { _%>
+        if (containsForbiddenSortedProperties(pageable)) {
+            return ResponseEntity.badRequest().build();
+        }
+        <% } %>
         final Page<<%= asDto('User') %>> page = userService.getAllManagedUsers(pageable);
         HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(ServletUriComponentsBuilder.fromCurrentRequest(), page);
         return new ResponseEntity<>(page.getContent(), headers, HttpStatus.OK);
         <%_ } else { _%>
     public Mono<ResponseEntity<Flux<<%= asDto('User') %>>>> getAllUsers(ServerHttpRequest request, Pageable pageable) {
+        <%_ if (authenticationType !== 'oauth2') { _%>
+        if (containsForbiddenSortedProperties(pageable)) {
+            return Mono.just(ResponseEntity.badRequest().build());
+        }
+        <% } %>
         return userService.countManagedUsers()
             .map(total -> new PageImpl<>(new ArrayList<>(), pageable, total))
             .map(page -> PaginationUtil.generatePaginationHttpHeaders(UriComponentsBuilder.fromHttpRequest(request), page))
             .map(headers -> ResponseEntity.ok().headers(headers).body(userService.getAllManagedUsers(pageable)));
         <%_ } _%>
     }
+
+    <%_ if (authenticationType !== 'oauth2') { _%>
+    private boolean containsForbiddenSortedProperties(Pageable pageable) {
+        return pageable.getSort()
+            .stream()
+            .map(Sort.Order::getProperty)
+            .anyMatch(FORBIDDEN_ORDERED_PROPERTIES::contains);
+    }
+    <%_ } _%>
 
     /**
      * Gets a list of all roles.
--- a/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
@@ -93,6 +93,7 @@ import java.util.*;
 <%_ if (reactive) { _%>
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Arrays;
 <%_ } _%>
 <%_ if (searchEngine === 'elasticsearch' && !reactive) { _%>
 import java.util.stream.Collectors;
@@ -129,7 +130,7 @@ import static org.elasticsearch.index.query.QueryBuilders.*;
 @RequestMapping("/api")
 public class UserResource {
     <%_ if (authenticationType !== 'oauth2') { _%>
-    private static final List<String> FORBIDDEN_ORDERED_PROPERTIES = Collections.unmodifiableList(Arrays.asList("resetKey", "password"));
+    private static final List<String> ALLOWED_ORDERED_PROPERTIES = Collections.unmodifiableList(Arrays.asList("id", "login", "firstName", "lastName", "email", "activated", "langKey"));
     <%_ } _%>
 
     private final Logger log = LoggerFactory.getLogger(UserResource.class);
@@ -293,7 +294,7 @@ public class UserResource {
         <%_ if (!reactive) { _%>
     public ResponseEntity<List<<%= asDto('User') %>>> getAllUsers(Pageable pageable) {
         <%_ if (authenticationType !== 'oauth2') { _%>
-        if (containsForbiddenSortedProperties(pageable)) {
+        if (!onlyContainsAllowedProperties(pageable)) {
             return ResponseEntity.badRequest().build();
         }
         <% } %>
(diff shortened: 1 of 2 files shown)


generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -287,18 +292,31 @@ public class UserResource {
...
+            return Mono.just(ResponseEntity.badRequest().build());
+        }
         return userService.countManagedUsers()
             .map(total -> new PageImpl<>(new ArrayList<>(), pageable, total))
             .map(page -> PaginationUtil.generatePaginationHttpHeaders(UriComponentsBuilder.fromHttpRequest(request), page))
             .map(headers -> ResponseEntity.ok().headers(headers).body(userService.getAllManagedUsers(pageable)));
         <%_ } _%>
     }
 
+    private boolean isContainsForbiddenSortedProperties(Pageable pageable) {
```
cbornet, line 313: maybe just containsForbiddenSortedProperties ?

cbornet:
Maybe, the sort parameters could be just ignored instead of returning an error ? See https://stackoverflow.com/a/53479298/4862883 for a possible solution.

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -149,6 +153,7 @@ public class UserResource {
         <%_ if (authenticationType !== 'oauth2') { _%>
         this.userRepository = userRepository;
         this.mailService = mailService;
+        this.forbiddenOrderedProperties = Collections.unmodifiableList(List.of("resetKey", "password"));
```
cbornet, line 156: Make it a constant ?

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -287,18 +292,31 @@ public class UserResource {
     <%_ if (databaseType === 'sql' || databaseType === 'mongodb' || databaseType === 'neo4j' || databaseType === 'couchbase') { _%>
         <%_ if (!reactive) { _%>
     public ResponseEntity<List<<%= asDto('User') %>>> getAllUsers(Pageable pageable) {
+        if(isContainsForbiddenSortedProperties(pageable)) {
```
murdos, line 295:
```suggestion
        if (isContainsForbiddenSortedProperties(pageable)) {
```

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -287,18 +292,31 @@ public class UserResource {
...
+            return ResponseEntity.badRequest().build();
+        }
         final Page<<%= asDto('User') %>> page = userService.getAllManagedUsers(pageable);
         HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(ServletUriComponentsBuilder.fromCurrentRequest(), page);
         return new ResponseEntity<>(page.getContent(), headers, HttpStatus.OK);
         <%_ } else { _%>
     public Mono<ResponseEntity<Flux<<%= asDto('User') %>>>> getAllUsers(ServerHttpRequest request, Pageable pageable) {
+        if(isContainsForbiddenSortedProperties(pageable)) {
```
murdos, line 303:
```suggestion
        if (isContainsForbiddenSortedProperties(pageable)) {
```

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
murdos, line 156: `List.of()` is only available since Java 9. I thought it was decided to stay with Java 8 syntax.
Also, `List.of()` already returns an immutable collection, there's no need for the additionnal `Collections.unmodifiableList`

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -31,6 +31,8 @@ import <%= packageName %>.repository.search.UserSearchRepository;
 import <%= packageName %>.security.AuthoritiesConstants;
 <%_ if (authenticationType !== 'oauth2') { _%>
 import <%= packageName %>.service.MailService;
+import org.springframework.data.domain.Sort;
```
qmonmert, line 34: tests are ❌ because this import is misplaced

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
avdev4j: You're right, I'll revert it to Java 8 syntax.

atomfrede: Should we add `activationKey`, `resetDate` and `persistentTokens` (for session, guess we remove it in 7.x) too?

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
avdev4j: not really :D. the private method (and the if statement) should not be applied when we choose Oauth2 (there si no password or resetKey with oauth2.

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
avdev4j: What do you think @laxaa ?

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -128,16 +130,19 @@ import static org.elasticsearch.index.query.QueryBuilders.*;
...
+    <%_ if (authenticationType !== 'oauth2') { _%>
+    private static final List<String> FORBIDDEN_ORDERED_PROPERTIES = Collections.unmodifiableList(Arrays.asList("resetKey", "password"));
+    <%_ } _%>
 
     @Value("${jhipster.clientApp.name}")
     private String applicationName;
 
     private final UserService userService;
 <%_ if (authenticationType !== 'oauth2') { _%>
-
```
cbornet, line 137: should not be removed

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
```diff
@@ -128,16 +130,19 @@ import static org.elasticsearch.index.query.QueryBuilders.*;
...
     private final UserRepository userRepository;
 
     private final MailService mailService;
+
```
cbornet, line 145: line should not be added

avdev4j:
> Maybe, the sort parameters could be just ignored instead of returning an error ? See https://stackoverflow.com/a/53479298/4862883 for a possible solution.

I'm not sure because I think it's cleaner to send the right Http code when a client asking for a parameters he is not supposed to used. Skipping those parameters can cause a misunderstood result, "why my result set is not sorted while I asked for a sorted one?"

generators/server/templates/src/main/java/package/web/rest/UserResource.java.ejs
laxaa: I would have advised to remove the unnecessary fields from the model if its possible.
Otherwise, I would rather go for a white-list rather than a black-list, might avoid future problems that way if your model evolves.

avdev4j: hmmm, it's a great idea, I'm not sure we can remove fields at the moment. Let's do a white list so.

mshima:
> > Maybe, the sort parameters could be just ignored instead of returning an error ? See https://stackoverflow.com/a/53479298/4862883 for a possible solution.
> 
> I'm not sure because I think it's cleaner to send the right Http code when a client asking for a parameters he is not supposed to used. Skipping those parameters can cause a misunderstood result, "why my result set is not sorted while I asked for a sorted one?"

I prefer to ignore, returning an error could reveal details of the implementation (not a big deal).
We could log the error.

avdev4j:
I understand this POV but with the witheList approach the message is a little bit different "This request is not correct because the fields you want to order from are nor supported".
The fields we are talking about could be "password" but "aWeirdField" too, so we don't give any information about the implementation details.

PR #12333, opened 2020-08-28
Datetime column type doesn't store sub-second informations on MySQL and MariaDB

gzsombor:
Datetime column type doesn't store sub-second informations on MySQL and MariaDB by default, here the precision defaults to 0.
As opposed to PostgreSQL and H2, where it assumes that 6 digit should be stored.

* https://mariadb.com/kb/en/microseconds-in-mariadb/
* https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html
* https://www.postgresql.org/docs/current/datatype-datetime.html
* http://h2database.com/html/datatypes.html

---

### DIFF
--- a/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
+++ b/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
@@ -76,9 +76,9 @@
                 } else if (fieldType === 'LocalDate') {
                     columnType="date";
                 } else if (fieldType === 'Instant') {
-                    columnType="datetime";
+                    columnType="${datetimeType}";
                 } else if (fieldType === 'ZonedDateTime') {
-                    columnType="datetime";
+                    columnType="${datetimeType}";
                 } else if (fieldType === 'Duration') {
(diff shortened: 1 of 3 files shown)


mshima:
This PR changed csv datatime data from datetime to string.
```
<column name="date" type="string"/>
```
Before it was:
```
<column name="date" type="datetime"/>
```
Should that be ${datetimeType}?

pascalgrimaud:
I think you're right, there is something to check here

gzsombor:
Sorry, I can't understand, how can I reproduce the issue?

mshima:
@gzsombor I saw this by https://github.com/jhipster/generator-jhipster/pull/12238 logs.
I really don't know if it's a problem at all.
Anyway, https://github.com/jhipster/generator-jhipster/pull/12238 probably will be merged today, and it handles this change.

Just for completeness, this comparison should be tested with `${datetimeType}`
https://github.com/jhipster/generator-jhipster/blob/360582c22965654520bf3d7615771f2ce4f6af7f/generators/entity-server/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs#L197

PR #12337, opened 2020-08-29
[webflux] Enable testcontainer support for MariaDB too

gzsombor:
and add converter from BitSet to Boolean which is needed by the MariaDB driver

---

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -2275,6 +2275,7 @@ module.exports = class extends PrivateBase {
         dest.searchEngine = config.searchEngine;
         dest.cacheProvider = config.cacheProvider;
         dest.enableHibernateCache = config.enableHibernateCache;
+        dest.reactiveSqlTestContainers = config.reactive && ['mysql', 'postgresql', 'mssql', 'mariadb'].includes(config.prodDatabaseType);
 
         dest.enableSwaggerCodegen = config.enableSwaggerCodegen;
         dest.messageBroker = config.messageBroker;
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -1410,7 +1410,7 @@ const serverFiles = {
             templates: [{ file: 'testcontainers/mariadb/my.cnf', method: 'copy', noEjs: true }],
         },
         {
-            condition: generator => ['mysql', 'postgresql', 'mssql'].includes(generator.prodDatabaseType) && generator.reactive,
+            condition: generator => generator.reactiveSqlTestContainers,
             path: SERVER_TEST_SRC_DIR,
             templates: [
                 {
--- a/.github/workflows/webflux.yml
+++ b/.github/workflows/webflux.yml
@@ -45,7 +45,8 @@ jobs:
         name: ${{ matrix.app-type }}
         runs-on: ${{ matrix.os }}
         if: >-
-            contains(github.event.head_commit.message, '[webflux]') &&
+            (contains(github.event.head_commit.message, '[webflux]') ||
+                contains(github.event.pull_request.title, '[webflux]')) &&
             !contains(github.event.head_commit.message, '[ci skip]') &&
             !contains(github.event.head_commit.message, '[skip ci]') &&
             !contains(github.event.pull_request.title, '[skip ci]') &&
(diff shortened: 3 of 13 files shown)


generators/server/files.js
```diff
@@ -1410,7 +1410,7 @@ const serverFiles = {
             templates: [{ file: 'testcontainers/mariadb/my.cnf', method: 'copy', noEjs: true }],
         },
         {
-            condition: generator => ['mysql', 'postgresql', 'mssql'].includes(generator.prodDatabaseType) && generator.reactive,
+            condition: generator => ['mysql', 'postgresql', 'mssql', 'mariadb'].includes(generator.prodDatabaseType) && generator.reactive,
```
mshima, line 1413:
```suggestion
            condition: generator.reactiveSqlTestContainers,
```

generators/generator-base.js
```diff
@@ -2284,6 +2284,7 @@ module.exports = class extends PrivateBase {
         dest.searchEngine = config.searchEngine;
         dest.cacheProvider = config.cacheProvider;
         dest.enableHibernateCache = config.enableHibernateCache;
+        dest.reactiveSqlTestContainers = config.reactiveSqlTestContainers;
```
mshima, line 2287: Can be dropped, this would load the config if it's stored at .yo-rc.json
```suggestion
```

gzsombor: If I remove it, the reactiveSqlTestContainers variable is not accessible from the templates.

generators/server/files.js
gzsombor: Thx, I forgot to update it

generators/generator-base.js
mshima: Ok, you are using to load into entity generator.
`reactiveSqlTestContainers` doesn't exists in `config`, you need to calculate it.
Updated the suggestion.

generators/server/index.js
```diff
@@ -272,6 +272,9 @@ module.exports = class extends BaseBlueprintGenerator {
                     } else {
                         this.containerVersion = 'latest';
                     }
+                    this.reactiveSqlTestContainers =
+                        this.jhipsterConfig.reactive &&
+                        ['mysql', 'postgresql', 'mssql', 'mariadb'].includes(this.jhipsterConfig.prodDatabaseType);
```
mshima, line 277: Can be removed, it's already loaded by loadAppConfig().
```suggestion
```

gzsombor:
Great, everything, but the webflux tests were run.

mshima:
Please change 
https://github.com/jhipster/generator-jhipster/blob/161519958d55863962abc03a8907ab24fc05ffbb/.github/workflows/webflux.yml#L48
to
```
            (contains(github.event.head_commit.message, '[webflux]') ||
                contains(github.event.pull_request.title, '[webflux]')) &&
```

MathieuAA:
I'm okay with `reactiveSqlTestContainers` as it's not a function (nor a function call) but a variable.

avdev4j:
As I remember we struggle with Webflux builds because the reactive option was totally broken (problem with compilation or backend tests...). That's why we had to merge the feature before having a fix. We did not want to wait more and avoid as much as possible conflicts (a little bit hard with branch with more than 100 commits :( ).

Does @nassimerrahoui or @adilabed any of you can help on this? I think we should open a specific PR to fix those tests.

mshima:
I think /account/password and /account/settings errors are related to the backend:
Eg: maybe https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/service/InvalidPasswordException.java.ejs returns error 401
while https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/src/main/java/package/web/rest/errors/InvalidPasswordException.java.ejs returns error 400.

The others seems to be cypress failing to login or keeping the user logged.

Anyway the errors should be fixed with spring 2.3 directly.

gzsombor:
Yes, as these change only touch the reactive-sql codepaths, it should not effect the MongoDB tests, or the oauth2 ones, I'm going to merge it, so I can finalize the reactive-entity support changes!

PR #12446, opened 2020-09-14
Use then-request Instead of sync-request

SudharakaP:
The library sync-request cannot be used in production applications (reference: https://www.npmjs.com/package/sync-request). In jhipster-online, sync-request is I think the root cause of generation being hanging. When I look at the processes always see a stuck process originating from this library. Therefore I would suggest replacing this with then-process instead.

Fixes https://github.com/jhipster/jhipster-online/issues/233

---

### DIFF
--- a/generators/openapi-client/prompts.js
+++ b/generators/openapi-client/prompts.js
@@ -93,7 +93,7 @@ function askActionType() {
             name: 'jhipsterEndpoint',
             message: 'Enter the URL of the running Jhipster instance',
             default: 'http://localhost:8080',
-            validate: async (input) => {
+            validate: async input => {
                 try {
                     const availableDocs = await fetchSwaggerResources(input);
 
@@ -133,7 +133,7 @@ function askActionType() {
             message: 'Where is your Swagger/OpenAPI spec (URL or path) ?',
             default: 'http://petstore.swagger.io/v2/swagger.json',
             store: true,
-            validate: async (input) => {
+            validate: async input => {
                 try {
                     if (/^((http|https):\/\/)/.test(input)) {
                         await request('GET', `${input}`, {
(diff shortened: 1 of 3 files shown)


mshima:
@SudharakaP I’ve adapted the api to async request. But did using mobile, no prettier available.
Probably this needs manual test.

generators/openapi-client/prompts.js
```diff
@@ -160,7 +160,7 @@ function askActionType() {
         this.props = props;
 
         if (props.jhipsterEndpoint !== undefined) {
-            return fetchSwaggerResources(props.jhipsterEndpoint).then(availableDocs => {
+            fetchSwaggerResources(props.jhipsterEndpoint).then(availableDocs => {
```
mshima, line 163: This return is required.
```suggestion
            return fetchSwaggerResources(props.jhipsterEndpoint).then(availableDocs => {
```

generators/openapi-client/prompts.js
```diff
@@ -152,31 +150,32 @@ function askActionType() {
...
+        if (props.jhipsterEndpoint !== undefined) {
+            fetchSwaggerResources(props.jhipsterEndpoint).then(availableDocs => {
+                props.availableDocs = availableDocs;
+            });
+        }
+        if (props.jhipsterDirectory !== undefined) {
+            props.inputSpec = `${props.jhipsterDirectory}/src/main/resources/swagger/api.yml`;
+        } else if (props.customEndpoint !== undefined) {
+            props.inputSpec = props.customEndpoint;
+        }
```
mshima, line 171:
```suggestion
        }
        return undefined;
```

SudharakaP:
@mshima : Thanks for modifying the openclient-api as well. I ran it on a project and it seems to work well without any errors. So I think we could merge this one if you are okay. 🤔

mshima:
@SudharakaP fell free to merge. I didn’t test.
The test focus should be jhipsterEndpoint prompt validation.

SudharakaP:
> @SudharakaP fell free to merge. I didn’t test.
> The test focus should be jhipsterEndpoint prompt validation.

@mshima : Yes it seems to work well. So I am merging. 😄

PR #12507, opened 2020-09-14
Add a Patch HTTP method when generating entities

### LINKED ISSUE #10077: Partial Update of objects using PATCH
Currently, only post and put are supported. When put becomes costly, or business rules do not always permit, partial update of data objects is useful

1. Based on authorisation rules, certain users can view/edit only some fields of the object while certain users have full access. Partial update support here becomes handy
2. Cost of large size and volume of DTOs can be minimised

PATCH method support for partial update #1315

-   [ Y ] Checking this box is mandatory (this is just to show you read everything)

qlereboursBS:
The aim of this PR is to allow developers to partially update the entities thanks to a Patch method in the Controller.

## How we plan to partially update an entity?
Let's say we have a `Customer` entity with two fields:
 - `name: String`
 - `age: Int`
If you want to update the customer that has the id `1`, you'll be able to call a `PATCH` on `localhost:8080/api/customers` with data:
```
{
  "id": 1,
  "name": "another name"
}
```

We're implementing [JSON merge patch RFC](https://tools.ietf.org/html/rfc7386.html).
**Note:** At the beginning, we won't support updating fields to set them null because Jackson can't understand the difference between a null field that is null because it's been set to null, and a null field that is null because it just not defined.
In other words: We won't be able to erase a field with the patch.

## Todo list
-   [x] Edit the EntityResource.java.ejs
-   [x] Extract patch code in a template `patch_template.ejs`
-   [x] Add integration tests
-   [x] Make it work when DTO are enabled
-   [x] Make it work when services are enabled
-   [x] Make it work with reactive streams
-   [x] Make it work with saveUserSnapshot condition
-   [x] Update documentation (if needed)
---

fix #10077 

- [ ] [All continuous integration](https://github.com/jhipster/generator-jhipster/actions) tests are green
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md](https://github.com/jhipster/generator-jhipster/blob/master/CONTRIBUTING.md) document are followed


cbornet:
Congrats @qlereboursBS on your first PR to JHipster :clap: :+1: !

atomfrede:
Congrats to your first contribution. Out of scope for this pr but https://github.com/OpenAPITools/jackson-databind-nullable it is possible to use null to erase a field via patch but this would mean huge changes in the entity generator

cbornet:
Yes we discussed this during JHipster Code. A dedicated DTO will be needed. This will be step 2 of PATCH support.

qlereboursBS:
We've got to fix and improve the tests now, but the whole code generation part seems to be done, and working.
I'm going to put some code samples generated to make it easier to discuss.
I didn't Webflux / Reactive streams before this PR, don't hesitate to comment if anything's wrong

My test entity is `Product(name, price, isAvailable)`

**Without service, with reactive streams:**
```
    /**
     * {@code PATCH  /products} : Updates given fields of an existing product.
     *
     * @param product the product to update.
     * @return the {@link ResponseEntity} with status {@code 200 (OK)} and with body the updated product,
     * or with status {@code 400 (Bad Request)} if the product is not valid,
     * or with status {@code 404 (Not Found)} if the product is not found,
     * or with status {@code 500 (Internal Server Error)} if the product couldn't be updated.
     * @throws URISyntaxException if the Location URI syntax is incorrect.
     */
    @PatchMapping(value = "/products", consumes = "application/merge-patch+json")
    public Mono<ResponseEntity<Product>> partialUpdateProduct(@NotNull @RequestBody Product product) throws URISyntaxException {
        log.debug("REST request to update Product partially : {}", product);
        if (product.getId() == null) {
        }

        Mono<Product> result = productRepository
            .findById(product.getId())
            .map(
                existingProduct -> {
                    if (product.getName() != null) {
                        existingProduct.setName(product.getName());
                    }
                    if (product.getPrice() != null) {
                        existingProduct.setPrice(product.getPrice());
                    }
                    if (product.isIsAvailable() != null) {
                        existingProduct.setIsAvailable(product.isIsAvailable());
                    }

                    return existingProduct;
                }
            )
            .flatMap(productRepository::save);

        return result
            .switchIfEmpty(Mono.error(new ResponseStatusException(HttpStatus.NOT_FOUND)))
            .map(
                res ->
                    ResponseEntity
                        .ok()
                        .headers(HeaderUtil.createEntityUpdateAlert(applicationName, false, ENTITY_NAME, res.getId()))
                        .body(res)
            );
    }
```

**With service interface and impl, with reactive streams:**

Resource:
```
    /**
     * {@code PATCH  /products} : Updates given fields of an existing product.
     *
     * @param product the product to update.
     * @return the {@link ResponseEntity} with status {@code 200 (OK)} and with body the updated product,
     * or with status {@code 400 (Bad Request)} if the product is not valid,
     * or with status {@code 404 (Not Found)} if the product is not found,
     * or with status {@code 500 (Internal Server Error)} if the product couldn't be updated.
     * @throws URISyntaxException if the Location URI syntax is incorrect.
     */
    @PatchMapping(value = "/products", consumes = "application/merge-patch+json")
    public Mono<ResponseEntity<Product>> partialUpdateProduct(@NotNull @RequestBody Product product) throws URISyntaxException {
        log.debug("REST request to update Product partially : {}", product);
        if (product.getId() == null) {
        }

        Mono<Product> result = productService.partialUpdate(product);

        return result
            .switchIfEmpty(Mono.error(new ResponseStatusException(HttpStatus.NOT_FOUND)))
            .map(
                res ->
                    ResponseEntity
                        .ok()
                        .headers(HeaderUtil.createEntityUpdateAlert(applicationName, false, ENTITY_NAME, res.getId()))
                        .body(res)
            );
    }
```

Service interface
```
    /**
     * Partially updates a product.
     *
     * @param product the entity to update partially.
     * @return the persisted entity.
     */
    Mono<Product> partialUpdate(Product product);
```

Service impl:
```
    @Override
    public Mono<Product> partialUpdate(Product product) {
        log.debug("Request to partially update Product : {}", product);

        return productRepository
            .findById(product.getId())
            .map(
                existingProduct -> {
                    if (product.getName() != null) {
                        existingProduct.setName(product.getName());
                    }
                    if (product.getPrice() != null) {
                        existingProduct.setPrice(product.getPrice());
                    }
                    if (product.isIsAvailable() != null) {
                        existingProduct.setIsAvailable(product.isIsAvailable());
                    }

                    return existingProduct;
                }
            )
            .flatMap(productRepository::save);
    }
```

**No reactive streams, only Resource**:
```
    /**
     * {@code PATCH  /products} : Updates given fields of an existing product.
     *
     * @param product the product to update.
     * @return the {@link ResponseEntity} with status {@code 200 (OK)} and with body the updated product,
     * or with status {@code 400 (Bad Request)} if the product is not valid,
     * or with status {@code 404 (Not Found)} if the product is not found,
     * or with status {@code 500 (Internal Server Error)} if the product couldn't be updated.
     * @throws URISyntaxException if the Location URI syntax is incorrect.
     */
    @PatchMapping(value = "/products", consumes = "application/merge-patch+json")
    public ResponseEntity<Product> partialUpdateProduct(@NotNull @RequestBody Product product) throws URISyntaxException {
        log.debug("REST request to update Product partially : {}", product);
        if (product.getId() == null) {
        }

        Optional<Product> result = productRepository
            .findById(product.getId())
            .map(
                existingProduct -> {
                    if (product.getName() != null) {
                        existingProduct.setName(product.getName());
                    }
                    if (product.getPrice() != null) {
                        existingProduct.setPrice(product.getPrice());
                    }
                    if (product.isIsAvailable() != null) {
                        existingProduct.setIsAvailable(product.isIsAvailable());
                    }

                    return existingProduct;
                }
            )
            .map(productRepository::save);

        return ResponseUtil.wrapOrNotFound(
            result,
            HeaderUtil.createEntityUpdateAlert(applicationName, false, ENTITY_NAME, product.getId().toString())
        );
    }
```

qlereboursBS:
@cbornet the PR is almost ready but we've got an issue:
The tests are not passing when the entity contains an image field (bytes).
We digged to understand why and we found what's wrong:
In `Entity.java.ejs`, there's a condition on line 255, that says that if the field is an image, we'll add an `@Column("theFieldContentType")` => That creates a new field to store the content type.

In the PUT endpoint, there's no problem because we're calling the `save` method of the repository that save all the fields.
But in our case (PATCH) we need to update all the fields one by one: `if(field != null) { entityToUpdate.setField(field) }`.
**The problem here** is that we're iterating on the `fields` variable that **doesn't contain** the `fieldContentType` field, that is added on the fly. You can find the PATCH generation code in `patch_template.ejs`.

One of the solutions to fix this would be adding the same condition (as Entity.java.ejs, line 255) when generating the Patch code (but honestly, I find that this would be a bit dirty).

What do you think about this?

qlereboursBS:
And maybe @deepu105 could have an idea (concerning the comment above) because I saw your name when annotating the code.
Is there a way to iterate on the fields including the fields generated on the fly (like `fieldContentType`)?

deepu105:
I don't think there is any better way than whats done already, so I would suggest doing the same, may be you can extract the check to a function and reuse. Also we are doing ejs in templates so dirty solutions are like the norm 😂

cbornet:
You could probably check the type of the field (field type I think) and add the fieldContentType copy if it's of type Blob.

cbornet:
@qlereboursBS something bad happened with the branch...

qlereboursBS:
Wow, sorry, I will check this today

Yoshio123 (approved):
Thanks helpful patch's.
May be multi bytes characters applications severely.
Heavy stalking came.
Career profile sending url fake.
May be cloud front mobile net with 2fa interfere system.
I pull out SIM .
Because, iOS 14 update after hidden camera FaceTime mail AppStore.
SIM pull out after useful camera.
I'm surprised and regret.
Repair couldn't be .

generators/entity-server/templates/src/main/java/package/common/utils/is_field_with_content_type.ejs
```diff
@@ -0,0 +1,3 @@
+<% isFieldWithContentType = function(field) {
```
mshima, line 1: You should put this function at generator-base.

mshima (changes requested):

mshima (changes requested):
Remove `jdl/bundling` changes

generators/entity-server/templates/src/main/java/package/common/utils/is_field_with_content_type.ejs
qlereboursBS: Thanks, I didn't know where to put it and I wanted to ask before asking for review! :+1: 
I just pushed the changes

qlereboursBS: I'll squash the commits tomorrow

mshima (approved):

generators/entity-server/templates/src/main/java/package/common/patch_template.ejs
```diff
@@ -0,0 +1,56 @@
...
+<%_ if(viaService) { %>
+   <%_ if(!reactive) { %>
+   Optional<<%= instanceType %>> result = Optional.ofNullable(<%= entityInstance %>Service.partialUpdate(<%= instanceName %>));
+   <%_ } else { %>
+   Mono<<%= instanceType %>> result = <%= entityInstance %>Service.partialUpdate(<%= instanceName %>);
+   <%_ } %>
+<%_ } else { %>
+<%- returnPrefix %> <%= entityInstance %>Repository.findById(<%= instanceName %>.getId())
+   .map(existing<%= entityClass %> -> {
+<%_ for (field of fields) { _%>
```
mshima, line 36: Since we are doing patch updates, maybe should randomize fields to include:
```
const includeField = faker.random.boolean()
if (includeField) { %> /* Randomized disable for patch test <%};
...
if (includeField) { %> */<%};
```

mshima: And is possible to add new fields, instead of just set the old value.
https://github.com/jhipster/generator-jhipster/blob/05e4ab9cec320c4e22260ae8a6d99f13aea4b315/generators/entity-client/templates/common/src/test/javascript/cypress/integration/entity/entity.spec.ts.ejs#L29
https://github.com/jhipster/generator-jhipster/blob/05e4ab9cec320c4e22260ae8a6d99f13aea4b315/generators/entity-client/templates/common/src/test/javascript/cypress/integration/entity/entity.spec.ts.ejs#L146

qlereboursBS: > Since we are doing patch updates, maybe should randomize fields to include:
> 
> ```
> const includeField = faker.random.boolean()
> if (includeField) { %> /* Randomized disable for patch test <%};
> ...
> if (includeField) { %> */<%};
> ```

Can you give me more details please? I don't really understand what you mean

Concerning your second comment, we didn't generate any view for the patch, I don't understand why you're quoting cypress tests, can you explain please?

mshima: Let's keep this way.

qlereboursBS:
The PR is ready for review :tada: 
@cbornet @mshima

generators/generator-base.js
```diff
@@ -2447,6 +2447,15 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
...
 
+    /**
+     * Allows to know if the given field needs an additional content type field to save it's content type.
+     * @param {object} field - the entity field
+     * @returns {boolean} true if it needs an additional field.
+     */
+    isFieldWithContentType(field) {
+        return (field.fieldType === 'byte[]' || field.fieldType === 'ByteBuffer') && field.fieldTypeBlobContent !== 'text';
+    }
+
```
mshima, line 2458: @qlereboursBS if we are using methods this should be the place for this method.
But I think we shouldn't add a method for this.
Put `fieldWithContentType` field after: https://github.com/jhipster/generator-jhipster/blob/3ded888de0ba40be807144620e078424e9b06db0/utils/field.js#L160

And use directly.

generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
```diff
@@ -1405,6 +1405,59 @@ class <%= entityClass %>ResourceIT <% if (databaseType === 'cassandra') { %>exte
...
+        <%= entityInstance %>Repository.<%= saveMethod %>(<%= asEntity(entityInstance) %>)<%= callBlock %>;
+
+        int databaseSizeBeforeUpdate = <%= entityInstance %>Repository.findAll()<%= callListBlock %>.size();
+
+        // Update the <%= entityInstance %> using partial update
+        <%= asEntity(entityClass) %> partialUpdated<%= asEntity(entityClass) %> = <%= entityInstance %>Repository.findById(<%= asEntity(entityInstance) %>.getId()).get() %>;
+        <%_ if (databaseType === 'sql') { _%>
+        // Disconnect from session so that the updates on updated<%= asEntity(entityClass) %> are not directly saved in db
+        em.detach(partialUpdated<%= asEntity(entityClass) %>);
+        <%_ } _%>
```
mshima, line 1423: Why don't create the instance directly instead of loading and detaching?
```
<%= asEntity(entityClass) %> partialUpdated<%= asEntity(entityClass) %> = new <%= asEntity(entityClass) %>();
partialUpdated<%= asEntity(entityClass) %>.setId(<%= asEntity(entityInstance) %>.getId());
```

mshima: And I am missing a test with a not existing id.

generators/generator-base.js
qlereboursBS: I totally agree, this is the best solution, thanks
I just pushed the fix for this one

generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs
qlereboursBS: @mshima the first reason why we did this is because it's done in the same way when updating the entity (Line 1319 in the same file).
If we dive into this, I think that the detach has been added to prevent getting "wrong working tests":
If you get an entity, then edit its fields, you editing the reference fields and you could get a test that is working just because the reference has been changed in the test, not because the Patch is working correctly.

I already encountered an issue like this:
 - I was getting an entity, lets say `User user1`
 - I was editing this entity: `user->setName("test")`
 - Then calling the repository to get all the entities with name `test`: `userRepository.findByName("test")`
 - And the entity I just got and edited **without saving** was returned from `findByName`

I can't be sure that it's the reason why it's been added to the update method, but to my mind, it would be safer to keep it. What do you think?

mshima: The update method is correct, we should load the entity so we don't miss and required field/relationship.
In this case I don't see the reason to load the entity, just create an empty instance.
PATCH should persist only changed fields.
When we load a persisted entity, edit it and save it, the workflow is identical to existing update method.
The old values will be passed to the patch method, so there is no guarantee if patch is loading and persisting not modified fields correctly.

IMO the correct approach would be to have 2 tests:
- Partial patch update
  - create an empty instance
  - set an updated value to half of the fields
  - execute patch.
  - load the updated entity
  - compare all field values, updated against updated value and non updated fields against old value.
- A complete patch update.
  - same as partial but with every field for completeness.

IMO the correct approach would be to use sql update with criteria api, but this would complicate non sql, and others use cases.

qlereboursBS: I understand, I'll develop the two different tests. We mentionned this when we were at JHipster code, but we forgot to do this :)

qlereboursBS: I'll also add a test with a missing id, to have a full coverage

mshima: I think the easier way is to extract your partialUpdate test to another template and use something like:
```
<% const fieldsToIncludeInPartialTest = fields.map(field => {
    const includeField = faker.random.boolean();
    const updateWithValue = includeField ? `UPDATED_${}` : 'null';
    const testWithConstant = includeField ? `UPDATED_${}` : `DEFAULT_${}`;
    return { includeField, updateWithValue, testWithConstant, field};
}); %>
<%- include('partial_it_test`, {fields: fieldsToIncludeInPartialTest}) -%>
```

qlereboursBS: Thank you very much for the advice, I just pushed the 3 tests.
I created the template in `tests/java/package/common`, is it ok for you? There was no folder to store templates for the tests.
Here is the commit link: https://github.com/jhipster/generator-jhipster/pull/12507/commits/c1927270545d7de526e5b75df0ba5f93b11c9b88

mshima: At first glance I would think `templates/src/tests/java/package/common` should go to `src/tests/java/package/common`.
I prefer:
- put in the same folder as the main file since it’s used only there
- put inside `templates/partials` and name it with a complete description like `it_patch_update.partial.java.ejs`.

qlereboursBS: I choosed the `templates/partials` solution: https://github.com/jhipster/generator-jhipster/pull/12507/commits/7f859d2cfa3358cea044683640e31f075ff917bc
Does it seems ok, or is there a bettery way than doing `../../......`?

mshima: > Does it seems ok, or is there a bettery way than doing `../../......`?

You can try adding.
```
    options = {
        root: generator.templatePath(),
        context: generator,
        ...options,
    };
```
to https://github.com/jhipster/generator-jhipster/blob/ad3e0fbc48a21369444595cf0c6d54d8151a28bb/generators/utils.js#L240

This should allow to include the file using absolute path (relative to templates path) `/partials/...ejs`
https://ejs.co/#docs

qlereboursBS: Awesome, I changed this and it works, thank you!

mshima: Great

mshima:
Thanks @qlereboursBS.

This PR doesn’t support updating relationships. That can be done later.

cbornet:
Congrats @qlereboursBS and Bear studio for this PR ! What an awesome outcome from the JHipster Code event !

qlereboursBS:
Thank you @cbornet and @mshima for helping us with this first PR :)

mshima:
@qlereboursBS do you want to implement remaining merge patch functionalities?

qlereboursBS:
@mshima I'm already assigned to an issue I'm working on, but I would be happy to develop the remaining functionalities when I will finish the other issue :)

mshima:
@qlereboursBS I needed this for 1 entity only, so I don't have plans to implement in jhipster right now.
So this is the implementation:

- Resource.java:
  - inject ConversionService.
```
    @PatchMapping(value = “/foos”, consumes = "application/merge-patch+json")
    public ResponseEntity<Foo> partialUpdateFoo(@NotNull @RequestBody HashMap<String, Object> foo) throws URISyntaxException {
        log.debug("REST request to update Foo partially : {}", foo);
        Object originalId = foo.remove("id");
        if (originalId == null) {
        }
        if (!conversionService.canConvert(originalId.getClass(), Long.class)) {
            throw new BadRequestAlertException("Invalid id", ENTITY_NAME, "idinvalid");
        }
        if (foo.isEmpty()) {
            throw new BadRequestAlertException("No changes requested", ENTITY_NAME, "nochanges");
        }

        Long id = conversionService.convert(originalId, Long.class);
        Optional<Foo> result = Optional.ofNullable(FooService.partialUpdate(id, foo));

        return ResponseUtil.wrapOrNotFound(
            result,
            HeaderUtil.createEntityUpdateAlert(applicationName, true, ENTITY_NAME, id.toString())
        );
    }
```

- Service.java
```
    /**
     * Partially udpates a Foo.
     *
     * @param Foo the entity to update partially.
     * @return the persisted entity.
     */
    public Foo partialUpdate(@NotNull Long id, @NotNull HashMap<String, Object> Foo) {
        log.debug("Request to partially update Foo : {}", foo);

        return FooRepository
            .findById(id)
            .map(
                existingFoo -> {
                    BeanWrapper wrapper = new PropertyAccessorFactory.forBeanPropertyAccess(existingFoo);
                    wrapper.setPropertyValues(foo);
                    return existingFoo;
                }
            )
            .map(FooRepository::validateAndSave)
            .get();
    }

```

- Repository:
```
@Validated
public interface FooRepository extends JpaRepository<Foo, Long> {
    @Validated
    default Foo validateAndSave(@Valid Foo foo) {
        return save(foo);
    }

```

Tests that should be implemented:
- Not valid properties.
- Request with no changes.

Tests that should be done:
- Others repository's methods should not be validated, otherwise we should implement a FooValidatedService exclusive to validate.

This should fix:
- Relationships partial update.
- Field removals.
- Beans with default values
```
class Foo {
    String name = "default";
}
```

Validation notes:
- Bean validation normally happens at Controller so we should validate in another step.
https://reflectoring.io/bean-validation-with-spring-boot/#validating-input-to-a-spring-service-method:
```
Note that we have to add Spring’s @Validated annotation to the controller at class level to tell Spring to evaluate the constraint annotations on method parameters.

The @Validated annotation is only evaluated on class level in this case
```
- https://stackoverflow.com/questions/19425221/spring-validated-in-service-layer

pascalgrimaud:
as suggested by @cbornet, this feature deserves a bounty.
Congrats and thanks for your work @qlereboursBS : plz claim the bounty

qlereboursBS:
I just submitted the expense: https://opencollective.com/generator-jhipster/expenses/25631

(I will have a look at the comment this evening or later)

pascalgrimaud:
@qlereboursBS : approved

Issue #12450, opened 2020-09-06
Vue: toast instead of alert custom


pascalgrimaud:
##### **Overview of the issue**

In a Vue.js application, the alert is managed by alert service and is custom.
We should use Toast instead, like it's done in JHipster Control Center

##### **Motivation for or Use Case**

Simpler and better UI

##### **Reproduce the error**

- generate a Vue.js application
- start the app
- login in
- go to user management
- edit an user

##### **Suggest a Fix**

- remove the alert.service.ts and alert.service.spec.ts
- replace by the toast (see an example at https://github.com/jhipster/jhipster-control-center/blob/master/src/main/webapp/app/applications/instance/instance.component.ts#L96-L101)
- the place where to change is:
    - user-management
    - entity

##### **JHipster Version(s)**

Current master

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "blueprints": [],
    "otherModules": [],
    "applicationType": "monolith",
    "reactive": false,
    "baseName": "jhipster",
    "jhipsterVersion": "6.10.1",
    "skipClient": false,
    "skipServer": false,
    "skipUserManagement": false,
    "skipCheckLengthOfIdentifier": false,
    "skipFakeData": false,
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "clientPackageManager": "npm",
    "testFrameworks": ["cypress"],
    "pages": [],
    "creationTimestamp": 1599409665813,
    "promptValues": {
      "packageName": "io.github.pascalgrimaud",
      "nativeLanguage": "en"
    },
    "serviceDiscoveryType": false,
    "authenticationType": "jwt",
    "packageName": "io.github.pascalgrimaud",
    "serverPort": "8080",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "mysql",
    "buildTool": "maven",
    "serverSideOptions": [],
    "websocket": false,
    "searchEngine": false,
    "messageBroker": false,
    "enableSwaggerCodegen": false,
    "clientFramework": "vue",
    "clientTheme": "cerulean",
    "clientThemeVariant": "primary",
    "enableTranslation": true,
    "nativeLanguage": "en",
[...]

PR #12508, opened 2020-09-14
Vue: use toast instead of alert custom

fdaudies:
Fix https://github.com/jhipster/generator-jhipster/issues/12450


pascalgrimaud:
thanks and congrats @fdaudies for your 1st contribution

Issue #12467, opened 2020-09-09
Sonar: Replace this "String" wrapper object with primitive type "string"


qmonmert:
##### **Overview of the issue**

Sonar issue in duration.pipe.ts.ejs

##### **Motivation for or Use Case**

Minor Sonar issue
 
##### **Reproduce the error**

- Generate a project with Angular like framework.
- Start Sonar `docker-compose -f src/main/docker/sonar.yml up -d`
- Launch analysis `./mvnw -Pprod clean verify sonar:sonar`

![image](https://user-images.githubusercontent.com/9989211/92623664-d9381880-f2c6-11ea-9524-c85bd0c5547b.png)

##### **Suggest a Fix**

Replace String by string in duration.pipe.ts.ejs

##### **JHipster Version(s)**

Current version

##### **Browsers and Operating System**

-   [x] Checking this box is mandatory (this is just to show you read everything)

ThomasCarca:
@pascalgrimaud I'll handle this one :)

PR #12509, opened 2020-09-14
fix useless usage of String wrapper in duration.pipe.ts, now using pr…

ThomasCarca:
Fix [#12467](https://github.com/jhipster/generator-jhipster/issues/12467)

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/shared/duration.pipe.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/shared/duration.pipe.ts.ejs
@@ -6,7 +6,7 @@ import * as moment from 'moment';
   name: 'duration',
 })
 export class DurationPipe implements PipeTransform {
-  transform(value: any): String {
+  transform(value: any): string {
     if (value) {
         return moment.duration(value).humanize();
     }


generators/client/templates/angular/src/main/webapp/app/shared/duration.pipe.ts.ejs
```diff
@@ -6,10 +6,7 @@ import * as moment from 'moment';
   name: 'duration',
 })
 export class DurationPipe implements PipeTransform {
-  transform(value: any): String {
-    if (value) {
-        return moment.duration(value).humanize();
-    }
-    return "";
+  transform(value: any): string {
+    return moment.duration(value).humanize();
```
qmonmert, line 10: why do you remove the `if (value) {` ?

ThomasCarca: Since it's a primitive type, it can never be null, hence no need to check this anymore !

qmonmert: `value` is not a primitive type

ThomasCarca: You are absolutely right, my bad :) Will revert

ThomasCarca: fixed

qmonmert (approved):

ThomasCarca:
@pascalgrimaud any idea what could have caused the pipeline to fail ?

qmonmert:
@pascalgrimaud you must close and reopen the PR (I had the same pb on my PR this morning) 
cc @ThomasCarca

pascalgrimaud:
no worry, I just restart the build, no need to close/reopen

qmonmert:
@pascalgrimaud I said that because me I can't just restart a build

PR #12510, opened 2020-09-14
Fix name convention

### LINKED ISSUE #12498: Name Convention Getter for "Boolean" and "boolean" is incorrect

In "User.java",

https://github.com/jhipster/generator-jhipster/blob/a91c744a3d093cb738825b369ab912325104c7f9/generators/server/templates/src/main/java/package/domain/User.java.ejs#L438

 following conventions the correct one should be 

```java
public boolean isActivated() {
```

In "LoginVM.java",
https://github.com/jhipster/generator-jhipster/blob/86a0ecdeca071ba4098d3a115f548841a821ab9e/generators/server/templates/src/main/java/package/web/rest/vm/LoginVM.java.ejs#L55

 following conventions the correct one should be 

```java
public Boolean getRememberMe() {
```
or change "Boolean" to primitive "boolean"

```java
public boolean isRememberMe() {
```

##### **Suggest a Fix**

- Change "Boolean" to "boolean"
or
- Follow convenction in method name

##### **JHipster Version(s)**

master

MarlonLuan:
Fixes #12498

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
@@ -431,7 +431,7 @@ public class <%= asEntity(entityClass) %> implements Serializable {
         const fieldInJavaBeanMethod = fields[idx].fieldInJavaBeanMethod; _%>
 
     <%_ if (fieldTypeBlobContent !== 'text') { _%>
-    public <%= fieldType %> <% if (fieldType.toLowerCase() === 'boolean') { %>is<% } else { %>get<%_ } _%><%= fieldInJavaBeanMethod %>() {
+    public <%= fieldType %> get<%= fieldInJavaBeanMethod %>() {
     <%_ } else { _%>
     public String get<%= fieldInJavaBeanMethod %>() {
     <%_ } _%>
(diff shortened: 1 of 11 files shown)


generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
```diff
@@ -431,7 +431,7 @@ public class <%= asEntity(entityClass) %> implements Serializable {
         const fieldInJavaBeanMethod = fields[idx].fieldInJavaBeanMethod; _%>
 
     <%_ if (fieldTypeBlobContent !== 'text') { _%>
-    public <%= fieldType %> <% if (fieldType.toLowerCase() === 'boolean') { %>is<% } else { %>get<%_ } _%><%= fieldInJavaBeanMethod %>() {
+    public <%= fieldType %> get<%= fieldInJavaBeanMethod %>() {
```
mshima, line 434: Instead of dropping the condition, remove `toLowerCase()`.
So if fieldType is primitive boolean, we will generate with `is`.

mshima (changes requested):

generators/entity-server/templates/src/main/java/package/domain/Entity.java.ejs
MarlonLuan: According to the current generator, it will never be primitive, even set "required"  still is "Boolean" with annotation "@NotNull"

mshima: Yep, that’s right.
Boolean primitive is useful due to checkbox without a default value would require the user to click twice to get false value.
I wanted to keep the template even if not used, blueprints could use this, but seems too much.

mshima (approved):

Issue #12465, opened 2020-09-06
Faker data: not correct for duration


pascalgrimaud:
##### **Overview of the issue**

When generating project with entities having duration, the faker data is not correct : so nothing is displayed

##### **Motivation for or Use Case**

The faker data should be correct for duration

##### **Reproduce the error**

- generate project
- generate entity with duration
- start the app
- go to the entity
- nothing is displayed for duration column

##### **Browsers and Operating System**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

ThomasCarca:
@pascalgrimaud I will take this one

mraible:
@ThomasCarca Any progress to report?

DanielFran:
Can someone confirm this has been fixed with #12427 and #12516 can be closed too?

PR #12516, opened 2020-09-14
fix bad parsing of duration for Vue.js projects

ThomasCarca:
Fix [#12465](https://github.com/jhipster/generator-jhipster/issues/12465)

### DIFF
--- a/generators/client/templates/vue/src/main/webapp/app/shared/date/filters.ts.ejs
+++ b/generators/client/templates/vue/src/main/webapp/app/shared/date/filters.ts.ejs
@@ -1,6 +1,6 @@
 import Vue from 'vue';
-import { format, parseISO, formatDuration } from 'date-fns';
-import * as duration from 'duration-fns';
+import { format, parseISO } from 'date-fns';
+import * as moment from 'moment';
 
 export const DATE_FORMAT = 'yyyy-MM-dd';
 export const DATE_TIME_FORMAT = 'yyyy-MM-dd HH:mm';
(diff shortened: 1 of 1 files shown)


ThomasCarca:
I was inspired by [this file](https://github.com/jhipster/generator-jhipster/blob/master/generators/client/templates/angular/src/main/webapp/app/shared/duration.pipe.ts.ejs) which works fine for Angular.

I just replicated the same logic to the util that parse duration for Vue; There should be, however, a way to mutualize this kind of code so we parse Durations the same way between frameworks

generators/client/templates/vue/src/main/webapp/app/shared/date/filters.ts.ejs
```diff
@@ -1,6 +1,6 @@
 import Vue from 'vue';
-import { format, parseISO, formatDuration } from 'date-fns';
-import * as duration from 'duration-fns';
+import { format, parseISO } from 'date-fns';
+import * as moment from 'moment';
```
MathieuAA, line 3: Why do we use both libs? One lib isn't enough?

MathieuAA: I'm creating an issue for that, it's not related to the PR

ThomasCarca: We previously used only date-fns in this util but to be inline with what is done for Angular, we can use moment (see my comment on the PR)

I think date-fns could be entirely replaced by moment but did not wanted to go through this process as it is not in the scope of this PR

MathieuAA: Thanks for the explanation @ThomasCarca :)

pascalgrimaud:
Yes, moment should not be used. I'm not sure we have the dependency either

ThomasCarca:
> Yes, moment should not be used. I'm not sure we have the dependency either

It is used for Angular. Should we use different time libraries between different frameworks ?

atomfrede:
We have not used moment for vue to have a very small bundle size. Of course as it is now part of core we could revisited the decision but I would not integrate moment.

deepu105:
Yes, we use moment very littel, ideally we should migrate to a smaller lib if possible for all frameworks

DanielFran:
Superseded by #12727

PR #12526, opened 2020-09-15
Test commands improvements

MathieuAA:
Removed failing and now useless test (about the JDL bundle stuff)

Implemented test commands:
  - npm test jdl => launches the jdl-related tests, in parallel
  - npm test jdl watch => the same, but with the watch mode on

One can replace jdl by whatever folder there is inside the test folder.
Not providing the test folder launches all the tests (cli in sequential mode
and noncli in parallel mode).
Passing cli as test folder doesn't launch them in parallel mode.


.mocharc.js
```diff
@@ -1,8 +1,10 @@
 module.exports = {
+    extension: 'js',
+    forbidOnly: true,
```
mshima, line 3: This will forbid `mocha` with .only.

.mocharc.js
```diff
@@ -1,8 +1,10 @@
...
+    noInsight: true,
```
mshima, line 4: Does not work.

package.json
```diff
@@ -47,13 +47,10 @@
         "eslint": "eslint .",
         "ejs-lint": "mocha test/ejslint.js",
         "ejslint": "ejslint",
-        "pretest": "npm run lint",
```
pascalgrimaud, line 50: is it launch somewhere ? Or do we lost the lint ?

mshima (changes requested):
We should fix cli test using parallel not doing some crazy workarounds.

package.json
MathieuAA: My bad, I forgot to add it back

.mocharc.js
MathieuAA: this was already the case, but we passed the options directly

.mocharc.js
MathieuAA: Interesting, thanks! I'll look into why it doesn't and put it back like before

.mocharc.js
mshima: exactly, `mocha` != `npm test`.

.mocharc.js
mshima: Just ask.
https://github.com/jhipster/generator-jhipster/blob/6b3cbfa43e5ee8a4faacc9fc624b2d4836c8c1a5/generators/statistics.js#L47

MathieuAA: awesome :) thanks

.mocharc.js
MathieuAA: fixed

scripts/launch-tests.js
```diff
@@ -0,0 +1,34 @@
+/**
```
mshima, line 1: This script is useless if we fix test/cli in parallel mode.
Much easier to learn:
```
mocha test/dir -p --watch
```
then
```
scripts/launch-tests.js dir --watch
```

MathieuAA: I disagree with you: I don't want to watch my test files all the time, and worse: not all the tests should be run in parallel.
The part where I agree is that the tests should be improved so that they can be run in parallel.

mshima: I really don't we should create a wrapper for this.
This will break
```
npm test -- test/app*
```
And there is very little to no improvements to test script.

MathieuAA: This won't break it: it's still possible by doing `npm test app`.
I still agree: we should **definitely** improve the CLI tests. If we can't run them in parallel, then we won't and my solution stands.

PR #12538, opened 2020-09-16
fix(Server Configuration): Rectify content type for serving JSON files

AlexandreCassagne:
Previously, static ".json" files had "text/html" [+ charset] as a MIME type.

This change ensures that JSON files are properly served using the standard "application/json" [+ charset] MIME type.

This change was briefly discussed with @jdubois in [this CloudFoundry issue](https://github.com/cloudfoundry/gorouter/issues/64#issuecomment-693584848).

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/WebConfigurer.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/WebConfigurer.java.ejs
@@ -134,21 +134,8 @@ public class WebConfigurer implements <% if (!reactive) { %>ServletContextInitia
      */
     @Override
     public void customize(WebServerFactory server) {
-        setMimeMappings(server);
         // When running in an IDE or with <% if (buildTool === 'gradle') { %>./gradlew bootRun<% } else { %>./mvnw spring-boot:run<% } %>, set location of the static web assets.
         setLocationForStaticAssets(server);
-    }
-
-    private void setMimeMappings(WebServerFactory server) {
-        if (server instanceof ConfigurableServletWebServerFactory) {
-            MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
-            // IE issue, see https://github.com/jhipster/generator-jhipster/pull/711
-            mappings.add("html", MediaType.TEXT_HTML_VALUE + ";charset=" + StandardCharsets.UTF_8.name().toLowerCase());
-            // CloudFoundry issue, see https://github.com/cloudfoundry/gorouter/issues/64
-            mappings.add("json", MediaType.TEXT_HTML_VALUE + ";charset=" + StandardCharsets.UTF_8.name().toLowerCase());
-            ConfigurableServletWebServerFactory servletWebServer = (ConfigurableServletWebServerFactory) server;
-            servletWebServer.setMimeMappings(mappings);
-        }
     }
 
     private void setLocationForStaticAssets(WebServerFactory server) {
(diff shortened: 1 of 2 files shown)


generators/server/templates/src/main/java/package/config/WebConfigurer.java.ejs
```diff
@@ -145,7 +145,7 @@ public class WebConfigurer implements <% if (!reactive) { %>ServletContextInitia
             // IE issue, see https://github.com/jhipster/generator-jhipster/pull/711
             mappings.add("html", MediaType.TEXT_HTML_VALUE + ";charset=" + StandardCharsets.UTF_8.name().toLowerCase());
             // CloudFoundry issue, see https://github.com/cloudfoundry/gorouter/issues/64
```
ruddell, line 147: This comment can be removed since it no longer applies

AlexandreCassagne: Yes apologies. Should I create a second commit or amend the previous one?

AlexandreCassagne: My first contribution. Starting small was a good idea. haha

murdos: @AlexandreCassagne : just amend the previous one ;)
And congrats for your first contribution !

AlexandreCassagne: @ruddell I amended my previous pull request.

AlexandreCassagne: @murdos  thank you ;-)

AlexandreCassagne:
Regarding the 

-  Documentation is added/updated where necessary

I don't think this change requires changing the documentation, because I haven't found any mention of the JSON → "text/html" MIME type mapping;

-  All continuous integration tests are green

There is at least one failing test, although it seems that the failure comes from upstream.

murdos (approved):

jdubois:
@AlexandreCassagne have you tested with totally removing the line? I guess that should be the standard mime type, we were just redefining it because of a bug, but isn't this "application/json" by default?
BTW we should do the same for the line above, IE is dead now.

AlexandreCassagne:
@jdubois  I'll run a test today in my development environment and amend my pull request if it works.

AlexandreCassagne:
@jdubois I ran the WebConfigurerTest cases and found the following error when I remove the two lines:

```Expecting:
 <"text/html">
to be equal to:
 <"text/html;charset=utf-8">
but was not.
org.opentest4j.AssertionFailedError: 
Expecting:
 <"text/html">
to be equal to:
 <"text/html;charset=utf-8">```

What is the importance of `charset=utf-8`?

AlexandreCassagne:
I'd suggest the following:

- For HTML: W3 [recommends keeping the charset in the content-type](https://www.w3.org/International/articles/http-charset/index), at least for text/html. 
- For JSON: [RFC-7159, section 11](https://tools.ietf.org/html/rfc7159#section-11), "_no 'charset' parameter is defined for this registration. Adding one really has no effect on compliant recipients_". If I understand this one well, the charset should probably not be set at all.

mshima:
Maybe we should remove custom config and let default.

jdubois:
If tests say it's good, then I'm trusting the Spring Boot team now has good default parameters, and so we would be able to remove this. It would be cool!

AlexandreCassagne:
I’m ok to make the change right now, just be aware that it would remove the “charset” parameter of the content-type.

Is everyone ok with that?

jdubois:
Yes @AlexandreCassagne please do! Let's not pollute all requests with useless headers

ruddell (approved):

AlexandreCassagne:
Alright! assuming all the checks pass, this amended commit removes the whole call to `setMimeMappings(...)` which customised the JSON and HTML content types.

The new tests expect:

- `application/json` for JSON content;
- `text/html` for HTML content.

PR #12551, opened 2020-09-18
[jdl] Export otherEntityField for every relationship.

mshima:
---

### DIFF
--- a/test/jdl/jdl-importer.spec.js
+++ b/test/jdl/jdl-importer.spec.js
@@ -57,6 +57,7 @@ describe('JDLImporter', () => {
                     ],
                     relationships: [
                         {
+                            otherEntityField: 'region',
                             relationshipType: 'one-to-many',
                             relationshipName: 'area',
                             otherEntityName: 'region',
@@ -120,6 +121,7 @@ describe('JDLImporter', () => {
                         {
                             relationshipType: 'one-to-many',
                             javadoc: 'A relationship',
+                            otherEntityField: 'id',
                             relationshipName: 'employee',
                             otherEntityName: 'employee',
                             otherEntityRelationshipName: 'department',
@@ -181,6 +183,7 @@ describe('JDLImporter', () => {
                     ],
                     relationships: [
                         {
+                            otherEntityField: 'id',
                             relationshipType: 'one-to-many',
                             relationshipName: 'job',
                             otherEntityName: 'job',
@@ -375,6 +378,7 @@ describe('JDLImporter', () => {
                     ],
                     relationships: [
                         {
+                            otherEntityField: 'id',
                             relationshipType: 'one-to-many',
                             relationshipName: 'country',
                             otherEntityName: 'country',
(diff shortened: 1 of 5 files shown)


mshima:
> Why do we need it? Except for consistency.

Consistency should be enough 😄, an error should be thrown instead of ignoring those informations, otherwise.
Actually I am doing some changes that allows those relationships to be shown, but the information is just lost.

MathieuAA:
Okay! I didn't know if there was an other reason, but I agree: consistency **is** enough.

test/jdl/jdl-importer.spec.js
```diff
@@ -57,6 +57,7 @@ describe('JDLImporter', () => {
                     ],
                     relationships: [
                         {
+                            otherEntityField: 'region',
```
MathieuAA, line 60: Shouldn't it be `id`?

MathieuAA (commented):
It's been a long time since I've had the displeasure of playing with relationships (and I'm putting it mildly ;)), but shouldn't `setRelationshipsToEntity` also be changed?

test/jdl/jdl-importer.spec.js
mshima: https://github.com/jhipster/generator-jhipster/blob/da8e8bae66c9a94260305b8233129274032c6912/test/jdl/test-files/big_sample.jdl#L85
It’s correct isn’t it?

MathieuAA: `region` (the inner one) should be the displayed field. There are two possibilities here: the JDL is wrong or the `otherEntityField` is weirder than I though. Or both.
It shouldn't have been `id` but `area`, right?

MathieuAA: Tell me If I'm getting confused :laughing: I hate relationships

mshima: It’s correct, `region` is the displayed name which is `otherEntityField`.
And `area` is the `relationshipName`.

MathieuAA: Ah! Thanks :)

MathieuAA (approved):

mshima:
> It's been a long time since I've had the displeasure of playing with relationships (and I'm putting it mildly ;)), but shouldn't `setRelationshipsToEntity` also be changed?

I didn’t realized `many-to-one` relationships added the other side, seems not necessary to me.
IMO it could be omitted, or I we should add `primary:true` to the main relationship so `relationship OneToMany{foo bar}` !== `relationship ManyToOne{bar foo}`
What do you think?
I will take a look at others cases.

mshima:
> > It's been a long time since I've had the displeasure of playing with relationships (and I'm putting it mildly ;)), but shouldn't `setRelationshipsToEntity` also be changed?
> 
> I didn’t realized `many-to-one` relationships added the other side, seems not necessary to me.
> IMO it could be omitted, or I we should add `primary:true` to the main relationship so `relationship OneToMany{foo bar}` !== `relationship ManyToOne{bar foo}`
> What do you think?
> I will take a look at others cases.

Never mind, this only happens when the other side is specified.

PR #12569, opened 2020-09-21
Add annotation to detect generated files

### LINKED ISSUE #12459: Annotation to detect generated files

There is no way to know if a java class has been generated by JHipster or not. It will help real applications for development

Add an annotation like `@Generated` to all generated classes from JHipster

##### **Suggest a Fix**

```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.METHOD, ElementType.TYPE, ElementType.CONSTRUCTOR })
public @interface Generated {}
```

##### **JHipster Version(s)**

Current master

jsone-studios:
Add a `@Generated("JHipster")` annotation from `jakarta.annotation:jakarta.annotation-api` to all generated
- [x] Java classes under `src/main/java`
- [ ] Java classes under `src/test/java`

I am a bit unsure if we want to use `@Generated("JHipster")` or `@Generated("generator-jhipster")`. Any feedback would be very welcome. @pascalgrimaud 

#### Update:
A custom `GeneratedByJHipster` annotation is now added with a transform script to each java class/interface.

---

Fixes #12459

---


pascalgrimaud:
Thanks @jsone-studios for this 1st draft !

About the `@Generated` annotation, I think it's better to use a custom one than the one from javax.annotation, as mentioned in the issue:

```java
package io.github.jhipster;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.METHOD, ElementType.TYPE })
public @interface Generated {
}
```

Because with this, the user can customize it easily. But it's my personal opinion. It can be discussed of course.

pascalgrimaud:
@jsone-studios : don't make the change immediatly, let's wait some feedbacks first :)

mshima:
What about
```java
package io.github.jhipster;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

import javax.annotation.Generated;

@Generated(string="JHipster", date="<%= creationTimestamp %>", comments="Generated by JHipster <%= jhipsterVersion %>");
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.TYPE })
public @interface GeneratedByJHipster {
}
```

pascalgrimaud:
I like it @mshima

vishal423:
I think you can change the retention policy to `@Retention(RetentionPolicy.SOURCE)`. Also, instead of modifying each file to add the annotation, can we have some scripts to apply like used for prettier transformation?

mshima:
> I think you can change the retention policy to `@Retention(RetentionPolicy.SOURCE)`. Also, instead of modifying each file to add the annotation, can we have some scripts to apply like used for prettier transformation?

That's possible, something like:
```
 registerGeneratedByJHipsterTransform() {
        this.registerTransformStream(
            through.obj(function (file, enc, cb) {
                if (path.extname(file.path) === '.java' && file.state) {
                    file.contents = Buffer.from(file.contents.toString('utf8').replace(/\n((\w*) class)/, '@io.github.jhipster.GeneratedByJHipster\n$1', 'utf8');
                }
                this.push(file);
                cb();
            })
        );
    }
```

pascalgrimaud:
Don't forget the java import

jsone-studios:
Thanks for your intensive feedback. I really like the concept of a transformation script. I am going to refactor my PR to use such a transformation script.

gmarziou:
Can someone clarify what is the use of this annotation? Is this for documentation purpose (for humans) or for tools (which ones)?

pascalgrimaud:
@gmarziou :
- it was asked long time ago by several people, to know which files are generated by JHipster or not
- it was suggested to add a "comment" at begining of the project, but nothing has been implemented yet
- then, the idea of an annotation comes from the different projects in my company
- you can have an example here (from DDD ticket): 
    - https://gitlab.com/cdamon/padbowl/-/blob/master/src/main/java/com/jhipster/padbowl/common/infrastructure/Generated.java
    - see the commit `Mark generated classes as so (ignore coverage)` https://gitlab.com/cdamon/padbowl/-/commit/48f2d0be2752a6f7af1dd922d35917388a8edd8b
    - with this, you can completely focus on the domain and be sure to have 100% coverage on your business code

Then my first suggestion was to create a flag, something like `jhipster --generated-flag`, for people who don't want this, instead of having by default. That's why it should be in discussion.

mraible:
I'm not against the idea. However, since I'm not a tools vendor, how would I benefit from it? It seems like noise.

pascalgrimaud:
@mraible :
- as said in previous comment, it's mainly for knowing which files are generated by JHipster or not
- in real project, for real customer, after months of development, when there are new developers in team, we advise them to avoid modifying JHipster generated code
- how to know it ? they can't, unless they know JHipster, or use git log, which is annoying
- using this kind of annotation will help

jsone-studios:
I have pushed a new draft with a custom `GeneratedByJHipster` annotation, that is added with a transform script to every Java class and interface.

mshima:
I think it's possible add:
```
public @interface GeneratedByJHipster {
    /**
     * Set true to make JHipster don't override this file
     * {@code @GeneratedByJHipster(true)}
     * 
     * @return
     */
    boolean value() default false;
}
```

generators/app/index.js
```diff
@@ -238,6 +238,7 @@ module.exports = class JHipsterAppGenerator extends BaseBlueprintGenerator {
 
         useBlueprints = !this.fromBlueprint && this.instantiateBlueprints('app');
 
+        this.registerGeneratedAnnotationTransform();
```
mshima, line 241: Following https://github.com/jhipster/generator-jhipster/pull/12569#issuecomment-697232896
```suggestion
        if (this.options.skipGeneratedByJhipster) {
            this.registerGeneratedAnnotationTransform();
        }
```
And add to `this.option('skip-generated-by-jhipster'...`

generators/generator-base-private.js
```diff
@@ -1469,6 +1469,16 @@ module.exports = class JHipsterBasePrivateGenerator extends Generator {
         generator.registerTransformStream([prettierFilter, prettierTransform(prettierOptions), prettierFilter.restore]);
     }
 
+    registerGeneratedAnnotationTransform() {
+        if (this.options.help) {
+            return;
+        }
+
```
mshima, line 1476: Can be dropped.
```suggestion
```

generators/generator-base-private.js
```diff
@@ -1469,6 +1469,16 @@ module.exports = class JHipsterBasePrivateGenerator extends Generator {
...
+        const packageName = this.jhipsterConfig.packageName;
+
+        this.registerTransformStream(generatedAnnotationTransform(packageName));
```
mshima, line 1479:
```suggestion
        this.registerTransformStream(generatedAnnotationTransform(this.jhipsterConfig.packageName));
```

generators/generator-transforms.js
```diff
@@ -50,7 +51,25 @@ const prettierTransform = function (defaultOptions) {
     });
 };
 
+const generatedAnnotationTransform = packageName => {
+    return through.obj(function (file, encoding, callback) {
+        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
+            const content = file.contents.toString('utf-8');
```
mshima, line 57: Not sure but I think encoding should be `utf8`.

jsone-studios:
I have applied the suggestions from  @mshima and moved the annotation generation behind an option. I have also added support for annotating all files from the entity, sping-controller and spring-service sub-generator.

mshima:
> I have applied the suggestions from @mshima and moved the annotation generation behind an option. I have also added support for annotating all files from the `entity`, `sping-controller` and `spring-service` sub-generator.

You can put the option in generator-base, so every derived generator will have this option.
No need to add one by one.

In my opinion it should be opt-out instead of opt-in.

And save the option for regeneration. Like:
```
if (this.options.skip !== undefined) {
  this.jhipsterConfig.skip = ...
}
If (this.jhipsterConfig.foog.skip) {
  this.register...
}
```

mshima:
I think opt out could be implemented by registering `no-` option.
And applying if the option is undefined.

generators/app/index.js
```diff
@@ -238,6 +238,9 @@ module.exports = class JHipsterAppGenerator extends BaseBlueprintGenerator {
 
         useBlueprints = !this.fromBlueprint && this.instantiateBlueprints('app');
 
+        if (!this.jhipsterConfig.skipGeneratedFlag) {
+            this.registerGeneratedAnnotationTransform();
+        }
```
mshima, line 243: Move to generator-base too, so it will be registered for every generator.

generators/generator-base.js
```diff
@@ -112,6 +112,14 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
             this.config.set(this.options.localConfig);
         }
 
+        this.option('skip-generated-flag', {
+            desc: 'Skip adding a GeneratedByJhipster annotation to all generated java classes and interfaces',
+            type: Boolean,
+        });
```
mshima, line 118: Move the option register to before `if (this.options.help)` otherwise it won't be shown at help and commander will not be aware of this option.

mshima (changes requested):
Make the PR as ready after this changes.

generators/app/index.js
jsone-studios: When moving this to generator-base, the `jhipster import-jdl` command does not work anymore (see failing Azure pipelines). Somehow `this.jhipsterConfig.packageName` is `undefined` at this point, but it is needed inside the `registerGeneratedAnnotationTransform()` function.

generators/generator-base-private.js
```diff
@@ -1469,6 +1469,13 @@ module.exports = class JHipsterBasePrivateGenerator extends Generator {
         generator.registerTransformStream([prettierFilter, prettierTransform(prettierOptions), prettierFilter.restore]);
     }
 
+    registerGeneratedAnnotationTransform(packageName) {
+        if (!packageName) {
+            packageName = this.jhipsterConfig.packageName;
+        }
+        this.registerTransformStream(generatedAnnotationTransform(packageName));
```
mshima, line 1476:
```suggestion
    registerGeneratedAnnotationTransform() {
        this.registerTransformStream(generatedJavaAnnotationTransform(this));
```

generators/generator-transforms.js
```diff
@@ -50,7 +51,25 @@ const prettierTransform = function (defaultOptions) {
     });
 };
 
+const generatedAnnotationTransform = packageName => {
+    return through.obj(function (file, encoding, callback) {
+        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
+            const content = file.contents.toString('utf8');
```
mshima, line 57:
```suggestion
const generatedAnnotationTransform = generator => {
    return through.obj(function (file, encoding, callback) {
        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
            const packageName = generator.jhipsterConfig.packageName;
            const content = file.contents.toString('utf8');
```

mshima (changes requested):

generators/app/index.js
mshima: We should not rely that `packageName` is defined that the constructor point.
It will never be defined when creating an app using prompts.

jsone-studios: Thank you very much @mshima for your extensive help with this PR

mshima (approved):

generators/generator-transforms.js
```diff
@@ -50,7 +51,26 @@ const prettierTransform = function (defaultOptions) {
     });
 };
 
+const generatedAnnotationTransform = generator => {
+    return through.obj(function (file, encoding, callback) {
+        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
+            const packageName = generator.jhipsterConfig.packageName;
+            const content = file.contents.toString('utf8');
+
```
mshima, line 59:
```suggestion
if (!new RegEx(`import ${packageName.replace('\.', '\\.')}\. GeneratedByJHipster;`).test(content)) ) {
```

mshima (changes requested):
Just occurred to me that when adding needles to existing file, the annotation can be duplicated.

mshima (approved):

DamnClin:
Hello, 

you have to keep the retention to RUNTIME so [jacoco will ignore those files ](https://github.com/jacoco/jacoco/issues/685#issuecomment-392020612). In my opinion removing this noise is one of the main reason for this annotation (at least I'm using it for this specific reason).

jsone-studios:
> Hello,
> 
> you have to keep the retention to RUNTIME so [jacoco will ignore those files ](https://github.com/jacoco/jacoco/issues/685#issuecomment-392020612). In my opinion removing this noise is one of the main reason for this annotation (at least I'm using it for this specific reason).

Changing the retention to `RUNTIME` will indeed make jacoco ignore all generated files. But for a fresh generated project it might be weird for users to see an empty coverage report. Also then, no code coverage would be reported to sonar.

pascalgrimaud:
Then, now we have this annotation, people can easily customize with their need, so I'm fine.
Let's merge this.
Thanks a lot @jsone-studios

PR #12598, opened 2020-09-27
Implement domain support at entity-server.

mshima:
This PR implements domain support for entities.
```
@domainName(Bar)
entity Foo {}
```

`domainName` is been used instead if `domain` due to conflict with exiting `domain` property (probably EventEmitter class).
To use `domain` would require us to stop using the generator as template data.

Related to https://github.com/jhipster/generator-jhipster/issues/11122

---

### DIFF
--- a/generators/app/index.js
+++ b/generators/app/index.js
@@ -398,6 +398,30 @@ module.exports = class JHipsterAppGenerator extends BaseBlueprintGenerator {
         return this._loading();
     }
 
+    _preparing() {
+        return {
+            createBuiltInEntitiesStubs() {
+                this.configOptions.sharedEntities = this.configOptions.sharedEntities || {};
+                if (this.jhipsterConfig.skipUserManagement && this.jhipsterConfig.authenticationType !== 'oauth2') return;
+                // Create entity definition for built-in entity to make easier to deal with relationships.
+                this.configOptions.sharedEntities.User = {
+                    entityClassPath: `${this.jhipsterConfig.packageName}.domain.${this.asEntity('User')}`,
+                    entityControllerClassPath: `${this.jhipsterConfig.packageName}.web.rest.UserResource`,
+                    relationships: [],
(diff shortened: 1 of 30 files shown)


pascalgrimaud:
thanks @mshima : I'll review it when possible. If I forget, plz remind me :)

pascalgrimaud:
@mshima : I just give a try and there is some good progress. Thanks a lot for working so fast.
Anyway, there is an important thing we should take into consideration:
In `domainName/domain/`, each class should not import any framework. It should be only default java import

It means:
- `domain/model/Entity.java` should be moved to infrastructure/secondary/entity/Entity.java
- service should be in `domainName/application`
- etc

mshima:
@pascalgrimaud thanks for trying it.

> In `domainName/domain/`, each class should not import any framework. It should be only default java import

I think it can have basic validations from `javax.validation.*`.

> * `domain/model/Entity.java` should be moved to infrastructure/secondary/entity/Entity.java

This is the step I was talking about, every adapter/port should have it's own DTO, or can have.
I am not sure if we should move current entity to secondary or create a new DTO there.

Anyway this is a domain kickstart, I would prefer to put in domain branch instead of my own, creating PRs and discussing there.

1 - If you agree I can create a PR without domain support. Only files, packages, etc configuration to be merged to master to prevent too many conflicts.
2 - And create domain branch.

or just create domain branch.

pascalgrimaud:
I think when using this feature, DTO is mandatory. In fact, I tried using this config:
```
    "entitySuffix": "Entity",
    "dtoSuffix": "",
```

It seems dtoSuffix can't have "" value.

Yes about "domain" branch, you should have the permission to do it
No, don't split your work here, it's already very good start and I don't want you lost too much time on splitting.
Let's create this domain branch and let's work on this !

mshima:
Closed in favor of domain branch.

Issue #10239, opened 2019-08-17
mvnw command throws Compilation failure


mrgenco:
##### **Overview of the issue**

I'm trying to use JHipster for the first time and after running mvnw command, build fails with following stack trace. I don't understand if the error is related to JHipster or Maven. 
Note : I opened a stackoverflow entry for this one but a team member suggested me to open a github issue. https://stackoverflow.com/questions/57526801/
```

[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  8.217 s
[INFO] Finished at: 2019-08-17T09:26:54+03:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project samplejhipster: Compilation failure: Compilation failure:
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[29,39] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[30,39] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[30,43] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[32,42] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[32,44] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[33,37] unmappable character (0xDD) for encoding UTF-8
[ERROR] /C:/Users/mehme/OneDrive/Masaüstü/samplejhipster/target/generated-sources/annotations/org/sample/blog/domain/User_.java:[36,36] unmappable character (0xDD) for encoding UTF-8
[...]

PR #12600, opened 2020-09-27
fixed turkish character problem

geist100:
Fixed jpa metamodel generation character encoding problem. Details: https://stackoverflow.com/questions/57182568/hibernate-jpa-2-metamodel-generator-turkish-char-problem

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1046,6 +1046,12 @@
                     <configuration>
                         <source>${java.version}</source>
                         <target>${java.version}</target>
+                        <fork>true</fork>
+                        <compilerArgs>
+                            <compilerArg>-J-Duser.language=en</compilerArg>
+                            <compilerArg>-J-Duser.country=US</compilerArg>
+                            <compilerArg>-J-Dfile.encoding=UTF-8</compilerArg>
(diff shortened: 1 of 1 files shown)


mshima:
@geist100 seems wrong to force user language.
You should try a less intrusive approach.
Adding <encoding>UTF-8</encoding> to compiler plugin like the second answer seems enough.

geist100:
> @geist100 seems wrong to force user language.
> You should try a less intrusive approach.
> Adding `<encoding>UTF-8</encoding>` to compiler plugin like the second answer seems enough.

Unfortunately not enough add this line <encoding>UTF-8</encoding> i tried many options, only add fork and compiler args works

mshima:
In my opinion this should not be merged as it is now.
This is a locale specific bug, by merging this, we can break many others locales and have unpredictable problems

I think this PR fixes https://github.com/jhipster/generator-jhipster/issues/10239, you should post your fix there.

pascalgrimaud:
Thanks for your PR @geist100 but as mentioned by @mshima we can't merge it, because it can break many others locales and have unpredictable problems on existing projects

Anyway, your proposal can be added to the documentation at https://github.com/jhipster/jhipster.github.io project, if you think it can help. So you're welcome to propose a PR there.

PR #12609, opened 2020-09-30
Add custom claim converter in a microservice with oauth2

### LINKED ISSUE #10975: SecurityUtils .getCurrentUserLogin in a microservice returns "null" with OAuth2 and Okta

Calling `SecurityUtils.getCurrentUserLogin()` in a microservice returns `null` instead of the user login

It should be possible to retrieve the user from the security context

##### **Reproduce the error**

Use `SecurityUtils.getCurrentUserLogin()` in a microservice using Okta for the authentication

##### **Related issues**

#9613

##### **Suggest a Fix**

```java
} else if (authentication instanceof JwtAuthenticationToken) {
    return (String) ((JwtAuthenticationToken)authentication).getToken().getClaims().get("preferred_username");
} else if (authentication.getPrincipal() instanceof DefaultOidcUser) {
```

`preferred_username` doesn't exist when using Okta, is it specific to Keycloak?

##### **JHipster Version(s)**

```
store@0.0.0 /Users/theo/Documents/perso/oauth2/store
└── generator-jhipster@6.5.1

```

<details>

<pre>
application {
  config {
    baseName gateway,
    packageName com.jhipster.gateway,
    applicationType gateway,
    authenticationType oauth2,
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka
  }
  entities Product
}

application {
  config {
    baseName store,
    packageName com.jhipster.store,
    applicationType microservice,
    authenticationType oauth2,
    prodDatabaseType postgresql,
    serviceDiscoveryType eureka
    serverPort 8081
  }
  entities Product
}

entity Product {
  title String
}

microservice Product with store
</pre>
</details>

openjdk version "11.0.4" 2019-07-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.4+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.4+11, mixed mode)

git version 2.23.0

node: v10.17.0

npm: 6.11.3

yeoman: 3.1.1

Docker version 19.03.5, build 633a0ea

docker-compose version 1.24.1, build 4667896b

Falydoor:
Add a custom claim converter to fetch the oauth2 `userinfo` endpoint and add custom claims. This only apply to a microservice using oauth2.

Fix #10975 @mraible (can this PR be merged after September for hacktoberfest 👀)

---

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -900,9 +900,9 @@ const serverFiles = {
             templates: [
                 {
                     file: 'package/security/oauth2/CustomClaimConverter.java',
-                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverter.java`
-                }
-            ]
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverter.java`,
+                },
+            ],
         },
         {
             condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
@@ -910,10 +910,10 @@ const serverFiles = {
             templates: [
                 {
                     file: 'package/security/oauth2/CustomClaimConverterIT.java',
-                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverterIT.java`
-                }
-            ]
-        }
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverterIT.java`,
+                },
+            ],
+        },
     ],
     serverJavaApp: [
         {
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -893,6 +893,28 @@ const serverFiles = {
             templates: ['config/bootstrap.yml', 'config/bootstrap-prod.yml'],
         },
     ],
+    serverMicroserviceAndOauth: [
+        {
+            condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
+            path: SERVER_MAIN_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/security/oauth2/CustomClaimConverter.java',
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverter.java`
+                }
+            ]
+        },
+        {
+            condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
+            path: SERVER_TEST_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/security/oauth2/CustomClaimConverterIT.java',
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverterIT.java`
+                }
+            ]
+        }
+    ],
     serverJavaApp: [
         {
             path: SERVER_MAIN_SRC_DIR,
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -895,7 +895,8 @@ const serverFiles = {
     ],
     serverMicroserviceAndOauth: [
         {
-            condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
+            condition: generator =>
+                !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
             path: SERVER_MAIN_SRC_DIR,
             templates: [
                 {
@@ -905,7 +906,8 @@ const serverFiles = {
             ],
         },
         {
-            condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
+            condition: generator =>
+                !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
             path: SERVER_TEST_SRC_DIR,
             templates: [
                 {
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -593,6 +593,36 @@ const serverFiles = {
                 },
             ],
         },
+        {
+            condition: generator =>
+                !generator.reactive &&
+                generator.authenticationType === 'oauth2' &&
+                (generator.applicationType === 'monolith' ||
+                    generator.applicationType === 'microservice' ||
+                    generator.applicationType === 'gateway'),
+            path: SERVER_MAIN_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/security/oauth2/CustomClaimConverter.java',
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverter.java`,
+                },
+            ],
+        },
+        {
+            condition: generator =>
+                !generator.reactive &&
+                generator.authenticationType === 'oauth2' &&
+                (generator.applicationType === 'monolith' ||
+                    generator.applicationType === 'microservice' ||
+                    generator.applicationType === 'gateway'),
+            path: SERVER_TEST_SRC_DIR,
+            templates: [
+                {
+                    file: 'package/security/oauth2/CustomClaimConverterIT.java',
+                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverterIT.java`,
+                },
+            ],
+        },
     ],
     serverJavaGateway: [
         {
@@ -891,30 +921,6 @@ const serverFiles = {
             condition: generator => generator.serviceDiscoveryType,
             path: SERVER_MAIN_RES_DIR,
             templates: ['config/bootstrap.yml', 'config/bootstrap-prod.yml'],
-        },
-    ],
-    serverMicroserviceAndOauth: [
-        {
-            condition: generator =>
-                !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
-            path: SERVER_MAIN_SRC_DIR,
-            templates: [
-                {
-                    file: 'package/security/oauth2/CustomClaimConverter.java',
-                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverter.java`,
-                },
-            ],
-        },
-        {
-            condition: generator =>
-                !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
-            path: SERVER_TEST_SRC_DIR,
-            templates: [
-                {
-                    file: 'package/security/oauth2/CustomClaimConverterIT.java',
-                    renameTo: generator => `${generator.javaDir}security/oauth2/CustomClaimConverterIT.java`,
-                },
-            ],
         },
     ],
     serverJavaApp: [
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -97,7 +97,7 @@ import <%= packageName %>.security.oauth2.JwtGrantedAuthorityConverter;
     <%_ if (authenticationType === 'jwt' && applicationType !== 'microservice') { _%>
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
     <%_ } _%>
-    <%_ if (authenticationType === 'oauth2' && applicationType === 'microservice') { _%>
+    <%_ if (authenticationType === 'oauth2' && (applicationType === 'monolith' || applicationType === 'microservice' || applicationType === 'gateway')) { _%>
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import <%= packageName %>.security.oauth2.CustomClaimConverter;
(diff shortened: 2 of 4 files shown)


mraible:
Thanks, @Falydoor! I believe you need to add a converter for WebFlux too.

Falydoor:
Unfortunately I can't have `((ServletRequestAttributes) RequestContextHolder.getRequestAttributes()).getRequest()` work with WebFlux 😢 so might be something to see for later.

mraible:
@Falydoor I might be able to help, but might not be for a few days. @cbornet is our resident Spring WebFlux expert.

cbornet:
`RequestContextHolder` is thread local so it can't be used in the reactive stack. I think the solution here is to use a `ServerWebExchangeContextFilter` and then get the ServerWebExchange from the subscriberContext.
See [this example](https://github.com/spring-projects/spring-hateoas/pull/983/commits/0a22b3847bb07af24004e5a9d429f0c71b0042c5)

Falydoor:
@cbornet I played with the `ServerWebExchangeContextFilter` but for some reason the subscriberContext is always empty in the `convert` method of the `CustomClaimConverter`.

Also even if the context contained `ServerWebExchange`, I don't think it would be possible to have it working because the `convert` method must return `Map<String, Object>` instead of `Mono<Map<String, Object>>` (which mean that `.block()` must be called when getting `ServerWebExchange`).

mraible:
@jgrandja Do you have any tips for how to implement a CustomClaimConverter for Spring WebFlux?

jgrandja:
@mraible A custom claim converter may be configured via `OidcUserService.setClaimTypeConverterFactory()` (Servlet) or `OidcReactiveOAuth2UserService.setClaimTypeConverterFactory()` (WebFlux). Both provide a default converter so take a look at that and customize to your needs.

mraible:
@Falydoor Does @jgrandja's advice help?

> A custom claim converter may be configured via `OidcUserService.setClaimTypeConverterFactory()` (Servlet) or `OidcReactiveOAuth2UserService.setClaimTypeConverterFactory()` (WebFlux). Both provide a default converter so take a look at that and customize to your needs.

Falydoor:
@mraible I tried using `OidcReactiveOAuth2UserService.setClaimTypeConverterFactory()` but I wasn't able to make it work unfortunately. The method `CustomClaimConverter.convert()` is never called for some reason 😬 .

mraible:
@Falydoor Is the code for reactive in this PR? If so, I can try and see if I can figure it out.

Falydoor:
@mraible I created a repo with the code, check the commit (I put some comments to help you) https://github.com/Falydoor/jhipster-reactive-oauth/commit/c44ff7b9a151dd870bd2d923cc044f0ec4e1c942

mraible:
@Falydoor I forked your project and there's no `src/main/docker/keycloak.yml` file. Any idea why this wasn't created?

Falydoor:
@mraible I'm using Okta and the keycloak file only comes with the gateway.

mraible:
@Falydoor As far as I can tell, the `oidcUserService()` is not used when the server is acting as a resource server. [This comment](https://github.com/spring-projects/spring-security/issues/6123#issuecomment-440783224) from @jgrandja makes me think it'll only be used when you've logged in with `oauth2Login()`.

Falydoor:
@mraible that would make sense! 

Did you try to uncomment this line https://github.com/Falydoor/jhipster-reactive-oauth/blob/main/src/main/java/com/mycompany/myapp/config/SecurityConfiguration.java#L164 ?

mraible:
Yes, I did try it with that line uncommented and it does call the converter.

Falydoor:
But then in the converter the `subscriberContext` is empty which means that it can't retrieve the token. Maybe there is another way to get the token?

I also tried using `ReactiveSecurityContextHolder.getContext().map(SecurityContext::getAuthentication)` but it's empty.

generators/server/files.js
```diff
@@ -893,6 +893,30 @@ const serverFiles = {
             templates: ['config/bootstrap.yml', 'config/bootstrap-prod.yml'],
         },
     ],
+    serverMicroserviceAndOauth: [
+        {
+            condition: generator =>
+                !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'microservice',
```
mraible, line 899: This conditional (and others) should include `applicationType === 'monolith'` and `gateway` too. We configure these as resource servers by default for Ionic and React Native apps.

mraible:
Using [this guide](https://www.baeldung.com/spring-webclient-oauth2), I tried to register a `WebClient`:

```java
@Bean
WebClient webClient(ReactiveClientRegistrationRepository clientRegistrations,
                    ServerOAuth2AuthorizedClientRepository authorizedClients) {
    ServerOAuth2AuthorizedClientExchangeFilterFunction oauth =
        new ServerOAuth2AuthorizedClientExchangeFilterFunction(
            clientRegistrations,
            authorizedClients);
    oauth.setDefaultOAuth2AuthorizedClient(true);
    oauth.setDefaultClientRegistrationId("oidc");
    return WebClient.builder()
        .filter(oauth)
        .build();
}
```

Then, I used the following line to create the `CustomClaimConverter`:

```java
@Bean
ReactiveJwtDecoder jwtDecoder(ReactiveClientRegistrationRepository clientRegistrationRepository, WebClient webClient) {
    ...
    jwtDecoder.setClaimSetConverter(new CustomClaimConverter(clientRegistrationRepository.findByRegistrationId("oidc").block(), webClient));

    return jwtDecoder;
}
```

Then, I modified the `CustomClaimConverter` to be the following:

```java
package com.mycompany.myapp.security;

import org.slf4j.Logger;
import org.springframework.core.convert.converter.Converter;
import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
import org.springframework.security.oauth2.client.registration.ClientRegistration;
import org.springframework.security.oauth2.jwt.MappedJwtClaimSetConverter;
import org.springframework.web.reactive.function.client.WebClient;
import reactor.core.publisher.Mono;

import java.util.Collections;
import java.util.Map;

public class CustomClaimConverter implements Converter<Map<String, Object>, Map<String, Object>> {
    private final Logger log = LoggerFactory.getLogger(CustomClaimConverter.class);

    private final MappedJwtClaimSetConverter delegate = MappedJwtClaimSetConverter.withDefaults(Collections.emptyMap());

    private final WebClient webClient;
    private final ClientRegistration registration;

    public CustomClaimConverter(ClientRegistration registration, WebClient webClient) {
        this.webClient = webClient;
        this.registration = registration;
    }

    public Map<String, Object> convert(Map<String, Object> claims) {
        Map<String, Object> convertedClaims = this.delegate.convert(claims);
        log.debug("convertedClaims : {} ", convertedClaims);

        Mono.subscriberContext()
            .subscribe(context -> {
                log.debug("CONTEXT SIZE : {}", context.size());
            });

        ClientRegistration.ProviderDetails.UserInfoEndpoint userInfoEndpoint = registration.getProviderDetails().getUserInfoEndpoint();
        log.debug("USER INFO : {}", userInfoEndpoint);

        Mono<String> retrievedResource = webClient.get()
            .uri(userInfoEndpoint.getUri())
            .retrieve()
            .bodyToMono(String.class);

        retrievedResource.subscribe(System.out::print);

        return convertedClaims;
    }
}
```

I get the following error:

```
reactor.core.Exceptions$ErrorCallbackNotImplemented: java.lang.IllegalArgumentException: serverWebExchange cannot be null
Caused by: java.lang.IllegalArgumentException: serverWebExchange cannot be null
        at org.springframework.security.oauth2.client.web.DefaultReactiveOAuth2AuthorizedClientManager.lambda$loadAuthorizedClient$9(DefaultReactiveOAuth2AuthorizedClientManager.java:102)
```

If I modify the WebClient bean to remove `setDefaultClientRegistrationId("oidc")`, it results in:

```
401 Unauthorized from GET https://dev-162929.okta.com/oauth2/default/v1/userinfo
```

It does make the call, but maybe it doesn't have an access token associated with it? I'm not sure.

I thought it might be possible to get the access token like we get the ID token in `LogoutResource`, but I'm not sure how that might work.

```
@AuthenticationPrincipal(expression = "idToken") OidcIdToken idToken
```

Maybe it's possible to get the `Authentication` from `ReactiveSecurityContextHolder` (like [this](https://stackoverflow.com/a/51543137/65681)), and then cast it to a [`Jwt`](https://docs.spring.io/spring-security/site/docs/5.4.1/api/org/springframework/security/oauth2/jwt/Jwt.html), which has a `getTokenValue()` method.

mraible:
I spent several hours working on this yesterday but didn't succeed. Maybe documenting what I tried will help. 

First of all, I found this concise blog post that makes it look easy to make a WebClient request with an access token. However, it uses `client_credentials`, whereas JHipster uses auth code flow. https://medium.com/@asce4s/oauth2-with-spring-webclient-761d16f89cdd

I found a Spring Security example that seems close to what we want. However, I need something that works without `oauth2Login()` since that’s not part of JHipster when it’s a resource server. https://github.com/spring-projects/spring-security/tree/master/samples/boot/oauth2webclient-webflux

Error I'm getting:

```
2020-11-11 16:08:46.649  INFO 36332 --- [ctor-http-nio-3] c.m.myapp.security.CustomClaimConverter  : userinfo endpoint https://dev-162929.okta.com/oauth2/default/v1/userinfo
2020-11-11 16:08:46.656 ERROR 36332 --- [ctor-http-nio-3] reactor.core.publisher.Operators         : Operator called default onErrorDropped
reactor.core.Exceptions$ErrorCallbackNotImplemented: java.lang.IllegalArgumentException: serverWebExchange cannot be null
Caused by: java.lang.IllegalArgumentException: serverWebExchange cannot be null
        at org.springframework.security.oauth2.client.web.DefaultReactiveOAuth2AuthorizedClientManager.lambda$loadAuthorizedClient$9(DefaultReactiveOAuth2AuthorizedClientManager.java:102)
        Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.MonoErrorSupplied] :
```

Adding this at the top of the security config doesn’t help: 

```
.addFilterAt(new ServerWebExchangeContextFilter(), SecurityWebFiltersOrder.FIRST)
```

To test everything, I'm using Ionic for JHipster since it surfaced the problem for me. I re-generated @Falydoor's project as a monolith running on 8080 so it would work with Ionic4J. You can see the difference between the projects at https://github.com/Falydoor/jhipster-reactive-oauth/compare/main...mraible:userinfo.

Search for "SecurityConfiguration.java" and "CustomerClaimConverter.java" to see the changes I tried.

mraible:
@Falydoor Maybe we should just merge this and open a new issue for reactive?

Falydoor:
@mraible yes I agree. You did more research than me for the reactive support and it is trickier than what we thought...

generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
```diff
@@ -341,14 +346,17 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
...
         NimbusJwtDecoder jwtDecoder = (NimbusJwtDecoder) JwtDecoders.fromOidcIssuerLocation(issuerUri);
 
         OAuth2TokenValidator<Jwt> audienceValidator = new AudienceValidator(jHipsterProperties.getSecurity().getOauth2().getAudience());
         OAuth2TokenValidator<Jwt> withIssuer = JwtValidators.createDefaultWithIssuer(issuerUri);
         OAuth2TokenValidator<Jwt> withAudience = new DelegatingOAuth2TokenValidator<>(withIssuer, audienceValidator);
 
         jwtDecoder.setJwtValidator(withAudience);
+        <%_ if (authenticationType === 'oauth2' && applicationType === 'microservice') { _%>
+        jwtDecoder.setClaimSetConverter(new CustomClaimConverter(clientRegistrationRepository.findByRegistrationId("oidc"), new RestTemplate()));
+        <%_ } _%>
```
Tcharl, line 359: Using new RestTemplate causes so much pains in real world (for example, when your app is behind an outbound proxy).
please prefer `RestTemplate rt = restTemplateBuilder.build();`

generators/server/files.js
mraible: @Falydoor I believe this comment still needs to be addressed.

generators/server/templates/src/test/java/package/security/oauth2/CustomClaimConverterIT.java.ejs
```diff
@@ -0,0 +1,90 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+@SpringBootTest(classes = {<%= mainClass %>.class, TestSecurityConfiguration.class})
```
pascalgrimaud, line 50: Use new `@IntegrationTest` plz, otherwise, it's broken for some config like Redis etc.

Falydoor: Sounds good! I opened https://github.com/jhipster/generator-jhipster/pull/13462

mraible:
After thinking about this a bit more, I'm not sure we need a `/userinfo` lookup in a microservice. I think it's only needed in a monolith or gateway - because that's where the access token will come in from the client. Thoughts?

Issue #12623, opened 2020-10-01
Add svelte support


heithemmoumni:
I suggest adding Svelte to the list of frontend frameworks

mraible:
This is currently being worked on. See https://github.com/jhipster/generator-jhipster-svelte. Help is most welcome!

PR #12625, opened 2020-10-01
feat: add support svelte

heithemmoumni:


murdos:
As already said [here](https://github.com/jhipster/generator-jhipster/issues/12623#issuecomment-702350561), there's already a work in progress support for Svelte outside main generator: https://github.com/jhipster/generator-jhipster-svelte

PR #12634, opened 2020-10-02
Export jdl with primary and secondary in relationship definition.

mshima:
Add `primary` or `secondary` option to relationship.
Relationship annotations are set at both sides.
When developing blueprints to customize one side of the relationship using jdl is not possible.
```
relationship OneToMany {
  @customPrimaryOption
  A to B
}
```

---


MathieuAA:
I don't get the concept here, why primary and secondary?

mshima:
> I don't get the concept here, why primary and secondary?

It's just to differentiate one side from the other, so I could use the annotation at the correct side.
@MathieuAA do you suggest another naming?

MathieuAA:
Oh, in this case this is okay I guess. The sides are called source and destination, but it would be verbose to use them here, don't you think?

mshima:
> Oh, in this case this is okay I guess. The sides are called source and destination, but it would be verbose to use them here, don't you think?

I used primary and secondary because, I prefer to put the owner at the left side.
But could be left\right, leftRelationship\rightRelationship, jdlSource\jdlDestination.
Not sure.

mshima:
Closing in favor of https://github.com/jhipster/generator-jhipster/issues/12949

PR #12674, opened 2020-10-07
Fix Okta login failure message and add Okta CLI to docs

mraible:
Question: If I can't find `Sign in failed!` in Cypress tests, does that mean they're not configured to test Okta?

Related: I was unable to get `run run e2e` to pass even after this change. I did set `E2E_USERNAME` and `E2E_PASSWORD` beforehand.

It seems to be because the failure happens too fast. Modifying `account.spec.ts` to have the following seems to fix it:

```ts
const error = element(by.css('.infobox-error'));      
await browser.sleep(1000);
expect(await error.getText()).to.eq('Unable to sign in');
```

The other error I found is the successful login test appends the correct password instead of replacing it. Not sure what changed. This code used to work when I last tried it.

### DIFF
--- a/generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
+++ b/generators/client/templates/angular/src/test/javascript/e2e/account/account.spec.ts.ejs
@@ -73,7 +73,7 @@ describe('account', () => {
         } else {
             // Okta
             const error = element(by.css('.infobox-error'));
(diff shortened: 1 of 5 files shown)


pascalgrimaud:
cc @avdev4j @nassimerrahoui @adilabed about Cypress

pascalgrimaud (approved):

PR #12681, opened 2020-10-08
Remove console log when generating an app with OAuth 2.0

### LINKED ISSUE #12671: Remove logging about user-management when creating app

When I run `jhipster jdl blog-oauth2`, it prints out a message that doesn't really help the user.

```
user-management will be handled by UAA app, oauth expects users to be managed in IpP
```

I think this should be removed.

Screen Shot 2020-10-07 at 11 03 39 AM

It kinda looks like something went wrong. I don't think info messages should be written to the console when creating an app. Maybe at the end.

mraible:
Fixes #12671.


DanielFran (approved):
LGTM

Issue #12090, opened 2020-07-17
JHipster Control Center


avdev4j:
##### **Overview of the feature request**
A few weeks ago, @PierreBesson has written an RFC to provide companion sidecar to manage JHipster apps (https://github.com/jhipster/generator-jhipster/issues/10913). This RFC is available here : https://github.com/jhipster/generator-jhipster/blob/master/rfcs/2-jhipster-rfc-jhipster-control-center.md

We decide to focus this feature on the V7. This is a different way to imagine how we can manage and control JHipster applications, monoliths or microservices.

This is how JHipster Control Center (aka JHCC) is born.

##### **Who are you JHCC?**
[JHipster Control Center](https://github.com/jhipster/jhipster-control-center) is a JHipster separated application that will allow you to manage and control your microservice or monolith applications from a single place.

Spring Cloud is used to reach your applications, 4 modes are supported:
- Eureka server
- Consul Server
- Kubernetes
- A static range URI

You can log to JHCC through:
- JWT
- Oauth2

You can test your API through a unified Swagger-UI.

The management UI is developed with Vue front-end framework.

##### **How to use JHCC**
We plan to provide JHCC through a Docker image or a jar file. JHipster applications will contains all the necessary Docker-compose configuration file to start JHCC out of the box.
JHCC is made to help developers during both development and production phases, that's why we want JHCC to be easy as possible to be ran.

##### **What are the consequences to generated JHipster applications?**
JHipster would no longer generate these features since they will be dedicated to JHCC:
- Open-api documentation
- Logs, Metrics, Configuration, properties and Health admin pages

These mean less generated code, to simplify the maintenance and avoid duplicate UI through 3 different frameworks.
Also, we want to, by this way, reduce the front-end compilation time. 

JHipster Registry will be deprecated for V7+.

##### **Conclusion**

The application is still under development and @nassimerrahoui and @adilabed are the main contributors since this is their traineeship subject. We are glad to know they are outperforming our expectations for this feature.
But, as usual, there is a lot of work to remain and we really need the community help to fully integrate this application into the V7.

Feel free to contribute and give us feedbacks.

PS: A list of what we will have to do on main generator
- [x] Don't always generate admin ui (Angular/React/Vue) https://github.com/jhipster/generator-jhipster/pull/12738
- [x] Add JHCC Docker-Compose files https://github.com/jhipster/generator-jhipster/pull/12126
[...]

PR #12738, opened 2020-10-13
Generate admin ui according to the flag value

avdev4j:
Add a --with-admin-ui flag to generate the admin ui pages. By default, those pages are not generated and the user should use the JHipster Control Center application.

EDIT: add prompt instead

#12167 
#12090

-   [x] Angular
-   [x] React
-   [x] Vue
-   [x] i18n
-   [x] JDL

---

-   [x] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary https://github.com/jhipster/jhipster-online/pull/257
-   [X] Documentation is added/updated where necessary https://github.com/jhipster/jhipster.github.io/pull/1026


mraible:
Does this mean we're going to require two apps to run one app (if you want to administer the one app)?

avdev4j:
> Does this mean we're going to require two apps to run one app (if you want to administer the one app)?

Unless we use the flag I have added ;).

generators/client/index.js
```diff
@@ -63,6 +63,12 @@ module.exports = class JHipsterClientGenerator extends BaseBlueprintGenerator {
             type: Boolean,
         });
 
+        // This adds support for a `--with-admin-ui` flag which can be used to generate the admin UI pages
+        this.option('with-admin-ui', {
+            desc: 'Generate the UI admin pages. By default thi feature is supported by the JHipster Control Center.',
```
MarlonLuan, line 68:
```suggestion
            desc: 'Generate the UI admin pages. By default this feature is supported by the JHipster Control Center.',
```

MarlonLuan (commented):
Typo

pascalgrimaud:
I'm not totally in favor of this, for now. 
I'd prefer to have no-admin-ui flag instead.

The reason is the JHipster Control Center is not finished yet today, and we don't have official lead on this. So it can become a dead project, if it's not well maintained.

avdev4j:
> I'm not totally in favor of this, for now.
> I'd prefer to have no-admin-ui flag instead.

This was my main issue yesterday, Maybe we should add a question, something like (Would like to administrate your application with the JHipster Control Center?(the admin ui will be not generated) y/N".

Because if we add a flag like no-admin-ui I'm sure that nobody will use it :D?

With this question we will be able to add the JHCC Docker compose configuration only when the answer is "yes". 

WDYT?

PierreBesson:
I would say let's generate the jhcc docker compose all the time so people can try it out in any case.... But I'm conflicted with Pascal's idea that maybe the admin pages should stay as default. As for if the admin UI should be generated by default or not, I think it would be better to actually do a vote about it on the mailing list.

generators/generator-base.js
```diff
@@ -2325,6 +2328,7 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
         dest.experimental = config.experimental;
         dest.logo = config.logo;
         dest.backendName = config.backendName || 'Java';
+        dest.withAdminUi = config.withAdminUi;
```
mshima, line 2331: Move to loadAppConfig.

generators/generator-base.js
```diff
@@ -2238,6 +2238,9 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
         if (options.recreateInitialChangelog) {
             this.configOptions.recreateInitialChangelog = options.recreateInitialChangelog;
         }
+        if (options.withAdminUi) {
+            this.configOptions.withAdminUi = options.withAdminUi;
```
mshima, line 2242:
```suggestion
        if (options.withAdminUi !== undefined) {
            this.jhipsterConfig.withAdminUi = options.withAdminUi;
```
For reproducibility.

generators/client/index.js
avdev4j: Done!

generators/generator-base.js
avdev4j: Done, what is the difference between configOptions and jhipsterConfig?

pascalgrimaud:
It needs to be discussed, when admin ui is not choosen, should we keep swagger or not ?
I'd say:
- with admin ui: like today
- without admin ui: user-management + swagger

Swagger is so important, and it would not required too much maintenance !

avdev4j:
after all, I think you're right @pascalgrimaud it's not a big effort to keep that so let's revert this changes.

avdev4j:
@MathieuAA I will maybe ping you to know how we can introduce that in the JDL Application (and if it's a good idea).

MathieuAA:
@avdev4j It's an application option, right? So if it's a matter of adding another option, it's no big deal.

test/jdl/jhipster/default-application-options.spec.js
```diff
@@ -60,6 +60,9 @@ describe('DefaultApplicationOptions', () => {
             it('should set the client theme variant option to none', () => {
                 expect(options.clientThemeVariant).to.equal('');
             });
+            it('should set withAdminUI option to true', () => {
+                expect(options.withAdminUi).to.equal(true);
```
MathieuAA, line 64: A small improvement: `expect(options.withAdminUi).to.equal(true);` => `expect(options.withAdminUi).to.be.true;`

MathieuAA (changes requested):
I've quickly reviewed the JDL part.
I'll do a more thorough review for this.

generators/client/templates/angular/src/main/webapp/app/layouts/navbar/navbar.component.html.ejs
```diff
@@ -80,6 +80,7 @@
                         </a>
                     </li>
                     <%_ } _%>
+                    <%_ if(withAdminUi){ _%>
```
MathieuAA, line 83:
```suggestion
                    <%_ if (withAdminUi) { _%>
```

generators/client/prompts.js
```diff
@@ -188,3 +189,23 @@ function askForClientThemeVariant() {
...
 }
+
+function askForAdminUi() {
+    if (this.existingProject) {
+        return undefined;
+    }
+
+    const skipClient = this.skipClient;
+
+    const PROMPT = {
```
MathieuAA, line 200: why is this uppercase?

generators/client/templates/react/src/main/webapp/app/modules/administration/administration.reducer.ts.ejs
```diff
@@ -71,31 +75,39 @@ export default (state: AdministrationState = initialState, action): Administrati
...
     <%_ } _%>
+  <%_ if (withAdminUi) { _%>
     case REQUEST(ACTION_TYPES.FETCH_METRICS):
     case REQUEST(ACTION_TYPES.FETCH_THREAD_DUMP):
     case REQUEST(ACTION_TYPES.FETCH_LOGS):
     case REQUEST(ACTION_TYPES.FETCH_CONFIGURATIONS):
     case REQUEST(ACTION_TYPES.FETCH_ENV):
     case REQUEST(ACTION_TYPES.FETCH_HEALTH):
+  <%_ } _%>
+  <%_ if (withAdminUi ||  (applicationType === 'gateway' && serviceDiscoveryType)) { _%>
```
MathieuAA, line 86:
```suggestion
  <%_ if (withAdminUi || (applicationType === 'gateway' && serviceDiscoveryType)) { _%>
```

generators/client/templates/react/src/main/webapp/app/modules/administration/administration.reducer.ts.ejs
```diff
@@ -71,31 +75,39 @@ export default (state: AdministrationState = initialState, action): Administrati
...
     <%_ } _%>
+  <%_ if (withAdminUi) { _%>
     case FAILURE(ACTION_TYPES.FETCH_METRICS):
     case FAILURE(ACTION_TYPES.FETCH_THREAD_DUMP):
     case FAILURE(ACTION_TYPES.FETCH_LOGS):
     case FAILURE(ACTION_TYPES.FETCH_CONFIGURATIONS):
     case FAILURE(ACTION_TYPES.FETCH_ENV):
     case FAILURE(ACTION_TYPES.FETCH_HEALTH):
+  <%_ } _%>
+  <%_ if (withAdminUi ||  (applicationType === 'gateway' && serviceDiscoveryType)) { _%>
```
MathieuAA, line 104:
```suggestion
  <%_ if (withAdminUi || (applicationType === 'gateway' && serviceDiscoveryType)) { _%>
```

test/jdl/grammar/validator.spec.js
```diff
@@ -503,6 +503,33 @@ describe('JDLSyntaxValidatorVisitor', () => {
...
+            context('an invalid value', () => {
+                it('should report a syntax error', () => {
+                    expect(() =>
+                        parse(`
+            application {
+              config {
+                withAdminUi 666
+              }
+            }`)
+                    ).to.throw(new RegExp('^A boolean literal is expected, but found: "666"'));
```
MathieuAA, line 529: Wouldn't the following work?

```suggestion
                    ).to.throw(/^A boolean literal is expected, but found: "666"/);
```

generators/client/prompts.js
avdev4j: I copied this from another prompt in the client prompt file. If you want I cant rename all the PROMPT to prompt.

test/jdl/grammar/validator.spec.js
avdev4j: Indeed we should use literal regexp, I do the modification for the whole file

generators/client/prompts.js
avdev4j: Ok so I think this have been done to avoir misunderstanding between PROMPT and prompt in the callback ;)

MathieuAA: Really? Wow. Talk about dev laziness... I know naming things is one of the most difficult things we devs have to do but still...

MathieuAA: The PROMPT thing is a param object, its name could be "promptParams". The prompt function param is regular param, which represents a prompt result, an answer.

avdev4j:
@MathieuAA I have tested many things and finally did not find the better way, except if we move to async functions.

Less code, easier to read and ofc less "PROMPT" ;).

I have added a commit, about that, in this PR but we can split into two different ones. 

This is related to this PR: https://github.com/jhipster/generator-jhipster/pull/12812

What do you think @mshima @sendilkumarn ?

mshima (approved):
LGTM.

2 non-blocker considerations:
- I would prefer --admin-ui instead of --with-admin-ui.
- httpsGet should be extracted into a util function instead of a generator-private member.

generators/client/prompts.js
```diff
@@ -113,59 +112,28 @@ function askForClientTheme() {
         { value: 'yeti', name: 'Yeti' },
     ];
 
-    const PROMPT = {
+    const bootSwatchChoices = await retrieveBootswatchThemes(self).catch(() =>
```
MathieuAA, line 115: Is it bootSwatch or bootswatch? I'm confused...

MathieuAA:
That's a much better solution @avdev4j :)

MathieuAA (approved):

generators/client/prompts.js
avdev4j: bootswatch indeed ;), let me fix that ;)

avdev4j:
> * I would prefer --admin-ui instead of --with-admin-ui.
 --with-admin-ui no longer exist because we move to a prompt instead of a flag.

> * httpsGet should be extracted into a util function instead of a generator-private member. Not the scope of this PR.
Not scope in this PR but let rework the function to easily extract it in the future.

pascalgrimaud:
congrats for your hard work @avdev4j !

pascalgrimaud:
lot of work done to achieve this, so increasing the bounty, it's well deserved

avdev4j:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/27873

pascalgrimaud:
@avdev4j : approved

Issue #12489, opened 2020-09-12
Generated CI with e2e


pascalgrimaud:
##### **Overview of the issue**

Our generated CI doesn't have e2e tests.
It would be nice if it was possible

##### **Motivation for or Use Case**

Better CI/CD

##### **Reproduce the error**

- generate a project with e2e options (protractor or cypress)
- generate CI with jhipster ci-cd
- look at the generated CI file -> no e2e tests

##### **Browsers and Operating System**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
@pascalgrimaud I think we need to integrate https://github.com/mshima/generator-jhipster-scripts for this.

pascalgrimaud:
@mshima : I had a look and there are some really great ideas.
So I'm sure we can improve our generated package.json, so it will be easier to start the database/service, without thinking about:
- what do I need to start, before starting my application ? oh yes, it is postgresql and elasticsearch
- then, what is the command ? docker-compose -f src/main/docker/....
- etc

JeroenAdam:
FWIW, I used the Jhipster 6.5.1 demo app for building my first Jenkins CI/CD pipeline. It included Protractor tests.
I went for the Docker in Docker approach with some config to keep only the latest few builds and with a cleanup of unused containers at the end of each build.

I'm not the strongest in automation, some notes I made:
 * Jenkins and Docker-in-Docker, to persist the docker permission denied issue: chmod 666 /var/run/docker.sock (should be set at boot time with sysv-rc-conf), not sure about this regarding security best-practices, but it works 
https://medium.com/swlh/quickstart-ci-with-jenkins-and-docker-in-docker-c3f7174ee9ff
https://www.digitalocean.com/community/questions/how-to-fix-docker-got-permission-denied-while-trying-to-connect-to-the-docker-daemon-socket
 * chmod +x mvnw  as described here https://github.com/jhipster/generator-jhipster/issues/6942 
 * (manual setup done once, to avoid downloading dependencies over and over again)
   docker exec -u root -it jenkins /bin/bash
   cd /var/jenkins_home/workspace/yourproject
   chmod +x mvnw  
   ./mvnw clean package -Pprod -DskipTests
   mkdir m2prod
   mv /root/.m2 m2prod (will be moved to /var/jenkins_home/workspace/yourproject/m2prod/.m2)
   rm -Rf target && mkdir target
[...]

PR #12765, opened 2020-10-15
Implement e2e at cicd generator.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/12489.

- Not tested
- The missing ci is gitlab, the implementation is quite different.

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -306,7 +306,7 @@ jobs:
             #----------------------------------------------------------------------
             - name: 'TESTS: Start docker-compose containers for e2e and backend tests'
               if: steps.tests-should-be-skipped.outputs.skip-tests != 'true' && matrix.testcontainers != 1
-              run: npm run ci:e2e:prepare:docker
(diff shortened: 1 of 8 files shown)


pascalgrimaud:
I'd prefer to have one additional question in the CI/CD sub generator. Is it possible ?
The question won't be there for GitLab CI / Jenkins, as it depends too much on the environment.
And by default, we won't generate e2e tests.

The main reason is because it won't work out of the box for Gateway:
- the gateway needs other microservices
So we could add a warning when the user selected e2e tests for Gateway

What do you think?

mshima:
> What do you think?

Yes you are right, we should disable e2e for gateway.
Omit or comment?

pascalgrimaud:
Good idea for putting it in comment !
In this case, no need to add a new question

Can you add this in comment too for GitLab CI and Jenkinsfile ?
So the user will have the model to update this part, if he really wants e2e

mshima:
I will change that.
I haven't implemented GitLab, is currently implemented in sections, needs more expertise or time.

PR #12776, opened 2020-10-16
Revert "Bump zone.js from 0.10.3 to 0.11.1 in /generators/client/templates/angular"

mshima:
Reverts jhipster/generator-jhipster#12775

Reference https://github.com/jhipster/generator-jhipster/pull/12762.


mshima:
dependabot should have found the conflict, I will have to change the angular dependency reference.

Issue #12665, opened 2020-10-06
Daily builds: JDL microservices broken with OAuth2+Cypress


pascalgrimaud:
##### **Overview of the issue**

See https://github.com/hipster-labs/jhipster-daily-builds

The daily builds for JDL Microservices + OAuth2 + Cypress are broken since days

##### **Browsers and Operating System**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

SudharakaP:
I can take a look at this issue if no one is on it already. 😄

pascalgrimaud:
yes plz, @SudharakaP 

Here some very quick analyze:

- for JDLMs in GitHub Actions, the registry started at the same time than Keycloak, and Keycloak took approximatively 20sec to be "up", so the registry failed and is not up, so the other services can't start as the registry is not up

- once, this 1st bug is fixed, Cypress can't access to Keycloak for unknown reason -> I manage to reproduce the issue locally. Maybe because cypress.json has "localhost:9080" hard coded, and here, we should use "keycloak:9080"

Easy to reproduce locally:
- create a gateway with OAuth2
- create a microservice with OAuth2
- generate entity for microservice
- generate the front part for this entity in gateway
- build image with docker
- run sub generator dockercompose
- then, start the stack -> it will fail because of registry
- once it's fixed, launch cypress -> it will fail because of config with cypress

SudharakaP:
@pascalgrimaud : Interestingly enough I think we didn't used to run the e2e tests when things were passing before; for example; https://github.com/hipster-labs/jhipster-daily-builds/runs/1127373854?check_suite_focus=true. Although just switching to Protractor and running the e2e seems to fix the issue; which probably means I think this is a mis-configuration in Cypress. 

Run with Protractor in my fork: https://github.com/SudharakaP/jhipster-daily-builds/runs/1227102969?check_suite_focus=true

pascalgrimaud:
@SudharakaP : yes, it's related to Cypress, as it works approximatively well with protractor (but with some random failures)

nassimerrahoui:
[...]

Issue #12784, opened 2020-10-18
React and OAuth2: bug with relationship to user


pascalgrimaud:
##### **Overview of the issue**

When saving an entity which has relationship to user, and letting user empty, it failed

##### **Motivation for or Use Case**

we should be able to save entity without user

##### **Reproduce the error**

- generate an app with React + OAuth2
- add entity with relationship to user
- start app
- save an entity with empty user

##### **Related issues**

Related to https://github.com/jhipster/generator-jhipster/pull/12777

##### **JHipster Version(s)**

current master

##### **JHipster Version(s)**

```
burger@0.0.1-SNAPSHOT /home/pgrimaud/tmp/28-react-oauth2
└── (empty)

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "blueprints": [],
    "otherModules": [],
    "applicationType": "monolith",
    "reactive": false,
    "baseName": "burger",
    "jhipsterVersion": "6.10.3",
    "skipClient": false,
    "skipServer": false,
    "skipUserManagement": true,
    "skipCheckLengthOfIdentifier": false,
    "skipFakeData": false,
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "clientPackageManager": "npm",
    "testFrameworks": ["cypress"],
    "pages": [],
    "creationTimestamp": 1603014850708,
    "promptValues": {
      "packageName": "io.github.pascalgrimaud",
      "nativeLanguage": "en"
    },
    "serviceDiscoveryType": false,
    "authenticationType": "oauth2",
    "packageName": "io.github.pascalgrimaud",
    "serverPort": "8080",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "buildTool": "maven",
    "serverSideOptions": [],
    "websocket": false,
    "searchEngine": false,
    "messageBroker": false,
    "enableSwaggerCodegen": false,
    "clientFramework": "react",
    "clientTheme": "none",
    "enableTranslation": true,
    "nativeLanguage": "en",
    "packageFolder": "io/github/pascalgrimaud",
    "clientThemeVariant": "primary",
    "useSass": true,
    "languages": ["en"]
  }
}

</pre>
</details>

[...]

Issue #12686, opened 2020-10-08
Add Okta support to Cypress tests


mraible:
##### **Overview of the feature request**

Our current Protractor tests support testing against Okta. Our Cypress tests should have the same ability. The major thing to make this happen is allowing `E2E_USERNAME` and `E2E_PASSWORD` to be read if they're set. Otherwise, default to `admin/admin` for Keycloak.

##### **Motivation for or Use Case**

I like to see if Okta works with the latest version of JHipster.

##### **Related issues or PR**

#12674

-   [x] Checking this box is mandatory (this is just to show you read everything)

mshima:
Rename:
- keycloackLogout to oauth2Logout or something generic.
- keycloackLogin to oauth2Login or something generic.

mraible:
Please keep this open as it's something I care about.

atomfrede:
I have maybe found and ugly workaround (will only work with chrome I think). Not sure we should further investigate that but it looks okaish to me as we don't need to interact with either keycloak or okta auth api directly https://github.com/cypress-io/cypress/issues/944#issuecomment-567453827

atomfrede:
@mraible Good news, I could make use of puppeteer to login into keycloak with the browser (headless or non headless) and use the resulting cookies inside the cypress tests. Basically following has changed:

`plugins/index.tx`:

```ts
  on('task', {
    login({ username, password }) {
      return (async () => {
        const browser = await puppeteer.launch({ 
          args: ['--no-sandbox', '--disable-setuid-sandbox'],
          devtools: true, 
          ignoreHTTPSErrors: true, 
          headless: true });
          
        const page = await browser.newPage();
   
        await page.goto('http://localhost:8080/oauth2/authorization/oidc', { // The app redirects to the login-page
          waitUntil: 'networkidle2' // Wait until login-page has been reached
        });
        
        await page.type('#username', username); // Insert username in form
        await page.type('#password', password); // Insert password
        await page.click('input[type="submit"]'); // Click login button

        await page.waitForNavigation({ waitUntil: 'networkidle2' }); // Wait until redirected back to the app
        
        const authCookies = await page.cookies();
        
        await browser.close(); // Close puppeteer
        return authCookies;
      })();
    },
    logout() {
      return (async () => {
      const browser = await puppeteer.launch({ ignoreHTTPSErrors: true });
      const page = await browser.newPage();
      await page.goto('http://localhost:8080/api/logout', { // The app redirects to the login-page
[...]

PR #12785, opened 2020-10-18
Fixes to cypress with oauth2.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/12665
- Change cypress to lazy load oauth2 data
- Fix oauth2 login.

---


azure-pipelines.yml
```diff
@@ -74,7 +74,7 @@ jobs:
                   JHI_APP: ms-react-gateway-consul-oauth2
                   JHI_ENTITY: sqllight
                   JHI_PROFILE: prod
-                  JHI_E2E: 0
+                  JHI_E2E: 1
```
pascalgrimaud, line 77: Not sure about this, as I'm pretty sure it still fails because of https://github.com/jhipster/generator-jhipster/issues/12784
it's not related to Cypress, but it's a bug with React when saving entities which have relationship with User

Protactor doesn't detect it as it's less completed than Cypress

mshima: I would have looked at it, there is a PR to fix it already.
IMO this use case should be tested by sqlfull if not tested yet.
Maybe add a sqlfull to github react+oauth and vue+oauth if missing.

pascalgrimaud: yes, this PR should fixed this -> https://github.com/jhipster/generator-jhipster/pull/12787

sqlfull was specially done for default app. I never tested it with OAuth2
But I think we tested too much entities, we should reduce and merge combinations

mshima: Resolving since it's already enabled at main.

generators/cypress/templates/src/test/javascript/cypress/support/keycloak-oauth2.ts.ejs
```diff
@@ -4,89 +4,91 @@
...
+    };
+    cy.wrap(data).as('oauth2Data');
+  });
+});
 
-  cy.fixture(`users/${user}`).then((userData) => {
-  const authBaseUrl = Cypress.env("auth_base_url");
-  const realm = Cypress.env("auth_realm");
-  const client_id = Cypress.env("auth_client_id");
+Cypress.Commands.add('keycloackLogin', (oauth2Data: any, user: string) => {
```
mraible, line 26: Why do we have to add something Keycloak specific? Why can't we just click on the Logout button like we can with Protractor?

mshima: Due to https://github.com/cypress-io/cypress/issues/944.
We must authenticate oauth2 (if the domain is not the same) using api.

mraible: We've tried not to use anything Keycloak or Okta-specific in the past. Can this be renamed to "OIDC" or similar? And can we make sure it works with Okta? Our Protractor e2e tests work with both identity providers.

mshima: Sure, I just didn't renamed the function, it was already there.
The one I created, I named `getOauth2Data`.

Just added this todo to https://github.com/jhipster/generator-jhipster/issues/12686.
I was going to implement it, but I forgot.

Issue #12809, opened 2020-10-21
[JDL] Stop setting languages as ['en', 'fr'].


mshima:
##### **Overview of the feature request**

Export empty languages by default.

##### **Motivation for or Use Case**

I want jdl samples to be localized in my language.
`jhipster jdl bug-tracker.jh`.

Next step is to use os-locale to set the language or fallback to ['en'] when locale is not valid or environment `CI===true`.

##### **Related issues or PR**

https://github.com/jhipster/generator-jhipster/blob/653e893014cf9768d38f7069db113395b98ac501/jdl/jhipster/application-options.js#L147

-   [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
CC @MathieuAA

MathieuAA:
How will the languages be set? If it's handled by the generator then I guess it's okay

MathieuAA:
PR approved, just one conflict to resolve.

mshima:
Thanks @MathieuAA

PR #12810, opened 2020-10-21
Change jdl to don't set nativeLanguage and languages if not present in jdl.

mshima:
Closes https://github.com/jhipster/generator-jhipster/issues/12809.

- try to set nativeLanguage to current locale
- defaults to en when environment variable CI is true

---


test/jdl/jhipster/default-application-options.spec.js
```diff
@@ -479,17 +479,12 @@ describe('DefaultApplicationOptions', () => {
             it('should set the jhipster prefix option to jhi', () => {
                 expect(options.jhiPrefix).to.equal('jhi');
             });
-            it('should set the languages option to en and fr', () => {
-                expect(options.languages).to.have.lengthOf(2);
-                expect(options.languages).to.include('en');
-                expect(options.languages).to.include('fr');
+            it('should set the languages option to empty', () => {
```
MathieuAA, line 482: should not set

MathieuAA (changes requested):
So the user has to specify a language if the locale one isn't the right one? all the time?
I don't think it's a good idea...

test/jdl/jhipster/default-application-options.spec.js
mshima: Empty array then.

mshima:
> So the user has to specify a language if the locale one isn't the right one? all the time?
> I don't think it's a good idea...

Why is better to set 'en' then guessing from user locale?

test/jdl/jhipster/default-application-options.spec.js
mshima:
```suggestion
            it('should set the languages option to empty array', () => {
```

test/jdl/jhipster/default-application-options.spec.js
```diff
@@ -479,17 +479,12 @@ describe('DefaultApplicationOptions', () => {
...
+                expect(options.languages).to.have.lengthOf(0);
```
mshima, line 483:
```suggestion
                expect(options.languages).to.be.eql([]);
```

MathieuAA:
Unless the user explicitly asks for a language, getting it fro the OS's locale may not work: my locale's set to French, but my apps are always in English.

mshima:
@MathieuAA dropped user language.

mshima:
> Unless the user explicitly asks for a language, getting it fro the OS's locale may not work: my locale's set to French, but my apps are always in English.

You just need to set your apps native language and languages at jdl.

Having default to user locale should mainly influence at examples like `jhipster jdl bug-tracker`.

MathieuAA (approved):

pascalgrimaud:
I didn't follow closely what is done here, but in my opinion, the command `jhipster jdl bug-tracker.jh` should generate the same project, no matter the machine

If it's different because of local, we'll have a lot of difficulty to reproduce, to guess the locale of the user etc...

Just my opinion

mshima:
@pascalgrimaud I planning to merge this PR so we can change the generator to make `jhipster jdl bug-tracker.jh --native-language xx` generate only the native language.

> I didn't follow closely what is done here, but in my opinion, the command `jhipster jdl bug-tracker.jh` should generate the same project, no matter the machine
> 
> If it's different because of local, we'll have a lot of difficulty to reproduce, to guess the locale of the user etc...
> 
> Just my opinion

Your concern is valid I will think about this.
Can you repost your comment at https://github.com/jhipster/generator-jhipster/issues/12900?

pascalgrimaud:
fine @mshima : I'm commenting the other ticket

PR #12811, opened 2020-10-21
Allow generators/blueprints to force skipGeneratedFlag

sendilkumarn:
Moving the `skip-generated-flag` check during the runtime. Adding a sub-generator for `fu` but this blocks me from generating the application. 

Also it looks like we are adding & checking this flag in the base generator, I think this should be in the server, entity-server, spring-{service, controller} generator only. 

---

### DIFF
--- a/generators/generator-transforms.js
+++ b/generators/generator-transforms.js
@@ -52,7 +52,9 @@ const prettierTransform = function (defaultOptions) {
 const generatedAnnotationTransform = generator => {
     return through.obj(function (file, encoding, callback) {
         if (
-            !generator.options.skipGeneratedFlag &&
+            !generator.configOptions.skipGeneratedFlag &&
+            !file.path.endsWith('package-info.java') &&
+            !file.path.endsWith('MavenWrapperDownloader.java') &&
             path.extname(file.path) === '.java' &&
             file.state !== 'deleted' &&
             !file.path.endsWith('GeneratedByJHipster.java')
--- a/generators/generator-transforms.js
+++ b/generators/generator-transforms.js
@@ -51,7 +51,12 @@ const prettierTransform = function (defaultOptions) {
 
 const generatedAnnotationTransform = generator => {
     return through.obj(function (file, encoding, callback) {
-        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
+        if (
+            !generator.options.skipGeneratedFlag &&
+            path.extname(file.path) === '.java' &&
+            file.state !== 'deleted' &&
+            !file.path.endsWith('GeneratedByJHipster.java')
+        ) {
             const packageName = generator.jhipsterConfig.packageName;
             const content = file.contents.toString('utf8');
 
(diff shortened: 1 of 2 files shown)


mshima (changes requested):
With this change skipGeneratedFlag will not be persisted.
Will require the user to pass --skip-generated-flag every time.

I don't get why this is failing.
Can you give an example?

mshima:
> Also it looks like we are adding & checking this flag in the base generator, I think this should be in the server, entity-server, spring-{service, controller} generator only.

This would create more problems.
- Conflicter is shared between every generator.
- Only the conflicter of the first generator is used.
- generators only register to it's own conflicter.

So:
- if app the generator doesn't register this generator-flag, it will not will be used at all when executing `jhipster`.
- if only app register the transformer, other generators will not use this transformer at all.

See https://github.com/jhipster/generator-jhipster/pull/12711.
Currently `jhipster entity Foo` generates java code without been prettified.

sendilkumarn:
@mshima so for example, I have a new generator say `foo`.  Now when I generate the application, I have the `@GeneratedByJHipster` annotation. But I do not need this annotation.  

Setting the flag inside the `foo` generator do not work because before invoking `foo` generator it is invoked in `base` generator.

sendilkumarn:
> Currently jhipster entity Foo generates java code without been prettified.

I do not want adding annotation but I do need prettier transform. Why `skip-generated-flag` influence the prettifying?

mshima:
> > Currently jhipster entity Foo generates java code without been prettified.
> 
> I do not want adding annotation but I do need prettier transform. Why `skip-generated-flag` influence the prettifying?

Nothing, just using as example since the same principles applies.
Transformers are global, they are applied when committing the file from mem-fs to the disk.

mshima:
> @mshima so for example, I have a new generator say `foo`. Now when I generate the application, I have the `@GeneratedByJHipster` annotation. But I do not need this annotation.

I don't understand the use case when you are saying that you don't want this.
If a `.java` file is written using the generator, it will add the flag.
Why do you want to prevent some `.java` file from having the flag?

mshima:
I saw 2 problem with current implementation:
- adds the flags do maven downloader 
- adds the flag to package-info.java.

We should prevent from adding the flag to:
- java files outside packageFolder.
- package-info.java.

sendilkumarn:
The whole point of fu is moving away from the `annotations` now we are adding them here. I am trying to circumvent that.

generators/generator-transforms.js
```diff
@@ -51,7 +51,12 @@ const prettierTransform = function (defaultOptions) {
 
 const generatedAnnotationTransform = generator => {
     return through.obj(function (file, encoding, callback) {
-        if (path.extname(file.path) === '.java' && file.state !== 'deleted' && !file.path.endsWith('GeneratedByJHipster.java')) {
+        if (
+            !generator.options.skipGeneratedFlag &&
```
mshima, line 55:
```suggestion
            !generator.configOptions.skipGeneratedFlag &&
            !file.path.endsWith('package-info.java') &&
            !file.path.endsWith('MavenWrapperDownloader.java') &&
```

mshima (changes requested):
You should revert the changes to generator-base.js.

And change from `options` to `configOptions`.
`options` is generator specific and will be applied for the entry generator only.
`configOptions` is shared between every generator so you can apply from every generator.

mshima (approved):

sendilkumarn:
Thanks @mshima ! There are a lot of changes!

PR #12818, opened 2020-10-22
[angular] Remove deprecated functionality

### LINKED ISSUE #12813: Remove deprecated functionality from Angular code

I would like to remove deprecated functionality from Angular code. As replacing `TestBed.get` with `TestBed.inject` was bigger task (without using type casting `as unknown` workaround) then I create this ticket to track different PR-s:

- [x] replace `TestBed.get` with `TestBed.inject`

 - [ ] replace other deprecated functionality

 - [ ] apply ESLint rule https://github.com/gund/eslint-plugin-deprecation to avoid further deprecated functionality

As we are hipsters then we should avoid using deprecated functionality.

Starting with PR-s shortly.

kaidohallik:
Closes #12813

I separated changes into 5 commits. Below is the commit history.

`rxjs` `flatMap` is synonym of `mergeMap` and is deprecated now, so replaced `flatMap` with `mergeMap`.  
In this process I noticed that metrics is loaded sequentially: if metrics loading is finished then thread dump loading starts. Now turned this to parallel load.  
After switching metrics to parallel loading noticed that `AuthExpiredInterceptor` is not working correctly if page loads several data at the same time - then first call routes to home page and second call saves home page as page where to return after relogin. This is fixed now.

Replaced deprecated `async` from `@angular/core/testing` with it's replacement `waitForAsync`. Noticed that some test files are missing license. Added license where it was missing. In some places license format was broken, fixed those formattings too.

`rxjs` `combineLatest` with `resultSelector` is deprecated. Transferred to another form.

Fix "Use an observer instead of error/complete callback" deprecation.

Add ESLint deprecation plugin and set deprecation to error level.

---

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/account/activate/activate.component.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/activate/activate.component.ts.ejs
@@ -18,7 +18,7 @@
 -%>
 import { Component, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
-import { flatMap } from 'rxjs/operators';
+import { mergeMap } from 'rxjs/operators';
 
 import { LoginModalService } from 'app/core/login/login-modal.service';
 import { ActivateService } from './activate.service';
@@ -38,7 +38,7 @@ export class ActivateComponent implements OnInit {
     ) {}
 
     ngOnInit(): void {
-        this.route.queryParams.pipe(flatMap(params => this.activateService.get(params.key))).subscribe(
+        this.route.queryParams.pipe(mergeMap(params => this.activateService.get(params.key))).subscribe(
             () => this.success = true,
             () => this.error = true
         );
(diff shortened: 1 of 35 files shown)


mshima:
Not sure we want to set `"deprecation/deprecation": "error"`.
How this will work?
A deprecation rule needs to be added to eslint-plugin-deprecation package, in this case it will only break updates to eslint-plugin-deprecation itself.
Or deprecations are auto detected by eslint-plugin-deprecation and a deprecation from other package will become a breaking change?

kaidohallik:
Deprecations are auto detected by `eslint-plugin-deprecation`. If some package deprecates functionality used by us then CI fails for this package upgrade and this upgrade needs to be handled manually: rewrite code so that deprecated functionality is not used any more.

mshima:
So I'm not in favor of `"deprecation/deprecation": "error"`.
With this setting, a minor release can become a breaking release.

That's the reason of deprecation, it should warned that will break at some point.

- We have too many frameworks (angular, react, vue, spring).
- Unless we have an official maintainer for angular (in this case), we should let minor releases with deprecations to be merged.
- Bugfixes/features is more important than removing deprecations.

kaidohallik:
OK, good point. I removed last commit which added deprecation plugin and adjusted issue title.

mshima:
> OK, good point. I removed last commit which added deprecation plugin and adjusted issue title.

If you want to keep the plugin but as `warning` instead of `error`, it can be nice. (Or useless really don't know).

mshima (approved):
LGTM

DanielFran:
@kaidohallik Is it ready to merge?

kaidohallik:
@DanielFran Yes, it's ready.

kaidohallik:
Thanks for merging @DanielFran !

PR #12829, opened 2020-10-23
Exclude translation of DataAccessException when there's no database

gmarziou:
Fix #12827

---

Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #12830, opened 2020-10-23
Exclude translation of DataAccessException when there's no database

gmarziou:
Backported from main branch to 6.x

Fix #12829

---


pascalgrimaud:
@gmarziou : do you think we should plan a new patch release ? or wait a little bit if we find some other bugs

gmarziou:
@pascalgrimaud  this is a very specific use case and the workaround is documented in the issue. So I don't think a patch release is required only for this bug, let's wait for more important fixes.

Issue #12764, opened 2020-10-15
Drop tslint in favor of eslint.


mshima:
##### **Overview of the feature request**

Migrate to eslint.

##### **Motivation for or Use Case**

```
npm WARN deprecated tslint@6.1.2: TSLint has been deprecated in favor of ESLint. Please see https://github.com/palantir/tslint/issues/4534 for more information.
```

##### **Related issues or PR**

-   [ ] Checking this box is mandatory (this is just to show you read everything)

pascalgrimaud:
The work has started for Vue.js by @atomfrede here: https://github.com/jhipster/jhipster-vuejs/pull/601 but was not finished.

About the other front, I think it's already done but I'm not sure

atomfrede:
I try to integrate the changes in the main generator this weekend. I am pretty sure for angular it is done already not sure about react though.

mshima:
@pascalgrimaud @atomfrede every framework still depends on tslint.
I plan to migrate angular soon.

atomfrede:
Good to know. So we need someone to do the react part 😁

vishal423:
If I remember correctly, last time the migration was done for `angular` and `react`. Though, `angular` code still has dependencies on `tslint` as the angular specific rules were not migrated to eslint by that time. If that transition path is available on angular, then, we can go ahead and migrate those rules to eslint and can remove the `tslint` dependencies

mshima:
I was wrong react doesn't depends on tslint.
Vue and angular are missing.
- angular: https://github.com/angular-eslint/angular-eslint#migrating-from-codelyzer-and-tslint

kaidohallik:
@mshima I would like to handle Angular part if this ok for you.

https://github.com/angular-eslint/angular-eslint is currently in beta but they want to release 1.0.0 this year. As `tslint` is deprecated then I think it's ok to migrate to `angular-eslint` already now. I have used this beta version of `angular-eslint` and for me this is working fine.

I would like to upgrade `@typescript-eslint` from version 2 to version 4. Our current Angular code doesn't meet to recommended rules of version 4, some rule violations are not easy to fix and those rules I turn off in config file.

I also would like to apply some additional rules in addition to the recommended ones.

mshima:
@kaidohallik sure. Just care to not conflict with angular cli PR too much. Or use that branch already.
We should merge it ASAP.

kaidohallik:
[...]

PR #12845, opened 2020-10-25
[angular] Improve linting

kaidohallik:
Follow up to work done for #12764

I compared our current config with ESLint and TypeScript ESLint recommended configs and removed our identical config and also fixed some recommended rules.

Below is commit history.

Sort ESLint rules alphabetically for better readability.

In the latest ESLint + TypeScript ESLint is `no-unused-vars` identical to our current config, so deleting unnecessary identical override.

Fix `@typescript-eslint/no-unsafe-return` errors.

Fix `@typescript-eslint/restrict-template-expressions` errors.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/eslint-recommended.ts and https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js doesn't contain `no-shadow`, so removing `"no-shadow": "off"`.

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/eslint-recommended.ts contains already `"no-var": "error"` and `"prefer-const": "error"` - so removing those overrides.

---

### DIFF
--- a/generators/client/templates/angular/.eslintrc.json.ejs
+++ b/generators/client/templates/angular/.eslintrc.json.ejs
@@ -116,7 +116,6 @@
     "no-console": ["error", { allow: ["warn", "error"] }],
     "no-new-wrappers": "error",
     "no-eval": "error",
-    "no-shadow": "off",
     "no-new": "error",
     "radix": "error",
     "eqeqeq": ["error", "always", { null: "ignore" }],
--- a/generators/client/templates/angular/.eslintrc.json.ejs
+++ b/generators/client/templates/angular/.eslintrc.json.ejs
@@ -105,7 +105,6 @@
     "@typescript-eslint/no-unsafe-assignment": "off",
     "@typescript-eslint/no-unsafe-call": "off",
     "@typescript-eslint/no-unsafe-member-access": "off",
-    "@typescript-eslint/no-unsafe-return": "off",
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "spaced-comment": ["warn", "always"],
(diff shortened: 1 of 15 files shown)


mshima (approved):

kaidohallik:
I found that for oauth2 user.id type is generated as number, should be string, fixing this shortly.

mshima:
@kaidohallik if you want to ignore User id.
It will be handled soon by composite pk.

kaidohallik:
Fixed `user.id` datatype for `oauth2` now. Just in case used function `getUserPkType` also in user-management tests (although it is not generated in `oauth2`).  
Removed [angular] from PR title to launch full tests as now also back-end part changed as using created `getUserPkType` also in back-end.

generators/generator-base-private.js
```diff
@@ -1262,6 +1262,20 @@ module.exports = class JHipsterBasePrivateGenerator extends Generator {
         return pk;
     }
 
+    /**
+     * Return the User entity primary key data type
+     *
+     * @param {any} databaseType - the database type
+     * @param {any} authenticationType - the authentication type
+     */
+    getUserPkType(databaseType, authenticationType) {
```
mshima, line 1271: The approach needed by composite id is quite different.
This function will be useless soon.
I prefer to don’t add it.

kaidohallik: OK, I made minimal changes, changed only `user.model.ts`, rewrote last commit with those minimal changes, restored [angular] in title to run only Angular tests.

PR #12946, opened 2020-11-03
[angular] Use tabWidth=2 and trailingComma and polish templates

kaidohallik:
Polished Angular templates under `src/main/webapp/app`:
* changed `tabWidth` from 4 to 2 because generated code uses `tabWidth=2` and if doing changes and copy-paste code then it's very annoying always to wrap from 2 to 4 and some newer templates are already using `tabWidth=2`
* added trailing commas to templates
* added missing license headers
* polished other noticed things

---

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/account/account.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/account/account.module.ts.ejs
@@ -20,7 +20,6 @@ import { NgModule } from '@angular/core';
 import { RouterModule } from '@angular/router';
 
 import { SharedModule } from 'app/shared/shared.module';
-
 <%_ if (authenticationType === 'session') { _%>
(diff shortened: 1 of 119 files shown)


mshima:
@kaidohallik please don't do this now.

mshima:
A bit too early.
There is no change on entity-client, so this should be safe.

kaidohallik:
OK, I want to do the same also in Angular tests part (only `client` subgenerator), is this also safe?  
`entity-client` part can be done sometime later after you have finished your work.

mshima:
> OK, I want to do the same also in Angular tests part (only `client` subgenerator), is this also safe?
> `entity-client` part can be done sometime later after you have finished your work.

Yes, thanks.

DanielFran (approved):

PR #12970, opened 2020-11-05
Change skipGeneratedFlag to withGeneratedFlag

pascalgrimaud:
After some discussions with @jdubois, we think it's better to not generate this annotation by default.
It will be available by flag + config.

Related to https://github.com/jhipster/generator-jhipster/issues/12459
Following these PRs: https://github.com/jhipster/generator-jhipster/pull/12569 https://github.com/jhipster/generator-jhipster/pull/12811

Is it ok for your blueprint @sendilkumarn ?

---

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -92,7 +92,7 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
         });
 
         this.option('with-generated-flag', {
-            desc: 'Adding a GeneratedByJHipster annotation to all generated java classes and interfaces',
+            desc: 'Add a GeneratedByJHipster annotation to all generated java classes and interfaces',
             type: Boolean,
         });
 
@@ -2362,9 +2362,6 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
         if (options.skipFakeData) {
             this.jhipsterConfig.skipFakeData = true;
         }
-        if (options.withGeneratedFlag) {
-            this.jhipsterConfig.withGeneratedFlag = true;
-        }
         if (options.skipUserManagement) {
             this.jhipsterConfig.skipUserManagement = true;
         }
@@ -2456,7 +2453,6 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
         dest.reactive = config.reactive;
         dest.jhiPrefix = config.jhiPrefix;
         dest.skipFakeData = config.skipFakeData;
-        dest.withGeneratedFlag = config.withGeneratedFlag;
         dest.entitySuffix = config.entitySuffix;
         dest.dtoSuffix = config.dtoSuffix;
         dest.skipUserManagement = config.skipUserManagement;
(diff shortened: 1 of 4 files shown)


DanielFran (approved):

mshima (changes requested):
Will post the review soon

generators/generator-base.js
```diff
@@ -91,8 +91,8 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
             hide: true,
         });
 
-        this.option('skip-generated-flag', {
-            desc: 'Skip adding a GeneratedByJhipster annotation to all generated java classes and interfaces',
+        this.option('with-generated-flag', {
+            desc: 'Adding a GeneratedByJHipster annotation to all generated java classes and interfaces',
```
mshima, line 95: Add instead of adding?

generators/generator-base.js
```diff
@@ -2362,6 +2362,9 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
         if (options.skipFakeData) {
             this.jhipsterConfig.skipFakeData = true;
         }
+        if (options.withGeneratedFlag) {
+            this.jhipsterConfig.withGeneratedFlag = true;
+        }
```
mshima, line 2367: Not needed, we already doing this at `generator-base` constructor.

generators/generator-base.js
```diff
@@ -2453,6 +2456,7 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
         dest.reactive = config.reactive;
         dest.jhiPrefix = config.jhiPrefix;
         dest.skipFakeData = config.skipFakeData;
+        dest.withGeneratedFlag = config.withGeneratedFlag;
```
mshima, line 2459: Not needed, we use this to expose configs to templates and make easier to use.
In this case we only use it at the constructor.

generators/generator-transforms.js
```diff
@@ -52,7 +52,7 @@ const prettierTransform = function (defaultOptions) {
 const generatedAnnotationTransform = generator => {
     return through.obj(function (file, encoding, callback) {
         if (
-            !generator.configOptions.skipGeneratedFlag &&
+            generator.jhipsterConfig.withGeneratedFlag &&
```
mshima, line 55: `!generator.configOptions.skipGeneratedFlag &&` condition is to let blueprints to skip the flag at runtime (post constructor) with `this.configOptions.skipGeneratedFlag = true`.

Since we are defaulting to disabled we can remove the condition.
Forcing true would need to queue the transform after constructor.

A alternative is to keep the condition and always queue the transform (remove condition from constructor).

mshima (changes requested):

generators/generator-transforms.js
pascalgrimaud: agree, let's remove this condition, you're right

mshima (approved):
LGTM

sendilkumarn (approved):
This is what I wanted to do initially. I perfectly fine with that 👍

PR #12985, opened 2020-11-08
Split AccountResourceIT template

murdos:
The template was very big and was in fact the aggregation of 3 templates for oauth2, skipUserManagement and other cases

---


mshima:
For consistency would be nice to have `${option}Filename` or `Filename_${option}`.
This PR uses the second, database changes uses the first.
I prefer the second.

murdos:
> 
> 
> For consistency would be nice to have `${option}Filename` or `Filename_${option}`.
> This PR uses the second, database changes uses the first.
> I prefer the second if it will drop the option from name.

Good point. I agree that the second is better. I'll do another PR that fixes database configuration and also other templates like reactive ones.

mshima (approved):
LGTM

PR #12986, opened 2020-11-08
Split UAA SecurityConfiguration in a distinct template

murdos:
And move security related templates in a dedicated template directory

---


generators/server/files.js
```diff
@@ -845,16 +855,6 @@ const serverFiles = {
                 },
             ],
         },
-        {
-            condition: generator => !generator.reactive && generator.authenticationType === 'oauth2',
```
murdos, line 849: This one was redundant with https://github.com/jhipster/generator-jhipster/blob/9ce17c10f78c8d36a0d034280b1bde43f86d44d1/generators/server/files.js#L450, which means that for a non reactive oauth2 application, the file was written twice.

mshima:
Great work on config files reorganisation.

Not convinced about the security folder though.
IMO it can be more confusing because we mimic destination package.

Same for database config, having `DatabaseConfiguration_cassandra.java.ejs` is already clean enough, don't need `database/DatabaseConfiguration_cassandra.java.ejs`.

murdos:
I've reverted the move to the `security` subdirectory and applied the Filename_${option} naming pattern for SecurityConfiguration.java

mshima (approved):
LGTM

pascalgrimaud:
I love what you're doing @murdos 
It will help a lot for v8, if we decide to use a clean architecture :)

murdos:
@pascalgrimaud: thanks! 
These were a few breaking improvements that I had in mind and wanted to do in a major release, inspired by annoyances or pitfalls that I encountered while working on new features or framework upgrades in generator-jhipster.
I hope indeed that it'll easy maintenance and progress of JHipster 😄

PR #12994, opened 2020-11-08
Fix rewrite when value is undefined

pascalgrimaud:
It is to fix the issue when `lines[otherwiseLineIndex]` is undefined.
I don't know how to reproduce it, but I encounter this issue several times.

This is for securing our code

---

### DIFF
--- a/generators/utils.js
+++ b/generators/utils.js
@@ -142,6 +142,11 @@ function rewrite(args) {
         }
     });
 
+    if (otherwiseLineIndex === -1) {
+        console.warn(`Needle ${args.needle} not found at file ${args.file}`);
+        return args.haystack;
+    }
+
     let spaces = 0;
     while (lines[otherwiseLineIndex].charAt(spaces) === ' ') {
         spaces += 1;


generators/utils.js
```diff
@@ -143,8 +143,10 @@ function rewrite(args) {
     });
```
mshima, line 144: Looking at the code, I would say that the needle was not found:
```
    If (otherwiseLineIndex === -1) {
        console.warn(`Needle ${args.needle} not found at file ${args.file}`);
        return args.haystack;
    }
```

pascalgrimaud: is it a suggested code ? instead of my change ?

mshima: Yes, but if you think it’s a different problem, you can go ahead and merge as it is.

pascalgrimaud: Looking at the code, I think you're right. I'll apply your change tomorrow, unless you want to do it :)
Thanks a lot

mshima: Maybe it’s a good idea to throw instead of just log.
Looks an error to me.

PR #13035, opened 2020-11-12
Angular 11 🔥

qmonmert:
Commands :
``` 
ng update
ng update @angular/cli @angular/core
``` 

---


generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
```diff
@@ -53,7 +53,7 @@ const LAYOUT_ROUTES = [navbarRoute, ...errorRoute];
         <%_ } _%>
         ...LAYOUT_ROUTES,
       ],
-      { enableTracing: DEBUG_INFO_ENABLED }
+      { enableTracing: DEBUG_INFO_ENABLED, relativeLinkResolution: 'legacy' }
```
kaidohallik, line 56: Do we have code which forces us to use `relativeLinkResolution: 'legacy'`?

If yes then can we changes those to use Angular 11 default `relativeLinkResolution: 'corrected'`?

qmonmert: I don't know, it's this command:`ng update @angular/cli @angular/core` to migrate which updates this file

kaidohallik: I think we should migrate from `'legacy'` to `'corrected'`.  
Tests are failing because `DatePipe` has stricter type checking in Angular 11 and code needs to be changed.  
@qmonmert are you planning to work on this PR to get CI green?

qmonmert: yes I'm working on it tomorrow, I have already started (but not push on this PR)

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics-modal-threads/metrics-modal-threads.component.html.ejs
```diff
@@ -17,85 +17,85 @@
...
+    <span class="badge badge-danger hand" (click)="threadStateFilter = ThreadState.Blocked">
+      <fa-icon *ngIf="threadStateFilter === ThreadState.Blocked" icon="check"></fa-icon>
+      Blocked&nbsp;<span class="badge badge-pill badge-default">{{ threadDumpBlocked }}</span>
+    </span>
+  </div>
 
-    <div class="pad" *ngFor="let entry of getThreads() | keyvalue">
-        <h6>
-            <span class="badge" [ngClass]="getBadgeClass(entry.value.threadState)">{{ entry.value.threadState }}</span>
+  <div class="pad" *ngFor="let entry of getThreads()">
```
kaidohallik, line 53: Can you rename variable `entry` to `thread`, because now this is not `Object` entry but `Thread` object.

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics-modal-threads/metrics-modal-threads.component.html.ejs
```diff
@@ -17,85 +17,85 @@
  limitations under the License.
 -%>
 <div class="modal-header">
-    <h4 class="modal-title" <%= jhiPrefix %>Translate="metrics.jvm.threads.dump.title">Threads dump</h4>
+  <h4 class="modal-title" <%= jhiPrefix %>Translate="metrics.jvm.threads.dump.title">Threads dump</h4>
```
kaidohallik, line 20: Currently we have 4 space spacing in HTML files. We should follow the same style.  
If we want to go for 2 space spacing in HTML files then this should be done systematically for all HTML files.

kaidohallik (changes requested):
Great unnecessary pipe removal!  
I have couple of comments.

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics-modal-threads/metrics-modal-threads.component.html.ejs
qmonmert: fix

generators/client/templates/angular/src/main/webapp/app/admin/metrics/metrics-modal-threads/metrics-modal-threads.component.html.ejs
qmonmert: done

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
qmonmert: @kaidohallik do you have an idea to fix `DatePipe` with `Dayjs` ?

kaidohallik: `DayJs`  has https://day.js.org/docs/en/display/format, maybe use this instead of `DatePipe`? If using methods in HTML then it's also good idea to switch to `changeDetection: ChangeDetectionStrategy.OnPush`, to avoid bad performance.

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
```diff
@@ -53,7 +53,7 @@ const LAYOUT_ROUTES = [navbarRoute, ...errorRoute];
         <%_ } _%>
         ...LAYOUT_ROUTES,
       ],
-      { enableTracing: DEBUG_INFO_ENABLED }
+      { enableTracing: DEBUG_INFO_ENABLED, relativeLinkResolution: 'corrected' }
```
ctamisier, line 56: the default is 'corrected' in angular 11, should we just take it out in this case ?

kaidohallik: We need to check, does we have code which relies on `relativeLinkResolution`, if yes then migrate that code and then remove this.

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
```diff
@@ -53,7 +53,7 @@ const LAYOUT_ROUTES = [navbarRoute, ...errorRoute];
         <%_ } _%>
         ...LAYOUT_ROUTES,
       ],
-      { enableTracing: DEBUG_INFO_ENABLED, relativeLinkResolution: 'corrected' }
+      { enableTracing: DEBUG_INFO_ENABLED }
```
ctamisier: squash this commit with the first one of this PR, as relativeLinkResolution has been added and then removed ?

qmonmert: done

generators/client/templates/angular/src/main/webapp/app/app-routing.module.ts.ejs
wmarques: The pipes are pure by default so for me the OnPush strategy is not mandatory, the pipe code will be executed only when the input changes even if the Component is not using the OnPush strategy

kaidohallik: Pipes are replaced by methods, so `OnPush` is needed for good performance.

qmonmert: when I use `changeDetection: ChangeDetectionStrategy.OnPush` on my component, the list of entities does not display

kaidohallik: On data load use `this.changeDetector.markForCheck();` as this is done in `metrics.component.ts`.

qmonmert: yes it works

wmarques: Maybe we can use custom date pipes using dayjs instead of the Angular default ones ? This way we can remove the OnPush and provide our users some dayJS pipes. 
For me formatting an input for the display is why pipes are made for., WDYT ?

That can be done on a separate PR

kaidohallik: For `DayJs` pipes is library available: https://github.com/ljmerza/ngx-dayjs - so if you think we should go for pipes then first need discuss, use this library or not.

wmarques: The lib doesn't look maintained so I won't go for it, I would vote for putting two custom pipes in our shared folder

kaidohallik: Agree with custom pipes in our shared folder, there is already `duration.pipe` for `DayJs`. So I suggest to create `shared/dayjs` subfolder, move `duration.pipe` there and add additional pipes.

qmonmert: ok let's do this

wmarques: Maybe `shared/dates` WDYT ? 😄  For me that should be an abstract thing so we can easily change our date lib in the future

kaidohallik: Yes, this looks better, but maybe not in plural, but `shared/date` like other folders are named?

generators/client/templates/angular/src/main/webapp/app/shared/date/format-medium-date.pipe.ts.ejs
```diff
@@ -0,0 +1,12 @@
+import { Pipe, PipeTransform } from '@angular/core';
```
wmarques, line 1: Can you add few unit tests to check that the date is correctly formatted ? That will help us if we switch lib like we did recently

Same goes for the other pipe 😄

wmarques (changes requested):

ctamisier:
Should these new formats be used in user-management.component.html.ejs ?

qmonmert:
@ctamisier I think there is a bug on the main branch with `user-management.component.html.ejs` because dates are not displayed

generators/client/templates/angular/src/main/webapp/app/shared/date/format-medium-date.pipe.ts.ejs
qmonmert: done

kaidohallik (commented):
Currently dates and datetimes in entity detail view are broken, already before this Angular 11 PR, but can you use those new pipes also there to fix entity detail view?

generators/client/files-angular.js
```diff
@@ -348,7 +348,9 @@ const files = {
             templates: [
                 'shared/shared.module.ts',
                 'shared/shared-libs.module.ts',
-                'shared/duration.pipe.ts',
+                'shared/date/duration.pipe.ts',
+                'shared/date/format-medium-date.pipe.ts',
+                'shared/date/format-medium.pipe.ts',
```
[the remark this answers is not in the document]
qmonmert: done

qmonmert:
> Currently dates and datetimes in entity detail view are broken, already before this Angular 11 PR, but can you use those new pipes also there to fix entity detail view?

done

qmonmert:
> Should these new formats be used in user-management.component.html.ejs ?

@ctamisier I don't think in this PR because `user.createdDate` is a `Date` and not a `dayjs.Dayjs`

qmonmert:
@pascalgrimaud why there is one cancelled checks ? 🤔

pascalgrimaud:
some timeout...

qmonmert:
you can restart the test ?

mshima:
It happens sometimes with that test. Restarted.

qmonmert:
thanks @mshima

kaidohallik (approved):
Now everything LGTM, thanks @qmonmert

qmonmert:
tests are ✅

qmonmert:
@pascalgrimaud I claim my money 😄 https://opencollective.com/generator-jhipster/expenses/28243

pascalgrimaud:
thanks for your work @qmonmert ! approved

mshima:
We should bump the dependency `@angular-builders/custom-webpack@11.0.0-beta.1`

wmarques:
@mshima id rather wait for a stable release

mshima:
@wmarques `@angular-builders/custom-webpack@10` has dependency and imports angular@10.
So we are using a mixed of angular-cli 11 with 10.

I would rather wait for the entire stack to update.
Since we updated to angular 11, IMO a beta supporting 11 is better then a stable supporting 10.

wmarques:
I do not agree here, the builder only add a "wrapper" around the CLI which will be in 11, so for me having the custom-webpack builder in 10 doesn't bother me a lot. 
I don't feel confortable by giving our user a beta dependency, I think the stable 11 will come pretty soon so we should better wait for it.

However I agree that the most simple solution was to wait for the release of the custom-webpack v11 😢

MathieuAA:
We should not use dependencies in beta versions, unless we know for sure there won't be any issues or the risks are measured and taken into account.
If we can wait for the dependency to use angular-cli 11, then let's do it.
The only warning I raise is about our release planning: if we wanna do a beta release, then we should think about whether this matter would be an issue for our release planning.

Ideally, this issue should have been brought up and discussed before merging this PR. These kinds of issues should be taken into account the next time we do something like this.

kaidohallik:
FYI: I made tests with HMR (`ng serve --hmr`) and:
* it works fully after upgrading `@angular-builders` components to `v11.0.0-beta.1`
* after upgrading `@angular-builders` components to `v11.0.0-beta.1` can remove `module.hot` stuff from `app.main.ts`, HMR works still fully
* HMR seems not working or working partially if not upgrading `@angular-builders` to `v11.0.0-beta.1` (our `module.hot` stuff in `app.main.ts` seems to be too simplified for full HMR support)
* in previous versions (Angular 10 and before) HMR seems not working or working partially

So upgrading Angular to v11 was great step, now everybody can switch manually `@angular-builders` dependencies to version `11.0.0-beta.1` and enjoy full HMR support.

PR #13039, opened 2020-11-13
fix heroku reprompting for app name when not found

ruddell:
When a Heroku app is not found, the user is prompted to re-run `jhipster heroku` to create a new deployment.  To enable regenerating the Heroku deployment, the `herokuAppName` config is set to `null`.  

The config is later saved with the value of `this.herokuAppName` in the [`saveConfig` step](https://github.com/jhipster/generator-jhipster/blob/e885becfb626467ce077e5fccb274eaa531255c2/generators/heroku/index.js#L308-L317), rather than `null`.

This commit fixes the behavior.  Now, when an app is not found, re-running `jhipster heroku` prompts you to create a new deployment.  It also skips other steps when finding the app fails (such as editing the pom.xml).

Related to [this StackOverflow question](https://stackoverflow.com/questions/64826441/getting-could-not-find-application-error-when-running-jhipster-heroku)

### DIFF
--- a/generators/heroku/index.js
+++ b/generators/heroku/index.js
@@ -108,13 +108,10 @@ module.exports = class extends BaseBlueprintGenerator {
                 if (this.herokuAppName) {
                     ChildProcess.exec('heroku apps:info --json', (err, stdout) => {
                         if (err) {
-                            this.config.set({
-                                herokuAppName: null,
-                                herokuDeployType: this.herokuDeployType,
-                            });
                             this.abort = true;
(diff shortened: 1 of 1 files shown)


mraible (approved):

mshima (approved):
An alternative for those abort conditions, is to use `cancelCancellableTasks()`.
Like:
https://github.com/jhipster/generator-jhipster/blob/a9c9a7692921cc548a350fd94f72c07f9d716ce1/generators/languages/index.js#L152

Current PR seems to fix the problem.

ruddell:
Will create a new PR for `cancelCancellableTasks` everywhere

Issue #13038, opened 2020-11-13
Avoid N+1 Selects on '*-to-many' relationships


ctamisier:
In `src/main/resources/config/application.yml`,
there is `hibernate.query.fail_on_pagination_over_collection_fetch: true`.
But this option is not in `src/test/resources/config/application.yml`.

Which means we could write a request that will work in a test but will fail in production.
A request of this kind is actually generated (ex: Foo many-to-many Bar):
![image](https://user-images.githubusercontent.com/766263/99065324-82410f00-25a7-11eb-87f8-f3686554e8ae.png)
(Pageable + fetch)

In this case, `@BatchSize` on bars field in Foo (like in User/Authority relation) should be used ? (+ remove the left join)

mshima:
I think that just removing left join for collections should be enough.
A should be fixed by https://github.com/jhipster/generator-jhipster/pull/13021/ and commit https://github.com/jhipster/generator-jhipster/pull/12971/commits/017be84ced6d406190f6c777392591e4a06f3065 from another PR.

ctamisier:
After the DTO improvement merge, we now have N+1 selects (with a Foo many-to-many Bar)
![image](https://user-images.githubusercontent.com/766263/99562230-1f48e100-29c8-11eb-9d20-711149ee3c38.png)

I don't know why but additional requests didn't show the 1st time I ran my app. But I can see them now... (screenshot)
It seems that `show-sql: true` is not the best way to debug SQL requests.. I will investigate.

So for the N+1 selects problem, what about @BatchSize(size = 20) ? (like User <-> Authority)

mshima:
> I don't know why but additional requests didn't show the 1st time I ran my app. But I can see them now... (screenshot)
> It seems that `show-sql: true` is not the best way to debug SQL requests.. I will investigate.

Probably due to caching.

We probably should implement a mix of https://vladmihalcea.com/fix-hibernate-hhh000104-entity-fetch-pagination-warning-message/ and https://vladmihalcea.com/hibernate-multiplebagfetchexception/

Something like
Fetch the entity:
```
List< Post > post = entityManager.createQuery("""
    select distinct p
    from Post p
    order by p.createdOn
    """, Post.class)
.setParameter(...)
.setMaxResults(5)
.setHint(QueryHints.PASS_DISTINCT_THROUGH, false)
.getResultList();
```

And then fetch every bag separated:
```
posts = entityManager.createQuery("""
    select distinct p
    from Post p
[...]

PR #13021, opened 2020-11-10
Improve dto mapper.

mshima:
- look at everything back relationship even if the entity does not contains the forward relationship.
Needed for completeness.
- implements another priority called prepararingRelationship.
at `default` priority, otherRelationships needs to be settled for a complete back reference.
- adds collection mapping.

Related to https://github.com/jhipster/generator-jhipster/issues/12844
Fixes https://github.com/jhipster/generator-jhipster/issues/13036.

---

Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13041, opened 2020-11-14
fix: add hibernate fail pagination option in test config as well

ctamisier:
Related to https://github.com/jhipster/generator-jhipster/issues/13038

Apply `hibernate.query.fail_on_pagination_over_collection_fetch: true` in test config as well.
This will avoid to write working SQL requests (just showing a warning) during tests but failing in production.

### DIFF
--- a/generators/server/templates/src/test/resources/config/application.yml.ejs
+++ b/generators/server/templates/src/test/resources/config/application.yml.ejs
@@ -93,6 +93,7 @@ spring:
             hibernate.generate_statistics: false
             hibernate.hbm2ddl.auto: validate
             hibernate.jdbc.time_zone: UTC
+            hibernate.query.fail_on_pagination_over_collection_fetch: true
     <%_ } _%>
     <%_ if (databaseType === 'mongodb' || databaseType === 'cassandra') { _%>
     data:


mshima (approved):
This is part of the fix, but not the fix itself.

- Owned many-to-many relations still generated with left join.
- Our test cases does not have any entity with owned many-to-many relation and pagination. 
Needs manual test for now.

ctamisier:
> Owned many-to-many relations still generated with left join.

You mentioned that this issue(https://github.com/jhipster/generator-jhipster/pull/13021) is taking care of this ?
(And you started working on this already ?: https://github.com/jhipster/generator-jhipster/commit/017be84ced6d406190f6c777392591e4a06f3065)
So maybe we should wait for all the related PRs affecting relationships to be merged and then check all this (and have a clearer overview, because even after removing left join, we will have the N+1 selects problem)

> Our test cases does not have any entity with owned many-to-many relation and pagination.
> Needs manual test for now.

There is this code https://github.com/jhipster/generator-jhipster/blob/main/generators/entity-server/templates/src/test/java/package/web/rest/EntityResourceIT.java.ejs#L896, which is a kind of an inactive test, that will fail (with the _fail_on_pagination_over_collection_fetch: true_ config) if it was an integration  test. I think we should remove this part (and remove all the eagerload (request parameter) mechanism as well)

So we can safely merge this current PR, right ? (Or am I missing something ?)
And I can create another ticket about handlng properly all this (fetch, pagination, N+1 Selects, eagerload, etc..)
(Except if there are already some PR including these issues that we can wait to be merged)

mshima:
The fix was done in 2 parts.
The second is in the other PR that removes eager when pagination is true at:
https://github.com/jhipster/generator-jhipster/blob/fac6e06cc24dd580f35831b34fe40008913a8ce4/generators/entity/index.js#L595-L599

I agree to remove eagerLoad parameter, let’s keep the bug open to discuss there.

PR #13043, opened 2020-11-15
fix: remove system and anonymoususer

ctamisier:
Remove **system** and **anonymoususer** as they are not used in a generated application.

For _auditing_, the **system** string is used as the editor when the action is not performed by a real connected user (`SpringSecurityAuditorAware.java`).
**system** user was a "loggable" user.

When a _reset password_ action is triggered, the string **anonymousUser** is used ('U' uppercase) and it is not related with the current **anonymoususer** user.

So everything seems safe to remove these 2 default accounts.


DanielFran (approved):
LGTM

PR #12736, opened 2020-10-13
[webflux] Re-enable blockhound and whitelist BoundedElasticScheduler.schedule

murdos:
Fixes #11599

---

Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13053, opened 2020-11-16
Revert "[webflux] Re-enable blockhound and whitelist BoundedElasticScheduler.schedule"

mshima:
Reverts jhipster/generator-jhipster#12736

Fixes webflux builds.


murdos:
What is the error in webflux builds? 
Maybe we need to whilelist another api rather than reverting the whole change

murdos (changes requested):
I disagree with reverting these changes, there are not the real issue IMO and doesn't seem to be related to the BadSqlGrammar errors.

mshima:
> I disagree with reverting these changes, there are not the real issue IMO and doesn't seem to be related to the BadSqlGrammar errors.

I am testing, webflux is failing for a week already. since merging https://github.com/jhipster/generator-jhipster/pull/12736.
There are 2 issues:
- BadSqlGrammar at postgres related to switching to sqlfull
- Mariadb failing with timeout since enabling blackhound see https://github.com/jhipster/generator-jhipster/pull/12736#issuecomment-725710000

Reverting both to test.

mshima:
@murdos `Webflux / webflux-mariadb-gradle` succeeds reverting blackhound.
Can you fix this? A alternative is to disable `Webflux / webflux-mariadb-gradle` test.

murdos:
> 
> 
> @murdos `Webflux / webflux-mariadb-gradle` succeeds reverting blackhound.
> Can you fix this? A alternative is to disable `Webflux / webflux-mariadb-gradle` test.

I'll try to take a look, but I'm not a reactive expert.
Meanwhile I think we should indeed disable `Webflux / webflux-mariadb-gradle` test and open a ticket to trace the issue, since having blockhound for other webflux builds brings some value by detecting blocking calls that could be introduced.

mshima:
Closing in favor of https://github.com/jhipster/generator-jhipster/pull/13071

PR #13073, opened 2020-11-17
[angular] Change delete dialog component to receive event name to bro…

DanielFran:
…adcast

This allow to reuse the delete component on other pages.
It simplifies also the definition of the event name, since it is defined only in the component that trigger the delete dialog.

---

### DIFF
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management-delete-dialog.component.ts.ejs
@@ -28,6 +28,7 @@ import { EventManager } from 'app/core/event-manager/event-manager.service';
 })
 export class <%= entityAngularName %>DeleteDialogComponent {
     <%= entityInstance %>?: I<%= entityAngularName %>;
+    eventName?: string;
 
     constructor(
         protected <%= entityInstance %>Service: <%= entityAngularName %>Service,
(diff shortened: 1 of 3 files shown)


kaidohallik (commented):
I understand that the delete component needs to be used somewhere else beside the list component.  
This seems a good solution for that, consumer has full control over message sent.  
Possible alternative is just to change event name to reflect better which action happened, name event for example as `'<%= entityInstance %>Delete'`. And multiple components can subscribe to this event without restrictions.

mshima:
A callback is cleaner in my opinion.
Instance field:
```
onDelete?: () => void;
```

Just before closing the callback
```
this.onDelete?.();
```

Move event broadcast to the callback:
```
modalRef.componentInstance.onDelete = () => {
  this.eventManager.broadcast('<%= entityInstance %>ListModification');
};
```
at:
https://github.com/jhipster/generator-jhipster/blob/3dcf7f29d884523e6c734245dc869ef90143e969/generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management.component.ts.ejs#L96

wmarques:
For me the solution is ok, a callback is indeed more generic but a little more complex to use for beginners. So I would stick with @DanielFran version, WDYT ?

wmarques:
After thinking of it, I would rather go on a callback or an output, it's a more generic way

mshima:
> After thinking of it, I would rather go on a callback or an output, it's a more generic way.
> We can also use the .then to check when the modal is closed, cf
> 
> ```
>   this.modalService.open(content).result.then((result) => {
>       this.closeResult = `Closed with: ${result}`;
>     }, (reason) => {
>       this.eventManager.broadcast('myEvent');
>     });
> ```
> 
> WDYT @mshima

Yep, output is better, is the built-in callback 😄 .

DanielFran:
Thanks for the feedback, will apply changes from @kaidohallik and @wmarques

DanielFran:
Superseded by #13111

PR #13083, opened 2020-11-18
[angular] Fix Protractor tests in Angular

kaidohallik:
Daily builds using Angular with Protractor are failing because of #13075  
For example https://github.com/hipster-labs/jhipster-daily-builds/runs/1417030608  
Reason: Jest and Mocha types are conflicting.  
This PR puts `skipLibCheck` back if Protractor tests are used which fixes Protractor tests in Angular.

---

### DIFF
--- a/generators/client/templates/angular/tsconfig.json.ejs
+++ b/generators/client/templates/angular/tsconfig.json.ejs
@@ -27,6 +27,9 @@
     "strict": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    <%_ if (protractorTests) { _%>
+    "skipLibCheck": true,
+    <%_ } _%>
     "outDir": "<%= DIST_DIR %>app",
     "lib": ["es7", "dom"],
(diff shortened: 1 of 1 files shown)


generators/client/templates/angular/tsconfig.json.ejs
```diff
@@ -27,6 +27,9 @@
     "strict": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    <%_ if (protractorTests) { _%>
```
mshima, line 30: Probably we should add a description for the reason.
```suggestion
    <%_ if (protractorTests) { // Prevents Jest and Mocha types conflicts. See https://github.com/jhipster/generator-jhipster/pull/13083 _%>
```

kaidohallik: It's good idea, I was afk and just 1 minute before clicking suggestion accept you merged this :) I'm doing new PR to add this comment here.

PR #13091, opened 2020-11-19
[angular] Remove core.module

kaidohallik:
Angular documentation doesn't mention `CoreModule` any more, all configuration is performed in `AppModule`. This is mentioned also in https://github.com/jhipster/generator-jhipster/pull/10521#issuecomment-538849407  
As after using interceptors by barrel and removing `ng-jhipster`, the size of the `CoreModule` is reduced then moving this to `AppModule` doesn't make this too big.

---


kaidohallik:
Some more thoughts about this change:
* some time ago we had configuration randomly split between `AppModule` and `CoreModule`, #10360 moved all to `CoreModule`
* `CoreModule` is not used in Angular documentation any more, so this is legacy object and we should not promote legacy principles
* if we have both `AppModule` and `CoreModule` then users can end with the same mess as we had some time ago - they can put configuration to `AppModule` (as Angular documentation suggests) and configuration is then split into 2 different places

DanielFran (approved):
LGTM

PR #13107, opened 2020-11-20
[angular] Clean up csrf.service

kaidohallik:
Follow up to #13098

As CsrfService is now used only in very specific case then we need to clean it up for most cases.

---

### DIFF
--- a/generators/cleanup.js
+++ b/generators/cleanup.js
@@ -118,6 +118,7 @@ function cleanupOldFiles(generator) {
             generator.removeFile(`${ANGULAR_DIR}shared/login/login.component.ts`);
             generator.removeFile(`${ANGULAR_DIR}shared/login/login.component.html`);
             generator.removeFile(`${ANGULAR_DIR}core/auth/user-route-access-service.ts`);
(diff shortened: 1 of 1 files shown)


generators/cleanup.js
```diff
@@ -118,6 +118,7 @@ function cleanupOldFiles(generator) {
             generator.removeFile(`${ANGULAR_DIR}shared/login/login.component.ts`);
             generator.removeFile(`${ANGULAR_DIR}shared/login/login.component.html`);
             generator.removeFile(`${ANGULAR_DIR}core/auth/user-route-access-service.ts`);
+            generator.removeFile(`${ANGULAR_DIR}core/auth/csrf.service.ts`);
```
mshima, line 121: You shouldn’t do this way, 2 options here:
- remove when not needed
- use `this.fs.delete(this.destinationPath(...));`

Otherwise any change to the file will be overwritten every regeneration.

mshima: cc @kaidohallik

kaidohallik: OK, adding condition, will do PR shortly.

PR #13096, opened 2020-11-19
[angular] Add multiple subscription feature to event manager

kaidohallik:
Closes #12778

---

Please make sure the below checklist is followed for Pull Requests.

-   [x] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13111, opened 2020-11-21
[angular] Reduce event manager importance and usage

kaidohallik:
This is follow up to #13096 - after thinking deeper about event manager and using ideas from #13073.

As there is not possible to apply strict type checking if using event manager then this usage should not be encouraged. So moved this to `util` subfolder instead of separate folder and put model to the same file where is service to reduce event manager visibility and importance.
Added also strict types instead of `any` to event manager. Now consumers must do type cast - so if they use event manager they take a conscious risk.
Added block "unreleased files and folders cleanup for v7 developers" to cleanup to delete files if developers regenerate with newer version in v7 development cycle.

This PR has 3 commits:
* move event manager to new location
* remove event manager from user delete (also simplifies user delete and list components)
* remove event manager from entity delete (also simplifies entity delete and list components)

@DanielFran : last commit in this PR is related to your PR #13073 - it basically implements suggestion from https://github.com/jhipster/generator-jhipster/pull/13073#issuecomment-729608862, if you disagree with last commit and want to do changes yourself then I can drop last commit and let this to you.

---


DanielFran:
@kaidohallik no problem, will close the other PR.

mshima (approved):
LGTM

generators/entity-client/templates/angular/src/main/webapp/app/entities/entity-management.component.ts.ejs
```diff
@@ -16,14 +16,14 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-import { Component, OnInit, OnDestroy } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
 import { <%_ if (pagination !== 'no') { _%>HttpHeaders, <% } %>HttpResponse } from '@angular/common/http';
 <%_ if (pagination === 'pagination') { _%>
 import { ActivatedRoute, Router } from '@angular/router';
+import { combineLatest } from 'rxjs';
```
qmonmert, line 23: this import seems unused

mshima: Looks right. Wondering why eslint didn’t complained.

kaidohallik: This is used in `pagination-template`, so this import is needed and is under right condition.  
This is migrated from deleted row below.

qmonmert: indeed ;)

DanielFran (approved):
LGTM

PR #13117, opened 2020-11-22
refactor: use a logger to debug SQL

ctamisier:
Use a logger to debug SQL requests.
`spring.jpa.show-sql` writes to standard output.

(https://thorben-janssen.com/hibernate-logging-guide)

### DIFF
--- a/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
@@ -35,6 +35,7 @@ logging:
     level:
         ROOT: DEBUG
         tech.jhipster: DEBUG
+        org.hibernate.SQL: DEBUG
         <%_ if (packageName !== 'tech.jhipster') { _%>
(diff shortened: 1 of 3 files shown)


DanielFran:
@ctamisier Should we also add org.hibernate.type.descriptor.sql?
_If you also want to get this information, you have to set the log level of org.hibernate.type.descriptor.sql to DEBUG_
````yml
org.hibernate.type.descriptor.sql: TRACE
````

ctamisier:
I would say what we have now is enough. the `org.hibernate.type.descriptor.sql: TRACE` logs the bound parameters and it's quite verbose as it's one line per parameters...
I check the current DEBUG logs to see the shape of the SQL requests and how many there are. I'm not much interested for the actual values.
But this is: IMHO :)

DanielFran (approved):
LGTM.

@ctamisier Maybe just remove the show-sql property since the default is false?

ctamisier:
Yes, indeed, I removed it in all configuration files.

PR #13143, opened 2020-11-25
Revert "[react] Fix default values in entity form"

ctamisier:
Simplify code, this part is no longer required.
Now, the use of normalized DTO ensures that a User is defined (linked to the entity) or not.
There is no "login" field in the handled Entity anymore that can be empty.

(The issue was only when DTO are used).
initial issue: https://github.com/jhipster/generator-jhipster/issues/12784

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
@@ -206,17 +206,10 @@ export const <%= entityReactName %>Update = (props: I<%= entityReactName %>Updat
         <%_ }) _%>
       <%_ } _%>
 
-      const cloneEntity = Object.assign({}, entity);
-      for (const property in entity) {
-        if (entity[property] === '') {
-          cloneEntity[property] = null;
-        }
-      }
-
       if (isNew) {
-        props.createEntity(cloneEntity);
+        props.createEntity(entity);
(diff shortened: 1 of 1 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
```diff
@@ -206,17 +206,10 @@ export const <%= entityReactName %>Update = (props: I<%= entityReactName %>Updat
...
-      const cloneEntity = Object.assign({}, entity);
-      for (const property in entity) {
-        if (entity[property] === '') {
-          cloneEntity[property] = null;
-        }
-      }
-
       if (isNew) {
-        props.createEntity(cloneEntity);
+        props.createEntity(entity);
```
qmonmert, line 210: the ticket https://github.com/jhipster/generator-jhipster/issues/12784 continue to work ? if yes, you can remove `cloneEntity` above

ctamisier: Yes it keeps working (Tried with oauth2 and keycloak, and tried before and after your fix).
Is there still a `cloneEntity` that I didn't remove ?

qmonmert: nope it's good :)

Issue #13125, opened 2020-11-23
Follow closer Angular Style Guide


kaidohallik:
### **Overview of the feature request**

In some important parts we don't follow Angular Style Guide. In this issue I propose 3 enhancement to follow closer Angular Style Guide (and therefore make generated application easier to develop).

#### 1. Move models away from shared folder

We generate entity models to `shared/model` folder. This violates against several rules in Angular Style Guide:
* https://angular.io/guide/styleguide#locate - code location should be intuitive, simple and fast - currently if I work with entity and want to look into it's model file then I must do extra effort to find it because it's not under entity folder
* https://angular.io/guide/styleguide#shared-feature-module - folder `shared` is for shared module this should contain only shared components, directives and pipes - so models doesn't fit here

My suggestion is:
1. generate entity models into `app/entities/<entity-name>/<entity-name>.model.ts`
2. for enums use `app/entity-enums`

#### 2. Polish template and style references

Follow https://angular.io/guide/styleguide#extract-templates-and-styles-to-their-own-files:
1. extract those inline templates to separate files where there is more than 3 lines in template (there are some, can find with `"@angular-eslint/component-max-inline-declarations": "error"`, in tests maybe should use inline templates even if more than 3 lines)
2. rename style files to satisfy format `[component-name].component.css`
3. in relative URLs specify `./` (references to style files are missing this), can apply `@angular-eslint` rule `"@angular-eslint/relative-url-prefix": "error"` to ensure that

#### 3. Move unit tests next to files they are testing

We have unit tests under `src/test/javascript/spec`, this violates with https://angular.io/guide/testing#place-your-spec-file-next-to-the-file-it-tests and I have ran into all those problems described there if test is not next to the file it tests.

So my suggestion is:
1. move tests next to the files they are testing as Angular Style Guide suggests (and as Angular CLI generates tests)
2. entities case: as entities folder is then too big then it's good idea to add 4 subfolders under entity (in entity root folder stay service, route, module and model files):
    - list
    - detail
    - update
    - delete
3. delete folder `src/test/javascript/spec` after that
4. move `jest.conf` to application root folder

### **Motivation for or Use Case**

If we follow Angular Style Guide then:
* our application is easier to develop
* if developers generate new classes using Angular CLI then content generated by Angular CLI is similar to content generated by JHipster

-   [x] Checking this box is mandatory (this is just to show you read everything)

mshima:
I agree.
[...]

PR #12939, opened 2020-11-02
Refactor angular folder structure

kaidohallik:
I have tried to clean up Angular folder structure in #10340 and #10521 but was rejected to wait for major version. So now it's right time to try again.

Below is general description about this refactoring.

* Currently config is present in 4 different places, besides `config` in root folder, also in `blocks`, `core` and `shared` - as this is very confusing then moving all config under `core/config` folder.
* @vishal423 mentioned in https://github.com/jhipster/generator-jhipster/pull/10521#issuecomment-538849407 that core module is no more recommended, but Angular style guide just says that construct logical folder structure. If moving folders from core to root folder then there is too much folders in root, so I kept core folder and kept core functionality under this folder (also Angular Style Guide sample folder structure contains core folder). In this PR I left `core.module` as this is. It can be considered to move all functionality from `core.module` to `app.module` and delete `core.module` file in another PR if someone thinks it's better.
* Deleting `ProdConfig` and putting it's content to `app.main.ts` (this block is generated into the main file also by Angular CLI).
* Moving interceptors from `blocks` folder to `core` folder - after that all is moved from `blocks` and `blocks` is deleted.
* Under `shared` folder if subfolder contains only 1 file then remove folder and move file directly to `shared` folder - improves readability a lot.
* Move `request-util` from `shared` to `core` as this is not contained in shared module but instead is core functionality. Also move request model definitions into separate file as this gives a better overview and is recommended. I put model definitions into `request-util.ts` file while dealing with #10631 because adding new files can be considered as a breaking change which can be done only in major version.
* `i18n.constants.ts` is not used any more - so just deleted that

So higher level file structure is now:
```
account
admin
core
  auth
  config
  interceptor
  request
  tracker
  user
entities
home
layouts
login
shared
```

---

Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13150, opened 2020-11-29
[angular] Revert using authorities as model and move config to root

kaidohallik:
This was discussed in #13125

This PR has 2 commits.

* First commit reverts https://github.com/jhipster/generator-jhipster/commit/a17093d490539c8916805a4cdf540d862885433f because:
    - authorities enum is not read from server, this is manually filled by developer and may not contain all authorities
    - authorities can even be calculated dynamically in server while authenticating user

* Second commit moves config from `core` folder to root folder because:
    - while working with #12939 decided to put config (which was in 4 different places) into `core` folder instead of root folder because config was mainly referenced in `core.module`
    - but #13091 removed `core.module` and now config fits better to root folder
    - if wanting to keep `core` folder as immutable as possible then this step is also needed, see https://github.com/jhipster/generator-jhipster/issues/13125#issuecomment-733894475 for final conclusion about that 

---


mshima (approved):
LGTM

Issue #7112, opened 2018-02-12
Composite primary key support


yelhouti:
##### **Overview of the issue**

Hi guys, I'm working on the support of composite primary keys in jhipster. the goal is to allow use cases where we for complex table relationships...
This thread will help follow advancement of the work and discuss design choices that need to be made.
I'll try to put different questions here, and discuss them in the comments:
1. How to specify primary keys in the entity.json
2. How to handle urls: GET /id1/id2/id3, vs /id1-id2-id3 ...  
3. How to handle Create vs Update (the keys is always specified since there is no AUTOINCREMENT...)

##### **Motivation for or Use Case**

When we want a many to many relationship with additional fields or more than two tables we need a lot of boiler plate code.

##### **Environment and Tools**

java version "1.8.0_161"
Java(TM) SE Runtime Environment (build 1.8.0_161-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.161-b12, mixed mode)

git version 2.14.3 (Apple Git-98)

node: v6.10.3

npm: 5.6.0

bower: 1.7.7

yeoman: 2.0.0

Docker version 17.12.0-ce, build c97c6d6

docker-compose version 1.18.0, build 8dd22a9

4.14
##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "promptValues": {
      "packageName": "com.xxx.yyy.zzz",
      "nativeLanguage": "en"
    },
    "jhipsterVersion": "4.14.0",
    "baseName": "zzz",
    "packageName": "com.xxx.yyy.zzz",
    "packageFolder": "com/xxx/yyy/zzz",
    "serverPort": "8081",
    "authenticationType": "oauth2",
    "cacheProvider": "hazelcast",
    "enableHibernateCache": true,
    "websocket": false,
    "databaseType": "sql",
    "devDatabaseType": "mysql",
    "prodDatabaseType": "mysql",
    "searchEngine": false,
    "messageBroker": false,
    "serviceDiscoveryType": "eureka",
    "buildTool": "gradle",
    "enableSocialSignIn": false,
    "enableSwaggerCodegen": false,
    "jwtSecretKey": "replaced-by-jhipster-info",
    "enableTranslation": true,
    "applicationType": "microservice",
    "testFrameworks": [],
    "jhiPrefix": "jhi",
    "nativeLanguage": "en",
[...]

PR #13173, opened 2020-12-01
[jdl] Don't ignore fields with id name.

mshima:
- Adopt field with id name as `@Id`

Related to https://github.com/jhipster/generator-jhipster/issues/7112

---


MathieuAA:
What about the one JHipster handles by default? If a user writes a JDL file with an id field present in an entity, what happens? Will it collide with the default one or be taken as default instead?

mshima:
> What about the one JHipster handles by default? If a user writes a JDL file with an id field present in an entity, what happens? Will it collide with the default one or be taken as default instead?

It will be taken as default. Done in the second commit.
But custom id support is limited right now.

MathieuAA (approved):

PR #13181, opened 2020-12-02
[angular] Fix UserRouteAccessService

kaidohallik:
This PR fixes the following problems related to `UserRouteAccessService` and it's usages.

Some routes require `ROLE_USER` role, but data shown in those routes and operations done there are allowed in back end without role restriction.
Developer can assume that if route is protected in front end then data and operations related to this route should be protected also in back end and this can lead to security issues.
This PR removes role restrictions from such routes in the front end.

If developer defines routes without restricting those to some roles then `UserRouteAccessService` allowed unauthenticated users to go to that page (but page was empty).
Correct is to route to login page and after successful login user is landing in requested page as this is after this PR.

---


mshima (approved):
LGTM

PR #13190, opened 2020-12-03
Do not wait 4 seconds and then display error if ui is loaded before.

lionnel:
---


DanielFran:
@lionnel Can you sign the CLA please?

mshima:
This PR fixes any bug?
In my opinion we should clear the timeout at page destroy instead.

mshima (changes requested):
`jhipster-error` exists at creation and may be removed from DOM after the application is loaded.
If that happens `checkStillError` will be executed forever.

If we need to resolve `setTimeout` before the timeout for some reason, we should should use `clearTimeout`.

IMO we should close this PR.

mshima:
Closing.
This PR doesn’t seems improve current implementation.

PR #13194, opened 2020-12-04
Improve page title logic for Angular client

kaidohallik:
Adds possibility to add page title through needle API to Angular client's entity and admin routes.

Uses this possibility to add page title to entity route in entity generation and removes page title 4-time copy from entity child routes.

---

### DIFF
--- a/generators/entity-client/files.js
+++ b/generators/entity-client/files.js
@@ -455,12 +455,14 @@ function writeFiles() {
             let clientMainSrcDir;
             let templatesDir;
             let microserviceName = this.microserviceName;
+            let pageTitle;
 
             if (this.clientFramework === ANGULAR) {
                 files = angularFiles;
                 clientMainSrcDir = ANGULAR_DIR;
                 templatesDir = CLIENT_NG2_TEMPLATES_DIR;
                 microserviceName = this.microserviceName;
+                pageTitle = this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural;
             } else if (this.clientFramework === REACT) {
                 files = reactFiles;
                 clientMainSrcDir = REACT_DIR;
@@ -497,7 +499,8 @@ function writeFiles() {
                     this.entityUrl,
                     this.clientFramework,
                     microserviceName,
-                    this.readOnly
+                    this.readOnly,
+                    pageTitle
                 );
                 this.addEntityToMenu(
                     this.entityStateName,
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -429,6 +429,7 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
      * @param {string} clientFramework - The name of the client framework
      * @param {string} microServiceName - Microservice Name
      * @param {boolean} readOnly - If the entity is read-only or not
+     * @param {string} pageTitle - The translation key or the text for the page title in the browser
      */
     addEntityToModule(
         entityInstance,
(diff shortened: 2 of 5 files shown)


DanielFran:
@kaidohallik trusting you on this

generators/entity-client/files.js
```diff
@@ -455,12 +455,14 @@ function writeFiles() {
...
             let templatesDir;
             let microserviceName = this.microserviceName;
+            let pageTitle;
 
             if (this.clientFramework === ANGULAR) {
                 files = angularFiles;
                 clientMainSrcDir = ANGULAR_DIR;
                 templatesDir = CLIENT_NG2_TEMPLATES_DIR;
                 microserviceName = this.microserviceName;
+                pageTitle = this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural;
```
mshima, line 465: Not needed

generators/entity-client/files.js
```diff
@@ -497,7 +499,8 @@ function writeFiles() {
                     this.entityUrl,
                     this.clientFramework,
                     microserviceName,
-                    this.readOnly
+                    this.readOnly,
+                    pageTitle
```
mshima, line 503: Not needed

generators/generator-base.js
```diff
@@ -439,17 +440,17 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
         entityUrl,
         clientFramework,
         microServiceName,
-        readOnly
+        readOnly,
+        pageTitle
```
mshima, line 444: ```pageTitle = this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural```

generators/generator-base.js
```diff
@@ -488,9 +489,10 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
      * @param {string} route - The route for the module. For example 'entity-audit'.
      * @param {string} modulePath - The path to the module file. For example './entity-audit/entity-audit.module'.
      * @param {string} moduleName - The name of the module. For example 'EntityAuditModule'.
+     * @param {string} pageTitle - The translation key or the text for the page title in the browser. For example 'entityAudit.home.title' or 'Entity audit'.
      */
-    addAdminRoute(route, modulePath, moduleName) {
-        this.needleApi.clientAngular.addAdminRoute(route, modulePath, moduleName);
+    addAdminRoute(route, modulePath, moduleName, pageTitle) {
```
mshima, line 494:
```sugestion
addAdminRoute(route, modulePath, moduleName, pageTitle = moduleName);
```

mshima (commented):
Make easier to modules/blueprints and other client frameworks to migrate or adopt.

generators/entity-client/files.js
kaidohallik: You are right, changes here not needed, because in `generator-base.js` in  `addEntityToModule` entity context is available. But question is: why we at all need all those parameters to `addEntityToModule`. Maybe it would be easier to have `addEntityToModule` with zero parameters and read all other information from entity context also there?
Why to act differently with `pageTitle` parameter?
Anyway, for now reverting changes here and using
```
this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural
```
in internal Angular method in PR coming shortly.

generators/generator-base.js
kaidohallik: Can't do this way because if i18n is enabled then this is translation key and this shows `translation-not-found` error in page title.
It should be undefined by default as this is now - then general page title is shown, which is better than `translation-not-found`.
Improving parameter comment in coming PR to give that additional information there.

generators/entity-client/files.js
mshima: > why we at all need all those parameters to addEntityToModule

This is actually a code smell, it should be an object instead.

> Maybe it would be easier to have addEntityToModule with zero parameters and read all other information from entity context also there?

context could be an optional parameter (defaults to this.context), but not sure context contains every corresponding parameters.
It's important to don't make `this.context` mandatory so modules/blueprints can add modules without need to create a this.context.

generators/generator-base.js
mshima: `this.enableTranslation ? undefined : moduleName`?

kaidohallik: I prefer not to add this, `moduleName` can be technical term, so better to show global page title rather than technical term.

generators/entity-client/files.js
kaidohallik: In #13198 gave default values to all `addEntityToModule` parameters so that here can use zero parameters and modules/blueprints can give all parameters - so use without creating a context.

PR #13195, opened 2020-12-04
Add support to `@MapstructExpression`.

mshima:
- Improve angular readonly fields support (mapstruct expression are readonly).
- Add support to dto only fields (transient fields doesn't generate an Entity field).
- Add support to `@MapstructExpression`.

Motivation:
`otherEntityField` (registry representation) fallback is the `id` field.
Composite id fallback is more than 1 field that would require changes to client handling.
`@MapstructExpression` allow us to map every id fields into 1 field to represent the registry.

Related to https://github.com/jhipster/generator-jhipster/issues/7112

---

### DIFF
--- a/generators/database-changelog-liquibase/index.js
+++ b/generators/database-changelog-liquibase/index.js
@@ -52,13 +52,17 @@ module.exports = class extends BaseGenerator {
                 }
 
                 if (databaseChangelog.type === 'entity-new') {
-                    this.fields = this.entity.fields.map(field => prepareFieldForLiquibaseTemplates(this.entity, field));
+                    this.fields = this.entity.fields
+                        .filter(field => !field.transient)
+                        .map(field => prepareFieldForLiquibaseTemplates(this.entity, field));
                 } else {
                     this.addedFields = this.databaseChangelog.addedFields
                         .map(field => prepareFieldForTemplates(this.entity, field, this))
+                        .filter(field => !field.transient)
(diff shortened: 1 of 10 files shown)


murdos:
Are you proposing a `@MapstructExpression` JDL annotation?

mshima:
@murdos yes, but the most important is the feature behind it.
It will allow to map multiple ids into 1 field for selects a alternative is to make otherEntityField required for composite id.

pascalgrimaud:
@mshima: a more global comment. I'd prefer having scripts shell in file, instead of hard coding command in the CI.
The reason why it was done like this:
- it will be easy to switch between CI: Travis CI, Azure Pipelines, GitHub Actions
- it's easy to copy/paste the CI, and override the shell files for another blueprint, for jhipster-bom, etc

With these changes, UAA can't be build any more in the CI with GitHub Actions.
I know, it's still in discussion for removing it or not... but not decided yet.

mshima:
@pascalgrimaud created another PR with workflow changes.

- UAA support was not removed from the script, just from workflow.
Since the workflow here contains diff, and diff is not portable due to PR requirement, it felt like there was no problem to use `jhipster --with-entities` instead of 12-generate-project.sh.
- Dropped UAA diff support for the simple reason that it's not portable and we are not using it here.

Anyway those comments are fixed at https://github.com/jhipster/generator-jhipster/pull/13204.

mshima:
It's open for long enough and don't affects existing features.
Merging.

PR #13198, opened 2020-12-04
Simplify addEntityToModule and document better addAdminRoute.pageTitle

kaidohallik:
Follow up to #13194 - changes after review from @mshima

---

### DIFF
--- a/generators/generator-base.js
+++ b/generators/generator-base.js
@@ -432,16 +432,16 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
      * @param {string} pageTitle - The translation key or the text for the page title in the browser
      */
     addEntityToModule(
-        entityInstance,
-        entityClass,
-        entityName,
-        entityFolderName,
-        entityFileName,
-        entityUrl,
-        clientFramework,
-        microServiceName,
-        readOnly,
-        pageTitle
+        entityInstance = this.entityInstance,
+        entityClass = this.entityClass,
+        entityName = this.entityAngularName,
+        entityFolderName = this.entityFolderName,
+        entityFileName = this.entityFileName,
+        entityUrl = this.entityUrl,
+        clientFramework = this.clientFramework,
+        microServiceName = this.microServiceName,
+        readOnly = this.readOnly,
+        pageTitle = this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural
     ) {
         if (clientFramework === ANGULAR) {
             this.needleApi.clientAngular.addEntityToModule(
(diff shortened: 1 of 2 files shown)


generators/generator-base.js
```diff
@@ -440,8 +439,7 @@ module.exports = class JHipsterBaseGenerator extends PrivateBase {
         entityUrl,
         clientFramework,
         microServiceName,
-        readOnly,
-        pageTitle
+        readOnly
```
mshima, line 442: Keep pageTitle parameter for modules and blueprints.

mshima: But default to `this.enableTranslation ? `${this.i18nKeyPrefix}.home.title` : this.entityClassPlural`.

kaidohallik: OK, keeping `pageTitle` parameter and giving default value to all parameters so that `addEntityToModule` can be called with zero parameters in `entity-client` subgenerator.

mshima (approved):
LGTM

pascalgrimaud:
@mshima @kaidohallik : all the JDL builds failed, with this stacktrace:

```
src/main/webapp/app/entities/entity-routing.module.ts:20:88 - error TS2339: Property 'ProductRoutingModule' does not exist on type 'typeof import("/home/runner/app/store/src/main/webapp/app/entities/crm/product/product-routing.module")'.

20         loadChildren: () => import('./crm/product/product-routing.module').then(m => m.ProductRoutingModule),
                                                                                          ~~~~~~~~~~~~~~~~~~~~
src/main/webapp/app/entities/entity-routing.module.ts:25:90 - error TS2339: Property 'CustomerRoutingModule' does not exist on type 'typeof import("/home/runner/app/store/src/main/webapp/app/entities/crm/customer/customer-routing.module")'.

25         loadChildren: () => import('./crm/customer/customer-routing.module').then(m => m.CustomerRoutingModule),
```

Can one of you have a look plz?
See:
- https://github.com/hipster-labs/jhipster-daily-builds
- https://github.com/hipster-labs/jhipster-daily-builds/runs/1505649488?check_suite_focus=true

kaidohallik:
`microserviceName` is missing from `this` in `generator-base.js` but is available in `files.js` (other variables are available in both places). So reverting partially this PR in the new PR shortly.

mshima:
That's strange.
Not sure why this happened.

kaidohallik:
Now I checked one more time and problem is that the parameter name is `microServiceName` but variable in `this` is `microserviceName` and I used variable name from `this` with uppercase S.
Should I create new simplification PR where correct variable name `microserviceName` is used?

kaidohallik:
I'll create PR where I rename also all other `microServiceName` occurrences to `microserviceName` to avoid such issues in the future.

Issue #13182, opened 2020-12-02
Can't save 2 entities consecutively (React + manytomany)


ctamisier:
##### **Overview of the issue**

Can't save 2 entities consecutively
Fails in the browser with 
```
Uncaught (in promise) TypeError: Cannot read property 'filter' of undefined
    at mapIdList (entity-utils.ts:23)
    at Object.saveEntity [as onSubmit] (foo-update.tsx:63)
    at AvForm._callee$ (AvForm.js:152)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:296)
    at Generator.prototype.<computed> [as next] (runtime.js:114)
    at step (asyncToGenerator.js:17)
    at asyncToGenerator.js:28
```

##### **Reproduce the error**

Create a Foo, and immediately after, create a new one.
On 2nd save, it fails

##### **JHipster Version(s)**

```
clement-2@0.0.1-SNAPSHOT /Users/ctamisier/projects/clement2
└── (empty)

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "blueprints": [],
    "otherModules": [],
    "applicationType": "monolith",
    "baseName": "clement2",
    "jhipsterVersion": "6.10.5",
    "skipClient": false,
    "skipServer": false,
    "skipUserManagement": false,
    "skipCheckLengthOfIdentifier": false,
    "skipFakeData": false,
    "jhiPrefix": "jhi",
    "entitySuffix": "",
    "dtoSuffix": "DTO",
    "testFrameworks": [],
    "pages": [],
    "creationTimestamp": 1606926921414,
    "promptValues": {
      "packageName": "com.mycompany.myapp",
      "nativeLanguage": "en"
    },
    "serviceDiscoveryType": false,
    "authenticationType": "jwt",
    "packageName": "com.mycompany.myapp",
    "serverPort": "8080",
    "cacheProvider": "ehcache",
    "enableHibernateCache": true,
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "prodDatabaseType": "postgresql",
    "buildTool": "maven",
    "serverSideOptions": [],
    "websocket": false,
    "searchEngine": false,
    "messageBroker": false,
    "enableSwaggerCodegen": false,
    "clientFramework": "react",
    "withAdminUi": true,
    "clientTheme": "none",
[...]

PR #13200, opened 2020-12-05
Fix many to many (React)

ctamisier:
Fix the root issue of https://github.com/jhipster/generator-jhipster/issues/13182
And fix an underlying issue where the relations from a many-to-many are set from the previous save. 

the `value` from `AvInput` should be set once.

This same "solution" is already applied with: `AvForm model={isNew ? {} : fooEntity}`

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/shared/util/entity-utils.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/shared/util/entity-utils.ts.ejs
@@ -38,12 +38,7 @@ export const cleanEntity = entity => {
  * @param idList Elements to map.
  * @returns The list of objects with mapped ids.
  */
-export const mapIdList = (idList: ReadonlyArray<any>) => {
-  if (idList) {
-    return idList.filter((entityId: any) => entityId !== '').map((entityId: any) => ({ id: entityId }));
-  }
-  return [];
-};
+ export const mapIdList = (idList: ReadonlyArray<any>) => idList.filter((entityId: any) => entityId !== '').map((entityId: any) => ({ id: entityId }));
 
  export const overridePaginationStateWithQueryParams = (paginationBaseState: IPaginationBaseState , locationSearch: string) => {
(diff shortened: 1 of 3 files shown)


generators/client/templates/react/src/main/webapp/app/shared/util/entity-utils.ts.ejs
```diff
@@ -38,12 +38,7 @@ export const cleanEntity = entity => {
  * @param idList Elements to map.
  * @returns The list of objects with mapped ids.
  */
-export const mapIdList = (idList: ReadonlyArray<any>) => {
-  if (idList) {
-    return idList.filter((entityId: any) => entityId !== '').map((entityId: any) => ({ id: entityId }));
-  }
-  return [];
-};
+ export const mapIdList = (idList: ReadonlyArray<any>) => idList.filter((entityId: any) => entityId !== '').map((entityId: any) => ({ id: entityId }));
```
qmonmert, line 41: and when the idList is undefined ???

ctamisier:
It's quite tricky..
Before this PR and before the fix on mapIdList we have this "workflow":
image
First [1, 2, 3] is set to the `value` of the `AvInput` (even for a create new Foo, as it gets the previous one (so I just have created a Foo with `[1, 2, 3]` bars before getting here)...)
Then the `reset` is called which set the `value` of the `AvInput` to `undefined`.
Then when we save, `values.bars` is `undefined` (which to the `undefined` of the `idList`)
It looks like the `AvInput` gets messed up when its value changes during component rendering.

With this PR:
image
We set once and only once the `value` of `AvInput` which is always `false` (it could have been always `undefined` it would have worked as well).
And with this, `values.bars` is equal to `[]` which is the same case that when we save a Foo the very first time.

ruddell (approved):
Tested locally, looks like a better fix to me. That list should not be undefined.  It also fixes the issue with the previously saved options being displayed as selected.

PR #13216, opened 2020-12-07
Fix application base name question in Windows Command Prompt

### LINKED ISSUE #13215: Application base name question is unstable in Windows Command Prompt

If in current main branch running linked `jhipster` then `askForModuleName` behaves inadequately in Windows Command Prompt. Question is asked but after entering answer new blank line is shown. After testing found out that must enter base name again into this blank line and this second input is used as application base name.

All works as expected in Git Bash in Windows.

Git bisect shows again first bad commit #13124 which upgraded Prettier to version 2.2.0 via `prettier-plugin-java` and for which solution in #13138 was to upgrade to Node.js LTS version.

If I remember correctly then upgrading Node.js to LTS version worked as expected for me when I was working with #13138. Maybe:
* I don't remember correctly
* or as meanwhile I updated npm to version npm@6.14.9 then this causes this problem
* or some transitive dependency meanwhile changed which causes this problem

I tried to upgrade node.js to v15.3.0 and npm to v7.1.0 but this didn't solve this issue.

`jhipster` should be stable also in Windows Command Prompt.

##### **Reproduce the error**

Run `jhipster` in Windows Command Prompt.

##### **Related issues**

#13138

##### **Suggest a Fix**

While debugging I found out that `askForInsightOptIn` is not asked because condition is not met, but some `async` operation is done internally. After deleting `askForInsightOptIn` question, Windows Command Prompt started to work as expected.
So I try to do PR where condition `shouldWeAskForOptIn()` is moved from `async function askForInsightOptIn()` to constructing questions array (this question is not added to questions array if condition is not met).

##### **JHipster Version(s)**

Current main branch.

Windows 10.

kaidohallik:
Fixes #13215 

---

### DIFF
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -30,6 +30,7 @@ module.exports = {
 };
 
 async function askForInsightOptIn() {
+    if (!statistics.shouldWeAskForOptIn()) return;
     const answers = await this.prompt({
         type: 'confirm',
         name: 'insight',
--- a/generators/app/prompts.js
+++ b/generators/app/prompts.js
@@ -31,7 +31,6 @@ module.exports = {
 
 async function askForInsightOptIn() {
     const answers = await this.prompt({
-        when: () => statistics.shouldWeAskForOptIn(),
         type: 'confirm',
         name: 'insight',
         message: `May ${chalk.cyan('JHipster')} anonymously report usage statistics to improve the tool over time?`,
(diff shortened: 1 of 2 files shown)


generators/app/prompts.js
```diff
@@ -31,7 +31,6 @@ module.exports = {
 
 async function askForInsightOptIn() {
     const answers = await this.prompt({
```
mshima, line 33:
```suggestion
    if (!statistics.shouldWeAskForOptIn()) return;
    const answers = await this.prompt({
```

mshima (changes requested):
Should be easier for blueprints.

kaidohallik:
> Should be easier for blueprints.

Thanks for the suggestions @mshima. Tested and those worked and used suggestions.

PR #13253, opened 2020-12-12
Change github actions caching

mshima:
Github actions cache limit is 5gb.
Each test creates ~500mb cache, with ~40 tests, the limit is reached by a lot.
So github cache approach is doing too many overwrites, making unstable.

Change to:
- Schedule a full build to create the cache.
- Change to 1 cache per related technology.

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -19,6 +19,8 @@
 
 name: Angular
 on:
+    schedule:
+        - cron: '0 0 * * *'
     push:
         branches-ignore:
             - 'dependabot/**'
@@ -74,6 +76,7 @@ jobs:
             matrix:
                 node_version: [14.15.0]
                 os: [ubuntu-20.04]
+                cache: [angular]
                 app-type:
                     - ngx-default
                     - ngx-mysql-es-noi18n-mapsid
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
That's what I said, few weeks ago, about the cache limit is 5gb, but I didn't correctly explain my idea, so my fault :-)

Maybe NPM is too big. What do you think about keeping cache only for backend?
As I'm not sure if NPM cache is used, as locally, I always download Internet when doing npm ci...

mshima:
Github actions cache became unstable some weeks ago.

ngx-mysql-es-noi18n-mapsid build:
- NPM: Cache Size: ~154 MB (161752912 B)
- Maven: Cache Size: ~218 MB (228606993 B)
- Cypress: Cache Size: ~164 MB (171914793 B)

ngx-gradle-fr build:
- NPM: Cache Size: ~154 MB (161714493 B)
- Maven: Cache Size: ~98 MB (102499987 B)
- Gradle: Cache Size: ~262 MB (274586907 B)
- Cypress: Same

Only maven/gradle would take ~300mb.
5gb is not enough for 1 cache per job.
The only solution is to have 1 cache for maven, 1 cache for gradle, 1 cache for cypress and 1 daily cache for npm.
Not sure how to accomplish this.
Anyway I think it's better to have an incomplete cache than no cache at all.

pascalgrimaud:
> Anyway I think it's better to have an incomplete cache than no cache at all.

Totally agree

mshima:
> Not sure how to accomplish this.

I mean that I don't know an easy way to create a more complete cache.
With this PR the first job will write the cache, and every other job will use that cache.

Issue #12441, opened 2020-09-12
Integrate automatic dependency scanning for security vulnerabilities.


bmvermeer:
Hi Folks, Brian here from Snyk

I had a chat with Matt Raible @mraible yesterday and he asked me to submit this issue.
We would love to integrate our free open-source dependency scanning capabilities into JHipster. As Snyk can scan issues in for multiple ecosystems like NodeJS, Java and .Net we feel that an integration makes sense.

What we would like to achieve is that JHipster users can use Snyk out of the box to scan their applications when they build. This to prevent dependencies with vulnerabilities entering the application or warn the user when there is a vulnerability found. Basically providing help to the developers out of the box to keep their application more secure.

Me and my team would be happy to discuss this further.

mraible:
@bmvermeer thanks for entering this issue. ❤️

If developers want to integrate Snyk w/o JHipster adding support, how would they do it?

Is it as simple as:

```
npm install -g snyk
snyk test
```

Will this test `package.json` and `pom.xml`? What about Gradle, will it auto detect `build.gradle` and `build.gradle.kts`?

Do you think we should install Snyk and run it as part of the app-generation process, or only when generating CI/CD files? If CI/CD, what commands should we run?

mraible:
Another thought: JHipster has support for generating K8s config and building Docker images with Jib. How can Snyk help scan those configurations and detect vulnerabilities?

bmvermeer:
@mraible 
It is indeed as easy as you say to use the snyk cli.
```
npm install -g snyk
snyk test
```
But before the first test a user needs to authenticate once with their free account.
By using 
```
snyk auth
```
you will get redirected to the browser and you can signing or signup. If you already have an account the token can be provided like
```
snyk auth <token>
```

The test command `snyk test` will look for a manifest file itself and use this for scanning this can be any manifest file we support. 
(package.json, pom.xml, build.gradle etc.) If you want to scan all. projects including the projects in subfolders you can supply the flag `--all-projects`. 

For docker images, you can use the same cli
```
snyk container test [options] [image]
```
To test docker images.

For both Snyk Open Source (the dependency scanning) and Snyk Container, you can also use the monitor function.
This takes a snapshot of that moment sends it to the Snyk platform and monitors it daily. So if some vuln is discovered over time (or a fix for an. existing vuln) you will be notified. This is normally used just before you app goes into production.

```
snyk monitor
snyk container monitor
```

[...]

PR #13254, opened 2020-12-12
CI-CD: fix CircleCI build

pascalgrimaud:
preparing work on #12441

---

### DIFF
--- a/generators/ci-cd/templates/circle.yml.ejs
+++ b/generators/ci-cd/templates/circle.yml.ejs
@@ -23,6 +23,17 @@ jobs:
             - image: jhipster/jhipster:v<%= jhipsterVersion %>
         steps:
             - checkout
+            - setup_remote_docker:
+                version: 19.03.13
+            - run:
+                name: Install docker and docker-compose
+                command: |
+                    echo jhipster | sudo -S apt update
+                    echo jhipster | sudo -S apt install -y docker.io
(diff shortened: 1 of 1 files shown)


SudharakaP (approved):

generators/ci-cd/templates/circle.yml.ejs
```diff
@@ -103,23 +95,23 @@ jobs:
...
-            -run:
+            - run:
                 name: Deploy to Heroku
                 command: './gradlew deployHeroku --no-daemon'
                 <%_ } _%>
             <%_ } _%>
             <%_ if (testFrameworks.includes('cypress') || testFrameworks.includes('protractor')) {
-                const commented = applicationType === 'gateway' ? '# ' : ''; _%>
-            <%= commented %>-run:
+                const commented = '# '; _%>
```
SudharakaP, line 104: @pascalgrimaud : Just a bit confused with this commented out section. Why do we have this e2e section if it doesn't work with CircleCI? Would we just remove it? 🤔

pascalgrimaud: to make it works, the user needs to remove the 127.0.0.1 part in docker-compose files, and as the change is easy, I thought it's a good idea to let this part.
But if you think we should remove it, I can too

SudharakaP: Ah, I get it. No, it's fine I think; makes sense. 👍🏽

Issue #13232, opened 2020-12-10
Sonar Analysis for Angular: bad results


pascalgrimaud:
##### **Overview of the issue**

The current Sonar Analysis for Angular application is bad:
- 13 bugs
- 1 vulnerability
- 15 security hotspots
- coverage at 63.5%

##### **Motivation for or Use Case**

- you can have a look at https://sonarcloud.io/dashboard?id=jhipster-sample-application
- the spec files should not be analyzed by Sonar https://sonarcloud.io/component_measures?id=jhipster-sample-application&metric=coverage&selected=jhipster-sample-application%3Asrc%2Fmain%2Fwebapp%2Fapp%2Fentities%2Ftest-root%2Flabel%2Flist%2Flabel.component.spec.ts
- the coverage dropped from 82% to 63% : https://sonarcloud.io/project/activity?custom_metrics=coverage&graph=custom&id=jhipster-sample-application

##### **Related issues**

Probably related to https://github.com/jhipster/generator-jhipster/pull/13171

##### **JHipster Version(s)**

current main branch

##### **Browsers and Operating System**

-   [x] Checking this box is mandatory (this is just to show you read everything)

kaidohallik:
I made #13233 which says to SonarCloud that `*.spec.ts` files are test files in Angular app and this should take them away from analyze and coverage.
This should fix coverage and most security hotspots.
Bugs, vulnerability and some security hotspots come from Java code

pascalgrimaud:
thanks a lot @kaidohallik for fixing this so fast ! let's merge it once the CI passed

avdev4j:
I supposed this is fixed through @kaidohallik PRs. I close the issue ;)

pascalgrimaud:
Not finished yet, @avdev4j as we still have :
- https://sonarcloud.io/project/issues?id=jhipster-sample-application&resolved=false&types=BUG
- https://sonarcloud.io/project/issues?id=jhipster-sample-application&resolved=false&types=VULNERABILITY

I'm removing the theme Angular, as the todo list is on Java part now

pascalgrimaud:
Closing this in favor of https://github.com/jhipster/generator-jhipster/issues/13291

PR #13256, opened 2020-12-12
Fix some sonar analysis: Call Optional#isPresent() before accessing the value

qmonmert:
#13232

Fix errors **Call Optional#isPresent() before accessing the value**

---

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/common/patch_template.ejs
+++ b/generators/entity-server/templates/src/main/java/package/common/patch_template.ejs
@@ -26,7 +26,7 @@ const mapper = entityInstance  + 'Mapper';
 _%>
 <%_ if(viaService) { %>
    <%_ if(!reactive) { %>
-   Optional<<%= instanceType %>> result = Optional.ofNullable(<%= entityInstance %>Service.partialUpdate(<%= instanceName %>));
+   Optional<<%= instanceType %>> result = <%= entityInstance %>Service.partialUpdate(<%= instanceName %>);
    <%_ } else { %>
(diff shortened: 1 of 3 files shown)


generators/entity-server/templates/src/main/java/package/common/patch_template.ejs
```diff
@@ -56,5 +56,5 @@ _%>
          <%_ } %>
       })<% } %><% if (dto !== 'no') { %>
    .map(<%= mapper %>::toDto)<% } %><% if (isService && !reactive) { %>
-   .get()<% } %>;
+   .orElse(null)<% } %>;
```
avdev4j, line 59: It sounds strange to return null after dealing with an optional, no?

Maybe throw an Exception or something like that will be better.

pascalgrimaud: I'd prefer to return the optional directly, and in Resource, instead of having:

```java
Optional<ChipsDTO> result = Optional.ofNullable(chipsService.partialUpdate(chipsDTO));
```

We'll have:

```java
Optional<ChipsDTO> result = chipsService.partialUpdate(chipsDTO);
```

avdev4j: Oh yea for sure, so far better.

qmonmert: ok guys, let's do this

DanielFran (approved):

PR #13332, opened 2020-12-19
Upgrade jhipster-registry to v6.5.0

pascalgrimaud:
---

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -55,7 +55,7 @@ const PRETTIER_JAVA_VERSION = packagejs.dependencies['prettier-plugin-java'];
 
 // Version of docker images
 const DOCKER_COMPOSE_FORMAT_VERSION = '3.8';
-const DOCKER_JHIPSTER_REGISTRY = 'jhipster/jhipster-registry:v6.4.0';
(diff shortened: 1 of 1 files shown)


mraible:
I've generated a couple of microservice architectures this morning to test the main branch. It seems I always have to start JHipster Registry twice to get it to stay up. Any idea why this is happening?

pascalgrimaud:
@mraible : is it with JWT or OAuth2 ? I suppose OAuth2 : is it with Keycloak or Okta ?

mraible:
It's with OAuth 2.0 and Keycloak.

pascalgrimaud:
it happens to me if I start registry at the same time than Keycloak, few weeks ago:
- start keycloak
- wait at least 20sec
- start registry

Tell me if it works for you ?

mraible:
That makes sense. Yes, if I start Keycloak and wait before starting the JHipster Registry, it works. Thanks, @pascalgrimaud!

pascalgrimaud:
@mraible : in fact, in an older version of JHipster, it would be possible, as the Registry waited Keycloak to be up, instead of failing immediatly. So something has changed with Spring Boot imo... or it can be a missing configuration

mraible:
In my experience, startup has always failed with Spring Boot + OAuth when the identity provider is not available. It might keep retrying when using WebFlux. I'm not sure.

PR #13336, opened 2020-12-20
Do not set location header (security issue)

ctamisier:
Remove _location_ header after a user or entity creation.

My understanding of this header is:
The URI that is passed could be used to **redirect** to this url from the browser right after creation, but we do not use this.
Also the current URI that is set is the API path (ex: /api/admin/users/user) and for me it is a not a URL that makes sense to be redirected on it. (I may be wrong be why redirect to the JSON resource instead of the "UI" of this resource).

Also it will fix the sonar security issue.

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/web/rest/EntityResource.java.ejs
@@ -66,6 +66,7 @@ import tech.jhipster.web.util.<% if (reactive) { %>reactive.<% } %>ResponseUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpStatus;
 <%_ if (pagination !== 'no') { _%>
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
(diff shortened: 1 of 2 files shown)


ruddell:
Sonar [flags this issue](https://sonarcloud.io/project/issues?id=jhipster-sample-application&open=AXWNjoEnzpLtv0cM4ct5&resolved=false&types=VULNERABILITY) because the `login` field is a user-controlled data.  We don't perform redirection and we already validate the user login through a regex, so it may be a false positive.  It's also not reported in the Sonar scan using the docker-compose file (may use different rules).

I think we should keep the header, it's valid for 201 responses (which don't redirect like 3xx responses).
> In cases of resource creation, it indicates the URL to the newly created resource.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location

> For 201 (Created) responses, the Location value refers to the primary resource created by the request.  For 3xx (Redirection) responses, the Location value refers to the preferred target resource for automatically redirecting the request.
https://tools.ietf.org/html/rfc7231#section-7.1.2

ctamisier:
Oh I see thanks.
I haven't seen the 201 specificity. So the path is correct.
But I still don't understand the added value on how this header can be used on client side through _location_ header.
I know there is a rfc but it looks like the first step to Hypermedia API world (but only for the 201..). (I know OHM would be a nice solution ;) )
I would have liked also to remove this code sections: 
`throws URISyntaxException {` 
and  
```
} catch (URISyntaxException e) {	
    throw new RuntimeException(e);	
}
```

So we can see for sonar rules for the security issue. 
We can close this ticket !
Thanks !

PR #13358, opened 2020-12-22
[angular] Use type-safe equality operator and non-null assertion

kaidohallik:
Current style is to use type-safe equality and non-null assertion for `editForm.get` where possible. This PR aligns code to this style.

---


kaidohallik:
I tried to achieve type safe equality also in entities update form, but this is problematic, because our model definitions in TypeScript are not correct.

Empty values in locally created instances are `undefined`.

Empty values coming from server are `null`.

But model definitions tell that `undefined` is allowed and `null` is not allowed value.

So giving up with entity update form type safe equality.

PR #13367, opened 2020-12-23
Fix entity model types in front end

kaidohallik:
This is follow up to #13358

In https://github.com/jhipster/generator-jhipster/pull/13358#issuecomment-749558087 I found that our entity model definitions are not correct, because non required fields can have value `null` (after reading from server) which is not allowed in model.
This PR fixes that, adds `| null` to non required fields type.

In the second commit using type safe equality operator in Angular entity update `.ts` file.

---


mshima (approved):
LGTM

DanielFran (approved):

PR #13369, opened 2020-12-23
Replace deprecated `StringUtils.isEmpty()` by `ObjectUtils.isEmpty()` and cover its usage with tests

ramnasidharta:
`StringUtils.isEmpty()` [has been deprecated](https://github.com/spring-projects/spring-framework/issues/25945), so this PR replaces it by `ObjectUtils.isEmpty()` in the constructor of `TokenProvider`.

Besides, as `isEmpty()` is used in a conditional block that is not being tested, I added two unit tests to `TokenProviderTest`.


DanielFran:
Thanks @ramnasidharta for tour contribution

DanielFran (approved):

PR #13196, opened 2020-12-04
Server common scripts

vishal423:
Closes #12697 
---

Please make sure the below checklist is followed for Pull Requests.

-   [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
-   [ ] Tests are added where necessary
-   [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
-   [ ] Documentation is added/updated where necessary
-   [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13429, opened 2021-01-08
Use prefix webapp instead of webpack in all npm build scripts

kaidohallik:
Follow up to #13196

I believe it's good to use one single prefix for front end build scripts.
This PR changes all `npm` build scripts from `webpack:` prefix to `webapp:` prefix.

I made just in case 2 commits:
* first for Angular, in Angular we use Angular CLI and `webapp:` prefix suits better for sure
* second for React and Vue - I believe it's better to use one single `webapp:` prefix also in React and Vue

---


vishal423:
The initial idea was to not break the backward compatibility on the nodejs scripts side in this version and then later in the next major release to move away completely (will give time to module/blueprint authors to migrate).

pascalgrimaud:
I'm trusting you here, as v7 is not released in final version, I think we can do some breaking changes.
So feel free to merge it if needed

PR #13440, opened 2021-01-08
Ensure /websocket/tracker/ cant bypass

### LINKED ISSUE #13439: Ensure spring security authorization for /websocket/tracker is not bypassed with /websocket/tracker/
When generating a jhipster project that uses Spring Websockets, jhipster in SecurityConfigurations.class
set authorization using antMatchers on /websocket/tracker to only allow ADMIN access. The issue is antMatchers do not account for trailing slash situations so /websocket/tracker/ would bypass this authorization.

As this issue has relation to security, I emailed Julien following the document security process. He directed me to put in an issue for it and a PR.

jhipster generator's default code should be as secure as possible even though users will adapt to their context.

##### **Related issues**
Not that I could tell.

##### **Suggest a Fix**
switch antMatchers to mvcMatchers which covers trailing slash situations

##### **JHipster Version(s)**
Found in 6.10 but present in latest code base

cgravatt:
Ensure /websocket/tracker/ cant bypass
AuthoritiesConstants.ADMIN restriction

Fix #13439

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -273,8 +273,7 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
             .antMatchers("/api/admin/**").hasAuthority(AuthoritiesConstants.ADMIN)
             .antMatchers("/api/**").authenticated()
     <%_ if (websocket === 'spring-websocket') { _%>
-            .antMatchers("/websocket/tracker").hasAuthority(AuthoritiesConstants.ADMIN)
-            .antMatchers("/websocket/**").permitAll()
+            .antMatchers("/websocket/**").authenticated()
     <%_ } _%>
             .antMatchers("/management/health").permitAll()
             .antMatchers("/management/health/**").permitAll()
--- a/generators/server/templates/src/main/java/package/config/UaaConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/UaaConfiguration.java.ejs
@@ -107,8 +107,7 @@ public class UaaConfiguration extends AuthorizationServerConfigurerAdapter imple
                 .antMatchers("/api/account/reset-password/init").permitAll()
                 .antMatchers("/api/account/reset-password/finish").permitAll()
                 .antMatchers("/api/**").authenticated()<% if (websocket === 'spring-websocket') { %>
-                .antMatchers("/websocket/tracker").hasAuthority(AuthoritiesConstants.ADMIN)
-                .antMatchers("/websocket/**").permitAll()<% } %>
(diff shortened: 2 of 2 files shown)


cgravatt:
Tagging @jdubois as requested.

ruddell:
Hi @cgravatt and thanks for the report.  

I tested your changes but I don't see any difference. I get the same result before and after on `/websocket/tracker/` - `Welcome to SockJS!`

Anyway, I don't think that line affects anything.  We don't connect directly to `/websocket/tracker` anywhere in the code.  It's used in the Tracker example ([here](https://github.com/jhipster/generator-jhipster/blob/770b5de4c63bc80ccf3f3e38ba193e08c0d47144/generators/client/templates/angular/src/main/webapp/app/core/tracker/tracker.service.ts.ejs#L61)), but that actually connects to `/websocket/tracker/info` so it's allowed ([requests screenshot](https://user-images.githubusercontent.com/4294623/104070795-b74b9580-51d5-11eb-8c01-baf3b0bc20d4.png)).  And it needs to be allowed so the activity info is sent (for ROLE_USERs).

WebSocket security is controlled in the WebsocketSecurityConfiguration, which blocks any connection/subscribe attempts if unauthenticated.  I traced the problematic lines back to their source, in [JHipster v0.7.0](https://github.com/jhipster/generator-jhipster/commit/1b6da47928c4b127dcd49b79a16f5cd9ca302d23#diff-559afee40f864b217f1c753b4a705028d89a01bfc831c61b694e76b0cb943745) - before SockJS and STOMP were added to JHipster.

I suggest to change the two WebSocket-related lines to `.antMatchers("/websocket/**").authenticated()` which works well.  It returns a 401 on WebSocket connection attempt, rather than on the first STOMP frame.

I only tested this with JWT though, need to verify it works for OAuth2 and Session.

cgravatt:
Thanks. This may be a non issue. I know in spring security antMatchers won't account for a trailing slash situation. So that authorization line is by-passable. I'll review your comment a little more this evening / over the weekend, but it does sound like in the grand scheme of things this is a non issue.

cgravatt:
@ruddell You are right. mvcMatchers will not fix the issue and will result in the same behavior as the existing antMatchers. I made a mistake.

I believe this is because mvcMatchers will typically handle trailing slash situations as it maps requests to request mappings; however, in this case this is not a controller endpoint with a request mapping. As a result, the mvcMatchers behavior falls back to the equivalent behavior of the antMatcher.

Here's demonstrating that the authorization is bypassable to begin with:

image

I think your suggestion makes sense to switch to .antMatchers("/websocket/**").authenticated()
based on WebSocketSecurityConfiguration.class.

I will test Oauth2 and Session with that change and update PR for additional review and thoughts.

cgravatt:
Session looked good to me

mraible:
@ruddell Can you please review again and merge if OK?

ruddell (approved):

PR #13445, opened 2021-01-09
Polish app config spacing

kaidohallik:
`.yml` files are formatted by Prettier with `tabWith=2`, using the same `tabWidth` also in templates.
In some places in template already was `tabWidth=2` used, now it's everywhere so.
Commented out sections are now with correct spacing after generation.
Fixed some odd number of spaces to correct ones.
Added one missing license header.

---

### DIFF
--- a/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application-dev.yml.ejs
@@ -32,232 +32,232 @@
 # ===================================================================
 
 logging:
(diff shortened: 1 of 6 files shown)


pascalgrimaud:
Agree with you @kaidohallik 
But it's more a global work to do on our side
- We have 2 spaces in generated ts, js files
- But we use 4 spaces in our yeoman code
- 4 spaces in our templates toi
- etc

We are not consistent at all

Issue #13510, opened 2021-01-14
Angular many-to-one and many-to-many relationships shows empty value.


mshima:
##### **Overview of the issue**

Angular many-to-one and many-to-many relationships may show empty value even if there is a value.
This problem happens when pagination at related Entity is enabled:
- With pagination only 20 items are returned from the server.
- If the saved value is not returned from server, select shows an empty value even when the value exists.

##### **Motivation for or Use Case**

Selects should not show empty value if there is a value.

##### **Reproduce the error**

- Download https://github.com/jhipster/jdl-samples/blob/main/bug-tracker.jh
- Add `Project` to pagination.
- Import `jhipster jdl bug-tracker.jh`
- Run the application
- Create 11 more Projects
- Go to `administration -> database`
- Select `Ticket`
- Set any Ticket's Project to the last project eg: 1011
- Go to the Ticket edit page.
- Project's select is empty.

##### **Suggest a Fix**

Add Project registry to the select if not returned from the server.

##### **JHipster Version(s)**

master

##### **JHipster configuration**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "authenticationType": "jwt",
    "cacheProvider": "ehcache",
    "clientFramework": "angularX",
    "serverPort": "8080",
    "serviceDiscoveryType": false,
    "skipUserManagement": false,
    "baseName": "jhipster",
    "buildTool": "maven",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "enableHibernateCache": true,
    "enableSwaggerCodegen": false,
    "enableTranslation": true,
    "jhiPrefix": "jhi",
    "languages": ["en"],
    "messageBroker": false,
    "nativeLanguage": "en",
    "prodDatabaseType": "postgresql",
    "searchEngine": false,
    "skipClient": false,
    "testFrameworks": [],
    "websocket": false,
    "jhipsterVersion": "7.0.0-beta.0",
    "skipServer": false,
    "packageFolder": "com/mycompany/myapp",
    "packageName": "com.mycompany.myapp",
    "clientPackageManager": "npm",
    "clientTheme": "none",
    "clientThemeVariant": "",
    "useSass": true,
    "applicationType": "monolith",
    "entities": ["Project", "Label", "Ticket", "Attachment", "Comment"],
    "skipJhipsterDependencies": true,
    "skipCheckLengthOfIdentifier": false,
    "skipFakeData": false,
[...]

PR #13592, opened 2021-01-18
[angular] Fix empty selects when the current value is not returned from the server.

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/13510

---

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/core/util/operators.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/core/util/operators.ts.ejs
@@ -16,21 +16,11 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-<%_
-  const idType = getTypescriptKeyType(user.primaryKey.type);
-_%>
-export interface IUser {
-  id?: <%= idType %>;
-  login?: string;
-}
 
-export class User implements IUser {
-  constructor(
-    public id: <%= idType %>,
-    public login: string,
-  ) {}
-}
-
-export function getUserIdentifier(user: IUser): <%= idType %> | undefined {
-  return user.id;
+/*
+ * Function used to workaround https://github.com/microsoft/TypeScript/issues/16069
+ * es2019 alternative `const filteredArr = myArr.flatMap((x) => x ? x : []);`
+ */
+export function isPresent<T>(t: T | undefined | null | void): t is T {
+  return t !== undefined && t !== null;
 }
--- a/generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.spec.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.spec.ts.ejs
@@ -32,12 +32,13 @@ describe('Service Tests', () => {
   describe('User Service', () => {
     let service: UserService;
     let httpMock: HttpTestingController;
+    let expectedResult: IUser | IUser[] | boolean | number | null;
 
     beforeEach(() => {
       TestBed.configureTestingModule({
         imports: [HttpClientTestingModule],
       });
-
+      expectedResult = null;
       service = TestBed.inject(UserService);
       httpMock = TestBed.inject(HttpTestingController);
     });
@@ -56,8 +57,6 @@ describe('Service Tests', () => {
       });
 
       it('should return Users', () => {
-        let expectedResult: IUser[] | null | undefined;
-
         service.query().subscribe(received => {
           expectedResult = received.body;
         });
@@ -68,8 +67,6 @@ describe('Service Tests', () => {
       });
 
       it('should propagate not found response', () => {
-        let expectedResult = 0;
-
         service.query().subscribe({
           error: (error: HttpErrorResponse) => expectedResult = error.status
         });
@@ -80,6 +77,58 @@ describe('Service Tests', () => {
           statusText: 'Inernal Server Error',
         });
         expect(expectedResult).toEqual(500);
+      });
+
+      describe('addUserToCollectionIfMissing', () => {
+        it('should add a User to an empty array', () => {
+          const user: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>;
+          expectedResult = service.addUserToCollectionIfMissing([], user);
+          expect(expectedResult).toHaveLength(1);
+          expect(expectedResult).toContain(user);
+        });
+
+        it('should not add a User to an array that contains it', () => {
+          const user: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>;
+          const userCollection: IUser[] = [
+            {
+              ...user,
+            },
+            <%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>,
+          ];
+          expectedResult = service.addUserToCollectionIfMissing(userCollection, user);
+          expect(expectedResult).toHaveLength(2);
+        });
+
+        it("should add a User to an array that doesn't contain it", () => {
+          const user: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>;
+          const userCollection: IUser[] = [<%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>];
+          expectedResult = service.addUserToCollectionIfMissing(userCollection, user);
+          expect(expectedResult).toHaveLength(2);
+          expect(expectedResult).toContain(user);
+        });
+
+        it('should add only unique User to an array', () => {
+          const userArray: IUser[] = [<%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>, <%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>, <%- generateTestEntityPrimaryKey(user.primaryKey) %>];
+          const userCollection: IUser[] = [<%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>];
+          expectedResult = service.addUserToCollectionIfMissing(userCollection, ...userArray);
+          expect(expectedResult).toHaveLength(3);
+        });
+
+        it("should accept varargs", () => {
+          const user: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>;
+          const user2: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>;
+          expectedResult = service.addUserToCollectionIfMissing([], user, user2);
+          expect(expectedResult).toHaveLength(2);
+          expect(expectedResult).toContain(user);
+          expect(expectedResult).toContain(user2);
+        });
+
+        it("should accept null and undefined values", () => {
+          const user: IUser = <%- generateTestEntityPrimaryKey(user.primaryKey, 0) %>;
+          expectedResult = service.addUserToCollectionIfMissing([], null, user, undefined);
+          expect(expectedResult).toHaveLength(1);
+          expect(expectedResult).toContain(user);
+        });
       });
     });
   });
--- a/generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.ts.ejs
@@ -22,8 +22,9 @@ import { Observable<% if (authenticationType !== 'oauth2' && databaseType !== 's
 
 import { SERVER_API_URL } from 'app/app.constants';
 import { createRequestOption } from 'app/core/request/request-util';
+import { isPresent } from 'app/core/util/operators';
 import { Pagination } from 'app/core/request/request.model';
-import { IUser } from './user.model';
+import { IUser, getUserIdentifier } from './user.model';
 <%_ if (authenticationType !== 'oauth2' && databaseType !== 'sql' && databaseType !== 'mongodb' && databaseType !== 'couchbase') { _%>
 import { Authority } from 'app/config/authority.constants';
 <%_ } _%>
@@ -37,5 +38,22 @@ export class UserService {
   query(req?: Pagination): Observable<HttpResponse<IUser[]>> {
     const options = createRequestOption(req);
     return this.http.get<IUser[]>(this.resourceUrl, { params: options, observe: 'response' });
+  }
+
+  addUserToCollectionIfMissing(userCollection: IUser[], ...usersToCheck: (IUser | null | undefined)[]): IUser[] {
+    const users: IUser[] = usersToCheck.filter(isPresent);
+    if (users.length > 0) {
+      const userCollectionIdentifiers = userCollection.map(userItem => getUserIdentifier(userItem)!);
+      const usersToAdd = users.filter(userItem => {
+        const userIdentifier = getUserIdentifier(userItem);
+        if (userIdentifier == null || userCollectionIdentifiers.includes(userIdentifier)) {
+            return false;
+        }
+        userCollectionIdentifiers.push(userIdentifier);
+        return true;
+      });
+      return [...usersToAdd, ...userCollection];
+    }
+    return userCollection;
   }
 }
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
@@ -235,6 +235,55 @@ describe('Service Tests', () => {
                 expect(expectedResult);
             });
             <%_ } _%>
+
+            describe('add<%= entityAngularName %>ToCollectionIfMissing', () => {
+                it('should add a <%= entityAngularName %> to an empty array', () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], <%= entityInstance %>);
+                    expect(expectedResult).toHaveLength(1);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                });
+
+                it('should not add a <%= entityAngularName %> to an array that contains it', () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    const <%= entityInstance %>Collection: I<%= entityAngularName %>[] = [{
+                        ...<%= entityInstance %>,
+                    }, <%- generateTestEntityPrimaryKey(primaryKey, 1) %>];
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection, <%= entityInstance %>);
+                    expect(expectedResult).toHaveLength(2);
+                });
+
+                it("should add a <%= entityAngularName %> to an array that doesn't contain it", () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    const <%= entityInstance %>Collection: I<%= entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(primaryKey, 1) %>];
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection, <%= entityInstance %>);
+                    expect(expectedResult).toHaveLength(2);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                });
+
+                it("should add only unique <%= entityAngularName %> to an array", () => {
+                    const <%= entityInstance %>Array: I<%= entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(primaryKey, 0) %>, <%- generateTestEntityPrimaryKey(primaryKey, 1) %>, <%- generateTestEntityPrimaryKey(primaryKey) %>];
+                    const <%= entityInstance %>Collection: I<%= entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(primaryKey, 1) %>];
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection, ...<%= entityInstance %>Array);
+                    expect(expectedResult).toHaveLength(3);
+                });
+
+                it("should accept varargs", () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    const <%= entityInstance %>2: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 1) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], <%= entityInstance %>, <%= entityInstance %>2);
+                    expect(expectedResult).toHaveLength(2);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                    expect(expectedResult).toContain(<%= entityInstance %>2);
+                });
+
+                it("should accept null and undefined values", () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], null, <%= entityInstance %>, undefined);
+                    expect(expectedResult).toHaveLength(1);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                });
+            });
         });
 
         afterEach(() => {
--- a/generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.ts.ejs
+++ b/generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.ts.ejs
@@ -30,6 +30,7 @@ import { map } from 'rxjs/operators';
 import * as dayjs from 'dayjs';
 <%_ } _%>
 
+import { isPresent } from 'app/core/util/operators';
 <%_ if (fieldsContainLocalDate) { _%>
 import { DATE_FORMAT } from 'app/config/input.constants';
 <%_ } _%>
@@ -38,7 +39,7 @@ import { createRequestOption } from 'app/core/request/request-util';
 <%_ if (searchEngine !== false) { _%>
 import { <%= searchType %> } from 'app/core/request/request.model';
 <%_ } _%>
-import { I<%= entityAngularName %> } from '../<%= entityFileName %>.model';
+import { I<%= entityAngularName %>, get<%= entityAngularName %>Identifier } from '../<%= entityFileName %>.model';
 
 export type EntityResponseType = HttpResponse<I<%= entityAngularName %>>;
 export type EntityArrayResponseType = HttpResponse<I<%= entityAngularName %>[]>;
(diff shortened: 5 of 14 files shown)


mshima:
@kaidohallik what do you think about this one?

kaidohallik:
I have mixed feelings here.

It's definitely good if we fix bug not showing data in the update form.

But current JHipster implementation has limitations which doesn't allow to produce such problematic situation. Maybe correct approach is to improve JHipster relationships related update form logic, like using typeahead with server side search.
Currently every developer must implement it's own logic for selects from paginated entities.

However, if I remember correctly then typeahead has been rejected several times, so probably need to go with this PR way.

I tested this branch. `addUserToCollectionIfMissing` and similar methods in entity side are throwing error if no item is selected. Selected item parameter should be optional.

mshima:
> I have mixed feelings here.
> 
> It's definitely good if we fix bug not showing data in the update form.

That's my opinion too.

> But current JHipster implementation has limitations which doesn't allow to produce such problematic situation. Maybe correct approach is to improve JHipster relationships related update form logic, like using typeahead with server side search.

Not entirely true, just load an external set of data for testing.

> Currently every developer must implement it's own logic for selects from paginated entities.
> 
> However, if I remember correctly then typeahead has been rejected several times, so probably need to go with this PR way.

Just looked for it and seems https://github.com/jhipster/generator-jhipster/pull/6618 was the candidate, but was never finished.
It's not difficult to implement typeahead now using https://ng-bootstrap.github.io/#/components/typeahead/examples.
The missing part is to expose filtering by the otherEntityFieldName.
 
> I tested this branch. `addUserToCollectionIfMissing` and similar methods in entity side are throwing error if no item is selected. Selected item parameter should be optional.

generators/client/templates/angular/src/main/webapp/app/core/util/operators.ts.ejs
```diff
@@ -0,0 +1,22 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+
+export function isPresent<T>(t: T | undefined | null | void): t is T {
+  return t !== undefined && t !== null;
+}
```
kaidohallik, line 22: It seems to me that this adds unnecessary complexity. Simple check can be used instead of this added `isPresent`.

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
```diff
@@ -20,8 +20,8 @@
   const idType = getTypescriptKeyType(user.primaryKey.type);
 _%>
 export interface IUser {
-  id: <%= idType %>;
-  login: string;
+  id?: <%= idType %>;
+  login?: string;
```
kaidohallik, line 24: Why this change? When can `id` or `login` can be undefined?

generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.ts.ejs
```diff
@@ -38,4 +39,21 @@ export class UserService {
     const options = createRequestOption(req);
     return this.http.get<IUser[]>(this.resourceUrl, { params: options, observe: 'response' });
   }
+
+  addUserToCollectionIfMissing(userCollection: IUser[], ...usersToCheck: (IUser | null)[]): IUser[] {
+    const users: IUser[] = usersToCheck.filter(isPresent);
```
kaidohallik, line 44: Can we simplify parameter `usersToCheck` and can we use assignment without `isPresent`?

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
```diff
@@ -30,3 +30,7 @@ export class User implements IUser {
     public login: string,
   ) {}
 }
+
+export function getUserIdentifier(user: IUser): <%= idType %> | undefined {
+  return user.id;
+}
```
kaidohallik, line 36: Why adding this function, why not using just `user.id`?

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.ts.ejs
```diff
@@ -99,6 +100,23 @@ export class <%= entityAngularName %>Service {
     }
     <%_ } _%>
 
+    add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection: I<%= entityAngularName %>[], ...<%= entityInstancePlural %>ToCheck: (I<%= entityAngularName %> | null)[]): I<%= entityAngularName %>[] {
+        const <%= entityInstancePlural %>: I<%= entityAngularName %>[] = <%= entityInstancePlural %>ToCheck.filter(isPresent);
```
kaidohallik, line 104: Can we simplify `ToCheck` parameter and construct assignment without `isPresent`?

kaidohallik (changes requested):

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
mshima: `login` can be undefined using dto:
https://github.com/jhipster/generator-jhipster/blob/ac3f14e815db7e419b4f67541d4c00573aed2bdc/generators/server/templates/src/main/java/package/service/mapper/UserMapper.java.ejs#L142-L152

Not sure about `id`, but I am wondering if it's worth to keep it as non-undefined and keep the quirk only for user entity..

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
mshima: I've added to entities too.
- I have a PR for composite ids that will change every `primaryKey.name` reference.
- Can be reused at `trackEntityByIdentifier` functions.

This should make easier for the upcoming PR, anyway if you prefer to keep references, I'm ok with it.

generators/client/templates/angular/src/main/webapp/app/core/util/operators.ts.ejs
kaidohallik: I tried too now and after playing a bit my conclusion is that your solution is a good solution. I couldn't do better. So marking this and related comments as resolved.

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
kaidohallik: OK, good point about keeping `User` entity similar to other entities, so this is good to go.

generators/client/templates/angular/src/main/webapp/app/entities/user/user.model.ts.ejs
kaidohallik: OK, good point about keeping User entity similar to other entities and this way simplify generator, so this is good to go.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+        describe('Loading relationships', () => {
+            describe('loadRelationshipsOptions', () => {
+    <%_ for (const relationshipsByEntityNeedingOptions of allRelationshipsByEntityNeedingOptions) { _%>
+        <%_ const relationshipsWithCustomUniqueOptions = relationshipsByEntityNeedingOptions.filter(rel => rel.relationshipType === 'one-to-one' && rel.otherEntityName !== 'user'); %>
+        <%_ const relationshipsWithCustomSharedOptions = relationshipsByEntityNeedingOptions.filter(rel => !relationshipsWithCustomUniqueOptions.includes(rel)); %>
+        <%_ if (relationshipsWithCustomSharedOptions.length > 0) { _%>
+            <%_ const otherEntity = relationshipsByEntityNeedingOptions[0].otherEntity _%>
+
+                it('Should call <%= otherEntity.entityAngularName %> query and add missing value', () => {
+                    fakeAsync(() => {
```
kaidohallik, line 143:
```suggestion
```
If I tested then this code was not executed if using `fakeAsync`. After removing `fakeAsync` test was ran and this test failed.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+            <%_ for (const relationship of relationshipsWithCustomSharedOptions) { _%>
+                <%_ const reference = relationship.reference _%>
+                            <% if (relationship.collection) { %>...<% } %><%= reference.name %>,
+            <%_ } _%>
+                        ];
+                        const expectedColection: I<%= otherEntity.entityAngularName %>[] = [...additional<%= otherEntity.entityAngularNamePlural %>, ...<%= otherEntity.entityInstance %>Colection];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'add<%= otherEntity.entityAngularName %>ToCollectionIfMissing').and.returnValues(expectedColection);
+
+                        comp['loadRelationshipsOptions']();
+                        tick();
```
kaidohallik, line 168:
```suggestion
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+
+                        expect(<%= otherEntity.entityInstance %>Service.query).toHaveBeenCalled();
+                        expect(<%= otherEntity.entityInstance %>Service.add<%= otherEntity.entityAngularName %>ToCollectionIfMissing).toHaveBeenCalledWith(<%= otherEntity.entityInstance %>Colection, additional<%= otherEntity.entityAngularNamePlural %>);
+                        expect(comp.<%= otherEntity.entityInstancePlural %>SharedCollection).toEqual(expectedColection);
+                    });
```
kaidohallik, line 173:
```suggestion
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+                        const additional<%= otherEntity.entityAngularNamePlural %> = [
+            <%_ for (const relationship of relationshipsWithCustomSharedOptions) { _%>
+                <%_ const reference = relationship.reference _%>
+                            <% if (relationship.collection) { %>...<% } %><%= reference.name %>,
+            <%_ } _%>
+                        ];
+                        const expectedColection: I<%= otherEntity.entityAngularName %>[] = [...additional<%= otherEntity.entityAngularNamePlural %>, ...<%= otherEntity.entityInstance %>Colection];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'add<%= otherEntity.entityAngularName %>ToCollectionIfMissing').and.returnValues(expectedColection);
+
+                        comp['loadRelationshipsOptions']();
```
kaidohallik, line 167:
```suggestion
                        comp.ngOnInit();
```
It's a good practice to test public interface. Can you give input to component through `ActivatedRoute` and run public `ngOnInit()` instead?

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+            <%_ for (const relationship of relationshipsWithCustomSharedOptions) { _%>
+                <%_ const reference = relationship.reference _%>
+                <%_ if (relationship.collection) { _%>
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>];
+                <%_ } else { _%>
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %> = <%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>;
+                <%_ } _%>
+                        <%= entityInstance %>.<%= reference.name %> = <%= reference.name %>;
+            <%_ } _%>
+                        comp.updateForm(<%= entityInstance %>);
```
kaidohallik, line 154: Can you give input through `ActivatedRoute` and then this functionality is called through `ngOnInit()`.
Method `updateForm` should be protected in component. If you agree to make this change as well in this PR then it would be great. If not then this change can be done in a separate PR.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+                        expect(comp.<%= otherEntity.entityInstancePlural %>SharedCollection).toEqual(expectedColection);
+                    });
+                });
+        <%_ } _%>
+        <%_ for (const relationship of relationshipsWithCustomUniqueOptions) { %>
+            <%_ const otherEntity = relationship.otherEntity _%>
+            <%_ const reference = relationship.reference _%>
+
+                it('Should call <%= reference.name %> query and add missing value', () => {
+                    fakeAsync(() => {
```
kaidohallik, line 181:
```suggestion
```
Test is not executed with `fakeAsync`. After removing test is failing.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+                        const <%= entityInstance %> : I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %> = <%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>;
+                        <%= entityInstance %>.<%= reference.name %> = <%= reference.name %>;
+                        comp.updateForm(<%= entityInstance %>);
```
kaidohallik, line 185: Give input through `ActivatedRoute`.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+
+                        const <%= reference.name %>Colection: I<%= otherEntity.entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'query').and.returnValue(of(new HttpResponse({ body: <%= reference.name %>Colection })));
+                        const expectedColection: I<%= otherEntity.entityAngularName %>[] = [<%= reference.name %>, ...<%= reference.name %>Colection];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'add<%= otherEntity.entityAngularName %>ToCollectionIfMissing').and.returnValue(expectedColection);
+
+                        comp['loadRelationshipsOptions']();
```
kaidohallik, line 192:
```suggestion
                        comp.ngOnInit();
```
Give input through `ActivatedRoute`.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+                        tick();
```
kaidohallik, line 193:
```suggestion
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+
+                        expect(<%= otherEntity.entityInstance %>Service.query).toHaveBeenCalled();
+                        expect(<%= otherEntity.entityInstance %>Service.add<%= otherEntity.entityAngularName %>ToCollectionIfMissing).toHaveBeenCalledWith(<%= reference.name %>Colection, <%= reference.name %>);
+                        expect(comp.<%= relationship.relationshipFieldNamePlural %>Collection).toEqual(expectedColection);
+                    });
```
kaidohallik, line 198:
```suggestion
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -124,6 +129,79 @@ _%>
...
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>];
+                <%_ } else { _%>
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %> = <%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>;
+                <%_ } _%>
+                        <%= entityInstance %>.<%= reference.name %> = <%= reference.name %>;
+            <%_ } _%>
+                        comp.updateForm(<%= entityInstance %>);
+
+                        const <%= otherEntity.entityInstance %>Colection: I<%= otherEntity.entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'query').and.returnValue(of(new HttpResponse({ body: <%= otherEntity.entityInstance %>Colection })));
```
kaidohallik, line 157: Typo, use Collection instead of Colection.

generators/client/templates/angular/src/main/webapp/app/core/util/operators.ts.ejs
kaidohallik, line 22: Can you add test to this method?

generators/client/templates/angular/src/main/webapp/app/entities/user/user.service.spec.ts.ejs
```diff
@@ -81,6 +78,42 @@ describe('Service Tests', () => {
...
+            {
+              ...user,
+            },
+            <%- generateTestEntityPrimaryKey(user.primaryKey, 1) %>,
+          ];
+          expectedResult = service.addUserToCollectionIfMissing(userCollection, user);
+          expect(expectedResult).toHaveLength(2);
+        });
+
+        it("should add a User to an array that doesn't contains it", () => {
```
kaidohallik, line 102:
```suggestion
        it("should add a User to an array that doesn't contain it", () => {
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
```diff
@@ -235,6 +235,55 @@ describe('Service Tests', () => {
...
+                it('should not add a <%= entityAngularName %> to an array that contains it', () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    const <%= entityInstance %>Collection: I<%= entityAngularName %>[] = [{
+                        ...<%= entityInstance %>,
+                    }, <%- generateTestEntityPrimaryKey(primaryKey, 1) %>];
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection, <%= entityInstance %>);
+                    expect(expectedResult).toHaveLength(2);
+                });
+
+                it("should add a <%= entityAngularName %> to an array that doesn't contains it", () => {
```
kaidohallik, line 256:
```suggestion
                it("should add a <%= entityAngularName %> to an array that doesn't contain it", () => {
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
```diff
@@ -235,6 +235,55 @@ describe('Service Tests', () => {
...
+                    const <%= entityInstance %>2: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 1) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], <%= entityInstance %>, <%= entityInstance %>2);
+                    expect(expectedResult).toHaveLength(2);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                    expect(expectedResult).toContain(<%= entityInstance %>2);
+                });
+
+                it("should accept null values", () => {
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], null, <%= entityInstance %>);
```
kaidohallik, line 282:
```suggestion
                it("should accept null and undefined values", () => {
                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], null, <%= entityInstance %>, undefined);
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
```diff
@@ -235,6 +235,55 @@ describe('Service Tests', () => {
...
+                });
+
+                it("should add only unique <%= entityAngularName %> to an array", () => {
+                    const <%= entityInstance %>Array: I<%= entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(primaryKey, 0) %>, <%- generateTestEntityPrimaryKey(primaryKey, 1) %>, <%- generateTestEntityPrimaryKey(primaryKey) %>];
+                    const <%= entityInstance %>Collection: I<%= entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(primaryKey, 1) %>];
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing(<%= entityInstance %>Collection, ...<%= entityInstance %>Array);
+                    expect(expectedResult).toHaveLength(3);
+                });
+
+                it("should accept varargs", () => {
```
kaidohallik, line 271: It's a good test. Can you add varargs test also for `user.service`?

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
```diff
@@ -235,6 +235,55 @@ describe('Service Tests', () => {
...
+                    const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 0) %>;
+                    const <%= entityInstance %>2: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 1) %>;
+                    expectedResult = service.add<%= entityAngularName %>ToCollectionIfMissing([], <%= entityInstance %>, <%= entityInstance %>2);
+                    expect(expectedResult).toHaveLength(2);
+                    expect(expectedResult).toContain(<%= entityInstance %>);
+                    expect(expectedResult).toContain(<%= entityInstance %>2);
+                });
+
+                it("should accept null values", () => {
```
kaidohallik, line 280: It's a good test. Can you add null + undefined test also for `user.service`?

kaidohallik (changes requested):
Now console error is gone and code is working. Good work @mshima!
I have some comments about polishing and tests.

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
mshima: Done

generators/entity-client/templates/angular/src/main/webapp/app/entities/service/entity.service.spec.ts.ejs
mshima: Done

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
mshima: Not an expert on angular tests.
So can take some time.

kaidohallik: Sample of providing `ActivatedRoute` is for example in https://github.com/jhipster/generator-jhipster/blob/main/generators/entity-client/templates/angular/src/main/webapp/app/entities/detail/entity-management-detail.component.spec.ts.ejs#L44-L49

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -26,64 +30,183 @@ import { HttpResponse } from '@angular/common/http';
...
 
 import { <%= entityAngularName %>UpdateComponent } from './<%= entityFileName %>-update.component';
 
 describe('Component Tests', () => {
     describe('<%= entityAngularName %> Management Update Component', () => {
+        const activatedRouteData = new Subject();
+        const activatedRouteStub: any = {
+            data: activatedRouteData.asObservable(),
+        };
+
```
kaidohallik, line 57:
```suggestion
        let activatedRouteData: Observable<I<%= entityAngularName %>>;
```
I think it's more natural to construct desired `ActivatedRoute` and then call `comp.ngOnInit();` as this is happening in real workflow.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -26,64 +30,183 @@ import { HttpResponse } from '@angular/common/http';
...
         beforeEach(() => {
             TestBed.configureTestingModule({
                 imports: [HttpClientTestingModule],
                 declarations: [<%= entityAngularName %>UpdateComponent],
-                providers: [FormBuilder, ActivatedRoute]
+                providers: [
+                    FormBuilder,
+                    {
+                        provide: ActivatedRoute,
+                        useValue: activatedRouteStub,
```
kaidohallik, line 74:
```suggestion
                        useValue: { data: activatedRouteData },
```

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -26,64 +30,183 @@ import { HttpResponse } from '@angular/common/http';
...
+                        const <%= reference.name %> : I<%= otherEntity.entityAngularName %> = <%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>;
+                        <%= entityInstance %>.<%= reference.name %> = <%= reference.name %>;
+
+                        const <%= reference.name %>Collection: I<%= otherEntity.entityAngularName %>[] = [<%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'query').and.returnValue(of(new HttpResponse({ body: <%= reference.name %>Collection })));
+                        const expectedCollection: I<%= otherEntity.entityAngularName %>[] = [<%= reference.name %>, ...<%= reference.name %>Collection];
+                        spyOn(<%= otherEntity.entityInstance %>Service, 'add<%= otherEntity.entityAngularName %>ToCollectionIfMissing').and.returnValue(expectedCollection);
+
+                        // Trigger an ActivatedRoute change
+                        activatedRouteData.next({ <%= entityInstance %> });;
```
kaidohallik, line 161:
```suggestion
                        activatedRouteData = of({ <%= entityInstance %> });
                        comp.ngOnInit();
```

kaidohallik (changes requested):
I made some suggestions to use `ngOnInit();` instead of `updateForm`.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.spec.ts.ejs
```diff
@@ -26,64 +30,183 @@ import { HttpResponse } from '@angular/common/http';
...
+                const <%= entityInstance %>: I<%= entityAngularName %> = <%- generateTestEntityPrimaryKey(primaryKey, 1) %>;
+<%_ for (const relationshipsByEntityNeedingOptions of allRelationshipsByEntityNeedingOptions) { _%>
+    <%_ for (const relationship of relationshipsByEntityNeedingOptions) { _%>
+        <%_ const otherEntity = relationship.otherEntity _%>
+                const <%= relationship.reference.name %>: I<%= otherEntity.entityAngularName %> = <%- generateTestEntityPrimaryKey(otherEntity.primaryKey) %>;
+                <%= entityInstance %>.<%= relationship.reference.name %> = <%= relationship.collection ? `[${relationship.reference.name}]` : relationship.reference.name %>;
+    <%_ } _%>
+<%_ } _%>
+
+                comp.updateForm(<%= entityInstance %>);
```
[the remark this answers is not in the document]
mshima:
```suggestion
                activatedRouteData = of({ <%= entityInstance %> });
                comp.updateForm(<%= entityInstance %>);
```

mshima:
@kaidohallik I've tried your approach, but `useValue` is created at beforeEach and `activatedRouteData = of({ <%= entityInstance %> });` is only applied at next test (beforeEach).

The Subject approach creates a dynamic Observable, so when `activatedRouteData.next({ <%= entityInstance %> });` is called, the route subscription is triggered.

Others alternatives are:
- Use only one `<%= entityInstance %>` for every test.
- Create `before` steps to set the Observable instance.

Current Subject approach seems better to me.

kaidohallik (approved):
This PR is good to go.
Thanks @mshima
We can do follow up PR if we want to improve entity update test.

mshima:
@kaidohallik thanks

PR #13601, opened 2021-01-19
Revert "chore(deps-dev): bump css-loader from 3.5.3 to 4.3.0 in /generators/client/templates/vue"

mraible:
Reverts jhipster/generator-jhipster#13556


mraible:
As an alternative to this change, we could modify the Vue build config to work with the latest version of css-loader. That's probably a better approach since this upgrade has caused issues twice on the last week.

PR #13663, opened 2021-01-21
update commands usage

deepu105:
---

### DIFF
--- a/cli/commands.js
+++ b/cli/commands.js
@@ -72,11 +72,13 @@ const defaultCommands = {
     options: [
       {
         option: '--fork',
-        desc: 'Run generators using fork',
+        desc:
+          'Generate multiple applications in parallel without any interaction. This is default when every application in the JDL is new. Files will be overwritten without prompt. The "--incremental-changelog" option might not work correctly when this option is used for existing applications',
       },
       {
         option: '--interactive',
-        desc: 'Run generation in series so that questions can be interacted with',
+        desc:
+          'Generate multiple applications in series so that questions can be interacted with. This is the default when there is an existing application configuration in any of the folders',
       },
       {
         option: '--json-only',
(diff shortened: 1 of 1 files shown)


mshima:
Worth noting:
- fork/interactive only applies for jdl with multiples applications (microservice).
- Fork is default for all new applications. While if 1 application existed interactive is used.

deepu105:
@mshima ah ok, I was running on top of existing app probably that's why I thought the default was changed. Let me clarify it in PR

cli/commands.js
```diff
@@ -108,7 +108,7 @@ const defaultCommands = {
             },
         ],
         desc: `Create entities from the JDL file/URL/content passed in argument.
-  By default everything is run in series. If you like to run things faster use '--fork' flag to run in parallel.`,
+  By default everything is run in parallel for new applications and in series when there is any existing application in the folder. If you like to run things faster use '--fork' flag to run in parallel or '--interactive' flag to run in series.`,
```
mshima, line 113:
```suggestion
  When re-generating multiple applications in same jdl consider using fork option.`,
```
Or drop it, IMO is just complicates everything, the regeneration is not too long to worth.

cli/commands.js
```diff
@@ -72,11 +72,11 @@ const defaultCommands = {
         options: [
             {
                 option: '--fork',
-                desc: 'Run generators in parallel without any interaction. Files will be overwritten without prompt',
+                desc: 'Run generators in parallel without any interaction. Files will be overwritten without prompt. This is default for new applications',
```
mshima:
```suggestion
                desc: `Generate multiple applications in parallel without any interaction (each jhipster instance will run in a separated process).
  Files will be overwritten without prompt. This is the default when every application is new. This option makes --incremental-changelog to not work correctly.`,
```

cli/commands.js
```diff
@@ -72,11 +72,11 @@ const defaultCommands = {
...
             },
             {
                 option: '--interactive',
-                desc: 'Run generation in series so that questions can be interacted with. This is the default',
+                desc: 'Run generation in series so that questions can be interacted with. This is the default when there is an existing application in the folder',
```
mshima:
```suggestion
                desc: `Generate multiple applications in series so that questions can be interacted with.
  This is the default when there is an existing application configuration in some of the applications folders`,
```

deepu105:
@mshima the descriptions should be good now

mshima (approved):
LGTM

PR #13718, opened 2021-01-25
improve check for existing heroku app

ruddell:
If the `herokuAppName` is set, we check for the existence of the app.  This only works if it has previously been deployed from the same device and directory (it checks for the Heroku git remote).  If a developer clones the project from a repository then tries to run `jhipster heroku`, it will not find the existing app and reset the `herokuAppName` in the `yo-rc.json`

By passing the `herokuAppName` as a param to `heroku apps:info`, we can find out if the app exists. We also find out if the user has access to the application, otherwise an error is returned.

---

### DIFF
--- a/generators/heroku/index.js
+++ b/generators/heroku/index.js
@@ -105,7 +105,7 @@ module.exports = class extends BaseBlueprintGenerator {
         const done = this.async();
 
         if (this.herokuAppName) {
-          ChildProcess.exec('heroku apps:info --json', (err, stdout) => {
+          ChildProcess.exec(`heroku apps:info --json ${this.herokuAppName}`, (err, stdout) => {
             if (err) {
               this.abort = true;
               this.log.error(`Could not find application: ${chalk.cyan(this.herokuAppName)}`);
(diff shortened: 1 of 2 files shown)


mraible:
Is it possible that this change could be causing [this failure in `npm test`](https://github.com/jhipster/generator-jhipster/pull/13739/checks?check_run_id=1773419768#step:9:12935)?

```
  1) JHipster Heroku Sub Generator
       microservice application
         with JAR deployment
           "before each" hook for "creates expected files":
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/generator-jhipster/generator-jhipster/test/heroku.spec.js)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

ruddell:
The only time `heroku.spec.js` reaches the [`heroku app:info` command](https://github.com/jhipster/generator-jhipster/blob/6dd921c9aa37541825d42ba3b33461ad3e50bcd7/generators/heroku/index.js#L108) is when there is an existing app, covered by the `with existing app` spec.

This PR originally failed CI ([logs](https://github.com/ruddell/generator-jhipster/runs/1765621994?check_suite_focus=true#step:9:12943)) at the expected spec `with existing app` with a similar log, but was fixed by my change to the test.

The `main` branch passes when I run `npm test test/heroku.spec.js `, so I don't think this is related.  

That error message is reproducible locally with the `spring-boot-2.4` branch.  I tried to find the issue but nothing stood out.  It looks like `end()` is never called for some reason.

PR #13740, opened 2021-01-27
[Protractor] Remove sleep because React doesn't need it

mraible:
Still trying to fix the failing [React + OAuth 2.0](https://github.com/hipster-labs/jhipster-daily-builds/actions/runs/513038454) daily build. 

NOTE: We might be able to reduce the e2e times on these jobs by specifying a lighter set of entities, or none at all. Currently, there's a lot.

Screen Shot 2021-01-26 at 6 55 42 PM


mraible:
Merging because [Monolith OAuth 2.0](https://github.com/hipster-labs/jhipster-daily-builds/runs/1774088023?check_suite_focus=true) passed with this branch.

PR #13754, opened 2021-01-28
Disable metrics as a Protractor test

mraible:


mraible:
Test run happening at https://github.com/hipster-labs/jhipster-daily-builds/actions/runs/519460070

generators/client/templates/angular/src/test/javascript/e2e/admin/administration.spec.ts.ejs
```diff
@@ -63,9 +63,9 @@ describe('administration', () => {
   });
   <%_ } _%>
 
+  <!-- This test tends to fail with Protractor, so we've disabled it. Please enable if it passes for you!
```
pascalgrimaud, line 66: the comment is not correct, as it should be `//` or `/*` instead

mraible: Doh!

mraible:
Test run passed! ✅ 🕺

PR #13824, opened 2021-02-03
[react] Enable use of custom id fields for react - not 100% finished

gzsombor:
---

### DIFF
--- a/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
+++ b/generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
@@ -53,7 +53,7 @@ export const <%= entityReactName %>Detail = (props: I<%= entityReactName %>Detai
     <Row>
       <Col md="8">
         <h2 data-cy="<%= entityInstance %>DetailsHeading">
-          <Translate contentKey="<%= i18nKeyPrefix %>.detail.title"><%= entityClass %></Translate> [<strong>{<%= entityInstance %>Entity.id}</strong>]
+          <Translate contentKey="<%= i18nKeyPrefix %>.detail.title"><%= entityClass %></Translate> [<strong>{<%= entityInstance %>Entity.<%= primaryKey.name %>}</strong>]
         </h2>
           <dl className="jh-entity-details">
           <%_ for (field of fields.filter(field => !field.id)) {
@@ -122,6 +122,7 @@ export const <%= entityReactName %>Detail = (props: I<%= entityReactName %>Detai
                 const relationshipFieldNamePlural = relationship.relationshipFieldNamePlural;
                 const relationshipNameHumanized = relationship.relationshipNameHumanized;
                 const otherEntityName = relationship.otherEntityName;
+                const otherEntityPkName = relationship.otherEntity.primaryKey.name;
                 const otherEntityStateName = relationship.otherEntityStateName;
                 const otherEntityField = relationship.otherEntityField;
                 const otherEntityFieldCapitalized = relationship.otherEntityFieldCapitalized;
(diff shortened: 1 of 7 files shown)


generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
```diff
@@ -230,7 +230,7 @@ export const <%= entityReactName %>Update = (props: I<%= entityReactName %>Updat
           { !isNew ?
             <AvGroup>
               <Label for="<%= entityFileName %>-id"><Translate contentKey="global.field.id">ID</Translate></Label>
-              <AvInput id="<%= entityFileName %>-id" type="text" className="form-control" name="id" required readOnly/>
+              <AvInput id="<%= entityFileName %>-id" type="text" className="form-control" name="<%= primaryKey.name %>" required readOnly/>
```
mshima, line 233: At angular I've removed the special case for id.
Removed the id filter, and added readOnly attribute when `field.id` is true.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity.tsx.ejs
```diff
@@ -303,7 +303,7 @@ export const <%= entityReactName %> = (props: I<%= entityReactName %>Props) => {
             <Table responsive>
               <thead>
                 <tr>
-                  <th <% if (pagination !== 'no') { %> className="hand" onClick={sort('id')} <%_ } _%>><Translate contentKey="global.field.id">ID</Translate><% if (pagination !== 'no') { %> <FontAwesomeIcon icon="sort" /><% } %></th>
+                  <th <% if (pagination !== 'no') { %> className="hand" onClick={sort('<%= primaryKey.name %>')} <%_ } _%>><Translate contentKey="global.field.id">ID</Translate><% if (pagination !== 'no') { %> <FontAwesomeIcon icon="sort" /><% } %></th>
                   <%_ for (field of fields.filter(field => !field.id)) { _%>
```
mshima, line 307: Remove this line and remove the id filter bellow.
```suggestion
                  <%_ for (field of fields) { _%>
```

mshima (commented):
Looks ok to me the only problem is the `generator-base-private.js` change.
Other changes are meant to prepare for composite id support, when `primaryKey` will be used for identifying the current entity.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
gzsombor: Good idea, but I think, that wont be enough, because we need to edit the id field - if it's a user assigned id, instead of the Long id, which generated by the database.

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
```diff
@@ -227,17 +227,11 @@ export const <%= entityReactName %>Update = (props: I<%= entityReactName %>Updat
...
-              <AvInput id="<%= entityFileName %>-id" type="text" className="form-control" name="id" required readOnly/>
-            </AvGroup>
-            : null
-          }
-          <%_ for (field of fields.filter(field => !field.id)) {
+          <%_ for (field of fields) {
               const fieldType = field.fieldType;
               const fieldName = field.fieldName;
               const fieldNameHumanized = field.fieldNameHumanized;
+              const readonlyIfGeneratedId = (field.id && field.fieldType === 'Long' ? 'readOnly' : '');
```
mshima, line 234: You will probably want to add support to UUID to readOnly too.

mshima (commented):
You will need to change translation key from `<%= i18nKeyPrefix %>.<%= fieldName %>` to `<%= field.fieldTranslationKey %>`

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-update.tsx.ejs
gzsombor: I think, it's better if I check for databaseType !== sql || field.jpaGeneratedValue - so if the id is not generated by the backend, the user could specify here.

mshima (approved):
LGTM

generators/entity-client/templates/react/src/main/webapp/app/entities/entity-detail.tsx.ejs
```diff
@@ -150,7 +151,7 @@ export const <%= entityReactName %>Detail = (props: I<%= entityReactName %>Detai
     {
         (<%= entityInstance %>Entity.<%= relationshipFieldNamePlural %>) ?
             (<%= entityInstance %>Entity.<%= relationshipFieldNamePlural %>.map((val, i) =>
-                <span key={val.id}><a>{val.<%= otherEntityField %>}</a>{(<%= entityInstance %>Entity.<%= relationshipFieldNamePlural %> && i === <%= entityInstance %>Entity.<%= relationshipFieldNamePlural %>.length - 1) ? '' : ', '}</span>
+                <span key={val.<%= otherEntityPkName %>}><a>{val.<%= otherEntityField %>}</a>{(<%= entityInstance %>Entity.<%= relationshipFieldNamePlural %> && i === <%= entityInstance %>Entity.<%= relationshipFieldNamePlural %>.length - 1) ? '' : ', '}</span>
```
mshima, line 154: @gzsombor there is no primaryKey for embedded entities.
See https://github.com/hipster-labs/jhipster-daily-builds/runs/1843126537?check_suite_focus=true

gzsombor: Interesting, I will have a look on it!

PR #13826, opened 2021-02-04
Improvements to frontend-maven-plugin

mshima:
- We don't support yarn so, this PR remove references that breaks yarn support even more.
npm 7 have support for yarn.lock, so if there is a yarn.lock, npm will use it just fine.
- Having node in the root of project is confusing, move it to target.

---


atomfrede:
Would you like to do the same things for yarn/npm in the [gradle build](https://github.com/jhipster/generator-jhipster/blob/1ef6be199e7558d860630870731698764b6298c3/generators/server/templates/gradle/profile_dev.gradle.ejs#L50) too? Otherwise I can have look.

vishal423:
This doesn't seem correct to me. If you move the node installation directory under the target, then, every clean build would require downloading a new nodejs binary. It would be better to explain the rationale behind this change and what issues it would prevent?

mshima:
> This doesn't seem correct to me. If you move the node installation directory under the target, then, every clean build would require downloading a new nodejs binary. It would be better to explain the rationale behind this change and what issues it would prevent?

Fixed webapp operations triggering backend cleanup in last commit.

It doesn't fixes any problem just some confusion about why it's there, node directory is a temporary folder just like everything inside target.
It can cause builds problem if it's broken, IMO a cleanup operation should remove it too.

gmarziou:
I agree with @vishal423, I can't see any big benefit of moving node folder under target but I can see several drawbacks: 

* unable to build offline
* slower CI builds
* more load on node mirror sites

Also, what is the equivalent for our gradle build?

pascalgrimaud:
I didn't pay attention to that.
Indeed, node folder should not be inside target folder as it is regularly clean during dev
No real benefits

mshima:
> I agree with @vishal423, I can't see any big benefit of moving node folder under target but I can see several drawbacks:
> 
> * unable to build offline
> * slower CI builds
> * more load on node mirror sites

@gmarziou I really don't see those points as valid.
- node directory is ignored by git, so cloning the repository CI will have to install it, doesn't change 'unable to build offline'
```
 git status --ignored
On branch master
Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
	.classpath
	.project
	.settings/
	node/
	node_modules/
	target/
```
- node and npm are cached, so no 'more load on node mirror sites'
```
$ find ~/.m2/repository/com/github/eirslett/node
/Users/mshima/.m2/repository/com/github/eirslett/node
/Users/mshima/.m2/repository/com/github/eirslett/node/12.18.3
/Users/mshima/.m2/repository/com/github/eirslett/node/12.18.3/node-12.18.3-darwin-x64.tar.gz
/Users/mshima/.m2/repository/com/github/eirslett/node/12.14.0

$ find ~/.m2/repository/com/github/eirslett/npm/
/Users/mshima/.m2/repository/com/github/eirslett/npm/
/Users/mshima/.m2/repository/com/github/eirslett/npm//7.5.2
/Users/mshima/.m2/repository/com/github/eirslett/npm//7.5.2/npm-7.5.2.tar.gz
/Users/mshima/.m2/repository/com/github/eirslett/npm//6.9.0
```
- for 'slower CI builds' it would have to remote target folder in the workflow, and each unpackage takes 3-4 secs?

IMO it's better to remove it and recreate, than risking a broken node/* without having clue how to fix it.

gmarziou:
You got a valid point, I did not know this plugin used maven local repository for caching binaries it downloaded from nodejs.org. So I change my mind and agree with your point of view.

Although I disagree on this point:

> IMO it's better to remove it and recreate, than risking a broken node/* without having clue how to fix it.

If you have a broken node/*, it's probably because file was corrupted in your local maven repository, in which case `mvnw clean` won't help either because you will get back the corrupted copy from cache. You'll know how to fix it only if you know about this feature which I did not find in the documentation but only in source code.

mshima:
I you want to revert, that’s ok to me, I can create the PR.
It won’t make too many difference.
I just think it’s a temporary folder so I would prefer to put inside target.

PR #13837, opened 2021-02-04
Fix Azure CI pipeline template for E2E tests

egvimo:
Our Azure Pipeline failed on E2E tests. These changes fixed it.

We had to enable the headless mode via env variable and install the latest Chrome, because the Azure Ubuntu VM does not include it.

Furthermore we had to change the `protractor.conf.js` file and add the `--no-sandbox` parameter to the Chrome args. I did not know whether I should include this change to the PR, so I left it out for now. But I update the PR and add this if you wish (probably all 3 template files should be changed - for Angular, React and Vue).

---

### DIFF
--- a/generators/ci-cd/templates/azure-pipelines.yml.ejs
+++ b/generators/ci-cd/templates/azure-pipelines.yml.ejs
@@ -26,6 +26,7 @@ jobs:
     SPRING_JPA_SHOW_SQL: false
     JHI_DISABLE_WEBPACK_LOGS: true
     NG_CLI_ANALYTICS: "false"
+    JHI_E2E_HEADLESS: true
 
   steps:
 <%_ if (!skipClient) { _%>
--- a/generators/ci-cd/templates/azure-pipelines.yml.ejs
+++ b/generators/ci-cd/templates/azure-pipelines.yml.ejs
@@ -38,6 +38,16 @@ jobs:
       versionSpec: $(NODE_VERSION)
     displayName: 'TOOLS: install Node.js'
 <%_ } _%>
+<%_ if (testFrameworks.includes('cypress') || testFrameworks.includes('protractor')) {
+  const commented = applicationType === 'gateway' ? '# ' : ''; _%>
+  <%= commented %>- task: Bash@3
+  <%= commented %>  inputs:
+  <%= commented %>    targetType: 'inline'
+  <%= commented %>    script: |
+  <%= commented %>      wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+  <%= commented %>      sudo apt install ./google-chrome-stable_current_amd64.deb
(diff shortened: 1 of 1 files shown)


generators/ci-cd/templates/azure-pipelines.yml.ejs
```diff
@@ -36,6 +37,16 @@ jobs:
     inputs:
       versionSpec: $(NODE_VERSION)
     displayName: 'TOOLS: install Node.js'
+<%_ } _%>
+<%_ if (testFrameworks.includes('cypress') || testFrameworks.includes('protractor')) {
+  const commented = applicationType === 'gateway' ? '# ' : ''; _%>
+  <%= commented %>- task: Bash@3
```
pascalgrimaud, line 43: To keep consistency, we could do the same than in our main azure-pipelines, see https://github.com/jhipster/generator-jhipster/blob/main/azure-pipelines.yml#L112-L117

```yaml
      - script: |
          wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
          sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
          sudo apt update
          sudo apt install google-chrome-stable
        displayName: 'TOOLS: install google-chrome-stable'
```

What's the best in your opinion ?
- the code you propose here
- or the one in our CI ?

egvimo: I would prefer the code I proposed, because Azure uses "throw away VMs" and it is not necessary to install the keys, update apt and so on...
But to keep consistency I can change it.

What do you think about the `--no-sandbox` flag? Should I also include these changes?

pascalgrimaud: I remember having some issue with no-sandbox... is it necessary ? If not, we can keep like this

pascalgrimaud:
Thanks and congrats for your 1st contribution @egvimo 
Let's keep your code here, then, we can update the main azure-pipelines in generator-jhipster later

generators/ci-cd/templates/azure-pipelines.yml.ejs
egvimo: We couldn't get it to run without the flag.
I will investigate further and do some tests, maybe there is another way. Let's keep it like this for now.

egvimo:
Thanks!
Should I create another PR for the change in the main Azure pipeline?

pascalgrimaud:
Yes plz @egvimo :)

generators/ci-cd/templates/azure-pipelines.yml.ejs
pascalgrimaud: Then, if it's needed, so we should add no-sandbox so

PR #13860, opened 2021-02-05
Implement json data generation for improved tests.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/7112 and https://github.com/jhipster/generator-jhipster/issues/14103

---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #13864, opened 2021-02-06
Implement composite id support at Java backend.

mshima:
This is a follow-up for https://github.com/jhipster/generator-jhipster/pull/13860.

Fixes https://github.com/jhipster/generator-jhipster/issues/7112.

---


yelhouti:
Hi @mshima , thanks again for all your help.

we took the time to read your code and we have a few remarks:
putting "trackByFileds" and "idMatrix" in entity.fields seems to be misleading because:
- trackByField is not really a field and in my opinion should be compted in the frontend at runtime
- idMatrix is only used and backend by JPA, not in DTO not in criteria, not in frontend, sincethe inod is already present in the relationship, and should be IMO not included in fields.
We would rather add an `ids` field/object in `primaryKey` with all the info needed to loop over it instead of `primaryKey.references` since reefrences do not contain some info we need and are used to group fields and relationships as you said we though we should name them differently.
we suggest to keep the getters and all the logic you created, to keep the yoeman flow even if the code is mich more complicated) and add our own without breaking the flow.
If you are ok with that we will fork create a branch locally from your PR with your getter logic and put and modify it to work with our template, but to avoid stepping on each others toes we can't work at the same time so please don't merge this for now.

What do you think?

mshima:
@yelhouti 

> putting "trackByFileds" and "idMatrix" in entity.fields seems to be misleading because:

trackByField is one field of entity.fields.
It's more likely a `@NaturalId`, so, a unique value to track the entity.

> * trackByField is not really a field and in my opinion should be compted in the frontend at runtime

trackByField is the field that should be used to identify a registry, so it should not be computed in the frontend.

> * idMatrix is only used and backend by JPA, not in DTO not in criteria, not in frontend, sincethe inod is already present in the relationship, and should be IMO not included in fields.
>   We would rather add an `ids` field/object in `primaryKey` with all the info needed to loop over it instead of `primaryKey.references` since reefrences do not contain some info we need and are used to group fields and relationships as you said we though we should name them differently.
>   we suggest to keep the getters and all the logic you created, to keep the yoeman flow even if the code is mich more complicated) and add our own without breaking the flow.
>   If you are ok with that we will fork create a branch locally from your PR with your getter logic and put and modify it to work with our template, but to avoid stepping on each others toes we can't work at the same time so please don't merge this for now.
> 
> What do you think?

- Composite id will not add too much value to our generated code.
- Our templates are already too complex and with too many features.
- Composite Ids calculated in the frontend will complicate the templates even more.
Unless we move the burden to the DTO like I did, I'm not sure we should proceed implementing composite id.
- This is the less intrusive implementation I could think of.
- References was created to simplify and unify fields and relationships, there are a few places it can be used, not sure if we should remove it.
- Manually created ids, has problems with our current persist strategy, it tries to updated an registry that it should insert throwing error. idMatrix can be used to differentiate an insert from an update.

You can see more details of trackByField at https://github.com/jhipster/generator-jhipster/pull/13860#issue-568555363.
If you see closely the entire https://github.com/jhipster/generator-jhipster/pull/13860 PR, you will not see a single reference to composite ids.
That PR adds angular support to:
- natural id field
- hidden fields
- autoGenerated fields

Those features were included at every id field (1 per entity), so by treating a id field just like an ordinary field, we need to add support to them for every field.
A complicated PR is unlikely to be merged unless the developer is confident and merge itself.
https://github.com/jhipster/generator-jhipster/pull/13860 is somewhat harmless, but not sure someone will actually merge it.

yelhouti:
@mshima I'd rather not handle a new `@NaturalId` in the backend, since I already manage to work without one, so please do not merge that PR.

> Composite id will not add too much value to our generated code.

People disagree it's one of the most used blueprints.

> Our templates are already too complex and with too many features

I agree but many or much less important in my opinion

> Composite Ids calculated in the frontend will complicate the templates even more.
Unless we move the burden to the DTO like I did, I'm not sure we should proceed implementing composite id.

I don't think so, please check my blueprint I have been using compositeId for more than a Year.

> This is the less intrusive implementation I could think of.

Please have a look at our template files they are not that different/complicated.

> References was created to simplify and unify fields and relationships, there are a few places it can be used, not sure if we should remove it.

I never said we should remove them, they are great in entity and result in much cleaner code, but you added them to primaryKey two, and I was saying primaryKey needs have something different (or even both).

> Manually created ids, has problems with our current persist strategy, it tries to updated an registry that it should insert throwing error. idMatrix can be used to differentiate an insert from an update.

We are already using idMatrix without having it in the frontend... please do not try to recode everything yourslef the code is already here and working and can be ported on top of the way you calculate primaryKey in the entity.js file, with some very minor changes.

> A complicated PR is unlikely to be merged unless the developer is confident and merge itself.

I am a developer and I am confident I can make a PR that works the way we all want if we just stop stepping on each other toes.

PR #13942, opened 2021-02-12
Add Okta message after generation as part of platinum sponsorship

mraible:
Before:

Screen Shot 2021-02-12 at 10 35 15 AM

After:

Screen Shot 2021-02-12 at 11 03 25 AM


mraible:
I'm not sure if cyan is the best color to use. Here's what it looks like when the terminal has a white background. Blue looks too dark (IMO) when the background is black.

Screen Shot 2021-02-12 at 11 23 37 AM

slade13:
Cyan looks good imo, and it's readable on the white background.

pascalgrimaud:
The cyan color depends on the Terminal and how the user customized it. So no worry about that:

Capture d’écran de 2021-02-13 09-58-28

PR #14147, opened 2021-03-03
Renames isCurrentUserInRole method

### LINKED ISSUE #13994: Use "hasRole" instead of "hasAuthority" in SecurityConfiguration?

I'm trying to understand the security model of spring-security, in particular I was looking for a clarification over the concepts of *authority* and *role*.

In [this article](https://www.baeldung.com/spring-security-granted-authority-vs-role) I've found the following explanation:
> ...in Spring Security, we can think of each Role as a coarse-grained GrantedAuthority that is represented as a String and prefixed with “ROLE“. When using a Role directly, such as through an expression like **hasRole(“ADMIN”)**, we are restricting access in a coarse-grained manner.

Since the generated project is not restricting resources in a fine-grained manner (it couldn't be otherwise, since it doesn't know the business requirements), I would expect it to use `hasRole(...)` instead of `hasAuthority(...)`.

This is one of the lines of code which could be affected by this change proposal:

https://github.com/jhipster/generator-jhipster/blob/5bdc374bcaf3553984524f19beaf90d1728aff43/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L278

The usage of `hasRole(...)` in the generated code would allow any new JHipster user to understand more easily the security model behind the scenes.

I can do the PR if the change proposal makes sense.

- [x ] Checking this box is mandatory (this is just to show you read everything)

nickshoe:
Closes #13994 

---

### DIFF
--- a/generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
@@ -137,14 +137,12 @@ public final class SecurityUtils {
     }
 
     /**
-     * If the current user has a specific authority (security role).
-     * <p>
-     * The name of this method comes from the {@code isUserInRole()} method in the Servlet API.
+     * Checks if the current user has a specific authority.
      *
      * @param authority the authority to check.
      * @return true if the current user has the authority, false otherwise.
      */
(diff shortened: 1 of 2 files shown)


pascalgrimaud:
Thanks @nickshoe

PR #14156, opened 2021-03-04
Moving forms to service

### LINKED ISSUE #10198: [Feature Request] - Frontend: Move FormGroup to service
When having entities with many fields, having them in the `entity-update.component.ts` file could be tedious. Also, when coding side-by-side it will force me to duplicate my code for creating the form, updating the form and for creating the entity from form (those can be big methods).

This change would allow anyone to work _side-by-side_ without repeating code, by just calling the service. For instance, I have moved the code to the service and I have a method like this:

``` 
createFromForm(editForm: FormGroup): IInvoice {
    return {
        ...new Invoice(),
        id: editForm.get(['id']).value,
        (...many other variables....)
        };
    }
``` 

and I would just call

`let invoice = this.invoiceService.createFromForm(this.editForm);`

I think this will improve readability, would make the code even cleaner, and again, it would stop repeated code in the front end.

None that I know of.

amatosg:
This PR fixes #10198 

By moving the create/update form group to the service layer, users will be able to reuse those methods avoiding code repetition and it will also ease the side-by-side development.

---


amatosg:
I'm getting `UnhandledPromiseRejectionWarning: TypeError: Converting circular structure to JSON` [here ](https://github.com/jhipster/generator-jhipster/pull/14156/checks?check_run_id=2028970768 )and I don't see why is this happening.

mshima:
IMO form doesn't belong to the service, it should be static methods in the update component.

amatosg:
the idea is to simplify the update component and promote side-by-side development, thus reducing code duplication

mshima:
Sure I understand, but IMO those operations don't belongs to the service.
They are methods helpers and can be used as:
```
import {createFromForm, createEditForm, updateForm} from "./entity-update.component";
```

Actually I have a branch that implements something like that, I've used for creating nested/modal forms.

amatosg:
When using `import {createFromForm, createEditForm, updateForm} from "../original-entity/entity-update.component";`  I get ` TS2305: Module '"../original-entity/new-entuty-update.component"' has no exported member 'createFromForm'.`

mshima:
I am trying to say that instead of implementing it in the service, I prefer as an exported method in `entity-update.component.ts` file.
If you want I can take a look in the branch and see if I can create a branch isolating those changes.

amatosg:
I would love to. Thanks!

mshima:
@amatosg https://github.com/jhipster/generator-jhipster/pull/14182

amatosg:
@mshima awesome! thank you. Well, bye bye bounty 😆

vishal423:
I used to implement something like this: https://github.com/vishal423/ngx-hipster/blob/master/demo/ngx-hipster-sample/src/app/movie/movie-detail/movie-form.service.ts

amatosg:
> I used to implement something like this: https://github.com/vishal423/ngx-hipster/blob/master/demo/ngx-hipster-sample/src/app/movie/movie-detail/movie-form.service.ts

@mshima what do you think about having this approach? it would make the update component cleaner reducing "noise" and would also comply with the single responsibility principle

mshima:
Creating a service for this seems overkill to me.

amatosg:
it might seem like overkill, but it means an improvement in terms of cleaner code and also complying with the single responsibility principle

mshima:
@amatosg I was going to split the form to another file and I‘ve changed my mind.
We should put in a service.
Due to:
https://github.com/jhipster/generator-jhipster/blob/dadae3f07991322f9133fa36cc6cbecdb14c3976/generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.ts.ejs#L47

For microfrontend support we will need to move constants to the AplicationConfigService to make sure the value is shared instead duplicated.

Do you want to implement it?
I prefer to switch away from form builder. But is a preference.

PR #14184, opened 2021-03-05
Upgrade Couchbase support for Spring Boot 2.4

mraible:
Fixes #13544.

Replaces https://github.com/jhipster/generator-jhipster/pull/13770.

---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
- [x] Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

Issue #13797, opened 2021-02-01
Updating an entity with a relationship deletes all the fields in the other entity (React + Neo4j)


utnaf:
Hey everyone! Thanks for the great job on JHipster!

##### Trying to update an entity with a relationship cause the other entity to be empty

When using Neo4j and React (at least happened to me with this combo) and with two entities with a OneToMany relationship (Foo - HAS_MANY -> Bar), the update of Bar with a Foo selected in the select box causes all Foo fields (except for the ID) to disappear.

##### **Motivation for or Use Case**

Updating an entity should not empty the related entity. Same setup but with Angular instead of React works fine.

##### **Reproduce the error**

So I created an app with React and Neo4j. Then I created two entities, let's say:

```
entity Exam {
	title String
}

entity Question {
	content String
}
```

and I related them

```
relationship OneToMany {
  Exam{questions} to Question{exam}
}
```

Then I create an Exam named "Foo", I create a Question and connect it with Foo using the select input, and I save.
Everything goes smoothly apparently.

<img width="943" alt="Screenshot 2021-02-01 at 21 35 36" src="https://user-images.githubusercontent.com/1318539/106515108-747b9580-64d5-11eb-8adf-7480f15ba336.png">

But when I got back to the exam

<img width="489" alt="Screenshot 2021-02-01 at 21 36 09" src="https://user-images.githubusercontent.com/1318539/106516507-6e86b400-64d7-11eb-88e7-edf0d344aee8.png">

The content of the entity is empty.

##### **Related issues**

Couldn't find any

##### **Suggest a Fix**

I think the bit that causes issues is `question-update.tsx:93-102`

```
<AvInput id="question-exam" type="select" className="form-control" name="exam.id">
  <option value="" key="0" />
  {exams
    ? exams.map(otherEntity => (
        <option value={otherEntity.id} key={otherEntity.id}>
          {otherEntity.id}
        </option>
      ))
    : null}
</AvInput>
```

Having the option value `{otherEntity.id}` leads to have a PUT request body like

```
{
  "id": "123",
  "text": "My first question",
  "exam": {
    "id": "456",
  }
}
```

The backend doesn't "refresh" the Exam entity so when it saves the question it just overwrite this json with the existing node on the db. This is my guess.

##### **JHipster Version(s)**

6.10.5

##### **JHipster configuration**

  <details>
  <summary>JDL definitions</summary>
  <pre>
application {
  config {
    applicationType monolith
    authenticationType jwt
[...]

PR #14186, opened 2021-03-05
Disable Couchbase and move WebFlux and Infinispan out of beta

mraible:
Disable Couchbase as a prompt so v7 does not prompt users. Can be re-enabled once [Couchbase for Spring Boot 2.4](https://github.com/jhipster/generator-jhipster/pull/14184) is fixed.

---

### DIFF
--- a/generators/server/prompts.js
+++ b/generators/server/prompts.js
@@ -150,7 +150,7 @@ function askForServerSideOpts() {
         }); */
         opts.push({
           value: 'neo4j',
-          name: 'Neo4j',
+          name: '[BETA] Neo4j',
         });
         opts.push({
           value: 'no',
--- a/generators/server/prompts.js
+++ b/generators/server/prompts.js
@@ -44,7 +44,7 @@ function askForServerSideOpts() {
       when: () => ['monolith', 'microservice'].includes(applicationType),
       type: 'confirm',
       name: 'reactive',
-      message: '[Beta] Do you want to make it reactive with Spring WebFlux?',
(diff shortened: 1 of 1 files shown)


mraible:
@jhipster/developers Are y'all OK with moving WebFlux, Neo4j, and Infinispan out of beta?

mshima:
https://github.com/jhipster/generator-jhipster/issues/13797 seems a blocker for Neo4j.

mraible:
@mshima I added a bug bounty to the issue. I'm also seeing issues with Neo4j when running `jhipster jdl reactive-ms` and running e2e tests. They fail when talking to the blog app that uses Neo4j. If I change the gateway to use Cypress, the tests still fail. These tests passed with beta 1.

jdubois:
At least Webflux should be stable now, but I’m not totally sure of its current quality level.

atomfrede:
Regarding the neo4j issue I will check with Michael and Gerrit. I remember something similar when we implement it and I think it's something how neo4j/neo4j spring integration handles relations. So maybe we need to prevent using dtos to always have the full object.

pascalgrimaud:
Webflux -> we have no choice, because our gateway is by default with Reactive. It's pretty well tested with the JDL daily builds, so I'm confident, we can remove "beta"
Infinispan -> very few feedbacks, so I think it works

Neo4J -> not sure it's ready yet, we still have some issues
Couchbase -> broken for a while

mraible:
@pascalgrimaud I re-added Beta to Neo4j.

PR #14273, opened 2021-03-11
Replace asEntity/asDto by pre calculated variables.

mshima:
---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #14275, opened 2021-03-12
primaryKey.ids to prepare blueprints for composite keys

yelhouti:
---


yelhouti:
@deepu105 I spent the night on it, I might still tweak some of the JS but not the templates anymore, so please have look and let me know if there is anything that you don't like.

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -48,7 +48,7 @@
             <%_ for (relationship of relationships) {
                 if ((relationship.relationshipType === 'many-to-one' || (relationship.relationshipType === 'one-to-one' && relationship.ownerSide === true))
                             && !relationship.id) {
-                    relationship.otherEntity.primaryKey.ownFields.forEach(idField => {
+                    relationship.otherEntity.primaryKey.ids.map(id => id.field).forEach(idField => {
```
deepu105, line 51: seems like `ids.map(id => id.field)` is repeated everywhere, can we do this instead
when populating `relationship.otherEntity.primaryKey.ids` & `entity.primaryKey.ids` do a `map(id => id.field)` and collect that into `relationship.otherEntity.primaryKey.ownFields` and `entity.primaryKey.fields` so that these usages need not be touched and would be much nicer to read and use.

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
```diff
@@ -73,19 +73,19 @@
         <%_ for (relationship of joinTableRelations) { _%>
 
         <createTable tableName="<%= relationship.joinTable.name %>">
-            <%_ for (field of relationship.otherEntity.primaryKey.fields) { _%>
```
deepu105, line 76: is `ownFields` and `fields` availabel from `ids`?

deepu105 (changes requested):
@mshima seems like internals changed a lot from when I last looked, can you also take a look at this and see if its OK please

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
yelhouti: In reality .ids should all be handled te same way, there should be no place where we handle ids that comme from a fields or a relationship. This is why I switched to ids.map(id => id.field) which is the original field in the entity it comes from.

But the current code contains distinctions, and If I want to have the least number of changes I have to sheet like that.
ownFields "represented" the id fields that come from the entity (it is why it was removed). Maybe we could have and `primaryKey.originalFields` which is easier to understand. But no matter what, this is not generalized enough to work with composite because you specifically instructed against that.

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
yelhouti: Please read comment above.

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
deepu105: What does it have to do with my comment. I don't want this map method repeated all over templates and would like to have it premapped on the field itself as seprate member thats it

yelhouti: @deepu105 you asked to rename it back to `ownFields` I proposed to renameIt `originalFields` and explained wh the previous variable ownField should not exist and tried to provide more context. Will make the change right away.

mshima:
> @mshima seems like internals changed a lot from when I last looked, can you also take a look at this and see if its OK please

@deepu105 I looked and there are some problems that I still don't agree.
- The fallback ID don't use the correct translation and humanized name, (easy to fix).
- It creates a `otherEntityFields`, which each frontend will have to implement a logic for showing them at selects.
IMO it's much easier to return the label ready to frontend, using mapstruct and keep compatible with 1 `otherEntityField`.
- It creates `ids` field, that is seems just like references to me.
- I still prefer to don't use eager calculation approach like `computeRelationshipOtherEntityFields` uses.

I had more thought and realized that all this primaryKey discussion is related to a server implementation and not to both server and client. The server must implement all the complexity and client just have to support it.
So to allow blueprint to easily customize primaryKey I will:
1. Move the all primaryKey related code out of `prepareEntityForTemplates` to `prepareEntityPrimaryKeyForTemplates`.
2. Move the task related to primaryKey from entity generator (which is used for server and backend) to entity-server generator. 
3. Create another priority called `preparingFields`, we currently have `preparing` and `preparingRelationships`.
`prepareEntityPrimaryKeyForTemplates` adds a field, so for the workflow to work correctly, `preparing` should be execute in every blueprint before moving to `preparingFields`.

A blueprint will just needs to override the prepareEntityPrimaryKeyForTemplates at entity-server and use it's own logic to generate the primaryKey.

I will change it once https://github.com/jhipster/generator-jhipster/pull/14273 is merged or denied 😄.

yelhouti:
@mshima thanks for the feedback, here is what I think:
> he fallback ID don't use the correct translation and humanized name, (easy to fix). 

I guess you are right in the case of OneToOne I could copy the translateKey from the originalField two since we are copying the name. I choose not to do so to allow for a different translation, but I am ok if you change that.

> It creates a otherEntityFields, which each frontend will have to implement a logic for showing them at selects.

IMO it's much easier to return the label ready to frontend, using mapstruct and keep compatible with 1 otherEntityField
I would rather not send the same info twice, and since the core generator doesn't support compositeKeys it is not concerned by this change, so please keep this fields for me or anyone else to use, and do not force added fields in DTOs. does who would want them in there blueprints would add them.

> It creates ids field, that is seems just like references to me.

I already talk about this in our exchanges, these are very different, an "id" represent ONE original field, for the same relationship we can have multiple ids. a reference represents one relationship which could results in MULTIPLE original fields.
When I was first implementing I though the same, but in complexe scenarios they are not. Please keep ids as a different thing. 

> I still prefer to don't use eager calculation approach like computeRelationshipOtherEntityFields uses.

You can change this I really don't mind.

> I had more thought and realized that all this primaryKey discussion is related to a server implementation and not to both server and client. 

It might seem that way, but in my experience I needed to know the ids in the frontend few times:
1- to send a request, the matrixVariable in the queryParams needs the info (for getOne and Delete)
2- to mark them as readOnly on update
...

> 2. Move the task related to primaryKey from entity generator (which is used for server and backend) to entity-server generator.

Please do not touch only the backend and expect the frontend to adapt, unless you don't have a fully working example with e2e tests I suggest you don't work on this.
Also, it seems like you are not interested on the this future, and I am more than willing to put on the work, so just assign it it to me or discuss it with me please if you don't like something, I am sure we can find a solution that suits everybody's needs.

> 3. Create another priority called preparingFields, we currently have preparing and preparingRelationships.
prepareEntityPrimaryKeyForTemplates adds a field, so for the workflow to work correctly, preparing should be execute in every blueprint before moving to preparingFields.

I am for some version of this, just please remember that in my blueprints I prefer to have the derivedField have a name like relationshipName+otherEntityField, and if you add the field too late I want be able to that, I am willing to help on this.

mshima:
> Please keep ids as a different thing.

The naming does not seem right.

> You can change this I really don't mind.

Are you aware how request changes works?
Anyway, with my proposed changes, it will be possible to implement this in the blueprint.

> 1- to send a request, the matrixVariable in the queryParams needs the info (for getOne and Delete)

As I said, if the server returns it ready, it's not needed.
Per discussion, composite id it not going to be fully implemented, adding a otherEntityFields in the future will not be a breaking change.

> 2- to mark them as readOnly on update

Every `@Id` field is already readOnly on update.
It should be implemented as a generic `updatable === false`.

> Please do not touch only the backend and expect the frontend to adapt

Implementation in the frontend should be generic.
The ideal frontend implementation would be to only customize getEntityIdentifier, and every other place use it.

> unless you don't have a fully working example with e2e tests I suggest you don't work on this.

I had a fully working composite id branch. Marked as don't merge by you.

> Also, it seems like you are not interested on the this future, and I am more than willing to put on the work, so just assign it it to me or discuss it with me please if you don't like something, I am sure we can find a solution that suits everybody's needs.

I'm not interested in discussing further.
And composite id may be implemented different in others backends, so 'suits everybody's needs' really need everybody to participate.

> I am for some version of this

So I am going to prepare it.

> just please remember that in my blueprints I prefer to have the derivedField have a name like relationshipName+otherEntityField and if you add the field too late I want be able to that, I am willing to help on this.

If the db column name is kept the same, it can be an option.
The db column name will be a bad breaking change for old `@MapsId` applications, and should be avoided, unless others developers thinks this is the way forward.

So:
- I will split the primaryKey so it can be more easily customized by blueprints.
- Following PRs should be as specific as possible, like adding support for autoGenerated, updatable...

generators/database-changelog-liquibase/templates/src/main/resources/config/liquibase/changelog/added_entity.xml.ejs
yelhouti: I don't know if answered your question, since I don't know if I understand it correctly, hopefully yes.

yelhouti:
> The naming does not seem right.

Anything other than references is good for me, (just to avoid confusion between the two)

> Are you aware how request changes works?

I don't know if you are being sarcastic or what, or maybe I am not understanding you well, so I will refriegn from answering that.

> As I said, if the server returns it ready, it's not needed.

Well I don't want my blueprint to return it, so please don't add something to fields like you did before to force me to.

> Anyway, with my proposed changes, it will be possible to implement this in the blueprint.

Exactly, so please don't add it in the main generator.

> I had a fully working composite id branch. Marked as don't merge by you.

I didn't work with my advanced example as I explained, and I didn't like how it forced to return the id to the frontend (since it's implementation detail IMO).

> I'm not interested in discussing further.
And composite id may be implemented different in others backends, so 'suits everybody's needs' really need everybody to participate.

Well sounds like an oxymoron to me, you can't want both: 
1- everybody to participate so that it suits everybody's needs
2-  don't discuss anymore

> I will split the primaryKey so it can be more easily customized by blueprints.

As soon as this is merged you go ahead please.

> Following PRs should be as specific as possible, like adding support for autoGenerated, updatable...

Good enough for me.

I will ask you to again, to please, just tag me `@yelhouti` if you make any PR related to primaryKey to make sure that my blueprints keep working. Thanks in advance.

deepu105:
Hold your horses guys 😂

1. Lets not get defensive here, we want a compromise solution that works
for all

2. Lets not get caught up in naming and stuff. I don't care about the name
as long as its understandable to core team and blueprint authors. If you
want, add more fields for specific purpose even if those are
duplicate/derived data of another fields. Its internal so its ok and is not
gonna add any noticeable overhead. Try not to break any existing fields

3. Lets get this PR merged before making any parellell changes. @mshima we
can do your proposal after this is merged or you can may be add that as a
PR to this PR branch so that its easy for everyone to follow

yelhouti:
@deepu105 I don' t know, I feel like he doesn't like me much :smile: 

Please tell me if there is anything I can do. I made the previous changes you asked for

mshima:
@deepu105 sure, your call, has you can see I've already said my opinion, and "I am free to change the PR" 😄.
I am against merging as it is, and the proposed changes would not be necessary if we delegate primaryKey to the blueprint.
Please review it and merge it at will.

yelhouti:
@mshima delegating to the blueprint force all blueprints to implement the same code + risk having inconsistencies between them (one breaking the other as explained in the issue) already face that. This si why we are try to generalize here :)

mshima:
> I feel like he doesn't like me much 😄

Really? I really don't know you enough, but you really don't pay attention to reviews and always do it your way.

> delegating to the blueprint force all blueprints to implement the same code + risk having inconsistencies between them (one breaking the other as explained in the issue) already face that.

I must quote you from another topic `I know what I am talking about`.
By the way, you should not try to force jhipster definitions as proxy for your blueprints.

@yelhouti please don't quote me anymore, you need to convince @deepu105, he is reviewing this PR.

yelhouti:
If you look at my PRs and your reviews on them, you can see that I always tried to change to match what you asked for, must of the time I find them very interesting and improve the quality of the code. But sometimes we can disagree, and other I don't totally understand what you want until you show me.

I do not understand your quote or what you are implying, maybe it's your english as you said before.

When you say: `don't quote me anymore` you mean tag `@mshima` or quote what you say like 
> this

and is this only for this PR, or all future PRs :smile: ? 
I hope we someday meet in person I am sure I are a really nice person :) !

deepu105:
@yelhouti @mshima you both really need to stop taking these personally. Remember we are doing OSS on our free time out of passion. Please stop digging into trivial things, its ok to have different opinions on stuff and to be opinionated. 

I request you both to take a step back, try to keep you feelings out of it and actually see just the code from the perspective of the project, is it useful for the community, does it add value?

conventions and stuff can be improved and someone else will always find better ways to do things in the future. Please stop marrying your code, especially in an OSS project. When you commit something to OSS you need to accept the fact that someone else will change it at some point and you might not like the change, but if it adds value to the project in some way you should accept that for the betterment of the project and thats how you build communities. The project can only keep growing if we accept that and make compromises when necessary, that is why I'm not bothered by the core changing even though I dont like all of the changes happening for example. 

So I again request you both to try and collaborate rather than trying to prove who is right and who is wrong by picking on trivial details. Come on guys chill, its just some code :joy: its not worth fighting over

yelhouti:
@deepu105 I a really don't mind him or anyone else changing stuff I even welcomed it, I just don't want a little bit of backward compatibilty. Anyway how should we proceed this on this PR @deepu105 ?

utils/entity.js
```diff
@@ -191,57 +186,46 @@ function prepareEntityForTemplates(entityWithConfig, generator) {
...
         // MapsId copy the id from the relationship.
         autoGenerate: true,
-        get fields() {
-          return this.derivedFields;
+        get ids() {
+          return relationshipToIds(relationshipId, generator);
         },
-        get derivedFields() {
-          return relationshipId.derivedPrimaryKey.derivedFields;
+        get originalFields() {
```
Tcharl, line 226: Really sorry, but originalFields is not spelling at all: what does it mean?
Isn't something like 'composingFields' would be more understandable

yelhouti: @Tcharl thanks for the feedback, I myself couldn't find a better name but maybe you could help. I will go with what the majority decides.

to answer your question a primaryKey boils down to the list of columns in the database that constitute the primary key of its entity. The column they refer to might be few entities away, originalFields represents the list of field in the original entities where they where first defined.

Ex: 
let's define the following JDL:
```

entity Employee{
	@id username String required
	fullname String required
}

entity EmployeeSkill{
	// compositeId employeeSkill.name, employee.id
	@id name String required
	level Integer required
}

relationship OneToMany{
	@id Employee{skill(name)} to EmployeeSkill{employee(fullname) required}
}

entity CertificateType{
	name String required
}

entity EmployeeSkillCertificate{
    // EmployeeSkill.employee.username, EmployeeSkill.name, CertficiateType.id
    grade Integer required
    date LocalDate required
}

relationship OneToMany{
	@id CertificateType to EmployeeSkillCertificate{type(name) required}
	@id EmployeeSkill to EmployeeSkillCertificate{skill required}
}
```

EmployeeSkillCertificate has primaryKey that contains 3 `ids`:
EmployeeSkill.employee.username, EmployeeSkill.name, CertficiateType.id

originalFields are:
- username (in the entity employee)
- name (in the entity employeeSkill)
- id (in the entity certificate type)

I do not personally like  `composingFields` since they are not really in the primaryKey and composing it, but if you agree on it I will use it.

deepu105:
I don't want this to block v7 so lets proceed with this for now. We can change it after release as I don't see any big breaking changes that could happen here except for internal logic may be and field name which we can avoid. Can you fix conflicts

utils/entity.js
Tcharl: more clear now, I thought that it was a deep copy instead of an alias.
Maybe being even more specific, like `referencedIdsComposingId`, WDYT?

yelhouti: I would prefer not to use reference in the name, since it's refers to something else in the code as you might know (kind of in intersection between field and relationship transformed) a good name would be <SOMETHING>Fields since they are fields and when you use then you know they have fieldName/type... but again we could change these later

yelhouti:
@deepu105 I just rebased, this can be merged.

Also while trying to use the blueprint I found out that our code was much cleaner when we used consistent naming while working with derived relationships (the current version code names differently for example if you have a derived oneToOne between user and employee, the id of employee would be `id` instead of `userId`. and in many places in the code we need to check for it.

What I propose is having a variable `generator.jhipsterConfig.backwardCompatibleDerivedName` that is true by default (for v7) and that blueprints (or maybe even in the jdl/.yo-rc) can switch to false to avoid dealing with these relationships differently. What do you think ?

deepu105 (approved):

yelhouti:
@deepu105 thanks a lot. what do you think about adding the backwardCompatibleDerivedName config varaiable ?

deepu105:
Well v7 is a major version with a lot of breaking changes so I dont see the
point of doing that just for a blueprint

Thanks & Regards,
Deepu

yelhouti:
Is there a way to add application options for blueprints ? (add in jdl...)

mshima:
@deepu105 just a few regression.

The merged code just ignores consistency.
Column name:
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/entity.js#L350
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/entity.js#L364-L367

Just ignores the other side and any validation.
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/field.js#L207-L220

`@Id any Long` with postgresql :collision:
`@fieldNameAsDatabaseColumn(my_custom_table_name) id Long` :collision:
Not even considering relationships.

```
entity Foo {
  @Id @entityTranslationKey("Foo Id") id Long
}

entity Bar {}

relationship OneToOne {
  Bar to @Id Foo
}
```
:collision: sorry, no correct translation for you, please duplicate your translation key.

The primaryKey was READY and working, without those problems.

MathieuAA:
I'm working on it, when I have some free time and when I'm not too tired to do so.

yelhouti:
thanks @MathieuAA

deepu105:
@yelhouti can you fix these problems identified by @mshima with a new PR

Column name is not consistent:
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/entity.js#L350
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/entity.js#L364-L367

Ignores the other side and any validation.
https://github.com/jhipster/generator-jhipster/blob/a25d5af967adc371d3c29730feccada471c6c9aa/utils/field.js#L207-L220

```
@Id any Long with postgresql 💥
@fieldNameAsDatabaseColumn(my_custom_table_name) id Long 💥
```
Not considering these relationships.

```
entity Foo {
  @Id @entityTranslationKey("Foo Id") id Long
}

entity Bar {}

relationship OneToOne {
  Bar to @Id Foo
}
```
Fails with sorry, no correct translation for you, please duplicate your translation key.

yelhouti:
@deepu105 sure on it

yelhouti:
@mshima I think I understand what you meant by inconsistent, I made a PR #14331 does it help.

about the other issues, I tried to make sure that the id field is added before prepareFieldForTemplate, so that they are handled the same as the others, but it looks like the templates handle them differently. Did I break something or was it broken before my changes? (the new introduced variables are use nowhere so I don't understand how I could have broken anything)

Also, It seems like these issues can be fixed my changing the templates, but I was instructed by @deepu105 not to fix anything else in the templates that didn't involve my changes, should I start making small PRs fixing stuff like mentioned above?  
ex: handle the derived Id field like all the other fields...

there are many changes/optimizations like this that I can port from the compositeKey branch.

deepu105:
I'm gonna look into this and fix this

yelhouti:
@deepu105 If you need any help with anything, please don't hesitate.

mshima:
@yelhouti the info is not there by the time the primaryKey is prepared, you could add getters for every calculated variable that is created afterwards.
Or if you let me, I will revert changes to the primaryKey, keeping your newly added variables just the way you created the tests.

yelhouti:
@mshima I don't know which info you are talking about, but I am sure you are right. please go ahead and make a PR I will be happy to check it out.

deepu105:
I already made a PR so please just add commits to it :)

Thanks & Regards,
Deepu

yelhouti:
@mshima maybe your proposed solution of having a step that adds the necessary "idField" if missing and computs the primaryKey would solve all these problems. Don't you think ?

PR #14307, opened 2021-03-13
OIDC /userinfo lookup improvements

mraible:
This changes things so microservices don't have a `CustomClaimConverter` for looking up a user's attributes. It should only be at the gateway or monolith level. Since it wasn't implemented for reactive, this means it only applies to non-reactive monoliths now. FYI @Falydoor. 

---

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -556,12 +556,7 @@ const serverFiles = {
       ],
     },
     {
-      condition: generator =>
-        !generator.reactive &&
-        generator.authenticationType === 'oauth2' &&
-        (generator.applicationType === 'monolith' ||
-          generator.applicationType === 'microservice' ||
-          generator.applicationType === 'gateway'),
+      condition: generator => !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'monolith',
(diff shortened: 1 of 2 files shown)


generators/server/files.js
```diff
@@ -556,12 +556,7 @@ const serverFiles = {
       ],
     },
     {
-      condition: generator =>
-        !generator.reactive &&
-        generator.authenticationType === 'oauth2' &&
-        (generator.applicationType === 'monolith' ||
-          generator.applicationType === 'microservice' ||
-          generator.applicationType === 'gateway'),
+      condition: generator => !generator.reactive && generator.authenticationType === 'oauth2' && generator.applicationType === 'monolith',
```
Tcharl, line 559: What about gateway?

mraible: This feature was never implemented for WebFlux. Since our gateway is now WebFlux-only, it doesn't make sense to include this. Adding it to WebFlux is being tracked by https://github.com/jhipster/generator-jhipster/issues/13461.

Tcharl: This explains that. Thank you!

PR #14312, opened 2021-03-14
Refactor github workflows.

mshima:
Since our concurrent builds dropped to 20, with must refactor our CI build.

- Move most of tests to backend workflow.
  - backend matrix workflow is broader and harder to split, so client test variants can be divided here.
- Keep only client specific test at angular/react/vue workflow.
  - Execute only with client specific changes that backend doesn't covers.
  - Test authentication (jwt/oauth2/session), theme, websocket.
  - Test sqlfull (most complete entities) and additional (specific features like custom id)

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -25,16 +25,30 @@ on:
     branches-ignore:
       - 'dependabot/**'
       - 'skip_ci*'
-    paths-ignore:
-      - 'generators/*client/templates/react/**'
-      - 'generators/*client/templates/vue/**'
+    paths:
+      # ci
+      - '.github/workflows/*'
+      - 'test-integration/**'
+
+      # paths not covered by backend tests
+      - 'package*.json'
+      - 'generators/*client/*'
+      - 'generators/*client/needle-api/*'
+      - 'generators/*client/templates/angular/**'
+
   pull_request:
     branches:
       - '*'
-    paths-ignore:
+    paths:
+      # ci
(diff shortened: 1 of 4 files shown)


pascalgrimaud:
Give me some time to review it plz :)

pascalgrimaud:
Just to let you know I don't forget this, I'll review it asap

pascalgrimaud:
> Since our concurrent builds dropped to 20, with must refactor our CI build

I know the CI builds take a lot of time, but I don't think we should reduce the number of builds in our CI

Sorry @mshima but I don't like moving all main builds into backend. I'll explain why :-)

I still think we should split:
- front: Angular / React / Vue
- back: Maven / Gradle

So, it's clear to have, and it will be easy to restart:
- Angular: Maven + Gradle
- React: Maven + Gradle
- Vue: Maven + Gradle

Webflux is different because it was still in beta.

But, we can improve the CI here.
For example, about infinispan -> do we need to test it for Angular + React + Vue ? No, only 1 is needed
About MariaDB -> same question. No, only 1 is needed, as it's similar to MySQL

Each options should be rework separately.

I agree with the idea behind this PR but we need to be smart :-)

mshima:
@pascalgrimaud the idea behind this PR is to be smart at frontend builds.
Frontend PRs are probably between 70-80% of the PRs due to our dependabot PRs.
I've grouped backend builds, but it's unrelated to the main idea of this PR.
maven/gradle should be divided equally at frontend tests.

PR #14348, opened 2021-03-16
Fix primary key

deepu105:
Trying to keep both @mshima and @yelhouti happy :wink:


deepu105:
@mshima @yelhouti I started putting both your changes in where I thought it matters, some of it definitely could be wrong. But I want all there of us to work on this until both of you think this works for your use cases in mind. Feel free to suggest changes or do PRs to my branch. My goal is to make sure both of you think this works for your usecase so no discussions on variable names or code style, just functionality please

deepu105:
also I have to accept that I dont understand the purpose of half the code, so it would nice for contributors if we add more code comments

mshima:
@deepu105 thanks, done at https://github.com/jhipster/generator-jhipster/pull/14351.

Issue #13796, opened 2021-02-01
npm install is slow and happens too often


mraible:
##### **Overview of the feature request**

I might not understand things, but it seems to be like `npm install` happens twice.

1. Once when you run `jhipster` (unless you use `--skip-install`)
2. Once when you start everything with Maven

##### **Motivation for or Use Case**

`npm install` is sooooo slow and it's really painful for it to happen twice in demos.

##### **Steps to reproduce**

With JHipster 7 beta 1:

```
jhipster jdl default
# Execution time: 1 min. 32 s
./mvnw
```

During the Maven process, there's a printout that it took another 60 seconds:

```
[INFO] --- frontend-maven-plugin:1.11.0:npm (npm install) @ jhipster ---
[INFO] Running 'npm install' in /Users/mraible/npm-install
[INFO] npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
[INFO]
[INFO] > husky@4.3.7 preuninstall /Users/mraible/npm-install/node_modules/husky
[INFO] > node husky uninstall
[INFO]
[INFO] husky > Uninstalling git hooks
[INFO] husky > Done
[INFO]
[INFO] > fsevents@1.2.13 install /Users/mraible/npm-install/node_modules/watchpack-chokidar2/node_modules/fsevents
[INFO] > node install.js
[INFO]
[INFO]   SOLINK_MODULE(target) Release/.node
[INFO]   CXX(target) Release/obj.target/fse/fsevents.o
[INFO]   SOLINK_MODULE(target) Release/fse.node
[INFO]
[INFO] > fsevents@1.2.13 install /Users/mraible/npm-install/node_modules/webpack-dev-server/node_modules/fsevents
[INFO] > node install.js
[INFO]
[INFO]   SOLINK_MODULE(target) Release/.node
[INFO]   CXX(target) Release/obj.target/fse/fsevents.o
[INFO]   SOLINK_MODULE(target) Release/fse.node
[INFO]
[INFO] > husky@4.3.7 install /Users/mraible/npm-install/node_modules/husky
[INFO] > node husky install
[INFO]
[INFO] husky > Setting up git hooks
[INFO] husky > Done
[INFO]
[INFO] > @fortawesome/fontawesome-common-types@0.2.34 postinstall /Users/mraible/npm-install/node_modules/@fortawesome/fontawesome-common-types
[...]

PR #14395, opened 2021-03-20
Run frontend tests with dev profile

vishal423:
Current integration supports execution of frontend tests with `prod` profile. This PR adds same support in the `dev` profile that's configured as default.
---

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1534,6 +1534,17 @@
                                 </goals>
                             </execution>
                             <execution>
+                                <id>webapp build test</id>
+                                <goals>
+                                    <goal>npm</goal>
+                                </goals>
+                                <phase>test</phase>
+                                <configuration>
+                                    <arguments>run webapp:test</arguments>
+                                    <npmInheritsProxyConfigFromMaven>false</npmInheritsProxyConfigFromMaven>
+                                </configuration>
+                            </execution>
+                            <execution>
                                 <id>webapp build dev</id>
                                 <goals>
                                     <goal>npm</goal>


pascalgrimaud:
Not sure about that
Does it mean : `mvn test` will launch front end test too?

vishal423:
>  Does it mean : `mvn test` will launch front end test too?

Yes. It doesn't look appropriate to install and build the frontend, but not run the test cases. With `prod` profile, we do run frontend tests, so, it should be done in the `dev` profile as well. In case we want to exclude frontend build/tests, then exclude in all phases by excluding `webapp` profile

pascalgrimaud:
The npm test are launch during `mvn verify -Pprod` because you package the app before deploying
So it's normal to launch backend unit tests, frontend unit tests, backend integration tests for build prod

But during the dev flow, I think it's a bad idea if `mvn test` launch front end test
As `mvn test` can launch unit test and `npm test` can launch front test, it's nice to keep these tests separately

IMO, adding your proposal will really slow down the workflow during development

vishal423:
If you launch `mvnw clean test`, then, in the current flow we already install node/npm, install packages and build the frontend code. Adding a step to run tests to tie along the maven test build phase will bring in consistency, IMO. 

If a developer doesn't care about the frontend code, then, the best way is to exclude the `webapp` profile to exclude all steps associated with frontend install/build/test.

pascalgrimaud:
> If you launch mvnw clean test, then, in the current flow we already install node/npm, install packages and build the frontend code

Yes and it's really bad. I was against this. There is an open ticket for that, and some suggestion to launch `mvn test -P-webapp`

If we decide to include `npm test` during `mvn test`, to be fair, we should include the backend test too, when we launch `npm test` too, which will be really strange

I'd like to discuss about it, because `mvn test` can be very fast and including additional tests will definitively be a bad user experience IMO

Tcharl:
@pascalgrimaud , are the Java unit test are run during `mvn test`: Yes! So to be consistent, we should trigger Jest test during the same phase.
Then, java integration tests are triggered during the `integration-test` lifecycle phase: it could be nice to trigger cypress/protractor tests here, but it's much more work (particularly for microservice) I made it in one of my project and can implement it in the jh pom, but it will result in a much bigger pom configuration

pascalgrimaud:
Here my point of view:
- I don't care about front test, when I want to launch "only" backend test :)
- If I want to launch Java unit test: mvn test
- If I want to launch Jest unit test: npm test
- Why should we mixte both ? I really don't understand here
- When I want to package, doing full tests (front + back unit tests + integration tests): mvn verify -Pprod

I don't like it because we start to include **too much front part**, in our different maven phases

Anyway, this PR won't change the current behaviour which is already really bad IMO. See this ticket by @mraible : https://github.com/jhipster/generator-jhipster/issues/13796

The current behavior with empty project, when you launch `mvn test`:
- npm install -> which can be slow...
- ng build -> 27sec
- mvn test -> 14sec
- Total: 45sec

With this PR:
- npm install -> which can be slow...
- ng build -> 27sec
- mvn test -> 14sec
- npm run lint -> 11sec
- npm run test -> 6sec
- Total: 1min

But as said, feel free to merge it if you think it's good for user experience to have slow phase :-D

No worry about me, I just need to remove the webapp profile in pom.xml and my projects are far better.

mraible:
I'm against anything that slows down npm or Maven/Gradle commands. It's already bad enough.

Tcharl:
@pascalgrimaud @mraible you convinced me :-), let's do something to improve the build time

vishal423:
I have something in slate that hopefully address slowness concerns. Closing this PR for now.

PR #14443, opened 2021-03-24
Fail to get jwt token

hwqllang:
Fix https://github.com/jhipster/generator-jhipster/issues/14488
- authenticationToken need parsed
- key name is wrong in localStorage

---

### DIFF
--- a/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
+++ b/generators/client/templates/angular/src/main/webapp/app/app.module.ts.ejs
@@ -73,7 +73,7 @@ import { ErrorComponent } from './layouts/error/error.component';
     // Set this to true to enable service worker (PWA)
     ServiceWorkerModule.register('ngsw-worker.js', { enabled: false }),
     HttpClientModule,
-    NgxWebstorageModule.forRoot({ prefix: '<%= jhiPrefixDashed %>', separator: '-' }),
+    NgxWebstorageModule.forRoot({ prefix: '<%= jhiPrefixDashed %>', separator: '-', caseSensitive: true }),
 <%_ if (enableTranslation) { _%>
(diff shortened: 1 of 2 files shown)


mshima:
@hwqllang thanks for your PR.
Can you sign the cla?

We need to check if `JSON.parse` is required for all 3 client frameworks.

gmarziou:
@hwqllang can you explain why token parsing is required in swagger-ui?

Personally, I don't understand it because next line is using it in header as bearer token so I would think that base64 encoding is expected.

pascalgrimaud:
I think there is indeed a bug in your current 7.0.0, but it needs to be confirmed:
- Angular / React: swagger broken
- Vue: swagger is fine

This PR seems to fix the issue for Angular and React, but it won't work for Vue.
So IMO, the issue comes from the way how the token is stored

mshima (changes requested):
@pascalgrimaud the problem seems to be that ngx-webstorage stores it encoded.
At this page we use bare localStorage.
Not sure we can fix it another way.

mshima:
@hwqllang can you fix the user of the first commit?

hwqllang:
@mshima sorry for that, the user of the first commit is wrong, it is default user of my server, and I don't know how to fix it.

mshima:
@hwqllang thanks and congrats for your first contribution.

mshima:
Tested with angular/react and remember me enabled/disabled.

pascalgrimaud:
@hwqllang : it's an important bug so I added a bounty on this. Don't forget to claim it

hwqllang:
Bounty claimed - https://opencollective.com/generator-jhipster/expenses/37328

pascalgrimaud:
@hwqllang : approved and thanks for your fix

PR #14493, opened 2021-03-29
fix: small fixes needed for composite key

yelhouti:
---

- [X] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary

### DIFF
--- a/generators/entity-server/index.js
+++ b/generators/entity-server/index.js
@@ -115,10 +115,9 @@ module.exports = class extends BaseBlueprintGenerator {
       },
 
       useMapsIdRelation() {
-        const jpaDerivedRelation = this.relationships.find(rel => rel.id === true);
-        if (jpaDerivedRelation) {
+        if (this.primaryKey && this.primaryKey.derived) {
           this.isUsingMapsId = true;
-          this.mapsIdAssoc = jpaDerivedRelation;
+          this.mapsIdAssoc = this.relationships.find(rel => rel.id);
           this.hasOauthUser = this.mapsIdAssoc.otherEntityName === 'user' && this.authenticationType === 'oauth2';
         } else {
           this.isUsingMapsId = false;
--- a/utils/relationship.js
+++ b/utils/relationship.js
@@ -112,16 +112,12 @@ function prepareRelationshipForTemplates(entityWithConfig, relationship, generat
   }
 
   relationship.relatedField = otherEntityData.fields.find(field => field.fieldName === relationship.otherEntityField);
-  if (!relationship.relatedField && !relationship.otherEntity.embedded) {
-    if (otherEntityData.primaryKey && otherEntityData.primaryKey.derived) {
-      Object.defineProperty(relationship, 'relatedField', {
-        get() {
-          return otherEntityData.primaryKey.derivedFields.find(field => field.fieldName === relationship.otherEntityField);
-        },
-      });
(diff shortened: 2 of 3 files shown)


utils/relationship.js
```diff
@@ -147,6 +147,10 @@ function prepareRelationshipForTemplates(entityWithConfig, relationship, generat
     relationshipNameCapitalized: _.upperFirst(relationshipName),
     relationshipNameHumanized: _.startCase(relationshipName),
     columnName: generator.getColumnName(relationshipName),
+    columnNamePrefix:
+      generator.jhipsterConfig.backwardCompatibleDerivedName && relationship.id && relationship.relationshipType === 'one-to-one'
```
yelhouti, line 151: This param can be used by blueprints to specify how they would like to name relationships for one-to-one derived ids

yelhouti:
@deepu105 or @mshima if you don't mind adding these small changes, so that composite key blueprint are easier to implement. Thanks a lot

generators/entity-server/index.js
```diff
@@ -115,7 +115,7 @@ module.exports = class extends BaseBlueprintGenerator {
       },
 
       useMapsIdRelation() {
-        const jpaDerivedRelation = this.relationships.find(rel => rel.id === true);
+        const jpaDerivedRelation = this.relationships.find(rel => rel.id && rel.relationshipType === 'one-to-one');
         if (jpaDerivedRelation) {
           this.isUsingMapsId = true;
           this.mapsIdAssoc = jpaDerivedRelation;
```
mshima, line 121:
```suggestion
        if (this.primaryKey && this.primaryKey.derived) {
          this.isUsingMapsId = true;
          this.mapsIdAssoc = this.relationships.find(rel => rel.id);
```

utils/relationship.js
```diff
@@ -112,7 +112,11 @@ function prepareRelationshipForTemplates(entityWithConfig, relationship, generat
   }
 
   relationship.relatedField = otherEntityData.fields.find(field => field.fieldName === relationship.otherEntityField);
-  if (!relationship.relatedField && !relationship.otherEntity.embedded) {
+  if (
+    !relationship.relatedField &&
+    !relationship.otherEntity.embedded &&
+    !(relationship.otherEntity.primaryKey && relationship.otherEntity.primaryKey.composite)
+  ) {
```
mshima, line 119: I suppose this change is related to having multiples otherEntityFields.
We don't support that in our templates. This change is specific to your blueprint.
You can extract lines 114-134 into another task in the generator and replace with your own implementation in your blueprint.

mshima (changes requested):

utils/relationship.js
yelhouti: Not really, the reason I am introducing this change, is that the blueprint throws without this line as there is no related Field. I would just like to avoid that. I can do this locally without redefining the whole function and calling  whenever it was called before.

generators/entity-server/index.js
yelhouti: done

utils/relationship.js
mshima: Unify with the condition below and drop the else throw.
So it will not be required anymore.

yelhouti: @mshima done

mshima (approved):

Issue #14235, opened 2021-03-08
Ease generator understanding


Tcharl:
##### **Overview of the feature request**

Not really a request but more a plan for what I'm trying to currently achieve.

The plan is divided into multiple steps and I think we gonna win some things at each of them.

1. Use constants in all js files.
As you may have noticed, I'm currently introducing options constants anywhere I can in the generators js files, replacing `buildTool == 'maven'` by `buildTool == MAVEN` and this for all constant.
As a developer with an IDE, I love the folding capability to see where a particular constant is used.
Using constants instead of plain string will help to identify all the generators, needles and helpers that uses a jhipster option.
Also, I think it will help the test maintenance and the removal of some jdl historic syntax mismatch (for example the `false`vs the `no`).

2. Align the generator's context feeding

Context is populated by options differently on each generator: 
for example in the [_loading()](https://github.com/jhipster/generator-jhipster/blob/e22ca7cfd00f2a905ae7996d7adfcf82096bb4e5/generators/client/index.js#L163) method in the client generator whereas its in the [getConfig()](https://github.com/jhipster/generator-jhipster/blob/e22ca7cfd00f2a905ae7996d7adfcf82096bb4e5/generators/ci-cd/index.js#L90) one in the ci-cd generator.

**Question**: Can someone tell me which is the generator that uses the best practices so that I'll replicate it in all of them? I'm not a Yeoman black belt, and I'm afraid to do it wrong.

So that we'll have something that is replicable and begin to mutualize some methods names and calls. Also it will be easier for everyone to have a standard canvas/template for method override and extension (I'm thinking of blueprints).

3. Use 'derived properties' in templates instead of using logic (equality tests, string tests)

The final step would be to use only boolean logic in templates, thus injecting derived attributes in generators context:
for example, injecting a `ìsBuildToolMaven` instead of using `buildTool == 'maven'`in all the ejs.
This last step will bring additional value on top of the two previous one (and can only be achieved well with the two previous one implemented):

- Easy folding: a search in the IDE will lead to all check achieved on an option
- Templates will be a little bit cleaner and just be dumb as possible (only containing OR and AND ejs checks)
- Also, any newcomer will be able to see where using the 'maven' 'buildTool' option is impacting the generation, so it will be easier for him to either add a similar option (sbt, bnd, ant :-p) or to override some option's behaviour.
[...]

PR #14561, opened 2021-04-02
Enable on disk build cache available at webpack 5.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/13796

---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #14563, opened 2021-04-02
use constants in generator docker

Tcharl:
contributes to [generator understanding](https://github.com/jhipster/generator-jhipster/issues/14235)

Use constants in generator docker
---


generators/docker-compose/index.js
```diff
@@ -223,11 +251,37 @@ module.exports = class extends BaseDockerGenerator {
...
   }
 
-  get writing() {
+  _loading() {
+    return {
+      loadSharedConfig() {
+        this.loadAppConfig();
+        this.loadClientConfig();
+        this.loadServerConfig();
+        this.loadTranslationConfig();
```
mshima, line 260: A custom config loader should be implemented.
Configuration is quite different at this generator and isn't supposed to be the same config file.

Tcharl: Just wanted here to load to base properties, in addition to the docker-compose generator ones.
Does it cause any issue?

mshima: Not sure, but docker-compose is not guarantee to be executed in the application folder, so .yo-rc.json doesn't exists in that folder.
IMO is misleading.

Tcharl: You're right, forgot about that... Going to rework it

generators/entity-client/index.js
```diff
@@ -59,6 +64,31 @@ module.exports = class extends BaseBlueprintGenerator {
...
+    this.searchEngineFalse = this.entity.searchEngine === false;
+  }
+
+  _loading() {
+    return {
+      loadSharedConfig() {
+        this.loadAppConfig();
+        this.loadClientConfig();
+        this.loadServerConfig();
+        this.loadTranslationConfig();
```
mshima, line 81: There is one detail about entity-client:
- at gateways, entities from microservices, should load the configuration of the microservice not the gateway one.

So IMO should be removed.

mshima (commented):
Would be nice if you limit PRs to ~20 files.
It will be a lot easier.

generators/entity-client/index.js
```diff
@@ -59,6 +64,31 @@ module.exports = class extends BaseBlueprintGenerator {
     return this._default();
   }
 
+  _loadEntityConfig() {
+    this.paginationPagination = this.entity.pagination === PAGINATION;
+    this.paginationInfiniteScroll = this.entity.pagination === INFINITE_SCROLL;
+    this.paginationNo = this.entity.pagination === NO_PAGINATION;
+
+    this.searchEngineFalse = this.entity.searchEngine === false;
```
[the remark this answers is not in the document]
Tcharl: If I remember well, we can also disable search at the entity level (not only globally) so I guess we should put everything in the entity generator.

Not a fan of `searchEngineFalse` though...

generators/entity-client/index.js
Tcharl: Removed because it was already in generator-entity.
I think that there's something to do on that part: we can't execute entity-client in a standalone mode (without executing entity). It looks against good code principles but can also be annoying for blueprints.
I can raise a ticket with my suggestion following what I encounter during the constant refactoring

Tcharl:
@mshima better now? FYI, I removed the elasticSearchFalse in favour of testing elasticSearch as it was set to `false` if disabled

generators/entity/index.js
```diff
@@ -272,6 +288,14 @@ class EntityGenerator extends BaseBlueprintGenerator {
...
         }
       },
+
+      reloadDerivedConfig() {
+        // May have been modified after options
+        this.loadDerivedAppConfig(_.defaults({}, this.jhipsterConfig), this.context);
+        this.loadDerivedClientConfig(_.defaults({}, this.microserviceConfig, this.jhipsterConfig), this.context);
+        // Try to load server config from microservice side, falling back to the app config.
+        this.loadDerivedServerConfig(_.defaults({}, this.microserviceConfig, this.jhipsterConfig), this.context);
+      },
```
mshima, line 298: @Tcharl the entity generator crashes without this?
- The api usage looks wrong `loadDerivedAppConfig(dest, config)`, dest/config is inverted.
- server config is from `this.microserviceConfig` while client config should be loaded from the gateway `this.jhipsterConfig` (while microfrontend is not implemented).

We don't have Microservices tests in our main repository, this is easy to break.
This will get easier when microfrontend lands.
```suggestion
```

generators/generator-base.js
```diff
@@ -2554,6 +2564,11 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
     dest.serviceDiscoveryType = config.serviceDiscoveryType;
 
     dest.embeddableLaunchScript = config.embeddableLaunchScript;
+
+    this.loadDerivedServerConfig(dest, config);
+  }
+
+  loadDerivedServerConfig(dest, config) {
```
mshima, line 2571: Probably it's better to change the parameter order to match loadServerConfig api.

generators/entity/index.js
```diff
@@ -465,7 +489,7 @@ class EntityGenerator extends BaseBlueprintGenerator {
         // Update current context with config from file.
         Object.assign(this.context, this.entityStorage.getAll());
         loadRequiredConfigIntoEntity(this.context, this.jhipsterConfig);
-
+        this._loadEntityDerivedProperties(this.context);
```
mshima, line 492:
```suggestion
        this.loadServiceDerivedMicroserviceProperties(this.context, this.context);
        this._loadEntityDerivedProperties(this.context);
```

generators/generator-base.js
```diff
@@ -2483,8 +2485,12 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
...
     dest.cypressTests = dest.testFrameworks.includes(CYPRESS);
 
-    dest.jhiPrefixCapitalized = _.upperFirst(this.jhiPrefix);
-    dest.jhiPrefixDashed = _.kebabCase(this.jhiPrefix);
+    this.loadDerivedAppConfig(dest, config);
+  }
+
+  loadDerivedAppConfig(dest, config) {
+    dest.jhiPrefixCapitalized = _.upperFirst(config.jhiPrefix);
+    dest.jhiPrefixDashed = _.kebabCase(config.jhiPrefix);
```
mshima, line 2493:
```suggestion
  loadDerivedMicroserviceAppConfig(dest, config) {
    dest.jhiPrefixCapitalized = _.upperFirst(config.jhiPrefix);
    dest.jhiPrefixDashed = _.kebabCase(config.jhiPrefix);
  }

  loadDerivedAppConfig(dest, config) {
    this.loadDerivedMicroserviceAppConfig(dest, config);
```

mshima (commented):
The entity* generators needs more work at derived properties due to Microservice support.
Some config is loaded from the gateway, others from Microservice.
Others seems ok.

IMO you should remove derived properties from entities at this PR, just use constants.
Microservice support is really complicated.

generators/docker-compose/index.js
```diff
@@ -22,9 +22,28 @@ const jsyaml = require('js-yaml');
...
 module.exports = class extends BaseDockerGenerator {
-  get initializing() {
+  constructor(args, options) {
+    super(args, options);
+    useBlueprints = !this.fromBlueprint && this.instantiateBlueprints(GENERATOR_DOCKER_COMPOSE);
+  }
+
+  _initializing() {
     return {
       ...super.initializing,
```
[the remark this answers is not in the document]
Tcharl: Didn't know I could call private variable from a child (but I knew the `_` was just a convention) ;-).

generators/entity/index.js
Tcharl: I use the same syntax as l#227. Anyway, after investigation, there's no option that can impact derived property: I find it a little bit risky, but this code can disappear

generators/generator-base.js
Tcharl: Ooops

Tcharl:
This entity-client generator's code is crazy: so much nifty details everywhere...

generators/entity/index.js
Tcharl: This one is for entity :-)

Tcharl:
@mshima or @pascalgrimaud  can you take a look and maybe quick (as any modification on the angular entity part will lead to a conflict)? I know it's tough and I'm sorry for that, but I think I found the elegant way to achieve it and now templates are way more clean.
I would like to continue on that refactoring to be globally applied and I guess we'll then have a way more maintainable codebase.

I guess that the overall refacto will need two or three more full refinement but as it will only be at the js level after the first pass we'll be able to make something way more scalable and modular (I can't wait to split up that 2.6 thousand-line BaseGenerator class :-p)

mshima (commented):
@Tcharl last commit is not ok, or it fixes bugs, not sure. The generated code is not identical anymore.

While testing, I've realized that micro service support was already buggy for entity due to a racing condition. We will fix this later.

Revert last commit and merge.
Create a new PR with the last commit.

mshima:
@Tcharl I've just merged https://github.com/jhipster/generator-jhipster/pull/14561 and realized that I've replaced builtToolUndefined with buildTookUnknown, that you may be using here.

Tcharl:
Sorry @mshima but I don't understand your remark: should I only accept your change and that's it?
What is different compared to previous generation? It shouldn't have modified anything (maybe some carriage return but should be the only difference).

mshima:
> Sorry @mshima but I don't understand your remark: should I only accept your change and that's it?

Yes, it stopped recalculating jhiPrefixDashed.

> What is different compared to previous generation? It shouldn't have modified anything (maybe some carriage return but should be the only difference).

See the builds bellow.
Builds taking more than 6min means that the source code have changed.
Eg. To to details of `Angular / ngx-default (pull_request)` at `Merge: merge project diff` step:

You will see the diff:
```
@@ -868,6 +868,13 @@
             >
               This field is required.
             </small>
+            <small
+              class="form-text text-danger"
+              [hidden]="!editForm.get('localDateRequiredTom')?.errors?.ZonedDateTimelocal"
+              myPrefixTranslate="entity.validation.ZonedDateTimelocal"
+            >
+              This field should be a date and time.
+            </small>
           </div>
         </div>
```

generators/entity/index.js
```diff
@@ -228,11 +249,6 @@ class EntityGenerator extends BaseBlueprintGenerator {
       setupSharedConfig() {
         const context = this.context;
 
-        context.protractorTests = context.testFrameworks.includes('protractor');
-        context.gatlingTests = context.testFrameworks.includes('gatling');
-        context.cucumberTests = context.testFrameworks.includes('cucumber');
-
-        context.jhiPrefixDashed = _.kebabCase(context.jhiPrefix);
         context.jhiTablePrefix = this.getTableName(context.jhiPrefix);
```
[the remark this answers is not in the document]
Tcharl: This one should be already configured: L#225 I'm calling loadAppConfig(), which calls `generator-base#loadDerivedAppConfig`, which calls `loadDerivedMicroserviceAppConfig()`which feeds the context

Tcharl:
I got the one you mentioned :-) and thank you for the TIPs!. It was a bug on my side (I used Temporal which included ZonedDateTime, LocalDate and Instant instead of Timed, which only includes ZonedDateTime and Instant).

Let's rebuild it and see if things are always modified: fingers crossed! Anyway, the review is still welcome: I'll hope I don't have included other side effect

Tcharl:
So, good to go? Ready for the next round (I promise, it will be less than 20 files this time ;-))

mshima (approved):

pascalgrimaud:
Go ahead @Tcharl !

PR #14682, opened 2021-04-14
Implement a cache workflow

mshima:
Implement a workflow exclusive to build our cache.
- npm: use our dependabots package.json for a more complete cache.
- maven: generate applications with most of our options and build the cache with go-offline goal.
- Gradle: use same applications from maven and execute some operations to build the cache.

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -19,8 +19,6 @@
 
 name: Angular
 on:
-  schedule:
-    - cron: '0 0 * * *'
   push:
     branches-ignore:
       - 'dependabot/**'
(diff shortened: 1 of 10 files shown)


atomfrede:
Could we use our gradle enterprise instance for the cache? Maybe even for the maven part?

mshima:
@atomfrede I think it will be complementary:
- Outside of GitHub infrastructure, it will be more stable if we use GitHub Actions cache at the main repository.
But an additional cache for making things faster would be great.
- It will be great for jhipster-bom and daily-builds, we don't have cache configured there.
- How the incremental cache works?
- Cache fetching? It requires pom.xml or build.gradle, or a key is enough to fetch it?
- Cache invalidation? If caching npm repository, since we don't have package-lock.json at our generated applications, we should invalidate the main cache daily (weekly at most), otherwise it can become really big.

The first step would be to build jhipster-bom cache using gradle enterprise inside the cache workflow.
It could contain npm/maven/gradle caches and be used at jhipster-bom/daily-builds at first.

I could be wrong, since I never tried gradle enterprise, I can help better when some kind of support for it lands.

pascalgrimaud:
Sorry @mshima : I don't have time the last days to review it, so don't hesitate to go ahead and merge it, like right now :)

atomfrede:
@mshima Sounds like we are on the same track. Will try to add gradle enterprise to the generated applications, such we can have buildscans etc. Afterwards the first project for caching would be indeed the bom, which should be fairly easy as it is a typical jvm project. Let's see how that works out.

PR #14703, opened 2021-04-18
use constants in entity-server

Tcharl:
Contributes to #14235

---

### DIFF
--- a/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
@@ -173,7 +173,7 @@ _%>
     public String toString() {
         return "<%= dtoClass %>{" +
 <%_ for ([idx, reference] of dtoReferences.entries()) {
-  const isNumeric = (reference.field && reference.field.fieldTypeNumeric) ||  (reference.relationship && reference.relationship.relatedField && reference.relationship.relatedField.fieldTypeNumeric)_%>
+  const isNumeric = (reference.field && reference.field.fieldTypeNumeric) ||  (reference.relationship && reference.relationship.otherEntity.primaryKey && reference.relationship.otherEntity.primaryKey.typeNumeric)_%>
             "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (!isNumeric) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (! isNumeric) { %>+ "'" <% } %>+
 <%_ } _%>
             "}";
--- a/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
@@ -174,7 +174,7 @@ _%>
         return "<%= dtoClass %>{" +
 <%_ for ([idx, reference] of dtoReferences.entries()) {
   const isQuoted = reference.field && !reference.field.fieldTypeNumeric _%>
-            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (isQuoted) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (!isQuoted) { %>+ "'" <% } %>+
+            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (isQuoted) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (isQuoted) { %>+ "'" <% } %>+
 <%_ } _%>
             "}";
     }
--- a/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
+++ b/generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
@@ -21,7 +21,7 @@ package <%= packageName %>.service.dto;
 import java.util.Objects;
 <%_ if (typeof javadoc != 'undefined') { _%>
 import io.swagger.annotations.ApiModel;
-<%_ } if (importApiModelProperty === true) { _%>
+<%_ } if (importApiModelProperty) { _%>
 import io.swagger.annotations.ApiModelProperty;
 <%_ } _%>
 <%_ if (fieldsContainInstant) { _%>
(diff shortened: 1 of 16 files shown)


generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -150,34 +148,34 @@ _%>
...
     // prettier-ignore
     @Override
     public String toString() {
         return "<%= dtoClass %>{" +
-            <%_ for ([idx, reference] of dtoReferences.entries()) {
-                const isNumeric = reference.type && ['integer', 'long', 'float', 'double', 'bigdecimal'].includes(reference.type.toLowerCase()); _%>
-            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (! isNumeric) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (! isNumeric) { %>+ "'" <% } %>+
-            <%_ } _%>
+<%_ for ([idx, reference] of dtoReferences.entries()) {
+  const isNumeric = (reference.field && reference.field.fieldTypeNumeric) ||  (reference.relationship && reference.relationship.relatedField && reference.relationship.relatedField.fieldTypeNumeric)_%>
```
mshima, line 176: relatedField is the label. We need to use otherEntity.primaryKey here.
```suggestion
  const isNumeric = (reference.field && reference.field.fieldTypeNumeric) ||  (reference.relationship && reference.relationship.otherEntity.primaryKey && reference.relationship.otherEntity.primaryKey.typeNumeric)_%>
```

generators/entity/index.js
```diff
@@ -753,12 +763,17 @@ class EntityGenerator extends BaseBlueprintGenerator {
         this.context.entityContainsCollectionField = this.context.relationships.some(relationship => relationship.relationshipCollection);
       },
 
+      processEntityPrimaryKeysDerivedProperties() {
+        derivedPrimaryKeyProperties(this.context.primaryKey);
```
mshima, line 767: primaryKey doesn't exist for embedded entities.
```suggestion
        if (!this.context.primaryKey) return;
        derivedPrimaryKeyProperties(this.context.primaryKey);
```

utils/field.js
```diff
@@ -183,7 +183,7 @@ const generateFakeDataForField = (field, faker, changelogDate, type = 'csv') =>
   return data;
 };
 
-function derivedProperties(field) {
+function _derivedProperties(field) {
```
mshima, line 186: Why the function rename?

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -150,34 +148,34 @@ _%>
...
     @Override
     public String toString() {
         return "<%= dtoClass %>{" +
-            <%_ for ([idx, reference] of dtoReferences.entries()) {
-                const isNumeric = reference.type && ['integer', 'long', 'float', 'double', 'bigdecimal'].includes(reference.type.toLowerCase()); _%>
-            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (! isNumeric) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (! isNumeric) { %>+ "'" <% } %>+
-            <%_ } _%>
+<%_ for ([idx, reference] of dtoReferences.entries()) {
+  const isNumeric = (reference.field && reference.field.fieldTypeNumeric) ||  (reference.relationship && reference.relationship.otherEntity.primaryKey && reference.relationship.otherEntity.primaryKey.typeNumeric)_%>
+            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (!isNumeric) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (! isNumeric) { %>+ "'" <% } %>+
```
mshima, line 177: We can simplify the condition.
```suggestion
  const isQuoted = reference.field && !reference.field.fieldTypeNumeric _%>
            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (isQuoted) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (!isQuoted) { %>+ "'" <% } %>+
```

utils/field.js
Tcharl: It's a private one: I thought it was a good idea

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
Tcharl: I really prefer the numeric keyword, which is more meaningful than the quoted one. Also, I tried that 'simplification', but the primaryKey use case was also to be taken into account in the loop.

Tcharl: I really prefer the numeric keyword, which is more meaningful than the quoted one. Also, I tried that 'simplification', but the primaryKey use case was also to be taken into account in the loop.

utils/field.js
mshima: We should start to document our public/private api:
```
/**
 * @private
 **
```


But in this case, if function is exported for this use case, it should be public (blueprints may need it), since it's not exported it is effectively private and no need to document.

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
mshima: Testing the relationship primaryKey or and field it will/can generate something like:
Relationship with User:
- For jwt it will generate `user=User{id=1}`
- For oauth2 it will generate `user='User{id='UUID-...'}'`

generators/entity-server/templates/src/main/java/package/service/dto/EntityDTO.java.ejs
```diff
@@ -150,34 +148,34 @@ _%>
...
     @Override
     public String toString() {
         return "<%= dtoClass %>{" +
-            <%_ for ([idx, reference] of dtoReferences.entries()) {
-                const isNumeric = reference.type && ['integer', 'long', 'float', 'double', 'bigdecimal'].includes(reference.type.toLowerCase()); _%>
-            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (! isNumeric) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (! isNumeric) { %>+ "'" <% } %>+
-            <%_ } _%>
+<%_ for ([idx, reference] of dtoReferences.entries()) {
+  const isQuoted = reference.field && !reference.field.fieldTypeNumeric _%>
+            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (isQuoted) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (!isQuoted) { %>+ "'" <% } %>+
```
mshima, line 177: Oops my bad.
```suggestion
            "<%= idx === 0 ? '' : ', ' %><%= reference.name %>=<% if (isQuoted) { %>'<% } %>" + <%= buildJavaGet(reference) %> <% if (isQuoted) { %>+ "'" <% } %>+
```

mshima (approved):
LGTM

PR #14854, opened 2021-05-03
JWTRelay allow no authorization header

iconben:
When relay requests, gateways should not assume that there is always authorization header. There are scenarios that services expose endpoints that allow anonymous access, which in most cases contain no authorization header. Current behavior of throwing an exception is not quite reasonable IMHO.

e.g. In my case, I use a user service instead of gateway to provide authentication service thus I permit all visits to /services/user/api/authenticate, no authorization header needed of course. 


mraible (approved):

DanielFran:
Thanks @iconben for the contribution

PR #15056, opened 2021-05-20
fix: missing dashes on Kafka configuration

fdelbrayelle:
---


DanielFran (approved):

pascalgrimaud:
The problem is simply consistency.
Sometimes we used dashed, sometimes not.

The current Kafka configuration work well. Your problem is because you tried to add a property with dash in your module, instead of not using a dash.
Not sure if generator-jhipster should change, to be compatible with your module.
I think, your module should change this part, to be compatible with generator-jhipster and the current config for existing projects

fdelbrayelle:
@pascalgrimaud Which part of the module are you making a reference of? I didn't modify any yml configuration recently :thinking:

pascalgrimaud:
I'm referring to your error:

```
Error parsing file src/main/docker/app.yml: SyntaxError: A collection cannot be both a mapping and a sequence (33:7)
  31 |       - 127.0.0.1:9092:9092
  32 |     environment:
> 33 |       KAFKA_BROKER_ID: 1
     |       ^^^^^^^^^^^^^^^^^^
> 34 |       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 35 |       - KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://kafka:29092,PLAINTEXT_HOST://localhost:9092
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 36 |       KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 37 |       KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 38 |       KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 39 |       KAFKA_ADVERTISED_HOST_NAME: kafka
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 40 |   zookeeper:
     | ^
```

You should add the property without dash, I think:
https://github.com/fdelbrayelle/generator-jhipster-kafka/blame/475b46eda587774108b3cae99923433dd8d36ceb/generators/app/files.js#L376

fdelbrayelle:
Thank you I didn't understand and notice that detail!

vishal423:
@pascalgrimaud , I think the consistency issue is on `generator-jhipster` side as few component configurations use `dash` notation. like: https://github.com/jhipster/generator-jhipster/blob/e39889ff66c4cd652493fe6fba46579dd15291f2/generators/docker-compose/templates/docker-compose.yml.ejs#L54

I believe we should be consistent on our approach so that module owners can extend without worrying about the notation to use

fdelbrayelle:
Yes @vishal423. Before it was like this in `app.yml` for Kafka : `- KEY=VALUE`. It became recently : `KEY: VALUE` and so the module builds fail as I update some of these keys/values.

pascalgrimaud:
Yes agree @vishal423

Issue #14566, opened 2021-04-02
Reduce CI builds runtime.


mshima:
##### **Overview of the issue**

Our CI builds are taking too much time.
Sometimes it can take more than 6 hours for PR merges tests to finish.

##### **Motivation for or Use Case**

Developer experience is really bad.

##### **Suggest a Fix**

Reduce tests runtime.
- Improve backend test time.
- Improve frontend test time.
- Improve e2e test time.
- Filter tests.

- [ ] Checking this box is mandatory (this is just to show you read everything)

CGarces:
Some builds check for the draft status at the end of the task.
https://github.com/jhipster/generator-jhipster/blob/main/.github/workflows/generator.yml#L71
Maybe should be check earlier?

deepu105:
This is another reason for the slowness https://github.com/jhipster/generator-jhipster/issues/15205

PR #14799, opened 2021-04-26
Conditionally skip frontend build tasks

vishal423:
Closes #13796 
---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) was updated if necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (bellow reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

PR #15209, opened 2021-06-03
Enable webpack cache at production builds for react and vue.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/14566.

Reduces React and Vue builds by ~1m.

---

### DIFF
--- a/generators/client/templates/react/webpack/webpack.common.js.ejs
+++ b/generators/client/templates/react/webpack/webpack.common.js.ejs
@@ -61,6 +61,21 @@ const getTsLoaderRule = env => {
 module.exports = options => merge(
 // jhipster-needle-add-webpack-config - JHipster will add custom config
 {
+  cache: {
+    // 1. Set cache type to filesystem
+    type: 'filesystem',
+    cacheDirectory: path.resolve(__dirname, '../<%= buildDir %>webpack'),
+    buildDependencies: {
+      // 2. Add your config as buildDependency to get cache invalidation on config change
+      config: [
+        __filename,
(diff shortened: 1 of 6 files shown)


deepu105 (commented):
I'm a bit worried about enabling cache for prod TBH. Shouldn't prod builds be as clean as possible? if we are doing this for prod, we should make every attempt to make sure all latest code gets in by enabling a postInstall cleanup of cache as recommended in webpack docs and by making sure all tips/warning [here](https://webpack.js.org/configuration/other-options/#cache) are considered

mshima:
@deepu105 I haven’t enabled on purpose when I introduced the filesystem cache. But we are already using up-2-date at gradle and a custom maven implementation https://github.com/jhipster/generator-jhipster/pull/14799.

IMO is better a webpack solution than the alternative.

deepu105:
I didnt realize we had it or prod builds with maven/gradle

vishal423:
Just to be clear, maven caching of webpack profile tasks does not apply to production build. If you do see that's being applied, then, do raise an issue and I will handle that out.

mshima:
It's right. The only dev.
Anyway seems safe to me.
The caching structure is stored at `target/webpack' and maven removes `target/class/static`.

atomfrede:
> Just to be clear, maven caching of webpack profile tasks does not apply to production build. If you do see that's being applied, then, do raise an issue and I will handle that out.

Same for gradle. On purpose that task is not configured to be cacheable or up-2-date. Is is always executed regardless of the previous build.

atomfrede (approved):

Issue #15074, opened 2021-05-24
DefinePlugin NODE_ENV conflicting values


rvong:
##### **Overview of the issue**
On fresh generation of JHipster 7 generated by JHipster online, upon `./mvnw`
I get 
```
[INFO] WARNING in DefinePlugin
[INFO] Conflicting values for 'process.env.NODE_ENV'
```
Why is this happening and what's the fix if it's needed?

##### **Reproduce the error**

Create JHipster7 package and run `./mvnw`

##### **Browsers and Operating System**

Manjaro 20.2

sdkman jdk 11
nvm nodejs 14

- [x] Checking this box is mandatory (this is just to show you read everything)

manoj-mass:
Found this.
https://stackoverflow.com/questions/55185601/webpack-process-env-undefined-using-defineplugin-and-dotenv

rvong:
Not sure what above bot comment means.

mshima:
@pascalgrimaud bounty claimed https://opencollective.com/generator-jhipster/expenses/50248.

pascalgrimaud:
@mshima : approved

PR #15210, opened 2021-06-04
Replace node especific process.env with injected variables at React.

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/15074.

Apply same fix used for angular https://github.com/jhipster/generator-jhipster/pull/15107. 

---

### DIFF
--- a/generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
+++ b/generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
@@ -16,14 +16,6 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -%>
-const config = {
-  VERSION: process.env.VERSION
-};
-
-export default config;
-
-export const SERVER_API_URL = process.env.SERVER_API_URL;
-
 export const AUTHORITIES = {
   ADMIN: 'ROLE_ADMIN',
   USER: 'ROLE_USER'
--- a/generators/client/templates/react/webpack/environment.js.ejs
+++ b/generators/client/templates/react/webpack/environment.js.ejs
@@ -1,0 +1,35 @@
+<%#
+ Copyright 2013-2021 the original author or authors from the JHipster project.
+
+ This file is part of the JHipster project, see https://www.jhipster.tech/
+ for more information.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+<%_ if (buildToolUnknown) { _%>
+const packageJson = require('./../package.json');
+<%_ } _%>
+
+module.exports = {
+<%_ if (buildToolUnknown) { _%>
+  VERSION: packageJson.version,
+<%_ } else { _%>
+  // APP_VERSION is passed as an environment variable from the Gradle / Maven build tasks.
+  VERSION: process.env.hasOwnProperty('APP_VERSION') ? process.env.APP_VERSION : 'DEV',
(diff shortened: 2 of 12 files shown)


generators/client/templates/react/webpack/webpack.common.js.ejs
```diff
@@ -111,24 +109,13 @@ module.exports = options => merge(
   },
   plugins: [
     new webpack.DefinePlugin({
-      'process.env': {
-        NODE_ENV: `'${options.env}'`,
```
deepu105, line 115: I think React and some libs like Redux for example internally still use `process.env.NODE_ENV` for certain logic and hence we should ensure its value is passes as well IMO, unless its automatically set

generators/client/templates/react/webpack/environment.js.ejs
```diff
@@ -0,0 +1,40 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-%>
+<%_ if (buildToolUnknown) { _%>
+const packageJson = require('./../package.json');
+<%_ } _%>
+
+module.exports = {
+  __ENVIRONMENT__: 'production',
```
deepu105, line 24: what about `development`?

deepu105 (changes requested):

generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
```diff
@@ -16,13 +16,25 @@
...
+declare const __BUILD_TIMESTAMP__: string;
+<%_ } _%>
+declare const __VERSION__: string;
+declare const __SERVER_API_URL__: string;
+declare const __ENVIRONMENT__: string;
+
+<%_ if (enableTranslation) { _%>
+export const BUILD_TIMESTAMP = __BUILD_TIMESTAMP__;
+<%_ } _%>
+export const ENVIRONMENT = __ENVIRONMENT__;
```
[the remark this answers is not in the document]
mshima: > Define plugin inlines the actual code on compilation and the minification will remove the unreachable code from what I recall. Would that still happen?

Yes, but it should smarter now and remove constants too, not sure.
We could use __ENVIRONMENT__ directly too, but I think it's cleaner to have a file for inputs.

generators/client/templates/react/webpack/webpack.common.js.ejs
mshima: I don't know Redux.
Use of `process.env` should be avoided: https://github.com/webpack/changelog-v5/blob/master/MIGRATION%20GUIDE.md#level-5-runtime-errors.

generators/client/templates/react/webpack/environment.js.ejs
mshima: This is used by jest, __ENVIRONMENT__ is replaced at webpack.common.js.

generators/client/templates/react/webpack/webpack.common.js.ejs
deepu105: Ya but many React frameworks might use it s its better to make sure its set. Redux definitely does it for enabling dev tools and stuff

generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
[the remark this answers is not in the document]
deepu105: can you look at a build and verify, if its replaced its fine else would be nice to use directly so production bundles will have unwanted code removed

generators/client/templates/react/webpack/webpack.common.js.ejs
mshima: According to https://webpack.js.org/guides/production/#specify-the-mode.
NODE_ENV is set by webpack when mode is set, so no need to set it manually.

generators/client/templates/react/src/main/webapp/app/config/constants.ts.ejs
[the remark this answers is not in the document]
mshima: Looks good.

mshima:
Dropped vue changes, it doesn't have the linked issue and it needs some cleanup.

mshima:
It's been open for too long. Merging.

PR #15225, opened 2021-06-05
Publish test results on Azure pipeline generator

egvimo:
Upates Azure DevOps subgenerator:

- Publish test results for backend and frontend tests
- Publish Cypress screenshots if E2E tests failed

---

### DIFF
--- a/generators/ci-cd/templates/azure-pipelines.yml.ejs
+++ b/generators/ci-cd/templates/azure-pipelines.yml.ejs
@@ -82,6 +82,13 @@ jobs:
       command: 'custom'
       customCommand: ' run ci:backend:test'
     displayName: 'TESTS: backend'
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFormat: 'JUnit'
+      testResultsFiles: '**/TEST-*.xml'
+      searchFolder: '$(Build.SourcesDirectory)/<%= buildDir %>test-results'
+    condition: succeededOrFailed()
+    displayName: 'TESTS: publish test results'
   <%_ } _%>
 <%_ if (!skipClient) { _%>
   - task: Npm@1
@@ -89,6 +96,12 @@ jobs:
       command: 'custom'
       customCommand: ' run ci:frontend:test'
     displayName: 'TESTS: frontend'
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFormat: 'JUnit'
+      testResultsFiles: '$(Build.SourcesDirectory)/<%= buildDir %>test-results/TESTS-results-jest.xml'
+    condition: succeededOrFailed()
+    displayName: 'TESTS: publish test results'
 <%_ } _%>
 <%_ if (!skipServer) { _%>
   - task: Npm@1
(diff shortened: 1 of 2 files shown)


DanielFran:
@egvimo Can you update snapshot files?

generators/ci-cd/templates/azure-pipelines.yml.ejs
```diff
@@ -119,6 +132,12 @@ jobs:
   <%= commented %>    CYPRESS_ENABLE_RECORD: <%= cicdIntegrationsCypressDashboard %>
   <%= commented %>    CYPRESS_PROJECT_ID: $(CYPRESS_PROJECT_ID)
   <%= commented %>    CYPRESS_RECORD_KEY: $(CYPRESS_RECORD_KEY)
+  <%= commented %>- task: PublishPipelineArtifact@1
+  <%= commented %>  inputs:
+  <%= commented %>    artifactName: 'cypress-screenshots'
+  <%= commented %>    targetPath: '$(System.DefaultWorkingDirectory)/build/cypress/screenshots'
```
mshima, line 138:
```suggestion
  <%= commented %>    targetPath: '$(System.DefaultWorkingDirectory)/*/cypress/screenshots'
```
maven uses target

generators/ci-cd/templates/azure-pipelines.yml.ejs
```diff
@@ -82,13 +82,26 @@ jobs:
...
 <%_ if (!skipClient) { _%>
   - task: Npm@1
     inputs:
       command: 'custom'
       customCommand: ' run ci:frontend:test'
     displayName: 'TESTS: frontend'
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFormat: 'JUnit'
+      testResultsFiles: '$(System.DefaultWorkingDirectory)/build/test-results/TESTS-results-jest.xml'
```
mshima, line 102:
```suggestion
      testResultsFiles: '$(System.DefaultWorkingDirectory)/*/test-results/TESTS-results-jest.xml'
```

generators/ci-cd/templates/azure-pipelines.yml.ejs
egvimo: @mshima Thanks for your suggestion. Unfortunately this does not work for all paths (Path does not exist error).

I use Gradle as build system and did not thought about Maven which would have `target`  as output directory.

How about declaring a variable on top of the Pipeline and using it for all paths?

```yaml
variables:
<%_ if (buildToolMaven) { _%>
  buildDirectory: $(Build.SourcesDirectory)/target
<%_ } else if (buildToolGradle) { _%>
  buildDirectory: $(Build.SourcesDirectory)/build
<%_ } _%>
```

mshima: And now? Using DIST_DIR?

egvimo: If we use the variables from the pipeline we could then use that variable for all paths:

```yaml
targetPath: '$(outputDirectory)/cypress/screenshots'
```

If we use your approach, then `BUILD_DIR` would be the correct variable.

We could also combine that to:

```yaml
variables:
  outputDirectory: $(Build.SourcesDirectory)/<%= BUILD_DIR %>

# ... and use it ...

targetPath: '$(outputDirectory)/cypress/screenshots'
```

Which one do you prefer?

mshima: Sorry, my suggestion was wrong.
The var I was looking for is buildDir.
It should have target or build value.

If you think it’s better to extract the variable, ok to me.

mshima: Your choice.
BUILD_DIR is correct.

mshima (approved):
LGTM

Issue #15205, opened 2021-06-03
Auto cancel redundant/stale GitHub action builds


deepu105:
I noticed that the Github action builds on our repos don't auto cancel on newer commits, like it used to do on Travis and this is really clogging up our pipelines and makes the GitHub builds really slow and more over this is wasting unnecessary resources and energy overall

So we need to try one of the solution mentioned [here](https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779/2)

- [x] Update for generator-jhipster
- [ ] Update for hipster/labs daily builds
- [x] Update for BOM project
- [ ] Update for JHipster online

deepu105:
@pascalgrimaud do you think 300 is good bounty for this or should we make it 500?

deepu105:
can be done along side https://github.com/jhipster/generator-jhipster/issues/15129 as well

SudharakaP:
I'll do this over the weekend; shouldn't be difficult; but if anyone wanna take it before that feel free to. 😄

mshima:
Last time I checked, wasn't possible, in the link you provided there are some alternatives.
I will take a look, it's holiday here today.

pascalgrimaud:
hehe, same as @mshima 
The last time I wanted this, it was not possible. Maybe it has been improved...?

mshima:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency should do the trick.
I in beta, but is upstream.

deepu105:
@mshima @SudharakaP seems to have assigned this to him as well so just make sure to coordinate so that both of you dont end up doing it :)

SudharakaP:
@mshima : I've assigned it to you. If you feel you need help or something feel free to ping me I can take a look.

mshima:
I don't think it's worth to add this to daily builds or to jhipster online.
The number of PRs and workflows is too low.
At generator-jhipster and jhipster-bom a PR can trigger almost 40 builds.

deepu105:
Ya those are just placeholders. Feel free to update to what you think is more applicable

DanielFran:
> 
> 
> I don't think it's worth to add this to daily builds or to jhipster online.
> The number of PRs and workflows is too low.
> At generator-jhipster and jhipster-bom a PR can trigger almost 40 builds.

@deepu105 Following last @mshima comment can this issue be closed?

deepu105:
[...]

Issue #15129, opened 2021-05-26
Create GitHub Actions to improve CI maintainability.


mshima:
##### **Overview of the feature request**

Implement custom actions to be shared between jhipster repositories.
- Implement actions locally in generator-jhispter.
- Migrate jhipster-bom to use generator-jhipster actions.
- Evaluate daily-build to use generator-jhipster actions.

##### **Motivation for or Use Case**

JHipster CI consists in:
 - 3 main repositories (generator-jhipster, jhipster-bom, daily-builds).
 - Each repository contains lots of workflows.
 - Each repository contains different scripts.
 - Many derived repositories.

Currently, to change some build process, each workflow must be updated manually to reflect the changes.

JHipster-BOM repository CI is failing, but should be identical to the generator-jhipster CI.

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
cc @pascalgrimaud @DanielFran

pascalgrimaud:
Agree with you @mshima 

When I coded this part, in the past, I tried to put everything inside shell scripts:
- to not depend on specific CI
- to be able to use the same shell scripts in different repo

Today, we have a lot of logic directly inside yml files. I think we should move logic into shell scripts. The yml files should only contain configuration (ngx, react, vue, database etc)

mshima:
I don't think we will have many benefits from up-2-date checks.
We already have our own up-2-date using diff.

When pushing to main branch is important to run the tests due to transitive dependencies.
Would be great to have it in others branchs and personal branch if possible.

mshima:
@pascalgrimaud bounty claimed https://opencollective.com/generator-jhipster/expenses/44491.

pascalgrimaud:
@mshima : approved

PR #15244, opened 2021-06-07
Minor adjusts to workflows.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/15205 and https://github.com/jhipster/generator-jhipster/issues/15129.

Description in each commit.

---


DanielFran (approved):
@mshima Just to be sure, if we want to use in generator a bom released version we just needs to update jhipster-bom-branch value to release?

Then this will apply to generator and daily builds if we use same action file?

How about in bom project? Should we define bom branch as local in each workflow to override the property?

mshima:
> @mshima Just to be sure, if we want to use in generator a bom released version we just needs to update jhipster-bom-branch value to release?

Yes, change here https://github.com/jhipster/generator-jhipster/blob/2156b1321893f7f290b5169248b2adeb082441dd/.github/actions/setup/action.yml#L74.\

> Then this will apply to generator and daily builds if we use same action file?

Yes, if the workflow doesn't override the default value.

> How about in bom project? Should we define bom branch as local in each workflow to override the property?

There is no local option for jhipster-bom-branch. It will try to load the branch local.
When executed by jhipster-bom repository it executes https://github.com/jhipster/generator-jhipster/blob/2156b1321893f7f290b5169248b2adeb082441dd/test-integration/scripts/10-install-jhipster.sh#L11-L22

Issue #15069, opened 2021-05-21
Investigate and fix refresh tokens for OAuth with Keycloak and Okta


mraible:
##### **Overview of the issue**

It should be possible to add `offline_access` to the scopes for Spring Security OAuth and get a refresh token along with an access token. JHipster should automatically get a new access token after 5 minutes (that's the default timeout we have set for access tokens in Keycloak). 

##### **Motivation for or Use Case**

https://devforum.okta.com/t/short-timeout-on-reactive-java-microservices-with-spring-boot-and-jhipster-blogpost/14270

##### **Reproduce the error**

Add `offline_access` to the list of scopes. Log in and wait 5+ minutes. You should be able to perform actions without having to log in again.

##### **JHipster Version(s)**

7.0.1

##### **JHipster configuration**

```
jhipster jdl reactive-ms
```

roh0sun:
I've investigated this issue / feature for some time.
I think there is no `RefreshToken` authorization flow in the Spring Boot server.
But all the needed components are already there.

For POC, I've tested a simple REST contoller.
This can successfully refresh the access token after 5+ minutes.

```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.security.core.Authentication;
import org.springframework.security.core.context.ReactiveSecurityContextHolder;
import org.springframework.security.core.context.SecurityContext;
import org.springframework.security.oauth2.client.OAuth2AuthorizationContext;
import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
import org.springframework.security.oauth2.client.RefreshTokenReactiveOAuth2AuthorizedClientProvider;
import org.springframework.security.oauth2.client.annotation.RegisteredOAuth2AuthorizedClient;
import org.springframework.security.oauth2.client.web.server.AuthenticatedPrincipalServerOAuth2AuthorizedClientRepository;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
import org.springframework.web.server.ServerWebExchange;
import reactor.core.publisher.Mono;

@RestController
class TokenRefreshController {

    private final Logger log = LoggerFactory.getLogger(TokenRefreshController.class);

    @Autowired
    private AuthenticatedPrincipalServerOAuth2AuthorizedClientRepository authorizedClientRepository;

[...]

PR #15424, opened 2021-06-21
Enable OAuth2 refresh token.

mshima:
- Implement OAuth2{Reactive}RefreshTokensWebFilter.java and instantiate dependencies.
- Inject OAuth token/user into required services for tests.
- Fix angular requiring api/account twice, causing problem with user synchronization.

Fixes https://github.com/jhipster/generator-jhipster/issues/15069.

---


generators/server/templates/src/main/resources/logback-spring.xml.ejs
```diff
@@ -45,6 +45,7 @@
     </root>
 -->
 
+    <logger name="<%= packageName %>" level="INFO"/>
```
DanielFran, line 48: This is not needed, already defined in application.yml

mshima:
```suggestion
```

Tcharl:
Right, using filter is way better than what we did.

generators/server/files.js
```diff
@@ -1165,6 +1165,35 @@ const serverFiles = {
       templates: ['META-INF/services/reactor.blockhound.integration.BlockHoundIntegration'],
     },
   ],
+  springBootOauth2: [
+    {
+      condition: generator => generator.authenticationType === 'oauth2' && generator.applicationType === 'monolith',
```
Tcharl, line 1170: constants?

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -999,7 +999,16 @@ public class UserService {
             })
             <%_ } _%>
             .collect(Collectors.toSet()));
-        return <% if (databaseType !== 'no' && !reactive) { %>new <%= asDto('AdminUser') %>(syncUserWithIdP(attributes, user))<% } else if (!reactive) { %>user<% } %><% if (databaseType === 'no' && reactive) { %>Mono.just(user)<% } else if (reactive) { %>syncUserWithIdP(attributes, user).flatMap(u -> Mono.just(new <%= asDto('AdminUser') %>(u)))<% } %>;
+
+    <%_ if (databaseType === 'no') { _%>
```
Tcharl, line 1003: constants?

Tcharl:
Everything looks good now: do you want I merge it?

mshima:
> Everything looks good now: do you want I merge it?

If it looks ok, then yes 😄.

generators/server/templates/src/test/java/package/web/rest/LogoutResourceIT.java.ejs
```diff
@@ -66,23 +65,30 @@ class LogoutResourceIT {
     @Autowired
     private <% if (!reactive) { %>Web<% } %>ApplicationContext context;
 
+    @Autowired
+    private <%= reactivePrefix %>OAuth2AuthorizedClientService authorizedClientService;
+
+    @Autowired
+	private ClientRegistration clientRegistration;
```
mraible, line 72: Indent 4 spaces

mraible (commented):
We should add `offline_access` as a scope and configure Keycloak to return a refresh token by default.

mraible (approved):
😃🤩✌️

mshima:
> We should add `offline_access` as a scope and configure Keycloak to return a refresh token by default.

@mraible in my tests `offline_access` is not required to enable refresh tokens, it's already enabled by default.

At least our angular client, uses cookies, which expires and I think this makes `offline_access` useless.
And we don't provide a mobile app, so IMO `offline_access` is not a sensible default.

mraible:
We have React Native, Ionic, and Flutter support! 😊

I think we should have it enabled by default in dev mode, maybe not in production.

Tcharl:
@mraible offline access is achieved at Keycloak level, not spring-sec one right? Also, the claim is already configured on  keycloak side.
Merging it, from what I've seen & tested everything is ok and much better than before

mraible:
The client (Spring Security in this case) needs to pass in `offline_access` as a scope in order to get a refresh token. If one is returned without this scope, that seems strange.

Issue #15490, opened 2021-06-28
New sub generators for modularity


pascalgrimaud:
##### **Overview of the feature request**

The discussion started here https://groups.google.com/g/jhipster-dev/c/y_sER_CM3Hc

##### **Motivation for or Use Case**

The goal will be to start to create some new sub generators. To resume quickly:

Pro:
- easier to maintain, no more if else in templates with too much combinations
- easier to add new options
- easier for new contributors

Cons:
- duplication of templates
- some options will be maintained better than others (but IMO, it's normal, if you compare an option used by 50% of projects, with an option used by less than 0.5%)

Then later, when the v8 will start, we'll migrate the existing sub generators into the new ones.
For now, no impact on existing projects as all these new sub generators are totally independent!

##### **Roadmap**

We can start to add these following sub generator:
- init : with default files, when you start a project
- maven
- gradle
- spring-boot

Then, we can think about:
- spring-boot-mvc
- spring-boot-reactive
- security-jwt
- security-oauth2
- database-postgresql
- database-mysql
- etc.

- [x] Checking this box is mandatory (this is just to show you read everything)

atomfrede:
Keep it open

mshima:
I have been thinking on how to better integrate entities.

Keeping same architecture that adds an entity-foo generator for each foo (eg sql, mongodb, cassandra, neo4j, couchbase, elasticsearch, angular, react, vue) generator, our generators will become too bloated.
And the workflow will not be as simple as a `this.composeWith('entity-foo')`, it would need some register to keep the entity-* generators order consistent.

So we should drop entity-* generators and replace with:
- new priorities called `preparingEntities` and another called `configuringEntities`.
- support methods at generator-base to be used in the priorities.
  - `this.configureEachEntity(({entity, entityConfig}) => {})` at `configuringEntities`
  - `this.prepareEachEntity(({entity, entityWithConfig}) => {})` at `preparingEntities` and  `configuringEntities`
  - `this.prepareEachField(({entity, entityWithConfig, field}) => {})` at `preparingFields`
  - `this.prepareEachRelationship(({entity, entityWithConfig, relationship}) => {})` at `preparingRelationships`
  - `this.writeEachEntity(({entity}) => {})` at `writing`
- `entity` generator should delegate to the project type generator (java, spring-boot, spring-boot-web, jhipster).

Tcharl:
[...]

PR #15540, opened 2021-07-03
Maven sub generator

pascalgrimaud:
Related to #15490 

---


mshima:
Should `jhipster maven` generate `jhipster init` files?

I am thinking about how can we reduce duplications in the `jhipster gradle` generator (prompts and config).
We could create a mixin to put common code, but I think se should create a unique `jhipster java-build-tool` (something like this), and add `jhipster maven/gradle` shortcuts for `jhipster java-build --build-tool maven/gradle`.
WDYT?

pascalgrimaud:
No, `jhipster maven` should not generate files from `jhipster init`, it is separated

About doing a common sub generator build tool, I don't think it's a good idea. I'd like to keep each options totally separated, with less common functions.
Later, if I improve the maven sub generator, there is no risk I can break the gradle sub generator. In the same time, I don't need to test gradle generation if I only modify the maven generator.

There will be common part, common prompt etc. But I think it's minor and we can live with it

Tcharl:
Agree with mshima: build tool options should be in common, as well as any layered option (platform, front, gateway, etc.).
Prompt: common 
Templates: not common
Js: common?
Constants: both

However the approach is really promising: +1.

pascalgrimaud:
If we regroup maven and gradle, then later if someone wants to add 'ant' for example, we need to maintain it.

If we do this for maven and gradle, we should do the same for security : jwt, oauth2, session, uaa.

Not sure you want to maintain all options...

Keeping stuff separated will force the stream lead to maintain the option.

Nothing against gradle, but I'm not a gradle user today. If I want to contribute and improve maven, I dont want to test and check if I broke gradle or not.

That's why I think sub generator should be simple and dedicated to only 1 option.

mshima:
Duplication is bad.

What about try to separate specialized generators and integration generators.
Where specialized generators is dump for one purpose while integration generators can provide decision about 2 similar specialized generators with similar configurations.

Maven is a specialized generator, we should keep it without prompts and make package name a required option.
And let the package name prompt to the generator that will choose the build tool.

I consider the init generator an integration generator that provides git, husky, prettier without any alternatives.

pascalgrimaud:
I understand your point of view.

Duplication is indeed bad, but if it will allow this:
- ease the creation of new generators, you just need to copy/past an existing one
- each generators are isolated, and do not depends on other generators
- if I modify one generator, I don't need to test others
- If I modify one generator, I won't be afraid to broke the others
- each stream lead / contributors will be responsible of his own generators

Doing specialized generators and integration generators could be a good idea, but I think it's too early in this roadmap.
I'd prefer to keep things simple.

My ideas are:
- each option = 1 sub generator, which is standalone, it means you can launch it separately
- at the end, 1 global generator with all questions (like today), which will use all these new generators

Doing a new sub generator build tool. Why not ?
But I don't know how to do it, I need to invest too much time to understand composing, but I don't have this time and I don't want sorry. I think about new contributors, which will be in the same case than me.

My personal opinion: for this use case, I worked on Maven, not Gradle. So why should I build another sub generator to add support for something I don't use ? Why should I think about another option when coding maven sub generator ?

So plz plz plz, let keep each generator simple ! :-)

mshima:
There is no way to be completely independent, in your PR maven generator already depends on init generator config/options.
Duplicate everything can become a maintenance nightmare.

> So plz plz plz, let keep each generator simple ! :-)

I will split a little more so we can make them more consistent and no duplications (only api reuse).
@pascalgrimaud do you want me to create a new PR or update this one?

pascalgrimaud:
Yes, you need to init project first, before using this generator, following the tree of options in my slides

Plz update this branch directly, no need new PR @mshima : thanks ;-)

mshima:
@pascalgrimaud done.

pascalgrimaud:
Thanks @mshima
I'll review and test it today

generators/maven/templates/.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,2 @@
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip
```
DanielFran, line 1:
```suggestion
distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.1/apache-maven-3.8.1-bin.zip
```

DanielFran (changes requested):
Use latest maven version 3.8.1

pascalgrimaud:
@mshima : just tested it, and the behaviour has changed.
- jhipster init : ok, it works as expected
- jhipster maven : no prompts, so I can't change the packageName

The generated files are correct, not the behaviour

pascalgrimaud:
Another comment, I forgot to add a new entry in `.yo-rc.json`, with buildTool=maven
How can we do this ? @mshima / @Tcharl ?

Honestly, I'm a little bit lost with the different refactoring / load common config (cc @avdev4j who encounted similar issues with BP)
See my previous comment about "simple" :D

generators/maven/templates/.mvn/wrapper/maven-wrapper.properties
pascalgrimaud: Thanks, I'll take care of it later, as soon as the behaviour is fixed

mshima:
> @mshima : just tested it, and the behaviour has changed.
> 
> * jhipster init : ok, it works as expected
> * jhipster maven : no prompts, so I can't change the packageName
> 
> The generated files are correct, not the behaviour

For java-package-manager generator, the flag that shows or hides the prompt can be the packageName.
For init generator should we use `prettierDefaultIndent`, `prettierJavaIndent`, a custom one? They are not a good alternatives for this flag.
To make it simpler, it's using baseName for every generator. This is not ideal for the generator independence approach you are trying right now, but IMO we shouldn't try to make every generator independent on a single project.

Current workaround is to use `--ask-answered` option.
The best fix I can think is to implement something like:
- new projects to use `jhipster init maven others...`
- existing projects to use `jhipster add maven others...` and force prompts to show

> Another comment, I forgot to add a new entry in `.yo-rc.json`, with buildTool=maven

That should be handled by the generator that will decide between maven or gradle.

> Honestly, I'm a little bit lost with the different refactoring / load common config (cc @avdev4j who encounted similar issues with BP)
> See my previous comment about "simple" :D

In my opinion this is the as simple as we can get. Our use case is not simple.
The only complication I can see is that we can have too many generators, but this is better than duplication.

Blueprints are not too complicated anymore to implement, but they have too much legacy code on old ones.
They are still using getAllJHipsterConfig, we have this.jhipsterConfig now.
The constructor can even be removed.

pascalgrimaud:
> That should be handled by the generator that will decide between maven or gradle

I think that's the main issue because with this approach, we'll have to do the same thing for database.
We'll have to add a generator which will ask which database we want. And inside this, we'll need to maintain ALL existing database: mysql, postgresql, mongodb, cassandra, couchbase, etc

Maintaining too much option is hard.
Keeping each generator totally independant will help us.

My goal here simple:
I want a sub generator or a module, which ask me which package name I want, and it generates for me all maven files I need, like in a real project.
I don't care about other option for now.
If someone wants to code the gradle generator, it will be done.

I think it's hard to share correctly my point of view, and maybe it's the same for you @mshima as my english is not so good, don't be offense by my words :-)

And don't hesitate to tell me if I'm totally wrong or if you don't understand something.

pascalgrimaud:
Another thing to understand this approach.
If I want to add specific question, exclusive to Maven, how can I do?

mshima:
I understand your current goal here, but I can't see this happening without becoming even more complicated than how it is now.
For example, we have a `pom.xml` in maven generator, the spring-boot generator will have another `pom.xml`. Postgres generator will have another `pom.xml`? How elastic search will integrate into Postgres's `pom.xml`? Frontend, caching, web socket, every generator will require some changes to `pom.xml`.
Same for Readme.md.

> We'll have to add a generator which will ask which database we want. And inside this, we'll need to maintain ALL existing database: mysql, postgresql, mongodb, cassandra, couchbase, etc

In my opinion this is the greatest jhipster feature and how open source works.

> Another thing to understand this approach.
> If I want to add specific question, exclusive to Maven, how can I do?

I didn't see the reason to add the basic structure for options/prompt at the maven generator.
Once added, just add the prompt, default config and option if needed.
https://github.com/jhipster/generator-jhipster/blob/2abe1e0b3a0d79c4a3a2ebc2f5a4f6cd49165820/generators/init/index.js#L84-L90

pascalgrimaud:
About the next generator spring boot, just ask yourself. What would you do in a real project ? No, you won't create a new pom.xml but instead:
- you'll add dependencies
- create main class
- create IT
- create application properties
That's all

mshima:
> About the next generator spring boot, just ask yourself. What would you do in a real project ? No, you won't create a new pom.xml but instead:

How do you do that?

pascalgrimaud:
Using needle, like we already do today in existing module

pascalgrimaud:
Here the repo I used when playing with these new generators:
https://github.com/pascalgrimaud/jhipster-hexa-app/pulls

That's what I'm trying to do here

mshima:
So I don't understand why do you want to execute 
`jhipster init; jhipster maven; jhipster spring-boot`
Instead of (any of these):
- `jhipster spring-boot` with spring-boot composing with init and maven due to generator dependency.
- `jhipster init maven spring-boot` with init composing with maven and spring-boot
- `jhipster init; jhipster add maven; jhipster add spring-boot`;

pascalgrimaud:
Because for my customer, I already have a specific init, and specific maven.
So I can use:
- jhipster init-personal
- jhipster maven-personal
- jhipster spring-boot, which does not depend too much on other generators

That's why I wanted to keep things inside these generators, without refactoring/common part. It's easier to copy/paste and customize

mshima:
We have modules and blueprints support, this is different.
It adds support to unrelated generators, by assuming the file exists and don't try to generate a complete working application.
I really don't like this approach.

In my opinion it goes agains jhipster goal.
Our goal is to generate a complete and modern web app or microservice architecture.

`jhipster spring-boot` should always try to generate a runnable application by default, even if it means we regenerate existing files.

pascalgrimaud:
Modules and Blueprints are not used enough by our users today. We don't have new releases since more than 1 year and no complains. It means, very few people code a new module or blueprint.
The main reason is the complexity.

Yes, I agree this approach is different, but I don't think it goes against the original goal of JHipster to produce full complete application, because the full complete application is still here, with the other generator: app, client, server.

These new generators are here to try a new approach, to check if:
- it is easier to maintain
- it matches better to our users

It's totally a new workflow, and I already discuss about this with people who hates JHipster. Why? Because:
- it generates too much files
- the generator itself is too complex

Finally, they like this approach, that's the reason I suggest it :-)

But I'm fine if we decide to not go in this way, and keep the current big workflow.
So let's remove the init generator and close this pull request, if we don't fully agree or convinced by this approach.

I'll use separate fork for my customers.

PierreBesson:
> In my opinion it goes agains jhipster goal.
Our goal is to generate a complete and modern web app or microservice architecture.

I would say this is one of the goals but not the absolute. Our goal is to help users develop their own applications. It's cool to be able to generate a "complete" application but it would be even better if this application contained just enough needed code for the users.

I don't see how changing the approach of how the generator works internally goes against this goal. I definitely see a lot of value of being able to generate JHipster applications "step by step" like @pascalgrimaud is suggesting. Moreover, this is a sentiment that is widely shared among the team.

@mshima As you know Yeoman better than us, maybe your issue is more of technicality of using yeoman subgenerators for this step-by-step approach as you feel generator/subgenerators should produce a runnable application. So maybe this could be implemented in a specific flag such as: `jhipster --add-component spring-boot`. In this case, the `add-component` invocations could be composed in a way different than yeoman composition (which might not be adapted to this way of working).

MathieuAA:
IMO, we should aim at being more Rails-y. I can't recall how many times I've said that, or heard others say the same thing.

One thing we haven't done, or maybe not enough, is talk about user experience, and what the user wants. As some of our core team members say, JHipster should be accessible to newcomers, more confirmed devs and seasoned devs alike.
As we understand the newcomers and confirmed devs, their goal is learning and have everything right away so they can learn how things are done, why etc. They may prefer a "complete" app instead of something shallower, so they can start right away and learn on the way.
The more seasoned ones have a tendency to throw what some may consider the "bloated part" of JHipster, because the generated code doesn't suit their needs. They may even prefer a more empty app instead of the real big deal so as not to be slowed down.

Both needs aren't incompatible as long as our approach goes their way but right now it doesn't, because we aim at generating the "big thing" right away. Instead, we should start small and let the user add things as she/he wants (thanks to the generators), like controllers, models etc.
Rails does this, and it works.

> I would say this is one of the goals but not the absolute. Our goal is to help users develop their own applications. It's cool to be able to generate a "complete" application but it would be even better if this application contained just enough needed code for the users.
+1 

> I don't see how changing the approach of how the generator works internally goes against this goal. I definitely see a lot of value of being able to generate JHipster applications "step by step" like @pascalgrimaud is suggesting. Moreover, this is a sentiment that is widely shared among the team.
+1

Another issue is code quality. Our code isn't clean, and even I dreaded the moment when I had to change something, like templates (I've paused my contributions due to the lack of energy, but that's another topic).
This issue prevents us from gaining more contributors.
The quality of what we generate must be the same as how JHipster is internally coded.
This is an issue for another time, but there are a lot of things to consider here.

mshima:
> But I'm fine if we decide to not go in this way, and keep the current big workflow.
> So let's remove the init generator and close this pull request, if we don't fully agree or convinced by this approach.

The discussion here is not about going modular. But the default behavior of the generators.

What files should this command generate?
`mkdir my-app; jhipster spring-boot`

What files should each command generate?
`mkdir my-app; jhipster init; jhipster maven; jhipster spring-boot`

The suggest behavior is to only generate the current generator files.
I don't agree with this.

Why?
It will complicate the regular user interaction and let it more subject to errors.
The following commands will fail to generate a valid application by default and the user will open an issue here:
- `mkdir my-app: jhipster spring-boot`
- `mkdir my-app: jhipster init; jhipster spring-boot`

Upgrade path:
The user uses an application generated with`mkdir my-app: jhipster init; jhipster maven; jhipster spring-boot`
A generator-jhipster is released and the user wants to upgrade.
What's the upgrade path?
- `npm i generator-jhipster@latest; jhipster spring-boot`? can fail on removed/rename needles, can fail due to dependency conflicts and probably others.
- `npm i generator-jhipster@latest; jhipster init; jhipster maven; jhipster spring-boot`?
The only generator that will be the upgraded is the first one, after that, the `.yo-rc.json` will have the new jhipsterVersion so cleanup will fail.
- `npm i -g generator-jhipster@latest; jhipster upgrade`?

Each behavior can be applied using flags or commands.
What I am discussion is the default behavior. A more advanced user can use more advanced flags.

deepu105:
> If we regroup maven and gradle, then later if someone wants to add 'ant' for example, we need to maintain it.
> 
> If we do this for maven and gradle, we should do the same for security : jwt, oauth2, session, uaa.
> 
> Not sure you want to maintain all options...
> 
> Keeping stuff separated will force the stream lead to maintain the option.
> 
> Nothing against gradle, but I'm not a gradle user today. If I want to contribute and improve maven, I dont want to test and check if I broke gradle or not.
> 
> That's why I think sub generator should be simple and dedicated to only 1 option.

Lot of excellent points in this thread. Let me share my thoughts. I 100% have same opinion as @MathieuAA about our user base and user/dev experience. I also understand the needs of the core team and contributors. 

We have to understand something. Most of us in the core team are power users and we fall into the second category of users @MathieuAA mentioned, so for us hexagonal architecture and separate generator that are manually composed will be fine. But for novice users and junior devs it would be daunting and confusing. Let me give you an example. Maven and Angular is more used than Gradle and React not because JHipster users like it more than others. Its because those are the default options and a lot of newcomers just pick the default. If we switch the default today and look the the stats after a year, I'm 100% sure that the stats will flip (We can look at session vs JWT for example or Postgres vs Mysql)

IMO those are the kind of users who appreciate JHipster and made us successful (not the power users, most power users who are not part of this community shrugs of JHipster as bloated or whatever) so personally I would focus more on the user base that made us successful and who finds JHipster more useful even if it means the power users are not 100% happy.

Also since there was discussion of goals, from what I remember there was one unwritten goal that was the core of JHipster from the beginning (`We make everything work out of the box regardless of the options you choose`), apart from the below we stated from v0.x (https://www.jhipster.tech/documentation-archive/v0.17.0/)
> Our goal is to generate a complete and modern Web app

Ok so history apart, I'm 100% ok with adding more sub gens that are specialised but what I wouldn't want to sacrifice would be below
1. The full web app/microservice flow should still work like today for novice devs
2. Developer experience is not sacrificed. means options if added later should still work with rest of the app (unless option are marked incompatible with certain others)

Another issue I see with adding specialised sub gen is the dependency. We currently are able to generate a complex web app with all options working well because of the dependency we have between sub gens. If we try to cut that this is gonna cause two issues IMO
1. Duplication of template code
2. Duplication of generator logic
3. Fragmentation of options (This IMO is a bigger issue as it would start driving the project based on the stack that core team likes rather than what the user community wants, which means some options would never be maintained well and we end up becoming more and more opinionated on stack choices)

A solution that could be a good middle ground would be something like below

1. Define a strict project model that is based on the JDL (all sub gens and blueprints must respect this model to be compatible)
2. Write our own composability mechanism that is compatible with blueprints and current generators and one which can automatically pass dependencies and options between each other using the project model
3. Have individual options as sub-generators like pascal suggests but these subs should also declare their compatibility for example if maven sub can work with elasticsearch, it needs to declare that as these will be used by master template for full project generation)
4. Have master templates that only composes those sub-generators together (sometimes overwrites a file from a sub based on combined options if required)
5. With this we could move away from Yeoman as well(We will still use the underlying libs for prompt, diff, etc)

Tcharl:
## Needles

Agrees with @mshima we can't rely on needles: most of our users don't understand needles and usually remove or move them in the code. Also, a concurrent technology called Acceleo already tried this kind of trocks and just failed badly (to have been a user of this technology during years, I can give the feedback). In that area, I would recommend the 'fragment' part that is currently being implemented by @mshima,  which is the equivalent in Eclipse of an 'extension point' (without the contract, but we'll tackle this aspect later on).

## Chaining

I also agree that generators at a higher level in the chain of responsibility should trigger previous generators of the chain . I won't be able to explain to newcomer that they will have to type `jhipster init maven springboot postgres liquibase oauth2 hazelcast dto react` to get something working while `jhipster` (or `jhipster react`) won't work at all. Also how will they know that `auth` is after `database`? It's a chain-of-responsibility pattern between generator beans from the gang of four and the core maintainers should provide the chains that are guaranteed (`jhipster.init().andTrigger(maven()).andTrigger(springboot().andTrigger(mapper(MAPSTRUCT)))`). May be we can provide a way for the guys that are doing custom stuff to provide register their own one.

## Duplication

Let me insist on avoiding duplication: it's always hard to maintain and IMHO will put the whole platform in the danger zone: we won't be able to track which generators are maintained and which aren't thus leading to random error.
Let's take an example: what if we decide to change the entity.java.ejs `String` attribute by a charsequence (for performance): how long it will take to have the Postgres, Mysql, Mssql, Neo4j, Elastic, couchbase... maintainers to upgrade their generators (generator-liquibase-postgres, generator-flyway-postgres, ...)  to handle that change? Here, two things can happen: we'll either loose options, or end users will face some errors depending on their choice (too bad, you chosed oracle) that we won't be able to spot upfront and also we'll spend days to change each template (including many bugs due to small differences between templates). Same will apply on different layers for fluent, partialUpdate, side-by-side, records, new attribute kinds, search, filters, Etc. so it's not this sample is not the exception.

We should identify what is a `contract` (api/spi/default impl), for example generic prompt and what is the implementation.

If we take the example of the database:
- prompt.database.js: contract, common to everyone, containing a common options that are printed to the end user that are common to each build tools (example `reactivedriver: true`, which  is not so relevant but you get the idea)
- prompt.database.postgres.js for additional implementation-specific options (example: `jsonb: true`).
Then, a simple loop on file depending on the options chosen by the end user.

Sample for js implementation: common stuff should go to the `DatabaseGenerator extends BlueprintBaseGenerator` while `PostgresGenerator extends DatabaseGenerator`.
My conclusion on that: blueprint are not perfect, but we can work hard on providing extensibility conventions in order to ease the maintenance. It will also help us reducing the amount of duplication.
Also, as you noticed, it's common that we modify core stuff and overall platform architecture: the more the duplication, the less we'll be able to improve it.

pascalgrimaud:
@Tcharl : I never said that:

> I won't be able to explain to newcomer that they will have to type jhipster init maven springboot postgres liquibase oauth2 hazelcast dto react to get something working while jhipster (or jhipster react) won't work at all

You'll be able to use `jhipster`, answer all questions and it will work like today. This workflow won't change

deepu105:
@pascalgrimaud would you be ok to consider this (its similar to what @mshima ) suggested

We split the generator like this into 3 layers
```
|_ Main generator (jhipter)
   |_ sub generator by layer  (init, build-tool, auth, database, client and so on) this layer will provide option prompts, hold any common JS logic in a util file that can be called by the generator below and any common template file that can be called by generator below
       |_ sub generator by tech (maven, gradle, mysql, cassandra, angular, react and so on) this layer will provide template and logic specific to that tech and all common logic/template will be just called here from above layer
```

Once we have this split, we can use the generator it in 3 ways
1. `jhipter` : same as today the main gen will compose sub-gen-by-layer. Each layer will compose the tech layes required based on options chosen
2. `jhipster add build-tool maven`: this will call the build-tool layer which inturn will compose the maven generator adding maven to existing app
3. `jhipster jdl { all custom layer stuff via JDL }`: This is for power users as you can compose different layers together by defining it as JDL and hence order etc wont be an issue

I believe this approach can provide what @pascalgrimaud wants and still kepp existing flow and keep duplication to a minimum

pascalgrimaud:
I never said I'm against this approach.

I'm ok with your proposition @deepu105 about point 2 and 3 (point 1 already exists).

The problem is: I don't know how to do it. It's too complex for me.
The only thing I know to do is a simple sub generator, with config, prompt, write.

That's why I have difficulty to take care of @mshima's comment. So if it's already too complex for me, I can't imagine for new contributors :-)

But I'm fine, I'm letting guys better than me to do it

deepu105:
Well new contributors rarely add new sub generators (honestly I wouldnt even want that, we already have enough options) so I don't think its that much of a problem. 

So for this I would suggest to wait and first split the second layer (means you split the entire build tool with maven and gradle part into a sub gen first and then split maven and gradle from there) may be @mshima or myself can do the first part

pascalgrimaud:
so @mshima : should we merge this first ?

deepu105:
@pascalgrimaud @mshima I thibnk we should do the second layer split before merging this else it would be a lot of conflicts and stuff IMO

deepu105:
Lets see if everyone agrees to the compromise, if so I suggest we create few tickets with bounties on them and the pick what you like based on interest. For example for this particular one I would do 3 tickets with 200$ on each
1. Split build tools into a sub-gen
2. Split maven out from build-tools
3. Split gradle out from build-tools

Tcharl:
@deepu105 do you think these three layers will be enough?
Won't we need one day or another `Postgresgenerator extends sqlgenerator extends databasegenerator`? It may be overengineered but I'm not able to answer this question.
I think we already have the case from what you provided: `KnativeGen extends KubernetesGen extends DockerGen`. Is it worth it to think about an unlimited generator-inheritance tree?

deepu105:
> @deepu105 do you think these three layers will be enough?
> Won't we need one day or another `Postgresgenerator extends sqlgenerator extends databasegenerator`? It may be overengineered but I'm not able to answer this question.
> I think we already have the case from what you provided: `KnativeGen extends KubernetesGen extends DockerGen`. Is it worth it to think about an unlimited generator-inheritance tree?

IMO that is different issue as the kubernetes one is extending for reusability which is more of a technical inheritance IMO. To me layers are based on logical grouping of the stack like build-tool, auth, db, client etc. Of course a specific 3rd layer item can have any number of inheritance and it would still work as long as you dont intent to call anything below the 3rd layer and IMO we should limit at this to avoid over engineering and IMO we dont want to be overly flexible as well as it will be causing more issues than the benefit

mshima:
IMO we should merge as it is now.
We should adjust later if needed.

> @pascalgrimaud would you be ok to consider this (its similar to what @mshima ) suggested
> 
> We split the generator like this into 3 layers
> 
> ```
> |_ Main generator (jhipter)
>    |_ sub generator by layer  (init, build-tool, auth, database, client and so on) this layer will provide option prompts, hold any common JS logic in a util file that can be called by the generator below and any common template file that can be called by generator below
>        |_ sub generator by tech (maven, gradle, mysql, cassandra, angular, react and so on) this layer will provide template and logic specific to that tech and all common logic/template will be just called here from above layer
> ```

This is what I mean by specialized generators and integration generators at https://github.com/jhipster/generator-jhipster/pull/15540#issuecomment-873470384.

> Once we have this split, we can use the generator it in 3 ways
> 
> 1. `jhipter` : same as today the main gen will compose sub-gen-by-layer. Each layer will compose the tech layers required based on options chosen
> 2. `jhipster add build-tool maven`: this will call the build-tool layer which in turn will compose the maven generator adding maven to existing app

This is what I mean by `jhipster add maven` at https://github.com/jhipster/generator-jhipster/pull/15540#issuecomment-874940139.

Glad you managed to make this simple at this point.

pascalgrimaud:
@deepu105 :  I don't think there can be conflict, as this generator is totally new

@mshima : if you think it's ready, then go ahead and let's merge this, so someone can start build tool

I'd like to have : init, build tool, maven for the next release. With that, we can code really simple Java libs
I wonder if we should rename maven to maven-java, as we could have maven-kotlin too

generators/maven/templates/.mvn/wrapper/maven-wrapper.properties
DanielFran: @pascalgrimaud I commited so it uses same version as in generator

generators/maven/templates/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+            });
+        }
+        URL website = new URL(urlString);
+        ReadableByteChannel rbc;
+        rbc = Channels.newChannel(website.openStream());
+        FileOutputStream fos = new FileOutputStream(destination);
+        fos.getChannel().transferFrom(rbc, 0, Long.MAX_VALUE);
+        fos.close();
+        rbc.close();
```
MathieuAA, line 114: we can use the try-with-resources syntax here, instead of manually closing and, but here it's missing, handling possible errors

test/generator-base.spec.js
```diff
@@ -350,7 +350,7 @@ describe('Generator Base', () => {
           jhipsterTemplatesFolders: [path.join(fixturesPath, 'templates', 'specific'), path.join(fixturesPath, 'templates', 'common')],
         });
       });
-      describe('exiting file in templates/specific and templates/common folders', () => {
+      describe('existing file in templates/specific and templates/common folders', () => {
```
MathieuAA, line 353: good catch, thanks :)

mshima:
Let's move on. I can take care of the build-tool generator.

PR #15618, opened 2021-07-09
Improve composing of new generators and tests.

mshima:
Many methods must be switched from sync to async in other to support ESM blueprints.
- Deprecate methods that should be removed for jhipster 8
- Implement async composeWithBlueprints to replace sync instantiateBlueprints.
- Fixes new generators with sbs blueprints.
- Implement dependsOnJHipster alternative to composeWithJHipster.
The first queues the composed generator before current one, the later queues after current one.
- Add common --reproducible option, in order to make non-reproducible options easier to test.
- Add common --add option. Should be used to compose with a generator at current project.
`jhipster add` command shortcut is planned.
- Implement common tests to test options, prompts and blueprint support.

Related to https://github.com/jhipster/generator-jhipster/issues/15490

---


generators/project-name/config.js
```diff
@@ -29,4 +35,4 @@ const defaultConfig = {
   ...requiredConfig,
 };
 
-module.exports = { requiredConfig, defaultConfig };
+module.exports = { requiredConfig, defaultConfig, reproducibleConfig };
```
Tcharl, line 38: idempotentConfig? defaultConfig? subGeneratorConfig?

mshima: - requiredConfig needs to be written to .yo-rc.json.
- defaultConfig is not written to .yo-rc.json to keep it clean.
- reproducibleConfig is for testing purpose.

mshima: Should requiredConfig be merged into defaultConfig?
What names?

MathieuAA: Guys, instead of naming them this way, we should be naming them by their actual function, not just their properties: how should each variable be used? For example, is it for tests? "testConfig" can be used, or "configForTests".

mshima: Not every test should use it.
reproducibleConfigForTests would be better.

MathieuAA: That'd be a better name indeed!

PR #15867, opened 2021-08-05
use contants in jdl/validators folder

renanfranca:
contributes to #14235

---

### DIFF
--- a/jdl/validators/jdl-with-application-validator.js
+++ b/jdl/validators/jdl-with-application-validator.js
@@ -35,6 +35,10 @@ const { isReservedFieldName } = require('../jhipster/reserved-keywords');
 const { isReservedTableName } = require('../jhipster/reserved-keywords');
 const { isReservedPaginationWords } = require('../jhipster/reserved-keywords');
 
+const USER = 'user';
+const AUTHORITY = 'authority';
+const builtInEntities = new Set([USER, AUTHORITY]);
+
 module.exports = {
   createValidator,
 };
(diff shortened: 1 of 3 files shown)


mshima (commented):
Please extract builtInEntities into a common place, maybe @Tcharl has a suggestion where to put or if it already exists..

jdl/validators/jdl-with-application-validator.js
```diff
@@ -107,7 +111,7 @@ function createValidator(jdlObject, logger = console, options = {}) {
   function checkForFieldErrors(entityName, jdlFields, jdlApplication) {
     const validator = new FieldValidator();
     const filtering =
-      jdlApplication.getConfigurationOptionValue('databaseType') === 'sql' &&
+      jdlApplication.getConfigurationOptionValue('databaseType') === DatabaseTypes.SQL &&
```
Tcharl, line 114: Destructuring like everywhere els?

jdl/validators/jdl-without-application-validator.js
```diff
@@ -202,7 +206,7 @@ function checkForAbsentEntities({ jdlRelationship, doesEntityExist, skippedUserM
   }
 }
 function isUserManagementEntity(entityName) {
-  return entityName.toLowerCase() === 'user' || entityName.toLowerCase() === 'authority';
+  return builtInEntities.has(entityName.toLowerCase());
```
Tcharl, line 209: For now this one works, but semantically speaking this change is a little bit risky: WDYT?

Tcharl:
for builtInEntities, they usually should go in jdl/jhipster/built-in-entities.js, shouldn't they?

jdl/validators/jdl-without-application-validator.js
renanfranca: I could undo this. I look up this approach from the Jhipster codebase. I don't if this is better or worse..

jdl/validators/jdl-with-application-validator.js
renanfranca: I don't understand. Should i use === SQL instead of DatabaseTypes.SQL?

Tcharl: Yes, totally
```
import { SQL } = require('../../jdl/jhipster/database-types');
...
jdlApplication.getConfigurationOptionValue('databaseType') === SQL && ...
```

PR #15918, opened 2021-08-11
Implement application template context at new generators.

mshima:
- Implement https://github.com/jhipster/generator-jhipster/pull/15883#discussion_r685798539.
- Implement async writeFiles to replace writeFilesToDisk. Supports custom context instead of this.
- Rename dependabotDependencies to nodeDependencies.
- Rename `prepareDerived*Properties` to `prepare*DerivedProperties`

---


generators/generator-base.js
```diff
@@ -2267,6 +2267,153 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
...
+   *
+   * @param {object} options
+   * @param {object} options.files - files to write
+   * @param {object} [options.context] - context for templates
+   * @param {string|string[]} [options.rootTemplatesPath] - path(s) to look for templates.
+   *        Single absolute path or relative path(s) between the templates folder and template path.
+   * @return {string[]}
+   */
+  async writeFiles(options) {
+    const { files, context = this } = options;
```
Tcharl, line 2281: is it possible to create a dedicated object for this?
example:
`utils/generation/model/templates-context.js`

generators/generator-base.js
```diff
@@ -2267,6 +2267,153 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
...
+   * write the given files using provided options.
+   *
+   * @param {object} options
+   * @param {object} options.files - files to write
+   * @param {object} [options.context] - context for templates
+   * @param {string|string[]} [options.rootTemplatesPath] - path(s) to look for templates.
+   *        Single absolute path or relative path(s) between the templates folder and template path.
+   * @return {string[]}
+   */
+  async writeFiles(options) {
```
Tcharl, line 2280: Can it be put in a dedicated files and called from there? 
for example:
* `utils/generation/processing/template-processor.js // for ejs`
* `utils/generation/processing/file-processor.js // for svg, png, or other static files` 
* `utils/generation/processing/templates-processor.js // calling the two above`

Ideally, if we could make the classes SOLID (including the S) we could get some new opportunity

generators/generator-base.js
```diff
@@ -2267,6 +2267,153 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
...
+            // Look for existing templates
+            const existingTemplates = rootTemplatesAbsolutePath
+              .map(rootPath => this.templatePath(rootPath, sourceFile))
+              .filter(templateFile => fs.existsSync(appendEjs ? `${templateFile}.ejs` : templateFile));
+
+            if (existingTemplates.length > 1) {
+              const moreThanOneMessage = `Multiples templates were found for file ${sourceFile}, using the first
+templates: ${JSON.stringify(existingTemplates, null, 2)}`;
+              if (existingTemplates.length > 2) {
+                this.warning(`Possible blueprint conflict detected: ${moreThanOneMessage}`);
```
Tcharl, line 2380: what about introducing a logger?

generators/init/index.cjs
```diff
@@ -52,6 +51,9 @@ module.exports = class extends MixedChain {
 
     if (this.options.help) return;
 
+    // Application context for templates
+    this.application = {};
```
Tcharl, line 55: what about creating a dedicated object for this?
`utils/generation/model/contexts/application-context.js`
`application = { baseName: undefined, baseFolder; undefined, ...}`

Tcharl: or in each generator if you prefer to keep them loosely coupled

generators/java/index.cjs
```diff
@@ -54,6 +54,9 @@ module.exports = class extends MixedChain {
       this.registerChainOptions();
     }
 
+    // Application context for templates
+    this.application = {};
+
```
Tcharl, line 59: same remark for model element

generators/java/mixin.cjs
```diff
@@ -65,7 +68,7 @@ module.exports.mixin = parent =>
      * @param {any} config - config to load config from
      * @param {any} into - destination context to use default is context
      */
-    loadJavaConfig(config = this.jhipsterConfig, into = this) {
+    loadJavaConfig(into = this, config = this.jhipsterConfig) {
```
Tcharl, line 71: `config =defaults({}, this.jhipsterConfig, defaultConfig)`?

generators/project-name/mixin.cjs
```diff
@@ -58,7 +58,7 @@ module.exports.mixin = parent =>
      * @param {any} config - config to load config from
      * @param {any} into - destination context to use default is context
      */
-    loadProjectNameConfig(config = this.jhipsterConfig, into = this) {
+    loadProjectNameConfig(into = this, config = this.jhipsterConfig) {
       config = defaults({}, config, defaultConfig);
```
Tcharl, line 62: `config = defaults({}, this.jhipsterConfig, defaultConfig)`?

generators/spring-boot/index.cjs
```diff
@@ -46,6 +46,9 @@ module.exports = class extends MixedChain {
 
     if (this.options.help) return;
 
+    // Application context for templates
+    this.application = {};
```
Tcharl, line 50: same remark regarding model element

Tcharl (commented):
Code start to become crystal clear and very efficient!
Thank you very much for that huge improvement!

generators/generator-base.js
mshima: IMO this will make more complicated.
I will try to make writeFiles more linear.

generators/generator-base.js
mshima: This is just a parameter grouping, no need.

generators/generator-base.js
mshima: If you think so, yeoman already has a very simple logger, IMO we should keep using it.

generators/init/index.cjs
mshima: We are using javascript not typescript, I don't see the need for this.

generators/java/mixin.cjs
mshima: This is exactly what is happening.
Do you want to drop the parameter?
The parameter is need for workspaces.
```
apps = appConfigs.map(config => this.loadChainConfig({}, config));
```

mshima (commented):
If we want types and objects more strict we should move to typescript otherwise it’s too much development burden for almost no benefits.

generators/init/index.cjs
Tcharl: giving some guidance and a clear contract for anyone which want to know the attributes that are recognized by the generators.
But anyway as you mentioned, javascript is not so pojo oriented...

generators/java/mixin.cjs
Tcharl: this was just to remove the line 72

mshima: Unless we save every config to .yo-rc.json, this is not possible.

PR #15966, opened 2021-08-13
Constants in jhipster tests

### LINKED ISSUE #14235: Ease generator understanding

Tcharl:
closes #14235 : You know what? I'm haaaaaappy :-)

---


test/cypress.spec.js
```diff
@@ -36,7 +34,7 @@ describe('jhipster:cypress', () => {
   });
 
   describe('generate cypress with React client with OAuth2', () => {
-    const options = { testFrameworks: ['cypress'], clientFramework: REACT, authenticationType: 'oauth2' };
+    const options = { testFrameworks: ['cypress'], clientFramework: REACT, authenticationType: OAUTH2 };
```
DanielFran, line 37:
```suggestion
    const options = { testFrameworks: [CYPRESS], clientFramework: REACT, authenticationType: OAUTH2 };
```

test/cypress.spec.js
```diff
@@ -60,7 +58,7 @@ describe('jhipster:cypress', () => {
   });
 
   describe('generate cypress with Angular client with JWT', () => {
-    const options = { testFrameworks: ['cypress'], clientFramework: ANGULAR, authenticationType: 'jwt' };
+    const options = { testFrameworks: ['cypress'], clientFramework: ANGULAR_X, authenticationType: JWT };
```
DanielFran, line 61:
```suggestion
    const options = { testFrameworks: [CYPRESS], clientFramework: ANGULAR_X, authenticationType: JWT };
```

test/cypress.spec.js
```diff
@@ -85,7 +83,7 @@ describe('jhipster:cypress', () => {
   });
 
   describe('generate cypress with Angular client with OAuth2', () => {
-    const options = { testFrameworks: ['cypress'], clientFramework: ANGULAR, authenticationType: 'oauth2' };
+    const options = { testFrameworks: ['cypress'], clientFramework: ANGULAR_X, authenticationType: OAUTH2 };
```
DanielFran, line 86:
```suggestion
    const options = { testFrameworks: [CYPRESS], clientFramework: ANGULAR_X, authenticationType: OAUTH2 };
```

DanielFran (changes requested):

Tcharl:
Thank you Daniel!

Tcharl:
Are you okl with this one @DanielFran ?

DanielFran (approved):

PR #16076, opened 2021-08-28
Vue: Upgrade Typescript / Remove IE support to open file?

qmonmert:
Fix: https://github.com/jhipster/generator-jhipster/pull/16071

But **IE support** is removed, is it a good solution?

**msSaveOrOpenBlob** is deprecated https://developer.mozilla.org/en-US/docs/Web/API/Navigator/msSaveOrOpenBlob

image

---

### DIFF
--- a/generators/client/templates/vue/src/main/webapp/app/shared/data/data-utils.service.ts.ejs
+++ b/generators/client/templates/vue/src/main/webapp/app/shared/data/data-utils.service.ts.ejs
@@ -44,25 +44,18 @@ export default class JhiDataUtils extends Vue {
    * Method to open file
    */
   openFile(contentType, data) {
-    if (window.navigator && window.navigator.msSaveOrOpenBlob) {
-      // To support IE and Edge
-      const byteCharacters = atob(data);
-      const byteNumbers = new Array(byteCharacters.length);
-      for (let i = 0; i < byteCharacters.length; i++) {
-        byteNumbers[i] = byteCharacters.charCodeAt(i);
-      }
-      const byteArray = new Uint8Array(byteNumbers);
-      const blob = new Blob([byteArray], {
-        type: contentType
-      });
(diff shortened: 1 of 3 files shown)


CGarces:
Please take a look at #15644
On angular the "iframe" solution was not working with big images (2.8 MB)  (tested on chrome 91.0.4472.124 or Microsoft Edge 91.0.864.67 under windows 10, but works fine on Firefox) should fail also on VUE implementation, right?

[createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) is supported since IE 10 (2012).

qmonmert:
@CGarces I did it, big images now work on Chrome
cc @mshima

generators/client/templates/vue/src/main/webapp/app/shared/data/data-utils.service.ts.ejs
```diff
@@ -44,25 +44,20 @@ export default class JhiDataUtils extends Vue {
...
+      byteNumbers[i] = byteCharacters.charCodeAt(i);
     }
+    const byteArray = new Uint8Array(byteNumbers);
+    const blob = new Blob([byteArray], {
+      type: contentType
+    });
+    const objectURL = URL.createObjectURL(blob);
+    const win = window.open();
+    win.document.write(
+      '<img src=\'' + objectURL + '\'>'
```
CGarces, line 59: It only will works with images... 
why not?

```
const win = window.open(objectURL);
win!.onload = function() {
        URL.revokeObjectURL(objectURL);
}  
```

qmonmert: oops you're right 👍

mshima (approved):
LGTM.
Thanks @qmonmert

PR #16077, opened 2021-08-29
Fix and enable back Couchbase tests

tchlyah:
When trying to fix Couchbase tests, I discovered that since Spring Boot 2.5.4, the application doesn't use `JacksonAutoConfiguration#jacksonObjectMapper()` bean, which causes the application to fail since the modules are not loaded anymore. Adding manually the bean fixed the problem.

---

### DIFF
--- a/.github/workflows/angular.yml
+++ b/.github/workflows/angular.yml
@@ -70,8 +70,8 @@ jobs:
           - ngx-mariadb-oauth2-infinispan
           - ngx-mongodb-kafka-cucumber
           - ngx-h2mem-ws-nol2
-          #- ngx-couchbase
-          #- ngx-gradle-couchbase-search
+          - ngx-couchbase
+          - ngx-gradle-couchbase-search
           - ngx-gradle-fr
           - ngx-gradle-mysql-es-noi18n-mapsid
           - ngx-gradle-mariadb-oauth2-infinispan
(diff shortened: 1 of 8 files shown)


generators/server/templates/src/main/java/package/config/JacksonConfiguration.java.ejs
```diff
@@ -71,4 +72,13 @@ public class JacksonConfiguration {
...
     }
+
+    /**
+     * Force create default Spring Boot jackson mapper to load custom modules, broken since Spring Boot 2.5.4
+     */
+    @Bean
+    @Primary
+    ObjectMapper jacksonObjectMapper(Jackson2ObjectMapperBuilder builder) {
+        return builder.createXmlMapper(false).build();
+    }
```
mshima, line 83: I couldn't see anything in the changelog and in the source code that indicates this is required.
Is this a couchbase only config?

tchlyah: I don't know, I didn't have time to test other configs. But I suppose that there is some new configuration that overrides SB default Object Mapper. There is a lot of issues that talking about this, for example: https://github.com/spring-projects/spring-boot/issues/26859

mshima: Can we drop to try to find the cause?
I think this breaks the backend to match a broken test.
Probably this https://github.com/jhipster/generator-jhipster/blob/3a2e16cab39dad196cbaebf52fe4d48d0f6f9c77/generators/server/templates/src/test/java/package/web/rest/TestUtil.java.ejs#L67 should use a injected ObjectMapper instead of a created one.
```suggestion
```

tchlyah:
No, it’s not just tests, it fails also on run. You can test locally without
it if you have time. I’ll try to dig when I have time. I prefer to work
first on reactive part.

Le lun. 30 août 2021 à 18:24, Marcelo Shima ***@***.***> a
écrit :

generators/server/templates/src/main/java/package/config/JacksonConfiguration.java.ejs
tchlyah: No, it’s not just tests, it fails also on run. You can test locally without
it if you have time. I’ll try to dig when I have time. I prefer to work
first on reactive part.

mshima: Can you move this change to CouchbaseConfig so we can merge it?
Jackson doesn’t looks broken at others databases.

tchlyah: I found the bug, I created an issue about it: https://github.com/spring-projects/spring-data-couchbase/issues/1209.

I didn't find a method to disable this bean, so the only way is to keep our `jacksonObjectMapper`. Do you want me to move it to `DatabaseConfiguration` for now ?

mshima: Great find.
Yes please, we should only add this where it is required.

mshima (approved):
LGTM

PR #16192, opened 2021-09-07
Authority check in SecurityUtils support array parameter

iconben:
To provide some handy methods in scenario:   
- want to check if a user has any authority in a list of authorities
- want to check if a user has none of authority in a list of authorities

---

### DIFF
--- a/generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
@@ -35,19 +35,23 @@ import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
 <%_ } _%>
-<%_ if (reactive) { _%>
-import reactor.core.publisher.Mono;
-<%_ } _%>
-
 <%_ if (authenticationTypeOauth2) { _%>
 import java.util.*;
 import java.util.stream.Collectors;
-<%_ } _%>
-<%_ if (!reactive) { _%>
-  <%_ if (!authenticationTypeOauth2) { _%>
-import java.util.Optional;
-  <%_ } _%>
+  <%_ if (reactive) { _%>
+import reactor.core.publisher.Mono;
+  <%_ } else { _%>
 import java.util.stream.Stream;
+  <%_ } _%>
+<%_ } else { _%>
+  <%_ if (reactive) { _%>
+import reactor.core.publisher.Mono;
+import java.util.Arrays;
+  <%_ } else { _%>
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.stream.Stream;
+  <%_ } _%>
 <%_ } _%>
 
 /**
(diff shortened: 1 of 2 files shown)


iconben:
Hi @mshima, I saw some checks failed due to network error, what should I do?

The error:
```
Could not download Saxon-HE-9.9.1-7.jar (net.sf.saxon:Saxon-HE:9.9.1-7)
      > Could not get resource 'https://plugins.gradle.org/m2/net/sf/saxon/Saxon-HE/9.9.1-7/Saxon-HE-9.9.1-7.jar'.
```

vishal423:
Since these are utility methods and are currently not used anywhere in code, I suggest moving this to server-side library code in  https://github.com/jhipster/jhipster-bom project.

iconben:
> Since these are utility methods and are currently not used anywhere in code, I suggest moving this to server-side library code in https://github.com/jhipster/jhipster-bom project.

Should we merge this first, then discuss whether the whole SecurityUtils class should be moved to jhipster-bom ?

vishal423:
I will rather close this PR as it would be surplus code for applications. JHipster server-side library is the perfect place to add these utility methods. Also, don't forget to add appropriate Junit tests.

pascalgrimaud:
Not sure about that @vishal423 

The current SecurityUtils is already generated by generator-jhipster and is used to check this kind of thing.
For my current customer, I added some specific method like this one, because the generated one is not enough.

So the good place is here. Anyway, this will add some dead code as it won't be used any where. Am I correct @iconben ?

iconben:
> Not sure about that @vishal423 
> 
> The current SecurityUtils is already generated by generator-jhipster and is used to check this kind of thing.
> For my current customer, I added some specific method like this one, because the generated one is not enough.
> 
> So the good place is here. Anyway, this will add some dead code as it won't be used any where. Am I correct @iconben ?

Yes and no, I added tests as well so they are used... in tests.😀

mshima:
IMO just point `hasCurrentUserThisAuthority` to `hasCurrentUserAnyOfAuthorities`.

While `hasCurrentUserNoneOfAuthorities` can be replaced with `!hasCurrentUserAnyOfAuthorities`.

The 3 implementations can reuse a single one.

generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
```diff
@@ -140,27 +144,52 @@ public final class SecurityUtils {
...
-     * Checks if the current user has a specific authority.
+     * Checks if the current user has any of the authorities.
      *
-     * @param authority the authority to check.
-     * @return true if the current user has the authority, false otherwise.
+     * @param authorities the authorities to check.
+     * @return true if the current user has any of the authorities, false otherwise.
      */
-    public static <% if (!reactive) { %>boolean<% } else { %>Mono<Boolean><% } %> hasCurrentUserThisAuthority(String authority) {
+    public static <% if (!reactive) { %>boolean<% } else { %>Mono<Boolean><% } %>  hasCurrentUserAnyOfAuthorities(String... authorities) {
```
DanielFran, line 152:
```suggestion
    public static <% if (!reactive) { %>boolean<% } else { %>Mono<Boolean><% } %> hasCurrentUserAnyOfAuthorities(String... authorities) {
```

generators/server/templates/src/main/java/package/security/SecurityUtils.java.ejs
```diff
@@ -140,27 +144,52 @@ public final class SecurityUtils {
...
 <%_ } _%>
     }
 
+    /**
+     * Checks if the current user has none of the authorities.
+     *
+     * @param authorities the authorities to check.
+     * @return true if the current user has none of the authorities, false otherwise.
+     */
+    public static <% if (!reactive) { %>boolean<% } else { %>Mono<Boolean><% } %>  hasCurrentUserNoneOfAuthorities(String... authorities) {
```
DanielFran, line 175:
```suggestion
    public static <% if (!reactive) { %>boolean<% } else { %>Mono<Boolean><% } %> hasCurrentUserNoneOfAuthorities(String... authorities) {
```

DanielFran (changes requested):

DanielFran (approved):

DanielFran:
Thanks @iconben for the contribution

PR #16207, opened 2021-09-08
Improve OIDC user handling

egvimo:
This PR contains some minor improvements for OIDC user handling, which I experienced recently connecting to Microsofts ADFS.

- Check if user info is not null before extracting authorities from it
- Remove else branch for setting the email of a user; the `sub` does not contain an email address and leads to a validation exception on the user entity because of the `@Email` annotation
- If the user already exists in the database, add existing (additional) roles to the returned object; this enables to maintain some roles in the app itself

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -65,6 +65,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.security.core.GrantedAuthority;
   <%_ if (!applicationTypeMicroservice) { _%>
 import org.springframework.security.core.authority.mapping.GrantedAuthoritiesMapper;
+import org.springframework.security.oauth2.core.oidc.OidcUserInfo;
 import org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority;
   <%_ } _%>
 import java.util.*;
@@ -319,7 +320,10 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
                 // each scope as a GrantedAuthority, which we don't care about.
                 if (authority instanceof OidcUserAuthority) {
                     OidcUserAuthority oidcUserAuthority = (OidcUserAuthority) authority;
-                    mappedAuthorities.addAll(SecurityUtils.extractAuthorityFromClaims(oidcUserAuthority.getUserInfo().getClaims()));
+                    OidcUserInfo userInfo = oidcUserAuthority.getUserInfo();
+                    if (userInfo != null) {
+                        mappedAuthorities.addAll(SecurityUtils.extractAuthorityFromClaims(userInfo.getClaims()));
+                    }
                 }
             });
             return mappedAuthorities;
--- a/generators/server/templates/src/main/java/package/service/UserService.java.ejs
+++ b/generators/server/templates/src/main/java/package/service/UserService.java.ejs
@@ -911,7 +911,7 @@ public class UserService {
                     updateUser(user.getFirstName(), user.getLastName(), user.getEmail(),
                         user.getLangKey(), user.getImageUrl());
                 }
-                // no last updated info, blindly update
+            // no last updated info, blindly update
             } else {
                 log.debug("Updating user '{}' in local database", user.getLogin());
                 updateUser(user.getFirstName(), user.getLastName(), user.getEmail(),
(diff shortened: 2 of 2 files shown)


generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
```diff
@@ -319,7 +320,10 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
                 // each scope as a GrantedAuthority, which we don't care about.
                 if (authority instanceof OidcUserAuthority) {
                     OidcUserAuthority oidcUserAuthority = (OidcUserAuthority) authority;
-                    mappedAuthorities.addAll(SecurityUtils.extractAuthorityFromClaims(oidcUserAuthority.getUserInfo().getClaims()));
+                    OidcUserInfo userInfo = oidcUserAuthority.getUserInfo();
+                    if (userInfo != null) {
```
vishal423, line 324: In which scenario you should get null over here? Isn't the UserInfo endpoint a requirement of `OIDC`?

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -911,12 +911,14 @@ public class UserService {
...
                 }
-                // no last updated info, blindly update
+            // no last updated info, blindly update
             } else {
                 log.debug("Updating user '{}' in local database", user.getLogin());
                 updateUser(user.getFirstName(), user.getLastName(), user.getEmail(),
                     user.getLangKey(), user.getImageUrl());
             }
+            // Add existing authorities stored in database to user
+            user.getAuthorities().addAll(existingUser.get().getAuthorities());
```
vishal423, line 921: It doesn't look appropriate. IDP manages the Roles/Groups association to a user account. On removing Role/Group association on the IDP side, we should also remove that from our app, and this synchronization code is trying to achieve the same.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -1042,8 +1044,6 @@ public class UserService {
         }
         if (details.get("email") != null) {
             user.setEmail(((String) details.get("email")).toLowerCase());
-        } else {
```
vishal423, line 1045: It was added in https://github.com/jhipster/generator-jhipster/pull/9947 Can you confirm that flow is not affected?

generators/server/templates/src/main/java/package/service/UserService.java.ejs
egvimo: You're right. That part is probably to specific for our use case. I'll remove that.

generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
egvimo: Spring Security loads the user info endpoint only if `email` or `profile` scope is provided (and if some other conditions are met).
Like I wrote in the description, I've connected our App to our ADFS. Unfortunately the ADFS does only return `odic` as scope and therefore Spring Security skips the user info resulting in a NPE.

This does not break anything, just wrapping it in a check.

mraible: We send [`oidc,profile,email`](https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/resources/config/application.yml.ejs#L296) as scopes. Are you saying ADFS doesn't work with this? It sounds like they're not OIDC-compliant if this is the case.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
egvimo: In my opinion mapping the `sub` claim to the email is wrong. This [claim is an Identifier](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).

In ADFS it looks similar to a base64 encoded string. Since the `email` claim is not provided by default in ADFS this leads to the validation error (see description).

However I did not try that with Okta, so I can't confirm.

vishal423: Unless you are willing to verify the said flow, I shall prefer to keep it as is.

generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
egvimo: Unfortunately yes. The ADFS has several quirks regarding OIDC (which I also stumbled upon googling for a solution).

vishal423: IMO, this is a useless check.

PR #16212, opened 2021-09-08
Add gradle-modernizer-plugin to detect legacy/deprecated API usage

gzsombor:
This plugin is a wrapper around modernizer-maven-plugin, but works in Gradle.
 
---

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -45,6 +45,7 @@ plugins {
 <%_ } _%>
     id "org.sonarqube"
     id "io.spring.nohttp"
+    id "com.github.andygoossens.gradle-modernizer-plugin"
     //jhipster-needle-gradle-plugins - JHipster will add additional gradle plugins here
 }
 
@@ -126,6 +127,11 @@ test {
     // see https://github.com/jhipster/generator-jhipster/pull/2771 and https://github.com/jhipster/generator-jhipster/pull/4484
     // ignoreFailures true
     reports.html.enabled = false
+}
+
+modernizer {
+    failOnViolations = true
+    includeTestClasses = true
 }
 
 task integrationTest(type: Test) {
--- a/generators/server/templates/gradle.properties.ejs
+++ b/generators/server/templates/gradle.properties.ejs
@@ -74,6 +74,7 @@ openapiPluginVersion=5.2.1
 <%_ } _%>
 noHttpPluginVersion=0.0.8
 checkstyleVersion=8.45.1
+modernizerPluginVersion=1.6.0
 
 # jhipster-needle-gradle-property - JHipster will add additional properties here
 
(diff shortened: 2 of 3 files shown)


atomfrede (approved):
Will restart the build. Failing ones are not gradle based.

mshima:
I am just wondering if we dropped Java 8 support with maven change.

gzsombor:
The JAVA_VERSION is already set to 11 here: 
https://github.com/jhipster/generator-jhipster/blob/f2d29b607ee7d66deb7e00cf4c672ceb9f7a63e7/generators/java/constants.cjs#L22

So I guess, we've already required Java 11 to build and run

mshima:
Default is Java 11. But compatible with java 8.
https://github.com/jhipster/generator-jhipster/blob/ec1cd00ca77bb10a7b16334ecf69b76954022b99/generators/java/constants.cjs#L23

But I am neutral to the Java 8 drop.

gzsombor:
I don't thinks JAVA_COMPATIBLE_VERSIONS is used anywhere, so it's not relevant as we set maven-compiler-plugin/target to 11, so you cant compile your app on Java 8 (the java compiler doesn't know the target version), and neither can you run on Java 8 - as the JVM doesn't understand the newer bytecode version.
 So, out of the box, a generated JHipster app already needs Java 11 at least, what has changed, is that it's a bit more work to make it compatible with Java 8.

atomfrede:
Agree, not sure if we have a hard policy. Java 11 by default (see https://github.com/jhipster/generator-jhipster/pull/12021) but it could be used with Java 8, at least that's was the intention. As Java 17 is around the corner I think we could drop the possibility to switch to java 8 without changing some code.

mshima:
> I don't thinks JAVA_COMPATIBLE_VERSIONS is used anywhere

https://github.com/jhipster/generator-jhipster/blob/d2df83756f7aa9c146c0c8fcd022626f5e063bdd/generators/server/templates/pom.xml.ejs#L1353
https://github.com/jhipster/generator-jhipster/blob/d2df83756f7aa9c146c0c8fcd022626f5e063bdd/generators/server/templates/build.gradle.ejs#L58

gzsombor:
> > I don't thinks JAVA_COMPATIBLE_VERSIONS is used anywhere
> 
> https://github.com/jhipster/generator-jhipster/blob/d2df83756f7aa9c146c0c8fcd022626f5e063bdd/generators/server/templates/pom.xml.ejs#L1353
> 
> https://github.com/jhipster/generator-jhipster/blob/d2df83756f7aa9c146c0c8fcd022626f5e063bdd/generators/server/templates/build.gradle.ejs#L58

Yes, that's what I'm talking about, the JAVA_COMPATIBLE_VERSIONS variable is not used anywhere, the maven-enforcer-config and the gradle config are hard coded

PR #16243, opened 2021-09-10
Update docker images version

DanielFran:
---


DanielFran:
I believe that the objective of snapshots is to compare with those that are created dynamically, so the purpose is to be fixed...
@mshima can you confirm?

mshima:
Versions are not important at snapshots, the structure are.
Tcharl did a huge refactoring and we would have a some or lot of breakages without the snapshots.
A flag to replace the version with a constant is acceptable, we already do that for timestamp and jwt secrets.

PR #16298, opened 2021-09-14
enabled build cache when gradle enterprise is selected

atomfrede:
Generate build cache confguration when gradle enterprise is selected.

---

### DIFF
--- a/generators/server/templates/settings.gradle.ejs
+++ b/generators/server/templates/settings.gradle.ejs
@@ -61,5 +61,14 @@ gradleEnterprise {
         publishAlways()
     }
 }
+
+buildCache {
+    local { enabled = true }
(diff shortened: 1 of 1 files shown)


generators/server/templates/settings.gradle.ejs
```diff
@@ -61,5 +61,14 @@ gradleEnterprise {
         publishAlways()
     }
 }
+
+buildCache {
```
mshima, line 65: Why not always enable it?

atomfrede: We can do so, but from my talk with Nelson from gradle not everyone wants to have it enabled by default, but of course we could do it (I don't see that much of a problem e.g. for our own use case)

PR #16612, opened 2021-10-08
Move some dependencies and plugins in their respective maven profile

### LINKED ISSUE #16485: Incompatibilities between some prod and dev databases

There are some incompatibility between some database:
Choosing H2 as dev database and mssql as Prod databse lead to an unability of liquibase to startup (sql error while executing create table).

fixing bug!

##### **Reproduce the error**

Choose H2 as dev db and Mssql as prod DB

##### **Suggest a Fix**

It's due to a dependency to `liquibase-mssql` in the pom in the main dependency section, whereas moving it in their relative profile (dev or prod) fix the thing. Not worth to mention that it's also a best practice :-)

I'll fix it when I'll have a minute

##### **JHipster Version(s)**

7.2.0

Tcharl:
fix #16485
---


Tcharl:
Hi experts,

Do you know why the test is failing?
It's a little bit strange to me, I think that this is a bug in the current behaviour of Jhipster

mshima:
It's interesting but I can see a feel problems:
- Liquibase is required for prod profile.
- We use dev profile to build docker image and then we use the image with prod profile
The docker file is compatible with dev and prod.

Tcharl:
Hello,

Any update on this one? someone to review?

Thank you very much in advance

mshima:
I think this is a good move.
But this is not a simple change and can be considered a breaking change.
Scripts and documentation must be updated:
https://www.jhipster.tech/development/#as-a-maven-project
Probably blueprints has they own scripts using mvnw directly.

I am in favor of making npm scripts the default for documentation and recommended way to run applications.
If we are going to do that, we should start migrating docs.
But the safe bet is to merge this at start of jhipster 8 cycle.

Tcharl:
Ending up with only few adjustment, without the need to update the doc, and solving @mshima mixed feelings

mshima:
Nice to see the requirement for `-Pdev` dropped.
This is quite big. I will take a look at maven soon, unless @DanielFran takes a look before me.
Would be nice if @atomfrede takes a look at gradle.

Tcharl:
Agreed, I'm not a gradle guy, so not totally sure it works like I think it works...

atomfrede:
I will have a look but didn't follow the issue.

generators/server/index.js
```diff
@@ -552,9 +552,9 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
             'backend:start': './mvnw -P-webapp',
             'java:jar': './mvnw -ntp verify -DskipTests --batch-mode',
             'java:war': './mvnw -ntp verify -DskipTests --batch-mode -Pwar',
-            'java:docker': './mvnw -ntp verify -DskipTests jib:dockerBuild',
+            'java:docker': './mvnw -ntp verify -DskipTests -Pprod jib:dockerBuild',
             'java:docker:arm64': 'npm run java:docker -- -Djib-maven-plugin.architecture=arm64',
-            'backend:unit:test': `./mvnw -ntp -P-webapp verify --batch-mode ${javaCommonLog} ${javaTestLog}`,
+            'backend:unit:test': `./mvnw -ntp  -P-webapp verify --batch-mode ${javaCommonLog} ${javaTestLog}`,
```
mshima, line 557:
```suggestion
            'backend:unit:test': `./mvnw -ntp -P-webapp verify --batch-mode ${javaCommonLog} ${javaTestLog}`,
```

generators/server/index.js
```diff
@@ -583,8 +583,8 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
           'java:war:prod': 'npm run java:war -- -Pprod',
           'java:docker:dev': 'npm run java:docker -- -Pdev,webapp',
           'java:docker:prod': 'npm run java:docker -- -Pprod',
-          'ci:backend:test': 'npm run backend:info && npm run backend:doc:test && npm run backend:nohttp:test && npm run backend:unit:test',
-          'ci:server:package': 'npm run java:$npm_package_config_packaging:$npm_package_config_default_environment',
```
mshima, line 587: Is this script removal intentional?

generators/server/index.js
```diff
@@ -583,8 +583,8 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
...
+          'ci:backend:test':
+            'npm run backend:info && npm run backend:doc:test && npm run backend:nohttp:test && npm run backend:unit:test -- -P$npm_package_config_default_environment',
```
mshima, line 587: Environment will be passed to the last command (test:unit) only.
Do we need to pass to others commands (info, nohttp) too?

generators/server/templates/pom.xml.ejs
```diff
@@ -37,7 +37,7 @@
             <url>https://repo.spring.io/milestone</url>
         </repository>
 <%_ } _%>
-<%_ if (JHIPSTER_DEPENDENCIES_VERSION.endsWith('-SNAPSHOT')) { _%>
+<%_ if (JHIPSTER_DEPENDENCIES_VERSION.endsWith('-SNAPSHOT')) {_%>
```
mshima, line 40:
```suggestion
<%_ if (JHIPSTER_DEPENDENCIES_VERSION.endsWith('-SNAPSHOT')) { _%>
```

mshima (commented):
Just saw that gradle changes are derived constants and ejs indentation only.
It’s ok. Reviewed it already.

generators/server/index.js
Tcharl: yes, this is the same than the one in normal mode (without `ci:` prefix)

generators/server/index.js
Tcharl: I don't think so (the previous steps are only to generate doc), but if someone can confirm I'll sleep a little bit better.
I'll test on my side if I get no answer

Tcharl: Tested on my side: It's ok for the doc and info without setting the maven profile: there's no change in the output at all

Tcharl:
Don't worry @atomfrede : I only used constants in the gradle descriptor, there's no impact regarding the way gradle works

Tcharl:
Shall we merge?

mshima:
Let's wait for 2.4.1 to be released.

pascalgrimaud:
Yes, I'm preparing the release, it should be good in few hours

mshima:
@Tcharl I think `backend:nohttp:test` script is executing client build.
At vue-default ci test, the backend step went from < 2min to 6-7min.

Tcharl:
Ouch, basically I don't get it: why compiling then building front (two steps) is more efficient than doing everything in a row?
I'll take a look of the process

PR #16863, opened 2021-10-28
Avoid using mavenLocal() in build.gradle?

nickshoe:

While reading the Gradle documentation to configure Gradle to use my private Nexus repository, I ran into this advice in the [official documentation](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local) regarding maven repositories:

> As a general advice, you should avoid adding `mavenLocal()` as a repository...

Existing code: 

https://github.com/jhipster/generator-jhipster/blob/1b76ae1b002eaf9498e9cbc190fb675de991624f/generators/server/templates/build.gradle.ejs#L283-L285

As stated in the doc's chapter ["Obtaining module metadata"](https://docs.gradle.org/current/userguide/dependency_resolution.html#obtaining_module_metadata):

> - Each repository is inspected in order.
>   - Depending on the type of repository, Gradle looks for metadata files describing the module (`.module`, `.pom` or `ivy.xml` file) or directly for artifact files.
>   - Modules that have a module metadata file (`.module`, `.pom` or `ivy.xml` file) are preferred over modules that have an artifact file only.
>   - Once a repository returns a _metadata_ result, following repositories are ignored.

(penultimate point)
> - All of the artifacts for the module are then requested from the same repository that was chosen in the process above.

(callout)
> The penultimate point above is what can make the integration with Maven Local problematic. As it is a cache for Maven, it will sometimes miss some artifacts of a given module. If Gradle is sourcing such a module from Maven Local, it will consider the missing artifacts to be missing altogether.

### DIFF
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -281,6 +281,8 @@ configurations {
 }
 
 repositories {
+    // Local maven repository is required for libraries built locally with maven like development jhipster-bom.
+    <%= !JHIPSTER_DEPENDENCIES_VERSION.includes('-SNAPSHOT') ? '// ' : '' %>mavenLocal()
     mavenCentral()
     <%_ if (SPRING_BOOT_VERSION.indexOf('M') > -1 || SPRING_BOOT_VERSION.indexOf('RC') > -1) { _%>
     maven { url 'https://repo.spring.io/milestone' }
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -281,7 +281,6 @@ configurations {
 }
 
 repositories {
-    mavenLocal()
     mavenCentral()
     <%_ if (SPRING_BOOT_VERSION.indexOf('M') > -1 || SPRING_BOOT_VERSION.indexOf('RC') > -1) { _%>
     maven { url 'https://repo.spring.io/milestone' }


mshima:
JHipster uses `mavenLocal()` for jhipster-bom at CI.
Would require us to start using snapshots (https://github.com/orgs/jhipster/packages?repo_name=jhipster-bom) instead.

nickshoe:
Hey @mshima I'm new to CI concepts and practices.. ':D
Can you point me to an explanation/tutorial of what jhipster-bom and snapshots are?
Let me know if I can do something to this PR.
Thanks

atomfrede:
Ah you are right @mshima . Didn't remember why we kept it.

generators/server/templates/build.gradle.ejs
```diff
@@ -281,7 +281,6 @@ configurations {
 }
 
 repositories {
-    mavenLocal()
     mavenCentral()
```
mshima, line 284: We can make it optional on jhipster-bom snapshot.
```suggestion
    // Local maven repository is required for libraries built locally with maven like development jhipster-bom.
    <%= !JHIPSTER_DEPENDENCIES_VERSION.includes('-SNAPSHOT') ? '// ' : '' %>mavenLocal()
    mavenCentral()
```

mshima:
@atomfrede what do you think about this now?

atomfrede (approved):

atomfrede:
Sorry missed the review request.

Issue #16318, opened 2021-09-17
[Warning] No dependency information available


tsendee0409:
I have recently installed Jhipster v7.2.0 and generated a simple microservice application. But some warnings appeared when it was running.

![image](https://user-images.githubusercontent.com/8222265/133712938-1d07d0d9-04b1-4294-bad3-807ae9690f0f.png)

`[WARNING] The POM for org.apache.logging.log4j:log4j-api-java9:zip:2.14.1 is missing, no dependency information available
Downloading from spring-libs-release: https://repo.spring.io/libs-release/org/springframework/roo/org.springframework.roo.annotations/1.2.3.RELEASE/org.springframework.roo.annotations-1.2.3.RELEASE.pom
Downloading from spring-libs-snapshot: https://repo.spring.io/libs-snapshot/org/springframework/roo/org.springframework.roo.annotations/1.2.3.RELEASE/org.springframework.roo.annotations-1.2.3.RELEASE.pom
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Could not build dependency tree Could not collect dependencies: mn.novelsoft:addressing:jar:0.0.1-SNAPSHOT ()`

![image](https://user-images.githubusercontent.com/8222265/133714555-012d8e5f-ba7e-45c5-bf81-8fecfe53dc82.png)

mraible:
I've see this too recently.

iconben:
I was upgrading a legacy app from jhipster 4.x to 7.2.0 and ran into this error too.

pmverma:
I have not looked deeper into it but I remember that this issue occurred when upgrading the maven enforcer plugin from `3.0.0-M3` to `3.0.0`
You can check by switching the versions.

iconben:
Yes, switching to 3.0.0-M3 of maven enforcer plugin works for now. Thanks for sharing @pmverma

rehmetjan:
in the POM.xml  add `-M3`
change this:
        <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
to: 
        <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
now it works!!

nomuna:
Ok, this is no longer a warning but an error/a build failure now... 

May be just add the working version for the enforcer in the template? I am using Jhipster 7.4.0 and getting this error and have to use the workaround to make the build work.

[...]

PR #16902, opened 2021-11-03
Fix error for expired oauth2 token in unit test

### LINKED ISSUE #16570: generated AuthorizationHeaderUtilTest   fails  on windows
##### **AuthorizationHeaderUtilTest  prodices failures on Windows due to timing issues**

While generating exipired token for test this code:

`            new OAuth2AccessToken(
                OAuth2AccessToken.TokenType.BEARER,
                "tokenVal",
                Instant.now(),
                accessTokenExpired ? Instant.now() : Instant.now().plus(Duration.ofMinutes(3))
            ),
`
may  produce the same instant in windows ( not in linux though )   and then an runtime exception comes from
spring boot code:

`ava.lang.IllegalArgumentException: expiresAt must be after issuedAt at org.springframework.util.Assert.isTrue(Assert.java:121) at org.springframework.security.oauth2.core.AbstractOAuth2Token.<init>(AbstractOAuth2Token.java:61) at org.springframework.security.oauth2.core.OAuth2AccessToken.<init>(OAuth2AccessToken.java:71) at org.springframework.security.oauth2.core.OAuth2AccessToken.<init>(OAuth2AccessToken.java:57) at de.vc.recorder.security.oauth2.AuthorizationHeaderUtilTest.getTestOAuth2AuthorizedClient(AuthorizationHeaderUtilTest.java:197) at`

Possible fix would be to use instant + 1ms and  waiting for one ms -   thus instant will be created, and also expired ( alternatively,  catch this exception and wait just in this case )

Some discussion about timer resolution:
https://stackoverflow.com/questions/20689055/java-8-instant-now-with-nanosecond-resolution

Having  failing unit tests  is generally bad and costs evaluation time. 

none found

crowne:
OAuth2AccessToken require expiryTime to be after issuedTime.
Also changed maven-enforcer-plugin.version to 3.0.0-M3,
because the central repo has an unresolved dependency issue with 3.0.0

Fix #16570

---

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -146,7 +146,7 @@
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-javadoc-plugin.version>3.3.1</maven-javadoc-plugin.version>
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
-        <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
+        <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
         <maven-failsafe-plugin.version>3.0.0-M5</maven-failsafe-plugin.version>
         <maven-idea-plugin.version>2.2.1</maven-idea-plugin.version>
         <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -146,7 +146,7 @@
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-javadoc-plugin.version>3.3.1</maven-javadoc-plugin.version>
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
-        <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
+        <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
         <maven-failsafe-plugin.version>3.0.0-M5</maven-failsafe-plugin.version>
         <maven-idea-plugin.version>2.2.1</maven-idea-plugin.version>
         <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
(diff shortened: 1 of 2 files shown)


mraible:
Is this one ready? If so, can you please move it out of draft mode.

generators/server/templates/pom.xml.ejs
```diff
@@ -146,7 +146,7 @@
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-javadoc-plugin.version>3.3.1</maven-javadoc-plugin.version>
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
-        <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
+        <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
```
mshima, line 149: Can you point a bug report to show the error.
Otherwise we should keep the release version.

crowne: Hi,
While I was experiencing problems with the build, I found the issue below
and the recommendation of switching to 3.0.0-M3 solved the build error.

[#16318](https://github.com/jhipster/generator-jhipster/issues/16318)

mshima: This may be resolved already.
There is nothing in the CI that indicates the problem.
What do you think @DanielFran

DanielFran: No, did not identify also any problem related with the versions of the maven plugin.

crowne: > This may be resolved already. There is nothing in the CI that indicates the problem. What do you think @DanielFran

Was it fixed in the POM?
Because the missing jar has not been resolved on maven central
mn.novelsoft:addressing:jar:0.0.1-SNAPSHOT

group mn doesn't exist
https://repo1.maven.org/maven2/mn/

crowne: I can't recreate the build error by setting the version back to version 3.0.0 in a generated project.
I also tested a new install on a separate Windows machine which never had jhipster installed previously and was not able to recreate the dependency issue with enforcer-3.0.0.
Perhaps some transitive dependency has been corrected on maven central.

So I will need to undo the POM change add another commit, then squash the commits, and push the change.
Or perhaps it might just be easier to redo a new PR and abandon this one.

mshima: No need.
```suggestion
        <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
```

crowne:
I'll redo the PR without reverting enforcer to 3.0.0-M3

crowne:
Replacement PR for this is #17004

PR #16974, opened 2021-11-10
Handle Auth0 claims when acting as an OAuth 2.0 resource server

mraible:
Add logic to handle `name` and `email` claims when acting as a resource server. Makes it so Ionic4J will work with Auth0. 

---

### DIFF
--- a/generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
@@ -34,6 +34,7 @@ import org.springframework.web.client.RestTemplate;
 import org.springframework.web.context.request.RequestContextHolder;
 import org.springframework.web.context.request.ServletRequestAttributes;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -91,20 +92,8 @@ public class CustomClaimConverter implements Converter<Map<String, Object>, Map<
                     String[] name = user.get("name").asText().split("\\s+");
                     if (name.length > 0) {
                         convertedClaims.put("given_name", name[0]);
-                        if (name.length == 2) {
-                            convertedClaims.put("family_name", name[1]);
-                        } else {
-                            StringBuilder lastName = new StringBuilder();
-                            int i = 1;
-                            while (i < name.length) {
-                                lastName.append(name[i]);
-                                i++;
-                                if (i < name.length) {
-                                    lastName.append(" ");
-                                }
-                            }
-                            convertedClaims.put("family_name", lastName.toString());
-                        }
+                        convertedClaims.put("family_name",
+                            String.join(" ", Arrays.copyOfRange(name, 1, name.length)));
                     }
                 }
                 if (user.has("groups")) {
--- a/generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
+++ b/generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
@@ -83,6 +83,15 @@ public class CustomClaimConverter implements Converter<Map<String, Object>, Map<
                 if (user.has("family_name")) {
                     convertedClaims.put("family_name", user.get("family_name").asText());
                 }
+                if (user.has("email")) {
+                    convertedClaims.put("email", user.get("email").asText());
+                }
+                // Allow full name in a name claim - happens with Auth0
+                if (user.has("name")) {
+                    String[] name = user.get("name").asText().split("\\s+");
+                    convertedClaims.put("given_name", name[0]);
+                    convertedClaims.put("family_name", name[1]);
+                }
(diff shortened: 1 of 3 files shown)


mraible:
Closing and reopening since tests seem to be hanging.

generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
```diff
@@ -83,6 +83,15 @@ public class CustomClaimConverter implements Converter<Map<String, Object>, Map<
                 if (user.has("family_name")) {
                     convertedClaims.put("family_name", user.get("family_name").asText());
                 }
+                if (user.has("email")) {
+                    convertedClaims.put("email", user.get("email").asText());
+                }
+                // Allow full name in a name claim - happens with Auth0
+                if (user.has("name")) {
+                    String[] name = user.get("name").asText().split("\\s+");
```
vishal423, line 91: By default, I guess for new users, `email` value is assigned to the `name` field on the `Auth0` side. If you update the `name` field value, then you can't be sure if that also contains the last name. I suggest including a length check before reading index-based values.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -1017,15 +1017,17 @@ public class UserService {
     private static <%= databaseTypeNo ? asDto('AdminUser') : asEntity('User') %> getUser(Map<String, Object> details) {
         <%= databaseTypeNo ? asDto('AdminUser') : asEntity('User') %> user = new <%= databaseTypeNo ? asDto('AdminUser') : asEntity('User') %>();
         Boolean activated = Boolean.TRUE;
+        String sub = String.valueOf(details.get("sub"));
+        String username = String.valueOf(details.get("preferred_username")).toLowerCase();
```
vishal423, line 1021: In the original code, we read this value in a null check enclosed block. In case of null, `toLowerCase()` invocation will throw NPE

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -1042,8 +1044,10 @@ public class UserService {
         }
         if (details.get("email") != null) {
             user.setEmail(((String) details.get("email")).toLowerCase());
+        } else if (sub.contains("|")) { // special handling for Auth0
```
vishal423, line 1047: Can we also check if username contains `@` ?

generators/server/templates/src/main/java/package/service/UserService.java.ejs
mraible: Should be fixed now.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
mraible: Added.

generators/server/templates/src/main/java/package/service/UserService.java.ejs
```diff
@@ -1042,8 +1047,11 @@ public class UserService {
         }
         if (details.get("email") != null) {
             user.setEmail(((String) details.get("email")).toLowerCase());
+        } else if (sub.contains("|") && (username != null && username.contains("@")) {
```
mraible, line 1050:
```suggestion
        } else if (sub.contains("|") && (username != null && username.contains("@"))) {
```

generators/server/templates/src/main/java/package/security/oauth2/CustomClaimConverter.java.ejs
```diff
@@ -83,6 +83,30 @@ public class CustomClaimConverter implements Converter<Map<String, Object>, Map<
...
+                            int i = 1;
+                            while (i < name.length) {
+                                lastName.append(name[i]);
+                                i++;
+                                if (i < name.length) {
+                                    lastName.append(" ");
+                                }
+                            }
+                            convertedClaims.put("family_name", lastName.toString());
+                        }
```
vishal423, line 107: for `length > 1`, you can probably write as `String.join(", ", Arrays.asList(name).subList(1, name.length).toArray(new String[0]))`

mraible: I'm a bit hesitant to change things now. However, I'd be happy to accept your changes if you modify this PR.

vishal423: I have updated comment with suggested replacement. If you agree, you can merge the change.

mraible: IntelliJ suggested an even shorter version with `Arrays.copyOfRange`. I just committed it.

vishal423: ok. I just wanted to avoid creation of a new array. Considering size of input array, I think copy should also be fine

Issue #17142, opened 2021-11-26
Elasticsearch is generated for all entities - not only for those  which are actually marked  with  it


ko5tik:
##### **Overview of the issue**

Only one entity in my project is marked with   "searchEngine": "elasticsearch" -   yet repositories  and code are generated for all those entities.   While it is not a show stopper,   it affects performance  and  misleads other developers. 

##### **Motivation for or Use Case**

I would like to have search repositories only for those classes  that are marked for it

##### **Related issues**

none  found

##### **Suggest a Fix**

improve logic  in generators/entity-server/files.js  ( and other places )   to generate search related stuff only if entity is marked for it.
Can do it myself  but there has to be general consensus   that it is worth doing  before I sink time into PR

##### **JHipster Version(s)**

7.4.0

##### **JHipster configuration**

elcome to the JHipster Information Sub-Generator

##### **JHipster Version(s)**

```
recorder@0.0.0 /home/ko5tik/customer/voicecollect/jhsandbox/recorder
└── generator-jhipster@7.3.0

```

##### **JHipster configuration, a `.yo-rc.json` file generated in the root folder**

<details>
<summary>.yo-rc.json file</summary>
<pre>
{
  "generator-jhipster": {
    "authenticationType": "oauth2",
    "cacheProvider": "ehcache",
    "serverPort": "8081",
    "serviceDiscoveryType": "eureka",
    "skipUserManagement": true,
    "baseName": "recorder",
    "buildTool": "gradle",
    "databaseType": "sql",
    "devDatabaseType": "h2Disk",
    "enableHibernateCache": true,
    "enableSwaggerCodegen": true,
    "enableTranslation": true,
    "jhiPrefix": "jhi",
    "languages": ["en"],
    "messageBroker": false,
    "prodDatabaseType": "postgresql",
    "searchEngine": "elasticsearch",
    "skipClient": true,
    "testFrameworks": [],
    "websocket": false,
    "packageName": "de.vc.recorder",
    "packageFolder": "de/vc/recorder",
    "applicationType": "microservice",
    "jhipsterVersion": "7.3.0",
    "clientPackageManager": "npm",
    "dtoSuffix": "DTO",
    "entitySuffix": "",
    "reactive": false,
    "skipServer": false,
    "skipCheckLengthOfIdentifier": false,
    "skipFakeData": false,
    "blueprints": [],
    "otherModules": [],
    "pages": [],
    "clientFramework": "angularX",
    "clientTheme": "none",
[...]

PR #17252, opened 2021-12-08
Microservice search artifacts shall be generated only when entity explicitely specifies then

ko5tik:
Search artifacts shall be generated only when entity  explicitely specifies then
See ticket:
Fix https://github.com/jhipster/generator-jhipster/issues/17142

---

- [x ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed


ko5tik:
Recreated pull request from other branch which was  out of sync.    Not sure where  and how document this -  IMHO  this creates behaviour which is consistent with existing documentation ( search engine not specified on entity means off )

generators/entity/index.js
```diff
@@ -1010,6 +1010,13 @@ class EntityGenerator extends BaseBlueprintGenerator {
     if (context.options.skipServer !== undefined) {
       this.entityConfig.skipServer = context.options.skipServer;
     }
+    // in case we are regenerating existing entities, and searchengine is not specified on
+    // entity file it shall be threated as if  search engine is set to false.
+    // also if search engine is set to false on the top level it shall not generate search artifacts for
+    // individual entities
+    if (this.context.entityExisted && (this.entityConfig.searchEngine === undefined || !this.jhipsterConfig.searchEngine)) {
+      this.entityConfig.searchEngine = false;
+    }
```
mshima, line 1019: With this condition a gateway without searchEngine will never expose search for a microservice entity.

mshima (changes requested):

generators/entity/index.js
ko5tik: Verified  it.   Will create unit tests and improve  behaviour.

ko5tik: Impemented and covered by unit test

ko5tik:
@mshima Please check  revised version.   Now there is a unit test for this use case :)

generators/entity/index.js
```diff
@@ -494,7 +494,15 @@ class EntityGenerator extends BaseBlueprintGenerator {
...
         // Update current context with config from file.
+        // copying everything will also overwrite search setting from the service itself
+        // which is wrong  for the microservice context.
+        this.context.disableElasticSearch =
+          !this.context.searchEngine && this.jhipsterConfig.applicationType === MICROSERVICE && this.context.entityExisted;
         Object.assign(this.context, this.entityStorage.getAll());
+        // in case search was deactiated on entity level (through  JSON file),   also disable  it in context
+        if (!this.entityConfig.searchEngine && this.context.entityExisted) {
+          this.context.searchEngine = false;
+        }
```
mshima, line 505: Replace current changes with:
```
       loadEntity() {
-        // Update current context with config from file.
+        loadRequiredConfigIntoEntity(this.context, this.jhipsterConfig);
+
         Object.assign(this.context, this.entityStorage.getAll());
+
         this.loadDerivedAppConfig(this.context);
+
+        // Entity searchEngine support should opt-in instead of opt-out for gateway/microservice.
+        // The entity definition should take precedence so frontend and backend are in sync.
+        // Doesn't apply to microfrontends.
+        if (
+          this.entityConfig.searchEngine === undefined &&
+          !this.context.microfrontend &&
+          (this.context.applicationTypeMicroservice || (this.entityConfig.microserviceName && this.context.applicationTypeGateway))
+        ) {
+          this.info(`searchEngine is missing in .jhipster/${this.entityConfig.name}.json, should opt-in for gateway/microservice entities`);
+          this.context.searchEngine = false;
+        }
+
         this.loadDerivedClientConfig(this.context);
         this.loadDerivedServerConfig(this.context);
         this.loadDerivedPlatformConfig(this.context);
-        loadRequiredConfigIntoEntity(this.context, this.jhipsterConfig);
+
```

generators/generator-base.js
```diff
@@ -2957,8 +2957,11 @@ templates: ${JSON.stringify(existingTemplates, null, 2)}`;
     dest.messageBrokerKafka = dest.messageBroker === KAFKA;
 
     dest.searchEngineCouchbase = dest.searchEngine === COUCHBASE;
-    dest.searchEngineElasticsearch = dest.searchEngine === ELASTICSEARCH;
-
+    // disable elastic search if it was deactivated on top level
+    dest.searchEngineElasticsearch = dest.searchEngine === ELASTICSEARCH && !dest.disableElasticSearch;
+    if (dest.searchEngine === ELASTICSEARCH && dest.disableElasticSearch) {
+      dest.searchEngine = false;
+    }
```
mshima, line 2964:
```suggestion
    dest.searchEngineElasticsearch = dest.searchEngine === ELASTICSEARCH;

```

mshima (changes requested):
@ko5tik
Ok, I think I understood the opt-in/opt-out problem here.
See my suggestions.

generators/entity/index.js
ko5tik: those changes break one of unit tests:

```
  1) JHipster generator for entity
       regeneration from json file
         microservice
           with dto, service, pagination
             shall not generate search specific artifacts because elastic search is false on top level:

      AssertionError [ERR_ASSERTION]: src/main/java/com/mycompany/myapp/repository/search/FooSearchRepository.java exists
      + expected - actual
```


Because
`Object.assign(this.context, this.entityStorage.getAll());`

Copies all the values from entity to context and thus  activates search engine on the top level implicitely.   Line 499  was meant to prevent this and save state

So question is:  shall entity opt in activate elastic search if  service has deactivated it?

ko5tik: Another issue,  check for this.context.entityExisted  is important to determine whether we are running interactively
( there are also unit tests for it ) - in this case entity shall take setting form the context

generators/entity/index.js
```diff
@@ -493,13 +493,27 @@ class EntityGenerator extends BaseBlueprintGenerator {
   _loading() {
     return {
       loadEntity() {
+        loadRequiredConfigIntoEntity(this.context, this.jhipsterConfig);
+        this.context.disableSearch =
+          !this.context.searchEngine && this.jhipsterConfig.applicationType === MICROSERVICE && this.context.entityExisted;
```
mshima, line 498:
```suggestion
```

In some cases the search engine option should be stored at entity level.
So gateways can pick it later.
And in some cases search engine should be forced to disabled at entity level.
See my proposed changes:

```
--- a/generators/entity/index.js
+++ b/generators/entity/index.js
@@ -38,6 +38,8 @@ const { prepareFieldForTemplates, fieldIsEnum } = require('../../utils/field');
 const { prepareRelationshipForTemplates } = require('../../utils/relationship');
 const { stringify } = require('../../utils');
 const { GATEWAY, MICROSERVICE } = require('../../jdl/jhipster/application-types');
+const { NO: CLIENT_FRAMEWORK_NO } = require('../../jdl/jhipster/client-framework-types');
+const { NO: SEARCH_ENGINE_NO } = require('../../jdl/jhipster/search-engine-types');
 const { CASSANDRA, COUCHBASE, MONGODB, NEO4J, ORACLE, SQL } = require('../../jdl/jhipster/database-types');
 const {
   GENERATOR_ENTITIES,
@@ -359,6 +361,28 @@ class EntityGenerator extends BaseBlueprintGenerator {
           throw new Error(validation);
         }

+        if (this.entityConfig.searchEngine === undefined) {
+          const backendOnlyMicroservice =
+            context.applicationType === MICROSERVICE && (!context.clientFramework || context.clientFramework === CLIENT_FRAMEWORK_NO);
+          if (backendOnlyMicroservice && context.searchEngine && context.searchEngine !== SEARCH_ENGINE_NO) {
+            // Propagate searchEngine to the entity, it's opt-in and should be propagated to gateway
+            this.entityConfig.searchEngine = true;
+          }
+        } else if (
+          (!context.searchEngine || context.searchEngine === SEARCH_ENGINE_NO) &&
+          (context.applicationType !== GATEWAY || !this.entityConfig.microserviceName)
+        ) {
+          if (this.entityConfig.searchEngine === SEARCH_ENGINE_NO) {
+            // Convert to boolean.
+            this.entityConfig.searchEngine = false;
+          }
+          // If search engine is disabled at application, and unless it's a gateway publishing a microservice entity.
+          if (this.entityConfig.searchEngine) {
+            this.warning('Search engine disabled at application, cannot be enabled at entity, disabling');
+            delete this.entityConfig.searchEngine;
+          }
+        }
+
         this.entityConfig.name = this.entityConfig.name || context.name;
         if (![SQL, MONGODB, COUCHBASE, NEO4J].includes(context.databaseType)) {
           this.entityConfig.pagination = NO_PAGINATION;
```

ko5tik: Implemented with one change - prooposed solution breaks test case when microservice has search engine disabled,  but entity has it enabled

mshima: @ko5tik it's a new review with more changes.

ko5tik: It also breaks this unit test case:
```

  2) JHipster generator for entity
       regeneration from json file
         microservice
           with dto, service, pagination
             shall not generate search specific artifacts because elastic search is false on top level:

```

Change  @497  is to satisfy this -   if   top search ist disable on  service level,   it has to be honored

ko5tik: Ok, found problem -  deleting from entityConfig is not enough.    It has to be set to false

ko5tik: changes are implemennted

mshima:
This PR has some behavior changes.
So I will try to explain:

Take the jdl sample:
https://github.com/jhipster/jdl-samples/blob/9d79a4d61b46e9eb32bd975aa10942df5705130d/microservices-blog-store.jh

Every application has elasticSearch enabled.
But the jdl is incorrect, it's missing the search option at entities:
```
search * with elasticsearch
```
The result is that the sample jdl will not generate search engine for entities at backend and frontend anymore.

Why current behavior is wrong?
Search frontend should be exposed when and only when the backend supports it.
At Microservices this needs to be enabled at the entity level, so the option is shared between gateway and microservice.
Otherwise the config can be conflicting.

This PR fixes to the correct behavior, but can be confusing and if merged must appear at jhipster changelog.

@mraible @pascalgrimaud

ko5tik:
IIRC  when you specify `search * with elasticsearch` all entitity json files will have search option enabled after importing jdl - so it woule be activated.

pascalgrimaud:
Ok thanks for the information @mshima 
We need to check quickly https://github.com/jhipster/jdl-samples but it should be OK

So if I'm correct, for the next release, it will be the 2nd break changes:
- springdoc
- JDL with search

mshima:
> Ok thanks for the information @mshima We need to check quickly https://github.com/jhipster/jdl-samples but it should be OK

👍

> 
> So if I'm correct, for the next release, it will be the 2nd break changes:
> 
> * springdoc
> * JDL with search

Possibly spring-boot 2.6 and angular 13.

mshima (approved):
LGTM

DanielFran:
Thanks @ko5tik for the contribution

PR #17376, opened 2021-12-22
Adjust server cleanup.

mshima:
https://github.com/jhipster/generator-jhipster/blob/b4ee21c6c26df280bde8d22486821d167dea4db3/generators/server/files.js#L1741
Follows tasks notation and returns an object and it's unused.
Convert to a plain function.

Fix variables.

---

### DIFF
--- a/generators/server/cleanup.js
+++ b/generators/server/cleanup.js
@@ -9,154 +9,141 @@
  * @param {string} testResourceDir - Test resources directory
  */
 function cleanupOldServerFiles(generator, javaDir, testDir, mainResourceDir, testResourceDir) {
-  return {
-    // eslint-disable-next-line consistent-return
-    cleanupCucumberTests() {
-      if (!this.cucumberTests) return undefined;
-      if (generator.isJhipsterVersionLessThan('7.4.1')) {
-        generator.removeFile(`${testResourceDir}cucumber.properties`);
-        generator.removeFile(`${testDir}../features/gitkeep`);
-        generator.removeFile(`${testDir}../features/user/user.feature`);
-      }
-    },
-    cleanupServer() {
-      if (generator.isJhipsterVersionLessThan('3.5.0')) {
-        generator.removeFile(`${javaDir}domain/util/JSR310DateTimeSerializer.java`);
-        generator.removeFile(`${javaDir}domain/util/JSR310LocalDateDeserializer.java`);
-      }
-      if (generator.isJhipsterVersionLessThan('3.6.0')) {
-        generator.removeFile(`${javaDir}config/HerokuDatabaseConfiguration.java`);
-      }
-      if (generator.isJhipsterVersionLessThan('3.10.0')) {
-        generator.removeFile(`${javaDir}config/CloudMongoDbConfiguration.java`);
-        generator.removeFile(`${javaDir}security/CustomAccessDeniedHandler.java`);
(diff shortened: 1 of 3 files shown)


generators/server/index.js
```diff
@@ -434,16 +445,16 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
     return {
```
Tcharl, line 445: why not moving this one in server/cleanup.js?

generators/server/cleanup.js
```diff
@@ -9,154 +9,141 @@
...
+  if (generator.isJhipsterVersionLessThan('3.6.0')) {
+    generator.removeFile(`${javaDir}config/HerokuDatabaseConfiguration.java`);
+  }
+  if (generator.isJhipsterVersionLessThan('3.10.0')) {
+    generator.removeFile(`${javaDir}config/CloudMongoDbConfiguration.java`);
+    generator.removeFile(`${javaDir}security/CustomAccessDeniedHandler.java`);
+    generator.removeFile(`${javaDir}web/filter/CsrfCookieGeneratorFilter.java`);
+  }
+  if (generator.isJhipsterVersionLessThan('3.12.0')) {
+    generator.removeFile(`${javaDir}config/hazelcast/HazelcastCacheRegionFactory.java`);
```
Tcharl, line 25: Could be nice to have multiple functions for each JH option: this one for example could test in the cacheManagerHazelcast has been activated.

It can help for generator componentization in the mid term WDYT?

mshima:
@Tcharl we should be careful with public api.
Moving from generators/cleanup to generators/server/cleanup breaks it and makes blueprints harder to maintain. See
https://github.com/jhipster/jhipster-kotlin/blob/25038425ea00562dddbd5ac932f8e95551666d39/generators/server/files.js#L21

For jhipster 8 we will define the exported api. It will be easier.
Couchbase is modularized at
https://github.com/jhipster/generator-jhipster/blob/a45da93bad22fc1d2859df311b5581a8dfb5e378/generators/server/files-couchbase.js#L126-L136
But it was broken and treated as a new implementation.

In the end generators/cleanup could be moved to a common generator and keep as legacy/deprecated.

Tcharl:
Totally understand your point and I've tendency to forget about blueprints, sorry for that. Regarding that context, strongly typed language switch would make sense (we could for example generate typings for the generator to easily check compatibility break).

Regarding the future, I would prefer to have a cleanup by generator (ideally in its own file) because to me it's part of the module itself. Setting cleanup in a common generator introduces strong coupling between the module  (I.e. couchbase) and the generator itself (because you'll have to run the common generator to manage your legacy)

mshima:
> Regarding that context, strongly typed language switch would make sense (we could for example generate typings for the generator to easily check compatibility break).

Sure, moving to typescript can be done.

> Regarding the future, I would prefer to have a cleanup by generator (ideally in its own file) because to me it's part of the module itself. Setting cleanup in a common generator introduces strong coupling between the module (I.e. couchbase) and the generator itself (because you'll have to run the common generator to manage your legacy)

Sure, just gave an example to reduce development, but can be done.

generators/server/index.js
mshima: With modularization, cleanup will be small. I don't see a need for a separated file.

DanielFran (approved):

Issue #17316, opened 2021-12-15
InsufficientAuthenticationException Needed To Be Handled


vw98075:
##### **Overview of the issue**

In an OAuth2 client app, InsufficientAuthenticationException will be threw after some idle time. The exception will be handled by ExceptionHandlerExceptionResolver after refreshing a browser. The followings are the related log messages for the scenario:
```
2021-12-13 10:50:15.381  WARN 68233 --- [  XNIO-1 task-1] o.z.problem.spring.common.AdviceTraits   : Unauthorized: Full authentication is required to access this resource
2021-12-13 10:50:15.427  WARN 68233 --- [  XNIO-1 task-1] .m.m.a.ExceptionHandlerExceptionResolver : Resolved [org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource]
```
It will be unhandled otherwise. Here is the exception track for the scenario:
```
2021-12-15 08:48:01.770 ERROR 6969 --- [  XNIO-1 task-2] o.z.problem.spring.common.AdviceTraits   : Internal Server Error

org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource
	at org.springframework.security.web.access.ExceptionTranslationFilter.handleAccessDeniedException(ExceptionTranslationFilter.java:194)
	at org.springframework.security.web.access.ExceptionTranslationFilter.handleSpringSecurityException(ExceptionTranslationFilter.java:173)
	at org.springframework.security.web.access.ExceptionTranslationFilter.doFilter(ExceptionTranslationFilter.java:142)
	at org.springframework.security.web.access.ExceptionTranslationFilter.doFilter(ExceptionTranslationFilter.java:115)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:336)
	at org.springframework.security.web.session.SessionManagementFilter.doFilter(SessionManagementFilter.java:126)
	at org.springframework.security.web.session.SessionManagementFilter.doFilter(SessionManagementFilter.java:81)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:336)
	at org.springframework.security.oauth2.client.web.OAuth2AuthorizationCodeGrantFilter.doFilterInternal(OAuth2AuthorizationCodeGrantFilter.java:168)
[...]

PR #17410, opened 2021-12-23
Handle expired session at vue.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/17316

---


generators/client/templates/vue/src/main/webapp/app/main.ts.ejs
```diff
@@ -130,3 +132,29 @@ new Vue({
 <%_ } _%>
   store,
 });
+
+setupAxiosInterceptors(
```
Tcharl, line 136: Why not in the config package?

mshima: Implemented here because it needs vue, store and loginService instances.
But it's possible to pass by parameters.

PR #17499, opened 2022-01-05
Add entity and i8n as core RFC.

mshima:
---

### DIFF
--- a/rfcs/4-jhipster-rfc-entity-i18n-as-core.md
+++ b/rfcs/4-jhipster-rfc-entity-i18n-as-core.md
@@ -1,0 +1,145 @@
+# JHipster-RFC-4: Entities and I18n as a core features.
+
+<!-- This is a RFC template based on the Rust RFC process but simplified: https://github.com/rust-lang/rfcs/ -->
+
+- Feature Name: Entities and I18n as a core features.
+- Start Date: 2022-01-05
+- Issue: [jhipster/generator-jhipster#0000](https://github.com/jhipster/generator-jhipster/0000)
+
+## Summary
+
+[summary]: #summary
+
+This RFC proposes to implement entities and i18n as JHipster core features instead of generators.
+
+## Motivation
+
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


rfcs/4-jhipster-rfc-entity-i18n-as-core.md
```diff
@@ -0,0 +1,145 @@
...
+
+Those generators and priorities will be replaced by entity/i18n focused priorities and internal methods to support the workflow. Some entity related generators may be kept for more specific purpose like prompts.
+
+Planed priorities includes:
+
+- `configuringEachEntity`: priority to manipulate entities config.
+- `preparingEachEntity`: priority to create derived properties for entities to be used by the templates.
+- `preparingEachEntityField`: priority to create derived properties for fields to be used by the templates.
+- `preparingEachEntityRelationship`: priority to create derived properties for relationships to be used by the templates.
+- `writingEachEntity`: priority to write entity related files.
```
Tcharl, line 52: May be a dumb question/suggestion, but if we'll use two round generation with needle, can't we also have:
- `writingEachFile`
- `writingEachFileAttribute`
- `writingEachFileMethods`
- `writeEachEntityFile`
- `writeEachEntityFileAttribute`
- `writeEachEntityFileRelationships`
- `writeEachEntityFileMethods`

mshima: I had second thought and changed the implementation.
See the new RFC that I dropped i18n (need to think more about it) https://github.com/jhipster/generator-jhipster/blob/4c12c3e7e06b6eead91b37d573b8d3032178250d/rfcs/4-jhipster-rfc-entity-i18n-as-core.md?plain=1#L91-L108.
Just realized it needs a rename.

Too many priorities makes the generator more complicated.
Needles are written at `postWriting`.
So maybe a `postWritingEntities`, but each entities, fields, relationship is just use a loop.

`preparing*` needs to be split because relationships depends on the other side entities and fields (primary key and label).

Issue #17520, opened 2022-01-08
Introduce more sophisticated ArchUnit architecture test


codecholeric:
##### **Overview of the feature request**

I was wondering if it would make sense to replace the predefined [ArchUnit test](https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/test/java/package/ArchTest.java.ejs) with one that is a little bit more sophisticated to give a better demo of what it can do. One thing that comes to mind would be to define a layered architecture on the technical components (as there is not too much domain to go on in a completely generic application). At the moment this could e.g. be such a test:

```
@AnalyzeClasses(packagesOf = ${root_class}.class)
class TechnicalStructureTest {

    @ArchTest
    static final ArchRule respects_technical_architecture_layers = layeredArchitecture()
        .layer("Config").definedBy("..config..")
        .layer("Web").definedBy("..web..")
        .layer("Security").definedBy("..security..")
        .layer("Service").definedBy("..service..")
        .layer("Persistence").definedBy("..repository..")
        .layer("Domain").definedBy("..domain..")

        .whereLayer("Web").mayNotBeAccessedByAnyLayer()
        .whereLayer("Service").mayOnlyBeAccessedByLayers("Web", "Config")
        .whereLayer("Persistence").mayOnlyBeAccessedByLayers("Service", "Security", "Web", "Config")
        .whereLayer("Domain").mayOnlyBeAccessedByLayers("Persistence", "Service", "Security", "Web", "Config");
}
```

(I just looked which allowed dependencies the test needs to be green, so I suppose this is the target architecture :wink:) 

##### **Motivation for or Use Case**

While forbidding dependencies `repository -> web` and `service -> web` obviously makes sense it still leaves a lot of holes. For example it would not complain if we create `root_package.SomeUtils` and then call `repository -> SomeUtils -> web`. Thus, I think in general it's a better approach to white-list what should be allowed instead of blacklist a specific sort of dependency. 

Also, if we have the dependency `archunit-junit5` on the classpath it makes sense to use the JUnit 5 support (`@AnalyzeClasses`, `@ArchTest`, ...) instead of importing the classes manually with `new ClassFileImporter()`. Because, for the current way of using ArchUnit the core dependency `archunit` would already be good enough.

If you think this makes sense I would be happy to create some PR, I just didn't want to go for it without discussion.

pascalgrimaud:
For your information, we have a complete Arch Unit Test for JHipster Lite project

[...]

PR #17521, opened 2022-01-08
replace generated `Archtest` by `TechnicalStructureTest`

codecholeric:
Since the generated application is modelled after a simple layered architecture we can provide a generated [ArchUnit](https://www.archunit.org/) test that uses `Architectures.layeredArchitecture()` as a nice example of a higher abstraction architecture rule. This will assert even more dependencies for correctness than before (for example no dependencies from repositories to services). Furthermore, by specifying an exact list of which layer is allowed to access which other layer, instead of forbidding specific dependencies from one layer to another, we remove one potential source of accidental violation. E.g., by forbidding `service -> web` we can still have accesses like `services -> some_utils_no_one_cares_about -> web`. If we specify that no other layer may access web (because all accesses are via REST, etc., and not from other Java code), we can make sure that there are no accidental dependencies through some unexpected packages.

 I've also replaced the manual import via `new ClassFileImporter()...importPackages(..)` by ArchUnit's JUnit 5 support. The dependency was already correct anyway, and it allows to write the test a little more concisely and automatically brings caching between multiple rules that import the same classes.

Related: #17520

---


codecholeric:
When I forgot to push `DoNotIncludeTests` on the arch test I noticed that a lot of tests have a dependency `domain -> web.rest.TestUtil.equalsVerifier(java.lang.Class)` which seems a little weird to me :thinking: 
The top level Javadoc says "Utility class for testing REST controllers" but then the Javadoc on `equalsVerifier` suddenly talks about the domain object, i.e. deals with objects from `domain`.
Would feel cleaner to me if this `equalsVerifier` method would reside in test utils within `domain` if it deals with domain objects :man_shrugging: But for now I've added back "ignore tests" so at least the arch test shouldn't fail anymore because of that...

codecholeric:
Hmm, I'm a little stuck I think. Because, the reactive configurations create the following dependency `repository.rowmapper.UserRowMapper -> service.ColumnConverter`. So this is in turn a cyclic constellation `service -> repository -> service`. Was it intended this way? It seems a little strange that logic to convert DB columns resides in the service layer? :thinking: 
Or I was wrong in the first place and this was never supposed to be layered this way, but then I still think a cycle between `repository` and `service` is weird in any case? (I would think one way or the other this should be directed on the static level?)
Anyway, I don't know exactly how I should go on, because I'm not sure what the original intentions were. Maybe somebody who knows why the code was designed like this originally can give me a hint? :pray:

pascalgrimaud:
From my point of view, your arch tests are correct.

If we have `service -> repository -> service`, it's a mistake and it means our ArchTests need to be improved to detect this kind of code.

mshima:
> Or I was wrong in the first place and this was never supposed to be layered this way, but then I still think a cycle between `repository` and `service` is weird in any case? (I would think one way or the other this should be directed on the static level?)
> Anyway, I don't know exactly how I should go on, because I'm not sure what the original intentions were. Maybe somebody who knows why the code was designed like this originally can give me a hint?

It’s open source, unless we have good tests in place and enforce code standards it’s hard to keep good code standards every time.

I think your PR looks correct, we should fix the implementation.
If I may, I will merge to a branch at main repository, so we can continue the work from there.

codecholeric:
@mshima Sure, feel free to do what you want! Does this mean you would take over the change from this PR and adjust the implementation then? In any case, just let me know if you want me to change something / take further action in any direction!

mshima:
Looks like there is only 1 failure. Should be easy to fix.

mshima:
https://github.com/jhipster/generator-jhipster/pull/17542 wasn’t enough, moving to a branch at main repository.

PR #17548, opened 2022-01-12
replace generated `Archtest` by `TechnicalStructureTest`

mshima:
Since the generated application is modelled after a simple layered architecture we can provide a generated ArchUnit test that uses `Architectures.layeredArchitecture()` as a nice example of a higher abstraction architecture rule. This will assert even more dependencies for correctness than before (for example no dependencies from repositories to services). Furthermore, by specifying an exact list of which layer is allowed to access which other layer, instead of forbidding specific dependencies from one layer to another, we remove one potential source of accidental violation. E.g., by forbidding `service -> web` we can still have accesses like `services -> some_utils_no_one_cares_about -> web`. If we specify that no other layer may access web (because all accesses are via REST, etc., and not from other Java code), we can make sure that there are no accidental dependencies through some unexpected packages.

 I've also replaced the manual import via `new ClassFileImporter()...importPackages(..)` by ArchUnit's JUnit 5 support. The dependency was already correct anyway, and it allows to write the test a little more concisely and automatically brings caching between multiple rules that import the same classes.

Resolves: #17520
Includes: https://github.com/jhipster/generator-jhipster/pull/17521

---


DanielFran (approved):

DanielFran:
@mshima It seems that for no database we have issues: https://github.com/hipster-labs/jhipster-daily-builds/runs/4891398745

PR #17552, opened 2022-01-13
Fix Node 16 deprecation error

mraible:
Fixes issue I saw today with Node v16.13.2 and Ionic for JHipster:

```
(node:51983) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./generators/" in the "exports" field 
module resolution of the package at /Users/mraible/dev/generator-jhipster-ionic/node_modules/generator-jhipster/package.json.
Update this package.json to use a subpath pattern like "./generators/*".
(Use `node --trace-deprecation ...` to show where the warning was created)
```

---


mraible:
This isn't really related, but I upgraded Ionic for JHipster to use the latest version of JHipster with Node.js 16 and there's a non-descript error message.

```
/home/runner/ionic4j/node_modules/generator-jhipster-ionic -> /opt/hostedtoolcache/node/14.18.2/x64/lib/node_modules/generator-jhipster-ionic -> /home/runner/work/generator-jhipster-ionic/generator-jhipster-ionic
INFO! The JDL is being parsed...
warn: An Entity name 'User' was used: 'User' is an entity created by default by JHipster. All relationships toward it will be kept but any attributes and relationships from it will be disregarded.
INFO! Found entities: Blog, Post, Tag.
INFO! The JDL has been successfully parsed.
INFO! Generating 3 entities.
Error jhipster-ionic:import-jdl /home/runner/backend/monolith-jwt.jdl --force 

Cannot read property 'name' of undefined
```

mshima:
I am seen lots for warnings like this developing the blueprint generator.
But I don't think it will work and probably will break blueprints.
`"./generators/": "./generators/"` is the old behavior like if we don't have exports field.

`"./generators/": "./generators/*"` is not covered by the manual https://nodejs.org/docs/latest-v16.x/api/packages.html#subpath-patterns.

But IFAIK `"./generators/*": "./generators/*"`, the conversion converts exactly from the left to the right. At the right, a extension is supposed to exist like:
`"./generators/*": "./generators/*/index.js"` that is ok for the generators but won't work for base classes like `"./generators/generator-base.js"`.

The only way is to expose every class manually, which is a lot of classes.
So unless we decide to do some breaking change regards blueprints, we will need rework the exports only at v8 as I plan to do.

mraible:
@mshima Should we close this issue?

mshima:
We need to test some blueprints with these changes.
I want to start v8 soon, so I don’t think this is a priority.

Issue #17488, opened 2022-01-04
Implement new interfaces for ElasticSearch Entity indexing


ko5tik:
##### **Overview of the feature request**

Current logic for entity indexing:

```java
    public RecordingDTO save(RecordingDTO recordingDTO) {
        log.debug("Request to save Recording : {}", recordingDTO);
        Recording recording = recordingMapper.toEntity(recordingDTO);
        recording = recordingRepository.save(recording);
        RecordingDTO result = recordingMapper.toDto(recording);
        recordingSearchRepository.save(recording);
        return result;
    }
```
Has a major flaw - it does not index  that is actually contained in database.   DTO coming from  REST layer may mot contain all the relations (which are managed by other entity)  and thus indexing result will not reflect actual data stored in database.

To overcome this situation,   indexing shall be performed over entity loaded from database in separate  transaction with sufficient isolation level to get real data contained in database.   This may ( shall ) be done asynchronously,   because it involves external rest call to elasticsearch service running in a separate JVM anyway. 

I propose adding 2 methods:

```java
    @Transactional(propagation =  Propagation.REQUIRES_NEW, readOnly = true, isolation = Isolation.READ_COMMITTED)
    public Recording indexRecording(Recording recording) {
        recordingRepository.findOneWithEagerRelationships(recording.getId()).ifPresent(recordingSearchRepository::save);
        return recording;
    }

    /**
     * asynch wrapper
     * @param toBeIndexed
     */
    @Async
    @Transactional(propagation =  Propagation.REQUIRES_NEW, readOnly = true, isolation = Isolation.READ_COMMITTED)
    public void asyncIndexRecording(Recording toBeIndexed) {
        indexRecording(toBeIndexed);
    }
```

And invoke them instead of calling search repository directly

     
##### **Motivation for or Use Case**

I would like index data  to be in synch with actual  data contained in database so search would be accurate. 

##### **Related issues or PR**

did not found

- [x ] Checking this box is mandatory (this is just to show you read everything)

ko5tik:
Actually I moved it out into separate service, so it does not collide with generated code when I have to regenerate - I am aware that special considerations shall be made when calling async methods and new transactions  on the same beans

ko5tik:
Additional improvement:  search repostory shall offer methods utilising  Query  ,  not only string query.

Tcharl:
Why do you want the exact same content from a sql db and an elastic db?

[...]

PR #17561, opened 2022-01-13
17488 indexing improvement

ko5tik:
---

This pull request improves querying and indexing of  entities -  #17488 

1. Additional query methods -   pass not only string but also real query objects
2.  Reindexing shall happen in separate transaction with proper isolation  - because reindexing incoming DTO will not always get all the data of the object. ( Relations managed on the other side are not transferred from frontend , and there could be changes in related objects )


generators/entity-server/templates/partials/save_template.ejs
```diff
@@ -46,18 +46,18 @@ if (isUsingMapsId) {
         <%= mapsIdAssoc.otherEntity.primaryKey.type %> <%= otherEntityName %>Id = <%= instanceName %>.get<%= mapsIdAssoc.relationshipNameCapitalized %>().get<%= primaryKey.nameCapitalized %>();
         <%= mapsIdRepoInstance %>.findById(<%= otherEntityName %>Id).ifPresent(<%= persistInstance %>::<%_ if (!fluentMethods) { _%>set<%= mapsIdAssoc.relationshipNameCapitalized %> <%_ } else { _%><%= mapsIdAssoc.relationshipName %><%_ } _%>);
     <%_ } _%>
-        <%= persistInstance %> = <%= entityInstance %>Repository.save(<%= persistInstance %>);
```
Tcharl, line 49: Save is really the keyword here: what about the rollback if there's something that fails afterwards?
We are loosing all the benefits of the Transactions here.

generators/entity-server/templates/src/main/java/package/repository/EntityRepository.java.ejs
```diff
@@ -69,6 +70,7 @@ import java.util.UUID;
 @SuppressWarnings("unused")
 <%_ } _%>
 @Repository
+@Primary
```
Tcharl, line 73: Not needed if I'm not wrong

generators/entity-server/templates/partials/save_template.ejs
ko5tik: Save and flush does not commit so transaction can still be rolled back.  And with just save  this data is not picked up while indexing later.  ( more about rationale in the other comment )

generators/entity-server/templates/src/main/java/package/repository/EntityRepository.java.ejs
ko5tik: When there is  another repository inheriting from this one ( say,  I need additional query methods etc ) ,   generated code can not be injected properly.  Alternative solution would be  explicit qualifier.

Tcharl: I think that end user would prefer to have their custom repositories picked up instead of the generated ones.
Also, as an alternative regarding custo,  take a look at what has been done on the reactive jdbc repositories

generators/entity-server/templates/partials/save_template.ejs
Tcharl: Always not a fan of it: stacking database queries is way better in term of perf.
Is it really mandatory?

generators/entity-server/templates/src/main/java/package/repository/EntityRepository.java.ejs
ko5tik: Sure,   I also like it this way.  But in case user derived from generated repository, there will be ambiguity  and  generated service can not be autowired properly -  this is a way to resolve  this ambiguity.

generators/entity-server/templates/partials/save_template.ejs
ko5tik: I like to ensure that indexing really gets complete updated database state -  because entity object here may not have all the relations initialized properly ( it was converted from DTO ).   That's why  I do indexing asynchronously,   and in separate transaction ( See EntitySearchRepository.java.ejs @ 164-171 ).   Save and flush ensures that READ_COMMITTED isolation level  kicks in even in asynchronous scenario.

generators/entity-server/templates/src/main/java/package/repository/EntityRepository.java.ejs
ko5tik: After some thinking about it - you are correct.  If user derives from the generated repository he can place @Primary on  hos own class and resolve ambiguity this way.

ko5tik:
Recreated this pull request as branch was too old:

https://github.com/jhipster/generator-jhipster/pull/17780

generators/entity-server/templates/partials/save_template.ejs
Tcharl: @ko5tik Sure :-).
The Thing that I'm not sure about is if it makes sense to persist the related elements in the ES database as well.
Let me try to elaborate:
- When you query an element, you basically just want that element and not necessarly its related ones.
- If you really want the siblings you could always 'findByIdWithEagerRelationship' from the database.

I'm not saying I'm against, both seems valid; The question being: do we want to have the exact same behaviour for search than the one we have with sql

ko5tik: Standard behaviour of elasticsearch would be to persist complete object tree (can be modified by annotations when necessary).    I my use case I need to be able to find entities by dependent objects ( say, customer conversations based on phone calls time of day )  -  thus  I need them indexed together.  And when  conversation object is updated and reindexed,  I will not lose related objects  from index - which is the actual case.

PS:  As this branch was too old and synch  brought in too many commits  I recreated it and raised a new pull request

https://github.com/jhipster/generator-jhipster/pull/17780

PR #17685, opened 2022-01-27
Use cassandra-test-container and kafka-testcontainers

Tcharl:
Use the official cassandra testcontainer module instead of generic module.
Use the official kafka testcontainer module instead of generic module.
Also unify extension location in tests (more to come).


generators/server/templates/src/test/java/package/config/TestContainersSpringContextCustomizerFactory.java.ejs
```diff
@@ -0,0 +1,60 @@
+package <%= packageName %>.config;
```
mshima, line 1: Split a file for Cassandra another for MongoDB, so we can move each to it's own generator later.

Tcharl: Not sure it will be good: there will be multiple combination if we use redis, elastic, couchbase or any dev database that won't be h2.

This is the first PR on the topic, but I plan to make many similar ones.

Tcharl:
Do you have an idea from where these OOM come from?

generators/server/templates/src/test/java/package/config/TestContainersSpringContextCustomizerFactory.java.ejs
```diff
@@ -0,0 +1,60 @@
...
+                .of("spring.data.mongodb.uri=" + databaseContainer.getReplicaSetUrl())
+<%_ } _%>
+<%_ if (databaseTypeCassandra) { _%>
+                .of("spring.data.cassandra.port=" +  databaseContainer.getMappedPort(CassandraContainer.CQL_PORT))
+                .and("spring.data.cassandra.contact-points=" +  databaseContainer.getHost())
+                .and("spring.data.cassandra.keyspace-name=" + CQLDataLoader.DEFAULT_KEYSPACE_NAME)
+                .and("spring.data.cassandra.local-datacenter=" + databaseContainer.getCluster().getMetadata().getAllHosts().iterator().next().getDatacenter())
+                .and("spring.data.cassandra.cluster-name=" + databaseContainer.getCluster().getMetadata().getClusterName())
+<%_ } _%>
+                .applyTo(context);
```
mshima, line 53: The original file had this:
```suggestion
                .applyTo(context.getEnvironment());
```

Tcharl: Yes, that was an improvement ^^

Tcharl:
Should be ok, but I don't know why it hickups...

Tcharl:
Stop and restart are too long on local machine, that's why I set the reuse to true by default.

I've just disabled reuse for CI: is it ok?

generators/server/templates/src/test/resources/testcontainers.properties.ejs
```diff
@@ -0,0 +1 @@
+testcontainers.reuse.enable=true
```
mshima, line 1: From what I've seem in docs, this won't work here.
It needs to be at the user home.

Tcharl: See https://www.testcontainers.org/features/configuration/

Tcharl:
Is it good to go? I hope that `testcontainairization` of redis, ES and Kafka will make that lights turn green again

generators/server/templates/src/test/resources/testcontainers.properties.ejs
mshima: This behavior should be documented.
I would prefer to disable by default.

As a jhipster developer I expect that my environment in left as it was, without a running container.

Is quite common to create a project, test and delete the project.

generators/server/index.js
```diff
@@ -598,7 +598,7 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
             'java:war': './mvnw -ntp verify -DskipTests --batch-mode -Pwar',
             'java:docker': './mvnw -ntp verify -DskipTests -Pprod jib:dockerBuild',
             'java:docker:arm64': 'npm run java:docker -- -Djib-maven-plugin.architecture=arm64',
-            'backend:unit:test': `./mvnw -ntp${excludeWebapp} verify --batch-mode ${javaCommonLog} ${javaTestLog}`,
+            'backend:test': `./mvnw -ntp${excludeWebapp} verify --batch-mode ${javaCommonLog} ${javaTestLog}`,
```
mshima, line 601: npm scripts should be stable to don’t break existing scripts.
A compatibility script should be added and dropped for v8.

mshima:
Maybe testcontainer (lack of) profile is ignored.
And maybe it’s starting various duplicate containers causing the out of memory.

generators/server/templates/src/test/java/package/config/CassandraTestContainerExtension.java.ejs
```diff
@@ -0,0 +1,132 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.CassandraContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+
+/**
+ * Base class for starting/stopping Cassandra during tests.
+ */
+public class CassandraTestContainerExtension implements BeforeAllCallback, AfterAllCallback {
```
mshima, line 40: Conditional on testcontainers profile?

generators/server/templates/src/test/java/package/config/MongoDbTestContainerExtension.java.ejs
```diff
@@ -1,31 +1,55 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.testcontainers.containers.MongoDBContainer;
-import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
 
-public class MongoDbTestContainerExtension implements BeforeAllCallback {
-    <%_ /* TODO, the smallest configuration that lead to only one error for webflux-gw-jwt was 0.6 RAM /0.8 SWAP/0.1 nanocore */ _%>
-    private static final ThreadLocal<MongoDBContainer> THREAD_CONTAINER = new ThreadLocal<>();
```
mshima, line 12: Conditional on testcontainers profile?

generators/server/templates/src/test/java/package/config/CassandraTestContainerExtension.java.ejs
Tcharl: I would prefer to drop the testcontainers profile and make it standard.
-> If the user chooses something that can be used with testcontainer (everything but h2) == testcontainer
-> Else depends on the maven profile (dev or prod)

WDYT?

Tcharl: And still, Cassandra testcontainer is a must have for testing stacks on cassandra

generators/server/templates/src/test/java/package/config/MongoDbTestContainerExtension.java.ejs
Tcharl: And still, Mongodb testcontainer is a must have for testing stacks on cassandra

generators/server/index.js
Tcharl: :'(. It's not only unit but integration tests too!
What about a 7.7.X? It's not breaking the blueprints API.

What's the guideline about Semantic versioning? I definitely don't get it.

To me, and according to semantic versioning spec:
- Major (user API incompatibility)=> Big change on the generated code (i.e. switching from service to modulith), removal of options or modification an option behavior
- Minor (developer API incompatibility) => Breaking blueprint API
- Patch => Everything else

mshima: > To me, and according to semantic versioning spec:

That's not what node/npm ecosystem uses, see https://docs.npmjs.com/about-semantic-versioning#incrementing-semantic-versions-in-published-packages.

We should break blueprints only at major versions.
And we usually don't release patch versions.

Tcharl: Ahah thanks for the pointer, I'm so old school I still followed the one of OSGi: http://docs.osgi.org/whitepaper/semantic-versioning/040-semantic-versions.html (which could have been nice in our context, if we weren't on js)

Still, this one doesn't break blueprints ^^. I'll make the change though

generators/server/templates/src/test/resources/testcontainers.properties.ejs
Tcharl: It is left like it was: this configuration bit just reuses the container among test case, but stop it after the whole test suite has been executed.

Tcharl:
> Maybe testcontainer (lack of) profile is ignored. And maybe it’s starting various duplicate containers causing the out of memory.

Not sure to understand how it works exactly.
From what I think: maven testcontainers profile if for sql only, not for the others (cassandra, mongo, redis, ...): thus there's no need to set it to '1' in github actions, right?

mshima:
> Not sure to understand how it works exactly.
> From what I think: maven testcontainers profile if for sql only, not for the others (cassandra, mongo, redis, ...): thus there's no need to set it to '1' in github actions, right?

If testcontainers is set to 1, containers will be started after backend tests. Otherwise, containers will be started before backend test to be used by the tests.

mongodb+kafka test is set to 0 because kafka doesn't support testcontainers. MongoDB will have a container + a testcontainer started.

I am not against enforce testcontainers, but IMO it would be better to be done for v8 or at least tests should pass.

mshima:
Why don't postpone MongoDB changes to after others services? It already gave a lot of trouble at SB2.6 migration.

Tcharl:
FYI, I'm currently dropping junit extension in favor of spring beans: containers will then be destroyed at the end of the test suite instead of being destroyed at the end of each test. Performance gains!

DanielFran:
@Tcharl all MS tests has been fixed in main branch... Can you update?

mraible:
Please resolve conflicts.

mshima:
Add Kafka to a microservice stack sample, it tests maven/gradle and reactive/imperative matrix in one test.
Angular and WebFlux workflows files are broken, if the stack sample passes, we can remove the angular/webflux broken samples.

generators/server/templates/src/main/resources/config/application.yml.ejs
```diff
@@ -189,6 +189,23 @@ spring:
         pool:
           max-connections: 1000
   <%_ } _%>
+  <%_ if (messageBrokerKafka) { _%>
+    stream:
+      kafka:
+        binder:
+          replicationFactor: 1
+          auto-create-topics: true
+          brokers: localhost:9092
```
mshima, line 198:
```suggestion
          # Add kafka host to your hosts
          brokers: kafka:9092
```

generators/server/templates/src/main/resources/config/application.yml.ejs
```diff
@@ -195,7 +195,8 @@ spring:
         binder:
           replicationFactor: 1
           auto-create-topics: true
-          brokers: localhost:9092
+          # Add kafka host to your hosts
```
Tcharl, line 198: I would prefer to avoid: otherwise we would have to do the same for kc, hazelcast, redis, ....
So 10 lines to add to your host before being able to startup your app locally...

mshima: Testing docker compose.
Fixed the docker env variable.

generators/server/templates/pom.xml.ejs
```diff
@@ -77,7 +77,7 @@
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
         <start-class><%= packageName %>.<%= mainClass %></start-class>
-        <argLine>-Djava.security.egd=file:/dev/./urandom -Xmx256m</argLine>
+        <argLine>-Djava.security.egd=file:/dev/./urandom -Xmx1G</argLine>
```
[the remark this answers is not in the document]
Tcharl: Not sure, this is to test.
The threshold was for mongo mostly (remember, we struggled on that). But changes introduced may have fixed this issue since then

generators/server/templates/src/main/resources/config/application.yml.ejs
```diff
@@ -298,7 +316,7 @@ spring:
             client-id: web_app
             client-secret: web_app
   <%_ } _%>
-            scope: openid,profile,email
+            scope: openid, profile, email, offline_access # last one for refresh tokens
```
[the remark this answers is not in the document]
Tcharl: It's the only way to refresh an expired access token. Also, it looks to be mainstream: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#offline_access

Tcharl:
Ahhhhhh, green: finally!

generators/server/templates/pom.xml.ejs
```diff
@@ -1029,7 +1031,7 @@
                                           fork="true"
                                           failonerror="true"
                                           newenvironment="true"
-                                          maxmemory="512m"
+                                          maxmemory="1G"
```
mshima, line 1034:
```suggestion
                                          maxmemory="512m"
```

generators/server/templates/src/main/resources/config/application.yml.ejs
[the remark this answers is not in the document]
mshima: Yes, but since the token is kept in the backend session, not in the client, there is no use for renewing expired tokens.
If we increase the clockSkew to match session timeout, the session token will never expire.

It's useful for ionic.

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -18,156 +18,85 @@
...
+import <%= packageName %>.IntegrationTest;
+import <%= packageName %>.config.EmbeddedKafka;
+import <%= packageName %>.config.KafkaSseConsumer;
+import <%= packageName %>.config.KafkaSseProducer;
+
+@IntegrationTest
+@AutoConfigureMockMvc
+@WithMockUser
+@EmbeddedKafka
+class SampleMongoKafkaKafkaResourceIT {
```
mshima, line 53:
```suggestion
class <%= capitalizedBasename %>KafkaResourceIT {
```

generators/server/templates/src/test/java/package/web/rest/KafkaResourceIT.java.ejs
```diff
@@ -18,156 +18,85 @@
...
+import <%= packageName %>.IntegrationTest;
+import <%= packageName %>.config.EmbeddedKafka;
+import <%= packageName %>.config.KafkaSseConsumer;
+import <%= packageName %>.config.KafkaSseProducer;
+
+@IntegrationTest
+@AutoConfigureMockMvc
+@WithMockUser
+@EmbeddedKafka
+class <%= capitalizedBasename %>KafkaResourceIT {
```
mshima, line 53:
```suggestion
class <%= capitalizedBaseName %>KafkaResourceIT {
```

Tcharl: <%= upperFirstCamelCase(baseName) %> ^^. I'm on it

Tcharl:
So do you consider that it is OK? May I continue to Redis, then ES, then DBs, ...

mshima (approved):

.github/workflows/angular.yml
```diff
@@ -118,7 +118,7 @@ jobs:
             environment: dev
             war: 0
             e2e: 1
-            testcontainers: 0
+            testcontainers: 1
```
mshima, line 121:
```suggestion
            testcontainers: 0
```

pascalgrimaud:
@Tcharl : I don't understand totally what happened here. I can see you reintroduce spring-cloud-stream here

image

There was a ticket to not use this:
- https://github.com/jhipster/generator-jhipster/issues/9287
- https://github.com/jhipster/generator-jhipster/pull/9567

Issue #17775, opened 2022-02-03
TechnicalStructureTest failure due to access to ApplicationProperties


OmarHawk:
##### **Overview of the issue**
Hi,

I just updated from 7.5.0 to 7.6.0.

When building my application, a generated test runs to perform a TechnicalStructureTest. This one complains about access to 

`layer 'Config' ('..config..')`

which is forbidden from anywhere. 

`where layer 'Config' may not be accessed by any layer`

But since the auto-generated (and by me filled with life) config file `ApplicationProperties.java` (to my understanding and also from the documentation supposed to be for such kind of things) resides exactly there and is referenced from other places in my code, the application fails to build.

##### **Reproduce the error**

Create a new application.
Fill the `ApplicationProperties.java` with some custom fields and methods.
Reference the `ApplicationProperties` from anywhere.
Build the application.
Build failure

##### **Related issues**

#17757 but with different reason.

##### **Suggest a Fix**

Either define an exception rule for the `ApplicationProperties` or move it to some other place where it can be used everywhere throughout the application.

##### **JHipster Version(s)**
7.6.0

##### **JHipster configuration**
Saving this up. This is a conceptual issue.

##### **Browsers and Operating System**

- [x] Checking this box is mandatory (this is just to show you read everything)

OmarHawk:
Btw. whitelisting the ApplicationProperties with a file "archunit_ignore_patterns.txt" within the src/main/resources folder containing 

`.*ApplicationProperties.*`

serves as a workaround for now.

Issue #17757, opened 2022-02-02
Architecture test failure with skip user management option


vishal423:
##### **Overview of the issue**
If you generate a new application with skip user management option, then, the generated application has Junit test failure related to Architecture tests.
Refer build # https://github.com/jhipster/generator-jhipster-svelte/runs/4917371747?check_suite_focus=true#step:13:577 for error details

Refer JDL at https://github.com/jhipster/generator-jhipster-svelte/blob/main/.github/workflows/scripts/monolithic-jwt-maven-skip-user.jdl

##### **Motivation for or Use Case**
Prevents migration of svelte blueprint to use `7.6.0` JHipster release.

##### **JHipster Version(s)**
7.6.0

##### **Browsers and Operating System**

- [ ] Checking this box is mandatory (this is just to show you read everything)

mshima:
Removing the service layer from the ArchitectureTest can be problematic, since it's possible to add an entity with service on an existing project.
@codecholeric is there a way to ignore empty groups?

codecholeric:
@mshima there is a method `optionalLayer(..)` (see https://www.javadoc.io/doc/com.tngtech.archunit/archunit/latest/com/tngtech/archunit/library/Architectures.LayeredArchitecture.html#optionalLayer(java.lang.String) ). If you define the layer this way it may be empty...

atomfrede:
As the PR is merged, we can close this I think.

PR #17778, opened 2022-02-03
Make Service layer optional and add Properties to ignoreDependency.

mshima:
Fixes https://github.com/jhipster/generator-jhipster/issues/17775 and https://github.com/jhipster/generator-jhipster/issues/17757

---


generators/server/templates/src/test/java/package/TechnicalStructureTest.java.ejs
```diff
@@ -38,32 +38,30 @@ class TechnicalStructureTest {
...
-        .whereLayer("Security").mayOnlyBeAccessedByLayers(<% if (applicationTypeMicroservice) { %>"Client", <% } %>"Web"<% if (!databaseTypeNo || authenticationTypeOauth2) { %>, "Service"<% } %>, "Config")
+        .whereLayer("Security").mayOnlyBeAccessedByLayers("Config", <% if (applicationTypeMicroservice) { %>"Client", <% } %>"Service", "Web")
 <%_ if (!databaseTypeNo) { _%>
         .whereLayer("Persistence").mayOnlyBeAccessedByLayers("Service", "Security", "Web", "Config")
         .whereLayer("Domain").mayOnlyBeAccessedByLayers("Persistence", "Service", "Security", "Web", "Config")
 <%_ } _%>
 
-        .ignoreDependency(belongToAnyOf(<%= mainClass %>.class), alwaysTrue());
+        .ignoreDependency(belongToAnyOf(<%= mainClass %>.class), alwaysTrue())
+        .ignoreDependency(alwaysTrue(), belongToAnyOf(
```
vishal423, line 61: Is it possible to exclude a set of classes by regex? Like classes with suffix `Properties` should be excluded. Otherwise, the generated code seems too restrictive.

vishal423: Also, I see a `Constants` class as another candidate for exclusion

mshima: > Is it possible to exclude a set of classes by regex? Like classes with suffix `Properties` should be excluded.

Could not find.

> Otherwise, the generated code seems too restrictive.

It’s meant to be restrictive for best practices.
A needle could be added for blueprints.

> Also, I see a `Constants` class as another candidate for exclusion

Indeed.

vishal423: It doesn't sound right to impose an architecture that requires you to modify the test for every additional properties class (don't be constrained with blueprints/generator). I would rather exclude complete config package from such checks.

mshima: Blueprints can remove the generated file, edit it, adapt to it or open issues/pr to improve it.
We must try to keep our best practices in our generated code.
Our review process is mainly based on tests. Imposing best practices inside tests, makes the review process easier.

DanielFran: @vishal423 we can exclude a set of classes by regex, see https://www.archunit.org/userguide/html/000_Index.html#_ignoring_violations
````
 A more generic approach is to ignore violations based on simple regex matches. For this one can put a file named archunit_ignore_patterns.txt in the root of the classpath. Every line will be interpreted as a regular expression and checked against reported violations. Violations with a message matching the pattern will be ignored. If no violations are left, the check will pass.
````

Using this way it seems pretty easy to exclude anything.
So for me we keep like that and we improve based on specific cases after.

DanielFran (approved):

mshima:
Let's merge to prepare for v7.7.0, we can change this later if appropriated.

PR #17811, opened 2022-02-07
npmw: revert to global node if local is not found.

mshima:
Local node is disabled for gradle and apple silicon, improves apple silicon support.

---

### DIFF
--- a/generators/common/files.js
+++ b/generators/common/files.js
@@ -62,8 +62,6 @@ const commonFiles = {
         '.lintstagedrc.js',
         {
           file: '.husky/pre-commit',
-          method: 'copy',
-          noEjs: true,
         },
       ],
     },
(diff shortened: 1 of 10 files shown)


pascalgrimaud:
Not related to this PR, but I think there is an issue at line 8, with webapp profile, which doesn't exist for microservice

mshima:
`npmw` isn’t used for microservice at the generation time.
https://github.com/jhipster/generator-jhipster/blob/687995104775c290ae4f0b2e506329e56a71149b/generators/server/index.js#L104-L106

If we write it, we can remove.

pascalgrimaud:
It is used here: https://github.com/jhipster/jhipster-sample-app-microservice/blob/main/npmw#L8

And if I'm not wrong, I saw some error when I upgraded my microservice with last version of JHipster, when commiting. So maybe some use by the prettier auto commit with husky

mshima:
I mean we should remove the `npmw` command and adjust husky.

Issue #14745, opened 2021-04-22
Jhipster 7.0.1 and couchbase 7


blackxMan:
Hello,
is there any chance to support couchbase 7 with scopes & collections feature.
thank you

mraible:
If you're willing to do the work, we'd love to have Couchbase support! We dropped it from JHipster 7 because we found Spring Data's support wasn't that great with Spring Boot 2.3 and 2.4.

https://github.com/jhipster/generator-jhipster/pull/14184

blackxMan:
hey @mraible,
Thanks for answer, i think they fix all problem in the new version Spring Data Couchbase 4.0  based on couchbase sdk 3

mraible:
@blackxMan If you'd like to see Couchbase support re-added, you're more than welcome to fork my repo and contribute to the `fix-couchbase` branch. I've spent too many hours on it that I'm no longer interested in working on it.

murdos:
> i think they fix all problem in the new version Spring Data Couchbase 4.0 based on couchbase sdk 3

On the contrary, things go broken since Spring Data Couchbase 4.0 release.
I also spent too many hours trying to fix it.

coderguy-tech:
Hey,
I'm working on this right now: 
1. Updating JHipster codebase to use the new version of Spring Data Couchbase 4 based on Couchbase SDK 3
2. Re-wiring some of the older code (wherever CB related database code is) to ensure that it works with Spring Data and Couchbase SDK API calls. 
3. Running through the test suites of JHipster to ensure that they pass — as we hit failures in the tests, we are resolving these issues as we go along

If I need help or get stuck along the way, will reach out.
Thanks

DanielFran:
@coderguy-tech dear not forget #14184, you can continue based on that...

coderguy-tech:
Thanks, I'll take a look at that

gkysaad:
As I was working on the fix I came across an issue with the Couchmove implementation. Couchbase's SearchIndex class' `toJson()` function does not output all of the custom data used to create a SearchIndex object, resulting in FTS indices not being created properly with all the required data in the database. So, I've created a custom SearchIndex class and created a PR in the couchmove repo to fix this bug there: https://github.com/differentway/couchmove/pull/13.

DanielFran:
@gkysaad I merged on BOM the update to couchmove 3.1.1

[...]

PR #17822, opened 2022-02-08
[Couchbase] Add support to scope and collections

tchlyah:
Add support to [scope and collections](https://docs.couchbase.com/server/current/manage/manage-scopes-and-collections/manage-scopes-and-collections.html) introduced in Couchbase 7.0.

Resolves https://github.com/jhipster/generator-jhipster/issues/14745.

Needs https://github.com/jhipster/jhipster-bom/pull/646

---


mshima:
Switched to the related bom branch.

mshima:
https://github.com/jhipster/generator-jhipster/blob/90008a0fa8d4eaaed484d2131023b989773bc868/.github/actions/setup/action.yml#L151
Try changing from CICD to SNAPSHOT.

DanielFran:
@tchlyah The issue that I saw prior your last changes was the couchmove version 3.2-SNAPSHOT was not found

tchlyah:
> @tchlyah The issue that I saw prior your last changes was that couchmove version 3.2-SNAPSHOT was not found

Yes, but it is in the ossrh-snapshot repo: https://oss.sonatype.org/service/local/repositories/snapshots/content/com/github/differentway/couchmove/3.2-SNAPSHOT/

> Try changing from CICD to SNAPSHOT.

Doesn't seem to work

mshima:
Repository should be ok now.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.ts.ejs
```diff
@@ -17,6 +17,8 @@
  limitations under the License.
 -%>
 <%_
+differentRelationships = differentRelationships.filter(rel => !rel.otherEntityIsEmbedded)
```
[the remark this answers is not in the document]
tchlyah: it seems that it is: [entity.js#L86](https://github.com/tchlyah/generator-jhipster/blob/feat/couchbase-collections/utils/entity.js#L86-L86)

I don't have this error locally 🤷‍♂️

generators/entity-server/files-couchbase.js
```diff
@@ -21,12 +21,23 @@ const { SERVER_MAIN_SRC_DIR, SERVER_MAIN_RES_DIR } = require('../generator-const
...
+    },
+    {
+      condition: generator => generator.searchEngineCouchbase && !generator.skipDbChangelog && !generator.embedded,
       path: SERVER_MAIN_RES_DIR,
       templates: [
         {
           file: 'config/couchmove/changelog/entity.fts',
-          renameTo: generator => `config/couchmove/changelog/V${generator.changelogDate}__${generator.entityInstance.toLowerCase()}.fts`,
+          renameTo: generator =>
+            `config/couchmove/changelog/V${parseInt(generator.changelogDate, 10) + 10}__${generator.entityInstance.toLowerCase()}.fts`,
```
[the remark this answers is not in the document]
tchlyah: Changelogs should have different versions, and since we can have both n1ql and fts changelong for an entity, we need to increment it a bit. I tried first with 1 but it doesn't work because a lot of entities in the sample have just 1s of difference. So I incremented by 10.

If you have a better idea on how to handle this I'm all ears.

generators/server/templates/src/test/java/package/web/rest/UserResourceIT.java.ejs
```diff
@@ -128,7 +128,7 @@ class UserResourceIT <% if (databaseTypeCassandra) { %>extends AbstractCassandra
 <%_ } _%>
 
 <%_ if (databaseTypeCouchbase) { _%>
-    private static final String DEFAULT_ID = <%= asEntity('User') %>.PREFIX + DEFAULT_LOGIN;
+    private static final String DEFAULT_ID = DEFAULT_LOGIN;
```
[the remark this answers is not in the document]
tchlyah: I kept it to minimize the modifications between Couchbase and other databases that needs that. Also, it's maybe important to understand that the id is now the login.

generators/entity-client/templates/angular/src/main/webapp/app/entities/update/entity-management-update.component.ts.ejs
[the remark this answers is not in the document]
mshima: > it seems that it is: [entity.js#L86](https://github.com/tchlyah/generator-jhipster/blob/feat/couchbase-collections/utils/entity.js#L86-L86)

That was a bug, see, the link now.
And the usage https://github.com/jhipster/generator-jhipster/blob/b474b1ea960d604da6fa19fc06cac43acbec434c/generators/entity/index.js#L727-L730

generators/entity-server/files-couchbase.js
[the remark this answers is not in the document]
mshima: This is not the best place, but since this is couchbase specific, and we are not modular yet, it's ok.

tchlyah:
I don't understand why tests are failing, everything is passing locally. I just updated jhipster-bom branch, can you run the tests again ?

DanielFran:
@tchlyah Can you update based on main branch, please?

.github/actions/setup/action.yml
```diff
@@ -67,11 +67,11 @@ inputs:
   jhipster-bom-repository:
     description: 'JHipster BOM repository'
     required: false
-    default: https://github.com/jhipster/jhipster-bom.git
+    default: https://github.com/tchlyah/jhipster-bom.git
```
DanielFran, line 70:
```suggestion
    default: https://github.com/jhipster/jhipster-bom.git
```

.github/actions/setup/action.yml
```diff
@@ -67,11 +67,11 @@ inputs:
...
   jhipster-bom-branch:
     description: 'JHipster BOM branch'
     required: false
-    default: auto
+    default: upgrade-couchbase
```
DanielFran, line 74:
```suggestion
    default: auto
```

.github/actions/setup/action.yml
```diff
@@ -148,7 +148,7 @@ runs:
         echo "NG_CLI_ANALYTICS=ci" >> $GITHUB_ENV
 
         if [[ "${{inputs.jhipster-bom-branch}}" != "release" && "${{inputs.jhipster-bom-branch}}" != "ignore" ]]; then
-          echo "JHI_BOM_VERSION=0.0.0-CICD" >> $GITHUB_ENV
+          echo "JHI_BOM_VERSION=0.0.0-SNAPSHOT" >> $GITHUB_ENV
```
DanielFran, line 151:
```suggestion
          echo "JHI_BOM_VERSION=0.0.0-CICD" >> $GITHUB_ENV
```

DanielFran (changes requested):

DanielFran (approved):

tchlyah:
It seems that there is some kind of regression on AccountResource IT tests introduced by latest commits. It seems infecting only Couchbase. I'm investigating.

tchlyah:
I found that: [SecurityMockMvcResultHandlersexporting-the-securitycontext#exporting-the-securitycontext](https://github.com/spring-projects/spring-security/blob/main/docs/modules/ROOT/pages/servlet/test/mockmvc/result-handlers.adoc#exporting-the-securitycontext)

It seems that MockMvc clear SecurityContext after performing requests. But What I can't understand, is why this is not happening for all tests. This should also affect MongoDB (https://github.com/spring-projects/spring-data-mongodb/issues/2906). Weird.

Adding `.andDo(exportTestSecurityContext())` to the failing tests fixes the problem. What do you think about that ? Should I add it everywhere to be safe, or just for Couchbase, or just for Couchbase for only those 3 methods ?

mshima:
Looks related to https://github.com/jhipster/generator-jhipster/pull/17901.
AFAIK `#{principal?.*}` is only used at sql here, and I doubt there is a test case for it:
https://github.com/jhipster/generator-jhipster/blob/503d6a66fc0f38e800f2e498104bf51d1541ffe8/generators/entity-server/templates/src/main/java/package/repository/EntityRepository.java.ejs#L76

So another option is to make that dependency sql only.
Have you introduced `#{principal?.*}` anywhere?

mshima:
I would go with only one solution.
Looks like they depend on each other, so makes more sense to add both or none.
If a developer adds `#{principal?.*}` and fails, it should be easy to find the solution.
But be free to do both if it feels appropriate.

DanielFran:
@tchlyah can you rebase from main?
@mshima made some adjustments for Integration tests last week...

tchlyah:
> @tchlyah can you rebase from main?
> @mshima made some adjustments for Integration tests last week...

Already did! doesn't change anything.

mshima:
A guess is that the generated username may not be safe for couchbase logic.
https://github.com/jhipster/generator-jhipster/blob/e471bf35d70513e90d4ad01fe99f434cae6451bc/generators/cypress/templates/src/test/javascript/cypress/integration/account/register-page.spec.ts.ejs#L134

tchlyah:
> A guess is that the generated username may not be safe for couchbase logic.
> 
> https://github.com/jhipster/generator-jhipster/blob/e471bf35d70513e90d4ad01fe99f434cae6451bc/generators/cypress/templates/src/test/javascript/cypress/integration/account/register-page.spec.ts.ejs#L134

No, I think I understand. Integration tests wait for all indexes to be built before running tests, which is not the case for PROD. I put the waiting on main code in order to validate this. If it is the case we need to discuss what is the best option to fix it.

mshima:
What about a property and a spring e2e profile?
Test property value can be added at https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/test/resources/config/application.yml.ejs.
And create a application-e2e.yml.

mshima (commented):
EJS ident to improve readability.

mshima:
Backend test and e2e are different and IMO should not be mixed.
A delay to start e2e will work? Maybe we should just delay it a bit.

tchlyah:
@mshima I implemented a first draft here: https://github.com/jhipster/generator-jhipster/pull/17822/commits/5b61835bcbc53668fb47585c9427874202602775
WDYT ?

mshima:
I still don't like a profile called test outside test folder.
I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.

tchlyah:
> I still don't like a profile called test outside test folder.
> I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.

Ok, now I understand what you mean, make sens. But we already have `test` profile on [JHipsterConstants#SPRING_PROFILE_TEST](https://github.com/jhipster/jhipster-bom/blob/a704aa4d511d5710e649267eb3135b728ce64c20/jhipster-framework/src/main/java/tech/jhipster/config/JHipsterConstants.java#L31), so it make it easier.

I can name it `e2e`, but it will mean that I'll also activate this profile in integration tests: https://github.com/jhipster/generator-jhipster/blob/5b61835bcbc53668fb47585c9427874202602775/generators/server/templates/src/test/resources/config/application.yml.ejs#L69

tchlyah:
Do you know if the timeout in npm tests are linked to my commit ? https://github.com/jhipster/generator-jhipster/runs/5291720283?check_suite_focus=true

Tcharl:
> I still don't like a profile called test outside test folder.
> I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.

Agree, profile could ideally be triggered for test classes annotated with @IntegrationTest.
A simple thing could be to extend this class Configuration class in src/test/java... and to activate it through the annotation

mshima:
> Do you know if the timeout in npm tests are linked to my commit ?

Probably not related.

tchlyah:
> > I still don't like a profile called test outside test folder.
> > I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> > Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.
> 
> Agree, profile could ideally be triggered for test classes annotated with @IntegrationTEST. A simple thing could be to extend this class Configuration class in src/test/java... and to activate it through the annotation

This was my previous version. It was working for integration tests but we need the same for e2e tests, which is running with production package, that doesn’t contain test classes.

Tcharl:
> > > I still don't like a profile called test outside test folder.
> > > I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> > > Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.
> > 
> > Agree, profile could ideally be triggered for test classes annotated with @IntegrationTEST. A simple thing could be to extend this class Configuration class in src/test/java... and to activate it through the annotation
> 
> This was my previous version. It was working for integration tests but we need the same for e2e tests, which is running with production package, that doesn’t contain test classes.

Sorry for misunderstanding:  in this case e2e profile is recommended there. Dumb question (again) is this index check not mandatory in any case? No matter if it's test or prod

tchlyah:
> > > > I still don't like a profile called test outside test folder.
> > > > I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> > > > Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.
> > > 
> > > 
> > > Agree, profile could ideally be triggered for test classes annotated with @IntegrationTEST. A simple thing could be to extend this class Configuration class in src/test/java... and to activate it through the annotation
> > 
> > 
> > This was my previous version. It was working for integration tests but we need the same for e2e tests, which is running with production package, that doesn’t contain test classes.
> 
> Sorry for misunderstanding: in this case e2e profile is recommended there.

I didn’t want to name it ’e2e’ for two reasons: we already have the constant as stated above, and it will be used on integration test also. e2e and integration are both tests so it make sens.

But if you both prefer to name it like that I don’t mind.

> Dumb question (again) is this index check not mandatory in any case? No matter if it's test or prod

We do not want to slow production with unnecessary checks that are only needed for tests. Checking indexes can be heavy queries depending on their number on the cluster.

generators/server/index.js
```diff
@@ -585,6 +585,7 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
         const javaCommonLog = `-Dlogging.level.ROOT=OFF -Dlogging.level.org.zalando=OFF -Dlogging.level.tech.jhipster=OFF -Dlogging.level.${this.jhipsterConfig.packageName}=OFF`;
         const javaTestLog =
           '-Dlogging.level.org.springframework=OFF -Dlogging.level.org.springframework.web=OFF -Dlogging.level.org.springframework.security=OFF';
+        const javaTestProfile = this.jhipsterConfig.databaseType === COUCHBASE ? ',test' : '';
```
mshima, line 588:
```suggestion
```

Tcharl:
> > > > > I still don't like a profile called test outside test folder.
> > > > > I would prefer a generic `e2e` profile (not couchbase specific) or Couchbase specific `couchmove_test`.
> > > > > Anyway, this is not a blocker for me if @DanielFran or @Tcharl think it's ok.
> > > > 
> > > > 
> > > > Agree, profile could ideally be triggered for test classes annotated with @IntegrationTEST. A simple thing could be to extend this class Configuration class in src/test/java... and to activate it through the annotation
> > > 
> > > 
> > > This was my previous version. It was working for integration tests but we need the same for e2e tests, which is running with production package, that doesn’t contain test classes.
> > 
> > Sorry for misunderstanding: in this case e2e profile is recommended there.
> 
> I didn’t want to name it `e2e`  for two reasons: we already have the constant as stated above, and it will be used on integration tests also. e2e and integration are both tests so it make sens.
> 
> But if you both prefer to name it like that I don’t mind.
> 
> > Dumb question (again) is this index check not mandatory in any case? No matter if it's test or prod
> 
> We do not want to slow production with unnecessary checks that are only needed for tests. Checking indexes can be heavy queries depending on their number on the cluster.
> 
> 

So what about disabling that check when prod profile is on?

generators/server/index.js
```diff
@@ -634,7 +635,7 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
           'ci:e2e:package':
             'npm run java:$npm_package_config_packaging:$npm_package_config_default_environment -- -Pe2e -Denforcer.skip=true',
           'preci:e2e:server:start': 'npm run docker:db:await --if-present && npm run docker:others:await --if-present',
-          'ci:e2e:server:start': `java -jar ${e2ePackage}.$npm_package_config_packaging --spring.profiles.active=$npm_package_config_default_environment ${javaCommonLog} ${javaTestLog} --logging.level.org.springframework.web=ERROR`,
+          'ci:e2e:server:start': `java -jar ${e2ePackage}.$npm_package_config_packaging --spring.profiles.active=$npm_package_config_default_environment${javaTestProfile} ${javaCommonLog} ${javaTestLog} --logging.level.org.springframework.web=ERROR`,
```
mshima, line 638:
```suggestion
          'ci:e2e:server:start': `java -jar ${e2ePackage}.$npm_package_config_packaging --spring.profiles.active=e2e,$npm_package_config_default_environment ${javaCommonLog} ${javaTestLog} --logging.level.org.springframework.web=ERROR`,
```

generators/server/templates/src/main/java/package/config/DatabaseConfiguration_couchbase.java.ejs
```diff
@@ -164,25 +165,34 @@ public class DatabaseConfiguration extends AbstractCouchbaseConfiguration {
...
+        Collection collection = bucket.scope(getScopeName()).collection(CHANGELOG_COLLECTION);
+        Couchmove couchmove = new Couchmove(collection, cluster, "config/couchmove/changelog");
         couchmove.migrate();
-        couchmove.buildN1qlDeferredIndexes();
+        couchmove.buildN1qlDeferredIndexes(getScopeName());
         return couchmove;
     }
 
+    @Bean
+    @Profile(JHipsterConstants.SPRING_PROFILE_TEST)
```
mshima, line 180:
```suggestion
    @Profile({JHipsterConstants.SPRING_PROFILE_TEST, JHipsterConstants.SPRING_PROFILE_E2E})
```

`SPRING_PROFILE_E2E` doesn't exist yet.

generators/server/templates/src/test/resources/config/application.yml.ejs
```diff
@@ -64,6 +64,9 @@ spring:
   profiles:
     # Uncomment the following line to enable tests against production database type rather than H2, using Testcontainers
     #active: testcontainers
+<%_ } else if (databaseTypeCouchbase) { _%>
+  profiles:
+    active: test
```
mshima, line 69:
```suggestion
```

If possible I would prefer adding
```
@ActiveProfiles(JHipsterConstants.SPRING_PROFILE_TEST)
```
To [IntegrationTest](https://github.com/jhipster/generator-jhipster/blob/ba83d069251ec488e93fceb2379ddf49be04ae7e/generators/server/templates/src/test/java/package/IntegrationTest.java.ejs#L61).

I don't think it needs be specific to couchbase.

mshima (commented):
Please add e2e profile too.

tchlyah:
> So what about disabling that check when prod profile is on?

Yes, that could work, on both DEV et PROD profile, avoiding to add any other profile, why not. So @Tcharl, @mshima, which solution do you prefer ? I'll implemented ASAP.

mshima:
> > So what about disabling that check when prod profile is on?
> 
> Yes, that could work, on both DEV et PROD profile, avoiding to add any other profile, why not.

I don’t get how this will work. dev or prod profiles are always on.

tchlyah:
>

tchlyah:
> > > So what about disabling that check when prod profile is on?
> > 
> > 
> > Yes, that could work, on both DEV et PROD profile, avoiding to add any other profile, why not.
> 
> I don’t get how this will work. dev or prod profiles are always on.

Yes you're right, sorry I got confused!

mshima:
We will have problems with Microservices. We use docker-compose with every service.
The profile would need to be passed through environment variable.

A easy solution is to wait indexes by default (so e2e and tests will always work out of the box) and add a production optimization recommendation note.

generators/server/templates/src/main/java/package/config/DatabaseConfiguration_couchbase.java.ejs
tchlyah: If you are happy with the proposed solution, I’ll add `SPRING_PROFILE_E2E ` to jhipster-framework

tchlyah:
@mshima @DanielFran @Tcharl I hope this time you’ll be ok with solution. This PR was more complicated than I though 😅 I’m eager to close it!

mshima (approved):

generators/server/templates/src/main/java/package/config/DatabaseConfiguration_couchbase.java.ejs
```diff
@@ -164,25 +165,34 @@ public class DatabaseConfiguration extends AbstractCouchbaseConfiguration {
...
+        Collection collection = bucket.scope(getScopeName()).collection(CHANGELOG_COLLECTION);
+        Couchmove couchmove = new Couchmove(collection, cluster, "config/couchmove/changelog");
         couchmove.migrate();
-        couchmove.buildN1qlDeferredIndexes();
+        couchmove.buildN1qlDeferredIndexes(getScopeName());
         return couchmove;
     }
 
+    @Bean
+    @Profile({JHipsterConstants.SPRING_PROFILE_TEST, "e2e"})
```
mshima, line 180:
```suggestion
    @Profile({JHipsterConstants.SPRING_PROFILE_TEST, JHipsterConstants.SPRING_PROFILE_E2E})
```

tchlyah:
> It's a huge work here, so I'm adding a big bounty
> plz claim it @tchlyah

Thanks a lot @pascalgrimaud ! I've claimed the bug bounty: https://opencollective.com/generator-jhipster/expenses/65435

pascalgrimaud:
@Tcharl : approved :-)

PR #17887, opened 2022-02-15
set x-frame-options header to deny in reactive stack

### LINKED ISSUE #17886: X-Frame-Options header not set in reactive stack

In the non-Reactive stack, X-Frame-Options are set to DENY:

https://github.com/jhipster/generator-jhipster/blob/980ca577b72a846208d2fec0d85267a995bc8e53/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L238-L239

In the reactive stack, it is not set at all (disabled):

https://github.com/jhipster/generator-jhipster/blob/b474b1ea960d604da6fa19fc06cac43acbec434c/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs#L258

I think, both stacks should behave same security wise... At least, the missing header is reported to me for my application...

To prevent clickjacking, X-Frame-Options header should be set in general, and ideally, we may set it the same way in all stacks...

##### **Suggest a Fix**

I'll prepare a PR to make change that LOC.

##### **JHipster Version(s)**
7.6.0

Saved myself from that.

OmarHawk:
fixes #17886

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
@@ -112,6 +112,7 @@ import org.springframework.security.web.server.authentication.HttpStatusServerEn
 import org.springframework.security.web.server.authentication.logout.HttpStatusReturningServerLogoutSuccessHandler;
 import org.springframework.security.web.server.csrf.CookieServerCsrfTokenRepository;
 <%_ } _%>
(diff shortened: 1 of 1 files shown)


OmarHawk:
Ah, I see, Seems like this is ssomewhat needed for the token refresh...

DanielFran (approved):

DanielFran:
Thanks @OmarHawk for the contribution

PR #18179, opened 2022-03-22
Add permissions to GitHub actions

naveensrinivasan:
- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Also, dependabot supports upgrading based on SHA.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

---

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -4,8 +4,14 @@ on:
     branches-ignore:
       - 'dependabot/**'
 
+permissions:
+  contents: read
+
 jobs:
   triage:
+    permissions:
+      contents: read  # for actions/labeler to determine modified files
+      pull-requests: write  # for actions/labeler to add labels to PRs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/labeler@v4
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -10,8 +10,8 @@ permissions:
 jobs:
   triage:
     permissions:
-      contents: read  # for actions/labeler to determine modified files
-      pull-requests: write  # for actions/labeler to add labels to PRs
+      contents: read # for actions/labeler to determine modified files
+      pull-requests: write # for actions/labeler to add labels to PRs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/labeler@v4
--- a/.github/workflows/lock-maintenance.yml
+++ b/.github/workflows/lock-maintenance.yml
@@ -4,8 +4,14 @@ on:
     - cron: 0 0 * * 6
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
     name: Bump transitional dependencies
     if: github.repository == 'jhipster/generator-jhipster'
     runs-on: ubuntu-latest
--- a/.github/workflows/lock-maintenance.yml
+++ b/.github/workflows/lock-maintenance.yml
@@ -10,8 +10,8 @@ permissions:
 jobs:
   build:
     permissions:
-      contents: write  # for peter-evans/create-pull-request to create branch
-      pull-requests: write  # for peter-evans/create-pull-request to create a PR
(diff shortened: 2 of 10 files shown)


mraible:
Why do you think this is a good idea? 

I'd prefer to use versions like `v2` and `v3` so minor releases are picked up without further maintenance. Pinning to commits seems like it'd be even more work than using patch versions.

mshima:
Using commits is more secure if we did third pard actions audits.
We don't audit GitHub actions and we have dependabot auto merge.
So IMO is not really useful.

Action permissions is welcome.

naveensrinivasan:
> Using commits is more secure if we did third pard actions audits. We don't audit GitHub actions and we have dependabot auto merge. So IMO is not really useful.
> 
> Action permissions is welcome.

Thanks, I would appreciate it if this could be merged sooner as it would avoid my re-work merge conflicts.

mshima:
@naveensrinivasan the idea is to keep using tags as sha will make the maintenance harder.

If you revert the actions to use tags, we can merge the permissions.

naveensrinivasan:
> @naveensrinivasan the idea is to keep using tags as sha will make the maintenance harder.
> 
> If you revert the actions to use tags, we can merge the permissions.

OK, I understand, I can do that.

naveensrinivasan:
> @naveensrinivasan the idea is to keep using tags as sha will make the maintenance harder.
> 
> If you revert the actions to use tags, we can merge the permissions.

@mshima  I have updated the PR with just permissions. PTAL.

.github/workflows/labeler.yml
```diff
@@ -4,8 +4,14 @@ on:
...
       - 'dependabot/**'
 
+permissions:
+  contents: read
+
 jobs:
   triage:
+    permissions:
+      contents: read  # for actions/labeler to determine modified files
+      pull-requests: write  # for actions/labeler to add labels to PRs
```
mshima, line 14:
```suggestion
      contents: read # for actions/labeler to determine modified files
      pull-requests: write # for actions/labeler to add labels to PRs
```

.github/workflows/lock-maintenance.yml
```diff
@@ -4,8 +4,14 @@ on:
...
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      contents: write  # for peter-evans/create-pull-request to create branch
+      pull-requests: write  # for peter-evans/create-pull-request to create a PR
```
mshima, line 14:
```suggestion
      contents: write # for peter-evans/create-pull-request to create branch
      pull-requests: write # for peter-evans/create-pull-request to create a PR
```

.github/workflows/release-drafter.yml
```diff
@@ -25,8 +25,14 @@ on:
...
       - main
 
+permissions:
+  contents: read
+
 jobs:
   update_release_draft:
+    permissions:
+      contents: write  # for release-drafter/release-drafter to create a github release
+      pull-requests: write  # for release-drafter/release-drafter to add label to PR
```
mshima, line 35:
```suggestion
      contents: write # for release-drafter/release-drafter to create a github release
      pull-requests: write # for release-drafter/release-drafter to add label to PR
```

mshima (commented):
Prettier

mshima:
@naveensrinivasan should we add
```
permissions:
   contents: read
```

to angular, react and vue workflows?

naveensrinivasan:
> @naveensrinivasan should we add
> 
> ```
> 
> permissions:
> 
>    contents: read
> 
> ```
> 
> 
> 
> to angular, react and vue workflows?

Can that be a separate PR? Need to analyze before answering the question.

mshima (approved):

mshima:
Thanks

PR #18211, opened 2022-03-25
Fix a typo in french

M4jor-Tom:
Correct "Etes" to "Êtes"


DanielFran:
@M4jor-Tom thank you for the contribution

PR #18212, opened 2022-03-25
ClientForwardController Change Regex Mapping

### LINKED ISSUE #18039: ClientForwardController intercepts request but he should not
Hi,

I think, I found a minor issue in ClientForwardController. By the way JHipster rocks !

    @GetMapping(value = "/**/{path:[^\\.]*}")
With that configuration of ClientForwardController,

And with a controller with this configuration

@RequestMapping("/api")

...

`@GetMapping("/{one}/{two}/{three}/{four}/{five}")
    public ResponseEntity<String> getValue(
        @PathVariable(name = "one") int one,
        @PathVariable(name = "two") int two,
        @PathVariable(name = "three") int three,
        @PathVariable(name = "four") int four,
        @PathVariable(name = "five") int five
    )  {
        return ResponseEntity.ok("hello world");
    }`

With this request : http://localhost:8080/api/1/2/3/4/5

I don't enter into my function getValue, it's ClientForwardController which intercepts the request.

If i remove "five" pathVariable, I enter into the function getValue. a bit strange !

##### **Suggest a Fix**

    @GetMapping(value = "/{path:[^\\.]*}")

Remove "/**" at start like the case communicationSpringWebsocket in ClientForwardController.java.ejs and getValue works with 5 pathVariable

maiido:
closes #18039

---

### DIFF
--- a/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
+++ b/generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
@@ -31,7 +31,7 @@ public class ClientForwardController {
 <%_ if (communicationSpringWebsocket) { _%>
     @GetMapping(value = {"/{path:[^\\.]*}", "/{path:^(?!websocket).*}/**/{path:[^\\.]*}"})
 <%_ } else { _%>
-    @GetMapping(value = "/**/{path:[^\\.]*}")
+    @GetMapping(value = "/{path:[^\\.]*}")
 <%_ } _%>
     public String forward() {
         return "forward:/";
--- a/generators/server/templates/src/test/java/package/web/rest/ClientForwardControllerTest.java.ejs
+++ b/generators/server/templates/src/test/java/package/web/rest/ClientForwardControllerTest.java.ejs
@@ -65,10 +65,9 @@ class ClientForwardControllerTest {
 
     @Test
     void getNestedClientEndpoint() throws Exception {
-        restMockMvc.perform(get("/admin/user-management"))
-            .andExpect(status().isOk())
-            .andExpect(forwardedUrl("/"));
+        restMockMvc.perform(get("/admin/user-management")).andExpect(status().isNotFound());
     }
+    
     @Test
     void getUnmappedDottedEndpoint() throws Exception {
         restMockMvc.perform(get("/foo.js")).andExpect(status().isNotFound());


generators/server/templates/src/main/java/package/web/rest/ClientForwardController.java.ejs
```diff
@@ -31,7 +31,7 @@ public class ClientForwardController {
 <%_ if (communicationSpringWebsocket) { _%>
     @GetMapping(value = {"/{path:[^\\.]*}", "/{path:^(?!websocket).*}/**/{path:[^\\.]*}"})
 <%_ } else { _%>
-    @GetMapping(value = "/**/{path:[^\\.]*}")
+    @GetMapping(value = "/{path:[^\\.]*}")
```
vishal423, line 34: have you tested with base path?

maiido: sorry, not sure to understand the question

pascalgrimaud: I think @vishal423 want you to test with base path, it means: localhost:8080/xxxx/

But honestly, it's just a pain to configure all the project, to make it work.

So for me, it's the correct fix and people who uses base path can test and propose an improvement here

pascalgrimaud:
@maiido : can you update the test plz? Otherwise the CI will fail

OmarHawk:
> @maiido : can you update the test plz? Otherwise the CI will fail

I have tested this change in a built application and this makes the admin interface unavailable - if we push f5 or use a direct link... so please also make sure that stuff like that will continue to work. ;-)

maiido:
> > @maiido : can you update the test plz? Otherwise the CI will fail
> 
> I have tested this change in a built application and this makes the admin interface unavailable - if we push f5 or use a direct link... so please also make sure that stuff like that will continue to work. ;-)

Hi, it's strange, I don't have this issue. My test : 
Compile with maven (profiles dev+webapp)
launch webservice in intellij as an application
I have access to http://localhost:8080 api+webapp, and if i go to http://localhost:8080/login and press f5 after no issue.

OmarHawk:
> > > @maiido : can you update the test plz? Otherwise the CI will fail
> > 
> > 
> > I have tested this change in a built application and this makes the admin interface unavailable - if we push f5 or use a direct link... so please also make sure that stuff like that will continue to work. ;-)
> 
> Hi, it's strange, I don't have this issue. My test : Compile with maven (profiles dev+webapp) launch webservice in intellij as an application I have access to http://localhost:8080 api+webapp, and if i go to http://localhost:8080/login and press f5 after no issue.

well, but the admin / management interface is behind something like this:

http://localhost:8080/admin/logs

maiido:
@OmarHawk you are right, it's not ok

Issue #18156, opened 2022-03-20
Sort doesn't work for an entity having many-to-many relationships


vishal423:
##### **Overview of the issue**
Sort doesn't work for an entity having many-to-many relationships

##### **Reproduce the error**

- Use this [JDL](https://github.com/jhipster/generator-jhipster-svelte/blob/main/.github/workflows/scripts/monolithic-session-maven.jdl) to generate a monolithic application.
- Check the Get all REST API generated for `SampleDomainEntity` entity and notice the default parameter of eagerload is `true`.
- Start application and try to sort records by `Id` field. You shall see records are not sorted on UI and can also verify same by API response.
- As a workaround, pass additional parameter to override the default to `false` and can see records are correctly sorted.
- This seems related to PR https://github.com/jhipster/generator-jhipster/pull/13448

##### **JHipster Version(s)**

v7.7.0
##### **Browsers and Operating System**

- [ ] Checking this box is mandatory (this is just to show you read everything)

Issue #17990, opened 2022-02-27
Upgrade to 7.7.0: No property fetchBagRelationships found for type Entity


tschig:
##### **Overview of the issue**
(Name of my Entity was replaced by "Entity", Name of my Project was replaced by "Example")

Backend won't start after updating to 7.7.0 from 7.6.0. 
Repositories now have a EntityRepositoryWithBagRelationships super class.
Starting the backend results in a "No property fetchBagRelationships found for type Entity!" Error.

##### **Motivation for or Use Case**

JHipster Backend won't start anymore

##### **Reproduce the error**

Upgrade a existing JHipster App from 7.6.0 to 7.7.0

##### **Entity configuration(s) `entityName.json` files generated in the `.jhipster` directory**

Please see example project in comments.

Stacktrace:

```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'exampleService' defined in file [F:\Development\example\MVP\Backend\target\classes\tv\example\service\example\exampleService.class]: Unsatisfied dependency expressed through constructor parameter 5; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'exampleEntityRepository' defined in tv.example.repository.example.exampleEntityRepository defined in @EnableJpaRepositories declared on DatabaseConfiguration: Invocation of init method failed; nested exception is org.springframework.data.repository.query.QueryCreationException: Could not create query for public abstract java.util.Optional tv.example.repository.EntityRepositoryWithBagRelationships.fetchBagRelationships(java.util.Optional)! Reason: Failed to create query for method public abstract java.util.Optional tv.example.repository.EntityRepositoryWithBagRelationships.fetchBagRelationships(java.util.Optional)! No property fetchBagRelationships found for type Entity!; nested exception is java.lang.IllegalArgumentException: Failed to create query for method public abstract java.util.Optional tv.example.repository.EntityRepositoryWithBagRelationships.fetchBagRelationships(java.util.Optional)! No property fetchBagRelationships found for type Entity!
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:800)
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:229)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1372)
[...]

PR #18307, opened 2022-04-06
Fix bag relationships sort

mshima:
Fix https://github.com/jhipster/generator-jhipster/issues/17990 and fix https://github.com/jhipster/generator-jhipster/issues/18156

---


DanielFran (dismissed):

DanielFran:
@vishal423 Can you confirm it is OK for you, please?

vishal423 (dismissed):
It would be good to add Junit tests to assert this custom logic. Also, in my opinion, we shouldn't do eager fetch in getAll apis and it should be optional as it was before.

DanielFran:
@mshima can you add some tests?

generators/entity-server/templates/src/main/java/package/repository/EntityRepositoryWithBagRelationshipsImpl.java.ejs
```diff
@@ -78,14 +80,19 @@ public class <%= entityClass %>RepositoryWithBagRelationshipsImpl implements <%=
     }
 
     List<<%= persistClass %>> fetch<%= relationship.relationshipNameCapitalizedPlural %>(List<<%= persistClass %>> <%= entityInstancePlural %>) {
-        return entityManager
+        HashMap<Object, Integer> order = new HashMap<Object, Integer>();
```
Tcharl, line 83: Could you add the order in a dedicated private method (or at least the sorting stuff at the bottom of the method)

pascalgrimaud:
> Also, in my opinion, we shouldn't do eager fetch in getAll apis and it should be optional as it was before.

totally agree with @vishal423 here

DanielFran:
@mshima, Can you update the PR to not fetch eagerly by default?

generators/entity-server/templates/src/main/java/package/repository/EntityRepositoryWithBagRelationshipsImpl.java.ejs
mshima: Does the createQuery function changes the original array?

vishal423 (approved):

Issue #17850, opened 2022-02-11
Update Keycloak to v18


DanielFran:
##### **Overview of the feature request**

https://www.keycloak.org/2022/04/keycloak-1800-released

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

DanielFran:
Cc @vishal423

vishal423:
I will spend some time next week on this.

pascalgrimaud:
adding a bounty as I know it's not so easy :D

yelhouti:
Does anyone mind me working on this ?

yelhouti:
@vishal423 thanks I ll do it :)

JeroenAdam:
What was their (Keycloak devteam) reasoning for this change? First tought, I would follow their standard without /auth, they must have good reason to have done so.

yelhouti:
I created an issue with keycloak: https://github.com/keycloak/keycloak/issues/10322
To see if the reason why the CI is failing is expected behavior and i should adapt the code, or I should try to fix it in keycloak.

The CI can also be changed to rerun keycloak after import for tests to pass.

mraible:
Any updates y'all? I'm adding a bug bounty in hopes of making this happen faster.

yelhouti:
My bad, I thought I was done with it. I lol try to finish it later

yelhouti:
@mraible all the code should be there, but for some reason e2e are either waiting for keycloak or app to start (and it is not happening).
Is there a way to see the logs of `docker-compose up` ? I tried reproducing locally for an hour, but there too much jhipster integration tests internals I don't fully know about.

yelhouti:
There is an issue with startup order it seems (registry isn't waiting enough for keycloak to be ready), I hope to fix it on wednesday.

yelhouti:
@mraible a fix is comming to keycloak in the next version, that will hopefully fix this: https://github.com/keycloak/keycloak/issues/10322#issuecomment-1068498992

DanielFran:
Closed by https://github.com/jhipster/generator-jhipster/pull/18441

PR #18441, opened 2022-04-22
chore: upgrade keycloak to 18.0.0

yelhouti:
---

- [X] The JDL part is updated if necessary
- [X] [jhipster-online](https://github.com/jhipster/jhipster-online) is updated if necessary


yelhouti:
#fixes: 17850

generators/server/templates/src/main/docker/app.yml.ejs
```diff
@@ -64,7 +64,7 @@ _%>
       # - JHIPSTER_CACHE_REDIS_CLUSTER=true
 <%_ } _%>
 <%_ if (authenticationTypeOauth2) { _%>
-      - SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_OIDC_ISSUER_URI=http://keycloak:9080/auth/realms/jhipster
+      - SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_OIDC_ISSUER_URI=http://keycloak:8080/realms/jhipster
```
atomfrede, line 67: Should it be 9080 as thats the exposed port?

yelhouti: I think the exposed port is keycloak:8080 but it is exposed to localhost:9080

yelhouti:
@DanielFran @vishal423 
can anyone give access to the cypress dashboard so I can see why some tests are sporadic. Ouath2/keycloak seems to work only half the time :s

generators/server/templates/src/main/docker/app.yml.ejs
yelhouti: @atomfrede by the way half the tests work so I am pretty sure this is correct

atomfrede: Yes guess you're right. Just wanted to make sure as it was 9080 before, looked strange.

vishal423: this should be `9080` as that's exposed for outside services to connect

yelhouti: @vishal423 did you try it ? It was 9080 before, and I changed it because of how docker exposes ports.

vishal423:
@yelhouti, I suggest trying failed workflows on local env to reproduce issues

yelhouti:
@vishal423 I could not reproduce locally, also they fail sporadically in the CI. This is why some ouath2 test pass and other fails. Is there a way to access cypress dashboard ?

generators/server/templates/src/main/docker/app.yml.ejs
vishal423: are you saying there is a recent change in `docker` ports mapping behaviour? If not, then, your configuration is incorrect. The previous configuration used to work and was thoroughly tested in local and CI (with docker compose).

yelhouti:
@vishal423 no what I am saying is the way keycloak starts changed (no way to have a port offset of 1000) https://github.com/jhipster/generator-jhipster/pull/18441/files#diff-1bfaf6e31fe218cd507027450c04913b93f510df3f63b4a2ea84163c6423e9fbL339

But we still want to offset ports to have docker and keycloak not fight for port 8080 on localhost so we map keycloak:8080 to localhost:9080 

https://github.com/jhipster/generator-jhipster/pull/18441/files#diff-8eef712ac6e760fcb9b7b54a8cb6d5579608b146d14323efa75a30bb1397f82bR107

Also please give others the benefit of the doubt :)

vishal423:
> > are you saying there is a recent change in docker ports mapping behaviour? If not, then, your configuration is incorrect. The previous configuration used to work and was thoroughly tested in local and CI (with docker compose).
> 
> @vishal423 no what I am saying is the way keycloak starts changed (no way to have a port offset of 1000) https://github.com/jhipster/generator-jhipster/pull/18441/files#diff-1bfaf6e31fe218cd507027450c04913b93f510df3f63b4a2ea84163c6423e9fbL339
> 
> But we still want to offset ports to have docker and keycloak not fight for port 8080 on localhost so we map keycloak:8080 to localhost:9080
> 
> https://github.com/jhipster/generator-jhipster/pull/18441/files#diff-8eef712ac6e760fcb9b7b54a8cb6d5579608b146d14323efa75a30bb1397f82bR107
> 
> Also please give others the benefit of the doubt :)

Got it. You are trying to access within the docker network. Failure tests seems related to micro-frontend and micro-services. Can you try those flows locally?

yelhouti:
I tried ms-angular locally which one would you recommend.

 Also --monolith and --worksoace doesn't seem to work correctly maybe I am doing something wrong

I use scripts 11 and 12 the same way they are used on the workflow

mraible:
Can your please fix conflicts @yelhouti?

yelhouti:
I spent the last few hours understanding the issue with my PR and it is as follows:

Since keycloak doesn't support port offset anymore we are "forced" to change the port mapping at the docker level.
The issue then becomes as follows:
Since our code uses the same issuer-uri to:
1. talk directly to the IDP (verify tokens...)
2. redirect the user for login
If we set it to:
1. keycloak:8080 the browser is unable to talk to it because it can't resolve keycloak correctly (and pointing to localhost as I suspect we do now doesn't solve the problem) as mentioned in docker-compose.yaml: # For Keycloak to work, you need to add '127.0.0.1 keycloak' to your hosts file (which is cheating and I don't know how github actions was configured to do that
2. localhost:9080 the gateway is not able to talk to localhost correctly because it is inside the container.

Possible solution:
The gateway should know 2 urls: 1 for internal use (to verify token, retrieve oidc config in Security Configuration...) and one to return to the frontend or redirect the frontend to.
There are other solutions I could think of but they are much more complicated, change the used authentication flow, or don't work in some edge cases...

What the you all think we should do about this ? @mraible @vishal423

mraible:
Can we convince Keycloak to add the port offset back in? Is there an issue
to track its removal?

yelhouti:
@mraible I don't think they should, the only reason our code was working is because we where using a "hack". I think its our code that needs fixing... Maybe there is a way to retrieve the IP of the keycloak container and make keycloak resolve to that instead of 127.0.0.1 which would be a correct way of doing this.

yelhouti:
@mraible 03-system.sh is the one pointing keycloak to 127.0.0.1 we could fix the ip of keycloak an change 03-system.sh. Or we could run a similar script once keycloak has started and the ip is read using docker container inspect | grep ...

mraible:
I agree the current requirement to edit your hosts file isn't ideal. However, it's only required when running everything with Docker or in CI. I've yet to figure out a better solution but would love to find one!

yelhouti:
As I mentioned, this can be achieved by specifying a different URL for server to server communication than the one used to redirect the user to... But for now what d you think I should do. Add a network to app.yml and set a fixed ip to keycloak or retrieve it and set to to host to just before running the e2e tests ?

mshima:
Making the keycloak host dynamic will be a pain for development.

Add `--http-port 9080` to entrypoint.
Tried locally seem to work:
```
17 14:25:33,330 INFO  [io.quarkus] (main) Keycloak 18.0.0 on JVM (powered by Quarkus 2.7.5.Final) started in 13.276s. Listening on: http://0.0.0.0:9080
```

yelhouti:
@mshima this is a hack, I would rather configure a network in app.yaml and set the IP in the compose file to a fixed one, and tell people to add it to /etc*hosts.
Also this is not needed for development since in development the app is not started using docker-compose so using localhost:9080 works perfectly.

yelhouti:
Also. the only raison I went against using a network is to avoid conflicts between project.

mshima:
> Also this is not needed for development since in development the app is not started using docker-compose so using localhost:9080 works perfectly.

You are only considering `./mvnw`.
It's possible to develop frontends:
```
docker-compose src/main/docker/app.yml
npm start
```

> ```shell
> sudo docker container inspect docker-compose_keycloak_1 | jq -r '.[0].NetworkSettings.Networks."docker-compose_default".IPAddress' | echo "`cat -` keycloak" | sudo tee -a /etc/hosts
> ```

I really don't think this is less hacky than the http port change.
Works on macOS? What about windows?

Before merging you need to update:
- documentation https://www.jhipster.tech/docker-compose/#7
- ionic blueprint https://github.com/jhipster/generator-jhipster-ionic/blob/b9c96386b28c459722732e4c50d2dd08052994f9/.github/workflows/ionic.yml#L107-L110.
- svelte blueprint https://github.com/jhipster/generator-jhipster-svelte/blob/d1a88ebd581e7f346791ea0041d2b598635e9c90/.github/workflows/applications.yml#L418
- Check every official blueprint
- Create descriptive text for breaking change to put at the release notes.

IMO you are trying to fix something that's not broken.

generators/docker-compose/files.js
```diff
@@ -40,7 +40,6 @@ function writeFiles() {
     writeKeycloakFiles() {
       if (this.authenticationType === OAUTH2 && this.applicationType !== MICROSERVICE) {
         this.template('realm-config/jhipster-realm.json.ejs', 'realm-config/jhipster-realm.json');
-        this.template('realm-config/jhipster-users-0.json.ejs', 'realm-config/jhipster-users-0.json');
```
mshima, line 43: This file should be removed from existing projects.
They will fail if not removed.

generators/server/files.js
```diff
@@ -356,11 +356,7 @@ const baseServerFiles = {
     {
       condition: generator => generator.authenticationTypeOauth2 && !generator.applicationTypeMicroservice,
       path: DOCKER_DIR,
-      templates: [
-        'keycloak.yml',
-        { file: 'config/realm-config/jhipster-realm.json', renameTo: () => 'realm-config/jhipster-realm.json' },
-        { file: 'config/realm-config/jhipster-users-0.json', method: 'copy', renameTo: () => 'realm-config/jhipster-users-0.json' },
```
mshima, line 362: Cleanup the file from existing projects.

mshima (changes requested):

mshima:
MacOS:
```
sudo docker container inspect docker-compose_keycloak_1
[]
Error: No such container: docker-compose_keycloak_1
```

It's started as docker-compose-keycloak-1

```
sudo docker container inspect docker-compose_keycloak_1 | jq -r '.[0].NetworkSettings.Networks."docker-compose_default".IPAddress' | echo "`cat -` keycloak" | sudo tee -a /etc/hosts
zsh: command not found: jq
Error: No such container: docker-compose_keycloak_1
```

yelhouti:
@mshima here you go: 

> documentation https://www.jhipster.tech/docker-compose/#7 

https://github.com/jhipster/jhipster.github.io/compare/main...yelhouti:patch-3
> ionic blueprint https://github.com/jhipster/generator-jhipster-ionic/blob/b9c96386b28c459722732e4c50d2dd08052994f9/.github/workflows/ionic.yml#L107-L110.

https://github.com/jhipster/generator-jhipster-ionic/pull/691

> svelte blueprint https://github.com/jhipster/generator-jhipster-svelte/blob/d1a88ebd581e7f346791ea0041d2b598635e9c90/.github/workflows/applications.yml#L418

https://github.com/jhipster/generator-jhipster-svelte/pull/1168

> native blueprint https://github.com/jhipster/generator-jhipster-native/blob/054dddd3a097b72062d2555cd01a234cda147ebf/.github/workflows/jdl.yml#L83

https://github.com/jhipster/generator-jhipster-native/pull/50

> Check every official blueprint

How do I know the others, also it is a one line change, it could be easily fixed by the blueprint maintainer using a copy paste.

yelhouti:
> It's started as docker-compose-keycloak-1
Seems like macos uses a different naming convention, shouldn't be a problem though right ?

generators/docker-compose/files.js
yelhouti: @mshima very good point I didn't think about it.

Though I am afraid removing it automatically would create problems for people who added users manually to that file and need to copy them to the other file.

What do you think ?

generators/docker-compose/templates/realm-config/jhipster-realm.json.ejs
```diff
@@ -2492,12 +2492,70 @@
...
+  "keycloakVersion": "18.0.0",
   "userManagedAccessAllowed": false,
   "clientProfiles": {
     "profiles": []
   },
   "clientPolicies": {
     "policies": []
-  }
+  },
+  "users": [
```
vishal423, line 2503: any specific reason to combine users data alongside realm?

yelhouti: @vishal423 splitting not allowed anymore starting from keylcoak 17

vishal423:
I was going over Keycloak migration guide and seems they have some changes on logout endpoint. @yelhouti, can you confirm the application logout works correctly with this PR? 

Also, can you confirm no realm configurations update available in this version as well (last time you confirmed for v17, I guess)

yelhouti:
> I was going over Keycloak migration guide and seems they have some changes on logout endpoint. @yelhouti, can you confirm the application logout works correctly with this PR?

No I can't :D thank for pointing this out fixing it.

> Also, can you confirm no realm configurations update available in this version as well (last time you confirmed for v17, I guess)

Yes I am pretty sure I have checked :)

yelhouti:
@mraible waiting for your review

mraible:
I'll give it a try in a few hours. I'm traveling this week in Iceland.

yelhouti:
@mraible please no rush, this can wait for next week. Enjoy your vacation !

mraible:
@yelhouti I'm not on vacation. This is work. ;0)

I tested it with `jhipster jdl reactive-ms` on my M1 since that's all I have with me.

It works when running all the subsystems with Docker and the gateway + microservices with Gradle. However, when you build Docker images and try to run everything with `docker compose up`, it fails.

It tries to redirect to `http://keycloak:8080/`, which doesn't exist. Keycloak is running at 9080, but both the JHipster Registry and the gateway app try to redirect to 8080.

yelhouti:
You need to change your host to point to the docker ip instead of 127.0.0.1 a commad for that is available in the docs PR

mraible:
@yelhouti Do you have a link to the docs PR or is it part of this one? If it's part of this one, can you tell me the file to look for changes in?

mshima:
I would prefer to keep the keycloak static.
Having to change the keycloak ip from time to time is bad.

yelhouti:
sudo docker container inspect docker-compose_keycloak_1 | jq -r '.[0].NetworkSettings.Networks."docker-compose_default".IPAddress' | echo "`cat -` keycloak" | sudo tee -a /etc/hosts

mraible:
IMO, this isn't a great developer experience because I can't memorize it. Nevertheless, I tried it:

```
sudo docker container inspect docker-compose_keycloak_1 | jq -r '.[0].NetworkSettings.Networks."docker-compose_default".IPAddress' | echo "cat - keycloak" | sudo tee -a /etc/hosts
```

It results in:

```
Error: No such container: docker-compose_keycloak_1
```

If I do a `docker ps`, it shows:

```
CONTAINER ID   IMAGE                               COMMAND                  CREATED          STATUS          PORTS                                            NAMES
fd9bfc31307b   blog                                "bash -c /entrypoint…"   15 minutes ago   Up 15 minutes   8081/tcp                                         docker-compose-blog-1
4dd3e779b620   store                               "bash -c /entrypoint…"   15 minutes ago   Up 15 minutes   8082/tcp                                         docker-compose-store-1
bc4acd679f4b   neo4j:4.4.6                         "tini -g -- /startup…"   15 minutes ago   Up 15 minutes   7473-7474/tcp, 7687/tcp                          docker-compose-blog-neo4j-1
b2d167b3ce8a   mongo:4.4.13                        "docker-entrypoint.s…"   15 minutes ago   Up 15 minutes   27017/tcp                                        docker-compose-store-mongodb-1
2ee88ba89092   jhipster/jhipster-registry:v7.3.0   "bash -c /entrypoint…"   15 minutes ago   Up 15 minutes   0.0.0.0:8761->8761/tcp                           docker-compose-jhipster-registry-1
7c4ad6afc410   gateway                             "bash -c /entrypoint…"   15 minutes ago   Up 15 minutes   0.0.0.0:8080->8080/tcp                           docker-compose-gateway-1
7247589108de   quay.io/keycloak/keycloak:18.0.0    "/opt/keycloak/bin/k…"   15 minutes ago   Up 15 minutes   0.0.0.0:9080->8080/tcp, 0.0.0.0:9443->8443/tcp   docker-compose-keycloak-1
58437475eacd   postgres:14.2                       "docker-entrypoint.s…"   15 minutes ago   Up 15 minutes   5432/tcp                                         docker-compose-gateway-postgresql-1
```

yelhouti:
@mshima we can do that we would have to use the same network in keycloak.yml and app.yml which not great. We should also be aware of conflicts between multiple projects... This is more flexible.
But whatever you guys decide

yelhouti:
For macOS it seems to be -1 not _1

vishal423:
@mraible, You can check exact name under `Names` column of `docker ps` output. From last pasted output, it seems to be `docker-compose-keycloak-1`

mraible:
This command does execute:

```
sudo docker container inspect docker-compose-keycloak-1 | jq -r '.[0].NetworkSettings.Networks."docker-compose_default".IPAddress' | echo "cat - keycloak" | sudo tee -a /etc/hosts
```

However, if I look in my `/etc/hosts/` file, I see:

```
cat - keycloak
cat - keycloak
cat - keycloak
```

This doesn't seem to be that easy. ;)

yelhouti:
It seems like `cat -` doesn't work the same in MacOS. Just docker run the first part to read the IP and set it manually in the file and it should work :)

yelhouti:
> @yelhouti Can you please explain why when I run Keycloak for the gateway (with `docker-compose -f src/main/docker/keycloak.yml up -d`), it can run on 9080. However, when I run with Docker Compose (from the `docker-compose` folder), it needs to run on 8080?

@mraible Sure !
I understand why you think that they run in different ports but in reality they don't. Here is what is happening:
docker run apps in a private network with it's own port management, dns resolution... In that network the keycloak container has a dns entry `keycloak` and an open port where the app runs on 8080.
All pods inside docker (in the same network, the default one when nothing is specified) can talk to keycloak using the "url" `keycloak:8080`. 
However your browser is running directly in the host, not inside docker and therefor can not talk to keycloak, unless the port is exposed. We do that by specifying `127.0.0.1:9080:8080` ie redirect all traffic coming to 127.0.0.1:9080 to this container's (keycloak) 8080 and now keycloak can be reached on 127.0.0.1:9080 .
But since in your /etc/hosts you have forced keycloak to resolve to 127.0.0.1 your browser can reach keycloak on keycloak:9080.
Why is using keycloak.yml and docker-compose.yml different.
When using keycloak.yml everything expect keycloak is running on the host, and keycloak is exposed for all component on 127.0.0.1:9080 So everything work great.
However, when using docker-compose.yml everything **expect the browser**  is running inside the private network and they can talk to it using: keycloak:8080. But the browser doesn't know how to resolve keycloak so it fails. To fix the browser, we just need to tell our machine where to resolve keycloak and everything should work great. How we so that is just by changing /etc/hosts `<PRIAVTE_KEYCLOAK_IP> keycloak` and we can reach it in the private network ip and port 8080. Of course we can still reach it in this case too on 127.0.0.1:9080 if it is exposed, or we can just choose not to expose it on a different port.

Finally keycloak:9080 should never work unless you still have the old `127.0.0.1 keycloak` in your /etc/hosts file.

mraible:
Can we make it so `docker-compose.yml` uses port 9080 like `keycloak.yml` does?

That way, your way (grab the IP from Docker and add it to the hosts file) and the current way both work.

yelhouti:
No it's not possible, because the app needs to talk to keycloak directly inside the private network. Going through the host is in theory possible but requires more tweaking than just modifying /etc/hosts like I do

mraible:
> Going through the host is in theory possible but requires more tweaking than just modifying /etc/hosts like I do

I'm just saying the current way is pretty much the same in that it makes the end user modify `/etc/hosts`. The new way in this PR just makes you calculate it every time instead of making it static.

yelhouti:
@mraible I can make it static just not 127.0.0.1 let me do that

yelhouti:
@mraible I fixed it to 172.31.0.200 and updated the docs accordingly https://github.com/jhipster/jhipster.github.io/compare/main...yelhouti:patch-3

I ll fix PRs on other repos once this is merged to the tests to pass

mraible:
I tried adding this IP to my `/etc/hosts` file. Then, I tried to run the following from the gateway:

```
docker-compose -f src/main/docker/keycloak.yml up -d
```

It fails with:

```
Error response from daemon: Invalid address 172.31.0.200: It does not belong to any of this network's subnets
```

yelhouti:
@mraible It this you issue is related to how networks in docker work/cached. can please try to docker-compose down before up ?

yelhouti:
FYI: since we want a fix IP we need to set the default subnet to ensure that the fixed IP is in it. And for all containers started with docker-compose to be able to communicate with each other we need to set the default network in all of them (this is why I had to change all the compose files).

yelhouti:
@mshima you requested changes, I replied do you agree with my assessment or do you think I should make the change you suggested ?

mraible:
I tried this PR on a new computer today:

```
Pulling keycloak (quay.io/keycloak/keycloak:18.0.0)...
18.0.0: Pulling from keycloak/keycloak
4752687a61a9: Pull complete
0344366a246a: Pull complete
bf8a68204bbb: Pull complete
c4db0a36467a: Pull complete
Digest: sha256:7f80b2c243c738761bb8702c16b7f30da40426604f2c18a5767492917ca886dc
Status: Downloaded newer image for quay.io/keycloak/keycloak:18.0.0
Creating docker_keycloak_1 ... error

ERROR: for docker_keycloak_1  Cannot start service keycloak: Invalid address 172.31.0.200: It does not belong to any of this network's subnets
```

I thought it might be because I didn't have `172.31.0.200` in my `/etc/hosts`. However, it still doesn't work after I added it. I'm just using `jhkeycloakup`, not Docker Compose. I tried removing all images from my machine, but that didn't help either. I'm using an Intel-based Mac.

generators/docker-compose/files.js
mshima: We should remove it the user should have it inside a repository.

mshima:
@yelhouti please revert keycloak to run at port 9080.
And open a new PR targeting v8 for the port change if you wish.

yelhouti:
@mshima I understand what you mean, although the approach of changing ports is wrong, I think its easier to do for most people, so sadly I think I will switch back to it.
@mraible I would gladly discuss what I think is going wrong, but since I am reverting to switching ports maybe you don't want to waster your time with this for now. Feel free to ask me anything if I am wrong.

yelhouti:
@mshima @mraible here you go :)

generators/cleanup.js
```diff
@@ -343,6 +343,10 @@ function cleanupOldFiles(generator) {
       generator.removeFile('.npmrc');
     }
   }
+  if (generator.isJhipsterVersionLessThan('7.8.2')) {
+    generator.removeFile(`${constants.DOCKER_DIR}realm-config/jhipster-users-0.json`);
```
mshima, line 347: Should go here:
https://github.com/jhipster/generator-jhipster/blob/main/generators/server/cleanup.js

generators/cleanup.js
```diff
@@ -343,6 +343,10 @@ function cleanupOldFiles(generator) {
...
+    generator.removeFile('realm-config/jhipster-users-0.json'); // docker-compose
```
mshima, line 348: Should create a cleanup task here:
https://github.com/jhipster/generator-jhipster/blob/dce15044b1f74a9345447ec8181c3945eb1da113/generators/docker-compose/files.js#L29

```
  cleanup() {
      this.removeFile('realm-config/jhipster-users-0.json');
  },
```

mshima (changes requested):

mraible:
I tested this today on my M1 and everything worked great. I have the following in my `/etc/hosts`.

```
127.0.0.1 keycloak
```

I tested with `jhipster jdl reactive-ms` and proved everything worked when starting Docker containers and each app using `./gradlew`. Then, I tested by running `docker compose up` in the `docker-compose` directory and running `npm run e2e`. All tests pass too.

mraible (approved):

yelhouti:
@mraible thanks a lot for you time :)

yelhouti:
@mshima I made the changes you requested for, could you check them please :)

vishal423 (approved):

mshima (approved):
Cleanup is fixed.

mshima:
Thanks @yelhouti 
The port change provides the best developer experience for now.

DanielFran:
Many thanks @yelhouti for the huge work.
Can you review the PR in other projects and the documentation, please?

yelhouti:
@DanielFran of course I will in few, but they can safely be closed

yelhouti:
@mshima indeed thank you for your feedback :) goes to prove that correct doesn't always mean better

yelhouti:
claiming bounty :) https://opencollective.com/generator-jhipster/expenses/78206

pascalgrimaud:
can https://github.com/jhipster/generator-jhipster/issues/17850 be closed ?
so the bounty can be claimed ?

mraible:
Can we make it so `docker-compose.yml` uses port 9080 like `keycloak.yml`
does?

That way, your way (grab the IP from Docker and add it to the hosts file)
and the current way both work.

yelhouti:
@mraible what is the exact change you are proposing please ? doesn't docker-compose.yml already use 9080 ?

mraible:
@yelhouti No changes need to be made as far as I know. When I made the comment above, it was using a different port.

PR #18449, opened 2022-04-23
Split server cleanup in multiple files

Tcharl:
This pr is meant to help the future split of server variant into multiple generator or sub folders
---

### DIFF
--- a/generators/server/cleanup-angular.js
+++ b/generators/server/cleanup-angular.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2013-2022 the original author or authors from the JHipster project.
+ *
+ * This file is part of the JHipster project, see https://www.jhipster.tech/
+ * for more information.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


DanielFran:
@Tcharl maybe rename method cleanupOldServerFiles to cleanupOldFiles?

DanielFran (approved):

Tcharl:
@DanielFran 
As I'm not sure about the future split I prefer to keep it like this.
i.e., the future modules split could contain only one `cleanup.js` file that will clean server, client and platform files for old versions. In this case we'll be happy to have a `cleanupoldserverfile` method instead of a one fits all I guess

generators/server/cleanup-reactive.js
```diff
@@ -0,0 +1,38 @@
+/**
```
mshima, line 1: If we split reactive, we may need to split every other tech into imperative/reactive.
We probably should keep it in server.

Tcharl:
Clear, I'll take it into account in the next PR (I've a 3 hours flight so will be fixed tonight)

generators/server/cleanup-angular.js
```diff
@@ -0,0 +1,37 @@
+/**
```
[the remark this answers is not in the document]
Tcharl: Not so sure for this one: in the future this method will be in the cleanup file of angular

generators/server/cleanup-reactive.js
Tcharl: I suggest to have a `clean-reactive` file for everything that is common, but use `if(reactive)`in the respective `cleanup-<tech>` for everything related to a particular technology (sql, angular, ...)

PR #18631, opened 2022-05-11
[JDL] Add otherEntityRelationshipName only if the relationship exists at otherEntity.

mshima:
- `otherEntityRelationshipName` is the relationshipName of the relationship at the otherEntity.
It should not be set if the otherEntity doesn't have the referred relationship.

- Filtering requires bidirectional relationships.
Adjust the templates to ignore one-to-one filtering if the relationship is unidirectional.

- Convert some missing jdl tests to snapshots

---


yelhouti:
Hi @mshima I am afraid this would break some of our blueprints, do you have time to talk about this please ?

mshima:
@yelhouti I suppose you are talking about the OneToOne relationship been bidirectional correct?

yelhouti:
@mshima I think also the otherEntityRelationshipName. I use it to know that I am working on the correct relationship by checking that they both match.

OmarHawk:
I also make use of that (I think) in my pr #18621 - so removing it would be not so nice. There, I do need to know in one side about the backreference from the other side, and I think this is what you get rid of here.

mshima:
> I use it to know that I am working on the correct relationship by checking that they both match.

@yelhouti main generator already does that.
I am cleaning this warning that happens at unidirectional one-to-one relationship:
https://github.com/jhipster/generator-jhipster/blob/a6e325217bba7f6d26d9a095f5e5de0299982d2e/utils/relationship.js#L77

> I can't be sure a 100% since I didn't rebase my blueprints to v8 yet. To see if you changes would work.

There is no v8 yet, and no timeline to start for now.

> There is this whole esm stuff that I need to understand and rebase on top of.

ESM changes the require/import and module.exports/export.
It's async now and may improve user experience by not requiring the entire dependency tree to load before executing.
You can start a new blueprint using the `generate-blueprint` command, it generates esm blueprint.

@OmarHawk it should not interfere, `otherEntityRelationshipName` will be there were it should be.

DanielFran (approved):

PR #18721, opened 2022-05-23
chore: Bypass svelte compiled components from reactive security conf

### LINKED ISSUE #18566: Support Svelte blueprint in a reactive stack backend
It is a placeholder issue to add support for the Svelte blueprint in a reactive stack backend application and eventually the micro-services. Micro-frontend support is out of the scope of this ticket

Support reactive stack monolithic, gateway in Svelte blueprint

vishal423:
Closes #18566 

----

A stop gap solution to allow svelte compiled components

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
@@ -215,7 +215,7 @@ public class SecurityConfiguration {
         // @formatter:off
         http
             .securityMatcher(new NegatedServerWebExchangeMatcher(new OrServerWebExchangeMatcher(
-                pathMatchers("/app/**", "/i18n/**", "/content/**", "/swagger-ui/**", "/v3/api-docs/**", "/test/**"),
+                pathMatchers("/app/**", "/_app/**", "/i18n/**", "/img/**", "/content/**", "/swagger-ui/**", "/v3/api-docs/**", "/test/**"),
                 pathMatchers(HttpMethod.OPTIONS, "/**")
             )))
             .csrf()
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
```diff
@@ -279,10 +279,12 @@ public class SecurityConfiguration {
...
+<%_ if (authenticationTypeOauth2) { _%>
             .pathMatchers("/api/auth-info").permitAll()
+<%_ } _%>
             .pathMatchers("/api/admin/**").hasAuthority(AuthoritiesConstants.ADMIN)
             .pathMatchers("/api/**").authenticated()
-<%_ if (applicationTypeGateway) { _%>
+<%_ if (applicationTypeGateway && microfrontend) { _%>
             // microfrontend resources are loaded by webpack without authentication, they need to be public
             .pathMatchers("/services/*/*.js").permitAll()
             .pathMatchers("/services/*/*.js.map").permitAll()
```
mshima, line 290: Swagger api should be exposed for non microfrontends.
```suggestion
<%_ if (applicationTypeGateway) { _%>
  <%_ if (microfrontend) { _%>
            // microfrontend resources are loaded by webpack without authentication, they need to be public
            .pathMatchers("/services/*/*.js").permitAll()
            .pathMatchers("/services/*/*.js.map").permitAll()
  <%_ } _%>
```

vishal423: I agree to some extent, but the current configurations don't point in that direction and are confusing. Current mapping protects the services `/v3/api-docs` endpoint with the ADMIN role. I initially thought of doing refactoring, but, hold onto that as we have another ticket to improve the imperative client security configurations and this can also be done along that or as a separate ticket.

mshima: I still don't get.
`/services/*/v3/api-docs` accessibility should be the same for gateways with microfrontends and without.
`.pathMatchers("/services/*/v3/api-docs").hasAuthority(AuthoritiesConstants.ADMIN)` should be removed in this case?

PR #18748, opened 2022-05-27
Make husky script executable

### LINKED ISSUE #18537: When adding a new project to Git, there's a Husky error

When you add a new project to Git, there's a warning from Husky:

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

We should set `.husky/pre-commit` as executable so users aren't surprised when they first try to check in their project.

##### **Reproduce the error**

Create a new project, then create a new repo on GitHub, then try to check it in and push it.

##### **JHipster Version(s)**

`main` branch as of May 1, 2022.

dbirks:
This fixes a error you get from husky after trying to commit for the first time after starting a new project:
```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

I think `_install()` is the right function to put this in. Tried putting it in `_postWriting()`, but the chmod was running too early it seems. Also tried `_end()`, but it was getting run after the git repo and initial commit were already made. But please let me know if there's a better place to put this.

And I formatted the log message similar to how I had seen it elsewhere:
https://github.com/jhipster/generator-jhipster/blob/v7.8.1/generators/kubernetes-helm/index.js#L177-L186

Fix #18537

---

### DIFF
--- a/generators/common/files.js
+++ b/generators/common/files.js
@@ -41,12 +41,10 @@ const commonFiles = {
         {
           file: 'gitattributes',
           renameTo: () => '.gitattributes',
-          method: 'copy',
         },
         {
           file: 'editorconfig',
           renameTo: () => '.editorconfig',
-          method: 'copy',
         },
         {
           file: 'sonar-project.properties',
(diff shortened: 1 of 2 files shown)


mshima:
@dbirks thanks for the PR.
I think we should fix it differently.
I will change your branch so you can keep as contributor.

PR #18756, opened 2022-05-28
Use correct env variable for Java options

egvimo:
The `JAVA_OPTS` environment variable is not picked up by the JVM.

Using `JAVA_TOOL_OPTIONS` instead.

Maybe this could also be changed for the Docker Compose files, so these options can be overridden by command line if necessary.

See [here](https://stackoverflow.com/questions/28327620/difference-between-java-options-java-tool-options-and-java-opts) for more details.

---

### DIFF
--- a/generators/kubernetes-knative/templates/service.yml.ejs
+++ b/generators/kubernetes-knative/templates/service.yml.ejs
@@ -147,7 +147,7 @@ spec:
             - name: SPRING_SLEUTH_PROPAGATION_KEYS
               value: "x-request-id,x-ot-span-context"
 <%_ } _%>
-            - name: JAVA_OPTS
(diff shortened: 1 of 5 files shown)


pascalgrimaud:
It depends on how you build your image.
Our entrypoint use JAVA_OPTS

https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/docker/jib/entrypoint.sh.ejs#L39

egvimo:
I'm using Jib to build the image and it contains the entrypoint, but the memory was still on the Java default. Therefore I changed it for our deployment. Another advantage of this environment variable is that it will be printed on startup.

But since `JAVA_OPTS` works for you, I'll close the PR.

PR #18780, opened 2022-06-01
fixes daily-build regressions related to npm scripts migration

mshima:
Fixes daily-build regressions related to npm scripts migration.

---

### DIFF
--- a/generators/server/index.js
+++ b/generators/server/index.js
@@ -30,7 +30,7 @@ const statistics = require('../statistics');
 const { defaultConfig } = require('../generator-defaults');
 const { JWT, OAUTH2, SESSION } = require('../../jdl/jhipster/authentication-types');
 
-const { CASSANDRA, COUCHBASE, ORACLE, SQL, MONGODB, NEO4J } = require('../../jdl/jhipster/database-types');
+const { CASSANDRA, COUCHBASE, ORACLE, SQL, MONGODB, NEO4J, MYSQL } = require('../../jdl/jhipster/database-types');
 const { CAFFEINE, EHCACHE, HAZELCAST, INFINISPAN, MEMCACHED, REDIS } = require('../../jdl/jhipster/cache-types');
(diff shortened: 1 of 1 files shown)


generators/server/index.js
```diff
@@ -459,6 +459,11 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
               'docker:db:await': `echo "Waiting for Couchbase to start" && wait-on -t ${WAIT_TIMEOUT} http-get://localhost:8091/ui/index.html && sleep 30 && echo "Couchbase started"`,
             });
           }
+          if (prodDatabaseType === MYSQL) {
```
Tcharl, line 462: I guess that now we should also add this line if devDatabaseType is Mysql too.
Maybe a liveliness probe in the docker-compose file and in the TestContainer implementation would be more efficient than an additional command

mshima: Testcontainer doesn’t have this problem.
Docker compose uses sleep at microservices.

A liveliness in entrypoint is nice but can be trick.

Tcharl: I remember that I made some liveliness probes for some jh services in one of my previous project, let me check

generators/server/index.js
```diff
@@ -490,13 +495,20 @@ module.exports = class JHipsterServerGenerator extends BaseBlueprintGenerator {
               scriptsStorage.set(`docker:${dockerConfig}:build`, `docker-compose -f ${dockerFile} build`);
               dockerBuild.push(`npm run docker:${dockerConfig}:build`);
             } else if (dockerConfig === 'jhipster-registry') {
-              dockerAwaitScripts.push(
+              if (authenticationTypeOauth2) {
+                dockerOthersUp.push('npm run docker:keycloak:await');
+              }
+              scriptsStorage.set(
+                `docker:jhipster-registry:await`,
```
DanielFran, line 502:
```suggestion
                'docker:jhipster-registry:await',
```

DanielFran (changes requested):

PR #18936, opened 2022-06-16
add .npmrc to force npm to use bash.

mshima:
---

### DIFF
--- a/generators/server/files.js
+++ b/generators/server/files.js
@@ -214,6 +214,11 @@ const baseServerFiles = {
       templates: ['entrypoint.sh'],
     },
   ],
+  node: [
+    {
+      templates: ['.npmrc'],
+    },
+  ],
   packageJson: [
     {
       condition: generator => generator.skipClient,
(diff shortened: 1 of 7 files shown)


DanielFran:
@mshima it will be needed to put back windows tests using prod profile?

mshima:
@DanielFran prod tests won’t work because they need docker now.
There is no way to install docker on GitHub action’s windows right now.

OmarHawk:
Puh, this took me a few hours to figure out that this here is the reason the new version actually broke my local build on windows, which was previously working and was set up with the default (git bash in my case). In my scenario, this change here now makes use of wsl bash, which does not work properly for me, as 

"bash -c node" 

will fail, due to the non-interactive shell not having access to PATH properly by default:

https://unix.stackexchange.com/questions/439151/passing-command-to-bash-from-cmd-exe-wsl-not-working

I "fixed" this now by removing the entry in the generated local .npmrc for now. 

Maybe this is some local mis-configuration of my wsl or some co-incidential changes on another place, idk, but in case someone else has the same trouble, this could serve as a starting point for trouble-shooting. For me, stuff is working again. ;-)

PR #19200, opened 2022-07-20
Revert "workaround intrusive warning due to deprecated bootstrap property"

mshima:
Reverts jhipster/generator-jhipster#19159

bootstrap@5.2.0 has been released.


atomfrede (approved):

mshima:
The problem was there for a couple of months.
I was waiting but decided to workaround due to 3.9.0.
So the timing was perfect. 😄

Issue #19267, opened 2022-07-29
authorities are not updated into the database


sbouchex:
##### **Overview of the issue**

I'm using jhipster with keycloak.

If the user does not exist in the local database, it is created (according to UserService.syncUserWithIdP code) and the user authorities are stored into the jhi_user_authority table.
If the user exists in the local database, user details (name, timestamp) are updated (according to UserService.syncUserWithIdP code) but the user authorities are not updated in the jhi_user_authority table.

So if the user has its authorities changed (due to keycloak user change), the user is not getting the new authorities at the next login.

##### **Motivation for or Use Case**

The behavior is different if the user is already existing in the local database or not.

##### **Suggest a Fix**

Replace: updateUser(user.getFirstName(), user.getLastName(), user.getEmail(), user.getLangKey(), user.getImageUrl());
by: 
userRepository.save(user);
this.clearUserCaches(user);

##### **JHipster Version(s)**

7.6.1

##### **JHipster configuration**

oauth2 configuration

##### **Browsers and Operating System**

Chrome

sbouchex:
Related to this change ? #7351

OmarHawk:
Bug Bounty Claim https://opencollective.com/generator-jhipster/expenses/117439

DanielFran:
@OmarHawk approved

PR #19301, opened 2022-08-01
use caffeine cache for caching user information

### LINKED ISSUE #18868: Caching of user information is not optimized for scale

When using OAuth 2.0, there is logic that will look up the user's information (from the OIDC `/userinfo` endpoint) to popular the user's first and last names. It caches the lookups by the `sub` claim so it only happens once per user. However, it's not optimized for scale.

```java
// Warning from Rob Winch: caching in a hashmap could be bad since there is no expiry.
// Changing attributes will never get updated, and it can grow indefinitely (run out of memory).
// Ideally, we could use @Cacheable(cacheNames="users", key="#token") on the `enrich()` method,
// but @Cacheable doesn't work on reactive types. See github.com/spring-projects/spring-framework/issues/17920
private final Map<String, Mono<Jwt>> users = new ConcurrentHashMap<>();
```

Both Spring MVC (in `CustomClaimConverter.java`) and WebFlux (in `SecurityConfiguration`) use such a map.

**NOTE**: This is not an issue for regular JHipster apps. It's only an issue when you're using JHipster as an OAuth 2.0 resource server and passing in a JWT access token to authenticate. Currently, this is only done by the Ionic and React Native blueprints. Keycloak will never do the lookup since `given_name` and `family_name` are included claims in the access token. For Auth0 and Okta, this auto-lookup makes things easier to configure.

This issue is to track converting the existing implementation to use the same cache abstraction that the user chooses. 

##### **Reproduce the error**

There is no error, it's just a hack currently.

##### **Related issues**

- https://github.com/jhipster/generator-jhipster/issues/13461
- https://github.com/jhipster/generator-jhipster/issues/11767

##### **Suggest a Fix**

Use a more optimized caching solution.

Related to https://github.com/spring-projects/spring-framework/issues/17920.

atomfrede:
This PR replaces the unbounded hashmap to cache user information with a simple in memory/node local caffeine cache which seems to be simplest and easiest to use caching solution. 

We could of course use the selected caching type, but it seems to complicated or use some distributed solution for even better scaleability, but I think having each node to do the cache lookup is fine and everything else would be too much.

closes #18868

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
@@ -134,6 +134,9 @@ import reactor.core.publisher.Mono;
   <%_ if (!applicationTypeMicroservice) { _%>
 import org.springframework.web.reactive.function.client.WebClient;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
 import java.util.Arrays;
 import java.util.Map;
   <%_ } _%>
@@ -141,7 +144,6 @@ import java.util.Map;
 import java.util.HashSet;
 import java.util.Set;
   <%_ if (!applicationTypeMicroservice) { _%>
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
   <%_ } _%>
 <%_ } _%>
@@ -170,8 +172,13 @@ public class SecurityConfiguration {
 
     private final ReactiveClientRegistrationRepository clientRegistrationRepository;
 
-    // todo: optimize for scale https://github.com/jhipster/generator-jhipster/issues/18868
-    private final Map<String, Mono<Jwt>> users = new ConcurrentHashMap<>();
+    // See https://github.com/jhipster/generator-jhipster/issues/18868
+    // We don't use a distributed cache or the user selected cache implementation here on purpose
+    private final Cache<String, Mono<Jwt>> users = Caffeine.newBuilder()
+        .maximumSize(10_000)
+        .expireAfterWrite(Duration.ofHours(1))
+        .recordStats()
+        .build();
 
   <%_ } _%>
 <%_ } _%>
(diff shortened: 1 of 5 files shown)


mshima:
I wonder if it's possible to change syncWithIdp to save only the user id and integrate with this cache to avoid https://github.com/jhipster/generator-jhipster/issues/19267

generators/server/templates/src/main/java/package/config/SecurityConfiguration_reactive.java.ejs
```diff
@@ -425,7 +432,7 @@ public class SecurityConfiguration {
                     return Mono.just(jwt);
                 }
                 // Retrieve user info from OAuth provider if not already loaded
-                return users.computeIfAbsent(jwt.getSubject(), s -> {
+                return users.get(jwt.getSubject(), s -> {
```
mraible, line 435: Does this method work the same as the previous one did? That is, it checks the cache and if it doesn't exist, then it does the lookup.

atomfrede: Yes. From the doc:

`Returns the value associated with the key in this cache, obtaining that value from the mappingFunction if necessary. This method provides a simple substitute for the conventional "if cached, return; otherwise create, cache and return" pattern.`

mraible (dismissed):

atomfrede:
> I wonder if it's possible to change syncWithIdp to save only the user id and integrate with this cache to avoid #19267

Will have a look afterwards. At sounds like a good idea. Not sure to have the authorities in this cache, but changing the role/authorities in idp should be reflected in our user database.

atomfrede:
Will have a look why the build is failing with infinispan

atomfrede:
Infinispan brings its own version of caffeine. Enforcing the latest version brings in an incompatible version. So we can use caffeine (from infinispan in this case) in custom claim converter.

mraible (approved):

atomfrede:
Bounty claimed https://opencollective.com/generator-jhipster/expenses/89309

DanielFran:
@atomfrede approved

PR #19363, opened 2022-08-07
configure spring-security to send client-related http security header…

### LINKED ISSUE #17035: Send ContentSecurityPolicy (CSP) header only if client is generated?
When using JHipster for backend-only development, or when splitting frontend and backend, the CSP header sent by Spring makes no sense. 
[Policy Delivery](https://www.w3.org/TR/CSP2/#policy-delivery) seems to make sense only for html apps running in the browser; in fact, another way to specify CSP is by setting an HTML `meta` element (as defined [here](https://www.w3.org/TR/CSP2/#delivery-html-meta-element)).

It should be the web tier serving the frontend application (web server, load balancer, reverse proxy, etc...) to deliver this header.

> ... the CSP is rather static and can be delivered from [web application tiers](https://en.wikipedia.org/wiki/Multitier_architecture) above the application, for example on [load balancer](https://en.wikipedia.org/wiki/Load_balancer) or [web server](https://en.wikipedia.org/wiki/Web_server).
> 
> &mdash; [Wikipedia](https://en.wikipedia.org/wiki/Content_Security_Policy)

The CSP header is set in the following line of code, during the App bootstrap:

https://github.com/jhipster/generator-jhipster/blob/b8c76e52cb8377158d9a0b32c44796b7a4555b8e/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L232

The value is read from the `application.yml` file:

https://github.com/jhipster/generator-jhipster/blob/91b29f57589a5b060ebeacd429361dfd924127a4/generators/server/templates/src/main/resources/config/application.yml.ejs#L333-L337

N.B.: The fact that the CSP header content depends on the client option of using a theme could be an indication that this aspect is a frontend-only matter.

**Propose a fix**: I would get rid of the CSP header set + config property altogether in the case where the client is not generated.

nickshoe:
…s only if client is generated

also generate the corresponding csp header value in the application config file only if client is generated

Fix #17035

---

### DIFF
--- a/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
+++ b/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
@@ -91,7 +91,9 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import <%= packageName %>.security.oauth2.CustomClaimConverter;
 <%_ } _%>
+<%_ if(!skipClient) { _%>
 import org.springframework.security.web.header.writers.ReferrerPolicyHeaderWriter;
+<%_ } _%>
 <%_ if (!applicationTypeMicroservice) { _%>
 import org.springframework.web.filter.CorsFilter;
 <%_ } _%>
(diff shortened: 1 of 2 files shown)


Tcharl:
LGTM when tests will pass

atomfrede (approved):

PR #19404, opened 2022-08-12
fix npmw.cmd failing to set the local node.js up

vgdm:
When Node.js is not already pre-installed globally on the system Windows building a generated project that relies on Node.js tooling for frontend fails. This is caused by NPM scripts not being able to locate Node.js, because the local installation is not added to the PATH by npmw.cmd. As the result NPM scripts that rely on Node.js fail to run.

The fix makes sure that the local Node.js installation, when present, is added to the PATH for all downstream scripts to see. The local Node.js is added temporarely,  for the duration of CMD session, leaving the global installation still accessible elsewhere.

**To recreate the issue, before applying this fix:**
  1. Make sure there is no Node.js installed on a Windows machine or global PATH variable does NOT contain path to a Node.js installation.
  2.  Take a simple generated project with an Angular frontend.
  3.  Attempt building the project using mvnw.cmd

**Expected result:**
  The project is built.

**Actual result:**
  npmw.cmd fails.

Following the fix, the same scenario should work as expected.

  

---

### DIFF
--- a/generators/server/templates/npmw.cmd
+++ b/generators/server/templates/npmw.cmd
@@ -1,29 +1,31 @@
 @echo off
 
-@setlocal
+setlocal
 
 set NPMW_DIR=%~dp0
 
-if exist "%NPMW_DIR%\mvnw.cmd" (
-  set NODE_EXE=""
-  set NPM_EXE=%NPMW_DIR%\target\node\npm.cmd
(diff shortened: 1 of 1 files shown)


mshima:
Looks good, but I don't have a Windows machine so maybe @OmarHawk could test.

mshima:
Thanks @vgdm for the contribution

vgdm:
@mshima, my pleasure!

PR #19427, opened 2022-08-15
remove unnecessary optional property

ertunga:
remove unnecessary optional properties from angular templates that cause warnings in the console

Fix #19373

---

Please make sure the below checklist is followed for Pull Requests.

- [ ] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [ ] Tests are added where necessary
- [ ] The JDL part is updated if necessary
- [ ] [jhipster-online](https://github.com/jhipster/jhipster-online) is updated if necessary
- [ ] Documentation is added/updated where necessary
- [ ] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (below reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

Issue #19373, opened 2022-08-08
Angular: NG8107 warning when compiling frontend in list and detail view with referenced other entities


OmarHawk:
##### **Overview of the issue**
In the following line, we do produce a ? character right before the element accessor:

https://github.com/jhipster/generator-jhipster/blob/1668772197f1e5fbc21d43b07abfa833c2e66a2d/generators/entity-client/templates/angular/src/main/webapp/app/entities/list/entity-management.component.html.ejs#L149

This produces such warnings during frontend build:

```
[INFO] Warning: src/main/webapp/app/entities/stream-data/list/stream-data.component.html:137:108 - warning NG8107: The left side of this optional chain operation does not include 'null' or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator.
[INFO] 
[INFO] 137               <a [routerLink]="['/stream-router', streamData.routerid.id, 'view']">{{ streamData.routerid?.info }}</a>
[INFO]                                                                                                                ~~~~
[INFO] 
[INFO]   src/main/webapp/app/entities/stream-data/list/stream-data.component.ts:18:16
[INFO]     18   templateUrl: './stream-data.component.html',
[INFO]                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[INFO]     Error occurs in the template of component StreamDataComponent.
```

Another place where the same thing is done is this:

https://github.com/jhipster/generator-jhipster/blob/1668772197f1e5fbc21d43b07abfa833c2e66a2d/generators/entity-client/templates/angular/src/main/webapp/app/entities/detail/entity-management-detail.component.html.ejs#L91

Producing such a warning
```
[INFO] Warning: src/main/webapp/app/entities/stream-log/detail/stream-log-detail.component.html:32:108 - warning NG8107: The left side of this optional chain operation does not include 'null' or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator.
[INFO] 
[INFO] 32             <a [routerLink]="['/stream-data', streamLog.streamData.uuid, 'view']">{{ streamLog.streamData?.uuid }}</a>
[INFO]                                                                                                               ~~~~
[INFO] 
[INFO]   src/main/webapp/app/entities/stream-log/detail/stream-log-detail.component.ts:8:16
[INFO]     8   templateUrl: './stream-log-detail.component.html',
[INFO]                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #19431, opened 2022-08-15
Revert "remove unnecessary optional property"

mshima:
Reverts jhipster/generator-jhipster#19427 once https://github.com/angular/angular/issues/46918 has been resolved.

Related to https://github.com/jhipster/generator-jhipster/issues/19373


ertunga:
@mshima I think we can merge this without waiting. Mentioned lines in the issue are wrapped with if checks that make sure the fields are there. Even the fields can be null or undefined, because of the ngIf ?. can be removed here.

Here is an example, we can remove the question mark here.

<div *ngIf="post.blog"> <a [routerLink]="['/blog', post.blog.id, 'view']">{{ post.blog?.name }}</a> </div>

Issue #19269, opened 2022-07-29
Add support to run a local blueprint


mshima:
##### **Overview of the feature request**

Blueprints are powerful to customize applications.
But it's not easy to create a blueprint package, publish (or install using git) to use it.
We may allow our users to implement a local blueprint at `.blueprint` folder.
This may improve interest in our blueprint system.

##### **Motivation for or Use Case**

Improve developer experience.

##### **Related issues or PR**

- [x] Checking this box is mandatory (this is just to show you read everything)

mshima:
Documentation implemented at https://github.com/jhipster/jhipster.github.io/pull/1223.

mshima:
@DanielFran bounty claimed https://opencollective.com/generator-jhipster/expenses/112872.

DanielFran:
@mshima approved

PR #19451, opened 2022-08-16
Implement local blueprint support.

mshima:
Resolves https://github.com/jhipster/generator-jhipster/issues/19269.

---


mraible:
@mshima Is there documentation that needs to be updated to show how to use this feature?

vishal423:
I am not sure about the usefulness of this feature.  As I understand, it's trying to remove the requirement to publish a blueprint. While doing so, it's encouraging to move and maintain the blueprint code within the generated application. IMO, it defies the purpose of a blueprint to reuse it across multiple applications. Further, I don't see the npm publishing as an issue as there are ways (like `npm link`) to handle that and are already used in other maintained blueprints during development.

vishal423:
How about identifying the pain points to improve the overall maintenance of blueprints? 

To avoid blueprint publishing, can we introduce an option like `--local <path-to-blueprint>`?

mshima:
> How about identifying the pain points to improve the overall maintenance of blueprints?
> 
> To avoid blueprint publishing, can we introduce an option like `--local <path-to-blueprint>`?

Publishing is not required, you can install using git or path.
But you still need to:
- maintain prettier, eslint
- maintain another repository
- keep another package.json updated
- make sure the blueprint's generator-jhipster is synchronized with the project generator-jhipster
- make sure to keep the blueprint git revision updated to keep the project reproducible

Most of the tasks are minor, but makes the developer experience bad.

vishal423:
I disagree that maintaining a blueprint in a separate repository is a bad developer experience. Infact, the opposite is a bad design choice and violates the single responsibility principle. It also increases the maintenance overhead as you need to keep each application template in sync. You can easily manage dependency updates using dependabot or equivalent tools. I don't see any challenge in that.

mshima:
> I disagree that maintaining a blueprint in a separate repository is a bad developer experience. Infact, the opposite is a bad design choice and violates the single responsibility principle. It also increases the maintenance overhead as you need to keep each application template in sync. You can easily manage dependency updates using dependabot or equivalent tools. I don't see any challenge in that.

Yes if you want to reuse the blueprint, but in this case it's application specific.

vishal423:
As it sounds, you want a provision to override the generated code with custom templates. Doing via current approach may be useful for a subset of specialized use cases, but it's not trivial. A blueprint is meant to be created once and reused across multiple applications. Following is what I propose:
- Create an option `--local <path-to-blueprint-template>` to allow you to specify the blueprint directory to be used to pick up and override the generated code. If you want this to be generated inside the current application, you can specify the path as `./.blueprint`.
- You can also specify the path to a full-fledged externalized blueprint as this option.
- Store the user-specified path inside `.yo-rc.json`. If the path is empty/no, then, the local blueprint option is not enabled for the current application. Generated application code need not be aware of the blueprint.
- When composed with external blueprints, a local blueprint ensures high priority.
- Create a sample repository within the `Jhipster` organization using this option with a useful override to explain/demonstrate the functional use.

mraible:
I'm thinking of doing a Twitter poll to help answer this question. How does this sound?

> Should JHipster support local blueprints that only apply to a single app?
> 1. Yes
> 2. No, all blueprints should require publishing
> 3. Blueprints are so confusing, I'm not sure
> 4. Don't share, show answers

Please provide alternatives if my question and answers won't help.

gmarziou:
@mraible I don't understand choice #4

mraible:
> @mraible I don't understand choice 4

I had a typo. I meant to say "Don't care".

gmarziou:
I see, never thought that to see survey results you have to vote. I misinterpreted the word "answers" in this context.

mraible:
I created and published the poll. I decided to eliminate option 3 about blueprints being confusing. 

https://twitter.com/mraible/status/1563257288504123392

mraible:
There's still 40 minutes left on my poll, but it seems like most people are in favor of local blueprints.

deepu105:
Here is my take on this.

1. I agree that for some usecases it might be useful to have a local blueprint that adds some customization.
2. But on the other hand, given that the popularity of blueprints isn't exactly what I expected it to be, the feature might not be used a lot by regular users. So we need to look at the maintenance burden of the feature. I didn't check the code, but I expect that this will mostly use the same code as normal blueprints except gor bootstrapping from the local path
3. I'm not against this but I don't expect it to be used a lot

mraible:
@mshima As @deepu105 mentioned, do you see this as a maintenance burden? If not, let's merge and release 7.9.3!

mshima:
It just bootstrapping local path like @deepu105 said, I don't see as a maintenance burden.

deepu105:
I can do the release if @DanielFran is busy

mshima:
Can this be merged? Or should we wait for a review?

mraible:
@deepu105 Can you please review?

deepu105:
> @deepu105 Can you please review?

I'll take a look this week

generators/generate-blueprint/generator.mjs
```diff
@@ -240,8 +275,8 @@ export default class extends BaseBlueprintGenerator {
 
   get postWriting() {
     return {
-      postWriting() {
-        if (this.shouldSkipFiles()) return;
+      packageJsom() {
```
deepu105, line 278: Shouldn't this be packageJson?
```suggestion
      packageJson() {
```

deepu105 (dismissed):
I didn't nitpick. Looks good overall except for a typo

PR #19489, opened 2022-08-21
Add translation keys for two translation-not-found errors on the admi…

mplawley:
Two translation keys were missing on the admin health panel across all i18n-related `health.json` files. The two keys with their values were added. It appears that no conditional logic is needed for these keys since they seem related to Spring itself.

In the contribution guidelines, I didn't see whether it is customary to first open and issue before opening a PR; it appears that opening a PR not linked to an issue is acceptable. If this is incorrect, please let me know so I can fix.

- [X] The JDL part is updated if necessary
- [X] [jhipster-online](https://github.com/jhipster/jhipster-online) is updated if necessary

Prior to fix:
i18n-bug-screenshot

After fix on this branch:
Fixed-i18n-bug-screenshot

### DIFF
--- a/generators/languages/templates/src/main/webapp/i18n/al/health.json.ejs
+++ b/generators/languages/templates/src/main/webapp/i18n/al/health.json.ejs
@@ -48,6 +48,8 @@
             "mail": "Email",
             "livenessState": "Liveness state",
             "readinessState": "Readiness state",
+            "discoveryComposite": "Discovery Composite",
+            "refreshScope": "Refresh Scope",
             "ping": "Application"<% if (searchEngineElasticsearch) { %>,
             "elasticsearch": "Elasticsearch"<% } %><% if (databaseTypeSql && !reactive) { %>,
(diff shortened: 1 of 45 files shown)


mshima:
PR is ok for obvious fixes.
In this case the fix looks something else, those keys are optional:
https://github.com/jhipster/generator-jhipster/blob/045e1841226e237aa62aeae631b0a8a559266fca/generators/languages/templates/src/main/webapp/i18n/en/health.json.ejs#L35-L37

mshima:
Is language specific?
Or you should post a way to reproduce.

mplawley:
> Is language specific? Or you should post a way to reproduce.

Thank you for your reply. I opened an issue referencing this PR to show steps to reproduce:

https://github.com/jhipster/generator-jhipster/issues/19491

mraible (approved):

mshima:
@mplawley thanks for the PR. I will close because the error is at serviceDiscoveryType property.
As workaround change serviceDiscoveryType from `no` to false.

PR #19559, opened 2022-08-28
Set up TypeScript

emilpaw:
Initial set up of TypeScript.

---

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
     "node": true,
     "es2022": true
   },
-  "extends": ["airbnb-base", "plugin:prettier/recommended", "'plugin:@typescript-eslint/recommended'"],
+  "extends": ["airbnb-base", "plugin:prettier/recommended", "plugin:@typescript-eslint/recommended"],
   "parser": "@typescript-eslint/parser",
   "plugins": ["mocha", "prettier", "chai-friendly", "@typescript-eslint"],
   "settings": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,8 +3,9 @@
     "node": true,
     "es2022": true
   },
-  "extends": ["airbnb-base", "plugin:prettier/recommended"],
-  "plugins": ["mocha", "prettier", "chai-friendly"],
+  "extends": ["airbnb-base", "plugin:prettier/recommended", "'plugin:@typescript-eslint/recommended'"],
+  "parser": "@typescript-eslint/parser",
+  "plugins": ["mocha", "prettier", "chai-friendly", "@typescript-eslint"],
   "settings": {
     "import/core-modules": ["generator-jhipster", "generator-jhipster/support"]
   },
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,15 @@
       "env": {
         "mocha": true
       }
+    },
+    {
+      "files": ["**/*.{c,m,}js", "**/*.{c,m,}js"],
+      "rules": {
+        "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-empty-function": "off",
+        "@typescript-eslint/no-unused-vars": "off",
+        "@typescript-eslint/no-this-alias": "off"
+      }
     }
   ],
   "rules": {
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ test/temp/
 .project
 .settings
 test/**/*.png
+dist
--- a/cli/environment-builder.js
+++ b/cli/environment-builder.js
@@ -116,7 +116,7 @@ module.exports = class EnvironmentBuilder {
    */
   _lookupJHipster() {
     // Register jhipster generators.
-    this.env.lookup({ packagePaths: [path.join(__dirname, '..')], lookups: ['generators'] }).forEach(generator => {
+    this.env.lookup({ packagePaths: [path.join(__dirname, '../..')], lookups: ['dist/generators'] }).forEach(generator => {
       // Verify jhipster generators namespace.
       assert(
         generator.namespace.startsWith(`${CLI_NAME}:`),
--- a/cli/environment-builder.js
+++ b/cli/environment-builder.js
@@ -116,7 +116,17 @@ module.exports = class EnvironmentBuilder {
    */
   _lookupJHipster() {
     // Register jhipster generators.
-    this.env.lookup({ packagePaths: [path.join(__dirname, '../..')], lookups: ['dist/generators'] }).forEach(generator => {
+    const sourceRoot = path.basename(path.join(__dirname, '..'));
+    let packagePath;
+    let lookup;
+    if (sourceRoot === 'generator-jhipster') {
+      packagePath = path.join(__dirname, '..');
+      lookup = 'generators';
+    } else {
+      packagePath = path.join(__dirname, '../..');
+      lookup = `${sourceRoot}/generators`;
+    }
+    this.env.lookup({ packagePaths: [packagePath], lookups: [lookup] }).forEach(generator => {
       // Verify jhipster generators namespace.
       assert(
         generator.namespace.startsWith(`${CLI_NAME}:`),
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "./esm/priorities": "./lib/constants/priorities.mjs",
     "./esm/support": "./lib/support/index.cjs"
   },
-  "main": "lib/index.js",
+  "main": "dist/lib/index.js",
   "bin": {
     "jhipster": "cli/jhipster.js"
   },
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "jsdoc": "jsdoc --configure jsdoc-conf.json",
     "lint": "npm run eslint && npm run ejs-lint",
     "lint-fix": "npm run prettier:format && npm run eslint -- --fix",
+    "prepare": "npm run build",
     "prettier:check": "prettier --check \"{,**/}*.{js,ts,cjs,mjs,json,md,yml,java}\"",
     "prettier:format": "prettier --write \"{,**/}*.{js,ts,cjs,mjs,json,md,yml,java}\"",
     "pretest": "npm run lint",
--- a/package.json
+++ b/package.json
@@ -123,6 +123,8 @@
     "yeoman-generator": "5.7.0"
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^5.35.1",
+    "@typescript-eslint/parser": "^5.35.1",
     "chai": "4.3.6",
     "ejs-lint": "1.2.2",
     "eslint": "8.23.0",
@@ -142,6 +144,7 @@
     "sinon": "14.0.0",
     "sinon-chai": "3.7.0",
     "sort-keys": "4.2.0",
+    "typescript": "^4.8.2",
     "yeoman-assert": "3.1.1",
     "yeoman-test": "6.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "build": "npm run compile && npm run copy-files",
     "compile": "tsc",
     "completion": "tabtab install --name jhipster --auto",
-    "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|ejs|mjs)' dist/generators && cpy 'package.json' dist",
+    "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|mjs)' dist/generators && cpy 'package.json' dist",
     "ejs-lint": "mocha test/ejslint.js",
     "ejslint": "ejslint",
     "eslint": "eslint . --ext .js,.cjs,.mjs",
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "!**/*.spec.?(c|m)js"
   ],
   "scripts": {
+    "build": "tsc",
     "completion": "tabtab install --name jhipster --auto",
     "ejs-lint": "mocha test/ejslint.js",
     "ejslint": "ejslint",
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
   "scripts": {
     "build": "tsc",
     "completion": "tabtab install --name jhipster --auto",
+    "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|ejs|mjs)' dist/generators",
     "ejs-lint": "mocha test/ejslint.js",
     "ejslint": "ejslint",
     "eslint": "eslint . --ext .js,.cjs,.mjs",
@@ -91,6 +92,7 @@
     "chevrotain": "10.3.0",
     "commander": "9.4.0",
     "conf": "10.2.0",
+    "cpy-cli": "^4.2.0",
     "debug": "4.3.4",
     "didyoumean": "1.2.2",
     "ejs": "3.1.8",
--- a/package.json
+++ b/package.json
@@ -45,11 +45,11 @@
     "./lib/": "./dist/lib/",
     "./utils/": "./dist/utils/",
     "./package.json": "./package.json",
-    "./esm/cli": "./dist/cli/index.mjs",
-    "./esm/generators/*": "./dist/generators/*/esm.mjs",
-    "./esm/generators": "./dist/lib/constants/generators.mjs",
-    "./esm/priorities": "./dist/lib/constants/priorities.mjs",
-    "./esm/support": "./dist/lib/support/index.cjs"
+    "./esm/cli": "./cli/index.mjs",
+    "./esm/generators/*": "./generators/*/esm.mjs",
+    "./esm/generators": "./lib/constants/generators.mjs",
+    "./esm/priorities": "./lib/constants/priorities.mjs",
+    "./esm/support": "./lib/support/index.cjs"
   },
   "main": "dist/lib/index.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
     "!**/*.spec.?(c|m)js"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "npm run compile && npm run copy-files",
+    "compile": "tsc",
     "completion": "tabtab install --name jhipster --auto",
     "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|ejs|mjs)' dist/generators",
     "ejs-lint": "mocha test/ejslint.js",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   },
   "main": "dist/lib/index.js",
   "bin": {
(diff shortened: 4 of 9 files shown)


.eslintrc.json
```diff
@@ -3,8 +3,9 @@
     "node": true,
     "es2020": true
   },
-  "extends": ["airbnb-base", "plugin:prettier/recommended"],
-  "plugins": ["mocha", "prettier", "chai-friendly"],
+  "extends": ["airbnb-base", "plugin:prettier/recommended", "'plugin:@typescript-eslint/recommended'"],
```
mshima, line 6:
```suggestion
  "extends": ["airbnb-base", "plugin:prettier/recommended", "plugin:@typescript-eslint/recommended"],
```

Tcharl:
Not an expert, but I would love to start contributing to jhipster in typescript!

.gitignore
```diff
@@ -20,3 +20,4 @@ test/temp/
 .project
 .settings
 test/**/*.png
+dist
```
mshima, line 23: we should keep dist at git.
It allows install from git and we won’t need to build at every test/integration test

emilpaw: Are you sure we want them to be committed? How do we make sure they always represent the current state of the source files? e.g. someone forgets to commit the build files after doing some changes to the source files.

mraible: I did the same thing in ionic-appauth so it's easy for developers to depend on the `main` branch if they want to be on the bleeding edge.

A precommit hook handles keeping the files up-to-date:

https://github.com/wi3land/ionic-appauth/blob/master/.husky/pre-commit

emilpaw:
At this point running `npm run build` compiles all the JavaScript code files and moves the needed assets to the build directory `dist`.

Running it does not work yet though. First linking with `npm link` and then running `jhipster info` throws `AssertionError [ERR_ASSERTION]: Error on the registered namespace dist:app, make sure your folder is called generator-jhipster.`

@mshima Could you help me understand what the error exactly means and what the issue is?

mshima:
> At this point running `npm run build` compiles all the JavaScript code files and moves the needed assets to the build directory `dist`.

Why copy every js file to dist?
We can keep old structure with additional dist and src folders for now.

> Running it does not work yet though. First linking with `npm link` and then running `jhipster info` throws `AssertionError [ERR_ASSERTION]: Error on the registered namespace dist:app, make sure your folder is called generator-jhipster.`
> 
> @mshima Could you help me understand what the error exactly means and what the issue is?

This is related to the generators lookup process.
We need to customize the lookup process when moving generators to dist.
IMO we should do this when they are migrated to typescript.

.gitignore
mshima: @mraible not sure, but prepare script can be a problem at libs:
https://github.com/wi3land/ionic-appauth/blob/fa7574a7388b7a612c0efddb800ac95d15908a99/package.json#L11
I had some experience were it is called at a child project without support for husky breaking `npm install` at this project.
That’s the reason `generate-blueprint` command doesn’t generate husky by default:
https://github.com/mshima/generator-jhipster/blob/498979ef2ddbb46ec9a29078b5a5f4dff54cff98/generators/generate-blueprint/generator.mjs#L167

We can add tests to see if it’s updated at GitHub actions workflow.

emilpaw:
> Why copy every js file to dist?

TypeScript compiles all source files, JavaScript and TypeScript ones. It's not just copying the files. Not sure what the migration to TS could look like if we only compile a selected part of the program and leave the rest uncompiled. 

I was thinking about letting TypeScript compile the whole program and then piece after piece migrate the JavaScript files to TypeScript.

What exactly did you have in mind?


> > 
> 
> This is related to the generators lookup process. We need to customize the lookup process when moving generators to dist.

I am still not sure what the code does but I just naively replace `..` with `../..` in https://github.com/jhipster/generator-jhipster/blob/b741f654858098e52307bdf0471e12f3fe93a30a/cli/environment-builder.js#L119 and so far it seems to work. At least generating a project works fine.





> IMO we should do this when they are migrated to typescript.

If there is a way to only compile parts of the programm and we decide we prefer that over letting TypeScript compile all source files we could do this. Otherwise this is needed to be done within the scope of this PR.

mraible:
Might be relevant to this conversation:

https://blog.jim-nielsen.com/2018/installing-and-building-an-npm-package-from-github/

mshima:
> > Why copy every js file to dist?
> 
> TypeScript compiles all source files, JavaScript and TypeScript ones. It's not just copying the files. Not sure what the migration to TS could look like if we only compile a selected part of the program and leave the rest uncompiled.
> 
> I was thinking about letting TypeScript compile the whole program and then piece after piece migrate the JavaScript files to TypeScript.
> 
> What exactly did you have in mind?

I was thinking to migrate jdl first to reduce the complexity of the migration.

> > This is related to the generators lookup process. We need to customize the lookup process when moving generators to dist.
> 
> I am still not sure what the code does but I just naively replace `..` with `../..` in
> 
> https://github.com/jhipster/generator-jhipster/blob/b741f654858098e52307bdf0471e12f3fe93a30a/cli/environment-builder.js#L119

The lookups should be changed to `dist/generators` too.

> and so far it seems to work. At least generating a project works fine.
> > IMO we should do this when they are migrated to typescript.
> 
> If there is a way to only compile parts of the programm and we decide we prefer that over letting TypeScript compile all source files we could do this. Otherwise this is needed to be done within the scope of this PR.

IFAIK jdl doesn't uses other folder, so compiling the jdl folder and moving it to src/jdl should be enough.
And change jdl imports to use `dist/jdl`.
https://github.com/jhipster/generator-jhipster/blob/b741f654858098e52307bdf0471e12f3fe93a30a/cli/import-jdl.js#L28

Anyway, feel free to compile everything if it works.

mshima (commented):
We will support node >=16, target and module can be bumped.

package.json
```diff
@@ -51,22 +51,21 @@
...
+    "dist",
     "!**/__*",
     "!**/*.snap",
     "!**/*.spec.?(c|m)js"
   ],
   "scripts": {
+    "build": "npm run compile && npm run copy-files",
+    "compile": "tsc",
     "completion": "tabtab install --name jhipster --auto",
+    "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|mjs)' dist/generators && cpy 'package.json' dist",
```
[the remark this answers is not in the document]
emilpaw: Some modules include the package.json and I think it's easier/cleaner to also copy it to `dist` so that every file needed to run the program is in `dist` without making an exception for the package.json.

tsconfig.json
```diff
@@ -0,0 +1,104 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects. */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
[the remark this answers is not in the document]
emilpaw: I agree that we can bump the target version. But according to the [recommended base config](https://github.com/tsconfig/bases/blob/main/bases/node16.json) for Node 16 and this [StackOverflow post](https://stackoverflow.com/a/67371788) we should rather set `target` to ES2021 and add `ES2021` to lib. Sounds good?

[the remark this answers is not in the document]
mshima: According to https://kangax.github.io/compat-table/es2016plus/#node16_11, node >= 16.11.0 has full es2022 support except a v8 bug at RegExp.
IMO, we can use es2022. [§](https://kangax.github.io/compat-table/es2016plus/#test-Error.cause_property)[Error.cause property](https://github.com/tc39/proposal-error-cause) can be useful.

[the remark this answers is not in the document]
emilpaw: Done

.gitignore
emilpaw: Some arguments I can think of against committing the build artifacts:
- there must be a system in place which makes sure the source files stay in sync with the build output
- the files will show in PR view and it will be harder to navigate
- VSCode will not ignore the build output in file searches by default

In my opinion, the downsides are pretty big so it'll like to explore some different solutions.

By installing from git you mean this, right? `npm install -g git://github.com:jhipster/generator-jhipster.git`

The blog post @mraible posted looks quite promising. Just running our `npm run build` script shouldn't fail like running `husky install` might. 

For optimizing the pipeline we could run the build and save the result (e.g. as a GitHub artifact) or publish the package (e.g. to GitHub packages).

mshima: > By installing from git you mean this, right? `npm install -g git://github.com:jhipster/generator-jhipster.git`

Install locally like this:
https://github.com/hipster-labs/generator-jhipster-entity-audit/blob/4779ae2fb859d1c2ddc40ab2cd974dc65c06c6b2/package.json#L57

The blog post approach won't work because:
- sources are not included in package.json files.
- blueprint can have different typescript versions, they can be source incompatible.
- `generator-jhipster` project is quite big.

Since the `dist` directory will be too big with every templates (duplicated), we should not commit it to the repository.
Our release release process is quite complicated, using a snapshot can be help with development process.
We should think in something else later like publishing a build branch based on main.

tsconfig.json
```diff
@@ -0,0 +1,104 @@
+{
+  "include": ["cli/**/*", "generators/**/*", "jdl/**/*", "lib/**/*", "test/**/*"],
```
mshima, line 2:
```suggestion
  "include": ["cli/**/*", "generators/**/*", "jdl/**/*", "lib/**/*", "test/**/*"],
  "exclude": ["**/*.spec.*"],
```

package.json
mshima, line 68:
```suggestion
    "copy-files": "cpy 'generators/**' '!**/*.(js|ts|cjs|mjs|snap)' dist/generators && cpy 'package.json' dist",
```

tsconfig.json
emilpaw: What do you want to do? Do you want the tests to stay written in JavaScript and not migrated to TypeScript?

package.json
emilpaw: Done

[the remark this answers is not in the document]
emilpaw: I am not sure I understand. These files include the package.json: 
image
(I blacked out files that include another package.json, not the one we are talking about)

Do you propose to not copy the `package.json`?

mshima:
> What do you want to do? Do you want the tests to stay written in JavaScript and not migrated to TypeScript?

If we copy tests to dist, how do we update snapshots?
We should move every directory to `src` before building with typescript. This will make import package.json work in both `dist` and `src` directories.
We could use mocha + ts-node to test ts files or other alternative for in place ts tests.

emilpaw:
> If we copy tests to dist, how do we update snapshots?

I haven't yet looked into how snapshots work, so I don't know.




> We should move every directory to `src` before building with typescript. This will make import package.json work in both `dist` and `src` directories.

You mean to move `cli`, `generators`, `jdl` and `lib` into a newly created directory `src` and then use the path `../../package.json` instead of `../package.json`? If you think that's better than copying the package.json we can do that. 


> We could use mocha + ts-node to test ts files or other alternative for in place ts tests.

Yes, I also looked into that but wasn't sure if that's the right approach.

The generator lookup seems to not work correctly yet. At least that's my guess based on the error message in the failed tests. So just changing the paths did not work. I am still looking into that.

mshima:
> > If we copy tests to dist, how do we update snapshots?
> 
> I haven't yet looked into how snapshots work, so I don't know.

`npm run update-snapshots` runs tests updating snapshots located near tests.

> > We should move every directory to `src` before building with typescript. This will make import package.json work in both `dist` and `src` directories.
> 
> You mean to move `cli`, `generators`, `jdl` and `lib` into a newly created directory `src` and then use the path `../../package.json` instead of `../package.json`? If you think that's better than copying the package.json we can do that.

That's not the only reason, having more than one root source folder can be confusing.
Improving the source tree is planned for v8.
My idea was to have `cli`, `lib` and `generators` folder for v8. Moving `utils` and `jdl` inside `lib`.
Migrating everything to typescript, we can have something like:
 - src
   - cli
   - jdl
   - generators
   - lib
     - utils

> > We could use mocha + ts-node to test ts files or other alternative for in place ts tests.
> 
> Yes, I also looked into that but wasn't sure if that's the right approach.
> 
> The generator lookup seems to not work correctly yet. At least that's my guess based on the error message in the failed tests. So just changing the paths did not work. I am still looking into that.

The lookup is working fine, all integration tests are working.
But tests are been executed in place, so the lookup is looking in the wrong place when executing tests due to the change.

cli/environment-builder.js
```diff
@@ -116,7 +116,7 @@ module.exports = class EnvironmentBuilder {
    */
   _lookupJHipster() {
     // Register jhipster generators.
-    this.env.lookup({ packagePaths: [path.join(__dirname, '..')], lookups: ['generators'] }).forEach(generator => {
+    this.env.lookup({ packagePaths: [path.join(__dirname, '../..')], lookups: ['dist/generators'] }).forEach(generator => {
```
mshima, line 119:
```suggestion
    const generatorsRoot = path.basename(path.join(__dirname, '..'));
    this.env.lookup({ packagePaths: [path.join(__dirname, '../..')], lookups: [`${generatorsRoot}/generators`] }).forEach(generator => {
```

mshima: commiting to check the build result.

emilpaw: Looks good

mshima:
> In both PRs the pipelines don't seem to start anymore though. Any idea why?

Due to conflicts, since this tests of this PR is succeeding, we can finish it and merge.
The directory rearrangement can be done after.

test-integration/scripts/10-install-jhipster.sh
```diff
@@ -54,10 +54,12 @@ if [[ "$JHI_REPO" == *"/generator-jhipster" || "$JHI_GEN_BRANCH" == "local" ]];
     cd "$JHI_CLI_PACKAGE_PATH"
     git --no-pager log -n 10 --graph --pretty='%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
     npm ci --prod
+    npm run build
     npm install -g "$JHI_CLI_PACKAGE_PATH"
 
 elif [[ "$JHI_GEN_BRANCH" == "release" ]]; then
     echo "*** generator-jhipster: use release version"
+    npm run build
```
mshima, line 62:
```suggestion
```

It’s installing from package, is necessary?

emilpaw: Good call. No, I don't think it needs to. I removed it.

emilpaw:
> Due to conflicts

I thought that also the pipeline in this PR was not starting but it just took a very long time.
Adapting the package.json exports to the new build output was not yet done. I changed the paths and then some tests broke. Currently, I am working on that. Other than that I think this PR is done.

package.json
```diff
@@ -36,37 +36,36 @@
...
+    "./generators/": "./dist/generators/",
+    "./jdl/": "./dist/jdl/",
+    "./lib/": "./dist/lib/",
+    "./utils/": "./dist/utils/",
     "./package.json": "./package.json",
     "./esm/cli": "./cli/index.mjs",
     "./esm/generators/*": "./generators/*/esm.mjs",
     "./esm/generators": "./lib/constants/generators.mjs",
     "./esm/priorities": "./lib/constants/priorities.mjs",
     "./esm/support": "./lib/support/index.cjs"
```
mshima, line 52: Is there a reason for not update?

```suggestion
    "./esm/cli": "./dist/cli/index.mjs",
    "./esm/generators/*": "./dist/generators/*/esm.mjs",
    "./esm/generators": "./dist/lib/constants/generators.mjs",
    "./esm/priorities": "./dist/lib/constants/priorities.mjs",
    "./esm/support": "./dist/lib/support/index.cjs"
```

package.json
```diff
@@ -79,6 +78,7 @@
     "prettier:format": "prettier --write \"{,**/}*.{js,ts,cjs,mjs,json,md,yml,java}\"",
     "pretest": "npm run lint",
     "test": "mocha test generators --no-insight --forbid-only --parallel --max-old-space-size=4096",
+    "test-single": "mocha test test/cli/cli.spec.js",
```
mshima, line 81: Is it important?
```suggestion
```

to test a single file just run
```
npx mocha  test/...
```

test-integration/scripts/10-install-jhipster.sh
```diff
@@ -54,6 +54,7 @@ if [[ "$JHI_REPO" == *"/generator-jhipster" || "$JHI_GEN_BRANCH" == "local" ]];
     cd "$JHI_CLI_PACKAGE_PATH"
     git --no-pager log -n 10 --graph --pretty='%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
     npm ci --prod
+    npm run build
```
mshima, line 57:
```suggestion
    npm install
```

test-integration/scripts/10-install-jhipster.sh
```diff
@@ -70,5 +71,6 @@ else
     git --no-pager log -n 10 --graph --pretty='%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
 
     npm ci --prod
+    npm run build
```
mshima, line 74:
```suggestion
    npm install
```

mshima (commented):
I took a look at `prepare` script and it seems to work for this case:
Adding `"prepare": "npm run build",` script to package.json and build should be executed every time npm install is executed.
Can you check please?

tsconfig.json
```diff
@@ -0,0 +1,105 @@
...
+  "exclude": ["**/*.spec.*"],
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
mshima, line 5: I think we can remove commented configs.

package.json
emilpaw: Yes, I explained it in the commit: https://github.com/jhipster/generator-jhipster/pull/19559/commits/5263183b9cb32368b6e306aaa44a72af6785919c

package.json
emilpaw: No, I added it by mistake. It's removed.

tsconfig.json
emilpaw: Done

mshima (approved):

PR #19632, opened 2022-09-04
increase sleep time for monolith oauth2

mshima:
Related to https://github.com/jhipster/generator-jhipster-ionic/pull/712#issuecomment-1236377397

---

### DIFF
--- a/generators/server/templates/src/main/docker/app.yml.ejs
+++ b/generators/server/templates/src/main/docker/app.yml.ejs
@@ -294,7 +294,11 @@ _%>
     # When run with the "prod" Spring profile, it will read the configuration from a Git repository
     # See https://www.jhipster.tech/jhipster-registry/#spring-cloud-config
     environment:
+<%_ if (authenticationTypeOauth2) { _%>
+      - JHIPSTER_SLEEP=40
+<%_ } else { _%>
       - JHIPSTER_SLEEP=20
+<%_ } _%>
       - _JAVA_OPTIONS=-Xmx512m -Xms256m
       - SPRING_PROFILES_ACTIVE=dev,api-docs<% if (authenticationTypeOauth2) { %>,oauth2<% } %>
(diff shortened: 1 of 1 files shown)


generators/server/templates/src/main/docker/app.yml.ejs
```diff
@@ -294,7 +302,11 @@ _%>
     # When run with the "prod" Spring profile, it will read the configuration from a Git repository
     # See https://www.jhipster.tech/jhipster-registry/#spring-cloud-config
     environment:
+<%_ if (authenticationTypeOauth2) { _%>
+      - JHIPSTER_SLEEP=40
```
Tcharl, line 306: Should really start to play with health probe and dependency...

mshima: Can be trick, do you have an idea on how to implement?
It depends on property files, environment variables, profile.

Tcharl: For keycloak I had something like this: 
```
healthcheck:
      test: ['CMD-SHELL', 'curl -f http://localhost:9080 || exit 1']
      interval: 1m
      timeout: 10s
      retries: 3
```
But you're right, it's case by case approach, the command is different by service and the dependencies as well

And to bootstrap the final class (in CI):
```
startClass="$1"
shift

for var in "$@"
do
    while [ true ]; do
        RUNNING=$(docker inspect --format="{{ .State.Health.Status }}" $var)
        if ["$RUNNING" != "healthy"]; then
            sleep 1
        else
            break
        fi
    done
done

>&2 echo "Container is going to start"
exec java ${JAVA_OPTS} -Djava.security.egd=file:/dev/./urandom -cp /app/resources/:/app/classes/:/app/libs/* "$startClass"
```

mshima: Would be nice.

DanielFran (approved):

PR #19242, opened 2022-07-26
Update liquibase version to 4.15.0

DanielFran:
Fix https://github.com/jhipster/generator-jhipster/issues/17529
---

Please make sure the below checklist is followed for Pull Requests.

- [x] [All continuous integration tests](https://github.com/jhipster/generator-jhipster/actions) are green
- [x] Tests are added where necessary
- [x] The JDL part is updated if necessary
- [x] [jhipster-online](https://github.com/jhipster/jhipster-online) is updated if necessary
- [x] Documentation is added/updated where necessary
- [x] Coding Rules & Commit Guidelines as per our [CONTRIBUTING.md document](https://github.com/jhipster/generator-jhipster/blob/main/CONTRIBUTING.md) are followed

When you are still working on the PR, consider converting it to Draft (below reviewers) and adding `skip-ci` label, you can still see CI build result at your branch.

Issue #17529, opened 2022-01-10
Upgrade H2 database to 2.0.206+


sbouchex:
##### **Overview of the issue**

A RCE has been reported on H2 database (see https://github.com/h2database/h2database/security/advisories/GHSA-h376-j262-vhq6), so an upgrade is needed

##### **Motivation for or Use Case**

RCE

##### **Reproduce the error**

see https://github.com/h2database/h2database/security/advisories/GHSA-h376-j262-vhq6

##### **Suggest a Fix**

Upgrade the H2 database

##### **JHipster Version(s)**

7.5.0

- [X] Checking this box is mandatory (this is just to show you read everything)

DanielFran:
@sbouchex Not sure we can do the upgrade since it includes some breaking changes. See spring-projects/spring-boot#29200

mshima:
This is partially done. Rever to https://github.com/jhipster/generator-jhipster/pull/18939#issuecomment-1167262967.

Blocked on https://github.com/liquibase/liquibase/issues/2306.

PR #19657, opened 2022-09-06
Fix JdbcSQLIntegrityConstraintViolationException  when running Gradle integration tests

jprinet:
Fix `JdbcSQLIntegrityConstraintViolationException` when running Gradle integration tests by forcing h2database dependency version

See the build scan for details:
https://scans.gradle.com/s/7s6fydvpkazyo/tests/overview?outcome=failed

### DIFF
--- a/generators/generator-constants.js
+++ b/generators/generator-constants.js
@@ -39,6 +39,7 @@ const SPRING_BOOT_VERSION = '2.7.3';
 const LIQUIBASE_VERSION = '4.15.0';
 // TODO v8: Remove this constant
 const LIQUIBASE_DTD_VERSION = 'latest';
+const H2_VERSION = '1.4.200';
 const HIBERNATE_VERSION = '5.6.10.Final';
 const JACOCO_VERSION = '0.8.8';
 const JACKSON_DATABIND_NULLABLE_VERSION = '0.2.3';
@@ -408,6 +409,7 @@ const constants = {
   LIQUIBASE_VERSION,
   // TODO v8: Remove this constant
   LIQUIBASE_DTD_VERSION,
+  H2_VERSION,
   HIBERNATE_VERSION,
   JIB_VERSION,
   JACOCO_VERSION,
--- a/generators/server/templates/build.gradle.ejs
+++ b/generators/server/templates/build.gradle.ejs
@@ -250,6 +250,7 @@ configurations {
             // Inherited version from Spring Boot can't be used because of regressions:
             // To be removed as soon as spring-boot use the same version
             force 'org.liquibase:liquibase-core:<%= LIQUIBASE_VERSION %>'
+            force 'com.h2database:h2:<%= H2_VERSION %>'
         }
     }
 }


atomfrede:
Are you sure this is needed? It would be a downgrade from the official boot managed version (see https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html), which we usually try to avoid.

jprinet:
> Are you sure this is needed? It would be a downgrade from the official boot managed version (see https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html), which we usually try to avoid.

I discovered some integration test failures on the latest commit of the `jhipster-sample-app-gradle`

You can reproduce with
`./gradlew clean test integrationTest -x webapp -x webapp_test`

[Here](https://scans.gradle.com/s/7s6fydvpkazyo/tests/overview?outcome=failed) is a build scan demonstrating the issue (JdbcSQLIntegrityConstraintViolationException)
[Here](https://scans.gradle.com/s/2uljix7bxccm6/dependencies?dependencies=com.h2&expandAll) is a build scan with H2 forced to 1.4.200 which fixes the issue.

I am not quite sure the downgrade is strictly required but this version makes the tests pass

jprinet:
The issue started with [latest commit](https://github.com/jhipster/jhipster-sample-app-gradle/commit/80e42418895e665cce2bbd7ffe7ccb3e6451d871), it was fine prior to that

atomfrede:
We have removed the force of h2 here for both maven and gradle https://github.com/jhipster/generator-jhipster/commit/4ffb31f64e89bb740170cd76029d9ae1353e686d We have h2 for both build systems in our ci

mshima:
We were holding on h2 at 1.x until recently. See https://github.com/jhipster/generator-jhipster/pull/19242/files  and https://github.com/jhipster/generator-jhipster/issues/17529.

But that spring-boot already uses v2 and liquibase was blocked too.
We have a few h2 integration tests implemented like:
https://github.com/jhipster/generator-jhipster/blob/1ac085bd577e9ae6ba049dfbae2057912c3f2609/test-integration/workflow-samples/angular.json#L53-L60

But looks like we don't have imperative + h2 + mysql test.

This particular error is mysql + h2 problem.
Unfortunately this is a hit-and-miss thing.

mshima:
The commit https://github.com/jhipster/generator-jhipster/pull/19242/commits/d46d4704132aa17e8aba66da9380826b11e1b4f9 fixed the same problem for reactive tests. You may try `MODE=MySQL` instead see http://www.h2database.com/html/features.html#compatibility

atomfrede:
Ah, didn't check for database type, good point @mshima

jprinet:
Is that something you will fix then @atomfrede ?

atomfrede:
@jprinet I will have a look. It should be quite easy to update the h2 configuration depending on the database type. So if you like to create a PR go ahead. I will have not time before the weekend. Should we close this PR then?

mshima:
@jprinet please open a new issue so we can track the bug.
Setting the hibernate dialect to h2 at test/testdev may fix the problem, I didn’t test.

jprinet:
> @jprinet please open a new issue so we can track the bug.

There you go https://github.com/jhipster/generator-jhipster/issues/19687

jprinet:
closing this PR, let's use the issue for further discussion

PR #19763, opened 2022-09-16
Fix docker image build

emilpaw:
We can't use the production install since we need typescript and typescript is a devDependency.

Related to #19559 

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ RUN \
 
 RUN \
   # install jhipster
-  cd /home/jhipster/generator-jhipster && npm ci --production && \
+  cd /home/jhipster/generator-jhipster && \
   npm install -g /home/jhipster/generator-jhipster && \
   # fix jhipster user permissions
   chown -R jhipster:jhipster \


emilpaw:
After fixing this I wondered why the `10-install-jhipster.sh` script can successfully install the package although it's using the production build. 

The difference is that the script uses `--prod` which I don't think work and is just ignored. So the production build never actually work. That means we can remove that there too, right?

mshima:
> The difference is that the script uses `--prod` instead of `--production` which I don't think works and is just ignored. So the production build never actually worked. That means we can remove that there too, right?

Indeed, I was confused it worked.

Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN \
 
 RUN \
   # install jhipster
-  cd /home/jhipster/generator-jhipster && npm ci --production && \
+  cd /home/jhipster/generator-jhipster && \
```
mshima, line 58: I think npm ci is required.
```suggestion
  cd /home/jhipster/generator-jhipster && npm ci &&\
```

emilpaw: As far as I understood calling `npm ci` beforehand doesn't really do anything. `npm install` installs the package anyways, regardless of what `npm ci` did.

Running it on my machine works.

mshima: lets test.

emilpaw: It did not work. I didn't properly clean my local folder so docker build copied some of the build artifacts over into the image and because of that it worked locally.

My guess is that `npm ci` is needed after all because altough `npm install` installs `tsc` it installs it in the global node_modules from where it might not be accessible by the prepare script. Not sure though.

I added `npm ci` again and it works locally, even after properly cleaning the folder before building :) #19766

PR #19798, opened 2022-09-21
Update jhipster-registry version to 7.4.0

juliensadaoui:
Update jhipster-registry version to 7.4.0


DanielFran:
@juliensadaoui Can you update the snapshot tests?

Issue #19113, opened 2022-07-08
Switch to ESM.


mshima:
##### **Overview of the feature request**

- [ ] app
- [ ] base
- [ ] bootstrap
- [ ] bootstrap-application
- [ ] ci-cd
- [ ] client
- [ ] common
- [ ] cypress
- [ ] database-changelog
- [ ] database-changelog-liquibase
- [ ] docker-compose
- [ ] generate-blueprint
- [ ] gradle
- [ ] heroku
- [ ] kubernetes
- [ ] java
- [ ] languages
- [ ] maven
- [ ] project-name
- [ ] server
- [ ] spring-boot
- [ ] upgrade
- [ ] workspaces

##### **Motivation for or Use Case**

Follow javascript standards

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

Tcharl:
Reopened: the jhipster cli is not working anymore due to a dependecy to constants.cjs, which itself needs jdl/validation.mjs...

mshima:
The generator has already switched to esm and every bit is esm except jdl and cli.

The cli problem is related to mocking, which I am working on.
The jdl can be switched except some definitions that the cli uses.

Please open a new issue to track cli and jdl switch to esm.

Tcharl:
The issue was labelled 'switch to ESM', thus wasn't mentioning that it was only targeting generator, still, the content did.

Why not reopening to get the complete picture? What's the added value of two tickets ?

I saw your current repo: great ambition which would really help, keep on rockin'! I'm trying through a simpler way, but not sure to succeed.

Best regards,

mshima:
@DanielFran bounty claimed https://opencollective.com/generator-jhipster/expenses/123269

DanielFran:
@mshima approved

PR #19856, opened 2022-09-27
Migrate JDL to TypeScript

### LINKED ISSUE #19182: Migrate JDL to typescript

JDL is quite complicated to understand and missing a few features.
We should migrate it to typescript to make it simpler and may receive more contributions.

Improve developer experience.

emilpaw:
Resolve #19182 

---


.mocharc.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   timeout: 30000,
   ui: 'bdd',
   extension: ['js', 'spec.cjs', 'spec.mjs', 'spec.ts', 'spec.mts'],
-  require: ['mocha-expect-snapshot/old-format'],
+  require: ['mocha-expect-snapshot/old-format', 'ts-node/register'],
```
mshima, line 8: Since we are using esbuild-kit/esm-loader for esm, probably we should use esbuild-kit/cjs-loader for commonjs.
https://github.com/esbuild-kit/cjs-loader

.mocharc.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
...
   // Resolve absolute path for test with fork and different cwd.
   loader: require.resolve('@esbuild-kit/esm-loader'),
```
mshima, line 10:
```suggestion
  loader: [require.resolve('@esbuild-kit/esm-loader'), require.resolve('@esbuild-kit/cjs-loader')],
```

.mocharc.js
emilpaw: I added the loader but with that mocha doesn't work anymore. The error message is 

> ✖ ERROR: /home/runner/work/generator-jhipster/generator-jhipster/node_modules/@cspotcode/source-map-support/source-map-support.js:49
>   var sharedData = this[sharedDataKey];
>                    ^
> 
> TypeError: Cannot read properties of undefined (reading 'Symbol(source-map-support/sharedData)')

See here: https://github.com/jhipster/generator-jhipster/actions/runs/3169199634/jobs/5160937434

Since mocha works without this package I removed it for now.

.mocharc.js
emilpaw: See here: https://github.com/jhipster/generator-jhipster/pull/19856#discussion_r985261292

emilpaw:
@mshima Some snapshot tests in `option-converter.spec.ts` and `jdl-importer.spec.ts` still fail. I have a hard time finding the bug. For example for the test case:
```
  1) OptionConverter
       convertOptions
         when passing options
           such as skipClient
             should convert it
```

it says the snapshot name is
```
Snapshot name: `OptionConverter convertOptions when passing options such as skipClient should convert it 1`
```
and it received
```
Received:
Array [
  JDLUnaryOption {
    "entityNames": Set {
      "A",
    },
    "excludedNames": Set {
      "B",
    },
    "name": "skipClient",
  },
]
```
In `option-converter.spec.js.snap` under the name `OptionConverter convertOptions when passing options such as skipClient should convert it 1` there is: 
```
Array [
  JDLUnaryOption {
    "entityNames": Set {
      "A",
    },
    "excludedNames": Set {
      "B",
    },
    "name": "skipClient",
  },
]
```
which looks identical to what it said it received. What's the issue here? Am I reading the test output wrong?

mshima:
@emilpaw you need to rename the snap file from `js.snap` to `ts.snap`.

emilpaw:
@mshima Thanks, that was it.

There are still things to properly type and some open TODOs, but I think it's best to merge this as is and then improve the code later. This PR is already too big.

mshima (commented):
Would be nice to have a basic type for `applicationConfig`, `entity`, `field`, `relationship`.

```
// config is generic because jdl should not be technology bound
type ApplicationConfig = Record<string, boolean | string | number | string[]>;

// Field and relationship needs to be defined and options are generic but fallback to generic
type Field = {
  fieldName: string;
  fieldType: string;
  options: Record<string, boolean | string | number>;
} & Record<string, boolean | string | number | string[]>

type Relationship = {
  relationshipName: string;
  relationshipType: string;
  options: Record<string, boolean | string | number>
} & Record<string, boolean | string | number | string[]>

type Entity = {
  javadoc?: string;
  fields?: Fields[];
  relationships?: Relationships[];
}
```

Don't need to be complete, but we need to reduce any usage and make easier to follow the types.

package.json
```diff
@@ -119,8 +119,10 @@
     "yeoman-generator": "5.7.0"
   },
   "devDependencies": {
-    "@esbuild-kit/esm-loader": "2.4.2",
+    "@chevrotain/types": "^10.3.0",
+    "@types/chai": "4.3.3",
     "@types/mocha": "10.0.0",
+    "@types/sinon-chai": "^3.2.8",
```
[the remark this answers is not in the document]
emilpaw: Done.

package.json
```diff
@@ -119,8 +119,10 @@
     "yeoman-generator": "5.7.0"
   },
   "devDependencies": {
-    "@esbuild-kit/esm-loader": "2.4.2",
+    "@chevrotain/types": "^10.3.0",
```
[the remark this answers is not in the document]
emilpaw: Done.

jdl/validators/entity-validator.ts
```diff
@@ -17,22 +17,21 @@
...
+import { isReservedClassName } from '../jhipster/reserved-keywords';
 
-class EntityValidator extends Validator {
+export default class EntityValidator extends Validator {
   constructor() {
     super('entity', ['name', 'tableName']);
   }
 
   validate(jdlEntity) {
+    console.log('in validate');
```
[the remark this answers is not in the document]
emilpaw: Yes, good catch. Removed.

.mocharc.cjs
```diff
@@ -7,6 +7,6 @@ module.exports = {
   extension: ['js', 'spec.cjs', 'spec.mjs', 'spec.ts', 'spec.mts'],
   require: ['mocha-expect-snapshot/old-format', 'test/mocha.config.cjs'],
   // Resolve absolute path for test with fork and different cwd.
-  loader: require.resolve('@esbuild-kit/esm-loader'),
+  loader: require.resolve('tsx'),
```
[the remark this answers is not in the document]
emilpaw: I don't see using tsx as a loader as an issue but I understand your reasoning and removed it. Using the loaders by itself looks quite convoluted.

What command should be run in the cli tests now? This obviously doesn't work without tsx: https://github.com/jhipster/generator-jhipster/blob/4ce31cec6fb5fc010f69206b549fffa25d813dea/test/utils/utils.cjs#L71
I already tried ``` let cmd = `node --require @esbuild/cjs-loader --loader @esbuild-kit/esm-loader ${cmdPath} `;``` but it does not work.

I think using tsx as a replacement for node in the tests similar to what you would do with ts-node (and like it was done in the cli test) could also be a good solution. With that we could also get rid of the warning: `(node:97101) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time` which is quite annoying. I could not get that to work though. And this would mean we need to keep tsx.

emilpaw:
I included the Field, Entity & Relationship types and improved the typing overall in https://github.com/jhipster/generator-jhipster/pull/19856/commits/164c070ecc7670c19bb3c3aadfe111e26b0bef84 and https://github.com/jhipster/generator-jhipster/pull/19856/commits/71e69ec79a5ba8b13f97f79d281a3f03356c7313 .

With the type `ApplicationConfig` do you mean the object which is passed here? https://github.com/jhipster/generator-jhipster/blob/71e69ec79a5ba8b13f97f79d281a3f03356c7313/jdl/models/jdl-application.ts#L29

.mocharc.cjs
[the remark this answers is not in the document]
mshima: I think we can use the cli from `dist` folder instead of using tsx.
It's just basic tests like `--version`, `unknown command` and a few more.
What do you think?

test/utils/utils.cjs
```diff
@@ -68,7 +68,7 @@ function shouldBeV3DockerfileCompatible(databaseType) {
 
 function getJHipsterCli() {
   const cmdPath = path.join(__dirname, '../../cli/jhipster.mjs');
```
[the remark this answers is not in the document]
emilpaw: Done

.eslintrc.json
```diff
@@ -62,12 +66,15 @@
...
-    "no-unused-vars": [2, { "vars": "local", "args": "none" }],
     "no-multi-str": 2,
     "no-promise-executor-return": "off",
     "no-irregular-whitespace": 2,
+    "@typescript-eslint/no-empty-function": "off",
     "comma-dangle": "off",
     "max-len": "off",
+    // TODO: Investigate if we can avoid the issue instead of
+    // deactivating the lint rule.
+    "import/no-duplicates": "off",
```
MathieuAA, line 77: Can you create an issue for this please?

AFAIK, this shouldn't be too much of a problem but to make the review and the processing of this PR easier, I'd rather have this in another task.

Link to the doc: https://eslint.org/docs/latest/rules/no-duplicate-imports

MathieuAA: Once the issue is created, you can replace this comment's content by the issue's link

.eslintrc.json
```diff
@@ -77,6 +84,7 @@
     "prefer-destructuring": "off",
     "no-multi-assign": "off",
     "no-param-reassign": "off",
+    "lines-between-class-members": [2, "always", { "exceptAfterSingleLine": true }],
```
MathieuAA, line 87: We can skip this rule, no?

generators/bootstrap-application-server/generator.mts
```diff
@@ -22,14 +22,15 @@ import { GENERATOR_BOOTSTRAP_APPLICATION_BASE } from '../generator-list.mjs';
...
-import FieldTypes from '../../jdl/jhipster/field-types.js';
 
-const {
-  CommonDBTypes: { LONG: TYPE_LONG },
-} = FieldTypes;
-const { OAUTH2 } = AuthentitcationTypes;
+const { CommonDBTypes } = fieldTypes;
+const { OAUTH2 } = authenticationTypes;
+
+const { LONG: TYPE_LONG } = CommonDBTypes;
```
MathieuAA, line 32: There should be an issue to improve the imports.

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
```diff
@@ -17,42 +17,45 @@
...
 
-module.exports = {
-  convert,
-};
+const { isBlobType } = fieldTypes;
+const { ANY_BLOB, BLOB, IMAGE_BLOB, TEXT_BLOB } = fieldTypes.CommonDBTypes;
+const { ANY, IMAGE, TEXT } = fieldTypes.BlobTypes;
+const { BYTES } = fieldTypes.RelationalOnlyDBTypes;
+
+export default { convert };
```
MathieuAA, line 36: Same suggestion as before, if we have an issue for the imports, there should be one for exports (inlining default exports and regular exports).

jdl/converters/jdl-to-json/jdl-to-json-option-converter.ts
```diff
@@ -30,22 +37,22 @@ const serviceClassOptionValue = BinaryOptions.Values.service.SERVICE_CLASS;
...
+type JDLOptionHolder = JDLObject | JDLApplication;
 
 /**
  * Converts JDL options to JSON content to be set in JSON entities.
  * This function takes a single parameter: a JDL option holder. This holder is a class or an object responding to these
  * calls:
  *   - forEachOption, to loop over JDL options
  *   - getEntityNames, to get the declared entity names
- * @param {JDLObject|JDLApplication} jdlOptionHolder - a JDL object (a JDLObject or a JDLApplication) containing the options.
+ * @param jdlOptionHolder - a JDL object (a JDLObject or a JDLApplication) containing the options.
```
MathieuAA, line 52: I didn't notice before but this change makes me realize that we can remove the comments altogether, as the IDEs can make use of TS typing system to assist the developers in knowing the signatures, etc.

We can have another issue to remove comments from this part of the code :)

jdl/converters/json-to-jdl-converter.ts
```diff
@@ -108,7 +115,7 @@ function cleanYoRcFileContent(yoRcFileContent) {
   return yoRcFileContent;
 }
 
-function getJSONEntityFiles(applicationDirectory) {
+function getJSONEntityFiles(applicationDirectory: string) {
   const entities = new Map();
```
MathieuAA, line 119: This would make the compiler cry. I think it tolerates not having parameterized types, but displays a warning about it, doesn't it?

We can create an issue to tackle it and use the right types

jdl/converters/jdl-to-json/jdl-without-application-to-json-converter.ts
```diff
@@ -71,21 +72,21 @@ function resetState() {
 }
 
 function setBasicEntityInformation() {
-  const convertedEntities = BasicEntityConverter.convert(jdlObject.getEntities());
+  const convertedEntities = BasicEntityConverter.convert(jdlObject!.getEntities());
```
MathieuAA, line 75: You're right to use the `!` and not fix it in this PR.
We have to create an issue to change how the design of this kind of files (like using a class, or another solution).

jdl/parsing/jdl-parser.ts
```diff
@@ -352,7 +351,7 @@ module.exports = class JDLParser extends CstParser {
     });
   }
 
-  enumProp() {
+  enumProp(): any {
```
MathieuAA, line 354: I'm putting this comment here, but we can replace the any here by void

MathieuAA: and in this file

jdl/parsing/jdl-ast-builder-visitor.ts
```diff
@@ -352,7 +357,7 @@ module.exports = class JDLAstBuilderVisitor extends BaseJDLCSTVisitor {
   }
 
   enumProp(context) {
-    const prop = {
+    const prop: any = {
```
MathieuAA, line 360: I think we need a new issue for handling the files related to chevrotain and the parsing system

MathieuAA: Because doing it in this PR would take a long time

jdl/chevrotain.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'chevrotain' {
```
MathieuAA, line 1: Why is it needed? Chevrotain is written in TS, there shouldn't be an issue here

MathieuAA (commented):
That's a lot of work, and I'd rather merge it if it all the tests pass and iterate later on

mshima (approved):
LGTM, it's typescript now and will make further development easier.

We need to define more types and improve others.
But IMO we should drop every validation and business logic out from jdl and make it only a parser for v8. It will allow blueprints to add custom properties.
Moving the validation and business logic to the generators will converge to the existing generators migration.

One direction to decide is wether we should implement jdl using an inheritable or configurable implementation.

But I will let @MathieuAA to decide when to merge.

MathieuAA:
@mshima with TS, there's one big action to do: redesign every file. Not just in the JDL module, but everywhere.
This is not gonna be easy as there are conventions to document and evolve.
About this PR, once the issues have been created, let's merge it.

mshima:
> @mshima with TS, there's one big action to do: redesign every file. Not just in the JDL module, but everywhere.
> This is not gonna be easy as there are conventions to document and evolve.
> About this PR, once the issues have been created, let's merge it.

@MathieuAA I know it won't be easy we have ~9 generators written in typescript right now. And most of the tests already are in typescript.
We are slowly: evolving the types, redesigning the module tree, cleaning up apis.

jdl/converters/json-to-jdl-converter.ts
emilpaw: What do you think is wrong with this? TypeScript does not complain about it and it's correct as far as I can tell.

.eslintrc.json
emilpaw: I am not activating the rule I am changing its configuration. The existing rule has "exceptAfterSingleLine" set to false, which means that eslint would require an empty line between class members like in this case: https://github.com/jhipster/generator-jhipster/blob/c81c32e79edf6ee1bcdace4fe632ad88c9ae3e22/jdl/models/jdl-relationship.ts#L27-L36

The rule didn't matter before because there were no single line class members but now there are some because TypeScript excepts that every property on a class is defined and so I had to add these.

IMO empty lines between single-line class members make the class needlessly large.

jdl/converters/jdl-to-json/jdl-without-application-to-json-converter.ts
emilpaw: Yes, I agree that this has to be solved. The need for this non-null assertion comes from the usage of global variables and how they are initialized. Some parts of the code are written in a way that makes it hard to rely on TypeScripts type safety.

.eslintrc.json
emilpaw: Done

generators/bootstrap-application-server/generator.mts
emilpaw: The imports are rather ugly in part because we use multiple module systems. This will improve when the project is completely migrated to ESM. See this issue: #19113

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
emilpaw: Same answer as before. :)

jdl/parsing/jdl-parser.ts
emilpaw: Unfortunately void does not work.

jdl/converters/jdl-to-json/jdl-to-json-option-converter.ts
emilpaw: What part of the comment do you mean exactly? 

We can remove the types in the JSDocs after the types are annotated using TypeScript. That is what I've done in this line and in some other places.

jdl/converters/json-to-jdl-converter.ts
MathieuAA: Oh? Wow, Last time I checked the compiler threw stones at me.
If there's no issue from the compiler, then let's just create an issue for this.
Having parameterized types, especially in types like the Map or the Set, is useful because TS won't infer the type. It will use "unknown" if memory serves. It's not optimal memory-wise. Plus, having types helps understanding the code

jdl/chevrotain.d.ts
emilpaw: There seems to be an issue with how chevrotain defines its types because without this TypeScript can't find the module declaration for `chevrotain`.

jdl/converters/json-to-jdl-converter.ts
emilpaw: I thought you are talking about the function argument while you were talking about the map a line below.

It works because TypeScript infered it as Map<any, any>. I pushed a commit improving the types.

emilpaw:
@mshima I just changed a comment and suddenly eslint fails because it apparently detects an unresolved import. Running the linter locally does not show this error and the tests also pass. Any idea? Does it work on your machine?

emilpaw:
@mshima Nevermind my last comment. Suddenly the issue is also shown on my local machine, not sure why it did not before. 

The issue is the newly added dynamic import of jdl-import. It does not work with the TypeScript setup. I am working on it.

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
```diff
@@ -98,7 +101,7 @@ function getConvertedFieldsForEntity(jdlEntity, jdlObject) {
 }
 
 function getBlobFieldData(fieldType) {
-  const blobFieldData = {
+  const blobFieldData: any = {
```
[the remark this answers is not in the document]
emilpaw: The `any` type is used in a lot of places in the codebase. Do we need an issue for every file containing `any`?

jdl/parsing/jdl-ast-builder-visitor.ts
emilpaw: I created an issue for that: https://github.com/jhipster/generator-jhipster/issues/20035

emilpaw:
The issues are created, comments are answered and the tests pass. Let's merge this. I'd rather work on improving the project than fixing merge conflicts.

jdl/converters/jdl-to-json/jdl-to-json-option-converter.ts
MathieuAA: All the comments :) they're not useful

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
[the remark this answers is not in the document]
MathieuAA: At first, having any types is useful (so that the project keeps working). Replacing the any types after that is mandatory. If we don't, what's the point of having TS? ;)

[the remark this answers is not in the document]
MathieuAA: Just one issue is enough

MathieuAA:
LGTM

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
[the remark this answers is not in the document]
emilpaw: Done: https://github.com/jhipster/generator-jhipster/issues/20037

Not sure if these kinds of vague and long-running issues are particularly useful though.

jdl/converters/jdl-to-json/jdl-to-json-option-converter.ts
emilpaw: The comments do provide important information in some cases and I think having good comments makes the code easier to work with. But obviously not every comment is useful.

jdl/converters/jdl-to-json/jdl-to-json-field-converter.ts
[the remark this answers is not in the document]
MathieuAA: They are, because they bring visibility to problems and the work being done to solve them :)
What you created is a good exemple of z good issue.

mshima:
Thanks @emilpaw

emilpaw:
Bug bounty claimed: https://opencollective.com/generator-jhipster/expenses/115035 Thanks :)

Issue #18778, opened 2022-05-31
Add frontend sort when no pagination


Tcharl:
##### **Overview of the feature request**

Jhipster lack of essential features to get really usable forms.
One of them is that there's no sort on fields when pagination is not selected.
It would be really easy to implement while being cool for end users.

##### **Motivation for or Use Case**

Improving the waow effect of a jhipster generation at low cost

- [x] Checking this box is mandatory (this is just to show you read everything)

Tcharl:
@deepu105 I get why the react-jhipster is complaining: some form React elements of the library are manipulating states and effects within the library itself, which is considered as a bad practice and unsupported in the last react versions.
It's very hard for me to adapt the lib as I didn't produce it by myself, so any help would be really appreciated.
Otherwise I'll continue to look at it, but it can take a bit of time...

deepu105:
I'll check when I find some time

Tcharl:
Created a branch with the current library achievement

https://github.com/Tcharl/react-jhipster/tree/sort_non_paginated
https://github.com/Tcharl/generator-jhipster/tree/sort_react

I made a bit of progress, but still some work to achieve

PR #19960, opened 2022-10-09
Add sort for react non paginated list

Tcharl:
I struggled a bit generating a sample project.
Also started to split small things that will prepare the sort for non paginated element in react

Contributes to https://github.com/jhipster/generator-jhipster/issues/18778
---


mshima:
I don’t know any other node package that have this behavior. IMO we should drop this.

`jhipster` for global and `npx jhipster` for local is the normal behavior.
And v8 global jhipster command will fail to invoke local v7. And global v7 will fail to invoke local v8.

Tcharl:
Still, the only way to make it work...
I also started another small thing on it...

Tcharl:
Can it be reviewed and then https://github.com/jhipster/react-jhipster/pull/89 merged so that I'll update that branch to be merged too?
I'll continue the sort stuff and try to export everything I can to the lib.
I would like to do it by baby step instead of big bang thing (I already failed twice implementing it)

Tcharl:
@qmonmert ?

Tcharl:
Hmmmm, I think I made a mistake, I merged react-jhipster after the review, but now I'm unable to build the branch.
It's unfortunately a breaking change, as the `getSortState`is replaced by `getPaginationState`. @all, really sorry for that: what is the procedure to fix it?

mshima:
@Tcharl no problem.
Just change the react-jhipster to the git main branch.

generators/client/templates/react/package.json.ejs
```diff
@@ -41,7 +41,7 @@
     "react": "<%= this.dependabotPackageJson.dependencies['react'] %>",
     "react-dom": "<%= this.dependabotPackageJson.dependencies['react-dom'] %>",
     "react-hook-form": "<%= this.dependabotPackageJson.dependencies['react-hook-form'] %>",
-    "react-jhipster": "<%= this.dependabotPackageJson.dependencies['react-jhipster'] %>",
+    "react-jhipster": "jhipster/react-jhipster",
```
mshima, line 44: change at package.json instead of the template file

Tcharl: Ahhhh, thanks!

Issue #17132, opened 2021-11-25
Unambiguous Priorities


mshima:
##### **Overview of the feature request**

RFC https://github.com/jhipster/generator-jhipster/pull/17131
Rendered https://github.com/jhipster/generator-jhipster/blob/main/rfcs/3-jhipster-rfc-unambiguous-priorities.md

Update:
Priorities name was moved as static class member.
Current import for `generator-jhipster/priorities` implies that we have same priorities for every generator, which is not true. So move priorities as a class member so a base generator will export the priorities it provides.

The syntax will be:
```
  get [BaseGenerator.INITIALIZING] () {
    return {};
  }
```

##### **Motivation for or Use Case**

Improve jhipster developer experience.

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

PR #19978, opened 2022-10-10
Finalize unambiguous priorities api changes.

mshima:
- change every generator to extend base/index.cjs instead of generator-base-blueprint or generator-base.
generator-base-blueprint and generator-base are in the inheritance tree, but should be dropped for v8.
- complete unambiguous api changes.
- drop deprecated preparingFields and preparingRelationships priorities
- cleanup a few unused generator-base methods.
- move lib/constants/priorities to near the generator that implements it.

Closes https://github.com/jhipster/generator-jhipster/issues/17132

---


generators/base-application/priorities.cjs
```diff
@@ -0,0 +1,138 @@
...
+const PRIORITY_NAMES_LIST = [
+  PRIORITY_NAMES.INITIALIZING,
+  PRIORITY_NAMES.PROMPTING,
+  PRIORITY_NAMES.CONFIGURING,
+  PRIORITY_NAMES.COMPOSING,
+  PRIORITY_NAMES.LOADING,
+  PRIORITY_NAMES.PREPARING,
+  CONFIGURING_EACH_ENTITY,
+  LOADING_ENTITIES,
+  PREPARING_EACH_ENTITY,
```
Tcharl, line 111: The EACH term is a bit confusing, it could mean that it will modify EACH Entities ^^.

Removing the EACH looks to be more meaningful (meaning that we are focusing on one specific entity).
Am I wrong with the meaning? If it really means EACH, I would go for PREPARING_ENTITIES.

generators/base-application/priorities.cjs
```diff
@@ -0,0 +1,138 @@
+const { PRIORITY_PREFIX, QUEUE_PREFIX, PRIORITY_NAMES, QUEUES } = require('../base/priorities.cjs');
```
Tcharl, line 1: Isn't it worth it to create a `base-platform/priorities.cjs` (a bunch of application), a `base-entity/priorities.cjs` and maybe a `base-field/priorities.cjs`?

generators/base-application/priorities.cjs
mshima: I think we can rename, but let's do it in another PR.

generators/base-application/priorities.cjs
mshima: > Isn't it worth it to create a `base-platform/priorities.cjs` (a bunch of application).

Yes, I think we should call it `base-workspaces`.
I will create a feature request.

> a `base-entity/priorities.cjs` and maybe a `base-field/priorities.cjs`?

I don't think so.
But I think we should implement a base-incremental-application for entity changes like liquibase and maybe entity removal.

generators/server/generator.cjs
```diff
@@ -508,7 +508,7 @@ module.exports = class JHipsterServerGenerator extends BaseApplicationGenerator
   }
 
   get [BaseApplicationGenerator.CONFIGURING_EACH_ENTITY]() {
-    return this.asPromptingTaskGroup(this.delegateToBlueprint ? {} : this.configuringEachEntity);
+    return this.asConfiguringEachEntityTaskGroup(this.delegateToBlueprint ? {} : this.configuringEachEntity);
```
Tcharl, line 511: Idem: asConfiguringEntityTaskGroup

mshima:
> Reviewed Please take the comments into consideration:
> 
> * Priority names should be crystal clear, the X_EACH_ENTITY is not that clear: we don't know if the processing should be on every entities (in this case X_ENTITIES would be more clear) or if the processing is applied on one specific entity (thus X_ENTITY would be more clear)

I would think that with `each` word the priority purpose is more clear and avoid confusion with the plural form `PREPARING_ENTITY` vs `PREPARING_ENTITIES`.
Now that we have types, it won't change much.
I am not a native english speaker, maybe @mraible can help us deciding.

Why I don't use the plural form for every priority?
```
get [BaseGenerator.PREPARING_EACH_ENTITY_FIELD]() {
  return {
    prepareField({ entity, field }) {
      ...
    }
  }
}
```

instead of 
```
get [BaseGenerator.PREPARING_ENTITIES_FIELDS]() {
  return {
    prepareFields({ entities }) {
      for (const entity of entities) {
        for (const field of fields) {
          ...
        }
      }
    }
  }
}
```

Why the plural form for others?
```
get [BaseGenerator.WRITING_ENTITIES]() {
  return {
    async writeEntities({ entities }) {
      await this.writeFiles({
        file: `menu.txt`,
        context: { entities },
      });
      await Promise.all(entities.map(entity => this.writeFiles({ file: 'entity-file.txt', context: { ...entity })));
    }
  }
}
```

Using a template like:
```
  <%_ for (const entity of entities) { %>
    <%= entity.name %>
  <%_ } %>
```

At this point this can be easily changed before final v8.

> * It may be cleaner to split the generator's lifecycle (and metamodel) of platform vs application vs entity vs fields. It would prepare future 'thiner' generators and improve the readability of the model and generator's code

I don't understand what do mean by this. What are you proposing?
Entities/Fields/Relationships depends on the application. For example databaseType defines the id field type, reserved words. The user entity depends on authenticationType/applicationType.
Entities/fields/relationships are related with each other, Entities depends on fields and relationships for primary key. Relationship depends on entities and fields for the type and label type.

Every lifecycle are optional. So IMO having `base-application, base-entity, base-field, base-relationship` will not improve the readability.
Platform lifecycle is non existent right now.

This PR is not changing priority names, it's moving more generic priorities to `generators/base` and application/entities priorities to `generators/base-application`.

@Tcharl this PR can be merged?

generators/base-application/priorities.cjs
Tcharl: Some users are complaining that generation is all or nothing. Splitting metamodels (and models) could ease to generate only entities and fields that have been updated

mshima: @Tcharl the application, entities, fields and relationships are dependent on each other, it's not possible to create a consistent application without preparing everything.
See https://github.com/jhipster/generator-jhipster/issues/19989.

DanielFran (approved):

Issue #19810, opened 2022-09-22
Modularize client and server generators


mshima:
##### **Overview of the feature request**

Split client generator into:
- client (for blueprints, maybe i18n)
- angular
- react
- vue

Split server generator into:
- server (for blueprints)
- spring-boot
- spring-boot-sql
- spring-boot-cassandra
- spring-boot-couchbase
- spring-boot-mongodb
- spring-boot-neo4j
- spring-boot-cache-provider
- spring-boot-gatling
- spring-boot-websocket
- maven
- gradle

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

PR #20079, opened 2022-10-22
Migrate remaining generators to esm

mshima:
Closes https://github.com/jhipster/generator-jhipster/issues/19113.

- migrate remaining generators to esm
- migrate base generator to esm
- reorganize some utils
- drop add/java/spring-boot generators. We can recreate later if needed.
- use maven/gradle generators at server workflow.
- drop mixing chain support.

We still have a few cjs files:
- cli (tests needs to be reimplemented)
- a few constants and legacy needle support that should be dismantled.
- aws lib. 

First commit just change imports from cjs to mjs.

Related to https://github.com/jhipster/generator-jhipster/issues/19810.

---


DanielFran:
@mshima We should move also maven wrapper to maven generator and use it in server?

generators/base/utils.mjs
```diff
@@ -0,0 +1,92 @@
...
+      callbacks = [ct => normalizeLineEndings(ct, '\n')].concat(callbacks).concat(ct => normalizeLineEndings(ct, '\r\n'));
+    }
+    for (const callback of callbacks) {
+      content = callback.call(this, content, filePath);
+    }
+    return content;
+  };
+};
+
+export function formatDateForChangelog(now) {
```
Tcharl, line 44: changelogutils?

generators/gradle/generator.mjs
```diff
@@ -0,0 +1,89 @@
...
+import { GENERATOR_GRADLE, GENERATOR_BOOTSTRAP_APPLICATION_SERVER } from '../generator-list.mjs';
+import files from './files.mjs';
+import { GRADLE } from './constants.mjs';
+import { GRADLE_VERSION } from './constants.mjs';
+
+/**
+ * @class
+ * @extends {BaseApplicationGenerator<import('../bootstrap-application-server/types.js').SpringBootApplication>}
+ */
+export default class MavenGenerator extends BaseApplicationGenerator {
```
Tcharl, line 33: Maven generator in gradle/generators?

generators/init/templates/package.json.ejs
```diff
@@ -32,7 +32,7 @@
     "prettier-plugin-packagejson": "<%= this.nodeDependencies['prettier-plugin-packagejson'] %>"
   },
   "engines": {
-    "node": ">=<%= NODE_VERSION %>"
+    "node": ">=<%= nodeVersion %>"
```
Tcharl, line 35: Why isn't it in constant format (for my own culture)?

generators/server/generator.mjs
```diff
@@ -222,10 +224,20 @@ export default class JHipsterServerGenerator extends BaseApplicationGenerator {
         await this.composeWithJHipster(GENERATOR_COMMON);
       },
 
-      async composeLanguages() {
+      async composing() {
+        const { buildTool, enableTranslation } = this.jhipsterConfigWithDefaults;
+        if (buildTool === GRADLE) {
```
Tcharl, line 229: Do you think that we can use derived constant right after the seed phase?

generators/server/templates/build.gradle.ejs
```diff
@@ -565,7 +565,7 @@ task cleanResources(type: Delete) {
 }
 
 wrapper {
-    gradleVersion = "<%= GRADLE_VERSION %>"
+    gradleVersion = "<%= gradleVersion %>"
```
Tcharl, line 568: Same question regarding constant format

generators/server/templates/package.json.ejs
```diff
@@ -32,9 +32,9 @@
     "<%= module.name %>": "<%= module.version %>",
   <%_ });
 } _%>
-    "prettier": "<%= this.dependabotPackageJson.devDependencies['prettier'] %>",
-    "prettier-plugin-java": "<%= this.dependabotPackageJson.devDependencies['prettier-plugin-java'] %>",
-    "prettier-plugin-packagejson": "<%= this.dependabotPackageJson.devDependencies['prettier-plugin-packagejson'] %>"
+    "prettier": "<%= this.dependabotPackageJson.devDependencies?.['prettier'] %>",
```
Tcharl, line 35: Don't get why it is optional

generators/base/utils.mjs
mshima: We will have at least 1 more function there, imo is ok to just call utils.
We can change later when the api become more complete.

File names are not important to apis anymore.
Package.json exports field defines the api now.
We can rearrange at will even after v8 release.

generators/init/templates/package.json.ejs
mshima: Used to be, but with typescript, I don't fell we should add a field called NODE_VERSION to the type..

mshima: Most of constants will be changed to variables.
Versions will be loaded from pom and Dockerfile.

generators/server/generator.mjs
mshima: Config files are supposed to be dirty at this point.
It's possible to create a class with getters for this purpose.
Or investigate the pros and cons of moving composing priority to after preparing priority.

Eg: the server generator changes the webappSourceDir and webappTestDir (dotnet blueprint), at preparing webappAppDir can be created with `${webappSourceDir}/app`. It may be be outdated if composing is done after preparing.

generators/server/templates/package.json.ejs
mshima: This is temporary.
For tests, `dependabotPackageJson` is loaded at the common generator, which are mocked.
So it's missing at tests.

In any case, we should create a `nodeDependencies` object to replace it. Like we have for java:
https://github.com/jhipster/generator-jhipster/blob/1ec2ebbe7fb7fb147654369d9f9164ca6c2e6553/generators/bootstrap-application-server/types.d.ts#L77

generators/base/utils.mjs
Tcharl: I really prefer to have clear file names (strict OOP) that tells what it does compared to surprises

PR #20107, opened 2022-10-25
Use Spring Security 6's authorizeHttpRequests

mraible:
Spring Security 6 deprecates `authorizeRequests()` and `authorizeHttpRequests()` adds an implicit `anyRequests().denyAll()` to the bottom of the mappings. Therefore, we need to allow all the patterns that can possibly occur in the UI. 

I think moving from allow-by-default to deny-by-default by Spring Security 6 is an excellent way to increase security awareness. 👍 

---


mraible:
I have a meeting with @rwinch next week to talk about this and determine the best path forward. 

FWIW, the WebFlux implementation in JHipster uses a filter for SPA routes and doesn't require all the extra allow rules. 

https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/java/package/web/filter/SpaWebFilter.java.ejs

mraible:
I had a zoom call with @rwinch this week and he provided a number of recommendations to improve JHipster's security configuration:

- `csrf().disable()` applies to everything, not just H2 - is this line still necessary? 
- Does `CorsFilter` really need to be registered or why does the default ordering not work?
- The line with `HttpMethod.OPTIONS` is probably not necessary since this is handled by `CorsFilter`
- We should re-order `requestMatchers` so more restrictive lines are first
- Do not use `securityContextHolder(Strategy).getContext().setAuthentication(authentication)` as it can cause race conditions.
- `SecurityContextHolder.setContext()` should be changed to strategy because allows observability. Currently, it seems we only use it in tests. 
- For serving up a SPA - is a filter or controller better? Rob said he'd mull on this one for a bit.

mraible:
I updated this PR to fix the allowed URLs and changed from a controller to a filter for serving up SPA files. I think there's a couple more enhancements that this needs, that's why I'm adding a bug bounty:

- Add a `SpaWebFilterTest` class that tests the Spring MVC version and the WebFlux version.
- Refactor both `SecurityConfiguration` files to use lambdas instead of `and()`.

atomfrede:
I remember something for integration tests to put certain test data in or change something we don't expose via regular apis. But I also think the endpointnis long gone

mraible:
It was used by `ClientForwardControllerTest`, which I removed in this PR. Speaking of that, where do I add/remove files for the upgrade sub-generator?

I'd still like to add tests for `SpaWebFilter`. I did some research today and it seems like we can do a unit test for MVC because there's MockMvc. However, for WebFlux, we might have to make it an integration test to use `WebTestClient`. 

On a related note: SPA Web seems kinda redundant. Should we rename it to `SpaFilter`?

mraible:
Does anyone know how to prevent line wrapping with Prettier? I refactored the SecurityConfiguration classes to use lambdas and the formatting makes the code hard to read now.

Before (desired format):

```java
.requestMatchers("/", "/index.html", "/*.js", "/*.map", "/*.css").permitAll()
.requestMatchers("/*.ico", "/*.png", "/*.svg", "/*.webapp").permitAll()
.requestMatchers("/app/**").permitAll()
.requestMatchers("/i18n/**").permitAll()
```

Prettier makes this all look like this instead:

```java
.requestMatchers("/", "/index.html", "/*.js", "/*.map", "/*.css")
.permitAll()
.requestMatchers("/*.ico", "/*.png", "/*.svg", "/*.webapp")
.permitAll()
.requestMatchers("/app/**")
.permitAll()
.requestMatchers("/i18n/**")
.permitAll()
```

mraible:
@atomfrede If I add it like the following:

```java
.authorizeHttpRequests(authz ->
    authz
        // prettier-ignore
        .requestMatchers("/", "/index.html", "/*.js", "/*.map", "/*.css").permitAll()
```

And run `npm run prettier:format`, it still wraps:

```java
.authorizeHttpRequests(authz ->
    authz
        // prettier-ignore
        .requestMatchers("/", "/index.html", "/*.js", "/*.map", "/*.css")
        .permitAll()
```

I like that this PR removed `// @formatter:off` and `and()`, but this formatting makes the new way almost worse.

atomfrede:
Have you tried an ignore block? https://prettier.io/docs/en/ignore.html#range-ignore

mraible:
I got it to work. It's all about where you put the comment. Here's what works:

```java
.authorizeHttpRequests(authz ->
    // prettier-ignore
    authz
        .requestMatchers("/", "/index.html", "/*.js", "/*.map", "/*.css").permitAll()
```

mraible:
I'm having a hard time getting WebSocket support to work. It seems to forward to `/login` when a WebSocket request is made and this results in an infinite loop. I think it should just return a 401, but not sure.

mraible:
I checked the `main` branch and WebSocket support is broken there too. It works if I generate the same app with v7.9.3. If I remove the following code from `SecurityConfiguration`, I see the same behavior in 7.9.3.

```java
.exceptionHandling()
    .authenticationEntryPoint(problemSupport)
    .accessDeniedHandler(problemSupport)
```

To me, this means the new Spring Boot 3 problem support is not fully integrated with Spring Security 6. I'm not sure how to configure exception handling with the new support. Does anyone else know?

mshima:
@mraible imperative microservices seem to be failing.

mraible:
@mshima Yeah, this PR is far from complete. I still need to figure out how to return a 401 instead of a 302 for WebSockets to work. I believe this is related to the new problem support. Also, I think it'd be good to include tests because manually testing everything is a real pain. I don't plan to dedicate much time to this before the holidays and I'll be on vacation until January 9.

mraible:
FWIW, I hope to look into the Spring Security + WebSocket issues this week. For some reason, the exception handlers are working for WebSocket security and it's returning a 302 redirect instead of a 401.

mraible:
Looking at the [vue-gradle-h2mem-ws-session job](https://github.com/jhipster/generator-jhipster/actions/runs/4027872247/jobs/6924320307#logs) the failure doesn't seem to be related to Spring Security anymore.

```
java.lang.ClassNotFoundException: org.springframework.boot.SpringApplication
```

Any ideas?

If I re-create `vue-gradle-ws-session` locally (just the shell, not with entities), it seems to work fine. Here's the `.yo-rc.json`:

```json
{
  "generator-jhipster": {
    "baseName": "jhipster",
    "blueprints": [],
    "buildTool": "gradle",
    "cacheProvider": "ehcache",
    "clientFramework": "vue",
    "clientTheme": "none",
    "clientThemeVariant": "primary",
    "creationTimestamp": 1594452153587,
    "databaseType": "sql",
    "devServerPort": 9060,
    "dtoSuffix": "DTO",
    "enableGradleEnterprise": false,
    "enableHibernateCache": true,
    "enableTranslation": true,
    "entities": [],
    "entitySuffix": "",
    "jhiPrefix": "",
    "jhipsterVersion": "7.9.3",
    "languages": ["en"],
    "messageBroker": "no",
    "nativeLanguage": "en",
    "otherModules": [],
    "packageFolder": "tech/jhipster/sample",
    "packageName": "tech.jhipster.sample",
    "pages": [],
    "prodDatabaseType": "mysql",
    "reactive": false,
    "searchEngine": "no",
    "serverPort": "8080",
    "serverSideOptions": [],
    "serviceDiscoveryType": "no",
    "skipCheckLengthOfIdentifier": false,
    "skipClient": false,
    "skipFakeData": false,
    "skipServer": false,
    "skipUserManagement": false,
    "testFrameworks": ["gatling", "cypress"],
    "withAdminUi": true
  }
}
```

mshima:
> Looking at the [vue-gradle-h2mem-ws-session job](https://github.com/jhipster/generator-jhipster/actions/runs/4027872247/jobs/6924320307#logs) the failure doesn't seem to be related to Spring Security anymore.
> 
> ```
> java.lang.ClassNotFoundException: org.springframework.boot.SpringApplication
> ```

This error started with spring-boot 3.0.2 update.

mshima:
By the way vue session was fixed at PR https://github.com/jhipster/generator-jhipster/pull/20749.
More specific at line https://github.com/jhipster/generator-jhipster/blob/8d55f965b4a51cc8e89f18318aaa2d11b11f76ba/generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs#L228.
Should be the same for websocket.

mraible:
@mshima I tried creating the `vue-gradle-ws-session` app locally again today. It does not have the following error like CI does.

```
ClassNotFoundException: org.springframework.boot.SpringApplication
```

Instead, it's `account/login-page.cy.ts` test fails.

login modal -- requires username (failed)

I'm testing it with the following commands:

```
npm run ci:e2e:package
npm run ci:e2e:prepare
npm run ci:e2e:run
```

mshima:
This commit may fix the SpringApplication problem: https://github.com/jhipster/jhipster-bom/commit/826367d66501bac14fe28eeb50bfcb0335eadf2d

generators/server/templates/src/main/java/package/config/SecurityConfiguration.java.ejs
```diff
@@ -232,7 +233,18 @@ public class SecurityConfiguration {
...
+                            new AntPathRequestMatcher("/websocket/**"),
+  <%_ } _%>
+                            new AntPathRequestMatcher("/api/**"),
+                            /*
+                             * FormLogin is redirecting to "/error" instead of just passtrough.
+                             * Error started at https://github.com/jhipster/generator-jhipster/pull/20107.
+                             * Once the redirect cause is found this can be dropped in favor of the above "/api/**" entrypoint.
+                             */
+                            new AntPathRequestMatcher("/error")
+                        )
```
mraible, line 247: This workaround seems like a bug. Should we enter something in the Spring Security project's issues?

mshima: I am investigating

mshima: It was hard, but is fixed.

mraible: Excellent! I'll increase the bug bounty because of the difficulty.

mraible:
It looks like several tests are still failing, mostly microservices. I'm going to create `mf-ng-eureka-jwt-psql-ehcache` locally and see if I can fix the problem. After all tests are passing, I'll work on writing unit tests for `SpaWebFilter`.

mshima:
> It looks like several tests are still failing, mostly microservices. I'm going to create `mf-ng-eureka-jwt-psql-ehcache` locally and see if I can fix the problem. After all tests are passing, I'll work on writing unit tests for `SpaWebFilter`.

@mraible the problem is imperative microfrontends.
For both jwt or oauth2.
Imperative microservices and reactive (microservice or microfrontend) are ok.

mshima:
I think I've found the problem, I will test.

mraible:
@mshima Generating the projects seems to be hanging at:

```
INFO! Generator app child process exited with code 0
INFO! Application successfully committed to Git from /Users/mraible/Downloads/mf-ng-eureka-jwt-psql-ehcache/blog.
Congratulations, JHipster execution is complete!
Sponsored with ❤️  by @oktadev.
INFO! Generator app child process exited with code 0
INFO! Generating 1 deployment.
INFO! Generating deployment docker-compose in a new parallel process
```

I'm using Node v18.13.0. 

Also, it seems like there's some line breaks now for logging messages:

```
INFO!
Server application generated successfully.

INFO! Run your Spring Boot application:
./mvnw
INFO! Start your Webpack development server with:
 npm start

INFO!
Client application generated successfully.
```

The `INFO!` and following message used to be on the same line. In reality, I don't think the `INFO!` part is needed, but maybe it's useful for terminals that don't have colorized output?

mshima:
@mraible worked for me on M1.
```
% node --version
v18.13.0
```

Maybe `npm ci`?

mshima:
I plan to take a look at the logger.

mshima:
> I think I've found the problem, I will test.

Looks like it worked.

mraible:
> Looks like it worked!

Nice! 

I started writing a `SpaWebFilterTest` for Spring MVC. All of the tests pass in the `ng-default` app, but many of the tests from the deleted `ClientForwardControllerTest.java` do not pass. 

Most of the failures are understandable because the URLs have dots in them, but the `getBackendEndpoint()` test is kinda strange. It seems like the filter is getting invoked before the controller where ideally it's the other way around. In the test below, the tests that don't pass are marked with `@Disabled`. 

```java
package tech.jhipster.sample.web.filter;

import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Disabled;
import org.junit.jupiter.api.Test;
import org.springframework.http.MediaType;
import org.springframework.test.web.servlet.MockMvc;
import org.springframework.test.web.servlet.ResultActions;
import org.springframework.test.web.servlet.setup.MockMvcBuilders;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;

import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.forwardedUrl;
import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;

class SpaWebFilterTest {

    private final SpaWebFilter filter = new SpaWebFilter();

    private MockMvc mockMvc;

    @BeforeEach
    public void setup() {
        this.mockMvc = MockMvcBuilders
            .standaloneSetup(new TestController())
            .addFilter(filter)
            .build();
    }

    @Test
    void testFilterForwardsToIndexHtml() throws Exception {
        mockMvc.perform(get("/"))
            .andExpect(status().isOk())
            .andExpect(forwardedUrl("/index.html"));
    }

    @Test
    void testFilterDoesNotForwardToIndexHtmlForApi() throws Exception {
        mockMvc.perform(get("/api"))
            .andExpect(status().isOk());
    }

    @Test
    void testFilterDoesNotForwardToIndexHtmlForManagement() throws Exception {
        mockMvc.perform(get("/management"))
            .andExpect(status().isOk());
    }

    @Test
    void testFilterDoesNotForwardToIndexHtmlForServices() throws Exception {
        mockMvc.perform(get("/services"))
            .andExpect(status().isOk());
    }

    @Test
    void testFilterDoesNotForwardToIndexHtmlForV3ApiDocs() throws Exception {
        mockMvc.perform(get("/v3/api-docs"))
            .andExpect(status().isOk());
    }

    @Test
    void testFilterDoesNotForwardToIndexHtmlForDotFile() throws Exception {
        mockMvc.perform(get("/file.js"))
            .andExpect(status().isOk());
    }

    @Test
    @Disabled
    void getBackendEndpoint() throws Exception {
        mockMvc.perform(get("/test"))
            .andExpect(status().isOk())
            .andExpect(content().contentTypeCompatibleWith(MediaType.TEXT_PLAIN_VALUE))
            .andExpect(content().string("test"));
    }

    @Test
    void forwardUnmappedFirstLevelMapping() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first-level"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    @Disabled
    void forwardUnmappedDottedFirstLevelWithTrailingSeparator() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first.level/"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    void forwardUnmappedSecondLevelMapping() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first-level/second-level"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    @Disabled
    void forwardUnmappedDottedSecondLevelWithTrailingSeparator() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first.level/second.level/"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    void forwardUnmappedThirdLevelMapping() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first-level/second-level/third-level"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    @Disabled
    void forwardUnmappedDottedThirdLevelWithTrailingSeparator() throws Exception {
        ResultActions perform = mockMvc.perform(get("/first.level/second.level/third.level/"));
        perform.andExpect(status().isOk()).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    void forwardUnmappedDeepMapping() throws Exception {
        mockMvc.perform(get("/1/2/3/4/5/6/7/8/9/10")).andExpect(forwardedUrl("/index.html"));
    }

    @Test
    void getUnmappedFirstLevelFile() throws Exception {
        mockMvc.perform(get("/foo.js")).andExpect(status().isNotFound());
    }

    @Test
    void getUnmappedSecondLevelFile() throws Exception {
        mockMvc.perform(get("/foo/bar.js")).andExpect(status().isNotFound());
    }

    @Test
    void getUnmappedThirdLevelFile() throws Exception {
        mockMvc.perform(get("/foo/another/bar.js")).andExpect(status().isNotFound());
    }

    @RestController
    public static class TestController {

        @RequestMapping(value = "/test")
        public String test() {
            return "test";
        }

        @RequestMapping(value = "/api")
        public String api() {
            return "api";
        }

        @RequestMapping(value = "/management")
        public String management() {
            return "management";
        }

        @RequestMapping(value = "/file.js")
        public String file() {
            return "file.js";
        }

        @RequestMapping(value = "/services")
        public String services() {
            return "services";
        }

        @RequestMapping(value = "/v3/api-docs")
        public String apiDocs() {
            return "api-docs";
        }
    }
}
```

mraible:
I should add the `getBackendEndpoint()` test fails with the following:

```
java.lang.AssertionError: Content type not set
```

So maybe the mapping is working? Changing the request mapping annotation to the following does not help.

```java
@RequestMapping(value = "/test", produces = MediaType.TEXT_PLAIN_VALUE)
```

mshima:
There is differences between the controller implementation and the filter one.
We should adjust the filter or adjust the test.
Due to `!path.contains(".")`, an url that contains `.` will never be redirected to `index.html`. It will always be considered a file resource instead of a page.

mraible:
I decided to leave the order of the filter as-is. If Spring Security 6 is deny-by-default (vs Spring Security 5 that was allow-by-default), I think it makes sense to return a 403 for a request path that's not accounted for. I changed some tests to confirm access is denied and included a comment on how to make it return a 404. 

```java
/**
 * This test verifies that any files that aren't permitted by Spring Security will be forbidden.
 * If you want to change this to return isNotFound(), you need to add a request mapping that
 * allows this file in SecurityConfiguration.
 */
@Test
void getUnmappedSecondLevelFile() throws Exception {
    mockMvc.perform(get("/foo/bar.js")).andExpect(status().isForbidden());
}
```

I created https://github.com/jhipster/generator-jhipster/issues/21003 and added a bug bounty to create a WebFlux version of this test.

mraible:
@mshima Does the removal of `ClientForwardController.java` and its test require changes to the upgrade sub-generator? I didn't see any examples of removing server files in the existing `cleanup.mjs`.

mshima:
> @mshima Does the removal of `ClientForwardController.java` and its test require changes to the upgrade sub-generator? I didn't see any examples of removing server files in the existing `cleanup.mjs`.

It have been modularized, so it's in generators/server/cleanup.mjs
`generators/cleanup` will be modularized soon.

generators/server/templates/src/test/java/package/web/filter/SpaWebFilterIT.java.ejs
```diff
@@ -0,0 +1,104 @@
...
+        mockMvc.perform(get("/"))
+            .andExpect(status().isOk())
+            .andExpect(forwardedUrl("/index.html"));
+    }
+<%_ if (!skipUserManagement) { _%>
+
+    @Test
+    void testFilterDoesNotForwardToIndexForApi() throws Exception {
+        mockMvc.perform(get("/api/<% if (!applicationTypeMicroservice) { %>authenticate<% } else { %>users<% } %>"))
+            .andExpect(status().isOk());
```
mshima, line 34: Should we make sure `.andExpect(forwardedUrl("/index.html"))` is false?

mshima (approved):

generators/server/templates/src/test/java/package/web/filter/SpaWebFilterIT.java.ejs
mshima: Great, look ok if passes.

mshima:
`v3/apidocs` should be available for microservices too, but we can adjust later.

mraible:
> `v3/apidocs` should be available for microservices too

I used the same logic that's in `SpaWebFilter`: https://github.com/jhipster/generator-jhipster/blob/main/generators/server/templates/src/main/java/package/web/filter/SpaWebFilter.java.ejs#L35

mraible:
@mshima Please collect the bug bounty for this since you figured out the hard parts.

mshima:
@mraible we should split.

mraible:
@mshima I'd agree with you if I worked on this after hours or on the weekends. However, I worked on it during work hours and since Okta is a sponsor, it seems weird for me to get paid twice. 🙂

mshima:
@DanielFran  bounty claimed https://opencollective.com/generator-jhipster/expenses/123272

DanielFran:
@mshima done

Issue #19878, opened 2022-09-30
Drop output path customizer.


mshima:
##### **Overview of the feature request**

It was added due to independent workflows for each generator.
We can use a shared clientSrcDir now.

##### **Motivation for or Use Case**

Cleanup

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

PR #20151, opened 2022-10-29
First PR - export some methods from the base generator

Tcharl:
Tries to centralize some common methods into their respective ts files
---


generators/base/generator-base.mjs
```diff
@@ -322,21 +314,12 @@ export default class JHipsterBaseGenerator extends PrivateBase {
    * @param {string} outputPath - Path to customize.
    */
   applyOutputPathCustomizer(outputPath) {
```
mshima, line 316: This will be removed, don't worth refactoring.
See https://github.com/jhipster/generator-jhipster/issues/19878.

generators/base/generator-base.mjs
```diff
@@ -59,6 +58,12 @@ import {
 import databaseData from '../sql-constants.mjs';
 import { CUSTOM_PRIORITIES } from './priorities.mjs';
 import { GENERATOR_BOOTSTRAP } from '../generator-list.mjs';
+import locateGenerator from './generator/blueprint-structure/generator-locator.mjs';
+import isBuiltInUserConfiguration from './generator/config/user.mjs';
+import isUsingBuiltInAuthorityConfiguration from './generator/config/authority.mjs';
```
mshima, line 63: We should extract those function.

- these methods should be moved to `base-application` generator.
- I prefer to name the file with the function name if we are using `export default`.
- add an index api aggregator to reduce imported files.
```
import { isBuiltInUserConfiguration } from './generator/config/index.mjs';
```
```suggestion
import isBuiltInUserConfiguration from './generator/config/user.mjs';
import isUsingBuiltInAuthorityConfiguration from './generator/config/authority.mjs';
```

generators/base/generator/config/authority.mts
```diff
@@ -0,0 +1,32 @@
...
+import { OAUTH2 } from '../../../../jdl/jhipster/authentication-types.js';
+import databaseTypes from '../../../../jdl/jhipster/database-types.js';
+
+const { COUCHBASE, MONGODB, NEO4J, SQL, NO: NO_DATABASE } = databaseTypes;
+
+const isUsingBuiltInAuthorityConfiguration = (config: any): boolean => {
+  return (
+    config ||
+    (!config.skipUserManagement && [SQL, MONGODB, COUCHBASE, NEO4J].includes(config.databaseType)) ||
+    (config.authenticationType === OAUTH2 && config.databaseType !== NO_DATABASE)
```
mshima, line 28: These conditions should be changed:

`skipUserManagement` should be split into something like `generateUserManagement` and `generateUserModel`.
It's quite confusing that oauth2 forces `skipUserManagement`, but adds a lot of workarounds to generate the user model for relationships.

generators/base/generator/config/user.mts
```diff
@@ -0,0 +1,33 @@
...
+import databaseTypes from '../../../../jdl/jhipster/database-types.js';
+
+const { NO: NO_DATABASE } = databaseTypes;
+const { OAUTH2 } = authenticationTypes;
+
+const isBuiltInUserConfiguration = (config: any): boolean => {
+  return (
+    !config ||
+    (!config.skipUserManagement && config.databaseType !== NO_DATABASE) ||
+    (config.authenticationType === OAUTH2 && config.databaseType !== NO_DATABASE)
```
mshima, line 29: same as the user logic

generators/base/generator-base.mjs
```diff
@@ -255,11 +255,7 @@ export default class JHipsterBaseGenerator extends PrivateBase {
    * @return {boolean} true if the User is built-in.
    */
   isUsingBuiltInUser() {
```
mshima, line 257: We should remove those apis.

generators/base/generator-base.mjs
Tcharl: The goal is to have multiple methods in the file, I'm at the very beginning.

An idea why the mts module isn't importable (javascript is far from being my specialty)?

mshima: > The goal is to have multiple methods in the file, I'm at the very beginning.

Don't use export default then.
 
> An idea why the mts module isn't importable (javascript is far from being my specialty)?

Bug at loader.
https://github.com/jhipster/generator-jhipster/blob/0751d385f0da95e3efdbda2a352ab1a31ccbdb55/generators/base/index.mjs#L27-L28

Tcharl: :'(

generators/base/generator-base.mjs
Tcharl: Let's do it step by step: I don't want to break anything during the refacto...

mshima: Implemented at https://github.com/jhipster/generator-jhipster/pull/20277.
Needed to change everywhere that uses constants with configurable veritable.

generators/base/generator-base.mjs
```diff
@@ -346,25 +322,21 @@ export default class JHipsterBaseGenerator extends PrivateBase {
    * @param {string} packageJsonSourceFile - Package json filepath with actual versions.
    */
   replacePackageJsonVersions(keyToReplace, packageJsonSourceFile) {
```
mshima, line 324: This is not use anywhere can safely be dropped.

Tcharl:
Hi @mshima ,

- breaks dependabot. Don't get it, can you please elaborate?
- changes apis that should be removed including: not the purpose of this PR, its purpose is just to extract logic to their relevant generators
- add app.jdl in root folder: My bad, will delete now

DanielFran:
> I don't understand why this was merged.
> 
>     * breaks dependabot. Snapshot cannot have versions.
> 
>     * changes apis that should be removed including [Drop output path customizer. #19878](https://github.com/jhipster/generator-jhipster/issues/19878) marked to be removed and will be removed at [Drop output path customizer and protractor. #20277](https://github.com/jhipster/generator-jhipster/pull/20277).
> 
>     * add app.jdl in root folder.
> 
> 
> IMO we should revert for now.

Yes, this is failing the main branch, @pascalgrimaud  already inform me that main was broken...
@Tcharl can you revert it please? Actually in windows I cannot test it...

Tcharl:
@DanielFran @mshima please tell me what's wrong?

To me snapshot should have version: it's snapshot, meaning that this should be exactly what the user sees (and for sure he sees versions in his package.json.
Also, I didn't change (or it was infortunate) anything regarding the actual business logic, just extracted methods: the misbehavior should be fixable.

I also fix the last bullet in the upcoming PR https://github.com/jhipster/generator-jhipster/pull/20286

mshima:
> @DanielFran @mshima please tell me what's wrong?
> 
> To me snapshot should have version: it's snapshot, meaning that this should be exactly what the user sees (and for sure he sees versions in his package.json. Also, I didn't change (or it was infortunate) anything regarding the actual business logic, just extracted methods: the misbehavior should be fixable.

While I agree with that, ideally versions should appear at the snapshot, dependabot will not update snapshots for us. So we must replace versions with placeholders, the base structure will be tested.

I will click at the revert button.

.mocharc.cjs
```diff
@@ -5,7 +5,7 @@ module.exports = {
   timeout: 30000,
   ui: 'bdd',
   extension: ['js', 'spec.cjs', 'spec.mjs', 'spec.ts', 'spec.mts'],
-  require: ['mocha-expect-snapshot/old-format', 'test/mocha.config.cjs'],
+  require: ['mocha-expect-snapshot/old-format', 'test/mocha.config.mjs'],
```
mshima, line 8: It’s not possible to require an esm (mjs) module. It must be cjs.

Tcharl: fixed on the new, non reverted PR

mshima: sorry but updating snapshots will not fix the dependabot problem.

PR #20201, opened 2022-11-05
fix cypress race condition that causes random failures at react frontend 

mshima:
A race condition is probably happening where cypress tries to clear an input that’s not yet loaded. The end result is the old value concatenated with the new value.

The fix is to wait the input to have the expected value before clearing it.

---


mshima:
Please squash when merging

DanielFran (approved):

PR #20316, opened 2022-11-15
Rework cypress settings page.

mshima:
---

### DIFF
--- a/generators/cypress/templates/src/test/javascript/cypress/e2e/account/settings-page.cy.ts.ejs
+++ b/generators/cypress/templates/src/test/javascript/cypress/e2e/account/settings-page.cy.ts.ejs
@@ -68,7 +68,10 @@ describe('/account/settings', () => {
     cy.authenticatedRequest({
       method: 'POST',
       url: '/api/account',
-      body: originalUserAccount,
+      body: {
+        ...originalUserAccount,
+        email: testUserEmail,
+      },
     })
       .its('status')
       .should('eq', 200);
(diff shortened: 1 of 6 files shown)


generators/cypress/templates/src/test/javascript/cypress/e2e/account/settings-page.cy.ts.ejs
```diff
@@ -22,22 +22,48 @@ import {
...
+
+  before(() => {
     cy.login(username, password);
-    cy.visit('/account/settings');
+
+    cy.getAccount().then(account => {
+      originalUserAccount = account;
+      testUserAccount = { ...account, email: testUserEmail };
+
+      // need to modify email because default email does not match regex in some frameworks
```
mraible, line 44: Why can't we just fix the regex in some frameworks?

mraible (approved):

generators/cypress/templates/src/test/javascript/cypress/e2e/account/settings-page.cy.ts.ejs
mshima: `localhost` is not a valid domain for emails.
Not sure which approach to use.

mraible: We could use `example.com`, since that's what it's designed for. https://example.com/

Or, any of the jhipster domains. For example, I own jhipster.org and just redirect it to jhipster.tech.

mshima: Looks good to use example.com.
I will create a issue about this.

PR #20442, opened 2022-11-23
remove unused testcontainers property

ctamisier:
`testcontainers.reuse.enable` in testcontainers.properties file is not used as it is a per environment property (not a per project property)

see here: https://github.com/testcontainers/testcontainers-java/issues/3780

We could also delete testcontainers.properties file but some developers may have defined some other properties in this file (?)

I realized this when I saw this docker logs during startup:
_Reuse was requested but the environment does not support the reuse of containers
To enable reuse of containers, you must set 'testcontainers.reuse.enable=true' in a file located at /Users/ctamisier/.testcontainers.properties_

So when `testcontainers.reuse.enable=true` is defined in ~/.testcontainers.properties the docker containers are indeed not removed after tests.

### DIFF
--- a/generators/cassandra/__snapshots__/generator.spec.mts.snap
+++ b/generators/cassandra/__snapshots__/generator.spec.mts.snap
@@ -452,9 +452,6 @@ exports[`generator - cassandra gateway-jwt-gradle-enableTranslation(true)-com.pa
   "src/test/resources/templates/mail/testEmail.html": {
     "stateCleared": "modified",
   },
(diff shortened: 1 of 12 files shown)


mshima:
I think we should drop the file entirely

DanielFran:
@ctamisier I agree with @mshima; better to remove the file since it is not used anywhere.

mraible:
@ctamisier Can you please delete the file `generators/server/templates/src/test/resources/testcontainers.properties.ejs` entirely for this PR?

PR #20463, opened 2022-11-28
sb3: bump elasticseach container to v8.5.0

mshima:
---

### DIFF
--- a/generators/docker-compose/__snapshots__/docker-compose.spec.mts.snap
+++ b/generators/docker-compose/__snapshots__/docker-compose.spec.mts.snap
@@ -582,6 +582,7 @@ eureka:
     environment:
       - ES_JAVA_OPTS=-Xms256m -Xmx256m
       - discovery.type=single-node
+      - xpack.security.enabled=false
     healthcheck:
       test:
         - CMD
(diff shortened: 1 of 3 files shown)


mraible (approved):

mshima:
@mraible actually, this is due to diff logic applied to the workflows. If there is no change to the generated source, the test is skipped.
It doesn’t fix anything, every elasticsearch project broke, while others didn’t change.

mraible:
So Elasticsearch v7.17.4 works fine with Spring Boot 3, but v8.5.0 does not? Maybe @xeraa can help.

mshima:
@mraible no it's just the docker health check that is not working.
https://github.com/jhipster/generator-jhipster/blob/75898e4860ed9850d697a5c0ba34f47b77559e14/generators/docker/templates/src/main/docker/elasticsearch.yml.ejs#L35

xeraa:
So, since 8.0 security is enabled by default. That means HTTPS only and a random password for the `elastic` user. Which is generally good, but would make your setup quite a bit more complex. And the health check is the first thing failing — not entirely surprising 😅

You could disable security for this scenario (**please, don't do this in general**)? I think this should get you sorted:

    environment:
      - xpack.security.enabled=false

mraible (approved):

PR #20498, opened 2022-12-02
Mark GenerateGitPropertiesTask as non cacheable

jprinet:
The `GenerateGitPropertiesTask` Gradle task pulled by the `com.gorylenko.gradle-git-properties` plugin is executed systematically as using `upToDateWhen { false }`.

Marking the task as non cacheable is interesting to first avoid overhead and second ease detection of performance regression as this one won't appear anymore as a cache miss.


atomfrede (approved):
Thanks for the contribution.

PR #20590, opened 2022-12-12
Revert "Update liquibase version to 4.18.0"

mshima:
Check if h2 regression is caused by liquibase.

Reverts jhipster/generator-jhipster#20532


mshima:
`vue-gradle-h2mem-ws-session` is still failing.

PR #20638, opened 2022-12-19
WIP - migrate cli to esm, second try

Tcharl:
Second try with  a focus on cli esm.

There are a lot of issues with the current jdl ts architecture that do not let the generator or the cli to resolve jdl files or compile (or is it the loader that does not resolves things well?).

I ended up with some errors that mentions that some modules do not have any 'default' export even if they have.

I'm not the 'node-loaders' initiator and really newbie on ESM, help appreciated (or I will for sure make some ugly hacks to fix the behavior)...

Also removes 'false' for [searchEngine and serviceDiscovery](https://github.com/jhipster/generator-jhipster/issues/20656) (unfortunately, was really difficult to avoid tackling it for many reasons)

---


mshima:
@Tcharl I checked this branch.
For the cli migration we will need a new version of yeoman-test.
I’ve reverted the jdl chances for now so we can focus on the cli migration.
I will push my branch and create a PR.

Tcharl:
@mshima don't revert everything please:
 - For sure, the use of named export instead of default export should be reverted: this was a hack due to some weird bugs I encountered
 - The lowercasing of const shouldn't be reverted, these are Typescript convention
 - The centralization of all calls on index.mts is lifesaving regarding what's tried to be achieved here: I'm banging my head on it since a month and this looks to be a life saver!

PR #20657, opened 2022-12-22
Convert cli to esm

mshima:
---


cli/import-jdl.mjs
```diff
@@ -16,22 +16,23 @@
...
+import EnvironmentBuilder from './environment-builder.mjs';
+import { CLI_NAME, GENERATOR_NAME, logger, toStringJoinArgs, printSuccess, getOptionAsArgs } from './utils.mjs';
+import { packageJson as packagejs } from '../lib/index.cjs';
+import statistics from '../generators/statistics.cjs';
+import { JHIPSTER_CONFIG_DIR } from '../generators/generator-constants.mjs';
+import { writeConfigFile } from './export-utils.cjs';
+
+const require = createRequire(import.meta.url);
+
+const jhipsterCli = require.resolve('./cli.mjs');
```
Tcharl, line 35: No solution to use `import()` here?

Tcharl:
Awesome! Thank you so much: the internals of esm are so hard to get

mshima:
Still needs some work.

cli/import-jdl.mjs
mshima: I will switch to path.join. Import.resolve is experimental.

generators/upgrade/generator.mjs
```diff
@@ -30,10 +30,9 @@ import BaseGenerator from '../base/index.mjs';
 import { upgradeFiles } from '../cleanup.mjs';
 import constants from '../generator-constants.cjs';
 import statistics from '../statistics.cjs';
-import blueprintUtils from '../../utils/blueprint.cjs';
+import { parseBluePrints } from '../../utils/blueprint.mjs';
```
Tcharl, line 33: `parseBlueprint`?

mshima: The original is parseBluePrints.

Tcharl: Yes, but we're preparing V8 aren't we?

mshima: Don’t analyzed if we need to rename. Can rename later if makes sense.

mshima:
Further improvements can be done later.
Should be ready now.

DanielFran (approved):

PR #20839, opened 2023-01-17
cleanup config before applying defaults

mshima:
When prompts are ignored, a null value is set to the store so subsequent prompts are ignored (`--ask-answered` option will show those questions again).

A null value at config will override the default value with the null value, causing a wrong config to be applied.
This PR removes undefined and null values so the default value will be correctly be returned.

---

### DIFF
--- a/generators/base/support/config.mts
+++ b/generators/base/support/config.mts
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2013-2023 the original author or authors from the JHipster project.
+ *
+ * This file is part of the JHipster project, see https://www.jhipster.tech/
+ * for more information.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const filterUndefinedAndNullValues = value => value !== undefined && value !== null;
+
+/**
+ * Copy and remove null and undefined values
+ * @param object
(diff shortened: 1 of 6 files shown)


generators/base/support/config.mts
```diff
@@ -0,0 +1,47 @@
...
+
+const filterUndefinedAndNullValues = value => value !== undefined && value !== null;
+
+/**
+ * Copy and remove null and undefined values
+ * @param object
+ * @returns
+ */
+// eslint-disable-next-line import/prefer-default-export
+export function deepCleanup(
```
Tcharl, line 28: This method should ideally be split into two: deepcleanup (which filters null and undefined, without any 'filtervalue' argument), relying on filterValue which has this signature

generators/entity/generator.mjs
```diff
@@ -149,7 +150,10 @@ export default class EntityGenerator extends BaseGenerator {
         this.loadClientConfig(undefined, this.application);
         this.loadTranslationConfig(undefined, this.application);
         // Try to load server config from microservice side, falling back to the app config.
-        this.loadServerConfig(getConfigWithDefaults({ ...this.jhipsterConfig, ...this.microserviceConfig }), this.application);
+        this.loadServerConfig(
+          getConfigWithDefaults({ ...deepCleanup(this.jhipsterConfig), ...deepCleanup(this.microserviceConfig ?? {}) }),
```
Tcharl, line 154: {...deepCleanup(getConfigWithDefaults(this.jhipsterConfig, this.microserviceConfig ?? {}))}

Tcharl: Also, should be added to all load, shouldn't it?

mshima: > {...deepCleanup(getConfigWithDefaults(this.jhipsterConfig, this.microserviceConfig ?? {}))}

No, cleanup should be applied before spreading and before defaults.

> Also, should be added to all load, shouldn't it?

Yes, but in this case my plan is to create a custom storage that applies defaults at read. So jhipsterConfig will have defaults applied and jhipsterConfigWithDefaults will be dropped.
This will be another PR.

Tcharl: forget about it

DanielFran (approved):

Issue #19114, opened 2022-07-08
Integrate languages inside feature specific generator.


mshima:
##### **Overview of the feature request**

Implement language inside feature specific generator like https://github.com/jhipster/generator-jhipster/blob/main/rfcs/4-jhipster-rfc-entity-as-core.md.

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

PR #20900, opened 2023-01-24
rework languages generator

mshima:
- rename LANGUAGES to `supportedLanguages`.
- `supportedLanguages` contains the calculated technology specific language/locale constant instead of relying on methods to convert at runtime.
- `supportedLanguages` allows a null value for technology specific language/locale so it can be ignored or have a fallback when the language is not officially supported.
- an error is thrown if the file containing the needle is not found.
- each language needle support was moved to the specific generator support (angular, react, vue, client - dayjs, server).
- generic languages utilities are moved to the languages/support and written in typescript.

---


generators/languages/generator.mjs
```diff
@@ -21,16 +21,19 @@ import chalk from 'chalk';
...
 import statistics from '../statistics.cjs';
 
 import { GENERATOR_LANGUAGES, GENERATOR_BOOTSTRAP_APPLICATION } from '../generator-list.mjs';
 import { clientI18nFiles } from './files.mjs';
 import { writeEntityFiles } from './entity-files.mjs';
-import { languageToJavaLanguage } from './utils.mjs';
-import { replaceContent } from '../utils.mjs';
 import TranslationData from './translation-data.mjs';
+import { findLanguageForTag, supportedLanguages } from './support/languages.mjs';
+import { updateLanguages as updateLanguagesInAngular } from '../angular/support/index.mjs';
```
Tcharl, line 32: Won't it be better to get the client (or any other related) technology to compose the language generator instead?

mshima: The technology generators are already composing with languages:
https://github.com/jhipster/generator-jhipster/blob/e7c95e52a6ba86b6a148a812d17fc6f6978bb38d/generators/angular/generator.mjs#L49-L55

This PR is moving the needles implementation there.

Tcharl: Thank you for your answer

I may have misspoken, I was trying to say that import from `angular` to `languages` looks normal (meaning you want to generate the i18n for your language) but the reverse looks less solid: we would be forced to introduce coupling from the language technology generator each time we want to introduce a new one (i.e. quarkus, svelte...) instead of 'just' implementing the language templates and abstract methods at the technology generator level

mshima: The languages generator have a particular behavior, it's possible to add a language to an already existing project.
It's possible for the blueprint to ignore this incremental behavior an iterate over the configured languages.
Unless we remove this incremental behavior or do some more advanced options passing (add something like --single-entity to generate a single language and ignore everything else) we should keep the languages generator like this.

This is the reason https://github.com/jhipster/generator-jhipster/issues/19114 will not be implemented.

DanielFran (approved):

PR #20916, opened 2023-01-25
Move fieldIsEnum and remove getEntityFolderName.

mshima:
- move `fieldIsEnum` to `base-application` generator. `entity` generator is a command to add new entities, should not contains utilities for others generators
- drop `getEntityFolderName`. The method is used only once and too simple. Implement inline.

cleanup https://github.com/jhipster/generator-jhipster/pull/20286.

---


Tcharl:
Not totally agree on this one.

The goal of keeping this stuff in the entity generator would be to be able to provide kind of 'jhipster lite' experience in the end: the end user could be able to generate app skeleton without any entity feature needed.

To me we have three different layers and guess that we should try to keep separate as much as we can:

 - The platform (kube, docker, ...)
 - The app generator (gateway, backend, microservice, frontends...)
 - The entities (including fields)

And the more we'll succeed to split these three concept, the more correct the overall architecture will be (just have to check if only 'entity' folders of generator-client, generator-server, ... have dependency to to this one.

mshima:
entities/support was merged without approval, past reviews were ignored.

Like I said before:
- we cannot split the basic entities support out of the `base-application` support without complicating the upper layers.
A full featured `generators/entity` would mean a full featured `generators/entity-angular`, `generators/entity-react`, `generators/entity-sql`, and so one like we had in v7 (without the modularizing).
- the basic entity support is dependent on the application for technology reasons like database, frameworks, authentication.
- entity support is optional.
- the entity model will be shared with jdl so will application model.

We could create another base generator like `generators/base-application-entity` but it's not worth IMO.

`Entity` support is optional so a jhipster-lite like implementation is completely independent.

DanielFran (approved):

mshima:
@Tcharl one example is enum support which only makes sense with entities that you put at `generators/base/support` https://github.com/jhipster/generator-jhipster/blob/main/generators/base/support/configuration-helpers/enum.mjs.
Base support is not related to generators.

Tcharl:
> entities/support was merged without approval, past reviews were ignored.

Sorry again, I definitely did my best in a time and material approach: I knew it was not perfect but assumed it. I've asked quite a few times for real show stoppers, which was ignored too ;-)

> 
> Like I said before:
> 
>     * we cannot split the basic entities support out of the `base-application` support without complicating the upper layers.
>       A full featured `generators/entity` would mean a full featured `generators/entity-angular`, `generators/entity-react`, `generators/entity-sql`, and so one like we had in v7 (without the modularizing).
I was thinking of an in-between: `generator-angular` and so on may be overkill, but `angular/support/entities` and `templates/entities` sounds appealing

> 
>     * the basic entity support is dependent on the application for technology reasons like database, frameworks, authentication.
Same way as the platform impacts the server (i.e. istio has/should have an impact on the service registry/config management). I may be wrong, but from an high level design: `platform` could compose `app`which could compose `entity`. May be time to graphiz the current dependency chain and exchange around it to be sure that the delegation chain is the correct one. I'm pretty sure it has a good logic, but convinced we can't improve it. I take the action to design the current state on a collaborative platform, and will create a dedicated ticket for brainstorm session if you agree
> 
>     * entity support is optional.
> 
>     * the entity model will be shared with jdl so will application model.
Today  I would say... When I was modeling with EMF, I was used to compose metamodels too
> 
> 
> We could create another base generator like `generators/base-application-entity` but it's not worth IMO.
100% agrees
> 
> `Entity` support is optional so a jhipster-lite like implementation is completely independent.

mshima:
> > entities/support was merged without approval, past reviews were ignored.
> 
> Sorry again, I definitely did my best in a time and material approach: I knew it was not perfect but assumed it. I've asked quite a few times for real show stoppers, which was ignored too ;-)

I've asked many times to split the PR.
And said that the generator-private was a todo list that needed reafactoring and api cleanup.
Now we need to go around to look for the apis to refactor and cleanup.

> > Like I said before:
> > ```
> > * we cannot split the basic entities support out of the `base-application` support without complicating the upper layers.
> >   A full featured `generators/entity` would mean a full featured `generators/entity-angular`, `generators/entity-react`, `generators/entity-sql`, and so one like we had in v7 (without the modularizing).
> > ```
> 
> I was thinking of an in-between: `generator-angular` and so on may be overkill, but `angular/support/entities` and `templates/entities` sounds appealing

The `support/entities` can be ok, but the entities support should have just a few apis. IMO is ok to just put in `support`.
IMO a single `templates` with a common hierarchy is cleaner. 

> > ```
> > * the basic entity support is dependent on the application for technology reasons like database, frameworks, authentication.
> > ```
> 
> Same way as the platform impacts the server (i.e. istio has/should have an impact on the service registry/config management). I may be wrong, but from an high level design: `platform` could compose `app`which could compose `entity`. May be time to graphiz the current dependency chain and exchange around it to be sure that the delegation chain is the correct one. I'm pretty sure it has a good logic, but convinced we can't improve it. I take the action to design the current state on a collaborative platform, and will create a dedicated ticket for brainstorm session if you agree

At current implementation any platform specific configuration that impacts the server is done at configuration level using environment variables.
Some flags may change at the application configuration (jhipster config).
The application/entity coupling is much deeper.

Tcharl:
> I've asked many times to split the PR.
And said that the generator-private was a todo list that needed reafactoring and api cleanup.
Now we need to go around to look for the apis to refactor and cleanup.

Tried this, definitely, but the legacy codebase pushed me to move more than what I expected. Also, I'm for sure not a black belt coder. The refactoring step was also planned to be achieved in a second step (due to the impact of the move) so everything is going as planned ;-) (still, sorry again for the curating part).

Much deeper for sure, still we should try to keep them as tight coupled as much as we can

Issue #20295, opened 2022-11-12
Extract the generator's logic into dedicated business-logic related structures


Tcharl:
##### **Overview of the feature request**

the goal would be to apply SOLID, OOP and all the guidelines explained in the 'Five Line of Code' book from Manning in this repository's codebase.

I propose to move all the business logic to each relevant bluprint's `logic` folder with an according split of responsibility splitting concerns in files (OOP-like).

##### **Motivation for or Use Case**

Current logic is split into `generator-*` file as well as some `utils`files. Method logic does many different things (mixing assignements, logic, assertions, ...) which complicates the introduction of new features as well as the tests.

##### **How to proceed**

In a first time, a dumb extraction of the logic in their respective files will be done (a bunch of PRs 'till it will be complete).
- It will transform the `generator` files into a dumb facade.
In a second time, we'll take a look at the usage and move the generator's method in their relevant facade.
In a third time, we'll decide if the API bring enough value to be kept and simplify the overall generator.

I currently have one full week of vacation to contribute to this feature

- [x] Checking this box is mandatory (this is just to show you read everything)

PR #20954, opened 2023-01-28
Move clientwebpack needle from base

Tcharl:
Move some methods from base to client generator
---
Contributes to #20295

### DIFF
--- a/generators/angular/generator.mjs
+++ b/generators/angular/generator.mjs
@@ -32,6 +32,7 @@ import {
   generateTestEntityPrimaryKey as getTestEntityPrimaryKey,
   generateTypescriptTestEntity as generateTestEntity,
 } from '../client/support/index.mjs';
+import AbstractJHipsterClientGenerator from '../client/abstract-client-generator.mjs';
 
 const { ANGULAR } = clientFrameworkTypes;
 
@@ -39,7 +40,7 @@ const { ANGULAR } = clientFrameworkTypes;
  * @class
  * @extends {BaseApplicationGenerator<import('../client/types.mjs').ClientApplication>}
  */
-export default class AngularGenerator extends BaseApplicationGenerator {
+export default class AngularGenerator extends AbstractJHipsterClientGenerator {
   async beforeQueue() {
     await this.dependsOnJHipster(GENERATOR_CLIENT);
     if (!this.fromBlueprint) {
--- a/generators/base/generator-base.mjs
+++ b/generators/base/generator-base.mjs
@@ -121,17 +121,6 @@ export default class JHipsterBaseGenerator extends PrivateBase {
 
   /**
    * @private
-   * Add external resources to root file(index.html).
-   *
-   * @param {string} resources - Resources added to root file.
-   * @param {string} comment - comment to add before resources content.
-   */
-  addExternalResourcesToRoot(resources, comment) {
-    this.needleApi.client.addExternalResourcesToRoot(resources, comment);
-  }
-
-  /**
-   * @private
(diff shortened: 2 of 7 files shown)


mshima:
If we add a needles to client and cypress, we will need to add same needle to angular, react and vue.
The idea is to use the `source` object which is shared between every generator, inject the needle and use it in any generator.
In this case the webpack needle should have a common api and angular, react and vue will inject its own needle.

You can see a needle injection here:
https://github.com/jhipster/generator-jhipster/blob/8d55f965b4a51cc8e89f18318aaa2d11b11f76ba/generators/docker/generator.mjs#L48-L59
And bellow it's using the needle:
https://github.com/jhipster/generator-jhipster/blob/8d55f965b4a51cc8e89f18318aaa2d11b11f76ba/generators/docker/generator.mjs#L110-L117

mshima:
I don't agree to add the same needle to 2 different generators even if temporary.

Tcharl:
Thanks for the explaination, it was indeed blurry to me to have something injected which wasn't explicitely called (is there a way to explicitely call it during one of the generator phase so that statement will be clear?

The only remaining blurry stuff is the fact that 'angular', 'vue' and 'react' generator needs it (of the reverse, why client generator needs it).

Either we should use 'client' subgen to be in charge of everything which is in common to all clients generators (cypress included), or either we let the leafs generators to be in charge of injecting needles and stuff. I would be mitigated on the fact that both handles should do.

I'm personally in favor of the second option, e.g. to get only angular, react, vue and cypress generator extending `AbstractJHipsterClientGenerator`.

Your thought?

mshima:
> Thanks for the explaination, it was indeed blurry to me to have something injected which wasn't explicitely called (is there a way to explicitely call it during one of the generator phase so that statement will be clear?

Needles should always be used at postWriting. The source parameter is only available at `preparing` for injection and `postWriting` for use.
https://github.com/jhipster/generator-jhipster/blob/ab599e070a346952290c87672b3428712a06e0fa/generators/base/generator.mts#L245-L248

> The only remaining blurry stuff is the fact that 'angular', 'vue' and 'react' generator needs it (of the reverse, why client generator needs it).

It's not about why client generator needs it, but which customization a blueprint can apply.
At jhipster v7 the needles api is available to every generators and a single generator can customize everything like it is stated in the docs:
https://www.jhipster.tech/modules/creating-a-blueprint/#local-blueprints.
`so a single sub-generator, the main one, is probably enough to customize your application (client and server).`

> Either we should use 'client' subgen to be in charge of everything which is in common to all clients generators (cypress included), or either we let the leafs generators to be in charge of injecting needles and stuff. I would be mitigated on the fact that both handles should do.

There is almost no common to all clients generators. We can abstract the api, but it will be always have diferences.
The technology generator should inject its own needle.

> I'm personally in favor of the second option, e.g. to get only angular, react, vue and cypress generator extending `AbstractJHipsterClientGenerator`.
> 
> Your thought?

At your current approach using `abstract-client-generator.mjs`,  a generator will require:
- to extend the client generator to have the needles api.
- if we migrate vue and/or react to vite, the `addWebpackConfig` will still be there.

This `source` approach is the simplest approach I can think of.

DanielFran:
> This `source` approach is the simplest approach I can think of.

@Tcharl Should we then keep the current approach and close both PRs?

Tcharl:
@DanielFran Working on it (I unfortunately have a jhipster project to migrate in parrallel, which is why I do not achieve that much progress). I'll close it when I'll totally understand what I should do ;-)

PR #20957, opened 2023-01-29
move entity to base-application

mshima:
Move entity operations from generators/base to generators/base-application.
Generate a temporary `generator-ts.mts` to keep ts implementation while `generator` is js.

---

### DIFF
--- a/generators/base-application/generator-ts.mts
+++ b/generators/base-application/generator-ts.mts
@@ -1,0 +1,51 @@
+import _ from 'lodash';
+import type Storage from 'yeoman-generator/lib/util/storage.js';
+
+import BaseGenerator from '../base/index.mjs';
+import { JHIPSTER_CONFIG_DIR } from '../generator-constants.mjs';
+import { getEntitiesFromDir } from './support/index.mjs';
+
+const { upperFirst } = _;
+
+// Temporary Generator with Typescript implementations
+export default class BaseApplicationTsGenerator extends BaseGenerator {
+  /**
+   * Get all the generator configuration from the .yo-rc.json file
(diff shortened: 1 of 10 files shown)


generators/base-application/generator-ts.mts
```diff
@@ -0,0 +1,51 @@
...
+  getEntityConfig(entityName: string, create = false): Storage | undefined {
+    const entityPath = this.destinationPath(JHIPSTER_CONFIG_DIR, `${upperFirst(entityName)}.json`);
+    if (!create && !this.fs.exists(entityPath)) return undefined;
+    return this.createStorage(entityPath, { sorted: true } as any);
+  }
+
+  /**
+   * get sorted list of entitiy names according to changelog date (i.e. the order in which they were added)
+   */
+  getExistingEntityNames(): string[] {
```
Tcharl, line 26: Should be in a support file isn't it?

mshima: IMO no.
Every core api should be implemented inline in the generator (except some utilities).
And we should avoid utilities to use any complex parameter like the generator and the log.

DanielFran (approved):

generators/base-application/generator-ts.mts
Tcharl: Why am I stating that: these APIs will certainly be used by many generators, so we'll avoid passing complex parameters (`generator`) reusing them by moving this method to a dedicated `support` package.

I agree `generator`should not be passed as an attribute. LogAdapter looks to be a must (quite systematic) as there's no slf4j-like static injection fwk

PR #21467, opened 2023-03-17
Run tests in random rather than alphabetical order

lmartelli:
Running tests in alphabetical order is a bad practice : it can hide unwanted dependencies between tests. Therefore I believe it is preferable to run them in random order so as that to spot problems as soon as possible. If some tests are flaky because of unwanted dependencies are incomplete cleanup, it's best to fix them rather than hide the problem and look away 🙈 
And if you really need a specific order for some tests, you can still use `@TestMethodOrder`.

---

### DIFF
--- a/generators/server/templates/pom.xml.ejs
+++ b/generators/server/templates/pom.xml.ejs
@@ -1391,8 +1391,9 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
                     <configuration>
-                        <!-- Force alphabetical order to have a reproducible build -->
-                        <runOrder>alphabetical</runOrder>
+                        <!-- Force random order. A fixed order hides potential problems with your tests -->
+			<!-- If you really need a specific order for some tests, use @TestMethodOrder -->
+                        <runOrder>random</runOrder>
                         <excludes>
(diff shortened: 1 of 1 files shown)


mshima:
I agree that would be best with random order.
But I think we should keep ordered at our CI due to complexity of JHipster.
A random failure is not reproducible. And:
- most contributors cannot merge PRs with failures.
- can break workflows like dependabot.

Random failures is really bad for JHipster CI.

mshima:
If a seed based on baseName is added https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#runOrderRandomSeed. I think it's acceptable.
Like this: 
https://github.com/jhipster/generator-jhipster/blob/6068ef26d40f95a52b3d4253533390e03135d554/generators/docker/generator.mjs#L82-L84

heilaschaefer:
Random order is interesting.  However if only one set of orders causes a failure, then it would be necessary to be able to reproduce that set of orders.  I suggest that a Seed for the random is clearly published so that it makes the order reproducible if need be.

mraible:
Closing and reopening to restart CI.

mshima:
I don’t think we should merge without adding a reproducible seed.

PR #21476, opened 2023-03-20
Fix broken Hibernate mapping of Instants

gzsombor:
So it will be stored as UTC values inside the database, instead of Local timezone. 
This was a breaking change introduced Hibernate 6, this fix restores the previous behavior, and re-enables the test which could warn about in the future, if similar thing happens.

---

### DIFF
--- a/generators/server/templates/src/main/resources/config/application.yml.ejs
+++ b/generators/server/templates/src/main/resources/config/application.yml.ejs
@@ -251,6 +251,7 @@ spring:
     open-in-view: false
     properties:
       hibernate.jdbc.time_zone: UTC
+      hibernate.type.preferred_instant_jdbc_type: TIMESTAMP
       hibernate.id.new_generator_mappings: true
       hibernate.connection.provider_disables_autocommit: true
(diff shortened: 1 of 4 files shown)


DanielFran (changes requested):

gzsombor:
Unfortunately, one test fails because of 
```
Caused by: org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: wrong column type encountered in column [renamed] in table [department]; found [text (Types#VARCHAR)], but expecting [oid (Types#CLOB)]
```

DanielFran:
> Unfortunately, one test fails because of
> 
> ```
> Caused by: org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: wrong column type encountered in column [renamed] in table [department]; found [text (Types#VARCHAR)], but expecting [oid (Types#CLOB)]
> ```

Is this same as this issue? https://stackoverflow.com/questions/73474448/schema-validation-wrong-column-type-encountered-in-column-name-in-table-test

gzsombor:
Unfortunately, it's not just that. After I've fixed to not use @Lob when the field is a string, we still have a bunch of other wrong column types

DanielFran:
Thanks @gzsombor

PR #21712, opened 2023-04-05
Modularize more pom dependencies

mshima:
---

### DIFF
--- a/generators/mongodb/generator.mjs
+++ b/generators/mongodb/generator.mjs
@@ -59,27 +59,36 @@ export default class MongoDBGenerator extends BaseApplicationGenerator {
       addDependencies({ application, source }) {
         const { reactive } = application;
         if (application.buildToolMaven) {
-          source.addMavenDependency?.([
-            {
-              groupId: 'org.springframework.boot',
-              artifactId: `spring-boot-starter-data-mongodb${reactive ? '-reactive' : ''}`,
-            },
-            {
-              groupId: 'org.testcontainers',
-              artifactId: 'junit-jupiter',
-              scope: 'test',
-            },
-            {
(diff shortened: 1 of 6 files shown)


generators/mongodb/generator.mjs
```diff
@@ -59,27 +59,36 @@ export default class MongoDBGenerator extends BaseApplicationGenerator {
...
+          source.addMavenDefinition?.({
+            dependencies: [
+              { groupId: 'io.mongock', artifactId: 'mongock-springboot-v3' },
+              { groupId: 'org.springframework.boot', artifactId: `spring-boot-starter-data-mongodb${reactive ? '-reactive' : ''}` },
+              { groupId: 'org.testcontainers', artifactId: 'junit-jupiter', scope: 'test' },
+              { groupId: 'org.testcontainers', artifactId: 'testcontainers', scope: 'test' },
+              { groupId: 'org.testcontainers', artifactId: 'mongodb', scope: 'test' },
+            ],
+            dependencyManagement: [
+              // Fix Mongock dependencies: https://github.com/mongock/mongock-jdk17/issues/6
```
DanielFran, line 72: Not sure we need that since we removed the reflections 0.10.2 in the bom project

DanielFran (changes requested):

DanielFran (approved):

generators/mongodb/generator.mjs
mshima: I think it won't break too.
But since the latest version (0.10.2) is known to be broken, IMO the safest bet is to force the version to make sure others dependencies won't bring an updated version.

mshima: We can remove if you feel confident.

PR #21747, opened 2023-04-08
update to gradle 8.0.2

atomfrede:
Updating gradle to 8.0.2. All blocking deprecation warnings are fixed. Also remove the buildscript dependency which I couln't find in the needles api and we should not support anymore.

Also fine tuned the input/outputs such that tasks are up-to-date and could be cached (with the help of our gradle enterprise instance)

---


generators/liquibase/templates/gradle/liquibase.gradle.ejs
```diff
@@ -1,7 +1,10 @@
+configurations {
+    liquibaseRuntime.extendsFrom sourceSets.main.compileClasspath
```
DanielFran, line 2: Only if database SQL?

DanielFran (changes requested):

generators/liquibase/templates/gradle/liquibase.gradle.ejs
atomfrede: Should not be needed as the file is only created for sql databases

DanielFran (approved):

Issue #21959, opened 2023-04-28
8.0.0-beta.1 release


mshima:
##### **Overview of the feature request**

Goals for Beta release:
- a typescript based yeoman-generator
https://github.com/jhipster/generator-jhipster/blob/main/generators/base/generator-base-definitions.mjs is exclusively to fix some typings errors at yeoman-generator types.
Updates mem-fs to v3 and typescript based mem-fs-editor too.
- rename generators to more meaningful names
the sql generator was called sql to see if it would be easier to blueprints to reuse, but there is almost nothing generic that is not related to spring-data-relational.
Let's call the generator spring-data-relational. Related to https://github.com/jhipster/generator-jhipster/issues/19810., https://github.com/jhipster/generator-jhipster/issues/21653.
- JDL cleanup:
jdl should not write defaults to application configuration, defaults are specific to the generator(blueprint). Defaults are calculated at use now.
Related to https://github.com/jhipster/generator-jhipster/issues/21555, https://github.com/jhipster/generator-jhipster/issues/21768

##### **Related issues or PR**

- [ ] Checking this box is mandatory (this is just to show you read everything)

mraible:
@mshima I'm doing a presentation on JHipster at Devoxx UK next week (May 11). Do you think it's possible to release before then? Anything I can do to help?

mshima:
@mraible if you want to rename the database generators to spring-data-*.
I will try to focus on the yeoman-generator side that should be the hardest task here.

mraible:
@mshima In #19810, it says the database generator for SQL should be `spring-boot-sql`. Is `spring-data-sql` preferred, or should it be `spring-data-jpa` and `spring-data-r2dbc`?

mshima:
spring-data-jpa and spring-data-r2dbc are joined at spring-data-relational.

mraible:
@mshima Since this issue is closed, does that mean we're ready for a beta release?

mshima:
@mraible yes, imo we can go ahead.

mraible:
@DanielFran Can you help facilitate a release this week?

benjaminbarbe10:
@mraible @DanielFran 
Hello, First thanks all for the time you spend on this wonderful project :)

Any news on the release date and is there something we can do to help ?

mshima:
@DanielFran bounty claimed https://opencollective.com/generator-jhipster/expenses/148965.

DanielFran:
@mshima approved

PR #22010, opened 2023-05-04
[jdl] don't validate and don't include defaults

mshima:
- Remove configuration defaults and validation from jdl.
Defaults and validation should be done at generators.
- Languages correctly defaults to [en, fr]
- Monorepositories correctly disables commit hooks.

Related to https://github.com/jhipster/generator-jhipster/issues/21959.
Includes https://github.com/jhipster/generator-jhipster/pull/21191.

---


generators/cassandra/generator.mjs
```diff
@@ -32,22 +32,36 @@ export default class CassandraGenerator extends BaseApplicationGenerator {
     }
   }
 
+  get configuringEachEntity() {
+    return this.asConfiguringEachEntityTaskGroup({
+      checkEntities({ entityConfig }) {
+        if (entityConfig.pagination && entityConfig.pagination !== 'no') {
```
Tcharl, line 38: Can't derived variable be loaded before configuring? Would avoid the '!== 'no''

Tcharl:
Nice spring cleanup!
LGTM

generators/cassandra/generator.mjs
mshima: No, but doing validation at a later step is something to consider.
Priorities exclusively for validations is something to consider.
We should open an issue for this.

DanielFran (approved):

PR #22105, opened 2023-05-12
Relationship with Authority - tests...

OmarHawk:
whether there's a problem to have a ManyToMany relationship with Authority (similar as User) using the new `with builtInEntity`

#22106
---

### DIFF
--- a/jdl/jdl-importer.spec.ts
+++ b/jdl/jdl-importer.spec.ts
@@ -59,10 +59,15 @@ entity A
 relationship OneToMany {
   A{user} to User with builtInEntity
 }
+entity AuthorityExtension
+relationship ManyToMany {
+  AuthorityExtension{authority} to Authority with builtInEntity
+}
 `,
           { applicationName: 'MyApp', databaseType: databaseTypes.SQL },
         ).import();
         jestExpect(importState.exportedEntities[0].relationships[0].relationshipWithBuiltInEntity).toBe(true);
+        jestExpect(importState.exportedEntities[1].relationships[0].relationshipWithBuiltInEntity).toBe(true);
       });
     });
     describe('when not parsing applications', () => {
--- a/jdl/__test-files__/big_sample.jdl
+++ b/jdl/__test-files__/big_sample.jdl
@@ -67,6 +67,11 @@ entity Region {
   name String
 }
 
+entity AuthorityExtension {
+  name String
+  description String
+}
+
 relationship OneToOne {
   @id Department{location} to @id(42) Location
 }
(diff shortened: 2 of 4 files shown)


jdl/jdl-importer.spec.ts
```diff
@@ -60,10 +60,15 @@ entity A
 relationship OneToMany {
   A{user} to User with builtInEntity
 }
+entity AuthorityExtension
+relationship ManyToMany {
+  X{authority} to Authority with builtInEntity
```
mshima, line 65:
```suggestion
  AuthorityExtension{authority} to Authority with builtInEntity
```

mshima:
Failure is related to snapshots only.

OmarHawk:
> Failure is related to snapshots only.

Mh... true, one of the Snapshots was incorrectly set up. Still it won't work unfortunately. I'd try to look into it next week unless you are faster with that and want to do that by yourself ;)

DanielFran:
@OmarHawk Can you please confirm if everything has been fixed?

OmarHawk:
I'll rebase on Thursday and let the tests (which is the only thing in this pull request) run again.

mshima:
It's needed to add this for Authority:
https://github.com/jhipster/generator-jhipster/blob/cc7950546c54903671e340fafc05a4a3da9c7b30/generators/bootstrap-application-base/utils.mjs#L31

OmarHawk:
> It's needed to add this for Authority:
> 
> https://github.com/jhipster/generator-jhipster/blob/cc7950546c54903671e340fafc05a4a3da9c7b30/generators/bootstrap-application-base/utils.mjs#L31

This is then really only about having it declared and then putting it like that?

https://github.com/jhipster/generator-jhipster/blob/cc7950546c54903671e340fafc05a4a3da9c7b30/generators/bootstrap-application-base/generator.mts#L200-L212

The rest of the User entity stuff is also happening somewhere else... I wonder, why the User entity was not "simply" ;-) declared as "actual" jdl String and merged with the user defined jdl (if existing), so that it would be treated just like any other entity and could then be used as such without special logic or requirement to use "with builtInEntity" - and without having special / kind of duplicated logic/templates in a few places...

mshima:
> This is then really only about having it declared and then putting it like that?

Yes, should be.
May need specific customizations to add management.

> The rest of the User entity stuff is also happening somewhere else... I wonder, why the User entity was not "simply" ;-) declared as "actual" jdl String and merged with the user defined jdl (if existing), so that it would be treated just like any other entity and could then be used as such without special logic or requirement to use "with builtInEntity" - and without having special / kind of duplicated logic/templates in a few places...

JDL does very few checks now. Consistency checks is one of them. You cannot have a relationship without both sides Entities.
`with builtInEntity` ignores that check.
v8 focused a lot in modularization and blueprints.
At v7 it was determined by application configs.
This causes problems like for e.g with oauth2 auth, builtIn User exists in JHipster. None of backend blueprints implements it.

Declaring User entity instead of `with builtInEntity` should work and is merged to built-in definition.
```
entity User {}
entity Foo {}
relationship ManyToOne {
  Foo to User
}
```

User template is separate. So any additional field or ownerSide relationship are silently ignored. Generating User using common templates and making it customizable is doable.

About all existing logic, we have a lot of legacy code that are slowly been cleaned up.

PR #22307, opened 2023-05-26
integrate jhipster log into yeoman log

mshima:
Having 2 logs is confusing, integrate jhipster custom logic into yeoman's.

---


Tcharl:
The issue using yeoman logger was that any `support` class, that is not part of the generator but related to jhipster-specific task was force to take the yeoman context or logger as an attribute, for each each method... Which is not so readable or optimal.

I miss the `private static Logger log = LoggerFactory.getLogger("...")` which injects logs into a class without any dependency to an execution context

mshima:
> The issue using yeoman logger was that any `support` class, that is not part of the generator but related to jhipster-specific task was force to take the yeoman context or logger as an attribute, for each each method... Which is not so readable or optimal.
> 
> I miss the `private static Logger log = LoggerFactory.getLogger("...")` which injects logs into a class without any dependency to an execution context

yeoman's log are mandatory and they are silenced automatically at tests.
There are others log libs, but they need to be integrated in the adapters io, queue and silenced at tests.
Current in generator jhipster specific log just has a few additional loggers and forwards to the yeoman log.

DanielFran (approved):

Issue #22579, opened 2023-06-17
JHipster timezone configuration fails with hibernate 6.2.x.


mshima:
##### **Overview of the issue**

Upgrading hibernate to v6.2.x breaks https://github.com/jhipster/generator-jhipster/blob/main/generators/spring-data-relational/templates/src/test/java/package/config/timezone/HibernateTimeZoneIT.java.ejs.

##### **Reproduce the error**

Reproducing may depend on system timezone.
```
jhipster --defaults --skip-install
<update to hibernate 6.2.x at pom.xml or build.gradle>
<set config `hibernate.timezone.default_storage=NORMALIZE_UTC`>
npm run backend:unit:test
```

##### **Related issues**

Looks related to https://hibernate.atlassian.net/browse/HHH-16778

##### **Suggest a Fix**

Keep hibernate at v6.1.7 as workaround.

##### **Browsers and Operating System**

- [ ] Checking this box is mandatory (this is just to show you read everything)

mraible:
Increasing bug bounty as it would be nice to support Hibernate 6.2 in 8.0 GA.

anarsultanov:
Please take a look at the proposed fix https://github.com/jhipster/generator-jhipster/pull/22946

PR #22646, opened 2023-06-22
Update spring-boot version to 3.1.1

DanielFran:
---

### DIFF
--- a/generators/generator-constants.mts
+++ b/generators/generator-constants.mts
@@ -25,7 +25,7 @@ export const JAVA_COMPATIBLE_VERSIONS = ['17', '18', '19'];
 export const ADD_SPRING_MILESTONE_REPOSITORY = false;
 
 // Version of Node, NPM
-export const NODE_VERSION = '18.16.0';
+export const NODE_VERSION = '18.16.1';
 export const OPENAPI_GENERATOR_CLI_VERSION = '2.6.0';
 
 export const javaDependencies: Record<string, string> = {
--- a/generators/generator-constants.mts
+++ b/generators/generator-constants.mts
@@ -26,7 +26,7 @@ export const ADD_SPRING_MILESTONE_REPOSITORY = false;
 
 // Version of Node, NPM
 export const NODE_VERSION = '18.16.0';
-export const OPENAPI_GENERATOR_CLI_VERSION = '2.5.1';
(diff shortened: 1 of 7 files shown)


generators/generator-constants.mts
```diff
@@ -25,20 +25,20 @@ export const JAVA_COMPATIBLE_VERSIONS = ['17', '18', '19'];
...
    */
-  'spring-boot': '3.1.0',
+  'spring-boot': '3.1.1',
   /*
    * hibernate version should match the one managed by https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-dependencies/SPRING_BOOT_VERSION
    * Required due to hibernate-jpamodelgen annotation processor.
    */
   // TODO update to spring-boot provided version. Refer to https://github.com/jhipster/generator-jhipster/issues/22579
-  hibernate: '6.1.7.Final',
+  hibernate: '6.1.7.Final', // Should be 6.2.5.Final
```
mraible, line 41: Is there a bug in Hibernate 6.2 to track this or is it something we need to change?

DanielFran: Related with https://github.com/jhipster/generator-jhipster/issues/22579

generators/init/constants.mjs
```diff
@@ -16,7 +16,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-export const NODE_VERSION = '14.18.1';
+export const NODE_VERSION = '18.16.1';
```
DanielFran, line 19: @mshima We should use only 1 constant for NODE_VERSION, no?
Maybe use only the one from init generator?

mshima: NODE_VERSION at init is used as required node version:
https://github.com/jhipster/generator-jhipster/blob/61ec0be217e40fa96c944bdc624921efb5104bb6/generators/init/templates/package.json.ejs#L35

I think it should match https://github.com/jhipster/generator-jhipster/blob/61ec0be217e40fa96c944bdc624921efb5104bb6/package.json#L179.
We probably should copy the package.json version.

generators/generator-constants.mts
BrayanMnz: May I ask if updating to 6.2.5 still failing?

PR #22744, opened 2023-07-02
feat: add spotless formatter

### LINKED ISSUE #22734: Feature: Make use of spotless maven plugin

For Java, I have made good experience with spotless. I saw in issue #6379 there was already a discussion about it.

@PierreBesson: I read your comment from December 12, 2017 in that discussion. The build does not necessarily have to fail when the formatting is not correct because spotless can apply the correct formatting, too.

Ensure clean source code files by using a formatted for all source files.
Introducing automated code formatting as early as possible in a project has some great benefits: 
* Better readability of source code because only necessary code parts are in the file (no unused imports, spotless can remove them). Also, the code will look more equal across different parts that might be written by different developers
* Less whitespace / reformatting noise when someone touches existing code. When automated formatting is used, the formatting should always look the same no matter who edited the code. This keeps the git diff clean because no unnecessary whitespace changes etc. should be present there when formatting is done automatically. 

#6379

CodingSpiderFox:
automatically remove unused imports

Spotless is added to pom.xml with minimalistic default configuration, bound to process-sources phase
After initial generation, process-sources phase is called so that the initially generated files are cleaned up right from the start.

Fixes #22734

---


atomfrede:
Can you add it to the gradle part too? @jhipster/developers Is this something prettier can do too? I only found organizing import, but nothing about removing unused imports.

CodingSpiderFox:
I was also looking into doing that with prettier but I did not find any working solution with prettier

CodingSpiderFox:
I am getting a lot of HTTP 403 when I create an application with gradle:
`Could not GET 'https://jcenter.bintray.com/org/sonarsource/scanner/api/sonar-scanner-api/2.16.2.588/sonar-scanner-api-2.16.2.588.pom'. Received status code 403 from server: Forbidden`

Is there something wrong with my project or is that a problem at bintray currently?

mraible:
bintray doesn't exist anymore, so it should be removed if it's referenced anywhere. 

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

CodingSpiderFox:
I wonder why my pc tries to only access it when I'm connected via my public VPN

CodingSpiderFox:
@atomfrede I added gradle support

generators/server/templates/build.gradle.ejs
```diff
@@ -32,6 +32,7 @@ plugins {
     id "com.github.node-gradle.node"
 <%_ } _%>
     id "org.sonarqube"
+    id "com.diffplug.spotless" version "<%- javaDependencies['spotless-gradle-plugin'] %>"
```
atomfrede, line 35: Just to be consistent, can you move the version number to `pluginManagement` block in `settings.gradle`, such that we can use it here without the version?

atomfrede (changes requested):
Small remark where to put the gradle plugin version

generators/server/templates/build.gradle.ejs
CodingSpiderFox: yes, sure :)

CodingSpiderFox: changed it

atomfrede (approved):
Looks fine to me.

mshima:
Sorry for the late review.
Current integration with `npmw` will cause differences in generation process.

`npmw` is executed if `--skip-install` is not passed and package.json has changes.
- `jhipster --skip-install` and `jhipster` may have differences in source source.
- `jhipster --defaults` (adds a package.json) following another `jhipster --defaults` (no changes to package.json) will cause a not clean tree since spotless will be executed at first command only.

Ideally source transformation should be executed at transform priority. (should be possible if spotless can be executed at a input an generate an output).
So spotless should be executed at server generator at install priority.
Something like:
```
  get [BaseGenerator.INSTALL] () {
    async runSpotless({ application }) {
      try {
        const spotlessCommand = application.buildToolGradle ? `./gradlew ...` : `./mvnw ...`;
        await this.spawnCommand(spotlessCommand);
      } catch(e) {
        this.log.warn(`Fail to execute spotless ${e.stdout}`);
      }
    },
  }
```

CodingSpiderFox:
Ok I will see if I can successfully integrate that

CodingSpiderFox:
haven't forgotten about my still open topics here ;) was just very busy but I guess I'll be able to find some time sunday or next week

mshima:
@CodingSpiderFox spotless is disabled by default.
We have switched to an implementation in js to remove unused imports.

PR #22780, opened 2023-07-05
Enable CORS when app is a gateway or monolith

mraible:
Fixes https://github.com/jhipster/generator-jhipster-ionic/issues/872. 

---


generators/server/templates/src/main/java/package/config/SecurityConfiguration_imperative.java.ejs
```diff
@@ -31,7 +31,6 @@ import org.springframework.http.HttpMethod;
 <%_ } _%>
 import org.springframework.security.config.annotation.method.configuration.EnableMethodSecurity;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
-import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
```
mraible, line 34: I talked with Rob Winch about this annotation, and he mentioned it's only needed when running in a non-Spring Boot app.

mraible:
@DanielFran It seems like most CI failures are from front-end tests, not Java tests. This PR seems like it would only affect Java tests since CORS isn't used except with Ionic and React Native.

DanielFran (approved):

PR #22816, opened 2023-07-10
Create publish workflow for publishing to npm registry

### LINKED ISSUE #22718: Scripts and binaries are not executable
On the latest v8 beta release, executables like `./gradlew`, `kubectl-apply-sh` etc are not executable and I had to do `chmod +x` on all of them. Can someone confirm if its just me or the file templates are indeed not executable

mraible:
Should fix #22718. 

---

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,19 @@
+name: Publish
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
+      - run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}


mraible:
@DanielFran This will need a `NPM_TOKEN` repository secret for it to work.

mshima:
Maybe we should create another repository for releases due to the token.
And add releases samples update.

mraible:
> Maybe we should create another repository for releases due to the token.

Why is this necessary? It seems like it should be easy enough to define the `NPM_TOKEN` in this repository. We use something similar in the [generator-jhipster-ionic](https://github.com/jhipster/generator-jhipster-ionic) project, and it works fine.

mshima:
We should be careful running PRs due to token leakage.
If we create a 'secure' workflow, malicious PRs will be much easier to identify.

pascalgrimaud:
Same opinion than @mshima 
Not sure it's a good idea for security reason

Another problem is maybe the token is not enough because you need OTP to publish

The original problem is executable, maybe it's related to Windows machine?

mraible:
FWIW, I used a similar workflow to publish the React Native and Micronaut blueprints recently. 

It'd be nice if we could fix the executable problem in an automated way. It'd also be good to add `--tag beta` when publishing if we do another beta. 

A workaround to fix things so `npm i -g generator-jhipster` doesn't install beta 2 is to release a new 7.x version, according to GitHub support:

> We have taken a closer look at `generator-jhipster` package and can confirm the latest version is `8.0.0-beta.2`.
>
> As `7.9.3` is not the latest version, it is not possible to call to that version when installing the package.
>
> If you want new users not to install the beta version, we recommend moving forward and publish a new version `7.9.4` to remediate this issue.

PR #22888, opened 2023-07-17
allow view/edit in new tabs from list view

### LINKED ISSUE #22801: [UX/angular] Allow view/edit in new tabs from list view

With the current implementation (angular, jhipster 7.9.3/8.0.0-beta.1) it is possible to open an item from the list view in a new tab (use linked ID in ID column). In the new tab switching to edit mode an editing is possible.
But the view and edit function on the right side is implemented with buttons. Therefore is not possible to open or edit in new tabs items in this way. If you simply replace `<button submit="true" ... </button>` with `<a ... </a>`, working in tabs would be possible, even without layout changes.

Is there any reason why viewing and editing in new tabs is allowed on the left side and disallowed on the right side?

A workflow where items could be opened and edited in new tabs would improve productivity, especially when items where filtered or sorted in list view. It would be possible to open all items to process in new tabs without loosing the filtered/sorted list/search result and to edit the items tab for tab.

##### **Reproduce the error**

Open list view of any entity. It is not possible to view/edit an item in a new tab with the buttons in the right side.

##### **Suggest a Fix**

Replacing the buttons with a link solves the problem.

##### **JHipster Version(s)**

7.9.3, 8.0.0-beta.1

No specific configuration needed. Just use angular and define at least one entity.

any

nitram84:
Fix #22801

It is interesting to note, that tabs were already supported by react, so there was nothing to do for react. With this PR the view and edit "buttons" for angular and vue (here only authentication type "session") work in the same way like react. Vue produces an error for links opened in new tabs when authentication type is not "session". Working with tabs works best in session mode, e.g. with JWT authentication a login for each tab is required. But even when a login is required, this PR makes it still easier to copy or share view/edit links.

---

### DIFF
--- a/generators/angular/templates/src/main/webapp/app/entities/_entityFolder_/list/_entityFile_.component.html.ejs
+++ b/generators/angular/templates/src/main/webapp/app/entities/_entityFolder_/list/_entityFile_.component.html.ejs
@@ -180,16 +180,14 @@ _%>
     <%_ } _%>
 <%_ } _%>
                             <a [routerLink]="['/<%= entityPage %>', <%= entityInstance %>.<%= primaryKey.name %>, 'view']"
-                                    class="btn btn-info btn-sm"
-                                    data-cy="entityDetailsButton">
+                               class="btn btn-info btn-sm" data-cy="entityDetailsButton">
                                 <fa-icon icon="eye"></fa-icon>
                                 <span class="d-none d-md-inline" <%= jhiPrefix %>Translate="entity.action.view">__jhiTransformTranslate__('entity.action.view')</span>
                             </a>
 <%_ if (!readOnly) { _%>
 
                             <a [routerLink]="['/<%= entityPage %>', <%= entityInstance %>.<%= primaryKey.name %>, 'edit']"
-                                    class="btn btn-primary btn-sm"
-                                    data-cy="entityEditButton">
+                               class="btn btn-primary btn-sm" data-cy="entityEditButton">
                                 <fa-icon icon="pencil-alt"></fa-icon>
                                 <span class="d-none d-md-inline" <%= jhiPrefix %>Translate="entity.action.edit">__jhiTransformTranslate__('entity.action.edit')</span>
                             </a>
--- a/generators/angular/templates/src/main/webapp/app/entities/_entityFolder_/list/_entityFile_.component.html.ejs
+++ b/generators/angular/templates/src/main/webapp/app/entities/_entityFolder_/list/_entityFile_.component.html.ejs
@@ -179,8 +179,7 @@ _%>
                             </button>
     <%_ } _%>
 <%_ } _%>
-                            <a
-                                    [routerLink]="['/<%= entityPage %>', <%= entityInstance %>.<%= primaryKey.name %>, 'view']"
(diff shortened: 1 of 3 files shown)


deepu105:
Can you check the failing tests plz. Probably snapshots need to be updated using the `npm run update-snapshots` command

nitram84:
@deepu105: The failing test cases weren’t mine. At the time I created my PR all builds had failing test cases (and those were accepted). The build of the previous PR of my PR (https://github.com/jhipster/generator-jhipster/pull/22886/checks) had the same failing test cases as my PR. So these test failures weren't caused by my PR. It seems that the failing tests are now fixed on main, so my PR shouldn't pose a problem anymore.

generators/angular/templates/src/main/webapp/app/entities/_entityFolder/list/_entityFile.component.html.ejs
```diff
@@ -181,22 +181,22 @@ _%>
...
                                     class="btn btn-info btn-sm"
                                     data-cy="entityDetailsButton">
                                 <fa-icon icon="eye"></fa-icon>
                                 <span class="d-none d-md-inline" <%= jhiPrefix %>Translate="entity.action.view"><%- getWebappTranslation('entity.action.view') %></span>
-                            </button>
+                            </a>
 <%_ if (!readOnly) { _%>
 
-                            <button type="submit"
+                            <a
```
mraible, line 193: The formatting looks a little off after this change. Can you try moving the `[routerLink]` line up with the `<a`? Running prettier might be a good idea too.

mraible (changes requested):

DanielFran (approved):

DanielFran:
@nitram84 it is missing only react. Can you take a look?

nitram84:
@DanielFran I didn't miss react. In react new tabs are supported by default for view and edit:

"It is interesting to note, that tabs were already supported by react, so there was nothing to do for react. With this PR the view and edit "buttons" for angular and vue (here only authentication type "session") work in the same way like react. [...]"

This PR reduces the differences between the supported UI libs.

For react there might be one point: React supports delete operations in new tabs, so users can copy or share a delete-link. For view and edit links this is ok - those links can be used as permalinks. This is not true for delete links, so perhaps react should use a button for delete instead of a link, if this behavior is unwanted.

mraible (dismissed):

DanielFran:
> 

@nitram84 I agree, for delete it should be a button.
Do you want to include those changes in this or?

nitram84:
I fixed the react part too. Now all changes should be consistent.

DanielFran (approved):

DanielFran:
@nitram84 Can you fix the conflicts please?

nitram84:
@DanielFran Thank you for reminding me. I had to rebase my changes. Rebasing worked, all changes are identical to previously reviewed commits.

DanielFran (approved):

DanielFran:
@nitram84 Can you take a look at the e2e tests, some are failing

DanielFran (approved):

DanielFran:
Thanks @nitram84 for the contribution

mshima:
This PR created a regression at react. See https://github.com/jhipster/generator-jhipster/issues/24115.

